PR #298, opened 2016-02-11
Bugfix:OS6135311: Incorrect base register assignment in inlined LdNewTarget

suwc:
Bugfix:OS6135311: Incorrect base register assignment in inlined LdNewTarget

LdNewTarget inliner uses a base operand that is not properly assigned for indirect load of 'new.target' from stack.
This change re-factors generation of said indirect load for use in both stack argument load (arguments[i]) and in LdNewTarget.
Add unit test.

### DIFF
--- a/Lib/Backend/Lower.cpp
+++ b/Lib/Backend/Lower.cpp
@@ -17913,6 +17913,38 @@ Lowerer::GenerateArgOutForStackArgs(IR::Instr* callInstr, IR::Instr* stackArgsIn
     return saveLenInstr->GetDst()->AsRegOpnd();
 }
 
+void
+Lowerer::GenerateLoadStackArgumentByIndex(IR::Opnd *dst, IR::RegOpnd *indexOpnd, IR::Instr *instr, int32 offset, Func *func)
+{
+    // Load argument set dst = [ebp + index].
+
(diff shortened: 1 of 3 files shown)


suwc:
@boingoing @rajatd blessed.
@abchatra @LouisLaf please take a look. Thanks!

abchatra:
:+1:

lib/Backend/Lower.cpp
```diff
@@ -17930,6 +17930,38 @@ Lowerer::GenerateArgOutForStackArgs(IR::Instr* callInstr, IR::Instr* stackArgsIn
...
+    //       this or new.target           <== EBP + formalParamOffset (4)
+    //       callinfo
+    //       function object
+    //       return addr
+    // EBP-> EBP chain
+
+    //actual arguments offset is LowererMD::GetFormalParamOffset() + 1 (this)
+
+    int32 actualOffset = GetFormalParamOffset() + offset;
+    Assert(GetFormalParamOffset() == 4);
```
LouisLaf: This returns 2 on ARM...

rajatd: This is the one defined in the lowerer, not the in machine dependent Lower(er)MD classes. This one returns 4 always and is fine as the offset is guaranteed to be from ebp.

PR #301, opened 2016-02-11
Use smaller data type to represent the counter fields of FunctionBody.

leirocks:
Use smaller data type to represent the counter fields of FunctionBody.

Most of the counters has value less than 256, use single byte to hold them. And can promote to short or full 32 bit integer when necessary.

### DIFF
--- a/lib/Runtime/Base/CompactCounters.h
+++ b/lib/Runtime/Base/CompactCounters.h
@@ -12,7 +12,7 @@ namespace Js
     template<class T, typename CountT = T::CounterFields>
     struct CompactCounters
     {
-        volatile uint8 fieldSize;
+        uint8 fieldSize;
 #if DBG
         mutable bool bgThreadCallStarted;
         bool isCleaningUp;
@@ -25,7 +25,6 @@ namespace Js
             WriteBarrierPtr<int16> i16Fields;
             WriteBarrierPtr<int32> i32Fields;
         };
-
 
         CompactCounters()
             :fieldSize(1)
--- a/lib/Runtime/Base/CompactCounters.h
+++ b/lib/Runtime/Base/CompactCounters.h
@@ -1,0 +1,274 @@
+//-------------------------------------------------------------------------------------------------------
+// Copyright (C) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE.txt file in the project root for full license information.
+//-------------------------------------------------------------------------------------------------------
+#pragma once
+
+#pragma warning(push)
+#pragma warning(disable:6200) // C6200: Index is out of valid index range, compiler complains here we use variable length array
+
+namespace Js
+{
+    template<class T, typename CountT = T::CounterFields>
+    struct CompactCounters
+    {
+        volatile uint8 fieldSize;
+#if DBG
+        mutable bool bgThreadCallStarted;
+        bool isCleaningUp;
+#endif
+        union {
+            uint8 u8fields[CountT::Max];
+            int8 i8fields[CountT::Max];
+            WriteBarrierPtr<uint16> u16Fields;
+            WriteBarrierPtr<uint32> u32Fields;
+            WriteBarrierPtr<int16> i16Fields;
+            WriteBarrierPtr<int32> i32Fields;
+        };
+
+
+        CompactCounters()
+            :fieldSize(1)
+#if DBG
+            , bgThreadCallStarted(false), isCleaningUp(false)
+#endif
+        {
+            memset(u8fields, 0, (uint8)CountT::Max);
+        }
+
+        void AllocCounters(T* host, uint8 newSize);
+
+        uint32 Get(CountT typeEnum) const
+        {
+#if DBG
+            if (ThreadContext::GetContextForCurrentThread() == nullptr)
+            {
+                bgThreadCallStarted = true;
+            }
+#endif
+            uint8 type = static_cast<uint8>(typeEnum);
+            uint8 localFieldSize = fieldSize;
+            uint32 value = 0;
+            if (localFieldSize == 1)
+            {
+                value = this->u8fields[type];
+            }
+            else if (localFieldSize == 2)
+            {
+                value = this->u16Fields[type];
+            }
+            else
+            {
+                Assert(localFieldSize == 4);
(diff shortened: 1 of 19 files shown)


lib/Runtime/Base/FunctionBody.h
```diff
@@ -1658,6 +1660,74 @@ namespace Js
...
+                LocalClosureRegister                    = 20,
+                LocalFrameDisplayRegister               = 21,
+                EnvRegister                             = 22,
+                ThisRegForEventHandler                  = 23,
+                FirstInnerScopeRegister                 = 24,
+                FuncExprScopeRegister                   = 25,
+
+                // Signed integers need keep the sign when promoting 
+                SignedFieldsStart                       = 26,
+                SerializationIndex                      = 26,
```
ianwjhalliday: These two are both 26

ianwjhalliday:
With this and your previous change how much does this shrink FunctionBody down on x64?

ianwjhalliday:
Ignoring the separately added data for the pointers and counters.  I'm curious about sizeof(FunctionBody) for curiosity's sake.

leirocks:
@ianwjhalliday it reduced from 648 to 384 bytes

leirocks:
@dotnet-bot test this please

lib/Runtime/Base/FunctionBody.h
leirocks: SignedFieldsStart  is a marker that from that point, it saves signed value. maybe I can remove this, instead, use a byte in the structure to save the position -- there're 3 bytes at the beginning is not used.

ianwjhalliday:
Nice, that is a significant reduction.

lib/Runtime/Base/FunctionBody.h
ianwjhalliday: Oh gotcha, I missed that SignedFieldsStart was a label marking the beginning of special fields.

leirocks:

lib/Runtime/Library/ScriptFunction.cpp
```diff
@@ -594,7 +594,7 @@ namespace Js
                     {
                         memset(this->m_inlineCaches[i], 0, sizeof(InlineCache));
                     }
-                    else if(!scriptContext->IsClosed())
+                    else if(scriptContext != nullptr && !scriptContext->IsClosed())
```
curtisman: In what situation can the scriptContext be nullptr?

lib/Runtime/Base/CompactCounters.h
```diff
@@ -0,0 +1,269 @@
...
+//-------------------------------------------------------------------------------------------------------
+#pragma once
+
+#pragma warning(push)
+#pragma warning(disable:6200) // C6200: Index is out of valid index range, compiler complains here we use variable length array
+
+namespace Js
+{
+    template<class T, typename CountT = T::CounterFields>
+    struct Counter
```
curtisman: This is name is kind of generic?  Rename to CompactCounters?

lib/Runtime/Base/CompactCounters.h
```diff
@@ -0,0 +1,269 @@
...
+
+            // check if main thread has updated the structure
+            if (localFieldSize == fieldSize) 
+            {
+                return value;
+            }
+            else 
+            {
+                Assert(!ThreadContext::GetContextForCurrentThread());
+                AutoCriticalSection autocs(T::GetLock());
```
curtisman: Is there any interpreter perf impact from this lock?  Is it rare enough to be have to use a larger size that it doesn't matter?

lib/Runtime/Base/CompactCounters.h
```diff
@@ -0,0 +1,269 @@
...
+                    AllocCounters(host, 4);
+                }
+                return this->SetSigned(typeEnum, val, host);
+            }
+
+            Assert(fieldSize == 4);
+            return this->i32Fileds[type] = val;
+        }
+
+        uint32 Increase(CountT typeEnum, T* host)
```
curtisman: Increment?

lib/Runtime/Base/CompactCounters.h
```diff
@@ -0,0 +1,269 @@
...
+{
+    template<class T, typename CountT = T::CounterFields>
+    struct Counter
+    {
+
+        volatile uint8 fieldSize;
+        union {
+            uint8 u8fields[CountT::Max];
+            int8 i8fields[CountT::Max];
+            WriteBarrierPtr<uint16> u16Fileds;
```
curtisman: Fileds = Fields?

lib/Runtime/Base/CompactCounters.h
```diff
@@ -0,0 +1,269 @@
...
+            uint8 type = static_cast<uint8>(typeEnum);
+            uint8 localFieldSize = fieldSize;
+            uint32 value = 0;
+            if (localFieldSize == 1)
+            {
+                value = this->u8fields[type];
+            }
+            else if (localFieldSize == 2)
+            {
+                value = this->u16Fileds[type];
```
curtisman: Since we didn't take a lock here, background thread could potentially read old value from u8fields here even after we switch to u16fields.   May guess it that it is probably ok, because we didn't have a lock before, so the background thread either is just access field that doesn't change after byte code gen, or some count that doesn't matter if we missed some.     But please verify that is true.

Please add a comment here too, so we won't think it is a bug in the future.

leirocks:
@curtisman as discussed, moved following mutable fields out of the structure. and added Assert in case there's any other fields missed.

        uint32 interpretedCount;
        uint32 loopInterpreterLimit;
        uint32 debuggerScopeIndex;
        uint32 savedPolymorphicCacheState;

lib/Runtime/Library/ScriptFunction.cpp
leirocks: ah, this is not intentional change. previously in order to solve the issue the when FunctionBody::Cleanup sometimes the scriptContext is already gone, I set this to null. So this is not needed any more because when cleaning up we should not be expanding the structure so should not need the recycler from scriptContext anymore.
let me remove the change. thanks.

curtisman:
LGTM

PR #305, opened 2016-02-12
Updates for 16-02

pleath:
#1: Fix bugs with non-canonical NaNs in asm.js

In asm.js, we don't canonicalize NaNs. We don't have vars in asm.js, so this is
ok. But we DO need to make sure to check for NaN if we are creating a
JavascriptNumber from it.
However, we were not doing this, which was an issue for argouts in asm.js->js
calls as well as return values from asm.js.
It was also an issue in one helper call on x64, where we were converting to var
before making the call.
This was being done because there was no support for x64 calling convention.
But with asm.js, I implemented that support, so I updated the helper call to
use the standard API. Other places were also hacking around this limitation, so
I updated them as well.

#2: add check against length for memset

#3: Issue:
There was a code path where we protect pages with PAGE_EXECUTE without the
PAGE_TARGETS_NO_UPDATE flag, which will control all the CFG bits in the
page.

Fix:
Fixed and ensured that we don't use PAGE_EXECUTE directly in the virtual
protect API. Instead #defines are used.
Cleaned up some complex code path in CustomHeap Protect Allocation
mechanism.
Removed isReadWrite flag from PageAllocation class, as it is not required.

Earlier, we were also letting RW pages to be added to the bucket list of
pages in custom heap, which lead the hacker to add his own RW page to the
list. Now it is made sure that only RX pages go into the bucket list.

Tests:
Unit tests passed.
Exprgen, Web crawler - Reran the tests and they look fine..


MikeHolman:
Thanks Paul, LGTM!

satheeshravi:
Thanks, Paul!

satheeshravi:
Thanks, Paul!

PR #307, opened 2016-02-12
Bugfix:OS5553123:Freeze while comparing valueOf()-injected null

suwc:
Bugfix:OS5553123:Freeze while comparing valueOf()-injected null

Chakra implementation of Abstract Relational Comparison operation fails to
handle 'null'. Specifically, it cause a browser tab to freeze when a comparison ("<")
is done with the left value being 'null' injected by valueOf() call through a prototype.
Add 'null' checking to fix this problem.
Add unit test.

### DIFF
--- a/test/Bugs/OS_5553123.js
+++ b/test/Bugs/OS_5553123.js
@@ -1,0 +1,31 @@
+//-------------------------------------------------------------------------------------------------------
+// Copyright (C) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE.txt file in the project root for full license information.
+//-------------------------------------------------------------------------------------------------------
+
+WScript.LoadScriptFile("..\\UnitTestFramework\\UnitTestFramework.js");
+
+function testRelationalComparison (retVal)
+{
+    var ObjectV = function ObjectV(v){ }
+
+    ObjectV.prototype = {
+        valueOf : function(){ return retVal; }
+    };
+
+    function f()
+    {
+        var x = new ObjectV(0);
+        x<"1";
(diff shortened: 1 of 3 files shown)


suwc:
@akroshg @tcare @abchatra could you take a look pls?

lib/Runtime/Language/JavascriptOperators.cpp
```diff
@@ -708,6 +708,7 @@ namespace Js
             dblRight = JavascriptConversion::ToNumber(aRight, scriptContext);
             break;
         case TypeIds_Boolean:
+        case TypeIds_Null:
```
akroshg: should handle the undefined and symbol cases here.

suwc: Good point. Updated in latest rev.

akroshg:
Looks good.

tcare:
LGTM, signing off.

tcare:
Actually, can we get tests for the undefined/symbol case? otherwise looks good.

test/Bugs/OS_5553123.js
```diff
@@ -0,0 +1,31 @@
...
+        x<"1";
+    }
+
+    f();
+    f();
+    f();
+}
+
+testRelationalComparison(null);
+testRelationalComparison(undefined);
```
suwc: @tcare, unit test covers 'undefined' and Symbol here.

tcare: Ah saw the wrong revision. thanks :)

PR #308, opened 2016-02-12
Fix AV in unittest when setting pointer to nullptr, we should never promote the AuxPtrs so remove the dependency on recycler on that case

leirocks:
Fix AV in unittest
when setting pointer to nullptr, we should never promote the AuxPtrs so remove the dependency on recycler on that case

### DIFF
--- a/lib/Runtime/Base/AuxPtrs.h
+++ b/lib/Runtime/Base/AuxPtrs.h
@@ -53,10 +53,10 @@ namespace Js
         AuxPtrs(uint8 capacity, AuxPtrs* ptr);      // called when expanding (i.e. promoting from AuxPtrs to bigger AuxPtrs)
         void* Get(FieldsEnum e);
         bool Set(FieldsEnum e, void* p);
-        static void AllocAuxPtrFix(T* pHost, uint8 size, Recycler* recycler);
-        static void AllocAuxPtr(T* pHost, uint8 count, Recycler* recycler);
-        static void* GetAuxPtr(const T* pHost, FieldsEnum e);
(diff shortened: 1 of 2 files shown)


lib/Runtime/Base/AuxPtrs.h
```diff
@@ -165,15 +165,16 @@ namespace Js
     }
 
     template<class T, typename FieldsEnum>
-    void AuxPtrs<T, FieldsEnum>::AllocAuxPtrFix(T* _this, uint8 size, Recycler* recycler)
+    void AuxPtrs<T, FieldsEnum>::AllocAuxPtrFix(T* pHost, uint8 size, Recycler* recycler)
```
MikeHolman: nit: imo, should just be `host`... `pHost` is pretty non-standard notation for our codebase.

leirocks: OK, thanks!

leirocks:
btw, another fix might be just set auxPtrs to nullptr instead of setting the pointers to null one by one. need to investigate if there's any side effect though. Use current fix to unblock test first

MikeHolman:
>another fix might be just set auxPtrs to nullptr instead of setting the pointers to null one by one

@leirocks Yes I was just thinking about this. Prior to your change, were we nulling out all the pointers? In either case, I think your current change is sufficient to unblock testing.

leirocks:
@MikeHolman from FunctionBody::CleanUp(), it's not nulling out all. but I'm not sure if there's pointer need to be accessed after that or just because those pointers won't cause circular reference and didn't null out them before. so need investigation and testing. I'm merging this change for now. thanks!

PR #315, opened 2016-02-15
Multiple bugfixes

kunalspathak:
Please refer to individual commits for bug fixes.


lib/jsrt/jsrtinternal.h
```diff
@@ -131,7 +131,7 @@ JsErrorCode ContextAPIWrapper(Fn fn)
     Js::ScriptContext *scriptContext = currentContext->GetScriptContext();
     try
     {
-        AUTO_NESTED_HANDLED_EXCEPTION_TYPE((ExceptionType)ExceptionType_JavascriptException);
```
Yongqu: Should we have same thing for stack overflow?

kunalspathak: Thought about that, but didn't want  to add without knowing if we will get OOS here. I will add if I see a reason to add it.

kunalspathak:
I had to rebase after `Lib` changes. Incorporated review comment by @Yongqu

Yongqu:
LGTM

Lib/Runtime/Base/WindowsGlobalizationAdapter.cpp
```diff
@@ -425,19 +425,68 @@ namespace Js
         return hr;
     }
 
-    void WindowsGlobalizationAdapter::ClearTimeZoneCalendars()
```
Yongqu: test. pls ignore

PR #202, opened 2016-01-26
Intl: Fail silently if windows.globalization.dll is missing

kunalspathak:
Intl: Fail silently if windows.globalization.dll is missing

Chakracore doesn't work on Windows 7 machines that doesn't have IE11.
We try to load `jsintl.dll` while trying to initialize `JavascriptLibrary` but throw
Error if we fail to load these binaries. This makes Chakracore unusable for
Win7 minus IE11 machines.

The fix is to ignore the error silently and keep executing. The downside to this is
`Intl` object will be empty and below properties won't be available because these
properties are populated based on globalization dlls.

```js
Intl.Collator
Intl.DateTimeFormat
Intl.NumberFormat
```

There are also specific locale specific APIs related to DateTime that won't work
without globalization support.

The upside is Chakracore will work on these machines for non-locale javascript code.

PR #318, opened 2016-02-16
Suppress error message if failed to load globalization dlls

kunalspathak:
In Microsoft/ChakraCore#202, we stop complaining if globalization dlls are missing
from the machine. However at one place we were printing the error message.
Removed the printing for such case. This was also causing all test to fail in
windows 7 because they were printing this message causing baseline mismatch.


Yongqu:
Lgtm

kunalspathak:
Thanks @Yongqu

dilijev:
I'll re-run the dailies to make sure this is fixed.

PR #321, opened 2016-02-17
Default related bug fixes.

akroshg:
1. Function at default are not fully qualified as the right name hint is not passed and used when there is a function - fixed that.
2. We don't create the deferredStub for function at param scope but we were trying to use m_currentDeferredStub when we are parsing functions at paramscope (inner function of the function at param scope). Using that  m_currentDeferredStub is plain wrong for the paramscope as we haven't created any stub for them so fixed that by setting m_currentDeferredStub to null when the current function is enclosed within param scope. (Unfortunately there was no smaller repro available to this - so I am not adding unittest)


akroshg:
@aneeshdk - could you look at this one?

aneeshdk:
LGTM

PR #322, opened 2016-02-17
Fixes `get` and `set` not parsing in object literal shorthand syntax

### LINKED ISSUE #320: `{get}` should be a valid object literal
We currently throw a syntax error for the following case:

```js
let get = 1;
let obj = { get };
```

ianwjhalliday:
Fixes #320

Also renamed wellKnownPropertyPids.getter/setter to get/set to reflect
the actual PID they represent.

### DIFF
--- a/test/es6/objlit.js
+++ b/test/es6/objlit.js
@@ -15,15 +15,58 @@ var tests = [
         }
     },
     {
-        name: "Method shorthand",
+        name: "Shorthand names `get` and `set` parse without error (they are not keywords in these cases)",
+        body: function () {
+            var a = 0;
+            var get = 1;
+            var set = 2;
+            var z = 3;
+
+            var o = { get };
+            var p = { set };
+            var q = { get, set };
+            var r = { set, get };
+            var s = { get, z };
+            var t = { a, set };
+            var u = { a, get, z };
+
+            assert.areEqual(1, o.get, "o.get = 1");
+            assert.areEqual(2, p.set, "p.set = 2");
+            assert.areEqual(1, q.get, "q.get = 1");
+            assert.areEqual(2, q.set, "q.set = 2");
+            assert.areEqual(2, r.set, "r.set = 2");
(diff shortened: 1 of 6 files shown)


ianwjhalliday:
@kunalspathak @boingoing @akroshg @aneeshdk please review
cc @abchatra @bterlson

kunalspathak:
:+1:

aneeshdk:
LGTM

kunalspathak:
@ianwjhalliday , once merged to master, please port it to release/1.1 branch and bump up the QFE version number.

ianwjhalliday:
Updated with fixes and refactored the get and set cases together into one if block.

test/es6/objlit.js
```diff
@@ -15,15 +15,58 @@ var tests = [
...
+            var a = 0;
+            var get = 1;
+            var set = 2;
+            var z = 3;
+
+            var o = { get };
+            var p = { set };
+            var q = { get, set };
+            var r = { set, get };
+            var s = { get, z };
```
akroshg: How about {get:get}?

ianwjhalliday: I did a grep of the tests and found that `{ get: ... }` is a very common pattern and therefore well tested already.  `get` appearing after the colon is the same as it appearing in any expression context and I figure is also already well tested.

akroshg: sounds good. This `var {get} = {}`  will not be error after your change, right?

ianwjhalliday: Correct.  I added a new test for `get` and `set` name use in object destructuring to `test/es6/destructuring_obj.js`

akroshg: Looks good.

ianwjhalliday:
Rebased the fix bugs commit as a fixup

ianwjhalliday:
@pleath would you mind taking a look at this one?

PR #326, opened 2016-02-17
Fixes this, arguments, and super in async functions

ianwjhalliday:
Makes the inner generator function of an async function a lambda so that
it captures this, arguments, and super from its parent.

Generator lambdas are not supported by the language syntax but we can
create one internally without restriction.

### DIFF
--- a/lib/Parser/Parse.cpp
+++ b/lib/Parser/Parse.cpp
@@ -7259,6 +7259,10 @@ void Parser::TransformAsyncFncDeclAST(ParseNodePtr *pnodeBody, bool fLambda)
     {
         GetCurrentFunctionNode()->sxFnc.SetStrictMode();
     }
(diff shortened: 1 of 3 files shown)


ianwjhalliday:
@pleath @akroshg @aneeshdk @boingoing please review
cc @abchatra @bterlson

pleath:
LGTM. Cool that it was that simple, eh?

abchatra:
:+1:

ianwjhalliday:
@pleath yeah! I wrote the tests first which took the bulk of the time.  When I first made the function a lambda the arguments test failed but the other two passed and I was very happy.  After a quick debug session I was able to get the arguments test to pass as well.  Hooray!

ianwjhalliday:
Rebased to fix merge conflict

PR #327, opened 2016-02-17
Move nested function array from the end of FunctionBody to seperate structure

leirocks:
Move nested function array from the end of FunctionBody to seperate
structure

Nested function array was allocated as a plus size at the end of
FunctionBody. FunctionBody is a fairly big structure, with random amount
of nested function at the end, FunctionBodys use different bucket from the
recycler and can cause memory fragmentation.
this change also remove field m_derivedSize which is not necessory any
more(it was there for locating nested function array), and m_nestedCount
will be moved to the new structure, it does not use memory for those
function does not have nested functions


curtisman:
How much does it improve?

leirocks:
This one is improvement for potential fragmentation. I didn't measure. let me try to measure it. I think the noise could be big. let me try and see

Lib/Runtime/Base/FunctionBody.cpp
```diff
@@ -1126,6 +1124,13 @@ namespace Js
 #endif
       isByteCodeDebugMode(false)
     {
+        if (nestedCount > 0)
+        {
+            NestedArray* nestedArray = RecyclerNewWithBarrierPlusZ(m_scriptContext->GetRecycler(),
+                nestedCount*sizeof(WriteBarrierPtr<FunctionProxy>), NestedArray, nestedCount);
+            SetAuxPtr(AuxPointerType::NestedArray, nestedArray);
```
curtisman: What is the data that support this to be in the Aux structure?  Are most function "leaf" functions?

Lib/Runtime/Base/FunctionBody.h
```diff
@@ -1211,12 +1212,42 @@ namespace Js
             SetDisplayNameFlagsRecyclerAllocated = 2
         };
 
+        struct NestedArray
+        {
+            NestedArray(uint32 count) :nestedCount(count) {}
+            uint32 nestedCount;
+            WriteBarrierPtr<FunctionProxy> array[1];
```
curtisman: "array" is too generic a name.
How about functionProxyArray?

Lib/Runtime/Base/FunctionBody.cpp
leirocks: I was also hesitate to put this into Aux structure or directly attach to functionProxy.
from my statistic, 3686 out of 5811 functionBodys have m_nestedCount equals 0. I didn't measure FunctionProxy and ParseableFunctionInfo. let me to that measurement as well to make better decision. thanks!

Lib/Runtime/Base/FunctionBody.cpp
```diff
@@ -1126,6 +1124,13 @@ namespace Js
 #endif
       isByteCodeDebugMode(false)
     {
+        if (nestedCount > 0)
```
curtisman: Is the nestedArray only needed for FunctionBody? 
Is it needed by the ParseableFunctionInfo when it is still in defer parsed state?

Lib/Runtime/Base/FunctionBody.cpp
```diff
@@ -4490,20 +4494,29 @@ namespace Js
 
         if (!fOnlyCurrent)
         {
-            for (uint uIndex = 0; uIndex < this->m_nestedCount; uIndex++)
+            this->ForEachNestedFunc([&](FunctionProxy* proxy, uint32 index)
             {
-                Js::ParseableFunctionInfo * pBody = this->GetNestedFunctionForExecution(uIndex);
+                FunctionProxy* currentNestedFunction = this->GetNestedFunc(index);
```
curtisman: Why is GetNestedFunctionForExecution expanded here?

Lib/Runtime/ByteCode/ByteCodeSerializer.cpp
```diff
@@ -2040,11 +2039,7 @@ class ByteCodeBufferBuilder
                 "Only the ErrorOnNew|SuperReference|Lambda|CapturesThis|Generator|ClassConstructor|Async|ClassMember attributes should be set on a serialized function");
 
         PrependInt32(builder, L"Offset Into Source", sourceDiff);
-        if (function->GetNestedCount() > 0)
-        {
-            definedFields.has_m_nestedCount = true;
-            PrependInt32(builder, L"Nested count", function->GetNestedCount());
-        }
+        PrependInt32(builder, L"Nested count", function->GetNestedCount());
```
curtisman: This grows the serialized by code if nested count is 0.

Lib/Runtime/Base/FunctionBody.cpp
leirocks: using the ForEach so it does not need to get the array everytime, just notice this is debugger only code path, let me change it back

Lib/Runtime/ByteCode/ByteCodeSerializer.cpp
leirocks: m_nestedCount is removed from the functionbody,  added a has_m_nestedCount in serializedFieldList and get the 4 bytes back

Lib/Runtime/Base/FunctionBody.cpp
leirocks: it's in ParseableFunctionInfo

Lib/Runtime/Base/FunctionBody.h
leirocks: fixed

leirocks:
@curtisman can you review this as well?

curtisman:
LGTM.  Still want to know how much this will improve things.

leirocks:
the improvement from comparing on facebook (baseline and fix each twice):
baseline:
-----------------------------------------------------------------------------------------
Total           :  3716  527160       1    47244320    19442656    66686976    70.84%
-----------------------------------------------------------------------------------------
Total           :  3667  522891       1    46512768    19129728    65642496    70.86%

with fix:
-----------------------------------------------------------------------------------------
Total           :  3661  535384       1    48547392    16771520    65318912    74.32%
-----------------------------------------------------------------------------------------
Total           :  3689  526514       1    48244240    18012656    66256896    72.81%


usually I see the usage is about 72-73%, may be I just hit a extreme cases for baseline today.

PR #334, opened 2016-02-18
[linux] Build Chakra.Common.Memory

digitalinfinity:
This change gets Chakra.Common.Memory compiling on linux. It includes
the following:
1. Disable concurrent and partial GC on non-Windows build
2. Disable stack trace collection in page heap mode by wrapping under
the existing STACK_BACK_TRACE macro
3. Added a declaration for a wrapper method GetCurrentThreadStackBounds,
which would abstract out the getting of thread stack bounds on at least
x86 base platforms. Replaced cases where we were using kernel APIs to
get the stack bounds with this API (not yet implemented)
4. Disable static valid pointer map on Linux
5. Replace string usage in valid pointer map generation code with
CH_WSTRs/wchar16
6. Added a few template specialization forward declarations

There are xplat-todos sprinkled around this change- I'll collate them
and start opening issues where it makes sense.

### DIFF
--- a/lib/Common/Memory/AutoAllocatorObjectPtr.h
+++ b/lib/Common/Memory/AutoAllocatorObjectPtr.h
@@ -15,7 +15,7 @@ private:
     AllocatorType* m_allocator;
 
 public:
-    AutoAllocatorObjectPtr(T* ptr, AllocatorType* allocator) : BasePtr(ptr), m_allocator(allocator)
+    AutoAllocatorObjectPtr(T* ptr, AllocatorType* allocator) : BasePtr<T>(ptr), m_allocator(allocator)
     {
         Assert(allocator);
     }
@@ -28,10 +28,10 @@ public:
 private:
     void Clear()
     {
-        if (ptr != nullptr)
+        if (this->ptr != nullptr)
         {
-            DeleteObject<TAllocator>(m_allocator, ptr);
-            ptr = nullptr;
+            DeleteObject<TAllocator>(m_allocator, this->ptr);
+            this->ptr = nullptr;
         }
     }
 };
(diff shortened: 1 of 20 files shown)


digitalinfinity:
@curtisman @jianchun could you guys take a look at this change please? Thanks!

lib/Common/Memory/AutoAllocatorObjectPtr.h
```diff
@@ -77,11 +77,14 @@ class AutoAllocatorArrayPtr : public BasePtr<T>
 //      TAllocator      The allocator type used to allocate/free the objects.
 //      ArrayAllocator  The allocator type used to allocate/free the array.
 //
-template <typename T, typename TAllocator, typename ArrayAllocator = ForceNonLeafAllocator<TAllocator>::AllocatorType>
+template <typename T, typename TAllocator, typename ArrayAllocator =
+typename ForceNonLeafAllocator<TAllocator>::AllocatorType>
```
jianchun: nit: indent this (can do later)

lib/Common/Memory/Recycler.cpp
```diff
@@ -1467,7 +1479,7 @@ Recycler::ScanStack()
     SAVE_THREAD_CONTEXT();
     void * stackTop = this->savedThreadContext.GetStackTop();
 
-    void * stackStart = GetStackBase();
+    void * stackStart = GetStackBase();
```
jianchun: nit: trailing spaces

lib/Common/Memory/RecyclerWriteBarrierManager.cpp
```diff
@@ -225,11 +231,6 @@ X64WriteBarrierCardTableManager::Initialize()
 
         LPVOID cardTableSpace = ::VirtualAlloc(NULL, _cardTableNumEntries, MEM_RESERVE, PAGE_READWRITE);
 
-        if (cardTableSpace == nullptr)
```
jianchun: Is this removed intentionally? cardTableSpace can't be nullptr here, or is not used?

jianchun:
LGTM!

lib/Common/Memory/Recycler.cpp
digitalinfinity: Fixed

lib/Common/Memory/AutoAllocatorObjectPtr.h
digitalinfinity: Fixed

lib/Common/Memory/RecyclerWriteBarrierManager.cpp
digitalinfinity: This was a weird one :) if it returned null, we were returning false (which clang didn't like)- if we remove the check, we'd return null anyway, so the check was unnecessary

PR #253, opened 2016-02-02
String.prototype.toLowerCase()/.toUpperCase() to use WINAPI

suwc:
Improve support for unicode, in particular for Deseret alphabet by
switching to Windows API CharLowerBuff() and CharUpperBuff().
Remove internal functions that are no longer used.
Add unit test.

PR #338, opened 2016-02-18
Don't allocate new strings when re-casing a string which didn't need it

FremyCompany:
This is my first ChakraCore pull request. As you can probably guess from the code I'm not a very experienced C++ developer. I am not even sure this code actually improves performance, since it makes it longer, but I wanted to see what you guys thought about the idea.

Being a Microsoft Employee I don't think I need to sign a CLA or something, but if I am wrong let me know.

### DIFF
--- a/lib/Runtime/Library/JavascriptString.cpp
+++ b/lib/Runtime/Library/JavascriptString.cpp
@@ -2195,7 +2195,7 @@ case_2:
 #endif
                 CharLowerBuffW(outStr + countToSkip, countToConvert);
 
-            Assert(converted == count);
+            Assert(converted == countToConvert);
         }
 
         return builder.ToString();
--- a/lib/Runtime/Library/JavascriptString.cpp
+++ b/lib/Runtime/Library/JavascriptString.cpp
@@ -2158,7 +2158,8 @@ case_2:
         }
         
         // if we reached the end of the string and didn't need recasing, we can return immediately
-        if(i == iLim) { 
+        if(i == iLim)
+        { 
             return pThis;
         }
         
--- a/lib/Runtime/Library/JavascriptString.cpp
+++ b/lib/Runtime/Library/JavascriptString.cpp
@@ -2136,38 +2136,68 @@ case_2:
         charcount_t count = pThis->GetLength();
 
         BufferStringBuilder builder(count, pThis->type->GetScriptContext());
-        const wchar_t *inStr = pThis->GetString();
-        wchar_t *outStr = builder.DangerousGetWritableBuffer();
+        const wchar_t* inStr = pThis->GetString();
+		wchar_t* i = inStr; 
+		wchar_t* iLim = i + count;
+		
+		// first, try to see if the string needs recasing at all
+		while(i < iLim)
+		{
+			
+			// copy the current char into a temporary variable, then case it
+			wchar_t inChar = *i;
+			wchar_t outChar = *i;
+			CharUpperBuffW(&outChar, 1);
+			
+			// if the output is different from the input, the string needs recasing from here
+			if(outChar != inChar) { break; }
+			
+			// otherwise, we can keep going on
+			i++;
+			
+		}
+		
+		// if we reached the end of the string and didn't need recasing, we can return immediately
+		if(i == iLim) { 
+			return pThis;
+		}
+		
+		// if not, we have to copy the first part of the string, then case the rest
+		charcount_t countToSkip = i - inStr;
+		charcount_t countToConvert = count - countToSkip; 
+		wchar_t* outStr = builder.DangerousGetWritableBuffer();
 
-        wchar_t* outStrLim = outStr + count;
-        wchar_t *o = outStr;
+		// start by copying the whole string in a new buffer
+		wchar_t* outStrLim = outStr + count;
+		wchar_t* o = outStr;
 
-        while (o < outStrLim)
-        {
-            *o++ = *inStr++;
-        }
-
-        if(toCase == ToUpper)
-        {
+		while (o < outStrLim)
+		{
+			*o++ = *inStr++;
+		}
+		
+		// then recase it
+		if(toCase == ToUpper)
+		{
 #if DBG
-            DWORD converted =
+			DWORD converted =
 #endif
-                CharUpperBuffW(outStr, count);
+				CharUpperBuffW(outStr + countToSkip, countToConvert);
 
-            Assert(converted == count);
-        }
-        else
-        {
-            Assert(toCase == ToLower);
+			Assert(converted == countToConvert);
+		}
+		else
+		{
+			Assert(toCase == ToLower);
 #if DBG
-            DWORD converted =
+			DWORD converted =
 #endif
-                CharLowerBuffW(outStr, count);
+				CharLowerBuffW(outStr + countToSkip, countToConvert);
 
-            Assert(converted == count);
-        }
+			Assert(converted == count);
+		}
 
-        return builder.ToString();
+		return builder.ToString();
     }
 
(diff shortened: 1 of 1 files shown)


dilijev:
@FremyCompany if you're a member of the Microsoft GitHub organization, you don't need to sign the CLA. Otherwise you do. The @msftclas bot indicates that you aren't (otherwise it would have tagged this PR as cla-not-required). Even if you're an MS employee, no harm in signing the CLA anyway. As for getting your GitHub account added to the MS organization on GitHub, you can send our team email and we'll send you instructions.

dilijev:
As for the code, no need to be too harsh of a critic on yourself. We'll evaluate this PR and if it doesn't hurt performance, we can help you get it where it needs to go. To see whether this improves performance, you should first convince yourself that is helps (or at least doesn't hurt) by doing many iterations of a test with many loops to exercise the types of cases you're trying to optimize. Run that test before and after your change and measure the performance change.

Note that longer code doesn't mean longer execution time. As a general rule, you make a program faster by it do fewer things (fewer iterations, less wasteful buffer copies, and so on). Adding more checks to remove unnecessary loop iterations and shortcuts is sometimes how we do that. Often, a faster program has more source code.

The only thing to worry about is whether your extra checks introduce a lot of overhead to the cases which do not hit your optimization path. In that case the situation you're trying to improve might get better but average program performance would get worse. That's the sort of thing you should try to evaluate before we consider accepting this change.

If you run some tests, please provide your test harness, method of testing, and results here in the comments.

Thanks!

lib/Runtime/Library/JavascriptString.cpp
```diff
@@ -2135,26 +2135,57 @@ namespace Js
...
 
-        BufferStringBuilder builder(count, pThis->type->GetScriptContext());
-        const wchar_t *inStr = pThis->GetString();
-        wchar_t *outStr = builder.DangerousGetWritableBuffer();
+        const wchar_t* inStr = pThis->GetString();
+        const wchar_t* i = inStr; 
+        const wchar_t* iLim = i + count;
+        
+        // first, try to see if the string needs recasing at all
+        while(i < iLim)
```
dilijev: `while` and `if` statements should have a space between the keyword and the `(`, to match the style elsewhere in this file.

lib/Runtime/Library/JavascriptString.cpp
```diff
@@ -2135,26 +2135,57 @@ namespace Js
...
+        wchar_t* o = outStr;
 
         while (o < outStrLim)
         {
             *o++ = *inStr++;
         }
-
+        
+        // then recase it
         if(toCase == ToUpper)
```
dilijev: Since you're making changes nearby, you can add the space here `if (` as well.

lib/Runtime/Library/JavascriptString.cpp
```diff
@@ -2135,26 +2135,57 @@ namespace Js
...
+        const wchar_t* iLim = i + count;
+        
+        // first, try to see if the string needs recasing at all
+        while(i < iLim)
+        {
+            
+            // copy the current char into a temporary variable, then case it
+            wchar_t inChar = *i;
+            wchar_t outChar = *i;
+            CharUpperBuffW(&outChar, 1);
```
dilijev: I don't think this is the right way to check for this. This method `ToCaseCore` can be used for converting to Upper- or Lower-case.

Additionally I don't think this method will be inlined, so there's a lot of function call overhead added here. What are the gains for pre-processing the string like this when you're attempting to run the conversion on every character anyway?

There might be something in the general idea of this pull request but I doubt this implementation will result in a performance improvement.

ianwjhalliday:
I was thinking about this earlier today, wondering how to make it fast without having to visit all characters in the string an extra time.  Is the GC allocation costly versus visiting the entire string an extra time?  Probably not.  If the Upper/LowerCharBufW APIs told us whether they changed the string or not we could decide to abandon the new allocation and return the original string, letting the new one get collected on the next GC, but would that help performance?  Perhaps not, unless there is a scenario that frequently calls `toLowerCase()` or `toUpperCase()`.

A change like this should be motivated by a real world scenario, either a poor performing code path in a real website (or product hosting the engine), or a poor performing path in a benchmark.

Without a real scenario in hand to test against there is no way to judge whether a change like this is making a difference.

I microbenchmark can help reveal a difference in performance, but making sure a microbenchmark models real scenarios can be tricky.

FremyCompany:
Yes, it is a bit sad the function does not tell you if it changed anything. I agree with you the frequent calls are likely to cause a slowdown. 

My use case was a data collection script which called toLowerCase() on a big string which actually was already lowercased very often, and which is actually not used after toLowerCase() as it is the result of a computation.

I figured out it would be a good idea to try to optimize this, but I didn't find a good way to fix this without trying to use our own function to "upper" or "lower" chars... What if we build a quick and dirty isProbablyLower and isProbablyUpper function, which would work for ASCII and return false for things outside ASCII or that are of the wrong case? That would allow to check the string efficiently, what do you think?

ianwjhalliday:
That might help.  If it is the toUpper case you could loop through and see if all the characters are ascii and outside the lowercase ascii range.  If you reach the end then there is no work to do and return the string as is.  Vice versa for toLower.  If you break out of this loop early, then you have a pointer to the character where you need to start checking, so you could call Upper/LowerCharBufW with this offset into the destination string.

FremyCompany:
Ok I'm going to try to implement that this weekend and see if it improves the perf on some test cases.

ianwjhalliday:
Rad, look forward to seeing your next update

abchatra:
Thank you for your pull request.  Here is my opinion:

Advantages of current code:
•	Simple (less code).
•	Performs better on strings which are not of intended case. 

Advantages of new change proposed:
•	Performs better on strings which are already in the intended case. 

I would not change the current built-in unless there is sufficient data to prove the new change is beneficial at the cost of simplicity (which helps maintainability). It would be great if you can collect that data before the code change. Something like 70% of the strings in the web are of intended case then this makes worthwhile effort. Anything less than 50% is not. 

Also there is a bug in the code as it doesn’t check for lowercase conversion.

lib/Runtime/Library/JavascriptString.cpp
```diff
@@ -2135,26 +2135,57 @@ namespace Js
...
-
+        
+        // then recase it
         if(toCase == ToUpper)
         {
 #if DBG
             DWORD converted =
 #endif
-                CharUpperBuffW(outStr, count);
+                CharUpperBuffW(outStr + countToSkip, countToConvert);
```
dreamer-dead: @dilijev , how do you think, it is correct to use WinAPI CharUpperBuff for string case conversion?
It is a common approach that length of uppercase string will be equal to length of lowercase string, but it is wrong.
For example http://stackoverflow.com/a/20301720

goyakin: @dreamer-dead You're right, `CharUpperBuffW` doesn't convert the characters that would increase the length of the string. We previously had our own mapping logic, which also didn't handle those characters, but decided to [replace](https://github.com/Microsoft/ChakraCore/pull/253) it with the Windows API to improve our Unicode support since this API handles a superset of our old implementation.

It doesn't look like we had a GitHub issue tracking these special characters, so I've created [one](https://github.com/Microsoft/ChakraCore/issues/421).

abchatra:
@FremyCompany  Any update on this PR?

FremyCompany:
Yes please don't close it just yet, I had to skip working on this last weekends due to other work, but I'm definitely going to work on it this weekend.

FremyCompany:
This discussion moved to #468 to start anew now that I have a better idea of what I am doing.

ianwjhalliday:
@FremyCompany is this PR superseded entirely by your new one?  If so could you close this PR please?

PR #340, opened 2016-02-18
Change the sym binding of debugger from bytecode binding to parser binding.

akroshg:
The debugger does the bytecode binding for the symbol even though it was already done during the parser time. Removing this unnecessary binding.


akroshg:
Unittest clean. Debugger (F12) clean. @pleath @abchatra  could you take a look at this one?

abchatra:
Please add more information on preference on parser binding instead of bytecode binding.

akroshg:
The sym binding under debugger is moved to Parse time to be consistent with non-debug mode binding. 
Parse time binding has some advantage over bytecode time binding. Some of the perf decision is naturally coupled with the information we get during the parse time. Also we don’t have to wait till the bytecode gen phase – this favors lot in deferring bytecode for multiple functions.

PR #348, opened 2016-02-19
Fix issue with monomorphic type checks

pleath:
With equivalent obj type spec on by default, certain operations still require monomorphic type checks. We've handled this in the dead store pass by locating any equivalent type check that protects an operation that requires a specific type and changing it to use a monomorphic check. But because of the vagaries of inline cache sharing, the operation that does the check may not appear to be monomorphic. Deal with this by having the dead store pass indicate the required type on the checking instruction.

### DIFF
--- a/test/fieldopts/fixedfieldmonocheck.js
+++ b/test/fieldopts/fixedfieldmonocheck.js
@@ -1,0 +1,55 @@
+//-------------------------------------------------------------------------------------------------------
+// Copyright (C) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE.txt file in the project root for full license information.
+//-------------------------------------------------------------------------------------------------------
+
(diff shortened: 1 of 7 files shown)


pleath:
@LouisLaf @rajatd

test/fieldopts/fixedfieldmonocheck.js
```diff
@@ -0,0 +1,50 @@
+function test0() {
```
rajatd: missing copyright

rajatd:
Looks good sans the missing copyright check.

pleath:
"sans"? Look at you. 😊

-- Paul

rajatd:
Hahaha :P

pleath:
@dotnet-bot please test

PR #349, opened 2016-02-19
Fixes global let shadowing non-configurable global properties

### LINKED ISSUE #165: `let undefined` in global scope should be an early Syntax error
```js
print('not early');
let undefined;
```

* Expected: Syntax error with nothing printed.
* Actual: not early is printed, then Syntax error is thrown.

See [GDI step 5.c+d](http://tc39.github.io/ecma262/#sec-globaldeclarationinstantiation).

ianwjhalliday:
Fixes #165

The ES6 spec states that an early error should be thrown if a global
lexical declaration conflicts with a non-configurable global property.

Fix by updating the global script preamble check for declaration
conflicts to also incorporate non-configurable globals.

### DIFF
--- a/test/es6/letconst_global_shadow_builtins.baseline
+++ b/test/es6/letconst_global_shadow_builtins.baseline
@@ -1,4 +1,6 @@
 [object Math]
+ReferenceError: Use before declaration
+ReferenceError: Use before declaration
 [object Math]
 3.141592653589793
 undefined
@@ -6,15 +8,11 @@ undefined
 undefined
 delicious
 [object JSON]
+ReferenceError: Use before declaration
+ReferenceError: Use before declaration
 [object JSON]
 function stringify() { [native code] }
 undefined
 [object Object]
 undefined
 no dice
-NaN
-NaN
-Infinity
-Infinity
-undefined
-undefined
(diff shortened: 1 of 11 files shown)


test/es6/letconst_global_shadow_builtins.baseline
```diff
@@ -1,4 +1,6 @@
 [object Math]
+ReferenceError: Use before declaration
```
pleath: Is Use before decl what we want for these cases? Or would Redeclaration be more intuitive?

ianwjhalliday: These cases are actual use before declarations, not redeclarations.  See below the new try/catches replacing the commented out test cases in `letconst_global_shadow_builtins.js`

abchatra:
Is this change compatible with Chrome?

ianwjhalliday:
Yes, Chrome already had this behavior for a while.

test/es6/letconst_global_shadow_builtins.baseline
pleath: Ah. Okay. LGTM, then.

PR #353, opened 2016-02-19
Change the term 'blacklist' to refer more specifically to unscopables.

dilijev:


ianwjhalliday:
Do we not like the term blacklist?  Does read more consistent with the feature.

LGTM

dilijev:
Policheck flagged it. From looking around the internet, there's some precedent for people being concerned by this.

PR #354, opened 2016-02-19
runtests.cmd: support extra command line flags

jianchun:
Add a few options to runtests.cmd:
```
-ExtraCCFlags "..."      Pass extra flags "..." directly to ch.exe
-DumpOnCrash             Pass "-DumpOnCrash" to ch.exe
-CrashOnException        Pass "-CrashOnException" to ch.exe
```

These could come in handy for investigating issues found in test suites.

### DIFF
--- a/test/runtests.cmd
+++ b/test/runtests.cmd
@@ -174,6 +174,11 @@ goto :main
   if /i "%1" == "-variants"         set _Variants=%~2&                                          goto :ArgOkShift2
   if /i "%1" == "-cleanupall"       set _CleanUpAll=1&                                          goto :ArgOk
 
+  ::Extra ch.exe command line flags
+  if /i "%1" == "-ExtraHostFlags"   set _ExtraHostFlags=%~2&                                    goto :ArgOkShift2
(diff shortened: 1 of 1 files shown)


test/runtests.cmd
```diff
@@ -174,6 +174,11 @@ goto :main
   if /i "%1" == "-variants"         set _Variants=%~2&                                          goto :ArgOkShift2
   if /i "%1" == "-cleanupall"       set _CleanUpAll=1&                                          goto :ArgOk
 
+  ::Extra ch.exe command line flags
+  if /i "%1" == "-ExtraCCFlags"     set _ExtraCCFlags=%~2&                                      goto :ArgOkShift2
```
ianwjhalliday: Can we call this `ExtraHostFlags` instead to move away from the C legacy, at least in the public facing interface?

ianwjhalliday:
LGTM!

test/runtests.cmd
jianchun: > Can we call this ExtraHostFlags instead to move away from the C legacy, at least in the public facing interface?

Good idea! Done.

PR #356, opened 2016-02-19
Silence misleading assert

pleath:
Prevent firing of a misleading assert, calling GetRootProperty without verifying that the instance is a root object. (Addresses a TODO in the source.)


ianwjhalliday:
LGTM

rajatd:
:+1:

PR #363, opened 2016-02-22
module runtime #2

Yongqu:
module runtime #2: merge change with parser; setup localexport slot to facilitate bytecode generation.  Setup ModuleRecord list in JavascriptLibrary to allow ModuleId lookup. Setup ModuleId for bytecode generation. When a new modulerecord is created, it is added to a list (moduleRecordList) in JavascriptLibrary. the index is the moduleID. There is a dependent childModuleRecord dictionary with specifier as key and modulerecord as value, this way we can generate the bytecode using the moduleId. All the local exports are stored in the localExportSlots, used in bytecode generator. Temporary disable anonymous function/class export. Need to fix the anonymous function binding.


Yongqu:
found the problem with class support in parser, thanks to help from @boingoing . I'll update next time.

lib/Parser/rterrors.h
```diff
@@ -336,3 +336,5 @@ RT_ERROR_MSG(JSERR_FunctionArgument_FirstCannotBeRegExp, 5640, "%s: first argume
 RT_ERROR_MSG(JSERR_RegExpExecInvalidReturnType, 5641, "%s: Return value of RegExp 'exec' is not an Object and is not null", "Return value of RegExp 'exec' is not an Object and is not null", kjstTypeError, 0)
 
 RT_ERROR_MSG(JSERR_ProxyTrapReturnedFalse, 5642, "Proxy trap `%s` returned false", "Proxy trap returned false", kjstTypeError, 0)
+RT_ERROR_MSG(JSERR_ModuleResolveExport, 5643, "module export %s cannot be resolved", "Module export cannot be resolved", kjstSyntaxError, 0)
```
boingoing: The casing for "module" in this area is inconsistent. Seems they should all be capitalized?

lib/Runtime/Language/SourceTextModuleRecord.cpp
```diff
@@ -50,15 +54,23 @@ namespace Js
...
+            return E_NOTIMPL;
+        }
         try
         {
             AUTO_NESTED_HANDLED_EXCEPTION_TYPE((ExceptionType)(ExceptionType_OutOfMemory | ExceptionType_StackOverflow));
             this->parser = (Parser*)AllocatorNew(ArenaAllocator, allocator, Parser, scriptContext);
+            this->srcInfo = {};
+            this->srcInfo.sourceContextInfo = scriptContext->CreateSourceContextInfo(sourceLength, Js::Constants::NoHostSourceContext);
+            // TODO: enable this for byte code generation.
+            // this->srcInfo.moduleID = moduleId;
```
boingoing: Is there a reason we need this to be commented-out now?

boingoing:
Looks good to me!

lib/Parser/rterrors.h
Yongqu: will do.

lib/Runtime/Language/SourceTextModuleRecord.cpp
Yongqu: it's causing assert in bytecode generator as it's confused with the old moduleroot stuff. We need to have different handling for moduleId when compile with module flag on.

PR #365, opened 2016-02-22
add remaining conversion ops + fix wasm JIT bugs

MikeHolman:
Adds remaining conversion ops. However, the truncation ops still need some work, as there is semantic difference in overflow scenario between wasm and js (wasm throws exception on overflow/NaN). I've also fixed a few bugs that came out of testing JIT code generation

[Review on Reviewable](https://reviewable.io/reviews/microsoft/chakracore/365)


MikeHolman:
Merging now so I can continue iterating. @abchatra @Cellule @LouisLaf feel free to review, and I'll later incorporate feedback.

PR #368, opened 2016-02-22
Improve JIT code page allocator virtual address use and reduced standing free page.

curtisman:
Share CodePageAllocator within a thread by moving it into ThreadContext.
This reduce the amount of virtual address we reserved, and tighter usage of the prereserved segment.

Also commit only page requested instead of the whole segment if the maxFreePageCount will be exceeded.
When we don't have a lot of page used within a page allocator, we waste about ~100K per page allocator,
because we commit the whole segment (128K), and only use a couple of page. So in general, we wastes
about 500K-1M per thread depending on how many script context is in the thread context.

Refactor the code that manages HeapPageAllocators in CustomHeap into CodePageAllocator.
Rationalize the locking and IsNativeAddress checks
Reimplement IsInRange as IsInHeap to work the heap list instead of the page allocator, instead of
relying on the page allocator, which is now shared.

### DIFF
--- a/lib/Backend/CodeGenAllocators.cpp
+++ b/lib/Backend/CodeGenAllocators.cpp
@@ -10,18 +10,21 @@ CodeGenAllocators::CodeGenAllocators(AllocationPolicyManager * policyManager, Js
                  PageAllocator::DefaultLowMaxFreePageCount :
                  PageAllocator::DefaultMaxFreePageCount))
 , allocator(L"NativeCode", &pageAllocator, Js::Throw::OutOfMemory)
-, emitBufferManager(policyManager, &allocator, scriptContext, L"JIT code buffer", ALLOC_XDATA)
+, emitBufferManager(&allocator, scriptContext->GetThreadContext()->GetCodePageAllocators(), scriptContext, L"JIT code buffer")
 #if !_M_X64_OR_ARM64 && _CONTROL_FLOW_GUARD
 , canCreatePreReservedSegment(false)
-#endif
-#ifdef PERF_COUNTERS
-, staticNativeCodeData(0)
 #endif
 {
 }
 
 CodeGenAllocators::~CodeGenAllocators()
 {
-    PERF_COUNTER_SUB(Code, StaticNativeCodeDataSize, staticNativeCodeData);
-    PERF_COUNTER_SUB(Code, TotalNativeCodeDataSize, staticNativeCodeData);
 }
+
+#if DBG
+void
+CodeGenAllocators::ClearConcurrentThreadId()
+{    
+    this->pageAllocator.ClearConcurrentThreadId();
+}
+#endif
--- a/lib/Common/Memory/CustomHeap.cpp
+++ b/lib/Common/Memory/CustomHeap.cpp
@@ -20,21 +20,18 @@ namespace CustomHeap
 
 #pragma region "Constructor and Destructor"
 
-Heap::Heap(AllocationPolicyManager * policyManager, ArenaAllocator * alloc, bool allocXdata):
+Heap::Heap(ArenaAllocator * alloc, CodePageAllocators * codePageAllocators):
     auxiliaryAllocator(alloc),
-    allocXdata(allocXdata),
+    codePageAllocators(codePageAllocators)
 #if DBG_DUMP
-    freeObjectSize(0),
-    totalAllocationSize(0),
-    allocationsSinceLastCompact(0),
-    freesSinceLastCompact(0),
+    , freeObjectSize(0)
+    , totalAllocationSize(0)
+    , allocationsSinceLastCompact(0)
+    , freesSinceLastCompact(0)
 #endif
 #if DBG
-    inDtor(false),
+    , inDtor(false)
 #endif
-    pageAllocator(policyManager, allocXdata, true /*excludeGuardPages*/),
-    preReservedHeapPageAllocator(policyManager, allocXdata, true /*excludeGuardPages*/),
-    cs(4000)
 {
     for (int i = 0; i < NumBuckets; i++)
     {
@@ -55,6 +52,7 @@ Heap::~Heap()
 #pragma region "Public routines"
 void Heap::FreeAll()
 {
+    CodePageAllocators::AutoLock autoLock(this->codePageAllocators);
     FreeBuckets(false);
 
     FreeLargeObjects();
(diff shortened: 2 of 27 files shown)


curtisman:
Code review please @ianwjhalliday @satheeshravi.

lib/Common/Memory/VirtualAllocWrapper.cpp
```diff
@@ -158,10 +136,81 @@ PreReservedVirtualAllocWrapper::GetPreReservedStartAddress()
...
 {
-    Assert(this);
+    Assert(this && IsPreReservedRegionPresent());
     return (char*)preReservedStartAddress + (PreReservedAllocationSegmentCount * AutoSystemInfo::Data.GetAllocationGranularityPageCount() * AutoSystemInfo::PageSize);
 }
 
+LPVOID PreReservedVirtualAllocWrapper::EnsurePreReservedRegion()
+{
+    LPVOID startAddress = preReservedStartAddress;
+    if (startAddress != NULL)
```
ianwjhalliday: nullptr?

lib/Common/Memory/CustomHeap.h
```diff
@@ -131,63 +131,100 @@ struct Allocation
...
+    public:
+        AutoLock(CodePageAllocators * codePageAllocators) : AutoCriticalSection(&codePageAllocators->cs) {};
+    };
 
-    template<typename T>
-    HeapPageAllocator<T>* GetPageAllocator(Page * page)
+    CodePageAllocators(AllocationPolicyManager * policyManager, bool allocXdata, PreReservedVirtualAllocWrapper * virtualAllocator) :
+        pageAllocator(policyManager, allocXdata, true /*excludeGuardPages*/, nullptr),
+        preReservedHeapPageAllocator(policyManager, allocXdata, true /*excludeGuardPages*/, virtualAllocator),
+        cs(4000)
```
ianwjhalliday: Where does this spin count value come from?  Arbitrary?  Is it a reasonable count?

lib/Common/Memory/CustomHeap.h
```diff
@@ -291,14 +337,67 @@ class Heap
...
+        AssertMsg(segment, "Why is segment null?");
+        Assert((HeapPageAllocator<PreReservedVirtualAllocWrapper>*)(segment->GetAllocator()) == &this->preReservedHeapPageAllocator);
+        return (HeapPageAllocator<PreReservedVirtualAllocWrapper> *)(segment->GetAllocator());
+    }
+
+    HeapPageAllocator<VirtualAllocWrapper>               pageAllocator;
+    HeapPageAllocator<PreReservedVirtualAllocWrapper>    preReservedHeapPageAllocator;
+    CriticalSection cs;
+};
+/*
```
ianwjhalliday: nit: blank line before the comment

lib/Common/Memory/CustomHeap.h
dilijev: I'd also prefer to define a macro instead of having a magic number inline.

lib/Backend/CodeGenAllocators.cpp
```diff
@@ -10,18 +10,21 @@ CodeGenAllocators::CodeGenAllocators(AllocationPolicyManager * policyManager, Js
...
-    PERF_COUNTER_SUB(Code, TotalNativeCodeDataSize, staticNativeCodeData);
 }
+
+#if DBG
+void
+CodeGenAllocators::ClearConcurrentThreadId()
+{    
+    this->pageAllocator.ClearConcurrentThreadId();
+}
+#endif
```
dilijev: nit: blank line at end of file

lib/Common/Memory/CustomHeap.h
ianwjhalliday: I don't mind the magic number inline so long as its value is well defended

lib/Common/Memory/CustomHeap.cpp
```diff
@@ -169,11 +167,50 @@ bool Heap::Decommit(__in Allocation* object)
...
 
-bool Heap::IsInRange(__in void* address)
+bool Heap::IsInHeap(DListBase<Page> const& bucket, __in void * address)
 {
-    AutoCriticalSection autocs(&this->cs);
+    DListBase<Page>::Iterator i(&bucket);
+    while (i.Next())
+    {
+        Page& page = i.Data();
+        if (page.address >= address || address < page.address + AutoSystemInfo::PageSize)
```
satheeshravi: This will return true for address <= page.address  - which is out of range ?

ianwjhalliday:
A few nitpicks, otherwise LGTM

lib/Common/Memory/CustomHeap.cpp
ianwjhalliday: Should be &&?

satheeshravi: {AncestorPath = 53674661}
also it should be  address >= page.address &&

lib/Common/Memory/CustomHeap.cpp
```diff
@@ -169,11 +167,50 @@ bool Heap::Decommit(__in Allocation* object)
...
+    return false;
+}
+
+bool Heap::IsInHeap(DListBase<Allocation> const& allocations, __in void *address)
+{
+    DListBase<Allocation>::Iterator i(&allocations);
+    while (i.Next())
+    {
+        Allocation& allocation = i.Data();
+        if (allocation.address >= address || address < allocation.address + allocation.size)
```
satheeshravi: same here ?

lib/Common/Memory/CustomHeap.cpp
curtisman: Yep, @satheeshravi.   Will fix.

lib/Common/Memory/CustomHeap.h
curtisman: Long, long time ago, I read somewhere that 4000 is a good number.
Did a bit of a search again and the MSDN doc for InitializeCriticalSectionAndSpinCount give an example where the OS heap uses 4000.  I guess that what I get the value from.
https://msdn.microsoft.com/en-us/library/windows/desktop/ms683476(v=vs.85).aspx.

There are other places where we use 4000 as the count, and that's where I got the number too.
Since this look should be very contentious, this value probably doesn't matter that much.

ianwjhalliday: I see.  I am fine with this.

lib/Common/Memory/CustomHeap.h
```diff
@@ -131,63 +131,100 @@ struct Allocation
...
-    HeapPageAllocator<PreReservedVirtualAllocWrapper> * GetPreReservedHeapPageAllocator()
+    bool IsInNonPreReservedPageAllocator(__in void *address)
     {
-        return &preReservedHeapPageAllocator;
+        Assert(this->cs.IsLocked());
+        return this->pageAllocator.IsAddressFromAllocator(address);
     }
 
-    HeapPageAllocator<VirtualAllocWrapper>* GetHeapPageAllocator()
+    char * Alloc(size_t * pages, void ** segment, bool canAllocInPreReservedHeapPageSegment, bool isAnyJittedCode, bool * isAllJITCodeInPreReservedRegion)
```
satheeshravi: nit: Is it ok to have definitions here ? Do you want to move to cpp ?

PR #369, opened 2016-02-22
a few '/bin' file name case changes

jianchun:
Change a few '/bin' file names to consistent CamelCase and update
references.

Update a few includes in lib/Common/Common.h (which my previous regex
match missed).


jianchun:
Not sure on "stdafx" vs. "StdAfx". Picked "stdafx" because most of our code uses it and a new VS2015 project uses it. @dilijev @digitalinfinity

dilijev:
LGTM

dilijev:
Agree on stdafx in all lowercase. It's alphabet soup anyway, not sure PascalCase would make a difference.

jianchun:
@dilijev Thank you!

PR #370, opened 2016-02-22
Fix auxPtrs using in background thread

leirocks:
Fix auxPtrs using in background thread

There's race problem when accessing auxPtrs from background thread while
auxPtrs is expanding in foreground thread to hold more pointers.
Adding a lock to sync between background access and foreground setter.

### DIFF
--- a/lib/Backend/CodeGenWorkItem.h
+++ b/lib/Backend/CodeGenWorkItem.h
@@ -333,7 +333,7 @@ struct JsLoopBodyCodeGen sealed : public CodeGenWorkItem
 
     uint GetLoopNumber() const override
     {
-        return functionBody->GetLoopNumber(loopHeader);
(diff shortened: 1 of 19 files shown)


leirocks:
@curtisman OK, ThreadContext::GetContextForCurrentThread uses tlsSlot, when it is called from background thread it might return null since we don't set the TLS slots for background thread right? but here for assertion purpose only I think it should be enough to tell if it is from background thread.

curtisman:
@leirocks.  I think the only case that a ThreadContext would be deleted on thread that doesn't have the thread in the TLS is ThreadBoundThreadContextManager::DestroyAllContextsAndEntries().  In that case, I think you should add a debug Boolean, to track whether you are in DestroyAllContextsAndEntries or not?

dilijev:
@dotnet-bot test this please

PR #371, opened 2016-02-22
Fix crawler assert

pleath:
Avoid deferring all code passed to ParseProcedureText in debug/rundown mode. (I was only avoiding deferral for code with event handler semantics.)


pleath:
@Yongqu, here's the one we were discussing.

Yongqu:
LGTM.

PR #372, opened 2016-02-22
Determine debug mode based on per script/source instead of ScriptContext

agarwal-sandeep:
Check if Utf8SourceInfo is in debug mode or not instead of ScriptContext
Utf8SourceInfo constructor now takes isLibraryCode flag instead of calling SetIsLibraryCode as it is always known where a source is library code or not while creating Utf8SourceInfo
Renamed all IsInDebug*Mode functions to Is<ClassName>InDebug*Mode so that it is explicit what are we checking and to identify all places this change applies
Remove isByteCodeDebugMode from FunctionBody

[Review on Reviewable](https://reviewable.io/reviews/microsoft/chakracore/372)


agarwal-sandeep:
@akroshg please take a look

akroshg:
Awesome to see this issue is taken care of.
I am going to take a look at the code change.

lib/Backend/BailOut.cpp
```diff
@@ -585,7 +585,7 @@ BailOutRecord::RestoreValues(IR::BailOutKind bailOutKind, Js::JavascriptCallStac
         }
 #endif
 
-        if (scriptContext->IsInDebugMode())
+        if (functionBody->GetUtf8SourceInfo()->IsUtf8SourceInfoInDebugMode())
```
akroshg: I understand that you have added the classname in the Is<classname>InDebugMode to make your refactor easier. But now it does not make sense to have this redundant information. What do you think?

agarwal-sandeep: I would still like to keep it for clarity sake if there are no strong objection.
Makes it explicit when searching for function calls. If all functions are named IsInDebugMode search shows all instance and then need to look at all to find which one applies.

lib/Backend/BailOut.cpp
```diff
@@ -1215,7 +1215,7 @@ BailOutRecord::BailOutHelper(Js::JavascriptCallStackLayout * layout, Js::ScriptF
     // Clear the disable implicit call bit in case we bail from that region
     functionScriptContext->GetThreadContext()->ClearDisableImplicitFlags();
 
-    bool isInDebugMode = functionScriptContext->IsInDebugMode();
+    bool isInDebugMode = executeFunction->GetUtf8SourceInfo()->IsUtf8SourceInfoInDebugMode();
     AssertMsg(!isInDebugMode || Js::Configuration::Global.EnableJitInDebugMode(),
```
akroshg: nit: you could add IsInDebugMode to FunctionBody (which is essentially doing redirection to Utf8SourceInfo) that way it is easier to read in many places.

lib/Backend/Func.cpp
```diff
@@ -978,7 +978,7 @@ bool Func::CanAllocInPreReservedHeapPageSegment ()
 {
 #ifdef _CONTROL_FLOW_GUARD
     return PHASE_FORCE1(Js::PreReservedHeapAllocPhase) || (!PHASE_OFF1(Js::PreReservedHeapAllocPhase) &&
-        !IsJitInDebugMode() && !GetScriptContext()->IsInDebugMode() && GetScriptContext()->GetThreadContext()->IsCFGEnabled()
+        !IsJitInDebugMode() && !this->m_workItem->GetFunctionBody()->GetUtf8SourceInfo()->IsUtf8SourceInfoInDebugMode() && GetScriptContext()->GetThreadContext()->IsCFGEnabled()
```
akroshg: what is this IsJitInDebugMode? does it require to be relooked?

lib/Parser/Parse.cpp
```diff
@@ -10369,7 +10369,7 @@ ParseNodePtr Parser::Parse(LPCUTF8 pszSrc, size_t offset, size_t length, charcou
     // Scanner should run in Running mode and not syntax coloring mode
     grfscr &= ~fscrSyntaxColor;
 
-    if (this->m_scriptContext->IsInDebugMode() || PHASE_OFF1(Js::Phase::DeferParsePhase)
+    if (this->m_scriptContext->IsScriptContextInDebugMode() || PHASE_OFF1(Js::Phase::DeferParsePhase)
```
akroshg: shouldn't we do this only when grfscr is not fscrIsLibraryCode? Essentially we are not deferring any library code when we are debugging.

lib/Backend/Func.cpp
agarwal-sandeep: When a CodeGenWorkItem is created we set if it is created in debug mode or not, See NativeCodeGenerator::NewFunctionCodeGen or NativeCodeGenerator::NewLoopBodyCodeGen.

lib/Parser/Parse.cpp
```diff
@@ -10682,7 +10682,7 @@ bool Parser::CheckAsmjsModeStrPid(IdentPtr pid)
         !m_pscan->IsEscapeOnLastTkStrCon() &&
         wcsncmp(pid->Psz(), L"use asm", 10) == 0);
 
-    if (isAsmCandidate && m_scriptContext->IsInDebugMode())
+    if (isAsmCandidate && m_scriptContext->IsScriptContextInDebugMode())
```
akroshg: So if we have asmjs code in the library file we will emit this error to the debugger. (we were doing it before your change as well). In the future we should not raise this message when the current code is library code - for the user it does not make sense.

lib/Runtime/Base/Utf8SourceInfo.cpp
```diff
@@ -199,13 +200,15 @@ namespace Js
...
         Utf8SourceInfo* newSourceInfo = Utf8SourceInfo::NewWithHolder(scriptContext, sourceInfo->GetSourceHolder()->Clone(scriptContext), sourceInfo->m_cchLength,
-             SRCINFO::Copy(scriptContext->GetRecycler(), sourceInfo->GetSrcInfo()));
+             SRCINFO::Copy(scriptContext->GetRecycler(), sourceInfo->GetSrcInfo()), sourceInfo->m_isLibraryCode);
         newSourceInfo->m_isXDomain = sourceInfo->m_isXDomain;
         newSourceInfo->m_isXDomainString = sourceInfo->m_isXDomainString;
         newSourceInfo->m_isLibraryCode = sourceInfo->m_isLibraryCode;
         newSourceInfo->SetIsCesu8(sourceInfo->GetIsCesu8());
         newSourceInfo->m_lineOffsetCache = sourceInfo->m_lineOffsetCache;
-        if (scriptContext->IsInDebugMode())
+        newSourceInfo->m_isInDebugMode = sourceInfo->m_isInDebugMode;
```
akroshg: this is not needed as the code below (setIndebugMode) will do that for you.

lib/Runtime/ByteCode/ByteCodeDumper.cpp
```diff
@@ -27,12 +27,6 @@ namespace Js
     }
     void ByteCodeDumper::Dump(FunctionBody* dumpFunction)
     {
-        if (!CONFIG_FLAG(DumpDbgControllerBytecode) && dumpFunction->GetSourceContextInfo() &&
```
akroshg: I didn't know there was special handling for test file. Good to get rid of that.

lib/Runtime/ByteCode/ByteCodeGenerator.cpp
```diff
@@ -1079,7 +1079,7 @@ FuncInfo * ByteCodeGenerator::StartBindGlobalStatements(ParseNode *pnode)
         byteCodeFunction->ResetByteCodeGenVisitState();
         if (byteCodeFunction->GetBoundPropertyRecords() == nullptr)
         {
-            Assert(!IsInNonDebugMode());
+            Assert(!IsByteCodeGeneratorInNonDebugMode());
```
akroshg: Nit: useless assert - there is an if condition above which checks the same condition.

lib/Runtime/Debug/ProbeContainer.cpp
```diff
@@ -110,7 +110,7 @@ namespace Js
                 ScriptContext* frameScriptContext = walker.GetCurrentScriptContext();
                 Assert(frameScriptContext);
 
-                if (!fMatchWithCurrentScriptContext && !frameScriptContext->IsInDebugMode() && tempFramePointers->Count() == 0)
+                if (!fMatchWithCurrentScriptContext && !frameScriptContext->IsScriptContextInDebugMode() && tempFramePointers->Count() == 0)
```
akroshg: Github does not let me comment for a line 106

will this (isLibraryFrameEnabledDebugger) ever be true? If this gets to true - I doubt that the debugger could handle the library frame in between.
Wondering if this can be removed in the future.

lib/Runtime/Language/InterpreterStackFrame.cpp
```diff
@@ -3603,7 +3603,7 @@ namespace Js
         Js::ImplicitCallFlags savedImplicitCallFlags = threadContext->GetImplicitCallFlags();
 
 #if DBG
-        if (scriptContext->IsInDebugMode())
+        if (scriptContext->IsScriptContextInDebugMode())
```
akroshg: shouldn't this be checking the ut8sourceinfo being debug mode?

lib/Runtime/Library/ThrowErrorObject.cpp
```diff
@@ -18,7 +18,7 @@ namespace Js
         ThrowErrorObject* throwErrorObject = ThrowErrorObject::FromVar(function);
 
         bool useExceptionWrapper =
-            scriptContext->IsInDebugMode() &&
+            scriptContext->IsScriptContextInDebugMode() &&
```
akroshg: Add a comment that it is intentional, the library code is also using the exception wrapper.

lib/Backend/BailOut.cpp
akroshg: Right now it seems fine as you want clarity on your refactoring. But it will get over-explicit in the future. We could leave the ScriptContextInDebugMode explicitly rest can be converted to IsInDebugMode as they are essentially redirection to the utf8sourceinfo.

lib/Backend/Func.cpp
akroshg: should we get rid of that?

lib/Backend/BailOut.cpp
agarwal-sandeep: Kept ScriptContext name and removed others

lib/Backend/BailOut.cpp
agarwal-sandeep: Good one, did thanks

lib/Parser/Parse.cpp
agarwal-sandeep: Kept a note, will open a new work item and do it separately

lib/Parser/Parse.cpp
agarwal-sandeep: Will do a separate work item

lib/Runtime/Base/Utf8SourceInfo.cpp
agarwal-sandeep: Fixed

lib/Runtime/Debug/ProbeContainer.cpp
agarwal-sandeep: Will do as a separate work item, took a note

akroshg:
Looks good to me.

dilijev:
@dotnet-bot test this please

agarwal-sandeep:
@dotnet-bot test this please

PR #376, opened 2016-02-23
Implement ES6 RegExp.prototype[@@match]

goyakin:


goyakin:
@abchatra Could you take a look?

test/UnitTestFramework/UnitTestFramework.js
```diff
@@ -81,7 +81,17 @@ var helpers = function helpers() {
             for (name in o) {
                 this.writeln(name, o.hasOwnProperty(name) ? "" : " (inherited)", ": ", o[name]);
             }
-        }
+        },
+
+        withPropertyDeleted: function withPropertyDeleted(object, propertyName, callback) {
```
akroshg: I saw this is a convention in this file, but why give it a name again when we have a name before column? may be this was written before Function.name implemented.

goyakin: I actually don't know. According to the [spec](https://tc39.github.io/ecma262/#sec-object-initializer-runtime-semantics-propertydefinitionevaluation), the `name` property would return the property name before the colon without the name in the function definition, and we implement this correctly. I'll see if removing them causes any test failures.

ljharb: If your function name inference is working properly then the `name` will still work. However, it's probably more robust to keep it explicitly for now.

lib/Runtime/Library/JavascriptRegularExpression.cpp
```diff
@@ -121,7 +136,18 @@ namespace Js
...
+        JavascriptError::ThrowTypeError(scriptContext, JSERR_This_NeedRegExp, varName);
+    }
+
+
+    RecyclableObject* JavascriptRegExp::GetThisObject(Arguments& args, PCWSTR varName, ScriptContext* scriptContext)
+    {
+        if (args.Info.Count == 0 || !JavascriptOperators::IsObject(args[0]))
+        {
+            JavascriptError::ThrowTypeError(scriptContext, JSERR_This_NeedObject, varName);;
+        }
```
akroshg: nit: extra semicolon at the end.

lib/Runtime/Library/JavascriptRegularExpression.cpp
```diff
@@ -121,7 +136,18 @@ namespace Js
...
         }
 
-        JavascriptError::ThrowTypeError(scriptContext, JSERR_This_NeedRegExp, propertyName);
+        JavascriptError::ThrowTypeError(scriptContext, JSERR_This_NeedRegExp, varName);
+    }
+
+
+    RecyclableObject* JavascriptRegExp::GetThisObject(Arguments& args, PCWSTR varName, ScriptContext* scriptContext)
+    {
+        if (args.Info.Count == 0 || !JavascriptOperators::IsObject(args[0]))
```
akroshg: can args.info.Count be 0? many places in this file we have that as assert.

abchatra: args.info.Count will be atleast 1 for this pointer

lib/Runtime/Library/JavascriptRegularExpression.cpp
```diff
@@ -635,21 +798,63 @@ namespace Js
...
+        }
 
-            return (JavascriptOperators::GetTypeId(result) == TypeIds_Null)
-                ? TaggedInt::ToVarUnchecked(-1)
-                : JavascriptOperators::GetProperty(RecyclableObject::FromVar(result), PropertyIds::index, scriptContext);
+        UnifiedRegex::RegexPattern* pattern = this->GetPattern();
+
+        UnifiedRegex::RegexFlags newFlags = pattern->GetFlags();
+        newFlags = SetRegexFlag(PropertyIds::global, newFlags, UnifiedRegex::GlobalRegexFlag, scriptContext);
+        newFlags = SetRegexFlag(PropertyIds::sticky, newFlags, UnifiedRegex::StickyRegexFlag, scriptContext);
```
akroshg: what if getting sticky flag changes the global flag? is it even possible?

lib/Runtime/Library/RegexHelper.cpp
```diff
@@ -347,9 +347,109 @@ namespace Js
         UnifiedRegex::Matcher* matcher;
     };
 
+    template <bool updateHistory>
+    Var RegexHelper::RegexMatchImpl(ScriptContext* scriptContext, RecyclableObject *thisObj, JavascriptString *input, bool noResult, void *const stackAllocationPointer)
+    {
+        ScriptConfiguration const * scriptConfig = scriptContext->GetConfig();
```
akroshg: nit: why not create local for just IsES6RegExSymbolsEnabled()? I see scriptConfig is used for only that purpose.

test/es6/regex-symbols.js
```diff
@@ -396,8 +425,184 @@ var tests = [
...
         body: verifySymbolMethodLength.bind(undefined, 1, Symbol.match)
     },
+    {
+        name: "RegExp.prototype[@@match] should run the built-in 'match' when none of the observable properties are overridden",
+        body: function () {
+            var pattern = "(a)-";
+
+            var nonGlobalRe = new RegExp(pattern);
+            var nonGlobalInput = "-a-a-";
+            var result = nonGlobalRe[Symbol.match](nonGlobalInput);
```
akroshg: what is nonGlobalRe[Symbol.match].name ?

lib/Runtime/Library/RegexHelper.cpp
```diff
@@ -1718,4 +1818,12 @@ namespace Js
         Var result = RegexHelper::RegexSplitImpl(entryFunctionContext, regularExpression, input, limit, noResult, stackAllocationPointer);
         return RegexHelper::CheckCrossContextAndMarshalResult(result, entryFunctionContext);
     }
+
+    int64 RegexHelper::AdvanceStringIndex(JavascriptString* string, int64 index, bool isUnicode)
```
akroshg: why string is even passed?

test/es6/regex-symbols.js
goyakin: `[Symbol.match]`. Why do you ask?

akroshg:
Overall code looks good to me.

test/es6/regex-symbols.js
akroshg: wondered if there is test for that?

lib/Runtime/Library/JavascriptRegularExpression.cpp
goyakin: It is possible. According to [`RegExpBuiltinExec`](https://tc39.github.io/ecma262/#sec-regexpbuiltinexec), we're supposed to `Get` `global` before `sticky`, so it won't take affect until `global` is used again.

lib/Runtime/Library/RegexHelper.cpp
goyakin: `scriptConfig` is also used in the `else` branch though.

test/es6/regex-symbols.js
goyakin: There is. In `regex-symbols.js`, `verifySymbolMethodName` gets called for `RegExp.prototype[@@match]` to verify the name.

lib/Runtime/Library/RegexHelper.cpp
goyakin: It'll be used to determine the size of the code point at `index`. I'm having all callers pass it so that  we'll only change this function when needed.

Just curious... Would the VS C++ compiler put an unused argument on the stack or would it actually eliminate the argument? @abchatra, do you know?

lib/Runtime/Library/JavascriptRegularExpression.cpp
goyakin: Looks like it would be at least 1, but it isn't only here we compare against `0`. I'll keep the check for now, but I've created #387 to track this.

goyakin:
@dotnet-bot test this please

PR #377, opened 2016-02-23
linux: skeleton CMakeLists.txt files

jianchun:
Ran vcxproj files through a script and generated skeleton CMakeLists.txt
files for most libs. The generated files are not used in build yet.


bin/ChakraCore/CMakeLists.txt
```diff
@@ -0,0 +1,12 @@
+# xplat-todo: This is a skeleton make file and not used in build yet.
+# Please add this to build and fix issues.
+
+add_library (ChakraCore
```
digitalinfinity: This is not a library but a dynamic load dll- we could change it later, but just calling it out. Same for rl

digitalinfinity:
Just one comment- LGTM otherwise!

bin/ChakraCore/CMakeLists.txt
jianchun: > This is not a library but a dynamic load dll- we could change it later, but just calling it out. Same for rl

Yes, I saw that as well, thought the file list would still be useful so included them. Thanks!

Issue #27, opened 2016-01-07
Object destructuring issue: Initializer on shorthand should not throw syntax error


abchatra:
Following should parse
```javascript
var x = 10;
({x = 10} = {});
```
Actual: Syntax error

Refer OS bug [3887967] (https://microsoft.visualstudio.com/web/wi.aspx?pcguid=cb55739e-4afe-46a3-970f-1b49d8ee7564&id=3887967)

caitp:
these sorts of frontend fixes are probably good first bugs, if you want to attract volunteer contributors to work on actual language features.

I'm hoping I can encourage some local youth to tackle some of these bugs.

kepta:
Hey, I'm a newbie. 
Can I work on this bug and also get some help on how to tackle thing bug ?

abchatra:
@kepta absolutely. Tracing from the parser is the easiest thing to do. Add breakpoint at Parser::Parse. Also look at the call sites of IsES6DestructuringEnabled. That code amounts to destructuring related. 

See [this](https://github.com/Microsoft/ChakraCore/wiki/Building-ChakraCore) for building instructions. Let me know if you have more questions.

ianwjhalliday:
@kepta also take a read through [test/runtests.cmd](https://github.com/Microsoft/ChakraCore/blob/master/test/runtests.cmd) to learn how to run tests.  Of particular note, consider providing the options `-dirs es6 -variant interpreter` to reduce the number of tests run to those surrounding the new language features and running them against front end, so you can have quicker code, build, test, debug loops.

So for example, after building x64 debug in VS, cd to the test directory and execute `runtests.cmd -x64debug -dirs es6 -variant interpreter` to run tests appropriate for this bug.

Of course once you're satisfied with your change you'll want to run all tests to make sure nothing broke by accident (run all tests by only specifying the build flavor, e.g. `runtests.cmd -x64debug`).

akroshg:
This is fixed.

PR #379, opened 2016-02-23
Few destructuring fixes

akroshg:
1. Object pattern can now support the initializer on the short-hand. ({x = 1} = {}); Thing to note that object literal cannot allow the initializer to shorthand but that object literal can be object pattern so we need to defer the error until we are sure that the object literal is not the object pattern. Current we are giving the favor to object pattern and defer the error. If the object literal on the LHS, the error will be ignored. Look at test for more scenarios.
2. bug fix: rest pattern cannot have initializer. We have done that for rest identifier but the rest pattern was missing (as the rest pattern was itself the late addition to the spec).
3. The spec says that for object pattern assignment - the AssignmentExpression must be object coercible. So {} = undefined/null; should not be allowed (raise TypeError). Fixed that by adding the bytecodes which is essentially checking the assignment expression with undefined or null.

### DIFF
--- a/test/es6/destructuring.js
+++ b/test/es6/destructuring.js
@@ -66,8 +66,8 @@ var tests = [
       assert.throws(function () { eval("let a; [++a] = [];"); },     SyntaxError, "Destructured let array assignment with an operator throws",    "Unexpected operator in destructuring expression");
       assert.doesNotThrow(function () { eval("var a = [1], i = 0; [a[i++]] = [];"); }, "Destructured var array assignment operators inside an identifier reference does not throw");
       assert.doesNotThrow(function () { eval("let a = [1], i = 0; [a[i++]] = [];"); }, "Destructured var array assignment operators inside an identifier reference does not throw");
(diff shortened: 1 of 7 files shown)


akroshg:
Unittest clean.
@tcare - could you take a look at this please?

test/es6/destructuring.js
```diff
@@ -202,14 +202,6 @@ var tests = [
       assert.doesNotThrow(function () { eval("var a, b; [[(a)], ((((((([b])))))))] = [[],[]];"); }, "Destructured var array assignment with valid mixed paren and array nesting does not throw");
       assert.doesNotThrow(function () { eval("let a, b; [[(a)], ((((((([b])))))))] = [[],[]];"); }, "Destructured let array assignment with valid mixed paren and array nesting does not throw");
 
-      // All the things
```
akroshg: They are redundant - we already have similar tests in different files. Best removed them.

akroshg:
Fixes #27

akroshg:
which variable you are talking about? Is it mayAllowForPattern?

tcare:
Signing off

akroshg:
will make the change and wait till Jenkins is green. Thanks @tcare

dilijev:
@dotnet-bot test this please

PR #380, opened 2016-02-23
Add assert to catch browser issue

pleath:
Invalid instances of LdLocalObj are showing up in browser dumps. Add an assert to catch those instruction emissions at compile time.


ianwjhalliday:
:+1:

PR #381, opened 2016-02-23
HTML Comment Spec Compliance

### LINKED ISSUE #20: Support HTML-style comments
ES [spec](http://tc39.github.io/ecma262/#sec-html-like-comments) has detailed semantics of HTML style comments.  

Test case
```javascript
print((function(){
--> A comment
<!-- Another comment
var a = 3; <!-- Another comment
return a === 3;
})());
```
Expected: true
Actual: Syntax error

tcare:
Removed custom logic from HTML comment handling and brought up to date with B.1.3 HTML-like Comments.
- Removed HTML comment flag and replaced with a flag for checking if we are parsing a module (the only place we don't allow HTML comments in the standard.)
- Removed custom logic involving looking for EOF and } when parsing HTML end comment and jump to single line comment handling instead.
- Added line separator / paragraph separator handling to the main scanner loop to keep track of whether we saw an EOL.
- Added negative parsing cases to modules syntax testing

The unit test added for this commit is a special file and must be treated like binary to preserve the mixed and non-standard line endings. I added a line to .gitattributes to ensure this.

Fixes #20.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,3 +1,4 @@
 *.baseline -crlf
 *.cmd -crlf
 test/*.js -crlf
+test/es6/HTMLComments.js binary diff=cpp
--- a/lib/Parser/Scan.cpp
+++ b/lib/Parser/Scan.cpp
@@ -202,7 +202,7 @@ void Scanner<EncodingPolicy>::SetText(EncodedCharPtr pszSrc, size_t offset, size
     m_startLine = lineNumber;
     m_pchStartLine = m_currentCharacter;
     m_ptoken->tk = tkNone;
-    m_fHtmlComments = (grfscr & fscrHtmlComments) != 0;
+    m_fIsModuleCode = (grfscr & fscrIsModuleCode) != 0;
     m_fHadEol = FALSE;
     m_fSyntaxColor = (grfscr & fscrSyntaxColor) != 0;
     m_DeferredParseFlags = ScanFlagNone;
@@ -1614,6 +1614,8 @@ tokens Scanner<EncodingPolicy>::SkipComment(EncodedCharPtr *pp, /* out */ bool* 
                 return tkNone;
             }
             break;
+
+        // ES 2015 11.3 Line Terminators
         case kchLS:         // 0x2028, classifies as new line
         case kchPS:         // 0x2029, classifies as new line
 LEcmaLineBreak:
@@ -1756,6 +1758,7 @@ tokens Scanner<EncodingPolicy>::ScanCore(bool identifyKwds)
     m_fHadEol = FALSE;
     CharTypes chType;
     charcount_t commentStartLine;
+    bool seenDelimitedCommentEnd = false;
 
     if (m_scanState && *p != 0)
     {
@@ -1934,16 +1937,19 @@ LEof:
             }
         case '(': Assert(chType == _C_LPR); token = tkLParen; break;
         case ')': Assert(chType == _C_RPR); token = tkRParen; break;
-        case ',': Assert(chType == _C_CMA); token = tkComma; break;
+        case ',': Assert(chType == _C_CMA); token = tkComma;  break;
         case ';': Assert(chType == _C_SMC); token = tkSColon; break;
         case '[': Assert(chType == _C_LBR); token = tkLBrack; break;
         case ']': Assert(chType == _C_RBR); token = tkRBrack; break;
-        case '~': Assert(chType == _C_TIL); token = tkTilde; break;
-        case '?': Assert(chType == _C_QUE); token = tkQMark; break;
-        case '{': Assert(chType == _C_LC); token = tkLCurly; break;
+        case '~': Assert(chType == _C_TIL); token = tkTilde;  break;
+        case '?': Assert(chType == _C_QUE); token = tkQMark;  break;
+        case '{': Assert(chType == _C_LC);  token = tkLCurly; break;
 
+        // ES 2015 11.3 Line Terminators
         case '\r':
         case '\n':
+        // kchLS:
+        // kchPS:
 LNewLine:
             m_currentCharacter = p;
             ScanNewLine(ch);
@@ -2087,36 +2093,11 @@ LIdentifier:
             case '-':
                 p++;
                 token = tkDec;
-                if (m_fHtmlComments)
+                if (!m_fIsModuleCode)
                 {
-                    int i = 0;
-                    while ('-' == PeekFirst(p + i, last)) //Have already seen --, skip any further - characters
-                        i++;
-                    if ('>' == PeekFirst(p + i++, last)) //This means we've got a --------------------------->.
+                    if ('>' == PeekFirst(p, last) && (m_fHadEol || seenDelimitedCommentEnd)) // --> HTMLCloseComment
                     {
-                        //If that precedes an EOF or }NWL (disregarding whitespace), then it is a comment.
-                        OLECHAR nextChar;
-                        nextChar = NextNonWhiteChar(&p[i], last);
-                        if (nextChar == 0)
-                        {
-                            //Treat the -----------------------------> EOF as if it were EOF
-                            token = tkEOF;
-                            ++p;
-                        }
-                        else if (nextChar == '}')
-                        {
-                            CharTypes nextNextCharType = this->charClassifier->GetCharType(NextNonWhiteCharPlusOne(&p[i], last));
-                            if (nextNextCharType == _C_NWL
-                                // Corner case: If we have reached the end of the source, either we are at the end of the file or the end of
-                                // a deferred function. We treat this case as NWL.
-                                // TODO(tcare): Update to ES6 spec. Tracked in Bug 1164686
-                                || (last == m_pchLast && nextNextCharType == _C_NUL))
-                            {
-                                //Treat the -----------------------------> }NWL as if it were }NWL
-                                p += i;
-                                continue;
-                            }
-                        }
+                        goto LSkipLineComment;
                     }
                 }
                 break;
@@ -2155,7 +2136,7 @@ LIdentifier:
             case '/':
                 if (p >= last)
                 {
-                    AssertMsg(m_fHtmlComments, "Do we have other line comment cases scanning pass last?");
+                    AssertMsg(!m_fIsModuleCode, "Do we have other line comment cases scanning pass last?");
 
                     // Effective source length may have excluded HTMLCommentSuffix "//... -->". If we are scanning
                     // those, we have passed "last" already. Move back and return EOF.
@@ -2251,6 +2232,7 @@ LMultiLineComment:
                     // of deciding whether to defer AST and byte code generation.
                     m_parser->ReduceDeferredScriptLength((ULONG)(pchT - m_pchMinTok));
                     p = pchT;
+                    seenDelimitedCommentEnd = true;
                     goto LLoop;
                 }
                 p = pchT;
@@ -2286,7 +2268,8 @@ LMultiLineComment:
                 }
                 break;
             case '!':
-                if (m_fHtmlComments && PeekFirst(p + 1, last) == '-' && PeekFirst(p + 2, last) == '-')
+                // ES 2015 B.1.3 -  HTML comments are only allowed when parsing non-module code.
+                if (!m_fIsModuleCode && PeekFirst(p + 1, last) == '-' && PeekFirst(p + 2, last) == '-')
                 {
                     // This is a "<!--" comment - treat as //
                     if (p >= last)
--- a/test/es6/HTMLComments.baseline
+++ b/test/es6/HTMLComments.baseline
@@ -1,0 +1,13 @@
+Code before CRLF--> is reachable
+Code before CR--> is reachable
+Code before LF--> is reachable
+Code before LS--> is reachable
+Code before PS--> is reachable
+Code before CRLS--> is reachable
+Code before CRPS--> is reachable
+Code before <!-- is reachable
+Code before <!-- --> is reachable
+Code before <!-- LineTerminator --> is reachable
+Code before /* */ --> is reachable
+Code before /* */--> is reachable
+Code after post-decrement with a greater-than comparison (-->) is reachable
--- a/test/es6/HTMLComments.js
+++ b/test/es6/HTMLComments.js
@@ -1,0 +1,66 @@
+﻿//-------------------------------------------------------------------------------------------------------
+// Copyright (C) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE.txt file in the project root for full license information.
+//-------------------------------------------------------------------------------------------------------
+
+/* NOTE: This file needs to be treated as binary. It contains mixed line endings, including non-standard
+ *       line endings. Most text editors will not handle the file correctly. If you need to edit this
+ *       file, make sure you do a binary compare to ensure the non-standard line endings have not been lost.
+ *
+ *       'LS' refers to Unicode Character 'LINE SEPARATOR' (U+2028)
+ *       'PS' refers to Unicode Character 'PARAGRAPH SEPARATOR' (U+2029)
+ */
+
+WScript.LoadScriptFile("..\\UnitTestFramework\\UnitTestFramework.js");
+
+/*
+ * Line terminator sequences - standard (11.3 LineTerminator)
+ */
+ 
+// CRLF
+WScript.Echo("Code before CRLF--> is reachable");
+--> WScript.Echo("Code after CRLF--> is unreachable");
+
+// CR
+WScript.Echo("Code before CR--> is reachable");--> WScript.Echo("Code after CR--> is unreachable");
+
+// LF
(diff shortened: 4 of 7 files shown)


tcare:
FYI you probably want to look at the test file in a hex editor or a text editor that can show placeholders for the non-standard line endings/whitespace.

Could you please take a look @abchatra @ianwjhalliday @boingoing @pleath

.gitattributes
```diff
@@ -1,3 +1,4 @@
 *.baseline -crlf
 *.cmd -crlf
 test/*.js -crlf
+test/es6/HTMLComments.js binary diff=cpp
```
dilijev: Wow very specific.

nit: EOL at end of file please

tcare: done

test/es6/HTMLComments.js
```diff
@@ -0,0 +1,72 @@
...
+ *       'PS' refers to Unicode Character 'PARAGRAPH SEPARATOR' (U+2029)
+ */
+
+/*
+ * Line terminator sequences - standard (11.3 LineTerminator)
+ */
+
+// CRLF
+WScript.Echo("Code before CRLF<-- is reachable");
+--> WScript.Echo("Code after CRLF<-- is unreachable");
```
ianwjhalliday: --> is used in the source but the description strings all say <--

Typo?

test/es6/HTMLComments.js
```diff
@@ -0,0 +1,72 @@
...
+WScript.Echo("CRLF<-- does not affect automatic semicolon insertion")
```
ianwjhalliday: Is this a sufficient test to prove that semicolon insertion is not affected?  Is it possible to use a grammar production that relies on [no LineTerminator here] to prove that automatic semicolon insertion is doing the right thing?

test/es6/HTMLComments.js
tcare: Oops bad string replace. Fixed.

test/es6/HTMLComments.js
```diff
@@ -0,0 +1,72 @@
...
+WScript.Echo("Code before <!-- LineTerminator --> is reachable"); <!-- WScript.Echo("Code after <!-- is unreachable");
+--> WScript.Echo("Code after <!-- LineTerminator --> is unreachable");
+
+
+// Multiline comments syntax
+/* Multi
+   Line
+   Comment */ --> WScript.Echo("Code after */ --> is unreachable");
+   
+
```
ianwjhalliday: Would be nice if the description for a-->a made it obvious that the --> is post-decrement greater than and not an html comment close syntax.

For some reason github thinks the last 6 lines of this file aren't new so I'm putting this comment up here, but I'm referring to line 78.

test/es6/HTMLComments.js
tcare: Sure, I can use something from 11.9.1Rules of Automatic Semicolon Insertion

test/es6/module-syntax.js
```diff
@@ -9,21 +9,21 @@ WScript.LoadScriptFile("..\\UnitTestFramework\\UnitTestFramework.js");
...
         // We can't use assert.throws here because the SyntaxError used to construct the thrown error
         // is from a different context so it won't be strictly equal to our SyntaxError.
         try {
             testfunc();
         } catch(e) {
             caught = true;
-            
+
             // Compare toString output of SyntaxError and other context SyntaxError constructor.
             assert.areEqual(e.constructor.toString(), SyntaxError.toString(), message);
```
ianwjhalliday: nit: The expected and actual arguments here are reversed.  Should be
```javascript
assert.areEqual(SyntaxError.toString(), e.constructor.toString(), message);
```

test/es6/HTMLComments.js
tcare: I had a description of that on the line above, did you mean something more specific?

lib/Parser/Scan.cpp
```diff
@@ -1934,16 +1936,19 @@ tokens Scanner<EncodingPolicy>::ScanCore(bool identifyKwds)
...
-        case '{': Assert(chType == _C_LC); token = tkLCurly; break;
+        case '~': Assert(chType == _C_TIL); token = tkTilde;  break;
+        case '?': Assert(chType == _C_QUE); token = tkQMark;  break;
+        case '{': Assert(chType == _C_LC);  token = tkLCurly; break;
 
+        // ES 2015 11.3 Line Terminators
         case '\r':
         case '\n':
+        case kchLS:
+        case kchPS:
```
ianwjhalliday: It would appear that this was already handled by the default case at the top of the switch with a goto LNewLine.

However, if the scanner is in utf8 mode these characters will not be handled because this switch is operating only on the first part of the character via ReadFirst().  Is this broken in utf8?

test/es6/module-syntax.js
tcare: Wontfix since it matches the rest of the tests in the file and is out of scope. Filing an issue might be the way to go.

lib/Parser/Scan.cpp
```diff
@@ -2087,36 +2092,14 @@ tokens Scanner<EncodingPolicy>::ScanCore(bool identifyKwds)
...
                 p++;
                 token = tkDec;
-                if (m_fHtmlComments)
+                if (!m_fIsModuleCode)
                 {
                     int i = 0;
                     while ('-' == PeekFirst(p + i, last)) //Have already seen --, skip any further - characters
                         i++;
-                    if ('>' == PeekFirst(p + i++, last)) //This means we've got a --------------------------->.
+                    if ('>' == PeekFirst(p + i++, last) && m_fHadEol) //This means we've got a --------------------------->.
```
ianwjhalliday: `i++` is subtle.  `i` is no longer used afterwards so its moot but I'd be happier reading this code if this was `p + i + 1` instead.

lib/Parser/Scan.cpp
```diff
@@ -2087,36 +2092,14 @@ tokens Scanner<EncodingPolicy>::ScanCore(bool identifyKwds)
             case '-':
                 p++;
                 token = tkDec;
-                if (m_fHtmlComments)
+                if (!m_fIsModuleCode)
                 {
                     int i = 0;
                     while ('-' == PeekFirst(p + i, last)) //Have already seen --, skip any further - characters
                         i++;
-                    if ('>' == PeekFirst(p + i++, last)) //This means we've got a --------------------------->.
```
ianwjhalliday: Are there already tests for `--->` or other odd number of dashes to make sure they do not scan into decrement and minus tokens?

lib/Parser/Scan.cpp
tcare: Agreed.

lib/Parser/Scan.cpp
tcare: Not in Core. I can add some.

test/es6/HTMLComments.js
ianwjhalliday: I mean so that the text in the baseline is as descriptive

tcare: Ah, sure.

test/es6/module-syntax.js
ianwjhalliday: @boingoing @Yongqu hey guys fyi the arguments are reversed on the assert.areEquals calls in this file

lib/Parser/Scan.cpp
dilijev: `i++` as an expression evaluates to `i` and then increments `i` afterwards as a side effect. The value of `p + i++` is equivalent to `p + i`.

test/es6/HTMLComments.js
```diff
@@ -0,0 +1,72 @@
...
+WScript.Echo("Code before PS<-- is reachable");
--> WScript.Echo("Code after PS<-- is unreachable");
+WScript.Echo("PS<-- does not affect automatic semicolon insertion")
--> WScript.Echo("Code after PS<-- is unreachable");
+-->
+
+/*
+ * Line terminator sequences - non-standard (11.3 LineTerminatorSequence <CR>[lookahead ≠ <LF>])
+ */
+
```
dilijev: To keep things as simple as possible in this file, could you change the `≠` sign to `!=`.

I assume that's not part of the test.

test/es6/HTMLComments.js
```diff
@@ -0,0 +1,72 @@
...

--> WScript.Echo("Code after CRLS<-- is unreachable");
+WScript.Echo("CRLS<-- does not affect automatic semicolon insertion")

-->
+
+// CRPS
+WScript.Echo("Code before CRPS<-- is reachable");

--> WScript.Echo("Code after CRPS<-- is unreachable");
```
dilijev: More instances of the message not corresponding to the code. I noticed this in a few places. It's probably a good idea to read over the whole file to check that everything is correct.

test/es6/HTMLComments.js
dilijev: I wonder if the LS and PS characters confused GitHub's diff tool and that's why those lines are showing up as past the number of lines GitHub thinks are in the file. We should file a bug report :P

tcare: It really depends on your interpretation of lines! :)

test/es6/HTMLComments.js
tcare: Copy pasted from the spec. Fixed.

lib/Parser/Scan.cpp
ianwjhalliday: oh right, my bad, being too hasty

ianwjhalliday: and all the more reason it was poor usage

test/es6/HTMLComments.js
tcare: Spoke with Ian offline. I'm exploring the compliance for this manually and not including the ASI tests here. Testing for automatic semicolon insertion in this context is a bit tedious and verbose. What needs to be verified is that a valid HTML end comment gets treated like a newline in an ASI context.

lib/Parser/Scan.cpp
tcare: They don't need to be there. I'll add a comment to be explicit.

test/es6/HTMLComments.js
```diff
@@ -0,0 +1,59 @@
...
+WScript.Echo("Code before LF--> is reachable");
+--> WScript.Echo("Code after LF--> is unreachable");
+
+// LS u+2028
+WScript.Echo("Code before LS--> is reachable");
--> WScript.Echo("Code after LS--> is unreachable");
+
+// PS u+2029
+WScript.Echo("Code before PS--> is reachable");
--> WScript.Echo("Code after PS--> is unreachable");
```
dilijev: You already gave the unicode sequences in the comment above, so I think this is redundant.

dilijev:
@dotnet-bot test this please

tcare:
Ping, I'd like to get this merged before EOW

lib/Parser/Scan.cpp
tcare: Sorry I missed this in my most recent commit. I don't think this is actually allowed in the spec.

tcare:
I missed support for single line case:

/* comment */ -->

Adding that now.

ianwjhalliday:
Hmm, since that case was missed it would be good to add another test file that enumerates all the branches of the grammar productions in simple cases to smoke out any other gotchas if there are any.

tcare:
Yep, I think it's important that we be exhaustive here.

test/es6/HTMLComments.js
tcare: Fixed

tcare:
Ping, could someone please take a look at the update.

test/es6/HTMLComments.baseline
```diff
@@ -0,0 +1,13 @@
...
+Code before LF--> is reachable
+Code before LS--> is reachable
+Code before PS--> is reachable
+Code before CRLS--> is reachable
+Code before CRPS--> is reachable
+Code before <!-- is reachable
+Code before <!-- --> is reachable
+Code before <!-- LineTerminator --> is reachable
+Code before /* */ is reachable
+Code before /* */ is reachable
```
dilijev: What about code after `/* */`?

test/es6/HTMLComments.js
```diff
@@ -0,0 +1,66 @@
...
+   Line
+   Comment */ --> WScript.Echo("Code after */ --> is unreachable");
+WScript.Echo("Code before /* */ is reachable"); /* Comment */ --> WScript.Echo("Code after /* */ --> is unreachable");
+WScript.Echo("Code before /* */ is reachable"); /* Comment */--> WScript.Echo("Code after /* */--> is unreachable"); // No WhiteSpaceSequence
+
+// Post-decrement with a greater-than comparison does not get interpreted as a comment
+var a = 1; a-->a; WScript.Echo("Code after --> is reachable");
+assert.areEqual(0, a, "Post decrement executes");
+
+assert.throws(function () { eval('/* */ --->'); }, SyntaxError, "HTMLCloseComment causes syntax error with an extra -", "Syntax error");
```
dilijev: The output message here is misleading because it's not the same as the end comment token when adjacent to a variable. In any other case it shouldn't be reachable. Just asking for a clearer message for the output.

test/es6/HTMLComments.baseline
tcare: Sorry that should be --> after the /* */

test/es6/HTMLComments.js
dilijev: Also in general I think it might be better to have the messages about unreachability displayed in the case that the code is unreachable (set a flag before and unset it in the supposedly unreachable code and then print "<incorrect situation> is reachable" if the flag is unset, or something like that).

That way incorrect output in the baseline is more suggestive of the problem.

test/es6/module-syntax.js
```diff
@@ -124,6 +124,16 @@ var tests = [
             assert.doesNotThrow(function () { WScript.LoadModuleFile('.\\module\\ValidReExportStatements.js', 'samethread'); }, "Valid re-export statements");
         }
     },
+    {
+        name: "HTML comments do not parse",
```
ianwjhalliday: nit: "... do not parse in module scripts"

Just to be clear about the intent and scope of this test, that yes, indeed, this is specific to modules.

lib/Parser/Scan.cpp
```diff
@@ -2087,36 +2093,11 @@ tokens Scanner<EncodingPolicy>::ScanCore(bool identifyKwds)
             case '-':
                 p++;
                 token = tkDec;
-                if (m_fHtmlComments)
+                if (!m_fIsModuleCode)
                 {
-                    int i = 0;
-                    while ('-' == PeekFirst(p + i, last)) //Have already seen --, skip any further - characters
-                        i++;
-                    if ('>' == PeekFirst(p + i++, last)) //This means we've got a --------------------------->.
```
ianwjhalliday: So the `-------->` is no longer handled, correct?

There should be a test verifying this.

Also, what do the other browsers do about this case?  If they also allow `------>` then we should probably disregard the spec and maintain this behavior for compatibility.

tcare: There is a test for this, but GitHub is cutting it off in the diff. You can see it if you view the whole file.

I'll look into what other browsers do.

ianwjhalliday: Ah, cool.

test/es6/HTMLComments.baseline
dilijev: I assume you mean "that should be 'after the /* */' " and not something involving the HTML comment token `-->` :)

tcare: No, the test is for the single line comment variant /* */ --> which should be the test description.

dilijev: Oh ok, that also makes sense.

lib/Parser/Scan.cpp
tcare: My comment seemed to get lost. Both Firefox and Chrome reject --->

test/es6/HTMLComments.js
tcare: Fixed the first, wontfix the second. I think it's easy enough to see where something broke for a baseline test.

tcare:
Updated based on feedback from Ian and Doug

test/es6/module-syntax.js
```diff
@@ -124,6 +124,16 @@ var tests = [
             assert.doesNotThrow(function () { WScript.LoadModuleFile('.\\module\\ValidReExportStatements.js', 'samethread'); }, "Valid re-export statements");
         }
     },
+    {
+        name: "HTML comments do not parse in module code,
```
dilijev: There was supposed to be a close quote at the end of the name string, right?

ianwjhalliday:
Looks good

tcare:
@dotnet-bot test this please

dilijev:
LGTM. If you had more time to invest on this I would still push for making the test output indicative of both cases where the code was reachable and cases where it was not, but it's fine the way it is.

tcare:
I think the baseline is fine and it's readable as is.

Issue #347, opened 2016-02-18
[linux] Get CustomHeap compiling on Linux


digitalinfinity:
CustomHeap.cpp is currently not built on Linux (for no particular reason other than it's not needed until we port the JIT/dynamically allocated code pages to Linux). This issue tracks re-enabling this in the CMakeList for lib\Common\Memory.

Eyas:
Shouldn't this be closed since #382 has been merged?

PR #367, opened 2016-02-22
Reduce the logic in the Copyright Check

Francososa:
Implements changes suggested in issue #166 

Fixes #166

PR #382, opened 2016-02-24
re-enable CustomHeap build on linux

bewest:
Should help with bug #347, get CustomHeap compiling on Linux.
This compiles, but without the patch to the type of address initially, I get
the following errors compiling:

  * https://gist.github.com/bewest/89a1fd7be74218a37f98

    In file included from /home/bewest/src/ChakraCore/lib/Common/Memory/CustomHeap.cpp:14:
    [1m/home/bewest/src/ChakraCore/lib/Common/Memory/CustomHeap.h:92:5: [0m[0;1;31merror: [0m[1munknown type name '__notnull'[0m
        __field_bcount(size) char* address;

    /home/bewest/src/ChakraCore/pal/inc/rt/specstrings.h:278:45: [0m[0;1;30mnote: [0mexpanded from macro '__field_bcount'[0m
    #define __field_bcount(size)                __notnull __byte_writableTo(size)

For some reason __notnull doesn't exist.  A quick grepping through the source
reveals that it's only used here, in CustomHeap, and that the rest of the
address variables are typed with either char * or void *.  There is one __in
annotation, but since I'm not sure what these do, the simplest thing for time
being is this change.

### DIFF
--- a/lib/Common/Memory/CMakeLists.txt
+++ b/lib/Common/Memory/CMakeLists.txt
@@ -5,9 +5,7 @@ add_library (Chakra.Common.Memory
     Allocator.cpp
     ArenaAllocator.cpp
 
-    # xplat-todo: This is needed for allocating jitted code memory
-    # CustomHeap.cpp
-
+    CustomHeap.cpp
     CommonMemoryPch.cpp
     EtwMemoryTracking.cpp
     ForcedMemoryConstraints.cpp
--- a/pal/inc/rt/no_sal2.h
+++ b/pal/inc/rt/no_sal2.h
@@ -25,6 +25,8 @@
 #ifndef _NO_SAL_2_H_
 #define _NO_SAL_2_H_
 
+#undef __notnull
+#define __notnull
 #undef _When_
 #define _When_(c,a)
 #undef _At_


lib/Common/Memory/CustomHeap.h
```diff
@@ -89,7 +89,8 @@ struct Allocation
         } largeObjectAllocation;
     };
 
-    __field_bcount(size) char* address;
```
dilijev: These annotations help with static analysis tools on Windows and might have other implications to the compiler as well. Would prefer not to outright remove it if possible.

A localized solution would be to check whether this is a linux build and use #ifdef to select your new version.

A better solution would be to define the problematic macros under linux builds, again using #ifdef so that the definition doesn't conflate with the definition already being used on Windows.

jianchun: Agree with @dilijev

I was trying to build other code today and happened to run into this too. I tried different solutions and chose to add the last line below to ```CommonPal.h```. You can consider this approach.

```
#define __in    _SAL1_Source_(__in, (), _In_)
#define __out   _SAL1_Source_(__out, (), _Out_)
#define __notnull
```

bewest: Agree, was thinking along those lines.  Thanks for the helpful pointers! I think I prefer the latter solution, I was having a hard time spotting where in the existing there may be an `#ifdef` somewhere that keeps these from being defined. Will take a closer look.

lib/Common/Memory/CMakeLists.txt
```diff
@@ -6,7 +6,7 @@ add_library (Chakra.Common.Memory
     ArenaAllocator.cpp
 
     # xplat-todo: This is needed for allocating jitted code memory
```
digitalinfinity: Remove this comment now since this is no longer something that needs to be done :smile:

digitalinfinity:
Code looks good to me- I'll give it a spin locally tomorrow morning- thanks for picking this up @bewest!

dilijev:
@bewest Would you mind rebasing this PR to a single commit? You can follow the instructions I posted here: https://github.com/Microsoft/ChakraCore/pull/367#issuecomment-187420508

bewest:
Cleaned up; I pushed to `wip/bewest/build-customheap` so you can have a clean branch if you prefer, but I noticed your instructions use `-f` push.  It appears your workflow can deal with any potential ensuing confusion.

Here's some steps I took, in case it helps anyone else.


```
2019  git grep 'include.*specstr'
 2020  git grep 'include.*palrt"
 2021  git grep 'include.*palrt
 2022  git grep 'include.*palrt'
 2023  git grep 'include.*CommonP'
 2024  git grep palrt
 2025  git grep 'include.*specstr'
 2026  git grep 'include.*palrt"
 2027  git grep 'include.*palrt'
 2028  git grep 'intsafe'
 2029  git grep 'include.*specstr'
 2030  git grep __LINUX
 2031  git grep __notnull
 2032  git grep 'include.*sal'
 2033  git diff
 2034  git status
 2035  git checkout -b wip/bewest/build-customheap
 2036  git fetch  upstream 
 2037  git rebase  -i upstream/linux 
 2038  git log -p upstream/linux...
 2039  git status
 2040  git push origin -u wip/bewest/build-customheap 
 2041  git push origin -f wip/bewest/build-customheap:wip/bewest/bug-347/build-customheap 
 2042  history
bewest@bewest-MacBookPro:~/src/ChakraCore$ 

```

dilijev:
@bewest Thanks! GitHub handles force-push cleanly, so no need for the extra branch.

digitalinfinity:
Thanks @bewest- confirmed that the PR builds cleanly on my machine- I'll merge your change in.

PR #384, opened 2016-02-24
Param scope and body scope as sibling scopes

aneeshdk:
This change set implements param scope and the body scope as parallel
scopes. The previous logic of having param scope as an inner scope to the
body scope does not work well in eval cases. This changelist only
addresses function definitions in the param scope. I will be working on
the eval cases next. This works in interpreter mode right now. Backend
changes are not done yet.

Two new bytecodes were added. BeginBodyScope marks the beginning of the
body scope where a new closure object is created for the body scope. Until
then closure created for the param scope is used. After BeginBodyScope the
param scope one is not used other than copying the initial value for the
body scope symbols from the param scope symbols. This required the
creation of another new bytecode LdParamSlot. This also include some
refactoring required for enabling these.


aneeshdk:
@pleath @akroshg @tcare @abchatra Can you please take a look?

lib/Parser/Parse.cpp
```diff
@@ -4609,6 +4609,12 @@ bool Parser::ParseFncDeclHelper(ParseNodePtr pnodeFnc, ParseNodePtr pnodeFncPare
         isEnclosedInParamScope = this->m_currentScope->GetEnclosingScope() && this->m_currentScope->GetEnclosingScope()->GetScopeType() == ScopeType_Parameter;
     }
 
+    if (isEnclosedInParamScope)
+    {
+        // We cannot just do HasDefaultArguments here becasue we may be inside a destructuring pattern.
+        Assert(pnodeFncParent && pnodeFncParent->sxFnc.HasNonSimpleParameterList());
+    }
```
agarwal-sandeep: If the 'if' block is only for assert why not put isEnclosedInParamScope inside the assert itself?

lib/Runtime/ByteCode/ByteCodeWriter.cpp
```diff
@@ -1633,7 +1633,9 @@ namespace Js
                     OpCodeUtil::ConvertNonCallOpToProfiled(op);
                 }
                 break;
-
+            case OpCode::LdParamSlot:
+                // No prfiled opcode present for LdParamSlot
```
agarwal-sandeep: spelling profiled

lib/Parser/Parse.cpp
akroshg: Agreed - also there is no need for a comment though. can pnodeFncParent be null in any case?

---
In reply to: [53990567](https://github.com/Microsoft/ChakraCore/pull/384#discussion_r53990567) [](ancestors = 53990567)

lib/Parser/Parse.cpp
```diff
@@ -4848,40 +4854,49 @@ bool Parser::ParseFncDeclHelper(ParseNodePtr pnodeFnc, ParseNodePtr pnodeFncPare
...
                 {
                     OUTPUT_TRACE_DEBUGONLY(Js::ParsePhase, L"The param and body scope of the function %s cannot be merged\n", pnodeFnc->sxFnc.pnodeName ? pnodeFnc->sxFnc.pnodeName->sxVar.pid->Psz() : L"Anonymous function");
-                    // Now add a new symbol reference for each formal in the param scope to the body scope.
+                    // Add a new symbol reference for each formal in the param scope to the body scope.
                     paramScope->ForEachSymbol([this](Symbol* param) {
                         OUTPUT_TRACE_DEBUGONLY(Js::ParsePhase, L"Creating a duplicate symbol for the parameter %s in the body scope\n", param->GetPid()->Psz());
-                        this->CreateVarDeclNode(param->GetPid(), param->GetSymbolType(), false, nullptr, false);
+                        ParseNodePtr paramNode = this->CreateVarDeclNode(param->GetPid(), STVariable, false, nullptr, false);
+                        Assert(paramNode && paramNode->sxVar.sym->GetScope()->GetScopeType() == ScopeType_FunctionBody);
+                        paramNode->sxVar.sym->SetHasInit(true);
```
akroshg: I am wondering if the debugger tracks (under TrackDebuggerMetaData) this if this is marked Init.

lib/Runtime/ByteCode/ByteCodeEmitter.cpp
```diff
@@ -1723,14 +1725,26 @@ void ByteCodeGenerator::InitScopeSlotArray(FuncInfo * funcInfo)
...
-    if (scopeSlotCount == 0)
+    uint scopeSlotCountForParamScope = funcInfo->paramScope ? funcInfo->paramScope->GetScopeSlotCount() : 0;
+    if (scopeSlotCount == 0 && scopeSlotCountForParamScope == 0)
     {
         return;
     }
 
     Js::FunctionBody *byteCodeFunction = funcInfo->GetParsedFunctionBody();
+    if (scopeSlotCount || scopeSlotCountForParamScope)
+    {
```
akroshg: nit : change this to scopeSlotCount >0 || scopesSlotCountForParamScope > 0

lib/Runtime/ByteCode/ByteCodeEmitter.cpp
```diff
@@ -1723,14 +1725,26 @@ void ByteCodeGenerator::InitScopeSlotArray(FuncInfo * funcInfo)
...
+        byteCodeFunction->SetScopeSlotArraySizes(scopeSlotCount, scopeSlotCountForParamScope);
+    }
+
+    // Need to add property ids for the case when scopeSlotCountForParamSCope is non-zero
```
akroshg: is it TODO ?

test/es6/default.js
```diff
@@ -203,7 +203,7 @@ var tests = [
     }
   },
   {
-    name: "OS 1583694: Arguments sym is not set correctly on undo defer",
+    name: "OS 1583694: Arguments sym is not set correctly on undo defer",
     body: function () {
```
ianwjhalliday: Looks like a whitespace only change here.  I can't see what it is, did the type of newline change?

test/es6/default-splitscope.js
```diff
@@ -6,260 +6,622 @@
...
-            function funcDecl(a = eval("function foo() { return 1; }"), b = foo()) { return [a, b, foo()]; }
-            assert.areEqual([undefined, 1, 1], funcDecl(), "Declaring a function inside an eval works at default parameter scope");
-
-            function genFuncDecl(a = eval("function * foo() { return 1; }"), b = foo().next().value) { return [a, b, foo().next().value]; }
-            assert.areEqual([undefined, 1, 1], genFuncDecl(), "Declaring a generator function inside an eval works at default parameter scope");
-
-            function funcExpr(a = eval("var f = function () { return 1; }"), b = f()) { return [a, b, f()]; }
-            assert.areEqual([undefined, 1, 1], funcExpr(), "Declaring a function inside an eval works at default parameter scope");
+    { 
+       name: "Split parameter scope in function definition",
```
ianwjhalliday: Indentation got weird here.  name starts seven spaces indented now instead of eight and the rest of the file seems to follow.

test/es6/default-splitscope.js
```diff
@@ -6,260 +6,622 @@
...
+               assert.areEqual(a1, 20, "Assignment to the symbol inside the function changes the value"); 
+               return b; 
+           } 
+           assert.areEqual(f6()(), 10, "Function in the param scope correctly binds to the outer variable"); 
+            
+           function f7(a = 10, b = { iFnc () { return a; } }) { 
+               a = 20; 
+               return b; 
+           } 
+           assert.areEqual(f7().iFnc(), 10, "Function definition inside the object literal should capture the formal from the param scope");
```
ianwjhalliday: expected and actual argument values are in the wrong order here and throughout this file.  They should be swapped.

Expected is first for readability.  And getting these in the wrong order makes the the error message when they fail incorrect, it'll have expected and actual reversed.

test/es6/default-splitscope.js
```diff
@@ -6,260 +6,622 @@
...
+               assert.areEqual(a, 20, "New assignment in the body scope of the method updates the variable's value in body scope"); 
+               return b; 
+           } 
+           assert.areEqual(f1()(), 10, "Function expression defined in the param scope captures the formals from the param scope not body scope"); 
             
-            assert.throws(function () { eval("function foo(a = eval('b'), b) {}; foo();"); }, ReferenceError, "Future default references using eval are not allowed", "Use before declaration");
+           function f1(a = 10, b = function c(recurse = true) { return recurse ? c(false) : a; }) { 
+               return b; 
+           } 
+           assert.areEqual(f1()(), 10, "Function expression defined in the param scope captures the formals from the param scope not body scope");
```
ianwjhalliday: This description is the same as the previous one.  What's the difference that warrants having another test here?

test/es6/default-splitscope.js
```diff
@@ -6,260 +6,622 @@
...
             
-            assert.throws(function () { eval("function foo(a = eval('b'), b) {}; foo();"); }, ReferenceError, "Future default references using eval are not allowed", "Use before declaration");
+           function f1(a = 10, b = function c(recurse = true) { return recurse ? c(false) : a; }) { 
+               return b; 
+           } 
+           assert.areEqual(f1()(), 10, "Function expression defined in the param scope captures the formals from the param scope not body scope"); 
+       } 
+   }, 
+   { 
+       name: "Split parameter scope in member functions",
```
ianwjhalliday: Is there a difference between function declarations and object literal concise function members that affects codes paths related to split scope?
This test case feels redundant to me but maybe I'm missing something.

test/es6/default-splitscope.js
```diff
@@ -6,260 +6,622 @@
...
+                   assert.areEqual(a, 20, "New assignment in the body scope of the method updates the variable's value in body scope"); 
+                    return b; 
+                } 
+            } 
+            assert.areEqual(o1.f()(), 10, "Function defined in the param scope of the object method captures the formals from the param scope not body scope"); 
+             
+            var o2 = { 
+                f1(a = 10, b = function () { return { f2 () { return a; } } }) { 
+                    var a = 20; 
+                    c = function () { return { f2 () { return a; } } };
```
ianwjhalliday: This test case seems interesting but is not specific to object literal concise methods.

dilijev:
@dotnet-bot test this please

lib/Runtime/ByteCode/OpCodes.h
```diff
@@ -381,6 +381,7 @@ MACRO_WMS_PROFILED(     LdSlot,                     ElementSlot,    OpTempNumber
 MACRO_WMS_PROFILED(     LdEnvSlot,                  ElementSlotI2,  OpTempNumberSources)
 MACRO_WMS_PROFILED(     LdInnerSlot,                ElementSlotI2,  OpTempNumberSources)
 MACRO_WMS_PROFILED(     LdLocalSlot,                ElementSlotI1,  OpTempNumberSources)
+MACRO_EXTEND_WMS(       LdParamSlot,                ElementSlotI1,  OpTempNumberSources)
```
pleath: Can you add a profiled LdParamSlot? It should be easy.

lib/Runtime/ByteCode/ByteCodeGenerator.cpp
```diff
@@ -1875,7 +1828,9 @@ Scope * ByteCodeGenerator::FindScopeForSym(Scope *symScope, Scope *scope, Js::Pr
 {
     for (scope = scope ? scope->GetEnclosingScope() : currentScope; scope; scope = scope->GetEnclosingScope())
     {
-        if (scope->GetFunc() != funcInfo && scope->GetMustInstantiate() && scope != this->globalScope)
+        if (scope->GetFunc() != funcInfo
+            && (scope->GetMustInstantiate() || (scope->GetScopeType() == ScopeType_Parameter && !scope->GetCanMergeWithBodyScope()))
```
pleath: Is the param scope special here? Can we just use the must-instantiate flag to indicate that it is not merged and has to be there?

lib/Runtime/ByteCode/ByteCodeEmitter.cpp
```diff
@@ -4794,7 +4868,7 @@ ByteCodeGenerator::GetLdSlotOp(Scope *scope, int envIndex, Js::RegSlot scopeLoca
     return op;
 }
 
-void ByteCodeGenerator::EmitPropLoad(Js::RegSlot lhsLocation, Symbol *sym, IdentPtr pid, FuncInfo *funcInfo)
+void ByteCodeGenerator::EmitPropLoad(Js::RegSlot lhsLocation, Symbol *sym, IdentPtr pid, FuncInfo *funcInfo, bool useParamScope)
```
pleath: Is it not possible to figure out that the scope we're loading from is the param scope and dispense with the extra flag parameter?

lib/Runtime/ByteCode/ByteCodeEmitter.cpp
```diff
@@ -3113,60 +3127,63 @@ void ByteCodeGenerator::EmitOneFunction(ParseNode *pnode)
...
-                // the enclosing scope. When emitting the formals the stack should be like this: outer scope -> param scope -> formals.
-                // When emitting the body the stack should be like this: outer scope -> body scope -> body statements.
-                // For that pop the param scope, pop the body scope and then push the param scope again.
-                PopScope();
-                Assert(this->GetCurrentScope()->GetScopeType() == ScopeType_FunctionBody);
+                byteCodeFunction->SetParamAndBodyScopeNotMerged();
+
+                HomeArguments(funcInfo);
+
+                // Pop the body scope before emitting the default args
```
pleath: Where is the body scope getting pushed at this point? Is it up in StartEmitFunction?

lib/Parser/Parse.cpp
aneeshdk: Done. When enclosed in param scope pnodeFncParent cannot be null.

lib/Runtime/ByteCode/ByteCodeWriter.cpp
aneeshdk: Fixed

lib/Runtime/ByteCode/ByteCodeEmitter.cpp

lib/Runtime/ByteCode/ByteCodeEmitter.cpp
aneeshdk: Yes.

test/es6/default.js
aneeshdk: I am not sure either. I don't see any change there. Not sure why it got added.

test/es6/default-splitscope.js
aneeshdk: Fixed it.

test/es6/default-splitscope.js
aneeshdk: Fixed.

test/es6/default-splitscope.js
aneeshdk: Second one is a recursive sanity check. Fixed the comment.

test/es6/default-splitscope.js
aneeshdk: No there is no specific difference between the two. I just added them for sanity check.

test/es6/default-splitscope.js
aneeshdk: I will keep it there as a nested level test case for object literals. I have more test cases for function definitions in a dedicated nested test cases section.

lib/Runtime/ByteCode/ByteCodeEmitter.cpp
aneeshdk: Yes.

lib/Runtime/ByteCode/OpCodes.h
aneeshdk: Done.

lib/Runtime/ByteCode/ByteCodeGenerator.cpp
aneeshdk: As the param scope is in the stack with body scope marking param scope with must intantiate flag seems to cause issues with the environment index calculation.

lib/Runtime/ByteCode/ByteCodeEmitter.cpp
aneeshdk: Fixed it.

lib/Runtime/ByteCode/ByteCodeEmitter.cpp
pleath: Okay. I remember we discussed restructuring the StartEmitFunction level. Is there a reason why we can't do that? It would be much better, IMO, to keep the scope stack logically consistent throughout.

lib/Runtime/ByteCode/OpCodes.h
pleath: Cool. Is the Profiled opcode now being emitted with your changes, or do we also need a ByteCodeWriter change to make that happen (and an interpreter change to consume it)?

lib/Runtime/ByteCode/ByteCodeGenerator.cpp
pleath: Ah. We have 2 choices here: have the body and param scope both in the scope stack, and both counted; or have the param scope popped off the stack before we begin emitting the body of the function. We settled on the first one for ease of implementation, right? In that case, the code that accesses outer scopes should count them both. Making the param scope a special case in this calculation is asking for trouble.

lib/Runtime/ByteCode/ByteCodeEmitter.cpp
pleath: Cool. Thanks.

lib/Runtime/ByteCode/ByteCodeEmitter.cpp
aneeshdk: Fixed it,

lib/Parser/Parse.cpp
```diff
@@ -4600,15 +4600,17 @@ bool Parser::ParseFncDeclHelper(ParseNodePtr pnodeFnc, ParseNodePtr pnodeFncPare
...
-    // Function definition is inside the parent function's parameter scope  
+    // Function definition is inside the parent function's parameter scope    
     bool isEnclosedInParamScope = this->m_currentScope->GetScopeType() == ScopeType_Parameter;
 
-    if (this->m_currentScope->GetScopeType() == ScopeType_FuncExpr)
+    if (this->m_currentScope->GetScopeType() == ScopeType_FuncExpr || this->m_currentScope->GetScopeType() == ScopeType_Block)
     {
-        // Or this is a function expression enclosed in a parameter scope  
+        // Or this is a function expression or class enclosed in a parameter scope  
         isEnclosedInParamScope = this->m_currentScope->GetEnclosingScope() && this->m_currentScope->GetEnclosingScope()->GetScopeType() == ScopeType_Parameter;
```
pleath: Will this check always work? Can it be fooled by deeper nesting within block, etc., scopes? Is it possible to discover this more reliably by tracking param-scope-ness as we descend?

lib/Runtime/ByteCode/ByteCodeDumper.cpp
```diff
@@ -838,6 +838,7 @@ namespace Js
                 Output::Print(L" [%d] = R%d ",data->SlotIndex, data->Value);
                 break;
             case OpCode::LdLocalSlot:
+            case OpCode::LdParamSlot:
```
pleath: Do we need to be able to dump the Profiled variant as well?

lib/Runtime/ByteCode/ByteCodeEmitter.cpp
```diff
@@ -1218,8 +1218,7 @@ Js::RegSlot ByteCodeGenerator::DefineOneFunction(ParseNode *pnodeFnc, FuncInfo *
     // "We need a closure for the nested function");
 
     // If we are in a parameter scope and it is not merged with body scope then we have to create the child function as an inner function
-    if ((this->GetCurrentScope()->GetScopeType() != ScopeType_Parameter || this->GetCurrentScope()->GetCanMergeWithBodyScope())
-            && (regEnv == funcInfoParent->frameDisplayRegister || regEnv == funcInfoParent->GetEnvRegister()))
+    if (regEnv == funcInfoParent->frameDisplayRegister || regEnv == funcInfoParent->GetEnvRegister())
```
pleath: (Glad to see this get simpler. Thanks.)

lib/Runtime/ByteCode/ByteCodeEmitter.cpp
```diff
@@ -3762,36 +3869,10 @@ void ByteCodeGenerator::StartEmitFunction(ParseNode *pnodeFnc)
...
-            if (funcInfo->isSuperCtorLexicallyCaptured)
-            {
-                funcInfo->EnsureSuperCtorScopeSlot();
-            }
-
-            if (funcInfo->isNewTargetLexicallyCaptured)
-            {
-                funcInfo->EnsureNewTargetScopeSlot();
-            }
+            this->EnsureSpecialScopeSlots(funcInfo, bodyScope);
```
pleath: Good refactor. Thanks.

lib/Runtime/ByteCode/ByteCodeGenerator.cpp
```diff
@@ -3130,13 +3064,39 @@ void VisitNestedScopes(ParseNode* pnodeScopeList, ParseNode* pnodeParent, ByteCo
                 i = 0;
                 ParseNodePtr containerScope = pnodeScope->sxFnc.pnodeScopes;
 
-                VisitNestedScopes(containerScope, pnodeScope, byteCodeGenerator, prefix, postfix, &i);
+                if (pnodeScope->sxFnc.HasNonSimpleParameterList() && !pnodeScope->sxFnc.funcInfo->GetParamScope()->GetCanMergeWithBodyScope())
+                {
+                    // Pop the body scope
+                    byteCodeGenerator->PopScope();
```
pleath: Where was the body scope pushed prior to this point? Is this situation similar to the one that we fixed in the Emit pass?

pleath:
Can you show me where the interpreter verifies the index when it executes [Profiled]LdParamSlot?

lib/Parser/Parse.cpp
aneeshdk: I can't think of any scenario where a block scope can be introduced in param scope without causing a syntax error, other than classes.

akroshg: this->m_currentScope->GetEnclosingScope() cannot be null as it will have global scope, no?

---
In reply to: [54613965](https://github.com/Microsoft/ChakraCore/pull/384#discussion_r54613965) [](ancestors = 54613965)

lib/Parser/Parse.cpp
```diff
@@ -4844,44 +4846,52 @@ bool Parser::ParseFncDeclHelper(ParseNodePtr pnodeFnc, ParseNodePtr pnodeFncPare
             pnodeFnc->sxFnc.pnodeVars = nullptr;
             m_ppnodeVar = &pnodeFnc->sxFnc.pnodeVars;
 
-            // We can't merge the param scope and body scope any more as the nested methods may be capturing params.
             if (pnodeFnc->sxFnc.HasNonSimpleParameterList() && !fAsync)
             {
                 Scope* paramScope = pnodeFnc->sxFnc.pnodeScopes->sxBlock.scope;
+                Assert(paramScope);
```
akroshg: nit: have it Assert(paramScope != nullptr)

lib/Runtime/ByteCode/ByteCodeDumper.cpp
aneeshdk: Profiled versions of the Ld bytecodes seems to be taken care by Write.

pleath: Yes, but is the dumper able to dump them?

lib/Parser/Parse.cpp
pleath: Okay. Thanks.

lib/Runtime/ByteCode/ByteCodeDumper.cpp
aneeshdk: Yes. I verified that.

lib/Parser/Parse.cpp

aneeshdk:
The LdParamSlot uses the same method used by LdLocalSlot. OP_LdSlot() has the index check.

pleath:
Okay. Thanks.

--Paul

lib/Runtime/ByteCode/ByteCodeDumper.cpp
pleath: Thanks.

--Paul

lib/Runtime/ByteCode/ByteCodeGenerator.cpp
aneeshdk: Sorry I missed that. Yes it was similar to the Emit pass. Fixed it in the last commit.

PR #385, opened 2016-02-24
One-line type-check fix

pleath:
My previous fix to monomorphic type checks had a subtle mistake. To handle the case where a downstream store adds a property, I grabbed the store's cached initial type to use as the guard type at the monomorphic check. But in some cases, the type we checked upstream is the final type, not the initial one. One-line fix for that.

### DIFF
--- a/test/fieldopts/fixedfieldmonocheck2.js
+++ b/test/fieldopts/fixedfieldmonocheck2.js
@@ -1,0 +1,45 @@
+//-------------------------------------------------------------------------------------------------------
+// Copyright (C) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE.txt file in the project root for full license information.
+//-------------------------------------------------------------------------------------------------------
+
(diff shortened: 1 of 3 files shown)


test/fieldopts/fixedfieldmonocheck2.js
```diff
@@ -0,0 +1,45 @@
...
+        for (var __loopvar1001 = 7; obj1.method1() ;) {
+        }
+    }
+    var v44 = v37();
+}
+test0();
+test0();
+test0();
+
+WScript.Echo('pass');
```
dilijev: nit: newline at end of file

dilijev:
@dotnet-bot test this please

pleath:
If anybody wants to take a chance on a one-liner in the JIT, this one will probably keep hitting in automated testing. (Thanks.)

satheeshravi:
lgtm

PR #388, opened 2016-02-24
Fix incremental LTCG disabling on TFS builds

tcare:
ChakraCore builds on TFS are meant to have incremental LTCG disabled. This reduces the build size by about half since we don't need the ipdb/iobj files.


tcare:
@curtisman @digitalinfinity

tcare:
There's an extra commit in here, removing.

PR #391, opened 2016-02-25
linux: make NumberUtilities_strtod.cpp compile

xanlpz:
Two issues here:

- wchar_t and WCHAR have different sizes on Linux (32bit and
  16bit). This can happen since wchar_t is compiler defined and WCHAR
  is defined as an unsigned short. To avoid ambiguities just change
  all uses of wchar_t for wchar16 in NumberUtilities*. This should
  probably done everywhere, including changing WCHAR to wchar16 too,
  but for now this is the minimum patch that makes this file compile
  while leaving it in a consistent state.

- The way the DECIMAL structure is accessed does not match its
  definition in pal/. Change the code to use the macros provided, that
  should work everywhere if they are correctly defined.

### DIFF
--- a/lib/Common/Common/CMakeLists.txt
+++ b/lib/Common/Common/CMakeLists.txt
@@ -10,7 +10,7 @@ add_library (Chakra.Common.Common
     Jobs.cpp
     MathUtil.cpp
     NumberUtilities.cpp
-    # NumberUtilities_strtod.cpp
+    NumberUtilities_strtod.cpp
     RejitReason.cpp
     SmartFpuControl.cpp
     Tick.cpp
--- a/lib/Common/Common/NumberUtilities.cpp
+++ b/lib/Common/Common/NumberUtilities.cpp
@@ -37,7 +37,7 @@ namespace Js
         return ch >= '0' && ch <= '9';
     }
 
-    BOOL NumberUtilities::FHexDigit(wchar_t ch, int *pw)
+    BOOL NumberUtilities::FHexDigit(wchar16 ch, int *pw)
     {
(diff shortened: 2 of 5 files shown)


xanlpz:
So the macros are not part of the public API. Just change the struct to match its definition in MSDN, that should work.

dilijev:
I'm not sure this is the recommended solution we've been using so far for issues like this. Also I don't think there should be any need to modify the PAL code, since it was already designed to compile cross-platform.

@digitalinfinity can confirm and offer advice if necessary.

xanlpz:
For the wchar_t/WCHAR issue I was basically following the advice given by jianchun some time ago, but I'll be happy to do something else if that's preferred. For the record he said:

> Currently we are replacing wchar_t with wchar16, and L"..." with CH_WSTR("..."). Will that solve the issues? PAL_STDCPP_COMPAT is a sensitive symbol (not sure what exactly it does either), yes, changing it breaks other things.

As for the tagDEC/DECIMAL structure, basically what's in pal/ does not match this: https://msdn.microsoft.com/en-us/library/windows/desktop/ms221061%28v=vs.85%29.aspx (the struct in pal/ does not have anonymous unions), so the code won't compile. If it was designed to be crossplatform then I guess I'm missing something here.

jianchun:
The change LGTM.

We also found PAL had issues. The case here is either CoreCLR uses custom macros to interact with DECIMAL or the def is not being used at all. @xanlpz 's change conforms to msdn and works better.

digitalinfinity:
@xanlpz change looks good to me too

PR #392, opened 2016-02-25
Removed the ParseType Reparse from the parser.

akroshg:
Essentially it is used for *not* doing the parse time binding. However now we have moved to the parse-time binding for the debugger we don't need this differentiation anymore. Once I did that there is no need to ParseType_Reparse. So I have removed the whole usage altogether.


akroshg:
@pleath @agarwal-sandeep  please take a look.

pleath:
LGTM. Nice that we can actually simplify things a little.

agarwal-sandeep:
Nice to see no more debugger specific logic

PR #393, opened 2016-02-25
Change CustomHeap to on demand update on whether a page should be in the full list or not.

curtisman:
Now that the page allocator for the code pages are shared amongst multiple CustomHeap, the state of the secondary allocator may change by other CustomHeap, making it impossible to maintain that full list are block that are really full (no more allocable space in the page and/or no more secondary space for the segment).  So, walking all the block to and move the pages between full list and the bucket list if the current CustomHeap when secondary allocator state changes, do it when we allocate on demand.

In PreReservedVirtualAllocWrapper::IsInRange, only check if the addres is committed when it is in the preserved region. We don't care otherwise.


curtisman:
@digitalinfinity Please code review.

digitalinfinity:
LGTM

PR #394, opened 2016-02-25
linux: partially builds lib/Parser

jianchun:
This enables partially building lib/Parser.
- Changed many lib/Runtime headers as lib/Parser includes "Runtime.h".
  Also changed some lib/Common headers and pulled in a few PAL headers to
  make it build.
- vtinfo.h: clang doesn't allow class initialization list
  ```Base(v), __VA_ARGS__``` if ```__VA_ARGS__``` is empty. Added an
  ```..._INIT``` version for extra initialization list.
- Added symbol ENABLE_GLOBALIZATION, ENABLE_SCRIPT_DEBUGGING,
  ENABLE_SCRIPT_PROFILING to #ifdef out features on Linux for now. Those
  need revisit in future.
- Added emmintrin.h include, #ifdef out conflicting __m128i from palrt.h.
- Added overloaded versions of InterlockedExchangeAdd and
  InterlockedExchangeSubtract, needed by our code.
- Forward enum declarations are not allowed unless the delcarations
  specifies enum storage type. Added storage type or int by default.
- "static" template method specializations are not allowed in class body.
  Moved specializations out of class and removed "static" keyword.
- "default", "and", "or", "xor", "not" are keywords not allowed as enum
  members. Renamed.
- function default parameter value is not allowed in implementation.
  Removed or moved to declaration.


digitalinfinity:
Ah man, that is a hard change to read- I'l be glad when we do the change to wrap all L""-strs with the macro. Lol, I have a lot of the WSTR changes locally too- and the datastructure changes :smile: this will be a fun merge 

One comment- I see you added InterlockedAdd* to CommonPal.h- I've been adding those functions to the PAL itself- I don't have a strong preference either way but we should be consistent here.

Otherwise- LGTM- please merge, would like to rebase on top of your changes :smile:

jianchun:
Thanks for review!

> Ah man, that is a hard change to read- I'l be glad when we do the change to wrap all L""-strs with the macro. Lol, I have a lot of the WSTR changes locally too- and the datastructure changes  this will be a fun merge 

That's exactly why I proposed to do batch str replacements.

> One comment- I see you added InterlockedAdd* to CommonPal.h- I've been adding those functions to the PAL itself- I don't have a strong preference either way but we should be consistent here.

One reason I didn't put them in PAL itself is PAL InterlockedAdd* are in an extern "C" section. Our InterlockedAdd* needs __cplusplus as they overload. Also I tried to reduce direct PAL change as it is hard to track what we deviate. No strong preference either, I can adapt to what we need.

PR #396, opened 2016-02-26
Fix stack closure boxing bug

pleath:
We need to box frame display and scope slots in a native frame before calling BoxNativeFrame so that the boxed versions can be found and written back to their known locations on the frame.


rajatd:
Looks good.

pleath:
Thanks.

PR #397, opened 2016-02-26
Use zero pages list before virtualalloc

leirocks:
Before allocating new pages, check if there's zero pages or pending zero pages from background sweeping, if yes use them instead of allocating new pages

### DIFF
--- a/lib/Common/Memory/PageAllocator.cpp
+++ b/lib/Common/Memory/PageAllocator.cpp
@@ -752,6 +752,8 @@ template<typename T>
 char *
 PageAllocatorBase<T>::TryAllocFromZeroPagesList(uint pageCount, PageSegmentBase<T> ** pageSegment, SLIST_HEADER& zeroPagesList, bool isPendingZeroList)
 {
+    FAULTINJECT_MEMORY_NOTHROW(this->debugName, pageCount*AutoSystemInfo::PageSize);
+
     char * pages = nullptr;    
     FreePageEntry* localList = nullptr;
     while (true) 
--- a/lib/Common/Memory/PageAllocator.cpp
+++ b/lib/Common/Memory/PageAllocator.cpp
@@ -749,70 +749,113 @@ PageAllocatorBase<T>::AddPageSegment(DListBase<PageSegmentBase<T>>& segmentList)
 }
 
 template<typename T>
+char *
+PageAllocatorBase<T>::TryAllocFromZeroPagesList(uint pageCount, PageSegmentBase<T> ** pageSegment, SLIST_HEADER& zeroPagesList)
+{
+    char * pages = nullptr;
+    FreePageEntry * freePage = (FreePageEntry *)::InterlockedPopEntrySList(&zeroPagesList);
+    FreePageEntry* localList = nullptr;
+    while (freePage != nullptr) {
+#if DBG
+        UpdateMinimum(this->debugMinFreePageCount, this->freePageCount);
+#endif
+
+        if (freePage->pageCount == pageCount)
+        {
+            *pageSegment = freePage->segment;
+            pages = (char *)freePage;
+            memset(pages, 0, sizeof(FreePageEntry));
+            this->FillAllocPages(pages, pageCount);
+            break;
+        }
+        else if (freePage->pageCount > pageCount)
+        {
+            *pageSegment = freePage->segment;
+            freePage->pageCount -= pageCount;
+            pages = (char *)freePage + freePage->pageCount * AutoSystemInfo::PageSize;
+            this->FillAllocPages(pages, pageCount);
+
+            ::InterlockedPushEntrySList(&backgroundPageQueue->freePageList, freePage);
+            break;
+        }
+        else
+        {
+            freePage->Next = localList;
+            localList = (FreePageEntry*)freePage;
+        }
+
+        freePage = (FreePageEntry *)::InterlockedPopEntrySList(&zeroPagesList);
+    }
+
+    while (localList != nullptr)
+    {
+        FreePageEntry* next = (FreePageEntry*)localList->Next;
+        ::InterlockedPushEntrySList(&zeroPagesList, localList);
+        localList = next;
+    }
+
+    return pages;
+}
+
+template<typename T>
+char *
+PageAllocatorBase<T>::TryAllocFromZeroPages(uint pageCount, PageSegmentBase<T> ** pageSegment, PageHeapMode pageHeapFlags)
+{
+    if (backgroundPageQueue != nullptr) 
+    {
+        return TryAllocFromZeroPagesList(pageCount, pageSegment, backgroundPageQueue->freePageList);
+    }
+    return nullptr;
+}
+
+template<typename T>
 template <bool notPageAligned>
 char *
 PageAllocatorBase<T>::TryAllocFreePages(uint pageCount, PageSegmentBase<T> ** pageSegment, PageHeapMode pageHeapFlags)
 {
     Assert(!HasMultiThreadAccess());
-    if (this->freePageCount < pageCount)
+    char* pages = nullptr;
+
+    if (this->hasZeroQueuedPages)
     {
-        return nullptr;
-    }
-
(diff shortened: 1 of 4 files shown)


leirocks:
@curtisman @digitalinfinity can you review please?

lib/Common/Memory/PageAllocator.cpp
```diff
@@ -749,70 +749,113 @@ PageAllocatorBase<T>::AddPageSegment(DListBase<PageSegmentBase<T>>& segmentList)
...
+                this->FillAllocPages(pages, pageCount);
+                return pages;
             }
-            this->FillAllocPages(pages, pageCount);
-            return (char *)pages;
         }
     }
-
-    return nullptr;
+
```
thefourtheye: Unnecessary whitespace change.

lib/Common/Memory/RecyclerPageAllocator.cpp
```diff
@@ -25,6 +25,24 @@ bool RecyclerPageAllocator::IsMemProtectMode()
     return recycler->IsMemProtectMode();
 }
 
+char * RecyclerPageAllocator::TryAllocFromZeroPages(uint pageCount, PageSegmentBase<VirtualAllocWrapper> ** pageSegment, PageHeapMode pageHeapFlags)
```
thefourtheye: inconsistent pointer syntax (check line 30 in the same file)

lib/Common/Memory/PageAllocator.cpp
```diff
@@ -749,70 +749,113 @@ PageAllocatorBase<T>::AddPageSegment(DListBase<PageSegmentBase<T>>& segmentList)
...
-            }
+        return TryAllocFromZeroPagesList(pageCount, pageSegment, backgroundPageQueue->freePageList);
+    }
+    return nullptr;
+}
 
-            this->freePageCount -= pageCount;
-            *pageSegment = freeSegment;
+template<typename T>
+template <bool notPageAligned>
```
thefourtheye: inconsistent with the previous line

lib/Common/Memory/RecyclerPageAllocator.cpp
```diff
@@ -25,6 +25,24 @@ bool RecyclerPageAllocator::IsMemProtectMode()
...
+
+    pages = TryAllocFromZeroPagesList(pageCount, pageSegment, this->zeroPageQueue.pendingZeroPageList);
+    if (pages != nullptr)
+    {
+        // zero out
+        memset(pages, 0, pageCount*AutoSystemInfo::PageSize);
+        return pages;
+    }
+
+    pages = TryAllocFromZeroPagesList(pageCount, pageSegment, this->zeroPageQueue.freePageList);
```
curtisman: We should try the freePageList first before the pendingZeroPageList,as it is cheaper without the need to zero the pages first.

lib/Common/Memory/PageAllocator.cpp
```diff
@@ -749,70 +749,113 @@ PageAllocatorBase<T>::AddPageSegment(DListBase<PageSegmentBase<T>>& segmentList)
...
+            break;
+        }
+        else if (freePage->pageCount > pageCount)
+        {
+            *pageSegment = freePage->segment;
+            freePage->pageCount -= pageCount;
+            pages = (char *)freePage + freePage->pageCount * AutoSystemInfo::PageSize;
+            this->FillAllocPages(pages, pageCount);
+
+            ::InterlockedPushEntrySList(&backgroundPageQueue->freePageList, freePage);
```
curtisman: If this is the pendingZeroPageList, you need to zero out all the page before putting it back to the freePageList.
You can't put it back to the pendingZeroPageList as it may be missed by the background GC thread's BackgroundZeroQueuedPages calls.

lib/Common/Memory/RecyclerPageAllocator.cpp
leirocks: yeah, I hesitated here. it was freePages first. Then I thought pendingZeroPage might have bigger chance to hit. But I think you are right, checking existence of freePages cost should be very low comparing to zero out total 4 pages. let me change back

lib/Common/Memory/PageAllocator.cpp
```diff
@@ -749,70 +749,113 @@ PageAllocatorBase<T>::AddPageSegment(DListBase<PageSegmentBase<T>>& segmentList)
...
+        }
+
+        freePage = (FreePageEntry *)::InterlockedPopEntrySList(&zeroPagesList);
+    }
+
+    while (localList != nullptr)
     {
-        return nullptr;
+        FreePageEntry* next = (FreePageEntry*)localList->Next;
+        ::InterlockedPushEntrySList(&zeroPagesList, localList);
```
curtisman: Same comment as before. You can't put it back to the pendingZeroPageList as it may be missed by the background GC thread's BackgroundZeroQueuedPages calls.

Alternatively, you will have to add a "last chance" ZeroQueuedPages on the main thread before we to ConcurrentTransferSwept.

lib/Common/Memory/PageAllocator.cpp
leirocks: hmm.. then I need to zero out all checked out pendingZero pages in below while loop..

lib/Common/Memory/PageAllocator.cpp
```diff
@@ -749,70 +749,147 @@ PageAllocatorBase<T>::AddPageSegment(DListBase<PageSegmentBase<T>>& segmentList)
...
+                this->FillAllocPages(pages, pageCount);
+                break;
+            }
+        }
+    }
+
+    if (localList != nullptr)
+    {
+        uint newFreePages = 0;
+        while (true)
```
curtisman: Why not "do { } while (localList != nullptr)"?

leirocks: hmm, localList can be null. but I can  remove the inner if check with while(localList!=nullptr){}

lib/Common/Memory/RecyclerPageAllocator.h
```diff
@@ -40,5 +40,6 @@ class RecyclerPageAllocator : public IdleDecommitPageAllocator
     Recycler* recycler;
 
     bool IsMemProtectMode();
+    virtual char * TryAllocFromZeroPages(uint pageCount, PageSegmentBase<VirtualAllocWrapper> ** pageSegment, PageHeapMode pageHeapFlags) override;
```
curtisman: Can we avoid the virtual here? 
I think you can just check for HasZeroPageQueue()?

leirocks: OK

leirocks: also looks there's a bug that I checked hasZeroQueuedPages before allocating from zero page list. should check for that only for pending zero page list allocating

lib/Common/Memory/PageAllocator.cpp
```diff
@@ -749,70 +749,141 @@ PageAllocatorBase<T>::AddPageSegment(DListBase<PageSegmentBase<T>>& segmentList)
...
+PageAllocatorBase<T>::TryAllocFreePages(uint pageCount, PageSegmentBase<T> ** pageSegment, PageHeapMode pageHeapFlags)
+{
+    Assert(!HasMultiThreadAccess());
+    char* pages = nullptr;
+
+    if (this->freePageCount >= pageCount)
     {
-        FreePageEntry * freePage = (FreePageEntry *)::InterlockedPopEntrySList(&backgroundPageQueue->freePageList);
-        if (freePage != nullptr)
+        FAULTINJECT_MEMORY_NOTHROW(this->debugName, pageCount*AutoSystemInfo::PageSize);
```
curtisman: Should this be outside of the "if"? Or we need to add it in TryAllocFromZeroPagesList?

leirocks: hmm. yes

curtisman:
LGTM

PR #399, opened 2016-02-26
Invalidate proto caches in the correct script context

rajatd:
If the prototype of an object is changed in a different context, we need to invalidate proto inline caches in the script context of the old prototype. We were only invalidating the caches in the current context, but the prototype being changed could have had property loads from it cached in a different context.

[Review on Reviewable](https://reviewable.io/reviews/microsoft/chakracore/399)

### DIFF
--- a/lib/Runtime/Base/CrossSiteObject.h
+++ b/lib/Runtime/Base/CrossSiteObject.h
@@ -223,21 +223,13 @@ namespace Js
     template <typename T>
     void CrossSiteObject<T>::RemoveFromPrototype(ScriptContext * requestContext)
     {
-        if (GetScriptContext() == requestContext)
-        {
-            __super::RemoveFromPrototype(requestContext);
-        }
-        // else do nothing because we never cache cross-context
+        __super::RemoveFromPrototype(GetScriptContext());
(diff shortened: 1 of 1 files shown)


lib/Runtime/Base/CrossSiteObject.h
```diff
@@ -223,21 +223,13 @@ namespace Js
     template <typename T>
     void CrossSiteObject<T>::RemoveFromPrototype(ScriptContext * requestContext)
     {
-        if (GetScriptContext() == requestContext)
-        {
-            __super::RemoveFromPrototype(requestContext);
-        }
-        // else do nothing because we never cache cross-context
+        __super::RemoveFromPrototype(GetScriptContext());
```
pleath: Can we actually just get rid of these methods, since they only forward to the parent?

jianchun: > Can we actually just get rid of these methods, since they only forward to the parent?

There is one change. It forwards but replaced ```requestContext``` to ```GetScriptContext()```.

pleath:
In that case, LGTM as is.

PR #400, opened 2016-02-26
Enable ES7 Array.prototype.includes by default and add to @@unscopables

tcare:
Enable ES7 Array.prototype.includes by default and add to @@unscopables

### DIFF
--- a/lib/Runtime/Library/JavascriptLibrary.cpp
+++ b/lib/Runtime/Library/JavascriptLibrary.cpp
@@ -1671,13 +1671,14 @@ namespace Js
         {
             DynamicType* dynamicType = DynamicType::New(scriptContext, TypeIds_Object, library->nullValue, nullptr, NullTypeHandler<false>::GetDefaultInstance(), false);
             DynamicObject* unscopablesList = DynamicObject::New(library->GetRecycler(), dynamicType);
-            unscopablesList->SetProperty(PropertyIds::find, JavascriptBoolean::ToVar(true, scriptContext), PropertyOperation_None, nullptr);
-            unscopablesList->SetProperty(PropertyIds::findIndex, JavascriptBoolean::ToVar(true, scriptContext), PropertyOperation_None, nullptr);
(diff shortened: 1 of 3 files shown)


tcare:
@abchatra please review

tcare:
Hmm, problem with ch. Investigating.

tcare:
Full rununittests.cmd doesn't run core tests :/ Will refile when I've fixed issues

tcare:
Ping - any takers? @ianwjhalliday @boingoing @akroshg

lib/Runtime/Library/JavascriptLibrary.cpp
```diff
@@ -1858,12 +1857,7 @@ namespace Js
         library->AddFunctionToLibraryObject(typedarrayPrototype, PropertyIds::keys, &TypedArrayBase::EntryInfo::Keys, 0);
         JavascriptFunction* valuesFunc = library->AddFunctionToLibraryObject(typedarrayPrototype, PropertyIds::values, &TypedArrayBase::EntryInfo::Values, 0);
         library->AddMember(typedarrayPrototype, PropertyIds::_symbolIterator, valuesFunc);
-
-        if (scriptContext->GetConfig()->IsES7BuiltinsEnabled())
-        {
-            library->AddFunctionToLibraryObject(typedarrayPrototype, PropertyIds::includes, &TypedArrayBase::EntryInfo::Includes, 1);
-        }
-
+        library->AddFunctionToLibraryObject(typedarrayPrototype, PropertyIds::includes, &TypedArrayBase::EntryInfo::Includes, 1);
```
goyakin: I think it would be good to have tests for function length and the property attributes. Test262 might have these, but I'm not sure. If it does, they might be enough :wink:

goyakin:
@tcare I added some minor comments inline, but the PR LGTM otherwise.

tcare:
Thanks Gorkem. I took the opportunity to fix a lot of the inconsistent spacing.

I'll look into the function length / property attributes but they may be out of scope for this change.

goyakin:
LGTM.

Regarding the function length and the property attributes, I misread your change. They would be out of scope for this PR.

PR #401, opened 2016-02-26
Page allocator used by CustomHeap's metadata don't need standing free page

curtisman:
The CustomHeap metadata isn't very big and we don't allocate them too often.
So we don't need standing free page for it (i.e. only commit if a page is in use)


digitalinfinity:
LGTM

PR #405, opened 2016-02-27
Replace malloc/free with HeapAlloc/HeapFree in HeapAllocator.cpp and use private heap for allocations. Add configuration flag PrivateHeap.

dilijev:
This enables Chakra to make allocations on a private heap instead of the process heap used by malloc/free. This reduces the likelihood of heap corruption from other modules and may improve fragmentation caused by interleaving allocations with other modules. (Allocations within our engine tend to be the same size so we would get good re-use of heap-allocated memory.)

Perflab on this change against the previous commit as a baseline shows that there are some marginal wins and losses in memory usage, within a reasonable standard deviation. i.e. No noticable perf impact.

[Review on Reviewable](https://reviewable.io/reviews/microsoft/chakracore/405)


dilijev:
@curtisman @digitalinfinity please review?

lib/Common/Memory/HeapAllocator.h
```diff
@@ -130,6 +130,35 @@ struct HeapAllocator
...
 
+private:
+    // TODO (doilij) add config flags to enable/disable PrivateHeap functionality
+    HANDLE PrivateHeap;
+
+    void InitPrivateHeap()
+    {
+        if (this->PrivateHeap == nullptr)
+        {
+            PrivateHeap = HeapCreate(0, 0, 0); // no options, default initial size, no max size
```
dilijev: I forgot `this->`. It's not required in MSVC. It might be required in clang but I think not because PrivateHeap is declared in this class, not a superclass. I'll add it to clean it up if there are any other changes to be made from this review.

lib/Common/Memory/HeapAllocator.h
```diff
@@ -130,6 +130,35 @@ struct HeapAllocator
     static HeapAllocator Instance;
     static HeapAllocator * GetNoMemProtectInstance();
 
+private:
+    // TODO (doilij) add config flags to enable/disable PrivateHeap functionality
```
dilijev: Whoops, this is done.

lib/Common/Memory/HeapAllocator.cpp
```diff
@@ -146,7 +153,15 @@ void HeapAllocator::Free(void * buffer, size_t byteSize)
         return;
     }
 #endif
-    free(buffer);
+
+    if (CONFIG_FLAG(PrivateHeap))
+    {
+        HeapFree(GetPrivateHeap(), 0, buffer);
```
Yongqu: If we are going this direction, maybe we can have different serialized/non-serialized heap? per thread heap can have HEAP_NO_SERIALIZE flag set. Only limited arena need cross-thread access.

lib/Common/Memory/HeapAllocator.h
```diff
@@ -130,6 +130,34 @@ struct HeapAllocator
...
+            BOOL success = HeapDestroy(this->PrivateHeap);
+            Assert(success);
+            this->PrivateHeap = nullptr;
+        }
+    }
+
+    HANDLE GetPrivateHeap()
+    {
+        InitPrivateHeap(); // will initialize PrivateHeap if not already initialized
+        return this->PrivateHeap;
```
Yongqu: will it make sense to have per threadcontext private heap?

lib/Common/Memory/HeapAllocator.h
```diff
@@ -130,6 +130,34 @@ struct HeapAllocator
     static HeapAllocator Instance;
     static HeapAllocator * GetNoMemProtectInstance();
 
+private:
+    HANDLE PrivateHeap;
```
Yongqu: why not camelCase here if this is not static?

dilijev: I could do that. According to our coding convention it should also start with `m_`, so I'll change it to `m_privateHeap`

lib/Common/Memory/HeapAllocator.h
curtisman: I don't think that will buy us much.  Most of the allocation that we use the heap for require multithread access (JIT bail record, recycler metadata, etc.)

lib/Common/Memory/HeapAllocator.h
```diff
@@ -130,6 +130,34 @@ struct HeapAllocator
...
+
+    void InitPrivateHeap()
```
curtisman: Implementation in CPP files?

dilijev: Ah, yep. I'll move it.

lib/Common/Memory/HeapAllocator.cpp
dilijev: I'll think about this for a future change, but I think this is something we want to get in early this week so I'd like to check in at first without making that change and re-evaluate with @curtisman

lib/Common/Memory/HeapAllocator.h
```diff
@@ -130,6 +130,14 @@ struct HeapAllocator
     static HeapAllocator Instance;
     static HeapAllocator * GetNoMemProtectInstance();
 
+private:
+    HANDLE m_privateHeap;
```
dreamer-dead: It seems that you should initialise this member with nullptr value here or in constructor initializer list.
Otherwise it may contain some random garbage value in Release build and then HeapAllocator::InitPrivateHeap() function will fail to initialize the heap.

dilijev: Thanks for catching that! I previously had this as a static member and initialized it as a static field but when I changed it I forgot to move the initialization to the constructor. I'll update shortly.

lib/Common/Memory/HeapAllocator.cpp
```diff
@@ -303,10 +339,27 @@ void NoThrowNoMemProtectHeapAllocator::ClearTrackAllocInfo(TrackAllocData* data
 #endif // TRACK_ALLOC
 #endif
 
-#if defined(HEAP_TRACK_ALLOC) || defined(INTERNAL_MEM_PROTECT_HEAP_ALLOC)
+HeapAllocator::HeapAllocator(bool useAllocMemProtect)
+    : m_privateHeap(nullptr)
```
dilijev: @dreamer-dead I now initialize m_privateHeap to nullptr correctly in the constructor.

dilijev:
@curtisman @digitalinfinity @tcare @Yongqu When ready, can one of you sign off on this PR?

curtisman:
LGTM

PR #407, opened 2016-02-27
add br_if, select, and memory export to wasm

MikeHolman:
[Review on Reviewable](https://reviewable.io/reviews/microsoft/chakracore/407)


MikeHolman:
@Cellule @abchatra @LouisLaf, can you please review?

Also, note that exporting as "memory" is a temporary attribute, which is likely to be superseded

LouisLaf:
LGTM

PR #408, opened 2016-02-29
Get GCStress building on Linux

digitalinfinity:
This change gets GCStress building and running on Linux. 
The harness doesn't pass yet- we segfault in the GC but the change is large enough already that I wanted to merge it in before proceeding further.
Since it's a large change, I've broken it up into the following 7 commits. They're mostly accurate but sometimes changes that should be in a particular commit end up in a later commit.

1. Update CMakeLists.txt
2. Add more #defines to control features
3. Add more of the PAL to build GCStress
4. Add amd64_save_registers.S
5. Added macro FLAG_STRING to ConfigFlagsList.
6. Build script for Unix platforms
7. Convert strings to use CH_WSTR

Details of changes:

**1. Update CMakeLists.txt**
Update CMakeLists with the following changes:
1) Add support for Debug Builds and building ASM files
2) Disable the JIT in linux builds explicitly
3) Remove dependency on PAL_STDCPP_COMPAT

**2. Add more #defines to control features**
Added a couple more ifdefs to control whether features are compiled in or not (they're not on linux). The new compile definitions are the following:
CONFIG_CONSOLE_AVAILABLE - Whether -Console config option is available
CONFIG_PARSE_CONFIG_FILE - Whether jscript.config should be parsed or not.
CONFIG_RICH_TRACE_FORMAT - Whether -RichTraceFormat config option is available
SYSINFO_IMAGE_BASE_AVAILABLE - Whether the linker emits the __ImageBase global variable
ENABLE_BACKGROUND_PAGE_FREEING - Whether the Page allocator supports freeing pages on a background thread
ENABLE_BACKGROUND_PAGE_ZEROING - Whether the Page allocator supports zeroing pages on a background thread
ENABLE_BACKGROUND_JOB_PROCESSOR - Whether the BackgroundJobProcessor is enabled
ENABLE_RECYCLER_TYPE_TRACKING - Whether the recycler keeps track of the type of objects allocated (for debugging/profiling)

The following existing #ifdefs were also disabled on Linux builds:
```
FAULT_INJECTION
GENERATE_DUMP
PROFILE_EXEC
PROFILE_RECYCLER_ALLOC
PROFILE_DICTIONARY
PDATA_ENABLED
HEAP_TRACK_ALLOC
```

I also cleaned up the STACK_BACK_TRACE macro to be used to disable more code that depended on capturing stack traces.

Finally, fixed some build errors on DEBUG builds

**3. Add more of the PAL to build GCStress**
This change brings in more of the PAL so that GCStress can build. Added a call to initialize the PAL runtime in AutoSystemInfo (the shutdown call is in main).
Added a #define to define priority for our static globals to simulate VC++'s #pragma init_seg

The PAL change also includes a dependency on libsodium for cryptographically secure random number generation- to install the package, run the following commands:
```
sudo apt-get install libsodium13
sudo apt-get install libsodium-dev
```

Specific comments on the PAL changes:
Bought in a large portion of the PAL + some AMD64 specific pieces
Added a new PAL_GLOBAL macro and annotated a bunch of statics/globals in the following files to ensure that they get initialized before our static variables:
```
cruntime\file.cpp
file\file.cpp
handlemgr\handleapi.cpp
init\pal.cpp
map\map.cpp, virtual.cpp
misc\dbgmsg.cpp
objmgr\shmobjectmanager.cpp
shmemory\shmemory.cpp
synchmgr\synchmanager.cpp
synchobj\mutex.cpp
thread\process.cpp, thread.cpp

```
Added in implementation for rand_s in misc\Random.cpp. This has a dependency on libsodium.

Other minor code changes in following files:
cruntime\file.cpp - remove unused headers
exception\seh-unwind.cpp - remove unused headers
file\file.cpp - Change call from GetFileAttributes to GetFileAttributesA in case where we were passing in an ANSI string
include\pal\misc.h - Add declaration to initialize the random number subsystem
init\pal.cpp - Rename PAL_InitializeCoreCLR to PAL_InitializeChakraCore
locale\unicode.cpp - Remove PAL_BindResources API
misc\sysinfo.cpp - Hard-code process max address space to 128 TB
safecrt\vsprintf.c - add implementation for vsprintf_s
synchmgr\synchmanager.cpp - remove non-coreclr code
thread\process.cpp - Remove support for running PE files in CreateProcess

**4. Add amd64_save_registers.S**
Added amd64_save_registers.S which reimplements amd64_save_registers.asm for the clang assembler.
The primary differences here are the following:
- Linux assembler syntax
- The Linux x64 ABI has the first parameter passed in in $rdi so this implementation uses that

**5. Added macro FLAG_STRING to ConfigFlagsList.**
Added a new FLAG_STRING macro. These are specifically for ConfigFlags with string default values, so that we can use the correct string type depending on the platform.

**6. Build script for Unix platforms**
Added a simple shell script to build on non-Windows platform.
./build.sh builds retail by default
./build.sh -d builds debug

**7. Convert strings to use CH_WSTR**
This change converts some more code to use 2-byte Unicode strings on Linux.
The subset of files in this commit are the files needed to get GCStress to build.


lib/Common/Memory/CommonMemoryPch.h
```diff
@@ -4,11 +4,23 @@
...
+// In Debug mode, the PALs definition of max and min are insufficient
+// since some of our code expects the template min-max instead, so
+// including that here
+#if defined(DBG) && !defined(_MSC_VER)
+#define NO_PAL_MINMAX
+#define _Post_equal_to_(x)
+#define _Post_satisfies_(x)
+
+#include "Core/CommonMinMax.h"
+#endif
```
jianchun: Can this section be moved to shared place, e.g. CommonMinMemory.h?

lib/Common/Common/Jobs.cpp
```diff
@@ -464,6 +465,9 @@ namespace JsUtil
         // Do nothing
     }
 
+// Xplat-todo: revive BackgroundJobProcessor- we need this for the JIT
+#if ENABLE_BACKGROUND_JOB_PROCESSOR
```
jianchun: nit: Hard to find out where this #if ends. Add to #endif // ENABLE_BACKGROUND_JOB_PROCESSOR

lib/Common/ConfigFlagsList.h
```diff
@@ -825,7 +846,7 @@ FLAGNR(Boolean, NoDynamicProfileInMemoryCache, "Enable in-memory cache for dynam
 FLAGNR(Boolean, ProfileBasedSpeculativeJit, "Enable dynamic profile based speculative JIT", DEFAULT_CONFIG_ProfileBasedSpeculativeJit)
 FLAGNR(Number,  ProfileBasedSpeculationCap, "In the presence of dynamic profile speculative JIT is capped to this many bytecode instructions", DEFAULT_CONFIG_ProfileBasedSpeculationCap)
 #ifdef DYNAMIC_PROFILE_MUTATOR
-FLAGNR(String,  DynamicProfileMutatorDll , "Path of the mutator DLL", L"DynamicProfileMutatorImpl.dll")
+FLAGNR_STRING(DynamicProfileMutatorDll , "Path of the mutator DLL", "DynamicProfileMutatorImpl.dll")
```
jianchun: Didn't get why you need this change... Why didn't it suffice by wrapping the default value as ```CH_WSTR("DynamicProfileMutatorImpl.dll")```?

jianchun:
LGTM!  :thumbsup:

lib/Common/Memory/CommonMemoryPch.h
digitalinfinity: Done

lib/Common/Common/Jobs.cpp
digitalinfinity: Done

lib/Common/ConfigFlagsList.h
digitalinfinity: That's what I'd initially done but then decided that FLAG_STRING looked a little neater that having CH_WSTR interspersed through that file. But since FLAG_STRING is inconsistent with the rest of the file having FLAGs taking in a type and there are only few places where we need CH_WSTR, I got rid of that commit and reverted back to FLAG.

PR #409, opened 2016-02-29
fix asm.js issue where fixnum was causing us to use uint div/mod

MikeHolman:
fixnum can be used as either signed or unsigned, so we need to look at both lhs and rhs to infer the correct context for binary operators.

[Review on Reviewable](https://reviewable.io/reviews/microsoft/chakracore/409)

### DIFF
--- a/lib/Runtime/Language/AsmJsByteCodeGenerator.cpp
+++ b/lib/Runtime/Language/AsmJsByteCodeGenerator.cpp
@@ -663,7 +663,8 @@ namespace Js
         {
             CheckNodeLocation( lhsEmit, int );
             CheckNodeLocation( rhsEmit, int );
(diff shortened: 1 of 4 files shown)


MikeHolman:
@Cellule @abchatra, can you please review?

Cellule:
The only thing I can see is that 
```js
  function f() {
    return 10 / 3 | 0;
  }
```
Currently does an unsigned division and with this change it will do a signed division.
I doubt it make a difference since both operation should be the same for `Fixnum`s  and this pattern is most likely not produced by emscripten.
I just wanted to point it out in case someone sees something I missed with this case.
Otherwise LGTM

MikeHolman:
I considered this, but fixnum should be ok to use either signed or unsigned division, since it is guaranteed to be in range [0, 2^31). Part of the idea of fixnum is that it doesn't matter.

PR #411, opened 2016-02-29
in asm.js ! operator should give type int, not signed

MikeHolman:
This is a bug which gives semantic difference between asm.js and normal js in case of return or ffi calls, where js semantics it should be `true` or `false`, but we were giving `1` or `0`. Correct result is that asm.js should not validate these cases.

[Review on Reviewable](https://reviewable.io/reviews/microsoft/chakracore/411)


MikeHolman:
@Cellule @abchatra: can you please review?

Cellule:
LGTM

PR #413, opened 2016-02-29
Copyright check: use python 3.x print function to make the script independent of the version of python installed.

dilijev:


goyakin:
LGTM

PR #414, opened 2016-02-29
Copyright check: add logic to handle empty files and files less than 3 lines.

dilijev:


goyakin:
@dilijev Did you consider using the [`zip`](https://docs.python.org/3/library/functions.html#zip) function instead? I think we could do:

```Python
linecount = 0
with open(file_name, 'r') as sourcefile:
    for (copyright_line, line) in zip(copyright_lines, sourcefile):
        linecount += 1
        # ...
if linecount == 0:
    # PASS
elif linecount != len(copyright_lines):
    # FAIL
```

dilijev:
@goyakin possible, but I don't want to step on PR #367 so I'll wait until that gets merged in to make the change you're suggesting and use `regexes` instead of copyright lines, in line with that change.

dilijev:
Alright, I made a few modifications based on @goyakin's suggestion. I also wanted to make the output easier to read.

The output will now look like this:

```
$ sh jenkins.check_copyright.sh
Check Copyright > Begin Checking...
Check Copyright > Checking jenkins.check_copyright.py
--------------
--- ERRORS ---
jenkins.check_copyright.py ... does not contain a correct Microsoft copyright notice.

     # icensed under the MIT license. See LICENSE.txt file in the project root for full license information.

--------------
```

I tested that this script works on the four scenarios:

* blank file (success)
* too short (failure)
* long enough but incorrect (failure)
* long enough and successful (success)

jenkins.check_copyright.py
```diff
@@ -30,13 +31,31 @@
     print("File does not exist:", file_name, "(not necessarily an error)")
     exit(0)
 
+def report_incorrect(file_name, pairs):
+    # found a problem so report the problem to the caller and exit
+    print(file_name, "... does not contain a correct Microsoft copyright notice.")
+    print()
+    # print the relevant lines to help the reader find the problem
+    for (regex, line) in pairs:
```
goyakin: Since you aren't using the `regex` variable, you can use `_` to indicate that.

jenkins.check_copyright.py
```diff
@@ -30,13 +31,31 @@
...
+        print("    ", line, end="")
+    print()
+    exit(1)
```
goyakin: I would move the call to `exit` outside this function since the purpose here is the report. That way, one won't need to read this function when he's looking at the main algorithm of the script.

jenkins.check_copyright.py
```diff
@@ -30,13 +31,31 @@
...
-        # TODO add a check for empty files (dummy.js etc), as they cause the script to crash here
-        line = next(sourcefile)
-        line = line.rstrip()
-        matches = regex.match(line)
-        if not matches:
-            print(file_name, "... does not contain a correct Microsoft copyright notice.")
-            # found a problem so exit and report the problem to the caller
-            exit(1)
+    for item in zip(regexes, sourcefile):
+        pairs.append(item)
```
goyakin: You can do `pairs += zip(regexes, sourcefile)` instead.

jenkins.check_copyright.py
dilijev: Good point.

jenkins.check_copyright.py
dilijev: Ok, good idea.

jenkins.check_copyright.py
dilijev: Oh! That's cool!

dilijev:
@goyakin How does it look now?

goyakin:
LGTM :smile:

PR #415, opened 2016-02-29
Destructuring redeclaration error under deferred mode.

akroshg:
In the deferparse mode we throw an error if the body has a variable which has the same name as formal’s. As the logic was under the buildAST check.
However we need to mark the formal as STFormal regardless of any mode. So we are walking the blockScope lex vars, when they appear at formals and change them appropriately.


akroshg:
@tcare - please take a look.

akroshg:
@aneeshdk - please take a look as well.

tcare:
Looks good

PR #416, opened 2016-02-29
Add RegExp.prototype[@@split]

goyakin:


goyakin:
@abchatra & @akroshg, could you take a look?

lib/Runtime/Library/JavascriptRegularExpression.cpp
```diff
@@ -765,11 +817,46 @@ namespace Js
...
+        // if the algorithm is observable. If it is, we go through the new ES6 algorithm, but otherwise, we
+        // run the faster ES5 version.
+        //
+        // According to the spec, we're supposed to process "limit" after we use some of the RegExp properties.
+        // However, there doesn't seem to be any reason why "limit" processing can't be pulled above the rest
+        // in the spec. Therefore, we should see if such a spec update is OK. If not, this would have to be
+        // moved to its correct place in the code.
+        uint32 limit = (args.Info.Count < 3 || JavascriptOperators::IsUndefinedObject(args[2], scriptContext))
+            ? UINT_MAX
+            : JavascriptConversion::ToUInt32(args[2], scriptContext);
```
goyakin: @bterlson This is what I talked to you about regarding where the `limit` argument is processed.

goyakin:
Ping...

lib/Runtime/Base/ScriptContext.cpp
```diff
@@ -850,17 +850,12 @@ namespace Js
         this->weakReferenceDictionaryList.Prepend(this->GeneralAllocator(), weakReferenceDictionary);
     }
 
-    RecyclableObject *ScriptContext::GetMissingPropertyResult(Js::RecyclableObject *instance, Js::PropertyId id)
+    RecyclableObject *ScriptContext::GetMissingPropertyResult()
```
akroshg: This is good.
While you are here - could you check and remove unwanted params for GetNullPropertyResult and GetNullItemResult below?

lib/Runtime/Library/JavascriptLibrary.cpp
```diff
@@ -3890,30 +3893,26 @@ namespace Js
 
         if (scriptConfig->IsES6RegExPrototypePropertiesEnabled())
         {
-            RuntimeFunction* globalGetter = library->CreateGetterFunction(PropertyIds::global, &JavascriptRegExp::EntryInfo::GetterGlobal);
-            library->AddAccessorsToLibraryObject(regexPrototype, PropertyIds::global, globalGetter, nullptr);
-            library->regexGlobalGetterFunction = globalGetter;
-
+            library->regexGlobalGetterFunction =
```
akroshg: Current function InitializeRegexPrototype should have passed right initSlotCapacity (typeHandler->Convert(, , 8) )?

lib/Runtime/Library/RegexHelper.cpp
```diff
@@ -1213,7 +1212,156 @@ namespace Js
     }
 
     // String.prototype.split (ES5 15.5.4.14)
-    Var RegexHelper::RegexSplitImpl(ScriptContext* scriptContext, JavascriptRegExp* regularExpression, JavascriptString* input, CharCount limit, bool noResult, void *const stackAllocationPointer)
+    Var RegexHelper::RegexSplitImpl(ScriptContext* scriptContext, RecyclableObject* thisObj, JavascriptString* input, CharCount limit, bool noResult, void *const stackAllocationPointer)
```
akroshg: is above spec reference valid?

test/es6/regex-symbols.js
```diff
@@ -600,9 +663,184 @@ var tests = [
...
+    },
+    {
+        name: "RegExp.prototype[@@split] should return an empty Array when the input size is 0 and the RegExp matches the empty string",
+        body: function () {
+            function assertResult(result) {
+                assert.areEqual(0, result.length, "result.length");
+            }
+            var re = /(?:)/;
+            var input = "";
+            verifySymbolSplitResult(assertResult, re, input);
```
akroshg: Did you have case where input and regex both are empty string ?

lib/Runtime/Library/RegexHelper.cpp
```diff
@@ -1813,17 +1960,24 @@ namespace Js
...
+    CharCount RegexHelper::AdvanceStringIndex(JavascriptString* string, CharCount index, bool isUnicode)
     {
         // TODO: Change the increment to 2 depending on the "unicode" flag and
         // the code point at "index". The increment is currently constant at 1
         // in order to be compatible with the rest of the RegExp code.
-        return index + 1;
+
+        if (index + 1 < index) // Overflow
+        {
+            return MaxCharCount;
```
akroshg: we should throw here - Js::Throw::OutOfMemory();

test/es6/regex-symbols.js
akroshg: Ok I meant new RegExp("")[Symbol.split](""), which is the same above case.

---
In reply to: [54801523](https://github.com/Microsoft/ChakraCore/pull/416#discussion_r54801523) [](ancestors = 54801523)

akroshg:
Looks good to me.

lib/Runtime/Base/ScriptContext.cpp
goyakin: It turns out we don't use these methods at all, so I removed them completely.

lib/Runtime/Library/JavascriptLibrary.cpp
goyakin: You're right, the initial slot capacity is wrong here. I changed it to `16` though, including the properties behind the `ES6RegExPrototypeProperties` flag.

lib/Runtime/Library/RegexHelper.cpp
goyakin: I created a new function called `RegexEs5SplitImpl` for the ES5 version, but looks like I forgot to move the spec reference :smile:

lib/Runtime/Library/RegexHelper.cpp
goyakin: I think it's better to cap the index at `MaxCharCount` instead. The new value is used for `lastIndex` and the next call to `exec` would result in **no** match, which would be the same as if we don't cap it.

goyakin:

goyakin:
Thanks @akroshg for the review.

PR #417, opened 2016-03-01
Fix SIMD alignment AV on Dev12 build.

nmostafa:
This fix separates the two SIMDValue structs.
Fix AV on Dev12 build. For SIMD value, we use a generic value (_SIMDValue) and an IA specific value with an intrinsic variable (_x86_SIMDValue). Initially, we had the IA specific one include the _SIMDValue struct. This causes the compiler to copy-prop pointers to _SIMDValue to _x86_SIMDValue, even though alignment is not guaranteed. For this particular case, it was when passing a parameter by reference to a SIMD op handler. A pointer to the stack slot in the caller holding the _SIMDValue is passed, it gets copied inside the hanlder to a _x86_SIMDValue. The compiler decides to eliminate the copy, and reference the pointer to the stack directly which causes hardware exception if it happens not to be aligned.

### DIFF
--- a/lib/Runtime/Language/SimdUtils.h
+++ b/lib/Runtime/Language/SimdUtils.h
@@ -18,19 +18,20 @@
 #define INT8_SIZE    1
 #define SIMD_INDEX_VALUE_MAX     5
 #define SIMD_STRING_BUFFER_MAX   1024
-
+#define SIMD_DATA 	\
+	int32   i32[4];\
+	int16   i16[8];\
+	int8    i8[16];\
+	uint32  u32[4];\
+	uint16  u16[8];\
+	uint8   u8[16];\
+	float   f32[4];\
+	double  f64[2];
 
 struct _SIMDValue
 {
     union{
-        int32   i32[4];
-        int16   i16[8];
-        int8    i8[16];
-        uint32  u32[4];
-        uint16  u16[8];
-        uint8   u8[16];
-        float   f32[4];
-        double  f64[2];
+		SIMD_DATA
     };
 
     void SetValue(_SIMDValue value)
@@ -79,7 +80,7 @@ struct DefaultComparer<_SIMDValue>
 struct _x86_SIMDValue
 {
     union{
-        _SIMDValue simdValue;
+		SIMD_DATA
         __m128  m128_value;
         __m128d m128d_value;
         __m128i m128i_value;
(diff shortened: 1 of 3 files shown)


nmostafa:
@dilijev @MikeHolman, please review.

dilijev:
@dotnet-bot test Windows daily_dev12_x64_test please

dilijev:
@dotnet-bot test Windows 7 daily_dev12_x64_test please

MikeHolman:
Interesting. I don't completely understand the VCs behavior here, but your change LGTM.

dilijev:
Whoops, there's a bug in the trigger regex that caused the full legacy set to be triggered. This is only a problem because the expected failure in daily_dev12_x64_debug will cause the checks for this change to fail.

I'll fix this ASAP and get it merged in. Then I will close and reopen this PR to get rid of the unnecessary checks and re-run the one we're interested in.

nmostafa:
@MikeHolman , here is what happens in more details. When handling this opcode:
`DEF2_WMS( SIMD_F4_2toF4_1  , Simd128_And_F4              , Js::SIMDFloat32x4Operation::OpAnd           )`
We call GetRegRawSIMD on the operands. The values returned are SIMDValues, and store in the callers stack framer. It is then passed by reference to the OpAnd handler:

```
    SIMDValue SIMDFloat32x4Operation::OpAnd(const SIMDValue& aValue, const SIMDValue& bValue)
    {
        X86SIMDValue x86Result;
        X86SIMDValue tmpaValue = X86SIMDValue::ToX86SIMDValue(aValue);
        X86SIMDValue tmpbValue = X86SIMDValue::ToX86SIMDValue(bValue);

        x86Result.m128_value = _mm_and_ps(tmpaValue.m128_value, tmpbValue.m128_value); // a & b

        return X86SIMDValue::ToSIMDValue(x86Result);
    }
```

The handler converts the value to an aligned value (X86SIMDValue), by copying it over. That way and_ps deals with an aligned value. VC decides to omit the copy code, and access the passed reference directly (a pointer to the caller stack slow, and is not necessarily aligned). Apparently VC decides to do this because one struct is subset of the other.

lib/Runtime/Language/SimdUtils.h
```diff
@@ -18,19 +18,20 @@
...
-
+#define SIMD_DATA 	\
+	int32   i32[4];\
+	int16   i16[8];\
+	int8    i8[16];\
+	uint32  u32[4];\
+	uint16  u16[8];\
+	uint8   u8[16];\
+	float   f32[4];\
+	double  f64[2];
```
dilijev: Please use spaces instead of tabs, here and elsewhere in this change.

dilijev:
Otherwise LGTM

dilijev:
LGTM now. Just waiting to see the results of http://dotnet-ci.cloudapp.net/job/Microsoft_ChakraCore/job/daily_dev12_x64_test_prtest/3/ before signing off.

dilijev:
That check passed. Signing off.

PR #422, opened 2016-03-01
Symbol.toString returns invalid string.

akroshg:
Call to JavascriptSymbol::ToString should throw an error - but we have a check for disabledimplicit call which return nullptr which leads to AV. Fixing that by checking RecordImplicitException. Now we will either throw or return an empty string.

### DIFF
--- a/test/Bugs/symbol_tostring.js
+++ b/test/Bugs/symbol_tostring.js
@@ -1,0 +1,20 @@
+//-------------------------------------------------------------------------------------------------------
+// Copyright (C) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE.txt file in the project root for full license information.
+//-------------------------------------------------------------------------------------------------------
+
+var k;
+function test0() {
+  try {
+    // Below will throw an error and we will try to do implicit toString call on Symbol which will also throw.
+    // Under disableImplicit call we were returning nullptr which was wrong.
+    [] = ab[Object(Symbol())] = null;
+    } catch (e) {
+  }
+  var ab = '';
+}
+test0();
+test0();
+test0();
+
+print("Pass");
--- a/lib/Runtime/Library/JavascriptSymbol.cpp
+++ b/lib/Runtime/Library/JavascriptSymbol.cpp
@@ -287,16 +287,12 @@ namespace Js
 
     JavascriptString* JavascriptSymbol::ToString(ScriptContext * requestContext)
     {
-        // Reject implicit call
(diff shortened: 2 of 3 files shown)


akroshg:
@abchatra @rajatd - please a take a look.

rajatd:
:+1:

test/Bugs/symbol_tostring.js
```diff
@@ -0,0 +1,20 @@
...
+// Copyright (C) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE.txt file in the project root for full license information.
+//-------------------------------------------------------------------------------------------------------
+
+var k;
+function test0() {
+  try {
+    // Below will throw an error and we will try to do implicit toString call on Symbol which will also throw.
+    // Under disableImplicit call we were returning nullptr which was wrong.
+    [] = ab[Object(Symbol())] = null;
```
ljharb: wait, this shouldn't throw an error - using a Symbol object as an object key is the same as using a Symbol primitive.

ljharb:
Just to clarify: `String(Symbol())` should work, and `String(Object(Symbol()))` and `Symbol().toString()` should throw?

test/Bugs/symbol_tostring.js
caitp: This doesn't seem to be true per https://tc39.github.io/ecma262/#sec-topropertykey / https://tc39.github.io/ecma262/#sec-symbol.prototype-@@toprimitive

This will just put a property with name `Symbol()` (type Symbol) in the object. This code should probably still throw, but not for the same reason

caitp:
> and String(Object(Symbol())) and Symbol().toString() should throw?

@ljharb no, but `"" + Symbol()` or `"" + Object(Symbol())` will throw.

ljharb:
aha, thanks

akroshg:
@ianwjhalliday 
>> I see, so the repro is to have an error thrown that causes an object wrapped symbol to be converted to a string for the error information (aside: what for exactly? the message string?).

yes - the message string.
Just to clarify 
```
var ab = []
ab[Object(Symbol())] = 1; 
```
will not do implicit call to toString.

I am filing an issue on Symbol expert @boingoing to revisit if the throw code path should also call toString.

Regarding the test case. I would have removed `[] = `if that was not needed :), our internal reducer did a nice job on giving me this repro. 
And yes `var ab = undefined `can be put in the beginning and I somehow missed that.

PR #423, opened 2016-03-01
linux: fix build

xanlpz:
commit 29bbe86ffa0912fc9c45 broke the build on some GNU/Linux
configurations, since CMAKE_CXX_FLAGS is not always used. Use
add_definitions() again for the flags we always need and use the
preferred way of selection the C++11 standard.


digitalinfinity:
LGTM- signing off

dilijev:
@dotnet-bot test this please

obastemur:
See https://github.com/Microsoft/ChakraCore/pull/438 before merging this. In case #438 is good to go, lets remove all the references etc. to `libsodium`

PR #428, opened 2016-03-02
Decoding offset based on alignment offset bit

naturaltransformation:
In the binary format, the offset may or may not be present as specified
by the alignment bitfield.


dilijev:
@naturaltransformation Why did you decide to close this PR?

commonlisp:
Replaced with pr#429 submitted under a handle with CLA.

PR #429, opened 2016-03-02
Decoding offset based on alignment offset bit

commonlisp:
In the binary format, the offset may or may not be present as specified
by the alignment bitfield.


MikeHolman:
lgtm

PR #430, opened 2016-03-02
Wasm.instantiateModule()

Cellule:
Added `WScript.LoadBinaryFile` that returns an ArrayBuffer with the bytes. 
Added a `Wasm` object on the global object with `instantiateModule` method to create a module from an ArrayBuffer


Cellule:
@dotnet-bot test this please

lib/Runtime/Library/WasmLibrary.cpp
```diff
@@ -0,0 +1,74 @@
+//-------------------------------------------------------------------------------------------------------
+// Copyright (C) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE.txt file in the project root for full license information.
+//-------------------------------------------------------------------------------------------------------
+#include "RuntimeLibraryPch.h"
+
+#ifdef ENABLE_WASM
+namespace Js
```
MikeHolman: would prefer to use the Wasm namespace here

MikeHolman:
lgtm

lib/Runtime/Library/WasmLibrary.cpp
Cellule: I don't think I can, the macro
```
BUILTIN(Wasm::WasmLibrary, instantiateModule, instantiateModule, FunctionInfo::ErrorOnNew)
```
doesn't accept the `Wasm` namespace because it uses this to make an enum and without the namespace it doesn't find the class.
I could do a `using namespace Wasm;` where the macro is used, but I doubt this is a very good idea

PR #432, opened 2016-03-03
module runtime change #3

Yongqu:
Implement the remaining operations in SourceTextModuleRecord, including   mainly ResolveExport and GetExportNames operations. Start implementation of ModuleNamespace. Implemented GetModuleNamespace. Detailed   namespace implementation is not in yet. Filled in some holes in earlier implementation, like error handling for top level operations,   notify the parent module if the current module failed in parse or ModuleDeclarationInitialization.   Export method for bytecode generator to retrieve the localexport slot information.   Fixed some potential memory leak by implementing the Finalize method for SourceTextModuleRecord. Temporary disable ModuleEvaluation for non-root modules, pending bytecode generator change.


Yongqu:
@pleath @boingoing @abchatra can you take a look?

lib/Runtime/Language/SourceTextModuleRecord.cpp
```diff
@@ -165,25 +217,245 @@ namespace Js
...
+        {
+            if (localNames == nullptr)
+            {
+                localNames = (ExportedNames*)AllocatorNew(ArenaAllocator, allocator, ExportedNames, allocator);
+            }
+            this->indirectExportRecordList->Map([=](ModuleExportEntry exportEntry) {
+                PropertyId exportedNameId = EnsurePropertyIdForIdentifier(exportEntry.exportName);
+                localNames->Prepend(exportedNameId);
+            });
+        }
```
boingoing: Do you feel strongly about merging these three lists of exports (indirectExportRecordList, localExportRecordList, starExportRecordList) ? Should be possible to determine which type of export without keeping them in different lists if duplicating this code becomes troublesome.

lib/Runtime/Language/SourceTextModuleRecord.cpp
```diff
@@ -165,25 +217,245 @@ namespace Js
...
+
+        exportStarSet->Prepend(this);
+        bool ambiguousResolution = false;
+        if (this->starExportRecordList != nullptr)
+        {
+            ModuleNameRecord* starResolution = nullptr;
+            starExportRecordList->MapUntil([&](ModuleExportEntry starExportEntry) {
+                ModuleNameRecord* currentResolution = nullptr;
+                SourceTextModuleRecord* childModule;
+                if (!childrenModuleSet->TryGetValue(starExportEntry.moduleRequest->Psz(), &childModule))
```
boingoing: Use this->GetChildModuleRecord() instead?

lib/Runtime/Language/SourceTextModuleRecord.cpp
```diff
@@ -312,6 +587,10 @@ namespace Js
                 }
             });
         }
+        CleanupBeforeExecution();
+        // BUGBUG: get rid of this after bytecode gen change.
+        if (!isRootModule) return scriptContext->GetLibrary()->GetUndefined();
```
boingoing: Will do

lib/Runtime/Language/SourceTextModuleRecord.cpp
```diff
@@ -333,6 +612,21 @@ namespace Js
         return NOERROR;
     }
 
+    SourceTextModuleRecord* SourceTextModuleRecord::GetChildModuleRecord(LPCOLESTR specifier) const
+    {
+        SourceTextModuleRecord* childModuleRecord = nullptr;
+        if (childrenModuleSet == nullptr)
+        {
+            AssertMsg(false, "there should have some child modulerecords first before trying to get child modulerecord.");
```
boingoing: Maybe "There should have been some... " or "We should have some... "

boingoing:
Looks good to me! Big change to module record, almost support everything besides namespace object. Cool!

lib/Runtime/Language/SourceTextModuleRecord.cpp
Yongqu: {AncestorPath = 54950339}
currently I'm not copying those lists, and it's better to keep them separate for different operations. We can change if perf becomes an issue here.

lib/Runtime/Language/ModuleNamespace.h
```diff
@@ -6,14 +6,78 @@
...
+        ModuleNamespace(ModuleRecordBase* moduleRecord, DynamicType * type);
+        static ModuleNamespace* New(ModuleRecordBase* moduleRecord);
     public:
-        ModuleNamespace();
-    private:
+        static ModuleNamespace* GetModuleNamespace(ModuleRecordBase* moduleRecord);
+
+        static bool Is(Var ns) { return ModuleNamespace::Is(ns); }
 
+        //virtual int GetPropertyCount() override;
```
pleath: Are you keeping the commented lines on purpose?

lib/Runtime/Language/SourceTextModuleRecord.cpp
```diff
@@ -37,16 +41,38 @@ namespace Js
...
         // The life time of a module record should be controlled by the module registry as defined in WHATWG module loader spec
         // in practice the modulerecord lifetime should be the same as the scriptcontext so it could be retrieved for the same
-        // site. DOM might hold a reference to the module as well after the initialize the module. 
+        // site. Host might hold a reference to the module as well after initializing the module. 
         // In our implementation, we'll use the moduleId in bytecode to identify the module.
         childModuleRecord->moduleId = scriptContext->GetLibrary()->EnsureModuleRecordList()->Add(childModuleRecord);
         return childModuleRecord;
     }
 
+    void SourceTextModuleRecord::Finalize(bool isShutdown)
```
pleath: Just for my info, what does "SourceTextModuleRecord" refer to?

lib/Runtime/Language/SourceTextModuleRecord.cpp
```diff
@@ -165,25 +217,245 @@ namespace Js
...
+            return true;
         }
-        Assert(false);
-        return false;
+        resolveSet->Prepend({ this, exportName });
+
+        if (localExportRecordList != nullptr)
+        {
+            PropertyId localNameId = Js::Constants::NoProperty;
+            localExportRecordList->MapUntil([&](ModuleExportEntry exportEntry) {
```
pleath: Does this walk the whole set of exports each time?

lib/Runtime/Language/ModuleNamespace.h
Yongqu: These are unimplemented feature. We'll uncomment these in upcoming changes.

---
In reply to: [55074978](https://github.com/Microsoft/ChakraCore/pull/432#discussion_r55074978) [](ancestors = 55074978)

lib/Runtime/Language/SourceTextModuleRecord.cpp
Yongqu: [https://tc39.github.io/ecma262/#sec-source-text-module-records](https://tc39.github.io/ecma262/#sec-source-text-module-records)

---
In reply to: [55075088](https://github.com/Microsoft/ChakraCore/pull/432#discussion_r55075088) [](ancestors = 55075088)

lib/Runtime/Language/SourceTextModuleRecord.cpp
Yongqu: Once every new resolution. I cache the resolved result.
I was thinking about creating additional dictionary for all different lists, but I'm leaving it like this for now before we have more real cases. It'll be a problem if there are a lot of exports from one module, and I hope the "interface", ie. the exported entries, are more concise.

---
In reply to: [55075711](https://github.com/Microsoft/ChakraCore/pull/432#discussion_r55075711) [](ancestors = 55075711)

pleath: Do you foresee a problem doing this as a dictionary search?

Yongqu: No but it might be better to come in from a later change.

pleath: Fair enough. LGTM.

--Paul

PR #436, opened 2016-03-03
Fix GCStress on Linux

digitalinfinity:
Complete the port of GCStress to linux.
1. Fix bugs in implementation of BitTestAndSet etc.
2. Support passing in command line flags to GCStress
3. Use sigtrap for assertion failures to break into both lldb and gdb
4. Lower threshold of initial object count/op count to show progress
   on my VM. We'll have a future investigation item to understand what
   operations are slower on linux and how they can be sped up.

### DIFF
--- a/lib/Common/Memory/Recycler.cpp
+++ b/lib/Common/Memory/Recycler.cpp
@@ -324,7 +324,7 @@ Recycler::Recycler(AllocationPolicyManager * policyManager, IdleDecommitPageAllo
 
 #ifdef ENABLE_DEBUG_CONFIG_OPTIONS
     // recycler requires at least Recycler::PrimaryMarkStackReservedPageCount to function properly for the main mark context
-    this->markContext.SetMaxPageCount(max(GetRecyclerFlagsTable().MaxMarkStackPageCount, Recycler::PrimaryMarkStackReservedPageCount));
+    this->markContext.SetMaxPageCount(max(static_cast<size_t>(GetRecyclerFlagsTable().MaxMarkStackPageCount), static_cast<size_t>(Recycler::PrimaryMarkStackReservedPageCount)));
     this->parallelMarkContext1.SetMaxPageCount(GetRecyclerFlagsTable().MaxMarkStackPageCount);
     this->parallelMarkContext2.SetMaxPageCount(GetRecyclerFlagsTable().MaxMarkStackPageCount);
     this->parallelMarkContext3.SetMaxPageCount(GetRecyclerFlagsTable().MaxMarkStackPageCount);
--- a/pal/inc/pal.h
+++ b/pal/inc/pal.h
@@ -5634,6 +5634,9 @@ InterlockedOr(
     return __sync_fetch_and_or(Destination, Value);
 }
 
+#define BITS_IN_BYTE 8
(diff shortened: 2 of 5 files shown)


digitalinfinity:
@curtisman @jianchun please review. @xanlpz this has some overlap with some of your changes- I'll merge your change first before merging this commit in.

lib/Common/Memory/Recycler.cpp
```diff
@@ -324,7 +324,7 @@ Recycler::Recycler(AllocationPolicyManager * policyManager, IdleDecommitPageAllo
 
 #ifdef ENABLE_DEBUG_CONFIG_OPTIONS
     // recycler requires at least Recycler::PrimaryMarkStackReservedPageCount to function properly for the main mark context
-    this->markContext.SetMaxPageCount(max(GetRecyclerFlagsTable().MaxMarkStackPageCount, Recycler::PrimaryMarkStackReservedPageCount));
+    this->markContext.SetMaxPageCount(max(static_cast<size_t>(GetRecyclerFlagsTable().MaxMarkStackPageCount), (size_t) Recycler::PrimaryMarkStackReservedPageCount));
```
jianchun: nit: inconsistent cast style

pal/inc/pal.h
```diff
@@ -5643,7 +5646,9 @@ InterlockedBitTestAndReset(
     IN OUT LONG volatile *Base,
     IN LONG Bit)
 {
-    return (InterlockedAnd(Base, ~(1 << Bit)) & (1 << Bit)) != 0;
+    volatile LONG* longToTest = Base + (Bit / BITS_IN_LONG);
+    LONG bitToTest  = Bit % BITS_IN_LONG;
+    return (InterlockedAnd(longToTest, ~(1 << bitToTest)) & (1 << bitToTest)) != 0;
```
jianchun: Add a comment somewhere to explain these are to support "Bit > bits in LONG"?

pal/inc/pal.h
```diff
@@ -5681,7 +5688,14 @@ BitTestAndSet(
     IN OUT LONG *Base,
     IN LONG Bit)
 {
-    return (__sync_fetch_and_or(Base, (1 << Bit)) & (1 << Bit)) != 0;
+    // xplat-todo: Benchmark whether this is better than
+    // using the inline assembler to generate a bts instruction
+    LONG* longToTest = Base + (Bit / BITS_IN_LONG);
+    LONG bitToTest  = Bit % BITS_IN_LONG;
+    
+    unsigned char res = (*longToTest >> bitToTest) & 1;
```
jianchun: UCHAR is better?

jianchun: To avoid understanding new code, maybe use ```((*longToTest) & (1 << bitToTest)) != 0``` same as in ```BitTest```.

jianchun:
LGTM

digitalinfinity:
@jianchun Thanks! Updated the PR to incorporate the review feedback. @obastemur see the comment on the gitter channel but I can't get lldb to break automatically (by default) when raise(SIGTRAP) is called- __builtin_trap does break in by default, which is helpful in speeding up debugging- so I'm going to continue to use that for now.

PR #438, opened 2016-03-03
cross platform: implement rand_s

obastemur:
Implement `rand_s` without bringing libsodium dependency. (Tested on OSX, FreeBSD, Ubuntu)

### DIFF
--- a/pal/src/misc/random.cpp
+++ b/pal/src/misc/random.cpp
@@ -17,35 +17,63 @@ Abstract:
 
 --*/
 
+#ifdef _CRT_RAND_S
+
 #include "pal/palinternal.h"
-#include <sodium.h>
+#include <sys/stat.h>
+#include <fcntl.h>
+#include <mutex>
 
-/*++
-Function :
-RANDOMInitialize
 
-Initialize libsodium
+#define RANDOM_CACHE_SIZE 8
+constexpr int rand_byte_len = sizeof(unsigned int) * RANDOM_CACHE_SIZE;
+static union {
+    char rstack[rand_byte_len];
+    unsigned int result[RANDOM_CACHE_SIZE];
+} random_cache;
+static unsigned cache_index(RANDOM_CACHE_SIZE);
+static std::mutex random_generator_mutex;
 
-(no parameters)
-
-Return value :
-TRUE  if libsodium initialization succeeded
-FALSE otherwise
---*/
-BOOL RANDOMInitialize(void)
+static bool GetRandom(unsigned int *result) noexcept
 {
-    if (sodium_init() == -1)
-    {
-        return FALSE;
+    std::lock_guard<std::mutex> guard(random_generator_mutex);
+    if (cache_index < RANDOM_CACHE_SIZE) {
+        *result = random_cache.result[cache_index++];
+        return true;
     }
-    
-    return TRUE;
+
+    const int rdev = open("/dev/urandom", O_RDONLY);
+    unsigned len = 0;
+
+    do
+    {
+        int added = read(rdev, random_cache.rstack + len, rand_byte_len - len);
+        if (added < 0) {
+            close(rdev);
+            return false;
+        }
+        len += added;
+    } while (len < rand_byte_len);
+
+    close(rdev);
+    *result = random_cache.result[0];
+    cache_index = 1;
+    return len == rand_byte_len;
 }
 
-
-// Implemented in stdlib in the universal CRT
-errno_t __cdecl rand_s(unsigned int* randomValue)
+errno_t __cdecl rand_s(unsigned int* randomValue) noexcept
 {
-    *randomValue = randombytes_random();
-    return 1;
+    if (randomValue == nullptr) return 1;
+    
+    if (GetRandom(randomValue)) 
+    {
+        return 0;
+    }
+    else
+    {
+        *randomValue = 0;
+        return 1;
+    }
 }
+
+#endif // _CRT_RAND_S
--- a/pal/src/include/pal/misc.h
+++ b/pal/src/include/pal/misc.h
@@ -60,17 +60,6 @@ FALSE otherwise
 BOOL TIMEInitialize( void );
 
 /*++
-Function:
-RANDOMInitialize
-
-Return value:
-TRUE if initialize succeeded
-FALSE otherwise
-
---*/
-BOOL RANDOMInitialize( void );
-    
-/*++
 Function :
     MsgBoxInitialize
 


digitalinfinity:
Thank you @obastemur. I think in general we're fine with this plan- can you change OSX to use /dev/urandom too? I'm told that on BSD, arc4random uses Yarrow, which is fine, but on OS X, it still uses RC4, which is not cryptographically secure

obastemur:
@digitalinfinity makes sense arc4random is less secure on OSX.

BTW; Added a basic caching to reduce the number of kernel traps. I'm not sure if that is a good idea in terms of memory usage (extra 80 bytes per process - including counter) ?? 

Saw ChakraCore calls `rand_s` four times in a row and uses the same method for Math.Random. Current cached version is ~5X faster 

If this PR is all fine, gonna make separate PR to cleanup the traces of sodium.

pal/src/misc/random.cpp
```diff
@@ -18,34 +18,47 @@ Module Name:
...
-TRUE  if libsodium initialization succeeded
-FALSE otherwise
---*/
-BOOL RANDOMInitialize(void)
+static bool GetRandom(unsigned int *result) noexcept
 {
-    if (sodium_init() == -1)
-    {
-        return FALSE;
+    std::lock_guard<std::mutex> guard(random_generator_mutex);
```
suwc: I recommend adding NULL checking for result pointer to match rand_s() behavior in win32.

obastemur: `GetRandom` is a local method used by `rand_s` below. Originally there was no `NULL` check thus I didn't add that. So, should add `nullptr` check to `rand_s` instead ?

pal/src/misc/random.cpp
```diff
@@ -18,34 +18,47 @@ Module Name:
...
+
+    close(rdev);
+    *result = random_cache.result[0];
+    cache_index = 1;
+    return len == rand_byte_len;
+}
 
-// Implemented in stdlib in the universal CRT
-errno_t __cdecl rand_s(unsigned int* randomValue)
+errno_t __cdecl rand_s(unsigned int* randomValue) noexcept
```
suwc: I would put rand_s() inside #ifdef _CRT_RAND_S .. #endif just to be consistent with in Windows.
https://msdn.microsoft.com/en-us/library/sxtz2fa8.aspx

pal/src/misc/random.cpp
```diff
@@ -18,34 +18,47 @@ Module Name:
...
-    return TRUE;
-}
 
+    const int rdev = open("/dev/urandom", O_RDONLY);
+    unsigned len = 0;
+
+    do
+    {
+        int added = read(rdev, random_cache.rstack + len, rand_byte_len - len);
+        if (added < 0) {
```
suwc: Are we ever going to keep getting 'added == 0' reading /dev/urandom?

pal/src/misc/random.cpp
obastemur: That would break the current build. @digitalinfinity  ? So I should also `#define _CRT_RAND_S` no matter what?

pal/src/misc/random.cpp
```diff
@@ -18,34 +18,47 @@ Module Name:
...
-Function :
-RANDOMInitialize
-
-Initialize libsodium
 
-(no parameters)
+#define RANDOM_CACHE_SIZE 8
+constexpr int rand_byte_len = sizeof(unsigned int) * RANDOM_CACHE_SIZE;
+static union {
+  char rstack[rand_byte_len];
```
suwc: nit: prefer 4-space soft tab

pal/src/misc/random.cpp
suwc: Right - nullptr check for rand_s()

pal/src/misc/random.cpp
obastemur: random device driver (random, urandom) is based on an entropy pool. When that pool is empty, it will block the execution until it has something. So, it is either positive or not.

obastemur: BTW; PR is updated with all the other changes asked

pal/src/misc/random.cpp
```diff
@@ -18,34 +18,49 @@ Module Name:
...
 
-/*++
-Function :
-RANDOMInitialize
-
-Initialize libsodium
 
-(no parameters)
+#define RANDOM_CACHE_SIZE 8
+constexpr int rand_byte_len = sizeof(unsigned int) * RANDOM_CACHE_SIZE;
```
suwc: @digitalinfinity, is Linux branch of ChakraCore only using C++ 11 and above compilers?
(assuming constexpr is)

obastemur: Yes.

pal/src/misc/random.cpp
suwc: I thought consumer of rand_s() is responsible for defining it:
`#define _CRT_RAND_S
#include <stdlib.h>`

The prototype of rand_s() should be inside inside `#define _CRT_RAND_S`, according to MSDN (it'd be curious if that's not the case).

pal/src/misc/random.cpp
digitalinfinity: Kind of :smile: - the PAL is currently used only on platforms supporting C++11 and above. Changes to actual ChakraCore code still needs to support VS2013's version of VC++. 
TLDR; in this case it's fine

pal/src/misc/random.cpp
obastemur: > I thought consumer of rand_s() is responsible for defining it:

Added `#ifdef _CRT_RAND_S`

digitalinfinity: Should be fine, no? @obastemur, we already define this in lib\Common\CommonBasic.h, so wrapping it should be fine.

obastemur: > Should be fine, no? @obastemur, we already define this in lib\Common\CommonBasic.h, so wrapping it should be fine. 

Yes. Already added.

pal/src/misc/random.cpp
ianwjhalliday: PAL is C++11 and up you say?  :trollface:

suwc:
LGTM. Thanks!

pal/src/misc/random.cpp
```diff
@@ -17,35 +17,54 @@ Module Name:
...
-RANDOMInitialize
+#include "pal/palinternal.h"
+#include <sys/stat.h>
+#include <fcntl.h>
+#include <mutex>
 
-Initialize libsodium
 
-(no parameters)
+#define RANDOM_CACHE_SIZE 8
```
digitalinfinity: Curious- why is this a #define while rand_byte_len is a constexpr- don't particularly care but was interested in the thinking here

pal/src/misc/random.cpp
digitalinfinity: I wonder if is another optimization you could do here- if you initialize cache_index to -1 in the case where the cache is initialized/reset, then you could skip the mutex enter and instead do an InterlockedIncrement first. If the incremented index (which is what is returned) is greater than RANDOM_CACHE_SIZE, you can then grab the mutex and repopulate the cache- that way, we have a fast path and a slow path (and you enter the mutex only in the slow path). Not sure if it saves us much, just a thought since you are trying to optimize this code anyway

pal/src/misc/random.cpp
```diff
@@ -17,35 +17,54 @@ Module Name:
...
+    std::lock_guard<std::mutex> guard(random_generator_mutex);
+    if (cache_index < RANDOM_CACHE_SIZE) {
+        *result = random_cache.result[cache_index++];
+        return true;
     }
-    
-    return TRUE;
-}
 
+    const int rdev = open("/dev/urandom", O_RDONLY);
```
digitalinfinity: This can fail, so should return false if it does?

pal/src/misc/random.cpp
```diff
@@ -17,35 +17,54 @@ Module Name:
...
+        len += added;
+    } while (len < rand_byte_len);
+
+    close(rdev);
+    *result = random_cache.result[0];
+    cache_index = 1;
+    return len == rand_byte_len;
+}
+
+errno_t __cdecl rand_s(unsigned int* randomValue) noexcept
```
digitalinfinity: We should initialize randomValue to 0 as per the docs for rand_s at https://msdn.microsoft.com/en-us/library/sxtz2fa8.aspx

digitalinfinity:
Hi @obastemur- so sorry for the delay with this change. I've finished reviewing, overall looks good, some minor comments- can you please address them and update the PR (please rebase your commit on top of the current Linux branch).

pal/src/misc/random.cpp
obastemur: @digitalinfinity if it fails, read will also fail. I had tested that scenario already.

pal/src/misc/random.cpp
obastemur: >  if you initialize cache_index to -1 in the case where the cache is initialized/reset, then you could skip the mutex enter and instead do an InterlockedIncrement first. If the incremented index (which is what is returned) is greater than RANDOM_CACHE_SIZE

`InterlockedIncrement` is Windows specific.. There you have 3 separate atomic operations.

IMO current implementation is the cleanest (cross platform).

pal/src/misc/random.cpp
obastemur: > We should initialize randomValue to 0

done. PR is updated

pal/src/misc/random.cpp
obastemur: `constexpr` lets compiler run some basic logic during compile time while `define` is just an alias

dilijev:
@digitalinfinity closing and reopening the issue fixed it. Since this was after master was merged into Linux, the checks should work now.

PR #439, opened 2016-03-03
Implement decoding of data segments

commonlisp:
A WASM module may contain multiple data segments. The decoder deserializes the offset and address from the binary.

### DIFF
--- a/lib/WasmReader/ModuleInfo.cpp
+++ b/lib/WasmReader/ModuleInfo.cpp
@@ -192,13 +192,43 @@ ModuleInfo::SetFunctionImport(uint32 i, uint32 sigId, wchar_t* modName, uint32 m
     m_imports[i].fnName = fnName;
 }
 
-Wasm::WasmImport* ModuleInfo::GetFunctionImport(uint32 i) const
+Wasm::WasmImport*
+ModuleInfo::GetFunctionImport(uint32 i) const
 {
     if (i >= m_importCount)
     {
         return nullptr;
     }
     return &m_imports[i];
+}
+
+void
+ModuleInfo::AllocateDataSegs(uint32 count)
+{
+    Assert(count != 0);
+    m_datasegCount = count;
+    m_datasegs = AnewArray(m_alloc, WasmDataSegment*, count);
+}
+
+bool
+ModuleInfo::AddDataSeg(WasmDataSegment* seg, uint32 index)
+{
+    if (index >= m_datasegCount)
+    {
+        return false;
+    }
+    m_datasegs[index] = seg;
+    return true;
+}
+
+WasmDataSegment*
+ModuleInfo::GetDataSeg(uint32 index) const
+{
+    if (index >= m_datasegCount)
+    {
+        return nullptr;
+    }
+    return m_datasegs[index];
 }
 
 } // namespace Wasm
--- a/lib/WasmReader/WasmBinaryReader.cpp
+++ b/lib/WasmReader/WasmBinaryReader.cpp
@@ -166,12 +166,18 @@ WasmBinaryReader::ProcessSection(SectionCode sectionId, bool isEntry /*= true*/)
         break; // This section is not used by bytecode generator, stay in decoder
     }
     case bSectDataSegments:
-        // TODO: Populate Data entry info
-        ReadConst<UINT32>();  // dest addr in module memory
-        ReadConst<UINT32>();  // source offset (?)
-        ReadConst<UINT32>();  // size
-        ReadConst<UINT8>();  // init
-        m_moduleState.count++;
+        m_moduleState.size = LEB128(length);
+        if (m_moduleState.size == 0)
+        {
+            // TODO: Probably can factor this out and validate all count fields.
+            ThrowDecodingError(_u("Illegal data segment entry count"));
+        }
+        m_moduleInfo->AllocateDataSegs(m_moduleState.size);
+        for (m_moduleState.count = 0; m_moduleState.count < m_moduleState.size; m_moduleState.count++)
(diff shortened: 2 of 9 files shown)


MikeHolman:
would prefer to propagate this up so after we allocate the arraybuffer we can go through data segments and memcopy them in. For example, it can be some list of these structs on ModuleInfo, with pointer, length, and target offset

MikeHolman:
There is an init field, but I don't see a way to delay initialize. So I guess this is more of a "reserved for future use" field?

lib/WasmReader/WasmBinaryReader.h
```diff
@@ -136,6 +147,9 @@ namespace Wasm
             void ThrowDecodingError(const wchar_t* msg);
             void Trace(const wchar_t* msg);
 
+            typedef JsUtil::GrowingArray<WasmTypes::DataSegment, ArenaAllocator> DataSegments;
```
Cellule: Since the size is known in advance and fixed, simply use a normal array.

MikeHolman:
data segments are now inline, so will need to update this PR accordingly.

lib/WasmReader/ModuleInfo.cpp
```diff
@@ -163,6 +163,37 @@ Wasm::WasmExport* ModuleInfo::GetFunctionExport(uint32 iExport) const
     return &m_exports[iExport];
 }
 
+void
+ModuleInfo::AllocateDataSegs(uint32 count)
+{
+    m_datasegCount = count;
+    if (count > 0)
```
Cellule: useless check, count is `uint32`

lib/WasmReader/WasmBinaryReader.cpp
```diff
@@ -177,12 +177,12 @@ WasmBinaryReader::ProcessSection(SectionCode sectionId, bool isEntry /*= true*/)
         break; // This section is not used by bytecode generator, stay in decoder
     }
     case bSectDataSegments:
-        // TODO: Populate Data entry info
-        ReadConst<UINT32>();  // dest addr in module memory
-        ReadConst<UINT32>();  // source offset (?)
-        ReadConst<UINT32>();  // size
-        ReadConst<UINT8>();  // init
-        m_moduleState.count++;
+        m_moduleInfo->AllocateDataSegs(m_moduleState.size);
```
Cellule: `m_moduleState.size` is not initialized

lib/WasmReader/ModuleInfo.cpp
MikeHolman: count might be 0...

lib/WasmReader/WasmBinaryReader.cpp
commonlisp: Good catch. Still used to the prior version where size was initialized outside the switch.

lib/WasmReader/WasmBinaryReader.cpp
```diff
@@ -685,12 +690,20 @@ WasmBinaryReader::FunctionBodyHeader()
     TRACE_WASM_DECODER(L"Function body header: i32 = %u, i64 = %u, f32 = %u, f64 = %u", i32Count, i64Count, f32Count, f64Count);
 }
 
+void
+WasmBinaryReader::DataSegment()
+{
+    UINT len = 0;
+    WasmDataSegment *dseg = Anew(&m_alloc, WasmDataSegment, &m_alloc, LEB128(len), LEB128(len), m_pc);
```
MikeHolman: don't you need to do `m_pc += size` (and ideally validate against reading beyond end)?

commonlisp: Ah, of course. Good catch.

lib/WasmReader/WasmBinaryReader.cpp
```diff
@@ -166,12 +166,18 @@ WasmBinaryReader::ProcessSection(SectionCode sectionId, bool isEntry /*= true*/)
...
-        ReadConst<UINT32>();  // dest addr in module memory
-        ReadConst<UINT32>();  // source offset (?)
-        ReadConst<UINT32>();  // size
-        ReadConst<UINT8>();  // init
-        m_moduleState.count++;
+        m_moduleState.size = LEB128(length);
+        if (m_moduleState.size == 0)
+        {
+            // TODO: Probably can factor this out and validate all count fields.
+            ThrowDecodingError(L"Illegal data segment entry count");
```
Cellule: Since you'll have to rebase on top of my integrate merge, might as well fix this to `_u("Illegal data segment entry count")`

commonlisp:
@dotnetbot test this please

commonlisp:
@dotnetbot test this please

commonlisp:
@dotnetbot test this please

lib/WasmReader/WasmBinaryReader.cpp
```diff
@@ -166,12 +166,18 @@ WasmBinaryReader::ProcessSection(SectionCode sectionId, bool isEntry /*= true*/)
...
+        m_moduleState.size = LEB128(length);
+        if (m_moduleState.size == 0)
+        {
+            // TODO: Probably can factor this out and validate all count fields.
+            ThrowDecodingError(_u("Illegal data segment entry count"));
+        }
+        m_moduleInfo->AllocateDataSegs(m_moduleState.size);
+        for (m_moduleState.count = 0; m_moduleState.count < m_moduleState.size; m_moduleState.count++)
+        {
+            TRACE_WASM_DECODER(L"Data Segment #%u", m_moduleState.count);
```
Cellule: use `_u()` macro

PR #443, opened 2016-03-03
Make jenkins.buildall.cmd accept extra parameters. Fixes #442.

### LINKED ISSUE #442: jenkins.buildall.cmd ignores extra parameters
jenkins.buildall.cmd is a convenience script for desktop users who want to build all of the flavors locallly in the same way that jenkins builds them. The script is not used by any automated systems. It defers to the same logic used by the other scripts but ignores any parameters passed.

Might be as easy as adding %* to each command invoked by that script. I'll look into it when I have the chance.

dilijev:


dilijev:
FYI @nmostafa @leirocks @digitalinfinity

PR #445, opened 2016-03-03
Wasm Memory section page count

Cellule:
The Memory section min/max now refers to pages instead of bytes.


Cellule:
Opened the pr against the wrong branch

PR #446, opened 2016-03-04
Wasm Memory section page count

Cellule:
The Memory section min/max now refers to pages instead of bytes.

[Review on Reviewable](https://reviewable.io/reviews/microsoft/chakracore/446)

### DIFF
--- a/lib/WasmReader/ModuleInfo.cpp
+++ b/lib/WasmReader/ModuleInfo.cpp
@@ -19,30 +19,21 @@ ModuleInfo::ModuleInfo(ArenaAllocator * alloc) :
 }
 
 bool
-ModuleInfo::InitializeMemory(size_t minSize, size_t maxSize, bool exported)
+ModuleInfo::InitializeMemory(uint32 minPage, uint32 maxPage, bool exported)
 {
     if (m_memory.minSize != 0)
     {
         return false;
     }
 
-    if (maxSize < minSize)
+    if (maxPage < minPage)
     {
         return false;
     }
 
-    if (minSize == 0 || minSize % AutoSystemInfo::PageSize != 0)
-    {
-        return false;
-    }
-
-    if (maxSize % AutoSystemInfo::PageSize != 0)
-    {
-        return false;
-    }
-
-    m_memory.minSize = minSize;
-    m_memory.maxSize = maxSize;
+    CompileAssert(Memory::PAGE_SIZE < INT_MAX);
+    m_memory.minSize = (uint64)minPage * Memory::PAGE_SIZE;
+    m_memory.maxSize = (uint64)maxPage * Memory::PAGE_SIZE;
     m_memory.exported = exported;
 
     return true;
--- a/lib/WasmReader/WasmBinaryReader.cpp
+++ b/lib/WasmReader/WasmBinaryReader.cpp
@@ -406,11 +406,11 @@ WasmBinaryReader::EndOfModule()
 void
 WasmBinaryReader::ReadMemorySection()
 {
-    // TODO: change to use multiple of page size
-    uint32 minSize = 1 << ReadConst<UINT8>();
-    uint32 maxSize = 1 << ReadConst<UINT8>();
+    UINT length = 0;
+    UINT minPage = LEB128(length);
+    UINT maxPage = LEB128(length);
     bool exported = ReadConst<UINT8>() != FALSE;
-    m_moduleInfo->InitializeMemory(minSize, maxSize, exported);
+    m_moduleInfo->InitializeMemory(minPage, maxPage, exported);
 }
 
 void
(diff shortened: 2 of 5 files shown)


Cellule:
@MikeHolman @commonlisp Can you guys review this ?

Cellule:

Cellule:

lib/WasmReader/WasmBinaryReader.cpp
```diff
@@ -406,11 +406,12 @@ WasmBinaryReader::EndOfModule()
 void
 WasmBinaryReader::ReadMemorySection()
 {
+    UINT length = 0;
     // TODO: change to use multiple of page size
-    uint32 minSize = 1 << ReadConst<UINT8>();
-    uint32 maxSize = 1 << ReadConst<UINT8>();
+    UINT minPage = 1 << LEB128(length);
```
MikeHolman: should remove `1 <<`

lib/WasmReader/WasmBinaryReader.cpp
```diff
@@ -406,11 +406,12 @@ WasmBinaryReader::EndOfModule()
 void
 WasmBinaryReader::ReadMemorySection()
 {
+    UINT length = 0;
     // TODO: change to use multiple of page size
```
MikeHolman: this completes the todo, so can remove it

MikeHolman:
lgtm

lib/WasmReader/WasmBinaryReader.cpp
```diff
@@ -406,11 +406,11 @@ WasmBinaryReader::EndOfModule()
 void
 WasmBinaryReader::ReadMemorySection()
 {
-    // TODO: change to use multiple of page size
-    uint32 minSize = 1 << ReadConst<UINT8>();
-    uint32 maxSize = 1 << ReadConst<UINT8>();
+    UINT length = 0;
+    UINT minPage = LEB128(length);
+    UINT maxPage = LEB128(length);
```
naturaltransformation: Is m_moduleState.count updated for these two lengths somewhere?

lib/WasmReader/ModuleInfo.cpp
```diff
@@ -19,30 +19,21 @@ ModuleInfo::ModuleInfo(ArenaAllocator * alloc) :
...
-    if (maxSize % AutoSystemInfo::PageSize != 0)
-    {
-        return false;
-    }
-
-    m_memory.minSize = minSize;
-    m_memory.maxSize = maxSize;
+    CompileAssert(Memory::PAGE_SIZE < INT_MAX);
+    m_memory.minSize = (uint64)minPage * Memory::PAGE_SIZE;
+    m_memory.maxSize = (uint64)maxPage * Memory::PAGE_SIZE;
```
naturaltransformation: This is targeting WASM32 initially, right? If so, should we check that minSize and maxSize aren't too large?

lib/WasmReader/ModuleInfo.cpp
```diff
@@ -19,30 +19,21 @@ ModuleInfo::ModuleInfo(ArenaAllocator * alloc) :
 }
 
 bool
-ModuleInfo::InitializeMemory(size_t minSize, size_t maxSize, bool exported)
+ModuleInfo::InitializeMemory(uint32 minPage, uint32 maxPage, bool exported)
 {
     if (m_memory.minSize != 0)
```
naturaltransformation: Is this a stale check? m_memory has an initializer for minSize(0). Is this trying to indirectly ensure that memory isn't initialized twice?

MikeHolman: Yes, that was the idea.

lib/WasmReader/WasmBinaryReader.cpp
MikeHolman: Currently it seems that we aren't doing real validation that section length == declared length, so doesn't really matter. But we will need to add in. Functionally it's fine, because the callee updates modulestate based on the declared section length.

lib/WasmReader/ModuleInfo.cpp
Cellule: Since there is no restriction by spec to the max number of pages, I wanted to keep it simple and avoid to check for overflow.
However, I do check for it when we try to allocate the buffer and will throw OutOfMemory if the size is too big.
That way we only have to make changes at one place if we want to allow more memory

lib/WasmReader/WasmBinaryReader.cpp
Cellule: Yeah the section length validation (and other validation items) will have to be done in a future change. 
For the time being we need to have functionality working and we assume the binaries are valid. 
Obviously we won't ship it like this though

Cellule:
Currently I have problem with our full build, I need to investigate if it is related to my change and/or make the fix before I merge this in.

Cellule:
This is merged, I just deleted my branch too soon

PR #448, opened 2016-03-04
Quick implementation of section and module headers

commonlisp:
This is an implementation of the string-form of section headers
and the new module header. It currently uses a suboptimal
chain of string comparisons, but this is a placeholder for
something more efficient as the discussion on whether to
use an LEB128-like encoding or brotli instead of general strings
is settled. Ideally, we would just hash the string using
JsUtil::CharacterBuffer<char>::StaticGetHashCode() or some faster
alternative and do lookup of the hash map. This commit will be
helpful to establish the section handling in preparation for
breaking up the function section into decls and defs.

### DIFF
--- a/lib/WasmReader/WasmBinaryReader.cpp
+++ b/lib/WasmReader/WasmBinaryReader.cpp
@@ -22,6 +22,7 @@ namespace Binary
 {
 
 bool WasmBinaryReader::isInit = false;
+bool WasmBinaryReader::seenModuleHeader = false;
 WasmTypes::Signature WasmBinaryReader::opSignatureTable[WasmTypes::OpSignatureId::bSigLimit]; // table of opcode signatures
 WasmTypes::OpSignatureId WasmBinaryReader::opSignature[WasmBinOp::wbLimit];                   // opcode -> opcode signature ID
 const Wasm::WasmTypes::WasmType WasmBinaryReader::binaryToWasmTypes[] = { Wasm::WasmTypes::WasmType::Void, Wasm::WasmTypes::WasmType::I32, Wasm::WasmTypes::WasmType::I64, Wasm::WasmTypes::WasmType::F32, Wasm::WasmTypes::WasmType::F64 };
@@ -72,6 +73,9 @@ WasmBinaryReader::ReadFromModule()
 
     SectionCode sectionId;
     UINT length = 0;
+
+    ModuleHeader();
+
     while (!EndOfModule())
     {
         if (m_moduleState.secId > bSectSignatures && m_moduleState.count < m_moduleState.size)
@@ -82,8 +86,7 @@ WasmBinaryReader::ReadFromModule()
         else
         {
             // next section
-            sectionId = m_moduleState.secId = (SectionCode)ReadConst<UINT8>();
-            m_moduleState.count = 0;
+            sectionId = m_moduleState.secId = SectionHeader();
             if (sectionId == bSectMemory || sectionId == bSectEnd)
             {
                 m_moduleState.size = 1;
@@ -100,6 +103,7 @@ WasmBinaryReader::ReadFromModule()
             m_visitedSections->Set(sectionId);
         }
 
+        // TODO: Skip unknown sections.
         Assert(sectionId >= bSectMemory && sectionId <= bSectEnd);
 
         switch (sectionId)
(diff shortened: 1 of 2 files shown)


lib/WasmReader/WasmBinaryReader.cpp
```diff
@@ -72,6 +72,17 @@ WasmBinaryReader::ReadFromModule()
 
     SectionCode sectionId;
     UINT length = 0;
+
+    if (ReadConst<UINT32>() != 0x6d736100)
```
MikeHolman: this will not work: it will be called for every function. should have some check to make sure this is only executed first time you call

lib/WasmReader/WasmBinaryReader.cpp
```diff
@@ -167,6 +178,35 @@ WasmBinaryReader::ReadFromModule()
...
+
+    sectionSize = LEB128(len);
+    m_moduleState.count += len;
+    idSize = LEB128(len);
+    m_moduleState.count += len;
+    const char *sectionName = (char*)(m_pc);
+    m_pc += idSize;
+    m_moduleState.count += idSize;
+
+    auto cmp = [&sectionName, &idSize](const char* n) { return !strncmp(n, sectionName, idSize); };
```
MikeHolman: this is ok for now, but I think we will eventually want to flip around, such that you cmp "expected" section with the actual, and return invalid if not expected. also, technically these names don't need to be valid utf-8 (e.g. embedded null is legal), so we should do memcmp instead

lib/WasmReader/WasmBinaryReader.cpp
naturaltransformation: Good catch. Since the SExprParser doesn't read any header, I'm using a bit of state to ensure checking this only once now. Having a separate ReadModuleHeader() would have been nicer, but let's consider that later.

lib/WasmReader/WasmBinaryReader.cpp
nmostafa: Is the encoding of Section names utf-8 ? I couldn't find this in the spec doc.

lib/WasmReader/WasmBinaryReader.cpp
```diff
@@ -167,6 +183,35 @@ WasmBinaryReader::ReadFromModule()
...
+    UINT idSize;
+
+    sectionSize = LEB128(len);
+    m_moduleState.count += len;
+    idSize = LEB128(len);
+    m_moduleState.count += len;
+    const char *sectionName = (char*)(m_pc);
+    m_pc += idSize;
+    m_moduleState.count += idSize;
+
```
nmostafa: You are incrementing m_moduleState.count here, but it gets reset to 0 at the call-site of SectionHeader. I think you should reset the count here, and also set the size.

lib/WasmReader/WasmBinaryReader.cpp
```diff
@@ -72,6 +73,21 @@ WasmBinaryReader::ReadFromModule()
...
+            ThrowDecodingError(L"Malformed WASM module header!");
+        }
+
+        if (ReadConst<UINT32>() != 10)
+        {
+            ThrowDecodingError(L"Invalid WASM version!");
+        }
+        seenModuleHeader = true;
+    }
+
```
nmostafa: Better to outline this (with the seenModuleHeader check) to ModuleHeader() ?

lib/WasmReader/WasmBinaryReader.cpp
naturaltransformation: Hmm, where are you seeing this? The section names are unrestricted and may contain null chars, as is the case with function names too, which we have previously assumed to be UTF8. This comparison should have already been rewritten to use memcmp.

lib/WasmReader/WasmBinaryReader.cpp
naturaltransformation: Ah, right. This shouldn't be updating count at all. However, we can't set the m_moduleState.size just yet because this size is a byte size of a section (exclusive of the varint size itself) and not a module section element count, which will follow the section name usually. We'll need a separate piece of state to track section byte size eventually.

lib/WasmReader/WasmBinaryReader.cpp
naturaltransformation: Good idea. Thanks!

MikeHolman:
lgtm

lib/WasmReader/WasmBinaryReader.cpp
nmostafa: Right. Better to rename those variables then (size: bytesize, len: entries count)

PR #449, opened 2016-03-04
Enabling JIT for split scope scenarios in default parameters

aneeshdk:
Enabling JIT for split scope scenarios in default parameters

The default parameter split scope introduced two new opcodes.
BeginBodyScope marks the end of param scope and the start of body scope.
In split scope we have to have separate scopes for both param and body. So
a new scope slot array is created for the body scope in this opcode.

At the beginning of the body the initial values of the body symbols
corresponding to the params should get the final value from the
param scope symbols. To load the old value from the old param scope slots
LdParamSlot opcode is used.

### DIFF
--- a/lib/Backend/IRBuilder.cpp
+++ b/lib/Backend/IRBuilder.cpp
@@ -588,8 +588,8 @@ IRBuilder::Build()
                 Js::OpCode op =
                     m_func->DoStackScopeSlots() ? Js::OpCode::NewStackScopeSlots : Js::OpCode::NewScopeSlots;
 
-                IR::Opnd * srcOpnd = IR::IntConstOpnd::New(
-                    m_func->GetJnFunction()->scopeSlotArraySize + Js::ScopeSlots::FirstSlotIndex, TyUint32, m_func);
+                uint size = m_func->GetJnFunction()->IsParamAndBodyScopeMerged() ? m_func->GetJnFunction()->scopeSlotArraySize : m_func->GetJnFunction()->paramScopeSlotArraySize;
+                IR::Opnd * srcOpnd = IR::IntConstOpnd::New(size + Js::ScopeSlots::FirstSlotIndex, TyUint32, m_func);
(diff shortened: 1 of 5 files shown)


aneeshdk:
@pleath @abchatra @rajatd Can you please take a look at this?

lib/Backend/IRBuilder.cpp
```diff
@@ -6615,6 +6624,36 @@ IRBuilder::BuildEmpty(Js::OpCode newOpcode, uint32 offset)
         }
         break;
 
+    case Js::OpCode::BeginBodyScope:
+        // This marks the end of a param socpe which is not merged with body scope.
+        // So we have to first cache the closure so that we can use it to copy the initial values for
+        // body syms from corresponding param syms (LdParamSlot). Body should get its own scope slot.
+        this->AddInstr(
```
pleath: Are we guaranteed at this point that local scope slots and/or frame display are required?

aneeshdk: Fixed it.

lib/Backend/IRBuilder.cpp
```diff
@@ -6615,6 +6624,39 @@ IRBuilder::BuildEmpty(Js::OpCode newOpcode, uint32 offset)
...
+                Js::Constants::NoByteCodeOffset);
+
+            this->AddInstr(
+                IR::Instr::New(
+                    Js::OpCode::LdFrameDisplay,
+                    this->BuildDstOpnd(this->m_func->GetJnFunction()->GetFrameDisplayRegister()),
+                    this->BuildSrcOpnd(this->m_func->GetJnFunction()->GetLocalClosureRegister()),
+                    this->BuildSrcOpnd(this->m_func->GetJnFunction()->GetFrameDisplayRegister()),
+                    m_func),
+                (uint32)-1);
```
akroshg: Should this be NoByteCodeOffset ?

aneeshdk: Fixed it.

PR #450, opened 2016-03-04
cross platform: force clang

obastemur:
Forces Clang. Discussed on gitter. 

In case user wants to target clang from a specific path, cmake provides a command line argument to override it. Otherwise, this update forces Clang from /usr/bin unless the actual compiler is not Clang.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -3,28 +3,100 @@
 # Licensed under the MIT license. See LICENSE.txt file in the project root for full license information.
 #-------------------------------------------------------------------------------------------------------
 
+SAFE_RUN() {
+    local SF_RETURN_VALUE=$($1 2>&1)
+
+    if [[ $? != 0 ]]; then
+        >&2 echo $SF_RETURN_VALUE
+        exit 1
+    fi
+    echo $SF_RETURN_VALUE
+}
+
+PRINT_USAGE() {
+    echo ""
+    echo "build.sh [options]"
+    echo ""
+    echo "options:"
+    echo "--help, -h : Show help"
+    echo "--debug, -d : Debug build"
+    echo "CXX=<path to clang++>"
+    echo "CC=<path to clang>"
+    echo ""
+    echo "example: ./build.sh CXX=/path/to/clang++ CC=/path/to/clang"
+}
+
+_CXX=""
+_CC=""
+DEBUG_BUILD=0
+
+while [[ $# -gt 0 ]]; do
+    if [[ "$1" =~ "CXX=" ]]; then
+        _CXX=$1
+        _CXX=${_CXX:4}
+    fi
+
+    if [[ "$1" =~ "CC=" ]]; then
+        _CC=$1
+        _CC=${_CC:3}
+    fi
+
+    if [[ "$1" == "--help" || "$1" == "-h" ]]; then
+        PRINT_USAGE
+        exit
+    fi
+
+    if [[ "$1" == "--debug" || "$1" == "-d" ]]; then
+        DEBUG_BUILD=1
+    fi
+
+    shift
+done
+
+if [[ ${#_CXX} > 0 && ${#_CC} > 0 ]]; then
+    echo "Custom CXX ${_CXX}"
+    echo "Custom CC  ${_CC}"
+
+    if [[ ! -f $_CXX || ! -f $_CC ]]; then
+        echo "ERROR: Custom compiler not found on given path"
+        exit 1
+    fi
+else
+    RET_VAL=$(SAFE_RUN 'c++ --version')
+    if [[ ! $RET_VAL =~ "clang" ]]; then
+        echo "Searching for Clang..."
+        if [[ -f /usr/bin/clang++ ]]; then
+            echo "Clang++ found at /usr/bin/clang++"
+            _CXX=/usr/bin/clang++
+            _CC=/usr/bin/clang
+        else
+            echo "ERROR: clang++ not found at /usr/bin/clang++"
+            echo ""
+            echo "You could use clang++ from a custom location."
+            PRINT_USAGE
+            exit 1
+        fi
+    fi
+fi
+
+CC_PREFIX=""
+if [[ ${#_CXX} > 0 ]]; then
+    CC_PREFIX="-DCMAKE_CXX_COMPILER=$_CXX -DCMAKE_C_COMPILER=$_CC"
+fi
+
 if [ ! -d "BuildLinux" ]; then
-    mkdir BuildLinux;
+    SAFE_RUN 'mkdir BuildLinux'
 fi
 
 pushd BuildLinux > /dev/null
 
-DEBUG_BUILD=0
-while getopts ":d" opt; do
-    case $opt in
-        d)
-        DEBUG_BUILD=1
-        ;;
-    esac
-done
-
 if [ $DEBUG_BUILD -eq 1 ]; then
     echo Generating Debug makefiles
-    cmake -DCMAKE_BUILD_TYPE=Debug ..
+    cmake $CC_PREFIX -DCMAKE_BUILD_TYPE=Debug ..
 else
     echo Generating Retail makefiles
     echo Building Retail;
-    cmake -DCMAKE_BUILD_TYPE=Release ..
+    cmake $CC_PREFIX -DCMAKE_BUILD_TYPE=Release ..
 fi
 
 make |& tee build.log


digitalinfinity:
Thanks @obastemur- I think this is a good change to make. However, couple points of feedback:
1. Should we have a similar check for the C compiler too to force clang?
2. I tried the change locally on my machine and I seem to be running into issues- I did the following steps:
 a) **`sudo update-alternatives --config c++`** (selected g++ here)
 b) **`./build.sh `**
I get a build error:
```
hiteshk@hiteshk-vplayer:~/code/ch_merge$ ./build.sh
Generating Retail makefiles
Building Retail
Current Compiler ID Clang
Detected Linux x86_64
-- Configuring done
-- Generating done
-- Build files have been written to: /home/hiteshk/code/ch_merge/BuildLinux
Scanning dependencies of target Chakra.Common.Codex
[  0%] Building CXX object lib/Common/Codex/CMakeFiles/Chakra.Common.Codex.dir/Utf8Codex.cpp.o
c++: error: unrecognized command line option ‘-fdelayed-template-parsing’
lib/Common/Codex/CMakeFiles/Chakra.Common.Codex.dir/build.make:54: recipe for target 'lib/Common/Codex/CMakeFiles/Chakra.Common.Codex.dir/Utf8Codex.cpp.o' failed
```
Do we need to do more to force the compiler to clang if update-alternatives is used?

obastemur:
Hmm. But it says `Current Compiler ID Clang`

obastemur:
Gonna check that, something is off.

obastemur:
@digitalinfinity updated `build.sh` instead. 

Sample usage below;
```
$> ./build.sh --help

build.sh [options]

options:
--help, -h : Show help
--debug, -d : Debug build
CXX=<path to clang++>
CC=<path to clang>

example: ./build.sh CXX=/path/to/clang++ CC=/path/to/clang
```

How it works:

- If CXX, CC are set from command line, build script uses them

otherwise

- If clang++ is default anyways, use it

finally

- In case clang++ is located at `/usr/bin/`, build script uses clang++ from that location instead of producing errors with default g++/gcc

- none of the options above ? then show an error message and exit.

Besides, `--debug, -d and --help, -h` options are added

digitalinfinity:
Thank you @obastemur. Change looks fine functionally- @dilijev can you also review please?

build.sh
```diff
@@ -3,28 +3,101 @@
...
 
-if [ ! -d "BuildLinux" ]; then
-    mkdir BuildLinux;
-fi
+SAFE_RUN() {
+    SF_RETURN_VALUE=$($1 2>&1)
 
-pushd BuildLinux > /dev/null
+    if [[ $? != 0 ]]
+    then
```
dilijev: To make this match the other if's can you move `; then` to the previous line?

build.sh
```diff
@@ -3,28 +3,101 @@
...
+if [[ ${#_CXX} > 0 && ${#_CC} > 0 ]]; then
+    echo "Custom CXX ${_CXX}"
+    echo "Custom CC  ${_CC}"
+
+    if [[ ! -f $_CXX || ! -f $_CC ]]; then
+        echo "ERROR: Custom compiler not found on given path"
+        exit 1
+    fi
+else
+    SAFE_RUN "c++ --version" XXXX
```
dilijev: Is the `XXXX` needed?

digitalinfinity: What is this XXXX for?

build.sh
```diff
@@ -3,28 +3,101 @@
...
+    if [[ ! $SF_RETURN_VALUE =~ "clang" ]]; then
+        echo "Searching for Clang.."
```
dilijev: Change to `...`?

build.sh
```diff
@@ -3,28 +3,101 @@
 # Licensed under the MIT license. See LICENSE.txt file in the project root for full license information.
 #-------------------------------------------------------------------------------------------------------
 
-if [ ! -d "BuildLinux" ]; then
-    mkdir BuildLinux;
-fi
+SAFE_RUN() {
+    SF_RETURN_VALUE=$($1 2>&1)
```
dilijev: Can you use `local` and return this string instead of setting it to a global variable in this function?

build.sh
```diff
@@ -3,28 +3,101 @@
...
+}
 
+_CXX=""
+_CC=""
 DEBUG_BUILD=0
-while getopts ":d" opt; do
-    case $opt in
-        d)
+
+while test $# -gt 0
```
dilijev: Can you do `while [ $# -gt 0 ]; then` as you did with `if` blocks?

build.sh
```diff
@@ -3,28 +3,101 @@
...
+    if [[ "$1" =~ "CC=" ]]; then
+        _CC=$1
+        _CC=${_CC:3}
+    fi
+
+    if [[ "$1" == "--help" || "$1" == "-h" ]]; then
+        PRINT_USAGE
+        exit
+    fi
+
```
dilijev: nit: trim trailing whitespace in this line

dilijev:
Functionally, it also worked for me. Aside from my comments LGTM.

build.sh
obastemur: > nit: trim trailing whitespace in this line
turned of automatic whitespace trimming on all the editors. there we have this. I'm still +100 on merging that whitespace fix PR though.

obastemur:
@dilijev @digitalinfinity Thanks for the in depth review. I've just update the PR.

build.sh
```diff
@@ -3,28 +3,100 @@
...
+}
 
+_CXX=""
+_CC=""
 DEBUG_BUILD=0
-while getopts ":d" opt; do
-    case $opt in
-        d)
+
+while test $# -gt 0; do
```
dilijev: I think I'd still prefer `[ ... ]` instead of `test` for legibility here.

build.sh
```diff
@@ -3,28 +3,100 @@
...
-if [ ! -d "BuildLinux" ]; then
-    mkdir BuildLinux;
-fi
+SAFE_RUN() {
+    local SF_RETURN_VALUE=$($1 2>&1)
 
-pushd BuildLinux > /dev/null
+    if [[ $? != 0 ]]; then
+        >&2 echo $SF_RETURN_VALUE
+        exit
```
dilijev: Can you set a non-zero exit value in this case?

build.sh
```diff
@@ -3,28 +3,100 @@
...
+            echo "ERROR: clang++ not found at /usr/bin/clang++"
+            echo ""
+            echo "You could use clang++ from a custom location."
+            PRINT_USAGE
+            exit 1
+        else
+            echo "Clang++ found at /usr/bin/clang++"
+            _CXX=/usr/bin/clang++
+            _CC=/usr/bin/clang
+        fi
```
dilijev: Can you flip the condition of this if block and put the success case first?

build.sh
```diff
@@ -3,28 +3,100 @@
...
+    if [[ "$1" == "--debug" || "$1" == "-d" ]]; then
         DEBUG_BUILD=1
-        ;;
-    esac
+    fi
+
+    shift
 done
 
+if [[ ${#_CXX} > 0 && ${#_CC} > 0 ]]; then
```
dilijev: Are the only options that either *both* or *neither* `CXX=` and `CC=` be supplied? That's probably safest, but I wonder if it is possible to deduce one given the other.

build.sh
```diff
@@ -3,28 +3,100 @@
...
+CC_PREFIX=""
+if [[ ${#_CXX} > 0 ]]; then
+    CC_PREFIX="-DCMAKE_CXX_COMPILER=$_CXX -DCMAKE_C_COMPILER=$_CC"
+fi
+
+if [ ! -d "BuildLinux" ]; then
+    mkdir BuildLinux;
+fi
+
+pushd BuildLinux > /dev/null
```
dilijev: I think you could change this to `&>` just in case of error output that you want to suppress, and then check that the working directory matches what was expected (that `BuildLinux` actually exists) before continuing.

build.sh
```diff
@@ -3,28 +3,100 @@
...
+    fi
+fi
+
+CC_PREFIX=""
+if [[ ${#_CXX} > 0 ]]; then
+    CC_PREFIX="-DCMAKE_CXX_COMPILER=$_CXX -DCMAKE_C_COMPILER=$_CC"
+fi
+
+if [ ! -d "BuildLinux" ]; then
+    mkdir BuildLinux;
```
dilijev: It's possible for this command to fail. You should check and return a non-zero exit code in that case.

build.sh
```diff
@@ -3,28 +3,100 @@
...
-    cmake -DCMAKE_BUILD_TYPE=Debug ..
+    cmake $CC_PREFIX -DCMAKE_BUILD_TYPE=Debug ..
 else
     echo Generating Retail makefiles
     echo Building Retail;
-    cmake -DCMAKE_BUILD_TYPE=Release ..
+    cmake $CC_PREFIX -DCMAKE_BUILD_TYPE=Release ..
 fi
 
 make |& tee build.log
```
dilijev: @digitalinfinity why `|&` instead of just `|`?

dilijev:
Hi @obastemur, I looked a little closer at your change after the corrections and noticed a couple more things I think would be a good idea to address.

build.sh
dilijev: Oh nevermind, of course make could have some output on stderr that you want to capture to a log.

build.sh
obastemur: @dilijev this wasn't part of the PR (part of the original file) and IMO it's just fine

build.sh
obastemur: This is also part of the original file but It can be executed with the new SAFE_RUN method I've implemented.

build.sh
obastemur: Can I ask you why?

build.sh
obastemur: IMO no other option makes sense

build.sh
obastemur: done

build.sh
obastemur: done and PR is updated

build.sh
dilijev: Agreed, just wanted to make sure we're on the same page.

build.sh
dilijev: 1. Positive if statement conditions are easier for the reader to parse, and there are both blocks here.
2. The success case is the goal, after all, makes sense for it to be the first thing to read and handle error as an alternative.
3. Personal preference in this case based on the above.

obastemur: done updated.

build.sh
```diff
@@ -3,28 +3,100 @@
...
+    fi
+fi
+
+CC_PREFIX=""
+if [[ ${#_CXX} > 0 ]]; then
+    CC_PREFIX="-DCMAKE_CXX_COMPILER=$_CXX -DCMAKE_C_COMPILER=$_CC"
+fi
+
+if [ ! -d "BuildLinux" ]; then
+    SAFE_RUN 'mkdir BuildLinux'
```
dilijev: Nice. Thanks.

dilijev:
Thanks for the changes! LGTM. @digitalinfinity good to merge.

PR #452, opened 2016-03-04
Exports property for function exports in WASM module

commonlisp:
Add a property exports to the WASM object in JS that maps to all the exported functions in addition to exporting it as part of the object itself.


lib/Runtime/Base/ScriptContext.cpp
```diff
@@ -1933,6 +1939,8 @@ namespace Js
                         HeapDeleteArray(utf16Len + 1, contents);
 
                         JavascriptOperators::OP_SetProperty(exportObj, propertyRecord->GetPropertyId(), funcObj, this);
```
MikeHolman: presumably we shouldn't add directly onto exportObj anymore now?

Cellule: right, at first I told him not to remove it because we didn't know how the format was going to progress.
I think at this point it is safe to simply add it to the `exports` object and remove this line.

Cellule:
It seems my commits got mixed up in your pr, could you fix that please

PR #453, opened 2016-03-04
Fix to allow debugger thread setting counter values on FunctionBody

leirocks:


leirocks:
looks wrong commit submitted together

PR #454, opened 2016-03-04
Avoid undo defer for function which has non simple param list.

akroshg:
There is a duplicated logic in the FinishDeferredFunction code path (can
be tried using ForceUndoDefer) for parsing function. It does not make
sense to write the code for handling the default scenario again. So avoid
undeferring such function immediately until they are called.

### DIFF
--- a/lib/Parser/Parse.cpp
+++ b/lib/Parser/Parse.cpp
@@ -6344,7 +6344,7 @@ void Parser::FinishFncNode(ParseNodePtr pnodeFnc)
                 if (m_token.tk == tkAsg)
                 {
                     // Eat the default expression
-                    m_pscan->ScanNoKeywords();
+                    m_pscan->Scan();
                     ParseExpr<false>(koplCma);
                 }
             }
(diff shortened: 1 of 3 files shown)


akroshg:
@tcare, @aneeshdk  please review only this commit 7d24143

tcare:
Was there a test case for this at all? We could do a -forceundodefer.

akroshg:
I got split mind while adding the test - will do that.

akroshg:
@pleath  - yes the ParseExpr below will give problem otherwise. (I know the whole thing is not exercise under non-simple-param list but I did just for completeness sake).

pleath:
Has ScanNoKeywords always been wrong? This isn’t a new requirement with your change, right?

--Paul

akroshg:
@pleath  - no it is not a requirement with my new change
Before my fix - this produced an assert
```
function (a = b) {
}
```

pleath:
Got it. Thanks. Change LGTM, then.

--Paul

PR #457, opened 2016-03-04
Implement interpreter support for module import/exports

boingoing:
Add bytecodes and interpreter handlers for StModuleSlot and LdModuleSlot.

Add parser support for exporting async functions.

Better parsing support for export default statements.


boingoing:
@pleath @Yongqu @abchatra would you mind taking a look?

Note: This change is built on top of Yong's work so build will fail until his work is pushed and this change is rebased.

lib/Parser/Parse.cpp
```diff
@@ -1398,15 +1398,52 @@ ParseNodePtr Parser::AddVarDeclNode(IdentPtr pid, ParseNodePtr pnodeFnc)
...
+    Symbol* sym = declNode->sxVar.sym;
+    
+    sym->SetIsModuleExportStorage(true);
 
     return declNode;
 }
 
+Symbol* Parser::MarkIdentifierReferenceIsModuleExport(IdentPtr localName)
+{
+    PidRefStack* pidRef = this->PushPidRef(localName);
```
pleath: Can you explain what this method does? It looks like you have a new identifier reference in a scope, and you look for the outermost reference to the same identifier that already has a binding.

lib/Parser/Parse.cpp
```diff
@@ -2252,7 +2301,26 @@ void Parser::AddModuleImportEntry(ModuleImportEntryList* importEntryList, IdentP
     importEntryList->Prepend(*importEntry);
 }
 
-void Parser::AddModuleExportEntry(ModuleExportEntryList* exportEntryList, IdentPtr importName, IdentPtr localName, IdentPtr exportName, IdentPtr moduleRequest)
+void Parser::CheckForDuplicateExportEntry(ModuleExportEntryList* exportEntryList, IdentPtr exportName)
+{
+    Js::PropertyId newPropertyId = EnsurePropertyId(exportName);
```
pleath: Can you explain what this is doing?

lib/Parser/Parse.cpp
```diff
@@ -2447,6 +2523,146 @@ IdentPtr Parser::ParseImportOrExportFromClause(bool throwIfNotFound)
...
+        }
+        goto LDefault;
+        break;
+    case tkFUNCTION:
+        {
+LFunction:
+            // We just parsed a function token but we need to figure out if the function 
+            // has an identifier name or not before we call the helper. 
+            RestorePoint parsedFunction;
+            m_pscan->Capture(&parsedFunction);
```
pleath: Would it work here to call ParseFncDecl with no flags and check the parsed result afterward to see if it has a name? I'd like to get rid of the rewind if we can, especially the special treatment of certain tokens we see between "function" and the point where we rewind.

lib/Runtime/ByteCode/ByteCodeEmitter.cpp
```diff
@@ -618,7 +618,7 @@ void ByteCodeGenerator::InitBlockScopedContent(ParseNode *pnodeBlock, Js::Debugg
                 Js::OpCode op = (sym->GetDecl()->nop == knopConstDecl) ? Js::OpCode::InitUndeclConsoleConstFld : Js::OpCode::InitUndeclConsoleLetFld;
                 this->m_writer.ElementScopedU(op, funcInfo->FindOrAddReferencedPropertyId(propertyId));
             }
-            else
+            else if (!sym->GetIsModuleExportStorage())
```
pleath: Can a module export be declared in a block scope? I'm wondering how sym->GetIsModuleExportStorage() can be true.

lib/Runtime/ByteCode/ByteCodeEmitter.cpp
```diff
@@ -6893,7 +6987,14 @@ void EmitMethodFld(ParseNode *pnode, Js::RegSlot callObjLocation, Js::PropertyId
     bool isScoped = (byteCodeGenerator->GetFlags() & fscrEval) != 0 ||
         (isRoot && callObjLocation != ByteCodeGenerator::RootObjectRegister);
 
-    EmitMethodFld(isRoot, isScoped, pnode->location, callObjLocation, propertyId, byteCodeGenerator, funcInfo, registerCacheIdForCall);
```
pleath: I wonder if this is the right place. Can you tell me how we got to EmitMethodFld with a symbol that shouldn't be loaded as an object property?

lib/Parser/Parse.cpp
```diff
@@ -2252,7 +2301,26 @@ void Parser::AddModuleImportEntry(ModuleImportEntryList* importEntryList, IdentP
...
+        if (existingPropertyId == newPropertyId)
+        {
+            return true;
+        }
+        return false;
+    });
+
+    if (findResult != nullptr)
+    {
+        Error(ERRsyntax);
```
ianwjhalliday: Please make a specific error message for this

lib/Runtime/ByteCode/ByteCodeEmitter.cpp
```diff
@@ -10229,6 +10330,11 @@ void Emit(ParseNode *pnode, ByteCodeGenerator *byteCodeGenerator, FuncInfo *func
             byteCodeGenerator->EmitPropStore(pnode->location, sym, nullptr, funcInfo, true, false);
         }
 
+        if (pnode->sxClass.IsDefaultModuleExport())
```
pleath: What's the syntax that would get us here?

lib/Runtime/Language/InterpreterStackFrame.cpp
```diff
@@ -7933,6 +7933,34 @@ const byte * InterpreterStackFrame::OP_ProfiledLoopBodyStart(const byte * ip)
...
+    template <class T>
+    Var InterpreterStackFrame::OP_LdModuleSlot(Var instance, const unaligned T* playout)
+    {
+        Js::SourceTextModuleRecord* moduleRecord = scriptContext->GetLibrary()->GetModuleRecord(playout->SlotIndex1);
+        Assert(moduleRecord != nullptr);
+
+        Var* moduleRecordSlots = moduleRecord->GetLocalExportSlots();
+        Assert(moduleRecordSlots != nullptr);
+
+        Var value = moduleRecordSlots[playout->SlotIndex2];
```
Yongqu: assert/verify the count is smaller than list size?

lib/Runtime/Language/InterpreterStackFrame.cpp
```diff
@@ -7933,6 +7933,34 @@ const byte * InterpreterStackFrame::OP_ProfiledLoopBodyStart(const byte * ip)
...
+    {
+        Assert(value != nullptr);
+
+        Js::SourceTextModuleRecord* moduleRecord = scriptContext->GetLibrary()->GetModuleRecord(slotIndex1);
+        Assert(moduleRecord != nullptr);
+
+        Var* moduleRecordSlots = moduleRecord->GetLocalExportSlots();
+        Assert(moduleRecordSlots != nullptr);
+
+        moduleRecordSlots[slotIndex2] = value;
```
Yongqu: similarly let's verify the slotIndex2 is valid.

lib/Runtime/ByteCode/ByteCodeWriter.cpp
```diff
@@ -1681,6 +1681,8 @@ namespace Js
             case OpCode::StEnvSlotChkUndecl:
             case OpCode::StEnvObjSlot:
             case OpCode::StEnvObjSlotChkUndecl:
+            case OpCode::StModuleSlot:
+            case OpCode::LdModuleSlot:
```
pleath: Just a future note. I expect we're going to want a Profiled variant of LdModuleSlot. Very easy change.

pleath: Oh, and I see below that we've already got one. Excellent. Thanks.

pleath:
Did I miss the JIT support, or is it coming later?

lib/Parser/Parse.cpp
```diff
@@ -2447,6 +2523,146 @@ IdentPtr Parser::ParseImportOrExportFromClause(bool throwIfNotFound)
...
+    ParseNodePtr pnode = nullptr;
+    ushort flags = fFncNoFlgs;
+    
+    switch (m_token.tk)
+    {
+    case tkCLASS:
+        {
+            if (!m_scriptContext->GetConfig()->IsES6ClassAndExtendsEnabled())
+            {
+                goto LDefault;
```
ianwjhalliday: @abchatra here is a reason why I'd prefer we remove this feature flags once we enable a feature.  Otherwise we need to make sure the feature flags are maintained and kept functionally correct when newer features come online.  @boingoing is doing that here, but in general either we need to make sure we are disciplined about ensuring the switches remain functionally correct (or at least don't crash), or we abandon them.

lib/Parser/Parse.cpp
boingoing: That's sort of what I'm trying to do. This method is trying to find the symbol for localName so that we can mark it as stored in an export slot. We use this method when we're exporting something via a binding identifier instead of having the object itself. 

An example:
function foo() { };
export { foo };
export { foo as bar };

In both export statements here, we use MarkIdentifierReferenceIsModuleExport to find and mark the symbol for identifier 'foo' as being an export.

lib/Runtime/Language/SourceTextModuleRecord.h
```diff
@@ -66,7 +66,8 @@ namespace Js
         }
         static SourceTextModuleRecord* Create(ScriptContext* scriptContext);
 
-        uint GetLocalExportSlotIndex(PropertyId exportNameId);
+        uint GetLocalExportSlotIndexByExportName(PropertyId exportNameId);
+        uint GetLocalExportSlotIndexByLocalName(PropertyId localNameId);
```
Yongqu: private is enough if this is only used by the verification code in namespace

lib/Parser/Parse.cpp
boingoing: It's a syntax error to export the same name twice. This method is checking to see if we already have an export entry in the list for the name exportName and if so, throwing an error.

lib/Runtime/Language/InterpreterStackFrame.cpp
pleath: I'm guessing that an assert should be sufficient here, since the index is computed internally and not provided by the user.

pleath: No, I take that back. It should be verified dynamically, with a FatalInternalError if it's not valid.

lib/Parser/Parse.cpp
pleath: Is this switch not an example of one we could remove, since it enables syntax and not just functionality?

lib/Runtime/ByteCode/ByteCodeEmitter.cpp
boingoing: This IsDefaultModuleExport pattern is to handle the case where we don't have a symbol bound for the thing we're trying to export. That only happens for default exports. The class version would look like this:

export default class { };

You can export an anonymous class (or function, generator, async function, etc) under the default export name. However, we still need to know that thing should be stored in module export var storage. So that's what this flag is meant to help us figure out.

lib/Runtime/Language/InterpreterStackFrame.cpp
boingoing: Sure, good call

lib/Parser/Parse.cpp
pleath: (Sorry, I meant the innermost reference.) What are the rules here? Is it not true that export statements can only appear at global scope?

boingoing:
JIT support is not added here but it should be a bug-level fix later.

lib/Parser/Parse.cpp
pleath: Okay. Two things, then. It's not necessary to verify this via property ID. You can just look at IdentPtr equality, because the parser's hash table will have exactly one entry per identifier. The other is that we should be able to avoid the n^2 search. Is it true that export statements can only occur at global scope?

lib/Runtime/ByteCode/ByteCodeEmitter.cpp
pleath: Got it. Thanks.

pleath:
Right, it shouldn't be complicated. Now, if people turn on the feature locally are they going to experience random death when we try to JIT? Or is it not possible to enable it?

boingoing:
I'm not sure if the JS experimental flag will turn the feature on or not. I think it will but, yes, we are going to experience random death if we try to JIT one of these until the JIT support is added.

lib/Parser/Parse.cpp
boingoing: It's true that export statements can only appear at global scope, yes. But they can export any local name which is why I'm looking up the symbol for the name. I found in testing that the Symbol* in the PidRefStack* I got back from PushPidRef was nullptr but I was able to walk back along the stack and find a Symbol* which I could use to mark the name as export storage. 

We can have some syntax like this:
var foo = 1; 
function bar() { foo++ };
export { foo, bar };

pleath: Well, a statement at global scope can't export a local name. In your example, we're exporting a global name. I think the solution we discussed originally (or maybe I only dreamed that we discussed it) was this: At the time we call PushPidRef for the identifier reference in the export statement, mark a bit on the pidref to indicate that the symbol it binds to is exported. Then, in BindPidRefsInScopeImpl, if the ref has the "exported" bit set, mark the symbol as exported. That should eliminate the search. At that point, we should be able to assert that the symbol is global.

lib/Parser/Parse.cpp
boingoing: I couldn't get ParseFncDecl to do what I wanted without being explicit about the flags. It always ended up throwing in one or the other types of function decls. I think it probably makes sense to add a fFncDefault (or similar) to indicate this grammar. The [+Default] grammar is only used by module default exports. ParseFncDecl is a big piece, though. I didn't want to destabilize it for other types of functions.

pleath: Okay. Do you remember why it threw when no flags were passed?

lib/Runtime/ByteCode/ByteCodeEmitter.cpp
boingoing: No, a module export can't be declared in a block scope (or anywhere except global scope). This line is for imports. When we import something, we create a const decl for it (Parser::CreateModuleImportDeclNode) and mark the symbol for that binding as symbol export storage. We get to this line when we're initializing const declarations at the top of the function. But for module imports, we don't want to initialize a root property (via InitUndeclRootConstFld) because the value is going to be loaded from module export var slots and not stored on the root object.

pleath: Aha. Got it. Sounds good.

lib/Runtime/ByteCode/ByteCodeEmitter.cpp
boingoing: This happens when we're trying to emit a call target when the symbol is an imported binding. We try to emit it as some kind of LdRootMethodFld but should be getting the method from the module storage instead. I can make the logic a little cleaner here, but not sure how to remove it completely.

lib/Parser/Parse.cpp
boingoing: Agree, I would like these syntax errors to all be much more descriptive.

Yongqu:
Change looks good from runtime part. I'll leave the remaining bytecode change to Paul.

lib/Runtime/Language/InterpreterStackFrame.cpp
boingoing: Added a runtime check and FatalInternalError in failure case for this one.

lib/Runtime/ByteCode/ByteCodeEmitter.cpp
pleath: I just think this should be handled upstream. What does the call stack look like in this case?

lib/Runtime/Language/InterpreterStackFrame.cpp
pleath: This should also get a FatalInternalError like the StModuleSlot case.

lib/Runtime/Language/InterpreterStackFrame.cpp
pleath: Thanks.

lib/Parser/Parse.cpp
boingoing: We probably discussed it but somehow I failed to actually implement it this way. This is a great simplification, though, I love it. Made the change.

lib/Parser/Parse.cpp
boingoing: With the declaration flag, it throws if the function doesn't have a name. With the no name flag, it throws if the function does have a name. I don't remember where it throws if you pass no flag at all, I think it throws if the function has no name (same as declaration flag).

lib/Runtime/ByteCode/ByteCodeEmitter.cpp
boingoing: The stack isn't very big. Inside EmitCallTarget, we have a direct call to EmitMethodFld which is how we arrived here. Stack looks like this (hope formatting doesn't look terrible):

	Chakra.dll!EmitMethodFld(ParseNode * pnode, unsigned int callObjLocation, int propertyId, ByteCodeGenerator * byteCodeGenerator, FuncInfo * funcInfo, bool registerCacheIdForCall) Line 6992	C++
 	Chakra.dll!EmitCallTarget(ParseNode * pnodeTarget, int fSideEffectArgs, unsigned int * thisLocation, unsigned int * callObjLocation, ByteCodeGenerator * byteCodeGenerator, FuncInfo * funcInfo) Line 7256	C++
 	Chakra.dll!EmitCall(ParseNode * pnode, unsigned int rhsLocation, ByteCodeGenerator * byteCodeGenerator, FuncInfo * funcInfo, int fReturnValue, int fEvaluateComponents, int fHasNewTarget, unsigned int overrideThisLocation) Line 7573	C++
 	Chakra.dll!Emit(ParseNode * pnode, ByteCodeGenerator * byteCodeGenerator, FuncInfo * funcInfo, int fReturnValue, bool isConstructorCall, ParseNode * bindPnode) Line 9881	C++
 	Chakra.dll!ByteCodeGenerator::EmitTopLevelStatement(ParseNode * stmt, FuncInfo * funcInfo, int fReturnValue) Line 956	C++

lib/Parser/Parse.cpp
boingoing: Sounds good about the IdentPtr equality. Yes, export statements are only valid at global scope.

lib/Parser/Parse.cpp
ianwjhalliday: @abchatra felt it was too early to remove support for disabling the class feature.  I disagree but cannot defend my opinion with strong reason to remove it.  It _might_ be beneficial in the future to help reduce website bugs because of eval of classes.  Seems weak, but removing it just for aesthetic and a little less distraction in the code doesn't persuade us enough to make the call on it.

ianwjhalliday: My instinct is that these switches will simply rot over time if they are not being regularly exercised and so that's why I'd prefer to remove them.

lib/Parser/Parse.cpp
ianwjhalliday: A module default parse flag would be appropriate to allow it to parse the +Default grammar.

I would prefer that to rewinding.  Scanner/parser rewind should only be used when absolutely necessary imo.

Alternatively we could consider adding token look ahead to the scanner/parser.  The async keyword in particular has me thinking about this because everywhere `async` can appear before `function` we need to scan two tokens instead of one to make a decision.  If we had one token look ahead we wouldn't need a save/rewind for parsing `async`.  Of course, in this case if we wanted to take advantage of that we would need two token look ahead because of the generator star possibility.

This look ahead idea is just a thought out loud though.

ianwjhalliday: And ditto on the module default flag for classes and other cases where +Default grammar rules apply.

pleath: It won’t throw if there’s no flag and no name. Anonymity is legal for non-declaration (function expression). I’d be curious to see what it is. If we can’t make that work for the initial check-in, can we open an issue to get rid of the rewind?

--Paul

lib/Runtime/ByteCode/ByteCodeEmitter.cpp
pleath: Okay. This must be in the knopName case of EmitCallTarget. EmitLoadInstance gives us the global object as the owning instance. Can you fix that in EmitLoadInstance (i.e., set instLocation to Js::Constants::NoRegister in the import/export case)? That should cause us to load the target the right way via Emit called on the knopName.

--Paul

lib/Parser/Parse.cpp
pleath: Cool.

--Paul

lib/Parser/Parse.cpp
pleath: Okay. Actually, we should talk about this more in person, so I understand the cases we have to handle. I don't want to hold up the change for this if it's functionally correct, but can we open an issue to get rid of the n^2 search?

lib/Runtime/ByteCode/ByteCodeEmitter.cpp
boingoing: Ok, cool. That seems to work ok.

lib/Parser/Parse.cpp
boingoing: Agreed about rewinding. I tested it out again and the issue is that, while we can parse both anonymous and named functions when we pass no flags to ParseFncDecl, we do not create a local binding for the named function without the fFncDeclaration flag. That means the exporting module won't be able to access the function via the local name (or any name for that matter).

export default function foo() { };
foo(); <-- throw ReferenceError

As I understand it, we should continue binding the default function to the local name foo so we need to pass the declaration flag if the function has a name.

pleath: Ah. Makes sense. Yes, I think the right fix is to create a flag that means, “Create a binding as though this were a declaration, but don’t require a name.”

--Paul

lib/Runtime/ByteCode/ByteCodeEmitter.cpp
pleath: Good. (I think fixing it this way may very likely save problems down the road where EmitLoadInstance is doing the wrong thing in other cases.)

--Paul

lib/Parser/Parse.cpp
boingoing: Sure, thanks

lib/Parser/Parse.cpp
boingoing: Agreed. That flag shouldn't be too hard to build and would clean up this case very nicely, but I plan to track it as a bugfix.

PR #458, opened 2016-03-04
Add RegExp.prototype[@@replace]

goyakin:
This change implements both versions of `RegExp.prototype[@@replace]` where the `replaceValue` argument could be either a String or a callable.


goyakin:

akroshg:
Could you provide the perf problem you have improved? Where and what benchmark the perf problem perceived and how much it is improved?

lib/Runtime/Library/JavascriptRegularExpression.cpp
```diff
@@ -642,155 +663,94 @@ namespace Js
...
-        }
-
-        return HasObservableFlag(
-            instance,
-            PropertyIds::unicode,
-            scriptContext->GetLibrary()->GetRegexUnicodeGetterFunction(),
-            scriptContext);
+        JavascriptLibrary* library = regexPrototype->GetLibrary();
+        return regexPrototype->GetSlot(library->GetRegexGlobalGetterSlotIndex())
+               != library->GetRegexGlobalGetterFunction();
```
akroshg: Is removal the check of IsES6UnitcodeExtensionsEnabled  intentional?

lib/Runtime/Library/RegexHelper.cpp
```diff
@@ -792,8 +782,234 @@ namespace Js
...
+            resultBuilder.Append(replace);
+        };
+        return RegexEs6ReplaceImpl(scriptContext, thisObj, input, appendReplacement, /* noResult */ false);
+    }
+
+    template<typename ReplacementFn>
+    Var RegexHelper::RegexEs6ReplaceImpl(ScriptContext* scriptContext, RecyclableObject* thisObj, JavascriptString* input, ReplacementFn appendReplacement, bool noResult)
+    {
+        bool global = JavascriptRegExp::GetGlobalProperty(thisObj, scriptContext);
+        bool unicode = false; // Dummy value. It isn't used below unless "global" is "true".
```
akroshg: should say default instead of dummy

lib/Runtime/Library/RegexHelper.cpp
```diff
@@ -792,8 +782,234 @@ namespace Js
...
+
+                int64 index = JavascriptConversion::ToLength(
+                    JavascriptOperators::GetProperty(resultObj, PropertyIds::index, scriptContext),
+                    scriptContext);
+                CharCount position = max(
+                    min(JavascriptRegExp::GetIndexOrMax(index), inputLength),
+                    (CharCount) 0);
+
+                // Capture groups can be referenced using at most two digits.
+                uint64 maxNumberOfCaptures = 99;
```
akroshg: you can move this variable outside of map callback and can make it const.

lib/Runtime/Library/RegexHelper.cpp
```diff
@@ -792,8 +782,234 @@ namespace Js
...
+                    Var nextCapture = JavascriptOperators::GetItem(resultObj, i, scriptContext);
+                    if (!JavascriptOperators::IsUndefined(nextCapture))
+                    {
+                        nextCapture = JavascriptConversion::ToString(nextCapture, scriptContext);
+                    }
+
+                    if (i <= numberOfCapturesToKeep)
+                    {
+                        captures[i] = nextCapture;
+                    }
```
akroshg: Don't follow why we are still looping if the i becomes > numberOfCaptures ? is it just for observability ?

lib/Runtime/Library/RegexHelper.cpp
akroshg: Just make sure if this cap needed if user have provided the replace function.

goyakin:
@akroshg In ES6, `RegExp.prototype[@@replace]` and, indirectly, `String.prototype.replace` are supposed to do lots of observable operations, e.g., calling the `exec` method. These operations slow down the execution quite a lot.

This wasn't the case in ES5. Since nothing on a RegExp instance could be overridden, once we went into C++, we stayed there until we returned a result.

The way I added support for the new version is that I first check, in `RegexHelper::RegexReplaceImpl`, whether anything is overridden and run the new version if so. I used to walk the whole prototype chain and call `GetProperty` for each property that could be overridden. This added some minor overhead to the previous RegExp symbol methods I implemented, but it was noticeable with `RegExp.prototype[@@replace]`. By directly retrieving properties from the slots, we're now almost on part with the original perf.

In Octane Regexp benchmark, the improvement is about 50%:

```
Octane            Left score        Right score       ∆ Score   ∆ Score %  Comment
----------------  ----------------  ----------------  --------  ---------  --------
Regexp              3756.67 ±0.30%    5662.50 ±0.15%   1905.83     50.73%  Improved
```

lib/Runtime/Library/RegexHelper.cpp
goyakin: Yes, since `ToString` is observable, we need to run that on all captures. However, we don't need to keep a capture if it isn't possible to refer to it.

lib/Runtime/Library/RegexHelper.cpp
goyakin: To me, _default_ would mean it'll be used. However, `unicode` isn't used at all anywhere in this function unless `global` is `true`. I just picked `false` to initialize `unicode` and wanted to mention why :wink:

lib/Runtime/Library/JavascriptLibrary.h
```diff
@@ -381,15 +382,22 @@ namespace Js
...
         JavascriptFunction* regexExecFunction;
+        int regexExecSlotIndex;
         JavascriptFunction* regexFlagsGetterFunction;
+        int regexFlagsGetterSlotIndex;
         JavascriptFunction* regexGlobalGetterFunction;
+        int regexGlobalGetterSlotIndex;
         JavascriptFunction* regexStickyGetterFunction;
+        int regexStickyGetterSlotIndex;
         JavascriptFunction* regexUnicodeGetterFunction;
+        int regexUnicodeGetterSlotIndex;
```
ianwjhalliday: I would prefer if these two groups of fields were not interleaved in their pairs, but a pair of groups.  I think that will be faster to read.

lib/Runtime/Library/JavascriptLibrary.h
```diff
@@ -718,11 +728,17 @@ namespace Js
...
         JavascriptFunction* GetRegexExecFunction() const { return regexExecFunction; }
+        int GetRegexExecSlotIndex() const { return regexExecSlotIndex;  }
         JavascriptFunction* GetRegexFlagsGetterFunction() const { return regexFlagsGetterFunction; }
+        int GetRegexFlagsGetterSlotIndex() const { return regexFlagsGetterSlotIndex;  }
         JavascriptFunction* GetRegexGlobalGetterFunction() const { return regexGlobalGetterFunction; }
+        int GetRegexGlobalGetterSlotIndex() const { return regexGlobalGetterSlotIndex;  }
         JavascriptFunction* GetRegexStickyGetterFunction() const { return regexStickyGetterFunction; }
+        int GetRegexStickyGetterSlotIndex() const { return regexStickyGetterSlotIndex;  }
         JavascriptFunction* GetRegexUnicodeGetterFunction() const { return regexUnicodeGetterFunction; }
+        int GetRegexUnicodeGetterSlotIndex() const { return regexUnicodeGetterSlotIndex;  }
```
ianwjhalliday: Same here, two groups instead of interleaved

lib/Runtime/Library/JavascriptLibrary.cpp
```diff
@@ -3873,10 +3876,14 @@ namespace Js
         JavascriptFunction ** builtinFuncs = library->GetBuiltinFunctions();
 
         library->AddMember(regexPrototype, PropertyIds::constructor, library->regexConstructor);
+        library->regexConstructorSlotIndex = 0;
+        Assert(regexPrototype->GetSlot(library->regexConstructorSlotIndex) == library->regexConstructor);
```
ianwjhalliday: Seeing the hard coding of these slot indices in the code for real now is making me reconsider.  It is rather clumsy looking isn't it?

I wouldn't block this PR on it, but it would be worth considering revisiting this to get the indices from the propertyid programatically instead.  Adding a new GetAccessorPropertyIndexes API probably won't be fun.

lib/Runtime/Library/JavascriptRegularExpression.cpp
```diff
@@ -642,155 +663,94 @@ namespace Js
...
+        JavascriptLibrary* library = instance->GetLibrary();
 
-    bool JavascriptRegExp::HasObservableExec(RecyclableObject* instance, ScriptContext* scriptContext)
-    {
-        return HasObservableValue(
-            instance,
-            PropertyIds::exec,
-            scriptContext->GetLibrary()->GetRegexExecFunction(),
-            scriptContext);
+        if (instance->GetType() != library->GetRegexType())
```
ianwjhalliday: `JavascriptLibrary` instances are per `ScriptContext` so in the case that the instance is a marshalled object from a different `ScriptContext` we must get the right context's type.  I think marshalled objects return their owning context's types and such, but you should verify with a sanity check unit test.

Same for the built-in function objects that are used for testing below.

ianwjhalliday:
I only took a skim of this PR.  Looks good.  I like the tests.

lib/Runtime/Library/JavascriptRegularExpression.cpp
goyakin: No, it wasn't. Fixed this in ef574a172a73ec4fb6a20381ddc93c5942dabad9.

lib/Runtime/Library/RegexHelper.cpp
goyakin: I made the variable `const`, but kept it where it currently is since, I think, it's better for it to be closer to where it is used.

lib/Runtime/Library/JavascriptLibrary.h
goyakin: I was thinking about that too. Readability now wins (dbd1646e6870c92818af6fef48cb2702952289cc).

lib/Runtime/Library/RegexHelper.cpp
goyakin: As discussed offline, I'll make the change regarding the limit for functional replace as a bug fix later.

lib/Runtime/Library/JavascriptLibrary.cpp
goyakin: :smile: I agree that the code doesn't look good with the hard coded values and would require updates if a new property is added in between. It doesn't seem like it would be quick to add such an API, so I think hard coding is enough for now. I'll get back to this later.

PR #461, opened 2016-03-04
Monorphic type check fix

pleath:
Since we don't want to perform monomorphic type checks at points where the type is already available, avoid clearing the monomorphic type check bucket in such a case.


pleath:
@rajatd @LouisLaf

rajatd:
:+1:

PR #462, opened 2016-03-05
Reset variable lifetime to be valid if not dead in DeadStorePhase

ThomsonTan:
CSE could add new reference to previous value def, thus extend its lifetime and make the previous use to be valid and not the last use. The DeadStorePhase is supposed to reset the last use information for each variable, but it assumes the variable is live and doesn't reset it if it is not dead. Miss resurrecting such variables will lead to incorrect lifetime calculated by SccLiveness and fail PrelowerPeeps optimization.


ThomsonTan:
@LouisLaf @pleath

rajatd:
Nice. Looks good. Could you add a unittest?

PR #463, opened 2016-03-05
Bug with missing StatementBoundary in native span sequence

rajatd:
When we get rid of the arguments overhead for an inlinee, we
are inadvertently missing to add the StatementBoundary for the inlinee
call to the native span sequence. If the inlinee throws, this leads to a difference in line
numbers in the stack trace between interpreter and jit.

Fixed by adding the missing StatementBoundary to the span sequence

[Review on Reviewable](https://reviewable.io/reviews/microsoft/chakracore/463)


rajatd:
@abchatra, could you review this?

abchatra:
LGTM

Issue #313, opened 2016-02-13
Optimize LdHeapArguments in presence of formal parameters


kunalspathak:
Today inside function that reads `arguments` object, we need to create heap object that copies arguments on the stack into this heap object. The reason being the function can return the `arguments` and caller should have access to those arguments. However for functions which don't return `arguments`, ChakraCore does optimization where it doesn't create heap object (and thus skip copying of stack arguments into this heap object). 

In below example, `foo` doesn't return `arguments` object and hence we optimize `... = arguments`cases by loading the values from stack instead of creating heap objects for them and then using it.

``` js
function foo()
{
  // ...
  x = arguments[i]
  // ...
}
```

But if function have formal parameters, we don't do this optimization. While investigating Microsoft/node#6, I noticed that there were lot of hot javascript functions that takes formal parameters and has reads from `arguments` objects. 

E.g. Here are some functions in node code :
- https://github.com/nodejs/node-chakracore/blob/chakracore-master/lib/timers.js#L421
- https://github.com/nodejs/node-chakracore/blob/chakracore-master/src/node.js#L506
- https://github.com/nodejs/node-chakracore/blob/chakracore-master/lib/net.js#L615
- https://github.com/nodejs/node-chakracore/blob/chakracore-master/lib/events.js#L118

For every call we create heap objects for arguments and hence spend lot of time in this instead of executing the functions. In my experiment, I replaced `arguments` reads with formal parameters and noticed the node+chakracore performance improves by 1.5X times.

We should optimize such scenarios to get better performance improvement.

kunalspathak:
@LouisLaf , @satheeshravi  - FYI.

PR #464, opened 2016-03-05
Stack Arguments Optimization extended - With Formal's presence.

satheeshravi:
Issue:

We optimize arguments object elements read by reading the values directly from the stack slot instead of creating a heap arguments object(as done in general case).
But this optimization works only in the absence of formals. We have few hot functions in node as reported by Kunal, that has arguments object and formals.

Fix:
This change extends the stack arguments optimization even when formals are
present. But we give up when there are writes to formals, as mapping will
be wrong.
Also, formals' accesses are still from the Activation object and hence is not optimized for this iteration and is the next workitem.

Test:
Exprgen looks fine. I have started other runs.

Perf:
Console perf runs look fine.

### DIFF
--- a/lib/Parser/Parse.cpp
+++ b/lib/Parser/Parse.cpp
@@ -685,6 +685,7 @@ void Parser::InitNode(OpCode nop,ParseNodePtr pnode) {
     pnode->notEscapedUse = false;
     pnode->isInList = false;
     pnode->isCallApplyTargetLoad = false;
+    pnode->isArgElemOrArgLenLoad = false;
 }
 
 // Create nodes using Arena
@@ -8252,11 +8253,11 @@ ParseNodePtr Parser::ParseExpr(int oplMin,
             if (nodeType & fnopBin)
             {
                 ParseNodePtr lhs = pnode->sxBin.pnode1;
-
                 Assert(lhs);
+
                 if (lhs->nop == knopDot)
                 {
-                    ParseNodePtr propertyNode = lhs->sxBin.pnode2;                    
+                    ParseNodePtr propertyNode = lhs->sxBin.pnode2;
                     if (propertyNode->nop == knopName)
                     {
                         propertyNode->sxPid.pid->PromoteAssignmentState();
--- a/lib/Parser/Parse.cpp
+++ b/lib/Parser/Parse.cpp
@@ -1744,6 +1744,10 @@ void Parser::BindPidRefsInScopeImpl(IdentPtr pid, Symbol *sym, int blockId, uint
 
         if (ref->IsAssignment())
         {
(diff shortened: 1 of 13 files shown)


satheeshravi:
@abchatra : FYI

satheeshravi:
https://github.com/Microsoft/ChakraCore/issues/313

lib/Parser/Parse.cpp
```diff
@@ -8252,7 +8256,7 @@ ParseNodePtr Parser::ParseExpr(int oplMin,
                 Assert(lhs);
                 if (lhs->nop == knopDot)
                 {
-                    ParseNodePtr propertyNode = lhs->sxBin.pnode2;
+                    ParseNodePtr propertyNode = lhs->sxBin.pnode2;
```
rajatd: nit: blank space at the end

lib/Runtime/Language/JavascriptOperators.cpp
```diff
@@ -6419,17 +6419,21 @@ namespace Js
         DynamicObject* frameObject,
         Js::PropertyIdArray *propIds,
         uint32 formalsCount,
-        ScriptContext* scriptContext)
+        ScriptContext* scriptContext,
+        bool isStackArgs)
```
rajatd: what is isStackArgs used for?

test/PerfHint/arguments1.baseline
```diff
@@ -1,7 +1,7 @@
-PerfHint: Not optimized : Arguments object not optimized due to formals {
+PerfHint: Not optimized : Arguments object not optimized {
```
rajatd: Do you know why arguments object is still not optimized in this function?

rajatd:
Are you planning to enable inlining scenarios which we don't currently catch because of presence of formals, in this iteration?

pleath:
I don't think we should try that until we've eliminated the ActivationObject creation.

lib/Runtime/Language/JavascriptOperators.cpp
satheeshravi: Removed it. Uncleaned change.

test/PerfHint/arguments1.baseline
satheeshravi: This message is changed again. With formals and arguments object access - This function will not be optimized with StackArgs

satheeshravi: I tried adding a PerfHint parser, but i was however not successful in figuring out the bytecode offset (probably this is not computed when I am accessing it!)

abchatra:
@satheeshravi Can we close this PR and reopen the new one later?

satheeshravi:
Yes. Sure.

Issue #424, opened 2016-03-01
Fix error message for changing [[Writable]] to true from false for non-configurable properties


rajatd:
Currently the error message we print is "Cannot modify non-writable property" while it should be "Cannot redefine non-configurable property"

ljharb:
By spec, nonconfigurable properties certainly can be changed from `writable: true` to `writable: false` - it's only that nonconfigurable nonwritable properties can't be made writable.

You can see this with:
```js
'use strict';
var obj = Object.defineProperty({}, 'foo', { configurable: false, writable: true });
console.log(Object.getOwnPropertyDescriptor(obj, 'foo'));
Object.defineProperty(obj, 'foo', { writable: false });
console.log(Object.getOwnPropertyDescriptor(obj, 'foo'));
```

ljharb:
oops, never mind, I misread your issue title :-) will delete my comment

ljharb:
It might be useful to say something more specific like "Cannot make nonconfigurable property writable" when attempting to set `writable` to `true` - to highlight the difference where you can change it to `false`.

rajatd:
Thanks @ljharb . Sure, we could have that. However, we have a macro for the error message to be thrown when trying to redefine non-configurable properties (except for the cases where it is allowed). Having a different one just for this case doesn't really seem worth it. Other browsers also just say "cannot redefine....".

ljharb:
Fair. I'd say every error message should be as helpful as possible, even if other browsers don't live up to that standard, but I don't truly understand the tradeoffs there.

EdMaurer:
Not enough impact to add to 1.4.1

PR #465, opened 2016-03-05
Fix error message per issue #424

mwksl:
This should fix the error messages when changing [[Writable]] as mentioned in #424.


rajatd:
Hi @mwksl , thanks for your pull request. I looked at your changes and it seems that it would while it would fix the issue #424 , it would give wrong error message for `obj.x = 1` where `x` was non-writable. We should detect the case when we are changing [[Writable]] to true from false for non-configurable properties, and change the error message for that case.

abchatra:
Thank you for your contribution. Can you please sign the cla?

abchatra:
@mwksl  any update on signing the cla and this PR? If not we will close the PR this week.

abchatra:
No update here, hence closing the PR

PR #466, opened 2016-03-05
Fix nested function array allocation

leirocks:


leirocks:

lib/Runtime/Base/FunctionBody.cpp
```diff
@@ -1143,7 +1143,7 @@ namespace Js
         if (nestedCount > 0)
         {
             nestedArray = RecyclerNewPlusZ(m_scriptContext->GetRecycler(),
-                nestedCount*sizeof(FunctionProxy), NestedArray, nestedCount);
+                (nestedCount-1)*sizeof(FunctionProxy*), NestedArray, nestedCount);
```
curtisman: Is it possible for NestedArray to have 0 length array to avoid this -1 here?

leirocks: C++ don't allow that? No?

curtisman: Have you tried? 
There are other example in the code base that have declares a 0 length array field.
e.g. TypePath's "assignment" field.

leirocks: looks c++ 11 does not allow it, but the compilers do allow zero length array in structure defination. updated the PR

curtisman:
LGTM

PR #471, opened 2016-03-06
Expect AssignmentExpression after 'of' in a for-of loop and inside a computed name

### LINKED ISSUE #147: for-of loop incorrectly recognizes Expression grammar for iterable expression; should be AssignmentExpression
See tc39/test262#473

For example this should be a `SyntaxError`:

```javascript
for (let x of 0, []) { }
```

### LINKED ISSUE #472: Expression shouldn't be allowed as computed property name
Chakra doesn't throw a SyntaxError for this code:
```js
({ [ 1, 2 ]: 3 });
```
but the specification defines [`ComputedPropertyName`](https://tc39.github.io/ecma262/#prod-ComputedPropertyName) as `[ AssignmentExpression ]`

nicolo-ribaudo:
Fixes #147
Fixes #472

### DIFF
--- a/lib/Parser/Parse.cpp
+++ b/lib/Parser/Parse.cpp
@@ -3675,7 +3675,7 @@ Parser::MemberNameToTypeMap* Parser::CreateMemberNameMap(ArenaAllocator* pAlloca
 template<bool buildAST> void Parser::ParseComputedName(ParseNodePtr* ppnodeName, LPCOLESTR* ppNameHint, LPCOLESTR* ppFullNameHint, ulong *pNameLength, ulong *pShortNameOffset)
 {
     m_pscan->Scan();
-    ParseNodePtr pnodeNameExpr = ParseExpr<buildAST>(koplNo, nullptr, TRUE, FALSE, *ppNameHint, pNameLength, pShortNameOffset);
+    ParseNodePtr pnodeNameExpr = ParseExpr<buildAST>(koplCma, nullptr, TRUE, FALSE, *ppNameHint, pNameLength, pShortNameOffset);
     if (buildAST)
     {
         *ppnodeName = CreateNodeT<knopComputedName>(pnodeNameExpr->ichMin, pnodeNameExpr->ichLim);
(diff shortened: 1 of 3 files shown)


nicolo-ribaudo:
This pr is ready to be reviewd

lib/Parser/Parse.cpp
```diff
@@ -3687,7 +3687,7 @@ template<bool buildAST> void Parser::ParseComputedName(ParseNodePtr* ppnodeName,
         *ppFullNameHint = FormatPropertyString(*ppNameHint, pnodeNameExpr, pNameLength, pShortNameOffset);
     }
 
-    ChkCurTokNoScan(tkRBrack, ERRsyntax);
+    ChkCurTokNoScan(tkRBrack, ERRnoRbrack);
```
ianwjhalliday: Nice

test/es6/objlit.js
```diff
@@ -123,7 +115,9 @@ var tests = [
             };
             assert.areEqual(protoObj.abc, 123, "__proto__ get assigned when used as a normal production");
 
-            assert.throws(function () { eval("var b = { ['str'] }"); },            SyntaxError, "Invalid computed identifier shorthand");
+            assert.throws(function () { eval("var b = { ['str'] }"); }, SyntaxError, "Invalid computed identifier shorthand");
```
ianwjhalliday: Would you mind adding the error message argument to this `assert.throws`?

test/es6/objlit.js
```diff
@@ -75,36 +75,28 @@ var tests = [
             var obj = {
                 ["foo" + "bar"]         : 1,
                 [1 * 10 * 10]           : 2,
-                ["foobar", "notfoobar"] : 3, // Should evaluate to notfoobar
```
ianwjhalliday: To ensure we're parsing the right starting grammar production we should have a test that uses one of the production rules that are unique to that production.  Simplest would be a simple assignment, so something like
`[a = 'baz']: 3,`.

Here and the other cases below.

ianwjhalliday:
Just a couple minor comments on the tests.  Looks great, nice a simple.  Thanks for taking the time to help us knock this issue off the list.

nicolo-ribaudo:
@ianwjhalliday I updated the tests as you asked

ianwjhalliday:
Sorry for the delay.  Looks great!  I'll get this merged in later today if I can.

ianwjhalliday:
@nicolo-ribaudo thanks again!

PR #474, opened 2016-03-07
LdSuperCtor: Update interpreter error message to match JIT

suwc:
LdSuperCtor: Update interpreter error message to match JIT

TypeError thrown by operator 'LdSuperCtor' differs in Interpreter vs. in JIT.
Change operator code to align the messages.

### DIFF
--- a/lib/Runtime/Language/JavascriptOperators.cpp
+++ b/lib/Runtime/Language/JavascriptOperators.cpp
@@ -9179,7 +9179,7 @@ CommonNumber:
 
         if (superCtor == nullptr || !IsConstructor(superCtor))
         {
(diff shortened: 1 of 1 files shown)


suwc:
@rajatd, @abchatra, this should be quick - could you take a look pls?
This is to fix the mismatch b.t.w. interpreter and JIT in TypeError message thrown by LdSuperCtor. Decided to take this route because the ROI of changing error-throwing helper call format from unary to binary (such that extra string can be passed) has not been justified.

rajatd:
The change itself looks good. I'm fine if you guys are fine changing the error message.

suwc:
Changing the error-throwing helper for JIT may make sense, although I see it as lower priority unless we need it in more often than isolated cases like this.

PR #475, opened 2016-03-07
Enabling Object.values and Object.entries by-default ON.

akroshg:


akroshg:
@abchatra - can you take a look at this please?

ljharb:
(ﾉ◕ヮ◕)ﾉ*:・ﾟ✧ yay!

lib/Common/ConfigFlagsList.h
```diff
@@ -868,6 +869,7 @@ FLAGPR_REGOVR_EXP(Boolean, ES6, ES6FunctionNameFull    , "Enable ES6 Full functi
 FLAGPR           (Boolean, ES6, ES6Generators          , "Enable ES6 generators"                                    , DEFAULT_CONFIG_ES6Generators)
 FLAGPR_REGOVR_EXP(Boolean, ES6, ES7ExponentiationOperator, "Enable ES7 exponentiation operator (**)"                , DEFAULT_CONFIG_ES7ExponentionOperator)
 FLAGPR_REGOVR_EXP(Boolean, ES6, ES7Builtins            , "Enable ES7 built-ins"                                     , DEFAULT_CONFIG_ES7Builtins)
+FLAGPR_REGOVR_EXP(Boolean, ES6, ES7ValuesEntries       , "Enable ES7 Object.values and Object.entries"              , DEFAULT_CONFIG_ES7ValuesEntries)
```
ianwjhalliday: `FLAGPR_REGOVR_EXP` means it is still off by default, only on when the experimental flag registry key is enabled.

If you meant real, true on-by-default you need to change this to `FLAGPR`

akroshg: ah, isn't it the same for ES7ExponentiationOperator? Also I have removed the flag from the test and test is still passing.

ljharb: (not that ti's a big deal, but exponentiation is in ES7 and values/entries are not - perhaps a prefix like `ESNext` would be best long-term)

ianwjhalliday: @ljharb agreed.  This is something we've mulled over a touch but keep punting on.  Going back and changing existing flags is slightly a pain because our test tools over time start to take dependency on them, so a change needs to make sure we don't break our tools.  Which makes me prefer some general always applicable convention, I'm thinking simply ES, but then they lose their groupings.

And there are features that miss revisions, like async functions will now be ESNext instead of 7, but we already have tests for it under test/es7.

Naming is hard.

ianwjhalliday: @akroshg hmm maybe I am mistaken and the registry key only overrides a default of false to true.  Regardless I would prefer we remove the experimental flag aspect from both this new flag and exponentiation, to avoid any gotchas.

ljharb:
Hooray! How can I be notified when this commit lands in a release?

PR #477, opened 2016-03-07
LEB128 and Function/Body Split

commonlisp:
- Added decoding of signed LEB128
- Using LEB128 for counts, indices, and offsets in several places
- Splitting up function declaration (function header)
  and function body

### DIFF
--- a/lib/WasmReader/WasmBinaryReader.cpp
+++ b/lib/WasmReader/WasmBinaryReader.cpp
@@ -85,6 +85,9 @@ WasmBinaryReader::ReadFromModule()
         }
         else
         {
+            // TODO: Should check that previous section, if any, was of the correct
+            // length by comparing against m_moduleState.byteLen.
+
             // next section
             sectionId = m_moduleState.secId = SectionHeader();
             if (sectionId == bSectMemory || sectionId == bSectEnd)
@@ -94,7 +97,7 @@ WasmBinaryReader::ReadFromModule()
             else
             {
                 m_moduleState.size = LEB128(length);
-                if (sectionId == bSectFunctions)
+                if (sectionId == bSectFunSigs)
                 {
                     m_moduleInfo->SetFunctionCount(m_moduleState.size);
                 }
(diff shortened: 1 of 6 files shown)


MikeHolman:
lgtm

lib/WasmReader/WasmBinaryReader.cpp
```diff
@@ -137,20 +137,28 @@ WasmBinaryReader::ReadFromModule()
...
             m_moduleState.count++;
+            break;
 
+        case bSectFunBodies:
+            if (!m_visitedSections->Test(bSectFunSigs))
+            {
+                ThrowDecodingError(L"Function signatures section missing before function bodies");
+            }
+            FunctionBodyHeader();
+            m_moduleState.count++;
```
nmostafa: Shouldn't we check that the count of FuncSigs and FuncBodies match ?

lib/WasmReader/WasmBinaryReader.cpp
```diff
@@ -530,39 +546,63 @@ WasmBinaryReader::FunctionHeader()
...
+            break;
+        default:
+            ThrowDecodingError(L"Unexpected local type");
         }
     }
 
+    for (UINT i = 0; i < i32Count; i++)
+    {
+        m_funcInfo->AddLocal(Wasm::WasmTypes::I32);
+    }
```
nmostafa: Maybe I am missing something, but shouldn't' we have a list of funcInfos signatures contructed from FuncSigs that we attribute function bodies to here ? 
See [Luke's PR](https://github.com/WebAssembly/design/blob/0d96ed985368fc636a38de31970dc35274018ada/BinaryEncoding.md#function-bodies-section)

MikeHolman: locals are defined in the body, not the signature

nmostafa: Not what I meant. There are two separate sections now: One for signatures, and another for bodies. They are associated by the order of the entries. So if the function signature section has 3 functions, the reader should internally keep a list of 3 functions, then go back an add locals to them. Currently, I see one funcInfo tracked by the reader.

MikeHolman: oh, you're absolutely right. We should do `moduleInfo->GetFuncInfo(n)` lookup.

lib/WasmReader/WasmBinaryReader.cpp
commonlisp: Good catch. I'm checking that now.

PR #478, opened 2016-03-08
Old issue in forward pass objtypespec analysis

pleath:
Old issue in forward pass objtypespec analysis exposed by recent backward pass fixes. If we have a monomorphic check upstream and reach a point where a monorphic check expects a different type, we can't replace the checked type in the value info. We have to indicate a type mismatch.


pleath:
@rajatd @LouisLaf

LouisLaf:
Looks good.

PR #480, opened 2016-03-08
Check if ScriptContext is actually closed instead of just marked for close in ThreadContext destructor

agarwal-sandeep:
Check if ScriptContext is actually closed instead of just marked for close
in ThreadContext destructor so that Debugmanager can be released properly
on call to ScriptContext internal close.
Don't delete debug manager in ThreadContext destructor and let ScriptContext
delete it on close because it is created on ScriptContext constructor and
is ref counterd.

### DIFF
--- a/lib/Runtime/Base/ThreadContext.cpp
+++ b/lib/Runtime/Base/ThreadContext.cpp
@@ -366,7 +366,7 @@ ThreadContext::~ThreadContext()
     {
         for (Js::ScriptContext *scriptContext = scriptContextList; scriptContext; scriptContext = scriptContext->next)
         {
-            if (!scriptContext->IsClosed())
+            if (!scriptContext->IsActuallyClosed())
             {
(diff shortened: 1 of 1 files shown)


lib/Runtime/Base/ThreadContext.cpp
```diff
@@ -1974,9 +1970,13 @@ void ThreadContext::ReleaseDebugManager()
         {
             this->recyclableData->returnedValueList = nullptr;
         }
-        this->debugManager->Close();
-        HeapDelete(this->debugManager);
-        this->debugManager = nullptr;
+
+        if (this->debugManager != nullptr)
```
akroshg: I do not understand the scenario here - this is the only place we delete the debugManager. which is when refcounter became 0. 
So at this point can it be null ? Or it is just conservative check (guarded by an assert above) ?

agarwal-sandeep:
@Yongqu can you please take a look

akroshg:
Looks good to me.

PR #481, opened 2016-03-08
Wasm section order

Cellule:
This uses the fact that sections needs to be in a certain order to be valid.
Will not validate if the order is wrong.
Tracing Wasm `-trace:wasmreader` will output all the sections in the binary file at startup.
Currently this breaks sections with a number of entries (except signatures section).
The reason being that each section should know how to parse the binary and we shouldn't make a common code with branch to read one uint32.
The Sexpr reader will not work either at this time since it is not our priority, I focused on the binary reader.


lib/WasmReader/BaseWasmReader.h
```diff
@@ -8,19 +8,36 @@
...
+    #include "WasmSections.h"
+        bSectLimit
+    };
+
+    class BaseWasmReader
+    {
+    public:
+        virtual void InitializeReader() = 0;
+        virtual bool ReadNextSection(SectionCode nextSection) = 0;
+        virtual bool ProcessSection(SectionCode nextSection) = 0;
```
Cellule: ProcessSection means the reader will take care of doing the work for that section. Still need to come up with a nice way to return back to the bytecode generator if needed.

lib/WasmReader/WasmBinaryReader.cpp
```diff
@@ -60,157 +69,208 @@ WasmBinaryReader::WasmBinaryReader(PageAllocator * alloc, byte* source, size_t l
...
+            m_pc = secHeader.end;
+            doRead = !EndOfModule() && secHeader.code != bSectEnd;
+        }
+        m_pc = startModule;
+    }
+#endif
+}
+
+void
+WasmBinaryReader::ThrowDecodingError(const wchar_t* msg, ...)
```
Cellule: I think I just moved this function around by mistake. Essentially I added the var args.

lib/WasmReader/WasmBinaryReader.cpp
```diff
@@ -658,6 +716,7 @@ WasmBinaryReader::LEB128(UINT &length, bool sgn)
     UINT result = 0;
     UINT shamt = 0;
     byte b;
+    length = 0;
```
Cellule: From what I saw in the way we used LEB128, we assumed that the length passed would always have the number of bytes read for this LEB128.
That wasn't the case and introduced bugs.
It could have been a valid use, but since at least 2 dev made that mistake, I think it better to always initialize the length to 0 when reading an LEB128

Cellule:
@commonlisp @MikeHolman could you review this please ?

lib/WasmReader/SExprParser.h
```diff
@@ -31,7 +31,11 @@ namespace Wasm
     public:
         SExprParser(PageAllocator * alloc, LPCUTF8 source, size_t length);
 
-        virtual WasmOp ReadFromModule() override;
+        virtual void InitializeReader() override;
+        virtual bool ReadNextSection(SectionCode nextSection) override;
+        virtual bool ProcessSection(SectionCode nextSection) override;
+        // todo:: Remove and use ReadNextSection and ProcessSection instead
+        virtual WasmOp ReadFromModule();
```
commonlisp: Do we want to keep the override specifier? But since ReadFromModule() is on its way out, it is probably a moot point.

MikeHolman:
this is ok with me. but would like if we get the generation of bodies under this scheme done soon

lib/WasmReader/WasmBinaryReader.cpp
```diff
@@ -60,157 +69,208 @@ WasmBinaryReader::WasmBinaryReader(PageAllocator * alloc, byte* source, size_t l
...
+    SectionHeader secHeader = ReadSectionHeader();
+    if (secHeader.code == bSectInvalid || sectionFlags[secHeader.code] == fSectIgnore)
+    {
+        TRACE_WASM_DECODER(L"Ignore this section");
+        m_pc = secHeader.end;
+        return ReadNextSection(nextSection);
+    }
+    if (secHeader.code < nextSection)
+    {
+        ThrowDecodingError(L"Invalid Section %s", sectionIds);
```
commonlisp: sectionId?

lib/WasmReader/SExprParser.h
Cellule: we can't keep it because it has already been removed from the BaseReader

lib/WasmReader/WasmBinaryReader.cpp
Cellule: @commonlisp  should this be initialized to 1? It seems small number returns 0 for length.

lib/WasmReader/WasmBinaryReader.cpp
```diff
@@ -60,157 +69,208 @@ WasmBinaryReader::WasmBinaryReader(PageAllocator * alloc, byte* source, size_t l
...
     idSize = LEB128(len);
+    if (sectionSize < idSize + len)
+    {
+        ThrowDecodingError(L"Invalid section size");
+    }
     const char *sectionName = (char*)(m_pc);
     m_pc += idSize;
+
     m_moduleState.count = 0;
     m_moduleState.byteLen = sectionSize;
```
commonlisp: Can drop m_moduleState.byteLen in favor of header.start and .end since byteLen is now redundant.

lib/WasmReader/WasmBinaryReader.cpp
```diff
@@ -60,157 +69,208 @@ WasmBinaryReader::WasmBinaryReader(PageAllocator * alloc, byte* source, size_t l
...
-    TRACE_WASM_DECODER(L"Decoding Module");
+    if (EndOfModule() || sectionFlags[nextSection] == fSectIgnore)
+    {
+        return false;
+    }
 
-    SectionCode sectionId;
-    UINT length = 0;
+    // TODO: Should check that previous section, if any, was of the correct
+    // length by comparing against m_moduleState.byteLen.
```
commonlisp: Stale comment referencing m_moduleState.byteLen which is now redundant with secHeader.start and secHeader.end.

lib/WasmReader/WasmBinaryReader.cpp
commonlisp: Good call.

lib/WasmReader/BaseWasmReader.h
```diff
@@ -8,19 +8,36 @@
...
-    WasmNode    m_currentNode;
-    ModuleInfo * m_moduleInfo;
-
-protected:
-    WasmFunctionInfo *  m_funcInfo;
-};
+    enum SectionFlag
+    {
+        fSectNone,
+        fSectIgnore,
```
commonlisp: How are you envisioning extending this flag beyond ignore?

lib/WasmReader/WasmBinaryReader.h
```diff
@@ -104,7 +95,7 @@ namespace Wasm
         private:
             struct ReaderState
             {
-                SectionCode secId;
+                SectionHeader secHeader;
                 UINT32 count;         // current entry
                 UINT32 size;          // number of entries
                 UINT32 byteLen;       // byte length of section excluding the LEB128 length field
```
commonlisp: Yeah, this is redundant now.

lib/WasmReader/WasmBinaryReader.cpp
```diff
@@ -442,7 +501,7 @@ WasmBinaryReader::ResetModuleData()
     m_visitedSections = BVFixed::New<ArenaAllocator>(bSectLimit + 1, &m_alloc);
```
commonlisp: With this ordering of sections, is m_visitedSections still necessary?

lib/Runtime/Base/ScriptContext.cpp
```diff
@@ -1792,6 +1792,8 @@ namespace Js
             Wasm::BaseWasmReader *reader = nullptr;
             Wasm::WasmBytecodeGenerator *bytecodeGen = nullptr;
 
+            bool traceReader = PHASE_TRACE1(WasmReaderPhase);
```
commonlisp: Is it possible to move this into WasmBinaryReader and dispense with m_trace altogether?

lib/WasmReader/WasmBinaryReader.cpp
```diff
@@ -60,157 +69,208 @@ WasmBinaryReader::WasmBinaryReader(PageAllocator * alloc, byte* source, size_t l
...
-    UINT len;
+    UINT len = 0;
     UINT32 sectionSize;
     UINT32 idSize;
 
+    SectionHeader header;
+    header.start = m_pc;
     sectionSize = LEB128(len);
+    header.end = m_pc + sectionSize;
+    if (header.end > m_end)
```
commonlisp: CheckBytesLeft(sectionSize)?

lib/WasmReader/BaseWasmReader.h
Cellule: I didn't know at this time, I just figure it will make it easy to extend if need be.
If we realize we don't need it (ie: when all sections are read) then we could drop it

lib/WasmReader/WasmBinaryReader.cpp
Cellule: It is to validate precedence.
Because although the order is fixed, the presence of a section is not mandatory.
Therefor, you could have section `function_bodies` but missing `function_signatures`

lib/Runtime/Base/ScriptContext.cpp
Cellule: You mean to remove the trace param in the reader constructor and simply use `PHASE_TRACE1(WasmReaderPhase)` in the WasmReader ?

Cellule:
So I decided to add reading the `function_signatures` and `function_bodies` section in this PR to keep it at the same quality that we had before

lib/Runtime/Base/ScriptContext.cpp
commonlisp: Yes, and the m_trace state

Cellule:

naturaltransformation:
LGTM

commonlisp:
LGTM

PR #483, opened 2016-03-08
Specialize pow when exponent is small integer

ThomsonTan:
The current CRT in Windows 10 specializes pow(double, int) by breaking the result in to product with power of 2, then aggregation the sub-results. This leads to impricision after many float multiply operations so it was just removed from CRT which will regress Chakra performance when the CRT fix reaches public. This change does the same optimization for smaller integer which improves imaging-darkrom in kraken by 11%.

### DIFF
--- a/lib/Runtime/Library/JavascriptNumber.cpp
+++ b/lib/Runtime/Library/JavascriptNumber.cpp
@@ -235,7 +235,35 @@ namespace Js
         }
         else if(TryGetInt32Value(y, &intY))
         {
-            return ::pow(x, intY);
+            // For exponent in [-8, 8], aggregate the product according to binary representation
+            // of exponent. This acceleration may lead to significant deviation for larger exponent
+            if (intY >= -8 && intY <= 8)
+            {
+                uint32 uexp;
+                if (intY >= 0)
+                {
+                    uexp = static_cast<uint32>(intY);
+                }
+                else
+                {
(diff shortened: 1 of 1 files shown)


lib/Runtime/Library/JavascriptNumber.cpp
```diff
@@ -235,7 +235,35 @@ namespace Js
...
+            {
+                uint32 uexp;
+                if (intY >= 0)
+                {
+                    uexp = static_cast<uint32>(intY);
+                }
+                else
+                {
+                    uexp = static_cast<uint32>(-intY);
+                }
```
goyakin: I think using the ternary operator here would make this shorter and easier to read, i.e. `uint32 uexp = static_cast<uint32>(intY >= 0 ? intY : -intY)`

ThomsonTan: Feedback taken, thanks.

dilijev:
What is the maximum deviation for the extremes of this range? Is it within epsilon 1e-6 or is there some other arbitrary number we're using for "good enough"?

ThomsonTan:
@dilijev , the current range [-8, 8] comes from analyzing benchmark call, and we found this range could cover most of the call to `Math.pow` with integer exponent. There will be very few multiply operations for this range so it is quite conservative. I am still tuning the integer range for optimization.

PR #485, opened 2016-03-08
Have certain String methods throw when passed a RegExp-like object

goyakin:
Existence of the @@match property on an Object indicates that it's a
RegExp-like object. Certain String methods now disallow use of such an
Object as their "searchString" argument. These methods are:

  * String.prototype.endsWith
  * String.prototype.includes
  * String.prototype.startsWith


goyakin:

akroshg:
Looks good to me.

PR #487, opened 2016-03-09
Decode Import Table Sections

commonlisp:
The new import table entries contain both an imported module name and function name as well as a signature index.

### DIFF
--- a/lib/Runtime/Base/ScriptContext.cpp
+++ b/lib/Runtime/Base/ScriptContext.cpp
@@ -1886,48 +1886,15 @@ namespace Js
             // TODO, refactor this function into smaller functions
             for (uint i = 0; i < wasmModule->functions->Count(); ++i)
             {
-                if (functionArray[i]->wasmInfo->Imported())
-                {
-                    PropertyRecord const * propertyRecord = nullptr;
-                    LPCUTF8 name = functionArray[i]->wasmInfo->GetName();
-
-                    utf8::DecodeOptions decodeOptions = utf8::doAllowInvalidWCHARs;
-
-                    UINT utf16Len = utf8::ByteIndexIntoCharacterIndex(name, strlen((const char*)name), decodeOptions);
-                    LPCWSTR contents = HeapNewArray(WCHAR, (utf16Len + 1));
-                    if (contents == nullptr)
-                    {
-                        Js::Throw::OutOfMemory();
-                    }
-                    utf8::DecodeIntoAndNullTerminate((wchar_t*)contents, name, utf16Len, decodeOptions);
-
-                    GetOrAddPropertyRecord(contents, utf16Len, &propertyRecord);
-                    HeapDeleteArray(utf16Len + 1, contents);
-                    if (!ffi)
-                    {
-                        // TODO: michhol give error message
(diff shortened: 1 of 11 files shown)


lib/WasmReader/WasmBinaryReader.cpp
```diff
@@ -652,6 +662,28 @@ WasmBinaryReader::FunctionBodyHeader()
...
+WasmBinaryReader::ImportEntry()
+{
+    UINT len = 0;
+    UINT sigId = LEB128(len);
+
+    m_funcInfo = Anew(&m_alloc, WasmFunctionInfo, &m_alloc);
+    m_funcInfo->SetModuleName(ReadString());
+    m_funcInfo->SetName(ReadString());
+
+    m_currentNode.func.info = m_funcInfo;
```
MikeHolman: this is not used... we should create another table of WasmFunctionInfo for imports on ModuleInfo.

lib/WasmReader/WasmByteCodeGenerator.cpp
```diff
@@ -87,7 +91,7 @@ WasmBytecodeGenerator::GenerateModule()
     {
         if (m_reader->ReadNextSection((SectionCode)sectionCode))
         {
-            if (!sectionProcess[sectionCode](this, (SectionCode)sectionCode))
+            if (sectionProcess[sectionCode](this, (SectionCode)sectionCode) == psrInvalid)
```
Cellule: This should actually be `!= psrEnd`

Cellule:
I think we should remove the `import` concept from the WasmFunctionInfo since they are no longer associated.
the operator `call_import` uses an index of the import table and not the function table, therefor they need to be stored separately.
I think we should do something similar to what I did for exports (see #489)
We'll have to allocate some space on the frame for the imports.

commonlisp:
Ok, I'll see what I can do to follow that closely.

lib/WasmReader/WasmByteCodeGenerator.cpp
```diff
@@ -496,7 +500,7 @@ WasmBytecodeGenerator::EmitCall()
 
     uint funcNum = Js::Constants::UninitializedValue;
     uint signatureId = Js::Constants::UninitializedValue;
-    bool imported;
+    bool imported = false;
```
MikeHolman: can remove this completely

lib/Runtime/Base/ScriptContext.cpp
```diff
@@ -1952,6 +1919,34 @@ namespace Js
...
+                    Js::Throw::InternalError();
+                }
+                Var modProp = JavascriptOperators::OP_GetProperty(ffi, modPropertyRecord->GetPropertyId(), this);
+                if (!JavascriptFunction::Is(modProp))
+                {
+                    Assert(UNREACHED);
+                    // TODO: michhol figure out correct error path
+                }
+                Var prop = JavascriptOperators::OP_GetProperty(modProp, propertyRecord->GetPropertyId(), this);
+                localModuleFunctions[i] = prop; // TODO: Fix this.
```
MikeHolman: fix this?

MikeHolman:
lgtm

PR #490, opened 2016-03-09
sectionProcess error handling issue

commonlisp:
It appears that sectionProcess is testing for a bool instead of
a psr.


MikeHolman:
lgtm

PR #491, opened 2016-03-09
linux: re-enable building lib/Parser

jianchun:
Re-enable building lib/Parser after completing char16 replacements.

- PAL code depends on min/max macros in pal.h. Other chakracore code uses
  safer template min/max functions. To work around the conflict, reordered
  pal,lib,bin and use NO_PAL_MINMAX for chakracore sources. Removed
  previous USING_PAL_MINMAX hack.

- template struct CompactCounters uses an enum class member as array size,
  which clang rejects as it must be an integral value. Changed the enum
  class to normal enum.

- Corrected some usages of #ifdefs (STACK_BACK_TRACE...)

- Moved a JavascriptArray "goto" target section out of an incorrect scope.
  The original scope has scoped variable initializations and cannot be
  skipped by goto target.

- Added 2 emmintrin.h functions into palrt.h, similar to existing changes.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -75,7 +75,7 @@ if(CLR_CMAKE_PLATFORM_UNIX)
     add_definitions(
         -DDISABLE_JIT=1
         )
-      
+
     # xplat-todo: revisit these
     # also, we should change this from add_definitions- the intent of that
     # is to just add -D flags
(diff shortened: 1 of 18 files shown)


digitalinfinity:
LGTM. Shame about the CounterFields- One option is that we could continue to use enum class, and in the case where we use CounterFields::Max, explicitly cast to an integral type. That we still get the type safety of enum class with the explicit case where we want to lose it. Thoughts?

jianchun:
> One option is that we could continue to use enum class, and in the case where we use CounterFields::Max, explicitly cast to an integral type. That we still get the type safety of enum class with the explicit case where we want to lose it. Thoughts?

@digitalinfinity You are right! Cast works. Restored to ```enum class```. Thanks!

PR #493, opened 2016-03-09
Split scope with eval in in body

aneeshdk:
Split scope with eval in in body

This changelist has,
1. Fix for issues with eval being present in a split scoped function body,
for both interpreter and JIT mode.
2. If a funciton definition in param scope refers to a formal, split scope
will be enabled by default.


aneeshdk:
@pleath @abchatra @akroshg @tcare can you please take a look?

lib/Runtime/ByteCode/ByteCodeGenerator.cpp
```diff
@@ -886,6 +886,10 @@ void ByteCodeGenerator::AssignFrameSlotsRegister()
     {
         top->frameSlotsRegister = NextVarRegister();
     }
+    if (top->GetParamScope() && !top->GetParamScope()->GetCanMergeWithBodyScope())
+    {
+        top->paramSlotsRegister = NextVarRegister();
```
akroshg: could you assert that paramSlotsRegister is noRegister here?

tcare:
What problems were you running into with eval being present in a split scoped function body?

For the changes in IRBuilder, were we just missing those instructions previously or are they IRBuilder specific?

aneeshdk:
@tcare When we have eval in function body we were wrongly marking param scope also as an object. Also when we create closures we were only looking at the body and creating a scope object. But in case of split where the param scope has function definition and no eval, but body has eval we have to create scope slots first. Then during OP_BeginBodyScope we have to create the scope object. Missing this logic was causing issues. Similar changes were required in the back end too. Also during bailout when we create a new interpreter stack frame from native frame we have to retrieve the paramClosure to use in the interpreter stack frame.

pleath:
Mmm, no, the param scope must also be an object when the function body calls eval. This is to facilitate accesses from within eval that have to traverse the param scope in order to load a property from an enclosing scope.

--Paul

aneeshdk:
Fixed this.

aneeshdk:
Fixed a comment from Paul. When emitting code to copy value from param scope to the corresponding body scope EmitPropLoad cannot be used directly as it walks the scope chain. We have to use the param scope directly.

lib/Runtime/ByteCode/ByteCodeEmitter.cpp
```diff
@@ -4971,22 +4986,32 @@ ByteCodeGenerator::GetLdSlotOp(Scope *scope, int envIndex, Js::RegSlot scopeLoca
...
-        // When param and body scopes are not merged the only place where you try to load the value from a param scoped symbol,
-        // while in body scope, is when we use its value to initialize the corresponding value in the body scope. That time we
-        // cannot use the normal LdLocalSlot as we have to use the param scope closure to load that field.
-        op = Js::OpCode::LdParamSlot;
-    }
     else if (scopeLocation != Js::Constants::NoRegister &&
              scopeLocation == funcInfo->frameSlotsRegister)
     {
-        op = Js::OpCode::LdLocalSlot;
+        if (!scope->GetCanMergeWithBodyScope() && this->currentScope == scope->GetFunc()->GetBodyScope())
```
pleath: Do we really need this method to return LdParam[Slot,ObjSlot]? Isn't that opcode only for the case where you're emitting the initialization of the body scope symbols?

lib/Runtime/ByteCode/ByteCodeGenerator.cpp
```diff
@@ -1391,13 +1398,16 @@ FuncInfo * ByteCodeGenerator::StartBindFunction(const char16 *name, uint nameLen
     if (pnode->sxFnc.CallsEval())
     {
         funcInfo->SetCallsEval(true);
+        
         bodyScope->SetIsDynamic(true);
         bodyScope->SetIsObject();
         bodyScope->SetCapturesAll(true);
         bodyScope->SetMustInstantiate(true);
+
+        paramScope->SetIsDynamic(true);
```
pleath: Is it true that the use of a default expression makes the parameter list obey strict mode rules? Does that extend to the behavior of eval? If so, then the paramScope is never dynamic, because eval can't leak declarations.

lib/Runtime/ByteCode/ByteCodeEmitter.cpp
aneeshdk: Removed it.

lib/Runtime/ByteCode/ByteCodeGenerator.cpp
aneeshdk: Fixed it.

pleath:
Code looks fine now. Do we have good test coverage for eval called from within param scope? We should check that eval loads the right values and that declarations from within eval in the param scope don't leak out.

aneeshdk:
I haven't done lot of testing with eval in the param scope as I haven't made the change in the parser to create split scope when param scope has eval. In merged scenario we already have test cases that verify leak scenario which are passing. I will be working on eval in param scope soon. I will add more test cases then.

akroshg:
Our internal tool has already producing scenario with eval on the param scope - I am sure are you running that for reliability purpose.

aneeshdk:
Yes they are running for reliability purpose. Until I set the flag for split scope in the parser those scenarios will result in syntax error without the new flag. Else they will work with merge scope.

PR #495, opened 2016-03-09
linux: initialize PAL in a way that has a chance of working everywhere

xanlpz:


jianchun:
@xanlpz Thanks!
@digitalinfinity This was the reason of segment fault on my box.

digitalinfinity:
Thanks @xanlpz :blush: sorry, not sure how that slipped through. @jianchun, since you've already tested this, can you merge this please? @xanlpz this change is fine for now but once we get ChakraCore actually building as a DLL, we'll need to switch back to using PAL_InitializeChakraCore (after a bit of cleanup) so that we can have our DLLMain get called after PAL's initialized. But for now, this change is fine.

jianchun:
Sure, I'll merge it.

PR #497, opened 2016-03-09
Debugger support for default parameters

akroshg:
When we stop at formals (params) location we need to show locals, visible
only at formals scope. For that purpose DebuggerScope is used (with a new
type DiagParamScope). That scope will contain the bytecode range and symbols. With
the help of that scope we will filter variables when we are halted at
formals location.


akroshg:
@aneeshdk @agarwal-sandeep - please have a look.

agarwal-sandeep:
Looks good to me

akroshg:
Thanks @agarwal-sandeep

PR #499, opened 2016-03-09
Wasm Indirect function table

Cellule:
Update the binary read to correctly read the indirect function table
@MikeHolman @commonlisp Can you review this please ?

---
This change is [Review on Reviewable](https://reviewable.io/reviews/microsoft/chakracore/499)


lib/WasmReader/WasmBinaryReader.cpp
```diff
@@ -770,7 +786,7 @@ WasmBinaryReader::LEB128(UINT &length, bool sgn)
         }
     }
 
-    if (b & 0x80)
+    if (b & 0x80 || m_pc > m_end)
```
MikeHolman: good to add this check, but probably better to have it in the loop

MikeHolman:
lgtm

PR #500, opened 2016-03-09
Destructuring pattern at formals with rest param.

akroshg:
We actually assign register to formals of destructuring pattern before than rest param. This makes the rest to fetch value from wrong register. Fixed that by reversing the order.


akroshg:
@tcare  - please have a look at this one.

tcare:
Looks good.

akroshg:
Thanks @tcare

PR #502, opened 2016-03-10
Cap new page count during background sweeping

leirocks:
When large amount of short living objects and small amount of long living objects interlaced on the same pages causes very bad fragmentation.
With background sweeping the problem becomes even worse because while background sweeping, the short living objects swept space can't be reused before the whole background sweeping completed. So the interlaced allocation keeps happening on new pages.
So far there's no easy way to totally solve the issue. This change mitigate the issue by setting a limit of new pages allowed during background sweeping, and wait for the background sweeping complete when limit is hit.

simple repro for the fragmentation:
```
var glo = [];
for (var k = 0; k < 258; k++) {
    for (var j = 0; j < 258; j++) {
        var x;
        for (var i = 0; i <1000 ; i++) {
             x = {}; // short living objects
        }
        glo.push({});// long living objects sharing same bucket/pages
    }
}
glo={};
```

### DIFF
--- a/lib/Common/Memory/HeapBucket.cpp
+++ b/lib/Common/Memory/HeapBucket.cpp
@@ -503,6 +503,17 @@ HeapBucketT<TBlockType>::SnailAlloc(Recycler * recycler, TBlockAllocatorType * a
     AllocationVerboseTrace(recycler->GetRecyclerFlagsTable(), _u("TryAlloc failed, forced collection on allocation [Collected: %d]\n"), collected);
     if (!collected)
     {
+        // wait for background sweeping finish if there are too many pages allocated during background sweeping
+        if (recycler->IsConcurrentSweepExecutingState() && this->heapInfo->uncollectedNewPageCount > (uint)CONFIG_FLAG(NewPagesCapDuringBGSweeping))
+        {
+            recycler->FinishConcurrent<ForceFinishCollection>();
+            memBlock = this->TryAlloc(recycler, allocator, sizeCat, attributes);
+            if (memBlock != nullptr)
+            {
+                return memBlock;
+            }
+        }
+
         // We didn't collect, try to add a new heap block
         memBlock = TryAllocFromNewHeapBlock(recycler, allocator, sizeCat, attributes);
         if (memBlock != nullptr)
--- a/lib/Common/ConfigFlagsList.h
+++ b/lib/Common/ConfigFlagsList.h
@@ -434,6 +434,7 @@ PHASE(All)
 #endif
 
 #define DEFAULT_CONFIG_LowMemoryCap         (0xB900000) // 185 MB - based on memory cap for process on low-capacity device
+#define DEFAULT_CONFIG_NewPagesCapDuringBGSweeping    (15000)
 
 #define DEFAULT_CONFIG_MaxCodeFill          (500)
 #define DEFAULT_CONFIG_MaxLoopsPerFunction  (10)
@@ -1168,6 +1169,7 @@ FLAGNR(Boolean, RecyclerProtectPagesOnRescan, "Temporarily switch all pages to r
 FLAGNR(Boolean, RecyclerVerifyMark    , "verify concurrent gc", false)
 #endif
 FLAGR (Number,  LowMemoryCap          , "Memory cap indicating a low-memory process", DEFAULT_CONFIG_LowMemoryCap)
+FLAGNR(Number,  NewPagesCapDuringBGSweeping, "New pages count allowed to be allocated during background sweeping", DEFAULT_CONFIG_NewPagesCapDuringBGSweeping)
 #ifdef RUNTIME_DATA_COLLECTION
 FLAGNR(String,  RuntimeDataOutputFile, "Filename to write the dynamic profile info", nullptr)
 #endif


leirocks:

lib/Common/Memory/HeapBucket.cpp
```diff
@@ -503,10 +504,42 @@ HeapBucketT<TBlockType>::SnailAlloc(Recycler * recycler, TBlockAllocatorType * a
...
+                memBlock = this->TryAlloc(recycler, allocator, sizeCat, attributes);
+                if (memBlock != nullptr)
+                {
+                    newAssignedPageCountDuringBGSweeping = 0;
+                    return memBlock;
+                }
+            }
+
+            recycler->CollectNow<CollectNowForceInThread>();
+            memBlock = this->TryAlloc(recycler, allocator, sizeCat, attributes);
```
curtisman: Why do we want to force a in thread GC here?

lib/Common/Memory/HeapBucket.cpp
```diff
@@ -503,10 +504,42 @@ HeapBucketT<TBlockType>::SnailAlloc(Recycler * recycler, TBlockAllocatorType * a
     AllocationVerboseTrace(recycler->GetRecyclerFlagsTable(), _u("TryAlloc failed, forced collection on allocation [Collected: %d]\n"), collected);
     if (!collected)
     {
+        // wait for background sweeping finish if there are too many pages allocated during background sweeping
+        BOOL isBackgroundSweeping = recycler->IsConcurrentSweepExecutingState();
+        if (newAssignedPageCountDuringBGSweeping > (uint)CONFIG_FLAG(NewPagesCapDuringBGSweeping))
```
curtisman: Only need to check if we are background sweeping?

lib/Common/Memory/HeapBucket.cpp
```diff
@@ -503,10 +504,42 @@ HeapBucketT<TBlockType>::SnailAlloc(Recycler * recycler, TBlockAllocatorType * a
...
+            }
+        }
+
         // We didn't collect, try to add a new heap block
         memBlock = TryAllocFromNewHeapBlock(recycler, allocator, sizeCat, attributes);
         if (memBlock != nullptr)
         {
+            if (isBackgroundSweeping)
+            {
+                newAssignedPageCountDuringBGSweeping += TBlockType::HeapBlockAttributes::PageCount;
```
curtisman: We can just always increment this count (inside TryAllocFromNewHeapBlock)
And 0 it out before we Sweep.

lib/Common/Memory/HeapBucket.cpp
leirocks: this can be hit at very low chance. I think I can remove it

lib/Common/Memory/HeapBucket.cpp
leirocks: I initially uses IsConcurrentExecutingState(), I remember it resulting worse perf in early-boyer (the code path will be hit more often). Let me try again since I made a few other adjustment to avoid the perf degradation.

lib/Common/Memory/HeapBucket.cpp
leirocks: good idea, will do

lib/Common/Memory/HeapBucket.h
```diff
@@ -284,6 +289,7 @@ HeapBucketT<TBlockType>::SweepBucket(RecyclerSweep& recyclerSweep, Fn sweepFn)
         // Every thing is swept immediately in non partial collect, so we can allocate
         // from the heap block list now
         StartAllocationAfterSweep();
+        this->newAssignedPageCountDuringBGSweeping = 0;
```
curtisman: Do we need to 0 it out here?

lib/Common/Memory/HeapBucket.h
```diff
@@ -267,6 +268,10 @@ template <bool pageheap, typename Fn>
 void
 HeapBucketT<TBlockType>::SweepBucket(RecyclerSweep& recyclerSweep, Fn sweepFn)
 {
+    if (recyclerSweep.IsBackground())
```
curtisman: I think you need to zero this out in SetupBackgroundSweep to catch all the page during background sweep.
Otherwise, you won't 0 it out until the background thread gets to it?

lib/Common/Memory/HeapBucket.cpp
```diff
@@ -379,6 +380,10 @@ HeapBucketT<TBlockType>::TryAllocFromNewHeapBlock(Recycler * recycler, TBlockAll
     // We just created a block we can allocate on
     char * memBlock = allocator->SlowAlloc<false /* disallow fault injection */>(recycler, sizeCat, attributes);
     Assert(memBlock != nullptr || IS_FAULTINJECT_NO_THROW_ON);
+    if (recycler->IsConcurrentSweepExecutingState())
```
leirocks: That will be counting all page count allocated. if there's intensive allocation before starting the background sweeping, and then right after started background sweeping it can reach the limit and result in wait. it will increase the chance to hit the waiting code path but how much chance it increases is uncertain, so result in high variation in perf. and need to set a bigger cap but not easy to find a good number for the cap

lib/Common/Memory/HeapBucket.h
leirocks: changed to just reuse the existing uncollectedNewPageCount, I tried with this the perf and fragmentation seems similar to before the change

lib/Common/Memory/HeapBucket.h
leirocks: changed to use uncollectedNewPageCount so should not zero out here any more. I thought about only count the new pages during sweeping.

curtisman:
LGTM

PR #503, opened 2016-03-10
Add JIT support to module import and exports

boingoing:
Add JIT support to module import and exports

Allow the export-side name to be a reserved word. This is by design and required for syntax such as 'export { foo as default };' Also allow import statements to use reserved words as the IdentifierName used to find the export needed for syntax like 'import { default as foo } from "./module.js"'

Fix an issue where using a value imported from a different module inside a nested function could cause AV

Fix a bug where indirect export resolution only performed resolution against the import identifier instead of the export name

### DIFF
--- a/lib/Backend/IRBuilder.cpp
+++ b/lib/Backend/IRBuilder.cpp
@@ -3662,7 +3662,11 @@ IRBuilder::BuildElementSlotI2(Js::OpCode newOpcode, uint32 offset, Js::RegSlot r
         case Js::OpCode::LdModuleSlot:
         case Js::OpCode::StModuleSlot:
         {
-            Js::Var* moduleExportVarArrayAddr = Js::JavascriptOperators::OP_GetModuleExportSlotArrayAddress(slotId1, m_func->GetScriptContext());
+            if ((uint)slotId1 >= m_func->GetJnFunction()->GetInnerScopeCount())
+            {
+                Js::Throw::FatalInternalError();
+            }
+            Js::Var* moduleExportVarArrayAddr = Js::JavascriptOperators::OP_GetModuleExportSlotArrayAddress(slotId1, slotId2, m_func->GetScriptContext());
             IR::AddrOpnd* addrOpnd = IR::AddrOpnd::New(moduleExportVarArrayAddr, IR::AddrOpndKindConstant, m_func, true);
             regOpnd = IR::RegOpnd::New(TyVar, m_func);
             instr = IR::Instr::New(Js::OpCode::Ld_A, regOpnd, addrOpnd, m_func);
--- a/lib/Backend/IRBuilder.cpp
+++ b/lib/Backend/IRBuilder.cpp
@@ -3662,10 +3662,6 @@ IRBuilder::BuildElementSlotI2(Js::OpCode newOpcode, uint32 offset, Js::RegSlot r
         case Js::OpCode::LdModuleSlot:
         case Js::OpCode::StModuleSlot:
         {
-            if ((uint)slotId1 >= m_func->GetJnFunction()->GetInnerScopeCount())
-            {
-                Js::Throw::FatalInternalError();
-            }
             Js::Var* moduleExportVarArrayAddr = Js::JavascriptOperators::OP_GetModuleExportSlotArrayAddress(slotId1, slotId2, m_func->GetScriptContext());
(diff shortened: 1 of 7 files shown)


boingoing:
@pleath @Yongqu @abchatra Can one of you take a glance over these changes, please? Thanks!

lib/Backend/IRBuilder.cpp
```diff
@@ -3659,6 +3659,36 @@ IRBuilder::BuildElementSlotI2(Js::OpCode newOpcode, uint32 offset, Js::RegSlot r
...
+        case Js::OpCode::LdModuleSlot:
+        case Js::OpCode::StModuleSlot:
+        {
+            Js::Var* moduleExportVarArrayAddr = Js::JavascriptOperators::OP_GetModuleExportSlotArrayAddress(slotId1, m_func->GetScriptContext());
+            IR::AddrOpnd* addrOpnd = IR::AddrOpnd::New(moduleExportVarArrayAddr, IR::AddrOpndKindConstant, m_func, true);
+            regOpnd = IR::RegOpnd::New(TyVar, m_func);
+            instr = IR::Instr::New(Js::OpCode::Ld_A, regOpnd, addrOpnd, m_func);
+            this->AddInstr(instr, offset);
+
+            fieldSym = PropertySym::FindOrCreate(regOpnd->m_sym->m_id, slotId2, (uint32)-1, (uint)-1, PropertyKindSlots, m_func);
```
pleath: We need to verify slotId2 here (i.e., verify that it's a valid slot ID for the given module) and call FatalInternalError if it isn't.

lib/Runtime/Language/InterpreterStackFrame.cpp
```diff
@@ -7935,39 +7935,12 @@ const byte * InterpreterStackFrame::OP_ProfiledLoopBodyStart(const byte * ip)
...
-        if (moduleRecord->GetLocalExportCount() <= (uint)playout->SlotIndex2)
-        {
-            Js::Throw::FatalInternalError();
-        }
-
-        Var value = moduleRecordSlots[playout->SlotIndex2];
-        Assert(value != nullptr);
-
-        return value;
+        return JavascriptOperators::OP_LdModuleSlot(playout->SlotIndex1, playout->SlotIndex2, scriptContext);
```
pleath: Does this path verify playout->SlotIndex2?

lib/Runtime/Language/InterpreterStackFrame.cpp
```diff
@@ -7935,39 +7935,12 @@ const byte * InterpreterStackFrame::OP_ProfiledLoopBodyStart(const byte * ip)
...
-        Var* moduleRecordSlots = moduleRecord->GetLocalExportSlots();
-        Assert(moduleRecordSlots != nullptr);
-
-        if (moduleRecord->GetLocalExportCount() <= (uint)slotIndex2)
-        {
-            Js::Throw::FatalInternalError();
-        }
-
-        moduleRecordSlots[slotIndex2] = value;
+        JavascriptOperators::OP_StModuleSlot(slotIndex1, slotIndex2, value, scriptContext);
```
pleath: Does this path verify slotIndex2?

lib/Backend/IRBuilder.cpp
boingoing: Added the check for valid slot id in OP_GetModuleExportSlotArrayAddress itself.

lib/Runtime/Language/InterpreterStackFrame.cpp
boingoing: It does check that both index values are correct in this path:

OP_LdModuleSlot->OP_GetModuleExportSlotAddress->OP_GetModuleExportSlotArrayAddress

lib/Runtime/Language/InterpreterStackFrame.cpp
boingoing: Yes, in the same way that OP_LdModuleSlot does - via checks in JavascriptOperators::OP_GetModuleExportSlotArrayAddress

lib/Backend/IRBuilder.cpp
pleath: Cool. Thanks.

pleath:
LGTM. Thanks.

PR #508, opened 2016-03-10
Makes `await` a FutureReservedWord in Modules only

### LINKED ISSUE #507: `await` incorrectly disallowed as identifier in strict mode
```javascript
"use strict";
var await = 0;
```

throws a SyntaxError.

`await` is being treated as a FutureReservedWord in all scripts, like `yield`, meaning it is disallowed in strict mode.  However, the ES6 spec states that `await` is only treated this way for module scripts.

ianwjhalliday:
Fixes #507 

The ES6 spec states that `await` is a _FutureReservedWord_ but then
refines this and says only when the goal symbol of the syntactic grammar
is _Module_.  Meaning that `await` cannot be used as an identifier in
ES6 modules.  The word is reserved to allow top-level await in modules
as a potential future feature addition to the language.

### DIFF
--- a/test/es6/rlexe.xml
+++ b/test/es6/rlexe.xml
@@ -1153,4 +1153,11 @@
         <baseline>HTMLComments.baseline</baseline>
     </default>
 </test>
+<test>
+  <default>
+    <files>await-futreserved-only-in-modules.js</files>
+    <compile-flags>-ES6Module</compile-flags>
+    <tags>exclude_dynapogo</tags>
+  </default>
+</test>
 </regress-exe>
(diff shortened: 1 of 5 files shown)


ianwjhalliday:
@boingoing could you review this please

test/es6/rlexe.xml
```diff
@@ -1153,4 +1153,11 @@
         <baseline>HTMLComments.baseline</baseline>
     </default>
 </test>
+<test>
+  <default>
+    <files>await-futreserved-only-in-modules.js</files>
+    <compile-flags>-ES6Module</compile-flags>
+    <tags>exclude_dynapogo</tags>
```
boingoing: Module JIT support was just added, it should work so you might want to try it

ianwjhalliday: Oh I made this `exclude_dynapogo` because it is a syntax test.  No need to hit the backend.

boingoing:
LGTM

test/es6/rlexe.xml
boingoing: That's fair

ianwjhalliday:
Thanks @boingoing

PR #509, opened 2016-03-10
Fix bad byte code for nested function in event handler

pleath:
Fixes an issue exposed on crawler. Don't mark a deferred nested function as an event handler. It doesn't require different code gen to support the semantics (the frame display given to it by the parent funciton does that), and we don't want to try to build an event-handler-style frame display for it.


pleath:
No, this is wrong. I'll fix the issue differently.

PR #511, opened 2016-03-10
Fix #263: `let let` and `const let` declarations are supposed to be a syntax error

### LINKED ISSUE #263: ES6 `let let` and `const let` declarations are supposed to be a syntax error
_Originally this issue mentioned only the syntax error for `for (let let ...` but the issue actually applies to all `let` and `const` declarations containing a variable named `let`.  See the general static semantics rule here: https://tc39.github.io/ecma262/#sec-let-and-const-declarations-static-semantics-early-errors_

---

```javascript
for (let let = 0; false; ) { } // should be early syntax error on identifier named let
for (let let in { }) { } // ditto
for (let let of { }) { } // ditto
// and likewise when destructuring patterns are used
```

I think this was at one time decidedly not going to be an error because we thought about this case and intentionally implemented it so that it would parse.

In any case it no longer matches the spec.  Spec says it should be an error.

We need to make sure we disallow the name 'let' in an arbitrary destructuring too.

https://tc39.github.io/ecma262/#sec-let-and-const-declarations-static-semantics-early-errors
https://tc39.github.io/ecma262/#sec-for-in-and-for-of-statements-static-semantics-early-errors

Ported issue from internal VSO bug [5055638](https://microsoft.visualstudio.com/DefaultCollection/OS/_workitems?_a=edit&id=5055638)

AlexElting:
I was finally able to get a chance to finish this. I added a check to the parser that makes sure an identifier used in a `let` or `const` declaration is not `let` itself (see the links to the spec in issue #263 ). The included tests check a few of the different ways that an identifier may be used in this manner. Thanks!


lib/Parser/perrors.h
```diff
@@ -92,9 +92,12 @@ LSC_ERROR_MSG(1082, ERRNonSimpleParamListInStrictMode, "Cannot apply strict mode
...
 LSC_ERROR_MSG(1084, ERRGetterMustHaveNoParameters, "Getter functions must have no parameters")
+
 LSC_ERROR_MSG(1085, ERRFuncRefFormalNotSupportedInParamScope, "Formals cannot contain functions definitions that reference them")
 LSC_ERROR_MSG(1086, ERREvalNotSupportedInParamScope, "'eval' is not allowed in the default initializer")
 LSC_ERROR_MSG(1087, ERRInvalidUseofExponentiationOperator, "Invalid unary operator on the left hand side of exponentiation (**) operator")
 
 LSC_ERROR_MSG(1088, ERRInvalidModuleImportOrExport, "Module import or export statement unexpected here")
 LSC_ERROR_MSG(1089, ERRInvalidExportName, "Unable to resolve module export name")
+
+LSC_ERROR_MSG(1090, ERRLetIDInLexicalDecl, "'let' is not an allowed identifier in lexical declarations")
```
ianwjhalliday: nit: missing a new line at the end of this file and `letlet.js` below.

test/LetConst/letlet.baseline
```diff
@@ -0,0 +1,13 @@
...
+PASSED
+*** Running test #3 (2): 'let' should not be an allowed name in destructuring let declarations
+PASSED
+*** Running test #4 (3): 'let' should not be an allowed name in destructuring const declarations
+PASSED
+*** Running test #5 (4): 'let' should not be an allowed name in 'for(let .. in ..)' declarations
+PASSED
+*** Running test #6 (5): 'let' is OK if used in var declarations
+PASSED
+Summary of tests: total executed: 6; passed: 6; failed: 0
```
ianwjhalliday: Instead of using a baseline you can use a summary argument to `ch.exe` to choose to print a terse "pass/fail" message that `rl.exe` will recognize in the absence of a baseline.

Take a look at another test that uses `UnitTestFramework.js` but doesn't have a baseline specified in `rlexe.xml`.  Not sure if there are examples under `test/LetConst`, but there are definitely examples under `test/es6`.

lib/Parser/Parse.cpp
```diff
@@ -8655,6 +8655,12 @@ ParseNodePtr Parser::ParseVariableDeclaration(
             nameHintLength = pid->Cch();
             nameHintOffset = 0;
 
+            if ((declarationType == tkCONST || declarationType == tkLET) &&
+                wcsncmp(pNameHint, L"let", 4) == 0)
```
ianwjhalliday: IdentPtrs (aka PIDs) are interned strings which means you need only do a pointer comparison with the PID you already know to be `let`.  No need to do a string comparison.

In this case you want to do `pid == wellKnownPropertyPids.let`.

`wellKnownPropertyPids` is a member of `Parser`.

ianwjhalliday:
There's a few minor things to touch up but otherwise looks great!  Thanks @AlexElting.

I was just reviewing this issue and wondering if it was coming soon when I noticed you had pushed a commit for it to your fork.  Weird.

AlexElting:
Thanks, @ianwjhalliday!! I suspected there was a better way to do that check.

ianwjhalliday:
This is perfect.  Will merge in a moment.

ianwjhalliday:
Boom, it's in.  Thanks again @AlexElting!

AlexElting:
Sweet, thank you!

PR #513, opened 2016-03-10
finish initial stardardization for reading wasm function bodies

MikeHolman:
This change should finalize our function body decoding for the current milestone. Changes here include: integration with the import table change from earlier, allows blocks/loops to yield values, loosens restrictions on if/else (instead of failing validation if lhs and rhs don't match, we only yield value if the types match), adds in a few math operations that were previously overlooked, and a couple bug fixes.

This change is [Review on Reviewable](https://reviewable.io/reviews/microsoft/chakracore/513)


Cellule:
LGTM

---

Reviewed 11 of 11 files at r1.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/microsoft/chakracore/513#-:-KCXHB0ii8jV7E-32e0o:676834233)*

PR #514, opened 2016-03-10
Break up TypePath to avoid false reference.

curtisman:
TypePath contains a TinyDictionary which use bytes to store the chain in the dictionary.  These bytes may combine to look like a pointer, more likely in 32-bit then 64-bits, causing memory leaks.  Move all the numerical fields from TypePath to a substructure, and allocate that as leaf.

Now that it is allocated separately from TypePath, we can also make TinyDictionary's next array dynamic size depending on the entries necessary to save some memory,

### DIFF
--- a/lib/Runtime/Types/TypePath.cpp
+++ b/lib/Runtime/Types/TypePath.cpp
@@ -8,18 +8,28 @@ namespace Js {
 
     TypePath* TypePath::New(Recycler* recycler, uint size)
     {
+        Assert(size <= MaxPathTypeHandlerLength);
         size = max(size, InitialTypePathSize);
-        size = PowerOf2Policy::GetSize(size);
+        
 
         if (PHASE_OFF1(Js::TypePathDynamicSizePhase))
         {
             size = MaxPathTypeHandlerLength;
         }
+        else
+        {
+            size = PowerOf2Policy::GetSize(size - TYPE_PATH_ALLOC_GRANULARITY_GAP);
+            if (size < MaxPathTypeHandlerLength)
+            {
+                size += TYPE_PATH_ALLOC_GRANULARITY_GAP;
+            }
+        }
 
         Assert(size <= MaxPathTypeHandlerLength);
 
         TypePath * newTypePath = RecyclerNewPlusZ(recycler, sizeof(PropertyRecord *) * size, TypePath);
-        newTypePath->pathSize = (uint16)size;
+        // Allocate enough space for the "next" for the TinyDictionary;
+        newTypePath->data = RecyclerNewPlusLeafZ(recycler, size, TypePath::Data, (uint8)size);
 
         return newTypePath;
     }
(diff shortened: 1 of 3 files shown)


curtisman:
@pleath please review.

lib/Runtime/Types/TypePath.cpp
```diff
@@ -8,18 +8,28 @@ namespace Js {
...
+                size += TYPE_PATH_ALLOC_GRANULARITY_GAP;
+            }
+        }
 
         Assert(size <= MaxPathTypeHandlerLength);
 
         TypePath * newTypePath = RecyclerNewPlusZ(recycler, sizeof(PropertyRecord *) * size, TypePath);
-        newTypePath->pathSize = (uint16)size;
+        // Allocate enough space for the "next" for the TinyDictionary;
+        newTypePath->data = RecyclerNewPlusLeafZ(recycler, size, TypePath::Data, (uint8)size);
```
leirocks: allocate the data first and then set the pathSize, otherwise the pathSize and the data pointer is not consistent when OOM

lib/Runtime/Types/TypePath.cpp
```diff
@@ -8,18 +8,28 @@ namespace Js {
...
-        size = PowerOf2Policy::GetSize(size);
+        
 
         if (PHASE_OFF1(Js::TypePathDynamicSizePhase))
         {
             size = MaxPathTypeHandlerLength;
         }
+        else
+        {
+            size = PowerOf2Policy::GetSize(size - TYPE_PATH_ALLOC_GRANULARITY_GAP);
```
pleath: Just for my info, what is this path doing?

lib/Runtime/Types/TypePath.cpp
leirocks: Ah sorry mistakenly read the change (thought the removed code is newly added code)

lib/Runtime/Types/TypePath.cpp
pleath: Got it. Thanks.

pleath:
LGTM. Thanks.

lib/Runtime/Types/TypePath.h
```diff
@@ -61,45 +52,71 @@ namespace Js
...
+#if defined(_M_X64_OR_ARM64)
+#define TYPE_PATH_ALLOC_GRANULARITY_GAP 0
+#else
+#define TYPE_PATH_ALLOC_GRANULARITY_GAP 2
+#endif
+#else
+#if defined(_M_X64_OR_ARM64)
+#define TYPE_PATH_ALLOC_GRANULARITY_GAP 1
+#else
+#define TYPE_PATH_ALLOC_GRANULARITY_GAP 3
```
Yongqu: seem risky with the constant here. Is there someway to calculate size of trailing gap and c_assert it?

curtisman: Unfortunately no. The constant is needed before the class TypePath is fully defined.
I already added compiler assert at the end to validate this constant.

PR #515, opened 2016-03-10
Restore context after beforeCollectCallback

Yongqu:
restore threadcontext & jsrt context after return from per-object pre collection callback.
Remove the requirement to SetCurrentContext for JsAddRef and JsRelease inside the callback.

Fixed JsrtContext leak by not pinning global object in JSRT scenario. GlobalObject is being
kept alive by the JsrtContext itself. We don't need to pin it in the ScriptContext anymore.


Yongqu:
@jianchun can you take a look?

lib/Common/Memory/Recycler.h
```diff
@@ -1966,9 +1972,12 @@ class Recycler
...
         ObjectBeforeCollectCallback callback;
         void* callbackState;
+        void* threadContext;
+        ObjectBeforeCollectCallbackWrapper callbackWrapper;
 
         ObjectBeforeCollectCallbackData() {}
-        ObjectBeforeCollectCallbackData(ObjectBeforeCollectCallback callback, void* callbackState) : callback(callback), callbackState(callbackState) {}
+        ObjectBeforeCollectCallbackData(ObjectBeforeCollectCallbackWrapper callbackWrapper, ObjectBeforeCollectCallback callback, void* callbackState, void* threadContext) : 
+            callbackWrapper(callbackWrapper), callback(callback), callbackState(callbackState), threadContext(threadContext) {}
     };
```
jianchun: nit pick: The order of these variables are a bit inconsistent and hard to read among function SetObjectBeforeCollectCallback, struct ObjectBeforeCollectCallbackData, and constructor ObjectBeforeCollectCallbackData.

Thanks for fixing this issue thoroughly! :+1:

PR #516, opened 2016-03-10
Fix for OOM when initially allocating counters structure on FunctionBody

leirocks:
When OOM happens initially allocating counters, the functionBody later cleaning up will reset the counters, which enters the setter path in the counters structure which is in wrong state (fieldSize is 1 but fields pointer is null)

### DIFF
--- a/lib/Runtime/Base/CompactCounters.h
+++ b/lib/Runtime/Base/CompactCounters.h
@@ -11,7 +11,7 @@ namespace Js
 {
     template<class T, typename CountT = T::CounterFields>
     struct CompactCounters
-    {    
+    {
         struct Fields {
             union {
                 uint8 u8Fields[CountT::Max];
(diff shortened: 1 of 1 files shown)


lib/Runtime/Base/CompactCounters.h
```diff
@@ -129,8 +135,13 @@ namespace Js
...
             }
 
-            Assert(fieldSize == 4);
-            return this->fields->u32Fields[type] = val;
+            if (fieldSize == 4)
+            {
+                return this->fields->u32Fields[type] = val;
+            }
+
+            Assert(fieldSize == 0 && this->isCleaningUp && this->fields == nullptr); // OOM when allocating the counters structure
```
leirocks: yes, I thought about that. but the register default value is unsigned max (-1). let me check if we reset those counters when cleaning up. and assert val==0||val==-1 if needed

lib/Runtime/Base/CompactCounters.h
```diff
@@ -253,7 +269,7 @@ namespace Js
             }
             else
             {
-                Assert(false);
+                Assert(this->fieldSize==0);
```
leirocks: ah, yes

curtisman:
LGTM

PR #519, opened 2016-03-10
delete the right indexed property

jianchun:
In one delete property code path "propertyRecord->IsNumeric()" (bool) is
passed for "uint32 index" parameter. Don't know why the compiler doesn't
complain about it. So if it is run we always delete property "1".

Seems the bug can only be hit through JSRT and also requires the object to
use SimpleDictionaryTypeHandler.


ianwjhalliday:
Oof that's nasty.  LGTM

curtisman:
LGTM.

jianchun:
Thanks @ianwjhalliday @curtisman

PR #520, opened 2016-03-10
WasmCompilationException cleanup for ByteCodeWriter

commonlisp:
When WasmCompilationException is thrown, LoadWasmScript()
can't clean up the bytecodeGen properly without a Reset().

### DIFF
--- a/lib/WasmReader/WasmByteCodeGenerator.cpp
+++ b/lib/WasmReader/WasmByteCodeGenerator.cpp
@@ -168,10 +168,19 @@ WasmBytecodeGenerator::GenerateFunction()
 
     WasmOp op;
     EmitInfo exprInfo;
-    while ((op = m_reader->ReadExpr()) != wnLIMIT)
+
+    try
     {
-        exprInfo = EmitExpr(op);
-        ReleaseLocation(&exprInfo);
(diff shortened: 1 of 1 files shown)


lib/WasmReader/WasmByteCodeGenerator.cpp
```diff
@@ -168,10 +168,19 @@ WasmBytecodeGenerator::GenerateFunction()
...
+
+    try
+    {
+        while ((op = m_reader->ReadExpr()) != wnLIMIT)
+        {
+            exprInfo = EmitExpr(op);
+            ReleaseLocation(&exprInfo);
+        }
+    }
+    catch (WasmCompilationException& e)
```
Cellule: I think we should catch all exception types, what do you think ?

commonlisp: Hmm, I try to keep exception handlers specific as possible to avoid unexpected behavior. I had considered making it a catch all. If the exception caught were from the ByteCodeWriter itself, would it be safe?

commonlisp: Upon second look, it ought to be ok. I'll make that change. Thanks.

commonlisp:
Ok, closed by #565.

PR #522, opened 2016-03-10
Line and column differences.

akroshg:
If the object or array literals are at the top level statement we didn't enclosed them in Start/End Statement. Fixed that.
Also for 'extends' we didn't enclose the expression in Start/End Statement so if that reproduce an error that error was reported on wrong place. Fixed that.

### DIFF
--- a/lib/Runtime/ByteCode/ByteCodeEmitter.cpp
+++ b/lib/Runtime/ByteCode/ByteCodeEmitter.cpp
@@ -21,6 +21,18 @@ bool CallTargetIsArray(ParseNode *pnode)
     return pnode->nop == knopName && pnode->sxPid.PropertyIdFromNameNode() == Js::PropertyIds::Array;
 }
 
+#define STARTSTATEMENET_IFTOPLEVEL(isTopLevel, pnode) \
+if ((isTopLevel)) \
+{ \
+    byteCodeGenerator->StartStatement(pnode); \
+}
+
+#define ENDSTATEMENET_IFTOPLEVEL(isTopLevel, pnode) \
+if ((isTopLevel)) \
+{ \
+    byteCodeGenerator->EndStatement(pnode); \
+}
+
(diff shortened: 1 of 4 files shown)


akroshg:
@agarwal-sandeep - please have a look at this.
I didn't add any unitttest as we don't add similar unittest in the past.

akroshg:
alright - will do.

agarwal-sandeep:
This seems to be a bigger issue than just object, array and extends. The issue also repros with
```javascript
try {
    function f1() {
        var k = 10;
        (typeof a.b); // Error thrown here.
    }
    f1();
} catch (e) {
    print(e.stack);
}
```

lib/Runtime/ByteCode/ByteCodeEmitter.cpp
```diff
@@ -21,6 +21,18 @@ bool CallTargetIsArray(ParseNode *pnode)
     return pnode->nop == knopName && pnode->sxPid.PropertyIdFromNameNode() == Js::PropertyIds::Array;
 }
 
+#define STARTSTATEMENET_IFTOPLEVEL(isTopLevel, pnode) \
+if ((isTopLevel)) \
+{ \
+    byteCodeGenerator->StartStatement(pnode); \
```
agarwal-sandeep: There are other usage of this statement in this file, change them to use macro instead?

akroshg: I wouldn't like to touch all of them. At this point it doesn't make sense go all for it.

agarwal-sandeep:
Rest looks good to me. Thanks for fixing all known cases.

lib/Runtime/ByteCode/ByteCodeEmitter.cpp
agarwal-sandeep: If it is just replacing with macro for code clarity I think it will be a good thing to do.

PR #523, opened 2016-03-10
Fix bad byte code for deferred nested event handlers

pleath:
Don't mark a nested function as a top-level event handler, since this can cause us to emit code to build a COM-hierarchy frame display from the this pointer, which should only be done for the function that is called by the host.


ianwjhalliday:
LGTM

PR #524, opened 2016-03-10
remove duplicated reporting of the freed pages

leirocks:
we already reported the freed pages when allocation failed


lib/Common/Memory/PageAllocator.cpp
```diff
@@ -144,7 +144,7 @@ SegmentBase<T>::Initialize(DWORD allocFlags, bool excludeGuardPages)
 
     if (this->address == nullptr)
     {
-        this->allocator->ReportFailure(totalPages * AutoSystemInfo::PageSize);
+        this->allocator->ReportFailure(0);
```
curtisman: This is miss reporting the size in the case where the first Alloc failed.

curtisman:
LGTM

Issue #510, opened 2016-03-10
yield* does not forward .next() argument to iterable's .next() call


ianwjhalliday:
```javascript
// repro, expected to print "X" but prints "undefined"

function* foo() {
    var x = yield 1;
    print(x);
}
function* bar() {
    yield* foo();
}
var it = bar();
it.next();
it.next("X");
```

From VSO bug [6790309](https://microsoft.visualstudio.com/web/wi.aspx?pcguid=cb55739e-4afe-46a3-970f-1b49d8ee7564&id=6790309)

This bug repros for any iterable passed to yield*.

nicolo-ribaudo:
Can I work on this issue?

ianwjhalliday:
Sure!  This looks like a bytecode issue so I suggest running the repro with `ch.exe -dump:bytecode` and taking a moment to take all the bytecode in.  `EmitYieldStar(...)` in `ByteCodeEmitter.cpp` is where you'll find the code that emits the bytecode for `yield*`.

I didn't look too closely at it but it seemed it is overwriting the next argument at the top of the loop with undefined always and/or it is missing the logic to get the argument value and load it into the register used for the inner next() call.

I would suggest fixing this issue standalone, and then considering #137 and #139 separately afterwards, if you're interested.

ianwjhalliday:
You might while looking at this also see a solution to #265.  If you do please feel free to tackle that one as well.

nicolo-ribaudo:
I just opened the pr - #525

> it seemed it is overwriting the next argument at the top of the loop with undefined always

The problem was exactly this :+1: 

> You might while looking at this also see a solution to #265. If you do please feel free to tackle that one as well.

I'll comment on the issue in few minutes if I can fix it.

ianwjhalliday:
Thanks for knocking this one off so quickly, @nicolo-ribaudo

PR #525, opened 2016-03-10
Make yield* forward .next() argument to iterable's .next() call

nicolo-ribaudo:
The argument of the first `.next()` was mistakenly initialized inside the `yield*`'s loop, so the correct argument was always overwritten. Move the initialization to `undefined` outside of the loop to fix gh-510


ianwjhalliday:
This is perfect.  I'll get this merged shortly.

PR #527, opened 2016-03-11
CheckFixedFld/CheckObjType fix

pleath:
Even in cases where there's another type check upstream (so the operand is marked type-available) we may do a monomorphic type check on the spot (e.g., fixed field check). So instead of checking type-available to see whether we need to push a monomorphic guard type up past a given instruction, look to see whether the instruction has a type check bailout.


pleath:
@rajatd @LouisLaf

LouisLaf:
LGTM

PR #533, opened 2016-03-11
FunctionBody: fix compiler error when doing implicit cast

xanlpz:
Clang is not happy about implicit casts to void*:

FunctionBody.h:1482:33: error: assigning to 'void *' from incompatible type 'Js::JavascriptMethod' (aka 'void *(*)(Js::RecyclableObject *, Js::CallInfo, ...)')
            entryPoint->address = address;
                                ^ ~~~~~~~

We only want the address of the function here so this is safe, add a
C-style cast.


digitalinfinity:
@xanlpz not sure what the build failures are here, @dilijev can you see if there's an issue here please?

dilijev:
Now that master is merged into Linux, re-running the tests should work.

dilijev:
@dotnet-bot test this please

PR #537, opened 2016-03-11
CodePageAllocator::AllocPages should pass thru the page count instead of have fix 1 page.o

curtisman:
During refactoring the page count of 1 was copy and paste there.  Since the caller only call with value 1 currently, it doesn't cause any functional issue.  But the request size should be honor when the API is used with different page count in the future.


curtisman:
@digitalinfinity please code review

digitalinfinity:
LGTM

Issue #435, opened 2016-03-03
Throwing an error causes ToString call on a property, and that property is a Symbol.


akroshg:
eg. 
```javascript
var a = undefined
a[Object(Symbol())] = 10;
```
as `a `is undefined we throw a type error. While throwing, we are trying to get string from that property to populate error information, which makes `ToString `call to that property (in this scenario that property is `Symbol`).
Is this implicit `ToString `call on `Symbol `correct and as per spec?
Look for 
`JavascriptError::ThrowTypeError(scriptContext, JSERR_Property_CannotSet_NullOrUndefined, GetPropertyDisplayNameForError(index, scriptContext));
`in the code base.

akroshg:
This issue raised from #422.

ljharb:
The spec doesn't talk about how error message text is generated - so an observable `toString` on that property is probably not something that's expected.

boingoing:
Fixed by #540

PR #540, opened 2016-03-11
Fix an implicit toString of symbol object which should not throw TypeError

boingoing:
Force serialize exposed a bug in JavascriptConversion::ToString which throws if it is passed a symbol wrapper object instead of correctly calling Symbol.prototype.toString.


akroshg:
Does this fix #435 as well?

boingoing:
Yes, looks like this also fixes #435

akroshg:
Awesome - looks good.

Issue #539, opened 2016-03-11
Incorrect name for class function member


abchatra:
See the following kangax test case
```js
class C { foo(){} };
print((new C).foo.name);
```
Expected: foo
Actual: prototype.foo

suwc:
Leave a comment

abchatra:
Thank you @farzonl

PR #541, opened 2016-03-11
We did not preserve the shortname offset of the last call to AppendNa…

farzonl:
…meHints, this fix does that

fix for #539


farzonl:
@digitalinfinity as per our conversation here is a contribution.
@suwc Hope you don't mind I stole this one from you, can you please review?

boingoing:
LGTM

suwc:
LGTM

PR #542, opened 2016-03-11
Fix compact counter in debugging case

leirocks:
In debugging mode function can be reparsed and set the numbers on function body after background code generator has already started. But when reparsing the background jit should have been stopped, so reset bgThreadCallStarted and allow the setter access for reparsing


leirocks:
@agarwal-sandeep @curtisman can you review please

agarwal-sandeep:
Looks ok to me

PR #545, opened 2016-03-11
Add BrOnClassConstructor to ARM LowerMD

boingoing:
LowerMD on ARM did not catch the case of BrOnClassConstructor which could cause an AV.


boingoing:
@abchatra @rajatd Simple change. Would you mind taking a quick look?

rajatd:
:+1:

PR #547, opened 2016-03-12
Correctly determine where to set breakpoint if cursor is at end of inner function

agarwal-sandeep:
While determining where to set breakpoint, if cursor (ibos) is just after
the closing braces of the inner function then we can't simply put
breakpoint there, we have to check if next statement starts
after a line break or not.


agarwal-sandeep:
@akroshg please take a look

akroshg:
Looks good to me

PR #549, opened 2016-03-12
SIMD.js - Fixed incorrect operand size in  lowering of shuffle op. Disabled failing shift test on jshost

nmostafa:
Incorrect opnd size was overwritting memory causing jshost failure. Disabled another failing test. Will investigate later.

### DIFF
--- a/lib/Backend/LowerMDSharedSimd128.cpp
+++ b/lib/Backend/LowerMDSharedSimd128.cpp
@@ -735,7 +735,12 @@ IR::Instr* LowererMD::Simd128LowerLdLane(IR::Instr *instr)
             if (laneType == TyInt8)
             {
                 IR::RegOpnd * tmp = IR::RegOpnd::New(TyInt8, m_func);
-                newInstr = IR::Instr::New(Js::OpCode::MOV, tmp, dst, m_func);
+#ifdef _M_IX86
+                tmp->SetReg(RegEBX);
+#else
+                tmp->SetReg(RegRBX);
+#endif
+                newInstr = IR::Instr::New(Js::OpCode::MOV, tmp->UseWithNewType(TyInt32, m_func), dst, m_func);
                 instr->InsertBefore(newInstr);
                 newInstr = IR::Instr::New(Js::OpCode::MOVSX, dst, tmp, m_func);
             }
@@ -758,7 +763,11 @@ IR::Instr* LowererMD::Simd128LowerLdLane(IR::Instr *instr)
     {
         IR::Instr* pInstr    = nullptr;
         IR::RegOpnd* tmp = IR::RegOpnd::New(TyInt8, m_func);
-
+#ifdef _M_IX86
+        tmp->SetReg(RegEBX);
+#else
+        tmp->SetReg(RegRBX);
+#endif
         // cmp      dst, -1
         pInstr = IR::Instr::New(Js::OpCode::CMP, m_func);
         pInstr->SetSrc1(dst);
@@ -766,8 +775,8 @@ IR::Instr* LowererMD::Simd128LowerLdLane(IR::Instr *instr)
         instr->InsertBefore(pInstr);
         Legalize(pInstr);
 
-        // mov     tmp(TyInt8), dst
-        instr->InsertBefore(IR::Instr::New(Js::OpCode::MOV, tmp, dst, m_func));
+        // mov     tmp(TyInt32), dst
+        instr->InsertBefore(IR::Instr::New(Js::OpCode::MOV, tmp->UseWithNewType(TyInt32, m_func), dst, m_func));
 
         // sete     tmp(TyInt8)
         pInstr = IR::Instr::New(Js::OpCode::SETE, tmp, tmp, m_func);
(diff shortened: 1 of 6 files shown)


nmostafa:
@MikeHolman @pleath @akroshg

dilijev:
Can you file a bug for the test you disabled and include the issue number in the comment?

dilijev:

dilijev:
@nmostafa I expect some of those daily tasks to fail. I just want to see what's left to fix. If you're planning to update this PR your next PR will not trigger the additional checks. Otherwise you can close and reopen the issue to clear the failing tasks and re-run the normal ones. Not the most elegant, I'll grant you, but for something like this where not all issues are resolved we don't really have a better mechanism.

nmostafa:
@LouisLaf @MikeHolman,
Some jshost failures were related to regAlloc generating bad code if casting to TyInt8. And I had to pin TyInt8 operands to a byteable register. What we want to do is cast reg2 to 1-byte then sign-extend:
```
        IR::RegOpnd *reg2 = IR::RegOpnd::New(TyInt32, m_func);
        IR::RegOpnd * tmp = IR::RegOpnd::New(TyInt8, m_func);
        instr->InsertBefore(IR::Instr::New(Js::OpCode::MOV, tmp->UseWithNewType(TyInt32, m_func), reg2, m_func));

        instr->InsertBefore(IR::Instr::New(Js::OpCode::MOVSX, reg2, tmp, m_func));
```
Unless we pin tmp to RegEBX, we could end up with tmp assigned to a non-byte-addressable register (e.g. esi). The code then looks like this:
```
mov         esi,eax 
movsx       eax,dh  // bad src reg encoding
```

Trying something like
```
        IR::RegOpnd *reg2 = IR::RegOpnd::New(TyInt32, m_func);
        IR::RegOpnd * tmp = IR::RegOpnd::New(TyInt8, m_func);
        instr->InsertBefore(IR::Instr::New(Js::OpCode::MOV, tmp, reg2, m_func));

        instr->InsertBefore(IR::Instr::New(Js::OpCode::MOVSX, reg2, tmp, m_func));
```
doesn't work either if reg2 gets assigned to esi. And we end up with bad encoding for the MOV still.

nmostafa:

dilijev:
@nmostafa did the tests not run for you or did you update the code review after your last comment?

lib/Backend/LowerMDSharedSimd128.cpp
```diff
@@ -1327,7 +1336,11 @@ IR::Instr* LowererMD::Simd128LowerShift(IR::Instr *instr)
         IR::RegOpnd * shamtReg = IR::RegOpnd::New(TyInt8, m_func);
         shamtReg->SetReg(LowererMDArch::GetRegShiftCount());
         IR::RegOpnd * tmp = IR::RegOpnd::New(TyInt8, m_func);
-
+#ifdef _M_IX86
```
pleath: Are we sure LinearScan tracks physical registers in a way that makes this safe? Do we understand why FilterRegIntSizeConstraints is not guaranteeing that we get a byteable register when we need one?

nmostafa:
@dilijev , the tests did not run. I have noticed they haven't run in #551 either.

lib/Backend/LowerMDSharedSimd128.cpp
nmostafa: This done elsewhere in LowererMDShared, and in situations where we need the shift amount to be in cl. So it should work here as well (albeit some potential perf loss, if we have to spill ebx).

I haven't looked into why LinearScan doesn't assign a byteable reg. My guess is that a sym will get byteable reg or not upon its def instr. So if we have Ld d(TyInt8), s(TyInt32), d will get a byteable reg, but the problem is that s will not be moved to a byteable reg, if it is not in one. In other words, I don't think value are moved to a byteable reg if a use demands it.

pleath: Yes, so I see. I wonder if the right fix isn't to require that a def and uses of a RegOpnd all have the same type, so that it's up to the lowered code gen to insert moves to guarantee this. We can look into that this week, but for now this emergency fix will do.

pleath: I can take care of pushing this once it's had test coverage. Doug, are watching that?

-- Paul

dilijev:

dilijev:
@nmostafa Looks like they ran on #551. Maybe you looked while they were still queued. I'll look into why yours don't seem to be triggering.

dilijev:

@nmostafa I forgot that I changed the trigger phrase for this set to fix #420. I maybe didn't need to remove the dev12 trigger because I added the requirement of the word "tests" afterwards.

dilijev:
@nmostafa Looks like they will run now.

dilijev:
@nmostafa looks like the legacy builds are failing the same set as they currently do in the dailies. Also you mentioned you disabled a test but I don't see that in the diff. Did you remove that change?

nmostafa:
@dilijev the legacy failures are expected as we discussed (asmjs link failure logic creates two dynamic profiles). I can disable SIMDLink-*.js tests if needed. 

Yes, I did submit another commit to this PR that fixes the failing test and re-enabled it.

nmostafa:

dilijev:

dilijev:
@nmostafa Things are looking good. Since my PreFAST warning fixes are not yet merged in, you should still see those failures. If that's all that fails, you can do the following:

1. Cause the checks to reset by either:
  1. Rebasing your change on master and force-pushing
  2. Closing and reopening this issue to get only the basic set of checks to run again
2. When green, merge it in.

nmostafa:
@dilijev , Just closed and reopened,  dev12 failing builds are still there :)

dilijev:
Maybe closing and reopening doesn't remove tests that are non-default (only retriggers the default ones). I think rebase and force push is what you need to do.

nmostafa:
@pleath  will handle the merge.

nmostafa:
@dotnet-bot test this please

PR #552, opened 2016-03-12
Assertion: default parameter with class expression

suwc:
Test case:  `function foo(a=class { }) { }`
     `Assertion failure: childScope == scope->GetEnclosingScope(), FuncInfo::OnStartVisitScope()`
     When parameter scope is merged with body scope, currentChildScope is pushed twice (vs. once otherwise)
     FuncInfo::OnEndVisitScope(), however, does not pop twice to recover the scope stack, causing the problem.
     Fix by adding a local flag and pop scope stack properly for merged scope case.
     Re-enable test case.

### DIFF
--- a/lib/Runtime/ByteCode/FuncInfo.cpp
+++ b/lib/Runtime/ByteCode/FuncInfo.cpp
@@ -427,8 +427,10 @@ void FuncInfo::OnEndVisitFunction(ParseNode *pnodeFnc)
     this->SetCurrentChildFunction(nullptr);
 }
 
-void FuncInfo::OnStartVisitScope(Scope *scope)
+void FuncInfo::OnStartVisitScope(Scope *scope, bool *pisMergedScope)
 {
+    *pisMergedScope = false;
+
     if (scope == nullptr)
     {
         return;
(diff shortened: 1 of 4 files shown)


suwc:
@aneeshdk @akroshg @abchatra , could you take a look pls?

lib/Runtime/ByteCode/FuncInfo.cpp
```diff
@@ -427,11 +427,13 @@ void FuncInfo::OnEndVisitFunction(ParseNode *pnodeFnc)
     this->SetCurrentChildFunction(nullptr);
 }
 
-void FuncInfo::OnStartVisitScope(Scope *scope)
+bool FuncInfo::OnStartVisitScope(Scope *scope)
```
akroshg: I would rather prefer *out* variable from the function, instead of returning bool from the function. like (Scope *scope, bool *isMergedScope).

aneeshdk:
LGTM

lib/Runtime/ByteCode/FuncInfo.cpp
suwc: Makes sense. Updated in latest rev.

PR #557, opened 2016-03-14
Change assigning to const to runtime TypeError

suwc:
Current ECMA spec spec says assigning to const is runtime TypeError, not early SyntaxError.
Fix by removing assign-to-const checks/throws in parser and changing runtime error type of ERRAssignmentToConst to TypeError.
Update unit test baselines.

Issue explained

Const declarations are created in an environment record via CreateImmutableBinding(name, true).
The second argument means that the binding is made "strict" which is important below.
Assignments to variables are evaluated via PutValue.
E.g. step 1.f for simple assignment in https://tc39.github.io/ecma262/#sec-assignment-operators-runtime-semantics-evaluation
In the case of a variable assignment, PutValue calls into SetMutableBinding on the corresponding environment record.
SetMutableBinding throws if the S parameter is true in step 6.
For const variables this parameter will be true (see CreateImmutableBinding above).
e.g.
`const x=0;`
`foo() { print('foo'); } // 'hello' should be printed`
`x = foo(); // foo() should be called, then TypeError thrown`
`// Instead we threw a SyntaxError during parse.`

### DIFF
--- a/lib/Parser/Hash.h
+++ b/lib/Parser/Hash.h
@@ -62,32 +62,25 @@ public:
 
 struct PidRefStack
 {
-    PidRefStack() : isAsg(false), isDynamic(false), id(0), span(), sym(nullptr), prev(nullptr), isModuleExport(false) {}
-    PidRefStack(int id) : isAsg(false), isDynamic(false), id(id), span(), sym(nullptr), prev(nullptr), isModuleExport(false) {}
+    PidRefStack() : isDynamic(false), id(0), sym(nullptr), prev(nullptr), isModuleExport(false) {}
+    PidRefStack(int id) : isDynamic(false), id(id), sym(nullptr), prev(nullptr), isModuleExport(false) {}
 
-    charcount_t GetIchMin()   { return span.GetIchMin(); }
-    charcount_t GetIchLim()   { return span.GetIchLim(); }
     int GetScopeId() const    { return id; }
     Symbol *GetSym() const    { return sym; }
     void SetSym(Symbol *sym)  { this->sym = sym; }
-    bool IsAssignment() const { return isAsg; }
     bool IsDynamicBinding() const { return isDynamic; }
     void SetDynamicBinding()  { isDynamic = true; }
     bool IsModuleExport() const { return isModuleExport; }
     void SetModuleExport()    { isModuleExport = true; }
-
-    void TrackAssignment(charcount_t ichMin, charcount_t ichLim);
 
     Symbol **GetSymRef()
     {
         return &sym;
     }
 
-    bool           isAsg;
     bool           isDynamic;
     bool           isModuleExport;
     int            id;
-    Span           span;
     Symbol        *sym;
     PidRefStack   *prev;
 };
@@ -169,18 +162,6 @@ public:
             }
         }
         return nullptr;
-    }
-
-    charcount_t GetTopIchMin() const
-    {
-        Assert(m_pidRefStack);
-        return m_pidRefStack->GetIchMin();
-    }
-
-    charcount_t GetTopIchLim() const
-    {
-        Assert(m_pidRefStack);
-        return m_pidRefStack->GetIchLim();
     }
 
     void PushPidRef(int blockId, PidRefStack *newRef)
--- a/lib/Parser/Parse.h
+++ b/lib/Parser/Parse.h
@@ -198,7 +198,6 @@ private:
     BOOL                m_uncertainStructure;
     bool                m_hasParallelJob;
     bool                m_doingFastScan;
-    Span                m_asgToConst;
     int                 m_nextBlockId;
 
     // RegexPattern objects created for literal regexes are recycler-allocated and need to be kept alive until the function body
@@ -930,7 +929,6 @@ public:
 
 private:
     void DeferOrEmitPotentialSpreadError(ParseNodePtr pnodeT);
-    template<bool buildAST> void TrackAssignment(ParseNodePtr pnodeT, IdentToken* pToken, charcount_t ichMin, charcount_t ichLim);
     PidRefStack* PushPidRef(IdentPtr pid);
     PidRefStack* FindOrAddPidRef(IdentPtr pid, int blockId, int maxScopeId = -1);
     void RemovePrevPidRef(IdentPtr pid, PidRefStack *lastRef);
--- a/test/LetConst/AssignmentToConst.js
+++ b/test/LetConst/AssignmentToConst.js
@@ -3,29 +3,29 @@
 // Licensed under the MIT license. See LICENSE.txt file in the project root for full license information.
 //-------------------------------------------------------------------------------------------------------
 
-try { eval("const x = 1; x = 2;"); } catch (e) { WScript.Echo(e); }
-try { eval("const x = 1; x += 2;"); } catch (e) { WScript.Echo(e); }
-try { eval("const x = 1; x -= 2;"); } catch (e) { WScript.Echo(e); }
-try { eval("const x = 1; x *= 2;"); } catch (e) { WScript.Echo(e); }
-try { eval("const x = 1; x /= 2;"); } catch (e) { WScript.Echo(e); }
-try { eval("const x = 1; x &= 2;"); } catch (e) { WScript.Echo(e); }
-try { eval("const x = 1; x |= 2;"); } catch (e) { WScript.Echo(e); }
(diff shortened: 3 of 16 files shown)


suwc:
@ianwjhalliday @pleath @abchatra @bterlson what do you think?

lib/Parser/Parse.h
```diff
@@ -198,7 +198,6 @@ class Parser
     BOOL                m_uncertainStructure;
     bool                m_hasParallelJob;
     bool                m_doingFastScan;
-    Span                m_asgToConst;
```
akroshg: There is a function TrackAssignment - is it still required in the parser side?

test/es6/forloops-per-iteration-bindings.js
```diff
@@ -89,10 +89,10 @@ var tests = [
     {
         name: "const variables in for-in and for-of loops cannot be assigned",
         body: function () {
-            assert.throws(function () { eval("for (const x in { a: 1 }) { x = 1; }"); }, SyntaxError, "assignment to const known at parse time in for-in loop", "Assignment to const");
-            assert.throws(function () { for (const x in { a: 1 }) { eval("x = 1;"); } }, ReferenceError, "assignment to const only known at run time in for-in loop", "Assignment to const");
-            assert.throws(function () { eval("for (const x of [ 0 ]) { x = 1; }"); }, SyntaxError, "assignment to const known at parse time in for-of loop", "Assignment to const");
-            assert.throws(function () { for (const x of [ 0 ]) { eval("x = 1;"); } }, ReferenceError, "assignment to const only known at run time in for-of loop", "Assignment to const");
+            assert.throws(function () { eval("for (const x in { a: 1 }) { x = 1; }"); }, TypeError, "assignment to const known at parse time in for-in loop", "Assignment to const");
```
akroshg: Do they required rewording?

akroshg: I meant the description of each Assert.

---
In reply to: [56048177](https://github.com/Microsoft/ChakraCore/pull/557#discussion_r56048177) [](ancestors = 56048177)

test/LetConst/AssignmentToConst.js
```diff
@@ -3,29 +3,29 @@
...
-try { eval("const x = 1; -- x;"); } catch (e) { WScript.Echo(e); }
-try { eval("const x = 1; {x++;}"); } catch (e) { WScript.Echo(e); }
-try { eval("const x = 1; {let x = 2; x++;}"); } catch (e) { WScript.Echo(e); }
-try { eval("const x = 1; {x++; let x = 2;}"); } catch (e) { WScript.Echo(e); }
-try { eval("const x = 1; {let x = 2;} x = 10"); } catch (e) { WScript.Echo(e); }
-try { eval("const x = 1; {const x = 2; x++;}"); } catch (e) { WScript.Echo(e); }
-try { eval("const x = 1; {const x = 2;} x++;"); } catch (e) { WScript.Echo(e); }
-try { eval("x = 1; {let x = 2;} const x = 10;"); } catch (e) { WScript.Echo(e); }
-try { eval("function f() {x = 1; {let x = 2;} const x = 10;} f();"); } catch (e) { WScript.Echo(e); }
+try { eval("WScript.Echo('test 1'); const x = 1; x = 2;"); WScript.Echo("passed"); } catch (e) { WScript.Echo(e); }
```
akroshg: Not related to your change - but this file can easily be changed to use the UnitTestFramework, so that we don't need to dump the baseline. What do you think?

test/LetConst/constreassign1.baseline
```diff
@@ -1,8 +1,8 @@
 3
-ReferenceError: Assignment to const
+TypeError: Assignment to const
```
akroshg: Same is true with this file - this test can easily be converted to use the UnitTestFramework, instead of dumping the baseline. I'll leave this up to you to make a call on this.

akroshg:
hmmm - so there are no destructuring case broken? I thought I have had destructuring scenarios covering the const.
eg.
const c = 10;
({c} = {c:11});

akroshg:
Rest looks good to me.

test/es6/forloops-per-iteration-bindings.js
suwc: All these should throw runtime TypeErrors now - but these are valid cases and their descriptions are still good.

test/LetConst/AssignmentToConst.js
suwc: I will make another PR to switch to UnitTestFramework so they will both be easier to read.

test/LetConst/constreassign1.baseline
suwc: Same as above - making pointed changes here and planning for UnitTestFramework switch in a separate PR.

pleath:
I'm surprised the byte code gen change is so small. What happens when you do simple "(function() { const x = 0; x = 1;})();" outside eval?

lib/Parser/Parse.h
suwc: No, TrackAssignment() is no longer necessary. Removed in latest rev.

suwc:
@akroshg I added your destructuring test case in the latest rev.
@pleath 
`(function() { const x=0; x=1; })();`
Before this change: SyntaxError
After this change: TypeError
These codepaths were used for eval, and now for non-eval cases as well.

pleath:
Great. I didn't notice any test cases where the const assignment takes place outside eval, but we can have such cases now. Could you add a big handful of them if you haven't already?

test/LetConst/AssignmentToConst.js
akroshg: Fine with me.

akroshg:
We probably have many test cases in the test262 - we can pick a few for our unittest.

test/es6/forloops-per-iteration-bindings.js
```diff
@@ -241,12 +241,12 @@ var tests = [
...
-            assert.throws(function () { for (const x = 0; eval("x++") < 3; ) { } }, ReferenceError, "assignment to const known at run time in the test expression", "Assignment to const");
-            assert.throws(function () { eval("for (const x = 0; x < 3; x += 1) { }"); }, SyntaxError, "assignment to const known at parse time in the increment expression", "Assignment to const");
-            assert.throws(function () { for (const x = 0; x < 3; eval("x += 1")) { } }, ReferenceError, "assignment to const known at run time in the increment expression", "Assignment to const");
-            assert.throws(function () { eval("for (const x = 0; x < 3; ) { x += 1; }"); }, SyntaxError, "assignment to const known at parse time in the body", "Assignment to const");
-            assert.throws(function () { for (const x = 0; x < 3; ) { eval("x += 1"); } }, ReferenceError, "assignment to const known at run time in the body", "Assignment to const");
+            assert.throws(function () { eval("for (const x = 0; x++ < 3; ) { }"); }, TypeError, "assignment to const known at parse time in the test expression", "Assignment to const");
```
akroshg: it says here - assignment to const known at parse time.

test/es6/forloops-per-iteration-bindings.js
ianwjhalliday: But some say "known at parse time" which is no longer true, right?

lib/Parser/Hash.h
```diff
@@ -76,8 +76,6 @@ struct PidRefStack
     bool IsModuleExport() const { return isModuleExport; }
     void SetModuleExport()    { isModuleExport = true; }
 
-    void TrackAssignment(charcount_t ichMin, charcount_t ichLim);
```
ianwjhalliday: `PidRefStack::isAsg` and `PidRefStack::span` look like they are also dead now, never set to anything but the constructor initialized values.

test/es6/forloops-per-iteration-bindings.js
suwc: Technically it is still true. We only choose not to track or throw at parse time by following the spec.

test/es6/forloops-per-iteration-bindings.js
suwc: These wordings seem okay in describing the differences between the lines ("known" != "thrown"), but we can certainly change them if there's a better way.

lib/Parser/Hash.h
suwc: Removed dead code.

suwc:
@pleath test cases with assignment to const outside of eval added.

PR #558, opened 2016-03-14
Fix for issue when marking a function as eval caller in defer and non-defer cases

aneeshdk:
Fix for issue when marking a function as eval caller in defer and non-defer cases

In defer parse case when parsing the argument list to a call we were
not returing the argument count. We mark the current function as
CallsEval only if the eval has at least one argument in the call. This
count check was missing in the defer parse case.

### DIFF
--- a/lib/Parser/Parse.cpp
+++ b/lib/Parser/Parse.cpp
@@ -3378,7 +3378,7 @@ ParseNodePtr Parser::ParsePostfixOperators(
                     }
                     else
                     {
-                        if (pToken->tk == tkID && pToken->pid == wellKnownPropertyPids.eval) // Detect eval
+                        if (pToken->tk == tkID && pToken->pid == wellKnownPropertyPids.eval && count > 0) // Detect eval
                         {
                             this->MarkEvalCaller();
                         }
@@ -3624,6 +3624,7 @@ ParseNodePtr Parser::ParseArgList( bool *pCallOfConstants, uint16 *pSpreadArgCou
             Error(ERRnoMemory);
         // Allow spread in argument lists.
         pnodeArg = ParseExpr<buildAST>(koplCma, nullptr, TRUE, /* fAllowEllipsis */TRUE);
+        ++count;
 
         if (buildAST)
         {
@@ -3639,7 +3640,6 @@ ParseNodePtr Parser::ParseArgList( bool *pCallOfConstants, uint16 *pSpreadArgCou
                 (*pSpreadArgCount)++;
             }
 
-            ++count;
             AddToNodeListEscapedUse(&pnodeList, &lastNodeRef, pnodeArg);
         }
         if (m_token.tk != tkComma)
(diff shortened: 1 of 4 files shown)


aneeshdk:

lib/Parser/Parse.cpp
```diff
@@ -4985,6 +4985,14 @@ bool Parser::ParseFncDeclHelper(ParseNodePtr pnodeFnc, ParseNodePtr pnodeFncPare
 
         this->ParseFncFormals<buildAST>(pnodeFnc, flags);
 
+        if (pnodeFnc->sxFnc.CallsEval() || pnodeFnc->sxFnc.ChildCallsEval())
+        {
+            if (!m_scriptContext->GetConfig()->IsES6DefaultArgsSplitScopeEnabled())
```
akroshg: Nit: both ifs can be combined.

lib/Parser/Parse.cpp
```diff
@@ -5094,14 +5102,7 @@ bool Parser::ParseFncDeclHelper(ParseNodePtr pnodeFnc, ParseNodePtr pnodeFncPare
...
 
-                if (pnodeFnc->sxFnc.CallsEval() || pnodeFnc->sxFnc.ChildCallsEval())
-                {
-                    if (!m_scriptContext->GetConfig()->IsES6DefaultArgsSplitScopeEnabled())
-                    {
-                        Error(ERREvalNotSupportedInParamScope);
-                    }
-                }
-                else
+                if (paramScope->GetCanMergeWithBodyScope())
```
akroshg: just curious how can this be false? aren't we setting SetCannotMerge here itself?

lib/Parser/Parse.cpp
```diff
@@ -3659,9 +3659,9 @@ ParseNodePtr Parser::ParseArgList( bool *pCallOfConstants, uint16 *pSpreadArgCou
         CHAKRATEL_LANGSTATS_INC_LANGFEATURECOUNT(SpreadFeatureCount, m_scriptContext);
     }
 
+    *pCount = (uint16)count;
```
tcare: I consider it optional since you just moved it but maybe we want to do a C++ style cast here

lib/Parser/Parse.cpp
aneeshdk: Fixed.

lib/Parser/Parse.cpp
aneeshdk: I am keeping it there for the eval case when I implement it later. If param scope has eval we will mark it as cannot merge from there.

lib/Parser/Parse.cpp
aneeshdk: Fixed it.

tcare:
We should add a defer parse variant to rl for default.js

akroshg:
Agree with @tcare 
rest looks good to me.

lib/Parser/Parse.cpp
```diff
@@ -4985,6 +4985,12 @@ bool Parser::ParseFncDeclHelper(ParseNodePtr pnodeFnc, ParseNodePtr pnodeFncPare
 
         this->ParseFncFormals<buildAST>(pnodeFnc, flags);
 
+        if ((pnodeFnc->sxFnc.CallsEval() || pnodeFnc->sxFnc.ChildCallsEval())
+                && !m_scriptContext->GetConfig()->IsES6DefaultArgsSplitScopeEnabled())
+        {
+                Error(ERREvalNotSupportedInParamScope);
+        }
```
ianwjhalliday: Could this go in ParserFncFormals instead?  Seems weird to be here in this higher level code.

PR #559, opened 2016-03-14
Fixing current deferred stub issue due to default on lambda

akroshg:
Default as function on lambda's formal location could be problematic when
they are undeferring. When we parse those default function we don't know
that they are part of lambda so those functions will try to use
m_currDeferredStub - which was not meant for them (as it was allocated for
the lamda). I have fixed that by using the function location (ichMin) as a
checkpoint, if they match we will use the current deferred stub otherwise
it is possible that is meant for lambda.
There is no simple repro and repro involes the page heap enabled, That is
why I didn't add any unittest case for that.

### DIFF
--- a/lib/Parser/Parse.cpp
+++ b/lib/Parser/Parse.cpp
@@ -35,9 +35,7 @@ struct DeferredFunctionStub
     uint fncFlags;
     uint nestedCount;
     DeferredFunctionStub *deferredStubs;
-#if DEBUG
     charcount_t ichMin;
-#endif
 };
 
 struct StmtNest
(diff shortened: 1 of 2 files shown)


akroshg:
@pleath  - please have a look. (Octane seems flat)

Here is the scenario (this reproes when page-heap enabled)
```javascript
function test2(){ };

function test3(){
    {
        var a = (b = function () {
            return {
                f1: function() { },
                f2: function() { },
            };
        }, c = function () {
        } ) => {};
        
    }
};

test2();
```
run this with forcedeferparse.

pleath:
Is this problem happening because of the reparse that happen when we find the arrow operator?

akroshg:
This is happening because we are un-deferring the test3 function (if you look a the example above).
Since the functions on default of lambda do not know they are part of lambda, so they were pulling the deferredstub from the parent function (parent of lambda, test3). After reparse of the lambda it will be fine.

pleath:
I see. So because we haven’t seen the arrow, we don’t know that these are default value expressions. Argh. Okay, seems fine.

--Paul

PR #561, opened 2016-03-14
Address perf regression from TypedArray constructor change related to iterable object

suwc:
Recent TypedArray constructor change to support iterable object as input had unintentionally
shadowed codepaths constructing a TypedArray from an array (iterable object), causing perf regression.

Fix by adding stricter checks to keep arrays out of the new codepath designed specifically for iterable object. This is done by retrieving `[Symbol.iterator]` property of the object and checking if it is not the built-in Array.prototype[Symbol.iterator], or if the object is not an array but still has an iterator (not `undefined`). Only if these conditions are met would the object be used as an iterable for constructing the `TypedArray`. Otherwise it will go down the existing codepaths for array or error-throwing.

### DIFF
--- a/lib/Runtime/Library/JavascriptLibrary.h
+++ b/lib/Runtime/Library/JavascriptLibrary.h
@@ -620,7 +620,7 @@ namespace Js
         JavascriptFunction* GetSetConstructor() const {return  setConstructor; }
         JavascriptFunction* GetSymbolConstructor() const {return symbolConstructor; }
         JavascriptFunction* GetEvalFunctionObject() { return evalFunctionObject; }
-        JavascriptFunction* GetArrayPrototypeValuesFunction() { return arrayPrototypeValuesFunction; }
+        JavascriptFunction* GetArrayPrototypeValuesFunction() { return EnsureArrayPrototypeValuesFunction(); }
         DynamicObject* GetMathObject() const {return mathObject; }
         DynamicObject* GetJSONObject() const {return JSONObject; }
         DynamicObject* GetReflectObject() const { return reflectObject; }
--- a/lib/Runtime/Library/TypedArray.cpp
+++ b/lib/Runtime/Library/TypedArray.cpp
@@ -143,7 +143,11 @@ namespace Js
             {
                 if (JavascriptOperators::IsObject(firstArgument))
                 {
-                    if (JavascriptOperators::IsIterable(RecyclableObject::FromVar(firstArgument), scriptContext))
+                    Var iterator = JavascriptOperators::GetProperty(RecyclableObject::FromVar(firstArgument), PropertyIds::_symbolIterator, scriptContext);
+
+                    if (!JavascriptOperators::IsUndefinedObject(iterator) &&
+                        (iterator != scriptContext->GetLibrary()->GetArrayPrototypeValuesFunction() ||
+                        !JavascriptArray::Is(firstArgument)))
                     {
                         return CreateNewInstanceFromIterableObj(RecyclableObject::FromVar(firstArgument), scriptContext, elementSize, pfnCreateTypedArray);
                     }
--- a/test/typedarray/TypedArrayBuiltins.js
+++ b/test/typedarray/TypedArrayBuiltins.js
@@ -83,6 +83,35 @@ var tests = [
                 assert.areEqual(2, arr[1], "TypedArray " + t + " created from iterable has element #1 == 2");
                 assert.areEqual(3, arr[2], "TypedArray " + t + " created from iterable has element #2 == 3");
             }
+
+            for(var t of TypedArray) {
+                var a = [1,2,3,4];
+                a[Symbol.iterator] = getIterableObj([99,0])[Symbol.iterator];
+                var arr = new this[t](a);
+                assert.areEqual(1, arr.length, "TypedArray " + t + " created from array with user-defined iterator has length == 1");
+                assert.areEqual(99, arr[0], "TypedArray " + t + " created from array with user-defined iterator has element #0 == 99");
+            }
+
+            function testTypedArrayConstructorWithIterableArray(t) {
+                Array.prototype[Symbol.iterator] = getIterableObj([99,0])[Symbol.iterator];
+                var arr = new this[t](a);
+                assert.areEqual(1, arr.length, "TypedArray " + t + " created from array with Array.prototype overridden has length == 1");
+                assert.areEqual(99, arr[0], "TypedArray " + t + " created from array with Array.prototype overriden has element #0 == 99");
+            }
+
+            var builtinArrayPrototypeIteratorDesc = Object.getOwnPropertyDescriptor(Array.prototype, Symbol.iterator);
+            var a = [1,2,3,4];
+            Object.defineProperty(Array.prototype, Symbol.iterator, {enumerable: false, configurable: true, writable: true});
+            testTypedArrayConstructorWithIterableArray('Int8Array');
+            testTypedArrayConstructorWithIterableArray('Uint8Array');
+            testTypedArrayConstructorWithIterableArray('Uint8ClampedArray');
+            testTypedArrayConstructorWithIterableArray('Int16Array');
+            testTypedArrayConstructorWithIterableArray('Uint16Array');
+            testTypedArrayConstructorWithIterableArray('Int32Array');
+            testTypedArrayConstructorWithIterableArray('Uint32Array');
+            testTypedArrayConstructorWithIterableArray('Float32Array');
+            testTypedArrayConstructorWithIterableArray('Float64Array');
+            Object.defineProperty(Array.prototype, Symbol.iterator, builtinArrayPrototypeIteratorDesc);
         }
     }
 ];


suwc:
@akroshg @abchatra @ianwjhalliday @curtisman could one of you take a look pls?

lib/Runtime/Library/TypedArray.cpp
```diff
@@ -175,6 +171,11 @@ namespace Js
                             Var lengthVar = JavascriptOperators::OP_GetProperty(jsArraySource, PropertyIds::length, scriptContext);
                             if (JavascriptOperators::GetTypeId(lengthVar) == TypeIds_Undefined)
```
abchatra: If there is a length property defined on the iterable object, is the new semantics right?

ianwjhalliday:
Special casing arrays is only legal if the `Array.prototype[Symbol.iterator]` property is still the built-in value.  If it has changed then we cannot special case array objects.

The constructor should look at the object's `[Symbol.iterator]` property and check if it is the array built-in, and if so, then check if the object is an array, and if so finally go down the array optimized path.

lib/Runtime/Library/TypedArray.cpp
suwc: Updated PR to use original codepath order. Gain perf by keeping arrays out of the slower codepath for iterables.

suwc:
@ianwjhalliday You are right. Updated in the new rev with additional test cases.

lib/Runtime/Library/JavascriptLibrary.h
```diff
@@ -448,6 +448,7 @@ namespace Js
         bool inProfileMode;
         bool inDispatchProfileMode;
         bool arrayObjectHasUserDefinedSpecies;
+        bool arrayObjectHasUserDefinedIterator;
```
ianwjhalliday: With the simplification I mentioned below, this should no longer be needed.

I am also wondering if the species flag can be eliminated through simplification where it is used.

test/typedarray/TypedArrayBuiltins.js
```diff
@@ -83,6 +83,23 @@ var tests = [
...
+                a[Symbol.iterator] = getIterableObj([99,0])[Symbol.iterator];
+                var arr = new this[t](a);
+                assert.areEqual(1, arr.length, "TypedArray " + t + " created from array with user-defined iterator has length == 1");
+                assert.areEqual(99, arr[0], "TypedArray " + t + " created from array with user-defined iterator has element #0 == 99");
+            }
+
+            for(var t of TypedArray) {
+                var a = [1,2,3,4];
+                Object.defineProperty(Array.prototype, Symbol.iterator,
+                    {enumerable: false, configurable: true, writable: true, value:getIterableObj([99,0])[Symbol.iterator]});
```
ianwjhalliday: Huge hazard here.  When you make a change to a global in a test you should undo that change afterwards so other tests are unaffected.

suwc: Thanks! This has been changed in a later rev.

lib/Runtime/Library/TypedArray.cpp
```diff
@@ -143,7 +143,11 @@ namespace Js
             {
                 if (JavascriptOperators::IsObject(firstArgument))
                 {
-                    if (JavascriptOperators::IsIterable(RecyclableObject::FromVar(firstArgument), scriptContext))
+                    if (JavascriptOperators::IsIterable(RecyclableObject::FromVar(firstArgument), scriptContext) &&
+                        (!JavascriptArray::Is(firstArgument) ||
+                        JavascriptOperators::HasOwnProperty(firstArgument, PropertyIds::_symbolIterator, scriptContext) ||
+                        DynamicObject::FromVar(firstArgument)->GetPrototype() != scriptContext->GetLibrary()->GetArrayPrototype() ||
+                        scriptContext->GetLibrary()->GetArrayObjectHasUserDefinedIterator()))
```
ianwjhalliday: This is unnecessarily complicated.

`IsIterable` is also suspicious.  To me it looks like for proxy objects we are going to call the `Symbol.iterator` property on proxy objects twice instead of once which have observable side effects that don't match the spec semantics.

We should do exactly as the spec says.  Get the `Symbol.iterator` property's value from the object.  At that point, before calling it, test to see if it is the built-in array iterator function.  If it is, _then_ check if the object to see if it is an array and do the fast path.  Otherwise continue with the rest of the spec algorithm.

Note that if the the array object did have its own `Symbol.iterator` but it was the built-in array iterator method we could still do the optimization, so there is no need to check only the prototype case.  There is no need to handle the cases separately at all since we always have to do a `Get` of `Symbol.iterator` on the object which will do the prototype lookup.  It is only the value returned from this `Get` that matters.

test/typedarray/TypedArrayBuiltins.js
```diff
@@ -83,6 +83,25 @@ var tests = [
...
+                var arr = new this[t](a);
+                assert.areEqual(1, arr.length, "TypedArray " + t + " created from array with user-defined iterator has length == 1");
+                assert.areEqual(99, arr[0], "TypedArray " + t + " created from array with user-defined iterator has element #0 == 99");
+            }
+
+            var builtinArrayPrototypeIteratorDesc = Object.getOwnPropertyDescriptor(Array.prototype, Symbol.iterator);
+            for(var t of TypedArray) {
+                var a = [1,2,3,4];
+                Object.defineProperty(Array.prototype, Symbol.iterator,
+                    {enumerable: false, configurable: true, writable: true, value:getIterableObj([99,0])[Symbol.iterator]});
```
ianwjhalliday: This doesn't need to be inside the loop.  Or resetting it should also be inside the loop.

lib/Runtime/Library/JavascriptLibrary.h
suwc: Removed in latest rev.
The species flag was there because we wanted to avoid any form of property look-up for perf reasons. So probably no.

lib/Runtime/Library/TypedArray.cpp
```diff
@@ -143,7 +143,11 @@ namespace Js
             {
                 if (JavascriptOperators::IsObject(firstArgument))
                 {
-                    if (JavascriptOperators::IsIterable(RecyclableObject::FromVar(firstArgument), scriptContext))
+                    Var iterator = JavascriptOperators::GetProperty(RecyclableObject::FromVar(firstArgument), PropertyIds::_symbolIterator, scriptContext);
+
+                    if (!JavascriptOperators::IsUndefinedObject(iterator) &&
```
ianwjhalliday: Is checking for undefined necessary?  Seems redundant at a glance

lib/Runtime/Library/TypedArray.cpp
suwc: I looked through `IsIterable()` and it only calls `JavascriptOperators::GetProperty()` or `::HasProperty()`, not doing any calling. Can't see any side effects.

Nevertheless, simplifying makes sense esp. since we don't mind property look-up here. Updated in the latest rev. Thanks!

lib/Runtime/Library/TypedArray.cpp
suwc: Yes it is.

lib/Runtime/Library/TypedArray.cpp
ianwjhalliday: `GetProperty` can call script code, on a getter property for instance, but particularly on a proxy object where property get can be trapped.  `GetProperty` and `SetProperty` can call script code and this something important to consider when using them.

Using `IsIterable` anywhere is a bug because of this.  I've opened issue #584 to track this separately.

lib/Runtime/Library/JavascriptLibrary.h
ianwjhalliday: Sounds good

lib/Runtime/Library/JavascriptLibrary.h
```diff
@@ -1216,8 +1216,9 @@ namespace Js
 
         JavascriptFunction* EnsureArrayPrototypeValuesFunction();
         
-
     public:
+        Var GetArrayPrototypeBuiltinIterator() { return EnsureArrayPrototypeValuesFunction(); }
```
ianwjhalliday: There is already a GetArrayPrototypeValuesFunction() method defined earlier in this class.  However it does not do the ensure call.  It should be updated to do the ensure and then you can use it instead of creating a new method.

test/typedarray/TypedArrayBuiltins.js
suwc: See the point. Getting rid of the loop here.

lib/Runtime/Library/JavascriptLibrary.h
suwc: Thanks. Updated.

lib/Runtime/Library/TypedArray.cpp
ianwjhalliday: Oh I see, `CreateNewInstanceFromIterableObj` doesn't handle the case where it is undefined.

Okay this is fine.

test/typedarray/TypedArrayBuiltins.js
ianwjhalliday: Oh, the loop itself was fine.  Just the override and reset of `Array#[Symbol.iterator]` should have been both inside or outside the loop, not mixed.

ianwjhalliday:
Looks good now.  Feel free to change the test to add back the loop or not.  I'm indifferent.

PR #563, opened 2016-03-14
Prevent "exec" results from getting GC'd during RegExp.prototype[@@replace]

goyakin:
[VSO bug](https://microsoft.visualstudio.com/web/wi.aspx?pcguid=cb55739e-4afe-46a3-970f-1b49d8ee7564&id=6857149)

### DIFF
--- a/lib/Runtime/Library/RegexHelper.cpp
+++ b/lib/Runtime/Library/RegexHelper.cpp
@@ -905,40 +905,41 @@ namespace Js
 
         JavascriptString* accumulatedResult = nullptr;
 
-        BEGIN_TEMP_ALLOCATOR(tempAlloc, scriptContext, _u("RegexHelper"))
+        Recycler* recycler = scriptContext->GetRecycler();
+
+        JsUtil::List<RecyclableObject*>* results = RecyclerNew(recycler, JsUtil::List<RecyclableObject*>, recycler);
+
+        while (true)
         {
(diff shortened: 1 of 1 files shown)


goyakin:

/cc @tcare

lib/Runtime/Library/RegexHelper.cpp
```diff
@@ -905,40 +905,41 @@ namespace Js
...
-            size_t previousNumberOfCapturesToKeep = 0;
-            Var* captures = nullptr;
+        CompoundString::Builder<64 * sizeof(void *) / sizeof(char16)> accumulatedResultBuilder(scriptContext);
+        CharCount inputLength = input->GetLength();
+        CharCount nextSourcePosition = 0;
+
+        size_t previousNumberOfCapturesToKeep = 0;
+        Var* captures = nullptr;
 
+        BEGIN_TEMP_ALLOCATOR(tempAlloc, scriptContext, _u("RegexHelper"))
```
abchatra: Is the use of tempAlloc in appendReplacement safe?

abchatra:
Looks good

lib/Runtime/Library/RegexHelper.cpp
goyakin: It's used to create an array of String indexes for the `ReplaceFormatString` function and won't be GC'd like `results`.

PR #565, opened 2016-03-15
Wasm Bug fixes

### LINKED ISSUE #440: WebAssembly pre-v1.0 update
The WebAssembly spec changed quite a bit recently and we have to update our implementation.
This is a list of things to do

- [x] : Add `Wasm` Object to the global object (#430)
- [x] : Add `instantiateModule` method to the `Wasm` Object. (In review #486 + #430) 
  - Takes 2 arguments
    - TypedArrayView with the bytes representing the wasm module
    - An Object representing the imports
- [x] : Modules return an `exports` (#452)
  - `var m = Wasm.instantiateModule(buf, imports); m.exports.foo();`
- [x] : Wasm magic number + version. [Module Header](https://github.com/WebAssembly/design/blob/master/BinaryEncoding.md#high-level-structure) (#448)
- [x] : Use LEB128 for module offsets and indexes (#477)
- [x] : Page count for memory section instead of size (In Review #446) [Memory Section](https://github.com/WebAssembly/design/blob/master/BinaryEncoding.md#memory-section)
- [x] : String for section instead of id. [Section Definition](https://github.com/WebAssembly/design/blob/master/BinaryEncoding.md#high-level-structure) (#448)
- [x] : Fixed order of section (#481)
  - We can assume the section will be in the order represented in the BinaryEncoding document, however we don't *have* to validate it for the time being (later work order).
- [x] : Export table. (#489) [Export Table Section](https://github.com/WebAssembly/design/blob/master/BinaryEncoding.md#export-table-section)
- [x] : Import table. [Import Table Section](https://github.com/WebAssembly/design/blob/master/BinaryEncoding.md#import-table-section) (@commonlisp)
  - [x] : Two level import
- [x] : Rename `tableswitch` to `br_table` (#407)
  - [x] : `br_table` changes (@MikeHolman)
- [x] : Split functions section into decls and definitions sections (#481 + #477) (https://github.com/WebAssembly/design/pull/591)
- [x] : Signed LEB128 for i32.const and i64.const (@commonlisp is this done ?) (https://github.com/WebAssembly/design/pull/592)
- [ ] : Data Segments section (In Review: #439)
  - [ ] : Inline data segments (https://github.com/WebAssembly/design/pull/590)
- [x] : Memory in exports object `module.exports.memory`

Cellule:
Closes #440
Closes #520 
- Cleanup bytecode writer if an exception occurs
- Implemented Lazy trap under `-on:WasmLazyTrap` flag. This will not throw if a function is invalid, instead it will throw when the function is called. Useful for feature detection and testing module with not yet implemented features.
- Implemented default exports. ie: `(export "" 0)` in sexpr. Only the first one will be used.
- Allows for 1 level import. ie : `imports.namespace.foo` and `imports.bar` are both valid imports
```js
const imports = {
  namespace: {
    foo(){}
  }, 
  bar() {}
}
```
- Expose the WasmModule to the WasmReader (this is not the best solution, but needed for lazy trap at this time)
- Added some tracing to help debug
- Yield last expression value if not a return statement
- Accept blocks with no expressions
- Fix bug in SLEB128 decoding for 4 bytes values
- Throwing Compilation Errors instead of asserts
@MikeHolman Could you comment on the things you changed

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/565)


lib/Runtime/Base/ScriptContext.cpp
```diff
@@ -1830,24 +1841,91 @@ namespace Js
...
+#endif
+            }
+
+            Js::Var exportsNamespace = nullptr;
+
+            // Check for Default export
+            for (uint32 iExport = 0; iExport < wasmModule->info->GetExportCount(); ++iExport)
+            {
+                Wasm::WasmExport* funcExport = wasmModule->info->GetFunctionExport(iExport);
+                if (funcExport && funcExport->nameLength == 0)
```
commonlisp: Curious, what is the rationale behind the empty export name being the default export? If we have both the default export and explicitly exported functions, what is the intended semantics? I don't recall this being specified in the design docs. Did they update that?

lib/Runtime/Base/ScriptContext.cpp
```diff
@@ -1830,24 +1841,91 @@ namespace Js
...
 
             Var* heap = moduleMemoryPtr + wasmModule->heapOffset;
 
             exportObj = JavascriptOperators::NewJavascriptObjectNoArg(this);
-            Js::Var exportsNamespace = JavascriptOperators::NewJavascriptObjectNoArg(this);
 
-            PropertyRecord const * exportsPropertyRecord = nullptr;
+            Var* localModuleFunctions = moduleMemoryPtr + wasmModule->funcOffset;
 
+            FrameDisplay * frameDisplay = RecyclerNewPlus(GetRecycler(), sizeof(void*), FrameDisplay, 1);
```
commonlisp: It could be helpful to break LoadWasmScript() into smaller functions as promised somewhere along the way...

lib/Runtime/Base/ScriptContext.cpp
```diff
@@ -1878,35 +1960,16 @@ namespace Js
...
-                entypointInfo->SetModuleAddress((uintptr_t)moduleMemoryPtr);
-                funcObj->SetEnvironment(frameDisplay);
-                localModuleFunctions[i] = funcObj;
-            }
-
             for (uint32 iExport = 0; iExport < wasmModule->info->GetExportCount(); ++iExport)
             {
                 Wasm::WasmExport* funcExport = wasmModule->info->GetFunctionExport(iExport);
-                if (funcExport)
+                if (funcExport && funcExport->nameLength > 0)
```
commonlisp: Would it make sense to emit an error if an function named "memory" was marked for export?

lib/WasmReader/WasmBinaryReader.cpp
```diff
@@ -382,6 +431,32 @@ WasmBinaryReader::CallNode()
     m_currentNode.var.num = funcNum;
 }
 
+void
+WasmBinaryReader::CallImportNode()
+{
+    UINT length = 0;
+    UINT32 funcNum = LEB128(length);
+    m_funcState.count += length;
+    if (funcNum >= m_moduleInfo->GetImportCount())
```
commonlisp: Can we factor out this validation code to consolidate the CallNode flavors?

lib/WasmReader/WasmBinaryReader.cpp
```diff
@@ -235,19 +273,24 @@ WasmBinaryReader::ReadFunctionBodies(FunctionBodyCallback callback, void* callba
         {
             UINT32 count = LEB128(len);
             m_funcState.count += len;
-            UINT8 type = ReadConst<UINT8>();
-            m_funcState.count++;
-            if (type >= Wasm::WasmTypes::Limit || type == Wasm::WasmTypes::Void)
+            Wasm::WasmTypes::WasmType type = ReadWasmType(len);
```
commonlisp: Are we anticipating that types will be larger than 1 byte?

lib/WasmReader/WasmByteCodeGenerator.cpp
```diff
@@ -485,17 +513,20 @@ WasmBytecodeGenerator::EmitBlock()
     if (m_reader->IsBinaryReader())
     {
         UINT blockCount = m_reader->m_currentNode.block.count;
-        if (blockCount <= 0)
+        if (blockCount > 0)
         {
-            throw WasmCompilationException(_u("Invalid block node count"));
```
commonlisp: Where does the validation of block node counts happen now? Perhaps it should be validated in the WasmBinaryReader.

lib/WasmReader/WasmByteCodeGenerator.cpp
```diff
@@ -485,17 +513,20 @@ WasmBytecodeGenerator::EmitBlock()
...
+            }
+            // block yields last value
+            blockInfo = EmitExpr(m_reader->ReadFromBlock());
         }
-        for (UINT i = 0; i < blockCount - 1; i++)
+        else
         {
-            EmitInfo tmpInfo = EmitExpr(m_reader->ReadFromBlock());
-            ReleaseLocation(&tmpInfo);
+            blockInfo = EmitInfo(Wasm::WasmTypes::Void);
```
commonlisp: Hmm, so an empty block with no count is a legal form or perhaps is blockCount being used for something else?

lib/WasmReader/WasmByteCodeGenerator.cpp
```diff
@@ -521,31 +552,34 @@ WasmBytecodeGenerator::EmitLoop()
 {
     Js::ByteCodeLabel loopTailLabel = m_writer.DefineLabel();
     m_labels->Push(loopTailLabel);
-
+
```
commonlisp: Nit: whitespace

lib/Runtime/Base/ScriptContext.cpp
Cellule: I don't think this is specified in the spec. At this time this is a convenience for testing.
If you have a default export and normal exports, we will attach on the default function.
```js
const m = Wasm.instantiateModule(convertToBinary(`(module (func (result i32) (i32.const 1)) (func (result i32) (i32.const 2)) (export "" 0) (export "t" 1))`), {});
print(m.exports());
print(m.exports.t());
```
will print
```
1
2
```

lib/Runtime/Base/ScriptContext.cpp
Cellule: I agree, there will be some effort in the near future to cleanup the current code base of web assembly branch before we move on to the next work items

lib/WasmReader/WasmBinaryReader.cpp
Cellule: @MikeHolman

lib/WasmReader/WasmBinaryReader.cpp
Cellule: I suppose we don't at this time, this is an effort to minimize possible errors if it does change, plus it looks like the way we use LEB128.

lib/WasmReader/WasmByteCodeGenerator.cpp
Cellule: We'll have to verify if an empty block is valid or not.
I changed it because I came across a test case that uses empty block. The test might have been out of date though.

lib/WasmReader/WasmByteCodeGenerator.cpp
Cellule: Well the block count is used to know how many expressions we have to read.
See my previous comment for empty block

Cellule:
I created the issue #568 to address most of the known issues and those discussed in this PR

lib/Runtime/Base/ScriptContext.cpp
Cellule: see (#568)

lib/Runtime/Base/ScriptContext.cpp
Cellule: Good point, we should do that (see #568)

lib/Backend/NativeCodeGenerator.cpp
```diff
@@ -1456,7 +1458,7 @@ NativeCodeGenerator::Prioritize(JsUtil::Job *const job, const bool forceAddJobTo
 
 ExecutionMode NativeCodeGenerator::PrejitJitMode(Js::FunctionBody *const functionBody)
 {
-    Assert(IS_PREJIT_ON());
+    Assert(IS_PREJIT_ON() || functionBody->IsAsmJSModule());
```
MikeHolman: I think it should be AsmJsMode, not module (IIRC module flag is only set for the closure itself).

lib/Runtime/Base/ScriptContext.cpp
MikeHolman: I don't know if we want this. I remember discussions about this, and as far as I can tell the conclusion was that empty string should not be treated as a special default export (and yes, I know SM has this right now).

lib/Runtime/Base/ScriptContext.cpp
```diff
@@ -1860,11 +1938,15 @@ namespace Js
...
                     const uint32 offset = segment->getDestAddr();
                     const uint32 size = segment->getSourceSize();
-                    if (offset > maxSize || UInt32Math::Add(offset, size) >= maxSize)
+                    if (offset > maxSize || UInt32Math::Add(offset, size) > maxSize)
                     {
                         throw Wasm::WasmCompilationException(_u("Data segment #%u is out of bound"), iSeg);
                     }
-                    js_memcpy_s(buffer + offset, (uint32)maxSize - offset, segment->getData(), size);
+
+                    if (size > 0)
```
MikeHolman: More a spec question, but I wonder if 0 size data segment should be considered compilation error?

lib/WasmReader/WasmBinaryReader.cpp
MikeHolman: Now that you mention it I'm not sure why check was here to begin with. I think in WasmBytecodeGenerator we already have this (or we should). Generally, I think semantic validation should not be done in the binary decoder. I'll factor this back out.

lib/WasmReader/WasmBinaryReader.cpp
```diff
@@ -346,9 +393,11 @@ WasmBinaryReader::ASTNode()
 #include "WasmBinaryOpcodes.h"
 
     default:
-        Assert(UNREACHED);
+        ThrowDecodingError(_u("Unknown opcode %u"), op);
     }
 
+    m_ops->AddNew(op);
```
MikeHolman: let's move this behind #if DBG

lib/WasmReader/WasmByteCodeGenerator.cpp
MikeHolman: empty blocks are currently considered valid

MikeHolman:
lgtm

PR #566, opened 2016-03-15
Simple long-standing objtypespec bug

pleath:
Types that are used in equivalent objtypespec guards must be locked to protect against, e.g., adding a property to a dictionary type handler, which will only change the type if it's locked.


rajatd:
Could there be other cases for which we would now start locking the type?

pleath:
Sharable type handlers are already going to be locked. This should just affect dictionary cases, and we need to lock those in case properties get added. (Weird that it took so long to find such a simple issue.)

-- Paul

rajatd:
Thanks, @pleath . Although, what I meant to ask was- for dictionary types themselves, could there be other operations than property adds, for which we will now lock types?

pleath:
You mean, are there cases where we will have locked the guard type and now have to promote the type when something other than a property add happens? Well, we do want to promote the type on a type ID change (see the comment that’s already there). I suppose we might now promote on a change to property attributes. For change to non-writable, promotion is probably a good idea anyway. Change in configurability or enumerability is bound to be a corner case, so I’d rather just accept the promotion for the sake of simplicity. Perf is flat, by the way. (I actually saw an improvement in gbemu.js, but I think it was flakiness in the run.)

--Paul

rajatd:
Ok, makes sense. Looks good.

PR #567, opened 2016-03-15
Fix PreservedVirtualAllocWrapper and CodeGenAllocators' HeapPageAllocator clean up order.

curtisman:
In the ThreadContext dtor sequence, PreservedVirtualAllocWrapper::Shutdown is called explicitly, before the CodeGenAllocators' HeapPageAllocator dtor is called.  The release should be in PreReservedVirtualAllocWrapper dtor instead, which will be the right order.


curtisman:
@satheeshravi   Please code review.

satheeshravi:
lgtm

PR #572, opened 2016-03-15
The 'async' function has inner function which has default threw assert.

akroshg:
In the async function transformation we set m_currentNodeDeferredFunc to the current node, which is wrong. This variable should be set if buildAST is false, which is not the case. I have removed that. Also I have put assert at one place where when we use the m_currentNodeFnc, the m_currentNodeDeferredFunc should be null.
found this in internal tool. Added a unittest.


akroshg:
@ianwjhalliday - please have a look.

ianwjhalliday:
LGTM

PR #574, opened 2016-03-15
Should throw error on initializer for shorthand of object literals

akroshg:
{x = 1} is not allowed for object literals but allowed for pattern. We have delayed that error. In obscure cases, though got the delayed-error but didn't raise them. Fixed that by raising it and added a test for clarity.

### DIFF
--- a/test/es6/destructuring_obj.js
+++ b/test/es6/destructuring_obj.js
@@ -206,6 +206,9 @@ var tests = [
             ({x:{a1 = 2}} = {x:{}});
             assert.areEqual(a1, 2);
         }
+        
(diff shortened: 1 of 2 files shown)


akroshg:
@tcare  please have a look at it.

test/es6/destructuring_obj.js
```diff
@@ -206,6 +206,9 @@ var tests = [
             ({x:{a1 = 2}} = {x:{}});
             assert.areEqual(a1, 2);
         }
+        
+        assert.throws(function () { eval("var a = 1; switch(true) {  case {a = 1} : break; };"); }, SyntaxError,"Object literal on case has initializer is not valid syntax", "Expected ':'");
```
tcare: Add a space to SyntaxError,"

tcare:
Looks good.

PR #576, opened 2016-03-15
Fix -0 tracking bugs in Add and Sub operations

rajatd:
The forward pass sets a bit, m_wasNegativeZeroPreventedByBailout, on an
operand if it was int-type-specialized and was protected by a
BailOutOnNegativeZero (which would happen if there are uses of that
operand downstream in manners which require distinction between -0 and +0
and the forward pass cannot say for sure that the operand can't be -0).
The Deadstore pass then looks at this bit to determine if it can remove
some -0 bailouts.

However, this bit being set to false for an operand doesn't necessarily mean that that
operand cannot be -0 if the operation that produced this operand wasn't type-spec'ed at all.
Negative zero tracking in the deadstore pass wasn't taking this into
account and was getting rid of more -0 bailouts than it should have,
resulting in wrong results.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/576)

### DIFF
--- a/lib/Backend/BackwardPass.cpp
+++ b/lib/Backend/BackwardPass.cpp
@@ -5158,15 +5158,34 @@ BackwardPass::TrackIntUsage(IR::Instr *const instr)
                 Assert(instr->GetSrc2()->IsRegOpnd() || instr->GetSrc2()->IsIntConstOpnd());
 
                 if(instr->ignoreNegativeZero ||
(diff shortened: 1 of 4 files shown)


rajatd:
Fixes [2918699] (https://microsoft.visualstudio.com/DefaultCollection/OS/_workItems?_a=edit&id=2918699&triage=true)

rajatd:
Discussed offline with @pleath - the fix is not correct as I'm walking the instructions lexically and not flow-based. Closing this PR, will open a new one with a robust fix.

PR #581, opened 2016-03-15
Restore the 'dev12 tests' trigger. The requirement of the word 'tests' serves as a disambiguator and calling out 'dev12' is more natural than 'legacy'.

dilijev:
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/581)


dilijev:
@mmitche FYI

PR #582, opened 2016-03-15
remove assert against asm.js code in strict mode

MikeHolman:
There is no reason you can't use asm.js under strict mode.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/582)


MikeHolman:
@abchatra @Cellule, can you please take a look.

abchatra:
LGTM

PR #583, opened 2016-03-15
Have RegExp constructor process RegExp-like objects correctly

### LINKED ISSUE #577: RegExp constructor should process RegExp-like Objects correctly
We differ from the spec in two aspects.

First, when the constructor is called as a function, we should use the `constructor` property of the passed RegExp-like object to determine whether we should return the argument as is. Kangax test case:

```JS
var re = /./;
re[Symbol.match] = false;
var foo = {constructor: RegExp};
foo[Symbol.match] = true;
return RegExp(re) !== re && RegExp(foo) === foo;
```

Second, we should use the `source` and `flags` properties of the passed RegExp-like object to initialize the new RegExp object. Kangax test case:

```JS
// RegExp -> Get -> [[Get]]
var get = [];
var re = { constructor: null };
re[Symbol.match] = true;
var p = new Proxy(re, { get: function(o, k) { get.push(k); return o[k]; }});
RegExp(p);
return get[0] === Symbol.match && get.slice(1) + '' === "constructor,source,flags";
```

goyakin:
Fixes #577

### DIFF
--- a/lib/Runtime/Library/JavascriptRegularExpression.cpp
+++ b/lib/Runtime/Library/JavascriptRegularExpression.cpp
@@ -258,7 +258,7 @@ namespace Js
         {
             pattern = scriptContext->GetLibrary()->GetEmptyRegexPattern();
         }
-        else if (JavascriptRegExp::Is(args[1]))
+        else if (JavascriptRegExp::IsRegExpLike(args[1], scriptContext))
         {
             if (!(callInfo.Flags & CallFlags_New) &&
                 (callInfo.Count == 2 || JavascriptOperators::IsUndefinedObject(args[2], scriptContext)) &&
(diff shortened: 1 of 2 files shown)


goyakin:

lib/Runtime/Library/JavascriptRegularExpression.cpp
```diff
@@ -258,48 +258,61 @@ namespace Js
         {
             pattern = scriptContext->GetLibrary()->GetEmptyRegexPattern();
         }
-        else if (JavascriptRegExp::Is(args[1]))
+        else if (JavascriptRegExp::IsRegExpLike(args[1], scriptContext))
```
akroshg: will you run the benchmark to verify the perf? (IsRegExpLike seems to costly at this point).

akroshg:
code wise looks good to me.

lib/Runtime/Library/JavascriptRegularExpression.cpp
goyakin: Benchmarks don't show any difference in perf.

PR #585, opened 2016-03-16
Make RegExp.prototype.toString generic

### LINKED ISSUE #578: RegExp.prototype.toString should be generic
Rather than requiring `this` to be a RegExp object, the `toString` method should be generic and use the `source` and `flags` properties to generate the string. Kangax test case:

```JS
// RegExp.prototype.toString -> Get -> [[Get]]
var get = [];
var p = new Proxy({}, { get: function(o, k) { get.push(k); return o[k]; }});
RegExp.prototype.toString.call(p);
return get + '' === "source,flags";
```

goyakin:
Fixes #578

### DIFF
--- a/lib/Runtime/Library/JavascriptRegularExpression.cpp
+++ b/lib/Runtime/Library/JavascriptRegularExpression.cpp
@@ -671,11 +671,37 @@ namespace Js
         Assert(!(callInfo.Flags & CallFlags_New));
         Assert(args.Info.Count > 0);
 
-        JavascriptRegExp* obj = GetJavascriptRegExp(args, _u("RegExp.prototype.toString"), scriptContext);
+        PCWSTR const varName = _u("RegExp.prototype.toString");
 
-        bool sourceOnly = false;
-        bool useFlagsProperty = scriptContext->GetConfig()->IsES6RegExPrototypePropertiesEnabled();
-        return obj->ToString(sourceOnly, useFlagsProperty);
+        const ScriptConfiguration* scriptConfig = scriptContext->GetConfig();
+
+        if (scriptConfig->IsES6RegExPrototypePropertiesEnabled())
+        {
+            RecyclableObject *thisObj = GetThisObject(args, varName, scriptContext);
+            JavascriptString* source = JavascriptConversion::ToString(
+                JavascriptOperators::GetProperty(thisObj, PropertyIds::source, scriptContext),
+                scriptContext);
+            JavascriptString* flags = JavascriptConversion::ToString(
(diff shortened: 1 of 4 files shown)


goyakin:

goyakin:
Benchmarks show no difference in perf.

goyakin:
Ping...

lib/Runtime/Library/JavascriptRegularExpression.cpp
```diff
@@ -671,11 +671,37 @@ namespace Js
...
+            builder->Append(flags);
+            return builder;
+        }
+        else
+        {
+            JavascriptRegExp* obj = GetJavascriptRegExp(args, varName, scriptContext);
+
+            bool sourceOnly = false;
+            bool useFlagsProperty = scriptConfig->IsES6RegExPrototypePropertiesEnabled();
+            return obj->ToString(sourceOnly, useFlagsProperty);
```
akroshg: Is this the only place the ::ToString gets called? wondering if we can remove useFlagProperty altogether. In the ToString itself can check of IsES6RegExPrototypePropertiesEnabled() and do the work.

test/es6/regex-prototype-properties.js
```diff
@@ -79,5 +79,20 @@ tests = tests.concat(nonGenericPropertyNames.map(function (name) {
...
+tests = tests.concat([
+    {
+        name: "RegExp.prototype.toString should be generic",
+        body: function () {
+            var re = {
+                source: "source",
+                flags: "gi"
+            };
+
+            var string = RegExp.prototype.toString.call(re);
```
akroshg: What if the source and flags are not there, say RegExp.prototype.toString.call({}) ?

akroshg:
code wise looks good to me.

lib/Runtime/Library/JavascriptRegularExpression.cpp
goyakin: This isn't the only place. It is also used to generate the string for the `source` property. However, now that `EntryToString` handles the `flags` property itself, the part of `ToString` handling it is unreachable. I've removed that section in commit 154d829011cd3290c7ba15a60bad3df250d5936b.

test/es6/regex-prototype-properties.js
goyakin: We should use the string `undefined` for them in that case. I didn't add a test specifically for this, but added some verifying the function when the properties are `undefined` in addition to non-string properties and when they are overridden on a subclass (ee246cce09c70b33b195c631050419b2785d8874).

akroshg:
Looks good to me.

PR #591, opened 2016-03-16
Default parameters pid ref stack issue

aneeshdk:
Default parameters pid ref stack issue

This is a fix for an assert happening when default param is referencing a
variable from outer scope and there is a function definition with the same
name in a block in the function body.

When we have a function definition in a block we add an additional var
declaration node for the same in the function body for backward
compatibility. When param scope also has a reference to the same name
we were adding this reference to the body scope in the wrong order.

### DIFF
--- a/lib/Parser/Hash.h
+++ b/lib/Parser/Hash.h
@@ -236,6 +236,7 @@ public:
 
         // Search for the corresponding PidRef, or the position to insert the new PidRef.
         PidRefStack *ref = m_pidRefStack;
+        PidRefStack *prevRef = nullptr;
         while (1)
         {
             // We may already have a ref for this scopeId.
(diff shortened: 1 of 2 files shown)


aneeshdk:
@tcare @akroshg @pleath Can you please take a look?

lib/Parser/Hash.h
```diff
@@ -266,8 +267,16 @@ struct Ident
...
-                    newRef->prev = m_pidRefStack;
-                    m_pidRefStack = newRef;
+                    if (prevRef)
+                    {
+                        newRef->prev = prevRef->prev;
+                        prevRef->prev = newRef;
+                    }
+                    else
+                    {
+                        newRef->prev = m_pidRefStack;
```
tcare: We were talking offline about how we weren't sure if this case would come about with the above condition. Do you have an example that breaks our assumption?

aneeshdk: Yes. if we have some code like function (a = x) { var x = 1; } then the prevRef will be null. So we can't get rid of the else part.

aneeshdk: I will add this as a comment there.

lib/Parser/Hash.h
```diff
@@ -266,8 +267,16 @@ struct Ident
                     // Without parameter scope, we would have just pushed the ref instead of inserting.
                     // We effectively had a false positive match (a parameter scope ref with no sym)
                     // so we need to push the Pid rather than inserting.
-                    newRef->prev = m_pidRefStack;
-                    m_pidRefStack = newRef;
+                    if (prevRef)
```
akroshg: Nit : change this to prevRef != nullptr

lib/Parser/Hash.h
```diff
@@ -266,8 +267,16 @@ struct Ident
                     // Without parameter scope, we would have just pushed the ref instead of inserting.
                     // We effectively had a false positive match (a parameter scope ref with no sym)
                     // so we need to push the Pid rather than inserting.
```
akroshg: I believe this comment required some update. As we are inserting in between two positions.

akroshg:
Looks good to me.

tcare:
Same here.

lib/Parser/Hash.h
aneeshdk: Done.

lib/Parser/Hash.h
aneeshdk: Done.

PR #592, opened 2016-03-16
Have RegExp.prototype.test delegate to "exec"

### LINKED ISSUE #579: RegExp.prototype.test should delegate to "exec"
Kangax test case:

```JS
// RegExp.prototype.test -> RegExpExec -> Get -> [[Get]]
var get = [];
var p = new Proxy({ exec: function() { return null; } }, { get: function(o, k) { get.push(k); return o[k]; }});
RegExp.prototype.test.call(p);
return get + '' === "exec";
```

goyakin:
Fixes #579

Benchmarks show no perf impact.


goyakin:

akroshg:
Looks good to me.

PR #593, opened 2016-03-16
remove special handling for atan2 from Js::Math::atan2

MikeHolman:
Using different methods for calculating atan2 was causing differences in result between JIT and interpreter for some inputs. Because it was unclear that the interpreter special handling of certain values is necessary, I removed the special cases and simply call the C++ library functions (as we do with all the other Math builtins).

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/593)

### DIFF
--- a/lib/Runtime/Library/MathLibrary.cpp
+++ b/lib/Runtime/Library/MathLibrary.cpp
@@ -294,149 +294,23 @@ namespace Js
 
     double Math::Atan2( double x, double y )
     {
-        double result = 0;
-        enum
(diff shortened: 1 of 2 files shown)


MikeHolman:
hmm... maybe this isn't correct fix. looks like this now leads to difference between x86 and x64 (where x64 still gives same result as before).

MikeHolman:
@abchatra, can you please review?

abchatra:
:+1:

PR #594, opened 2016-03-16
Fix assertion in EntryPointInfo expiring

leirocks:
Previously when fixing a bug that when codegen failed, entryPointInfo is cleanedup with wrong(previously jitted code/thunk) address and causes the jit code freed prematurely.
I added the assert to capture we should not be expiring the entryPointInfo which has the address still installed on the functionType.
Here's another situation that the entryPointInfo on deferredPrototypeType (which is same as the default m_defaultEntryPointInfo) is already cleanedup due to some other reason (most likely hit the JitLimit, added Unknown to capture other cases). In this situation the Assertion can be wrong.


leirocks:
@curtisman @digitalinfinity can you review

curtisman:
LGTM

PR #595, opened 2016-03-16
OS:6523257: Enable CollectGarbage functionality for WWA and Webview.

dilijev:
During testing, I temporarily introduced some output into the function after it would have done nothing and returned if it were disabled.

The function now performs GC for WWA and Webview host types (2 and 3 respectively), and does nothing for HostType 1 (Browser). Default is HostType:1 (Browser). Applies to both internal test harness and hosting. ch.exe always acts as a Browser host.

As before, function is always present on the GlobalObject for legacy reasons (in all binaries and hosts), even if it is set to do nothing.

This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/595)

### DIFF
--- a/lib/Runtime/Library/JavascriptLibrary.cpp
+++ b/lib/Runtime/Library/JavascriptLibrary.cpp
@@ -1264,7 +1264,12 @@ namespace Js
         AddFunctionToLibraryObject(globalObject, PropertyIds::escape, &GlobalObject::EntryInfo::Escape, 1);
         AddFunctionToLibraryObject(globalObject, PropertyIds::unescape, &GlobalObject::EntryInfo::UnEscape, 1);
 
-        if (scriptContext->GetConfig()->SupportsCollectGarbage())
+        if (scriptContext->GetConfig()->SupportsCollectGarbage()
+#ifdef ENABLE_PROJECTION
+            || scriptContext->GetConfig()->GetHostType() == HostType::HostTypeApplication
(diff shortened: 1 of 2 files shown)


lib/Runtime/Library/JavascriptLibrary.cpp
```diff
@@ -1264,7 +1264,12 @@ namespace Js
         AddFunctionToLibraryObject(globalObject, PropertyIds::escape, &GlobalObject::EntryInfo::Escape, 1);
         AddFunctionToLibraryObject(globalObject, PropertyIds::unescape, &GlobalObject::EntryInfo::UnEscape, 1);
 
-        if (scriptContext->GetConfig()->SupportsCollectGarbage())
+        if (scriptContext->GetConfig()->SupportsCollectGarbage()
+#ifdef ENABLE_PROJECTION
+            || scriptContext->GetConfig()->GetHostType() == HostType::HostTypeApplication
+            || scriptContext->GetConfig()->GetHostType() == HostType::HostTypeWebview
+#endif
+            )
```
dilijev: SupportsCollectGarbage is hardcoded to true. This check is a bit of future proofing against changes to this value as long as we still want to support WWA and WebView host types even if we were to remove the API from the browser.

dilijev:
FYI @curtisman @digitalinfinity @Yongqu - please review?

curtisman:
LGTM

Issue #139, opened 2016-01-19
Incorrect .return() result on generator with yield* to another generator with try/finally


bterlson:
```js
function i(o) { print(JSON.stringify(o)) }
function* test() {
  try {
    yield 1;
  } finally {
    yield 2;
  }
  yield 3;
}

function* wrap() {
  yield* test();
  yield 4;
}

iter = wrap();
i(iter.next()); // { value: 1, done: false }
i(iter.return(42)); // { value: 2, done: false }
i(iter.next()); // { value: 4, done: false }
i(iter.next()); // { value: undefined, done: true }
```

Expected is above. However in our implementation we give { value: 2, done: true} for the 2nd result. I believe spec says .return(42) should invoke .return(42) on the inner generator, whose signal is handled by try/finally and yields { done: false, value: 2 }, which is then yielded as-is in [step 5.c.viii](https://tc39.github.io/#sec-generator-function-definitions-runtime-semantics-evaluation) of yield *.

ianwjhalliday:
Reviewing `EmitYieldStar` I see that we simply do not look for nor call `return` on the iterator operand.  We call `next` always, regardless of the way the generator was called into.  So same problem for `throw`.

This is related to #137 and #510

nicolo-ribaudo:
I fixed this issue while fixing #137 (pr: #597)

ianwjhalliday:
Fixed by #665

Issue #137, opened 2016-01-19
Incorrect .throw() result on generator with yield* to another generator with try/catch


bterlson:
Repro:

```js
function* test() {
  try {
    yield;
  } catch (e) {
    return;
  }
}

function* wrap() {
  yield* test();
  yield 3;
}

var iter = wrap();
iter.next();
print(JSON.stringify(iter.throw(Error("fail"))));
```

Expected per spec is `iterator result object { value: undefined, done: true }` as yield* has a return completion value when received is a throw completion. However, v8/sm do differently and https://github.com/tc39/ecma262/issues/293 proposes changing the expected result to `{value: 3, done: false }` by returning a normal completion for this case.

bterlson:
The v8/sm semantics won the day here, so the expected result is `{value: 3, done: false}`.

abchatra:
This is failing a kangax test case as well. 
```js
print((function()
{
function * generator() {
  yield * (function * () {
    try {
      yield 'foo';
    }
    catch(e) {
      return;
    }
  }());
  yield 'bar';
}
var iter = generator();
iter.next();
return iter['throw']().value === 'bar';
})());
```

nicolo-ribaudo:
Can I tackle this issue?

ianwjhalliday:
@nicolo-ribaudo the 1.2 milestone bugs need to be fixed asap.  I'm happy to let you take a stab at it.  Please start on it immediately and let me know within the next couple of days what your plan for the fix is.

Let me know if you do not have enough time to complete this bug fix within the next week or so.  I will want to start on it first thing next week if that is the case.

ianwjhalliday:
@nicolo-ribaudo also note that the due date on the 1.2 milestone is a rough estimate.  Lockdown on our bug fixing for the next release may occur sooner which is why this needs to be addressed asap.

nicolo-ribaudo:
@ianwjhalliday I'm working on this now. Is it ok to move the call to the iterator's `.next()` method from the bytecode to the `JavascriptOperators::OP_ResumeYield` function? It probably will be a little less efficient, but I don't know if there is a way to know in the bytecode how the generator has been resumed.

nicolo-ribaudo:
Should this code throw?
```js
function* gen() {
  yield* {
    [Symbol.iterator]() {
      return {
        next() { return { value: 3, done: false } },
        throw() { return 11; } // <-- 11 is not an object!!!
      }
    }
  };
};
var g = gen();
[...]

PR #597, opened 2016-03-17
Update the yield* implementation to match the spec

nicolo-ribaudo:
1. If the iterator's `.return()` method returns an object whose `done` property is `false`, do not terminate the yield*'s loop.
2. If the iterator's `.throw()` method doesn't throw, return the value it returned instead of throwing an exception

Fixes gh-139, fixes gh-137

### DIFF
--- a/lib/Runtime/Language/JavascriptOperators.cpp
+++ b/lib/Runtime/Language/JavascriptOperators.cpp
@@ -743,7 +743,7 @@ namespace Js
                 aLeft = JavascriptConversion::ToPrimitive(aLeft, JavascriptHint::HintNumber, scriptContext);
             }
             //BugFix: When @@ToPrimitive of an object is overridden with a function that returns null/undefined
-            //this helper will fall into a inescapable goto loop as the checks for null/undefined were outside of the path 
+            //this helper will fall into a inescapable goto loop as the checks for null/undefined were outside of the path
             return RelationalComparisonHelper(aLeft, aRight, scriptContext, leftFirst, undefinedAs);
         }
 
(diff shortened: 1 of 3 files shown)


ianwjhalliday:
@nicolo-ribaudo fantastic tests! Concise targeted and clear.

ianwjhalliday:
I've taken a look through and this looks reasonable at a glance.  I have some concerns but no time today to investigate further.  I'll look into this more tomorrow.

lib/Runtime/Language/JavascriptOperators.cpp
```diff
@@ -9386,83 +9386,112 @@ namespace Js
...
-                    // Ld_A R0 Rx
-                    // Br $exitFinallyAndReturn
-                    //
-                    // This would probably give better performance for the common case of calling next() on generators since we wouldn't
-                    // have to wrap the call to the generator code in a try catch.
-                    yieldData->exceptionObj->SetThrownObject(value);
+                    // Use ThrowExceptionObject() to get debugger support for breaking on throw
+                    JavascriptExceptionOperators::ThrowExceptionObject(yieldData->exceptionObj, yieldData->exceptionObj->GetScriptContext(), true);
+                }
+                else if (!isNext)
```
ianwjhalliday: Can you check this in the enclosing if condition above to make it clear earlier to the reader that this entire body only applies to throw and return cases.

lib/Runtime/Language/JavascriptOperators.cpp
```diff
@@ -9386,83 +9386,112 @@ namespace Js
...
                     {
-                        JavascriptError::ThrowTypeError(scriptContext, JSERR_NeedObject);
+                        RecyclableObject* method = RecyclableObject::FromVar(prop);
+                        Var args[] = { iterator, yieldData->data };
+                        CallInfo callInfo(CallFlags_Value, _countof(args));
+                        Var result = JavascriptFunction::CallFunction<true>(method, method->GetEntryPoint(), Arguments(callInfo, args));
+
+                        if (!JavascriptOperators::IsObject(result))
+                        {
+                            JavascriptError::ThrowTypeError(scriptContext, JSERR_NeedObject);
```
ianwjhalliday: I think we might want a new error message for this case.

Something like "yield* operand does not have a throw method".  That's awkward though, don't use that directly.

lib/Runtime/Language/JavascriptOperators.cpp
```diff
@@ -9386,83 +9386,112 @@ namespace Js
...
+                }
+                else if (!isNext)
+                {
+                    // Do not use ThrowExceptionObject for return() API exceptions since these exceptions are not real exceptions
+                    throw yieldData->exceptionObj;
                 }
             }
-            else if (!isReturn)
+
+            RecyclableObject* method = RecyclableObject::FromVar(prop);
```
ianwjhalliday: It is not safe to call `RecyclableObject::FromVar` on a random `Var`.  The `Var` could be a tagged int or float.  `RecyclableObject::Is` should be used first to check that it is indeed a `RecyclableObject`.  If it isn't that would be equivalent to the spec's `GetMethod` line that checks if the result is callable, so a TypeError should be thrown if it isn't.  There is a helper, `JavascriptConversion::IsCallable` that you could use to check for this and make the code read nicer than `RecyclableObject::Is`.

lib/Runtime/Language/JavascriptOperators.cpp
```diff
@@ -9386,83 +9386,112 @@ namespace Js
...
+            if (isThrow || isNext)
+            {
+                // 5.b.ii.2
+                // NOTE: Exceptions from the inner iterator throw method are propagated.
+                // Normal completions from an inner throw method are processed similarly to an inner next.
+                return result;
+            }
+            else
+            {
+                RecyclableObject* obj = RecyclableObject::FromVar(result);
```
ianwjhalliday: Another unsafe `FromVar`.

In this particular case there is also a missing `TypeError` throw when `result` is not an object, from the spec at 5.c.v.

This case and the above case should both get unit test coverage.

lib/Runtime/Language/JavascriptOperators.cpp
```diff
@@ -9386,83 +9386,112 @@ namespace Js
...
+            }
+
+            if (isThrow || isNext)
+            {
+                // 5.b.ii.2
+                // NOTE: Exceptions from the inner iterator throw method are propagated.
+                // Normal completions from an inner throw method are processed similarly to an inner next.
+                return result;
+            }
+            else
```
ianwjhalliday: style nit.  I prefer else blocks attached to if/if-elses that all return to be unwrapped and not in an else.

i.e. instead of
```C++
if (...)
{
    return;
}
else
{
    ...
}
```

do
```C++
if (...)
{
    return;
}

...
```

This makes it easier to see where control flow continues and terminates.

ianwjhalliday:
A few comments.  Looks good overall.

lib/Runtime/Language/JavascriptOperators.cpp
nicolo-ribaudo: > In this particular case there is also a missing `TypeError` throw when `result` is not an object, from the spec at 5.c.v.

It is thrown at line 9437

lib/Runtime/Language/JavascriptOperators.cpp
nicolo-ribaudo: I think it would be confusing. It is legal for the iterable to not have a throw method, but if there isn't it and there is the return method, it is an error if that method doesn't return an object.

ianwjhalliday: Oh oops, I misread this.  I was reading this as the throw corresponding to 5.b.iii.4.  Looks like that throw is missing?  There should be a new TypeError thrown in this case signifying the violation of the yield* protocol.

lib/Runtime/Language/JavascriptOperators.cpp
ianwjhalliday: You're right, my bad, I missed that.

ianwjhalliday:
Looking good other than the one missing TypeError for the yield* protocol violation.

I think this will break some internal debugging tests we have, so this will take some time to get merged once you've added the TypeError.  I am also busy with some higher pri stuff so bear with me.

lib/Runtime/Language/JavascriptOperators.cpp
nicolo-ribaudo: You are right, I just fixed it. Let me know when I can squash these commits.

test/es6/generators-functionality.js
```diff
@@ -1355,16 +1371,16 @@ var tests = [
             assert.areEqual({value: 2, done: true}, g.return(2), "As the return property is missing the yield* just returns as is");
             g = gf();
             assert.areEqual({value: 1, done: false}, g.next(), "Get the first yield value from the inner iterator");
-            assert.throws(function () { g.throw(new ExpectedException()); }, ExpectedException, "As the throw property is missing the yield* just throws as is");
+            assert.throws(function () { g.throw(new ExpectedException()); }, TypeError, "As the throw property is missing a TypeError is thrown");
```
ianwjhalliday: Sorry I just noticed that some of the assert.throws you've modified or added do not have the expected exception message string at the end as the (unfortunately optional) fourth argument.

Could you add those in.  Then feel free to squash.

I did a preliminary test run and indeed found that our debugger internal tests need to be updated before this is merged which will take me some time.  I'll get to that next week.

lib/Runtime/Language/JavascriptOperators.cpp
```diff
@@ -9395,7 +9395,7 @@ namespace Js
             PropertyId propertyId = isNext ? PropertyIds::next : isThrow ? PropertyIds::throw_ : PropertyIds::return_;
             Var prop = JavascriptOperators::GetProperty(iterator, propertyId, scriptContext);
 
-            if (!isNext && prop == iterator->GetLibrary()->GetUndefined())
+            if (!isNext && JavascriptOperators::IsUndefinedOrNull(prop))
```
nicolo-ribaudo: This is because `GetMethod` (which is used to get iterator's `.return`/`.throw`) returns `undefined` either if the property is `null` or if it is `undefined`

ianwjhalliday: Good catch!

test/es6/generators-functionality.js
```diff
@@ -1518,7 +1694,7 @@ var tests = [
             g1.return = function() { closed = true; return {done: true}; }
             g2 = gf2();
             g2.next();
-            assert.throws(function() { g2['throw'](new ExpectedException()) }, ExpectedException, "Throw is propagated back to the caller");
+            assert.throws(function() { g2['throw'](new ExpectedException()) }, TypeError, "As the throw property is missing a TypeError is thrown");
```
ianwjhalliday: This assert.throws is also missing the error message text

nicolo-ribaudo:
I added the missing error message in the test and squashed

ianwjhalliday:
Alright looking good.  Sorry I have to request one more thing before we merge this in.  At the end of last week we created a new branch for 1.2 fixes to allow master to open up for post-1.2 fixes.

This pull request now needs to be reopened against that branch, release/1.2.  Once you've created the new PR I'll go ahead and merge it in.

Thanks for your help and patience on this one, @nicolo-ribaudo, I know this has been longer than it normally would take.

nicolo-ribaudo:
No problem :wink:

PR #600, opened 2016-03-17
Remove the registering of scriptContexts those have inlineCaches

leirocks:
ThreadContext uses different list to keep tracking of those scriptContexts which have inlineCaches, when adding to this list it can OOM and then causes some of the inlineCaches not be able to cleared up.
Changing the design to not using this kind of registering mechanism, instead just use the flags on scriptContext to determine if there's inlineCaches need to be cleared or not. with this change there should be less work to do while allocating inlineCache, but longer list to walk while shutting down. usually we do not have many scriptContexts (my guess is in most web sites we have less than 100 scriptContexts) so it should not slow down the shutdown process.

### DIFF
--- a/lib/Runtime/Base/ScriptContext.cpp
+++ b/lib/Runtime/Base/ScriptContext.cpp
@@ -118,10 +118,8 @@ namespace Js
 #endif
         inlineCacheAllocator(_u("SC-InlineCache"), threadContext->GetPageAllocator(), Throw::OutOfMemory),
         isInstInlineCacheAllocator(_u("SC-IsInstInlineCache"), threadContext->GetPageAllocator(), Throw::OutOfMemory),
-        hasRegisteredInlineCache(false),
-        hasRegisteredIsInstInlineCache(false),
-        entryInScriptContextWithInlineCachesRegistry(nullptr),
-        entryInScriptContextWithIsInstInlineCachesRegistry(nullptr),
+        hasInlineCache(false),
+        hasIsInstInlineCache(false),
         registeredPrototypeChainEnsuredToHaveOnlyWritableDataPropertiesScriptContext(nullptr),
         cache(nullptr),
         bindRefChunkCurrent(nullptr),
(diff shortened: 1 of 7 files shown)


leirocks:

lib/Runtime/Base/ScriptContext.cpp
```diff
@@ -118,10 +118,8 @@ namespace Js
 #endif
         inlineCacheAllocator(_u("SC-InlineCache"), threadContext->GetPageAllocator(), Throw::OutOfMemory),
         isInstInlineCacheAllocator(_u("SC-IsInstInlineCache"), threadContext->GetPageAllocator(), Throw::OutOfMemory),
-        hasRegisteredInlineCache(false),
-        hasRegisteredIsInstInlineCache(false),
-        entryInScriptContextWithInlineCachesRegistry(nullptr),
-        entryInScriptContextWithIsInstInlineCachesRegistry(nullptr),
+        hasInlineCache(false),
```
curtisman: The condition doesn't mean we have inline cache, but have populated (or in use) inline cache.  
Maybe hasUsedInlineCache?

lib/Runtime/Base/ScriptContext.cpp
```diff
@@ -4122,10 +4070,11 @@ namespace Js
     {
         // Prevent reentrancy for the following work, which is not required to be done on every call to this function including
         // reentrant calls
-        if (this->isPerformingNonreentrantWork)
+        if (this->isPerformingNonreentrantWork || !this->hasInlineCache)
```
curtisman: Why is this added?

lib/Runtime/Language/InlineCache.cpp
```diff
@@ -26,8 +26,6 @@ namespace Js
         // We may, however, have a flags cache (setter) change to add property cache.
         Assert(typeWithoutProperty == nullptr || !IsProto());
 
-        requestContext->RegisterAsScriptContextWithInlineCaches();
```
curtisman: This would have added the script context into the list to be zero before.
Now it doesn't set the flag to indicate the script context needed to be zeroed?

lib/Runtime/Base/ScriptContext.cpp
leirocks: this was only called for those scriptContexts in the zero list, from ThreadContext::ClearScriptContextCaches(), adding this to keep the same logic as before. I didn't look into the code to understand what isPerformingNonreentrantWork is though, do you think we don't need hasInlineCache here?

lib/Runtime/Language/InlineCache.cpp
leirocks: no, it was not setting the flag before, it only check and allocate the item in the zero list, and save the item address back to scriptContext. the flag was set in ScriptContext::RegisterProtoInlineCache and ScriptContext::RegisterStoreFieldInlineCache, which is the call right after

lib/Runtime/Base/ScriptContext.cpp
leirocks: will change

lib/Runtime/Language/InlineCache.cpp
curtisman: Well.. that would be different then before in ThreadContext::ClearScriptContextCaches, where it would have cleared this cache before, but not now. Although it is just an heuristic to indicate whether a script context was used....?

It is also wrong that if threadContext->ClearInlineCaches() before, it would clear this script context, now it doesn't.

leirocks:
@curtisman can you review again?

curtisman:
LGTM

PR #551, opened 2016-03-12
Fix PreFAST errors in dev12 release builds.

dilijev:
Fixes #553 

This change is [Review on Reviewable](https://reviewable.io/reviews/microsoft/chakracore/551)

PR #603, opened 2016-03-18
Fix AsmJs link failure scenario

### LINKED ISSUE #560: Asm.js tests failing link-time validation crash
Observable only with ch.exe using Dev'12. 
```
E:\nagy\git\chakra5\core\test\AsmJs>..\..\Build\VcBuild\bin\x86_debug\ch.exe -forceAsmJsLinkFail -DumpOnCrash -WERExceptionSupport -ExtendedErrorStackForTestHost -maxInterpretCount:1 -maxSimpleJitRunCount:1 -bgjit- -dynamicprofilecache:profile.dpl.UnnamedTest160  -bgjit- -on:asmjs -testtrace:asmjs -simdjs -AsmJsStopOnError -maic:0 constTest.js
Successfully compiled asm.js code
Asm.js Runtime Error : Forcing link failure
asm.js linking failed.
2
1
FATAL ERROR: ch.exe failed due to exception code c0000005
```

1. When calling the AsmJsModule function, the call-logic invokes EnsureDynamicProfileInfo on the function body. This is before ProcessAsmJsModule gets called. This adds a profile to profileInfoList in the ScriptContext.

2. In ProcessAsmJsModule, link fails and we call ReparseAsmJsModule, which generates a new dynamicProfile and attach it to the function body, and prepends it to the ScriptContext profileInfoList.
Now we end up with two profileInfos for the same function, the last of which is the one gets updated, since it is attached to the functionBody.

3. On ScriptContext close, we copy all entries in the profileInfoList over to sourceDynamicProfileManager->dynamicProfileInfoMap. This copy doesn't check if a profilInfo already exists in the map, so we overwrite the valid profileInfo with invalid one, and we end up dumping empty profile Info. Causing memcpy_s to fail on src != NULL assertion.

I could fix this by having a check if the profileInfo already exists in the sourceDynamicProfileManager->dynamicProfileInfoMap upon copying, but not sure if that's the right thing to do. Ideally, we want to avoid recreating a profileInfo here:
```
    void JavascriptFunction::ReparseAsmJsModule(ScriptFunction** functionRef)
    {
        ParseableFunctionInfo* functionInfo = (*functionRef)->GetParseableFunctionInfo();

        Assert(functionInfo);
        Assert(functionInfo->HasBody());
        functionInfo->GetFunctionBody()->AddDeferParseAttribute();
        functionInfo->GetFunctionBody()->ResetEntryPoint();
        functionInfo->GetFunctionBody()->ResetInParams();

        FunctionBody * funcBody = functionInfo->Parse(functionRef);
```

Currently #549 disables SIMDLink tests to avoid this failure.

nmostafa:
- Disabled profiled interpretation for AsmJs modules functions. This was causing two profiles to be created if AsmJs linking fails. 
- Re-enabled SIMD.js linking tests, and one AsmJs test with -forceAsmJsLinkFail

Fixes #560. More detail on the bug there.


nmostafa:

test/AsmJs/uint.js
```diff
@@ -12,10 +12,9 @@ let m = function (stdlib, foreign, heap) {
   }
   return f;
 }({}, {}, asmHeap);
-print(m(4294967295));
-print(m(4294967295));
 
-m = function (stdlib, foreign, heap) {
+
+m2 = function (stdlib, foreign, heap) {
```
nmostafa: For some reason, the Compiler and execution messages got re-ordered after this change (but execution output remains correct). This happens consistently with rl.exe, not when run from command-line. The new output is:
```
Successfully compiled asm.js code
0
0
Successfully compiled asm.js code
1
1
```

I had to re-order the test to get the original baseline output:
```
Successfully compiled asm.js code
Successfully compiled asm.js code
0
0
1
1
```
@MikeHolman , any idea what might be going on here ? And if it could be an issue ?

nmostafa:
@MikeHolman , please review.
/cc @dilijev

test/AsmJs/uint.js
MikeHolman: yup, it is because of defer parsing. depending whether we defer or not the message will be in different place (this is why so many of the asm.js tests have some -off:deferparse or -forcedeferparse flag)

nmostafa: I thought so. Any idea why my change caused deferred parsing on the second module though ? Can the deferred parsing order be non-deterministic and sensitive to execution times (I am guessing because the module function has no profile setup overhead, it is faster for the interpreter to get to module execution code)  ?

nmostafa:

test/AsmJs/uint.js
MikeHolman: no, it should be deterministic. my guess is it's because m2 is now becoming a property on global object rather than a var. if you change to `let m2 = ...`, I suspect behavior will revert back.

MikeHolman:
lgtm

test/AsmJs/uint.js
```diff
@@ -23,5 +22,9 @@ m = function (stdlib, foreign, heap) {
...
 }({}, {}, asmHeap);
-print(m(4294967295));
-print(m(4294967295));
\ No newline at end of file
+
+print(m1(4294967295));
+print(m1(4294967295));
+
+print(m2(4294967295));
+print(m2(4294967295));
```
dilijev: nit: newline at end of file

dilijev:
LGTM. Sorry about the dev12 PreFAST failures; still waiting on #551

dilijev:
@nmostafa I've fixed the PreFAST issues. Now if you rebase and rerun the tests *everything* should pass :)

nmostafa:

lib/Runtime/Base/FunctionBody.cpp
```diff
@@ -7089,7 +7090,7 @@ namespace Js
     {
 #if ENABLE_PROFILE_INFO
         // Switch off profiling is asmJsFunction
-        if (this->GetIsAsmJsFunction())
+        if (this->GetIsAsmJsFunction() || this->GetAsmJsModuleInfo())
```
leirocks: can we use GetIsAsmjsMode() instead? otherwise it need a lock while accessing GetAsmJsModuleInfo() from background thread

leirocks: looks not, GetIsAsmjsMode is not for AsmJsModuleInfo. we have m_isAsmJsFunction, but don't have that for AsmJsModuleInfo

nmostafa: Is locking necessary here ? Even asmJsModuleInfo is set once by the AsmJs compiler, and remains read-only ?

leirocks: because now the pointer is saved in a dynamically expanding structure, so while background thread is accessing, the foreground thread might be expanding the structure, so it can get wrong value. 
so when GetIsAsmJsFunction() returns false, we can still have AsmJsModuleInfo on the function body, right? if so let me submit a PR to use lock for background accessing

nmostafa: Yes, Module functions won't return GetIsAsmJsFunction() == true.

Issue #390, opened 2016-02-25
Map.get bug with hash code calc


mybios:
test case
```javascript
var map = new Map();
map.set(1 , "value");
console.assert(map.get(1) , "ok");

var key = 1.1;
key -= 0.1;
// will assert because "key" is double , and 1 is int
console.assert(map.get(key) , "not ok");
```

dilijev:
Confirmed no throw from v8 in Canary (50) and Chrome (48).

ianwjhalliday:
Good catch.  Yep, a quick look shows that `SameValueComparerCommon::GetHashCode` incorrectly returns different hash values for tagged ints and equivalent untagged ints.  Probably also has this error with Int64 and UInt64 integers.

dilijev:
Tried another repro.

```js

(function() {

var _1 = 1;
var _1dot1 = 1.1;
var _1dot1minus1 = _1dot1 - 0.1;
if (_1 === _1dot1minus1) {
    print("1 === 1.1 - 0.1");
} else {
    print("0: not good");
}
    
var map = new Map();
map.set(1, "test value");

try {
    print(map.get(_1));
} catch (e) {
    print("1: not good");
}

var key = 1.1;
key -= 0.1;

if (map.get(key)) {
    print(map.get(key));
} else {
    print("2: not good");
}

})();
```

Replace print with console.log and it runs with no repro in Canary and Chrome.
Repros in latest Chakra by printing "2: not good"

ianwjhalliday:
This is my bug from years old code. I've assigned it to myself but if someone wants to grab it off me before I get to it I'm cool with that.  Probably will be a couple weeks at minimum before I can get to this.

PR #604, opened 2016-03-18
Fixes mismatched hashing of ints versus doubles

ianwjhalliday:
`SameValueComparerCommon::GetHashCode` was hashing tagged ints
differently from javascript numbers (doubles).  E.g. 1 would hash
differently from 1.0.  This made Map, Set, WeakMap, and WeakSet lookups
fail when the input key was double and the stored key was int, and vice
versa.

Fixed by hashing integers as doubles so long as those integers fit in a
double.  int64 values that do not fit in a double are hashed as int64
values.

Fixes gh-390

### DIFF
--- a/lib/Runtime/Library/SameValueComparer.h
+++ b/lib/Runtime/Library/SameValueComparer.h
@@ -31,39 +31,56 @@ namespace Js
             }
         }
 
+        static hash_t HashDouble(double d)
+        {
+            if (JavascriptNumber::IsNan(d))
+            {
+                return 0;
+            }
+
+            if (zero)
+            {
+                // SameValueZero treats -0 and +0 the same, so normalize to get same hash code
+                if (JavascriptNumber::IsNegZero(d))
+                {
+                    d = 0.0;
+                }
+            }
+
+            __int64 v = *(__int64*)&d;
+            return (uint)v ^ (uint)(v >> 32);
+        }
+
         static hash_t GetHashCode(Var i)
         {
             switch (JavascriptOperators::GetTypeId(i))
             {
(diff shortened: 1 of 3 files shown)


ianwjhalliday:
@abchatra @curtisman would either of you mind looking at this please?

lib/Runtime/Library/SameValueComparer.h
```diff
@@ -31,39 +31,56 @@ namespace Js
...
+            if (zero)
+            {
+                // SameValueZero treats -0 and +0 the same, so normalize to get same hash code
+                if (JavascriptNumber::IsNegZero(d))
+                {
+                    d = 0.0;
+                }
+            }
+
+            __int64 v = *(__int64*)&d;
```
curtisman: Don't we have problem here with different representation of a double of the same value here?

ianwjhalliday: I don't believe so.  Only NaN has multiple representations.  Number values are all uniquely represented, including the denormalized numbers.  Since we canonicalize NaN we don't have to worry about accidentally giving NaN multiple hash values.

Actually, I wonder then, we probably can also get rid of that special case for NaN above.

ianwjhalliday: We should also not be duplicating the raw double hash function algorithm.  Instead we should call `DefaultComparer<double>` to do that work.  I'll update the PR after I convince myself that the NaN special case is or is not necessary.

ianwjhalliday:
After speaking with Curtis offline we agreed that this hash function works fine for our needs.

Issue #538, opened 2016-03-11
Symbol.species does not work with Promise.prototype.then


abchatra:
See the following kangax test case

```js
print((function()
{
var promise      = new Promise(function(resolve){ resolve(42); });
var FakePromise1 = promise.constructor = function(exec){ exec(function(){}, function(){}); };
var FakePromise2 = function(exec){ exec(function(){}, function(){}); };

Object.defineProperty(FakePromise1, Symbol.species, {value: FakePromise2});

return promise.then(function(){}) instanceof FakePromise2;
})());

```
Expected: true
Actual: false

boingoing:
Fixed by #605

PR #605, opened 2016-03-18
Add support to Promise.prototype.then for Symbol.species

boingoing:
Add support to Promise.prototype.then for Symbol.species

Fixes a couple of issues with Promise and how it handled @@species.

First, Promise.prototype.then did not correctly use the @@species property to construct the promise stored in the PromiseCapability. See NewPromiseCapability (https://tc39.github.io/ecma262/#sec-newpromisecapability).

Next, we were looking at the @@species property in Promise.all and Promise.race when we are supposed to ignore it. Those API should use the 'this' argument (Promise in the nominal case) to construct the object they return.

Also cleaned up some stale comments, updated a few small details (like length property of anonymous functions used by Promise machinery), and fixed a bug where the executor function passed to the Promise constructor via Promise.all or Promise.race could not be called twice even if the first call passed undefined for arguments. This should be allowed.


boingoing:
@abchatra @suwc Care to take a look? This fixes #538

boingoing:
@akroshg Updated this PR to address 6127686, want to take a quick look over it?

test/es6/ES6Promise.js
```diff
@@ -91,10 +91,10 @@ var tests = [
     {
         name: "Promise constructor throwing behavior",
         body: function () {
-            assert.throws(function() { Promise.call(); }, TypeError, "Promise throws when not called as a new expression with no this parameter", "Promise: 'this' is not a Promise object");
-            assert.throws(function() { Promise.call(undefined); }, TypeError, "Promise throws when not called as a new expression if the this parameter is undefined", "Promise: 'this' is not a Promise object");
-            assert.throws(function() { Promise.call(null); }, TypeError, "Promise throws when not called as a new expression if the this parameter is null", "Promise: 'this' is not a Promise object");
-            assert.throws(function() { Promise.call({}); }, TypeError, "Promise throws when not called as a new expression if the this parameter is not a promise", "Promise: 'this' is not a Promise object");
+            assert.throws(function() { Promise.call(); }, TypeError, "Promise throws when not called as a new expression with no this parameter", "'new.target' is undefined");
```
akroshg: Nit : somehow I don't like the new error as well - it is not intuitive. shouldn't this be "Promise constructor to be called as 'new' or 'Promise constructor requires the promise object' ?

lib/Runtime/Library/JavascriptPromise.cpp
```diff
@@ -167,11 +165,6 @@ namespace Js
             JavascriptError::ThrowTypeError(scriptContext, JSERR_This_NeedObject, _u("Promise.all"));
         }
 
-        // 3. Let S be Get(C, @@species).
-        // 4. ReturnIfAbrupt(S).
-        // 5. If S is neither undefined nor null, let C be S.
-        constructor = JavascriptOperators::GetSpecies(RecyclableObject::FromVar(constructor), scriptContext);
-
         JavascriptLibrary* library = scriptContext->GetLibrary();
         Var undefinedVar = library->GetUndefined();
```
akroshg: Nit: you can get rid of this var as well ? it is just used once I think.

lib/Runtime/Library/JavascriptPromise.cpp
```diff
@@ -815,9 +782,18 @@ namespace Js
...
-        Assert(args.Info.Count > 1);
+        Var arg;
+
+        if (args.Info.Count > 1 && args[1] != nullptr)
+        {
+            arg = args[1];
+        }
+        else
+        {
+            arg = function->GetScriptContext()->GetLibrary()->GetUndefined();
```
akroshg: Should we still assert here?

akroshg:
Rest looks good to me.

test/es6/ES6Promise.js
boingoing: The meaning of the message is "Promise cannot be called without the 'new' keyword" but spec uses the case where new.target is undefined to indicate this. I don't think there is a way for new.target to actually be undefined if we do have the new flag, though. Even Reflect.construct(Promise,[],undefined) throws. I'll change the error message here.

lib/Runtime/Library/JavascriptPromise.cpp
boingoing: We used this var a few times before the refactor here. Nice catch.

lib/Runtime/Library/JavascriptPromise.cpp
boingoing: Spec says the Thrower function should return a completion with the value equal to the argument. So, I think it's correct for this function to be called with no argument and return undefined. I'll add assert that the arg is not nullptr here.

PR #606, opened 2016-03-18
remove libsodium dependency from cmake

obastemur:


jianchun:
Apologize for delay, verified and merging now.

Issue #610, opened 2016-03-19
[Test] [Daily] [DisableJIT] test\Bugs\bug934443.js failing after commit 353e7d9 with AV (c0000005)


aneeshdk:
test\Bugs\bug934443.js failing after commit 353e7d9 with AV (c0000005)

To repro, build the binary with
`jenkins\buildone.cmd x64 debug "/p:BuildJIT=false"`

Failing test is:
`Build\VcBuild.NoJIT\bin\x64_debug\ch.exe  -DumpOnCrash  -WERExceptionSupport -ExtendedErrorStackForTestHost -nonative  -force:fieldcopyprop -off:dynamicprofile test\Bugs\bug934443.js`

dilijev:
@abchatra Do you think this should be tagged as [Bug] and/or Milestone 1.2?

PR #611, opened 2016-03-19
Disabling the failing disable JIT test case for now

aneeshdk:
Disabling the test case mentioned in the issue https://github.com/Microsoft/ChakraCore/issues/610 temporarily to unblock the build.


dilijev:
LGTM.

PR #612, opened 2016-03-19
module bugfix

Yongqu:
fix debugger break in dynamic attach scenario
When debugger is dynamically attached, we reparse the tree and
regenerate the byte code. During the time we might need to use
the auxilary data. Disable the early release of auxilary data
for now and we'll have a better memory management when we have
full debugger support.
Also fixed a bug when the host pass in nullptr source indicating
download failure.

### DIFF
--- a/lib/Runtime/Language/SourceTextModuleRecord.cpp
+++ b/lib/Runtime/Language/SourceTextModuleRecord.cpp
@@ -91,38 +91,42 @@ namespace Js
         if (sourceText == nullptr)
         {
             Assert(sourceLength == 0);
+            hr = E_FAIL;
             JavascriptError *pError = scriptContext->GetLibrary()->CreateError();
-            JavascriptError::SetErrorMessageProperties(pError, E_FAIL, _u("empty module"), scriptContext);
+            JavascriptError::SetErrorMessageProperties(pError, hr, _u("empty module"), scriptContext);
+            *exceptionVar = pError;
         }
-        try
+        else
         {
-            AUTO_NESTED_HANDLED_EXCEPTION_TYPE((ExceptionType)(ExceptionType_OutOfMemory | ExceptionType_StackOverflow));
-            this->parser = (Parser*)AllocatorNew(ArenaAllocator, allocator, Parser, scriptContext);
-            this->srcInfo = {};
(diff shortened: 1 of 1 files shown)


Yongqu:

lib/Runtime/Language/SourceTextModuleRecord.cpp
```diff
@@ -91,38 +91,42 @@ namespace Js
         if (sourceText == nullptr)
         {
             Assert(sourceLength == 0);
+            hr = E_FAIL;
             JavascriptError *pError = scriptContext->GetLibrary()->CreateError();
-            JavascriptError::SetErrorMessageProperties(pError, E_FAIL, _u("empty module"), scriptContext);
+            JavascriptError::SetErrorMessageProperties(pError, hr, _u("empty module"), scriptContext);
```
boingoing: Should the message say "download failed" or be more descriptive?

lib/Runtime/Language/SourceTextModuleRecord.cpp
```diff
@@ -560,7 +565,6 @@ namespace Js
         }
         Js::AutoDynamicCodeReference dynamicFunctionReference(scriptContext);
         Assert(this == scriptContext->GetLibrary()->GetModuleRecord(srcInfo.moduleID));
-        uint sourceIndex = scriptContext->SaveSourceNoCopy(this->pSourceInfo, static_cast<charcount_t>(this->pSourceInfo->GetCchLength()), /*isCesu8*/ true);
         CompileScriptException se;
         this->rootFunction = scriptContext->GenerateRootFunction(parseTree, sourceIndex, this->parser, this->pSourceInfo->GetParseFlags(), &se, _u("module"));
```
boingoing: Where does sourceIndex come from here if not from SaveSourceNoCopy?

lib/Runtime/Language/SourceTextModuleRecord.cpp
Yongqu: will do.

lib/Runtime/Language/SourceTextModuleRecord.cpp
Yongqu: It's saved in the modulerecord already. I should have deleted this line :(

boingoing:
LGTM

Issue #265, opened 2016-02-04
Generator function yield* incorrectly unwraps and rewraps iterator results instead of passing them through


ianwjhalliday:
When `yield*` iterates over an iterable, it is supposed to pass the iterator result object on to the caller of the generator as it is, when its done property is falsy.  Looks like we are currently unwrapping the iterator result from the iterable, taking its value, and rewrapping it in a new iterator result object, that happens to have a `done: false` property added to it.

This was discovered via test262, we fail [`test/language/expressions/yield/star-iterable.js`](https://github.com/tc39/test262/blob/master/test/language/expressions/yield/star-iterable.js)

The spec states this at step 5.a.iv of _YieldExpression_: __yield *__ _AssignmentExpression_ Runtime Semantics here https://tc39.github.io/ecma262/#sec-generator-function-definitions-runtime-semantics-evaluation

Issue ported from VSO bug [5116152](https://microsoft.visualstudio.com/DefaultCollection/OS/_workitems?_a=edit&id=5116152)

ianwjhalliday:
@nicolo-ribaudo does your PR #597 fix this bug?

nicolo-ribaudo:
No, it doesn't

ianwjhalliday:
Kk.  Does it look like this will invalidate or require change to your fix for the other bugs or are does this look orthogonal to you?  I'm wondering if this bug should be incorporated or if it can be handled separately.

nicolo-ribaudo:
There might be a little merge conflict inside the `EmitYieldStar` function, but this issue is unrelated to the other one.

nicolo-ribaudo:
@ianwjhalliday I can work on this issue today/tomorrow.

ianwjhalliday:
@nicolo-ribaudo awesome! Go for it! Thanks for helping with ask these bugs. Greatly appreciated

PR #615, opened 2016-03-21
yield* shouldn't unwrap and rewrap iterator results

nicolo-ribaudo:
Commit message:
> This commit makes the bytecode generated by the EmitYield function return the IteratorResult object (instead of its `value` property) to the JavascriptGenerator::CallGenerator function.

> Fix gh-265

### DIFF
--- a/lib/Runtime/Library/JavascriptGenerator.cpp
+++ b/lib/Runtime/Library/JavascriptGenerator.cpp
@@ -67,25 +67,20 @@ namespace Js
             }
         }
 
-        if (this->IsCompleted())
-        {
-            result = library->CreateIteratorResultObject(result, library->GetTrue());
-        }
-        else
+        if (!this->IsCompleted())
         {
             int nextOffset = this->frame->GetReader()->GetCurrentOffset();
             int endOffset = this->frame->GetFunctionBody()->GetByteCode()->GetLength();
 
             if (nextOffset != endOffset - 1)
             {
-                result = library->CreateIteratorResultObjectValueFalse(result);
-            }
-            else
-            {
-                result = library->CreateIteratorResultObject(result, library->GetTrue());
-                this->SetState(GeneratorState::Completed);
+                // Yielded values are already wrapped in an IteratorResult object, so we don't need to wrap them.
+                return result;
             }
         }
+
+        result = library->CreateIteratorResultObject(result, library->GetTrue());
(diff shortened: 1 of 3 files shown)


nicolo-ribaudo:
This pr introduces an inconsistency between the `yield`'s and `return`'s behavior (the first one passes an IteratorResult to the `CallGenerator` function, the letter just its value). If you prefer to not have this inconsistency, I can close this pr and add a commit to #597 (otherwise, there will be some conflicts between these prs)

test/es6/generators-functionality.js
```diff
@@ -1378,7 +1378,7 @@ var tests = [
             var iteratorReturningWithoutValue = CreateIterable(() => { return {done: false}; }, () => { return {done: true}; }, () => { return {done: true}; });
             gf = function* () { yield* iteratorReturningWithoutValue; }
             g = gf();
-            assert.areEqual({value: undefined, done: false}, g.next(), "As the value property is missing undefined is returned");
+            g.next();
```
ianwjhalliday: Why did this change?

lib/Runtime/Library/JavascriptGenerator.cpp
```diff
@@ -67,26 +67,21 @@ namespace Js
...
-                result = library->CreateIteratorResultObject(result, library->GetTrue());
-                this->SetState(GeneratorState::Completed);
+                // Yielded values are already wrapped in an IteratorResult object, so we don't need to wrap them.
+                return result;
             }
         }
 
+        result = library->CreateIteratorResultObject(result, library->GetTrue());
+        this->SetState(GeneratorState::Completed);
+
```
ianwjhalliday: Love how this is simpler now

test/es6/generators-functionality.js
nicolo-ribaudo: Because it tested the wrong behavior fixed by this pr. I removed the assertion because the correct behavior is tested by the test I added.

ianwjhalliday: Right.  Because `g.next()` returns the `{done: false}` from the first arrow function passed to `CreateIterable`, correct?

ianwjhalliday:
Running tests on this one now.  I will merge it before #597 if it runs clean.

ianwjhalliday:
Thanks again @nicolo-ribaudo

PR #616, opened 2016-03-21
Rename JsParseScriptWithFlags to JsParseScriptWithAttributes

agarwal-sandeep:
The API name and the mask passed to API is inconsistent in naming
mask is called JsParseScriptAttributes and API have WithFlags.
Change it to match the mask name.

### DIFF
--- a/lib/Jsrt/ChakraCommon.h
+++ b/lib/Jsrt/ChakraCommon.h
@@ -379,7 +379,7 @@
     } JsMemoryEventType;
 
     /// <summary>
-    ///     Attribute mask for JsParseScriptWithFlags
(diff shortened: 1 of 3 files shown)


agarwal-sandeep:
@Yongqu @liminzhu please take a look

Yongqu:
Code looks good; don't know the potential impact for API change at this time.

agarwal-sandeep:
This have not made into any official release of ChakraCore as well as Windows yet. I think it should be ok to change it at this point.

liminzhu:
LGTM. I'm okay with the name change given there is no use case for this API yet without debugging APIs in master and it's not part of any release. Need to be careful in the future though.

PR #620, opened 2016-03-21
Fix bug with recursive inlining

rajatd:
We do recursive inlining only when all the callsites in the function are recursive. But, the checks weren't taking into account the presence of getters and setters which are not exposed as "call" sites. This leads to data about a getter/setter present in the inliner but not in the inlinee version of the recursive function. This causes us to assert when we try to get data for the getter/setter when trying to inline it in the inlinee.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/620)

### DIFF
--- a/lib/Backend/NativeCodeGenerator.cpp
+++ b/lib/Backend/NativeCodeGenerator.cpp
@@ -2289,7 +2289,7 @@ NativeCodeGenerator::GatherCodeGenData(
                     // Share the jitTime data if i) it is a recursive call, ii) jitTimeData is not from a polymorphic chain, and iii) all the call sites are recursive
                     if (functionBody == inlineeFunctionBody   // recursive call
                         && jitTimeData->GetNext() == nullptr        // not from a polymorphic  call site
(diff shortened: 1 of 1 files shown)


rajatd:
@abchatra could you review this, please?

lib/Backend/NativeCodeGenerator.cpp
```diff
@@ -2422,34 +2422,30 @@ NativeCodeGenerator::GatherCodeGenData(
...
-
+                if (inlineeRuntimeData->GetFunctionBody() != inlineeFunctionBody)
+                {
+                    //There are obscure cases where profileData has not yet seen the polymorphic LdFld but the inlineCache has the newer object from which getter is invoked.
+                    //In this case we don't want to inline that getter. Polymorphic bit will be set later correctly.
+                    //See WinBlue 54540
+                    continue;
+                }
 
-                    if (!isJitTimeDataComputed)
```
abchatra: Why is this removed?

rajatd: Because this bool will always be false for non-recursive inlining and I disabled recursive inlining for functions having accessor calls.

abchatra:
:+1:

PR #622, opened 2016-03-22
[OS5427497] Parser mistakes 'new.target' as in global function under -forceundodefer

suwc:
Global function is not allowed to have 'new.target' expression. Detection of being in the global function
depends on m_currentNodeNonLamdaFunc being nullptr for the global function and pointing to an actual
function pnode otherwise. During undo of deferred parsing, however, m_currentNodeNonLambdaFunc is not
properly set to a function pnode, causing parser to mistake 'new.target' as if it is in the global function.
Fix by assigning m_currentNodeNonLambdaFunc to pnodeFnc.

### DIFF
--- a/lib/Parser/Parse.cpp
+++ b/lib/Parser/Parse.cpp
@@ -10473,7 +10473,13 @@ void Parser::FinishDeferredFunction(ParseNodePtr pnodeScopeList)
             pnodeFnc->sxFnc.pnodeVars = nullptr;
             m_ppnodeVar = &pnodeFnc->sxFnc.pnodeVars;
 
+            Assert(m_currentNodeNonLambdaFunc == nullptr);
+            m_currentNodeNonLambdaFunc = pnodeFnc;
+
             this->FinishFncNode(pnodeFnc);
+
+            Assert(pnodeFnc == m_currentNodeNonLambdaFunc);
+            m_currentNodeNonLambdaFunc = nullptr;
 
             m_ppnodeExprScope = ppnodeExprScopeSave;
 
--- a/test/es6/rlexe.xml
+++ b/test/es6/rlexe.xml
@@ -1059,6 +1059,13 @@
       <tags>BugFix</tags>
     </default>
 </test>
+<test>
+    <default>
+      <files>ES6NewTarget_bugfixes.js</files>
+      <compile-flags>-ES6Classes -forceundodefer -args summary -endargs</compile-flags>
+      <tags>BugFix</tags>
+    </default>
+</test>
 
 <test>
     <default>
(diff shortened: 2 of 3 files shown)


suwc:
@akroshg @boingoing @abchatra could you take a look pls?

test/es6/rlexe.xml
```diff
@@ -1059,6 +1059,13 @@
       <tags>BugFix</tags>
     </default>
 </test>
+<test>
+    <default>
+      <files>ES6NewTarget_bugfixes.js</files>
+      <compile-flags>-ES6Classes -forceundodefer -args summary -endargs</compile-flags>
```
akroshg: forceundodefer works for top level global function - so the test you have implemented in es6newtarget_bugfixes.js is not exercising the scenario.

lib/Parser/Parse.cpp
```diff
@@ -10473,8 +10473,16 @@ void Parser::FinishDeferredFunction(ParseNodePtr pnodeScopeList)
             pnodeFnc->sxFnc.pnodeVars = nullptr;
             m_ppnodeVar = &pnodeFnc->sxFnc.pnodeVars;
 
+            ParseNodePtr pnodeFncSaveNonLambda = nullptr;
+
+            pnodeFncSaveNonLambda = m_currentNodeNonLambdaFunc;
```
akroshg: m0currentNodeNonLamdaFunc should be null here - should you put the null assert here?

akroshg: combine both 10476 and 10478 together.

akroshg: On a side note : we don't need to have `GetCurrentNonLamdaFunctionNode `at all. You can always determine that the current function (`funcInfo`) is global or not. We should get rid of this function and underlying variables.

suwc: @akroshg funcInfo at that point is not available (null).

test/es6/rlexe.xml
suwc: The function that has `new.target` is defined in the global function (`tests[2]["body"]`).
Without the change in `parse.cpp`, this test fails with `-forceundodefer`

lib/Parser/Parse.cpp
akroshg: `GetCurrentFunctionNode`() should not be null. In any case there are better ways to check if the current function you are parsing is the global function.

test/es6/rlexe.xml
akroshg: ah - good to know. I thought entire test is wrapped with anonymous function.

suwc:
@akroshg updated in latest rev.

lib/Parser/Parse.cpp
suwc: `GetCurrentFunctionNode()` returns the lambda function node while inside a lambda. This is probably why `GetCurrentNonLambdaFunctionNode()` was created in the first place.

akroshg:
Looks good to me.

tcare:
@dotnet-bot test this please

Issue #618, opened 2016-03-21
string.bold() not available for WWA (ES3 non standard). Now added back on ES6


joalmeid:
string.bold() is non-standard es3 extension methods and hence not available in WWA.

ES6 adds them back to spec as part of Annex B.

dilijev:
@abchatra did #623 completely resolve this issue?

abchatra:
Fixed by #623

PR #623, opened 2016-03-22
Enable es3 extension methods for WWA

abchatra:
See github issue #618


tcare:
Looks good.

akroshg:
So we are not getting rid of `SupportES3Entensions `altogether ?

Yongqu:
annex b only have 2 String.prototype methods: substr and anchor.
What's the backward compat impact for existing wwa apps?

abchatra:
@akroshg SupportES3Entensions still guards getVarDate which is not a standard. 
@Yongqu This is a known bug with existing WWA App which fails due to non-availability of  these string built-in. This issue\pull request is to address the same. I agree, this has a potential to break existing apps, though chances are remote. Is there a way to test the compat?

bterlson:
It is not worth testing compat for these. These exist on the web everywhere, so we're talking about some code breaking that exists in WWA only that is polyfilling these methods but in a way that isn't web compatible. I think this is fine to go in as-is.

PR #624, opened 2016-03-22
Setter is throwing SyntaxError when the parameter is default parameter

aneeshdk:
Setter is throwing SyntaxError when the parameter is default parameter

We were wrongly throwing SyntaxError when setter has only one default
parameter.


aneeshdk:

akroshg:
Looks good to me.

Issue #619, opened 2016-03-21
Discuss JsRTDebugging APIs


agarwal-sandeep:
I have implemented significant part of new APIs to debug ChakraCore and flushed most of bugs using debugger implemented in ch.

Before opening the PR request for merging code from JsRTDebugging to master it will be good to discuss the APIs name, signature and arguments. 

Please take a look at https://github.com/Microsoft/ChakraCore/blob/JsRTDebugging/lib/Jsrt/ChakraDebug.h and provide any feedback you have.

agarwal-sandeep:
@akroshg @liminzhu @yongqu please provide your feedback

dilijev:
@agarwal-sandeep Did you mean the contents here? https://github.com/Microsoft/ChakraCore/blob/JsRTDebugging/lib/Jsrt/ChakraDebug.h

agarwal-sandeep:
Yes, thanks. Fixed

akroshg:
@agarwal-sandeep how do I resume as continue?

agarwal-sandeep:
Execution resumes after call to JsDiagDebugEventCallback return. Host needs to call JsDiagSetStepType before returning from callback.

akroshg:
line 47/51 change 'Indicates break ' to 'Indicates a break'

akroshg:
>> JsDiagDebugEventCallback 
Please embed the description about that.

akroshg:
line 86 : please reword the description : eg. Request the runtime to break on next javascript statement.

akroshg:
line 98 : what do you mean by 'active' breakpoints ?

akroshg:
line 100, change breakPoints to breakpoints.

akroshg:
line 128 : the structure of the breakpoint is different from defined at line 104. that is not intentional, right?

akroshg:
line 274 : do you get handle as well?

akroshg:
line 282 : Shouldn't this be called JsDiagGetScript ?

akroshg:
line 309 : change func to function

akroshg:
line 296 : why do you need filename when we have scriptId ?

akroshg:
line 299 : change the names -> firstStatementLine/Column.
Also this is relative to the current function or absolute ? as the line is 1 above but the stmtStartLine is 0.

akroshg:
[...]

PR #625, opened 2016-03-22
Corrected spelling "but" to "put" in XML doc

ninlar:
Changed "but" to "put" in the XML doc on JsSetBreakpoint

### DIFF
--- a/lib/Jsrt/ChakraDebug.h
+++ b/lib/Jsrt/ChakraDebug.h
@@ -118,7 +118,7 @@
     /// <summary>
     ///     Sets breakpoint in the specified script at a location
     /// </summary>
-    /// <param name="scriptId">Id of script from JsDiagGetScripts or JsDiagGetSource to but breakpoint</param>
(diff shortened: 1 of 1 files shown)


dilijev:
Hi @ninlar, Thanks for pointing this out. We're discussing further changes to this document here: #619, so I raised this issue. I think it will probably be easiest to tackle this as part of that change as well, but I'll leave that up to @agarwal-sandeep in case he wants to merge this now.

Before we can accept this change, the Microsoft copyright header at the top of the file also needs to be updated. I guess somehow it was missed before we added the Jenkins check. The other various warnings will be fixed if you rebase this change on top of the latest master.

agarwal-sandeep:
Thanks @ninlar for pointing this out. Based on discussion in #619 I have made other changes to ChakraDebug.h, If you don't mind it will be easier to make this change instead of merging?

agarwal-sandeep:
I have made the change and pushed eb31241f59877629b7acbff33708026136c872ec

PR #627, opened 2016-03-22
Fix for background call into AuxPtrs to get AsmJsModuleInfo

leirocks:
PR #603 introduced a new background call to get AsmJsModuleInfo, this part need a lock

### DIFF
--- a/lib/Runtime/Base/FunctionBody.cpp
+++ b/lib/Runtime/Base/FunctionBody.cpp
@@ -7079,9 +7079,20 @@ namespace Js
             !IsGenerator(); // Generator JIT requires bailout which SimpleJit cannot do since it skips GlobOpt
     }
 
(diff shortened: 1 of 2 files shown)


leirocks:
@nmostafa @curtisman can you review

lib/Runtime/Base/FunctionBody.cpp
```diff
@@ -7103,6 +7114,23 @@ namespace Js
 #endif
     }
 
+    bool FunctionBody::DoInterpreterProfileWithLock() const
+    {
```
nmostafa: Shouldn't calls to DoInterpreterProfile to call this instead ?

leirocks: only for background thread. do you think there's other instance that background jit will call into DoInterpreterProfile ?

curtisman:
LGTM

PR #629, opened 2016-03-23
Initialize prototype type's defer initialize type handler with the prototype variant.

curtisman:
Avoid the need to switch to the prototype variant when the type that we are create is going to be a prototype anyways.
Reduce binary size and script context initialization costs

### DIFF
--- a/lib/Runtime/Library/JavascriptLibrary.h
+++ b/lib/Runtime/Library/JavascriptLibrary.h
@@ -605,7 +605,6 @@ namespace Js
         JavascriptFunction* GetStringConstructor() const {return stringConstructor; }
         JavascriptFunction* GetArrayBufferConstructor() const {return arrayBufferConstructor; }
         JavascriptFunction* GetErrorConstructor() const { return errorConstructor; }
-        JavascriptFunction* GetTypedArrayConstructor() const { return typedArrayConstructor; }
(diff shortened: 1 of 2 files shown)


lib/Runtime/Library/JavascriptLibrary.h
```diff
@@ -605,7 +605,6 @@ namespace Js
         JavascriptFunction* GetStringConstructor() const {return stringConstructor; }
         JavascriptFunction* GetArrayBufferConstructor() const {return arrayBufferConstructor; }
         JavascriptFunction* GetErrorConstructor() const { return errorConstructor; }
-        JavascriptFunction* GetTypedArrayConstructor() const { return typedArrayConstructor; }
```
ianwjhalliday: Was this related to binary size or initialization cost?

ianwjhalliday:
:+1:

lib/Runtime/Library/JavascriptLibrary.h
curtisman: Not this line. This is duplicated from JavascriptLibraryBase.h
Yes for the rest.  Noticed it when I was looking at the profile the other day.  This reduce the binary size by ~50K in x86 and ~80K in amd64 (release build), because all these templated DeferTypeHandler code for the non-prototype variant is gone.

abchatra:
:+1:

PR #630, opened 2016-03-23
xplat: build more lib/Runtime/Base files

jianchun:
Added more remaining files to build (Debug.cpp, ScriptContext.cpp,
StackProber.cpp, ThreadContext.cpp, Utf8SourceInfo.cpp).

The rest probably need new implementations for xplat (
DelayLoadLibrary.cpp, Entropy.cpp, HiResTimer.cpp,
WindowsFoundationAdapter.cpp, WindowsGlobalizationAdapter.cpp).

Added a few script debugging dependencies and left some script
debugging/profiling code outside of #ifdef to reduce code noise (otherwise
too many places to #ifdef).


lib/Common/CommonPal.h
```diff
@@ -119,6 +119,28 @@ inline void DebugBreak()
 // activscp.h
 #define SCRIPT_E_RECORDED                _HRESULT_TYPEDEF_(0x86664004L)
 
+
+typedef
+enum tagBREAKPOINT_STATE
```
digitalinfinity: Confused- why is this needed? Are we planning on supporting IDebugDocument on Linux? Shouldn't the code that depends on this be just ifdefed out?

lib/Runtime/Base/ThreadContext.cpp
```diff
@@ -1414,7 +1418,7 @@ ThreadContext::SetForceOneIdleCollection()
 BOOLEAN
 ThreadContext::IsOnStack(void const *ptr)
 {
-
+#ifdef _WIN32
```
digitalinfinity: probably should be ifdef _MSC_VER here, not _WIN32?

digitalinfinity:
LGTM otherwise- thanks!

lib/Common/CommonPal.h
jianchun: Script debugging/profiling are deeply embedded in many places (all the entryPoint Asserts, references from .asm...). To completely ifdef out will incur lots of code noise. I asked Sandeep and was told the current script debugging code will largely be reused for future JSRT debugging. That gave me less incentive to cleanly ifdef out all the debugging/profiling code. These enum types/constants are used in internal debugger logic. By adding them here the related code builds. We can remove them when future debugging code is rewritten to use new JSRT debugging types/constants.

lib/Runtime/Base/ThreadContext.cpp
jianchun: Windows code either reading ```FS``` or calling ```GetCurrentThreadStackLimits```. Those are platform behavior, not VC support. On Linux we call ```GetCurrentThreadStackBounds``` which is currently declared in CommonPal.h under ```!_WIN32```. Looks _WIN32 is more appropriate here?

jianchun: Discussed offline, @digitalinfinity meant __readfsdwords/etc. are VC intrinsic functions and should be under _MSC_VER. Will add.

PR #631, opened 2016-03-23
Temporary memop fix for breaking sites

pleath:
Removal of some overly conservative code in mem op preparation exposed a bug in induction variable analysis. We can select the wrong value as the upper bound of a loop, which can lead us to perform a memcopy with the wrong number of elements. On some popular sites this caused massive array growth, with hangs and memory spikes. Restoring the deleted code as a short-term fix while investigation of the deeper issue continues.


pleath:
@Cellule

Cellule:
LGTM

PR #632, opened 2016-03-23
xplat: build lib/Runtime/ByteCode

jianchun:
Moved ```DefaultComparer<ByteBuffer*>``` template specialization to global
scope as ```DefaultComparer``` was declared in global scope.

Added a (default) parameter to ```EncodeT()``` call as otherwise clang
matches it to a different function and failes.

Others are simple changes. All "ByteCode" files build successfully.


digitalinfinity:
LGTM

PR #634, opened 2016-03-23
Function with rest is not marked as HasNonSimpleParameter when deferred

aneeshdk:
Function with rest is not marked as HasNonSimpleParameter when deferred

When function has rest parameter we were marking it as
HasNonSimpleParameter only in non-defer parse mode. So if strict mode is
declared inside the function the Syntax Error is not thrown in deferred
mode.


aneeshdk:

akroshg:
Looks good to me.

tcare:
Looks Good

PR #635, opened 2016-03-23
TypedArray construction from array with user-defined %ArrayIteratorPrototype%.next

suwc:
TypedArray constructor uses the fast path for array on an array even if %ArrayIteratorPrototype%.next
is changed by user code. Correct behavior is to treat it as an iterable object and invoke
TypedArray(object).
Fix by adding extra conditions (essentially testing if %ArrayIteratorPrototype%.next is the built-in version)
under which TypedArray(object) will be used.
Tweak unit test to cover this case.

### DIFF
--- a/lib/Runtime/Library/TypedArray.cpp
+++ b/lib/Runtime/Library/TypedArray.cpp
@@ -94,6 +94,15 @@ namespace Js
         return newArr;
     }
 
+    bool TypedArrayBase::ArrayIteratorPrototypeHasUserDefinedNext(ScriptContext *scriptContext)
+    {
+        Var arrayIteratorPrototypeNext = nullptr;
+        ImplicitCallFlags flags = scriptContext->GetThreadContext()->TryWithDisabledImplicitCall(
+            [&] () { arrayIteratorPrototypeNext = JavascriptOperators::GetProperty(scriptContext->GetLibrary()->GetArrayIteratorPrototype(), PropertyIds::next, scriptContext); });
+
+        return (flags != ImplicitCall_None) || arrayIteratorPrototypeNext != scriptContext->GetLibrary()->GetArrayIteratorPrototypeBuiltinNextFunction();
+    }
+
     Var TypedArrayBase::CreateNewInstance(Arguments& args, ScriptContext* scriptContext, uint32 elementSize, PFNCreateTypedArray pfnCreateTypedArray)
     {
         uint32 byteLength = 0;
@@ -147,7 +156,7 @@ namespace Js
 
                     if (!JavascriptOperators::IsUndefinedObject(iterator) &&
                         (iterator != scriptContext->GetLibrary()->GetArrayPrototypeValuesFunction() ||
-                        !JavascriptArray::Is(firstArgument)))
+                            !JavascriptArray::Is(firstArgument) || ArrayIteratorPrototypeHasUserDefinedNext(scriptContext) ))
                     {
                         return CreateNewInstanceFromIterableObj(RecyclableObject::FromVar(firstArgument), scriptContext, elementSize, pfnCreateTypedArray);
                     }
--- a/test/typedarray/TypedArrayBuiltins.js
+++ b/test/typedarray/TypedArrayBuiltins.js
@@ -64,6 +64,21 @@ var tests = [
                 };
             }
 
+            function getIterableObjNextDesc (array)
+            {
+                return  {
+                    get: function () {
+                        array.shift(); // side effect
+                        return function () {
+                            return {
+                                value: array.shift(),
+                                done:  array.length == 0
+                            };
+                        };
+                    }
+                };
+            }
+
             var TypedArray = [
                 'Int8Array',
                 'Uint8Array',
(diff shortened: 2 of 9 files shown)


suwc:
@evilpie @ianwjhalliday @abchatra could you take a look pls?

lib/Runtime/Library/TypedArray.cpp
```diff
@@ -144,10 +144,13 @@ namespace Js
                 if (JavascriptOperators::IsObject(firstArgument))
                 {
                     Var iterator = JavascriptOperators::GetProperty(RecyclableObject::FromVar(firstArgument), PropertyIds::_symbolIterator, scriptContext);
+                    Var arrayIteratorPrototypeNext = nullptr;
 
                     if (!JavascriptOperators::IsUndefinedObject(iterator) &&
                         (iterator != scriptContext->GetLibrary()->GetArrayPrototypeValuesFunction() ||
-                        !JavascriptArray::Is(firstArgument)))
+                            !JavascriptArray::Is(firstArgument) ||
+                            !RuntimeFunction::Is(arrayIteratorPrototypeNext = JavascriptOperators::GetProperty(scriptContext->GetLibrary()->GetArrayIteratorPrototype(), PropertyIds::next, scriptContext)) ||
```
ianwjhalliday: `JavascriptOperators::GetProperty` can execute a getter or proxy trap.  It can't be used here because of that unfortunately.

I don't know off the top of my head if we have an API to do a get that will get a property value without executing a trap nor getter.  That is what we need here though.

E.g. this code will behave badly:

```javascript
var iter = [][Symbol.iterator]();
var count = 0;
Object.defineProperty(iter.__proto__, 'next', {
    get: function () {
        count++;
        return function () {
            return { done: true };
        };
    }
});

var i32a = new Int32Array([1, 2, 3]);

for (let i = 0; i < i32a.length; i += 1) {
    print(i32a[i]);
}
print(count);
```

`count` will be one too large.

lib/Runtime/Library/TypedArray.cpp
```diff
@@ -144,10 +144,13 @@ namespace Js
...
+                            &JavascriptArrayIterator::EntryInfo::Next != RuntimeFunction::FromVar(arrayIteratorPrototypeNext)->GetFunctionInfo()))
```
ianwjhalliday: I'd rather we compare the `JavascriptFunction` objects instead, getting the array iterator next object from the library for the comparison.

test/typedarray/TypedArrayBuiltins.js
```diff
@@ -92,28 +93,49 @@ var tests = [
...
+
+            var overrideBuiltinArrayPrototypeIterator = function() {
+                Array.prototype[Symbol.iterator] = getIterableObj([99,0])[Symbol.iterator];
+            };
+            testAllTypedArrayConstructorsWithIterableArray(overrideBuiltinArrayPrototypeIterator);
             Object.defineProperty(Array.prototype, Symbol.iterator, builtinArrayPrototypeIteratorDesc);
+
+            // change built-in array iterator's next function
+            var arrayIterator = [][Symbol.iterator]();
+            var builtinArrayPrototypeIteratorNext = arrayIterator.__proto__.next;
```
ianwjhalliday: Don't use `__proto__`.  It is deprecated.

Use these APIs instead:
```javascript
Object.getPrototypeOf(obj)
Object.setPrototypeOf(obj, proto)
```

lib/Runtime/Library/TypedArray.cpp
tcare: We have the same problem with spread. Let's chat today about a solution.

suwc: Good point.
I can add a flag in JavascriptLibrary called `hasUserDefinedArrayiteratorPrototypeNext` to get around having to call `GetProperty()`.

test/typedarray/TypedArrayBuiltins.js
suwc: Updated.

lib/Runtime/Library/TypedArray.cpp
suwc: Comparison is removed and replaced with flag checking.

lib/Runtime/Library/TypedArray.cpp
suwc: Updated in latest rev.

lib/Runtime/Language/JavascriptOperators.cpp
```diff
@@ -2307,6 +2307,11 @@ namespace Js
 
     BOOL JavascriptOperators::SetProperty(Var receiver, RecyclableObject* object, PropertyId propertyId, Var newValue, PropertyValueInfo * info, ScriptContext* requestContext, PropertyOperationFlags propertyOperationFlags)
     {
+        if (propertyId == PropertyIds::next && object == requestContext->GetLibrary()->GetArrayIteratorPrototype())
+        {
+            requestContext->GetLibrary()->SetArrayIteratorHasUserDefinedNext(true);
+        }
+
```
ianwjhalliday: We should not be in the business of adding special cases to SetProperty.  Same goes for DefineProperty below.  Likewise I'm also not keen on the special case in DefineProperty for species, but I don't fully understand that situation so will not argue it here.

@abchatra and I discussed this briefly and one potential solution is to use the disable implicit calls feature.  However I think that might still be heavyweight.  Essentially we want to do what @goyakin did for Regex properties.  If the type matches what we knew it to be at creation, then we can cache the slot index at creation time and use that to directly access the property on the object.

However, if the object's type changes, e.g. the user adds a new method to Array.prototype, then we cannot assume anything.  We would need to update the type we compare with and get the new slot (in case it changed, very low chance but not impossible).

Let's chat about this offline.

suwc: @abchatra shall we do this for 1.2 and see if we can find a better solution later?

abchatra: I agree with @ianwjhalliday here. Overriding the next here won't work with the inlinecache. There are too many SetProperty overrides which can land in next being added to the prototype object.  We should find and cost the correct solution. To take it for 1.2 depends on the cost of the solution. 

---
In reply to: [57463986](https://github.com/Microsoft/ChakraCore/pull/635#discussion_r57463986) [](ancestors = 57463986)

abchatra:
:+1:

lib/Runtime/Language/JavascriptOperators.cpp
suwc: Makes sense. Returning to using GetProperty() and comparison of `%ArrayIteratorPrototype%.next` to its built-in function. Calling GetProperty() with implicit call disabled. Updated in latest rev.

suwc:
@abchatra latest rev uses GetProperty() to compare `%ArrayIteratorPrototype%.next with stored `JavascriptFunction` object while implicit call disabled. Verified no perf regression.

abchatra:
:-1:

lib/Runtime/Library/TypedArray.cpp
```diff
@@ -94,6 +94,15 @@ namespace Js
         return newArr;
     }
 
+    bool ArrayIteratorPrototypeHasUserDefinedNext(ScriptContext *scriptContext)
```
abchatra: Can you add this to TypedArrayBase class?

lib/Runtime/Library/JavascriptLibrary.inl
```diff
@@ -49,4 +49,22 @@ namespace Js
         // dummy template function
     }
 #endif
+
+    template <class Fn>
+    ImplicitCallFlags TryWithDisabledImplicitCall(ThreadContext *threadContext, Fn fn)
```
abchatra: Can this method be in ThreadContext?

abchatra:
:+1:

lib/Runtime/Library/TypedArray.cpp
suwc: Yes. Updated.

lib/Runtime/Library/JavascriptLibrary.inl
suwc: Yes. Updated.

suwc:
@ianwjhalliday @abchatra thanks for the feedback. If there's no more comments I plan to merge this one soon.

PR #636, opened 2016-03-23
Returning nullptr in preReserved Virtual Alloc on low virtual memory(OOM)

satheeshravi:
We are currently asserting when we can't allocate any more memory from the
pre-reserved region due to low virtual memory on the system.
We have hit low virtual memory limits on web crawler runs.
We should ideally return nullptr, and the call site will take care of reporting the OOM failure.


digitalinfinity:
LGTM- can you explain what the comment that was previously there was trying to say?

satheeshravi:
That comment is wrong and should not be there.

curtisman:
LGTM

LouisLaf:
LGTM

PR #639, opened 2016-03-24
Assert with Array.prototype.map and ES5Array.

akroshg:
In the EntryMap function, we did not handle the ES5Array correctly, as in the case of ES5Array the newArr will be null. Fixed that as it is done in other APIs.


akroshg:
@suwc please take a look.

boingoing:
Looks good to me

akroshg:
Thanks @boingoing

PR #641, opened 2016-03-24
xplat: build lib/Runtime/Language, lib/Runtime/Math

jianchun:
Notable changes:
- InterpreterStackFrame.cpp: refactor code to get rid of a "goto", which
  was illegal as it skipped other variable initialization.
- TaggedInt.cpp: replaced __try/__except with explict test to deal with a
  floating point exception case.
- JavascriptMath.cpp: "and", "or" are reserved keywords and not allowed
  for variable names.
- pal_mstypes.h: LARGE_INTEGER should have an unnamed union member.

To be addressed:
- implement _ultow_s, _ui64tow_s
- design/implement xplat stack walker (RtlVirtualUnwind, ...)
- Simd only partially builds. The rest needs lots of functions declared in
  emmintrin.h/xmmintrin.h. But PAL conflicts with it. We have been copying
  some functions by hand into palrt.h, but that seems not maintainable.

### DIFF
--- a/lib/Common/CommonPal.h
+++ b/lib/Common/CommonPal.h
@@ -31,7 +31,6 @@ typedef wchar_t char16;
 #else // !_WIN32
 
 #define USING_PAL_STDLIB 1
-
 #include "inc/pal.h"
 #include "inc/rt/palrt.h"
 #include "inc/rt/no_sal2.h"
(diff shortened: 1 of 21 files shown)


jianchun:
@digitalinfinity Could you please review? Thanks!

digitalinfinity:
LGTM- the TaggedInt part, is there any perf impact?

jianchun:
> LGTM- the TaggedInt part, is there any perf impact?

Thanks! I inquired Louis about the TaggedInt change last week and was told it should be fine as it is now covered by type-specialization.

PR #643, opened 2016-03-25
Fixing a bug with forcing type-spec on a loop header.

rajatd:
In case of more than one loop back edges, we can force type-spec on the loop header only for the first back edge.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/643)


rajatd:
@LouisLaf , could you review this please?

rajatd:
Closing this. Will create a fresh PR based on release/1.2

PR #644, opened 2016-03-25
Type specialize Math.pow in JIT

ThomsonTan:
JIT does type specialization for all parameters of Math.pow to float. This causes unnecessary int->double->int conversion if the input exponent is integer. This change tries to do int specialization for its parameters.  Exponentiation operator (**) and Math.pow(int, int) will be included later.


lib/Runtime/Library/JavascriptNumber.cpp
```diff
@@ -162,6 +162,29 @@ namespace Js
         return Is_NoTaggedIntCheck(object) && TryGetInt32Value(GetValue(object), &i) ? i : 0;
     }
 
+    double JavascriptNumber::DirectPowDoubleInt(double x, int32 y)
```
abchatra: Who uses this?

lib/Backend/GlobOpt.cpp
```diff
@@ -9585,6 +9584,32 @@ GlobOpt::TypeSpecializeInlineBuiltInBinary(IR::Instr **pInstr, Value *src1Val, V
             break;
         }
 
+        case Js::OpCode::InlineMathPow:
+        {
+#ifndef _M_ARM
```
abchatra: Why not do this for ARM as well?

lib/Runtime/Library/JavascriptNumber.cpp
ThomsonTan: This is referenced as `IR::HelperDirectMath_PowDoubleInt` in JIT.

lib/Backend/GlobOpt.cpp
ThomsonTan: ARM has its specific MD (it doesn't use LowerMDShared.cpp) specific lowering and requires separated implementation, so doesn't fall into this general path.

abchatra: Yes Can we fix LowerMd::GenerateFastInlineBuiltInCall for ARM as well?

---
In reply to: [57507152](https://github.com/Microsoft/ChakraCore/pull/644#discussion_r57507152) [](ancestors = 57507152)

lib/Runtime/Library/JavascriptNumber.cpp
ThomsonTan: This is referenced as `IR::HelperDirectMath_PowDoubleInt` in JIT.

---
In reply to: [57507059](https://github.com/Microsoft/ChakraCore/pull/644#discussion_r57507059) [](ancestors = 57507059)

lib/Backend/GlobOpt.cpp
ThomsonTan: ARM has its specific MD (it doesn't use LowerMDShared.cpp) specific lowering and requires separated implementation, so doesn't fall into this general path.

---
In reply to: [57507086](https://github.com/Microsoft/ChakraCore/pull/644#discussion_r57507086) [](ancestors = 57507086)

ThomsonTan: Will consider to apply the same optimization for arm later.

---
In reply to: [57507236](https://github.com/Microsoft/ChakraCore/pull/644#discussion_r57507236) [](ancestors = 57507236)

rajatd: I would rather that we do the optimization for all architectures together to maintain parity for ARM. ARM has a way of lagging behind when we defer ARM specific stuff for later.

ThomsonTan: This sounds reasonable. I'd like to create a separate work item on this and follow it up after the current one for x86/amd64 because of the difference in details.

---
In reply to: [57610176](https://github.com/Microsoft/ChakraCore/pull/644#discussion_r57610176) [](ancestors = 57610176)

lib/Backend/LowerMDShared.cpp
```diff
@@ -8684,6 +8684,18 @@ void LowererMD::GenerateFastInlineBuiltInCall(IR::Instr* instr, IR::JnHelperMeth
         Assert(helperMethod == (IR::JnHelperMethod)0);
         return GenerateFastInlineBuiltInMathAbs(instr);
 
+    case Js::OpCode::InlineMathPow:
+#ifdef _M_IX86
+        if (!instr->GetSrc2()->IsFloat())
```
nmostafa: If the exponent int32 type-spec can happen for both x86/x64, why restrict this to x86 ?

ThomsonTan: The function called by Math.pow(double, double) has naked calling convention on x86 so it need to fallthough to previous path.

---
In reply to: [57631625](https://github.com/Microsoft/ChakraCore/pull/644#discussion_r57631625) [](ancestors = 57631625)

nmostafa: Ah, I see you handle both in GenerateFastInlineBuiltInMathPow.

ThomsonTan: Yes, GenerateFastInlineBuiltInMathPow should handle all except the customized calling convention.

---
In reply to: [57632884](https://github.com/Microsoft/ChakraCore/pull/644#discussion_r57632884) [](ancestors = 57632884)

lib/Backend/LowerMDShared.cpp
```diff
@@ -9321,6 +9332,39 @@ void LowererMD::GenerateFastInlineBuiltInMathAbs(IR::Instr* inlineInstr)
...
+#ifdef _M_IX86
+    AssertMsg(!instr->GetSrc2()->IsFloat(), "Math.pow(*, double) needs customized lowering!");
+#endif
+
+    IR::JnHelperMethod directPowHelper = (IR::JnHelperMethod)0;
+
+    if (instr->GetSrc2()->IsFloat())
+    {
+        AssertMsg(instr->GetSrc1()->IsFloat(), "Math.Pow(int, double) should not generated by GlobOpt!");
+        directPowHelper = IR::HelperDirectMath_Pow;
```
nmostafa: Better to ifdef this for x64, for clarity and to not have unreachable code on x86 ?

ThomsonTan: This looks good. Will enclose instr->GetSrc2()->IsFloat() into ifdef.

---
In reply to: [57633793](https://github.com/Microsoft/ChakraCore/pull/644#discussion_r57633793) [](ancestors = 57633793)

lib/Backend/GlobOpt.cpp
```diff
@@ -9585,6 +9584,32 @@ GlobOpt::TypeSpecializeInlineBuiltInBinary(IR::Instr **pInstr, Value *src1Val, V
...
+
+                IR::Opnd* src1 = instr->GetSrc1();
+                this->ToFloat64(instr, src1, this->currentBlock, src1Val, nullptr, IR::BailOutPrimitiveButString);
+
+                IR::Opnd* src2 = instr->GetSrc2();
+                this->ToInt32(instr, src2, this->currentBlock, src2Val, nullptr, lossy);
+
+                IR::Opnd* dst = instr->GetDst();
+                this->ToFloat64Dst(instr, dst->AsRegOpnd(), this->currentBlock);
+            }
```
nmostafa: I think it is better to call TypeSpecializeFloatDst() here instead. That way the dst gets the correct Value propagated downstream.

ThomsonTan: Good catch.

---
In reply to: [57645551](https://github.com/Microsoft/ChakraCore/pull/644#discussion_r57645551) [](ancestors = 57645551)

PR #647, opened 2016-03-25
Fixing a subtle Switch Opt Assert

satheeshravi:
Issue:
The interpreter profiles the switch expression as a string. But the
globopt sees the value of the expression as an Integer, due to effects
of constant folding and BailOnNoProfile.

Fix:
When such incompatibility is seen, we do a compile time rejit.
This is already done when we have a sym. But we were just asserting for
this in constant folding path.

### DIFF
--- a/lib/Backend/GlobOpt.cpp
+++ b/lib/Backend/GlobOpt.cpp
@@ -8988,7 +8988,11 @@ GlobOpt::OptConstFoldUnary(
     case Js::OpCode::Ld_A:
         if (instr->HasBailOutInfo())
         {
-            Assert(instr->GetBailOutKind() == IR::BailOutExpectingInteger);
+            //The profile data for switch expr can be string and in GlobOpt we realize it is an int.
+            if(instr->GetBailOutKind() == IR::BailOutExpectingString)
+            {
+                throw Js::RejitException(RejitReason::DisableSwitchOptExpectingString);
+            }
             instr->ClearBailOutInfo();
         }
         value = intConstantValue;


satheeshravi:
@LouisLaf @rajatd : please review.
For a pattern like 
if(a)
{
}
else
{
// a is assumed to be 0 here. 
}
This is the reason, why switch's Ld_A source operand was being constant folded with 0. This is correct and it is just not compatible with switch being optimized for string cases. Hence throwing a rejit exception for this very corner case.

lib/Backend/GlobOpt.cpp
```diff
@@ -8988,7 +8988,11 @@ GlobOpt::OptConstFoldUnary(
     case Js::OpCode::Ld_A:
         if (instr->HasBailOutInfo())
         {
-            Assert(instr->GetBailOutKind() == IR::BailOutExpectingInteger);
+            //The profile data for switch expr can be string and in GlobOpt we realize it can be a int.
```
rajatd: nit:Sicne, we are constant folding, the a" should be "is an" to convey the right information

rajatd:
:+1:

lib/Backend/GlobOpt.cpp
```diff
@@ -8988,7 +8988,11 @@ GlobOpt::OptConstFoldUnary(
     case Js::OpCode::Ld_A:
         if (instr->HasBailOutInfo())
         {
-            Assert(instr->GetBailOutKind() == IR::BailOutExpectingInteger);
```
LouisLaf: You should keep the assert after the throw.

satheeshravi: You mean, after the if condition ?

LouisLaf: If not string, on the non throwing path, assert we have ExpectingInt.

PR #648, opened 2016-03-25
Assert when no identifier on object pattern.

akroshg:
We were allowing the empty identifier (or no names) on the object pattern. The code path is common between Array and Object destructuring. the Array can have empty ones. Fixed that by not allowing it on object pattern.


akroshg:
@tcare  - please have a look.

tcare:
Looks good. Out of interest, was this asserting or actually getting to codegen?

akroshg:
we have assert and then check.

PR #650, opened 2016-03-25
fix compact counter on detecting background access

leirocks:
previously we assume that we don't change these counters after jit begins. however if there's prejit, we mark the counters has been accessed from background thread already and does not allow update, and then with force searialize bytecode, it is updating the counter (searialize ID).

fixing this by block the counter updating only when there's background thread JIT is ongoing, and unblock when the JIT completed


leirocks:
@curtisman can you review?

leirocks:
see #712 targeting release/1.2

Issue #589, opened 2016-03-16
Error.stack undefined when Error.stackTraceLimit == 0


jianchun:
Small compatibility issue.

```
Error.stackTraceLimit = 0
new Error('abc').stack
```

Chrome: ```"Error: abc"```
Edge: ```undefined```

Yongqu:
what's FF's behavior?

jianchun:
> what's FF's behavior?

FF doesn't seem to support this. This is a Chrome/V8 API.

PR #655, opened 2016-03-26
Fixes `Error.stack` is being `undefined` when `Error.stackTraceLimit==0`

obastemur:
Fixes https://github.com/Microsoft/ChakraCore/issues/589

### DIFF
--- a/lib/Runtime/Language/JavascriptExceptionOperators.cpp
+++ b/lib/Runtime/Language/JavascriptExceptionOperators.cpp
@@ -682,52 +682,52 @@ namespace Js
         // NOTE Don't set the throwing exception here, because we might need to box it and will cause a nested stack walker
         // instead, return it to be set in WalkStackForExceptionContext
 
-        if (stackCrawlLimit == 0)
-        {
-            return caller;
-        }
-
-        const bool crawlStackForWER = CrawlStackForWER(scriptContext);
+        JavascriptExceptionContext::StackTrace *stackTrace = nullptr;
         // If we take an OOM (JavascriptException for OOM if script is active), just bail early and return what we've got
         HRESULT hr;
-        JavascriptExceptionContext::StackTrace *stackTrace = NULL;
-
         BEGIN_TRANSLATE_EXCEPTION_AND_ERROROBJECT_TO_HRESULT_NESTED
         {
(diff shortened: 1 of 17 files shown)


obastemur:
Ping.. This can also go into 1.2 ? Change is very simple actually. IMHO Git diff messed up a bit.

lib/Runtime/Language/JavascriptExceptionOperators.cpp
```diff
@@ -682,50 +682,46 @@ namespace Js
         // NOTE Don't set the throwing exception here, because we might need to box it and will cause a nested stack walker
         // instead, return it to be set in WalkStackForExceptionContext
 
-        if (stackCrawlLimit == 0)
+        JavascriptExceptionContext::StackTrace *stackTrace = RecyclerNew(scriptContext.GetRecycler(), JavascriptExceptionContext::StackTrace, scriptContext.GetRecycler());
+        if (stackCrawlLimit > 0)
```
jianchun: You moved RecyclerNew(...StackTrace...) outside of BEGIN/END_TRANSLATE_EXCEPTION block. This changed OOM logic. This code is called during exception throw. If it OOMs here and throws a different exception, that would be undesirable.

obastemur: > You moved RecyclerNew(...StackTrace...) outside of BEGIN/END_TRANSLATE_EXCEPTION block. 

I had checked couple of usages of RecyclerNew and it had looked fine reaching outside of the block. 

I guess the intention is to make this part safe from any possible throw. @jianchun Thanks for the review. I've updated the PR and kept that in.

jianchun:
Change LGTM. Thanks @obastemur

jianchun:
@obastemur Apologize for delay. I was told this issue is not critical and fix should go to master (post rs1). Could you please rebase this fix to latest master?

In addition, could you please also update related unit tests in  `test\StackTrace`? There are existing tests on `Error.stackTraceLimit == 0`. You can consider in `LongCallStackThrow.js` change `Dump(TrimStackTracePath(e.stack) || e)` to just `Dump(TrimStackTracePath(e.stack))` and see what changes it brings. It looks the `|| e` part was to workaround this issue.

PR #658, opened 2016-03-27
Proxies ignore traps whose value is null

### LINKED ISSUE #646: Proxy should ignore traps whose value is null
```js
const obj = new Proxy({}, {
    getPrototypeOf: null
});

print(Object.getPrototypeOf(obj));
```

This code should print `[object Object]`, but it throws `TypeError: 'getPrototypeOf' is not a function`
The trap should be ignored if it is either `undefined` or `null`, but ChackraCore only ignores it in the first case.

Specification:
```
9.5.1 [[GetPrototypeOf]] ( )
  When the [[GetPrototypeOf]] internal method of a Proxy exotic object O is called, the following steps are taken:
    ...
    5. Let trap be ? GetMethod(handler, "getPrototypeOf").
    6. If trap is undefined, then
        a. Return ? target.[[GetPrototypeOf]]().


7.9.3 GetMethod (V, P)
    1. Assert: IsPropertyKey(P) is true.
    2. Let func be ? GetV(V, P).
    3. If func is either undefined or null, return undefined. <---
```

nicolo-ribaudo:
Fixes #646

### DIFF
--- a/lib/Runtime/Language/JavascriptConversion.cpp
+++ b/lib/Runtime/Language/JavascriptConversion.cpp
@@ -472,29 +472,27 @@ CommonNumber:
         RecyclableObject *const recyclableObject = RecyclableObject::FromVar(aValue);
         ScriptContext *const scriptContext = recyclableObject->GetScriptContext();
 
-        /*7.3.7 GetMethod (O, P)
-        The abstract operation GetMethod is used to get the value of a specific property of an object when the value of the property is expected to be a function.
-        The operation is called with arguments O and P where O is the object, P is the property key. This abstract operation performs the following steps:
-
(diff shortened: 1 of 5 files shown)


nicolo-ribaudo:
I just pushed another commit which fixes the same bug but related to @@toPrimitive

dilijev:
Hi @nicolo-ribaudo, since issue #646 is targeted for Milestone 1.2, could you close this PR and make this change on the `release/1.2` branch, and file a new pull request against that branch?

Don't worry about the Jenkins failures on this PR btw. Jenkins was down earlier.

/cc @abchatra @ianwjhalliday

nicolo-ribaudo:
Moved to #677

PR #659, opened 2016-03-28
[wasm] f32.min/max interpreter and inline support

commonlisp:
Working on supporting missing f32 opcodes in wasm

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/659)


lib/Backend/IRBuilderAsmJs.cpp
```diff
@@ -2872,6 +2872,14 @@ IRBuilderAsmJs::BuildFloat3(Js::OpCodeAsmJs newOpcode, uint32 offset, Js::RegSlo
         instr = IR::Instr::New(Js::OpCode::Div_A, dstOpnd, src1Opnd, src2Opnd, m_func);
         break;
 
+    case Js::OpCodeAsmJs::Min_Flt:
+        instr = IR::Instr::New(Js::OpCode::InlineMathMin, dstOpnd, src1Opnd, src2Opnd, m_func);
```
MikeHolman: I took a look and lowering codepath also needs to be changed, i.e. LowerMDShared.cpp around line 9200 we need a case for float32.

MikeHolman: if you want to wait and implement in another checkin that is ok, but I would like if we have explicit failure here. Otherwise if someone tries to JIT this it might lead to confusing failure (crash will not happen until lowering or maybe even further)

abchatra:
Any update on this PR?

commonlisp:
This PR is unfortunately blocked on #831 for now.

MikeHolman:
@commonlisp Is this still blocked?

Cellule:
Could you remove the `.amplxeproj` files from the commit ?

---

Reviewed 13 of 16 files at r1, 1 of 4 files at r2, 1 of 3 files at r3, 13 of 13 files at r4.
Review status: all files reviewed at latest revision, 8 unresolved discussions.

---

*[lib/Backend/LowerMDShared.cpp, line 9292 \[r4\]](https://reviewable.io:443/reviews/microsoft/chakracore/659#-KMRZDNCi-11F6glTPMm:-KMRZDNCi-11F6glTPMn:-1815160871) ([raw file](https://github.com/microsoft/chakracore/blob/7d42bdf4208c7c01a71357ea7b8819b329ced5b4/lib/Backend/LowerMDShared.cpp#L9292)):*
> ```cpp
>                 //      JMP $doneLabel
>                 //
>                 // $labelNegZeroAndNaNCheckHelper
> ```

Could you update this comment to reflect the changes you made

---

*[lib/Backend/LowerMDShared.cpp, line 9333 \[r4\]](https://reviewable.io:443/reviews/microsoft/chakracore/659#-KMRaEkisbuw_b_JGC5F:-KMRaEkisbuw_b_JGC5G:-1133608446) ([raw file](https://github.com/microsoft/chakracore/blob/7d42bdf4208c7c01a71357ea7b8819b329ced5b4/lib/Backend/LowerMDShared.cpp#L9333)):*
> ```cpp
>                     if (src1->IsFloat32())
>                     {
>                         src1f64 = IR::RegOpnd::New(TyFloat64, m_func);
> ```

You should call `EmitFloat32ToFloat64` which contains additional checks

---

*[lib/Backend/LowerMDShared.cpp, line 9418 \[r4\]](https://reviewable.io:443/reviews/microsoft/chakracore/659#-KMRYMIHqbTB70Fr3Uvj:-KMRYMIHqbTB70Fr3Uvk:-741373506) ([raw file](https://github.com/microsoft/chakracore/blob/7d42bdf4208c7c01a71357ea7b8819b329ced5b4/lib/Backend/LowerMDShared.cpp#L9418)):*
> ```cpp
> IR::Opnd* LowererMD::IsOpndNaN(IR::Opnd* opnd, IR::Instr* instr)
> {
>     IR::Opnd * isNaN = IR::RegOpnd::New(TyInt32, this->m_func);
> ```

I think it would be worth to rewrite the NaN implementation here instead of a helper call.
Maybe right now, but a `//todo:: ` if not

---

*[lib/Runtime/ByteCode/OpCodes.h, line 759 \[r4\]](https://reviewable.io:443/reviews/microsoft/chakracore/659#-KMRMs27afX-6ZYfjDbx:-KMRMs27afX-6ZYfjDby:1256888354) ([raw file](https://github.com/microsoft/chakracore/blob/7d42bdf4208c7c01a71357ea7b8819b329ced5b4/lib/Runtime/ByteCode/OpCodes.h#L759)):*
> ```objc
> MACRO_BACKEND_ONLY(     Trunc_A,            Empty,          OpTempNumberSources | OpCanCSE | OpProducesNumber)
> MACRO_BACKEND_ONLY(     Nearest_A,          Empty,          OpTempNumberSources | OpCanCSE | OpProducesNumber)
> MACRO_BACKEND_ONLY(MinNaN, Empty, OpTempNumberSources | OpCanCSE | OpProducesNumber)
> ```

nit: follow alignment of the file

---

*[lib/Runtime/Library/JavascriptNumber.h, line 51 \[r4\]](https://reviewable.io:443/reviews/microsoft/chakracore/659#-KMRbBeI8Fmhvr_jRxo1:-KMRbBeI8Fmhvr_jRxo2:2123198958) ([raw file](https://github.com/microsoft/chakracore/blob/7d42bdf4208c7c01a71357ea7b8819b329ced5b4/lib/Runtime/Library/JavascriptNumber.h#L51)):*
> ```objc
>         static bool TryToVarFastWithCheck(double value, Var* result);
> 
>         inline static BOOL IsNan(double value) { return NumberUtilities::IsNan(value); }
> ```

No need to change this.  `NumberUtilities::IsNan` returns `bool`

---

*[lib/Runtime/Math/AsmJsMath.inl, line 9 \[r4\]](https://reviewable.io:443/reviews/microsoft/chakracore/659#-KMRSq_nyi7RS7R6guUL:-KMRSq_nyi7RS7R6guUM:1034030020) ([raw file](https://github.com/microsoft/chakracore/blob/7d42bdf4208c7c01a71357ea7b8819b329ced5b4/lib/Runtime/Math/AsmJsMath.inl#L9)):*
> ```cpp
> 
>     template<typename T>
>     __inline T minCheckNan(T aLeft, T aRight)
> ```

We are using `inline` now instead of `__inline`

---

*[test/wasm/f32.js, line 6 \[r4\]](https://reviewable.io:443/reviews/microsoft/chakracore/659#-KMRSRsimiOO9Spc3v-t:-KMRSRsimiOO9Spc3v-u:-856997972) ([raw file](https://github.com/microsoft/chakracore/blob/7d42bdf4208c7c01a71357ea7b8819b329ced5b4/test/wasm/f32.js#L6)):*
> ```JavaScript
> 
> const blob = WScript.LoadBinaryFile('f32.wasm');
> ```

nit: trailing whitespaces

---

*[test/wasm/rlexe.xml, line 45 \[r4\]](https://reviewable.io:443/reviews/microsoft/chakracore/659#-KMRSETVGzYNcPqewVKM:-KMRSETVGzYNcPqewVKN:-1094245624) ([raw file](https://github.com/microsoft/chakracore/blob/7d42bdf4208c7c01a71357ea7b8819b329ced5b4/test/wasm/rlexe.xml#L45)):*
> ```XML
>        <files>f32.js</files>
>        <baseline>f32.baseline</baseline>
>        <compile-flags>-On:Wasm</compile-flags>
> ```

Why have a version without `-on:wasm` ? I think we'll need to have all our tests with `-on:wasm` if we are to merge with master sometime soon.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/microsoft/chakracore/659#-:-KMRJQTpOFBden9iLoUb:53887596)*

commonlisp:
Review status: 8 of 18 files reviewed at latest revision, 8 unresolved discussions.

---

*[lib/Runtime/Library/JavascriptNumber.h, line 51 \[r4\]](https://reviewable.io:443/reviews/microsoft/chakracore/659#-KMRbBeI8Fmhvr_jRxo1:-KMSHa44AOjybHbZ-Lp5:833741382) ([raw file](https://github.com/microsoft/chakracore/blob/7d42bdf4208c7c01a71357ea7b8819b329ced5b4/lib/Runtime/Library/JavascriptNumber.h#L51)):*
> No need to change this.  `NumberUtilities::IsNan` returns `bool`
</details>
The subsequent instructions (especially the comparison) are expecting a uint32 which BOOL is defined to be. It breaks with bool.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/microsoft/chakracore/659)*

Cellule:
:lgtm:

---

Reviewed 10 of 20 files at r5.
Review status: all files reviewed at latest revision, 7 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/microsoft/chakracore/659#-:-KMV3OviSB2xSUeqlrQ-:1417088721)*

lib/Backend/IRBuilderAsmJs.cpp
```diff
@@ -2921,6 +2921,14 @@ IRBuilderAsmJs::BuildFloat3(Js::OpCodeAsmJs newOpcode, uint32 offset, Js::RegSlo
         instr = IR::Instr::New(Js::OpCode::Copysign_A, dstOpnd, src1Opnd, src2Opnd, m_func);
         break;
 
+    case Js::OpCodeAsmJs::Min_Flt:
+        instr = IR::Instr::New(Js::OpCode::MinNaN, dstOpnd, src1Opnd, src2Opnd, m_func);
```
MikeHolman: I don't really like calling this MinNaN. IMO would rather call this something like WasmMin, or not create new opcode and in lowering check to see if the function is a wasm function

lib/Backend/LowerMDShared.cpp
```diff
@@ -9307,15 +9324,52 @@ void LowererMD::GenerateFastInlineBuiltInCall(IR::Instr* instr, IR::JnHelperMeth
...
+                }
+
+                IR::Opnd * src1f64 = src1;
+                IR::Opnd * src2f64 = src2; 
+                if (propagateNaN)
+                {
+                    if (src1->IsFloat32())
+                    {
+                        src1f64 = IR::RegOpnd::New(TyFloat64, m_func);
+                        EmitFloat32ToFloat64(src1f64, src1, instr);
```
MikeHolman: shouldn't the result be a f32 still? we shouldn't emit a cvtss2sd...

lib/Backend/LowerMDShared.cpp
```diff
@@ -9307,15 +9324,52 @@ void LowererMD::GenerateFastInlineBuiltInCall(IR::Instr* instr, IR::JnHelperMeth
...
 
+                        src2f64 = IR::RegOpnd::New(TyFloat64, m_func);
+                        EmitFloat32ToFloat64(src2f64, src2, instr);
+                    }
+
+                    IR::Opnd* isSrc1NaN = IsOpndNaN(src1f64, instr);
+                    IR::Opnd* isSrc2NaN = IsOpndNaN(src2f64, instr);
```
MikeHolman: This is pretty inefficient. Can't you do `JP` instead after the below `COMISD`?

lib/Backend/IRBuilderAsmJs.cpp
commonlisp: Ok.

lib/Backend/LowerMDShared.cpp
commonlisp: The result is ultimately still f32. The f64 is for getting the same comparisons for both f32 and f64. I could alternatively have f32 and f64-specific paths but that has so far turned out to be messy.

lib/Backend/LowerMDShared.cpp
```diff
@@ -9307,15 +9324,52 @@ void LowererMD::GenerateFastInlineBuiltInCall(IR::Instr* instr, IR::JnHelperMeth
...
                 Assert(!dst->IsEqual(src1));
-                this->m_lowerer->InsertMove(dst, src1, instr);
+                
+                IR::Opnd * opndNaN = nullptr;
+
+                bool isF32 = dst->IsFloat32();
+
+                if (isF32)
+                {
+                    opndNaN = IR::MemRefOpnd::New((double*)&(Js::JavascriptNumber::k_Nan32), IRType::TyFloat32, this->m_func);
```
MikeHolman: nit: should be float*

lib/Backend/LowerMDShared.cpp
MikeHolman: i think having different paths is better, because cvtss2sd isn't exactly free. If instead of checking for NaN beforehand you do a JP after the compare, all you need to do for float32 to work is change the first if condition in InsertCompareBranch to accept float32

---
In reply to: [70673548](https://github.com/Microsoft/ChakraCore/pull/659#discussion_r70673548) [](ancestors = 70673548)

lib/Backend/LowerMDShared.cpp
MikeHolman: actually, it looks like we already do this. i think the only change in lowering that we need is for InsertCompareBranch to accept float32

---
In reply to: [70487092](https://github.com/Microsoft/ChakraCore/pull/659#discussion_r70487092) [](ancestors = 70487092)

lib/Backend/LowerMDShared.cpp
```diff
@@ -9239,8 +9240,11 @@ void LowererMD::GenerateFastInlineBuiltInCall(IR::Instr* instr, IR::JnHelperMeth
                 instr->Remove();
             }
             break;
-        }
+        }  
 
+    case Js::OpCode::MinNaN:
+    case Js::OpCode::MaxNaN:
+        propagateNaN = true;
```
MikeHolman: from what i can tell, this is same semantics as for JS, so i think we can use InlineMathMin/InlineMathMax (mostly) as-is

MikeHolman:
:shipit:

PR #661, opened 2016-03-28
Fix bug with forcing type spec on loop header

### LINKED ISSUE #664: Bug with forcing type-specialization on loop header
When forcing type-spec on a loop header, globopt wasn't taking multiple loop back edges into account. In case of more than one loop back edges, we can only force type-spec on the loop header for the first edge.

Created from [Bug 4506788](https://microsoft.visualstudio.com/DefaultCollection/OS/_workItems?_a=edit&id=4506788&triage=true)

rajatd:
We can force type-spec on a loop header only for the first back edge.

Fixes #664 

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/661)


rajatd:
@LouisLaf , could you review this?

LouisLaf:
:+1:

PR #662, opened 2016-03-28
Fix for deserialization issue with split scope

aneeshdk:
Fix for deserialization issue with split scope

When deserializing a split scoped function, the flag to indicate the
function is not merged was not updated. This was causing issues when we
try to creat the slots.


aneeshdk:

lib/Runtime/ByteCode/ByteCodeEmitter.cpp
```diff
@@ -1765,7 +1765,7 @@ void ByteCodeGenerator::InitScopeSlotArray(FuncInfo * funcInfo)
 {
     // Record slots info for ScopeSlots/ScopeObject.
     uint scopeSlotCount = funcInfo->bodyScope->GetScopeSlotCount();
-    uint scopeSlotCountForParamScope = funcInfo->paramScope ? funcInfo->paramScope->GetScopeSlotCount() : 0;
+    uint scopeSlotCountForParamScope = funcInfo->paramScope != nullptr && !funcInfo->paramScope->GetCanMergeWithBodyScope() ? funcInfo->paramScope->GetScopeSlotCount() : 0;
```
akroshg: why paramScope->GetScopeSlotCount not be 0 when GetCanMergeWithbodyScope is true ? I believe the paramScope->GetScopeSlotCount will be populated only when the scope is split.

tcare:
Do we have a test case for this at all?

lib/Runtime/ByteCode/ByteCodeEmitter.cpp
aneeshdk: Fixed it. Added an assert too.

aneeshdk:
As discussed I ran all the test cases with forceserialized flag and fixed the issues.

tcare:
Looks good. Opened #670 to track periodic testing of other variants including forceserialize.

aneeshdk:
@dotnet-bot test disablejit tests please

Issue #217, opened 2016-01-28
linux: unix version of _mul128 to optimize Int64::Mul


jianchun:
In #206, @obastemur has implemented unix version of _mul128 using inline assembly. Consider getting this in to optimize Int64::Mul.

> Prepared a unix/linux version of `__mul128` https://gist.github.com/obastemur/48125aacb960a5b66730

obastemur:
I've been testing the solution on other platforms / compilers. What would be the best place (header file) to define linux/unix version of `__mul128` ?

jianchun:
Since it will be used only by Int64Math.cpp and other code should probably just use Int64Math wrapper, maybe also put it in lib/common/common? (No strong opinion.)

obastemur:
This was implemented while ago. Closing.

PR #668, opened 2016-03-28
cross platform: Implement _mul128

obastemur:
See https://github.com/Microsoft/ChakraCore/issues/217

### DIFF
--- a/lib/Common/Common/Int64Math.cpp
+++ b/lib/Common/Common/Int64Math.cpp
@@ -5,8 +5,24 @@
 #include "CommonCommonPch.h"
 #include "Common/Int64Math.h"
 
-#if defined(_M_X64) && defined(_MSC_VER)
-#pragma intrinsic(_mul128)
+#if defined(_M_X64)
+#if defined(_MSC_VER)
+    #pragma intrinsic(_mul128)
+#else
+    static int64 _mul128(const int64 left, const int64 right, int64 *high) noexcept
+    {
+        int64 low;
+        __asm__
+        (
+            "imulq %[right]\n"
+            // (I)MUL (Q/64) R[D/A]X <- RAX * r/m64
+            : "=d"(*high), "=a"(low)
+            : [right]"rm"(right), "1"(left)
+        );
+
+        return low;
+    }
+#endif
 #endif
 
 bool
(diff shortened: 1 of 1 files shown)


obastemur:
Is there a problem with CI ?

tcare:
Is down for maintenance I believe

lib/Common/Common/Int64Math.cpp
```diff
@@ -5,8 +5,23 @@
...
+    #pragma intrinsic(_mul128)
+#else
+    static int64 _mul128(const int64 left, const int64 right, int64 *high) noexcept
+    {
+        int64 low;
+        __asm__
+        (
+            "imulq %[right]\n"
+            : "=d"(*high), "=a"(low)
+            : [right]"rm"(right), "1"(left)
```
curtisman: I don't know much about gcc's inline asm.  
Looking at the doc, "1"(left) means that you require left to be at the same location as output 1 (which rax).
Am I right?  Why?

curtisman: Oh.  I am going to answer my own question.  :)
It's because you are using the IMUL r/m64 (right) form of the instruction which does RDX (high):RAX (low) ← RAX (left) ∗ r/m64. (right)

Please add a comment.  But it looks good otherwise.

obastemur: @curtisman thanks for the review. 

As cross platform work emerges, there will be more inline assembly code pieces like this one. I mean there is nothing special, it's just imul* and at&t syntax. So, I'm not sure what to comment there though.

curtisman: Just that you are using IMUL r/m64 form that do RDX:RAX <- RAX * r/m64.  That's why it expect RDX:RAX out and (left) on RAX (or output "1"). Avoid having to bring up the Intel instruction manual.

lib/Common/Common/Int64Math.cpp
```diff
@@ -5,8 +5,24 @@
...
+#if defined(_MSC_VER)
+    #pragma intrinsic(_mul128)
+#else
+    static int64 _mul128(const int64 left, const int64 right, int64 *high) noexcept
+    {
+        int64 low;
+        __asm__
+        (
+            "imulq %[right]\n"
+            // (I)MUL (Q/64) R[D/A]X <- RAX * r/m64
```
obastemur: @curtisman okay, done.

dilijev:
@dotnet-bot test this please

PR #669, opened 2016-03-28
Get rest of Lib/Parser building

digitalinfinity:
Compile the following files:
    - CharClassifier.cpp
    - CharSet.cpp
    - Parse.cpp
    - RegexParser.cpp
    - RegexPattern.cpp
    - RegexRuntime.cpp
    - Scan.cpp
    - StandardChars.cpp
    - TextbookBoyerMoore.cpp

Get scan.cpp compiling

Couple noteworthy items:
    - added _ltow_s/_ltow to the PAL
    - Exposed ULongMult from the PAL
    - Switched a __try/__finally in Scan.cpp to TryFinally

### DIFF
--- a/lib/Parser/Parse.cpp
+++ b/lib/Parser/Parse.cpp
@@ -7254,7 +7254,7 @@ ParseNodePtr Parser::ParseStringTemplateDecl(ParseNodePtr pnodeTagFnc)
 
         // We are not able to pass more than a ushort worth of arguments to the tag
         // so use that as a logical limit on the number of string constant pieces.
-        if (stringConstantCount >= USHORT_MAX)
+        if (stringConstantCount >= USHRT_MAX)
         {
             Error(ERRnoMemory);
         }
@@ -7802,7 +7802,7 @@ LPCOLESTR Parser::AppendNameHints(LPCOLESTR left, IdentPtr right, ulong *pNameLe
 
     Assert(leftLen <= ULONG_MAX); // name hints should not exceed ULONG_MAX characters
 
-    if (left == nullptr || leftLen == 0 && !wrapInBrackets)
+    if (left == nullptr || (leftLen == 0 && !wrapInBrackets))
     {
         if (right != nullptr)
         {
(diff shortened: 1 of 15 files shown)


digitalinfinity:
@Jianchun can you review please?

lib/Parser/Parse.cpp
```diff
@@ -10868,7 +10868,7 @@ ParseNodePtr Parser::Parse(LPCUTF8 pszSrc, size_t offset, size_t length, charcou
 
     m_scriptContext->AddSourceSize(m_length);
 
-    if(!m_parseType != ParseType_Deferred)
+    if (m_parseType != ParseType_Deferred)
```
jianchun: Does this fix a bug?

lib/Parser/RegexParser.cpp
```diff
@@ -2948,7 +2949,7 @@ namespace UnifiedRegex
     {
         Assert(program->source == 0);
 
-        program->source = _u("");
+        program->source = const_cast<Char*>(_u(""));
```
jianchun: Can "program->source" be made const Char*?

lib/Parser/StandardChars.h
```diff
@@ -225,6 +225,8 @@ namespace UnifiedRegex
             return true;
         }
 
+        TrivialCaseMapper() {}
+
```
jianchun: nit pick: trailing spaces. What makes this constructor needed... the CompileAssert?

pal/inc/rt/palrt.h
```diff
@@ -1076,6 +1076,33 @@ errno_t __cdecl _itow_s(int _Value, WCHAR *_Dst, size_t _SizeInWords, int _Radix
...
+errno_t __cdecl _ltow_s(long _Value, WCHAR (&_Dst)[_SizeInWords], int _Radix)
+{
+    return _ltow_s(_Value, _Dst, _SizeInWords, _Radix);
+}
+#endif
+
+#if _SAFECRT_USE_INLINES
+
+__inline
+errno_t __cdecl _ltow_s(long _Value, WCHAR *_Dst, size_t _SizeInWords, int _Radix)
```
jianchun: We will also need a bunch other forms... ultow_s, ui64tow_s...

lib/Parser/Parse.cpp
digitalinfinity: Yeah, I think so- I think that was a typo there (we check for m_parseType != ParseType_Deferred when we fire the start event)

lib/Parser/RegexParser.cpp
digitalinfinity: That was the first thing I tried but quite a bit of code assumes its not const so went with the cheaper fix of maintaining the current assumption

pal/inc/rt/palrt.h
digitalinfinity: ui64 already exists I think- ultow_s, we can add- I vote for adding them as and when we need it

jianchun:
LGTM! Thanks for completing these!

lib/Parser/StandardChars.h
digitalinfinity: No, we have a const instance of TrivialCaseMapper. C++ requires that any const objects that are default constructed should have a default constructor- see http://stackoverflow.com/questions/7411515/why-does-c-require-a-user-provided-default-constructor-to-default-construct-a
I can add a comment

jianchun: Good to know! Thanks!

lib/Parser/Parse.cpp
digitalinfinity: Confirmed with @abchatra that this is an existing bug

jianchun: Agree, it should be matching line 10681.

PR #672, opened 2016-03-28
If reparse fails return failure to host, in test build fail fast

agarwal-sandeep:


agarwal-sandeep:
@akroshg please take a look

lib/Runtime/Debug/DebugContext.cpp
```diff
@@ -260,7 +260,7 @@ namespace Js
             return false;
         });
 
-        if (this->scriptContext != nullptr && !this->scriptContext->IsClosed())
+        if (SUCCEEDED(hr) && this->scriptContext != nullptr && !this->scriptContext->IsClosed())
```
akroshg: So what is the idea now? if we fail on one we return the failures to host? but we have changed our scriptcontext already - so this is more unstable.

agarwal-sandeep:
Host should fail fast. Or I can do fail fast in retail build as well.

agarwal-sandeep:
This change looks risky. Shouldn't target release/1.2 will create a PR for master for future release.

PR #674, opened 2016-03-29
Deserializing anonymous function

pleath:
Track anonymous functions bodies when we serialize them in the byte code. Otherwise it's impossible on deserialization to tell the difference between an anonymous function and a function with the name, 'Anonymous function.' The difference determines, for instance, whether instances of the function body have a name property.

### DIFF
--- a/lib/Runtime/ByteCode/ByteCodeSerializer.cpp
+++ b/lib/Runtime/ByteCode/ByteCodeSerializer.cpp
@@ -125,7 +125,7 @@ struct SerializedFieldList {
 #include "SerializableFunctionFields.h"
     bool has_m_lineNumber: 1;
     bool has_m_columnNumber: 1;
-	bool has_m_nestedCount: 1;
+    bool has_m_nestedCount: 1;
 };
 
 C_ASSERT(sizeof(GUID)==sizeof(DWORD)*4);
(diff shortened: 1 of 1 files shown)


lib/Runtime/SerializableFunctionFields.h
```diff
@@ -80,6 +80,7 @@ PRIVATE_FIELDS
     DECLARE_SERIALIZABLE_ACCESSOR_FIELD(uint, LiteralRegexCount, UInt32);
     DECLARE_SERIALIZABLE_ACCESSOR_FIELD(uint, InnerScopeCount, UInt32);
     DECLARE_SERIALIZABLE_ACCESSOR_FIELD(RegSlot, LocalClosureRegister, RegSlot);
+    DECLARE_SERIALIZABLE_ACCESSOR_FIELD(RegSlot, ParamClosureRegister, RegSlot);
```
akroshg: I believe @aneeshdk  already fixed this.

lib/Runtime/ByteCode/ByteCodeSerializer.cpp
```diff
@@ -3743,6 +3751,10 @@ class ByteCodeBufferReader
             (*functionBody)->m_isAsmjsMode = (bitflags & ffIsAsmJsMode) ? true : false;
 #endif
 
+            if ((*functionBody)->paramScopeSlotArraySize > 0)
```
akroshg: @aneeshdk  already fixed this.

akroshg:
Storing and retrieving isAnonymous looks good to me.

lib/Runtime/ByteCode/ByteCodeSerializer.cpp
pleath: Yeah, this commit somehow swallowed Aneesh's when I rebased on his change. Only my own commit is relevant.

-- Paul

PR #677, opened 2016-03-29
Ignore proxy traps and @@toPrimitive if their value is null

### LINKED ISSUE #646: Proxy should ignore traps whose value is null

nicolo-ribaudo:
Fixes #646

### DIFF
--- a/lib/Runtime/Language/JavascriptConversion.cpp
+++ b/lib/Runtime/Language/JavascriptConversion.cpp
@@ -472,29 +472,27 @@ CommonNumber:
         RecyclableObject *const recyclableObject = RecyclableObject::FromVar(aValue);
         ScriptContext *const scriptContext = recyclableObject->GetScriptContext();
 
-        /*7.3.7 GetMethod (O, P)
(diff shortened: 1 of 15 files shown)


Yongqu:
Thanks for fixing both issues. Both code change and test are nice. I'll merge the change in.

PR #678, opened 2016-03-29
Destructuring pattern at param has arguments as declaration.

akroshg:
This was producing assert while generating bytecode since the 'arguments' is not marked as declaration. Fixed that by setting the flag as the arguments is overridden. Added unittest.


akroshg:
@tcare - please have a look

tcare:
Looks good.

akroshg:
This PR was manually submitted.

PR #679, opened 2016-03-29
fix for page heap  

leirocks:
- fix stack capturing for which page heap partially enabled 
- fix for memprotect page heap configuration, not sharing the config with chakra. there's still some other shared recycler configs, like LeakDetect etc. may seperate out those in the future.


leirocks:
@digitalinfinity @curtisman

curtisman:
LGTM

PR #680, opened 2016-03-29
Fix for issues with default parameters in async functions

aneeshdk:
Fix for issues with default parameters in async functions

This changelist fix two issues
1. When the inner function of an async function was calling eval we were
not marking the parent function as ChildCallsEval causing SyntaxError
related to eval were missed.
2. The parameters of async function were not marked as
hasNonLocalReference, so when the parameter is referenced inside the body
wrong environment index calculation was happening.


aneeshdk:
This change is followed up on the PR https://github.com/Microsoft/ChakraCore/pull/681 after the branch update.

PR #681, opened 2016-03-29
Fix for issues with default parameters in async functions

aneeshdk:
This changelist fixes two issues
1. When the inner function of an async function was calling eval we were
not marking the parent function as ChildCallsEval causing SyntaxError
related to eval being missed.
2. The parameters of async function were not marked as
hasNonLocalReference, so when the parameter is referenced inside the body
wrong environment index calculation was happening.

### DIFF
--- a/lib/Parser/Parse.cpp
+++ b/lib/Parser/Parse.cpp
@@ -7558,6 +7558,10 @@ void Parser::TransformAsyncFncDeclAST(ParseNodePtr *pnodeBody, bool fLambda)
     {
         GetCurrentFunctionNode()->sxFnc.SetUsesArguments();
     }
+    if (pnodeFncGenerator->sxFnc.CallsEval() || pnodeFncGenerator->sxFnc.ChildCallsEval())
+    {
+        GetCurrentFunctionNode()->sxFnc.SetChildCallsEval();
+    }
     lastNodeRef = NULL;
 }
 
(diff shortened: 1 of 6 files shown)


aneeshdk:
@ianwjhalliday @akroshg Can you please take a look?

lib/Parser/Parse.cpp
```diff
@@ -7558,6 +7558,10 @@ void Parser::TransformAsyncFncDeclAST(ParseNodePtr *pnodeBody, bool fLambda)
     {
         GetCurrentFunctionNode()->sxFnc.SetUsesArguments();
     }
+    if (pnodeFncGenerator->sxFnc.CallsEval() || pnodeFncGenerator->sxFnc.ChildCallsEval())
```
akroshg: will CallsEval be true for this dummy generator function?

aneeshdk: Yes the body can have eval.

akroshg: isn't this pnodeFncGenerator the dummy wrapper around the another function inside?

aneeshdk: pnodeFncGenerator is the inner function of the async function.

akroshg: ok - looks good to me.

tcare:
Looks good

PR #653, opened 2016-03-26
Declare m_displayShortNameOffset as a serializable field. Fixes #652.

dilijev:
With -forceserialized, the name of the function was being printed incorrectly because the short name offset was not being serialized.

Marking the offset as serializable fixes the issue.

Fixes #652.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/653)

PR #682, opened 2016-03-29
OS 6471427: Declare m_displayShortNameOffset as a serializable field. Fixes #652.

### LINKED ISSUE #652: With -forceserialized, the name of a function is not displayed correctly.
Repro code:

```
function test0() {
  class class6 {
    static func88() {}
  }
  WScript.Echo(class6.func88.name);
}
// generate profile
test0();
```

With -forceserialized, the output is:
`class6.func88`

This is incorrect. It should be just `func88`, as it is without the flag.

Converted from [OS 6471427](https://microsoft.visualstudio.com/web/wi.aspx?pcguid=cb55739e-4afe-46a3-970f-1b49d8ee7564&id=6471427)

dilijev:
With -forceserialized, the name of the function was being printed incorrectly because the short name offset was not being serialized.

Marking the offset as serializable fixes the issue.

Fixes #652


dilijev:
Note to self: still needs a unit test.

dilijev:
Note: Changes not including bytecode change and unit test were previously signed off in #653.

@digitalinfinity @curtisman Does the unit test look good?

dilijev:
@curtisman I moved the unit test to its own file. Confirmed it fails without this change and passes with this change. I'll merge this in when checks pass.

PR #683, opened 2016-03-29
MSFT: 6982641 6982542 6982434 6982395 6982802 7033315

Yongqu:
Fix a couple of fault injection failures. We should notify the host if
importing child module failure for some reason, and don't continue down the execution
pipeline.
We should also skip dtor at shutdown time and the allocator might have been gone.
assert dispose when in noscriptscope
don't do console output if console is not supported by the host.


Yongqu:

dilijev:
LGTM

boingoing:
Looks good

PR #684, opened 2016-03-29
README: Clarify and make more readable information about the path at which binaries can be found.

dilijev:


liminzhu:
Looks good :D

PR #685, opened 2016-03-29
X64 JIT: bad code with ToVar and register spills

pleath:
Use MOV_TRUNC in place of MOV when generating ToVar from i32 on x64. We rely on the register move to clear the upper 32 bits, so make sure it doesn't get folded into a 32-bit spill.


pleath:
@LouisLaf

LouisLaf:
LGTM

PR #686, opened 2016-03-30
update interpreter to use SSE2 div when possible

MikeHolman:
In investigating a difference between JIT and interpreter, I found that interpreter is using x87 for dividing, which is significantly slower than sse2 div, and may not give same result.
To help bring consistency between JIT and interpreter, and to benefit from SSE2, I have added a path to do SSE2 div in the interpreter where possible.

With this, I'm seeing a couple small but consistent perf wins on x86: 3% improvement in Kraken crypto-aes, and 100% (1 ms to 0ms) in sunspider crypto-sha1.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/686)

### DIFF
--- a/lib/Runtime/Math/JavascriptMath.inl
+++ b/lib/Runtime/Math/JavascriptMath.inl
@@ -2,6 +2,9 @@
 // Copyright (C) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE.txt file in the project root for full license information.
 //-------------------------------------------------------------------------------------------------------
+
+#include "RuntimeMathPch.h"
+
 namespace Js
 {
 #ifdef SSE2MATH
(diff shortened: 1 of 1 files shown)


lib/Runtime/Language/InterpreterStackFrame.cpp
```diff
@@ -4532,7 +4533,17 @@ namespace Js
     template <bool doProfile>
     Var InterpreterStackFrame::ProfiledDivide(Var aLeft, Var aRight, ScriptContext* scriptContext, ProfileId profileId)
     {
-        Var result = JavascriptMath::Divide(aLeft, aRight,scriptContext);
+        Var result;
+#ifdef _M_IX86
+        if (AutoSystemInfo::Data.SSE2Available())
```
nmostafa: Why only for x86 ? Also, is it not better to move the SSE2Available check into JavascriptMath::Divide ?

MikeHolman: For x64 we should be compiling with sse2 already. I was afraid extra call overhead (if not inlined) would lose any perf benefit, though I should try because that seems like better general solution e.g. if there are any other ops we want to do this with.

MikeHolman: Seems that doesn't matter for perf, so made the change.

MikeHolman:
While I was here, I added SSE2 multiply as well, which improves sunspider math-partial-sums by 2.5ms, giving an overall win of 2.8% on sunspider on my machine.

MikeHolman:
@abchatra @LouisLaf, can you please review?

Issue #584, opened 2016-03-15
`JavascriptOperators::IsIterable` can lead to script code getting executed twice instead of once


ianwjhalliday:
`JavascriptOperators::IsIterable` calls `JavascriptOperators::GetProperty` when the object in question is a `JavascriptProxy`.  This will lead to code execution in the case that the property is a getter or the proxy has a `get` trap handler installed.

The use of `IsIterable` appears to be to check whether or not to go down the slow iterable iteration code paths rather than fast paths.  These code paths however will then again do a `GetProperty` on the property, which in the proxy case would call that getter or trap handler again.

IsIterable should be removed and uses of it fixed to follow the correct spec semantics.

suwc:
Is this a 1.2 issue? Would be great to have some 2nd opinions.
@abchatra @bterlson ?

abchatra:
What is the test case for this?

suwc:
We don't have a real-world test case right now.
The idea is to make a getter function that has side effect and use it to override [Symbol.iterator] of an object, then use the object (possibly through a proxy) to construct a TypedArray.

ianwjhalliday:
This has an easy repro.  This bug affects any clients of `JavascriptOperators::IsIterable` which fortunately is few for now, including `JavascriptArray::EntryFrom`, `TypedArrayBase::EntryFrom`, and `TypedArrayBase::CreateNewInstance`.

Here is an example repro using `Array.from()`

```javascript
let iterable = {
    count: 0,
    [Symbol.iterator]() {
        print("iterable[Symbol.iterator]() called: " + this.count);
        return {
            i: 0,
            a: [ 'a', 'b', 'c', 'd', 'e' ],
            next() {
                return {
                    value: this.a[this.i],
                    done: this.i++ >= this.a.length
                };
            }
        };
    }
};

let handler = {
    count: 0,
    get(target, propKey, receiver) {
        if (propKey === Symbol.iterator) {
            print("get trap propKey=Symbol.iterator called: " + ++this.count);
        }
        return target[propKey];
    }
}

let p = new Proxy(iterable, handler);
let a = Array.from(p);
print(a.join(''));
```

Running this with `ch.exe` prints the following output

```
get trap propKey=Symbol.iterator called: 1
get trap propKey=Symbol.iterator called: 2
iterable[Symbol.iterator]() called: 0
abcde
```

Notice that the trap handler is called twice but `iterable[Symbol.iterator]()` is only invoked once.

[...]

PR #687, opened 2016-03-30
Array.from should ignore @@iterator if it is undefined or null / Prevent Array.from and TypedArray from getting @@iterator twice

### LINKED ISSUE #667: Array.from should ignore @@iterator if it is undefined
```js
print(JSON.stringify(Array.from({
    [Symbol.iterator]: undefined,
    0: "a",
    1: "b",
    length: 2
})));
```

Expected output: `[ "a", "b" ]`
Actual: `TypeError: Function expected`

nicolo-ribaudo:
Fixes #667

### DIFF
--- a/lib/Runtime/Language/JavascriptOperators.cpp
+++ b/lib/Runtime/Language/JavascriptOperators.cpp
@@ -743,7 +743,7 @@ namespace Js
                 aLeft = JavascriptConversion::ToPrimitive(aLeft, JavascriptHint::HintNumber, scriptContext);
             }
             //BugFix: When @@ToPrimitive of an object is overridden with a function that returns null/undefined
-            //this helper will fall into a inescapable goto loop as the checks for null/undefined were outside of the path 
+            //this helper will fall into a inescapable goto loop as the checks for null/undefined were outside of the path
             return RelationalComparisonHelper(aLeft, aRight, scriptContext, leftFirst, undefinedAs);
         }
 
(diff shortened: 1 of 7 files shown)


lib/Runtime/Language/JavascriptOperators.cpp
```diff
@@ -10175,22 +10175,8 @@ namespace Js
 
     BOOL JavascriptOperators::IsIterable(RecyclableObject* instance, ScriptContext* scriptContext)
     {
-        if (JavascriptProxy::Is(instance))
```
abchatra: Can you run all the benchmarks to see performance impact of this?

Yongqu:
Change looks great. Thanks for fixing this. I'll merge it in.

Yongqu:
oh I think we can get this in release 1.2. Can you help port this to release/1.2 branch? thanks.

lib/Runtime/Language/JavascriptOperators.cpp
```diff
@@ -10175,22 +10175,8 @@ namespace Js
...
-            else
-            {
-                return TRUE;
-            }
-        }
-        else
-        {
-            return JavascriptOperators::HasProperty(instance, PropertyIds::_symbolIterator);
-        }
+        Var func = JavascriptOperators::GetProperty(instance, PropertyIds::_symbolIterator, scriptContext);
```
ianwjhalliday: This is not allowed. See #584.

I want to remove IsIterable().  It leads to code that does not match the spec.

@nicolo-ribaudo can you find a different way.  Possibly by fixing #584?

nicolo-ribaudo: Sorry, I didn't see that issue. I pushed a new commit which fixes both #667 and #584

lib/Runtime/Language/JavascriptOperators.cpp
```diff
@@ -10175,55 +10175,47 @@ namespace Js
...
-        }
-        else
-        {
-            return JavascriptOperators::HasProperty(instance, PropertyIds::_symbolIterator);
-        }
-    }
-
     // GetIterator as described in ES6.0 (draft 22) Section 7.4.1
-    RecyclableObject* JavascriptOperators::GetIterator(Var iterable, ScriptContext* scriptContext)
+    RecyclableObject* JavascriptOperators::GetIterator(Var iterable, ScriptContext* scriptContext, bool optional)
```
suwc: I would use a more descriptive name than `optional`, e.g. `noThrowOnNonCallable`.

ianwjhalliday: `optional` is okay I think.  `returnNullptrOnUndefOrNull` feels too verbose to me but I can't think of a terser equally descriptive name, other than `optional` and looking at the context.

I wonder if maybe instead of a bool parameter there should simply be two separate APIs.  I'd say its not important enough to delay getting this merged in though.

ianwjhalliday:
LGTM!

lib/Runtime/Language/JavascriptOperators.cpp
suwc: Agreed.

suwc:
@nicolo-ribaudo, thanks for your contribution!

PR #688, opened 2016-03-30
[wasm] eqz, copysign, trunc, and nearest ops

commonlisp:
Implements eqz, copysign, trunc, and nearest ops in the interpreter


lib/Backend/IRBuilderAsmJs.cpp
```diff
@@ -2553,6 +2553,10 @@ IRBuilderAsmJs::BuildInt2(Js::OpCodeAsmJs newOpcode, uint32 offset, Js::RegSlot
         }
 
         break;
+    
+    case Js::OpCodeAsmJs::Eqz_Int:
+        instr = IR::Instr::New(Js::OpCode::CmEqz_I4, dstOpnd, srcOpnd, m_func);
```
MikeHolman: Do we really need a new opcode for this?  I think cmeq_i4 with opnd2 of 0 should be fine

lib/Backend/IRBuilderAsmJs.cpp
```diff
@@ -2826,6 +2830,10 @@ IRBuilderAsmJs::BuildFloat2(Js::OpCodeAsmJs newOpcode, uint32 offset, Js::RegSlo
     case Js::OpCodeAsmJs::I_Conv_VTF:
         instr = IR::Instr::New(Js::OpCode::Ld_A, dstOpnd, srcOpnd, m_func);
         break;
+    case Js::OpCodeAsmJs::Copysign_Flt:
+    case Js::OpCodeAsmJs::Trunc_Flt:
+    case Js::OpCodeAsmJs::Nearest_Flt:
```
MikeHolman: can you add an Assert(UNREACHED) here? I'm pretty sure this will crash later with nullptr dereference if we come here, but would be good to have more explicit error if trying to jit these that they are not yet implemented

lib/Runtime/Math/AsmJsMath.inl
```diff
@@ -167,4 +167,27 @@ namespace Js
...
+    template<typename T>
+    inline T AsmJsMath::Copysign(T aLeft, T aRight)
+    {
+        return copysign(aLeft, aRight);
+    }
+
+    template<typename T>
+    inline T AsmJsMath::Trunc(T value)
+    {
+        return value;
```
MikeHolman: can you also Assert(UNREACHED) here and in Nearest? If you're not going to implement these yet, it's important that we have explicit failure in case someone tries to use them.

lib/Runtime/Math/AsmJsMath.inl
```diff
@@ -167,4 +167,27 @@ namespace Js
...
     }
+
+    inline int AsmJsMath::Eqz(int value)
+    {
+        return value == 0;
+    }
+
+    template<typename T>
+    inline T AsmJsMath::Copysign(T aLeft, T aRight)
+    {
```
commonlisp: Sounds good. Has that PR been merged in?

lib/Backend/IRBuilderAsmJs.cpp
commonlisp: Perhaps. I'm still thinking of whether this will require a real comparison at all. It seems to be introducing an extra layer of branching since this instruction ought only be used with if, thus together they are really an if0. I think that was the CG's intention anyway.

lib/Backend/IRBuilderAsmJs.cpp
commonlisp: Good point. Thanks.

lib/Backend/IRBuilderAsmJs.cpp
MikeHolman: I believe the optimizer will recognize this and change to a test reg, reg; jne (if used in an if condition). But we should validate. If it doesn't, better to update optimizer, which may have more general benefit, than add new op for this special case.

commonlisp:
I'm going to finish up the codegen for this before continuing.

PR #690, opened 2016-03-30
Type specialize Math.pow(int, int)

ThomsonTan:
Type specialize Math.pow for both magnitude and exponent, use powers of 2 to compute the exponentiation. Only integer integer multiplication involved. Both overflow and negative magnitude trigger bailout. Negative magnitude bails out because double is required to return and this scenario looks rare.


lib/Backend/BailOut.cpp
```diff
@@ -2135,6 +2135,18 @@ void BailOutRecord::ScheduleFunctionCodeGen(Js::ScriptFunction * function, Js::S
                 }
                 break;
             }
+            case IR::BailOutOnPowIntIntOverflow:
```
rajatd: you should also add similar code in ScheduleLoopBodyCodeGen

lib/Backend/GlobOpt.cpp
```diff
@@ -9591,13 +9594,29 @@ GlobOpt::TypeSpecializeInlineBuiltInBinary(IR::Instr **pInstr, Value *src1Val, V
...
 
+                this->ToInt32(instr, instr->GetSrc2(), this->currentBlock, src2Val, nullptr, lossy);
+
                 IR::Opnd* src1 = instr->GetSrc1();
-                this->ToFloat64(instr, src1, this->currentBlock, src1Val, nullptr, IR::BailOutPrimitiveButString);
+                int32 valueMin, valueMax;
+                if (src1Val->GetValueInfo()->IsLikelyInt() &&
+                    this->DoPowIntIntTypeSpec() &&
+                    src2Val->GetValueInfo()->GetIntValMinMax(&valueMin, &valueMax, this->DoAggressiveIntTypeSpec()) &&
+                    valueMax >= 0)
```
rajatd: shouldn't this be checking for `valueMin > 0`

rajatd:
You should also update the method `GetBailOutKindName` to handle your new bailout kind

lib/Backend/GlobOpt.cpp
ThomsonTan: Good catch.

---
In reply to: [57935934](https://github.com/Microsoft/ChakraCore/pull/690#discussion_r57935934) [](ancestors = 57935934)

rajatd:
:+1:

PR #692, opened 2016-03-30
fix page alignment when allocating from pending zero list 

leirocks:
Previous while allocating pages from pending zero list, didn't check the page alignment. with page heap it's very easy to hit such issue because in page heap we alloc extra guard page.


leirocks:
@curtisman

PR #693, opened 2016-03-30
Issue with this object handling in param scope

aneeshdk:
Issue with this object handling in param scope

We were allocating scope slot for this object in the body scope. So when a lamdda tries to access this object it was using the wrong scope slot. Another issue was when we try to find the enclosing non-lambda function from a lambda function we always compare with the body scope. But when the lambda is defined inside the param scope, we have to compare it with the param scope.


aneeshdk:
I will reopen this under the right branch

PR #694, opened 2016-03-30
Issue with this object handling in param scope

aneeshdk:
We were allocating scope slot for this object in the body scope. So when a lamdda tries to access this object it was using the wrong scope slot. Another issue was when we try to find the enclosing non-lambda function from a lambda function we always compare with the body scope. But when the lambda is defined inside the param scope, we have to compare it with the param scope.

### DIFF
--- a/test/es6/default-splitscope.js
+++ b/test/es6/default-splitscope.js
@@ -207,12 +207,19 @@ var tests = [
         } 
         assert.areEqual(30, f2.call({y : 10}), "Properties are accessed from the right this object"); 
 
-        var thisObj = {x : 1, y: 20};             
-        function f3(b = () => { this.x = 10; return this.y; }) { 
+        var thisObj = {x : 1, y : 20 };
+        function f3(a, b = () => { a; this.x = 10; return this.y; }) {
+            assert.areEqual(1, this.x, "Assignment from the param scope has not happened yet");
+            assert.areEqual(20, this.y, "y property of the this object is not affected");
             return b; 
         } 
         assert.areEqual(20, f3.call(thisObj)(), "Lambda defined in the param scope returns the right property value from thisObj"); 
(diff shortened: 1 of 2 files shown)


aneeshdk:
@ianwjhalliday @pleath @akroshg @tcare Can you please take a look?

akroshg:
looks good to me

test/es6/default-splitscope.js
```diff
@@ -207,12 +207,19 @@ var tests = [
         } 
         assert.areEqual(30, f2.call({y : 10}), "Properties are accessed from the right this object"); 
 
-        var thisObj = {x : 1, y: 20};             
-        function f3(b = () => { this.x = 10; return this.y; }) { 
+        var thisObj = {x : 1, y: 20 };
```
tcare: nit: space

test/es6/default-splitscope.js
```diff
@@ -207,12 +207,19 @@ var tests = [
...
+        function f3(a, b = () => { a; this.x = 10; return this.y; }) {
+            assert.areEqual(1, this.x, "Assignment from the param scope has not happened yet");
+            assert.areEqual(20, this.y, "y property of the this object is not affected");
             return b; 
         } 
         assert.areEqual(20, f3.call(thisObj)(), "Lambda defined in the param scope returns the right property value from thisObj"); 
         assert.areEqual(10, thisObj.x, "Assignment from the param scope method updates thisObj's property"); 
+
+        function f4(a, b = () => { a; return this}) {
```
tcare: nit: ; and spacing

tcare:
Looks good

test/es6/default-splitscope.js
aneeshdk: Fixed.

test/es6/default-splitscope.js
aneeshdk: Fixed.

PR #696, opened 2016-03-30
Fix a few module linking issues

boingoing:
Fix module cyclic dependencies

Make import and export link errors into syntax errors instead of runtime
errors.

Enable re-export of a locally-bound import. This is re-export without an
"export ... from ..." statement. Instead we allow an import to be
re-exported implicitly. Looks like "import { foo } from 'm'; export { foo
};"

Also fixed a few cases where accessing module exports via slots was
failing. One case here was typeof which has a special case for loading the
target of the typeof statement.

Fix spelling mistake in error code name

### DIFF
--- a/lib/Runtime/Language/SourceTextModuleRecord.cpp
+++ b/lib/Runtime/Language/SourceTextModuleRecord.cpp
@@ -261,7 +261,7 @@ namespace Js
         if (this->localExportRecordList != nullptr)
         {
             localNames = (ExportedNames*)AllocatorNew(ArenaAllocator, allocator, ExportedNames, allocator);
-            this->localExportRecordList->Map([=](ModuleExportEntry exportEntry) {
+            this->localExportRecordList->Map([=](ModuleImportOrExportEntry exportEntry) {
                 PropertyId exportNameId = EnsurePropertyIdForIdentifier(exportEntry.exportName);
                 localNames->Prepend(exportNameId);
             });
@@ -272,7 +272,7 @@ namespace Js
             {
                 localNames = (ExportedNames*)AllocatorNew(ArenaAllocator, allocator, ExportedNames, allocator);
             }
-            this->indirectExportRecordList->Map([=](ModuleExportEntry exportEntry) {
+            this->indirectExportRecordList->Map([=](ModuleImportOrExportEntry exportEntry) {
                 PropertyId exportedNameId = EnsurePropertyIdForIdentifier(exportEntry.exportName);
                 localNames->Prepend(exportedNameId);
             });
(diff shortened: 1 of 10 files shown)


boingoing:
@Yongqu Can you take a look?

lib/Runtime/Language/SourceTextModuleRecord.cpp
```diff
@@ -576,12 +601,27 @@ namespace Js
...
 
         if (this->errorObject != nullptr)
         {
             NotifyParentsAsNeeded();
-            return; 
+            return;
+        }
+
+        SetWasDeclarationInitialized();
+        if (childrenModuleSet != nullptr)
```
Yongqu: childmoduleset need to be initialized first

lib/Runtime/Language/SourceTextModuleRecord.cpp
```diff
@@ -665,6 +705,23 @@ namespace Js
...
+
+        importRecordList->Map([&](ModuleImportOrExportEntry& importEntry) {
+            Js::PropertyId importName = EnsurePropertyIdForIdentifier(importEntry.importName);
+
+            SourceTextModuleRecord* childModule = this->GetChildModuleRecord(importEntry.moduleRequest->Psz());
+            ModuleNameRecord* importRecord = nullptr;
+            if (!childModule->ResolveExport(importName, nullptr, nullptr, &importRecord)
+                || importRecord == nullptr)
+            {
+                JavascriptError::ThrowSyntaxError(scriptContext, JSERR_ModuleResolveImport, importEntry.importName->Psz());
```
Yongqu: probably not worth enter and throw here. you can just create errorobject and return and we'll just return the error object back.

lib/Runtime/Language/SourceTextModuleRecord.cpp
```diff
@@ -373,7 +393,16 @@ namespace Js
             });
             if (localNameId != Js::Constants::NoProperty)
             {
-                resolvedExportMap->AddNew(exportName, { this, localNameId });
+                ModuleRecordBase* sourceModule = this;
+                ModuleNameRecord* importRecord = nullptr;
+                if (this->importRecordList != nullptr
```
Yongqu: can importRecordList be null?

lib/Runtime/Language/SourceTextModuleRecord.cpp
```diff
@@ -373,7 +393,16 @@ namespace Js
...
+                    && this->ResolveImport(localNameId, &importRecord)
```
Yongqu: Add a comment? this is import {exportName as localName} from "module.js" right?

lib/Runtime/Language/SourceTextModuleRecord.cpp
boingoing: I tried but found that if we try to call ModuleDeclarationInstantiation on the child module set before initializing our exports, the cyclical case asserts because child module tries to resolve an export in parent module before slots are allocated. Maybe we should ensure the slots are allocated before resolving exports?

lib/Runtime/Language/SourceTextModuleRecord.cpp
boingoing: Yeah, that makes sense.

lib/Runtime/Language/SourceTextModuleRecord.cpp
boingoing: Yes, I suppose so. This module might have exports but no imports.

lib/Runtime/Language/SourceTextModuleRecord.cpp
boingoing: This is checking for the case of exporting a binding we imported but didn't use the re-export syntax. It would look like:
import { foo } from 'module';
export { foo };

I'll add a comment, it's a little confusing.

PR #697, opened 2016-03-30
Allow built-in getters without side-effect to be called for objects on the stack

goyakin:
Fixes VSO bug #[6149637](https://microsoft.visualstudio.com/web/wi.aspx?pcguid=cb55739e-4afe-46a3-970f-1b49d8ee7564&id=6149637).


goyakin:

abchatra:
:+1:

PR #698, opened 2016-03-30
Fix OS 4721262: FailFast when JsAddRef fails in CallbackMessage ctor.

dilijev:
Allowing the code to continue after a failed JsAddRef will cause corruption of the datastructure which will cause issues later, notable during the dtor when it looks for the Ref which should have been added and should exist.

When OOM occurs and this call fails, we should terminate the process because the OOM is not recoverable in this case. Assert in debug builds and raise the FailFast error in any case.


dilijev:
Reopened as #699 from my own fork

PR #700, opened 2016-03-30
Conversion from uint32 to int32 should be marked lossy

rajatd:
If one of the values in a binary operation is a float constant with a value that fits in a uint32 but not an int32, and the instruction can ignore int overflow, the source value for the purposes of int specialization would have been changed to an int constant value by ignoring overflow. But, the conversion is still lossy.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/700)

### DIFF
--- a/lib/Backend/GlobOpt.cpp
+++ b/lib/Backend/GlobOpt.cpp
@@ -10962,7 +10962,14 @@ GlobOpt::TypeSpecializeBinary(IR::Instr **pInstr, Value **pSrc1Val, Value **pSrc
 
             // Try to type specialize to int32
 
-            lossy = false;
+            // If one of the values is a float constant with a value that fits in a uint32 but not an int32, 
+            // and the instruction can ignore int overflow, the source value for the purposes of int specialization 
+            // would have been changed to an int constant value by ignoring overflow. But, the conversion is still lossy.
+            if (!(src1OriginalVal && src1OriginalVal->GetValueInfo()->IsFloatConstant() && src1Val && src1Val->GetValueInfo()->HasIntConstantValue()) &&
+                !(src2OriginalVal && src2OriginalVal->GetValueInfo()->IsFloatConstant() && src2Val && src2Val->GetValueInfo()->HasIntConstantValue()))
+            {
+                lossy = false;
+            }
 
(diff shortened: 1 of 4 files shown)


rajatd:
@LouisLaf

satheeshravi:
lgtm

LouisLaf:
:+1:

lib/Backend/GlobOpt.cpp
```diff
@@ -10961,7 +10961,14 @@ GlobOpt::TypeSpecializeBinary(IR::Instr **pInstr, Value **pSrc1Val, Value **pSrc
 
             // Try to type specialize to int32
 
-            lossy = false;
+            // If one of the values is a float constant with a value that fits in a uint32 but not an int32, 
+            // and the instruction can ignore int overflow, the source value for the purposes of int specialization 
+            // would have been changed to an int constant value by ignoring overflow. But, the conversion is still lossy.
+            if (!(src1OriginalVal && src1OriginalVal->GetValueInfo()->IsFloatConstant() && src1Val && src1Val->GetValueInfo()->HasIntConstantValue()) &&
```
ThomsonTan: No need to check this?

lib/Backend/GlobOpt.cpp
```diff
@@ -10961,7 +10961,14 @@ GlobOpt::TypeSpecializeBinary(IR::Instr **pInstr, Value **pSrc1Val, Value **pSrc
 
             // Try to type specialize to int32
 
-            lossy = false;
+            // If one of the values is a float constant with a value that fits in a uint32 but not an int32,
```
ThomsonTan: It sounds like we can construct a repro case?

rajatd: Yes, I'll add a unittest. Thanks

lib/Backend/GlobOpt.cpp
ThomsonTan: Seems both src1Val and src2Val need to be checked. Pls ignore this comment.

---
In reply to: [58146247](https://github.com/Microsoft/ChakraCore/pull/700#discussion_r58146247) [](ancestors = 58146247)

Issue #703, opened 2016-03-31
Switch optimizations are not disabled when we rejit because of them.


rajatd:
When we rejit (which means that some assumption we made in the jit is no longer true and we have bailed out enough times to realize that we want updated profile info) we disable the optimization that caused the bailouts and the rejit. Otherwise, there is a high chance that we will keep bailing out and rejitting because of the same reason. This is not being done for switch opts.

dilijev:
Resolved by #704

PR #704, opened 2016-03-31
Misc minor fixes to Switch Optimization

satheeshravi:
- Disabled Switch optimization upon Rejit.
- Cleaned up an API which is no longer called.
- Added back the Assert in the Ld_A const folding API (It was removed
  earlier as it can have only two BailOutKinds, but it is good to have the
  old assert here.)


satheeshravi:
@rajatd : please take a look.
This also resolves  [Issue #703](https://github.com/Microsoft/ChakraCore/issues/703#event-609085223)

rajatd:
Looks good

satheeshravi:
Closing this due to some ChakraHub issue and re-opening https://github.com/Microsoft/ChakraCore/pull/707

PR #706, opened 2016-03-31
Simple X64 JIT issue

pleath:
X64 JIT: Constant prop during register allocation can produce a MOV_TRUNC with an immediate src. The encoder wasn't handling this form, but it's simple to do so.

### DIFF
--- a/lib/Backend/amd64/EncoderMD.cpp
+++ b/lib/Backend/amd64/EncoderMD.cpp
@@ -815,7 +815,7 @@ EncoderMD::Encode(IR::Instr *instr, BYTE *pc, BYTE* beginCodeAddress)
 
             if (opr2->IsImmediateOpnd())
             {
-                Assert(instr->m_opcode == Js::OpCode::MOV);
+                Assert(EncoderMD::IsMOVEncoding(instr));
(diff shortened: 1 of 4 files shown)


pleath:
@LouisLaf @rajatd

lib/Backend/amd64/EncoderMD.cpp
```diff
@@ -1672,7 +1672,7 @@ bool EncoderMD::TryConstFold(IR::Instr *instr, IR::RegOpnd *regOpnd)
 
     bool isNotLargeConstant = Math::FitsInDWord(regOpnd->m_sym->GetLiteralConstValue_PostGlobOpt());
 
-    if (!isNotLargeConstant && (instr->m_opcode != Js::OpCode::MOV || !instr->GetDst()->IsRegOpnd()))
+    if (!isNotLargeConstant && (!EncoderMD::IsMOVInstr(instr) || !instr->GetDst()->IsRegOpnd()))
```
ThomsonTan: Is there a test which causes error if MOV_TRUN is not handled?

pleath:
It's a very large and unreadable auto-generated script. I haven't got a reduced UT.

LouisLaf:
:+1:

PR #707, opened 2016-03-31
Misc minor fixes to Switch Optimization

satheeshravi:
- Disabled Switch optimization upon Rejit.
- Cleaned up an API which is no longer called.
- Added back the Assert in the Ld_A const folding API (It was removed
  earlier as it can have only two BailOutKinds, but it is good to have the
  old assert here.)


satheeshravi:
https://github.com/Microsoft/ChakraCore/pull/704 was closed due to some issue with ChakraHub. So recreating a new (duplicate PR).

PR #711, opened 2016-03-31
use large heap block for all pageheap allocation

leirocks:
- remove the pageheap implementation in small and medium block
- fix couple issues in large heap pageheap implementation (OOM while allocating LargeHeapBlock or adding to heapBlockMap)
- decommit guard page instead of setting to NOACCESS to save memory.
- implement pattern check for corruption
- TrackBit is not supported by LargeHeapBlock today, so disable pageheap for such allocation
- added assert on non-alignment page allocated for recycler, this is found by page heap run

### DIFF
--- a/lib/Common/Memory/HeapBucket.inl
+++ b/lib/Common/Memory/HeapBucket.inl
@@ -18,14 +18,6 @@ HeapBucketT<TBlockType>::RealAlloc(Recycler * recycler, size_t sizeCat)
         memBlock = SnailAlloc(recycler, &allocatorHead, sizeCat, attributes, nothrow);
         Assert(memBlock != nullptr || nothrow);
     }
-    else
-    {
-#ifdef RECYCLER_PAGE_HEAP
-        Assert(allocatorHead.heapBlock == nullptr || !allocatorHead.heapBlock->InPageHeapMode());
-#else
-        Assert(allocatorHead.heapBlock == nullptr);
-#endif
-    }
 
     // If this API is called and throwing is not allowed,
     // check if we actually allocated a block before verifying
--- a/lib/Common/Memory/HeapBucket.inl
+++ b/lib/Common/Memory/HeapBucket.inl
@@ -7,7 +7,7 @@
 template <typename TBlockType>
 template <ObjectInfoBits attributes, bool nothrow>
 __inline char *
-HeapBucketT<TBlockType>::RealAlloc(Recycler * recycler, size_t sizeCat)
+HeapBucketT<TBlockType>::RealAlloc(Recycler * recycler, size_t sizeCat, size_t size)
 {
     Assert(sizeCat == this->sizeCat);
 
@@ -15,7 +15,7 @@ HeapBucketT<TBlockType>::RealAlloc(Recycler * recycler, size_t sizeCat)
 
     if (memBlock == nullptr)
     {
-        memBlock = SnailAlloc(recycler, &allocatorHead, sizeCat, attributes, nothrow);
+        memBlock = SnailAlloc(recycler, &allocatorHead, sizeCat, size, attributes, nothrow);
         Assert(memBlock != nullptr || nothrow);
     }
 
@@ -38,9 +38,17 @@ HeapBucketT<TBlockType>::RealAlloc(Recycler * recycler, size_t sizeCat)
 #endif
         )
     {
-        // Skip the first and the last pointer objects- the first may have next pointer for the free list
-        // the last might have the old size of the object if this was allocated from an explicit free list
-        recycler->VerifyZeroFill(memBlock + sizeof(FreeObject), sizeCat - (2 * sizeof(FreeObject)));
+        if (this->IsPageHeapEnabled(attributes))
+        {
+            // in page heap there's no free list, and using size instead of sizeCat
+            recycler->VerifyZeroFill(memBlock, size);
+        }
+        else
+        {
+            // Skip the first and the last pointer objects- the first may have next pointer for the free list
+            // the last might have the old size of the object if this was allocated from an explicit free list
+            recycler->VerifyZeroFill(memBlock + sizeof(FreeObject), sizeCat - (2 * sizeof(FreeObject)));
+        }
     }
 #endif
 
--- a/lib/Common/Memory/LargeHeapBlock.cpp
+++ b/lib/Common/Memory/LargeHeapBlock.cpp
@@ -168,6 +168,9 @@ LargeHeapBlock::Delete(LargeHeapBlock * heapBlock)
 
 LargeHeapBlock::LargeHeapBlock(__in char * address, size_t pageCount, Segment * segment, uint objectCount, LargeHeapBucket* bucket)
     : HeapBlock(LargeBlockType), pageCount(pageCount), allocAddressEnd(address), objectCount(objectCount), bucket(bucket), freeList(this)
+#ifdef RECYCLER_PAGE_HEAP
+    , pageHeapAllocStack(nullptr), pageHeapFreeStack(nullptr)
+#endif
 {
     Assert(address != nullptr);
     Assert(pageCount != 0);
@@ -197,7 +200,10 @@ LargeHeapBlock::~LargeHeapBlock()
 #ifdef RECYCLER_PAGE_HEAP
     if (this->pageHeapAllocStack != nullptr)
     {
-        this->pageHeapAllocStack->Delete(&NoCheckHeapAllocator::Instance);
+        if (this->pageHeapAllocStack != (StackBackTrace*)1)
+        {
+            this->pageHeapAllocStack->Delete(&NoThrowHeapAllocator::Instance);
+        }
         this->pageHeapAllocStack = nullptr;
     }
 
@@ -205,7 +211,7 @@ LargeHeapBlock::~LargeHeapBlock()
     // Is this okay? Should we delay freeing heap blocks till process/thread shutdown time?
     if (this->pageHeapFreeStack != nullptr)
     {
-        this->pageHeapFreeStack->Delete(&NoCheckHeapAllocator::Instance);
+        this->pageHeapFreeStack->Delete(&NoThrowHeapAllocator::Instance);
         this->pageHeapFreeStack = nullptr;
     }
 #endif
@@ -280,52 +286,104 @@ LargeHeapBlock::ReleasePagesShutdown(Recycler * recycler)
 #endif
 }
 
-template void LargeHeapBlock::ReleasePagesSweep<true>(Recycler * recycler);
-template void LargeHeapBlock::ReleasePagesSweep<false>(Recycler * recycler);
-
-template<bool pageheap>
 void
 LargeHeapBlock::ReleasePagesSweep(Recycler * recycler)
 {
     recycler->heapBlockMap.ClearHeapBlock(this->address, this->pageCount);
 
-    ReleasePages<pageheap>(recycler);
+    ReleasePages(recycler);
 }
 
+#ifdef RECYCLER_PAGE_HEAP
+__declspec(noinline)
+void LargeHeapBlock::VerifyPageHeapPattern()
+{
+    Assert(InPageHeapMode());
+    if (this->allocAddressEnd > this->address) // in case OOM while allocate heapBlockMap
+    {
+#if _M_X64
+        byte* addr = (byte*)this->allocAddressEnd;
+        uint64* alignAddr = (uint64*)Math::Align(((uint64)addr), sizeof(void*));
 
-template void LargeHeapBlock::ReleasePages<true>(Recycler * recycler);
-template void LargeHeapBlock::ReleasePages<false>(Recycler * recycler);
+        for (int i = 0; addr+i<(byte*)alignAddr; i++)
+        {
+            byte current = addr[i];
+            if (current != 0xF0u)
+            {
+                Assert(false);
+                ReportFatalException(NULL, E_FAIL, Fatal_Recycler_MemoryCorruption, 2);
+            }
+        }
 
-template<bool pageheap>
+        for (int i = 0; alignAddr + i < (uint64*)this->allocAddressEnd; i++)
+        {
+            uint64 current = alignAddr[i];
+            if (current != 0xF0F0F0F0F0F0F0F0u)
+            {
+                Assert(false);
+                ReportFatalException(NULL, E_FAIL, Fatal_Recycler_MemoryCorruption, 2);
+            }
+        }
+#else
+        byte* addr = (byte*)this->allocAddressEnd;
+        uint32* alignAddr = (uint32*)Math::Align(((uint32)addr), sizeof(void*));
+
+        for (int i = 0; addr + i<(byte*)alignAddr; i++)
+        {
+            byte current = addr[i];
+            if (current != 0xF0u)
+            {
+                Assert(false);
+                ReportFatalException(NULL, E_FAIL, Fatal_Recycler_MemoryCorruption, 2);
+            }
+        }
+
+        for (int i = 0; alignAddr + i < (uint32*)this->allocAddressEnd; i++)
+        {
+            uint32 current = alignAddr[i];
+            if (current != 0xF0F0F0F0u)
+            {
+                Assert(false);
+                ReportFatalException(NULL, E_FAIL, Fatal_Recycler_MemoryCorruption, 2);
+            }
+        }
+#endif
+    }
+}
+#endif
+
 void
 LargeHeapBlock::ReleasePages(Recycler * recycler)
 {
     Assert(segment != nullptr);
 
-    char* pageAddress = address;
-    size_t realPageCount = pageCount;
+    char* blockStartAddress = this->address;
+    size_t realPageCount = this->pageCount;
 #ifdef RECYCLER_PAGE_HEAP
-    if (pageheap)
+    if (InPageHeapMode())
     {
-        if (InPageHeapMode())
+        Assert(((LargeObjectHeader*)this->address)->isPageHeapFillVerified || this->allocCount == 0);
+        VerifyPageHeapPattern();
+
+        if (guardPageAddress != nullptr)
         {
-            if (guardPageAddress != nullptr)
+
+            if (this->pageHeapMode == PageHeapMode::PageHeapModeBlockStart)
             {
-                DWORD noAccess;
-                if (::VirtualProtect(static_cast<LPVOID>(guardPageAddress), AutoSystemInfo::PageSize, guardPageOldProtectFlags, &noAccess) == FALSE)
-                {
-                    AssertMsg(false, "Unable to set permission for guard page.");
-                    return;
-                }
-                AssertMsg(noAccess == PAGE_NOACCESS, "Guard page should be PAGE_NOACCESS");
-
-                if (this->pageHeapMode == PageHeapMode::PageHeapModeBlockStart)
-                {
-                    pageAddress = guardPageAddress;
-                }
-
-                realPageCount = actualPageCount;
+                blockStartAddress = guardPageAddress;
             }
+            realPageCount = actualPageCount;
+
+            MEMORY_BASIC_INFORMATION memInfo;
+            VirtualQuery(this->address, &memInfo, sizeof(memInfo));
+            if (::VirtualAlloc(guardPageAddress, AutoSystemInfo::PageSize, MEM_COMMIT, memInfo.Protect) == NULL)
+            {
+                // failed to commit the guard page again. decommit all pages
+                RecyclerPageAllocator* pageAllocator = (RecyclerPageAllocator*)this->GetPageAllocator(recycler);
+                pageAllocator->PartialDecommitPages(blockStartAddress, realPageCount, this->address, this->pageCount, this->segment);
+                this->segment = nullptr;
+                return;
+            }            
         }
     }
 #endif
@@ -333,7 +391,7 @@ LargeHeapBlock::ReleasePages(Recycler * recycler)
 #ifdef RECYCLER_FREE_MEM_FILL
     memset(this->address, DbgMemFill, AutoSystemInfo::PageSize * pageCount);
 #endif
-    recycler->recyclerLargeBlockPageAllocator.Release(pageAddress, realPageCount, segment);
+    recycler->GetRecyclerLargeBlockPageAllocator()->Release(blockStartAddress, realPageCount, segment);
     RECYCLER_PERF_COUNTER_SUB(LargeHeapBlockPageSize, pageCount * AutoSystemInfo::PageSize);
 
     this->segment = nullptr;
@@ -482,7 +540,7 @@ LargeHeapBlock::AllocFreeListEntry(size_t size, ObjectInfoBits attributes, Large
 char*
 LargeHeapBlock::Alloc(size_t size, ObjectInfoBits attributes)
 {
-    Assert(HeapInfo::IsAlignedSize(size));
+    Assert(HeapInfo::IsAlignedSize(size) || InPageHeapMode());
     Assert((attributes & InternalObjectInfoBitMask) == attributes);
     AssertMsg((attributes & TrackBit) == 0, "Large tracked object collection not implemented");
 
@@ -513,6 +571,7 @@ LargeHeapBlock::Alloc(size_t size, ObjectInfoBits attributes)
         memset(header, 0, sizeof(LargeObjectHeader));
     }
 #endif
+
     header->objectIndex = allocCount;
     header->objectSize = size;
     header->SetAttributes(recycler->Cookie, (attributes & StoredObjectInfoBitMask));
@@ -556,9 +615,28 @@ LargeHeapBlock::Mark(void* objectAddress, MarkContext * markContext)
         return;
     }
 
+    if (header->isPageHeapAlloc)
+    {
+        this->VerifyPageHeapPattern();
+    }
+
     DUMP_OBJECT_REFERENCE(markContext->GetRecycler(), objectAddress);
 
-    if (!UpdateAttributesOfMarkedObjects(markContext, objectAddress, header->objectSize, header->GetAttributes(this->heapInfo->recycler->Cookie),
+    unsigned char attributes = header->GetAttributes(this->heapInfo->recycler->Cookie);
+    size_t objectSize = header->objectSize;
+    if (this->InPageHeapMode())
+    {
+        // trim off the trailing part which is not a pointer
+        objectSize = objectSize - (objectSize%sizeof(void*));
+        if (objectSize == 0)
+        {
+            // finalizable object must bigger than a pointer size because of the vtable
+            Assert((attributes & (TrackBit | FinalizeBit)) == 0);
+            return;
+        }
+    }
+
+    if (!UpdateAttributesOfMarkedObjects(markContext, objectAddress, objectSize, attributes,
         [&](unsigned char attributes) { header->SetAttributes(this->heapInfo->recycler->Cookie, attributes); }))
     {
         // Couldn't mark children- bail out and come back later
@@ -979,9 +1057,18 @@ LargeHeapBlock::RescanOnePage(Recycler * recycler)
         RECYCLER_STATS_INC(recycler, markData.rescanLargeObjectCount);
         RECYCLER_STATS_ADD(recycler, markData.rescanLargeByteCount, header->objectSize);
 
-        if (!recycler->AddMark(objectAddress, header->objectSize))
+        size_t objectSize = header->objectSize;
+        if (this->InPageHeapMode())
         {
-            this->SetNeedOOMRescan(recycler);
+            // trim off the trailing part which is not a pointer
+            objectSize = objectSize - (objectSize%sizeof(void*));
+        }
+        if (objectSize > 0) // otherwize the object total size is less than a pointer size
+        {
+            if (!recycler->AddMark(objectAddress, objectSize))
+            {
+                this->SetNeedOOMRescan(recycler);
+            }
         }
     }
     return true;
@@ -1090,11 +1177,20 @@ LargeHeapBlock::RescanMultiPage(Recycler * recycler)
 #ifdef RECYCLER_STATS
         bool objectScanned = false;
 #endif
+
+        size_t objectSize = header->objectSize;
+        if (this->InPageHeapMode())
+        {
+            // trim off the trailing part which is not a pointer
+            objectSize = objectSize - (objectSize%sizeof(void*));
+        }
+
+        Assert(objectSize > 0);
         Assert(oldNeedOOMRescan || !header->markOnOOMRescan);
         // Avoid writing to the page unnecessary by checking first
         if (header->markOnOOMRescan)
         {
-            if (!recycler->AddMark(objectAddress, header->objectSize))
+            if (!recycler->AddMark(objectAddress, objectSize))
             {
                 this->SetNeedOOMRescan(recycler);
                 header->markOnOOMRescan = true;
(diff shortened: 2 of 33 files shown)


leirocks:
@curtisman @digitalinfinity

leirocks:
@dotnet-bot test this please

leirocks:
@dotnet-bot test   Windows x64_release please

abchatra:
Any update on this PR?

leirocks:
I'm running pageheap for memGC. Hitesh is reviewing it.

lib/Common/Memory/HeapBucket.inl
```diff
@@ -7,25 +7,17 @@
...
-        memBlock = SnailAlloc(recycler, &allocatorHead, sizeCat, attributes, nothrow);
+        memBlock = SnailAlloc(recycler, &allocatorHead, sizeCat, size, attributes, nothrow);
         Assert(memBlock != nullptr || nothrow);
     }
-    else
-    {
-#ifdef RECYCLER_PAGE_HEAP
-        Assert(allocatorHead.heapBlock == nullptr || !allocatorHead.heapBlock->InPageHeapMode());
-#else
-        Assert(allocatorHead.heapBlock == nullptr);
```
digitalinfinity: Why is this assert removed?

lib/Common/Memory/PageAllocator.cpp
```diff
@@ -479,6 +462,25 @@ PageSegmentBase<T>::DecommitPages(__in void * address, uint pageCount)
 }
 
 template<typename T>
+void
+PageSegmentBase<T>::PartialDecommitPages(__in void * address, size_t totalPageCount, __in void* addressToDecommit, size_t pageCountToDecommit)
+{
+
+    Assert(address >= this->address);
+    Assert(totalPageCount <= allocator->maxAllocPageCount);
+    Assert(((uint)(((char *)address) - this->address)) <= (allocator->maxAllocPageCount - totalPageCount) * AutoSystemInfo::PageSize);
```
digitalinfinity: This cast should probably be to uintptr_t for safety

lib/Common/Memory/PageAllocator.cpp
```diff
@@ -479,6 +462,25 @@ PageSegmentBase<T>::DecommitPages(__in void * address, uint pageCount)
 }
 
 template<typename T>
+void
+PageSegmentBase<T>::PartialDecommitPages(__in void * address, size_t totalPageCount, __in void* addressToDecommit, size_t pageCountToDecommit)
+{
+
+    Assert(address >= this->address);
```
digitalinfinity: Should you assert that the address is within the segments end address too?

lib/Common/Memory/PageAllocator.cpp
```diff
@@ -1387,6 +1394,37 @@ PageAllocatorBase<T>::AddFreePageCount(uint pageCount)
...
+{
+    // TODO: use a specialized PageHeapPageAllocator to simplify the page allocating logic for pageheap
+    if (pageCountTotal > this->maxAllocPageCount) 
+    {
+        SegmentBase<T> * segment = (SegmentBase<T>*) segmentParam;
+        PageTracking::ReportFree((PageAllocator*)this, segment->GetAddress(), AutoSystemInfo::PageSize * segment->GetPageCount());
+        LogFreePages(segment->GetPageCount());
+        LogFreeSegment(segment);
+
+        // when deleteing segement, it call VirtualFree with MEM_RELEASE, so it should be OK
```
digitalinfinity: Nit: spelling- should be deleting the segment

lib/Common/Memory/PageAllocator.cpp
```diff
@@ -1387,6 +1394,37 @@ PageAllocatorBase<T>::AddFreePageCount(uint pageCount)
 
 template<typename T>
 void
+PageAllocatorBase<T>::PartialDecommitPages(__in void * address, size_t pageCountTotal, __in void* decommitAddress, size_t pageCountToDecommit, __in void * segmentParam)
+{
+    // TODO: use a specialized PageHeapPageAllocator to simplify the page allocating logic for pageheap
+    if (pageCountTotal > this->maxAllocPageCount) 
+    {
+        SegmentBase<T> * segment = (SegmentBase<T>*) segmentParam;
```
digitalinfinity: In this case, should we assert that the pageCountTotal == segment->pageCount?

lib/Common/Memory/Recycler.inl
```diff
@@ -280,7 +296,27 @@ Recycler::RealAllocFromBucket(HeapInfo* heap, size_t size)
 #endif
         )
     {
-        VerifyZeroFill(memBlock + sizeof(FreeObject), sizeCat - (2 * sizeof(FreeObject)));
+        // TODO: looks the check has been done already
+
+        bool isPageHeapAlloc = false;
```
digitalinfinity: Isn't there a big switch indicating whether PageHeap is enabled on the recycler at all? Can we wrap this code under that so that we dont need to do any of this checking unless page heap is enabled overall?

lib/Common/Memory/Recycler.cpp
```diff
@@ -1298,6 +1314,28 @@ bool Recycler::AllowNativeCodeBumpAllocation()
 }
 
 
+#if DBG
+void Recycler::VerifyPageHeapFillAfterAlloc(char* memBlock)
```
digitalinfinity: Consider passing in the attributes for the allocation here and asserting that either the found heap block is large or pageheap is not enabled for this object?

lib/Common/Memory/Recycler.cpp
```diff
@@ -1298,6 +1314,28 @@ bool Recycler::AllowNativeCodeBumpAllocation()
...
+{
+    if (IsPageHeapEnabled())
+    {
+        HeapBlock* heapBlock = this->FindHeapBlock(memBlock);
+        if (heapBlock->IsLargeHeapBlock())
+        {
+            LargeHeapBlock* largeHeapBlock = (LargeHeapBlock*)heapBlock;
+            if (largeHeapBlock->InPageHeapMode())
```
digitalinfinity: Should this be an assert?

lib/Common/Memory/LargeHeapBucket.cpp
```diff
@@ -126,17 +124,22 @@ LargeHeapBucket::PageHeapAlloc(Recycler * recycler, size_t size, ObjectInfoBits
         return nullptr;
     }
 
-    size_t actualPageCount = pageCount + 1; // for page heap
+    if(size<sizeof(void*))
```
digitalinfinity: Nit: spacing- also, what scenario is this optimizing for?

lib/Common/Memory/HeapBucket.inl
leirocks: let me add non-pageheap part back

---
In reply to: [60987129](https://github.com/Microsoft/ChakraCore/pull/711#discussion_r60987129) [](ancestors = 60987129)

lib/Common/Memory/LargeHeapBlock.cpp
```diff
@@ -197,15 +200,18 @@ LargeHeapBlock::~LargeHeapBlock()
 #ifdef RECYCLER_PAGE_HEAP
     if (this->pageHeapAllocStack != nullptr)
     {
-        this->pageHeapAllocStack->Delete(&NoCheckHeapAllocator::Instance);
+        if (this->pageHeapAllocStack != (StackBackTrace*)1)
```
digitalinfinity: Eh? What is this for?

lib/Common/Memory/LargeHeapBlock.cpp
```diff
@@ -280,60 +286,112 @@ LargeHeapBlock::ReleasePagesShutdown(Recycler * recycler)
...
 
+#ifdef RECYCLER_PAGE_HEAP
+__declspec(noinline)
+void LargeHeapBlock::VerifyPageHeapPattern()
+{
+    Assert(InPageHeapMode());
+    if (this->allocAddressEnd > this->address) // in case OOM while allocate heapBlockMap
+    {
+#if _M_X64
+        byte* addr = (byte*)this->allocAddressEnd;
```
digitalinfinity: Most of this code is the same across platforms with the exception of a couple lines- consider refactoring into a template function

lib/Common/Memory/LargeHeapBlock.cpp
leirocks: :) OOM when allocating allocStack, I set the pointer to non-null to indicated we did try to allocate for allocStack. when allocating for freeStack, we have check if allocStack is non-null, that check will fail if OOM while allocating allocStack

---
In reply to: [60992519](https://github.com/Microsoft/ChakraCore/pull/711#discussion_r60992519) [](ancestors = 60992519)

lib/Common/Memory/LargeHeapBlock.cpp
```diff
@@ -280,60 +286,112 @@ LargeHeapBlock::ReleasePagesShutdown(Recycler * recycler)
...
-    ReleasePages<pageheap>(recycler);
+    ReleasePages(recycler);
 }
 
+#ifdef RECYCLER_PAGE_HEAP
+__declspec(noinline)
+void LargeHeapBlock::VerifyPageHeapPattern()
+{
+    Assert(InPageHeapMode());
+    if (this->allocAddressEnd > this->address) // in case OOM while allocate heapBlockMap
```
digitalinfinity: I don't understand this comment- isn't this true for almost all cases?

lib/Common/Memory/LargeHeapBlock.cpp
```diff
@@ -280,60 +286,112 @@ LargeHeapBlock::ReleasePagesShutdown(Recycler * recycler)
...
+    {
+#if _M_X64
+        byte* addr = (byte*)this->allocAddressEnd;
+        uint64* alignAddr = (uint64*)Math::Align(((uint64)addr), sizeof(void*));
```
digitalinfinity: Consider renaming addr/alignAddr to objectEndAddress/alignedObjectEndAddress

lib/Common/Memory/LargeHeapBlock.cpp
```diff
@@ -280,60 +286,112 @@ LargeHeapBlock::ReleasePagesShutdown(Recycler * recycler)
...
+        {
+            byte current = addr[i];
+            if (current != 0xF0u)
+            {
+                Assert(false);
+                ReportFatalException(NULL, E_FAIL, Fatal_Recycler_MemoryCorruption, 2);
+            }
+        }
+
+        for (int i = 0; alignAddr + i < (uint64*)this->allocAddressEnd; i++)
```
digitalinfinity: What is this second loop for? Isn't alignedAddr always >= allocAddressEnd (since it's basically allocAddressEnd aligned to the next pointer boundary?)

lib/Common/Memory/LargeHeapBlock.cpp
leirocks: while failed to add the block to heapblock map, we didn't call Alloc() yet, allocAddressEnd should be equal to address. and then freeing the block will trigger the verification

---
In reply to: [60993191](https://github.com/Microsoft/ChakraCore/pull/711#discussion_r60993191) [](ancestors = 60993191)

lib/Common/Memory/LargeHeapBlock.cpp
```diff
@@ -280,60 +286,112 @@ LargeHeapBlock::ReleasePagesShutdown(Recycler * recycler)
...
-                }
+        if (guardPageAddress != nullptr)
+        {
 
-                realPageCount = actualPageCount;
+            if (this->pageHeapMode == PageHeapMode::PageHeapModeBlockStart)
+            {
+                blockStartAddress = guardPageAddress;
             }
+            realPageCount = actualPageCount;
```
digitalinfinity: Is actualPageCount always initialized (even in non-page heap cases)? If so, can we just initialize realPageCount as const size_t realPageCount = this->actualPageCount?

lib/Common/Memory/LargeHeapBlock.cpp
```diff
@@ -280,60 +286,112 @@ LargeHeapBlock::ReleasePagesShutdown(Recycler * recycler)
...
+
+            MEMORY_BASIC_INFORMATION memInfo;
+            VirtualQuery(this->address, &memInfo, sizeof(memInfo));
+            if (::VirtualAlloc(guardPageAddress, AutoSystemInfo::PageSize, MEM_COMMIT, memInfo.Protect) == NULL)
+            {
+                // failed to commit the guard page again. decommit all pages
+                RecyclerPageAllocator* pageAllocator = (RecyclerPageAllocator*)this->GetPageAllocator(recycler);
+                pageAllocator->PartialDecommitPages(blockStartAddress, realPageCount, this->address, this->pageCount, this->segment);
+                this->segment = nullptr;
+                return;
```
digitalinfinity: Do we need to sub the perf counter here?

lib/Common/Memory/LargeHeapBlock.cpp
leirocks: good catch, should be this->addressEnd. the second loop is for optimizing to move with pointer size instead of single byte

---
In reply to: [60993705](https://github.com/Microsoft/ChakraCore/pull/711#discussion_r60993705) [](ancestors = 60993705)

lib/Common/Memory/LargeHeapBlock.cpp
```diff
@@ -280,60 +286,112 @@ LargeHeapBlock::ReleasePagesShutdown(Recycler * recycler)
...
-                realPageCount = actualPageCount;
+            if (this->pageHeapMode == PageHeapMode::PageHeapModeBlockStart)
+            {
+                blockStartAddress = guardPageAddress;
             }
+            realPageCount = actualPageCount;
+
+            MEMORY_BASIC_INFORMATION memInfo;
+            VirtualQuery(this->address, &memInfo, sizeof(memInfo));
+            if (::VirtualAlloc(guardPageAddress, AutoSystemInfo::PageSize, MEM_COMMIT, memInfo.Protect) == NULL)
```
digitalinfinity: Curious- is this necessary? That is, if we leave the page decommitted, and call ReleasePages on the page allocator, what happens?

lib/Common/Memory/LargeHeapBlock.cpp
```diff
@@ -556,9 +615,28 @@ LargeHeapBlock::Mark(void* objectAddress, MarkContext * markContext)
...
+
     DUMP_OBJECT_REFERENCE(markContext->GetRecycler(), objectAddress);
 
-    if (!UpdateAttributesOfMarkedObjects(markContext, objectAddress, header->objectSize, header->GetAttributes(this->heapInfo->recycler->Cookie),
+    unsigned char attributes = header->GetAttributes(this->heapInfo->recycler->Cookie);
+    size_t objectSize = header->objectSize;
+    if (this->InPageHeapMode())
+    {
+        // trim off the trailing part which is not a pointer
+        objectSize = objectSize - (objectSize%sizeof(void*));
```
digitalinfinity: Nit: spacing

lib/Common/Memory/LargeHeapBlock.cpp
```diff
@@ -556,9 +615,28 @@ LargeHeapBlock::Mark(void* objectAddress, MarkContext * markContext)
         return;
     }
 
+    if (header->isPageHeapAlloc)
```
digitalinfinity: In #if DBG or some other similar macro?

lib/Common/Memory/LargeHeapBlock.h
```diff
@@ -29,11 +29,11 @@ struct LargeObjectHeader
 #endif
 
 public:
-    bool markOnOOMRescan;
+    bool markOnOOMRescan:1;
+    bool isExplicitFreed:1;
+    bool isPageHeapAlloc:1;
```
digitalinfinity: Heap block already has page heap mode- what do we need this for? What scenario do we have for storing this bit that we couldn't infer from the other flag?

lib/Common/Memory/LargeHeapBlock.h
```diff
@@ -29,11 +29,11 @@ struct LargeObjectHeader
 #endif
 
 public:
-    bool markOnOOMRescan;
+    bool markOnOOMRescan:1;
```
digitalinfinity: This structure appears to have alignment concerns before- can you confirm that the alignment hasn't changed? (see use of unused2/unused3/unused4)

lib/Common/Memory/LargeHeapBlock.cpp
```diff
@@ -556,9 +615,28 @@ LargeHeapBlock::Mark(void* objectAddress, MarkContext * markContext)
...
+    unsigned char attributes = header->GetAttributes(this->heapInfo->recycler->Cookie);
+    size_t objectSize = header->objectSize;
+    if (this->InPageHeapMode())
+    {
+        // trim off the trailing part which is not a pointer
+        objectSize = objectSize - (objectSize%sizeof(void*));
+        if (objectSize == 0)
+        {
+            // finalizable object must bigger than a pointer size because of the vtable
+            Assert((attributes & (TrackBit | FinalizeBit)) == 0);
```
digitalinfinity: I thought Large block didn't support TrackBit?

lib/Common/Memory/LargeHeapBlock.cpp
```diff
@@ -979,9 +1057,18 @@ LargeHeapBlock::RescanOnePage(Recycler * recycler)
         RECYCLER_STATS_INC(recycler, markData.rescanLargeObjectCount);
         RECYCLER_STATS_ADD(recycler, markData.rescanLargeByteCount, header->objectSize);
 
-        if (!recycler->AddMark(objectAddress, header->objectSize))
+        size_t objectSize = header->objectSize;
+        if (this->InPageHeapMode())
         {
-            this->SetNeedOOMRescan(recycler);
+            // trim off the trailing part which is not a pointer
+            objectSize = objectSize - (objectSize%sizeof(void*));
```
digitalinfinity: Spacing- and also check for underflow

lib/Common/Memory/LargeHeapBlock.cpp
```diff
@@ -1888,3 +1979,56 @@ LargeHeapBlock::GetTrackerDataArray()
...
+        }
+        else
+        {
+            this->pageHeapAllocStack = StackBackTrace::Capture(&NoThrowHeapAllocator::Instance,
+                Recycler::s_numFramesToSkipForPageHeapAlloc, Recycler::s_numFramesToCaptureForPageHeap);
+        }
+        
+        if (this->pageHeapAllocStack == nullptr) 
+        {
+            this->pageHeapAllocStack = (StackBackTrace*)1; // allocate failed, mark it we have tried
```
digitalinfinity: Ah, I see- consider changing this to a sentinel constant rather than using magic numbers

digitalinfinity:
One more comment (can't comment on the line since it's not a part of the patch set)- LargeHeapBucket::Initialize sets isPageHeapEnabled. I think that line needs to be updated too since without that I think -PageHeapBlockType config flag is broken

lib/Common/Memory/Recycler.h
```diff
@@ -1502,7 +1502,7 @@ class Recycler
     template <typename SmallHeapBlockAllocatorType>
     void RemoveSmallAllocator(SmallHeapBlockAllocatorType * allocator, size_t sizeCat);
     template <ObjectInfoBits attributes, typename SmallHeapBlockAllocatorType>
-    char * SmallAllocatorAlloc(SmallHeapBlockAllocatorType * allocator, size_t sizeCat);
+    char * SmallAllocatorAlloc(SmallHeapBlockAllocatorType * allocator, size_t sizeCat, size_t size);
```
digitalinfinity: What is the perf impact of taking in an additional argument? Please do a perf run to make sure that non-page heap builds are not impacted

lib/Common/Memory/PageAllocator.cpp
leirocks: done

---
In reply to: [60987860](https://github.com/Microsoft/ChakraCore/pull/711#discussion_r60987860) [](ancestors = 60987860)

lib/Common/Memory/PageAllocator.cpp
leirocks: done

---
In reply to: [60987690](https://github.com/Microsoft/ChakraCore/pull/711#discussion_r60987690) [](ancestors = 60987690)

lib/Common/Memory/PageAllocator.cpp
leirocks: should be. let me try

---
In reply to: [60988863](https://github.com/Microsoft/ChakraCore/pull/711#discussion_r60988863) [](ancestors = 60988863)

lib/Common/Memory/PageAllocator.cpp
leirocks: done

---
In reply to: [60988264](https://github.com/Microsoft/ChakraCore/pull/711#discussion_r60988264) [](ancestors = 60988264)

lib/Common/Memory/Recycler.inl
leirocks: this is for 'is pageheap enabled on the specified bucket', yes we can wrap this to recycler for given attributes and sizeCat

---
In reply to: [60989559](https://github.com/Microsoft/ChakraCore/pull/711#discussion_r60989559) [](ancestors = 60989559)

lib/Common/Memory/Recycler.cpp
leirocks: done

---
In reply to: [60989848](https://github.com/Microsoft/ChakraCore/pull/711#discussion_r60989848) [](ancestors = 60989848)

lib/Common/Memory/Recycler.cpp
leirocks: added assert

---
In reply to: [60989873](https://github.com/Microsoft/ChakraCore/pull/711#discussion_r60989873) [](ancestors = 60989873)

lib/Common/Memory/LargeHeapBucket.cpp
leirocks: if alloc size is less than pointer size(eg: ask for 3 bytes), it must be a leaf object. we have runtime code and trident code currently not passing leaf attribute for such allocation. 
in page heap mode, I force it to be a leaf, so don't mark it.

I think we should assert or force it to be leaf in non-page heap mode as well. this reduces false positive in x86


---
In reply to: [60990204](https://github.com/Microsoft/ChakraCore/pull/711#discussion_r60990204) [](ancestors = 60990204)

lib/Common/Memory/LargeHeapBlock.cpp
leirocks: done

---
In reply to: [60992932](https://github.com/Microsoft/ChakraCore/pull/711#discussion_r60992932) [](ancestors = 60992932)

lib/Common/Memory/LargeHeapBlock.cpp
leirocks: done

---
In reply to: [60993373](https://github.com/Microsoft/ChakraCore/pull/711#discussion_r60993373) [](ancestors = 60993373)

lib/Common/Memory/LargeHeapBlock.cpp
leirocks: done, and refactored to a function. that always true since size_t is unsigned and objectSize is always positive, right?

---
In reply to: [60994659](https://github.com/Microsoft/ChakraCore/pull/711#discussion_r60994659) [](ancestors = 60994659)

lib/Common/Memory/LargeHeapBlock.cpp
leirocks: done

---
In reply to: [60996418](https://github.com/Microsoft/ChakraCore/pull/711#discussion_r60996418) [](ancestors = 60996418)

lib/Common/Memory/LargeHeapBlock.cpp
leirocks: done

---
In reply to: [60996139](https://github.com/Microsoft/ChakraCore/pull/711#discussion_r60996139) [](ancestors = 60996139)

lib/Common/Memory/LargeHeapBlock.cpp
leirocks: removed TrackBit

---
In reply to: [60996026](https://github.com/Microsoft/ChakraCore/pull/711#discussion_r60996026) [](ancestors = 60996026)

lib/Common/Memory/LargeHeapBlock.h
leirocks: removing, introduced in initial design, not referencing this field in many places now

---
In reply to: [60995282](https://github.com/Microsoft/ChakraCore/pull/711#discussion_r60995282) [](ancestors = 60995282)

lib/Common/Memory/LargeHeapBlock.h
leirocks: adding static assertion for alignment

---
In reply to: [60995415](https://github.com/Microsoft/ChakraCore/pull/711#discussion_r60995415) [](ancestors = 60995415)

lib/Common/Memory/LargeHeapBlock.cpp
leirocks: removing, using this->InPageHeapMode()

---
In reply to: [60995050](https://github.com/Microsoft/ChakraCore/pull/711#discussion_r60995050) [](ancestors = 60995050)

lib/Common/Memory/Recycler.h
leirocks: will do, this needs pogo build, most of these function should be inlined so no stack arguments passing overhead I think

---
In reply to: [60997858](https://github.com/Microsoft/ChakraCore/pull/711#discussion_r60997858) [](ancestors = 60997858)

lib/Common/Memory/LargeHeapBlock.cpp
leirocks: don't need realPageCount  actually

---
In reply to: [60994174](https://github.com/Microsoft/ChakraCore/pull/711#discussion_r60994174) [](ancestors = 60994174)

lib/Common/Memory/LargeHeapBlock.cpp
leirocks: this is for page reusing. I think we should implement a mechanism in pageAllocator to only commit part of the pages, like when asking 2 pages but only commit first page for me. and let the pageAllocator to manage the commit status. 
with that we don't even need to decommit while pageheap allocating. that is also good for some of the medium block, which last 1 or more page is not allocatable, we ask for 8 pages but only commit first 7 pages.

---
In reply to: [60994617](https://github.com/Microsoft/ChakraCore/pull/711#discussion_r60994617) [](ancestors = 60994617)

lib/Common/Memory/LargeHeapBlock.cpp
leirocks: >> If we OOM while allocation, we should just not call this function?
when OOM while adding to heapBlockMap, we are going through regular cleaning up path which calls into VerifyPageHeapPattern()

lib/Common/Memory/LargeHeapBlock.cpp
leirocks: no, perf counter add only called in LargeHeapBlock constructor, guard page is not counted

lib/Common/Memory/LargeHeapBlock.cpp
leirocks: looks we do need realPageCount. actualPageCount  is only initialized in pageheap case.
also, I hit issue that actualPageCount is quite different with pageCount, like pageCount is 0x42 but actualPageCount is 0x50. the pageAllocator returned more pages than wanted. let me look into how the non-page heap alloc handle this case, to see if we can merge these counters

lib/Common/Memory/LargeHeapBlock.cpp
leirocks: let me change to "this->allocCount > 0" which is easier to understand

lib/Common/Memory/LargeHeapBlock.cpp
leirocks: removing usage of actualPageCount, we can just use pageCount+1 for pageheap case

lib/Common/Memory/LargeHeapBlock.cpp
leirocks: done

leirocks:
fixed  -PageHeapBlockType for largeBlockType

lib/Common/Memory/LargeHeapBlock.cpp
leirocks: looks that won't work, for some cases looks the pageAllocator returned aligned page count, 
like if we ask for 0x42 pages, it returns 0x50 pages. need to record this down.

also, since we already switched to LargeHeapBlock design. I think it's just easier to use guard page in both front and back of the object. and if pageAllocator returned more than asked pages, we can set all the extra pages as guard pages

lib/Common/Memory/LargeHeapBlock.h
```diff
@@ -258,6 +258,19 @@ class LargeHeapBlock sealed : public HeapBlock
...
+    PageHeapMode pageHeapMode;
+    char* guardPageAddress;
+    StackBackTrace* pageHeapAllocStack;
+    StackBackTrace* pageHeapFreeStack;
+
+public:
+    __inline bool InPageHeapMode() const { return pageHeapMode != PageHeapMode::PageHeapModeOff; }
+    void CapturePageHeapAllocStack();
+    void CapturePageHeapFreeStack();
+    const StackBackTrace* failedAllocStackTracePtr = (StackBackTrace*)1;
```
digitalinfinity: Nit: consider making this static const and prefix with s_. In any case, change this to StackTraceAllocFailed since it's not a real ptr and constants are PascalCased.

lib/Common/Memory/LargeHeapBlock.cpp
```diff
@@ -280,60 +286,93 @@ LargeHeapBlock::ReleasePagesShutdown(Recycler * recycler)
...
+__declspec(noinline)
+void LargeHeapBlock::VerifyPageHeapPattern()
+{
+    Assert(InPageHeapMode());
+    if (this->allocAddressEnd > this->address) // in case OOM while allocate heapBlockMap
+    {
+        byte* objectEndAddress = (byte*)this->allocAddressEnd;
+#if _M_X64
+        const uint64 pattern = 0xF0F0F0F0F0F0F0F0u;
+        uint64* alignedObjectEndAddress = (uint64*)Math::Align(((uint64)objectEndAddress), sizeof(void*));
```
digitalinfinity: Can probably abstract this out of the ifdef if you use uintptr_t for both cases?

lib/Common/Memory/LargeHeapBlock.cpp
```diff
@@ -280,60 +286,93 @@ LargeHeapBlock::ReleasePagesShutdown(Recycler * recycler)
...
+        uint64* addrEnd = (uint64*)this->addressEnd;
+#else
+        const uint32 pattern = 0xF0F0F0F0u;
+        uint32* alignedObjectEndAddress = (uint32*)Math::Align(((uint32)objectEndAddress), sizeof(void*));
+        uint32* addrEnd = (uint32*)this->addressEnd;
+#endif
 
-template void LargeHeapBlock::ReleasePages<true>(Recycler * recycler);
-template void LargeHeapBlock::ReleasePages<false>(Recycler * recycler);
+        for (int i = 0; objectEndAddress +i<(byte*)alignedObjectEndAddress; i++)
```
digitalinfinity: Nit: spacing

lib/Common/Memory/LargeHeapBlock.cpp
digitalinfinity: Good idea- consider filing an issue for this (or do it as part of the decommit unallocable page change, although probably better to do in a future seperate change)

---
In reply to: [61011532](https://github.com/Microsoft/ChakraCore/pull/711#discussion_r61011532) [](ancestors = 61011532,60994617)

digitalinfinity:
:+1:

lib/Common/Memory/LargeHeapBlock.cpp
leirocks: OK

lib/Common/Memory/LargeHeapBlock.h
leirocks: done

PR #712, opened 2016-03-31
fix compact counter on detecting background access

leirocks:
previously we assume that we don't change these counters after jit begins. however if there's prejit, we mark the counters has been accessed from background thread already and does not allow update, and then with force searialize bytecode, it is updating the counter (searialize ID).

fixing this by block the counter updating only when there's background thread JIT is ongoing, and unblock when the JIT completed

### DIFF
--- a/lib/Runtime/Base/FunctionBody.cpp
+++ b/lib/Runtime/Base/FunctionBody.cpp
@@ -410,6 +410,59 @@ namespace Js
     }
 
 
+    uint32 FunctionBody::GetCountField(FunctionBody::CounterFields fieldEnum) const
+    {
+        // when getting counters from bg thread, assert that it has entered background thread call
+        // this is in order to flush out those call not guarded by the 
+        Assert(ThreadContext::GetContextForCurrentThread() != nullptr || this->bgThreadRank > 0);
+        return counters.Get(fieldEnum);
+    }
+    uint32 FunctionBody::SetCountField(FunctionBody::CounterFields fieldEnum, uint32 val)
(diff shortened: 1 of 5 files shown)


leirocks:
@curtisman  can you review

lib/Runtime/Base/FunctionBody.cpp
```diff
@@ -409,6 +409,24 @@ namespace Js
 #endif
     }
 
+#if DBG
+    FunctionBody::AutoResetThreadState::AutoResetThreadState(Js::FunctionBody* fb)
+        :functionBody(fb), foreground(ThreadContext::GetContextForCurrentThread() != nullptr)
```
abchatra: Good to initialize alreadyInBackground as well

leirocks: good point

abchatra:
:+1:

leirocks:
@dotnet-bot test Windows x86_debug  please

leirocks:

leirocks:

leirocks:
@abchatra @curtisman can you take a look again. I changed the design a bit to fix the assert in unittest in Jenkins machine. the assert is because we have two background jit threads, they can do codegen on same function body for different loop body

leirocks:
with running crawler I found it's hard to make it correct this way, because while JIT on one function, it sometimes trying to get counters from another function body. I'm going to do a spot fix for the prejit+forceserialized and etw rundown scenarios

PR #715, opened 2016-04-01
Scope slot positions for destructuring patterns on param.

akroshg:
We were not assigning the slot position to the destructuring pattern itself. eg funciton foo({}, a), we were not assigning the slot for the pattern, so the 'a' used to get the first slot. This was creating the problem for the arguments object as while creating the arguments object we were having formals count as 2 but only 1 slot assigned so the arguments object to frameObject interaction is messed up. For that reason I have created the slots for the pattern itself.
While doing that I have also made sure the order of assigning slots should be consistent.
such as
1. normal params and patterns
2. rest
3. arguments object
4. Identifiers inside the pattern.

Since I have touched the way 'rest' used to get slots I have added tests for 'rest' as well for validation.

### DIFF
--- a/lib/Runtime/ByteCode/ByteCodeEmitter.cpp
+++ b/lib/Runtime/ByteCode/ByteCodeEmitter.cpp
@@ -1577,15 +1577,20 @@ void ByteCodeGenerator::EmitScopeObjectInit(FuncInfo *funcInfo)
                     Assert(sym->GetScopeSlot() != Js::Constants::NoProperty && sym->GetScopeSlot() > slot);
                     propIds->elements[slot] = Js::Constants::NoProperty;
                 }
-                slot++;
             }
+            else
+            {
+                // This is for patterns
+                propIds->elements[slot] = Js::Constants::NoProperty;
+            }
+            slot++;
         };
         MapFormalsWithoutRest(pnodeFnc, initArg);
 
-        // initArg assumes the sym is in a slot, but this may not be true for the rest parameter.
+        // If the rest is in the slot - we need to keep that slot.
         if (pnodeFnc->sxFnc.pnodeRest != nullptr && pnodeFnc->sxFnc.pnodeRest->sxVar.sym->IsInSlot(funcInfo))
         {
-            initArg(pnodeFnc->sxFnc.pnodeRest);
+            Symbol::SaveToPropIdArray(pnodeFnc->sxFnc.pnodeRest->sxVar.sym, propIds, this);
         }
     }
     else
--- a/lib/Runtime/ByteCode/Scope.cpp
+++ b/lib/Runtime/ByteCode/Scope.cpp
@@ -130,9 +130,10 @@ void Scope::MergeParamAndBodyScopes(ParseNode *pnodeScope, ByteCodeGenerator *by
 
     // Reassign non-formal slot positions. Formals need to keep their slot positions to ensure
     // the argument object works properly. Other symbols need to be reassigned slot positions.
+    // The sym belonging to arguments will use the same slot.
     paramScope->ForEachSymbol([&](Symbol * sym)
     {
-        if (sym->GetSymbolType() != STFormal && sym->GetScopeSlot() != Js::Constants::NoProperty)
+        if (sym->GetSymbolType() != STFormal && sym->GetScopeSlot() != Js::Constants::NoProperty && !sym->GetIsArguments())
         {
             sym->SetScopeSlot(Js::Constants::NoProperty);
             sym->EnsureScopeSlot(pnodeScope->sxFnc.funcInfo);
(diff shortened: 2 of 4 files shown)


akroshg:
@tcare @aneeshdk could you look at this please.

lib/Runtime/ByteCode/ByteCodeEmitter.cpp
```diff
@@ -1578,15 +1578,19 @@ void ByteCodeGenerator::EmitScopeObjectInit(FuncInfo *funcInfo)
...
+            else
+            {
+                propIds->elements[slot] = Js::Constants::NoProperty;
+            }
+            slot++;
         };
         MapFormalsWithoutRest(pnodeFnc, initArg);
 
-        // initArg assumes the sym is in a slot, but this may not be true for the rest parameter.
+        // If the rest is not the slot - we need to keep that slot.
```
akroshg: typo - if the rest is `in` the slot

lib/Runtime/ByteCode/ByteCodeEmitter.cpp
```diff
@@ -1578,15 +1578,19 @@ void ByteCodeGenerator::EmitScopeObjectInit(FuncInfo *funcInfo)
                     Assert(sym->GetScopeSlot() != Js::Constants::NoProperty && sym->GetScopeSlot() > slot);
                     propIds->elements[slot] = Js::Constants::NoProperty;
                 }
-                slot++;
             }
+            else
```
aneeshdk: Is there a way to do this only in case of pattern and not affect the other cases?

test/es6/destructuring_bugs.js
```diff
@@ -120,6 +120,129 @@ var tests = [
...
+            assert.areEqual(arguments[3], 6);
+            assert.areEqual(k, 21);
+        })(1, {x2:2, x3:3}, [4, 5], 6);
+        
+        (function (x1, {x2, x3}, [x4, x5], x6) {
+            (function() {
+                assert.areEqual(x1 + x2 + x3 + x4 + x5 + x6, 21);
+            })();
+        })(1, {x2:2, x3:3}, [4, 5], 6);
+
```
aneeshdk: Can you add a test case with accessing arguments in the param scope? Also changing value of arguments in the param scope and validating it in the body scope?

test/es6/destructuring_bugs.js
```diff
@@ -120,6 +120,129 @@ var tests = [
...
+            eval("");
+            assert.areEqual(arguments[1].x, 1);
+            assert.areEqual(a, 2);
+        })({}, {x:1}, {}, {}, {}, 2);
+    }
+  },
+  {
+    name: "Destructuring patterns (multiple identifiers in the pattern) at param with arguments/eval at function body",
+    body: function () {
+        (function (x1, {x2, x3}, [x4, x5], x6) {
```
aneeshdk: Lambdas accessing arguments in the body may also be interesting.

aneeshdk:
Rest looks good to me.

lib/Runtime/ByteCode/ByteCodeEmitter.cpp
tcare: It's a bit unclear to me here from just reading the code what this else is meant to handle. Adding something to restrict to the pattern case or assert that would helpful here.

---
In reply to: [58244592](https://github.com/Microsoft/ChakraCore/pull/715#discussion_r58244592) [](ancestors = 58244592)

lib/Runtime/ByteCode/Scope.cpp
```diff
@@ -130,9 +130,10 @@ void Scope::MergeParamAndBodyScopes(ParseNode *pnodeScope, ByteCodeGenerator *by
 
     // Reassign non-formal slot positions. Formals need to keep their slot positions to ensure
     // the argument object works properly. Other symbols need to be reassigned slot positions.
+    // The sym belongs to arguments will use the same slot.
```
tcare: belonging*?

lib/Runtime/ByteCode/ByteCodeEmitter.cpp
akroshg: Either it will be var sym or it will be pattern. Putting explicit assert is for pattern is redundant. GetFormalNext() already has assert.

lib/Runtime/ByteCode/Scope.cpp
akroshg: will do.

lib/Runtime/ByteCode/ByteCodeEmitter.cpp
tcare: Alright. It may be worth putting a small comment.

akroshg: sure - will do.

test/es6/destructuring_bugs.js
akroshg: I'll add that.

test/es6/destructuring_bugs.js
akroshg: will add that.

PR #716, opened 2016-04-01
Change GenByteCode.cmd to allow generating bytecodes from alternately-named binaries in different locations.

dilijev:
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/716)


akroshg:
Change looks good to me - but I am curious what alternate binary you want to use this for?

dilijev:
To support building bytecode from jshost, and as a side effect, it should also support building ch to a different output location.

Issue #608, opened 2016-03-18
TypedArray constructor optimizes construction from array when %ArrayIteratorPrototype%.next is no longer the built-in


ianwjhalliday:
Thanks to @evilpie for catching this looking at PR #561 

```javascript
var iter = [][Symbol.iterator]()
iter.__proto__.next = function() {
    return {done: true};
};

var i32a = new Int32Array([1, 2, 3]);

for (let i = 0; i < i32a.length; i += 1) {
    print(i32a[i]);
}
```

Currently prints 1, 2, 3, but should print nothing.  The Int32Array should be created empty because the array iterator's next() method is now supposed to always return {done: true} on first call.

The optimization legality check @suwc added in #561 needs to be fixed up to include verifying that the %ArrayIteratorPrototype%.next slot is still equal to the built-in method.

suwc:
https://github.com/Microsoft/ChakraCore/pull/635

suwc:
Merged

PR #717, opened 2016-04-01
ES6 Spread over zealous optimization for Arrays ignores user changes to %ArrayIteratorPrototype%.next property

tcare:
This is a similar bug that was reported in #608 and fixed by @suwc for TypedArray in #635.

The core problem is that we do not check for changes to %ArrayIteratorPrototype%.next that could have side effects when optimizing spread item copying. Using a function Suwei added, we attempt to get the value without invoking implicit calls, e.g. a getter. In this case we will fall back to calling the iterator for each element.

I moved out the ArrayIteratorPrototypeHasUserDefinedNext function from TypedArray to JavascriptLibrary since we now use it in JavascriptOperators. I moved it there because all the ArrayIteratorPrototype related code seems to be there. I am open to moving it to JavascriptArray instead if people prefer that.

### DIFF
--- a/lib/Runtime/Language/JavascriptOperators.cpp
+++ b/lib/Runtime/Language/JavascriptOperators.cpp
@@ -1347,12 +1347,11 @@ CommonNumber:
     {
         RecyclableObject* function = GetIteratorFunction(aRight, scriptContext);
         JavascriptMethod method = function->GetEntryPoint();
-        if ((JavascriptArray::Is(aRight) && method == JavascriptArray::EntryInfo::Values.GetOriginalEntryPoint()) ||
-            (TypedArrayBase::Is(aRight) && method == TypedArrayBase::EntryInfo::Values.GetOriginalEntryPoint()))
+        if (((JavascriptArray::Is(aRight) && method == JavascriptArray::EntryInfo::Values.GetOriginalEntryPoint())
+                || (TypedArrayBase::Is(aRight) && method == TypedArrayBase::EntryInfo::Values.GetOriginalEntryPoint()))
+            // We can't optimize away the iterator if the array iterator prototype is user defined.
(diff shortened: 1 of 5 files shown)


tcare:
@abchatra @ianwjhalliday @suwc  please review

akroshg:
Looks good to me.

lib/Runtime/Language/JavascriptOperators.cpp
```diff
@@ -1347,12 +1347,11 @@ namespace Js
     {
         RecyclableObject* function = GetIteratorFunction(aRight, scriptContext);
         JavascriptMethod method = function->GetEntryPoint();
-        if ((JavascriptArray::Is(aRight) && method == JavascriptArray::EntryInfo::Values.GetOriginalEntryPoint()) ||
-            (TypedArrayBase::Is(aRight) && method == TypedArrayBase::EntryInfo::Values.GetOriginalEntryPoint()))
+        if (((JavascriptArray::Is(aRight) && method == JavascriptArray::EntryInfo::Values.GetOriginalEntryPoint())
+                || (TypedArrayBase::Is(aRight) && method == TypedArrayBase::EntryInfo::Values.GetOriginalEntryPoint()))
+            // We can't optimize away the iterator if the array iterator prototype is user defined.
+            && !JavascriptLibrary::ArrayIteratorPrototypeHasUserDefinedNext(scriptContext))
```
abchatra: Does TypedArray dependent on ArrayIteratorPrototype as well?

suwc:
I think `ArrayIteratorPrototypeHasUserDefinedNext` in `JavascriptLibrary` is reasonable.
LGTM.

lib/Runtime/Language/JavascriptOperators.cpp
tcare: Yes as per Suwei's change. Is that what you meant?

PR #718, opened 2016-04-01
Harden Promise code against OOM after catching exception

boingoing:
JavascriptExceptionObject::GetThrownObject returns nullptr if it hits OOM.
We previously did not check for this case which caused assert or AV in
retail builds when a JavascriptError object fails to allocate with OOM.

Made our implementation more robust by catching the case wehre we hit an
OOM and pass undefined to the promise reject handler instead of nullptr in
that case. This avoids potential AV or assert but does ignore the OOM.


boingoing:
@abchatra, @ianwjhalliday  Can you take a look at this?

abchatra:
:+1:

PR #628, opened 2016-03-22
SIMD.js - Fix regalloc to not fold SIMD opnds

nmostafa:
Fixed RegAlloc to not use SIMD opnds from memory if lifetime is spilled.
@LouisLaf , @pleath can either of you please review this ?

PR #720, opened 2016-04-01
SIMD.js - Fix regalloc to not fold SIMD opnds

### LINKED ISSUE #695: SIMD.js - RegAlloc may fold SIMD operands
With register pressure, RegAlloc may fold SIMD operands to be loaded from memory. Since we currently don't ensure 16-byte alignment for SIMD stack syms, this might trigger an AV.

```
// Try folding if there are no registers available
 if (!sym->IsConst() && !this->RegsAvailable(regOpnd->GetType()) && EncoderMD::TryFold(instr, regOpnd))
 {
                    return;
}
```

nmostafa:
Fixed RegAlloc to not use SIMD opnds from memory if lifetime is spilled. Fixes #695


nmostafa:
Instead of #628.

PR #722, opened 2016-04-01
Get the last error code immediately after a Windows API call in IntlEngineInterfaceExtensionObject::EntryIntl_CompareString

goyakin:
In certain cases, END_TEMP_ALLOCATOR would clear the last error code
after the CompareStringEx call, causing us to hit an assert in
JavascriptError::SetErrorMessage since the function expects a failure.


goyakin:

abchatra:
:+1:

PR #723, opened 2016-04-02
Fix optimized element access that incorrectly reports that the instance has no such property

pleath:
Reimplement shortcut in element access on unknown property name. The optimization requires a prototype chain walk to confirm that the property being looked for can't be found on an external object with a deferred type handler. For now the prototype walk is done on every attempted access where no property record exists for the index name. This could be optimized with a scheme similar to the one we use for non-writable properties in the prototype chain, but I'd like to see a needful use case before I saddle us with another registration mechanism. Benchmarks are flat.

### DIFF
--- a/lib/Runtime/Language/JavascriptOperators.cpp
+++ b/lib/Runtime/Language/JavascriptOperators.cpp
@@ -76,7 +76,7 @@ namespace Js
         }
     }
 
-    IndexType GetIndexType(Var indexVar, ScriptContext* scriptContext, uint32* index, PropertyRecord const ** propertyRecord, JavascriptString ** propertyNameString, bool createIfNotFound, bool preferJavascriptStringOverPropertyRecord)
+    IndexType GetIndexType(Var indexVar, ScriptContext* scriptContext, uint32* index, PropertyRecord const ** propertyRecord, JavascriptString ** propertyNameString, bool createIfNotFound, RecyclableObject *object, bool preferJavascriptStringOverPropertyRecord)
     {
         indexVar = JavascriptConversion::ToPrimitive(indexVar, JavascriptHint::HintString, scriptContext);
 
@@ -96,6 +96,11 @@ namespace Js
                 char16 buffer[20];
                 ::_itow_s(indexInt, buffer, sizeof(buffer) / sizeof(char16), 10);
                 charcount_t length = JavascriptString::GetBufferLength(buffer);
+                if (!preferJavascriptStringOverPropertyRecord)
+                {
+                    createIfNotFound = createIfNotFound &&
+                        (!object || !JavascriptOperators::CanShortcutOnUnknownPropertyName(object));
+                }
                 if (createIfNotFound || preferJavascriptStringOverPropertyRecord)
                 {
                     // When preferring JavascriptString objects, just return a PropertyRecord instead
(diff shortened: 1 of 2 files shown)


pleath:
@ianwjhalliday

ianwjhalliday:
Might not have time to look at this until late Friday. If urgent we'll need another reviewer and I can look at it after checkin.

pleath:
Thanks. Friday is fine.

--Paul

lib/Runtime/Language/JavascriptOperators.cpp
```diff
@@ -4932,6 +4930,41 @@ namespace Js
...
+        if (!CanShortcutInstanceOnUnknownPropertyName(instance))
+        {
+            return false;
+        }
+        return CanShortcutPrototypeChainOnUnknownPropertyName(instance->GetPrototype());
+    }
+
+    bool JavascriptOperators::CanShortcutInstanceOnUnknownPropertyName(RecyclableObject *instance)
+    {
+        if (!instance->HasDeferredTypeHandler())
```
ianwjhalliday: Doesn't this preclude the rest of the conditions checked after this early out?

ianwjhalliday: E.g. an instance that is a Proxy will not have a deferred type handler.  So it'll come here and return true instead of returning false below?

---
In reply to: [59070623](https://github.com/Microsoft/ChakraCore/pull/723#discussion_r59070623) [](ancestors = 59070623)

lib/Runtime/Language/JavascriptOperators.cpp
```diff
@@ -4932,6 +4930,41 @@ namespace Js
...
+        return !(instance->IsExternal() || 
+                 JavascriptProxy::Is(instance) ||
+                 (JavascriptFunction::Is(instance) && JavascriptFunction::FromVar(instance)->IsExternalFunction()));
+    }
+
+    bool JavascriptOperators::CanShortcutPrototypeChainOnUnknownPropertyName(RecyclableObject *prototype)
+    {
+        Assert(prototype);
+
+        TypeId typeId;
```
ianwjhalliday: typeId isn't used, remove

lib/Runtime/Language/JavascriptOperators.cpp
pleath: Actually, Proxy is the only case where it doesn't matter what kind of type handler we have. I'll refactor this.

lib/Runtime/Language/JavascriptOperators.cpp
pleath: (No, I need it in the loop condition.)

ianwjhalliday: But it is only assigned in the loop condition and not used later anywhere.

Can't the loop condition change to just `prototype->GetTypeId() != TypeIds_Null`?

pleath: Oh, I see what you’re saying. Sure.

--Paul

lib/Runtime/Language/JavascriptOperators.cpp
```diff
@@ -4932,6 +4930,43 @@ namespace Js
...
+    {
+        if (JavascriptProxy::Is(instance))
+        {
+            return false;
+        }
+        if (!instance->HasDeferredTypeHandler())
+        {
+            return true;
+        }
+        return !(instance->IsExternal() ||
```
ianwjhalliday: I still think this is wrong.  `IsExternal` is orthogonal to `HasDeferredTypeHandler` and I believe external objects will always have non-deferred type handlers, so `IsExternal` implies `!HasDeferredTypeHander` which means this will return true above when `IsExternal` is true.  Is that not the case?

pleath: External objects with deferred type handlers do exist. They’re the case that’s the source of the bug. The DOM has property names that the JS engine hasn’t seen, and if the type handler is deferred then the external object may have one of those unknown property names.

--Paul

ianwjhalliday: Ah.  Is it not possible for a DOM object to still have properties that the current undeferred type has not seen though?  Or do we fill all properties once we undefer the type?  Seems like we couldn't guarantee that the type fully matches the DOM object since the host could add new properties after we have undeferred the type, or do I misunderstand external objects?

pleath:
Re-implemented from a new branch, replaced by https://github.com/Microsoft/ChakraCore/pull/841.

PR #724, opened 2016-04-02
Add new cross-platform APIs to Jsrt

digitalinfinity:
This change introduces the cross-platform API shape for JSRT.
It starts with the following assumptions:
-  Changing the definition of existing APIs is ok, as long as the effective shape doesnt change
   (so it's fine to use new macros etc as long as signature/calling convention/linkage etc doesn't change)
-  Existing APIs which take wchar_t* as a parameter will be exposed only on Windows
-  New variants of these APIs that take in UTF-8 strings will be exposed on *all* platforms

Here is a summary of the changes:

New macros introduced:
```CHAKRA_CALLBACK``` - Calling convention attribute of callback functions consumed by Jsrt APIs. Used in place of Win32 ```CALLBACK``` macro.
```CHAKRA_API``` - Calling convention, linkage and return type of Jsrt APIs (always returns JsErrorCode). Used in place of Win32 ```STDAPI_``` macro.

New types introduced
```ChakraCookie``` - Type of cookies consumed by Jsrt Apis. Used in place of Win32 DWORD_PTR datatype.
```JsSerializedScriptLoadUtf8SourceCallback``` - Callback called to load source code of serialized script encoded as utf8

Following APIs become WIN32 only:
```
JsParseScript
JsParseScriptWithFlags
JsRunScript
JsExperimentalApiRunModule
JsSerializeScript
JsParseSerializedScriptWithCallback
JsRunSerializedScriptWithCallback
JsParseSerializedScript
JsRunSerializedScript
JsGetPropertyIdFromName
JsGetPropertyNameFromId
JsPointerToString
JsStringToPointer
```

New APIs introduced (available on all platforms):
```
JsParseScriptUtf8
JsParseScriptWithAttributesUtf8
JsRunScriptUtf8
JsExperimentalApiRunModuleUtf8
JsRunScriptUtf8
JsExperimentalApiRunModuleUtf8
JsSerializeScriptUtf8
JsParseSerializedScriptUtf8
JsRunSerializedScriptUtf8
JsGetPropertyIdFromNameUtf8
JsGetPropertyNameFromIdUtf8
JsPointerToStringUtf8
JsStringToPointerUtf8
```

Couple notes:
-      All utf8 strings are passed in as char*
-      All string parameters in the new APIs are utf8 (e.g source, url etc).

### DIFF
--- a/lib/Jsrt/ChakraCommon.h
+++ b/lib/Jsrt/ChakraCommon.h
@@ -32,21 +32,39 @@
 #define CHAKRA_API STDAPI_(JsErrorCode)
 
 typedef DWORD_PTR ChakraCookie;
+typedef BYTE* ChakraBytePtr;
 #else
 #include <cstdint>
 
 // SAL compat
 #define _Return_type_success_(x)
 #define _In_
+#define _In_z_
 #define _In_opt_
+#define _Inout_
 #define _Out_
 #define _Out_opt_
+#define _In_reads_(x)
+#define _Pre_maybenull_
+#define _Pre_writable_byte_size_(byteLength)
+#define _Outptr_result_buffer_(byteLength)
+#define _Outptr_result_bytebuffer_(byteLength)
+#define _Outptr_result_z_
+#define _Ret_maybenull_
+#define _Out_writes_to_opt_(byteLength, byteLength2)
 
 // Header macros
-#define CHAKRA_CALLBACK __attribute__((stdcall))
+// REVIEW: What should we do here??
+#ifdef __i386___
+#define CHAKRA_CALLBACK __attribute__((cdecl))
+#else
+#define CHAKRA_CALLBACK
+#endif
+
 #define CHAKRA_API extern "C" JsErrorCode 
 
 typedef uintptr_t ChakraCookie;
+typedef unsigned char* ChakraBytePtr;
 #endif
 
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by digitalinfinity** [comment]
@curtisman @jianchun @ianwjhalliday @Yongqu @liminzhu and anyone else who's interested- please review. Also, this has only the API definitions, no implementations- I'll implement in a future PR since no one consumes these APIs yet.

**Comment by digitalinfinity** [comment]
Updated the PR with the following changes:
1. Replaced use of BYTE* with a typedef (ChakraBytePtr)
2. Stubbed out more SAL on Linux 

Confirmed that the latest change compiles cleanly on Linux

File: lib/Jsrt/ChakraCommon.h [hunk 1]
```diff
@@ -819,13 +891,13 @@
...
     /// </returns>
-    STDAPI_(JsErrorCode)
+    CHAKRA_API
         JsParseScript(
             _In_z_ const wchar_t *script,
             _In_ JsSourceContext sourceContext,
             _In_z_ const wchar_t *sourceUrl,
             _Out_ JsValueRef *result);
-
+
```
dilijev: nit: trailing spaces

dilijev:
Looks great from the code change point of view. I didn't proof the doc comments. I assume they are copied from the non-utf8 APIs with an extra note about the encoding of the string being in utf8.

Since other authors like @agarwal-sandeep might make changes to these APIs aside from within the linux branch, maybe it would be better to make this PR against the master branch so that the new macros propagate which will make merging the changes easier later on.

obastemur:
> Since other authors like @agarwal-sandeep might make changes to these APIs aside from within the linux branch, 

Agree with @dilijev on targeting the master branch for some of the changes. i.e. Most of the work I've been doing on Runtime/types are changing the actual core code for the sake of cross platform / compiler. IMHO, by targeting the linux branch for such changes we let other authors to break linux work unconsciously.

digitalinfinity:
@curtisman any objection to retargeting this PR to master?

curtisman:
@digitalinfinity sounds good.

digitalinfinity:
Closing this PR, using #766 instead (which is targeted at master)

PR #725, opened 2016-04-02
SIMD.js - unpin 1-byte opnds to BL

nmostafa:
This change removes pinning of byte operands to R/EBX in Simd128 lowerer, and makes a copy of the original opnd in Opnd::UseWithNewType()

We initially had this:
```
IR::RegOpnd *reg2 = IR::RegOpnd::New(TyInt32, m_func);
        IR::RegOpnd * tmp = IR::RegOpnd::New(TyInt8, m_func);
        instr->InsertBefore(IR::Instr::New(Js::OpCode::MOV, tmp->UseWithNewType(TyInt32, m_func), reg2, m_func));

        instr->InsertBefore(IR::Instr::New(Js::OpCode::MOVSX, reg2, tmp, m_func));
```

One would expect tmp->UseWithNewType to create a copy of the opnd, but instead it checks if the opnd is in use by an instruction, and if not, it will use it and change its type (tmp will be TyInt32 afterwards). That makes the MOVSX use an TyInt32 copy of tmp causing the regAlloc to not restrict register assignment to byteable regs. 

@LouisLaf , can you take a look ?

### DIFF
--- a/lib/Backend/LowerMDSharedSimd128.cpp
+++ b/lib/Backend/LowerMDSharedSimd128.cpp
@@ -745,18 +745,13 @@ IR::Instr* LowererMD::Simd128LowerLdLane(IR::Instr *instr)
             if (laneType == TyInt8)
             {
                 IR::RegOpnd * tmp = IR::RegOpnd::New(TyInt8, m_func);
-#ifdef _M_IX86
-                tmp->SetReg(RegEBX);
-#else
-                tmp->SetReg(RegRBX);
-#endif
-                newInstr = IR::Instr::New(Js::OpCode::MOV, tmp->UseWithNewType(TyInt32, m_func), dst, m_func);
+                newInstr = IR::Instr::New(Js::OpCode::MOV, tmp, dst, m_func);
                 instr->InsertBefore(newInstr);
+                Legalize(newInstr);
                 newInstr = IR::Instr::New(Js::OpCode::MOVSX, dst, tmp, m_func);
             }
             else
             {
-                // any register can be 2-byte addressable
                 newInstr = IR::Instr::New(Js::OpCode::MOVSXW, dst, dst->UseWithNewType(laneType, m_func), m_func);
             }
         }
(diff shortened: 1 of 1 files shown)


LouisLaf:
I never liked that API, but don't really like your modification either, as we will now create unnecessary copies for opnds that are not being used.
The common case usage for it is:
opnd = newOpnd(type1);
use(opnd);
opnd2 = opnd->UseWithNewType(type2);
use(opnd2);

So I think ideally, the API should be called CopyWithNewType(), and the API should assert that the opnd is not InUse().

This will require you change you code pattern in SYMD.

Nagy, do you mind making this API change?

Thanks.

nmostafa:
I will take care of it. I suggested the assertion in #727.

nmostafa:
@LouisLaf 
> So I think ideally, the API should be called CopyWithNewType(), and the API should assert that the opnd is not InUse().

I assume you mean assert that the opnd _is InUse()_ ? 
It seems existing uses of this API in the lowerer can call it on the first use of the operand. For example, see InsertMove below. It doesn't know if src is used.

The simplest fix here is to keep the API as-is, and I can fix the SIMD lowerer to not use the API (legalizing MOV TyInt8, TyInt32 should be sufficient). 

```C++
IR::Instr *Lowerer::InsertMove(IR::Opnd *dst, IR::Opnd *src, IR::Instr *const insertBeforeInstr, bool generateWriteBarrier)
{
...
    if(TySize[dst->GetType()] < TySize[src->GetType()])
    {
        src = src->UseWithNewType(dst->GetType(), func);
    }
```

LouisLaf:
You're right.  This is probably the better choice.

nmostafa:
@LouisLaf , if it looks good, please sign off. Thanks

nmostafa:
@abchatra , I will merge this into release/1.2. There is no repro for this since this is just a more appropriate fix for a previous jshost bug.

abchatra:
Sure go ahead.

PR #726, opened 2016-04-03
cross platform: Runtime/types Part 2/2

obastemur:


lib/Common/DataStructures/WeakReferenceDictionary.h
```diff
@@ -20,20 +20,19 @@ namespace JsUtil
     class WeakReferenceDictionary: public BaseDictionary<TKey, RecyclerWeakReference<TValue>*, RecyclerNonLeafAllocator, SizePolicy, Comparer, WeakRefValueDictionaryEntry>,
                                    public IWeakReferenceDictionary
     {
-        typedef BaseDictionary<TKey, RecyclerWeakReference<TValue>*, RecyclerNonLeafAllocator, SizePolicy, Comparer, WeakRefValueDictionaryEntry> Base;
-        
+        typedef BaseDictionary<TKey, RecyclerWeakReference<TValue>*, RecyclerNonLeafAllocator, SizePolicy, Comparer, WeakRefValueDictionaryEntry> Base;        
     public:
         WeakReferenceDictionary(Recycler* recycler, int capacity = 0):
-          BaseDictionary(recycler, capacity)
+          BaseDictionary<TKey, RecyclerWeakReference<TValue>*, RecyclerNonLeafAllocator, SizePolicy, Comparer, WeakRefValueDictionaryEntry>(recycler, capacity)
```
digitalinfinity: Can you just call Base(recycler, capacity) here?

lib/Runtime/Types/ES5ArrayTypeHandler.cpp
```diff
@@ -271,25 +279,34 @@ namespace Js
             doneConversion = true;
 #endif
         }
+#ifdef _MSC_VER
         __finally
         {
             if (AbnormalTermination())
```
digitalinfinity: @Yongqu can you review this part of the change please? Can we convert this to using the TryFinally primitive that @jianchun introduced? (That is, do we really need SEH here or can we use just regular C++ exception handling mechanisms)

lib/Runtime/Types/SimpleDictionaryUnorderedTypeHandler.h
```diff
@@ -37,7 +37,14 @@ namespace Js
         SimpleDictionaryUnorderedTypeHandler(Recycler * recycler, int slotCapacity, uint16 inlineSlotCapacity, uint16 offsetOfInlineSlots);
         SimpleDictionaryUnorderedTypeHandler(ScriptContext * scriptContext, SimplePropertyDescriptor* propertyDescriptors, int propertyCount, int slotCapacity, uint16 inlineSlotCapacity, uint16 offsetOfInlineSlots);
         SimpleDictionaryUnorderedTypeHandler(Recycler* recycler, int slotCapacity, int propertyCapacity, uint16 inlineSlotCapacity, uint16 offsetOfInlineSlots);
-        DEFINE_VTABLE_CTOR_NO_REGISTER(SimpleDictionaryUnorderedTypeHandler, SimpleDictionaryTypeHandlerBase);
+
+        // Clang requeries template params are explicitely given. while it breaks the macro below
```
digitalinfinity: I think you can still use the macro if you do the following:
typedef SimpleDictionaryTypeHandlerBase<TPropertyIndex, TMapKey, IsNotExtensibleSupported> Base;
DEFINE_VTABLE_CTOR_NO_REGISTER(SimpleDictionaryUnorderedTypeHandler, Base);

lib/Runtime/Types/ES5ArrayTypeHandler.cpp
obastemur: IMO, TryFinally wouldn't fit here as is. `if (AbnormalTermination())` 

If SEH is not needed `try/catch(...)` should be fine. See https://github.com/Microsoft/ChakraCore/pull/726/files#diff-ffccee5f5c8edef39aeaad0f043b66d8R252

obastemur:
https://github.com/Microsoft/ChakraCore/pull/726/commits/68482b2b4c51f25073ce5cc7e392be553726c049 also fixes https://github.com/Microsoft/ChakraCore/issues/758

Yongqu:
we are using SEH for asm.js to avoid the outofbound check. I am not sure if it's easy to get rid of that SEH usage.

digitalinfinity:
@obastemur please separate out the commits for #758 into its own PR to keep this review focused on Runtime/Types.

obastemur:
> we are using SEH for asm.js to avoid the outofbound check. I am not sure if it's easy to get rid of that SEH usage. I don't know what does FF do for their 64bit asm.js support, should be something similar.

In that case; how about keeping as is and putting this info as a TODO comment there? Looks like getting rid of SEH is curicial for asm.js support on linux.

digitalinfinity:
Actually, @obastemur, Curtis has a better suggestion. Use a destructor to execute the code in the finally part and get rid of the catch block. We still need the __try/__finally for Windows because of the behavior Yong talks about, but we could just have the finally block call the same method as the destructor (and use a flag for the destructor to know whether an exception was thrown or not)

lib/Runtime/Types/ES5ArrayTypeHandler.cpp
jianchun: > IMO, TryFinally wouldn't fit here as is. if (AbnormalTermination())

The [TryFinally](https://github.com/Microsoft/ChakraCore/blob/linux/lib/Common/CommonPal.h#L388) calls `finallyFunc(hasException)`. `hasException` was intended to indicate `AbnormalTermination()` here. Would it work?

digitalinfinity:
Ok, talked to @jianchun and @Yongqu. My mistake, Yong was talking about other places where we use __try/__finally, not in this particular case. Here, we care only about C++ exceptions, the SEH usage for asm.js is not relevant here. In this case, please convert to using the TryFinally pattern that @jianchun introduced. You can replace the AbnormalTermination check with checking the hasException flag.

obastemur:
> the SEH usage for asm.js is not relevant here. 

> please convert to using the TryFinally pattern 

Could you please elaborate why? In that case I don't see a reason to maintain `__try` , `__finally` there. IMO, the `try/catch(...)` option implemented already takes care of the task cross platform. 

BTW; @digitalinfinity thanks for the review!

digitalinfinity:
You're right, I think try/catch should be fine. Although, instead of catch (...), for this case, better to scope to catch (Js::ExceptionBase) (with the rethrow). And you can remove __try/__finally on windows too 😄

obastemur:
@digitalinfinity thanks again. PR is updated.

lib/Runtime/Types/ES5ArrayTypeHandler.cpp
```diff
@@ -271,25 +272,23 @@ namespace Js
             doneConversion = true;
 #endif
         }
-        __finally
+        catch(...)
```
digitalinfinity: Is this commit slightly stale or is there a reason you decided to go with catch(...) instead of catch (Js::ExceptionBase)?

obastemur: It just carries the functionality as is. Previous logic was defined as to run cleanup part no matter what and do not catch anything. Current version does the same. catch everything, cleanup. and throw. So, IMHO there was no point for hosting exception info inside the scope stack.

PR is updated with Js::ExceptionBase.

PR #728, opened 2016-04-04
Fixing a potential OOM point

rajatd:
I stumbled across this piece of code while inestigating another issue. We are calling CloneIfStaticExceptionObject inside a catch. This could potentially throw OOM if the caught exception was an out of memory exception. Changing the code here to conform to the pattern used at other places - save the exception object in a local and take action on that local outside the catch. This reduces the chances of an OOM being thrown as the stack would have been unwound when we get outside the catch block.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/728)


akroshg:
Looks good to me.

PR #729, opened 2016-04-04
OS4927797: TypeofElem fastpath mishandling out-of-range scenario

suwc:
Fastpaths for opcode 'TypeofElem' mistakenly return 'number' for certain out-of-range situations.
Fix by adding check 'undefined' for return values in the fastpaths.

### DIFF
--- a/lib/Runtime/Language/JavascriptOperators.cpp
+++ b/lib/Runtime/Language/JavascriptOperators.cpp
@@ -2458,8 +2458,11 @@ CommonNumber:
             Var member;
 
             // If the item is found in the array own body, then it is a number
-            if (JavascriptOperators::GetOwnItem(object, index, &member, scriptContext))
+            if (JavascriptOperators::GetOwnItem(object, index, &member, scriptContext)
+                && !JavascriptOperators::IsUndefined(member))
+            {
                 return TRUE;
+            }
         }
         return FALSE;
     }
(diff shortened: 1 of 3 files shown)


suwc:
@goyakin @boingoing @abchatra could you take a quick look?

lib/Runtime/Language/JavascriptOperators.cpp
```diff
@@ -2459,7 +2459,7 @@ namespace Js
             Var member;
 
             // If the item is found in the array own body, then it is a number
-            if (JavascriptOperators::GetOwnItem(object, index, &member, scriptContext))
+            if (JavascriptOperators::GetOwnItem(object, index, &member, scriptContext) && member != scriptContext->GetLibrary()->GetUndefined())
```
boingoing: Can you use JavascriptOperators::IsUndefined here?

goyakin: As far as I now, typed arrays can't contain `undefined`. Instead of checking if the value is `undefined` here, I think `JavascriptOperators::GetOwnItem` should return `false` when the item doesn't exist.

It might be OK to keep it the way it currently is for the `1.2` milestone and look into returning `false` later since `GetItem` (called by `GetOwnItem`) is used in a bunch of places. @abchatra, what do you think?

suwc: @boingoing, will do.
@goyakin, I'd like to open a non-bug-non-1.2 issue to find out if it makes sense to change the behavior of GetItem(). In the meantime, I think it is prudent to avoid that in 1.2

abchatra: Can you move this check to the separate next line? It is bit confusing for me.   
This is safe fix for 1.2  If GetItem returns false for undefined, other code paths might walk the prototype chain, which is not desirable for typedarray.

abchatra:
:+1:

suwc:
Addressed review comments and ready for merge...

PR #731, opened 2016-04-04
Fixing debug-only race condition during RecordNativeCodeSize

satheeshravi:
The VirtualQuery in CodeGenWorkItem is unprotected without lock. Hence the
assert that is checking for the protect state of the page might not always
be true.

Fix:
Moved the VirtualQuery to EmitBuffer under a lock. This query is still
debug-only.


satheeshravi:
@digitalinfinity : can you please take a look?

digitalinfinity:
LGTM

satheeshravi:
Thanks Hitesh.

PR #734, opened 2016-04-04
Fix one use of JavascriptExceptionObject::GetThrownObject in the async spawn function which I missed during an earlier round of cleanup

boingoing:
This is fault-injection case which causes us to call a rejection handler
with nullptr instead of Javascript undefined value.

### DIFF
--- a/lib/Runtime/Language/JavascriptOperators.cpp
+++ b/lib/Runtime/Language/JavascriptOperators.cpp
@@ -9492,7 +9492,7 @@ CommonNumber:
 
         if (e != nullptr)
         {
-            reject->GetEntryPoint()(reject, CallInfo(CallFlags_Value, 2), library->GetUndefined(), e->GetThrownObject(scriptContext));
+            JavascriptPromise::TryRejectWithExceptionObject(e, reject, scriptContext);
         }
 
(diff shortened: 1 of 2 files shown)


lib/Runtime/Language/JavascriptOperators.cpp
```diff
@@ -9489,7 +9489,7 @@ namespace Js
 
         if (e != nullptr)
         {
-            reject->GetEntryPoint()(reject, CallInfo(CallFlags_Value, 2), library->GetUndefined(), e->GetThrownObject(scriptContext));
+            JavascriptPromise::TryRejectWithExceptionObject(e, reject, scriptContext);
```
akroshg: Many places in the file we call the TryRejectWithExceptionObject under catch handler - which is the not correct way to do that. Current change looks good. Make sure you re-look other places and potentially fix them.

boingoing: Thanks for the heads-up. I have fixed all the other places TryRejectWithExceptionObject is called. Can you take a look?

boingoing:

akroshg:
Looks good to me

PR #735, opened 2016-04-05
[wasm] Fixed return type check bug

commonlisp:
Previously, the type of the return was not passed up.
This precluded expressions with returns in the body.

### DIFF
--- a/lib/WasmReader/WasmByteCodeGenerator.cpp
+++ b/lib/WasmReader/WasmByteCodeGenerator.cpp
@@ -1068,6 +1068,7 @@ WasmBytecodeGenerator::GetConstReg(T constVal)
 EmitInfo
 WasmBytecodeGenerator::EmitReturnExpr(EmitInfo *lastStmtExprInfo)
 {
+    EmitInfo retExprInfo;
     if (m_funcInfo->GetResultType() == WasmTypes::Void)
     {
         // TODO (michhol): consider moving off explicit 0 for return reg
@@ -1075,7 +1076,6 @@ WasmBytecodeGenerator::EmitReturnExpr(EmitInfo *lastStmtExprInfo)
     }
     else
     {
-        EmitInfo retExprInfo;
         if (lastStmtExprInfo)
         {
             retExprInfo = *lastStmtExprInfo;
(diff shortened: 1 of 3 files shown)


lib/WasmReader/WasmByteCodeGenerator.cpp
```diff
@@ -1107,14 +1107,9 @@ WasmBytecodeGenerator::EmitReturnExpr(EmitInfo *lastStmtExprInfo)
...
         m_writer.Conv(retOp, 0, retExprInfo.location);
-        if (!lastStmtExprInfo)
-        {
-            ReleaseLocation(&retExprInfo);
-        }
     }
     m_writer.AsmBr(m_funcInfo->GetExitLabel());
-
-    return EmitInfo();
+    return retExprInfo;
```
MikeHolman: I believe this isn't valid.
>The  `nop`, `if`, `br`, `br_if`, and  `return` constructs do not yield values
From: https://github.com/WebAssembly/design/blob/master/AstSemantics.md#yielding-values-from-control-constructs

It does prevent your example from working, but your example could easily be changed by moving one parenthesis,  like so:
    (func  (result i32) (if (i32.const 1) (return (i32.const 2))) (return (i32.const 3)))

commonlisp: Very good point. Somewhat misleading was the fact that the original example actually type checked under spec. I'll have to double check what is going on in spec. Closing.

commonlisp: It seems that sexpr-wasm also passes the original example. Moreover, labels.wast's misc2 also exhibits internal return, which surprisingly type checks.

commonlisp: I confirm: in both the spec and sexpr-wasm, return does yield a value via desugaring into a value-yielding form of br. This seems to be a discrepancy between the design docs, specs, and implementation. I posted an issue on spec https://github.com/WebAssembly/spec/issues/272 to get some clarity.

commonlisp:
Hmm, the discussion in spec suggests that "do not yield value" refers to the operational semantics and not the static validation so both the original if_else example and spec test case ought to type check. Previously, it was failing in the GenerateFunction() return type check.

commonlisp:
That said, still may need to change some things. Internal returns should type check because any unconditional control transfer should be polymorphic, so as to allow (block (if_else (i32.const 1) (br 0) (i32.const 2))), which we do type check, not because the body type happens to match the other branch value.

MikeHolman:
Thanks for following up. I think we can achieve desired semantics by having `return` and `br` yield an `Unreachable` type, and at merge points `Unreachable` is dominated by all other types. This is somewhat similar to what we are doing now with `Void`.
The difficulty I see is that we will need to do slightly better tracking of labels (which we were going to have to do anyway once we have `br` with value) to be able to merge correctly at the end of blocks.

commonlisp:
That sounds like a good approach. I'm putting together a batch of test cases exploring the various edge cases where we are incorrectly rejecting functions and also where it is correct to reject per the spec. Will be revisiting this shortly.

PR #736, opened 2016-04-05
Profiler for call with flags does not check the call flags to see if the call is a new call

boingoing:
This only really happens when we are doing a super call in a derived class
constructor. That call is using ProfiledCallIFlags opcode to pass the new
flag and new target flags in the call info. When we get to the helper,
OP_ProfileCallCommon, we need to check the flags to see if we have the new
flag and pass that to RecordCallSiteInfo.

Previously we ignored the flags and always passed false for
isConstructorCall. This could lead to an assert as in the attached test
case.

### DIFF
--- a/lib/Runtime/Language/InterpreterStackFrame.cpp
+++ b/lib/Runtime/Language/InterpreterStackFrame.cpp
@@ -3761,7 +3761,8 @@ namespace Js
         DynamicProfileInfo * dynamicProfileInfo = functionBody->GetDynamicProfileInfo();
         FunctionInfo* functionInfo = function->GetTypeId() == TypeIds_Function?
             JavascriptFunction::FromVar(function)->GetFunctionInfo() : nullptr;
-        dynamicProfileInfo->RecordCallSiteInfo(functionBody, profileId, functionInfo, functionInfo ? static_cast<JavascriptFunction*>(function) : nullptr, playout->ArgCount, false, inlineCacheIndex);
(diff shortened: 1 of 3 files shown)


boingoing:
@abchatra Can you take a look?

lib/Runtime/Language/InterpreterStackFrame.cpp
```diff
@@ -3761,7 +3761,8 @@ namespace Js
         DynamicProfileInfo * dynamicProfileInfo = functionBody->GetDynamicProfileInfo();
         FunctionInfo* functionInfo = function->GetTypeId() == TypeIds_Function?
             JavascriptFunction::FromVar(function)->GetFunctionInfo() : nullptr;
-        dynamicProfileInfo->RecordCallSiteInfo(functionBody, profileId, functionInfo, functionInfo ? static_cast<JavascriptFunction*>(function) : nullptr, playout->ArgCount, false, inlineCacheIndex);
+        bool isConstructorCall = (CallFlags_New & flags) == CallFlags_New;
+        dynamicProfileInfo->RecordCallSiteInfo(functionBody, profileId, functionInfo, functionInfo ? static_cast<JavascriptFunction*>(function) : nullptr, playout->ArgCount, isConstructorCall, inlineCacheIndex);
```
boingoing: Yes, looks like simple jit and full jit were already doing this right.

abchatra:
👍

PR #738, opened 2016-04-05
Class extends clause has different source locations between defer/non-defer parse modes

tcare:
Tracked in OS 7030846.

When emitting a class statement with an extends clause, we made the assumption that the start and end of the bytecode for that statement happened in the same place in defer and non-defer parse modes. In fact, it is possible to have the load happen in different locations, e.g. if we cache a previous load's register location in non-defer parse mode and have to load from a local slot in defer parse mode.

The fix is to set the statement location to the InitClass bytecode.

### DIFF
--- a/lib/Runtime/ByteCode/ByteCodeEmitter.cpp
+++ b/lib/Runtime/ByteCode/ByteCodeEmitter.cpp
@@ -10371,9 +10371,9 @@ void Emit(ParseNode *pnode, ByteCodeGenerator *byteCodeGenerator, FuncInfo *func
         // Extends
         if (pnode->sxClass.pnodeExtends)
         {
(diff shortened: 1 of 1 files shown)


tcare:
@akroshg @boingoing please review.

I considered adding a test for this, but the differences in the bytecode would be fragile.

akroshg:
Change looks good to me.
There are some cases added for line and column difference under `test\error\validate_line_column.js`
There is one test for `extends `already.

boingoing:
Looks good to me, what's the symptom of this bug? Can you add a unit test that throws an exception and checks the column number?

tcare:
We would get a different line number and offset in defer and non-defer parse mode.

As we discussed offline, the test mentioned by Akrosh will cover this.

PR #739, opened 2016-04-05
Fixing bug with lossy type spec from uint32 to int32

rajatd:
Track lossyness separately for src1 and src2 of an instruction so that we
dont miss aggressive int type spec bailouts on one source because the
other source was being lossy type spec'ed.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/739)


rajatd:
@LouisLaf , could you take a look?

LouisLaf:
:+1:

PR #744, opened 2016-04-05
Trivial CallFlags issue in jitted code

pleath:
Fix an inconsistency between interpreter and JIT in the passing of call flags. Constructor calls with spread arguments were getting CallFlags_Value in jitted code but not interpreter, which exposed some weirdness in the Error constructor. (Suwei and/or Taylor are looking into that issue.)


pleath:
@tcare

tcare:
Looks good.

PR #746, opened 2016-04-06
Fix inline cache assert with BoundFunction

pleath:
When we create a bound function, we set its prototype to match the prototype of the target function. Typically this is the Function prototype. But if it isn't, we can't set the prototype of the new instance without changing the prototype of all the other bound functions that are sharing the type. If the type is shared, and the new prototype doesn't match, create a new type for the new instance.


pleath:
@abchatra, this is the one we discussed this afternoon.

abchatra:
:+1:

PR #747, opened 2016-04-06
Fix for the pid ref issues with default parameters

aneeshdk:
When param and body scopes are merged, when a formal is referenced in the
body we were using an additional argument named maxScopeId to look up
until the param scope block. This logic was causing issues when we try to
insert new references into the pid ref linked list. This changelist gets
rid of the requirement of maxScopeId by adding additional pid refs for
each param scope symbol into the body before we start parsing the body.

### DIFF
--- a/lib/Parser/Parse.cpp
+++ b/lib/Parser/Parse.cpp
@@ -820,31 +820,7 @@ Symbol* Parser::AddDeclForPid(ParseNodePtr pnode, IdentPtr pid, SymbolType symbo
         blockInfo = blockInfo->pBlockInfoOuter;
     }
 
-    int maxScopeId = blockInfo->pnodeBlock->sxBlock.blockId;
-
-    // The body of catch may have let declared variable. In the case of pattern, found at catch parameter level,
-    // we need to search the duplication at that scope level as well - thus extending the scope lookup range.
-    if (IsES6DestructuringEnabled()
-        && fBlockScope
-        && blockInfo->pBlockInfoOuter != nullptr
-        && blockInfo->pBlockInfoOuter->pnodeBlock->sxBlock.scope != nullptr
-        && blockInfo->pBlockInfoOuter->pnodeBlock->sxBlock.scope->GetScopeType() == ScopeType_CatchParamPattern)
-    {
-        maxScopeId = blockInfo->pBlockInfoOuter->pnodeBlock->sxBlock.blockId;
-    }
-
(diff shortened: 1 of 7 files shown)


aneeshdk:
@pleath @tcare @akroshg Can you please take a look?

lib/Parser/Parse.cpp
```diff
@@ -7453,6 +7441,14 @@ void Parser::TransformAsyncFncDeclAST(ParseNodePtr *pnodeBody, bool fLambda)
     // meaning post-parsing that won't match the actual parameter list of the generator.
     pnodeFncGenerator->sxFnc.SetHasNonSimpleParameterList(hasNonSimpleParameterList);
 
+    // We always merge the param scope and body scope for async methods right now.
+    // So adding an additional reference for the param symbols to the body.
+    paramScope->ForEachSymbol([this] (Symbol* param)
```
pleath: Just curious -- is there a specific reason we always merge these scopes? Can async methods not have default parameter values? Or is the support a TODO?

lib/Runtime/ByteCode/Scope.cpp
```diff
@@ -124,8 +124,10 @@ void Scope::MergeParamAndBodyScopes(ParseNode *pnodeScope, ByteCodeGenerator *by
             return;
         }
 
-        Assert(paramScope->m_symList == nullptr || paramScope->FindLocalSymbol(sym->GetName()) == nullptr);
-        paramScope->AddNewSymbol(sym);
+        if (paramScope->FindLocalSymbol(sym->GetName()) == nullptr)
```
pleath: Can you explain the need to add arguments to both param and body scopes? There can only be one binding. Can we have it only in the param scope?

lib/Parser/Parse.cpp
aneeshdk: That support is TODO.

lib/Runtime/ByteCode/Scope.cpp
aneeshdk: Updated this implementation. Can you please take a look at the new commit?

pleath:
Some nice simplification. LGTM.

tcare:
After this change, one of the idioms I have used of there only being one symbol object referenced doesn't hold true any more. Would you mind adding a note somewhere that we expect to see the same symbol object used twice in the stack in this case?

Otherwise LGTM too. Nice elegant solution.

aneeshdk:
@tcare Added a comment

tcare:
Perfect, thanks!

PR #748, opened 2016-04-06
Fixes `Error.stack` is being `undefined` when `Error.stackTraceLimit==0`

### LINKED ISSUE #589: Error.stack undefined when Error.stackTraceLimit == 0

obastemur:
Fixes #589 See https://github.com/Microsoft/ChakraCore/pull/655

### DIFF
--- a/lib/Runtime/Language/JavascriptExceptionOperators.cpp
+++ b/lib/Runtime/Language/JavascriptExceptionOperators.cpp
@@ -682,52 +682,52 @@ namespace Js
         // NOTE Don't set the throwing exception here, because we might need to box it and will cause a nested stack walker
         // instead, return it to be set in WalkStackForExceptionContext
 
-        if (stackCrawlLimit == 0)
-        {
-            return caller;
-        }
-
-        const bool crawlStackForWER = CrawlStackForWER(scriptContext);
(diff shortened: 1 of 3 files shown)


obastemur:
@jianchun tests are failed after editing `Dump(TrimStackTracePath(e.stack) || e);`

Is there a test case for throwing non Error object ? That could be the reason. In that case, no point updating that line ?

jianchun:
Yes, test fails because one case the getter throws a String. Just need to update the baseline:

```
--Throw in getter
** Custom stackTraceLimit getter Called, throw
My error in custom stackTraceLimit getter
```

Replace the last line with `undefined`. Now the tests cover both `throw new Error` and `throw 'String'` cases, and the later case `e.stack === undefined` (same as V8).

The `|| e` part in `TrimStackTracePath(e.stack) || e` was to work around `e.stack === undefined` case. Removing `|| e` makes the test to cover your fix.

obastemur:
@jianchun thanks for the details. PR is updated accordingly.

jianchun:
Thank you @obastemur!

PR #750, opened 2016-04-06
Debugger support for split param/body scope case during default.

akroshg:
For the split scope we were still using the older way of getting information from the function body when we walk the slot array walker. However that is not true as the propertyidslotarraycontainer was meant for slots for body's variable and also the regslotcontainer will not have right information as well.
Fixed that by
The variable which are copied from param scope to body scope (due to split scope) will be put regslotcontainer if needed.
While walking we need to distinguish that the scope is split and we need to make use of DebuggerScope.
Also the paramscope could be activation object so we need to have DiagParamScopeInObject to distinguish while locals walking.
Linear scan gave the problem as for the split case the formals are duplicated as variables. Our intention was to restore only formals, so I fixed that by limiting the loop till the formalsCount only.

### DIFF
--- a/lib/Backend/LinearScan.cpp
+++ b/lib/Backend/LinearScan.cpp
@@ -1626,7 +1626,9 @@ LinearScan::FillBailOutRecord(IR::Instr * instr)
 
         if (hasFormalArgs)
         {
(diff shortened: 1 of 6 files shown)


akroshg:
@aneeshdk @abchatra @agarwal-sandeep  please review.

agarwal-sandeep:
LGTM but @aneeshdk should also take a look.

aneeshdk:
Looks good

PR #751, opened 2016-04-06
Add a cast when using scoped enum as an array size.

Rastaban:
This is not allowed in standard C++, and will be enforced in future relases of MSVC.

There is already casts when the array is indexed, so no additional changes
are needed at this time.


dilijev:
LGTM

PR #753, opened 2016-04-06
should only do debug reparse if we are actually reparsing

MikeHolman:
In some cases this was causing us to disable asm.js even in SourceRundown mode, because we were incorrectly identifying the function as being debug reparsed, when it was really just being defer parsed.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/753)

### DIFF
--- a/lib/Runtime/Base/FunctionBody.cpp
+++ b/lib/Runtime/Base/FunctionBody.cpp
@@ -1656,10 +1656,7 @@ namespace Js
         Recycler* recycler = this->m_scriptContext->GetRecycler();
         propertyRecordList = RecyclerNew(recycler, Js::PropertyRecordList, recycler);
 
-        bool isDebugReparse = m_scriptContext->IsScriptContextInSourceRundownOrDebugMode() && !this->GetUtf8SourceInfo()->GetIsLibraryCode();
-        bool isAsmJsReparse = false;
-        bool isReparse = isDebugReparse;
-
+        bool isDebugOrAsmJsReparse = false;
         FunctionBody* funcBody = nullptr;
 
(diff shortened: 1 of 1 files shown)


MikeHolman:
@agarwal-sandeep, can you please take a look?

lib/Runtime/Base/FunctionBody.cpp
```diff
@@ -1656,10 +1656,7 @@ namespace Js
         Recycler* recycler = this->m_scriptContext->GetRecycler();
         propertyRecordList = RecyclerNew(recycler, Js::PropertyRecordList, recycler);
 
-        bool isDebugReparse = m_scriptContext->IsScriptContextInSourceRundownOrDebugMode() && !this->GetUtf8SourceInfo()->GetIsLibraryCode();
-        bool isAsmJsReparse = false;
-        bool isReparse = isDebugReparse;
-
+        bool isReparse = false;
```
agarwal-sandeep: Lets rename this to _isDebugOrAsmJsReparse_ to be explicit

agarwal-sandeep:
Except rename suggestion LGTM.

abchatra:
Can you please add the bug or issue innformation here? Is this for 1.2?

MikeHolman:
@abchatra Yes, this is for 1.2. Internal tracking bug is OS: 5393978.

Repro steps:
1. Go to https://dev.windows.com/en-us/microsoft-edge/testdrive/demos/chess/
2. Open F12
3. Disconnect debugger 
4. Refresh the page

Message "asm.js compilation succeeded. Successfully compiled asm.js code (total compilation time XXXms)" is expected but not seen in the F12 console.

PR #754, opened 2016-04-06
Do not unnecessarily rescan a string for a literal in a RegExp disjunction

goyakin:
Assume we have the RegExp /<(foo|bar)/ and the input string "<0bar<0bar<0bar".
When we try to match the string, we first scan it fully for "foo", but can't
find it. Then we scan for "bar" and find it at index 2. However, since there is
no '<' right before it, we continue with our search. We currently do the same
thing two more times starting at indexes 7 and 12 (since those are where the
"bar"s are), each time scanning the rest of the string fully for "foo".

However, if we cache the furthest offsets we tried, we can skip the searches
for "foo" after the first time.

This change has no effect on the general perf.

VSO bug: [4540543](https://microsoft.visualstudio.com/web/wi.aspx?pcguid=cb55739e-4afe-46a3-970f-1b49d8ee7564&id=4540543)
jQuery issue: https://github.com/jquery/jquery/issues/2563

### DIFF
--- a/lib/Parser/RegexRuntime.cpp
+++ b/lib/Parser/RegexRuntime.cpp
@@ -2130,6 +2130,7 @@ namespace UnifiedRegex
 
         if (matcher.literalNextSyncInputOffsets == nullptr)
         {
+            Assert(numLiterals <= MaxNumSyncLiterals);
             matcher.literalNextSyncInputOffsets =
                 RecyclerNewArrayLeaf(matcher.recycler, CharCount, ScannersMixin::MaxNumSyncLiterals);
         }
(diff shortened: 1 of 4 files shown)


goyakin:
@abchatra & @akroshg, could you take a look?

goyakin:
Ping...

lib/Parser/RegexRuntime.cpp
```diff
@@ -2128,9 +2128,28 @@ namespace UnifiedRegex
         int besti = -1;
         CharCount bestMatchOffset = 0;
 
+        if (matcher.literalNextSyncInputOffsets == nullptr)
+        {
+            matcher.literalNextSyncInputOffsets =
+                RecyclerNewArrayLeaf(matcher.recycler, CharCount, ScannersMixin::MaxNumSyncLiterals);
+        }
+        CharCount* literalNextSyncInputOffsets = matcher.literalNextSyncInputOffsets;
```
akroshg: Please add the assert here for numLiterals is lesser than MaxNumSincLiterals.
Rest looks good to me.

PR #755, opened 2016-04-06
Simple fix to stack closures in jitted loop bodies

pleath:
In jitted loop bodies, we may have stack-allocated closures but not create stack-allocated functions inside block scopes. Let BoxScopeSlots call BoxStackFunction for functions not on stack to make sure that any stack closures get boxed.


rajatd:
Looks good to me.

pleath:
Thanks.

--Paul

PR #761, opened 2016-04-07
handle stack overflow in case OS is in low resource state

leirocks:


leirocks:
@curtisman

lib/Common/Core/FaultInjection.cpp
```diff
@@ -1398,11 +1405,22 @@ namespace Js
...
         }
 
-        struct AutoValue {
-            AutoValue() { inExceptionHandler = true; }
-            ~AutoValue() { inExceptionHandler = false; }
-        } autoVal;
-        FaultInjection::Global.FaultInjectionAnalyzeException(ExceptionInfo);
+        inExceptionHandler = true;
+
+        __try
```
obastemur: can it be handled with standard try, catch ?

i.e. if you use try/catch instead, it wouldn't catch stack overflow ?

obastemur: The reason I asked because of the cross platform work. This will break the build.

leirocks: the purpose of the code here is to catch hard stack overflow, mostly the OS is out of resource. it does not throw a C++ exception so c++ try/catch won't work. and c++ try/catch itself (the CRT code) can also cause the stack overflow to happen.

On Linux this might end up with a segment fault, not sure how Linux handle this. can we just #ifdef out this part of code? what's the specific Macro for Linux build?

obastemur: IMHO 'ifdef out' sounds like the best option then. Thanks!

leirocks:
@digitalinfinity

digitalinfinity:
:+1:

PR #762, opened 2016-04-07
Fix for issue with using this object in the param scope of a lambda function

aneeshdk:
When there are function definitions in the param scope of lambda functions with split scope, those functions belong to the param scope. When we were doing the look up for a non-lambda function, like in the case of looking up for this object, we were always starting from the body which was causing issues.


aneeshdk:
@ianwjhalliday @tcare @pleath @akroshg Can you please take a look?

akroshg:
Looks good to me.

pleath:
LGTMT.

PR #763, opened 2016-04-07
Fix exception path when excpeption is thrown while processing jit transfer data

rajatd:
After generating the jitted code for a function, we pin type references and install property guards on the entryPointInfo. If either one of these throws (OOM/SO, for example), we execute an object destructor that calls a function that allocates a new entryPointInfo. This is a problem if we were throwing an OOM exception. Apart from exceptions, this function can be called when we lazily bailout - in that case, it is ok to create a new entry point info and generate jitted code for it. In fact, we have to do this, otherwise we'll keep bailing out. For OOM scenarios, we dont really have to create a new default entry point info- we could just change the entry point to CheckCodeGenThunk. If the exception was caught and the function was called again, it'll re-use the jitted code and attempt to pin type refs and install property guards again.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/763)

### DIFF
--- a/lib/Runtime/Base/FunctionBody.cpp
+++ b/lib/Runtime/Base/FunctionBody.cpp
@@ -8189,6 +8189,9 @@ namespace Js
             } autoCleanup(this);
 
             ScriptContext* scriptContext = GetScriptContext();
+
+            // If we start transferring more data from the jit thread, corresponding fields on the
+            // entryPointInfo should be rolled back in EntryPointInfo::OnNativeCodeInstallFailure
             PinTypeRefs(scriptContext);
             InstallGuards(scriptContext);
             FreeJitTransferData();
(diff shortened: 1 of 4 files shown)


rajatd:
@LouisLaf @digitalinfinity , could you guys take a look?

lib/Runtime/Base/FunctionBody.cpp
```diff
@@ -8913,7 +8905,9 @@ namespace Js
     }
 
 #if ENABLE_NATIVE_CODEGEN
-    void EntryPointInfo::ResetOnNativeCodeInstallFailure()
+    // This function needs review when we enable lazy bailouts- 
+    // Is calling Reset enough? Does Reset sufficiently resets the state of the entryPointInfo?
+    void EntryPointInfo::ResetOnLazyBailoutFailure()
     {
```
LouisLaf: Should we have an assert here?

LouisLaf:
Rest looks goof.

PR #764, opened 2016-04-07
Clean up logic in the Error constructors. 

pleath:
Specifically, a. prevent the base Error ctor entry point from wiping out new.target before calling the helper that does the construction work, b. get rid of the reliance on the CallFlags_Value flag.

### DIFF
--- a/lib/Runtime/Library/JavascriptError.cpp
+++ b/lib/Runtime/Library/JavascriptError.cpp
@@ -53,19 +53,16 @@ namespace Js
         SetEnumerable(propertyId, false);
     }
 
-    Var JavascriptError::NewInstance(RecyclableObject* function, JavascriptError* pError, CallInfo callInfo, Arguments args)
+    Var JavascriptError::NewInstance(RecyclableObject* function, JavascriptError* pError, CallInfo callInfo, Var newTarget, Var message)
     {
         ScriptContext* scriptContext = function->GetScriptContext();
 
-        AssertMsg(args.Info.Count > 0, "Should always have implicit 'this'");
-
-        Var newTarget = callInfo.Flags & CallFlags_NewTarget ? args.Values[args.Info.Count] : args[0];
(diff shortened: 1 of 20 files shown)


pleath:
@boingoing @suwc @abchatra This is the Error constructor cleanup we discussed.

lib/Runtime/Library/JavascriptError.cpp
```diff
@@ -53,19 +53,16 @@ namespace Js
...
-    Var JavascriptError::NewInstance(RecyclableObject* function, JavascriptError* pError, CallInfo callInfo, Arguments args)
+    Var JavascriptError::NewInstance(RecyclableObject* function, JavascriptError* pError, CallInfo callInfo, Var newTarget, Var message)
     {
         ScriptContext* scriptContext = function->GetScriptContext();
 
-        AssertMsg(args.Info.Count > 0, "Should always have implicit 'this'");
-
-        Var newTarget = callInfo.Flags & CallFlags_NewTarget ? args.Values[args.Info.Count] : args[0];
-        bool isCtorSuperCall = (callInfo.Flags & CallFlags_New) && (callInfo.Flags & (CallFlags_Value|CallFlags_NewTarget)) && newTarget != nullptr && RecyclableObject::Is(newTarget);
+        bool isCtorSuperCall = (callInfo.Flags & (CallFlags_New|CallFlags_NewTarget)) == (CallFlags_New|CallFlags_NewTarget) && newTarget != nullptr && !JavascriptOperators::IsUndefined(newTarget);
```
abchatra: Why is this not a check for RecyclableObject::Is as everywhere else?

abchatra:
:+1:

lib/Runtime/Library/JavascriptError.cpp
pleath: Good point. @boingoing we discussed the fact that this should be checking for undefined. Should we change the check in the other built-ins? RecyclableObject::Is will return false for tagged numbers but true for undefined.

suwc:
LGTM

pleath:
@boingoing Updated the new.target checks in all the subclassable built-ins for consistency.

boingoing:
Nice, looks good to me

pleath:
Thanks.

--Paul

PR #768, opened 2016-04-08
Module script sources should be visible in the debugger

akroshg:
Scipts within modules were not registered to the debugger. This work enables
them to register them properly. The dwHostSourceContext was already setup
correctly, the work remaining is to create the proper sourceContextInfo
and link them to utf8sourceinfo. As we create the correct
sourceContextInfo the module code exposed the inliner bug (earlier we
were not inlining - after current work we are now) - fixed that as well.

Note : Locals experience is also not optimum for the module - but that
will be tackled in the later stages.

### DIFF
--- a/lib/Runtime/Base/ScriptContext.cpp
+++ b/lib/Runtime/Base/ScriptContext.cpp
@@ -1627,7 +1627,9 @@ namespace Js
             // We do not own the memory passed into DefaultLoadScriptUtf8. We need to save it so we copy the memory.
             if (*ppSourceInfo == nullptr)
             {
-                *ppSourceInfo = Utf8SourceInfo::New(this, script, parser->GetSourceIchLim(), cb, pSrcInfo, isLibraryCode);
+                // the 'length' here is not correct - we will get the length from the parser - however parser hasn't done yet.
+                // Once the parser is done we will update the utf8sourceinfo's lenght correctly with parser's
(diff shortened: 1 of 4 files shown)


akroshg:
@boingoing @agarwal-sandeep @abchatra

lib/Runtime/Base/ScriptContext.cpp
```diff
@@ -1627,7 +1627,7 @@ namespace Js
             // We do not own the memory passed into DefaultLoadScriptUtf8. We need to save it so we copy the memory.
             if (*ppSourceInfo == nullptr)
             {
-                *ppSourceInfo = Utf8SourceInfo::New(this, script, parser->GetSourceIchLim(), cb, pSrcInfo, isLibraryCode);
```
agarwal-sandeep: Why is a bug? if not length is always guaranteed to be same as parser->GetSourceIchLim()? can we assert

agarwal-sandeep:
LGTM

lib/Runtime/Base/ScriptContext.cpp
akroshg: It looks like this is a bug - parse->GetSourceIchLim was returning random numbers. may be module was not doing something right to have that? @Yongqu - do you know about this ? the line below we use the length but here we were using GetSourcveIchLim.

boingoing:
Cool, looks good to me

PR #770, opened 2016-04-08
OS:4899199 If exception thrown while evalutating arguments to RecyclerNewEnumClass, we can end up tracking garbage object

boingoing:
Fault injection introduced OOM at DynamicType::New but we already started
tracking the allocation for the constructor.

Fix is simple, just evaluate the arguments before executing the macro.

### DIFF
--- a/lib/Runtime/Library/JavascriptLibrary.cpp
+++ b/lib/Runtime/Library/JavascriptLibrary.cpp
@@ -4528,9 +4528,9 @@ namespace Js
         ConstructorCache* ctorCache = ((functionInfo->GetAttributes() & FunctionInfo::Attributes::SkipDefaultNewObject) != 0) ?
             this->builtInConstructorCache : &ConstructorCache::DefaultInstance;
 
-        return RecyclerNewEnumClass(this->GetRecycler(), EnumFunctionClass, RuntimeFunction,
-            DynamicType::New(scriptContext, TypeIds_Function, prototype, functionInfo->GetOriginalEntryPoint(), typeHandler),
-            functionInfo, ctorCache);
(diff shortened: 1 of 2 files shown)


boingoing:
@digitalinfinity @curtisman @abchatra Pretty simple change

lib/Runtime/Library/JavascriptLibrary.cpp
```diff
@@ -4528,9 +4528,9 @@ namespace Js
         ConstructorCache* ctorCache = ((functionInfo->GetAttributes() & FunctionInfo::Attributes::SkipDefaultNewObject) != 0) ?
             this->builtInConstructorCache : &ConstructorCache::DefaultInstance;
 
-        return RecyclerNewEnumClass(this->GetRecycler(), EnumFunctionClass, RuntimeFunction,
-            DynamicType::New(scriptContext, TypeIds_Function, prototype, functionInfo->GetOriginalEntryPoint(), typeHandler),
-            functionInfo, ctorCache);
+        DynamicType* type = DynamicType::New(scriptContext, TypeIds_Function, prototype, functionInfo->GetOriginalEntryPoint(), typeHandler);
+
+        return RecyclerNewEnumClass(this->GetRecycler(), EnumFunctionClass, RuntimeFunction, type, functionInfo, ctorCache);
```
akroshg: We have other places similar kind of pattern. Like 
```
symbolPrototype = DynamicObject::New(recycler,
            DynamicType::New(scriptContext, TypeIds_Object, objectPrototype, nullptr,
            DeferredTypeHandler<InitializeSymbolPrototype, DefaultDeferredTypeFilter, true>::GetDefaultInstance()));
```
they will also have potential problem though.
However current changes looks good

boingoing: That's a good point. I added the same fix to a few other places with the same pattern in JavascriptLibrary.

akroshg:
Looks good to me.

PR #771, opened 2016-04-09
Profile a callsite after the call returns

rajatd:
In the interpreter, we profile a call site first and then call the callee. It may happen that the callee was defer-parsed and the its bytecode generation failed due to some reason (OOM?) upon undeferral. Since, the call won't execute any code in this case, we should not record call site information - during work item creation, we would erroneously think that this a potential inlining candidate. Fix is to record the call site info after the call returns

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/771)


rajatd:
Submitting a different fix for the issue. This one will impact recursive inlining

Issue #758, opened 2016-04-07
FunctionBody.cpp won't compile with clang >= 3.8.0


xanlpz:
A number of errors like this happen:
```

/home/xan/git/ChakraCore/lib/Runtime/Base/FunctionBody.cpp:1070:83: error: no matching constructor for initialization of 'Js::ProxyEntryPointInfo'
        this->m_defaultEntryPointInfo = RecyclerNew(scriptContext->GetRecycler(), ProxyEntryPointInfo, DefaultDeferredDeserializeThunk);
                                                                                  ^                    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/xan/git/ChakraCore/lib/Common/Memory/Recycler.h:117:88: note: expanded from macro 'RecyclerNew'
#define RecyclerNew(recycler,T,...) AllocatorNewBase(Recycler, recycler, AllocInlined, T, __VA_ARGS__)
                                                                                       ^  ~~~~~~~~~~~
/home/xan/git/ChakraCore/lib/Common/Memory/Allocator.h:83:206: note: expanded from macro 'AllocatorNewBase'
  ...alloc, AllocFunc, T, ...) VALIDATE_OBJECT(T, new (TRACK_ALLOC_INFO(static_cast<AllocatorType *>(alloc), T, AllocatorType, 0, (size_t)-1), &AllocatorType::AllocFunc) T(__VA_ARGS__))
                                                                                                                                                                          ^ ~~~~~~~~~~~
/home/xan/git/ChakraCore/lib/Common/Memory/Allocator.h:79:33: note: expanded from macro 'VALIDATE_OBJECT'
#define VALIDATE_OBJECT(T, obj) obj
                                ^~~
/home/xan/git/ChakraCore/lib/Runtime/Base/FunctionBody.h:314:11: note: candidate constructor (the implicit copy constructor) not viable: no known conversion from 'Var (Js::RecyclableObject *, Js::CallInfo, ...)'
      (aka 'void *(Js::RecyclableObject *, Js::CallInfo, ...)') to 'const Js::ProxyEntryPointInfo' for 1st argument
    class ProxyEntryPointInfo:  public ExpirableObject
          ^
/home/xan/git/ChakraCore/lib/Runtime/Base/FunctionBody.h:314:11: note: candidate constructor (the implicit move constructor) not viable: no known conversion from 'Var (Js::RecyclableObject *, Js::CallInfo, ...)'
[...]

PR #772, opened 2016-04-09
Fix FunPtr to voidPtr casting

obastemur:
Fixes https://github.com/Microsoft/ChakraCore/issues/758

### DIFF
--- a/lib/Runtime/Base/FunctionBody.h
+++ b/lib/Runtime/Base/FunctionBody.h
@@ -316,13 +316,28 @@ namespace Js
     public:
         // These are public because we don't manage them nor their consistency;
         // the user of this class does.
-        void * address;
+        union address_type_ {
+            address_type_(void *ptr):asPtr(ptr) {}
+            address_type_(Js::JavascriptMethod method):asJsMethod(method) {}
 
-        ProxyEntryPointInfo(void* address, ThreadContext* context = nullptr):
+            void *asPtr;
+            Js::JavascriptMethod asJsMethod;
+        } address;
+        static_assert(sizeof(Js::JavascriptMethod) == sizeof(uintptr_t),
+                                "This compiler / platform is not supported");
+
+        explicit ProxyEntryPointInfo(void* address, ThreadContext* context = nullptr):
             ExpirableObject(context),
             address(address)
         {
         }
+
+        explicit ProxyEntryPointInfo(Js::JavascriptMethod address, ThreadContext* context = nullptr):
+            ExpirableObject(context),
+            address(address)
+        {
+        }
+
         static DWORD GetAddressOffset() { return offsetof(ProxyEntryPointInfo, address); }
         virtual void Expire()
         {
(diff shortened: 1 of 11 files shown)


lib/Runtime/Base/FunctionBody.h
```diff
@@ -316,13 +316,28 @@ namespace Js
...
-        void * address;
+        union address_type_ {
+            address_type_(void *ptr):asPtr(ptr) {}
+            address_type_(Js::JavascriptMethod method):asJsMethod(method) {}
 
-        ProxyEntryPointInfo(void* address, ThreadContext* context = nullptr):
+            void *asPtr;
+            Js::JavascriptMethod asJsMethod;
+        } address;
+        static_assert(sizeof(Js::JavascriptMethod) == sizeof(uintptr_t),
```
obastemur: IMO the issue was most likely an alignment related optimization issue. It don't see this static_assert is ever happening though.

lib/Runtime/Base/FunctionBody.h
```diff
@@ -316,13 +316,28 @@ namespace Js
     public:
         // These are public because we don't manage them nor their consistency;
         // the user of this class does.
-        void * address;
+        union address_type_ {
+            address_type_(void *ptr):asPtr(ptr) {}
+            address_type_(Js::JavascriptMethod method):asJsMethod(method) {}
 
-        ProxyEntryPointInfo(void* address, ThreadContext* context = nullptr):
+            void *asPtr;
```
dreamer-dead: IIRC, writing one member of a union and then inspecting it through another member leads to undefined behavior in C++ (but not in C). See a good discussion at http://stackoverflow.com/questions/2310483/purpose-of-unions-in-c-and-c

obastemur: > writing one member of a union and then inspecting it through another member leads to undefined behavior in C++ 

Right. See also (https://isocpp.org/wiki/faq/cpp11-language-types#generalized-unions) for a better reference. AFAIK, it is undefined due to different/unknown alignment approaches. 

However, it does not apply here. There is no casting from dynamically allocated memory to union / fun ptr.

obastemur: Besides, this is just a (IMO quickest alignment friendly solution) I could come up. If there is other tricks to mend this issue, welcome.

dreamer-dead: > However, it does not apply here.

Maybe I'm missing something, but I see the following code

```
Assert(entryPoint->address.asPtr != nullptr);
return entryPoint->address.asJsMethod;
```

Doesn't it means that both fields of the union are accessed at the same time?

obastemur: > Doesn't it means that both fields of the union are accessed at the same time?

I'm not sure what you are referring ? Could you please explain what exactly could go wrong there?

dreamer-dead: At the page https://isocpp.org/wiki/faq/cpp11-language-types#generalized-unions there is a sentence: `it’s illegal to write one member and then read another`.
So in my point of view, it is incorrect to use any of the `address_type_` union constructor and then execute a code like 

```
Assert(entryPoint->address.asPtr != nullptr);
return entryPoint->address.asJsMethod;
```
because it will read a field that wasn't directly initialised. Am I wrong?

obastemur: > because it will read a field that wasn't directly initialized. Am I wrong?

Perfectly yes.
AFAIK, prior to C++11 all the members of union are initialized to their default constructors. This is no longer an issue.

REF: http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2008/n2544.pdf 

IMO, what happens if they were initialized is a better question to ask. 

AFAIK; it is illegal most because the alignment issue. However, the usage here is most likely an `aliasing`..

digitalinfinity:
Hi @obastemur- @LouisLaf took a look at this change and prefers obastemur/ChakraCore@57cd553710b50a7a814f4abb95928e029fe9d051. Also, can you clarify the question at the end of #758 please?

LouisLaf:
Yes, the void* ptrs we have here are all really pointing to JS methods.  They do not point to data.  Your second change is more in line with this.  Thanks.

obastemur:
> Also, can you clarify the question at the end of #758 please?

Two PRs and two approaches. Which suits best for the actual structure. Actually @LouisLaf kind of answered it. 

@digitalinfinity @LouisLaf Thanks for the review. See https://github.com/Microsoft/ChakraCore/pull/836

PR #776, opened 2016-04-11
cross platform: linux branch compiles on OSX

obastemur:
This PR fixes compile time issues on OSX.


obastemur:
Linux branch in general fails with latest Clang (hence it fails with latest Apple LLVM 7.3.0) See https://github.com/Microsoft/ChakraCore/issues/758

The assumption for PR is either https://github.com/Microsoft/ChakraCore/pull/772 is merged or Clang < 3.8 / Apple LLVM < 7.3.0 is used.

pal/src/CMakeLists.txt
```diff
@@ -153,6 +153,21 @@ add_library(Chakra.Pal
   ${ARCH_SOURCES}
 )
 
+if(CMAKE_SYSTEM_NAME STREQUAL Darwin)
+  add_definitions(-D_TARGET_MAC64)
+  add_definitions(-DPLATFORM_UNIX)
```
jianchun: Are these 2 lines needed... aren't they populated from top-level CMakeLists.txt change?

pal/src/thread/thread.cpp
```diff
@@ -2788,21 +2788,37 @@ int CorUnix::CThreadMachExceptionHandlers::GetIndexOfHandler(exception_mask_t bm
...
     size_t stacksize;
     void* stackend;
+    int status;
 
-    pthread_getattr_np(currentThreadHandle, &attr);
-    pthread_attr_getstack(&attr, &stackend, &stacksize);
+#if HAVE_PTHREAD_ATTR_GET_NP
+        status = pthread_attr_get_np(currentThreadHandle, &attr);
+#elif HAVE_PTHREAD_GETATTR_NP
+        status = pthread_getattr_np(currentThreadHandle, &attr);
```
jianchun: Please fix indentations of the 2 `status = ...` lines.

jianchun:
LGTM, minor comments.

obastemur:
@jianchun Thanks for the review. PR is updated (with some additional updates. i.e. Now it has MachExceptions included to OSX build etc.)

pal/src/exception/machmessage.h
```diff
@@ -55,6 +55,21 @@ typedef	mach_exception_data_type_t	mach_exception_subcode_t;
 
 struct MachExceptionHandler;
 
+class MachMessage;
+
+// Contains all the exception and thread state information needed to forward the exception.
```
digitalinfinity: Can you update the PR description to explain the new set of changes around exception handling? I'm not too familiar with OS X exception handling to start out with and some more context about the motivation/some pointers to docs to read up on would be nice. Also, I see you have functions defined in the below struct but where are they implemented?

pal/src/thread/thread.cpp
```diff
@@ -2788,21 +2788,37 @@ int CorUnix::CThreadMachExceptionHandlers::GetIndexOfHandler(exception_mask_t bm
 
 int GetCurrentThreadStackBounds(char** stackBase, char** stackEnd)
 {
+#ifdef _TARGET_MAC64
```
digitalinfinity: What's the difference between __APPLE__ and _TARGET_MAC64?

pal/src/misc/dbgmsg.cpp
```diff
@@ -834,7 +834,7 @@ bool DBG_ShouldCheckStackAlignment()
 #include "CoreFoundation/CFString.h"
 #include "Security/AuthSession.h"
 
-static const char * PAL_DISPLAY_DIALOG PAL_GLOBAL = "PAL_DisplayDialog";
+#define PAL_DISPLAY_DIALOG "PAL_DisplayDialog"
```
digitalinfinity: Why is this change needed? Also, can we just ifdef this entire code out?

pal/src/arch/i386/activationhandlerwrapper.S
```diff
@@ -0,0 +1,31 @@
+// -------------------------------------------------------------------------------------------------------
+// Copyright (C) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE.txt file in the project root for full license information.
+// -------------------------------------------------------------------------------------------------------
+
+.intel_syntax noprefix
+#include "unixasmmacros.inc"
+#include "asmconstants.h"
+
+#ifdef BIT64
```
digitalinfinity: If this is 64 bit code, shouldn't this file be under pal/src/arch/amd64?

pal/inc/pal_mstypes.h
```diff
@@ -219,7 +219,7 @@ extern "C" {
 #ifndef PAL_STDCPP_COMPAT
 // Defined in gnu's types.h. For non PAL_IMPLEMENTATION system
 // includes are not included, so we need to define them.
-#ifndef PAL_IMPLEMENTATION
+#if !defined(PAL_IMPLEMENTATION) && !defined(__APPLE__)
```
digitalinfinity: Consider adding a comment here on why this is not defined for MacOS

pal/src/arch/i386/activationhandlerwrapper.S
jianchun: I was reading around here too and realized these were pulled from https://github.com/dotnet/coreclr/tree/master/src/pal/src/arch/i386. So that could be the answer. PAL did that.

pal/src/exception/machmessage.h
jianchun: This is also pulled from latest PAL: https://github.com/dotnet/coreclr/blob/775003a4c72f0acc37eab84628fcef541533ba4e/src/pal/src/exception/machmessage.h#L86, commit https://github.com/dotnet/coreclr/commit/9f6654b6faec5730f71d6c547ea497e406ae4d47, dated 1/26/2016.

The 3 `.S` assembly files are same as PAL master.

pal/src/arch/i386/activationhandlerwrapper.S
digitalinfinity: Ah, that makes sense- @obastemur can you confirm?

obastemur: Yes. They are from latest PAL

pal/src/misc/dbgmsg.cpp
obastemur: Clang OSX 7.3.0 doesn't support custom priority for non member variables and no reason that particular member is being initialized with a special priority

pal/src/thread/thread.cpp
obastemur: TARGET_MAC64 is from PAL. Kept that as is. Maybe we should only use '__APPLE__'?

pal/inc/pal_mstypes.h
obastemur: > Consider adding a comment here on why this is not defined for MacOS

done.

I'm not sure any other update is needed for this PR?

jianchun:
@obastemur Could you please sync to latest and resolve a merge conflict in `build.sh`?

obastemur:
@jianchun done

PR #777, opened 2016-04-11
Dont attempt to get profiled iterations of a functionbody if it's bytecode hasn't been generated yet.

rajatd:
A number of conditions required to be met for this issue to repro:
1. A callee at a polymorphic callsite was defer-parsed.
2. Upon undeferring, its bytecode generation failed.
3. While gathering codegen data for the parent function, we recorded that the said callee cannot be inlined but we still needed to save the info that we will need to emit a call to this function when we inline the rest of the calls at this callsite. We try to save the number of profiled iterations for the callee on the info that we collect for it - that's when we assert that execution mode and limits have not been initialized (because the bytecode wasn't generated)

This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/777)

### DIFF
--- a/lib/Runtime/Language/FunctionCodeGenJitTimeData.cpp
+++ b/lib/Runtime/Language/FunctionCodeGenJitTimeData.cpp
@@ -812,7 +812,7 @@ namespace Js
 #ifdef FIELD_ACCESS_STATS
         inlineCacheStats(nullptr),
 #endif
-        profiledIterations(GetFunctionBody() ? GetFunctionBody()->GetProfiledIterations() : 0),
+        profiledIterations(GetFunctionBody() && GetFunctionBody()->GetByteCode() ? GetFunctionBody()->GetProfiledIterations() : 0),
         next(0)
     {
     }


rajatd:
@abchatra

lib/Runtime/Language/FunctionCodeGenJitTimeData.cpp
```diff
@@ -812,7 +812,7 @@ namespace Js
 #ifdef FIELD_ACCESS_STATS
         inlineCacheStats(nullptr),
 #endif
-        profiledIterations(GetFunctionBody() ? GetFunctionBody()->GetProfiledIterations() : 0),
+        profiledIterations(GetFunctionBody() && GetFunctionBody()->GetByteCode() ? GetFunctionBody()->GetProfiledIterations() : 0),
```
akroshg: Curious, why not use `IsFunctionParsed` ?

rajatd: Because the function might have been parsed alright, but the bytecode gen failed

abchatra:
:+1:

PR #779, opened 2016-04-12
xplat: build lib/Runtime/Library

jianchun:
This change enables building `lib/Runtime/Library`.
- Added a few unimplemented stubs into `CommonPal.h`.
- Replaced a bunch of `__try/__finally(__leave)` with `TryFinally`.
- `JavascriptBuiltInFunctions.cpp`: Added `BUILTIN_TEMPLATE` to work with
  all TypedArray types as they are template specializations and need
  `template<>`.
- `TypedArray.cpp`: Needed to move `Is`/`FromVar` specializations forward
  before being used.

TODO:
- Figure out how to work around one __try/__except use -- to resume from
  out of bound array access AV.

2 files excluded from build:
- JavascriptErrorDebug.cpp: using interfaces
- MathLibrary.cpp: using intrinsics conflicting with stdlib/math.h


lib/Runtime/Library/JavascriptFunction.cpp
```diff
@@ -621,6 +618,11 @@ namespace Js
         }
 #endif
 
+// xplat-todo: This __try/__except catches AV and resumes from out of bound TypedArray access.
```
digitalinfinity: I think for now, the plan should be to disable this feature on linux. I think you should change JavascriptArrayBuffer's constructor to use malloc on linux always, and also assert that JavascriptArrayBuffer::AllocWrapper is never used on linux. Also, in TypedArray::NewInstance, compile assert virtualAllocated is false on linux.

digitalinfinity:
Just one comment re TypedArray usage. Other than that, LGTM

lib/Common/CommonPal.h
```diff
@@ -274,6 +274,10 @@ inline T InterlockedDecrement(
     return __sync_sub_and_fetch(Addend, T(1));
 }
 
+inline __int64 _abs64(__int64 n)
```
obastemur: can we use stl `llabs` instead ?

jianchun: We haven't used any STL in chakracore for various reasons. Maybe in future we could reevaluate.

lib/Runtime/Library/JavascriptFunction.cpp
jianchun: Thanks for referring to the related virtualAlloc dependencies. From the code, IsValidVirtualBufferLength always returns false and JavascriptArrayBuffer::AllocWrapper already has Assert(false) on non _WIN64 platforms. This feature is already disabled. I'll change the comment and remove this xplat-todo.

lib/Common/CommonPal.h
obastemur: Isn't that (_abs64) already part of standard library? (MS flavored)
BTW; If I'm not mistaking, llabs, abs are already used in the project.

jianchun: My mistake, I thought you were talking about STL.

Yes, _abs64 is on MS system, but not on non _WIN32. I added it to non _WIN32 only.

You are right I could use llabs. I couldn't find existing usage of it in the project. Since `long long` doesn't guarantee to be same as `__int64`, a version for __int64 might be ok?

obastemur: `abs` from `pal.h` looks like a good candidate ? something like `#define _abs64 abs`

PR #780, opened 2016-04-12
Globalization Assert

Cellule:
Relax assert in Globalization IsWellFormed. It is valid to have nullptr as a string as it will correctly throw a javascript Error that the Local is not well-formed.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/780)

### DIFF
--- a/lib/Runtime/Base/WindowsGlobalizationAdapter.cpp
+++ b/lib/Runtime/Base/WindowsGlobalizationAdapter.cpp
@@ -349,12 +349,15 @@ namespace Js
     {
         boolean retVal;
         HRESULT hr;
-        HSTRING hString;
+        HSTRING hString = nullptr;
         HSTRING_HEADER hStringHdr;
         // OK for languageTag to get truncated as it would pass incomplete languageTag below which
         // will be rejected by globalization dll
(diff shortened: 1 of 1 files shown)


Cellule:
@akroshg Can you review this please ?

akroshg:
looks good to me.

lib/Runtime/Base/WindowsGlobalizationAdapter.cpp
```diff
@@ -349,12 +349,15 @@ namespace Js
...
         boolean retVal;
         HRESULT hr;
-        HSTRING hString;
+        HSTRING hString = nullptr;
         HSTRING_HEADER hStringHdr;
         // OK for languageTag to get truncated as it would pass incomplete languageTag below which
         // will be rejected by globalization dll
         IfFailThrowHr(GetWindowsGlobalizationLibrary(scriptContext)->WindowsCreateStringReference(languageTag, static_cast<UINT32>(wcslen(languageTag)), &hStringHdr, &hString));
-        AnalysisAssert(hString);
+        if (hString == nullptr)
```
Cellule: I added this because Prefast wouldn't let it pass.
For some reason the function `IsWellFormed` doesn't allow `null` in the signature, but supports it.

Cellule:
@akroshg is my additional change ok ?

akroshg:
Yes - infact this is better as we are future proofing ourselves in the case where IsWellFormed could not handle null itself.
LGTM

PR #781, opened 2016-04-12
OS:7112681 Promise callback handler code is not marked as being handled

boingoing:
If that code throws, the exception gets propagated back to the host via ReportError which prints it into the console in Edge. We can fix this by setting a flag before calling into the callback code which indicates that the exception should be treated as handled and not handed back to the host. Also removes one dead function from JavascriptPromise.

### DIFF
--- a/lib/Runtime/Library/JavascriptPromise.cpp
+++ b/lib/Runtime/Library/JavascriptPromise.cpp
@@ -69,7 +69,7 @@ namespace Js
         InitializePromise(promise, &resolve, &reject, scriptContext);
 
         JavascriptExceptionObject* exception = nullptr;
-
+        
         // 9. Let completion be Call(executor, undefined, « resolvingFunctions.[[Resolve]], resolvingFunctions.[[Reject]] »).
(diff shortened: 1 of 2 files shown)


boingoing:
@Yongqu Can you take a look?

lib/Runtime/Library/JavascriptPromise.cpp
```diff
@@ -808,16 +811,19 @@ namespace Js
...
         {
-            return thenFunction->GetEntryPoint()(thenFunction, Js::CallInfo(Js::CallFlags::CallFlags_Value, 3),
-                thenable,
-                resolve,
-                reject);
-        }
-        catch (JavascriptExceptionObject* e)
-        {
-            exception = e;
+            Js::JavascriptExceptionOperators::AutoCatchHandlerExists autoCatchHandlerExists(scriptContext);
```
Yongqu: even nested calls we worried about were taken care of already. :)

Yongqu:
:+1:

PR #782, opened 2016-04-12
Build Lib\Jsrt and ch on Linux

digitalinfinity:
Build Lib\Jsrt and ch on Linux

This commit gets ch and Lib\Jsrt building on Linux.
ch doesn't really do anything yet since ChakraCore isn't present yet.

Of note are the following changes:
- The change builds on the change to define the xplat API for Jsrt (however the API is still not implemented)
- ch's dependency on STL multimap is replaced with a sorted doubly linked list
- ch uses narrow strings for the source file name instead of wide
- JsrtHolder is templatized on the callback function

One design choice here is to use the PAL for compiling ch. The work needed to remove the PAL dependency is understood but for expediency, I've opted to go with the PAL.


digitalinfinity:
Closing this PR- not sure why its targeted to master...

PR #783, opened 2016-04-12
Build Lib\Jsrt and ch on Linux

digitalinfinity:
This commit gets ch and Lib\Jsrt building on Linux.
ch doesn't really do anything yet since ChakraCore isn't present yet.

Of note are the following changes:
- The change builds on the change to define the xplat API for Jsrt (however the API is still not implemented)
- ch's dependency on STL multimap is replaced with a sorted doubly linked list
- ch uses narrow strings for the source file name instead of wide
- JsrtHolder is templatized on the callback function

One design choice here is to use the PAL for compiling ch. The work needed to remove the PAL dependency is understood but for expediency, I've opted to go with the PAL.


digitalinfinity:
@jianchun can you review please? @akroshg can you review the changes in ch please?

bin/ch/CMakeLists.txt
```diff
@@ -0,0 +1,23 @@
+add_executable (ch
+  ch.cpp
+  ChakraRtInterface.cpp
+  CodexAssert.cpp
+  Helpers.cpp
+  HostConfigFlags.cpp
+  WScriptJsrt.cpp
+    )
```
obastemur: typo: extra space

bin/ch/ch.cpp
```diff
@@ -442,12 +452,17 @@ HRESULT ExecuteTest(LPCWSTR fileName)
         ChakraRTInterface::JsDisposeRuntime(runtime);
     }
 
+    if (libraryNameUtf8 != nullptr)
```
akroshg: where is this libraryNameUtf8 used?

bin/ch/ch.cpp
```diff
@@ -531,6 +555,40 @@ int _cdecl wmain(int argc, __in_ecount(argc) LPWSTR argv[])
         }
         ChakraRTInterface::UnloadChakraDll(chakraLibrary);
     }
+#else
+    // TODO: Remove this once ChakraCore is compiling on Linux
+    Assert(chakraLibrary == nullptr);
```
akroshg: since we have different main function for non-win32 how do we reach here?

bin/ch/Helpers.cpp
```diff
@@ -5,10 +5,63 @@
...
+    
+    utf8char_t* destString = (utf8char_t*)malloc(cbDestString);
+    if (destString == nullptr)
+    {
+        return E_OUTOFMEMORY;
+    }
+    if (cchSourceString >= MAXUINT32)
+    {
+        return E_OUTOFMEMORY;
+    }
```
akroshg: can we move this check above - before we do malloc. Also if we can check the cbDestString as well.

akroshg:
rest looks good to me.

bin/ch/ChakraRtInterface.h
```diff
@@ -193,7 +201,7 @@ class ChakraRTInterface
     static JsErrorCode WINAPI JsAddRef(JsRef ref, unsigned int* count) { return m_jsApiHooks.pfJsrtAddRef(ref, count); }
     static JsErrorCode WINAPI JsGetValueType(JsValueRef value, JsValueType *type) { return m_jsApiHooks.pfJsrtGetValueType(value, type); }
     static JsErrorCode WINAPI JsSetIndexedProperty(JsValueRef object, JsValueRef index, JsValueRef value) { return m_jsApiHooks.pfJsrtSetIndexedProperty(object, index, value); }
-    static JsErrorCode WINAPI JsSerializeScript(const char16 *script, BYTE *buffer, unsigned long *bufferSize) { return m_jsApiHooks.pfJsrtSerializeScript(script, buffer, bufferSize); }
+    static JsErrorCode WINAPI JsSerializeScript(const char16 *script, BYTE *buffer, unsigned int *bufferSize) { return m_jsApiHooks.pfJsrtSerializeScript(script, buffer, bufferSize); }
```
dilijev: Maybe we should be using size_t for this type instead?

bin/ch/WScriptJsrt.cpp
```diff
@@ -225,33 +238,39 @@ JsValueRef WScriptJsrt::LoadScriptHelper(JsValueRef callee, bool isConstructCall
...
-    wchar_t fullPath[_MAX_PATH];
-    if (_wfullpath(fullPath, fileName, _MAX_PATH) == nullptr)
+    if (_fullpath(fullPathNarrow, fileName, _MAX_PATH) == nullptr)
     {
         IfFailGo(E_FAIL);
     }
     // canonicalize that path name to lower case for the profile storage
-    size_t len = wcslen(fullPath);
+    // REVIEW: This doesn't work for UTF8...
+    len = strlen(fullPathNarrow);
```
dilijev: I know windows has `_mbstrlen` but a quick look around doesn't reveal an obvious Linux version of the same, but there are apparently libraries like UTF8-CPP that would contain the requisite function. I'm not sure whether it would be necessary to know *a priori* whether the string is ASCII or UTF-8 since ASCII characters with binary sequences starting with 1{1,4}0 could be interpreted as part of a UTF-8 multi-byte sequence.

bin/ch/ch.cpp
```diff
@@ -500,20 +520,24 @@ int _cdecl wmain(int argc, __in_ecount(argc) LPWSTR argv[])
 
     HostConfigFlags::pfnPrintUsage = PrintUsageFormat;
 
+    // The following code is present to make sure we don't load
+    // jscript9.dll etc with ch. Since that isn't a concern on non-Windows
+    // builds, it's safe to conditionally comment it out.
```
dilijev: "conditionally compiler it out"?

bin/ch/ch.cpp
dilijev: `main` calls `wmain`

lib/Jsrt/JsrtPch.cpp
```diff
@@ -2,5 +2,4 @@
 // Copyright (C) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE.txt file in the project root for full license information.
 //-------------------------------------------------------------------------------------------------------
-#pragma once
```
dilijev: Weird... Was JsrtPch.cpp `#include`d somewhere at some point?

pal/src/misc/bstr.cpp
```diff
@@ -119,6 +119,24 @@ STDAPI_(BSTR) SysAllocStringLen(const OLECHAR *psz, UINT len)
...
+ *  BSTR to free
+ *
+ *Exit:
+ *  return value = void
+ *
+ ***********************************************************************/
+STDAPI_(void) SysFreeString(const OLECHAR* psz)
+{
+    if (psz != NULL)
+        HeapFree(GetProcessHeap(), 0, (LPVOID) psz);
```
dilijev: { } around if body

dilijev:
LGTM

bin/ch/MessageQueue.h
```diff
@@ -22,36 +22,183 @@ class MessageBase
...
+                InsertAfter(node, prevOfCurr);
+                return;
+            }
+            curr = curr->next;
+        }
+
+        // If we reach this point, we didn't find a node that satisfied the predicate
+        // That means, we're either inserting to the end of the list, or the list is empty 
+        // or we're inserting after head (the insert before head case is taken care of in the loop)
+        InsertAfter(node, (prevOfCurr != nullptr ? prevOfCurr->next : head));
```
jianchun: Seems this function can be simplified a bit.
```
  prev = nullptr;

  while (curr != nullptr) {
    if (T::compare(data, curr->data)) {
      break;
    }
    prev = curr;
    curr = curr->next;
  }

  InsertAfter(node, prev);
```

bin/ch/MessageQueue.h
```diff
@@ -22,36 +22,183 @@ class MessageBase
...
+    {
+        DListNode<T>* curr = head;
+        DListNode<T>* node = new DListNode<T>(data);
+        DListNode<T>* prevOfCurr = nullptr; 
+
+        // Now, if we have to insert, we have to insert *after* some node
+        while (curr != nullptr)
+        {
+            prevOfCurr = curr->prev;
+            if (T::Compare(data, curr->data))
```
jianchun: Change `T::Compare` to `T::LessThan` would be more clear.

lib/CMakeLists.txt
```diff
@@ -1,3 +1,4 @@
 add_subdirectory (Common)
 add_subdirectory (Parser)
 add_subdirectory (Runtime)
+add_subdirectory (Jsrt)
```
jianchun: nit: add a blank line at the end

lib/Jsrt/Jsrt.cpp
```diff
@@ -2697,24 +2699,31 @@ JsErrorCode RunSerializedScriptCore(const wchar_t *script, JsSerializedScriptLoa
     });
 }
 
+#ifdef _WIN32
+template
+JsErrorCode RunSerializedScriptCore<JsSerializedScriptLoadSourceCallback, JsSerializedScriptUnloadCallback>(const wchar_t *script, JsSerializedScriptLoadSourceCallback scriptLoadCallback,
+    JsSerializedScriptUnloadCallback scriptUnloadCallback, unsigned char *buffer, JsSourceContext sourceContext,
+    const wchar_t *sourceUrl, bool parseOnly, JsValueRef *result);
```
jianchun: Is this needed? Wouldn't the compiler automatically figure out needed template instantiations from the callers below?

jianchun:
LGTM, just a few nits/questions.

bin/ch/CMakeLists.txt
digitalinfinity: Fixed

bin/ch/ch.cpp
digitalinfinity: Looks vestigial- removed.

bin/ch/ch.cpp
digitalinfinity: Yup- @dilijev is correct here

bin/ch/Helpers.cpp
digitalinfinity: Done

bin/ch/ChakraRtInterface.h
digitalinfinity: Not a bad suggestion but out of scope for this change (I just changed ChakraRtInterface to match the JSRT API definition here)- consider opening an issue for changing the JSRT api to use size_t instead of unsigned int for this API (although I'm not sure if we can change this API for compat reasons- talk to @liminzhu to confirm)

bin/ch/WScriptJsrt.cpp
digitalinfinity: Actually, the comment was less about using strlen and more about how we use the length to convert the string to its lower-case equivalent. This approach doesn't work with UTF8 strings- since it's null-terminated, we could do a single pass conversion (read a char, convert to lower case, replace) until we hit the null char. However, not sure if we can do it in-place in all cases or if there are cases where the lower case version of a character takes up more bytes than the upper case one. Anyway, it's academic for now which is why I left it as a review comment in case anyone actually runs into the issue using an UTF8 path to a file :)

Side note: Not sure if _mbstrlen works with UTF8 strings (I think it predates utf8 and the mb prefix refers to some other multi-byte encodings but could be wrong)

bin/ch/ch.cpp
digitalinfinity: Fixed

lib/Jsrt/JsrtPch.cpp
digitalinfinity: Don't think so- think it was just a copy-paste error

pal/src/misc/bstr.cpp
digitalinfinity: Done

bin/ch/MessageQueue.h
digitalinfinity: Nice- changed.

bin/ch/MessageQueue.h
digitalinfinity: Done

lib/CMakeLists.txt
digitalinfinity: Done

lib/Jsrt/Jsrt.cpp
digitalinfinity: Hmm, thought I needed this at some point but removing it doesn't seem to make any difference- so removed it.

bin/ch/ChakraRtInterface.h
dilijev: Opened issue. :)

PR #785, opened 2016-04-12
Don't scan finalizable leaf objects

leirocks:
We are scanning leaf objects if it is finalizable today.
this should improve perf a little bit and prevent many false positives because PropertyRecord which has random string falls into this category.


leirocks:
looks we are not actually scanning it. but handled differently than normal leaf object

PR #787, opened 2016-04-12
Fix for CheckFixedFld hoisting bug OS:6773420

LouisLaf:
Hoist CheckObjType as well since there is a dependency with CheckFixedFld, and
don't hoist propertySym opnd if type is checked.  It break obj type spec and has no benefits.


LouisLaf:
@pleath and @MikeHolman, please review.

pleath:
LGTM.

PR #788, opened 2016-04-12
Update arguments count for an inlinee at InlineeStart instead of InlineeEnd

rajatd:
We currently do this when we see InlineeEnd, and have special cases for
opcodes (such as InlineThrow, InlineRuntimeError) which may make
InlineeEnd unreachable. Updating the maxInlineeArgoutCount on the func
upon seeing an InlineeStart is safer and more sustainable. We wouldn't
have to keep adding cases in which InlineeEnd might get removed.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/788)


rajatd:
@abchatra @LouisLaf @nmostafa

abchatra:
Have you run the ExprGen on this?

nmostafa:
:+1:

rajatd:
@abchatra , its running. So far so good.

PR #789, opened 2016-04-12
Loop upper/lower bound issue

pleath:
Fix the underlying loop count identification bug that was exposed by a memop issue that hit on the web. The fix was suggested by Kount: given two possibly symbols that could be the bound for the loop count, if one symbol is itself a relative bound of the other, choose the other.


pleath:
@LouisLaf @rajatd

rajatd:
Looks good to me.

PR #791, opened 2016-04-12
Fix for issue with aux slots reuse in default parameter

aneeshdk:
When the function has default parameters we do the LdSlotArr before the body starts. When the body, with eval, has InitFld it can update the size of the slot array thus allocating a new slot array. In this bug, in the body we were still using the old aux slots array. With this fix we kill the slot array when InitFld occurs.


aneeshdk:
@abchatra @pleath @tcare Can you please take a look?

tcare:
Looks good.

abchatra:
:+1:

PR #793, opened 2016-04-13
MemCopy modified source length

Cellule:
When doing a MemCopy, do the copy from only 1 segment to another and make sure they are both long enough.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/793)

### DIFF
--- a/lib/Runtime/Library/JavascriptArray.inl
+++ b/lib/Runtime/Library/JavascriptArray.inl
@@ -955,45 +955,46 @@ SECOND_PASS:
             return true;
         }
 
-        SparseArraySegment<T> *toSegment = PrepareSegmentForMemOp<T>(toStartIndex, length);
-        if (toSegment == nullptr)
+        const auto isSegmentValid = [length](Js::SparseArraySegment<T>* segment, uint32 startIndex) {
+            uint32 end, segmentEnd;
+            // Check the segment is long enough
+            return (
+                segment &&
+                !UInt32Math::Add(startIndex, length, &end) &&
+                !UInt32Math::Add(segment->left, segment->length, &segmentEnd) &&
+                startIndex >= segment->left &&
+                startIndex < segmentEnd &&
+                segmentEnd >= end
+            );
+        };
+        //Find the segment where itemIndex is present or is at the boundary
+        Js::SparseArraySegment<T>* fromSegment = (Js::SparseArraySegment<T>*)fromArray->GetBeginLookupSegment(fromStartIndex, false);
+        if (!isSegmentValid(fromSegment, fromStartIndex))
         {
             return false;
         }
-        Assert(fromArray->head);
-        Assert(fromArray->length >= (fromStartIndex + length));
 
-        //Find the segment where itemIndex is present or is at the boundary
-        SparseArraySegmentBase* current = fromArray->GetBeginLookupSegment(fromStartIndex, false);
-        Assert(current);
-        Assert(GetSegmentMap() == nullptr  && fromArray->GetSegmentMap() == nullptr);
-
-        while (current && length)
+        // Check the from segment first so we don't prepare the toSegment in case it is invalid
+        SparseArraySegment<T> *toSegment = PrepareSegmentForMemOp<T>(toStartIndex, length);
+        if (!isSegmentValid(toSegment, toStartIndex))
         {
-            int memcopySize = length;
-            int startOffset;
-            if (fromStartIndex >= current->left && fromStartIndex < (current->left + current->length))
-            {
-                startOffset = fromStartIndex - current->left;
-                if ((startOffset + length) > current->length)
-                {
-                    memcopySize = current->length - startOffset;
-                }
-
-
-                js_memcpy_s(toSegment->elements + toStartIndex, memcopySize * sizeof(T), (((Js::SparseArraySegment<T>*)current)->elements + startOffset), memcopySize * sizeof(T));
-
-                fromArray->SetLastUsedSegment(current);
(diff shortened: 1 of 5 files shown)


Cellule:
@LouisLaf @pleath @rajatd Can one of you guys review this please ?

lib/Runtime/Library/JavascriptArray.inl
```diff
@@ -955,45 +955,45 @@ SECOND_PASS:
...
 
-        SparseArraySegment<T> *toSegment = PrepareSegmentForMemOp<T>(toStartIndex, length);
-        if (toSegment == nullptr)
+        const auto isSegmentValid = [length](Js::SparseArraySegment<T>* segment, uint32 startIndex) {
+            uint32 end, segmentEnd;
+            // Check the segment is long enough
+            return (
+                segment &&
+                !UInt32Math::Add(startIndex, length, &end) &&
+                !UInt32Math::Add(segment->left, segment->length, &segmentEnd) &&
```
rajatd: For the dst array, shouldn't we be calculating segmentEnd as segment->left + segment->size (instead of segment->length)?

lib/Runtime/Library/JavascriptArray.inl
```diff
@@ -955,45 +955,45 @@ SECOND_PASS:
...
         }
+        Assert(GetSegmentMap() == nullptr && fromArray->GetSegmentMap() == nullptr);
+
+        int memcopySize = length;
+        int startOffset = fromStartIndex - fromSegment->left;
+        Assert((startOffset + length) <= fromSegment->length);
 
-        Assert(length == 0);
+        js_memcpy_s(
+            toSegment->elements + toStartIndex,
```
rajatd: Should this also be an offset from the start of the segment?

lib/Runtime/Library/JavascriptArray.inl
Cellule: No because `PrepareSegmentForMemOp` should modify the length of the segment to be long enough for the copy.
That particular check could actually be an assert for the dst, but I figured it was safer to do the check.

I still don't know what happens if this check fails since the segment has already been modified. 
I don't know the `PrepareSegmentForMemOp` code and since there has been no bug with it so far, I assume it correctly prepares the segment (or return `nullptr` which is fine to bailout then)
The only problematic case is if `PrepareSegmentForMemOp` returns a segment that is not big enough, which I think cannot happen

lib/Runtime/Library/JavascriptArray.inl
Cellule: i think you are right, I think this used to work because we only do the memcopy with the head segment, but since there is no actual code protecting it we could possibly use another segment. I'll make the change

lib/Runtime/Library/JavascriptArray.inl
rajatd: Hmm.. if PrepareSegmentFromMemOp does the right thing, then this part should be fine.

---
In reply to: [59632827](https://github.com/Microsoft/ChakraCore/pull/793#discussion_r59632827) [](ancestors = 59632827)

rajatd: Actually, if the segment was big enough to begin with, but its length was smaller that its allocated size, your current code will call it as invalid if we try to write beyond segment->length but within segment->size.

lib/Runtime/Library/JavascriptArray.inl
```diff
@@ -955,45 +955,46 @@ SECOND_PASS:
...
+
+        int memcopySize = length;
+        int toStartOffset = toStartIndex - toSegment->left;
+        int fromStartOffset = fromStartIndex - fromSegment->left;
+        Assert((fromStartOffset + length) <= fromSegment->length);
 
-        Assert(length == 0);
+        js_memcpy_s(
+            toSegment->elements + toStartOffset,
+            (toSegment->size - toStartOffset) * sizeof(T),
```
rajatd: This seems fishy- this can potentially try to copy more elements than memcopySize, and will throw internal error. I think if you have verified the validity of both segments, you could replace this with memcopySize.

Cellule: well the whole point of `js_memcpy_s` is to make an additional check between the buffer size and the copy length.
I remember we had issues with this in the past and determined this is the best course of action.
@LouisLaf might have more input about this

lib/Runtime/Library/JavascriptArray.inl
Cellule: `PrepareSegmentFromMemOp ` is supposed to change the length to match. Since I make no additional change to the segment length after, it has to already by valid and thus this check is just to make sure PrepareSegmentFromMemOp did its job right.

rajatd: ok. sounds good

---
In reply to: [59797832](https://github.com/Microsoft/ChakraCore/pull/793#discussion_r59797832) [](ancestors = 59797832)

lib/Runtime/Library/JavascriptArray.inl
LouisLaf: js_memcpy will copy memcopySize bytes, unless that is larger than toSegment->size -toStartOffset.  If it is larger, it will abort the process.  I assume we have the right checks with the isSegmentValid calls above to prevent this.

---
In reply to: [59797587](https://github.com/Microsoft/ChakraCore/pull/793#discussion_r59797587) [](ancestors = 59797587)

LouisLaf:
:+1:

PR #794, opened 2016-04-13
MemCopy Native Float Arrays

Cellule:
Enable Native Float arrays for memcopy

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/794)

### DIFF
--- a/lib/Backend/GlobOpt.cpp
+++ b/lib/Backend/GlobOpt.cpp
@@ -4121,9 +4121,7 @@ GlobOpt::IsAllowedForMemOpt(IR::Instr* instr, bool isMemset, IR::RegOpnd *baseOp
     if (
         !indexValueType.IsInt() ||
             !(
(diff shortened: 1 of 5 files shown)


Cellule:
@LouisLaf @pleath @rajatd Can on of you guys review this please ?

rajatd:
Do you know if there were strong reasons to not have it enabled for float arrays?

Cellule:
Simply that when I enabled it for memset, I hadn't done any testing for memcopy so I held off doing it for memcopy.
Apparently there was a bug with the check that prevented memcopy for native float arrays (it would produce memcopy then bailout because it wasn't supported).
I considered reinforcing the check to avoid this issue, but I decided it was actually easier to enable it. I tested it some more and determined it was ok to do a memcopy on native float arrays.

rajatd:
:+1:

PR #795, opened 2016-04-13
Yield keyword usage in default parameter was not handled well in defer parse mode

aneeshdk:
It is a syntax error to use yield keyword in the default parameter
expression in a generator function. We were using the wrong function node
to check this condition in defer parse mode.

### DIFF
--- a/test/es6/rlexe.xml
+++ b/test/es6/rlexe.xml
@@ -751,6 +751,13 @@
   </test>
   <test>
     <default>
+      <files>generators-deferparse.js</files>
+      <compile-flags>-force:deferparse -ES6Generators -JitES6Generators -ES6Classes -ES6DefaultArgs</compile-flags>
+      <tags>exclude_arm</tags>
+    </default>
+  </test>
+  <test>
+    <default>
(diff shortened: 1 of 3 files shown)


aneeshdk:
@ianwjhalliday @tcare Can you please take a look?

test/es6/rlexe.xml
```diff
@@ -751,6 +751,13 @@
   </test>
   <test>
     <default>
+      <files>generators-deferparse.js</files>
+      <compile-flags>-force:deferparse -ES6Generators -JitES6Generators -ES6Classes -ES6DefaultArgs -args summary -endargs</compile-flags>
```
akroshg: Above test did not use the unittestframework - so passing -args summary -endargs is not useful. How about you have this test in the generators.js itself and call that in force deferparse mode as well?

akroshg:
Rest looks good to me.

test/es6/rlexe.xml
aneeshdk: Updated.

PR #799, opened 2016-04-13
Fixing an assert on double to string coversion

akroshg:
The assert is an overzealous assert as the exponent can be higher then
number of digits can be. Removed that assert.


akroshg:
@abchatra - please have a look at this one.

akroshg:
I am going to hold on this change for now.

PR #801, opened 2016-04-13
Guard deleting of DebugContext in critical section

agarwal-sandeep:
OnBreakFlagChange from PDM can be called while debugcontext is being
deleted, so gaurd the deletion in critical section and if getting critical
section fails during OnBreakFlagChange return error to PDM.


agarwal-sandeep:
@akroshg @Yongqu please take a look

lib/Runtime/Base/ScriptContext.cpp
```diff
@@ -621,11 +624,19 @@ namespace Js
 
         if (this->debugContext != nullptr)
         {
+            // Guard the closing and deleting of DebugContext as in meantime PDM might
+            // call OnBreakFlagChange
+            EnterCriticalSection(&debugContextCloseCS);
```
akroshg: could be changed to AutoCriticalSection?

akroshg:
Looks good to me.

PR #803, opened 2016-04-13
Destructuring issue with assignment.

akroshg:
We were passing the rhs regslot location while emitting bytecode for the object's member. However if the first member changes the variable which is stored in that rhslocation, then the second member will work on the updated one instead of the original value on the rhs location.
For fixing that we storing the rhslocation to a temp and use that temp for emitting the members of the object pattern.


akroshg:
@tcare please have a look.

tcare:
Looks good.

PR #805, opened 2016-04-14
Add FailFast for SEH handling

MikeHolman:
Doing SEH over Chakra frames can cause us to go into a bad state because C++ destructors will not be called.
This change adds code to explicitly FailFast in case a host tries this.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/805)

### DIFF
--- a/bin/ch/ch.cpp
+++ b/bin/ch/ch.cpp
@@ -109,9 +109,14 @@ int HostExceptionFilter(int exceptionCode, _EXCEPTION_POINTERS *ep)
     }
 
     fwprintf(stderr, _u("FATAL ERROR: %ls failed due to exception code %x\n"), hostName, exceptionCode);
-    fflush(stderr);
 
-    return EXCEPTION_EXECUTE_HANDLER;
+    _flushall();
+
+    // Exception happened, so we probably didn't clean up properly,
+    // Don't exit normally, just terminate
+    TerminateProcess(::GetCurrentProcess(), exceptionCode);
+
+    return EXCEPTION_CONTINUE_SEARCH;
 }
 
 void __stdcall PrintUsageFormat()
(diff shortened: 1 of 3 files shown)


MikeHolman:
@Yongqu @LouisLaf, can you please review?

bin/ch/ch.cpp
```diff
@@ -109,9 +109,14 @@ int HostExceptionFilter(int exceptionCode, _EXCEPTION_POINTERS *ep)
...
 
     fwprintf(stderr, _u("FATAL ERROR: %ls failed due to exception code %x\n"), hostName, exceptionCode);
-    fflush(stderr);
 
-    return EXCEPTION_EXECUTE_HANDLER;
+    _flushall();
+
+    // Exception happened, so we probably didn't clean up properly,
+    // Don't exit normally, just terminate
+    TerminateProcess(::GetCurrentProcess(), exceptionCode);
```
Yongqu: when debugger not present?

lib/Runtime/Library/JavascriptFunction.cpp
```diff
@@ -621,21 +619,38 @@ namespace Js
         }
 #endif
 
+        // mark volatile, because otherwise VC will incorrectly optimize away load in the finally block
+        volatile ulong exceptionCode = 0;
+        volatile int exceptionAction = EXCEPTION_CONTINUE_SEARCH;
+        EXCEPTION_POINTERS exceptionInfo = EXCEPTION_POINTERS();
```
Yongqu: is this basically {0}?

Yongqu:
:+1:

bin/ch/ch.cpp
MikeHolman: I am trying to keep with current behavior, which had this in the except handler before.

PR #806, opened 2016-04-14
Do not throw implicit exceptions when an implicit call happened

Cellule:
BailOutOnImplicitCallsPreOp means anything unexpected should bailout and try again in the interpreter.
Since this is placed on hoisted instructions, it is possible that we won't actually execute this instruction.
We have to bailout on exception and let the interpreter determine if we actually need to throw or not.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/806)

### DIFF
--- a/lib/Runtime/Base/ThreadContext.cpp
+++ b/lib/Runtime/Base/ThreadContext.cpp
@@ -3580,8 +3580,19 @@ bool ThreadContext::HasNoSideEffect(Js::RecyclableObject * function, Js::Functio
 bool
 ThreadContext::RecordImplicitException()
 {
+    // Record the exception in the implicit call flag
     AddImplicitCallFlags(Js::ImplicitCall_Exception);
-    return !IsDisableImplicitException();
+    if (IsDisableImplicitException())
+    {
+        // Indicate that we shouldn't throw if ImplicitExceptions have been disabled
+        return false;
+    }
+    // Disabling implicit exception when disabling implicit calls can result in valid exceptions not being thrown.
+    // Instead we tell not to throw only if an implicit call happened and they are disabled. This is to cover the case
+    // of an exception being thrown because an implicit call not executed left the execution in a bad state.
+    // Since there is an implicit call, we expect to bailout and handle this operation in the interpreter instead.
+    bool hasImplicitCallHappened = IsDisableImplicitCall() && (GetImplicitCallFlags() & ~Js::ImplicitCall_Exception);
+    return !hasImplicitCallHappened;
 }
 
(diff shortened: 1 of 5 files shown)


Cellule:
@Yongqu @rajatd Can you review this please ?

Cellule:
After discussion, I decided to change the behavior to not throw only if an implicit call happens.
The difference is, that if we decide to throw and no implicit calls happened, then we will throw.

@pleath Can you review this please ?

lib/Runtime/Language/JavascriptOperators.cpp
```diff
@@ -2152,8 +2152,8 @@ namespace Js
                 {
                     Assert((flags & Data) == Data && (flags & Writable) == None);
 
-                    JavascriptError::ThrowCantAssignIfStrictMode(propertyOperationFlags, requestContext);
                     requestContext->GetThreadContext()->AddImplicitCallFlags(ImplicitCall_NoOpSet);
+                    JavascriptError::ThrowCantAssignIfStrictMode(propertyOperationFlags, requestContext);
```
Cellule: It is a little confusing how this case should be handled.
The implicit call possibly already happened and we want to throw/bailout.
With my change, in this scenario it means that if we disabled implicit calls we won't throw and bailout instead.
The question is, should we prioritize bailing out or throwing ?

Cellule:
Ping

lib/Runtime/Language/JavascriptOperators.cpp
pleath: I think bailing out (and then throwing from the interpreter if appropriate) is actually preferable, because it gives us a chance to rejit with the knowledge that the function will trigger implicit calls. That won’t keep us from throwing in future, but it’ll keep us from optimizing in a way that requires us to bail out.

--Paul

lib/Runtime/Base/ThreadContext.cpp
```diff
@@ -3581,7 +3581,13 @@ bool
 ThreadContext::RecordImplicitException()
 {
     AddImplicitCallFlags(Js::ImplicitCall_Exception);
-    return !IsDisableImplicitException();
+    if (IsDisableImplicitException())
+    {
+        return false;
+    }
+    bool hasImplicitCallHappened = IsDisableImplicitCall() && (GetImplicitCallFlags() & ~Js::ImplicitCall_Exception);
+    // Do not throw if any implicit call flag is set
```
pleath: Can you add a more explanatory comment here? Something about the reason for not just disabling implicit exceptions everywhere that implicit calls are disabled?

Cellule:
ping

pleath:
LGTM. Thanks.

PR #807, opened 2016-04-14
MSFT: 5470396 sourceList entry needs to be guarded against etw rundown

Yongqu:
We can be compiling new source code while rundown thread is reading from the list, causing AV on the reader thread
lock the list during write when new code is compiled. The lock should be mostly noop as the rundown open happen
during debugger dynamic attach & etw data collection start after the script is running. Don't see ieperf difference.


agarwal-sandeep:
ETW rundown is taking the lock at beginning and for each script context doing too much work, enumerating all source contexts, all function body, all entry points and all loop headers so during that time any code in UI thread which is waiting on etw lock is blocked. We should consider moving the lock at a granular level so as not to block UI thread.

This change LGTM.

PR #808, opened 2016-04-14
get rid of jscript9lock check in chakracore

Yongqu:
There is valid scenario where jscript9.dll and chakracore.dll can be loaded in the same
process. The chance of misuse of jscript9+chakracore scenario should be really rare. Let's
remove the lock and see if there is any strange combined usage.


Yongqu:
@digitalinfinity can you take a look?

digitalinfinity:
LGTM

PR #810, opened 2016-04-14
CAS:FaultInjection: ASSERT:  Ensure var is actually a 'JavascriptFunction'

suwc:
chakra!Js::JavascriptFunction::FromVar+7f [d:\th\inetcore\jscript\lib\runtime\library\javascriptfunction.inl @ 50]
Need to change the type of the new typehandler before changing type handler of an instance.

### DIFF
--- a/lib/Runtime/Types/DeferredTypeHandler.cpp
+++ b/lib/Runtime/Types/DeferredTypeHandler.cpp
@@ -65,24 +65,18 @@ namespace Js
 
         if (isSimple)
         {
-            newTypeHandler = ConvertToSimpleDictionaryType(instance, initSlotCapacity);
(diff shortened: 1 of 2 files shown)


suwc:
@akroshg @abchatra could you take a look pls?

rajatd:
Change itself looks good. Although, I'm curious to know what issue is this fixing.

suwc:
The issue is captured in OS2127188.

ianwjhalliday:
External to Microsoft GitHub users cannot see our internal bug database.  It would be neighborly to provide more details on this issue for all to see.

ianwjhalliday:
Also makes historical research easier in the future if we ever look back at this.

abchatra:
:+1:

PR #812, opened 2016-04-14
xplat: turn on and customize PAL_STDCPP_COMPAT mode

jianchun:
We use PAL's stdlib implementations which are incompatible with std headers.
This was a big headache as we couldn't include std headers such as
`<stdio.h>` or `<emmintrin.h>`, directly or indirectly. To work around we
used to copy some declarations directly (into palrt.h or CommonPal.h), till
it no longer worked for some reason (copied declarations dependency?).

This change attemps to turn on `PAL_STDCPP_COMPAT` mode and customize it,
so that we can use PAL and be compatible with std headers.

The primary change is very simple: turn on `PAL_STDCPP_COMPAT` mode, and
in `pal.h` enable some non-STDCPP-COMPAT customizations that we need. (e.g.,
`printf` is defined to `PAL_printf`, thus Chakra code that uses `printf`
are all redirected to call `PAL_printf`.)

With this change we are able to `#include <emmintrin.h>` and build
`Runtime/Library/MathLibrary.cpp` and a list of `Runtime/Language/Simd???`
files.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -59,28 +59,19 @@ if(CLR_CMAKE_PLATFORM_UNIX)
         add_definitions(-D_M_X64 -D_M_AMD64)
     endif(CMAKE_SYSTEM_PROCESSOR STREQUAL x86_64 OR CMAKE_SYSTEM_PROCESSOR STREQUAL amd64)
 
-    add_definitions("-D__STDC_WANT_LIB_EXT1__=1")
     add_definitions(
         -DUNICODE
         -D_SAFECRT_USE_CPP_OVERLOADS=1
+        -D__STDC_WANT_LIB_EXT1__=1
+        -DDISABLE_JIT=1  # xplat-todo: enable the JIT for Linux
         )
 
-    add_compile_options(-fms-extensions)
-
-    # Disable some warnings
-    add_compile_options(-Wno-tautological-constant-out-of-range-compare)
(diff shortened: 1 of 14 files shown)


jianchun:
@digitalinfinity Could you review please?

digitalinfinity:
Overall, looks good. The one feedback is to make the model as explicit as possible. So it seems like the new model is as follows:
- PAL can use stdlib/stl. It can also provide *some* stdlib functions
- ChakraCore can use stdlib/stl, but in the cases where the PAL provides a function, direct calls from ChakraCore will use that variant instead of the official one. Indirect calls will still use the official one (not sure if this can cause problems).

Does that seem right? Anything else that we need to think through here?

Oh, one other thing- keep the Simd operations code out of the build for now- we can re-enable it after we have ChakraCore working

jianchun:
To be clear this change itself is of less scope than @digitalinfinity discussed. It didn't touch PAL itself (if/how PAL uses stdlib) or impact STL usage. This is my understanding:
- PAL and ChakraCore can both use stdlib.
- For cases PAL provides necessary replacements, ChakraCore uses PAL's replacements transparently. E.g. there is official stdlib `printf`. PAL provides a `PAL_printf` to support extended features (printing char16) and `#define printf PAL_printf`. Thus ChakraCore source actually calls `PAL_printf` and links to PAL's implementation.

I will revert `Simd` from CMakeLists.txt.

PR #813, opened 2016-04-14
Do not recompile a RegExp pattern before calling "exec" from Symbol methods

### LINKED ISSUE #759: RegExpBuiltinExec should no longer access sticky and global via getters
It should instead consult the [[OriginalFlags]] internal slot. See https://github.com/tc39/ecma262/pull/494.

goyakin:
RegExpBuiltinExec in the spec has been updated to infer "global" and
"sticky" using [[OriginalFlags]] instead of "Get"ting them via their
respective properties.

The "exec" implementation in our codebase wasn't previously updated to
use the properties due to perf concerns, so it already uses [[OriginalFlags]].
However, we used to recompile the RegExp pattern before calling "exec"
from Symbol methods (e.g., RegExp.prototype[@@match]), so this change
removes that recompilation.

Fixes #759.

### DIFF
--- a/lib/Runtime/Library/JavascriptRegularExpression.cpp
+++ b/lib/Runtime/Library/JavascriptRegularExpression.cpp
@@ -870,21 +870,6 @@ namespace Js
 
     Var JavascriptRegExp::CallExec(RecyclableObject* thisObj, JavascriptString* string, PCWSTR varName, ScriptContext* scriptContext)
     {
-        JavascriptRegExp* regExObj = nullptr;
-
-        // TODO: The built-in "exec" in ES6 is supposed to access the RegExp flags. Since the flags
-        // can be overridden and return a different value than what's passed to the constructor,
-        // the pattern needs to be recompiled. However, the way the observability check is currently
-        // implemented, it reduces the perf quite a lot.
-        //
-        // Therefore, we recompile the pattern here before calling "exec" for the time being, but
-        // this will be moved to "exec".
-        if (JavascriptRegExp::Is(thisObj))
-        {
-            regExObj = JavascriptRegExp::FromVar(thisObj);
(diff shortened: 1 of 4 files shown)


goyakin:
@akroshg, could you take a look?

goyakin:
Ping...

lib/Runtime/Library/JavascriptRegularExpression.cpp
```diff
@@ -870,21 +870,6 @@ namespace Js
...
-        // can be overridden and return a different value than what's passed to the constructor,
-        // the pattern needs to be recompiled. However, the way the observability check is currently
-        // implemented, it reduces the perf quite a lot.
-        //
-        // Therefore, we recompile the pattern here before calling "exec" for the time being, but
-        // this will be moved to "exec".
-        if (JavascriptRegExp::Is(thisObj))
-        {
-            regExObj = JavascriptRegExp::FromVar(thisObj);
-            regExObj->RecompilePatternForExecIfNeeded(scriptContext);
```
akroshg: I see this function is called only here, who else is using it? I don't see that in our codebase. Do you need to remove the function itself?

goyakin: This was the only place it was referenced, so it is now removed with commit 97f7269d2e40b69c33da24f7ff9df51805b288e2. I did the same for [`JavascriptRegExp::HasObservableStickyFlag`](https://github.com/Microsoft/ChakraCore/pull/813/commits/97f7269d2e40b69c33da24f7ff9df51805b288e2#diff-8dbc33a433cf59be6a51b9a3318ad7d2L71).

akroshg:
alright - looks good to me.

PR #814, opened 2016-04-15
Fix assertion for background access on function body counter fields

leirocks:
- in case of -prejit -forceserialized combination, we assumed since background jit started, these counters should not be modified any more. however with prejit and forceserialized, it updates serializationId and hit the assertion. Looks we don't use the serializeID and functionTable fields at all. just remove them

- When Etw rundown in background thread, the foreground thread can be updating the counters and can lead to expansion. we need a lock for the expansion.


leirocks:
@abchatra @curtisman 

I removed serializeId counter field which seems we don't actually use it for identifying anything. also removed the functionTable/functionTableSize which don't even use the value passed from IASD/JSRT. removing serializeID is enough to solve the background checking assertion on -prejit and -forcesearialized combination.

another possible background checking assertion can happen on ETW rundown. I fixed this by entering the lock before expanding the counters structure.

leirocks:
@digitalinfinity

lib/Runtime/Base/CompactCounters.h
```diff
@@ -235,45 +164,42 @@ namespace Js
...
-            this->fieldSize = sizeof(FieldT);
-            this->fields = fieldsArray;
+            if (this->fieldSize == 0)
+            {
+                this->fieldSize = sizeof(FieldT);
+                this->fields = fieldsArray;
+            }
+            else
+            {
+                AutoCriticalSection autoCS(ThreadContext::GetThreadContextList()->GetEtwRundownCriticalSection());
```
digitalinfinity: Careful here- you already hold the ThreadContext's critical section- please trace through the code and make sure you don't end up deadlocking

digitalinfinity:
:+1:

PR #815, opened 2016-04-15
ShrU_A int type spec fix

pleath:
The int range for a ShrU_A result was using an int32 min/max. We detect ShrU_A whose result may not be an int32 (source may be negative, shift amount may be zero), but because the value type assigned to the result is 'Likely Int', we still use an int32 range. We could bail out on non-int32 result, or we could assign an 'Any Number' value type. I've gone with the latter, since there may be no operation downstream that is affected by the range. Benchmarks are flat.


pleath:
@LouisLaf @rajatd

rajatd:
Why are we using the int-range if the value type also contains "Number"?

pleath:
This happens in ValueInfo::TryGetIntConstantUpperBound. It will return a range based on int32 if the ValueInfo is "Likely Int" and the caller wants to include likely int (usually because aggressive int type spec is on). I tried to change that and got asserts. I think the rule is that it's up to the caller to create a bail out point to guard against incorrect use of an int range.

rajatd:
I see, ok. Lgtm

LouisLaf:
:+1:

PR #816, opened 2016-04-15
Disabling eval in async functions in non-strict mode

aneeshdk:
If a var declaration occurs in eval in a non-strict async function the
variable should ideally bind to the inner function. But right now if the
var declaration is shadowing one of the formals then the binding goes
wrong. This changelist makes it a SyntaxError to use eval inside async
functions in non-strict mode. Fix for the binding issue will be addressed later.


aneeshdk:
@pleath @ianwjhalliday @abchatra @tcare Can you please take a look?

lib/Parser/perrors.h
```diff
@@ -100,3 +100,4 @@ LSC_ERROR_MSG(1087, ERRInvalidModuleImportOrExport, "Module import or export sta
 LSC_ERROR_MSG(1088, ERRInvalidExportName, "Unable to resolve module export name")
 
 LSC_ERROR_MSG(1089, ERRLetIDInLexicalDecl, "'let' is not an allowed identifier in lexical declarations")
+LSC_ERROR_MSG(1090, ERREvalNotSupportedInNonStrictAsync, "'eval' is not allowed in async function in non-strict mode")
```
ianwjhalliday: `not allowed` sounds misleading, as if we are correct and this is the spec

test/es6/default.js
```diff
@@ -284,7 +284,7 @@ var tests = [
         assert.throws(function () { eval("function f(a, b = function () { function f() { eval('1'); } }) { }") }, SyntaxError, "Evals in nested child functions are not allowed in the parameter scope", "'eval' is not allowed in the default initializer");
         assert.throws(function () { eval("function f(a, b = eval('a')) { }") }, SyntaxError, "Eval is not allowed in the parameter scope", "'eval' is not allowed in the default initializer");
         assert.throws(function () { eval("async function f(a = eval('b')) { }"); }, SyntaxError, "Eval is not allowed in the param scope of async functions", "'eval' is not allowed in the default initializer");
-        assert.throws(function () { eval("function f(a = async function(y) { eval('b'); }) { }"); }, SyntaxError, "Eval is not allowed in the param scope of nested async functions", "'eval' is not allowed in the default initializer");
```
ianwjhalliday: Unrelated to this change but I just noticed the default param eval error message

`'eval' is not allowed in the default initializer`

This is not specific enough.  Which default initializer? Is there only one?

`'eval' is not allowed in a default initializer` or
`'eval' is not allowed in the default initializer of a parameter`

ianwjhalliday:
Looks good to me.  We need to make sure this is what we want to do before merging though.

pleath:
LGTM.

tcare:
Looks good.

aneeshdk:
Followed up on the thread https://github.com/Microsoft/ChakraCore/pull/859

PR #817, opened 2016-04-15
Have RegExp.prototype.test work correctly when lastIndex > MaxCharCount

goyakin:


goyakin:
@jianchun, could you take a look?

jianchun:
LGTM

PR #818, opened 2016-04-15
MSFT: 6967505 setInterval(alert(), 10000/237) throws Invalid calling object exception after domtimeline script is injected

Yongqu:
In debugger when we used to put all globaobject properties in the ActivationObject and use that
as "this" pointer for global object. We should separate out the ActivationObject from GlobalObject and
use the GlobalObject as "this"

### DIFF
--- a/lib/Runtime/Debug/DiagObjectModel.h
+++ b/lib/Runtime/Debug/DiagObjectModel.h
@@ -144,11 +144,12 @@ namespace Js
 
     enum FrameWalkerFlags
     {
-        FW_None                 = 0x0,
-        FW_MakeGroups           = 0x1,  // Make groups such as [Scope], [Globals] etc.
-        FW_EnumWithScopeAlso    = 0x2,  // While walking include the with scope as well.
-        FW_AllowLexicalThis     = 0x4,  // Do not filter out Js::PropertyIds::_lexicalThisSlotSymbol
-        FW_AllowSuperReference  = 0x8,  // Allow walking of Js::PropertyIds::_superReferenceSymbol and Js::PropertyIds::_superCtorReferenceSymbol
+        FW_None                   = 0x0,
+        FW_MakeGroups             = 0x1,  // Make groups such as [Scope], [Globals] etc.
+        FW_EnumWithScopeAlso      = 0x2,  // While walking include the with scope as well.
+        FW_AllowLexicalThis       = 0x4,  // Do not filter out Js::PropertyIds::_lexicalThisSlotSymbol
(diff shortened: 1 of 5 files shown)


Yongqu:
@agarwal-sandeep can you take a look?

lib/Runtime/Debug/DiagObjectModel.h
```diff
@@ -149,6 +149,7 @@ namespace Js
         FW_EnumWithScopeAlso    = 0x2,  // While walking include the with scope as well.
         FW_AllowLexicalThis     = 0x4,  // Do not filter out Js::PropertyIds::_lexicalThisSlotSymbol
         FW_AllowSuperReference  = 0x8,  // Allow walking of Js::PropertyIds::_superReferenceSymbol and Js::PropertyIds::_superCtorReferenceSymbol
+        FW_SkipGlobal           = 0x10, // Do not add global object directly.
```
agarwal-sandeep: I think more appropriate name is FW_DontAddGlobalsDirectly as this is not skipping globals as the name suggest without reading the comment.

agarwal-sandeep:
Beside name change it looks good to me

lib/Runtime/Debug/DiagObjectModel.h
```diff
@@ -353,10 +354,20 @@ namespace Js
                             }
                             else
                             {
+                                PropertyIndex propertyIndex = activeScopeObject->GetPropertyIndex(resolveObject.propId);
```
agarwal-sandeep: This will always return invalid property index as activeScopeObject doens't have this propId. We need to call GetPropertyIndex on resolveObject.obj or resolveObject.originalInstance

lib/Runtime/Debug/DiagObjectModel.h
```diff
@@ -353,10 +354,20 @@ namespace Js
...
+                                PropertyAttributes propertyAttribute;
+                                if (!activeScopeObject->GetAttributesWithPropertyIndex(resolveObject.propId, propertyIndex, &propertyAttribute))
+                                {
+                                    propertyAttribute = resolveObject.isConst ? PropertyConstDefaults : PropertyDynamicTypeDefaults;
+                                }
+                                else
+                                {
+                                    Assert(((propertyAttribute & (PropertyLetConstGlobal | PropertyConst)) == (PropertyLetConstGlobal | PropertyConst)) == !!resolveObject.isConst);
+                                }
```
agarwal-sandeep: Please add tests for this change

PR #819, opened 2016-04-18
Build lib/Runtime/Debug

digitalinfinity:
1. Remove dependency on StrCmpLogicalW
2. Define a few enums in CommonPal
3. Convert __try/__finally to TryFinally
4. Run make in parallel build mode to speed up local builds


digitalinfinity:
@jianchun can you review please?

jianchun:
LGTM 👍

build.sh
```diff
@@ -99,5 +99,5 @@ else
     cmake $CC_PREFIX -DCMAKE_BUILD_TYPE=Release ..
 fi
 
-make |& tee build.log
+make -j $(nproc) |& tee build.log
```
obastemur: can we make/keep this optional please ? i.e. we can extend build.sh with `-j` option ? (I can take that item)

digitalinfinity: Sure, you can extend build.sh- but curious, @obastemur, why do you want this optional?

obastemur: Multi-core compile option (make) is really good for building from scratch, building after a small change etc.  IMO, the output is not desirable while updating a common header file with template of a template inside. It could be annoying to update `build.sh` to see what's going on.

digitalinfinity:
Rebased the PR commit on top of latest Linux branch- thanks for the review @Jianchun. @obastemur, feel free to issue a PR to make -j optional

PR #828, opened 2016-04-19
Simd.js:Enforce simd value coercion in asmjs return

### LINKED ISSUE #827: SIMD.js: Asmjs validation allows returning simd values from functions without coercion.
Asmjs validation should require coercion for return values. Simd values needs to be coreced using the check() builtin before returning from asmjs functions. currently this is not enforced. 

**Repro Case**
```javascript
function module(stdlib) {
    "use asm"
    var b = stdlib.SIMD.Uint8x16;
    function foo() {
        var abc = b(1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1);
        return abc;
    }
    return { foo:foo }
}
```

**Expected results**
>	Asm.js Compilation Error function : module::foo
 	Expression for returning simd values must be coerced using check()
  Asm.js compilation failed.

**Actual Results**
Successfully compiled asm.js code

arunetm:
Simd values needs to be coreced using the check() builtin before returning
from asmjs functions.

Fixes #827

### DIFF
--- a/lib/Runtime/Language/AsmJsByteCodeGenerator.cpp
+++ b/lib/Runtime/Language/AsmJsByteCodeGenerator.cpp
@@ -84,6 +84,17 @@ namespace Js
 #define CheckNodeLocation(info,type) if(!mFunction->IsValidLocation<type>(&info)){\
     throw AsmJsCompilationException( _u("Invalid Node location[%d] "), info.location ); }
 
+    template<class NodeType>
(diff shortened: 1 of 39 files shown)


arunetm:
@abchatra can you please review?

nmostafa:
Let's close this one then. Coercion is optional per https://github.com/tc39/ecmascript_simd/issues/307

arunetm:
Closing this PR as coercion has been made optional for return types in asm.js

PR #831, opened 2016-04-19
Introduce double extended prefix for opcodes

commonlisp:
For certain upcoming work (e.g., WebAssembly and perhaps SIMD.js later on), we are starting to exhaust the extended prefix space. One option is to introduce another prefix to expand the opcode space.


commonlisp:
@MikeHolman @Cellule Could you please review this when you get a chance? Thanks!

MikeHolman:
+@pleath, @ianwjhalliday for visibility

lib/Runtime/ByteCode/OpLayoutsAsmJs.h
```diff
@@ -61,6 +61,9 @@ namespace Js
         return (uint)(Js::OpCodeAsmJs::Simd128_End - Js::OpCodeAsmJs::Simd128_Start) + 1 + (uint)(Js::OpCodeAsmJs::Simd128_End_Extend - Js::OpCodeAsmJs::Simd128_Start_Extend) + 1;
     }
 
+    inline bool IsByteOpCode(Js::OpCodeAsmJs op)        { return op < ((uint)Js::OpCodeAsmJs::MaxByteSizedOpcodes + 1); }                                       /* op == [0 - 256) ?       */
+    inline bool IsExtendedOpCode(Js::OpCodeAsmJs op)    { return !IsByteOpCode(op) && (uint)op < 2 * ((uint)Js::OpCodeAsmJs::MaxByteSizedOpcodes + 1); }        /* op == [0 - 2 * 256) ?  */
```
nmostafa: Nit: misleading comment. Should read: op == [256 - 2 * 256) ? Same for the one below.

MikeHolman:
lgtm

commonlisp:
@ianwjhalliday To make room for the prefix, we have to move three single byte ops to extended. We chose SetConcatStrMultiItem* and SpreadArrayLiteral thus far, but would there be three better candidates for moving over to extended?

lib/Runtime/Language/InterpreterLoop.inl
```diff
@@ -157,14 +157,25 @@ SWAP_BP_FOR_OPCODE:
                 return nullptr;
 
             case INTERPRETER_OPCODE::ExtendedOpcodePrefix:
+            case INTERPRETER_OPCODE::DblExtendedOpcodePrefix:
```
MikeHolman: Discussing offline with @curtisman, he made a good point that this should have its own case arm.

nmostafa: To reduce branch mispredictions on the if-condition below ? Anything else ?

nmostafa: I agree, I think we should also surround it with `#ifdef INTERPRETER_ASMJS` as well, to not increase non-asmjs interpreter code size which doesn't use DblExtended prefixes.

lib/Runtime/Language/InterpreterLoop.inl
```diff
@@ -232,15 +243,27 @@ SWAP_BP_FOR_OPCODE:
                 break;
             }
             case INTERPRETER_OPCODE::ExtendedMediumLayoutPrefix:
+            case INTERPRETER_OPCODE::DblExtendedMediumLayoutPrefix:
             {
 #ifndef INTERPRETER_ASMJS  // Asmjs doesn't have any extended opcodes for now, remove that case
```
nmostafa: @MikeHolman, I see Extended opcodes for Medium/Large layouts are disabled for AsmJs interpreter. Is that still valid ? If so, then perhaps those case arms should be removed entirely since neither Asmjs or normal interpreter use them ?

MikeHolman: Hmm... I didn't know this was here, but it doesn't seem valid. I think this will cause bug in interpreter dispatch on SIMD in asm.js when you have medium/large bytecode register number.

nmostafa: Yes, and WASM. I think we just never ran anything with over 256 SIMD bytecode regs. Will fix this separately against release/1.2.

lib/Runtime/Language/InterpreterLoop.inl
MikeHolman: I don't think this should increase code size for non-asm.js, because there shouldn't be any cases using the double extended macro (the only case would be default).

nmostafa: But if DblExtendedOpcodePrefix gets its own case arm per your suggestion above, then we will have to replicate the code below, and that case is never taken on non-asm.js interpreter.

MikeHolman: Ah yeah, that is true. I agree with you, having this behind #ifdef is good idea.

nmostafa: @commonlisp , can you please take care of this along with having a separate case arm for DblExtendedOpcodePrefix ?

commonlisp:
@dotnet-bot test this please

commonlisp:
@dotnet-bot test this please

commonlisp:
@dotnet-bot test this please

nmostafa:
@commonlisp , @ianwjhalliday
please take a look at https://github.com/Microsoft/ChakraCore/pull/908
We cleaned up some unused SIMD opcodes. With that, I think you don't have to move opcodes to extended space.

ianwjhalliday:
@nmostafa rad, thanks!  @commonlisp if we don't have to move the opcodes anymore that'd be my preference before merging this change.

commonlisp:
We going to handle the extended opcode prefix a little differently. We'll have the prefix reader pull in two bytes instead of one for a more efficient and general solution.

PR #833, opened 2016-04-20
Adding back the flag for enabling opHelper checks only on jshost and ch hosts

satheeshravi:
OpHelper Checks will be disabled on WebCrawler runs. These asserts do not seem make enough sense when they come from the crawler runs - as reverse engineering the scenario seems pretty tough.
These asserts in DbCheckPostLowerer do not imply that there is a functional bug - it will indicate more of a code layout issue.
With Louis' advice, these checks are enabled only on jshost and ch hosts.

### DIFF
--- a/bin/ch/ChakraRtInterface.h
+++ b/bin/ch/ChakraRtInterface.h
@@ -136,6 +136,9 @@ public:
     static HRESULT SetCheckMemoryLeakFlag(bool flag) { return CHECKED_CALL(SetCheckMemoryLeakFlag, flag); }
     static HRESULT SetEnableCheckMemoryLeakOutput(bool flag) { return CHECKED_CALL(SetEnableCheckMemoryLeakOutput, flag); }
 #endif
+#ifdef DEBUG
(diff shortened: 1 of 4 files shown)


satheeshravi:
@LouisLaf : please review.

lib/Backend/GlobOpt.cpp
```diff
@@ -7303,9 +7303,11 @@ GlobOpt::ValueNumberDst(IR::Instr **pInstr, Value *src1Val, Value *src2Val)
...
         {
-            // Src1 may be too large to represent as a signed int32, and src2 may be zero. Since the result can therefore be too
-            // large to represent as a signed int32, include Number in the value type.
-            return CreateDstUntransferredValue(ValueType::GetNumberAndLikelyInt(true), instr, src1Val, src2Val);
+            // Src1 may be too large to represent as a signed int32, and src2 may be zero. 
+            // Since the result can therefore be too large to represent as a signed int32, 
+            // include Number in the value type.
+            return CreateDstUntransferredValue(
+                ValueType::AnyNumber.SetCanBeTaggedValue(true), instr, src1Val, src2Val);
         }
```
LouisLaf: Why is this showing up here?

LouisLaf:
:+1:

PR #839, opened 2016-04-20
SIMD.js/Asm.js : Adding codegen sequences for Float32x4.min/max and tests

### LINKED ISSUE #837: SIMD.js/Asm.js : Float32x4.min/Float32x4.max may produce wrong result if one of the first opnd's lanes is NaN
Two SIMD.js ops: **Float32x4.min()** and **Float32x4.max()** produce a wrong result when there's a NaN value in the first operand and a number in the second operand. 

For example, if the following two SIMD values are compared both **mix** and **max** ops will set the fourth lane of the resulting value to 42. 

```
    var one = SIMD.Float32x4(1, 1,  1, NaN)
    var two = SIMD.Float32x4(1, 1,  1,  42)
```

Here's a simple test case exposing this issue: 

[b208.js.txt](https://github.com/Microsoft/ChakraCore/files/228422/b208.js.txt)

Expected Output:
```
testMin = SIMD.Float32x4(NaN, NaN, NaN, NaN)
testMax = SIMD.Float32x4(NaN, NaN, NaN, NaN)
```

Actual Output:
```
testMin = SIMD.Float32x4(NaN, NaN, NaN, 3.140000104904175)
testMax = SIMD.Float32x4(NaN, NaN, NaN, 3.140000104904175)
```

Krovatkin:
Fixes #837

### DIFF
--- a/lib/Backend/LowerMDSharedSimd128.cpp
+++ b/lib/Backend/LowerMDSharedSimd128.cpp
@@ -353,6 +353,10 @@ IR::Instr* LowererMD::Simd128LowerUnMappedInstruction(IR::Instr *instr)
     case Js::OpCode::Simd128_Max_U8:
         return Simd128LowerMinMax(instr);
 
+    case Js::OpCode::Simd128_Min_F4:
+    case Js::OpCode::Simd128_Max_F4:
+        return Simd128LowerMinMax_F4(instr);
+
     case Js::OpCode::Simd128_MinNum_F4:
     case Js::OpCode::Simd128_MaxNum_F4:
         return Simd128LowerMinMaxNum(instr);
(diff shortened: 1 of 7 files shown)


Krovatkin:
This is a fix for #837

Krovatkin:
@MikeHolman Hello Michael! Could you please take a look when you have a chance? Thanks! FYI: @nmostafa

dilijev:
Hi @Krovatkin would you mind rebasing your changes to a single commit before this is merged in?

Krovatkin:
@dilijev will do!)

test/Bugs/b208_asmjs.js
```diff
@@ -0,0 +1,67 @@
...
+    }
+   
+    return {func1:func1, func2:func2};
+}
+
+var m = asmModule(this, {});
+
+m.func1();
+m.func1();
+m.func1();
```
MikeHolman: I think you shouldn't need all these calls

nmostafa: Please enable this test with type-spec ON. For which two calls / function are enough.
-bgjit- -simdjs -simd128typespec -mic:1 -lic:1 -off:simplejit

Krovatkin: @MikeHolman I don't unless we want to re-use this test for -simd128typespec which @nmostafa wants to do.  

Three calls don't seem to be enough for -simd128typespec to even kick off compilations:

```
m.func1();
m.func1();
m.func2();
m.func2();

print (m.func1());
print (m.func2());

```
$ ../../Build/VcBuild/bin/x86_debug/ch.exe -bgjit- -simdjs -simd128typespec -mic:1 -lic:1 -off:simplejit -trace:backend  b208_asmjs.js
SIMD.Float32x4(NaN, NaN, NaN, NaN)
SIMD.Float32x4(NaN, NaN, NaN, NaN)
```

File: lib/Backend/LowerMDSharedSimd128.cpp [hunk 2]
```diff
@@ -2320,6 +2324,68 @@ IR::Instr* LowererMD::Simd128LowerMinMax(IR::Instr* instr)
...
+    IR::RegOpnd* tmp1 = IR::RegOpnd::New(src1->GetType(), m_func);
+    IR::RegOpnd* tmp2 = IR::RegOpnd::New(src2->GetType(), m_func);
+
+    if (instr->m_opcode == Js::OpCode::Simd128_Min_F4)
+    {
+        pInstr = IR::Instr::New(Js::OpCode::MINPS, tmp1, src1, src2, m_func);
+        instr->InsertBefore(pInstr);
+        Legalize(pInstr);
+        //
+        pInstr = IR::Instr::New(Js::OpCode::MINPS, tmp2, src2, src1, m_func);
```
MikeHolman: Why not do what intel manual suggests for this case, which is to do comparison followed by AND, ANDN, and OR? Something like:
`
tmp1 = CMPLTPS src1, src2
tmp2 = ANDPS src1, tmp1
tmp3 = ANDNPS src2, tmp1
dst = ORPS tmp2, tmp3
`

test/Bugs/b208_asmjs.js
MikeHolman: Ok, sounds reasonable to me. For asm.js there is different max interpret count flag (-maic). If you want to test typespec, you probably want to do -asmjs- for this case.

lib/Backend/LowerMDSharedSimd128.cpp
Krovatkin: That's a very good question :-)))) I ported our polyfill implementation and it runs this intrinsics sequence: 

```
        tmp1.m128_value = _mm_min_ps(tmpaValue.m128_value, tmpbValue.m128_value);
        tmp2.m128_value = _mm_min_ps(tmpbValue.m128_value, tmpaValue.m128_value);
        x86Result.m128_value = _mm_or_ps(tmp1.m128_value, tmp2.m128_value);
``` 

Unless there's a clear benefit of using CMP/AND/ANDN/OR we might want to keep the two in sync for consistency and maintainability reasons. If there's indeed a substantial benefit from using CMP/AND/ANDN/OR we would probably want to re-implement our polyfill too? 

Do you guys @MikeHolman @nmostafa think it might be worthwhile to do bench both implementations?

test/Bugs/b208_asmjs.js
Krovatkin: @MikeHolman Ah, my bad! I didn't realize I was missing  ```-asmjs-``` 2 calls are good enough to expose failiures in both, asmjs and simd128typespec modes :-)

lib/Backend/LowerMDSharedSimd128.cpp
Krovatkin: @MikeHolman The existing polyfill implementation seems to be using one less register and one less instruction, though I don't know how expensive  _mm_min_ps actually is :-)

MikeHolman: I guess it is question of whether or not `MINPS` is more expensive than `ANDPS+ANDNPS`. And I think it is probably close enough that sticking with the current method is reasonable approach.

MikeHolman:
lgtm

lib/Backend/LowerMDSharedSimd128.cpp
nmostafa: CMPLTPS considers -0.0 and +0.0 equal, while MINPS returns the second operand. The ORPS will guarnatee the -0.0 is picked.

PR #840, opened 2016-04-20
decommit unusable page(s) in medium heap blocks (bug 5822302)

### LINKED ISSUE #742: Decommit unused page in medium block
In medium block we use 8 pages for one block, in some cases we can have the whole last page not used. for example, if object size is 0x1600, we only uses first 7 pages and can hold 5 objects.

leirocks:
in medium heap blocks, if the object size is bigger than 1 page, it can cause the whole last 1~3 page(s) never been allocated. when such case is hit, decommit those pages to save memory as well as capture corruption. when returning the pages back to page allocator, we should commit those pages again for reuse, in case of OOM here, just decommit all pages in the heap block and let the page allocator to manage the decommitted pages.
in rescan code, asserting that we never scan those unallocatable pages

Fixes #742

### DIFF
--- a/lib/Common/Memory/HeapInfo.cpp
+++ b/lib/Common/Memory/HeapInfo.cpp
@@ -505,10 +505,11 @@ HeapInfo::Initialize(Recycler * recycler
     if (pageheapmode == PageHeapMode::PageHeapModeOff)
     {
 #ifdef ENABLE_DEBUG_CONFIG_OPTIONS
-        isPageHeapEnabled = recycler->GetRecyclerFlagsTable().PageHeap != PageHeapMode::PageHeapModeOff;
-        pageheapmode = (PageHeapMode)recycler->GetRecyclerFlagsTable().PageHeap;
-        blockTypeFilter = (PageHeapBlockTypeFilter)recycler->GetRecyclerFlagsTable().PageHeapBlockType;
-        pBucketNumberRange = &recycler->GetRecyclerFlagsTable().PageHeapBucketNumber;
+        auto& flags = recycler->GetRecyclerFlagsTable();
+        isPageHeapEnabled = flags.PageHeap != PageHeapMode::PageHeapModeOff;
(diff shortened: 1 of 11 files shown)


leirocks:
@digitalinfinity

digitalinfinity:
The behavior on free is a little concerning in the case where commit failed. Is it possible to just call PartialDecommit on the decommitted page to update the PageAllocator state and then call release on the entire set of pages? Release doesn't always just decommit, it can also call VirtualFree in some cases, so just decommitting if the commit fails might not be enough.

dilijev:
Is this tied to https://github.com/Microsoft/ChakraCore/issues/742?

leirocks:
yes

lib/Common/Memory/HeapInfo.cpp
```diff
@@ -1766,12 +1766,61 @@ BOOL SmallAllocationBlockAttributes::IsAlignedObjectSize(size_t sizeCat)
...
 {
     return HeapInfo::IsAlignedMediumObjectSize(sizeCat);
 }
+/* static */
+uint MediumAllocationBlockAttributes::GetUnusablePageCount(size_t sizeCat)
+{
+    return ((MediumAllocationBlockAttributes::PageCount*AutoSystemInfo::PageSize) % sizeCat) / AutoSystemInfo::PageSize;
+}
+/* static */
+void MediumAllocationBlockAttributes::ProtectUnusablePages(HeapBlock* heapBlock)
```
digitalinfinity: I'm not sure why this is here and not in the HeapBlock- GetUnusablePageCount can be argued as getting an attribute of the block, this is definitely not an attribute. Also, nitpicky but please fix spacing of usage of *- not sure why its the only operator that you use with no spaces

leirocks: done

digitalinfinity:
:shipit:

leirocks:
need to merge to master, see #985

PR #841, opened 2016-04-20
Fix optimized element access that incorrectly reports that the instance has no such property

pleath:
Reimplement shortcut in element access on unknown property name. The optimization requires a prototype chain walk to confirm that the property being looked for can't be found on an object such as a CEO that may use names unknown to the JS engine, or that, like a proxy, may require that the lookup be done regardless. For now the prototype walk is done on every attempted access where no property record exists for the index name. This could be optimized with a scheme similar to the one we use for non-writable properties in the prototype chain, but I'd like to see a needful use case before I saddle us with another registration mechanism. Benchmarks are flat.

### DIFF
--- a/lib/Runtime/Language/JavascriptOperators.cpp
+++ b/lib/Runtime/Language/JavascriptOperators.cpp
@@ -76,10 +76,8 @@ namespace Js
         }
     }
 
-    IndexType GetIndexType(Var indexVar, ScriptContext* scriptContext, uint32* index, PropertyRecord const ** propertyRecord, JavascriptString ** propertyNameString, bool createIfNotFound, bool preferJavascriptStringOverPropertyRecord)
+    IndexType GetIndexTypeFromPrimitive(Var indexVar, ScriptContext* scriptContext, uint32* index, PropertyRecord const ** propertyRecord, JavascriptString ** propertyNameString, bool createIfNotFound, bool preferJavascriptStringOverPropertyRecord)
     {
-        indexVar = JavascriptConversion::ToPrimitive(indexVar, JavascriptHint::HintString, scriptContext);
-
         // CONSIDER: Only OP_SetElementI and OP_GetElementI use and take advantage of the
         // IndexType_JavascriptString result. Consider modifying other callers of GetIndexType to take
         // advantage of non-interned property strings where appropriate.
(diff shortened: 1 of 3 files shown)


pleath:
@Yongqu @ianwjhalliday Reimplemented this so that we only do extra work when we've discovered that no PropertyRecord exists for the given property name. Seems to get rid of perf regressions.

pleath:
Ping @Yongqu @ianwjhalliday

lib/Runtime/Language/JavascriptOperators.cpp
```diff
@@ -450,26 +458,35 @@ namespace Js
...
+                    Assert(propertyRecord != nullptr);
+                }
 
+                if (propertyRecord != nullptr)
+                {
+                    if (!JavascriptOperators::GetProperty(instance, object, propertyRecord->GetPropertyId(), &member, scriptContext))
+                    {
+                        // If the instance doesn't have the property, typeof result is "undefined".
+                        threadContext->CheckAndResetImplicitCallAccessorFlag();
+                        threadContext->AddImplicitCallFlags(savedImplicitCallFlags);
```
Yongqu: how do we know there is implicit call?

lib/Runtime/Language/JavascriptOperators.cpp
```diff
@@ -4959,6 +5010,50 @@ namespace Js
...
+            return false;
+        }
+        if (DynamicType::Is(instance->GetTypeId()) && 
+            static_cast<DynamicObject*>(instance)->GetTypeHandler()->IsStringTypeHandler())
+        {
+            return false;
+        }
+        return !(instance->HasDeferredTypeHandler() &&
+                 JavascriptFunction::Is(instance) && 
+                 JavascriptFunction::FromVar(instance)->IsExternalFunction());
```
Yongqu: can you remind me again which externalfunction doesn't have normal type handlers?

Yongqu:
:+1:

lib/Runtime/Language/JavascriptOperators.cpp
pleath: I think this is conservative; I don't know why it was originally added.

PR #843, opened 2016-04-21
Fix bug with handling of OOM while processing jit transfer data

rajatd:
We allocate equivalent type caches and register the current entry point info on the thread context as having equivalent type caches. Each of these operations could throw OOM. 
The resetting of entrypoint's state in the OOM handling code should be agnostic of the success or failure of the aforementioned operations.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/843)


rajatd:
@pleath @LouisLaf

pleath:
LGTM.

rajatd:
Thanks!

PR #844, opened 2016-04-21
Disable obj-type spec for Init*LetFld and Init*ConstFld opcodes

rajatd:
The lowerer doesn't expect InitLetFld to be obj-type speced. While it is
legal to obj-type spec InitLetFld and changing the lowerer would be one
way to fix this issue, it would not be performant as we don't share types
of activation objects yet, leading to repeated Obj-type spec bailouts for
InitLetFld. Same reasoning applies to other opcodes Init*LetFld and
Init*ConstFld opcodes.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/844)


pleath:
LGTM.

PR #845, opened 2016-04-21
Remove JavascriptErrorDebug from Linux builds

digitalinfinity:
This is used only for WinRT errors, so remove use of it on non-Windows builds

### DIFF
--- a/lib/Runtime/Library/JavascriptError.cpp
+++ b/lib/Runtime/Library/JavascriptError.cpp
@@ -300,37 +300,51 @@ namespace Js
         Assert(pError != nullptr);
 
         PCWSTR varName = (pei ? pei->bstrDescription : nullptr);
+#ifdef _WIN32
         if (useErrInfoDescription)
         {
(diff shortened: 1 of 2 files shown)


lib/Runtime/Library/JavascriptError.cpp
```diff
@@ -300,37 +300,51 @@ namespace Js
...
+#else
+#define CREATE_ERROR(create_method, get_type_method, err_type)   \
+    { \
+        Assert(perrinfo == nullptr); \
+        pError = library->create_method(); \
+    }
+#endif
+
 #define THROW_ERROR_IMPL(err_method, create_method, get_type_method, err_type) \
     static JavascriptError* create_method(ScriptContext* scriptContext, IErrorInfo* perrinfo, RestrictedErrorStrings * proerrstr) \
```
jianchun: Unrelated, `err_method` isn't used?

jianchun:
LGTM

digitalinfinity:
Not in that particular function but it's used as the method name for several other methods generated by THROW_ERROR_IMPL. Thanks for the review!

PR #847, opened 2016-04-21
[wasm] Refactor LoadWasmScript()

commonlisp:
Refactor LoadWasmScript() into several smaller functions. The smaller functions aren't private member functions of ScriptContext because that would complicate the ScriptContext header dependencies and order. In the very least, this helps to scope all the locals better.


MikeHolman:
Thanks for the refactor. This LGTM.

PR #848, opened 2016-04-21
Undo defer parse mode issue with formals

aneeshdk:
In undo defer parse mode we reconstruct the pid ref stack for the formals.
With my last change to insert a new reference for each param symbol in the
body I missed making change in the undo defer parse mode. This change adds
the code to do that for both split scope and non-split scope.

### DIFF
--- a/test/es6/default-splitscope-undodeferparse.js
+++ b/test/es6/default-splitscope-undodeferparse.js
@@ -1,0 +1,64 @@
+//-------------------------------------------------------------------------------------------------------
+// Copyright (C) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE.txt file in the project root for full license information.
+//-------------------------------------------------------------------------------------------------------
+
+function foo1(a, b) {
+    if (b != 10) {
+        print("FAILED")
+    } else {
+        print("PASSED");
+    }
+    if (eval('b') != 10) {
+        print("FAILED")
+    } else {
+        print("PASSED");
+    }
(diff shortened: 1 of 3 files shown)


aneeshdk:
@pleath @abchatra @tcare Can you please take a look?

test/es6/default-splitscope-undodeferparse.js
```diff
@@ -0,0 +1,64 @@
+//-------------------------------------------------------------------------------------------------------
+// Copyright (C) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE.txt file in the project root for full license information.
+//-------------------------------------------------------------------------------------------------------
+
+function foo1(a, b) {
+    if (b != 10) {
+        print("FAILED")
+    } else {
+        print("PASSED");
```
tcare: I don't we can have tests in RL like this because it only cares if the last line is 'PASSED'. If any previous tests fail, their failure will be ignored.

Can we use the unit test framework or use something other than 'PASSED' for the intermediate status?

aneeshdk: No actually RL does look at each line and if any of them is not "Passed" it will report the test as failed. So we are fine on this.

tcare: Ah, must have been fixed a while back :)

pleath:
LGTM. Thanks.

tcare:
LGTM

PR #851, opened 2016-04-21
Disable an invalid assert in the array check hoisting code (and correct and clarify the assert in case we decide to try and re-enable it at some point).

pleath:


rajatd:
:+1:

PR #852, opened 2016-04-21
OS 7249217: Rest is able to be in a slot in arguments optimization case

tcare:
We hit an assert with rest in a case where we trigger an arguments scope object optimization. In the case where we don't actually need to create a scope object, rest was hard coded to not be in a slot. This was based on an assumption from a very long time ago; since then rest is no different to other formals when determining whether it is in a slot. I removed the special casing and added the test.


tcare:
@pleath @goyakin @akroshg please review

pleath:
LGTM.

tcare:
Forgot to call the test function.

akroshg:
Looks good to me as well.

PR #855, opened 2016-04-21
Fix bug with negative 0 handling in global optimizer

rajatd:
The forward pass sets a bit, m_wasNegativeZeroPreventedByBailout, on an
operand if it was int-type-specialized and was protected by a
BailOutOnNegativeZero (which would happen if there are uses of that
operand downstream in manners which require distinction between -0 and +0
and the forward pass cannot say for sure that the operand can't be -0).
The Deadstore pass then looks at this bit to determine if it can remove
some -0 bailouts.

However, this bit being set to false for an operand doesn't necessarily mean that that
operand cannot be -0 if the operation that produced this operand wasn't type-spec'ed at all.
Negative zero tracking in the deadstore pass wasn't taking this into
account and was getting rid of more -0 bailouts than it should have,
resulting in wrong results.

Fix being proposed here is,
- for the current instruction, detect the source operand for whose defintion we can remove negative zero bailout.
- if there is any bailout instruction between this definition and the current instruction, then don't remove the negative zero bailout.

Reasoning behind this is that if we removed -0 bailout for, say, s1 and 
- a bailout happened between def of s1 and an int specialized use of s1, and
- s1 was actually -0 (but we removed the bailout),
s1 will be +0 in the interpreter leading to wrong results.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/855)


lib/Backend/BackwardPass.cpp
```diff
@@ -5096,38 +5162,55 @@ BackwardPass::TrackIntUsage(IR::Instr *const instr)
...
+                if (tag == Js::BackwardPhase)
                 {
-                    // -0 does not matter for dst, or this instruction does not generate -0 since one of the srcs is not -0
-                    // (regardless of -0 bailout checks)
-                    SetNegativeZeroDoesNotMatterIfLastUse(instr->GetSrc1());
-                    SetNegativeZeroDoesNotMatterIfLastUse(instr->GetSrc2());
-                    break;
+                    // Could have type specialized opcodes in the Backward Pass for asmjs
+                    if (instr->ignoreNegativeZero ||
+                        !(instr->GetSrc1()->IsRegOpnd() && instr->GetSrc1()->AsRegOpnd()->m_wasNegativeZeroPreventedByBailout) ||
```
rajatd: I could ideally get rid of this condition all together since m_wasNegativeZeroPreventedByBailout starts off as false and is only set in the forward pass

rajatd:
@LouisLaf - could you review this change?

lib/Backend/BackwardPass.cpp
```diff
@@ -4934,6 +4957,26 @@ BackwardPass::TrackBitWiseOrNumberOp(IR::Instr *const instr)
 }
 
 void
+BackwardPass::RemoveNegativeZeroBailout(IR::Instr* instr)
+{
+    Assert(instr->HasBailOutInfo() && (instr->GetBailOutKind() & IR::BailOutOnNegativeZero));
+    IR::BailOutKind bailOutKind = instr->GetBailOutKind();
+    bailOutKind -= IR::BailOutOnNegativeZero;
```
LouisLaf: WOuldn't bit mask be cleaner than -=?

lib/Backend/BackwardPass.cpp
```diff
@@ -4967,29 +5010,52 @@ BackwardPass::TrackIntUsage(IR::Instr *const instr)
...
-            if(tag == Js::DeadStorePhase && instr->HasBailOutInfo())
+            if (negativeZeroDoesNotMatterBySymId->Test(dstSym->m_id))
             {
-                IR::BailOutKind bailOutKind = instr->GetBailOutKind();
-                if(bailOutKind & IR::BailOutOnNegativeZero)
+                instr->ignoreNegativeZero = true;
+            }
+
+            if (tag == Js::DeadStorePhase)
+            {
```
LouisLaf: A comment on why we clear the whole bv would be good.

lib/Backend/BackwardPass.cpp
```diff
@@ -5149,26 +5232,47 @@ BackwardPass::TrackIntUsage(IR::Instr *const instr)
...
+                        SetNegativeZeroDoesNotMatterIfLastUse(instr->GetSrc1());
+                        SetNegativeZeroDoesNotMatterIfLastUse(instr->GetSrc2());
+                        break;
+                    }
+                    goto NegativeZero_Sub_Default;
                 }
-                goto NegativeZero_Sub_Default;
-
+                else
+                {
```
LouisLaf: Another comment here as to why this is different from BackwardPhase.

LouisLaf:
:+1:

PR #856, opened 2016-04-21
OS 2970825: Spread should throw a syntax error with unary operators

tcare:
This is a bug is around from an early spec draft. Spread should throw a syntax error when it is not immediately preceeded by a comma, LParen, or LBrack. This fixes an assert that happens when we try to do codegen. There is no fre impact as it gets caught by assignment to rvalue.

### DIFF
--- a/lib/Parser/Parse.cpp
+++ b/lib/Parser/Parse.cpp
@@ -8007,6 +8007,11 @@ ParseNodePtr Parser::ParseExpr(int oplMin,
 
         m_pscan->Scan();
 
+        if (m_token.tk == tkEllipsis) {
+            // ... cannot have a unary prefix.
+            Error(ERRUnexpectedEllipsis);
+        }
+
         if (nop == knopYield && !m_pscan->FHadNewLine() && m_token.tk == tkStar)
         {
             m_pscan->Scan();
@@ -8026,8 +8031,8 @@ ParseNodePtr Parser::ParseExpr(int oplMin,
         }
         else
         {
-            // Disallow spread after an Ellipsis token. This prevents chaining, and ensures spread is the top level expression.
-            pnodeT = ParseExpr<buildAST>(opl, &fCanAssign, TRUE, nop != knopEllipsis && fAllowEllipsis, nullptr /*hint*/, nullptr /*hintLength*/, nullptr /*hintOffset*/, &operandToken, true);
+            // Disallow spread after a unary operator.
+            pnodeT = ParseExpr<buildAST>(opl, &fCanAssign, TRUE, FALSE, nullptr /*hint*/, nullptr /*hintLength*/, nullptr /*hintOffset*/, &operandToken, true);
         }
 
         if (nop != knopYieldLeaf)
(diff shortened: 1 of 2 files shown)


tcare:
@ianwjhalliday In the bug for this you mentioned that we should refactor the code by extracting out the ParseExpr handling of the ... operator, since it's not a real expression operator. Instead, we would explicitly scan ... in ParseArgList and ParseArrayLiteral.

This would be nice, but it requires more refactoring than that. The problem is the following case:
(...args) => blah

When we first scan the ..., we are treating the current context as a comma separated expression list. We mark the ... as a potential error and defer emitting it. When we see a =>, the error is no longer valid as it is now a rest parameter. At that point, we rewind and start reparsing the arg list as a lambda arg list. Because of this ambiguity it is not sufficient to handle ... just in ParseArgList and ParseArrayLiteral.

@pleath suggested that we could work around this by extracting out the parsing of the paren comma separated expression lists. We could then check for the ... explicitly before parsing every potential expression. Unfortunately this is a little invasive for refactoring, because of the flow of parsing the expressions in ParseExpr recursively.

Because of the complexity of handling the deferred (...a) error case, I am going to skip the refactoring for now. This might be a good issue to accept pull requests for as you mentioned.

As for getting rid of fAllowEllipsis, I would like to change that to be a member variable that we save and restore the state on the frame that requires the change. I also decided not to do this because we have several fAllowWhatever parameters around the parser that I think we could also do this with. I would also like to file an issue to fix these and reduce the amount of parameters we have to keep passing true/false to.

tcare:
@abchatra please review (or anyone else who sees this :))

tcare:
@Microsoft/chakra-es please review (trying out the new team mention!)

boingoing:
This fix looks good to me.

abchatra:
:+1: Please run test262 and make sure you haven't regressed any tests.

ianwjhalliday:
Sorry I am a little late to this party :)

Good write up about what happened, @tcare, much appreciated.

I am thinking about the ambiguity with `...` in a parenthesized list.  If I am not mistaken the `...` only can appear in a parameter list, so seeing it during parenthesized list parsing implies that either 1) the list is actually a parameter list for an arrow function, or 2) it is a syntax error.  That means we could stop parsing as a parameter list when encountering a `tkEllipis` token (specifically after a `tkLParen` or `tkComma`) and rewind back to the `tkLParen` and begin parsing as an arrow function, just like we do when we see the `tkDArrow` at the end the parenthesized list.

If it were in fact a syntax error because there was no `tkDArrow` afterward, then we would error out there and all is well.  Then we can do the refactoring I'd like to see.  The deferred potential error would also no longer be necessary.

No need to do this now for 1.2 release of course.  Just thoughts for simplifying the parsing code in the future.

pleath:
If we stop parsing as a parenthetical expression immediately on seeing … , does that not mean that we’ll report errors specific to parameter lists, even when we really do have a parenthetical expression, and the … is a syntax error?

--Paul

ianwjhalliday:
@pleath I don't follow.  What would be an example?

PR #857, opened 2016-04-22
PRE bookkeeping fix

pleath:
When we clear out live fields on discovering that implicit calls prevent loop optimization, we leave the same syms in the loop's initialValueFieldMap, resulting in later asserts. Two things: should use the KillAllFields API to clear out the live fields BV, and that API should walk the loop hierarchy setting the allFieldsKilled bits.

### DIFF
--- a/lib/Backend/GlobOptFields.cpp
+++ b/lib/Backend/GlobOptFields.cpp
@@ -310,7 +310,10 @@ GlobOpt::KillLiveFields(StackSym * stackSym, BVSparse<JitArenaAllocator> * bv)
         bv->Clear(propertySym->m_id);
         if (this->IsLoopPrePass())
         {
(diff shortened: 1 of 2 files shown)


pleath:
@LouisLaf @rajatd

lib/Backend/GlobOptFields.cpp
```diff
@@ -382,7 +382,10 @@ GlobOpt::KillAllFields(BVSparse<JitArenaAllocator> * bv)
     bv->ClearAll();
     if (this->IsLoopPrePass())
     {
-        this->rootLoopPrePass->allFieldsKilled = true;
+        for (Loop * loop = this->rootLoopPrePass; loop != nullptr; loop = loop->parent)
+        {
+            loop->allFieldsKilled = true;
```
LouisLaf: We have the same issue when setting loop->fieldKilled.  Can you please fix this as well?

PR #858, opened 2016-04-22
Implement ICU based platform agnostic interface on Linux builds

digitalinfinity:
This change implements the UnicodeText Platform Agnostic interface on Linux builds.
With this change, Linux builds will temporarily have an ICU dependency
(we'll do another change later to make it optional but it shouldn't be too difficult).

Additionally, one new API is introduced to the UnicodeText interface- ChangeStringCaseInPlace.
This is a thin wrapper around CharUpperBuffer/CharLowerBuff on Win32.
On Linux it just falls back to the ICU based function ChangeStringLinguisticCase.

Note that while the ICU functions are generally analogous to their Windows counterparts, they
need not be identical in behavior. This can cause differences between Windows and Linux but we'll
deal with them as we encounter them.

Finally, one additional function was re-implemented for Linux. LogicalStringCompareImpl is a
reimplementation of CompareStringW(NORM_IGNORECASE | SORT_DIGITSASNUMBERS)

### DIFF
--- a/lib/Runtime/PlatformAgnostic/Platform/Common/UnicodeText.Common.cpp
+++ b/lib/Runtime/PlatformAgnostic/Platform/Common/UnicodeText.Common.cpp
@@ -1,0 +1,282 @@
+//-------------------------------------------------------------------------------------------------------
+// Copyright (C) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE.txt file in the project root for full license information.
+//-------------------------------------------------------------------------------------------------------
+
+#include "RuntimePlatformAgnosticPch.h"
+#include "UnicodeText.h"
+#include <ctype.h>
+
+namespace PlatformAgnostic
+{
+namespace UnicodeText
+{
+CharacterTypeFlags GetLegacyCharacterTypeFlags(char16 ch)
+{
+    Assert(ch >= 128);
+
+    const char16 lineSeperatorChar = 0x2028;
+    const char16 paraSeperatorChar = 0x2029;
+
+    if (ch == lineSeperatorChar || ch == paraSeperatorChar)
+    {
+        return LineCharGroup;
+    }
+
+    CharacterClassificationType charClass = GetLegacyCharacterClassificationType(ch);
+
+    if (charClass == CharacterClassificationType::Letter)
+    {
+        return LetterCharGroup;
+    }
+    else if (charClass == CharacterClassificationType::DigitOrPunct)
+    {
+        return IdChar;
+    }
+    else if (charClass == CharacterClassificationType::Whitespace)
+    {
+        return SpaceChar;
+    }
+
+    return UnknownChar;
+}
+
+#if !defined(_WIN32) || ENABLE_TEST_PLATFORM_AGNOSTIC
+namespace Internal
+{
+template <typename CharType>
+inline bool isDigit(__in CharType c)
+{
+    return c >= '0' && c <= '9';
+}
+
+template <typename CharType>
+inline int readNumber(__inout CharType* &str)
+{
+    int num = 0;
+
+    // Count digits on each string
+    while (isDigit(*str))
+    {
+        int newNum = (num * 10) + (*str - '0');
+        Assert(newNum > num);
+        if (newNum <= num)
+        {
+            return num;
+        }
+
+        num = newNum;
+
+        str++;
+    }
+
+    return num;
+}
+
+///
+/// Implementation of CompareString(NORM_IGNORECASE | SORT_DIGITSASNUMBERS)
+/// This code is in the common library so that we can unit test it on Windows too
+///
+/// Basic algorithm is as follows:
+/// 
+/// Iterate through both strings
+///  If either current character is not a number, compare the rest of the strings lexically
+///  else if they're both numbers:
+///     Count the leading zeroes in both strings and skip over them
+///     Read the two numbers
+///     If they're not the same, return accordingly
+///     If they are the same, but the leading zeroes are different, return whichever has fewer zeroes as "larger"
+///     Otherwise, they're the same to continue scanning
+///  else they're both not numbers:
+///     Compare lexically till we find a number
+/// The algorithm treats the characters in a case-insensitive manner
+///
+template <typename CharType>
+int LogicalStringCompareImpl(__in const CharType* p1, __in const CharType* p2)
+{
+    Assert(p1 != nullptr);
+    Assert(p2 != nullptr);
+
+    while (*p1 && *p2)
+    {
+        bool isDigit1 = isDigit(*p1);
+        bool isDigit2 = isDigit(*p2);
+
+        if (isDigit1 ^ isDigit2)
+        {
+            // If either character exclusively is not a number, compare just the characters
+            // since that's enough to sort the string
+            CharType c1 = tolower(*p1);
+            CharType c2 = tolower(*p2);
+
+            Assert(c1 != c2);
+
+            if (c1 < c2)
+            {
+                return -1;
+            }
+
+            return 1;
+        }
+        else if (isDigit1 && isDigit2)
+        {
+            // Both current characters are digits, so we'll compare the numbers
+            int numZero1 = 0, numZero2 = 0;
+
+            // Count leading zeroes on each string
+            while (*p1 == '0')
+            {
+                p1++; numZero1++;
+            }
+
+            while (*p2 == '0')
+            {
+                p2++; numZero2++;
+            }
+
+            int num1 = readNumber(p1);
+            int num2 = readNumber(p2);
+
+            if (num1 != num2)
+            {
+                // If the numbers are unequal, just compare the numbers
+                return (num1 > num2) ? 1 : -1;
+            }
+            else if (numZero1 != numZero2)
+            {
+                // The numbers are equal but the number of leading zeroes are not
+                // The one with the fewer number of leading zeroes is considered
+                // "larger" (gets sorted earlier)
+                return (numZero1 < numZero2 ? 1 : -1);
+            }
+            // else both numbers are the same, keep going
+        }
+        else
+        {
+            // Both characters are not digits - scan till we find a
+            // digit in the same position in each string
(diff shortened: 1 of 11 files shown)


digitalinfinity:
@jianchun Please review

lib/Runtime/PlatformAgnostic/RuntimePlatformAgnosticPch.h
```diff
@@ -15,6 +17,42 @@
...
 #endif
 
+// Minimal definitions to use AssertMsg in the PAIL
+#ifndef _WIN32
+#define DbgRaiseAssertionFailure() __builtin_trap()
+#define __analysis_assume(x)
+#define __in
+#define __inout
+
+#ifndef TRUE
```
obastemur: `TRUE` and `FALSE` are already defined from `pal/inc/pal_mstypes.h`

lib/Runtime/PlatformAgnostic/Platform/Linux/UnicodeText.ICU.cpp
```diff
@@ -101,13 +113,172 @@ namespace PlatformAgnostic
...
+            }
+            else
+            {
+                Assert(false);
+            }
+
+            if (U_FAILURE(errorCode))
+            {
+                *pErrorOut = TranslateUErrorCode(errorCode);
+                return -1;
```
obastemur: Is `-1` here intended? function return type is `uint32`

obastemur: Actually ICU `u_strToLower` and `u_strToUpper` return types are `int32_t`, why don't we use it instead?

lib/Runtime/PlatformAgnostic/Platform/Common/UnicodeText.Common.cpp
```diff
@@ -0,0 +1,278 @@
...
+///  else they're both not numbers:
+///     Compare lexically till we find a number
+///
+template <typename CharType>
+int LogicalStringCompareImpl(__in const CharType* str1, __in const CharType* str2)
+{
+    Assert(str1 != nullptr);
+    Assert(str2 != nullptr);
+
+    CharType* p1 = const_cast<CharType*>(str1);
```
dreamer-dead: Why this cast?
It seems that isDigit function can take just a |const CharType*|

lib/Runtime/PlatformAgnostic/Platform/Common/UnicodeText.Common.cpp
```diff
@@ -0,0 +1,278 @@
...
+
+    while (*p1 && *p2)
+    {
+        bool isDigit1 = isDigit(*p1);
+        bool isDigit2 = isDigit(*p2);
+
+        // If either character is not a number, compare lexically
+        if (isDigit1 ^ isDigit2)
+        {
+            while (*p1 && *p2)
```
dreamer-dead: Are STL functions allowed in this project?
E.g. http://en.cppreference.com/w/cpp/algorithm/lexicographical_compare

lib/Runtime/PlatformAgnostic/RuntimePlatformAgnosticPch.h
digitalinfinity: I tried to avoid taking a dependency on the PAL here

obastemur: Actually I was also suspecting that ( i.e. You have removed a ref to common pal) 

Could you please share the reason? Probably it's not like some parts use PAL, others custom?

digitalinfinity: This is supposed to be the layer that replaces the PAL long term (hence called Runtime/PlatformAgnostic). It would be odd for it to take a dependency on the PAL.

lib/Runtime/PlatformAgnostic/Platform/Linux/UnicodeText.ICU.cpp
digitalinfinity: Done- I had originally defined ChangeStringLinguisticCase on Windows as uint- I don't remember why, even the Windows variant works fine with the int32 return type, so changed it.

lib/Runtime/PlatformAgnostic/Platform/Common/UnicodeText.Common.cpp
digitalinfinity: Yup, unnecessary- removed.

lib/Runtime/PlatformAgnostic/Platform/Common/UnicodeText.Common.cpp
digitalinfinity: Thank you @dreamer-dead - wasn't aware of lexicographical_compare, very interesting.
Generally, STL functions are ok on a case-by-case basis, STL classes/types are generally avoided for a variety of reasons (memory model inconsistencies, perf in some cases, etc). In this particular case, I would have been ok using lexicographical_compare. Looking at its definition though, couple things stand out:
- It needs the beginning and end of the sequence
- It returns a bool (so either strictly less than or not)

LogicalStringCompareImpl checks for equality, greater than or less than. Given this requirement, using lexicographical_compare would involve the following (I think- please tell me if there is a simpler way):
- Create two std::string instances from p1 and p2
- Check first for equality
- Check next with lexicographical_compare

It seems like we'd have done at least three passes at this point (once to determine the beginning/end of the string, once for the equality check and once for the lexicographical_compare. The current implementation just does a single pass.

Is there a simpler way that I'm not seeing? If not, I'd prefer to keep the current implementation.

lib/Runtime/PlatformAgnostic/Platform/Common/UnicodeText.Common.cpp
```diff
@@ -0,0 +1,275 @@
...
+                {
+                    return -1;
+                }
+                else
+                {
+                    return 1;
+                }
+            }
+
+            return 0;
```
jianchun: This seems incorrect. They can be unequal. Need to check if *p1 or *p2 is not null terminator.

lib/Runtime/PlatformAgnostic/Platform/Common/UnicodeText.Common.cpp
```diff
@@ -0,0 +1,275 @@
...
+
+template <typename CharType>
+inline int readNumber(__inout CharType* &str)
+{
+    int num = 0;
+
+    // Count digits on each string
+    while (isDigit(*str))
+    {
+        num = (num * 10) + (*str - '0');
```
jianchun: Wonder what behavior is expected when int overflows.

lib/Runtime/PlatformAgnostic/Platform/Common/UnicodeText.Common.cpp
```diff
@@ -0,0 +1,275 @@
...
+            {
+                p2++; numZero2++;
+            }
+
+            int num1 = readNumber(p1);
+            int num2 = readNumber(p2);
+
+            if (num1 != num2)
+            {
+                // The two numbers have the same number of digits
```
jianchun: This comment confuses me.

lib/Runtime/PlatformAgnostic/Platform/Common/UnicodeText.Common.cpp
jianchun: Actually this branch looks weird. If `isDigit(c1) != isDigit(c2)`, then `c1 != c2`. This doesn't need a `while` loop and only needs to compare c1 with c2 and return.

lib/Runtime/PlatformAgnostic/RuntimePlatformAgnosticPch.h
jianchun: Do we need `TRUE`/`FALSE` on Linux? Maybe they are only used on `Windows` implementation?

obastemur: Actually I was planning to replace all `TRUE/FALSE` to simple `true/false` ?

digitalinfinity: The Assertion code uses TRUE/FALSE today, as does parts of the ICU. I could remove the definitions and include <Unicode/umachine.h> instead to get TRUE/FALSE from there. If @obastemur changes the TRUE/FALSE in the Assertion code to using true/false, then we could move this #include out of the pch and into UnicodeText.ICU.cpp since that would be the only code that would need it.

digitalinfinity: Actually, no, not going to do that since I don't want a dependency on an ICU header in the pch (we want ICU to be optional). I'd say leave this as is and we can remove as part of cleaning up the Assertions code later.

lib/Runtime/PlatformAgnostic/Platform/Common/UnicodeText.Common.cpp
digitalinfinity: Agreed that code could be simplified- done

lib/Runtime/PlatformAgnostic/Platform/Common/UnicodeText.Common.cpp
digitalinfinity: Wasn't sure if overflow needed to be handled but decided that it's probably better to just do it- added a truncating overflow behavior

lib/Runtime/PlatformAgnostic/Platform/Common/UnicodeText.Common.cpp
digitalinfinity: Vestigial- cleaned up

digitalinfinity:
Thanks for the review @Jianchun. Updated PR.

digitalinfinity:
Updated the PR to rebase on top of the latest branch and squashed out the code review comment commits- merging now.

PR #859, opened 2016-04-22
Moving async functionality under experimental flag

aneeshdk:
Moving the async functions feature under experimental flag


ljharb:
sad :-( why?

bterlson:
@ljharb do not despair, it will make its return! We found a few rather severe issues with our internal fuzzer recently. We could fix them with our current implementation but they suggest a different implementation strategy (closer to generators rather than a source transform). We could have left them enabled but only in strict mode, or we could have made a bunch of error conditions that we relax once the bugs are fixed, but I think it's better to just hold off for 1.2. The proposal is also still stage 3 which we'd prefer not to ship :-P

abchatra:
:+1:

PR #861, opened 2016-04-23
put asm.js output message behind trace flag

MikeHolman:
This was causing fuzzers to give error on difference in output.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/861)


MikeHolman:
@Cellule can you take a look?

abchatra:
👍

Cellule:
lgtm

satheeshravi:
I understand that diffing on trace output is bad. But why is the property name different in this case for a given test ? - Is it just printed in a different order ?

MikeHolman:
It's different if you compare against -asmjs-

PR #862, opened 2016-04-25
Few destructuring fixes

akroshg:
The IsArgument optimization code did not handle the pattern nodes. Added
'cases' for those nop. This happens when we try to find out the
'arguments' usage from a parse node. Here in this case we have an empty pattern and this leads to the assert.
We have place holder slots for destructuring patterns - however when we
generate the scopeinfo for deferred function we didn't account that piece
of information. Fixed that by using the same counter (earlier used for same name args place holder) and increment that for destructuring patterns.
Added unittest for those.


akroshg:
@tcare @aneeshdk please have a look.

aneeshdk:
LGTM

tcare:
Looks good

akroshg:
Thanks

PR #864, opened 2016-04-26
cross platform: extend build script

obastemur:


build.sh
```diff
@@ -18,17 +18,23 @@ PRINT_USAGE() {
...
+    echo "-j          : Multicore build (i.e. -j:3 for 3 cores)"
+    echo "-icu        : Path to ICU include folder (see example below)"
+    echo "--help, -h  : Show help"
     echo "--debug, -d : Debug build"
     echo "CXX=<path to clang++>"
     echo "CC=<path to clang>"
     echo ""
-    echo "example: ./build.sh CXX=/path/to/clang++ CC=/path/to/clang"
+    echo "example : ./build.sh CXX=/path/to/clang++ CC=/path/to/clang -j:2"
+    echo "with icu: ./build.sh -icu:/usr/local/Cellar/icu4c/version/include/"
```
jianchun: Command line switch styles seem to become inconsistent. Can we consider sth. like `--CXX=... --CC=... --icu-path=...`? Similar original make style `-j [N], --jobs[=N]`, or only `--jobs=N`.

build.sh
```diff
@@ -18,17 +18,23 @@ PRINT_USAGE() {
...
+    echo ""
 }
 
 _CXX=""
 _CC=""
 DEBUG_BUILD=0
+MULTICORE_BUILD=""
```
jianchun: Can we make `-j $(nproc)` the default? Having the default to support multi-core makes most scenarios faster.

build.sh
obastemur: Agree on that but shouldn't that change be part of another commit/pr?

build.sh
obastemur: nproc is not cross platform.

Apart from dealing with cross platform nproc.. 

Having multicore mode enabled; I have failed to see my system doesn't have uchar.h. Enabling multicore by default is no good for development and this work is still under development.

build.sh
obastemur: Adding '--jobs' option

jianchun: > Agree on that but shouldn't that change be part of another commit/pr?

Thanks, either is fine with me.

build.sh
jianchun: Ok.

dilijev:
Hey @obastemur, I think the typical style for Unix scripts and tools is to use `--option=value` rather than using a colon. Could you make that switch?

obastemur:
@dilijev good point. It should accept both of them now.

build.sh
```diff
@@ -18,27 +18,33 @@ PRINT_USAGE() {
...
+    echo "--jobs, -j  : Multicore build (i.e. -j:3 for 3 cores)"
+    echo "--icu       : Path to ICU include folder (see example below)"
+    echo "--help, -h  : Show help"
     echo "--debug, -d : Debug build"
-    echo "CXX=<path to clang++>"
-    echo "CC=<path to clang>"
+    echo "--ccx=<path to clang++>"
+    echo "--cc=<path to clang>"
+    echo ""
+    echo "example : ./build.sh --ccx=/path/to/clang++ --cc:/path/to/clang -j=2"
```
jianchun: Could you please also change `:` to `=` in `--cc:...`?

`ccx` looks fine but deviates from `CXX` and harder to read the script. Maybe `cxx` is more consistent and better?

The printed options are now in descendent order. Should we reverse to ascendant?

obastemur:
Updated the styling and help output accordingly.

@jianchun, @dilijev Thanks for the reviews.

Issue #865, opened 2016-04-26
Simd.js Bug: Incorrect reuse of unused simd temp registers in IRBuilder register allocation


arunetm:
Incorrect register reuse for unused simd temp registers. If the simd function call is not part of an assignment a temp dst register is created and since its unused its marked for reuse. However the stack symbol associated to this tmp register is typed and will assert when reused for a simd value with a different type. 

**Repro Case**
```javascript
function asmModule(stdlib) {
    "use asm";
    var i4 = stdlib.SIMD.Int32x4;
    var i4check = i4.check;
    var u4 = stdlib.SIMD.Uint32x4;
    var u4add = u4.add;
    function func1(a, b)
    {
        a = i4check(a);
        b = i4check(b);
        var x = u4(-1, -2, -3, -4);
        u4add(x,x);
        return i4check(a);
    }
    return {func1:func1};
}
var m = asmModule(this);
var ret2 = m.func1(SIMD.Int32x4(-1, -2, -3, -4),SIMD.Int32x4(-1, -2, -3, -4));
```

**Expected Results**
>Successfully compiled asm.js code

**Actual Results**
>Successfully compiled asm.js code
ASSERTION 15960: (G:\arun\git\chakra-411\core\lib\Backend\Sym.cpp, line 382) stackSym->GetType() == type
 Failure: (stackSym->GetType() == type)
FATAL ERROR: ch.exe failed due to exception code c0000420

Issue #867, opened 2016-04-26
Simd.js Bug: Destination register type mismatch assertion while returning simd types from asm functions. 


arunetm:
This seems like a register allocation bug when nested calls are involved. The issue happens when creating/finding register for dstOpnd during the irbuilding phase of the SIMD128ReturnU4 opcode. There seem to be a mismatch in the return register type (Expecting U4 and finding I4)

**Repro Case**

>
```javascript
function asmModule(stdlib, imports) {
    "use asm";
    var i4 = stdlib.SIMD.Int32x4;
    var i4check = i4.check;
    var u4 = stdlib.SIMD.Uint32x4;
    var u4check = u4.check;
    function func1(a)
    {
        a = i4check(a);
        return a;
    }
    function func2()
    {
        var x = u4(-1, -2, -3, -4);
        var k = i4(-1, -2, -3, -4);;
        k = i4check(func1(k));
        return u4check(x);  //Deliberate use of x here
    }
    return {func1:func1, func2:func2};
}
var m = asmModule(this, {g1:SIMD.Float32x4(90934.2,123.9,419.39,449.0), g2:SIMD.Uint32x4(-1065353216, -1073741824,-1077936128, -1082130432)});
var ret2 = m.func2();

**Flags**
>x64_debug
>-DumpOnCrash  -bgjit-  -on:asmjs -testtrace:asmjs -simdjs -AsmJsStopOnError -maic:0 -lic:0


**Actual result**
>
Successfully compiled asm.js code
ASSERTION 30176: (G:\arun\git\chakra-310\core\lib\Backend\Sym.cpp, line 382) stackSym->GetType() == type
 Failure: (stackSym->GetType() == type)
dump filename C:\Users\b-arpuru\AppData\Local\Temp\\CH_30176_18664.dmp
FATAL ERROR: ch.exe failed due to exception code c0000420

abchatra:
@arunetm  Can this issue be closed now?

arunetm:
Yes. The fix had already been merged to Release/1.2

Issue #866, opened 2016-04-26
Simd.js Bug: Simd type conversions in asmjs sets the dst type incorrectly


arunetm:
The helper method used to build IR for simd type conversion operations sets the destination register type to 'TySimd128I4' always. 

**Repro Case**
```javascript
function asmModule(stdlib) {
    "use asm";
    var i4 = stdlib.SIMD.Int32x4;
    var i4check = i4.check;
    var u4 = stdlib.SIMD.Uint32x4;
    var u4swizzle = u4.swizzle;
    var u4shuffle = u4.shuffle;
    var u4fi4 = u4.fromInt32x4Bits;
    var i4fu4 = i4.fromUint32x4Bits;
    function func1(a)
    {
        a = i4check(a);
        var x = u4(0, 0, 0, 0);
        x = u4swizzle(x, 3, 2, 1, 0);
        x = u4shuffle(u4fi4(a), x, 0, 1, 7, 6);
        return i4check(i4fu4(x));
    }
    return {func1:func1};
}
var m = asmModule(this);
var ret2 = m.func1(SIMD.Int32x4(1,2,3,4));
```

**Expected Results**
>Successfully compiled asm.js code
PASS

**Actual Results**
>Successfully compiled asm.js code
ASSERTION 16232: (G:\arun\git\chakra-411\core\lib\Backend\Sym.cpp, line 382) stackSym->GetType() == type
 Failure: (stackSym->GetType() == type)
FATAL ERROR: ch.exe failed due to exception code c0000420

abchatra:
@arunetm Can this issue be closed now?

arunetm:
@abchatra, yes the fix has already been merged.

PR #868, opened 2016-04-26
SIMD.js Bugfix: IRBuilder Register allocation in Asm.js

arunetm:
1. Type assertion while reusing function retrun registers for SIMD values.
2. Incorrect reuse of unused simd temp registers.
3. Incorrect destination register type for simd conversion operations.

Fixes: #865, #866 and  #867


arunetm:
@MikeHolman @abchatra Please review.

MikeHolman:
lgtm

PR #869, opened 2016-04-26
Fixing bug with FromVar hoisting

rajatd:
A combination of math type spec, copy-prop, nested loops and invariant hoisting is required to repro this bug. We end up hoisting a FromVar from a region where it doesn't need a bailout to one where it does. We should then use the value type of the src in the destination region to put the right bailout on FromVar.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/869)

### DIFF
--- a/lib/Backend/GlobOpt.cpp
+++ b/lib/Backend/GlobOpt.cpp
@@ -13734,57 +13734,60 @@ GlobOpt::ToTypeSpecUse(IR::Instr *instr, IR::Opnd *opnd, BasicBlock *block, Valu
             Assert(loop->bailOutInfo);
         }
 
-        if(toType == TyInt32 && opcode == Js::OpCode::FromVar)
+        if (opcode == Js::OpCode::FromVar)
         {
-            Assert(valueInfo);
-            if(lossy)
+            
+            if (toType == TyInt32)
             {
-                if(!valueInfo->IsPrimitive() && !IsTypeSpecialized(varSym, block))
+                Assert(valueInfo);
+                if (lossy)
                 {
-                    // Lossy conversions to int32 on non-primitive values may have implicit calls to toString or valueOf, which
-                    // may be overridden to have a side effect. The side effect needs to happen every time the conversion is
-                    // supposed to happen, so the resulting lossy int32 value cannot be reused. Bail out on implicit calls.
-                    Assert(DoLossyIntTypeSpec());
+                    if (!valueInfo->IsPrimitive() && !IsTypeSpecialized(varSym, block))
+                    {
+                        // Lossy conversions to int32 on non-primitive values may have implicit calls to toString or valueOf, which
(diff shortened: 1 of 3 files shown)


rajatd:
@LouisLaf @pleath

nmostafa:
*[lib/Backend/GlobOpt.cpp, line 18503 \[r1\]](https://reviewable.io:443/reviews/microsoft/chakracore/869#-KGKERA-xYNsERx9McjE:-KGKERA-xYNsERx9McjF:-3253831) ([raw file](https://github.com/microsoft/chakracore/blob/b1a0cf7cec3d17535070e1725ce72afc4fcc10cc/lib/Backend/GlobOpt.cpp#L18503)):*
If this is the src of a FromVar, shouldn't the sym always be a Var (i.e. IsTypeSpec() == false) ?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/microsoft/chakracore/869)*

rajatd:
@nmostafa You're right. Thanks

lib/Backend/GlobOpt.cpp
```diff
@@ -18456,34 +18459,68 @@ GlobOpt::OptHoistInvariant(
...
+                Assert(src1Val->GetValueNumber() == landingPadSrc1val->GetValueNumber());
+
+                ValueInfo *src1ValueInfo = src1Val->GetValueInfo();
+                ValueInfo *landingPadSrc1ValueInfo = landingPadSrc1val->GetValueInfo();
+                IRType dstType = dst->GetType();
+
+                // A definite type in the source position and not a definite type in the destination (landing pad)
+                // and no bailout on the instruction; we should put a bailout on the hoisted instruction.
+                if ((dstType == TyInt32 && src1ValueInfo->IsInt() && !landingPadSrc1ValueInfo->IsInt()) ||
+                    (dstType == TyFloat64 && src1ValueInfo->IsNumber() && !landingPadSrc1ValueInfo->IsNumber()) ||
```
LouisLaf: Need lossy check here...

lib/Backend/GlobOpt.cpp
```diff
@@ -18456,34 +18459,68 @@ GlobOpt::OptHoistInvariant(
...
+            }
+            break;
+
+        case Js::OpCode::FromVar:
+        {
+            StackSym* src1StackSym = IR::RegOpnd::TryGetStackSym(instr->GetSrc1());
+
+            // We may be hoisting FromVar from a region where it didn't need a bailout (src1 had a definite value type) to a region
+            // where it would. In such cases, the FromVar needs a bailout based on the value type of src1 in its new position.
+            if (src1StackSym && !instr->HasBailOutInfo() && bailoutKind != IR::BailOutInvalid)
```
LouisLaf: Probably should have an assert that if it has bailout, it is one of the expected ones.

LouisLaf:
:+1:

PR #870, opened 2016-04-26
MSFT: 7334254

Yongqu:
I wanted to use recycler temp allocator because we can have shorter lifetime for different objects in
moduleRecord than the scriptContext. At it turned out we need to keep various helper structures alive
longer due to debugger reparsing of source code, and the module will be kept alive by the JavascriptLibrary
anyhow. Recycler tempallocator is not designed to be used for this long period. Change it to use the
general arena allocator.


Yongqu:

boingoing:
LGTM

PR #871, opened 2016-04-27
Have getters on RegExp.prototype return default values when called with RegExp.prototype

### LINKED ISSUE #760: RegExp.prototype getters and methods need a web-compat hack
ES6 specified that RegExp.prototype is no longer a RegExp object and as a result, eg. RegExp.prototype.global would result in an exception (as is our current semantics under /es6all). However this change is not web compatible and, rather than going back and making RegExp.prototype a RegExp object, the various methods simply return default values if `this` doesn't have an [[OriginalFlags]] internal slot.

See also https://github.com/tc39/ecma262/issues/262 and https://github.com/tc39/ecma262/issues/511.

goyakin:
Fixes #760.

### DIFF
--- a/lib/Runtime/Library/JavascriptRegularExpression.cpp
+++ b/lib/Runtime/Library/JavascriptRegularExpression.cpp
@@ -233,6 +233,12 @@ namespace Js
         }
     }
 
+    bool JavascriptRegExp::ShouldApplyPrototypeWebWorkaround(Arguments& args, ScriptContext* scriptContext)
+    {
+        return scriptContext->GetConfig()->IsES6PrototypeChain() && \
+               args.Info.Count >= 1 && args[0] == scriptContext->GetLibrary()->GetRegExpPrototype();
+    }
+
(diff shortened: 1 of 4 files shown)


goyakin:

lib/Runtime/Library/JavascriptRegularExpression.cpp
```diff
@@ -963,6 +969,12 @@ namespace Js
         ARGUMENTS(args, callInfo);
         Assert(!(callInfo.Flags & CallFlags_New));
 
+        ScriptContext* scriptContext = function->GetScriptContext();
+        if (ShouldApplyPrototypeWebWorkaround(args, scriptContext))
+        {
+            return scriptContext->GetLibrary()->GetUndefined();
+        }
+
         return GetJavascriptRegExp(args, _u("RegExp.prototype.options"), function->GetScriptContext())->GetOptions();
```
akroshg: nit: you could change this to scriptContext.

akroshg:
Rest looks good to me.

PR #873, opened 2016-04-27
Build ChakraCore shared library

digitalinfinity:
For easier reviewing, I've split up this PR into five different commits. Each commit has a more detailed description of its changes. 

With this change, ChakraCore builds on debug builds in Linux and can run hello world (have not tried release builds yet)

### DIFF
--- a/bin/ChakraCore/CMakeLists.txt
+++ b/bin/ChakraCore/CMakeLists.txt
@@ -1,4 +1,4 @@
-add_library (ChakraCoreLib SHARED
+add_library (ChakraCore SHARED
     ChakraCoreDllFunc.cpp
     ConfigParserExternals.cpp
     TestHooks.cpp
@@ -6,10 +6,10 @@ add_library (ChakraCoreLib SHARED
 
 # By default, don't export any symbols from this library
 # We'll manually export the relevant individual functions
-set_target_properties(ChakraCoreLib PROPERTIES CXX_VISIBILITY_PRESET hidden)
+set_target_properties(ChakraCore PROPERTIES CXX_VISIBILITY_PRESET hidden)
   
 target_include_directories (
-  ChakraCoreLib PUBLIC
+  ChakraCore PUBLIC
   ${CMAKE_CURRENT_SOURCE_DIR}
   ${CHAKRACORE_SOURCE_DIR}/lib/Common
   ${CHAKRACORE_SOURCE_DIR}/lib/Runtime
@@ -32,8 +32,8 @@ target_include_directories (
 #  dl: For shared library loading related functions
 #  icuuc: For the ICU (xplat-todo: Make this optional)
 #
-target_link_libraries(ChakraCoreLib
-  -Wl,--no-undefined
+target_link_libraries(ChakraCore
+  -Wl,-undefined,error
   -Wl,--start-group
   -Wl,--whole-archive
   Chakra.Jsrt
@@ -60,7 +60,6 @@ target_link_libraries(ChakraCoreLib
   pthread
   stdc++
   dl
-  gcc_s
   icuuc
   -Wl,--version-script=${CMAKE_CURRENT_SOURCE_DIR}/libChakraCoreLib.version
 )
@@ -68,7 +67,7 @@ target_link_libraries(ChakraCoreLib
 #
 # Post build step to copy the built shared library
 # to BuildLinux (or whatever the CMakeBuildDir is)
-add_custom_command(TARGET ChakraCoreLib POST_BUILD
+add_custom_command(TARGET ChakraCore POST_BUILD
   COMMAND ${CMAKE_COMMAND} -E copy_if_different 
-  "${CHAKRACORE_BINARY_DIR}/bin/ChakraCore/libChakraCoreLib.so"
+  "${CHAKRACORE_BINARY_DIR}/bin/ChakraCore/libChakraCore.so"
    ${CHAKRACORE_BINARY_DIR}/)
--- a/bin/ChakraCore/CMakeLists.txt
+++ b/bin/ChakraCore/CMakeLists.txt
@@ -1,9 +1,74 @@
-add_library (ChakraCore
-    ChakraCore.def
-    ChakraCoreDLLFunc.cpp
+add_library (ChakraCoreLib SHARED
+    ChakraCoreDllFunc.cpp
     ConfigParserExternals.cpp
     TestHooks.cpp
-    )
+)
 
+# By default, don't export any symbols from this library
+# We'll manually export the relevant individual functions
+set_target_properties(ChakraCoreLib PROPERTIES CXX_VISIBILITY_PRESET hidden)
+  
 target_include_directories (
-    ChakraCore PUBLIC ${CMAKE_CURRENT_SOURCE_DIR})
+  ChakraCoreLib PUBLIC
+  ${CMAKE_CURRENT_SOURCE_DIR}
+  ${CHAKRACORE_SOURCE_DIR}/lib/Common
+  ${CHAKRACORE_SOURCE_DIR}/lib/Runtime
+  ${CHAKRACORE_SOURCE_DIR}/lib/Parser
+  ${CHAKRACORE_SOURCE_DIR}/lib/Jsrt
+  )
+
+#
+# Link step for the ChakraCore shared library
+#
+# This requires that the library have no undefined
+# symbols after linking
+# Chakra.Jsrt and Chakra.Jsrt.Core have their archives
+# imported wholesale since all functions used in these
+# libraries are considered to be used by Chakra
+#
+# External libraries we link with are the following:
+#  pthread: For threading
+#  stdc++/gcc_s: C++ runtime code
+#  dl: For shared library loading related functions
+#  icuuc: For the ICU (xplat-todo: Make this optional)
+#
+target_link_libraries(ChakraCoreLib
+  -Wl,--no-undefined
+  -Wl,--start-group
+  -Wl,--whole-archive
+  Chakra.Jsrt
+  Chakra.Jsrt.Core
+  -Wl,--no-whole-archive
+  Chakra.Runtime.Types
+  Chakra.Runtime.Math
+  Chakra.Runtime.Library
+  Chakra.Runtime.Language
+  Chakra.Runtime.Debug
+  Chakra.Runtime.ByteCode
+  Chakra.Runtime.PlatformAgnostic
+  Chakra.Runtime.Base
+  Chakra.Parser
+  Chakra.Common.Util
+  Chakra.Common.Memory
+  Chakra.Common.Common
+  Chakra.Common.Core
+  Chakra.Common.DataStructures
+  Chakra.Common.Exceptions
+  Chakra.Common.Codex
+  -Wl,--end-group
+  Chakra.Pal
+  pthread
+  stdc++
+  dl
+  gcc_s
+  icuuc
+  -Wl,--version-script=${CMAKE_CURRENT_SOURCE_DIR}/libChakraCoreLib.version
+)
+
+#
+# Post build step to copy the built shared library
+# to BuildLinux (or whatever the CMakeBuildDir is)
+add_custom_command(TARGET ChakraCoreLib POST_BUILD
+  COMMAND ${CMAKE_COMMAND} -E copy_if_different 
+  "${CHAKRACORE_BINARY_DIR}/bin/ChakraCore/libChakraCoreLib.so"
+   ${CHAKRACORE_BINARY_DIR}/)
--- a/bin/ch/ChakraRtInterface.cpp
+++ b/bin/ch/ChakraRtInterface.cpp
@@ -8,7 +8,7 @@
 LPCSTR chakraDllName = "chakracore.dll";
 #else
 #include <dlfcn.h>
-LPCSTR chakraDllName = "libChakraCoreLib.so";
+LPCSTR chakraDllName = "libChakraCore.so";
 #endif
 
 bool ChakraRTInterface::m_testHooksSetup = false;
--- a/bin/ch/ChakraRtInterface.cpp
+++ b/bin/ch/ChakraRtInterface.cpp
@@ -4,7 +4,12 @@
 //-------------------------------------------------------------------------------------------------------
 #include "stdafx.h"
 
-LPCWSTR chakraDllName = _u("chakracore.dll");
+#ifdef _WIN32
+LPCSTR chakraDllName = "chakracore.dll";
+#else
+#include <dlfcn.h>
+LPCSTR chakraDllName = "libChakraCoreLib.so";
+#endif
 
 bool ChakraRTInterface::m_testHooksSetup = false;
 bool ChakraRTInterface::m_testHooksInitialized = false;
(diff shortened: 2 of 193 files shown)


digitalinfinity:
@LouisLaf @curtisman please review 7453224
@jianchun please review the rest

obastemur:
@digitalinfinity can we make `ch` using ChakraCore as a static library instead? Having ChakraCore as a shared library for ch will be a headache. i.e. osx problem with shared library from same folder. OR, ch is referencing wrong shared binary by mistake etc..

pal/src/cruntime/misc.cpp
```diff
@@ -287,6 +287,45 @@ PAL_qsort(void *base, size_t nmemb, size_t size,
...
+// Lambda doesn't work here as qsort_r needs a function pointer
+//
+typedef int (__cdecl * qsort_s_comparer)(void*, const void *, const void *);
+struct CompareWrapperContext
+{
+    qsort_s_comparer real_comparer;
+    void* real_context;
+};
+
+static int qsort_compare_wrapper(const void* e1, const void* e2, void* context)
```
obastemur: In order not to break OSX/BSD compat, please consider the implementation below;

```c++
#if !defined(__APPLE__) && !defined(__FreeBSD__)
static int qsort_compare_wrapper(const void* e1, const void* e2, void* context)
{
    CompareWrapperContext* wrapper = (CompareWrapperContext*) context;
    return wrapper->real_comparer(wrapper->real_context, e1, e2);
}
#else
static int qsort_compare_wrapper(void* context, const void* e1, const void* e2)
{
    CompareWrapperContext* wrapper = (CompareWrapperContext*) context;
    return wrapper->real_comparer(wrapper->real_context, e1, e2);
}
#endif

PALIMPORT 
void __cdecl 
PAL_qsort_s(void *base, size_t nmemb, size_t size, 
    int (__cdecl * compar)(const void*, const void *, void *), void* context)
{
    ASSERT(compar != nullptr);
    CompareWrapperContext qsort_s_context = { reinterpret_cast<qsort_s_comparer>(compar), context };

    qsort_r(
        base,
        nmemb,
        size,
#if !defined(__APPLE__) && !defined(__FreeBSD__)
        qsort_compare_wrapper,
        &qsort_s_context);
#else
        &qsort_s_context,
        qsort_compare_wrapper);
#endif
}
```

bin/ch/ChakraRtInterface.cpp
```diff
@@ -4,7 +4,12 @@
 //-------------------------------------------------------------------------------------------------------
 #include "stdafx.h"
 
-LPCWSTR chakraDllName = _u("chakracore.dll");
+#ifdef _WIN32
+LPCSTR chakraDllName = "chakracore.dll";
+#else
+#include <dlfcn.h>
+LPCSTR chakraDllName = "libChakraCoreLib.so";
```
ThomsonTan: Naming it libChakraCore might be more consistent to Linux?

lib/Common/DataStructures/DictionaryEntry.h
```diff
@@ -87,15 +87,15 @@ namespace JsUtil
     template<class TKey, class TValue>
     struct ValueToKey
     {
-        static TKey ToKey(const TValue &value) { return static_cast<TKey>(value); }
+        static inline TKey ToKey(const TValue &value) { return static_cast<TKey>(value); }
```
jianchun: Curious why add `inline` here and below?

lib/Parser/CharTrie.cpp
```diff
@@ -6,10 +6,12 @@
 
 namespace UnifiedRegex
 {
+    const int CharTrie::initCapacity;
```
jianchun: Header file has 'static const int initCapacity = 4`. It works by init in header file and not here?

lib/Common/CommonPal.h
```diff
@@ -31,17 +31,26 @@ typedef wchar_t char16;
...
 #include "inc/pal.h"
 #include "inc/rt/palrt.h"
 #include "inc/rt/no_sal2.h"
 #include "inc/rt/oaidl.h"
 
+// In general, we don't need to force-inline any of Chakra's functions
+// on Linux- we'd rather let the compiler figure this out for us
+// If, in the future, perf investigations reveal that force inlining
+// would help, we can address on a case-by-case basis.
+#define __forceinline inline
```
jianchun: What's the reason to define `__forceinline` this way and on case-by-case #ifdef to `__attribute__((always_inline))` on Linux? It makes all those case-by-cases code uglier. Why not let it mean *force inline*?

lib/Common/Common.h
```diff
@@ -29,7 +29,11 @@ typedef _Return_type_success_(return >= 0) LONG NTSTATUS;
 #include <malloc.h>
 #endif
 
+// _AddressOfReturnAddress is declared as (forced) inline in CommonPal
+// Don't need this declaration outside of VC
+#ifdef _MSC_VER
 extern "C" void * _AddressOfReturnAddress(void);
```
jianchun: Consider move this whole thing to CommonPal.h

lib/Common/CommonPal.h
```diff
@@ -78,13 +87,23 @@ inline void DebugBreak()
 #define _bittestandset BitTestAndSet
 #define _interlockedbittestandset InterlockedBitTestAndSet
 
+#if defined(__GNUC__) || defined(__clang__)
+inline void * __attribute__((__always_inline__)) _AddressOfReturnAddress()
+{
```
jianchun: Consider move this down to be next or merge with `_ReturnAddress`.

lib/Common/CommonPal.h
```diff
@@ -433,11 +456,92 @@ extern "C" PVOID _ReturnAddress(VOID);
...
+        cchMax = cchDest - 1;
+
+        iRet = _vsnwprintf(pszDest, cchMax, pszFormat, argList);
+        // ASSERT((iRet < 0) || (((size_t)iRet) <= cchMax));
+
+        if ((iRet < 0) || (((size_t)iRet) > cchMax))
+        {
+            // need to null terminate the string
+            pszDest += cchMax;
+            *pszDest = L'\0';
```
jianchun: Should these be _u('\0')?

pal/src/misc/bstr.cpp
```diff
@@ -139,6 +139,28 @@ STDAPI_(void) SysFreeString(const OLECHAR* psz)
...
+ *
+ ***********************************************************************/
+STDAPI_(UINT) SysStringLen(const OLECHAR* psz)
+{
+    if (psz == NULL)
+    {
+        return 0;
+    }
+
+    return PAL_wcslen(psz);
```
jianchun: Why not read the DWORD before `psz`?

lib/Runtime/Base/StackProber.cpp
```diff
@@ -63,7 +63,7 @@ StackProber::Initialize()
 #endif
 #else // !_WIN32
     void *lowLimit, *highLimit;
-    ::GetCurrentThreadStackBounds((char**)&lowLimit, (char**)&highLimit);
+    ::GetCurrentThreadStackBounds((char**)&highLimit, (char**)&lowLimit);
```
jianchun: Would it be better to reimplement this function (or even rename it) similar to Windows `GetCurrentThreadStackLimits`, so that all calls are similar?

lib/Runtime/Language/Arguments.h
```diff
@@ -4,6 +4,43 @@
...
+//
+// xplat-todo: Investigate whether we can do something
+// cleverer here to avoid this stack copying. Perhaps a
+// combination of making up our own calling convention between
+// amd64_JavascriptCallFunction and its callees, and having
+// argument reader manipulate the stack pointer directly
+// instead of using va_list/va_args. However, the following
+// solution is probably the cleanest/most portable.
+#define VA_LIST_TO_VARARRAY(vl, va, callInfo)                              \
+    Js::Var* va = (Var*) alloca(callInfo.Count * sizeof(Var*));            \
```
jianchun: Do we need to worry about stack overflow here? Maybe corresponding PROBE_STACK needs to add needed stack space for this.

bin/ch/ChakraRtInterface.cpp
digitalinfinity: Done- I had originally named the target as ChakraCoreLib to avoid conflicting with the project name, but switched them to be the same and it seems to work :)

lib/Common/DataStructures/DictionaryEntry.h
digitalinfinity: Agreed- unnecessary- removed

lib/Common/CommonPal.h
digitalinfinity: Changed. I keep going back and forth on this, I wanted it to be inline initially so that we switch it to always_inline only after we've understood the perf. Switched to your model for now but might change my mind again in the future 😄

lib/Common/Common.h
digitalinfinity: Done

lib/Common/CommonPal.h
digitalinfinity: Done

lib/Common/CommonPal.h
digitalinfinity: Done

pal/src/misc/bstr.cpp
digitalinfinity: Good catch- fixed.

lib/Runtime/Base/StackProber.cpp
digitalinfinity: Renamed to GetCurrentThreadStackLimits and implemented those semantics instead.

lib/Runtime/Language/Arguments.h
digitalinfinity: Probably. But we have to worry about that in more places than just here, I've been ignoring the stack probing completely 😄. We should open a separate issue for that and deal with it after this change.

digitalinfinity:
@obastemur ch is a thin driver for testing the ChakraCore dynamic library through static entry points. The problems that you describe are applicable on Windows as well, and in practice, at least on Windows, it hasn't mattered. We should continue to use ChakraCore as a dynamic library with ch to simulate how ChakraCore would be used by its consumers, but if you feel strongly about switching ch to be statically linked with ChakraCore, please open a separate issue and we can have that discussion there.

pal/src/thread/thread.cpp
```diff
@@ -2786,12 +2786,12 @@ int CorUnix::CThreadMachExceptionHandlers::GetIndexOfHandler(exception_mask_t bm
...
 
-int GetCurrentThreadStackBounds(char** stackBase, char** stackEnd)
+void GetCurrentThreadStackLimits(ULONG_PTR* lowLimit, ULONG_PTR* highLimit)
 {
 #ifdef _TARGET_MAC64
     // This is a Mac specific method
-    *stackBase = (char*)pthread_get_stackaddr_np(pthread_self());
-    *stackEnd = (char*)(stackBase - pthread_get_stacksize_np(pthread_self()));
+    *highLimit = (ULONG_PTR)pthread_get_stackaddr_np(pthread_self());
+    *lowLimit = (ULONG_PTR)(stackBase - pthread_get_stacksize_np(pthread_self()));
```
obastemur: why `lowLimit` and `highLimit`? The order of the arguments are also changed? i.e. since there is no stackBase local parameter now, what stackBase at 2794 needs to be?

lib/Runtime/Library/amd64/JavascriptFunctionA.S
```diff
@@ -0,0 +1,144 @@
+//-------------------------------------------------------------------------------------------------------
+// Copyright (C) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE.txt file in the project root for full license information.
+//-------------------------------------------------------------------------------------------------------
+.intel_syntax noprefix
+
+.extern _ZN2Js18JavascriptFunction13DeferredParseEPPNS_14ScriptFunctionE
```
obastemur: Where this method is defined ?

lib/Runtime/Library/amd64/JavascriptFunctionA.S
```diff
@@ -0,0 +1,144 @@
...
+
+.global amd64_CallFunction
```
obastemur:
```c++
#ifdef __APPLE__
#define FNAME(x) _##x
#else
#define FNAME(x) x
#endif
.global FNAME(amd64_CallFunction)
```

lib/Runtime/Library/amd64/JavascriptFunctionA.S
```diff
@@ -0,0 +1,144 @@
...
+.global	_ZN2Js18JavascriptFunction20DeferredParsingThunkEPNS_16RecyclableObjectENS_8CallInfoEz
```
obastemur: I also couldn't find a reference for this one? (Apart from the one defined under this source file)

lib/Runtime/Library/amd64/JavascriptFunctionA.S
```diff
@@ -0,0 +1,144 @@
...
+
+.type amd64_CallFunction, @function
```
obastemur: can we define `.type` (if we need to do so at all) only for linux ?

```
#ifndef __APPLE__
```

lib/Runtime/Library/amd64/JavascriptFunctionA.S
```diff
@@ -0,0 +1,144 @@
...
+.type _ZN2Js18JavascriptFunction20DeferredParsingThunkEPNS_16RecyclableObjectENS_8CallInfoEz, @function
+        
+.text
+amd64_CallFunction:
```
obastemur: `FNAME(amd64_CallFunction):`

lib/Runtime/Library/amd64/JavascriptFunctionA.S
```diff
@@ -0,0 +1,144 @@
...
+        pop r12
+        pop r13
+        pop r14
+        pop r15
+        pop rbx
+
+        pop rbp
+        ret
+
+// extrn ?DeferredParse@JavascriptFunction@Js@@SAP6APEAXPEAVRecyclableObject@2@UCallInfo@2@ZZPEAPEAVScriptFunction@2@@Z : PROC
```
obastemur: I couldn't find any ref to source below this line. Safe to remove?

bin/ChakraCore/CMakeLists.txt
```diff
@@ -1,9 +1,74 @@
...
+# libraries are considered to be used by Chakra
+#
+# External libraries we link with are the following:
+#  pthread: For threading
+#  stdc++/gcc_s: C++ runtime code
+#  dl: For shared library loading related functions
+#  icuuc: For the ICU (xplat-todo: Make this optional)
+#
+target_link_libraries(ChakraCore
+  -Wl,--no-undefined
```
obastemur: can we use `-Wl,-undefined,error` instead?

bin/ChakraCore/CMakeLists.txt
```diff
@@ -1,9 +1,74 @@
...
+  Chakra.Common.Core
+  Chakra.Common.DataStructures
+  Chakra.Common.Exceptions
+  Chakra.Common.Codex
+  -Wl,--end-group
+  Chakra.Pal
+  pthread
+  stdc++
+  dl
+  gcc_s
```
obastemur: why do we need this explicitly defined?

pal/src/thread/thread.cpp
digitalinfinity: I had switched to using the Win32 GetCurrentThreadStackLimits semantics as per @jianchun's feedback. That returns the smaller value as lowLimit and higher value as highLimit (generally used for figuring out if an address belongs to the current threads stack). What you had as stackEnd is less than stackBase so that becomes the lowLimit and stackBase becomes highLimit.

bin/ChakraCore/CMakeLists.txt
digitalinfinity: Hmm, don't remember- removed.

bin/ChakraCore/CMakeLists.txt
digitalinfinity: Sure, switched to that- curious, why?

lib/Runtime/Library/amd64/JavascriptFunctionA.S
digitalinfinity: Switched to using C_NAME

digitalinfinity:
Thanks @obastemur. Added new commits to address your comments. Let me know if its good to merge (cc @Jianchun)

pal/src/thread/thread.cpp
obastemur: could you please change `stackBase`(2794) to `highLimit` then?

obastemur:
@digitalinfinity :+1:

jianchun:
👍

pal/src/thread/thread.cpp
digitalinfinity: Hah! sorry, missed that- I understand your comment now, thanks!

lib/Common/Exceptions/ReportError.cpp
```diff
@@ -5,7 +5,7 @@
 #include "CommonExceptionsPch.h"
 
 #ifdef _MSC_VER
-__inline
+inline
```
LouisLaf: Just remove the inline altogether here.  VC does not inline function with try/except anyway.

lib/Common/Memory/ArenaAllocator.cpp
```diff
@@ -8,7 +8,13 @@
 
 const uint Memory::StandAloneFreeListPolicy::MaxEntriesGrowth;
 
+#ifdef _MSC_VER
 template __forceinline BVSparseNode * BVSparse<JitArenaAllocator>::NodeFromIndex(BVIndex i, BVSparseNode *** prevNextFieldOut, bool create);
+#else
+// Clang/GCC support
+// We need this function to be inlined for perf
+template __attribute__((always_inline)) BVSparseNode * BVSparse<JitArenaAllocator>::NodeFromIndex(BVIndex i, BVSparseNode *** prevNextFieldOut, bool create);
+#endif
```
LouisLaf: This can be removed if you now define __forceinline as always_inline?

PR #875, opened 2016-04-27
Fixing an issue with class definition in the param scope

aneeshdk:
When visiting the pnodescopes I had a check to avoid visiting the body scope as part of the param scope. The condition I was using for the check does not seem to cover all scenarios. This changelist fixes that issue.

### DIFF
--- a/lib/Runtime/ByteCode/ByteCodeGenerator.h
+++ b/lib/Runtime/ByteCode/ByteCodeGenerator.h
@@ -284,7 +284,7 @@ public:
 
     void DefineLabels(FuncInfo *funcInfo);
     void EmitProgram(ParseNode *pnodeProg);
-    void EmitScopeList(ParseNode *pnode, bool breakOnNonFunc = false);
+    void EmitScopeList(ParseNode *pnode, ParseNode *breakOnBodyScopeNode = nullptr);
     void EmitDefaultArgs(FuncInfo *funcInfo, ParseNode *pnode);
     void EmitOneFunction(ParseNode *pnode);
     void EmitGlobalFncDeclInit(Js::RegSlot rhsLocation, Js::PropertyId propertyId, FuncInfo * funcInfo);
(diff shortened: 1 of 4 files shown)


aneeshdk:

lib/Runtime/ByteCode/ByteCodeGenerator.h
```diff
@@ -284,7 +284,7 @@ class ByteCodeGenerator
 
     void DefineLabels(FuncInfo *funcInfo);
     void EmitProgram(ParseNode *pnodeProg);
-    void EmitScopeList(ParseNode *pnode, bool breakOnNonFunc = false);
+    void EmitScopeList(ParseNode *pnode, ParseNode *breakOnBodyScope = nullptr);
```
akroshg: nit: can you rename this as breakOnBodyScopeNode ?

test/es6/default-splitscope.js
```diff
@@ -315,7 +315,36 @@ var tests = [
                 return y; 
             } 
         } 
-        assert.areEqual(10, (new c4()).f({})(), "The method defined as the default destructured value of the parameter should capture the formal from the param scope"); 
+        assert.areEqual(10, (new c4()).f({})(), "The method defined as the default destructured value of the parameter should capture the formal from the param scope");
+        
+        function f3(a = 10, d, b = (function () { return a; }, class { method1() { return a; } }), c) {
```
akroshg: could you have test case where class has a name and also class has a 'extends' ?

akroshg:
Rest looks good to me

lib/Runtime/ByteCode/ByteCodeGenerator.h
aneeshdk: Fixed.

test/es6/default-splitscope.js
aneeshdk: Done.

PR #876, opened 2016-04-27
Remove duplicated code in FunctionBody::SetNativeEntryPoint

rajatd:
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/876)

### DIFF
--- a/lib/Runtime/Base/FunctionBody.cpp
+++ b/lib/Runtime/Base/FunctionBody.cpp
@@ -3180,24 +3180,6 @@ namespace Js
         }
         TraceExecutionMode();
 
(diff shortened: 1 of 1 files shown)


lib/Runtime/Base/FunctionBody.cpp
```diff
@@ -3180,24 +3180,6 @@ namespace Js
         }
         TraceExecutionMode();
```
ThomsonTan: As mentioned by Michael, also removes this line introduced by incorrect merge?

rajatd: There's no harm in keeping this call- just does some tracing.

ThomsonTan:
:+1:

PR #878, opened 2016-04-27
Object.prototype should be an immutable prototype exotic object.

### LINKED ISSUE #261: Object.prototype should be an immutable prototype exotic object
You should not be able to change its prototype with `Object.setPrototypeOf`, `__proto__`, or any other means. See also: https://github.com/tc39/ecma262/pull/308.

dilijev:
This means that any attempts to set the [[Prototype]] slot of the Object Prototype Object (the object which is initially pointed to by `Object.prototype`) notably by using `Object.prototype.__proto__` or `Object.setPrototypeOf(Object.prototype, otherObject)` should fail with a `TypeError` as per spec.

This is the only exotic behavior of this object.

I chose to implement this by adding the virtual method `IsProtoImmutable` to `RecyclableObject` with the default return value `false`, and to override this method in `ObjectPrototypeObject` to return `true`. This allows for a future implementation of `Object.setImmutablePrototype` which would allow other objects to exhibit this behavior. A flag could be added which can be set, and this method would return the value of that flag. In that case, the implementation in `ObjectPrototypeObject` should remain hardcoded to false to avoid accidentally allowing the value if `IsProtoImmutable` to change.

See [ES2016 19.1.3](https://tc39.github.io/ecma262/#sec-properties-of-the-object-prototype-object):
The Object prototype object is the intrinsic object %ObjectPrototype%.
The Object prototype object is an immutable prototype exotic object.
See [ES2016 9.4.7](https://tc39.github.io/ecma262/#sec-immutable-prototype-exotic-objects):
An immutable prototype exotic object is an exotic object that has an immutable [[Prototype]] internal slot.

Fixes #261

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/878)

### DIFF
--- a/lib/Parser/rterrors.h
+++ b/lib/Parser/rterrors.h
@@ -275,6 +275,8 @@ RT_ERROR_MSG(JSERR_DeletePropertyWithSuper, 5146, "Unable to delete property '%s
 RT_ERROR_MSG(JSERR_DetachedTypedArray, 5147, "%s: The ArrayBuffer is detached.", "The ArrayBuffer is detached.", kjstTypeError, 0)
 RT_ERROR_MSG(JSERR_AsmJsCompileError, 5148, "%s: Compiling asm.js failed.", "Compiling asm.js failed.", kjstError, 0)
 
+RT_ERROR_MSG(JSERR_ImmutablePrototypeSlot, 5149, "%s: Can't set the prototype of this object.", "Can't set the prototype of this object.", kjstTypeError, 0)
+
 /* Error messages for misbehaved Async Operations for use in Promise.js */
 RT_ERROR_MSG(ASYNCERR_NoErrorInErrorState, 5200, "", "Status is 'error', but getResults did not return an error", kjstError, 0)
 RT_ERROR_MSG(ASYNCERR_InvalidStatusArg, 5201, "", "Missing or invalid status parameter passed to completed handler", kjstError, 0)
--- a/lib/Runtime/Library/JavascriptObject.cpp
+++ b/lib/Runtime/Library/JavascriptObject.cpp
@@ -176,6 +176,16 @@ namespace Js
             return FALSE;
         }
 
+        if (object->IsProtoImmutable())
+        {
+            // ES2016 19.1.3:
+            // The Object prototype object is the intrinsic object %ObjectPrototype%.
+            // The Object prototype object is an immutable prototype exotic object.
+            // ES2016 9.4.7:
+            // An immutable prototype exotic object is an exotic object that has an immutable [[Prototype]] internal slot.
+            JavascriptError::ThrowTypeError(scriptContext, JSERR_ImmutablePrototypeSlot);
+        }
+
         // 6.   If V is not null, then
         //  a.  Let p be V.
         //  b.  Repeat, while p is not null
--- a/test/es7/immutable-prototype.js
+++ b/test/es7/immutable-prototype.js
@@ -1,0 +1,45 @@
+//-------------------------------------------------------------------------------------------------------
+// Copyright (C) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE.txt file in the project root for full license information.
+//-------------------------------------------------------------------------------------------------------
+
+// ES7 Object Prototype object has an immutable [[Prototype]] internal slot
+// See: 19.1.3 Properties of the Object Prototype Object
+// See: 9.4.7 Immutable Prototype Exotic Objects
+
+WScript.LoadScriptFile("..\\UnitTestFramework\\UnitTestFramework.js");
+
+var tests = [
+    {
+        name: "Not okay to set Object.prototype.[[Prototype]] using __proto__",
+        body: function () {
+            var objectPrototypeObject = Object.getPrototypeOf(Object.prototype)
+            var b = Object.create(null)
+
+            assert.throws(function () { Object.prototype.__proto__ = b },
+                TypeError,
+                "It should not be okay to set Object.prototype.[[Prototype]] using __proto__",
+                "Can't set the prototype of this object.")
+
+            assert.areEqual(objectPrototypeObject, Object.prototype.__proto__, "Object.prototype.__proto__ is unchanged")
+            assert.areEqual(objectPrototypeObject, Object.getPrototypeOf(Object.prototype), "Object.getPrototypeOf(Object.prototype) is unchanged")
+        }
+    },
+    {
+        name: "Not okay to set Object.prototype.[[Prototype]] using Object.setPrototypeOf",
+        body: function () {
+            var objectPrototypeObject = Object.getPrototypeOf(Object.prototype)
+            var b = Object.create(null)
+
+            assert.throws(function () { Object.setPrototypeOf(Object.prototype, b) },
+                TypeError,
+                "It should not be okay to set Object.prototype.[[Prototype]] using Object.setPrototypeOf",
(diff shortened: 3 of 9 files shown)


dilijev:
@bterlson @ljharb @abchatra @curtisman 
Please review?

lib/Parser/rterrors.h
```diff
@@ -275,6 +275,8 @@ RT_ERROR_MSG(JSERR_DeletePropertyWithSuper, 5146, "Unable to delete property '%s
 RT_ERROR_MSG(JSERR_DetachedTypedArray, 5147, "%s: The ArrayBuffer is detached.", "The ArrayBuffer is detached.", kjstTypeError, 0)
 RT_ERROR_MSG(JSERR_AsmJsCompileError, 5148, "%s: Compiling asm.js failed.", "Compiling asm.js failed.", kjstError, 0)
 
+RT_ERROR_MSG(JSERR_ImmutablePrototypeSlot, 5149, "%s: Can't set the prototype slot of this object.", "Can't set the prototype slot of this object.", kjstTypeError, 0)
```
goyakin: I wouldn't mention _slot_ here since users don't need to know the language internals. I think it would be better to say `Can't set the prototype of this object` instead.

lib/Runtime/Library/JavascriptObject.cpp
```diff
@@ -176,6 +176,16 @@ namespace Js
             return FALSE;
         }
 
+        if (object->IsProtoImmutable())
+        {
+            // 19.1.3:
```
goyakin: Section numbers change between spec versions, so you should specify "ES2016" here.

test/es7/immutable-prototype.js
```diff
@@ -0,0 +1,77 @@
...
+        name: "Okay to set arbitrary object's [[Prototype]] using Object.setPrototypeOf",
+        body: function () {
+            var a = {}
+            var b = {'b': 2}
+
+            assert.areEqual(Object.prototype, a.__proto__, "a.__proto__ === Object.prototype")
+            assert.areEqual(Object.prototype, Object.getPrototypeOf(a), "Object.getPrototypeOf(a) === Object.prototype")
+
+            assert.doesNotThrow(function () { eval("Object.setPrototypeOf(a, b)") },
+                "It should be okay to set the [[Prototype]] of an arbitrary object using __proto__.")
```
akroshg: using *setPrototypeOf * ?

test/es7/immutable-prototype.js
```diff
@@ -0,0 +1,77 @@
...
+
+// ES7 Object Prototype object has an immutable [[Prototype]] internal slot
+// See: 19.1.3 Properties of the Object Prototype Object
+// See: 9.4.7 Immutable Prototype Exotic Objects
+
+WScript.LoadScriptFile("..\\UnitTestFramework\\UnitTestFramework.js");
+
+var tests = [
+    {
+        name: "Okay to set arbitrary object's [[Prototype]] using __proto__",
```
goyakin: Looks like we already test this in `test/es6/proto_basic.js`. Look for the test with name `Change object.__proto__ to an object`.

test/es7/immutable-prototype.js
```diff
@@ -0,0 +1,77 @@
...
+
+            assert.doesNotThrow(function () { eval("a.__proto__ = b") },
+                "It should be okay to set the [[Prototype]] of an arbitrary object using __proto__.")
+
+            assert.areEqual(b, a.__proto__, "a.__proto__ === b")
+            assert.areEqual(b, Object.getPrototypeOf(a), "Object.getPrototypeOf(a) === b")
+        }
+    },
+    {
+        name: "Okay to set arbitrary object's [[Prototype]] using Object.setPrototypeOf",
```
goyakin: Already tested in `test/es6/ES6Proto.js`. Look for the test with name `Simple validation of Object.setPrototypeOf`.

test/es7/immutable-prototype.js
```diff
@@ -0,0 +1,77 @@
...
+            assert.areEqual(b, Object.getPrototypeOf(a), "Object.getPrototypeOf(a) === b")
+        }
+    },
+    {
+        name: "Not okay to set Object.prototype.[[Prototype]] using __proto__",
+        body: function () {
+            var objectPrototypeObject = Object.getPrototypeOf(Object.prototype)
+            var b = Object.create(null)
+
+            assert.throws(function () { eval("Object.prototype.__proto__ = b") },
```
goyakin: Is there a particular reason why `eval` used here instead of doing a plain assignment to `__proto__`?

test/es7/immutable-prototype.js
```diff
@@ -0,0 +1,77 @@
...
+            assert.areEqual(objectPrototypeObject, Object.getPrototypeOf(Object.prototype), "Object.getPrototypeOf(Object.prototype) is unchanged")
+        }
+    },
+    {
+        name: "Not okay to set Object.prototype.[[Prototype]] using Object.setPrototypeOf",
+        body: function () {
+            var objectPrototypeObject = Object.getPrototypeOf(Object.prototype)
+            var b = Object.create(null)
+
+            assert.throws(function () { eval("Object.setPrototypeOf(Object.prototype, b)") },
```
goyakin: Can we directly call `setPrototypeOf` without `eval`?

test/es7/immutable-prototype.js
dilijev: Ah, I figured there would be a test already. Since I wanted to make sure the new behavior didn't break the old behavior I added these tests here, but I see that they are redundant so I'll remove them.

test/es7/immutable-prototype.js
dilijev: I had only seen assert.throws that used eval. I thought it might have been required for some reason. I don't need it, I'll get rid of it.

test/es7/immutable-prototype.js
dilijev: Yes, I'll make the change.

test/es7/immutable-prototype.js
dilijev: Removing.

test/es7/immutable-prototype.js
dilijev: Thanks!

lib/Parser/rterrors.h
dilijev: Makes sense. I'll make this change.

goyakin:
Other than minor inline comments, everything else LGTM.

> I chose to implement this by adding the virtual method IsProtoImmutable to RecyclableObject with the default return value false, and to override this method in ObjectPrototypeObject to return true.

This would be useful when we implement `[[SetPrototypeOf]]` for [module namespace objects](https://tc39.github.io/ecma262/#sec-module-namespace-exotic-objects-setprototypeof-v) :+1:

lib/Runtime/Library/JavascriptObject.cpp
dilijev: Fixing this as well.

dilijev:
There is a failing test because of defunct behavior (directly contradicts that `Object.prototype.__proto__` should be immutable).

dilijev:
Removed defunct test and updated spec references to specifically refer to ES2016. Thanks @goyakin.

/cc @jianchun @bterlson on test removal.

`Cyclic __proto__ ` was tested in another test in proto_basic.js which didn't involve Object.prototype. All of the other assignments don't make sense anymore since the [[Prototype]] slot is immutable so removing that unit test made sense.

PR #882, opened 2016-04-28
Fixing the static warning by adding sufficient validation for the buffer.

akroshg:


akroshg:
@abchatra @LouisLaf - please have a look.

abchatra:
:+1:

PR #885, opened 2016-04-28
Fix #884: Proxy: ToPropertyDescriptor assert; when assert not hit, should throw TypeError.

### LINKED ISSUE #884: Proxy: ToPropertyDescriptor assert; when Proxy has() always lies and returns true, should result in TypeError
Repro test case:

```js
var h = { has : function() { WScript.Echo("has"); return true; }};
var p = new Proxy({}, h);
var o = {};
Object.defineProperty(o, "x", p);
```
 
Output:
```
has
ASSERTION 3060: (D:\src\Chakra\core\lib\Runtime\Language\JavascriptOperators.cpp, line 8661) Proxy : HasProperty and GetProperty's result don't match for 'enumerable'.  Failure: (0)
FATAL ERROR: jshost.exe failed due to exception code c0000420
```
 
Expected output:
```
has
has
has
has
has
has
TypeError: Property cannot have both accessors and a value
```

Each `has` is for the Proxy properties: configurable, enumerable, value, writable, get, and set that are checked when there is a proxy around the PropertyDescriptor being used with Object.defineProperty.

See: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy
See: https://tc39.github.io/ecma262/2016/#sec-object.defineproperty
See: https://tc39.github.io/ecma262/2016/#sec-topropertydescriptor

The reason there should be a `TypeError` is because according to spec semantics, when the proxy says `has()` is true for a property, we should try to look it up, and if we discover that it is not present, then we should set it to the default value (falsy: undefined for functions and false for non-functions). Since certain combinations of these PropertyDescriptor properties are disallowed, we see a `TypeError` in this case.

Converted from [7019635](https://microsoft.visualstudio.com/web/wi.aspx?pcguid=cb55739e-4afe-46a3-970f-1b49d8ee7564&id=7019635)

/cc @bterlson to double check details.

dilijev:
Fixes #884 (see issue for details).

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/885)

### DIFF
--- a/lib/Parser/rterrors.h
+++ b/lib/Parser/rterrors.h
@@ -164,7 +164,7 @@ RT_ERROR_MSG(JSERR_Property_CannotDelete_NullOrUndefined, 5049, "Unable to delet
 RT_ERROR_MSG(JSERR_Property_VarDate, 5050, "Unable to access property '%s': type 'VarDate' does not support user-defined properties", "Object expected", kjstTypeError, JSERR_NeedObject)
 RT_ERROR_MSG(JSERR_Property_NeedFunction, 5051, "The value of the property '%s' is not a Function object", "Function expected", kjstTypeError, JSERR_NeedFunction)
 RT_ERROR_MSG(JSERR_Property_NeedFunction_NullOrUndefined, 5052, "The value of the property '%s' is null or undefined, not a Function object", "Function expected", kjstTypeError, JSERR_NeedObject)
-RT_ERROR_MSG(JSERR_Property_CannotHaveAccessorsAndValue, 5053, "", "Property cannot have both accessors and a value", kjstTypeError, VBSERR_ActionNotSupported)
+RT_ERROR_MSG(JSERR_Property_CannotHaveAccessorsAndValue, 5053, "", "Invalid property descriptor: cannot both specify accessors and a 'value' attribute", kjstTypeError, VBSERR_ActionNotSupported)
 
 RT_ERROR_MSG(JSERR_This_NullOrUndefined, 5054, "%s: 'this' is null or undefined", "'this' is null or undefined", kjstTypeError, JSERR_NeedObject) // {Locked="\'this\'"}
 RT_ERROR_MSG(JSERR_This_NeedObject, 5055, "%s: 'this' is not an Object", "Object expected", kjstTypeError, JSERR_NeedObject) // {Locked="\'this\'"}
--- a/test/es6/proxy-issue884.js
+++ b/test/es6/proxy-issue884.js
@@ -1,0 +1,24 @@
+//-------------------------------------------------------------------------------------------------------
+// Copyright (C) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE.txt file in the project root for full license information.
+//-------------------------------------------------------------------------------------------------------
+
+var proxyHandler = {
+    has() { WScript.Echo("has"); return true; }
+};
+var p = new Proxy({}, proxyHandler);
+var obj = {};
+
+try {
+    Object.defineProperty(obj, "x", p);
+} catch (e) {
+    if (e instanceof TypeError) {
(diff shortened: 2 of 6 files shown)


dilijev:
@bterlson @digitalinfinity Review please? :)

bterlson:
:+1:

lib/Parser/rterrors.h
```diff
@@ -164,7 +164,7 @@ RT_ERROR_MSG(JSERR_Property_CannotDelete_NullOrUndefined, 5049, "Unable to delet
 RT_ERROR_MSG(JSERR_Property_VarDate, 5050, "Unable to access property '%s': type 'VarDate' does not support user-defined properties", "Object expected", kjstTypeError, JSERR_NeedObject)
 RT_ERROR_MSG(JSERR_Property_NeedFunction, 5051, "The value of the property '%s' is not a Function object", "Function expected", kjstTypeError, JSERR_NeedFunction)
 RT_ERROR_MSG(JSERR_Property_NeedFunction_NullOrUndefined, 5052, "The value of the property '%s' is null or undefined, not a Function object", "Function expected", kjstTypeError, JSERR_NeedObject)
-RT_ERROR_MSG(JSERR_Property_CannotHaveAccessorsAndValue, 5053, "", "Property cannot have both accessors and a value", kjstTypeError, VBSERR_ActionNotSupported)
+RT_ERROR_MSG(JSERR_Property_CannotHaveAccessorsAndValue, 5053, "", "Invalid property descriptor: cannot both specify accessors and a 'value' attribute", kjstTypeError, VBSERR_ActionNotSupported)
```
ianwjhalliday: cannot specify both

test/es6/proxy-issue884.js
```diff
@@ -0,0 +1,24 @@
+//-------------------------------------------------------------------------------------------------------
+// Copyright (C) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE.txt file in the project root for full license information.
+//-------------------------------------------------------------------------------------------------------
+
+var proxyHandler = {
+    has() { WScript.Echo("has"); return true; }
```
ianwjhalliday: consider printing out the queried property name here too in case that makes debugging a future bug easier

lib/Parser/rterrors.h
dilijev: I was worried about potential confusion here. The term "both" in "specify both" could refer to the two accessors: "get" and "set". The grammar as written is technically correct and avoids that potential confusion.

Also this matches the grammar of v8's error message: "VM79:4 Uncaught TypeError: Invalid property descriptor. Cannot both specify accessors and a value or writable attribute, [object Object](…)"

ianwjhalliday:
:+1:

lib/Parser/rterrors.h
ianwjhalliday: Huh, its like v8's implementer shared your same concern.

I don't think it is ambiguous because of the `and` but I'm only one sample point.  It's fine.

test/es6/proxy-issue884.js
dilijev: I'll try that.

lib/Parser/rterrors.h
dilijev: I'll try to reword this to avoid both grammatical and semantic ambiguity. :)

test/es6/proxy-issue884.js
dilijev: Output will then be:

```
has enumerable
has configurable
has value
has writable
has get
has set
PASS
```

I wonder if this order will be stable?

dilijev: Yes, this order will be stable because it comes from the order of checks/accesses in the C++ code, not the canonical ordering of those named properties.

I'll make the change to this test.

dilijev:
@dotnet-bot test this please

PR #886, opened 2016-04-28
Fix 7019631: Check whether a Proxy object is anywhere in the prototype chain before choosing to do the optimized ToPropertyDescriptor for non-Proxy objects.

dilijev:
See: https://tc39.github.io/ecma262/2016/#sec-topropertydescriptor

---

This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/886)

### DIFF
--- a/lib/Runtime/Language/JavascriptOperators.cpp
+++ b/lib/Runtime/Language/JavascriptOperators.cpp
@@ -8670,6 +8670,7 @@ CommonNumber:
 
         return true;
     }
+
     BOOL JavascriptOperators::ToPropertyDescriptorForProxyObjects(Var propertySpec, PropertyDescriptor* descriptor, ScriptContext* scriptContext)
     {
         if (!JavascriptOperators::IsObject(propertySpec))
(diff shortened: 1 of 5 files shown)


dilijev:
@Yongqu @curtisman please review?

@Yongqu do you think this walk of the prototype chain as written is a good way to answer this question or should we use another approach? Any other thoughts?

lib/Runtime/Language/JavascriptOperators.cpp
```diff
@@ -10033,6 +10032,24 @@ namespace Js
...
+        {
+            return false;
+        }
+        else if (prototype->GetTypeId() == TypeIds_Proxy)
+        {
+            return true;
+        }
+        else
+        {
+            return CheckIfPrototypeChainContainsProxyObject(prototype->GetPrototype());
```
dilijev: Even though prototype chains are typically fairly short, I think it is probably a good idea to convert this function to an interative version to avoid taking up O(n) stack frames where n is the length of the prototype chain.

ianwjhalliday: Agreed.  We use iterative loops for prototype chain walking almost everywhere.  I think we have one or two recursive walks, but I would favor a loop.

---
In reply to: [61649427](https://github.com/Microsoft/ChakraCore/pull/886#discussion_r61649427) [](ancestors = 61649427)

ianwjhalliday:
:shipit:

Yongqu:
:+1:

PR #892, opened 2016-04-29
Fixing scope issues with function expressions with split scope

aneeshdk:
This change fixes multiple issues with using function expression symbols in split scope.

One issue was if the function expression is not captured and has not eval
then we were adding the function expression symbol to body scope. In split
scope case I changed it to add to the param scope. But we don't have to
duplicate the function expression name symbol in the body.

Function expression should be pop out only after taking out the param scope. Another issue was while creating the new frame display during BeginBodyScope we have to skip the function expression.

In split scope the body scope was always getting marked as must
instantiate even though no closure was there in the body scope. I added a
new flag to the scope object to mark whether a closure is there.

### DIFF
--- a/lib/Runtime/ByteCode/ByteCodeEmitter.cpp
+++ b/lib/Runtime/ByteCode/ByteCodeEmitter.cpp
@@ -3278,9 +3278,9 @@ void ByteCodeGenerator::EmitOneFunction(ParseNode *pnode)
         {
             // Emit bytecode to copy the initial values from param names to their corresponding body bindings.
             // We have to do this after the rest param is marked as false for need declaration.
-            paramScope->ForEachSymbol([this, funcInfo, paramScope, byteCodeFunction](Symbol* param) {
+            paramScope->ForEachSymbol([&](Symbol* param) {
                 Symbol* varSym = funcInfo->GetBodyScope()->FindLocalSymbol(param->GetName());
-                Assert(varSym || param->GetIsArguments());
+                Assert(varSym || param->GetIsArguments() || pnode->sxFnc.pnodeName->sxVar.sym == param);
                 Assert(param->GetIsArguments() || param->IsInSlot(funcInfo));
                 if (varSym && varSym->GetSymbolType() == STVariable && (varSym->IsInSlot(funcInfo) || varSym->GetLocation() != Js::Constants::NoRegister))
                 {
@@ -3570,22 +3570,7 @@ void ByteCodeGenerator::EmitScopeList(ParseNode *pnode, ParseNode *breakOnBodySc
                 this->StartEmitFunction(pnode);
 
                 Scope* paramScope = pnode->sxFnc.funcInfo->GetParamScope();
-                Scope* bodyScope = pnode->sxFnc.funcInfo->GetBodyScope();
 
-                if (paramScope && !paramScope->GetCanMergeWithBodyScope())
-                {
-                    ParseNodePtr paramBlock = pnode->sxFnc.pnodeScopes;
-                    Assert(paramBlock->nop == knopBlock && paramBlock->sxBlock.blockType == Parameter);
-
-                    PushScope(paramScope);
-
-                    // While emitting the functions we have to stop when we see the body scope block.
-                    // Otherwise functions defined in the body scope will not be able to get the right references.
-                    this->EmitScopeList(paramBlock->sxBlock.pnodeScopes, pnode->sxFnc.pnodeBodyScope);
-                    Assert(this->GetCurrentScope() == paramScope);
-                }
-
-                PushScope(bodyScope);
                 // Persist outer func scope info if nested func is deferred
                 if (CONFIG_FLAG(DeferNested))
                 {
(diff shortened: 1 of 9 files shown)


aneeshdk:
@pleath @akroshg @tcare Can you please take a look?

test/es6/default-splitscope.js
```diff
@@ -86,7 +86,43 @@ var tests = [
...
+            if (a == 20) {
+                return 10;
+            }
+            var a = function () { return f8(20); };
+            return a;
+        }
+        assert.areEqual(10, f7()(), "Recursive call to the function from the body scope returns the right value");
+        
+        var f9 = function f10 (a = function ( ) { b; return f10(20); }, b) {
+            eval("");
```
akroshg: have a one more test where eval is inside the inner function.

lib/Runtime/ByteCode/ScopeInfo.h
```diff
@@ -41,6 +41,7 @@ namespace Js {
         BYTE isGlobalEval : 1;
         BYTE areNamesCached : 1;
         BYTE canMergeWithBodyScope : 1;
+        BYTE hasLocalInClosure : 1;
```
akroshg: initialized this value in the constructor below.

lib/Runtime/ByteCode/ByteCodeEmitter.cpp
```diff
@@ -3768,7 +3762,14 @@ void ByteCodeGenerator::StartEmitFunction(ParseNode *pnodeFnc)
         else
         {
             Symbol *sym = funcInfo->root->sxFnc.GetFuncSymbol();
-            funcInfo->bodyScope->AddSymbol(sym);
+            if (funcInfo->paramScope->GetCanMergeWithBodyScope())
```
akroshg: why not always add it to param scope?

lib/Runtime/ByteCode/ByteCodeEmitter.cpp
```diff
@@ -4016,6 +4017,13 @@ void ByteCodeGenerator::StartEmitFunction(ParseNode *pnodeFnc)
                 this->EnsureLetConstScopeSlots(pnodeFnc->sxFnc.pnodeBodyScope, funcInfo);
             }
         }
+
+        if (!paramScope->GetCanMergeWithBodyScope() && bodyScope->GetScopeSlotCount() == 0 && !bodyScope->GetHasOwnLocalInClosure())
+        {
+            // When we have split scope the body scope may be wrongly marked as must instantiate even though the capture occurred
+            // in param scope. This check is to make sure if no capture occurs in body scope make in not must instantiate.
+            bodyScope->SetMustInstantiate(false);
```
akroshg: so what happens in the case of eval?
you could split scope and there are no variable in the body scope.
Also why there is different condition to change the mustinstantiate why not do it on line 3807

lib/Runtime/ByteCode/ByteCodeEmitter.cpp
```diff
@@ -3605,12 +3605,6 @@ void ByteCodeGenerator::EmitScopeList(ParseNode *pnode, ParseNode *breakOnBodySc
 
                 this->EmitOneFunction(pnode);
                 this->EndEmitFunction(pnode);
-
-                if (paramScope && !paramScope->GetCanMergeWithBodyScope())
```
akroshg: Why are you not "pushing" the paramscope in the StartEmitFunction (line 3579) - then it will be consistent and easy to read.

akroshg: Nit: Similarly the pushing of bodyscope can also be moved to StartEmitFunction. 

---
In reply to: [61647086](https://github.com/Microsoft/ChakraCore/pull/892#discussion_r61647086) [](ancestors = 61647086)

test/es6/default-splitscope.js
aneeshdk: Done.

lib/Runtime/ByteCode/ScopeInfo.h
aneeshdk: Done.

lib/Runtime/ByteCode/ByteCodeEmitter.cpp
aneeshdk: The merge happens before this so I cannot add to param scope in non-split scope case.

lib/Runtime/ByteCode/ByteCodeEmitter.cpp
aneeshdk: If the eval is in the param scope body scope does not have to be an object. If the eval is in the body then the scope slot count will be updated accordingly and it will be dynamic so in FindScopeForSym we will anyway return from the body scope.

We cannot have this check in line 3807 because we have to wait till all the EnsureScopeSlots are done before checking the scope's scope slot count.

akroshg: Then what is the point on doing it on line 3807? why not move that here?

---
In reply to: [61648793](https://github.com/Microsoft/ChakraCore/pull/892#discussion_r61648793) [](ancestors = 61648793)

lib/Runtime/ByteCode/ByteCodeEmitter.cpp

lib/Runtime/ByteCode/ByteCodeEmitter.cpp

dilijev:
@dotnet-bot test this please

akroshg:
:+1:

dilijev:
@dotnet-bot test this please

PR #893, opened 2016-04-29
Fixing bug with hoisting a type check bailout

rajatd:
For a type check bailout (could be on a LdFld, CheckFixedFld, CheckObjType) hoisted out from an inlinee to outside the loop containing the inlinee, the polymorphicCacheIndex on the bailout record associated with the type check is set from the inlinee while the bailout itself is marked to be from the inliner. Both these things are fine on their own but cause problems when in conjunction. The inline cache index set on the bailout record (from the inlinee) might not be in the range of inline cache indices on the inliner function body. 
FIxing this by, on the bailout path, setting the correct function body on the bailout record for a shared bailout

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/893)


rajatd:
@LouisLaf , could you take a look?

rajatd:
@abchatra @pleath

rajatd:
@pleath @abchatra @LouisLaf , can one of you guys review this?

pleath:
LGTM.

PR #896, opened 2016-04-29
cross platform: fixes output message for ch

obastemur:


bin/ch/ch.cpp
```diff
@@ -121,17 +125,17 @@ int HostExceptionFilter(int exceptionCode, _EXCEPTION_POINTERS *ep)
...
 }
 
 void __stdcall PrintUsage()
 {
-#ifndef DEBUG
-    wprintf(_u("\nUsage: ch.exe filename")
-            _u("\n[flaglist] is not supported for Release mode\n"));
+#if !defined(DEBUG) || !defined(_WIN32)
+    wprintf(_u("\nUsage: %s <source file> %s"), hostName,
+            _u("\n[flaglist] is only supported for debug builds on Windows\n"));
```
jianchun: Why change this? `[flaglist]` should/would be support on Linux as well?

obastemur: Yes but now it's not (most probably for a while it may stay that way) and the current output gives a wrong impression. i.e. I thought process was crashing after taking `-?` argument. However the reason was the difference in initialization. We may enable it back when `flaglist` is actually supported?

obastemur:
This PR also fixes minor `ch` build issue with clang3.9

bin/ch/ch.cpp
jianchun: I see, LGTM then.

digitalinfinity:
LGTM- merging

digitalinfinity:
Well- whenever the CI system allows me too 😄

PR #900, opened 2016-04-30
param scope symbols should not go the propIdsForScopeSlotArray container.

akroshg:
This metadata is not required for param-scope symbols.


akroshg:
@aneeshdk - please have a look

aneeshdk:
LGTM

PR #901, opened 2016-04-30
cross platform: fix conditional behavior for x64 target

obastemur:
- `_M_X64_OR_ARM64` is a Windows specific and widely used macro
definition in ChakraCore. Define it for appropriate *nix platforms

 - Replace `_TARGET_MAC64` to `__APPLE__`


obastemur:
#905 replaces this one.

PR #902, opened 2016-04-30
cross platform: fix platform endianness check

obastemur:
By default(mostly), both `BIG_ENDIAN` and `LITTLE_ENDIAN` are defined
on *nix platforms. Unlike Windows, comparing the value of `BYTE_ORDER`
to `BIG_ENDIAN` or `LITTLE_ENDIAN` tells the endianness of the system.

P.S. : I couldn’t see any other usage of `BIG_ENDIAN` definition, otherwise
macro could be located under a common header file.

### DIFF
--- a/lib/Common/Common/NumberUtilities.inl
+++ b/lib/Common/Common/NumberUtilities.inl
@@ -7,11 +7,44 @@
 #define NUMBER_UTIL_INLINE
 #endif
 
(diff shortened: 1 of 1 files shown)


digitalinfinity:
:+1:

digitalinfinity:
Theoretically, the ifdefs should be in CommonPal but I guess we can refactor it out later

PR #904, opened 2016-05-02
cross platform: fix stack/heap overflow (ulong)

obastemur:
`ulong` (unsigned long) is 4 bytes on Windows while 8 bytes on *nix.

There is/are routines within ChakraCore assume casting from `double`
to `ulong` and then reaching to second item `[1]` is safe. However,
this ends up corrupting stack/heap on *nix.

PAL has a definition for ULONG as (unsigned int) and ULONG is already
widely used. For this reason, replacing all the `ulong` to `ULONG`


digitalinfinity:
Generally, this is fine. Couple comments:
1) Instead of replacing ulong, lets use uint32. We can leave existing instances of ULONG as is, but everywhere you've changed ulong, change it to uint32.
2) We need to do this for long as well (replace with int32)
3) In NumberUtilities, consider renaming FDblIsLong to FDblIsInt32.

jianchun:
+1 to avoid long/ulong/LONG/ULONG.

What do you think of int, uint? At least for some places? I understand int32, uint32 are safer. But int, uint look neater...

digitalinfinity:
I'm ambivalent because we also have int64 already so int32 feels more natural, plus int is also not necessarily defined to be the same everywhere (in practice, in all platforms we care about it is, but we once thought the same about long too :smile:)

obastemur:
Okay. I was thinking about this whole night. IMHO lets keep this PR as is and as simple. This doesn't mean that I'm not agree with @digitalinfinity or @jianchun but read the reasons below;

Reasons

1- If we change all the instances of ulong, ULONG to int32, we need to deal with (additional casting, redefining etc.) on Windows for WINAPI

2 - The proposed change is so big, I'm not sure if we have the right reasons to do that? Apart from the look and feel, ULONG should be just fine to let us move on.

3 - If we are going to refactor one type to another, I prefer doing that on master instead of linux branch. Otherwise, we will be keep fixing and maintaining the type diff for all the upcoming merges from master.

4 - This PR/Commit fixes the problem it's after without introducing big changes. Even if we want to refactor all the types, they can be part of a separate and more controlled task. IMO we have more types to refactor.

Apologies for the long read..

obastemur:
As suggested the instances of `ulong` are replaced with `uint32`. However I found keeping ULONG matching with the usage on some places hence used ULONG there instead.

Updating long to int32 will be part of another PR so feel free to review this and merge.

jianchun:
@dotnet-bot test Ubuntu ubuntu_linux_debug Ubuntu ubuntu_linux_release please

jianchun:

PR #905, opened 2016-05-02
cross platform: support amd64 on *nix

obastemur:
Currently *nix binary contains a portion of code from x86 specific implementation. This, 
generates an unstable behavior / crashes.

 - `_M_X64_OR_ARM64` is a Windows specific and a widely used macro
definition in ChakraCore. Define it for appropriate *nix platforms

 - Replace `_TARGET_MAC64` to `__APPLE__`

 - ~~Implement `amd64_Thunks.S`~~
 
- Profiling related code temporarily disabled for *nix builds

### DIFF
--- a/lib/Runtime/Base/ScriptContext.cpp
+++ b/lib/Runtime/Base/ScriptContext.cpp
@@ -3399,6 +3399,7 @@ namespace Js
         {
             return ProfileDeferredParsingThunk;
         }
+
         if (entryPoint == DefaultDeferredDeserializeThunk || entryPoint == ProfileDeferredDeserializeThunk)
         {
             return ProfileDeferredDeserializeThunk;
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -19,7 +19,6 @@ if(CMAKE_SYSTEM_NAME STREQUAL Linux)
     endif()
     set(CLR_CMAKE_PLATFORM_LINUX 1)
 elseif(CMAKE_SYSTEM_NAME STREQUAL Darwin)
-    add_definitions(-D_TARGET_MAC64)
     add_definitions(-DPLATFORM_UNIX)
 
     set(CLR_CMAKE_PLATFORM_UNIX 1)
@@ -34,14 +33,15 @@ else()
 endif()
 
 if(CLR_CMAKE_PLATFORM_UNIX_TARGET_AMD64)
-  set(CLR_CMAKE_PLATFORM_ARCH_AMD64 1)
+    add_definitions(-D_M_X64_OR_ARM64)
+    set(CLR_CMAKE_PLATFORM_ARCH_AMD64 1)
 endif()
 
 if (CMAKE_CXX_COMPILER_ID STREQUAL AppleClang
     OR CMAKE_CXX_COMPILER_ID STREQUAL Clang
     OR CMAKE_CXX_COMPILER_ID STREQUAL GNU)
-  # Color diagnostics for g++ and clang++
-  add_definitions("-fdiagnostics-color=always")
+    # Color diagnostics for g++ and clang++
+    add_definitions("-fdiagnostics-color=always")
 endif()
 
 if(CLR_CMAKE_PLATFORM_UNIX)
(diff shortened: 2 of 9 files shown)


lib/Backend/NativeCodeGenerator.cpp
```diff
@@ -1167,7 +1167,9 @@ NativeCodeGenerator::CheckCodeGen(Js::ScriptFunction * function)
             (
                 originalEntryPoint == DefaultEntryThunk
              || scriptContext->IsDynamicInterpreterThunk(originalEntryPoint)
+#ifdef ENABLE_SCRIPT_PROFILING
```
digitalinfinity: Assert is a macro and VC doesn't like macros within macros. You'll need to ifdef out the whole assert. But more generally, you shouldn't need this here yet since lib/Backend isn't being built on linux.

lib/Runtime/Base/FunctionBody.cpp
```diff
@@ -4653,11 +4653,14 @@ namespace Js
         if (!IsIntermediateCodeGenThunk(defaultEntryPointInfo->jsMethod)
             && defaultEntryPointInfo->jsMethod != DynamicProfileInfo::EnsureDynamicProfileInfoThunk)
         {
+#ifdef ENABLE_SCRIPT_PROFILING
```
digitalinfinity: Is this needed? This function looks like it's already wrapped under #ifdef ENABLE_SCRIPT_PROFILING

lib/Runtime/Base/ScriptContext.cpp
```diff
@@ -3395,10 +3395,12 @@ namespace Js
 #if ENABLE_NATIVE_CODEGEN
         Assert(!IsIntermediateCodeGenThunk(entryPoint));
 #endif
+#ifdef ENABLE_SCRIPT_PROFILING
```
digitalinfinity: Is this needed? It looks like this entire function is already under ENABLE_SCRIPT_PROFILING?

obastemur:
@digitalinfinity thanks for the review. PR is updated.

lib/Backend/NativeCodeGenerator.cpp
```diff
@@ -1163,6 +1163,7 @@ NativeCodeGenerator::CheckCodeGen(Js::ScriptFunction * function)
         // it either corresponds with this codegen episode (identified by function->entryPointIndex) of the function body
         // or one that was scheduled after. In the latter case originalEntryPoint will get updated if and when
         // that last episode completes successfully.
+#ifdef ENABLE_SCRIPT_PROFILING
```
digitalinfinity: As mentioned earlier- this is not necessary right now since Backend doesn't build

lib/Runtime/Base/FunctionBody.cpp
```diff
@@ -4848,16 +4848,21 @@ namespace Js
     void FunctionBody::SetEntryToDeferParseForDebugger()
     {
         ProxyEntryPointInfo* defaultEntryPointInfo = this->GetDefaultEntryPointInfo();
-        if (defaultEntryPointInfo->jsMethod != DefaultDeferredParsingThunk && defaultEntryPointInfo->jsMethod != ProfileDeferredParsingThunk)
-        {
 #ifdef ENABLE_SCRIPT_PROFILING
+        if (defaultEntryPointInfo->jsMethod != DefaultDeferredParsingThunk
```
digitalinfinity: Whoops- ended up fixing this already to unblock JC so you can just merge this change with my change

digitalinfinity:
Rest looks good- please rebase and I'll merge.

obastemur:
re-based to latest

lib/Backend/NativeCodeGenerator.cpp
digitalinfinity: @obastemur this is still showing up in my diff- can you remove this for now please, since we're not looking at touching Backend yet (or if you're going to keep it, you need to add an #else case to assert the entry point state...

obastemur:
@digitalinfinity thanks for the review. Updated the PR

digitalinfinity:
:+1:

PR #909, opened 2016-05-02
Don't try to create a RecyclableObject::FromVar without checking that the Var is in fact a RecyclableObject first.

dilijev:
Fixes issue introduced in #886

Fixes VSO 7399962

### DIFF
--- a/lib/Runtime/Language/JavascriptOperators.cpp
+++ b/lib/Runtime/Language/JavascriptOperators.cpp
@@ -8823,8 +8823,9 @@ CommonNumber:
 
     BOOL JavascriptOperators::ToPropertyDescriptor(Var propertySpec, PropertyDescriptor* descriptor, ScriptContext* scriptContext)
     {
-        if (JavascriptProxy::Is(propertySpec) ||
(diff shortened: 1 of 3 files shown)


dilijev:
@digitalinfinity quick review & sign off please?

dilijev:
Here are some reduced repros. I'm not sure if I should create unit tests to cover this. It seems like this should have been caught in unit tests initially.

```

Object.create({}, {a: 0});

// ---

Object.defineProperty({}, "x", 0);

// ---

Object.create({}, [0])

// --- 

a = 0;
x = function() { Object.create({}, this) }
x() // for sloppy-mode functions, it uses the global object as 'this', which has member 'a', and that can't be converted via RecyclableObject::FromVar
// x.call() // equivalent 
// x.call(null) // no repro using null
```

dilijev:
@MikeHolman FYI The last repro above is why the inclusion/exclusion of `"use strict"` would hide or reveal the bug. It changed the behavior of the call w.r.t. `this`

MikeHolman:
lgtm

dilijev:
@MikeHolman thanks!

PR #910, opened 2016-05-03
Assorted cleanup commits

digitalinfinity:
1) Cleanup some CMakeList filesl
2) Fix unnecessary reference to profiler thunk
3) Add simple test runner for hello world so that CI can run it

PR #911, opened 2016-05-03
xplat: support build with Ninja

jianchun:
Add build.sh `-n, --ninja` option to build with Ninja instead of make.
(Ninja uses multicore build by default and doesn't mix output from
different cores.)


jianchun:
This depends on a fix in #910 (`$` escape issue in ninja.build generation).

digitalinfinity:
LGTM

obastemur:
:+1:

PR #912, opened 2016-05-03
Reenable warning C4463, C4311, C4499 and C4091

curtisman:
Fix and reenable these warnings:
C4091 - ignore on left of '<type>' when no variable is declared
C4311 - 'type cast': pointer truncation from '' to ''
C4463  - overflow; assigning 1 to bit-field that can only hold values from -1 to 0
C4499 - explicit specialization cannot have a storage class (ignored)

### DIFF
--- a/lib/Parser/Scan.h
+++ b/lib/Parser/Scan.h
@@ -667,7 +667,7 @@ private:
     BOOL m_doubleQuoteOnLastTkStrCon :1;
     bool m_OctOrLeadingZeroOnLastTKNumber :1;
     BOOL m_fSyntaxColor : 1;            // whether we're just syntax coloring
-    BOOL m_EscapeOnLastTkStrCon:1;
+    bool m_EscapeOnLastTkStrCon:1;
     BOOL m_fNextStringTemplateIsTagged:1;   // the next string template scanned has a tag (must create raw strings)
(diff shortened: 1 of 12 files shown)


lib/Parser/Scan.h
```diff
@@ -667,7 +667,7 @@ class Scanner : public IScanner, public EncodingPolicy
     BOOL m_doubleQuoteOnLastTkStrCon :1;
     bool m_OctOrLeadingZeroOnLastTKNumber :1;
     BOOL m_fSyntaxColor : 1;            // whether we're just syntax coloring
```
agarwal-sandeep: Should all BOOL be changed to bool?

curtisman:
@digitalinfinity please review.

lib/Parser/Scan.h
curtisman: Yeah... that get rids of the warning.
I guess the rest of the fields should be change to bool too..... but those didn't cause warnings, so I didn't touch it.

dilijev:
LGTM

lib/Parser/Scan.h
ianwjhalliday: This pattern of mixing types on bit fields might be bloating our struct sizes: http://stackoverflow.com/questions/308364/c-bitfield-packing-with-bools

PR #913, opened 2016-05-03
cross platform: fix linux branch build break

obastemur:
Fixes the build break introduced with #910

Reason: CMake version needs to be 3.3 for COMPILE_LANGUAGE QUERY support

Solution could be fixing the minimum CMake version to 3.3 or implementing a
similar functionality without changing the CMake requirements.

This PR prefers the second option due to lack of availability of CMake 3.3 (from
package managers)

Current aproach is sufficient enough without introducing a new dependency.

Tested separately with `ninja-build` in addition to `make`.

~~Also, replaced `-Wl,--no-undefined` to it's old value `-Wl,-undefined,error`~~

~~`-Wl,-undefined,error` is cross platform `clang` compatible alternative~~

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -71,43 +71,49 @@ if(CLR_CMAKE_PLATFORM_UNIX)
 
     set(CMAKE_CXX_STANDARD 11)
 
-    set(c_warn_flags
-        -Wno-implicit-function-declaration
+    # CC WARNING FLAGS
+    set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} \
+        -Wno-implicit-function-declaration"
     )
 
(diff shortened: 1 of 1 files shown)


digitalinfinity:
I think this is fine but it appears that `-Wl,--no-undefined` is not the same as `-Wl,-undefined,error`. Specifically, the latter doesn't seem to cause a build failure when an import is unresolved, at least on clang 3.8. I recommend leaving that one as is or finding a solution that actually works there- we would have caught the unresolved dependency around the script profiling thunk sooner otherwise.

obastemur:
Ok. Updated the PR

digitalinfinity:
:+1:

Issue #626, opened 2016-03-22
Object.defineProperty assigns the wrong name to anonymous getter/setter functions


goyakin:
```JS
var o = {};
Object.defineProperty(o, 'test', {get: function () {}, set: function () {}});
var desc = Object.getOwnPropertyDescriptor(o, 'test');
desc.get.name === ' test';
desc.set.name === ' test';
```

The names should instead be:

```JS
desc.get.name === 'get';
desc.set.name === 'set';
```

akroshg:
@goyakin  is already working on this one.

PR #914, opened 2016-05-03
Function.name on the getter/setter on defineProperty.

akroshg:
For getter/setter function in the object literal we were initially giving them get/set name respectively. However if they appear in the defineProperty we were changing them for diagnostics purpose. In the diagnostics side we were expected to give <propname>.get but due to function.name we have changed it to "get <propname>"., which is not needed and not consistent.
This change is removing that and putting the <propname>.get as full name for the current function. The short name will now automatically be given as get/set correctly.

### DIFF
--- a/lib/Runtime/Library/JavascriptObject.cpp
+++ b/lib/Runtime/Library/JavascriptObject.cpp
@@ -2047,60 +2047,26 @@ namespace Js
     }
 
     /*static*/
-    char16 * JavascriptObject::ConstructAccessorNameES6(const PropertyRecord * propertyRecord, const char16 * getOrSetStr, ScriptContext* scriptContext)
-    {
-        Assert(propertyRecord);
-        Assert(scriptContext);
-        char16 * finalName = nullptr;
-        size_t propertyLength = static_cast<size_t>(propertyRecord->GetLength() + 1); //+ 1 (for null terminator)
-        if (propertyLength > 0)
-        {
-            size_t totalChars;
-            const size_t getSetLength = 4;    // 4 = 3 (get or set) +1 (for space)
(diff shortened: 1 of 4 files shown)


akroshg:
Fixes #626

akroshg:
@abchatra please review.

test/es6/function.name.js
```diff
@@ -750,6 +750,17 @@ var tests = [
             class C { foo(){} };
             assert.areEqual("foo",(new C).foo.name);
         }
+    },
+	{
+        name: "Getter and setter have correct name in defineProperty",
+        body: function()
+        {
+            var obj = {};
+            Object.defineProperty(obj, 'test', {get : function () {}, set : function () {} });
```
ljharb: should your test functions have names themselves, so the logic is explicit about which name gets priority?

lib/Runtime/Library/JavascriptObject.cpp
```diff
@@ -2082,25 +2082,20 @@ namespace Js
...
         {
+            charcount_t propertyLength = propertyRecord->GetLength();
+
             if (descriptor.GetterSpecified()
                 && Js::ScriptFunction::Is(descriptor.GetGetter())
                 && _wcsicmp(Js::ScriptFunction::FromVar(descriptor.GetGetter())->GetFunctionProxy()->GetDisplayName(), _u("get")) == 0)
             {
                 // modify to name.get
-                char16* finalName;
-                if (scriptContext->GetConfig()->IsES6FunctionNameEnabled())
```
akroshg: they are - they will have different PR.

abchatra:
:+1:

PR #916, opened 2016-05-03
MSFT: 7363961 Array.prototype.concat doesn't check if a Proxy is revoked when the Proxy is "this"

Yongqu:
There are places we directly access proxy's target & potentially handler after the proxy is revoked.
At check when non-proxy code access the handler & target, throw if proxy was revoked already.


Yongqu:

agarwal-sandeep:
LGTM

PR #917, opened 2016-05-03
Disable stack nested functions for split scope

aneeshdk:
When we have non-escaping functions in both param and body scope we are
trying to create stack nested functions for them. The problem is we try to
reuse the local frame display. In case of split scope we cannot do that as
both functions should get different frame displays. This changelist
disables stack nested functions inside split scoped functions.

### DIFF
--- a/lib/Runtime/ByteCode/ByteCodeGenerator.cpp
+++ b/lib/Runtime/ByteCode/ByteCodeGenerator.cpp
@@ -1819,6 +1819,17 @@ bool ByteCodeGenerator::CanStackNestedFunc(FuncInfo * funcInfo, bool trace)
         return false;
     }
 
+    if (funcInfo->paramScope && !funcInfo->paramScope->GetCanMergeWithBodyScope())
+    {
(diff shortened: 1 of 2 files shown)


aneeshdk:
@pleath @abchatra @akroshg @tcare Can you please take a look?

akroshg:
LGTM

lib/Runtime/ByteCode/ByteCodeGenerator.cpp
```diff
@@ -1829,6 +1829,11 @@ bool ByteCodeGenerator::CanStackNestedFunc(FuncInfo * funcInfo, bool trace)
             funcInfo->byteCodeFunction->GetDebugNumberSet(debugStringBuffer));
     }
 
+    if (funcInfo->paramScope && !funcInfo->paramScope->GetCanMergeWithBodyScope())
```
abchatra: Please add tracking using trace

abchatra:
:+1:

PR #919, opened 2016-05-04
Memset: Native arrays typecheck and missing item check

Cellule:
Memset with native float arrays needs to check the type of the source because we can do `nativearray[i] = var` and bailout after the fact if the source was not a float (ie: changes the type of array). 

Since memset will not change the type of the array, it needs to check before hand that the source is actually a float or int and bailout if not.

Additionally, I added a check for `IsMissingItem` for native int and float arrays.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/919)

### DIFF
--- a/lib/Runtime/Language/JavascriptOperators.cpp
+++ b/lib/Runtime/Language/JavascriptOperators.cpp
@@ -4685,7 +4685,13 @@ CommonNumber:
                 }
                 else if (instanceType == TypeIds_NativeIntArray)
                 {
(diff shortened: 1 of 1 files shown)


Cellule:
@LouisLaf can you review this please ?

Cellule:

rajatd:
:+1:

MikeHolman:
lgtm

PR #920, opened 2016-05-04
xplat: wmemset should return a value

jianchun:
wmemset implementation was missing return value. Trivial fix.


obastemur:
:+1:

jianchun:
@obastemur Thanks!

PR #922, opened 2016-05-04
xplat: fix SysFreeString implementation

jianchun:
PAL's SysFreeString implementation does not match SysAllocStringLen.
Should offset DWORD length before HeapFree.

### DIFF
--- a/pal/src/misc/bstr.cpp
+++ b/pal/src/misc/bstr.cpp
@@ -119,7 +119,7 @@ STDAPI_(BSTR) SysAllocStringLen(const OLECHAR *psz, UINT len)
 }
 
 /***
- *BSTR SysFreeString(char*)
+ *void SysFreeString(BSTR)
  *Purpose:
  *  Free a bstr using the passed in string
  *
@@ -130,16 +130,20 @@ STDAPI_(BSTR) SysAllocStringLen(const OLECHAR *psz, UINT len)
  *  return value = void
  *
  ***********************************************************************/
-STDAPI_(void) SysFreeString(const OLECHAR* psz)
+STDAPI_(void) SysFreeString(BSTR bstr)
 {
-    if (psz != NULL)
+    if (bstr != NULL)
     {
-        HeapFree(GetProcessHeap(), 0, (LPVOID) psz);
+        bstr = (BSTR) ((char*) bstr - sizeof(DWORD));
+#if defined(_WIN64)
+        bstr = (BSTR) ((char*) bstr - sizeof(DWORD));
+#endif
+        HeapFree(GetProcessHeap(), 0, (LPVOID) bstr);
     }
 }
 
 /***
- *BSTR SysStringLen(char*)
+ *UINT SysStringLen(BSTR)
  *Purpose:
  *  Return the length of the string in characters (not including null terminator)
  *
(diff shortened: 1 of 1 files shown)


pal/src/misc/bstr.cpp
```diff
@@ -130,16 +130,20 @@ STDAPI_(BSTR) SysAllocStringLen(const OLECHAR *psz, UINT len)
...
-STDAPI_(void) SysFreeString(const OLECHAR* psz)
+STDAPI_(void) SysFreeString(BSTR bstr)
 {
-    if (psz != NULL)
+    if (bstr != NULL)
     {
-        HeapFree(GetProcessHeap(), 0, (LPVOID) psz);
+        bstr = (BSTR) ((char*) bstr - sizeof(DWORD));
+#if defined(_WIN64)
+        bstr = (BSTR) ((char*) bstr - sizeof(DWORD));
```
obastemur: both 137 and 139 are exactly same. Do we need both for `_WIN64` ?

jianchun: This is because of extra offset done for `_WIN64` on line 105. Actually I don't understand line 105 or its comment. Maybe it was done for alignment on x64? It looks without the extra offset it works fine on Linux Amd64.

obastemur: Actually, AFAIK PAL (ChakraCore) is not for `_WIN64` anyways.

jianchun: Yes, that's why I said without it Linux x64 worked fine. There might be a perf difference. But BSTR isn't used a lot here and doesn't matter.

digitalinfinity:
LGTM

pal/src/misc/bstr.cpp
jianchun: > update: so maybe we should also remove 99 to 106 too ?

I thought about the same, but didn't make the change since we are not using this on _WIN64 anyway. Left it there in case one day we found we need it for Linux x64.

jianchun:
Thanks @obastemur @digitalinfinity

pal/src/misc/bstr.cpp
```diff
@@ -130,16 +130,20 @@ STDAPI_(BSTR) SysAllocStringLen(const OLECHAR *psz, UINT len)
  *  return value = void
  *
  ***********************************************************************/
-STDAPI_(void) SysFreeString(const OLECHAR* psz)
+STDAPI_(void) SysFreeString(BSTR bstr)
```
ThomsonTan: Preserve const?

jianchun: > Preserve const?

The original API signature on Windows and also https://github.com/Microsoft/ChakraCore/blob/linux/pal/inc/rt/palrt.h#L444 does not have const.

pal/src/misc/bstr.cpp
```diff
@@ -130,16 +130,20 @@ STDAPI_(BSTR) SysAllocStringLen(const OLECHAR *psz, UINT len)
...
 {
-    if (psz != NULL)
+    if (bstr != NULL)
     {
-        HeapFree(GetProcessHeap(), 0, (LPVOID) psz);
+        bstr = (BSTR) ((char*) bstr - sizeof(DWORD));
```
ThomsonTan: Instead of hard-coding to use char*, use or define some macro to access the heap address in BSTR? This looks a little weird because BSTR usually points to WCHAR.

jianchun: These are low level address manipulation code cloned and modified from existing SysAllocStringLen. Using "char*" to manipulate byte addresses are common elsewhere. Defining macro here looks overkill to me.

PR #923, opened 2016-05-04
add failfast if entryExitRecord list is messed up

MikeHolman:
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/923)

### DIFF
--- a/lib/Common/Exceptions/ReportError.cpp
+++ b/lib/Common/Exceptions/ReportError.cpp
@@ -106,4 +106,16 @@ __declspec(noinline) void FromDOM_NoScriptScope_fatal_error()
     ReportFatalException(NULL, E_UNEXPECTED, EnterScript_FromDOM_NoScriptScope, scenario);
 }
 
+__declspec(noinline) void EntryExitRecord_Corrupted_fatal_error()
+{
+    int scenario = 6;
+    ReportFatalException(NULL, E_UNEXPECTED, Fatal_EntryExitRecordCorruption, scenario);
+}
+
+__declspec(noinline) void UnexpectedExceptionHandling_fatal_error(EXCEPTION_POINTERS * originalException)
+{
+    int scenario = 7;
+    ReportFatalException(NULL, E_UNEXPECTED, Fatal_UnexpectedExceptionHandling, scenario);
+}
+
 #pragma optimize("",on)
--- a/lib/Runtime/Base/ThreadContext.cpp
+++ b/lib/Runtime/Base/ThreadContext.cpp
@@ -1874,9 +1874,9 @@ ThreadContext::PushEntryExitRecord(Js::ScriptEntryExitRecord * record)
         record->next = lastRecord;
 
         // these are on stack, which grows down. if this condition doesn't hold, then the list somehow got messed up
-        if ((uintptr_t)record > (uintptr_t)lastRecord)
+        if (!IsOnStack(lastRecord) || (uintptr_t)record >= (uintptr_t)lastRecord)
         {
-            RaiseFailFastException(nullptr, nullptr, 0);
+            EntryExitRecord_Corrupted_fatal_error();
         }
     }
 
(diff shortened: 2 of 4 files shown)


MikeHolman:
@LouisLaf can you take a look?

akroshg:
do we want to this on release/1.2 ?

MikeHolman:
Yes, I think so. It should help our analysis of reliability bugs.

lib/Runtime/Base/ThreadContext.cpp
```diff
@@ -1872,6 +1872,12 @@ ThreadContext::PushEntryExitRecord(Js::ScriptEntryExitRecord * record)
         Assert(lastRecord->leaveForHost || lastRecord->leaveForAsyncHostOperation);
         lastRecord->hasReentered = true;
         record->next = lastRecord;
+
+        // these are on stack, which grows down. if this condition doesn't hold, then the list somehow got messed up
+        if ((uintptr_t)record > (uintptr_t)lastRecord)
+        {
+            RaiseFailFastException(NULL, NULL, NULL);
```
ThomsonTan: Use nullptr? Furthermore, could we record some value/flag here for analysis?

MikeHolman: I think in our codebase we generally use NULL rather than nullptr when calling Windows APIs. Dumps from FailFast exception should contain same information as any other crash, so I think there's no need to record anything special here.

ThomsonTan: There is no such convesion. You could see "RaiseFailFastException(nullptr, nullptr, 0);" in many similar projects.

---
In reply to: [62121674](https://github.com/Microsoft/ChakraCore/pull/923#discussion_r62121674) [](ancestors = 62121674)

MikeHolman: Ok, that was just my observation of our code. Anyway, I'll change it.

lib/Runtime/Base/ThreadContext.cpp
```diff
@@ -1872,6 +1872,12 @@ ThreadContext::PushEntryExitRecord(Js::ScriptEntryExitRecord * record)
         Assert(lastRecord->leaveForHost || lastRecord->leaveForAsyncHostOperation);
         lastRecord->hasReentered = true;
         record->next = lastRecord;
+
+        // these are on stack, which grows down. if this condition doesn't hold, then the list somehow got messed up
+        if ((uintptr_t)record > (uintptr_t)lastRecord)
+        {
+            RaiseFailFastException(nullptr, nullptr, 0);
```
agarwal-sandeep: Why not use ReportFatalException similar to Binary_Inconsistency_fatal_error with a different ErrorReason

lib/Runtime/Base/ThreadContext.cpp
```diff
@@ -1872,6 +1872,12 @@ ThreadContext::PushEntryExitRecord(Js::ScriptEntryExitRecord * record)
         Assert(lastRecord->leaveForHost || lastRecord->leaveForAsyncHostOperation);
         lastRecord->hasReentered = true;
         record->next = lastRecord;
+
+        // these are on stack, which grows down. if this condition doesn't hold, then the list somehow got messed up
+        if ((uintptr_t)record > (uintptr_t)lastRecord)
```
LouisLaf: That's a good check, but why not call IsOnStack() as well?

lib/Runtime/Base/ThreadContext.cpp
MikeHolman: That's a good idea. I will make this change.

lib/Common/Exceptions/ReportError.cpp
```diff
@@ -106,4 +106,16 @@ __declspec(noinline) void FromDOM_NoScriptScope_fatal_error()
     ReportFatalException(NULL, E_UNEXPECTED, EnterScript_FromDOM_NoScriptScope, scenario);
 }
 
+__declspec(noinline) void EntryExitRecord_Corrupted_fatal_error()
+{
+    int scenario = 6;
+    ReportFatalException(NULL, E_UNEXPECTED, Fatal_EntryExitRecordCorruption, scenario);
+}
+
+__declspec(noinline) void UnexpectedExceptionHandling_fatal_error(EXCEPTION_POINTERS * originalException)
```
agarwal-sandeep: What's the use of this argument, this is not used?

MikeHolman: The point is so that the original context will be very clearly available when debugging.

LouisLaf:
:shipit:

PR #924, opened 2016-05-05
xplat: custom calling convention

jianchun:
In order to support JavascriptStackWalker and RUNTIME_ARGUMENTS for
JavascriptMethod, we need all arguments on stack. This change implements
a custom calling convention to invoke JavascriptMethod. JavascriptMethod
entry stack layout:

```
      [Return Address]
      [function]              == RDI
      [callInfo]              == RSI
      [arg0]
      [arg1]
      ...
```

RDI/RSI are same as System V AMD64 ABI, thus `function/callInfo` can be
accessed in C++ code. Stack layout is the same as Windows x86.

Also annotated `amd64_CallFunction` assembly code with EHFrame/debugging
info using unix asm macros. Now JS exception throw/catch works, and gdb
can unwind the stack correctly.

Fixed `JavascriptFunction::DeferredParsingThunk` assembly code to pass
correct argument, `&function`, to call `JavascriptFunction::DeferredParse`.


lib/Runtime/Library/amd64/JavascriptFunctionA.S
```diff
@@ -2,27 +2,41 @@
...
+//------------------------------------------------------------------------------
+NESTED_ENTRY amd64_CallFunction, _TEXT, NoHandler
         // Var amd64_CallFunction(RecyclableObject *function,
         //	JavascriptMethod entryPoint, CallInfo callInfo, uint argc, Var *argv) ;
 
         // Save existing rbp
-        push rbp
+        push_nonvol_reg rbp
+        mov rbp, rsp
+        set_cfa_register rbp, (2*8)
```
digitalinfinity: What does this do- can you add a comment around this please?

lib/Runtime/Library/amd64/JavascriptFunctionA.S
```diff
@@ -2,27 +2,41 @@
...
+//------------------------------------------------------------------------------
+// Invoke
+//
+//      JavascriptMethod(RecyclableObject* function, CallInfo callInfo, ...)
+//
+// with a custom calling convention in order to support JavascriptStackWalker
+// and RUNTIME_ARGUMENTS. JavascriptMethod entry stack layout:
+//
+//      [Return Address]
+//      [function]              == RDI
```
digitalinfinity: Arent the callee save registers in between the Return address and function object?

digitalinfinity: Oh I see- you meant the stack layout on entering JavascriptMethod- got it.

---
In reply to: [62139178](https://github.com/Microsoft/ChakraCore/pull/924#discussion_r62139178) [](ancestors = 62139178)

lib/Runtime/Library/amd64/JavascriptFunctionA.S
```diff
@@ -37,112 +51,94 @@ C_FUNC(amd64_CallFunction):
...
+LOCAL_LABEL(copy_args_to_stack):
+        mov r10, rcx
+        dec r10
+        shl r10, 3              // r10 = (argc - 1) * 8
+        lea rax, [r8]           // rax = argv
+        lea rbx, [rsp + 10h]    // rbx = &rsp[2]
+
+        // Copying from rax+r10 to rbx+r10
+LOCAL_LABEL(copy_args):
         mov r12, qword ptr [rax + r10]
```
digitalinfinity: Curious- since we're copying all arguments to the stack- can we use rep movsq here?

lib/Runtime/Library/amd64/JavascriptFunctionA.S
```diff
@@ -2,27 +2,41 @@
...
+//      [arg1]
+//      ...
+//------------------------------------------------------------------------------
+NESTED_ENTRY amd64_CallFunction, _TEXT, NoHandler
         // Var amd64_CallFunction(RecyclableObject *function,
         //	JavascriptMethod entryPoint, CallInfo callInfo, uint argc, Var *argv) ;
 
         // Save existing rbp
-        push rbp
+        push_nonvol_reg rbp
```
digitalinfinity: When do we use push_nonvol_reg and should we do this for the rest of the callee saves?

digitalinfinity:
:+1:

lib/Runtime/Library/amd64/JavascriptFunctionA.S
jianchun: Yes we can. Didn't try that for sake of less change; and most of the time we have only a handful of args.

lib/Runtime/Library/amd64/JavascriptFunctionA.S
jianchun: These are macros. Will add comment.

lib/Runtime/Library/amd64/JavascriptFunctionA.S
jianchun: > When do we use push_nonvol_reg and should we do this for the rest of the callee saves?

It is a macro to push reg and adjust CFA offset accordingly. The next annotation changed CFA relative register to rbp and rbp remains unchanged in this function, thus all the rest push/pop doesn't affect CFA offset and doesn't need any adjustment. That's why all the rest are direct push/pop. (Copied this from a PAL .S file.That's my understanding.)

lib/Runtime/Library/amd64/JavascriptFunctionA.S
```diff
@@ -2,147 +2,122 @@
...
+//      [function]              == RDI
+//      [callInfo]              == RSI
+//      [arg0]
+//      [arg1]
+//      ...
+//------------------------------------------------------------------------------
+NESTED_ENTRY amd64_CallFunction, _TEXT, NoHandler
+        // Var amd64_CallFunction(
+        //      RecyclableObject *function,             (rdi)
+        //	JavascriptMethod entryPoint,            (rsi)
```
digitalinfinity: Nit: tab to space

jianchun: > Nit: tab to space

That's why the diff was not aligned... lost my VS Code settings during Ubuntu upgrade.

lib/Runtime/Library/amd64/JavascriptFunctionA.S
```diff
@@ -2,147 +2,122 @@
...
+        //      CallInfo callInfo,                      (rdx)
+        //      uint argc,                              (rcx)
+        //      Var *argv) ;                            (r8)
+
+        push_nonvol_reg rbp             // push rbp and adjust CFA offset
```
digitalinfinity: I would add a little about CFA = Canonical Frame as described by the DWARF spec (and maybe a link to it) since most of the team would probably not be familiar with it

jianchun: Done

PR #915, opened 2016-05-03
cross platform: Enable DateTime

obastemur:
This PR enables DateTime

PR #925, opened 2016-05-05
cross platform: Enable Date features with system default globalization

obastemur:
This PR may require the presence of #915

### DIFF
--- a/lib/Runtime/Base/ScriptContext.cpp
+++ b/lib/Runtime/Base/ScriptContext.cpp
@@ -2337,13 +2337,17 @@ namespace Js
             return si;
     }
 
-#ifdef ENABLE_GLOBALIZATION
     void ScriptContext::UpdateTimeZoneInfo()
     {
         GetTimeZoneInformation(&timeZoneInfo);
+        
+        // todo: check possible winrt issue
+        // !defined(__cplusplus_winrt)
+        // see https://msdn.microsoft.com/en-us/library/90s5c885.aspx
+#if defined(_WIN32)
         _tzset();
-    }
 #endif
+    }
 
 #ifdef PROFILE_EXEC
     void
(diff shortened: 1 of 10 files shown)


lib/Runtime/Base/ScriptContext.cpp
```diff
@@ -2337,13 +2337,13 @@ namespace Js
             return si;
     }
 
-#ifdef ENABLE_GLOBALIZATION
     void ScriptContext::UpdateTimeZoneInfo()
     {
         GetTimeZoneInformation(&timeZoneInfo);
+#if !defined(__cplusplus_winrt) && defined(_WIN32)
```
jianchun: Why adding `!defined(__cplusplus_winrt)`? It worked before without it.

Linux has `tzset()`. Does it work here?

obastemur: Saw the `important` message from https://msdn.microsoft.com/en-us/library/90s5c885.aspx

AFAIK using `tzset` on *nix may depend on the choice of localization.

jianchun: > Saw the important message from https://msdn.microsoft.com/en-us/library/90s5c885.aspx

I'm unfamiliar with it. It might be possible that chakra.dll can use this API? Maybe better avoid side-fixing it here and file an issue to investigate it separately?

obastemur: > I'm unfamiliar with it. It might be possible that chakra.dll can use this API? Maybe better avoid side-fixing it here and file an issue to investigate it separately?

Okay, added a comment there instead.

jianchun:
LGTM, thanks!

obastemur:
Updated this PR. Since we haven't merged #915 yet, combination was enabling some windows only features. After both PRs merged, we should have a functioning DateTime support.

jianchun:

obastemur:
This was cherry-picked into #915

PR #931, opened 2016-05-05
If debugger attach/detach fails terminate the process as we can't recover from failure during attach/detach

agarwal-sandeep:
Fixes VSO bug 4158821

### DIFF
--- a/lib/Common/Exceptions/ReportError.cpp
+++ b/lib/Common/Exceptions/ReportError.cpp
@@ -106,4 +106,10 @@ __declspec(noinline) void FromDOM_NoScriptScope_fatal_error()
     ReportFatalException(NULL, E_UNEXPECTED, EnterScript_FromDOM_NoScriptScope, scenario);
 }
 
+__declspec(noinline) void Debugger_AttachDetach_fatal_error()
+{
+    int scenario = 5;
+    ReportFatalException(NULL, E_UNEXPECTED, Fatal_Debugger_AttachDetach_Failure, scenario);
+}
+
 #pragma optimize("",on)
--- a/lib/Common/Exceptions/Throw.h
+++ b/lib/Common/Exceptions/Throw.h
@@ -228,8 +228,4 @@ namespace Js {
     catch (ex) \
     {
 
-#ifdef ENABLE_DEBUG_CONFIG_OPTIONS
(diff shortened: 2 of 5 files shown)


lib/Common/Exceptions/ReportError.cpp
```diff
@@ -106,4 +106,10 @@ __declspec(noinline) void FromDOM_NoScriptScope_fatal_error()
     ReportFatalException(NULL, E_UNEXPECTED, EnterScript_FromDOM_NoScriptScope, scenario);
 }
 
+__declspec(noinline) void Debugger_AttachDetach_fatal_error()
+{
+    int scenario = 5;
```
akroshg: what is this magic number?

lib/Common/Exceptions/Throw.h
```diff
@@ -228,8 +228,4 @@ namespace Js {
     catch (ex) \
     {
 
-#ifdef ENABLE_DEBUG_CONFIG_OPTIONS
-#define RAISE_FATL_INTERNAL_ERROR_IFFAILED(hr) if (hr != S_OK) Js::Throw::FatalInternalError();
-#else
-#define RAISE_FATL_INTERNAL_ERROR_IFFAILED(hr)
-#endif
+#define DEBUGGER_ATTACHDETACH_FATAL_ERROR_IF_FAILED(hr) if (hr != S_OK) Debugger_AttachDetach_fatal_error();
```
akroshg: oh - we are doing fatal even in the fre build?

lib/Common/Exceptions/ReportError.cpp
agarwal-sandeep: It is an unused parameter which was put with the idea to group different kinds of fata error, but is not used. All other functions have this. I can put any number.

akroshg:
ensure web-crawler before you commit.

lib/Common/Exceptions/Throw.h
akroshg: I realized we are doing this for RS2.

agarwal-sandeep: I would say yes to prevent random issue which can come if this failed. Since this is going in master and will have time to flush any issues.

agarwal-sandeep:
Currently WebCrawler doesn't run on master, I will merge this change and when we start running WebCrawler using master will fix any issue which comes.

akroshg:
looks good to me.

PR #935, opened 2016-05-06
Specify path to clang on Ubuntu 16.04 machines.

dilijev:
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/935)


dilijev:
@mmitche @digitalinfinity FYI

netci.groovy
```diff
@@ -124,7 +124,7 @@ def CreateLinuxBuildTasks = { machine, configTag, linuxBranch, nonDefaultTaskSet
 
             def infoScript = 'bash jenkins/get_system_info.sh'
             def debugFlag = buildType == 'debug' ? '--debug' : ''
-            def buildScript = "bash ./build.sh -j=`nproc` ${debugFlag}"
+            def buildScript = "bash ./build.sh -j=`nproc` ${debugFlag} --cxx=/usr/bin/clang++-3.5 --cc=/usr/bin/clang-3.5"
```
obastemur: It should be clang++-3.7 or something newer. 

Actually we may need to test both 3.7 and 3.9

Internally, there are some parts don't comply with the latest updates to clang, however keeping 3.7 is supported is also important.

dilijev: At the moment, 3.5 was all we had to work with. I'm talking to the owners to add clang 3.7

PR #936, opened 2016-05-06
Fix numerous issues with Use Before Declaration

### LINKED ISSUE #505: Use before declaration error has wrong line number.
```javascript
function test3() {
  var k = 1;
  d = 2;
  let d;
}
test3();

```
Emits 
```
ReferenceError: Use before declaration
   at test3 (e:\chgittest\bug8.js:2:3)
   at Global code (e:\chgittest\bug8.js:6:1)
```
The line number should be 3.
`EmitUseBeforeDeclaration `does not have enclosing `Start/End Statement`.

ianwjhalliday:
Fixes #505 and VSO 5425942

Emit() needs to visit all AST nodes so that various algorithms can
collect data and set info on symbols and such, assign locations etc.

In our implementation of the let/const lexical variable Use Before
Declaration error, we had a call to EmitUseBeforeDeclaration() at the
beginning of Emit() and would exit early if it decided to emit the error
into the bytecode.  This would prevent visitation of subtrees in certain
cases, like e.g. issue #505.  It would do this for assignments, i.e.
stores.  Detection for Use Before Declaration for loads (and typeof)
are done in EmitPropLoad and EmitPropTypeof.

This fix pushes Use Before Declaration detection for stores down into
EmitPropStore and fixes up where we clear the `needsDeclaration` flag on
symbols that we weren't doing exactly right for EmitPropStore to work
this way.

Summary of changes

 - Tweak parse tree tracing to note deferred funcs
 - Change EmitPropStore to emit static UseBeforeDecl
 - Consolidate Emit() of knop[Var|Let|Const]Decl
 - Reduce code duplication in TrackMemberNodesInObjectForIntConstants
 - Add missing EmitUseBeforeDeclaration in EmitReference for knopCall
 - Remove EmitUseBeforeDeclaration for knopVarDecl case in Emit()
 - Remove redundant parameter fLoadUndef from
   EmitUseBeforeDeclarationRuntimeError().  Use the presence of the
   location parameter to indicate that a LdUndef is needed instead.
 - Remove fAcquire param from EmitUseBeforeDeclaration
 - Remove pnodeNeedUndef parameter from EmitUseBeforeDeclaration
 - Remove always true fAssignRegs parameter from EmitApplyCall.
 - Removed call to EmitUseBeforeDeclaration in
   EmitCallTargetNoEvalComponents.  This was the only remaining call
   that used pnodeNeedUndef, hence ability to remove that parameter
   from EmitUseBeforeDeclaration.
 - Added three test cases to unittests

### DIFF
--- a/lib/Parser/Parse.cpp
+++ b/lib/Parser/Parse.cpp
@@ -12915,6 +12915,12 @@ void PrintPnodeWIndent(ParseNode *pnode,int indentAmt) {
       PrintFormalsWIndent(pnode->sxFnc.pnodeParams, indentAmt + INDENT_SIZE);
       PrintPnodeWIndent(pnode->sxFnc.pnodeRest, indentAmt + INDENT_SIZE);
       PrintPnodeWIndent(pnode->sxFnc.pnodeBody, indentAmt + INDENT_SIZE);
+      if (pnode->sxFnc.pnodeBody == nullptr)
+      {
+          Output::Print(_u("[%4d, %4d): "), pnode->ichMin, pnode->ichLim);
+          Indent(indentAmt + INDENT_SIZE);
+          Output::Print(_u("<parse deferred body>\n"));
+      }
       break;
       //PTNODE(knopProg       , "program"    ,None    ,Fnc  ,fnopNone)
   case knopProg:
--- a/lib/Runtime/ByteCode/ByteCodeEmitter.cpp
+++ b/lib/Runtime/ByteCode/ByteCodeEmitter.cpp
@@ -12,8 +12,8 @@ void EmitLoad(ParseNode *rhs, ByteCodeGenerator *byteCodeGenerator, FuncInfo *fu
 void EmitCall(ParseNode* pnode, Js::RegSlot rhsLocation, ByteCodeGenerator* byteCodeGenerator, FuncInfo* funcInfo, BOOL fReturnValue, BOOL fEvaluateComponents, BOOL fHasNewTarget, Js::RegSlot overrideThisLocation = Js::Constants::NoRegister);
 void EmitSuperFieldPatch(FuncInfo* funcInfo, ParseNode* pnode, ByteCodeGenerator* byteCodeGenerator);
 
-bool EmitUseBeforeDeclaration(ParseNode *pnode, ByteCodeGenerator *byteCodeGenerator, FuncInfo *funcInfo);
-void EmitUseBeforeDeclarationRuntimeError(ByteCodeGenerator *byteCodeGenerator, Js::RegSlot location, bool fLoadUndef = true);
+bool EmitUseBeforeDeclaration(Symbol *sym, ByteCodeGenerator *byteCodeGenerator, FuncInfo *funcInfo);
+void EmitUseBeforeDeclarationRuntimeError(ByteCodeGenerator *byteCodeGenerator, Js::RegSlot location);
 void VisitClearTmpRegs(ParseNode * pnode, ByteCodeGenerator * byteCodeGenerator, FuncInfo * funcInfo);
 
 bool CallTargetIsArray(ParseNode *pnode)
@@ -2220,7 +2220,7 @@ void ByteCodeGenerator::EmitScopeSlotLoadThis(FuncInfo *funcInfo, Js::RegSlot re
             // been called inside a lambda so we can check to see if we called
             // super and assigned to the this register already. If not, this should trigger
             // a ReferenceError.
-            EmitUseBeforeDeclarationRuntimeError(this, regLoc, false);
+            EmitUseBeforeDeclarationRuntimeError(this, Js::Constants::NoRegister);
         }
     }
     else if (this->flags & fscrEval && (funcInfo->IsGlobalFunction() || (funcInfo->IsLambda() && nonLambdaFunc->IsGlobalFunction()))
@@ -2397,7 +2397,7 @@ void ByteCodeGenerator::EmitClassConstructorEndCode(FuncInfo *funcInfo)
     {
         // This is the case where we don't have any references to this or super in the constructor or any nested lambda functions.
         // We know 'this' must be undecl so let's just emit the ReferenceError as part of the fallthrough.
-        EmitUseBeforeDeclarationRuntimeError(this, ByteCodeGenerator::ReturnRegister, true);
+        EmitUseBeforeDeclarationRuntimeError(this, ByteCodeGenerator::ReturnRegister);
     }
 }
 
@@ -2694,7 +2694,12 @@ void ByteCodeGenerator::EmitFunctionBody(FuncInfo *funcInfo)
                         Output::Print(_u("--- DelayCapture: Committed symbol '%s' to slot.\n"), sym->GetName());
                         Output::Flush();
                     }
+                    // REVIEW[ianhall]: HACK to work around this causing an error due to sym not yet being initialized
+                    // what is this doing? Why are we assigning sym to itself?
+                    bool old = sym->GetNeedDeclaration();
+                    sym->SetNeedDeclaration(false);
                     this->EmitPropStore(sym->GetLocation(), sym, sym->GetPid(), funcInfo, decl->nop == knopLetDecl, decl->nop == knopConstDecl);
+                    sym->SetNeedDeclaration(old);
                 }
             }
             NEXT_SLIST_ENTRY;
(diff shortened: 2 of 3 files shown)


ianwjhalliday:
@pleath @Microsoft/chakra-es please review

pleath:
I like it already. (But I’ll look in detail tomorrow.)

--Paul

ianwjhalliday:
I forgot to note that I ran test262 with and without my changes.  No new failures were introduced by my change.

lib/Runtime/ByteCode/ByteCodeEmitter.cpp
```diff
@@ -6026,16 +6037,6 @@ void EmitDestructuredElement(ParseNode *elem, Js::RegSlot sourceLocation, ByteCo
         break;
 
     case knopName:
```
pleath: Nice. Can this case just go away?

lib/Runtime/ByteCode/ByteCodeEmitter.cpp
```diff
@@ -9429,22 +9396,6 @@ void Emit(ParseNode *pnode, ByteCodeGenerator *byteCodeGenerator, FuncInfo *func
         return;
     }
 
-    if (EmitUseBeforeDeclaration(pnode, byteCodeGenerator, funcInfo))
```
pleath: Boom!

pleath:
Very nice cleanup. Ship it.

test/LetConst/tdz2.js
```diff
@@ -52,4 +52,80 @@ test1();
...
+        };
+        class u3056 {static b(){ var x = z; }}(window).bind;
+    };
+    f();
+
+    if (w !== undefined) {
+        WScript.Echo('test4 failed, w not undefined second time');
+    }
+}
+test4();
```
dilijev: Are you missing calls to test2() and test3()?

lib/Runtime/ByteCode/ByteCodeEmitter.cpp
```diff
@@ -9983,7 +9934,7 @@ void Emit(ParseNode *pnode, ByteCodeGenerator *byteCodeGenerator, FuncInfo *func
             if (pnode->sxCall.isApplyCall && funcInfo->GetApplyEnclosesArgs())
             {
                 // TODO[ianhall]: Can we remove the ApplyCall bytecode gen time optimization?
```
dilijev: Have you revisited this question?

lib/Runtime/ByteCode/ByteCodeEmitter.cpp
```diff
@@ -2694,7 +2694,12 @@ void ByteCodeGenerator::EmitFunctionBody(FuncInfo *funcInfo)
                         Output::Print(_u("--- DelayCapture: Committed symbol '%s' to slot.\n"), sym->GetName());
                         Output::Flush();
                     }
+                    // REVIEW[ianhall]: HACK to work around this causing an error due to sym not yet being initialized
+                    // what is this doing? Why are we assigning sym to itself?
+                    bool old = sym->GetNeedDeclaration();
+                    sym->SetNeedDeclaration(false);
```
dilijev: I just want to draw attention to this `REVIEW` comment (as I don't have an answer). Should this question block completion of the review?

dilijev:
Otherwise LGTM

tcare:
Appreciate the cleanup effort. Looks good.

lib/Runtime/ByteCode/ByteCodeEmitter.cpp
ianwjhalliday: Thanks for calling this out, @dilijev.

@pleath I meant to ask you in person about this on Friday but then I was sick.  Do you have any ideas about how to handle this better, or is this a reasonable solution?

lib/Runtime/ByteCode/ByteCodeEmitter.cpp
ianwjhalliday: Not recently.  It is not a priority for me at this time though.

test/LetConst/tdz2.js
ianwjhalliday: Oops, yes I am!  Good catch

lib/Runtime/ByteCode/ByteCodeEmitter.cpp
ianwjhalliday: We can remove it and let it fall into the default call to EmitReference, which will then also do nothing for knopName.

---
In reply to: [62540326](https://github.com/Microsoft/ChakraCore/pull/936#discussion_r62540326) [](ancestors = 62540326)

lib/Runtime/ByteCode/ByteCodeEmitter.cpp
```diff
@@ -4875,6 +4880,11 @@ void ByteCodeGenerator::EmitPropStore(Js::RegSlot rhsLocation, Symbol *sym, Iden
     }
 
     // Arrived at the scope in which the property was defined.
+    if (sym && sym->GetNeedDeclaration() && scope->GetFunc() == funcInfo)
+    {
+        EmitUseBeforeDeclarationRuntimeError(this, Js::Constants::NoRegister);
+    }
```
ianwjhalliday: I want to call out here that fall through into the rest of `EmitPropStore` is intentional, whereas if you look at `EmitPropLoad/Typeof` they both stop at this point.

The intention is that this way we don't need `EmitUseBeforeDeclarationRuntimeError` to do something special to make the bytecode well-formed.  Instead we just let `EmitPropStore` do the thing its always done, except it is preceeded by a `RuntimeError` opcode.

We could consider taking this further and removing the `LdUndef` from `EmitUseBeforeDeclarationRuntimeError` and making `EmitPropLoad/Typeof` also continue after emitting the `RuntimeError` bytecode, but I think that is out of scope at this point in the release.  No need to stir that up.

ianwjhalliday: I'll leave a comment here to call out the intentional fall through

---
In reply to: [62564629](https://github.com/Microsoft/ChakraCore/pull/936#discussion_r62564629) [](ancestors = 62564629)

lib/Runtime/ByteCode/ByteCodeEmitter.cpp
ianwjhalliday: Oh yeah.  I've been wanting to make this change for a long while.

I had a change that did it part way from years ago but I couldn't find the pack.  I think the work I did here to redo it resulted in a cleaner change though.

---
In reply to: [62540552](https://github.com/Microsoft/ChakraCore/pull/936#discussion_r62540552) [](ancestors = 62540552)

lib/Runtime/ByteCode/ByteCodeEmitter.cpp
```diff
@@ -8850,50 +8851,28 @@ void EmitBinaryReference(ParseNode *pnode1, ParseNode *pnode2, ByteCodeGenerator
...
+
+    if (location != Js::Constants::NoRegister)
     {
+        // Optionally load something into register in order to do not confuse IRBuilder. This value will never be used.
         byteCodeGenerator->Writer()->Reg1(Js::OpCode::LdUndef, location);
     }
 }
 
-bool EmitUseBeforeDeclaration(ParseNode *pnode, ByteCodeGenerator *byteCodeGenerator, FuncInfo *funcInfo)
+bool EmitUseBeforeDeclaration(Symbol *sym, ByteCodeGenerator *byteCodeGenerator, FuncInfo *funcInfo)
```
akroshg: nit: we do not use the return value (bool) of this function EmitUseBeforeDeclaration. You could remove it to make it simple.

lib/Parser/Parse.cpp
```diff
@@ -12915,6 +12915,12 @@ void PrintPnodeWIndent(ParseNode *pnode,int indentAmt) {
       PrintFormalsWIndent(pnode->sxFnc.pnodeParams, indentAmt + INDENT_SIZE);
       PrintPnodeWIndent(pnode->sxFnc.pnodeRest, indentAmt + INDENT_SIZE);
       PrintPnodeWIndent(pnode->sxFnc.pnodeBody, indentAmt + INDENT_SIZE);
+      if (pnode->sxFnc.pnodeBody == nullptr)
+      {
+          Output::Print(_u("[%4d, %4d): "), pnode->ichMin, pnode->ichLim);
```
akroshg: Nit: We have already printed this in the beginning of this function. Seems redundant to me.

akroshg:
Looks good to me as well.

lib/Parser/Parse.cpp
ianwjhalliday: Originally I thought that too but it turns out this is the prefix for each line of output in the AST trace.  Since I was printing a new line it looked more consistent having the prefix there, and also made it easier to do proper alignment of the text.

---
In reply to: [62569197](https://github.com/Microsoft/ChakraCore/pull/936#discussion_r62569197) [](ancestors = 62569197)

lib/Runtime/ByteCode/ByteCodeEmitter.cpp
ianwjhalliday: True! I will make it void.

`EmitUseBeforeDeclaration` is now only used in one location and I couldn't reason about it to sufficiently convince myself that I could reduce down that call site to a direct call to `EmitUseBeforeDeclarationRuntimeError`.  In particular the `HasStaticPathToAncestor` guard had me concerned.

Instead of inlining this function at that one call site I decided to leave it as a proper function.

I would be thrilled if we can reduce this down further and eliminate it altogether at some point.

---
In reply to: [62568830](https://github.com/Microsoft/ChakraCore/pull/936#discussion_r62568830) [](ancestors = 62568830)

PR #945, opened 2016-05-06
Fix for an issue with split scope in dead store optimization

aneeshdk:
Storing the arg in for the parameter and the copy logic of initial value
of the param to the corresponding symbol in the body was happening in the
same block. During dead store check were were only checking the property
sym for usage. So the initial store was getting removed. This fix adds
usage check for the stack sym too.


aneeshdk:
@abchatra @LouisLaf @pleath Can you please take a look?

abchatra:
:+1:

PR #948, opened 2016-05-07
Make fixes in ChakraCore

digitalinfinity:
The last few commits in ChakraCore Linux branch have broken the NTBUILD build. This fixes those breaks.

### DIFF
--- a/lib/Runtime/Library/CompoundString.cpp
+++ b/lib/Runtime/Library/CompoundString.cpp
@@ -146,7 +146,9 @@ namespace Js
 
     #endif
 
-    CharCount CompoundString::Block::PointerLengthFromCharLength(const CharCount charLength)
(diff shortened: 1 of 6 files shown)


lib/Runtime/Library/CompoundString.cpp
```diff
@@ -146,6 +146,11 @@ namespace Js
 
     #endif
 
+    // ChakraDiag includes CompoundString.cpp as a header file so this method needs to be marked as inline 
+    // to handle that case
+#ifdef IsJsDiag
+    inline
```
curtisman: Use a macro instead?

digitalinfinity: Done

curtisman:
LGTM

PR #949, opened 2016-05-08
Experimental: enable x64 nojit configuation to build using clang on windows

curtisman:
Build x64 no jit configurations using clang on windows.
Disable a bunch of warning,
Fix issue with PCH builds.

Since clang on windows doesn't support /homeparam, any javascript that requires stack walking doesn't work (e.g. exception)

### DIFF
--- a/lib/Common/CommonPal.h
+++ b/lib/Common/CommonPal.h
@@ -4,6 +4,12 @@
 //-------------------------------------------------------------------------------------------------------
 #pragma once
 
+// Define _ALWAYSINLINE for template that want to always inline, but doesn't allow inline linkage in clang
+#if defined(__GNUC__) || defined(__clang__)
+#define _ALWAYSINLINE __attribute__((always_inline))
+#else
+#define _ALWAYSINLINE __forceinline
+#endif
 
 #ifdef _WIN32
 #pragma warning(push)
@@ -27,6 +33,13 @@ typedef wchar_t char16;
 #define INIT_PRIORITY(x)
 
 #define get_cpuid __cpuid
+
+#if defined(__clang__)
+__forceinline void  __int2c()
+{
+    __asm int 0x2c
+}
+#endif
 
 #else // !_WIN32
 
@@ -402,7 +415,7 @@ DWORD __cdecl CharUpperBuffW(const char16* lpsz, DWORD  cchLength);
 #endif
 
 
-#ifdef _MSC_VER
+#if defined(_MSC_VER) && !defined(__clang__)
 // ms-specific keywords
 #define _ABSTRACT abstract
 // MSVC2015 does not support C++11 semantics for `typename QualifiedName` declarations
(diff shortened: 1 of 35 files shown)


lib/Runtime/Library/JavascriptErrorDebug.cpp
```diff
@@ -235,10 +239,7 @@ namespace Js
         memset(proerrstr, 0, sizeof(*proerrstr));
         *pperrinfo = nullptr;
 
-#ifndef _WIN32
-        // xplat-todo: Find out if we can implement richer error info on Linux
-        return hr;
-#else
+#if defined(_MSC_VER) && !defined(__clang__)
```
obastemur: AFAIK '_MSC_VER' defined means '__clang__' is not defined. What option we are trying to rule out here?

lib/Common/Memory/HeapInfo.h
```diff
@@ -279,7 +279,7 @@ class HeapInfo
     {
         // xplat-todo: fix up vpm.64b.h generation to generate correctly
         // templatized code
-#ifdef _WIN32
+#if defined(_WIN32) && !defined(__clang__)
```
obastemur: Can we say '#ifdef '_MSC_VER' ' instead?

lib/Common/Exceptions/Throw.cpp
```diff
@@ -64,7 +64,7 @@ extern "C"{
 
 namespace Js {
 #if defined(GENERATE_DUMP) && defined(STACK_BACK_TRACE)
-    StackBackTrace * Throw::stackBackTrace = nullptr;
+    __declspec(thread) StackBackTrace * Throw::stackBackTrace = nullptr;
```
obastemur: AFAIK We can't use '__declspec' on non Windows.

lib/Common/Core/FaultInjection.cpp
```diff
@@ -349,7 +349,7 @@ namespace Js
     const auto& globalFlags = Js::Configuration::Global.flags;
     PVOID FaultInjection::vectoredExceptionHandler = nullptr;
     DWORD FaultInjection::exceptionFilterRemovalLastError = 0;
-    int(*Js::FaultInjection::pfnHandleAV)(int, PEXCEPTION_POINTERS) = nullptr;
+    __declspec(thread) int(*Js::FaultInjection::pfnHandleAV)(int, PEXCEPTION_POINTERS) = nullptr;
```
obastemur: #ifdef WIN32?

lib/Runtime/Library/JavascriptErrorDebug.cpp
digitalinfinity: Apparently clang defines msc_ver on Windows in VC compat mode. I guess Curtis is trying to genuinely check for VC

lib/Common/CommonPal.h
```diff
@@ -554,3 +567,61 @@ void TryFinally(const TryFunc& tryFunc, const FinallyFunc& finallyFunc)
 
     finallyFunc(hasException);
 }
+
+
+namespace PlatformAgnostic
+{
+    __forceinline unsigned char _BitTestAndSet(LONG *_BitBase, int _BitPos)
+    {
+#if defined(__clang__)
```
digitalinfinity: Probably should check for x86 compatible arch? Does clang on windows target arm right now?

lib/Common/CommonPal.h
```diff
@@ -554,3 +567,61 @@ void TryFinally(const TryFunc& tryFunc, const FinallyFunc& finallyFunc)
 
     finallyFunc(hasException);
 }
+
+
+namespace PlatformAgnostic
+{
+    __forceinline unsigned char _BitTestAndSet(LONG *_BitBase, int _BitPos)
```
digitalinfinity: Can we use int32 instead of LONG here?

lib/Common/Memory/RecyclerPageAllocator.cpp
```diff
@@ -197,7 +197,7 @@ size_t
 RecyclerPageAllocator::GetAllWriteWatchPageCount(DListBase<T> * segmentList)
 {
     size_t totalCount = 0;
-    DListBase<T>::Iterator i(segmentList);
+    _TYPENAME DListBase<T>::Iterator i(segmentList);
```
digitalinfinity: Curious, why not use just typename here?

lib/Common/Memory/SmallFinalizableHeapBucket.h
```diff
@@ -167,39 +167,65 @@ class SmallHeapBlockType<FinalizableWithBarrierBit, MediumAllocationBlockAttribu
...
-public:
     template <ObjectInfoBits objectAttributes>
-    typename SmallHeapBlockType<objectAttributes, TBlockAttributes>::BucketType& GetBucket()
+    class BucketGetter
     {
-        CompileAssert(objectAttributes & FinalizeBit);
-        return this->finalizableHeapBucket;
-    }
+    public:
+        static typename SmallHeapBlockType<objectAttributes, TBlockAttributes>::BucketType& GetBucket(HeapBucketGroup<TBlockAttributes> * heapBucketGroup)
```
digitalinfinity: Could this be slightly more readable if you typedef BucketType here- can we use a typedef defined in the template class in the specialized class?

digitalinfinity:
:+1:

lib/Common/CommonPal.h
```diff
@@ -4,6 +4,12 @@
 //-------------------------------------------------------------------------------------------------------
 #pragma once
 
+// Define _ALWAYSINLINE for template that want to always inline, but doesn't allow inline linkage in clang
+#if defined(__GNUC__) || defined(__clang__)
+#define _ALWAYSINLINE __attribute__((always_inline))
```
jianchun: Why don't we `#define __forceinline  __attribute__((always_inline))` and then use `__forceinline` cross-platform?

digitalinfinity: I suspect, its because, as Curtis' comment says, __forceinline implies inline linkage in some situations- __attribute__((always_inline) does not.

jianchun: So I can assume `__forceinline` is already supported by clang, and doesn't work for these cases? LGTM then.

lib/Runtime/Library/JavascriptErrorDebug.cpp
curtisman: What @digitalinfinity  said :)

lib/Common/CommonPal.h
curtisman: We need inline linkage for __forceinline when they are defined in the header.
But template can't have inline linkage in clang.  That's why we need a different define to only have the always_inline attribute.

lib/Common/Memory/HeapInfo.h
curtisman: Yeah, probably.   Will change

lib/Common/Exceptions/Throw.cpp
curtisman: Supported in clang under -fms-extensions mode.
http://clang.llvm.org/docs/AttributeReference.html#thread

lib/Common/Core/FaultInjection.cpp
curtisman: __declspec(thread) is supported in clang under -fms-extensions mode.
http://clang.llvm.org/docs/AttributeReference.html#thread

lib/Common/CommonPal.h
curtisman: This builds for x86 too.
But I have probably with x86 with clang in general because of existing inline asm.

lib/Common/Memory/RecyclerPageAllocator.cpp
curtisman: I am following the pattern in DList.h line 456 and 466.
@jianchun  It may be for vs2013 compiler compat?

lib/Common/CommonPal.h
curtisman: Call to the intrinsic below require this to be a LONG *

PR #951, opened 2016-05-09
cross platform: python script to run test files [suggestion]

obastemur:
This PR

- adds a Python based test script to run all the test files or the ones from
a given folder. In addition to executing JS files, it also compares the results
to predefined baseline files.
- replaces the functionality of `runtests.sh` to call `runtests.py`

Try:

```
./build.sh
./test/runtests.py Basics
```

In order to run a single file
```
./test/runtests.py Basics/hello.js
```

### DIFF
--- a/test/runtests.py
+++ b/test/runtests.py
@@ -1,0 +1,120 @@
+#!/usr/bin/env python
+#-------------------------------------------------------------------------------------------------------
+# Copyright (C) Microsoft. All rights reserved.
+# Licensed under the MIT license. See LICENSE.txt file in the project root for full license information.
+#-------------------------------------------------------------------------------------------------------
+
+import sys
+import os
+import subprocess as SP
+
+test_all = True
+test_root = os.path.dirname(os.path.realpath(__file__))
+
+# ugly trick
+ch_path = os.path.join(os.path.dirname(test_root), "BuildLinux/ch")
+
+if not os.path.isfile(ch_path):
+    print "BuildLinux/ch not found. Did you run ./build.sh already?"
+    sys.exit(1)
+
+if len(sys.argv) > 1:
+    if sys.argv[1] in ['-?', '--help']:
+        print "ChakraCore *nix Test Script\n"
+
+        print "Usage:"
+        print "test.py <optional test path>\n"
+
+        print "-?, --help    : Show help\n"
+
+        print "Samples:"
+        print "test only Array:"
+        print "\t./test.py Array\n"
+        
+        print "test a single file:"
+        print "\t./test.py Basics/hello.js\n"
+
+        print "test all folders:"
+        print "\t./test.py"
+        sys.exit(0)
+    test_all = None
+
+test_dirs=['']
+if test_all:
+    test_dirs = os.listdir(test_root)
+else:
+    test_dirs[0] = sys.argv[1]
+
+def show_failed(filename, output, exit_code, expected_output):
+    print "\nFailed ->", filename
+    if expected_output == None:
+        print "\nOutput:"
+        print "----------------------------"
+        print output
+        print "----------------------------"
+    else:
+        lst_output = output.split('\n')
+        lst_expected = expected_output.split('\n')
+        ln = min(len(lst_output), len(lst_expected))
+        for i in range(0, ln):
+            if lst_output[i] != lst_expected[i]:
+                print "Output: (at line " + str(i) + ")" 
+                print "----------------------------"
+                print lst_output[i]
+                print "----------------------------"
+                print "Expected Output:"
+                print "----------------------------"
+                print lst_expected[i]
+                print "----------------------------"
+                break
+
+    print "exit code:", exit_code
+    print "\nFailed!"
+    sys.exit(exit_code)
+
+def test_path(folder, is_file):
+    files=['']
+    if is_file == False:
+        print "Testing ->", os.path.basename(folder)
+        files = os.listdir(folder)
+    else:
+        files[0] = folder
+    
+    for js_file in files:
+        if is_file or os.path.splitext(js_file)[1] == '.js':
+            js_file = os.path.join(folder, js_file)
+            js_output = ""
+
+            if not os.path.isfile(js_file):
+                print "Javascript file doesn't exist (" + js_file + ")"
+                sys.exit(1)
+
+            p = SP.Popen([ch_path, js_file], stdout=SP.PIPE, stderr=SP.STDOUT, close_fds=True)
+            js_output = p.communicate()[0].replace('\r','')
+            exit_code = p.wait()
+
+            if exit_code != 0:
+                show_failed(js_file, js_output, exit_code, None)
+            else: #compare outputs
+                baseline = os.path.splitext(js_file)[0] + '.baseline'
+                baseline = os.path.join(folder, baseline)
+                if os.path.isfile(baseline):
+                    expected_output = None
+                    with open(baseline, 'r') as bs_file:
+                        expected_output = bs_file.read().replace('\r', '')
+                    # todo: compare line by line and use/implement wild cards support
+                    # todo: by default we discard line endings (xplat), make this optional
+                    if expected_output.replace('\n', '') != js_output.replace('\n', ''):
+                        show_failed(js_file, js_output, exit_code, expected_output)
+
+            if not is_file:
+                print "\tPassed ->", os.path.basename(js_file)
+
+is_file = len(test_dirs) == 1 and os.path.splitext(test_dirs[0])[1] == '.js'
+
+for folder in test_dirs:
+    full_path = os.path.join(test_root, folder)
+    if os.path.isdir(full_path) or is_file:
+        test_path(full_path, is_file)
+
+print 'Success!'
--- a/test/runtests.sh
+++ b/test/runtests.sh
@@ -3,13 +3,10 @@
 # Licensed under the MIT license. See LICENSE.txt file in the project root for full license information.
 #-------------------------------------------------------------------------------------------------------
 #
-# Eventual test running harness on *nix for ChakraCore
-# Today, it's simply there to make sure hello.js doesn't regress
-#
+# todo-CI: REMOVE THIS AFTER ENABLING runtests.py on CI
 
 test_path=`dirname "$0"`
 ch_path="$test_path/../BuildLinux/ch"
-hello_path="$test_path/Basics/hello.js"
 
 if [ ! -f $ch_path ]; then
     echo 'ch not found- exiting'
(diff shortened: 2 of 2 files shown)


obastemur:
@dilijev I guess Copyright check because of py file?

obastemur:
@dilijev I've updated netci.groovy to use runtests.py instead of runtests.sh (which is no longer exist) but somehow CI doesn't take this change into consideration?

dilijev:
@obastemur Yes, changes to the CI script need to be merged in to the branch and then the tasks need to be regenerated before the CI will use them. netci.groovy is not used during the CI checks, it is only used to generate them. (That's part of why changes to the CI are so difficult to test.)

Probably the fastest thing to do for now is to have runtests.sh wrap runtests.py.

dilijev:
@obastemur  as for the copyright failure, yes, you can see in the logs:

>test/runtests.py ... does not contain a correct Microsoft copyright notice.

.gitignore
```diff
@@ -42,6 +42,7 @@ install_manifest.txt
 *.xcworkspace
 
 # additional *nix generated files
+/ch
```
dilijev: Why `/ch` instead of just `ch` doesn't the leading slash only exclude the file if it is in the same directory as the `.gitignore`?

netci.groovy
```diff
@@ -125,7 +125,7 @@ def CreateLinuxBuildTasks = { machine, configTag, linuxBranch, nonDefaultTaskSet
             def infoScript = 'bash jenkins/get_system_info.sh'
             def debugFlag = buildType == 'debug' ? '--debug' : ''
             def buildScript = "bash ./build.sh -j=`nproc` ${debugFlag} --cxx=/usr/bin/clang++-3.8 --cc=/usr/bin/clang-3.8"
-            def testScript = "bash test/runtests.sh"
+            def testScript = "bash test/runtests.py Basics/hello.js"
```
dilijev: I think it is better to keep the `runtests.sh` script and add any tests you want to run to that script as we go (presumably it will grow?). It is difficult to test changes to the netci.groovy script as I noted in another comment. For tests which will change often, it is best to keep the information about which tests to run in another file, like `runtests.sh`. Additionally, this way we don't need the build definitions tied to the specific way we run tests and we are free to change them later.

obastemur:
> Probably the fastest thing to do for now is to have runtests.sh wrap runtests.py.

and send an empty config to delete `runtests.sh`? 

Do we have to do this? BTW; Where do we update that Copyright Check algorithm to cover py files ?

test/runtests.py
```diff
@@ -0,0 +1,120 @@
+#!/usr/bin/env python
```
dilijev: Yeah, I figured this would come up. Files that need shebangs will break the copyright check. This is arguably a bug in the copyright check that needs a special case. Previously I had specified that the copyright notice had to be the VERY FIRST thing in the file. But since semantically, the shebang needs to come first when it is present, I'll add an exception.
@digitalinfinity does this exception seem reasonable?

test/runtests.sh
```diff
@@ -1,28 +0,0 @@
-#-------------------------------------------------------------------------------------------------------
```
dilijev: I think you should keep this file, or add it back. Make sure it is set to executable or the CI will not be able to run it.

.gitignore
obastemur: That's my intention. new build script creates a symbolic link for ch on the root. I want it to be ignored on git.

dilijev:
@obastemur Why do you want to delete `runtests.sh`? It seems better to have a script which could run any test harness in any configuration and abstract the details of running the tests from the build definition?

Anyway changes to the CI script should be done as separate pull requests (and in a separate branch, as you've noted, to avoid disruption while we sort out the details).

.gitignore
dilijev: To be more general, we could ignore that file in any directory it happens to appear by simply removing the leading `/`

dilijev:
The copyright check files are:
https://github.com/Microsoft/ChakraCore/blob/master/jenkins/check_copyright.py
https://github.com/Microsoft/ChakraCore/blob/master/jenkins/check_copyright.sh

The fix to ignore a shebang line at the beginning of the file will be in
https://github.com/Microsoft/ChakraCore/blob/master/jenkins/check_copyright.py

obastemur:
> Why do you want to delete runtests.sh? It seems better to have a script which could run any test harness in any configuration and abstract the details of running the tests from the build definition?

Both are executable scripts, I don't see any reason to keep both. 

> Anyway changes to the CI script should be done as separate pull requests (and in a separate branch, as you've noted, to avoid disruption while we sort out the details).

Great. Gonna cherry-pick this to separate branch and let's see what's going to happen

build.sh
```diff
@@ -148,4 +148,12 @@ echo Generating $BUILD_TYPE makefiles
 cmake $CMAKE_GEN $CC_PREFIX $ICU_PATH -DCMAKE_BUILD_TYPE=$BUILD_TYPE ..
 
 $MAKE $MULTICORE_BUILD 2>&1 | tee build.log
+
+result=$?
+if [[ $result == 0 && ! -f ../ch ]]; then
+    ln -s BuildLinux/ch ../ch
```
jianchun: I assume you did this change to save some typing while testing manually? I tried it out and this broke my personal settings. I have "." at head of PATH. When I build at repo root and run "ch", it prints
```
dlopen() failed; dlerror says 'ch: cannot open shared object file: No such file or directory'
```

The same happens if I have `<repo>/BuildLinux` at head of PATH.

My workaround was a wrapper script "ch" in my PATH to invoke BuildLinux/ch. That doesn't have this problem.

So maybe a top-level symbolic link to "ch" isn't very useful?
- It has above problem. (maybe we can investigate and fix it.)
- The primary binary is "chakracore", not "ch". "ch" is mostly a testing binary. A top-level link feels over kill for it. 
- We might consider build type subdirs like BuildLinux/x64_debug, BuildLinux/x64_release. Top-level ch link doesn't work with that.

dilijev: I agree with the above concerns. It's fairly straightforward to know where the test binary is. Especially as long as this is intended to use with already-written test code, saving typing is a moot point. For manual testing, there's not that much extra typing to do, and a wrapper script might be more useful.

obastemur: >  I tried it out and this broke my personal settings. I have "." at head of PATH. When I build at repo root and run "ch", it prints

local one is available at `./ch`

> I agree with the above concerns. It's fairly straightforward to know where the test binary is.

you can call `ls -l ch`

> We might consider build type subdirs like BuildLinux/x64_debug, 

We could have `ch_d` or `ch_debug` etc..

> Especially as long as this is intended to use with already-written test code, saving typing is a moot point.

Sorry, didnt' understand. Could you please explain more?

jenkins/check_copyright.py
```diff
@@ -42,6 +42,8 @@ def report_incorrect(file_name, pairs):
 linecount = 0
 pairs = []
 with open(file_name, 'r') as sourcefile:
+    if not re.match(r'#!/usr/bin/env.*', sourcefile.readline(), 0):
+        sourcefile.seek(0)
```
dilijev: I want to be more general than this. I've got a patch ready.

obastemur: can you be more precise ?

dilijev: Actually I like your sourcefile.seek method. Can you change this to be more general as follows?

```python
pairs = []
with open(file_name, 'r') as sourcefile:
    shebang = sourcefile.readline()
    if not shebang.startswith("#!"):
        sourcefile.seek(0)
    pairs += zip(regexes, sourcefile)
```

obastemur: How this one is being more general? Would we also accept one line license texts on top of Microsoft copyright notice?

obastemur:
@jianchun  `ch` is removed from root for now. 
@dilijev see `env` is removed. Do you think that should cover?

Although I don't expect any problem from this PR, I'm still +1 on cherry-picking this PR to separate branch and run through CI first. However, I'm also okay if you don't see any reason for that too J

dilijev:
@obastemur I still strongly oppose modifying `netci.groovy` as part of this change, or removing `runtests.sh` at all, until the test harness for linux becomes more mature. We might change mechanisms and having a single script which will automate whatever we need will remove churn from the netci.groovy file, which is difficult to manage, as you've seen. I think removing that script and making a change here is jumping the gun.

jenkins/check_copyright.py
```diff
@@ -42,6 +42,8 @@ def report_incorrect(file_name, pairs):
 linecount = 0
 pairs = []
 with open(file_name, 'r') as sourcefile:
+    if not re.match(r'#!/usr/bin/env.*', sourcefile.readline(), 0):
```
dilijev: Why not just `sourcefile.readline().startswith("#!")`? I think we're imposing too much structure on the shebang.

jenkins/check_copyright.py
dilijev: Changes will still be code reviewed and we can catch that sort of thing. In fact, there's nothing stopping someone from adding a comment at the end of the first shebang line to do the same no matter what structure you impose here. You don't know for a fact where the binaries will live and making assumptions like that seems bad.

dilijev: Also since we're matching the beginning of the line, a regex is overkill.

test/runtests.sh
```diff
@@ -1,28 +1,8 @@
...
 # Copyright (C) Microsoft. All rights reserved.
```
dilijev: We require the first line of dashes as part of the copyright notice.

obastemur:
CI related file changes are removed from the PR. I guess that's all ?

test/runtests.sh
```diff
@@ -1,28 +1,8 @@
...
 # Licensed under the MIT license. See LICENSE.txt file in the project root for full license information.
 #-------------------------------------------------------------------------------------------------------
 #
-# Eventual test running harness on *nix for ChakraCore
-# Today, it's simply there to make sure hello.js doesn't regress
-#
+# REMOVE THIS AFTER MERGING runtests.py in!
```
dilijev: I don't think there's any need for this comment. We can make this decision later.

dilijev:
Unless you include an exception to not return a failure error code for now, the linux tasks will keep failing until clang 3.8 is installed on the build machines, which should happen later today. Since that change is incoming, it's probably best to wait until it is done and then re-run the tests and merge this in.

test/runtests.sh
obastemur: AFAIK the decision is done. Is there any `technical` reason to keep this file?

dilijev: The idea is that, since netci.groovy is much slower to update and the way we run the tests could change, I'm ultimately trying to reduce resistance to changes by keeping this file as a wrapper for whatever mechanism we actually use to run tests. I think that's a fair enough technical reason (based on a technical limitation of the `netci.groovy` task generation mechanism.

I'm not aware of any decision regarding `netci.groovy` task changes or this file in particular. I'll start an email thread with @curtisman to sort out the details.

dilijev:

dilijev:
Following completion of the Copyright Check, LGTM and this is good to merge.

dilijev:
Since the toolchain has been updated...
@dotnet-bot 

PR #952, opened 2016-05-09
My cleanup of the shortcut-property-access-on-unknown-property-name optimization relied on IsExternal() to tell us whether an object is eligible. But we must also exclude HostDispatch, which returns false on IsExternal().

pleath:


pleath:
@Yongqu

PR #958, opened 2016-05-09
Fixing bug with shared bailouts

rajatd:
Shared bailouts for InlineMathFloor/Ceil/Round were bypassing the code on the bailout path that sets up data specific to a bailout (bailout kind, polymorphic inline cache index etc.) on the bailout record and were jumping directly to the bailout label. Fixed that.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/958)

### DIFF
--- a/lib/Backend/Lower.cpp
+++ b/lib/Backend/Lower.cpp
@@ -12157,11 +12157,10 @@ Lowerer::GenerateObjectTestAndTypeLoad(IR::Instr *instrLdSt, IR::RegOpnd *opndBa
 }
 
 IR::LabelInstr *
-Lowerer::GenerateBailOut(IR::Instr * instr, IR::BranchInstr * branchInstr, IR::LabelInstr *bailOutLabel)
+Lowerer::GenerateBailOut(IR::Instr * instr, IR::BranchInstr * branchInstr, IR::LabelInstr *bailOutLabel, IR::LabelInstr * collectRuntimeStatsLabel)
 {
     BailOutInfo * bailOutInfo = instr->GetBailOutInfo();
     IR::Instr * bailOutInstr = bailOutInfo->bailOutInstr;
-    IR::LabelInstr *collectRuntimeStatsLabel = nullptr;
     if (instr->IsCloned())
     {
         Assert(bailOutInstr != instr);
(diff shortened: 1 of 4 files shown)


rajatd:
@pleath @MikeHolman @LouisLaf

lib/Backend/Lower.cpp
```diff
@@ -12174,14 +12173,21 @@ Lowerer::GenerateBailOut(IR::Instr * instr, IR::BranchInstr * branchInstr, IR::L
         return bailOutLabel;
     }
 
+    // Add helper label to trigger layout.
+    if (collectRuntimeStatsLabel)
+    {
+        collectRuntimeStatsLabel->Unlink();
```
pleath: I wonder if we're asking for trouble unlinking a label that the caller already put in the IR stream (presumably for reasons of its own). Would it work to assert that the label is already unlinked? Is there a caller that requires unlinking now? If so, I wonder if the caller shouldn't unlink it itself.

rajatd: Hmm.. The lowering functions for InlineMathFloor et.al. require unlinking now. I could, however, not insert the label in those routines and just pass it to GenerateBailout as collectRuntimeStatsLabel and insert it before the current instruction there.

pleath: That sounds better to me. The logic in the caller shouldn’t require/expect the label to remain where it was put in the IR stream.

--Paul

lib/Backend/LowerMDShared.cpp
```diff
@@ -8851,7 +8851,10 @@ void LowererMD::GenerateFastInlineBuiltInCall(IR::Instr* instr, IR::JnHelperMeth
                 sharedBailout = (instr->GetBailOutInfo()->bailOutInstr != instr) ? true : false;
                 if (sharedBailout)
                 {
-                    bailoutLabel = instr->GetBailOutInfo()->bailOutInstr->AsLabelInstr();
+                    // In case of a shared bailout, we should jump to the code that sets some data on the
```
ThomsonTan: I only see an extra label created here, what is "the code that sets some data on the bailout record"?

rajatd: That is in Lowerer::GenerateBailout: when bailoutInstr != instr

pleath:
LGTM.

rajatd:
Thanks, merging.

PR #960, opened 2016-05-10
Silence Clang 3.9 warnings

obastemur:
warning: destructor called on ******** that is abstract but has non-virtual destructor [-Wdelete-non-virtual-dtor]
    obj->~T();

and couple of other type related / unused code warnings.

See @jianchun’s review (thanks again!)
https://github.com/obastemur/ChakraCore/commit/c4084f642c0bc658bf349242c7618311fdf525b0 

@curtisman  @jianchun @digitalinfinity any concerns with this targeting
master branch instead of linux?

### DIFF
--- a/lib/Runtime/Base/CharStringCache.cpp
+++ b/lib/Runtime/Base/CharStringCache.cpp
@@ -18,7 +18,7 @@ namespace Js
     {
         AssertMsg(JavascriptString::IsASCII7BitChar(c), "GetStringForCharA must be called with ASCII 7bit chars only");
 
-        PropertyString * str = charStringCacheA[c];
+        PropertyString * str = charStringCacheA[(int)c];
         if (str == nullptr)
         {
             PropertyRecord const * propertyRecord;
@@ -26,7 +26,7 @@ namespace Js
             JavascriptLibrary * javascriptLibrary = JavascriptLibrary::FromCharStringCache(this);
             javascriptLibrary->GetScriptContext()->GetOrAddPropertyRecord(&wc, 1, &propertyRecord);
             str = javascriptLibrary->CreatePropertyString(propertyRecord);
-            charStringCacheA[c] = str;
+            charStringCacheA[(int)c] = str;
         }
 
         return str;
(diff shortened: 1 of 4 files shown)


dilijev:
@obastemur seems fine. I don't have any issue with this going into master, but I wonder why master instead of linux since this seems more directly related to the linux port?

lib/Runtime/Base/CharStringCache.cpp
```diff
@@ -18,15 +18,15 @@ namespace Js
     {
         AssertMsg(JavascriptString::IsASCII7BitChar(c), "GetStringForCharA must be called with ASCII 7bit chars only");
 
-        PropertyString * str = charStringCacheA[c];
+        PropertyString * str = charStringCacheA[(int)c];
```
dilijev: I don't like casting the index inside of the array index operator. I think, especially since you use this twice, you should assign the result of the cast to a new index variable and index with that variable instead.

lib/Runtime/DetachedStateBase.h
```diff
@@ -20,6 +20,10 @@ namespace Js
         {
         }
 
+        virtual ~DetachedStateBase()
+        {
+        }
+
```
dilijev: Since this whole method could be a single line, I think it would be better to have the braces on the same line, like you did with `~InterruptPoller`

lib/Runtime/Base/CharStringCache.cpp
obastemur: What would be the purpose of doing that? Styling?

lib/Runtime/DetachedStateBase.h
obastemur: I have matched the style with constructor on both places.

obastemur:
> @obastemur seems fine. I don't have any issue with this going into master, but I wonder why master instead of linux since this seems more directly related to the linux port?

This PR is to cleanup platform independent clang warnings.

dilijev:
Ok, sounds fine. LGTM.

Re: style comments - you'll get a lot of those from a few people on the team in code reviews. It's fine to take under advisement and you can decide to take no action at your discretion, if it's clearly only a style-related concern :)

obastemur:
@jianchun @dilijev thanks for the reviews

Issue #535, opened 2016-03-11
Proxy traps not triggered for __lookupGetter and  __lookupSetter_


abchatra:
Expected true from both the following test cases. Actual: false
This is a kangax test case

```js
print((function()
{
// Object.prototype.__lookupGetter__ -> [[GetOwnProperty]]
// Object.prototype.__lookupGetter__ -> [[GetPrototypeOf]]
var gopd = [];
var gpo = false;
var p = new Proxy({},
  { getPrototypeOf: function(o) { gpo = true; return Object.getPrototypeOf(o); }},
  { getOwnPropertyDescriptor: function(o, v) { gopd.push(v); return Object.getOwnPropertyDescriptor(o, v); }});
Object.prototype.__lookupGetter__.call(p, "foo");
return gopd + '' === "foo" && gpo;
})());
```
```js
print((function()
{
// Object.prototype.__lookupSetter__ -> [[GetOwnProperty]]
// Object.prototype.__lookupSetter__ -> [[GetPrototypeOf]]
var gopd = [];
var gpo = false;
var p = new Proxy({},
  { getPrototypeOf: function(o) { gpo = true; return Object.getPrototypeOf(o); }},
  { getOwnPropertyDescriptor: function(o, v) { gopd.push(v); return Object.getOwnPropertyDescriptor(o, v); }});
Object.prototype.__lookupSetter__.call(p, "foo");
return gopd + '' === "foo" && gpo;
})());
```

leirocks:
looks there's problem in the repro code?
it's new-ing  Proxy with 3 parameters, not two:
1. {},
2.  { getPrototypeOf: function(o) { gpo = true; return Object.getPrototypeOf(o); }},
3.  { getOwnPropertyDescriptor: function(o, v) { gopd.push(v); return Object.getOwnPropertyDescriptor(o, v); }}

the code should be like this:?

`
print((function()
{
// Object.prototype.__lookupGetter__ -> [[GetOwnProperty]]
// Object.prototype.__lookupGetter__ -> [[GetPrototypeOf]]
var gopd = [];
var gpo = false;
var p = new Proxy({},
  { 
      getPrototypeOf: function(o) { gpo = true; return Object.getPrototypeOf(o); },
      getOwnPropertyDescriptor: function(o, v) { gopd.push(v); return Object.getOwnPropertyDescriptor(o, v); 
});
Object.prototype.__lookupGetter__.call(p, "foo");
[...]

PR #961, opened 2016-05-10
#535 - Proxy traps not triggered for __lookupGetter and __lookupSetter_

leirocks:
just need to call GetPrototype instead of GetPrototypeNoTrap in JavascriptOperators::GetAccessors

### DIFF
--- a/test/es7/lookupgettersetter.js
+++ b/test/es7/lookupgettersetter.js
@@ -1,0 +1,49 @@
+//-------------------------------------------------------------------------------------------------------
+// Copyright (C) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE.txt file in the project root for full license information.
+//-------------------------------------------------------------------------------------------------------
+
+WScript.LoadScriptFile("..\\UnitTestFramework\\UnitTestFramework.js");
+
+var tests = [
+    {
+        name: "Object.prototype.__lookupGetter__ -> [[GetOwnProperty]], [[GetPrototypeOf]]",
+        body: function () {
+            assert.isTrue((function()
+            {
+                // Object.prototype.__lookupGetter__ -> [[GetOwnProperty]]
+                // Object.prototype.__lookupGetter__ -> [[GetPrototypeOf]]
+                var gopd = [];
+                var gpo = false;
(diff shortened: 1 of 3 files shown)


leirocks:
@Yongqu can you review?

test/es7/lookupgettersetter.js
```diff
@@ -0,0 +1,49 @@
...
+                var gopd = [];
+                var gpo = false;
+                var p = new Proxy({}, 
+                {
+                    getPrototypeOf: function(o) { gpo = true; return Object.getPrototypeOf(o); },
+                    getOwnPropertyDescriptor: function(o, v) { gopd.push(v); return Object.getOwnPropertyDescriptor(o, v); }
+                });
+                Object.prototype.__lookupGetter__.call(p, "foo");
+                return gopd + '' === "foo" && gpo;
+            })());
```
ianwjhalliday: This is poor use of asserts and `UnitTestFramework.js`.  Instead of wrapping the test in a function and returning true/false, please `assert.areEqual` and `assert.isTrue` the results that are expected.

E.g. should look something like this, and the same goes for the other test below too:

```javascript
{
    name: "...",
    body: function () {
        // Object.prototype.__lookupGetter__ -> [[GetOwnProperty]]
        // Object.prototype.__lookupGetter__ -> [[GetPrototypeOf]]
        var gopd = [];
        var gpo = false;
        var p = new Proxy({},
        {
            getPrototypeOf: function(o) { gpo = true; return Object.getPrototypeOf(o); },
            getOwnPropertyDescriptor: function(o, v) { gopd.push(v); return Object.getOwnPropertyDescriptor(o, v); }
        });
        Object.prototype.__lookupGetter__.call(p, "foo");

        assert.areEqual(1, gopd.length, "getOwnPropertyDescriptor should only be called once");
        assert.areEqual("foo", gopd[0], "description of what is being asserted equal here");
        assert.isTrue(gpo, "description of what is expected to be true");
    }
}
```

leirocks: Fixed. thanks!

ianwjhalliday:
:+1:

Yongqu:
:shipit:

PR #968, opened 2016-05-11
xplat: fix direct entryPoint calling convention

jianchun:
For cross-plat we required custom calling convention on JavascriptMethods.
This change replaces all direct entryPoint calls with macros. On non-Win32,
the macros expand to call with extra parameters to meet the custom calling
convention.


digitalinfinity:
:shipit:

PR #970, opened 2016-05-11
Reenable and fix warning 4457 - declaration of '' hides function parameter

curtisman:


lib/Runtime/Base/EtwTrace.cpp
```diff
@@ -387,8 +387,7 @@ void EtwTrace::LogLoopBodyLoadEvent(FunctionBody* body, LoopHeader* loopHeader,
         {
             methodInfo.method_id = iJIT_GetNewMethodID();
             size_t len = utf8::EncodeInto(utf8MethodName, methodName, (charcount_t)methodLength);
-            uint loopNumber = body->GetLoopNumber(loopHeader) + 1;
```
dilijev: If this `loopNumber` is the same as the parameter and that data is contained in `loopHeader`, should we remove that parameter and extract that value from `loopHeader`? Otherwise, is there an assertion that should be made about whether the parameter `loopNumber` matches the value in `body->GetLoopNumber(loopHeader)`?

lib/Runtime/ByteCode/ByteCodeEmitter.cpp
```diff
@@ -7264,9 +7264,8 @@ void EmitCallTarget(
         }
 
         Emit(pnodeTarget->sxBin.pnode1, byteCodeGenerator, funcInfo, false);
-        Js::PropertyId propertyId = pnodeTarget->sxBin.pnode2->sxPid.PropertyIdFromNameNode();
-        Js::RegSlot callObjLocation = pnodeTarget->sxBin.pnode1->location;
-        Js::RegSlot protoLocation = callObjLocation;
```
dilijev: Nice. I wonder how this ended up getting into the code in the first place?

lib/Runtime/Library/RegexHelper.cpp
```diff
@@ -966,7 +966,7 @@ namespace Js
 
         BEGIN_TEMP_ALLOCATOR(tempAlloc, scriptContext, _u("RegexHelper"))
         {
-            results->Map([&](int i, RecyclableObject* resultObj) {
+            results->Map([&](int resultIndex, RecyclableObject* resultObj) {
```
dilijev: Is this lambda parameter used at all, here?

dilijev:
LGTM :shipit:

lib/Runtime/Library/RegexHelper.cpp
curtisman: Nope.  But I thought it would be nice to leave it there for debugging.

lib/Runtime/ByteCode/ByteCodeEmitter.cpp
curtisman: Dunno.

lib/Runtime/Base/EtwTrace.cpp
curtisman: Added assert

PR #973, opened 2016-05-11
MSFT: 7429352 JSRT context leak regression fix

Yongqu:
We start supporting cross-context scenarios in JSRT in chakracore. However we are having
leaks in different scenarios due to pinned object. This change cleans up some basic leak scenarios.
We might need to have more cleanup in the future to make sure that there is only one
root for a JSRTContext or script context.

One major leak source is the guest arena in ScriptContext. objects allocated in the guestArena are
hard pinned till the scriptcontext is closed. We need to move objects off the guestArena:
    scriptContext::Cache. it's allocated using normal RecyclerNew, linked in the JavascriptLibrary, and
keep the reference in scriptcontext to minimize the code change.
    BuiltInFunctionMap. We might need to have a full cleanup of BindReference, but for now we just move
out the builtInFunctionMap to JavascripLibrary.
    Convert ScriptContext::BindReference to use linklist in JavascriptLibrary
    Allocate JSRTContext using RecyclerNew (no more leaf); have JSRTContext have a different reference to
JavascriptLibrary to keep it alive, and masked out other fields for potential false reference.

### DIFF
--- a/lib/Parser/RegexPattern.cpp
+++ b/lib/Parser/RegexPattern.cpp
@@ -41,7 +41,7 @@ namespace UnifiedRegex
             Assert(
                 !GetScriptContext()->GetDynamicRegexMap()->TryGetValue(
                     RegexKey(source.GetBuffer(), source.GetLength(), GetFlags()),
-                    &p) ||
+                    &p) || ( source.GetLength() == 0 ) ||
                 p != this);
         }
 #endif
--- a/lib/Jsrt/Core/JsrtContextCore.cpp
+++ b/lib/Jsrt/Core/JsrtContextCore.cpp
@@ -33,22 +33,22 @@ JsrtContextCore::JsrtContextCore(JsrtRuntime * runtime) :
 /* static */
 JsrtContextCore *JsrtContextCore::New(JsrtRuntime * runtime)
 {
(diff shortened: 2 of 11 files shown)


Yongqu:
@digitalinfinity @kunalspathak can you take a look?

lib/Parser/RegexPattern.cpp
```diff
@@ -41,7 +41,7 @@ namespace UnifiedRegex
             Assert(
                 !GetScriptContext()->GetDynamicRegexMap()->TryGetValue(
                     RegexKey(source.GetBuffer(), source.GetLength(), GetFlags()),
-                    &p) ||
+                    &p) || ( source.GetLength() == 0 ) ||
```
kunalspathak: what is this about?

Yongqu: We have a builtin regexp stored in the map that are not cleaned up by default. It's ok for it to be there.

Yongqu:
@dotnet-bot please test

Yongqu:
@dotnet-bot test this

Yongqu:
@dotnet-bot test this please

digitalinfinity:
:shipit:

PR #978, opened 2016-05-12
Clear singletoneinstance on activation object clone

abchatra:
Recreating PR as the old one failed to merge


abchatra:
Targeted fix for 7225606

PR #979, opened 2016-05-12
FreeLoopBodyJobs race condition.

LouisLaf:
FreeLoopBodyJobs also need a processed flag to avoid race conditions.
To free data associated with a JIT loopbody, we create a FreeLoopBodyJob.  If we can't allocate one of these (OOM), we create one on the stack, and wait for it to be processed.
There was a race condition if we processed the job before starting to wait for it...  The 'processed' flag avoids this.


LouisLaf:
@abchatra and @rajatd, please review.

rajatd:
:shipit:

abchatra:
:shipit:

PR #981, opened 2016-05-13
Changes for MicroBuild v2.

dilijev:
* Don't sign PGI builds to save time in the signing step.
* Disable a broken test and reference internal bug number.
* Only take dependency on MicroBuild v2 plugins when building in MicroBuild v2. This allows us to remove from the repo the plugin metadata files that MicroBuild v2 will download automatically, while simultaneously unblocking legacy builds which would have depended on those files.


dilijev:
@curtisman @digitalinfinity @Cellule code review please?

bin/ChakraCore/ChakraCore.vcxproj
```diff
@@ -148,7 +148,7 @@
     <ClInclude Include="TestHooks.h" />
   </ItemGroup>
   <!-- Conditionally include the group of files to sign because MBv2 knows what to do with this but the old build definition gets confused. -->
-  <ItemGroup Condition="'$(VSO_MICROBUILD_V2)'=='True'">
+  <ItemGroup Condition="'$(VSO_MICROBUILD_V2)'=='True' AND '$(POGO_TYPE)'!='PGI'">
```
Cellule: Any specific reason not to sign PGI? Is the signing process costly ?

I know we don't need it now, but we might if we plan to do pogo on arm.

test/es6/rlexe.xml
```diff
@@ -95,13 +95,15 @@
       <tags>BugFix</tags>
     </default>
   </test>
+  <!-- Disabling for now because the test is flaky and fails in some situations
```
Cellule: Is there a bug filed for this? 
It seems out of scope of this PR, although I understand you wanted all tests to pass to test the build process properly.

Build/scripts/compose_build.ps1
```diff
@@ -14,15 +14,6 @@ param (
 )
 
 #
-# Clean up the sentinel which previously marked this build as incomplete.
-#
-
-$buildIncompleteFile = Join-Path $rootPath "build.incomplete"
```
Cellule: This is legacy right ? We don't have a `build.incomplete` file anymore?

bin/ChakraCore/ChakraCore.vcxproj
dilijev: Real-signing takes 3-5 minutes per binary and (I guess) depending on load on the signing server. We never release the PGI binaries and they can be used to produce PGO counts even unsigned so there's no reason to sign them. Right now we sign ChakraCore.dll and ch.exe twice. This cuts out a fair amount of time when the _pogo build definition runs.

Also, I think they would prefer we don't unnecessarily Real-sign anything we're not planning on actually releasing.

test/es6/rlexe.xml
dilijev: Uh yeah, actually. Thanks for reminding me to go look it up. We made this change in another version of the source a while back when the issue initially came up. There's a bug tracking it. Also it's definitely a compile bug and not just test flakiness.

bin/ChakraCore/ChakraCore.vcxproj
Cellule: Got it, fair enough

Build/scripts/compose_build.ps1
dilijev: Right. `build.incomplete` didn't provide any new information but there's sometimes contention for locks on the file from the different parallel builds, so having it is a net loss. It's also another moving piece to deal with for the rebuild scenario.

Build/scripts/init_build.ps1
```diff
@@ -76,39 +76,35 @@ if ($YearAndMonth) {
...
+New-Item -ItemType Directory -Force -Path (Join-Path $BinariesDirectory "logs")
 
-$buildIncompleteFile = Join-Path $FullOutputPath "build.incomplete"
-$flavorBuildIncompleteFile = Join-Path $FullOutputPath "${FlavorName}.incomplete"
+$flavorBuildIncompleteFile = Join-Path $DropPath "${BuildType}.incomplete"
 
-if (-not (Test-Path $buildIncompleteFile)) {
-    ($buildIncompleteFileContentsString -f "Build") `
-        | Out-File $buildIncompleteFile -Encoding Ascii
-}
```
dilijev: @Cellule I also deleted the code here which created the file in the first place.

test/es6/rlexe.xml
dilijev: Fixed.

Cellule:
:+1:

tcare:
Looks good

Build/MicroBuildv2.Build.ProjectConfiguration.props
```diff
@@ -1,12 +1,16 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Project DefaultTargets="Build" ToolsVersion="12.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
-  <Import Project="$(SolutionDir)packages\MicroBuild.Core.0.2.0\build\MicroBuild.Core.props" Condition="Exists('$(SolutionDir)packages\MicroBuild.Core.0.2.0\build\MicroBuild.Core.props')" />
+  <Import Project="$(SolutionDir)packages\MicroBuild.Core.0.2.0\build\MicroBuild.Core.props"
```
digitalinfinity: Can we group the imports in this file under ImportGroup guarded by the new condition we added? The other conditions need to be still in their individual imports i guess. Btw, should the Microbuild package name be a declared variable so that if the version/name changes we just have one place to update?

Build/scripts/init_build.ps1
```diff
@@ -46,7 +46,7 @@ $remote = $remote.replace("mshttps", "https")
 # Get the pushId and push date time to use that for build number and build date time
 $uri = ("{0}/commits/{1}?api-version=1.0" -f $remote, $commitHash)
```
digitalinfinity: should this be $CommitHashs? Also, i dont like that this file has variables with inconsistent casing but i guess its fine for now- would probably be better to just have camelCased variables

digitalinfinity:
>     $YearAndMonth = Get-Date $BuildPushDate -Format yyMM

Why not just have YearAndMonth set to "\" + Get-Date blah instead of having a seperate YearAndMonthSeperator variable?

---
Refers to: Build/scripts/init_build.ps1:70 in 5b5fb1c. [](commit_id = 5b5fb1c48d41c3d476dbae778a71787da19e4eb2, deletion_comment = False)

Build/scripts/init_build.ps1
```diff
@@ -76,39 +76,35 @@ if ($YearAndMonth) {
...
 if (-not (Test-Path $flavorBuildIncompleteFile)) {
-    ($buildIncompleteFileContentsString -f "Build of ${FlavorName}") `
+    ($buildIncompleteFileContentsString -f "Build of ${BuildType}") `
         | Out-File $flavorBuildIncompleteFile -Encoding Ascii
 }
 
 $PogoConfig = "False"
-if (((${Env:BuildPlatform} -eq "x64") -or (${Env:BuildPlatform} -eq "x86")) `
-    -and (${Env:BuildConfiguration} -eq "release"))
+if ((($Env:BuildPlatform -eq "x64") -or ($Env:BuildPlatform -eq "x86")) `
```
digitalinfinity: Consider just having an array of pogo supported architectures, and then just checking whether "{Env:BuildPlatform}{Env:BuildConfiguration}" is in the hash table

Build/scripts/post_build.ps1
```diff
@@ -34,6 +34,8 @@ param (
     [string[]]$pogo = @(),
     [string]$pogoscript = "",
 
+    [string]$buildTypeSuffix = "",
```
digitalinfinity: Can you add a comment here about what a buildType suffix is? (I assume pogo?)

digitalinfinity:
:shipit:

Build/scripts/post_build.ps1
dilijev: Yeah, to support build configs where the output folder gets the _pogo suffix and the script needs to know about that directory specifically.

Build/scripts/init_build.ps1
dilijev: powershell's variables are case-insensitive. Unfortunately that makes it possible to be inconsistent. I don't want to get into a huge code cleanup at this time but there's another round of changes coming after this and that might be a good time.

---
In reply to: [63264018](https://github.com/Microsoft/ChakraCore/pull/981#discussion_r63264018) [](ancestors = 63264018)

Build/scripts/init_build.ps1
dilijev: Excellent idea. I've decided to make those arrays into script parameters as well so that we can customize them from the build definition if need be.

---
In reply to: [63264916](https://github.com/Microsoft/ChakraCore/pull/981#discussion_r63264916) [](ancestors = 63264916)

dilijev:
>     $YearAndMonth = Get-Date $BuildPushDate -Format yyMM

Fair point. Made the change.

---
In reply to: [219183483](https://github.com/Microsoft/ChakraCore/pull/981#issuecomment-219183483) [](ancestors = 219183483)

---
Refers to: Build/scripts/init_build.ps1:70 in 8515958. [](commit_id = 85159583442db336d3f90cc245e40a0c4da04a69, deletion_comment = False)

dilijev:
Thanks @curtisman @digitalinfinity @Cellule @tcare for your reviews. This is now checked in! :)

PR #969, opened 2016-05-11
linux-ci branch: Remove exception for missing clang 3.8 toolchain.

dilijev:

PR #982, opened 2016-05-13
Remove exception for missing clang 3.8 toolchain; update CI config.

dilijev:
Tested these changes in the linux-ci branch and no breaks resulted.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/982)

### DIFF
--- a/netci.groovy
+++ b/netci.groovy
@@ -29,7 +29,7 @@ def dailyRegex = 'dailies'
 // Only generate PR check triggers for the version of netci.groovy in the master branch
 // since those PR checks will apply for all branches.
 def jobTypesToGenerate = [false]
-if (branch == 'master') {
+if (branch.startsWith('master')) {
     // OK to generate PR checks (this ensures we only generate one set of them)
     jobTypesToGenerate += true
 }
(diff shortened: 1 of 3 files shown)


dilijev:
FYI @digitalinfinity @obastemur @jianchun 

See #969 #966

netci.groovy
```diff
@@ -122,6 +122,8 @@ def CreateLinuxBuildTasks = { machine, configTag, linuxBranch, nonDefaultTaskSet
             // params: Project, BaseTaskName, IsPullRequest (appends '_prtest')
             def jobName = Utilities.getFullJobName(project, config, isPR)
 
+            def testableConfig = buildType in ['debug']
+
             def infoScript = 'bash jenkins/get_system_info.sh'
             def debugFlag = buildType == 'debug' ? '--debug' : ''
```
obastemur: Do you think that the `buildType`, `debugFlag` and the `testableConfig` above could be arranged/defined a bit more efficiently?

obastemur:
LGTM :+1:

netci.groovy
dilijev: I followed the pattern used when there were more options. This works, but I'm open to suggestions.

dilijev: Specifically, @digitalinfinity indicated that there are plans to add a `test` config like we have on Windows, where we will change line 125 to be: `def testableConfig = buildType in ['debug', 'test']`

digitalinfinity:
:shipit:

PR #983, opened 2016-05-13
xplat: enhance test/runtests.py to use rlexe.xml

jianchun:
Use `argparse` module to handle command line arguments. (Now this also
works on Windows -- needs **python 2.7**.)

Read `rlexe.xml` to load test metadata and honor `compile-flags`.

Run all loaded tests and print summary at the end.

Note this is only one step towards current rl test harness. There are lots
of test harness features not implemented yet.

### DIFF
--- a/test/runtests.py
+++ b/test/runtests.py
@@ -7,43 +7,64 @@
 import sys
 import os
 import subprocess as SP
+import argparse, textwrap
+import xml.etree.ElementTree as ET
 
-test_all = True
+parser = argparse.ArgumentParser(
+    description='ChakraCore *nix Test Script',
+    formatter_class=argparse.RawDescriptionHelpFormatter,
+    epilog=textwrap.dedent('''\
+        Samples:
+
+        test all folders:
+            {0}
+
+        test only Array:
+            {0} Array
+
+        test a single file:
+            {0} Basics/hello.js
+    '''.format(sys.argv[0]))
+    )
+parser.add_argument('folders', metavar='folder', nargs='*',
+                    help='folder subset to run tests')
+parser.add_argument('-b', '--binary', metavar='binary', help='ch full path');
+parser.add_argument('-d', '--debug', action='store_true',
+                    help='use debug build');
+parser.add_argument('-t', '--test', action='store_true', help='use test build');
+parser.add_argument('--x86', action='store_true', help='use x86 build');
+parser.add_argument('--x64', action='store_true', help='use x64 build');
+args = parser.parse_args()
+
+
 test_root = os.path.dirname(os.path.realpath(__file__))
+repo_root = os.path.dirname(test_root)
 
-# ugly trick
-ch_path = os.path.join(os.path.dirname(test_root), "BuildLinux/ch")
+# arch: x86, x64
+arch = 'x86' if args.x86 else ('x64' if args.x64 else None)
+if arch == None:
+    arch = os.environ.get('_BuildArch', 'x86')
 
-if not os.path.isfile(ch_path):
-    print "BuildLinux/ch not found. Did you run ./build.sh already?"
+# flavor: debug, test, release
+flavor = 'debug' if args.debug else ('test' if args.test else None)
+if flavor == None:
+    flavor = {'chk':'debug', 'test':'test', 'fre':'release'}\
+                    [os.environ.get('_BuildType', 'fre')]
+
+# binary: full ch path
+binary = args.binary
+if binary == None:
+    if sys.platform == 'win32':
+        binary = os.path.join(repo_root,
+            'Build/VcBuild/bin/{}_{}/ch.exe'.format(arch, flavor))
+    else:
+        binary = os.path.join(repo_root, "BuildLinux/ch")
+if not os.path.isfile(binary):
+    print '{} not found. Did you run ./build.sh already?'.format(binary)
     sys.exit(1)
 
-if len(sys.argv) > 1:
-    if sys.argv[1] in ['-?', '--help']:
-        print "ChakraCore *nix Test Script\n"
-
-        print "Usage:"
-        print "test.py <optional test path>\n"
-
-        print "-?, --help    : Show help\n"
-
-        print "Samples:"
-        print "test only Array:"
-        print "\t./test.py Array\n"
-        
-        print "test a single file:"
-        print "\t./test.py Basics/hello.js\n"
-
-        print "test all folders:"
-        print "\t./test.py"
-        sys.exit(0)
-    test_all = None
-
-test_dirs=['']
-if test_all:
-    test_dirs = os.listdir(test_root)
-else:
-    test_dirs[0] = sys.argv[1]
+pass_count = 0
+fail_count = 0
 
 def show_failed(filename, output, exit_code, expected_output):
     print "\nFailed ->", filename
@@ -58,7 +79,7 @@ def show_failed(filename, output, exit_code, expected_output):
         ln = min(len(lst_output), len(lst_expected))
         for i in range(0, ln):
             if lst_output[i] != lst_expected[i]:
-                print "Output: (at line " + str(i) + ")" 
+                print "Output: (at line " + str(i) + ")"
                 print "----------------------------"
                 print lst_output[i]
                 print "----------------------------"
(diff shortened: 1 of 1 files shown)


test/runtests.py
```diff
@@ -69,52 +90,88 @@ def show_failed(filename, output, exit_code, expected_output):
...
+
+    print "Testing ->", os.path.basename(folder)
+    for test in tests:
+        test_one(folder, test)
+
+def test_one(folder, test):
+    js_file = os.path.join(folder, test['files'])
+    js_output = ""
+
+    cmd = [x for x in [binary,
```
digitalinfinity: Not too familiar with this syntax- can you add a comment on whats going on here?

test/runtests.py
```diff
@@ -69,52 +90,88 @@ def show_failed(filename, output, exit_code, expected_output):
...
+            '-WERExceptionSupport',
+            '-ExtendedErrorStackForTestHost',
+            test.get('compile-flags'),
+            js_file]  if x != None]
+    p = SP.Popen(cmd, stdout=SP.PIPE, stderr=SP.STDOUT)
+    js_output = p.communicate()[0].replace('\r','')
```
digitalinfinity: I think there are tests which need to be compared in binary mode- @tcare will have details but there are parser tests that rely on the line ending being a particular way

digitalinfinity:
LGTM- couple comments and couple meta comments:
- Could do with more comments throughout the script in general
- Might be worth adding exclude-tags support up-front so that we can exclude tests that require the JIT.

test/runtests.py
dilijev: Standard list comprehension. The outer brackets define a list comprehension scope, x is the loop body and the loop control statement follows it on a line. It's supposed to read like natural language. The loop puts an element into the list for every value of x that passes the test. The inner brackets are just a list of values.

dilijev: I think you could rewrite this logically as:

```python
cmd = []
for x in [ ... ]:
    if x != None:
        cmd.append(x)
```

It's not too bad written that way, but the long list explicitly in the for loop would be difficult to read, and it makes sense / is more pythonic to do the list comprehension.

test/runtests.py
```diff
@@ -7,43 +7,64 @@
 import sys
 import os
 import subprocess as SP
+import argparse, textwrap
+import xml.etree.ElementTree as ET
+
+parser = argparse.ArgumentParser(
+    description='ChakraCore *nix Test Script',
+    formatter_class=argparse.RawDescriptionHelpFormatter,
+    epilog=textwrap.dedent('''\
```
dilijev: I think you could just define the text as a script-level constant instead of defining it inline at this indentation level and taking a dependency on textwrap.

Side note: I'd prefer one package per import line.

test/runtests.py
```diff
@@ -7,43 +7,64 @@
...
+# arch: x86, x64
+arch = 'x86' if args.x86 else ('x64' if args.x64 else None)
+if arch == None:
+    arch = os.environ.get('_BuildArch', 'x86')
+
+# flavor: debug, test, release
+flavor = 'debug' if args.debug else ('test' if args.test else None)
+if flavor == None:
+    flavor = {'chk':'debug', 'test':'test', 'fre':'release'}\
+                    [os.environ.get('_BuildType', 'fre')]
```
dilijev: This would be easier to read if the map of flavors was a variable of its own and the whole access expression fit on one line.

test/runtests.py
```diff
@@ -7,43 +7,64 @@
...
+
+# binary: full ch path
+binary = args.binary
+if binary == None:
+    if sys.platform == 'win32':
+        binary = os.path.join(repo_root,
+            'Build/VcBuild/bin/{}_{}/ch.exe'.format(arch, flavor))
```
dilijev: This whole statement is less than 100 characters. I don't think we need to try so hard to fit within 80 columns. In python especially, I prefer not to split lines unless it specifically helps readability.

dilijev:
Echo @digitalinfinity's concerns. Besides comments, LGTM

test/runtests.py
obastemur: > unless it specifically helps readability..

+1 for keeping it 80 (also useful with the console based text editors!)

test/runtests.py
```diff
@@ -7,43 +7,64 @@
...
-        sys.exit(0)
-    test_all = None
-
-test_dirs=['']
-if test_all:
-    test_dirs = os.listdir(test_root)
-else:
-    test_dirs[0] = sys.argv[1]
+pass_count = 0
+fail_count = 0
```
obastemur: How about defining a class object for test results and keeping all the test related global information inside? The global variables are usually tend to mix-up at some point.

test/runtests.py
dilijev: I think a while back, around the time of the line-endings saga, I discovered that there are some tests that rely on the text of the source code being printed for a function, and since the source code contains `\r\n`, the test output needed to have that as well.

test/runtests.py
dilijev: I'm ambivalent on this particular point, but I'll point out that a lot of the code in this project breaks any reasonable expectation of max line length.

test/runtests.py
dilijev: :+1:

test/runtests.py
tcare: @digitalinfinity is correct - we have scanner tests that need binary comparison due to mixed and non-standard line endings. See the files marked in .gitattributes.

test/runtests.py
tcare: The rest of our code base actively doesn't conform to 80.

test/runtests.py
dilijev: Side note: something to watch out for is that python on windows will replace any `\n` in text strings with `\r\n` so for that we would need to read and compare the files using binary buffers to avoid that annoying behavior.

test/runtests.py
jianchun: Thanks for the suggestion! I'll try simplify this -- actually only "compile-flags" could be empty and if so can't be passed to `Popen`.

test/runtests.py
obastemur: I guess there is a mix with '\r` from tests due to CRLF and a string with CR. See https://github.com/Microsoft/ChakraCore/pull/984/ for discussion and comments

jianchun: Yes we may need to polish this. This is the part I inherited from previous script and didn't touch at all in this change.

jianchun:
> LGTM- couple comments and couple meta comments:
> 
> 
> Could do with more comments throughout the script in general
> Might be worth adding exclude-tags support up-front so that we can exclude tests that require the JIT.

Will add some and continue to enhance it. Sent this out to unblock "test/Basics" tests first.

test/runtests.py
jianchun: Will do. Copied this from python docs example.

test/runtests.py
jianchun: There are many tools broken on long lines. I'd like to stick to 80-column myself...

test/runtests.py
jianchun: Yes agreed.

test/runtests.py
```diff
@@ -69,52 +90,88 @@ def show_failed(filename, output, exit_code, expected_output):
...
+    js_output = ""
+
+    cmd = [x for x in [binary,
+            '-WERExceptionSupport',
+            '-ExtendedErrorStackForTestHost',
+            test.get('compile-flags'),
+            js_file]  if x != None]
+    p = SP.Popen(cmd, stdout=SP.PIPE, stderr=SP.STDOUT)
+    js_output = p.communicate()[0].replace('\r','')
+    exit_code = p.wait()
```
ianwjhalliday: So this test runner runs the tests in serial right now if I am understanding this correctly, is that right?

Plans to to make it parallel?

Also timeout functionality will be desirable if there is a plan to get this in continuous integration at some point.

jianchun: Yes, we will make it parallel. Timeout is needed. Currently this is still quite rudimental.

test/runtests.py
```diff
@@ -7,114 +7,208 @@
...
+# arch: x86, x64
+arch = 'x86' if args.x86 else ('x64' if args.x64 else None)
+if arch == None:
+    arch = os.environ.get('_BuildArch', 'x86')
+
+# flavor: debug, test, release
+type_flavor = {'chk':'debug', 'test':'test', 'fre':'release'}
+flavor = 'debug' if args.debug else ('test' if args.test else None)
+if flavor == None:
+    flavor = type_flavor[os.environ.get('_BuildType', 'fre')]
```
dilijev: If this is hardcoded, why have the map lookup?

test/runtests.py
```diff
@@ -7,114 +7,208 @@
...
-                print lst_expected[i]
-                print "----------------------------"
+                print("Output: (at line " + str(i) + ")")
+                print("----------------------------")
+                print(lst_output[i])
+                print("----------------------------")
+                print("Expected Output:")
+                print("----------------------------")
+                print(lst_expected[i])
+                print("----------------------------")
```
dilijev: Are you planning on using `from __future__ import print_function`?

Otherwise using what looks like function syntax with the print function is actually printing a tuple with one element.

test/runtests.py
```diff
@@ -7,114 +7,208 @@
...
+        test_one(folder, test)
+
+def test_one(folder, test):
+    js_file = os.path.join(folder, test['files'])
+    js_output = ""
+
+    flags = test.get('compile-flags')
+    cmd = [binary, '-WERExceptionSupport', '-ExtendedErrorStackForTestHost'] \
+          + (flags.split() if flags else []) \
+          + [js_file]
```
dilijev: I think if you put the pluses at the end of the line it will look to the next line for the rest of the expression so you wouldn't need the `\` at the ends of the lines.

test/runtests.py
```diff
@@ -7,114 +7,208 @@
...
+    p = SP.Popen(cmd, stdout=SP.PIPE, stderr=SP.STDOUT)
+    js_output = p.communicate()[0].replace('\r','')
+    exit_code = p.wait()
+
+    if exit_code != 0:
+        return show_failed(js_file, js_output, exit_code, None)
+    else: #compare outputs
```
dilijev: I think you should move this comment onto its own line at the start of the block.

dilijev:
Looks like you're delaying the exclude tags work to a future PR? Sounds good to me, just want to make sure that was intentional.

test/runtests.py
dilijev: Alternatively, since the if block ends with a return, you can eliminate the `else:` entirely and de-indent this block. Leave a blank line, and put the comment at the beginning of the section.

test/runtests.py
jianchun: > If this is hardcoded, why have the map lookup?

It is not hardcoded. It reads env variable `_BuildType`, and default to `fre` if no env variable.

dilijev: Oh, thanks for clarifying.

test/runtests.py
jianchun: I did not know about that. Merely tried to make this work for both 2.7 and 3.x. Maybe we can add that line?

test/runtests.py
jianchun: Yes, I tried that as well, couldn't tell which looks better...

jianchun:
> Looks like you're delaying the exclude tags work to a future PR? Sounds good to me, just want to make sure that was intentional.

Yes, future work. This one doesn't even have "print pass skip baseline file" support which we are using everywhere.

jianchun:
@digitalinfinity @dilijev @obastemur @ianwjhalliday Thanks for review and suggestions!

PR #984, opened 2016-05-13
runtests.py: improve CRLF handling and popen working folder 

obastemur:
- Simplify help output
- Improve CRLF handling
- Set Javascript file path as a cwd for POpen to prevent the output/baseline path related differences

### DIFF
--- a/test/runtests.py
+++ b/test/runtests.py
@@ -9,6 +9,7 @@ import os
 import subprocess as SP
 import argparse
 import xml.etree.ElementTree as ET
+import re
 
 # handle command line args
 parser = argparse.ArgumentParser(
@@ -18,14 +19,14 @@ parser = argparse.ArgumentParser(
 Samples:
 
 test all folders:
-    {0}
+    runtests.py
 
 test only Array:
-    {0} Array
+    runtests.py Array
 
 test a single file:
-    {0} Basics/hello.js
-'''.format(sys.argv[0]))
+    runtests.py Basics/hello.js
+''')
 
 parser.add_argument('folders', metavar='folder', nargs='*',
                     help='folder subset to run tests')
@@ -63,7 +64,6 @@ if binary == None:
 if not os.path.isfile(binary):
     print('{} not found. Did you run ./build.sh already?'.format(binary))
     sys.exit(1)
-
 
 # records pass_count/fail_count
 class PassFailCount(object):
@@ -146,10 +146,15 @@ def test_one(folder, test):
     js_output = ""
 
     flags = test.get('compile-flags')
+
+    working_path = os.path.dirname(js_file)
+    file_name = os.path.basename(js_file)
+
     cmd = [binary, '-WERExceptionSupport', '-ExtendedErrorStackForTestHost'] \
           + (flags.split() if flags else []) \
-          + [js_file]
-    p = SP.Popen(cmd, stdout=SP.PIPE, stderr=SP.STDOUT)
+          + [file_name]
+
+    p = SP.Popen(cmd, stdout=SP.PIPE, stderr=SP.STDOUT, cwd=working_path)
     js_output = p.communicate()[0].replace('\r','')
     exit_code = p.wait()
 
@@ -160,10 +165,15 @@ def test_one(folder, test):
         if os.path.isfile(baseline):
             expected_output = None
             with open(baseline, 'r') as bs_file:
-                expected_output = bs_file.read().replace('\r', '')
-            # todo: compare line by line and use/implement wild cards support
-            # todo: by default we discard line endings (xplat), make this optional
-            if expected_output.replace('\n', '') != js_output.replace('\n', ''):
+                baseline_output = bs_file.read()
+
+            # Cleanup carriage return
+            # todo: remove carriage return at the end of the line
+            #       or better fix ch to output same on all platforms
+            expected_output = re.sub('[\r]+\n', '\n', baseline_output)
+
+            # todo: implement wild cards support
+            if expected_output != js_output:
                 return show_failed(js_file, js_output, exit_code, expected_output)
 
     print("\tPassed -> " + os.path.basename(js_file))


test/runtests.py
```diff
@@ -89,7 +89,11 @@ def test_path(folder, is_file):
                 print "Javascript file doesn't exist (" + js_file + ")"
                 sys.exit(1)
 
-            p = SP.Popen([ch_path, js_file], stdout=SP.PIPE, stderr=SP.STDOUT, close_fds=True)
+            working_path = os.path.dirname(js_file)
+            file_name = os.path.basename(js_file)
+            p = SP.Popen([ch_path, file_name], stdout=SP.PIPE, stderr=SP.STDOUT, 
+                                                             cwd=working_path,
+                                                             close_fds=True)
```
jianchun: Question about `close_fds=True`: I deleted it in my PR, because python complains this parameter combination is invalid:

```
If close_fds is true, all file descriptors except 0, 1 and 2 will be closed before the child process is executed. (Unix only). Or, on Windows, if close_fds is true then no handles will be inherited by the child process. Note that on Windows, you cannot set close_fds to true and also redirect the standard handles by setting stdin, stdout or stderr.
```

What were your purpose to use this parameter? To close the new pipe? Do we need to worry about closing the new pipe?

obastemur: I thought this X in mind. Good point on Linux. The initial plan was to make testing compatible to MT runs. i.e. Test multiple JS files at a time to reduce testing time. Still I can remove that and it should be fine.

test/runtests.py
```diff
@@ -99,9 +103,19 @@ def test_path(folder, is_file):
...
-                        expected_output = bs_file.read().replace('\r', '')
+                        baseline_output = bs_file.read()
+
+                    # clear carriage return
+                    # todo: remove carriage return at the end of the line
+                    #       or better fix ch to output same on all platforms
+                    expected_output = baseline_output.replace('\r', '')
+
+                    # normalize tabs on both outputs to 2 spaces
+                    expected_output = expected_output.replace('\t', '  ')
```
jianchun: I didn't get why you need to process `\t`. It doesn't matter what is displayed. If both baseline file and captured output use `\t`, they should match.

If you replaced them here and `show_failed` prints to log file, people will actually be confused when comparing with baseline files, because now the log has `'  '` while baseline has `\t`.

obastemur: I didn't understand? on both sides they are replaced to two char spaces. The difference is due to output from ch on a particular platform. This issue is similar to `\r`

jianchun: > The difference is due to output from ch on a particular platform. This issue is similar to \r

What's the difference of `\t` on different platforms? Isn't it the same character?

You replaced here only in the log, but actual `.baseline` file contains `\t`. Thus people will be confused reading the log file and the baseline file as they don't match.

obastemur: It's not due our our output on different platforms. 

`2020 2061 7420  Error: xx.   at ` .. there is no 09 here..

jianchun: > 2020 2061 7420  Error: xx.   at .. there is no 09 here..
> 
> update:
> This is an output on Linux Ubuntu for throw new Error("xx")

Is it possible due to missing command line flag `-ExtendedErrorStackForTestHost`? If you add that flag, output should have `\t`. That flag is default in test harness and also now in runtests.py. (I don't know the history why this flag was added and affects this behavior.)

```
if (CONFIG_FLAG(ExtendedErrorStackForTestHost))
{
    bs->AppendChars(_u("\n\tat "));
}
else
{
    bs->AppendChars(_u("\n   at "));
}
```

obastemur: Yes. I'm not sure if we use a workaround or flag ? I guess with that flag in use it is safe to close this PR?

test/runtests.py
```diff
@@ -18,14 +19,14 @@
...
+    runtests.py
 
 test only Array:
-    {0} Array
+    runtests.py Array
 
 test a single file:
-    {0} Basics/hello.js
-'''.format(sys.argv[0]))
+    runtests.py Basics/hello.js
```
obastemur: When running from root folder, help was showing `test/runtests.py Basics/hello.js` however it should be `tests/runtests.py test/Basics/hello.js` .. simplified the help output instead.

obastemur:
Updated and re-based the PR. 

- Simplify help output
- Improve CRLF handling
- Set Javascript file path as a cwd for POpen to prevent the output/baseline path related differences

test/runtests.py
jianchun: Since the flag is in test harness (runtests.cmd/rl.exe), this script can be compatible and provide the flag.

Now I think on it, the history of this flag might be: We used to always print `\t` here, so all our baselines contain `\t`. But then we wanted to be compatible and use 3 spaces instead. To avoid updating all baselines, we chose this test flag to keep printing `\t` in all tests. (I think we should change this, no point to print `\t`. Should update all baselines to 3 spaces.)

test/runtests.py
jianchun: Alternatively can use basename(sys.argv[0]) to avoid hardcode the script name.

Currently tab completion doesn't work well with test scripts, because there are 2 scripts with similar file names -- runtests.sh and runtests.py. May we consider rename this one to `test.py`?

test/runtests.py
```diff
@@ -64,6 +65,9 @@
     print('{} not found. Did you run ./build.sh already?'.format(binary))
     sys.exit(1)
 
+if len(sys.argv) > 1:
+    if sys.argv[1] in ['-?', '--help']:
+        print "ChakraCore *nix Test Script\n"
```
jianchun: Did you keep this intentionally, or was it merge error? This doesn't seem needed. The argparse module handles `-h, --help` automatically. It doesn't handle `-?` as I recall, but also prints help on that.

test/runtests.py
obastemur: from runtests.py to test.py ? 👍

obastemur:

test/runtests.py
jianchun: > from runtests.py to test.py ? 👍

Yes

jianchun:
@obastemur Signed off

PR #986, opened 2016-05-13
Param debug scope metadata.

akroshg:
The debugger scopes are created in the non-debug mode as well. Removed the only-debug-mode check for creating debug scope for parameters.


akroshg:
@agarwal-sandeep have a look.

agarwal-sandeep:
Looks good to me.

PR #987, opened 2016-05-13
flush print/WScript.Echo output

jianchun:
Flush stdout after print/WScript.Echo calls (don't know why we didn't have
one already). On Linux we found different stdout/stderr output order when
redirecting/piping output in python script.


digitalinfinity:
:shipit:

dilijev:
LGTM

obastemur:
👍

PR #988, opened 2016-05-14
Implement 0xB binary format

commonlisp:
- Postorder opcodes and compilation
- Shorter section names
- Section name/section length order swap
- Arities in call, return, and break
- Signature in signature section has a type constructor
- if/else introduces blocks
- Merge if and if_else to handle postorder opcodes


lib/WasmReader/WasmBinaryReader.cpp
```diff
@@ -423,7 +421,7 @@ WasmBinaryReader::ModuleHeader()
         ThrowDecodingError(_u("Malformed WASM module header!"));
     }
 
-    if (version != 10)
+    if (version != 0xb)
```
dilijev: Changing from decimal 10 (0xA) to 11 (0xB) intentional?

commonlisp: Yeah, that is how the Community Group specs are written now.

lib/WasmReader/WasmBinaryReader.cpp
```diff
@@ -574,7 +573,7 @@ WasmBinaryReader::GetWasmToken(WasmBinOp op)
 bool
 WasmBinaryReader::EndOfFunc()
 {
-    return m_funcState.count == m_funcState.size;
+    return m_funcState.count >= m_funcState.size;
```
MikeHolman: why >= now?

commonlisp:
One thing I'm definitely working on changing is the data structure for the postorder m_evalStack. It really ought not be an SList.

lib/WasmReader/WasmBinaryReader.cpp
commonlisp: It really should have been >= to begin with. With ==, one can get subtle bug conditions where if m_funcState.count wasn't maintained correctly (e.g., it skipped m_funcState.size or just started out > then m_funcState.size), the decoder and bytecode compiler keeps running. At least that seems to be how we are using m_funcState.count right now. There is another question whether such conditions should be checked separately.

MikeHolman: I was thinking in such bug conditions we should immediately fail compilation.

commonlisp: Agreed! I'll look around to see where would be the best place for such a check.

Cellule:
Could you update your fork so it sees that your previous PR has been merged.

lib/WasmReader/WasmByteCodeGenerator.cpp
```diff
@@ -211,6 +213,7 @@ WasmBytecodeGenerator::GenerateFunction()
             {
                 throw WasmCompilationException(_u("Last expression return type mismatch return type"));
             }
+            m_reader->m_currentNode.ret.arity = 1;
```
Cellule: The arity should be 0 if the return type is Void

lib/WasmReader/WasmByteCodeGenerator.cpp
```diff
@@ -197,12 +198,13 @@ WasmBytecodeGenerator::GenerateFunction()
...
         WasmOp op = wnLIMIT, newOp;
         EmitInfo exprInfo;
+        EnterEvalStackScope();
         while ((newOp = m_reader->ReadExpr()) != wnFUNC_END)
         {
-            ReleaseLocation(&exprInfo);
             op = newOp;
             exprInfo = EmitExpr(op);
         }
+        ExitEvalStackScope();
```
Cellule: You are exiting the eval scope too soon. We might emit a return before leaving that function.

lib/WasmReader/WasmByteCodeGenerator.cpp
```diff
@@ -211,6 +213,7 @@ WasmBytecodeGenerator::GenerateFunction()
             {
                 throw WasmCompilationException(_u("Last expression return type mismatch return type"));
             }
+            m_reader->m_currentNode.ret.arity = 1;
             EmitReturnExpr(&exprInfo);
```
Cellule: You need to push `exprInfo` to the eval stack if the return type is not `Void` before Emitting the Return expression.

Cellule:
I am currently updating sexpr-wasm test runner to work with ch and I am running it with your changes. I will report as I find issues.
Once the changes are stable and done, I will share it with you so you can test it yourself to find and fix issues

commonlisp:
Thanks. I'll be creating some test cases to capture those issues you kindly flagged.

Cellule:
There quite a few places where we call `EmitExpr` which is invalid with Post Order.
We should change these calls to `PopEvalStack`.
Notably I found a bug in `EmitMemRead` where we try to emit the next op to get the index, this is wrong because the index has already been emitted

lib/WasmReader/WasmBinaryReader.h
```diff
@@ -94,6 +94,7 @@ namespace Wasm
             virtual WasmOp ReadFromBlock() override;
             virtual WasmOp ReadFromCall() override;
             virtual WasmOp ReadExpr() override;
+            virtual void Unread() override { m_pc--; m_funcState.count--; }
```
Cellule: It looks like `Unread` is used only to get the last read opcode. Why not just have a variable that keeps track of last opcode read ?
Also, `Unread` is confusing since you don't really "unread" it, you just go back so you can read it again.

dilijev: Agreed with this concern. If keeping this function, maybe rename to `Backtrack`?

commonlisp: I'll revisit this and see if we can avoid this function altogether.

commonlisp:
@dotnet-bot test this please

commonlisp:
Closed by #1091

PR #991, opened 2016-05-15
#110 Convert rlexe.xml tags to use new build architecture and type terms

### LINKED ISSUE #110: rlexe.xml tags: add support for new build architecture and type terms
Tests use legacy build architecture and type terms for exclusion tags:
- `exclude_amd64`
- `exclude_chk`
- `exclude_fre`

[`test/runtests.cmd`](https://github.com/Microsoft/ChakraCore/blob/master/test/runtests.cmd) uses the new terms as input but shims them to the old terms for the exclusion tags.  It would be nicer to change all the rlexe.xml files to use the new terms and remove the shim from `runtests.cmd`.

New tags should be:
- `exclude_x64`
- `exclude_debug`
- `exclude_test`

----

@dilijev: Update 12/18/2016

Due to the pervasive use of the old tag names across multiple dev branches and both internally and externally, the initial work item here should be to add support for the new tags, and ensure that we continue to support the old tags.

At the time of the PR which adds support, all of the existing uses of the old terms should be updated to the new terms. New tests added via other branches which do not yet support the new terms will still use the old terms, so supporting both for a time is necessary.

This means any place we would exclude either the old or new tag, we must exclude both. For example, we must exclude BOTH `exclude_x64` (new) and `exclude_arm64` (old), and never only one of them.

At some point in the future (when all active development branches support the new terms), we will deprecate the use of the old terms and remove support for them. At that time, all remaining uses of the old terms will be replaced with the new terms and support will be dropped from the scripts.

See follow-up work item #2240.

RexSong:
Fixes #110


test/runtests.cmd
```diff
@@ -323,15 +323,10 @@ goto :main
...
-  set _BuildArchMapped=%_BuildArch%
   set _BuildTypeMapped=%_BuildType%
 
-  :: Map new build arch and type names to old names until rl test tags are
-  :: updated to the new names
-  if "%_BuildArchMapped%" == "x64" set _BuildArchMapped=amd64
-  if "%_BuildTypeMapped%" == "debug" set _BuildTypeMapped=chk
-  if "%_BuildTypeMapped%" == "test" set _BuildTypeMapped=fre
-  if "%_BuildTypeMapped%" == "codecoverage" set _BuildTypeMapped=fre
+  :: codecoverage need to mapped to test config
```
dilijev: Change this to "codecoverage needs to be mapped to test config"

dilijev:
Big thanks for this! I believe @Cellule was talking about doing this a while back -- take a look?

test/runtests.cmd
```diff
@@ -323,15 +323,10 @@ goto :main
     set _DIRS=-all
   )
 
-  set _BuildArchMapped=%_BuildArch%
   set _BuildTypeMapped=%_BuildType%
 
-  :: Map new build arch and type names to old names until rl test tags are
-  :: updated to the new names
-  if "%_BuildArchMapped%" == "x64" set _BuildArchMapped=amd64
```
Cellule: I think we should keep the map, but reverse it to ease up transition. At a later time we'll remove it.

bin/rl/rl.cpp
```diff
@@ -228,7 +228,7 @@ TARGETINFO TargetInfo[] = {
    "sh5c",       FALSE, FALSE, FALSE, "sh5sim",       "libcsim.lib", "x86asm",
    "sh5m",       FALSE, FALSE, FALSE, "sh5sim",       "libcsim.lib", "x86asm",
    "ia64",       FALSE, TRUE,  TRUE,  NULL,           NULL, "x86asm",
-   "amd64",      FALSE, TRUE,  TRUE,  NULL,           NULL, "x86asm",
```
Cellule: I think we should keep both, that way it will work regardless of which we use. I imagine `ia64` is just an alias for `amd64` at this time, we should have the same behavior to map `amd64` to `x64` to keep all existing tools working

ianwjhalliday: Agree, we should have both.  Internal tests may not be converted and rl still needs to work with them.

`ia64` is different from `amd64` though, I wouldn't expect those to be aliases.

test/runtests.cmd
ianwjhalliday: There's no need to keep the reverse mapping.  No one uses this script with the old names.

That said, our `unittest` tests will need to have their `exclude_*` tags converted to be compatible with this change to `runtests.cmd`, which would then also require `unittest\RunAllRLTests.cmd` to be updated to do the reverse mapping so that snap still works.

I believe the internal `ProjectionTests` suite has a completely separate script for calling rl, and so should not be affected by this change. So fortunately merging this PR shouldn't require any work in the `ProjectionTests` suite.

test/runtests.cmd
ianwjhalliday: I was hoping we could remove both `Mapped` environment variables.  `codecoverage` puts an annoying thorn in that.  Maybe this can be a separate option?

bin/rl/rl.h
```diff
@@ -73,6 +73,7 @@ typedef enum {
     TM_SH5M,
     TM_IA64,
     TM_AMD64,
+	TM_X64,
```
Cellule: nit: use space for tabs

dilijev:
@RexSong any update on this?

RexSong:
Updated for the comments. but I saw there is big change for current source. do I need to update for current source?

dilijev:
@RexSong yes, it looks like there are conflicts that need to be resolved, so you'll need to resolve them before we can merge this PR. Were there any other updates that were needed to the code that haven't been made yet?

dilijev:
@RexSong also if there was nothing left for you to do at the time - our sincerest apologies for not merging this in sooner.

RexSong:
@dilijev I will resolve the conflicts and commit this PR ASAP.

dilijev:
Thanks @RexSong

RexSong:
Commit a new PR [Updated] #110 Convert rlexe.xml tags to use new build architecture and type terms
 instead of this, because it is hard to merge. Close this PR.

dilijev:
PSA: you can use git push --force to push to the same branch (no need to open a new PR). That's fine though. We'll merge that one instead.

RexSong:
@dilijev Thank you for sharing the tips.

PR #992, opened 2016-05-16
xplat: enable -test build type

jianchun:
Simple changes to enable "test" build type on cross-platform.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -20,14 +20,15 @@ PRINT_USAGE() {
     echo "  build.sh [options]"
     echo ""
     echo "  options:"
-    echo "    --cxx         : Path to Clang++ (see example below)"
-    echo "    --cc          : Path to Clang   (see example below)"
-    echo "    --debug, -d   : Debug build (by default Release Build)"
-    echo "    --help, -h    : Show help"
-    echo "    --icu         : Path to ICU include folder (see example below)"
-    echo "    --jobs, -j    : Multicore build (i.e. -j=3 for 3 cores)"
-    echo "    --ninja, -n   : Build with ninja instead of make"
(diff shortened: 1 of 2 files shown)


build.sh
```diff
@@ -27,6 +27,7 @@ PRINT_USAGE() {
     echo "    --icu         : Path to ICU include folder (see example below)"
     echo "    --jobs, -j    : Multicore build (i.e. -j=3 for 3 cores)"
     echo "    --ninja, -n   : Build with ninja instead of make"
+    echo "    --test, -t    : Test build (by default Release Build)"
```
obastemur: `--test` feels like it's going to run tests. `--test-build` ?

digitalinfinity: agree with @obastemur. Alternatively, --fretest if you want something shorter? Although that has msft connotations with the usage of fre, so I'm fine with whatever here.

jianchun: Updated to --test-build. Thanks!

build.sh
```diff
@@ -27,6 +27,7 @@ PRINT_USAGE() {
     echo "    --icu         : Path to ICU include folder (see example below)"
     echo "    --jobs, -j    : Multicore build (i.e. -j=3 for 3 cores)"
     echo "    --ninja, -n   : Build with ninja instead of make"
+    echo "    --test-build, -t  : Test build (by default Release Build)"
```
dilijev: I think you could update the whole stretch of options here so the colons are aligned.

jianchun: > I think you could update the whole stretch of options here so the colons are aligned.

Done. Thanks!

dilijev:
LGTM

jianchun:
Thanks @obastemur @digitalinfinity @dilijev

PR #994, opened 2016-05-16
MSFT: 7563889 stress AV when javascriptLibrary is nullptr

Yongqu:
JavascriptLibrary can null be if the script context initialization threw.
move the cleanup code inside the null check block.


ianwjhalliday:
Presume the missing word in your description is 'null' :)

:shipit:

Yongqu:
@ianwjhalliday oops. You are right and the comment updated. thanks!

PR #995, opened 2016-05-16
Fix for pid ref stack issue in split scope when a lambda shadows a param

aneeshdk:
When a lambda in the param scope has a parameter that shadows a parameter in the parent function an additional pid ref is inserted into the ref stack as we parse the arguments list of lambda initially as an expression. This causes the parent function to be marked as split scope and the symbol requires scope slots. But as there is no non-local reference a slot register was not getting allocated for the function. The fix is to assign frame slots register when the function has split scope.


aneeshdk:
@ianwjhalliday @pleath @akroshg @tcare Can you please take a look?

pleath:
LGTM.

PR #996, opened 2016-05-16
Adding a flag to force all functions to have split scope

aneeshdk:
-ForceSplitScope will force all functions to have unmerged body and param
scope. This flag will be used in our test infrastructure to create more
split scope scenarios.


aneeshdk:
@abchatra Can you please take a look?

abchatra:
:shipit:

pleath:
Can we make this a phase switch? (Those work in the parser now.)

PR #997, opened 2016-05-17
xplat: enhance runtests.py

jianchun:
Run tests in parallel.
Support tags, not_tags (excludes).
Support `pass` baselines.
Support test time out.
Support test variants, currently only interpreted.

Print elapsed time for each test and entire run.

Load and filter test folders with test_root/rlexedirs.xml.

Temporary: Find and use correct filename on disk if rlexe.xml filename
case mismatches. Print a warning.

### DIFF
--- a/bin/ch/ch.cpp
+++ b/bin/ch/ch.cpp
@@ -392,7 +392,7 @@ HRESULT ExecuteTest(const char* fileName)
 
     IfJsErrorFailLog(ChakraRTInterface::JsCreateContext(runtime, &context));
     IfJsErrorFailLog(ChakraRTInterface::JsSetCurrentContext(context));
-    
+
 #ifdef DEBUG
     ChakraRTInterface::SetCheckOpHelpersFlag(true);
 #endif
@@ -583,7 +583,11 @@ int main(int argc, char** argv)
         Helpers::NarrowStringToWideDynamic(argv[i], &args[i]);
     }
 
-    int ret = wmain(argc, args);
+    // Call wmain with a copy of args, as HostConfigFlags may change argv
+    char16** argsCopy = new char16*[argc];
+    memcpy(argsCopy, args, sizeof(args[0]) * argc);
+    int ret = wmain(argc, argsCopy);
+    delete[] argsCopy;
 
     for (int i = 0; i < argc; i++)
     {
--- a/test/runtests.py
+++ b/test/runtests.py
@@ -4,6 +4,10 @@
 # Licensed under the MIT license. See LICENSE.txt file in the project root for full license information.
 #-------------------------------------------------------------------------------------------------------
 
+from __future__ import print_function
+from datetime import datetime
+from multiprocessing import Pool, Manager
+from threading import Timer
 import sys
 import os
 import subprocess as SP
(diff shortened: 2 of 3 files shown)


obastemur:
It was really useful for me to see what is expected vs output for the first line failed. It doesn't really matter if many other things failed on the same file since we can run that file and see output / compare the baseline individually right?

jianchun:
> It was really useful for me to see what is expected vs output for the first line failed. It doesn't really matter if many other things failed on the same file since we can run that file and see output / compare the baseline individually right?

I didn't change that logic. When baseline mismatches it should still dump the first diff line. Doesn't it work that way?

obastemur:
Just tested, actually It does `in general`. Maybe I'm missing something? IMHO, The output is a bit mixed when lots of tests are failing. Tried with `test/Array`

bin/ch/ch.cpp
```diff
@@ -583,7 +583,11 @@ int main(int argc, char** argv)
         Helpers::NarrowStringToWideDynamic(argv[i], &args[i]);
     }
 
-    int ret = wmain(argc, args);
+    // Call wmain with a copy of args, as HostConfigFlags may change argv
+    char16** argsCopy = new char16*[argc];
+    memcpy(argsCopy, args, sizeof(args[0]) * argc);
```
obastemur: This is an interesting use of `memcpy` to convert from `char` to `char16` but I'm not sure if it works?

digitalinfinity: I dont understand- he's not converting from char to char16, he's simply copying from the char16 array called args to the char16 array called newArgs?

---
In reply to: [63589023](https://github.com/Microsoft/ChakraCore/pull/997#discussion_r63589023) [](ancestors = 63589023)

jianchun: This is one problem of reviewing on GitHub -- hard to read code context. This is not copying chars. It is copying pointers.

obastemur: Yep. :+1:

jianchun:
@obastemur @digitalinfinity I've also fixed multi-process output issue (by queueing to main process to print) and updated PR. Please review. Thanks!

test/runtests.py
```diff
@@ -98,126 +130,302 @@ def log(self, filename=None, folder=None, fail=False):
...
+        tags = set(tags.split(',')) if tags else self._empty_set
+        if not tags.isdisjoint(self.not_tags):
+            return False
+        if self.tags and not self.tags.issubset(tags):
+            return False
+        return True
+
+    # queue a test result from multi-process runs
+    def _log_result(self, filename, fail=False):
+        self.msg_queue.put((self.MSG_TEST_RESULT, filename, fail))
```
obastemur: It could be nice to implement lock here.

obastemur:
Apart from a single comment LGTM 👍

test/runtests.py
obastemur: Actually, we collect the outputs from a particular run/process at a time right? In that case no lock is needed but the comment on line 149 is confusing

jianchun: As I understand from reading multiprocessing docs, Manager().Queue() is "thread and process safe". Each time I sent a single complete message -- a test pass/fail result, or a piece of output. So no lock is needed.

jianchun:
@obastemur Thanks for review!

test/runtests.py
jianchun: I missed one thing -- currently we have multiple messages (_print calls) per test result. We can indeed use a lock to group all these together so messages from different processes don't mix. I'll do that later.

Issue #319, opened 2016-02-16
Fix timezone-sensitive tests (disabled in Jenkins)


dilijev:
There are a handful of tests which by their nature are timezone-sensitive (testing Date APIs). Because the execution of some Javascript Date APIs is dependent on the system time zone, the tests only work when run in the time zone in which the baselines were created (in our case, Pacific Time [UTC-8]).

The Jenkins machines are configured with UTC time and so we noticed that these tests would fail in any other timezone. To work around the problem for now, we disabled these tests in Jenkins.

It would be nice to add a way to enable these tests.

My thoughts:
- Some of these tests may not need to depend on the time zone. If possible, modify the tests to make them more general (as long as we keep exercising the same code paths).
- Add a flag to the test runner (`rl.exe`) and/or `ch.exe` which provides the ability to spoof the system timezone for tests, so that we can test the right code paths within ChakraCore which do depend on the system's reported time zone.

To locate the affected tests which were disabled in Jenkins, you can run:
`git grep -n -C 3 "timezone-sensitive"`

/cc @digitalinfinity @tcare @ianwjhalliday @kunalspathak FYI

Converted from [VSO 5239718](https://microsoft.visualstudio.com/web/wi.aspx?pcguid=cb55739e-4afe-46a3-970f-1b49d8ee7564&id=5239718)

tcare:
This is something that we've been meaning to fix for a while. @ianwjhalliday encountered some pain over this when he was working remotely a while back :)

ianwjhalliday:
@akroshg noticed it too on when he was working remote recently.

Afaik there is no way to change the timezone per process for easy spoofing.  The system's timezone can be changed but that would affect new processes and not be a robust solution in the face of concurrent testing via rl.exe (it would work if all tests used the same timezone for spoofing, which they do currently, but it would lead to race condition bugs if new tests used other timezones).  It's also not friendly to the user.

Spoofing if possible would be best done in the engine itself, in the date and time APIs.

ianwjhalliday:
One test I ran into was using a Date to get a long string for use as a property name.  It didn't depend on the date itself, nor the timezone, so I hardcoded the string that new Date(...) produced to make it work across timezones.

dilijev:
@tcare Yep I had a bug open internally from when we set up Jenkins and I noticed this originally. I decided to move it here since it seems to be ChakraCore-specific. I should have added a link. Added that now.

[...]

PR #1001, opened 2016-05-18
Fixes #804 Json parsing errors mention JSON.parse in errors

### LINKED ISSUE #804: JSON.parse("") gives very ambigious "Syntax Error".
```javascript
JSON.parse("")
```

```
$ ch.exe jsonParseEmpty.js
SyntaxError: Syntax error
   at Global code (d:\personal\chakracore\build\vcbuild\bin\x64_debug\test.js:1:1)
```

'Syntax Error' is very ambiguous error. It could mean a lot of things. When its part of a large callstack, I would think its a javascript syntax error somewhere in my js files. The callstack doesn't even mention JSON.parse.

Chrome gives
```
VM145:1 Uncaught SyntaxError: Unexpected end of JSON input(…) 
```

Ideal error should mention something like "JSON.parse Syntax error: Empty string" or "JSON.parse Syntax Error: unexpected end of json input."

nojvek:
I added JSON.parse specific errors. It still doesn't feel very clean as there is no column info. Would this be relatively easy to add?

### DIFF
--- a/lib/Runtime/Library/JSONParser.cpp
+++ b/lib/Runtime/Library/JSONParser.cpp
@@ -35,7 +35,7 @@ namespace JSON
         Js::Var ret = ParseObject();
         if (m_token.tk != tkEOF)
         {
-            Js::JavascriptError::ThrowSyntaxError(scriptContext, ERRsyntax);
+            m_scanner.ThrowSyntaxError(JSERR_JsonSyntax);
         }
         return ret;
     }
@@ -232,7 +232,7 @@ namespace JSON
             }
             else
             {
-                Js::JavascriptError::ThrowSyntaxError(scriptContext, ERRbadNumber);
+                m_scanner.ThrowSyntaxError(JSERR_JsonBadNumber);
             }
 
         case tkLBrack:
@@ -260,11 +260,11 @@ namespace JSON
                     Scan();
                     if(tkRBrack == m_token.tk)
                     {
-                        Js::JavascriptError::ThrowSyntaxError(scriptContext, ERRillegalChar);
+                        m_scanner.ThrowSyntaxError(JSERR_JsonIllegalChar);
                     }
                 }
                 //check and consume the ending ']'
-                CheckCurrentToken(tkRBrack, ERRnoRbrack);
+                CheckCurrentToken(tkRBrack, JSERR_JsonNoRbrack);
                 return arrayObj;
 
             }
@@ -311,7 +311,7 @@ namespace JSON
                     //pick "name"
                     if(tkStrCon != m_token.tk)
                     {
-                        Js::JavascriptError::ThrowSyntaxError(scriptContext, ERRsyntax);
+                        m_scanner.ThrowSyntaxError(JSERR_JsonIllegalChar);
                     }
 
                     // currentStrLength = length w/o null-termination
@@ -332,7 +332,7 @@ namespace JSON
                             //check and consume ":"
                             if(Scan() != tkColon )
                             {
-                                Js::JavascriptError::ThrowSyntaxError(scriptContext, ERRnoColon);
+                                m_scanner.ThrowSyntaxError(JSERR_JsonNoColon);
                             }
                             Scan();
 
@@ -365,7 +365,7 @@ namespace JSON
                     //check and consume ":"
                     if(Scan() != tkColon )
                     {
-                        Js::JavascriptError::ThrowSyntaxError(scriptContext, ERRnoColon);
+                        m_scanner.ThrowSyntaxError(JSERR_JsonNoColon);
                     }
                     Scan();
                     Js::Var value = ParseObject();
@@ -404,12 +404,12 @@ namespace JSON
                 }
 
                 // check  and consume the ending '}"
-                CheckCurrentToken(tkRCurly, ERRnoRcurly);
+                CheckCurrentToken(tkRCurly, JSERR_JsonNoRcurly);
                 return object;
             }
 
         default:
-            Js::JavascriptError::ThrowSyntaxError(scriptContext, ERRsyntax);
+            m_scanner.ThrowSyntaxError(JSERR_JsonSyntax);
         }
     }
 } // namespace JSON
--- a/lib/Runtime/Library/JSONScanner.h
+++ b/lib/Runtime/Library/JSONScanner.h
@@ -20,9 +20,16 @@ namespace JSON
             ::Js::ScriptContext* sc, const char16* current, ArenaAllocator* allocator);
 
         void Finalizer();
-        char16* GetCurrentString(){return currentString;}
-        uint GetCurrentStringLen(){return currentIndex;}
+        char16* GetCurrentString() { return currentString; } 
+        uint GetCurrentStringLen() { return currentIndex; }
+        uint GetScanPosition() { return uint(currentChar - inputText); }
 
+        void __declspec(noreturn) ThrowSyntaxError(int wErr)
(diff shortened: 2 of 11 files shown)


ianwjhalliday:
Thanks @nojvek, these new error messages are definitely more helpful.

The error stack will have the line column information for the source of the throw, which will be where the JSON.parse call occurred, so we do have that already.

Could also be useful to have character offset (or line/column?) information for the source of the error in the JSON string passed to parse.  I don't know if that would require much additional work on in the JSON parser or not.  Certainly this can be raised as a new separate `Suggestion` issue if you think it is a good idea.

@abchatra any concerns over adding new error messages?

nojvek:
Right now there seems to be a LSC_ERROR_MSG(..) macro definition of a
syntax error.

How do I throw a syntax error with a dynamically created string ?

ianwjhalliday:
Look at `rterrors.h` where we have errors with two forms of their message string.  The first form has a %s parameter that will be replaced by the given string to the `JavascriptError::Throw*` methods.

I believe you should be able to move the errors to `rterrors.h` and still use them for `ThrowSyntaxError` without any issues.

abchatra:
> @abchatra any concerns over adding new error messages?

I am happy to see improvement in the error messages.

nojvek:
Thanks abchatra for the help today. I have made the fixes to show the position just like v8 does for JSON.parse. wdyt?

ianwjhalliday:
Looks great!  Thanks @nojvek, I'll get this merged.

ianwjhalliday:
@nojvek turns out we have some JSON tests that do not run in jenkins currently because they depend on timezone.

You will need to manually run these two tests, `test/JSON/jx2.js` and `test/es5/jx2.js`, and update their baselines to match your change before we can merge this in.  These two tests happen to run with no ch.exe command line switches, but in general when running tests manually you'd want to confirm you are using the right switches by examining the `test/.../rlexe.xml` file to see if it specifies switches and if so what they are.  Also this will show you what the baseline file is named, sometimes it doesn't match the source file.

I also just noticed that your new test `syntaxerror.js` is not added to `test/JSON/rlexe.xml`.  Please add it there so that it runs in the test runner.

ianwjhalliday:
Note the json tests that don't run in jenkins because of their timezone dependency is an understood issue that we have not fixed yet.  It depends on PST timezone.

lib/Runtime/Library/JSONParser.cpp
```diff
@@ -404,12 +404,13 @@ namespace JSON
...
                 // check  and consume the ending '}"
-                CheckCurrentToken(tkRCurly, ERRnoRcurly);
+                CheckCurrentToken(tkRCurly, JSERR_JsonNoRcurly);
                 return object;
             }
 
         default:
-            Js::JavascriptError::ThrowSyntaxError(scriptContext, ERRsyntax);
+            m_scanner.ThrowSyntaxError(JSERR_JsonSyntax);
+            return NULL;
```
abchatra: Not required. This statement won't be reached.

lib/Runtime/Library/JSONScanner.h
```diff
@@ -20,9 +20,16 @@ namespace JSON
...
-        char16* GetCurrentString(){return currentString;}
-        uint GetCurrentStringLen(){return currentIndex;}
+        char16* GetCurrentString() { return currentString; } 
+        uint GetCurrentStringLen() { return currentIndex; }
+        uint GetScanPos() { return uint(currentChar - inputText); }
 
+        void ThrowSyntaxError(int wErr)
+        {
+            char16 scanPos[16];
+            ::_itow_s(GetScanPos(), scanPos, sizeof(scanPos) / sizeof(char16), 10);
```
abchatra: Use _countof(scanPos)

lib/Runtime/Library/JSONScanner.h
```diff
@@ -20,9 +20,16 @@ namespace JSON
             ::Js::ScriptContext* sc, const char16* current, ArenaAllocator* allocator);
 
         void Finalizer();
-        char16* GetCurrentString(){return currentString;}
-        uint GetCurrentStringLen(){return currentIndex;}
+        char16* GetCurrentString() { return currentString; } 
+        uint GetCurrentStringLen() { return currentIndex; }
+        uint GetScanPos() { return uint(currentChar - inputText); }
```
abchatra: Nit: Rename to GetScanPosition?

lib/Runtime/Library/JSONScanner.h
```diff
@@ -20,9 +20,16 @@ namespace JSON
...
 
+        void ThrowSyntaxError(int wErr)
```
abchatra: This should be long right?

lib/Runtime/Library/JSONParser.cpp
nojvek: Without this C++ won't compile.

>e:\personal\chakracore\lib\runtime\library\jsonparser.cpp(415): warning C4715: 'JSON::JSONParser::ParseObject': not all control paths return a value

lib/Runtime/Library/JSONScanner.h
nojvek: the function takes a int32 hcode

err_method(ScriptContext* scriptContext, int32 hCode, PCWSTR varName = nullptr)

lib/Runtime/Library/JSONParser.cpp
ianwjhalliday: That is because `JSONScanner::ThrowSyntaxError` is not marked with `__declspec(noreturn)`.

Probably reasonable to change the definition of `JSONScanner::ThrowSyntaxError` to:
```c++
void __declspec(noreturn) ThrowSyntaxError(int wErr)
```

nojvek: done.

lib/Runtime/Library/JSONScanner.h
nojvek: done.

lib/Runtime/Library/JSONScanner.h
nojvek: done.

nojvek:
Ensured tests pass in  runtests.cmd -x64 -debug and added syntaxError.js to rl.xml

ianwjhalliday:
@nojvek you forgot to update the baseline for `test\es5\jx2.js` as well.

nojvek:
Oh forgot that. I updated Json/jx2.js

ianwjhalliday:
@nojvek when you get a chance please update `test\es5\jx3.baseline`.  I've updated internal test baselines so once you've done that I believe we'll be all ready to merge your change.

nojvek:
Done. Ran all 1655 tests and confirmed they passed.

Not sure why baseline for jx2.js was called jx3.js. I've renamed to
jx2.baseline.

Soz for delay

ianwjhalliday:
Awesome, thanks for the rename.  I don't know why it was different either.  Probably some historical reason that didn't get cleaned up nicely.

nojvek:
Would be great to see this go in today. Let me know if you need anything else from my side. Thanks

ianwjhalliday:
@nojvek we're all good, merged!  Thanks for improving these error messages

nojvek:
That's awesome. Thanks Ian.

abchatra:
Thank you @nojvek

dilijev:
Note that we're tracking the timezone-sensitive tests issue in #319

PR #1004, opened 2016-05-18
MemOp loop opcode whitelist & BailOnNotArray instr removal

Cellule:
Check for a list of instruction that are valid for memop and refuse the optimization if it is unknown. 

Add BailOnNotArray in the list of instruction to remove when emitting a memop. This check if done in the helper and is thus not needed.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/1004)


Cellule:
@LouisLaf @pleath

lib/Backend/GlobOpt.cpp
```diff
@@ -4528,6 +4529,23 @@ GlobOpt::CollectMemOpInfo(IR::Instr *instr, Value *src1Val, Value *src2Val)
         // Fallthrough if not an induction variable
     }
     default:
+        // List of instruction that are valid with memop (ie: instr that gets removed if memop is emitted)
+        if (
+            this->currentBlock != loop->GetHeadBlock() &&
```
pleath: Could we extract the opcode list as an API? That would probably make it friendlier for maintenance down the road. Otherwise LGTM.

Cellule:
@pleath I extracted the list into a function and it does look cleaner. What do you think ?
That particular new condition is similar to a check we do in the backward pass. I wanted to make a function that would cover both cases, but I couldn't find one that would not look hacky.
I decided to leave it as it is.

pleath:
LGTM. Thanks.

PR #1006, opened 2016-05-18
Port "Changes for MicroBuild v2" to release/1.1

dilijev:
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/1006)

### DIFF
--- a/test/es6/rlexe.xml
+++ b/test/es6/rlexe.xml
@@ -121,6 +121,7 @@
       <tags>BugFix</tags>
     </default>
   </test>
+  <!-- OS BUG 7117880
   <test>
     <default>
       <files>toStringTag.js</files>
@@ -128,6 +129,7 @@
       <tags>exclude_dynapogo</tags>
     </default>
   </test>
+  -->
   <test>
     <default>
       <files>proto_basic.js</files>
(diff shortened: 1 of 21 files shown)


test/es6/rlexe.xml
```diff
@@ -121,13 +121,15 @@
       <tags>BugFix</tags>
     </default>
   </test>
+  <!-- OS BUG 7117880
```
Cellule: Should this be in release/1.1 ?

Cellule:
:+1: for pgo related changes

test/es6/rlexe.xml
dilijev: Since the compiler issue will affect any builds in MicroBuild v2 this could be a problem for release/1.1 as well. Not sure if it's strictly necessary.

dilijev:
FYI @ianwjhalliday 

I don't think any changes are needed as this passed in all the various CIs. Feel free to merge this in for me while I'm out?

dilijev:
PGO related changes are good and are signed off. Disabling the test due to the compiler bug will affect this branch so I think it's good to check in this change. We will re-enable that test when a fix to the tooling is available or that unit test becomes critical.

Merging this change.

Thanks @Cellule @ianwjhalliday

PR #1008, opened 2016-05-19
Array Destructuring issue on .next function.

akroshg:
The iterator.next should not be called if the last slot is empty slot. Fixed that and added the test.


akroshg:
@abchatra @aneeshdk @tcare please have a look.

aneeshdk:
LGTM

PR #1009, opened 2016-05-19
xplat: fix rlexe.xml filename case

jianchun:
Change all file names (files/baseline) in rlexe.xml to match real file
names on disk.

### DIFF
--- a/test/Array/rlexe.xml
+++ b/test/Array/rlexe.xml
@@ -373,8 +373,8 @@
   </test>
   <test>
     <default>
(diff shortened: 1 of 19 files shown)


obastemur:
IMO this should target master and we merge it back from there.

jianchun:
> IMO this should target master and we merge it back from there.

Targeting master works better; But since nobody modifies these lines and we only need them in Linux, making the change here is more convenient.

digitalinfinity:
:shipit:

PR #1010, opened 2016-05-19
xplat: runtests.py: prevent mixed output, misc improvements

jianchun:
Prevent mixed output: Previously a test run sends/queues multiple result
messages. This can result in intertwined messages from different test runs
thus mixed output. Change test run _print to buffer outputs, and sends
whole test pass/fail/elapsed_time together with output as one message.

Add `exclude_x86/x64' to global `not_tags` to better filter tests.

Support multiple `tags` in rlexe.xml. Used by some tests.

Support test specific `timeout`.

Changed log format slightly to log [#current_test/#total_test] to indicate
progress.


digitalinfinity:
:shipit:

PR #1015, opened 2016-05-20
MemOp Loop Memory allocation

Cellule:
Allocate Memop memory only if needed.
I did some tracing on the benchmarks.
Before this change there were over 1000 occurrences of the memop memory to be allocated, but not used.
After this change there were 85 occurrences.

The main reason is that right now any loop with 2 blocks will have all the memop memory info allocated.
The downside of this change is that it adds a boolean to every loops.
I was thinking doing something funky on the pointer instead (like -1 means do not do memop), but I fear this might be confusing just to save 1 boolean.

This is not really fixing a bug, and there doesn't seems to be a perf impact, but I came across this as I was fixing something else and it bugged me.
I think it's worth changing.
Maybe I should target the master branch instead ?

@LouisLaf @pleath @rajatd What do you guys think ?

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/1015)

### DIFF
--- a/lib/Backend/GlobOpt.cpp
+++ b/lib/Backend/GlobOpt.cpp
@@ -4187,30 +4187,33 @@ GlobOpt::IsAllowedForMemOpt(IR::Instr* instr, bool isMemset, IR::RegOpnd *baseOp
     Assert(iv->IsChangeDeterminate() && iv->IsChangeUnidirectional());
     const IntConstantBounds & bounds = iv->ChangeBounds();
 
-    // Only accept induction variables that increments by 1
-    Loop::InductionVariableChangeInfo inductionVariableChangeInfo = { 0, 0 };
-    inductionVariableChangeInfo = loop->memOpInfo->inductionVariableChangeInfoMap->Lookup(indexSymID, inductionVariableChangeInfo);
-
-    if (
-        (bounds.LowerBound() != 1 && bounds.LowerBound() != -1) ||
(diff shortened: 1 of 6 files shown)


lib/Backend/GlobOpt.cpp
```diff
@@ -4348,11 +4359,6 @@ bool GlobOpt::CollectMemcopyStElementI(IR::Instr *instr, Loop *loop)
     SymID srcSymID = GetVarSymID(src1->GetStackSym());
 
     // Prepare the memcopyCandidate entry
-    if (loop->memOpInfo->candidates->Empty())
-    {
-        // There is no ldElem matching this stElem
-        return false;
-    }
     Loop::MemOpCandidate* previousCandidate = loop->memOpInfo->candidates->Head();
```
rajatd: Don't we need a call to EnsureMemOpVariablesInitialized before this line?

rajatd:
Looks good to me otherwise.

rajatd:
And yes, target master :)

lib/Backend/GlobOpt.cpp
Cellule: We don't because at the beginning of this function we return false if there are no candidates (or no memopInfo allocated)

rajatd: ok, makes sense.

---
In reply to: [65933989](https://github.com/Microsoft/ChakraCore/pull/1015#discussion_r65933989) [](ancestors = 65933989)

PR #1016, opened 2016-05-20
Kill 'caller' fields when entering and leaving an inlinee

rajatd:
Function.caller should only be copy-propped when it is invoked from the same call chain. As such, it cannot be copy-propped from a parent function into an inlinee or from an inlinee to outside the inlinee.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/1016)


pleath:
LGTM. Thanks.

rajatd:
Thanks @pleath . Closing this and opening a new PR against master.

PR #1017, opened 2016-05-20
Tweaking rejit heuristics

rajatd:
Our current threshold of call-count-to-bailout-count ratio beyond which a rejit is triggered, is 20. Also, we don’t have such a ratio for loop bodies, we just rejit after the second bailout for the same bailout record. I had been looking into tuning the ratio for function rejits and also to make loop body rejits heuristics based instead of a hard bailout limit of 2. 

Making rejits less aggressive should generally help with CPU time for websites. However, making them too conservative starts to show regressions on benchmarks due to increased number of bailouts. From my experiments, sweet spot for call-count to bailout-count ration for rejit seems to 10.

_Loop body rejits:_
1. Implemented a counter for loop bodies to keep track of how many iterations we run in the jitted code before bailing out of a jitted loop body. Then, I use the ratio of jitted-iterations-count to bailout-count to trigger a rejit. Settling on a value of 50 as the threshold for this ratio.
2. Ported some preemptive rejit strategies that we use for functions, to the loop body rejitting code.


rajatd:
Closing to re-open against master

PR #1018, opened 2016-05-21
cross platform: fix Math.pow -Infinity case

obastemur:
- fix -> on some platforms `pow` doesn't support `-inf`
 - add test case for non-integer negative exponent for -0
 - add test case for negative integer exponent for +0


obastemur:
Updated the PR and added a test case. A non integer negative exponent for -0 would return `+Infinity` instead of `-Infinity`.

lib/Runtime/Library/MathLibrary.cpp
```diff
@@ -816,6 +816,23 @@ namespace Js
     {
         double result = 0;
 
+#ifdef POW_DOESNT_SUPPORT_NEGATIVE_INFINITY
+        if ( signbit( x ) && 0 == NumberUtilities::LuLoDbl( x ) && 0 == ( NumberUtilities::LuHiDbl( x ) & 0x7FFFFFFF ) )
```
digitalinfinity: More curiosity than anything else- can we instead of this do *((int64*) &x) == Js::NumberConstants::k_NegZero? Either way, it would be nice to have this refactored out into its own function in NumberUtilities for readability.

lib/Runtime/Library/MathLibrary.cpp
```diff
@@ -816,6 +816,23 @@ namespace Js
...
+        {
+            // On some platforms pow doesn't support (-Infinity)
+            // When x is -0, and y is an odd `integer` < 0
+            // The expected result is -Infinity
+            if ( y < 0 && (int64)y & 1 )
```
digitalinfinity: Nit: wrap ((int64) y & 1) in parentheses for readability

digitalinfinity:
:shipit:

lib/Runtime/Library/MathLibrary.cpp
obastemur: IIRC having `-0.0` here is important.

>  Either way, it would be nice to have this refactored out into its own function in NumberUtilities for readability.

This is a `pow` specific case as `POW_DOESNT_SUPPORT_NEGATIVE_INFINITY` and it's under `Math::Pow`

curtisman: @obastemur.  I think @digitalinfinity  means you should refactor the -0.0 detect into NumberUtilities.    But think some thing similar already exists.  See JavascriptNumber::IsNegZero.  That one is an instance method, so we can't really use it, but it is implemented as calling `NumberUtilities::IsSpecial(value, k_NegZero)`

So either provide a static version of NumberUtilties::IsNegZero or just directly call `NumberUtilities::IsSpecial(value, NumberUtilities::k_NegZero)` here to detect negative zero

lib/CMakeLists.txt
```diff
@@ -1,5 +1,20 @@
 add_compile_options(-fPIC)
 
+set(CMAKE_REQUIRED_LIBRARIES m)
+check_cxx_source_runs("
+#include <stdlib.h>
+#include <math.h>
+int main()
+{
+    double f = pow(-0, -1);
```
curtisman: You might need -0.0 instead of -0, here?

obastemur:
So the problem here was the implementation from PAL. After cleaning up that part there was no need for dealing platform `-inf` issues.

@curtisman @digitalinfinity Thanks for the reviews!

curtisman:
:shipit:

PR #1021, opened 2016-05-22
cross-platform: Enable non-ICU toLocateString

obastemur:
`wcscmp` produces positive or negative results based on the inputs. However we expect -1, 0, +1

Also, normalized another usage of `wcscmp` although no consumer found for `Compare`.

### DIFF
--- a/lib/Common/DataStructures/Comparer.h
+++ b/lib/Common/DataStructures/Comparer.h
@@ -155,9 +155,10 @@ struct StringComparer
         return hash;
     }
 
-    static int Compare(T str1, T str2)
+    inline static int Compare(T str1, T str2)
     {
-        return ::wcscmp(str1, str2);
+        const int result = ::wcscmp(str1, str2);
+        return result > 0 ? 1 : result == 0 ? 0 : -1;
     }
 };
 
--- a/lib/Runtime/Library/JavascriptString.cpp
+++ b/lib/Runtime/Library/JavascriptString.cpp
@@ -1339,6 +1339,9 @@ case_2:
         const char16* pThatStr = pThat->GetString();
         int thatStrCount = pThat->GetLength();
 
+        // xplat-todo: doing a locale-insensitive compare here
+        // but need to move locale-specific string comparison to
+        // platform agnostic interface
 #ifdef ENABLE_GLOBALIZATION
         LCID lcid = GetUserDefaultLCID();
         int result = CompareStringW(lcid, NULL, pThisStr, thisStrCount, pThatStr, thatStrCount );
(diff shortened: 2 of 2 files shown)


jianchun:
> However we expect -1, 0, +1

What code expects this? Is it possible to fix the caller not to expect -1/+1, as https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare states `DO NOT rely on exact return values of -1 or 1`?

obastemur:
Our test cases expect. It is nice to have a consisten behavior cross platform.

ljharb:
As far as I'm aware, all comparators on all languages and platforms tend to return 0, a negative, or a positive. It's fine to clamp but don't the larger magnitudes convey information?

obastemur:
>  It's fine to clamp but don't the larger magnitudes convey information?

I'm not sure if that information is any useful at the first place. First we don't know for sure if it's going to be available on any platform / clib implementation. Second, it would return the difference between the non equal first characters those we don't know the location. 

ChakraCore on Windows returns -1, 0, +1 although MSDN states that wcscmp may return > 0 and <0 based on the comparison. node.js also does the same for Windows/*nix and returns -1, 0, +1. 

I would appreciate if ChakraCore would give me the same output regardless from platform / lib-c implementation.

obastemur:
I don't have a strong opinion on output. My reasons is detailed in a comment above. 

More importantly, this PR fixes an assert crash with non ICU string comparison.

digitalinfinity:
I'm in the "no strong opinion" boat too. I don't think we have too many tests that depend on this so I see @jianchun's point about just fixing up the test to expect the right thing. On the other hand, the ES5 spec states that the return values just need to be negative, zero, or positive, with the actual return values being implementation dependent. So @obastemur is also fine fixing it in this manner. So yeah guys- you can toss a coin to decide 😄 💸

obastemur:
I still don't have the signoff working. Appreciate signoff in that case.

@jianchun @ljharb @digitalinfinity Thanks for the review.

jianchun:
I'm inclined to fix the test instead because the test is basically non-conformant to spec. The test is implementation dependent thus not a great test. Somebody reading or copying the test in future might think +1/-1 is the standard; code following this may break on other browsers.

obastemur:
> code following this may break on other browsers.

This PR fixes `cross platform implementation only`  and makes it similar to actual Windows base. (node has same output cross platform)

So there is no change for browser.

jianchun:
I meant imagine somebody learn from this test and write "if (localCompareResult == 1) ..." which works on Edge and breaks on other browsers.

obastemur:
> I meant imagine somebody learn from this test and write "if (localCompareResult == 1) ..." which works on Edge and breaks on other browsers.

FF, CH, and EDGE returns -1, 0, +1 

If we are into changing +1 to sub(diff1, diff2) on all the platforms, I'm also okay with that. However that would be different than other browsers.

We could leave it as is and depend on libc on platform but IMO that wouldn't be a developer friendly solution.

jianchun:
I'm fine with it if matching browsers. Signing off.

lib/Common/DataStructures/Comparer.h
```diff
@@ -155,9 +155,10 @@ struct StringComparer
         return hash;
     }
 
-    static int Compare(T str1, T str2)
+    inline static int Compare(T str1, T str2)
     {
-        return ::wcscmp(str1, str2);
+        const int result = ::wcscmp(str1, str2);
```
curtisman: @obastemur I am curious why this needs too change as well?

obastemur: I had mentioned this one on the details of this PR. I couldn't find any consumer for that interface yet still normalized the output from libc call to prevent future consistency

PR #1022, opened 2016-05-23
Stack Arguments Optimization with presence of Formals

satheeshravi:
- First Cut with creation of scope object and no heap arguments creation.
- This PR is not check-in ready but is sent as a checkpoint towards the final implementation.
- This change contains the following
     - Tracking of writes to formals, in the Parser
    - Removal of Heap arguments object creation in the presence of formals(in the Lowerer).
     - Optimization is disabled when there are any writes to the formals.
     - Bail out path is fixed by recreating the heap arguments object and pointing to the right scope object.

Addl. notes:
- Parser, LoadHeapArguments helper APIs and Backend changes are ready for review.
- Changes pertaining to BailOut will go off with my subsequent iteration. Bail out change was done to complete a temporary milestone(as we still have scope object creation). 
- Next change will contain removal of scope object creation when formals are present without any writes.

Perf details to follow after this.


satheeshravi:
Will resubmit the PR. It is branched off master.

PR #1024, opened 2016-05-23
xplat: 2 test changes

jianchun:
runtests.py: "hide" passed tests on the terminal

```
While running tests we are more interested in failed tests. Printing
passed results on the screen makes it harder to read failed tests.

Change `runtests.py` slightly so that "Pass" tests are still printed on the
terminal, but immediately overwritten by the next result. Now we can still
read test progress but passed tests no longer flush the screen.
```

xplat: exclude more unsupported tests 

```
Exclude more Simd tests (require_backend), Debugger tests (
require_debugger), and apply exclude_chk/exclude_fre.

Exclude tests that require compile-flags: -serialized, -simdjs (neither
implemented yet on cross-platform).
```

### DIFF
--- a/test/rlexedirs.xml
+++ b/test/rlexedirs.xml
@@ -278,24 +278,24 @@
 <dir>
   <default>
     <files>SIMD.float32x4</files>
-    <tags>exclude_arm,exclude_arm64</tags>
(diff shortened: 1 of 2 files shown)


obastemur:
LGTM

digitalinfinity:
:shipit:

PR #1025, opened 2016-05-23
cross platform: fix FWrite output buffer size

obastemur:
Fixes unicode output to file stream.

```
print("ab\u00A0cd");

before fix : ab c
after fix  : ab cd
```


digitalinfinity:
:shipit:

Issue #1014, opened 2016-05-20
Method defined in object literal should not have 'prototype' property


suwc:
var o = {get m() {}};
var f = Object.getOwnPropertyDescriptor(o, 'm').get;
console.log(f.hasOwnProperty('prototype'));  // expected: false, actual: true

suwc:
https://github.com/Microsoft/ChakraCore/pull/1027

PR #1027, opened 2016-05-23
Getter/setter methods defined in object literal should not have 'prototype' property

suwc:
Fix following bug:
    var o = {get m() {}};
    var f = Object.getOwnPropertyDescriptor(o, 'm').get;
    console.log(f.hasOwnProperty('prototype')); // expected: false, actual: true

### DIFF
--- a/lib/Runtime/Library/JavascriptLibrary.cpp
+++ b/lib/Runtime/Library/JavascriptLibrary.cpp
@@ -4962,13 +4962,13 @@ namespace Js
 #if DEBUG
             if (!function->GetFunctionProxy()->GetIsAnonymousFunction())
             {
-                Assert(!function->GetFunctionInfo()->IsLambda() ?
+                Assert(function->GetFunctionInfo()->IsConstructor() ?
                     function->GetDynamicType()->GetTypeHandler() == JavascriptLibrary::GetDeferredPrototypeFunctionTypeHandler(this->GetScriptContext()) :
                     function->GetDynamicType()->GetTypeHandler() == JavascriptLibrary::GetDeferredFunctionTypeHandler());
             }
             else
             {
-                Assert(!function->GetFunctionInfo()->IsLambda() ?
+                Assert(function->GetFunctionInfo()->IsConstructor() ?
                     function->GetDynamicType()->GetTypeHandler() == JavascriptLibrary::GetDeferredAnonymousPrototypeFunctionTypeHandler() :
                     function->GetDynamicType()->GetTypeHandler() == JavascriptLibrary::GetDeferredAnonymousFunctionTypeHandler());
             }
--- a/lib/Runtime/Types/ScriptFunctionType.cpp
+++ b/lib/Runtime/Types/ScriptFunctionType.cpp
@@ -31,7 +31,7 @@ namespace Js
             scriptContext, functionPrototype,
             address,
             proxy->GetDefaultEntryPointInfo(),
-            library->ScriptFunctionTypeHandler(proxy->IsLambda() || proxy->IsAsync() || proxy->IsClassMethod(), proxy->GetIsAnonymousFunction()),
+            library->ScriptFunctionTypeHandler(!proxy->IsConstructor() || proxy->IsAsync() || proxy->IsClassMethod(), proxy->GetIsAnonymousFunction()),
             isShared, isShared);
     }
 };
--- a/test/es6/object_literal_bug.js
+++ b/test/es6/object_literal_bug.js
@@ -37,7 +37,20 @@ var tests = [
             }
         }
     }
-  }
+  },
+  {
+    name: "Getter/setter methods defined in object literal should not have 'prototype' property",
+    body: function () {
+        var o = {
+            get m() {},
+            set m(v) {}
+        };
+        var g = Object.getOwnPropertyDescriptor(o, 'm').get;
+        var s = Object.getOwnPropertyDescriptor(o, 'm').set;
+        assert.areEqual(false, g.hasOwnProperty('prototype'), "Getter method defined in object literal should not have 'prototype' property");
+        assert.areEqual(false, s.hasOwnProperty('prototype'), "Setter method defined in object literal should not have 'prototype' property");
+    }
+  },
 ];
 
 testRunner.runTests(tests, { verbose: WScript.Arguments[0] != "summary" });


suwc:
https://github.com/Microsoft/ChakraCore/issues/1014

suwc:
@ianwjhalliday @boingoing could you take a look?

lib/Runtime/Types/ScriptFunctionType.cpp
```diff
@@ -31,7 +31,7 @@ namespace Js
             scriptContext, functionPrototype,
             address,
             proxy->GetDefaultEntryPointInfo(),
-            library->ScriptFunctionTypeHandler(proxy->IsLambda() || proxy->IsAsync() || proxy->IsClassMethod(), proxy->GetIsAnonymousFunction()),
+            library->ScriptFunctionTypeHandler(proxy->IsLambda() || !proxy->IsConstructor() || proxy->IsAsync() || proxy->IsClassMethod(), proxy->GetIsAnonymousFunction()),
```
ianwjhalliday: `IsConstructor` implies `!(IsLambda || IsAsync || IsClassMethod)` so you can remove the checks for `IsLambda` etc in favor of just checking `IsConstructor`

lib/Runtime/Library/JavascriptLibrary.cpp
```diff
@@ -4962,13 +4962,13 @@ namespace Js
...
             {
-                Assert(!function->GetFunctionInfo()->IsLambda() ?
+                Assert(!function->GetFunctionInfo()->IsLambda() && function->GetFunctionInfo()->IsConstructor() ?
                     function->GetDynamicType()->GetTypeHandler() == JavascriptLibrary::GetDeferredPrototypeFunctionTypeHandler(this->GetScriptContext()) :
                     function->GetDynamicType()->GetTypeHandler() == JavascriptLibrary::GetDeferredFunctionTypeHandler());
             }
             else
             {
-                Assert(!function->GetFunctionInfo()->IsLambda() ?
+                Assert(!function->GetFunctionInfo()->IsLambda() && function->GetFunctionInfo()->IsConstructor() ?
```
ianwjhalliday: Likewise in these two cases `IsConstructor` implies `!IsLambda` so the `!IsLambda` here is redundant

ianwjhalliday:
Missing a regression test to cover these case, and a couple comments about simplifying the logic checks.

:+1:

suwc:
@ianwjhalliday: updated. Thanks!

lib/Runtime/Types/ScriptFunctionType.cpp
```diff
@@ -31,7 +31,7 @@ namespace Js
             scriptContext, functionPrototype,
             address,
             proxy->GetDefaultEntryPointInfo(),
-            library->ScriptFunctionTypeHandler(proxy->IsLambda() || proxy->IsAsync() || proxy->IsClassMethod(), proxy->GetIsAnonymousFunction()),
+            library->ScriptFunctionTypeHandler(!proxy->IsConstructor() || proxy->IsAsync() || proxy->IsClassMethod(), proxy->GetIsAnonymousFunction()),
```
ianwjhalliday: `IsAsync` and `IsClassMethod` both imply `!IsConstructor` so are also redundant and should also be removed.

test/es6/object_literal_bug.js
```diff
@@ -1,43 +1,56 @@
...
+        };
+        var g = Object.getOwnPropertyDescriptor(o, 'm').get;
+        var s = Object.getOwnPropertyDescriptor(o, 'm').set;
+        assert.areEqual(false, g.hasOwnProperty('prototype'), "Getter method defined in object literal should not have 'prototype' property");
+        assert.areEqual(false, s.hasOwnProperty('prototype'), "Setter method defined in object literal should not have 'prototype' property");
+    }
+  },
+];
+
+testRunner.runTests(tests, { verbose: WScript.Arguments[0] != "summary" });
```
ianwjhalliday: Looks like the EOL character changed in this test file.  Did it change to be consistent with other test files or is this a mistake?

lib/Runtime/Types/ScriptFunctionType.cpp
suwc: I will push another PR to remove them.

test/es6/object_literal_bug.js
suwc: Oops. I didn't intend to change EOL. It may be from Notepad++.
Do you see a compelling reason to change it back?

ianwjhalliday: It might mean your autocrlf setting is inverted.

Consistency is a compelling reason to change it.  Since you'll submit another PR its worth checking out if what the EOL encoding is in this file checked in vs other files.  If it is the odd one out then let's change it back.

If we remain lax about this sort of thing then inconsistencies grow and creep up on us over time and things get messy.

lib/Runtime/Types/ScriptFunctionType.cpp
suwc: https://github.com/Microsoft/ChakraCore/pull/1038

test/es6/object_literal_bug.js
suwc: I am not able to find inconsistencies after this change is merged.
The following command returned 'true' for me - seems correct:
git config --get core.autocrlf

@dilijev: Is Jenkins able to catch EOL that is messed up? This PR passed all Jenkins.

Changing it back is not justified at this point.

suwc: Thanks to @dilijev for clarifying. Let me recap in case I misunderstood:
1) Jenkins EOL check doesn't operate on .js files because there are cases that are intentional
2) This file: test/es6/object_literal_bug.js , had Unix EOL (\n) and was changed to Windows EOL (\r\n) by git when it checked out. This is expected with autocrlf==true
3) When git checks in this file, it didn't complain because we have specifically set it such that .js files' EOL shall be left alone (same reason as 1)
4) Given 2) and 3), it would've helped reducing the size of this commit if I had explicitly run a 'dos2unix' on this file.
5) Sampling of two other .js files showed they have Windows EOL. There may be files with Unix EOL out there but we hadn't seen it. So far they are not causing any issues.
6) There's no compelling value to change test/es6/object_literal_bug.js back to Unix EOL. It might have become more consistent with other .js files in the same directory.

Therefore the recommended course of action is not to change back to Unix EOL for test/es6/object_literal_bug.js

@ianwjhalliday thanks for asking good questions. Let me know if you have other questions or concerns.

dilijev: Most JS files had CRLF when we migrated to GitHub, and a few relied on the EOL style of the source files for the baselines to be correct. A few had just LF. Since we tell GitHub to NOT convert EOL formatting in these files, and Windows checks out JS files with CRLF regardless of saved line endings thanks to the bug in .gitattributes that @tcare observed last week, when editing these files on Windows, if a file is changed, Git will record the file as it is on disk (with EOL changed to CRLF) without converting EOL to match what was previously in the file.

.gitattributes bug fix (see commit 40af6b8 in the `linux` branch)
```
- test/*.js -crlf
+ test/**/*.js -crlf
```

If no baselines broke as a result of this change, there's no reason to change it back as probably improved consistency rather than hurting it.

dilijev: Note EOL Check does not check JS files because they are allowed to have CRLF.

ianwjhalliday: Huh.  We should probably have special cased only those test files that had special requirements and not blanket ignored EOL on all test .js files.

For this specific test it is fine.

Note some editors provide functionality to convert between EOL conventions.  I believe Notepad++ has this in its menus somewhere.  I identify and convert EOL with vim personally.

dilijev: I agree that we should have been specific but there were more than a handful of tests where it mattered and considering how busy we already were at that time it would have been a huge time sink.

dilijev: Also, adding specific exceptions to the EOL check would get annoying. I originally wanted to put all of those patterns in a file and read them dynamically so that we wouldn't have to add exceptions via directly modifying the script, but that seemed like a pain to set up.

PR #1032, opened 2016-05-24
Assorted Linux fixes

digitalinfinity:
- Turn off fpo- stack walking depends on it on retail builds
- Speed up debug builds by improving performance of IsAddressOnStack
- Add support for debug and retail builds being in different directories

### DIFF
--- a/pal/src/thread/thread.cpp
+++ b/pal/src/thread/thread.cpp
@@ -2820,3 +2820,32 @@ void GetCurrentThreadStackLimits(ULONG_PTR* lowLimit, ULONG_PTR* highLimit)
     *highLimit = (ULONG_PTR) stackbase;
 #endif
 }
+
+static thread_local ULONG_PTR s_cachedThreadHighLimit = 0;
+
+bool IsAddressOnStack(ULONG_PTR address)
+{
+    // Assumption: Stack always grows, never shrinks and the high limit is stable
+    // Assumption: pthread_getattr_np is slow, so we need to cache the current stack
+    // bounds to speed up checking if a given address is on the stack
+    // The semantics of IsAddressOnStack is that we care if a given address is
+    // in the range of the current stack pointer
+    if (s_cachedThreadHighLimit == 0)
+    {
+        ULONG_PTR lowLimit, highLimit;
+        GetCurrentThreadStackLimits(&lowLimit, &highLimit);
+
+        s_cachedThreadHighLimit = highLimit;
+    }
+
+    ULONG_PTR currentStackPtr = 0;
+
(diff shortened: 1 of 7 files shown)


pal/src/thread/thread.cpp
```diff
@@ -2820,3 +2820,32 @@ void GetCurrentThreadStackLimits(ULONG_PTR* lowLimit, ULONG_PTR* highLimit)
     *highLimit = (ULONG_PTR) stackbase;
 #endif
 }
+
+static thread_local ULONG_PTR s_cachedThreadHighLimit = 0;
```
jianchun: nit pick: `s_cachedThreadHighLimit` variable name doesn't indicate it's stack related.

jianchun:
Changing binary drop location may require runtests.py update?

LGTM otherwise.

obastemur:
> Speed up debug builds by improving performance of IsAddressOnStack

Curious, how did we do that?

'fno-omit-frame-pointer' does otherwise.

digitalinfinity:
@obastemur the two are inrelated- -fno-omit-frame-pointer should not noticeably affect build time (it might slow down *runtime* on functions with high register pressure). The debug build speed up was done by improving the performance of IsAddressOnStack- it used to call GetThreadStackBounds very frequently and the Linux implementation of that function is slow- I optimized IsAddressOnStack by not having to call GetThreadStackBounds and can do a much cheaper check.

@jianchun looks like runtests.sh needs to be updated too- I'll update the PR

pal/src/thread/thread.cpp
```diff
@@ -2820,3 +2820,32 @@ void GetCurrentThreadStackLimits(ULONG_PTR* lowLimit, ULONG_PTR* highLimit)
...
+
+bool IsAddressOnStack(ULONG_PTR address)
```
obastemur: Should it be better to check if there is an upper limit in case of `ulimit -s unlimited` ? `getrlimit(RLIMIT_STACK..`

obastemur: Discard this one. In that case (if system doesn't support limits etc.) `pthread_attr_getstack` should fail.

obastemur:
> fno-omit-frame-pointer should not noticeably affect build time (it might slow down runtime on functions with high register pressure).

I'm more optimistic with runtime performance. 

> The debug build speed up was done by improving the performance of IsAddressOnStack- it used to call GetThreadStackBounds very frequently and the Linux implementation of that function is slow

@digitalinfinity I didn't understand how that would improve debug build time? Are we using this method on build ?

LGTM

jianchun:
> I didn't understand how that would improve debug build time?

He didn't mean speed up "build time". He meant speed up debug build binary.

PR #1035, opened 2016-05-24
perf: Too many bailouts for BailOutOnImplicitCalls

kunalspathak:
In below test case, function `test` always bailout and never get rejited.
```js
function test(req, res) {
    req.__proto__ = obj.prop; // bailout here every time after JIT
    res.__proto__ = obj.prop;
  };

var obj = { prop: {} };
for (var i = 0; i < 1000; i++) {
    test({}, {});
}
```

During interpretation we record that `req.__proto__` was a setter (implicit call) and set implictflag to ImplicitFlag_Accessor for that instruction. We assume that the setter would get inlined
and so we ignore `ImplicitFlag_Accessor` flag in `ImplicitFlagAllowOpts()`. However in this case, setter was not getting inlined and every time we called, we bailed out. In bailout, we just checked
that the implicitFlag was set to NONE. Here the implicitFlag was set to ImplicitFlag_Accessor and it never changed. Because of this assumption, would not REJIT and  keep bailing out for the function.

With my change, in bailout code path, we would see if that BailoutReason was BailoutOnImplicitCall and if old and new flags were not changed and they were `ImplicitFlag_Accessor`. We would rejit
and record this information in the profile info to stop ignoring `ImplicitFlag_Accessor` for that function for next JIT.

Test: UnitTest, benchmarks and internal test tools pass.


kunalspathak:
@LouisLaf , @rajatd , @pleath - Please CR.

pleath:
LGTM.

LouisLaf:
I don't think this is the right fix here.  Like Rajat mentioned over email, there shouldn't be a bailout on that instruction to start with.  Fixing that issue is a better way to go.

kunalspathak:
Makes sense now. I will close this PR for now.

PR #1036, opened 2016-05-24
perf: Fix memory leak happening in unique property guards of propertyRecords

kunalspathak:
Repro:
```js
function ctor() {
    this.p = null;
    this.q = null;
    this.r = null;
    this.s = null;
    this.t = null;
    this.u = null;
    this.v = null;
    this.w = null;
    this.x = null;
    this.y = null;
    this.z = null;
}

var superObj1 = { x: '1' };
var superObj2 = { blah: '1' };

function stress() {
    var req = Object.create(superObj1);
    var a = new ctor();
    req.__proto__ = superObj2;
    // throw away req, but cache still remains around on tc
}

function test(iter) {
    for (var i = 0; i < 1000000; i++) {
        stress();
    }
}

test(0);
print('done, sleeping...');
while(1){} // Before fix, at this point ch shows 240MB memory usage. After the fix it comes down to 170MB
```
Cause :
 1. Create an object `req` whose prototype has a property `x`.
 2. Let's say we create another object from `ctor()`.
    The ctor also has property `x` in addition to other properties.
 3. When we change the prototype of `req`, we invalidate all the properties in it's prototype as well.
    We also invalidate the propertyGuards of the properties of prototype. Since propertyGuards are stored
    at `threadContext`, we don't care which type was the property was associated with. We simply search for
    propertyGuardEntry corresponding to the propertyRecord of a property and invalidate the propertyGuard
    i.e. ctor cache of a constructor. In this case, we invalidate the propertyGuard for property `x` which
    was on prototype of `req`. However the propertyGuard of `x` corresponds to ctor cache for `ctor()` from
    which we created another object.
 4. Next time, when an object is created from `ctor()`, it sees that the ctor cache is invalid and it creates new ctor cache.
    When creating new ctor cache, all the properties of `ctor()` are again registered in propertyGuard table for new ctor cache.
 5. What is missing here is when invalidating the ctor cache in step 3 above, we should also walk through property guards table
     of other properties and clear the invalidated property guard entry from the unique property guard table. Since these are
	 not removed today, the table keeps growing retaining property guard entries of invalidated ctor cache leading to memory leak.

Fix: While invalidating a unique property guard, walk through the guards of other properties and remove the entry of property guard
getting invalidated from their table.

Test: UnitTest, benchmarks and internal test tools pass.

### DIFF
--- a/lib/Runtime/Base/ThreadContext.cpp
+++ b/lib/Runtime/Base/ThreadContext.cpp
@@ -88,7 +88,8 @@ ThreadContext::RecyclableData::RecyclableData(Recycler *const recycler) :
     typesWithProtoPropertyCache(recycler),
     propertyGuards(recycler, 128),
     oldEntryPointInfo(nullptr),
-    returnedValueList(nullptr)
+    returnedValueList(nullptr),
+    constructorCacheInvalidationCount(0)
 {
 }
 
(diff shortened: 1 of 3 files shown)


kunalspathak:
@pleath , please review.

lib/Runtime/Base/ThreadContext.cpp
```diff
@@ -2993,6 +2996,29 @@ ThreadContext::InvalidatePropertyGuardEntry(const Js::PropertyRecord* propertyRe
             }
 
             guard->Invalidate();
+
+            // If a propertyGuard is invalidated, make sure to also remove it's entry from other unique property guard table of other property records.
+            if (performCleanupOfUniquePropertyGuards)
+            {
+                guards.Map([=](const Js::PropertyRecord* otherPropertyRecord, PropertyGuardEntry* guardEntry, const Memory::RecyclerWeakReference<const Js::PropertyRecord >* otherPropertyRecordWeakRef)
```
curtisman: This looks very expensive. Is there any perf impact? Can this be done only periodically?

kunalspathak: I thought about periodically cleaning this, but running benchmarks, I didn't see any regression. All benchmarks are clean. I verified benchmark code and none of them has "change the prototype" scenario and that could be the reason.

curtisman: Well, it depends on how many InvalidatePropertyGuards you have. 
Can you do a perf lab run?

kunalspathak: > Can you do a perf lab run?

I have started one sometime back.

curtisman:
:shipit:

PR #1039, opened 2016-05-25
[wasm] Wasm.experimentalVersion

commonlisp:
To support websites serving different versions of a WASM binary
(i.e., different binary format versions), we use an experimentalVersion
property to indicate the binary format version supported.


lib/Runtime/Library/JavascriptLibrary.cpp
```diff
@@ -1117,6 +1117,9 @@ namespace Js
         globalObject->SetPrototype(globalObjectPrototype);
         Recycler * recycler = this->GetRecycler();
 
+#ifdef ENABLE_WASM
+        experimentalVersion = JavascriptNumber::New(WasmLibrary::experimentalVersion, scriptContext);
```
Cellule: Could we put this in `InitializeWasmObject` ?

lib/Runtime/Library/JavascriptLibraryBase.h
```diff
@@ -269,6 +269,7 @@ namespace Js
         Var minValue;
         Var maxValue;
         Var negativeZero;
+        Var experimentalVersion; // Wasm.experimentalVersion
```
Cellule: Put this in `JavascriptLibrary.h` changing `JavascriptLibraryBase` will break api compatibility with Edge which is unnecessary in this case

commonlisp: Got it. Thanks.

commonlisp:
@dotnet-bot test this please

lib/Runtime/Library/JavascriptLibrary.h
```diff
@@ -1236,6 +1236,9 @@ namespace Js
 #if DBG
         void DumpLibraryByteCode();
 #endif
+
+        Var experimentalVersion; // Wasm.experimentalVersion
```
Cellule: Could you put this next to the `wasmObject` inside the `#ifdef ENABLE_WASM` (line 380) of this file

Cellule: Actually I was wondering if we needed to keep this as a field in Library. Couldn't it just be a member of the Wasm object and that's it?
What I mean is, do we need to keep a reference to the member in JavascriptLibrary?

commonlisp: Good point. I dropped that field. Thanks.

Cellule:
LGTM

PR #1041, opened 2016-05-25
Decouple VTune and ETW Trace, enable VTune JIT profiling on ChakraCore

litian2025:
Decouple VTune an dETW Trace, enable VTune JIT profiling on Charka Core, using explicit dll link mechansim so that end user can build chakracore/test flavor with VTune JIT profiling enabled, VTune samples will be taken when VTune is present.

### DIFF
--- a/lib/Runtime/Base/FunctionBody.cpp
+++ b/lib/Runtime/Base/FunctionBody.cpp
@@ -12,6 +12,9 @@
 
 #include "ByteCode/ScopeInfo.h"
 #include "Base/EtwTrace.h"
+#ifdef VTUNE_PROFILING
+#include "Base/VTuneChakraProfile.h"
+#endif
 
 #ifdef DYNAMIC_PROFILE_MUTATOR
 #include "Language/DynamicProfileMutator.h"
@@ -3237,6 +3240,9 @@ namespace Js
         TraceExecutionMode();
 
         JS_ETW(EtwTrace::LogMethodNativeLoadEvent(this, entryPointInfo));
+#ifdef VTUNE_PROFILING
+        VTuneChakraProfile::LogMethodNativeLoadEvent(this, entryPointInfo);
+#endif
 
 #ifdef _M_ARM
         // For ARM we need to make sure that pipeline is synchronized with memory/cache for newly jitted code.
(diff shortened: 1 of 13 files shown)


lib/Runtime/Base/FunctionBody.cpp
```diff
@@ -4384,7 +4393,7 @@ namespace Js
 #define ORIGINAL_CDECL CDECL
 #undef CDECL
 #endif
-    // Not enabled in ChakraCore
+    // Enabled in ChakraCore
```
LouisLaf: Maybe just delete this now?

lib/Runtime/Base/jitprofiling.cpp
```diff
@@ -0,0 +1,266 @@
...
+// Copyright (C) Microsoft Corporation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE.txt file in the project root for full license information.
+//-------------------------------------------------------------------------------------------------------
+#include "RuntimeBasePch.h"
+
+#include "ittnotify_config.h"
+
+#if ITT_PLATFORM==ITT_PLATFORM_WIN
+#include <windows.h>
+#pragma optimize("", off)
```
LouisLaf: Why??

lib/Runtime/Base/FunctionBody.cpp
litian2025: Yes, I can delete this line.

lib/Runtime/Base/jitprofiling.cpp
litian2025: Good finding, thanks! It maybe be a workaround for some compiler issues and it's no longer relevant. We have tested by removing this line on VTune code base as well as ChakraCore with and without VTune, no negative impact has been found. So will remove this line: #pragma optimize("", off) and update the PR.

LouisLaf:
:shipit:

lib/Runtime/Base/jitprofiling.cpp
```diff
@@ -0,0 +1,265 @@
+//-------------------------------------------------------------------------------------------------------
+// Copyright (C) Microsoft Corporation and contributors. All rights reserved.
+// Licensed under the MIT license. See LICENSE.txt file in the project root for full license information.
+//-------------------------------------------------------------------------------------------------------
```
LouisLaf: Can the code in this file also be under #ifdef VTUNE_PROFILING?

litian2025: We definitely can do that, I will add it in and update the PR.

PR #1043, opened 2016-05-25
Clean up build error when SIMD is disabled with JIT enabled.

curtisman:

PR #1045, opened 2016-05-25
Fix error reporting when rl crashes.

dilijev:
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/1045)

### DIFF
--- a/test/runtests.cmd
+++ b/test/runtests.cmd
@@ -442,9 +442,10 @@ goto :main
   set REGRESS=%CD%
 
   call :do rl %_rlArgs%
-  if ERRORLEVEL 1 set _HadFailures=1
+  if %ERRORLEVEL% NEQ 0 set _HadFailures=1
 
   call :do move /Y %_logsRoot%\*.log %_logsRoot%\%_BuildArch%_%_BuildType%\%_TESTCONFIG%
+  if %ERRORLEVEL% NEQ 0 set _HadFailures=1
 
   set EXTRA_CC_FLAGS=%_OLD_CC_FLAGS%
 
@@ -474,6 +475,7 @@ goto :main
 
   echo ^>^> %*
   cmd /s /c "%*"
+  if %ERRORLEVEL% NEQ 0 exit /b %ERRORLEVEL%
 
   goto :eof
 


dilijev:
@dotnet-bot test this please

dilijev:
@digitalinfinity @curtisman review please?

The configs that run tests failed correctly before #1043 was merged in and they pass now.

digitalinfinity:
LGTM

dilijev:
@digitalinfinity Thanks, merging.

ianwjhalliday:
Why was this change needed?  Sorry I missed it before it merged.

test/runtests.cmd
```diff
@@ -474,6 +475,7 @@ goto :main
 
   echo ^>^> %*
   cmd /s /c "%*"
+  if %ERRORLEVEL% NEQ 0 exit /b %ERRORLEVEL%
```
ianwjhalliday: This line seems at odds with the addition on line 448 above.  This will prevent that line from executing, no?

More importantly though, `:do` is meant to just print out its command before executing it.  It should not be taking control of control flow.

dilijev: `call :do` somewhat unintuitively acts like calling a new instance of the batch script with label `:do` as the entry point, from which `exit /b` will exit and return the error code as `%ERRORLEVEL%` to the context of the call site (`exit` without `/b` would have exited the entire execution environment).

Because everything about cmd scripts always makes perfect sense.

ianwjhalliday: lol, oh yeah.  my bad, you're right.

let us... stop using batch scripts as soon as reasonably possible.

dilijev:
@ianwjhalliday There was a test that was causing RL to crash and because that was wrapped by the `runtests.cmd` script which didn't handle that situation at all, the error code from the crash was not being reported at the exit of `runtests.cmd` so that Jenkins could indicate the tests failed, even though semantically the tests did fail.

This change exposes the crash as an error. #1043 fixes the test that caused the crash. So far the cause of the crash in RL has not been addressed, but since the `linux` branch is rewriting RL in Python, it's not as big of a priority right now to fix crashes in the native version of RL.

PR #1046, opened 2016-05-26
Create log file during test run

digitalinfinity:
Update runtests.py to generate a log file of all the tests run and the
failures encountered.
Log files are by default placed in test/logs, and are named
testrun.{arch}{flavor}.log.
The filename/location can be overridden with the -l commandline switch.

### DIFF
--- a/test/runtests.py
+++ b/test/runtests.py
@@ -117,7 +117,7 @@ class LogFile(object):
         if log_file_path is None:
             # Set up the log file paths
             # Make sure the right directory exists and the log file doesn't
-            log_file_name = "testrun.{0}{1}.log".format(arch_alias, flavor_alias)
+            log_file_name = "testrun.{0}{1}.log".format(arch, flavor)
             log_file_directory = os.path.join(repo_root, "test", "logs")
 
             if not os.path.exists(log_file_directory):
@@ -141,12 +141,12 @@ class LogFile(object):
 
 log_file = LogFile(args.logfile)
 
-def print_and_log(msg = ""):
-    print(msg)
-    log_file.log(msg + "\n")
-
 def log_message(msg = ""):
     log_file.log(msg + "\n")
+
+def print_and_log(msg = ""):
+    print(msg)
+    log_message(msg)
 
 # records pass_count/fail_count
 class PassFailCount(object):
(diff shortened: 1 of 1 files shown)


test/runtests.py
```diff
@@ -107,6 +108,44 @@
...
     if sys.platform != 'win32' else None
 
+class LogFile(object):
+    def __init__(self, log_file_path = None):
+        self.file = None
+
+        if log_file_path is None:
+            # Set up the log file paths
+            # Make sure the right directory exists and the log file doesn't
+            log_file_name = "testrun.{0}{1}.log".format(arch_alias, flavor_alias)
```
jianchun: `arch, flavor` may be better than _alias here

test/runtests.py
```diff
@@ -107,6 +108,44 @@
...
+
+            if not os.path.exists(log_file_directory):
+                os.mkdir(log_file_directory)
+
+            self.log_file_path = os.path.join(log_file_directory, log_file_name)
+
+            if os.path.exists(self.log_file_path):
+                os.remove(self.log_file_path)
+        else:
+            self.log_file_path = log_file_path
```
jianchun: Does this also need `os.mkdir` code above if the dir doesn't exist?

test/runtests.py
```diff
@@ -107,6 +108,44 @@
...
+
+    def __del__(self):
+        if not (self.file is None):
+            self.file.close()
+
+log_file = LogFile(args.logfile)
+
+def print_and_log(msg = ""):
+    print(msg)
+    log_file.log(msg + "\n")
```
jianchun: nit: Can call `log_message`

test/runtests.py
```diff
@@ -197,10 +236,11 @@ def _process_msg(self, msg):
             'Failed' if fail else 'Passed',
             self._short_name(filename))
         padding = self._last_len - len(line)
+        log_message(line)
```
jianchun: nit: Move this `log_message` down one line, so that we always follow `print and log` order, and let `padding` calculation go together with consumption in print.

jianchun:
LGTM

digitalinfinity:
Thanks @Jianchun. Yup, took all feedback except for the mkdir one- the assumption is that if the user explicitly passed in the path, we assume it to be valid, we don't do anything to make it valid.

PR #1047, opened 2016-05-26
cross platform: fix display name for dst

obastemur:
issue: previously implemented name caching was adding an additional character to
the end of display name.

extra: consumer for GetStandardName and GetDaylightName asks only for one of them
while the previous implementation was copying both. Keep timeZoneInfo as
is and bring the result back from that.


digitalinfinity:
:shipit:

PR #1049, opened 2016-05-26
Fix build errors using clang on windows with JIT

curtisman:
Just to flush out compiler errors when using clang.
Build and link clean for x64 debug only.
No guarentees on the resulting binaries.


curtisman:
@digitalinfinity @jianchun @obastemur please code review.

lib/Common/DataStructures/BaseDictionary.h
```diff
@@ -1239,7 +1239,7 @@ namespace JsUtil
         public:
             EntryIterator(TDictionary &dictionary) : Base(dictionary, 0), entryCount(dictionary.count)
             {
-                if(IsValid() && IsFreeEntry(entries[entryIndex]))
+                if(IsValid() && IsFreeEntry(__super::entries[entryIndex]))
```
jianchun: I recall last time `__super` wasn't supported and we used base class name directly (`Base::` here).

jianchun:
LGTM

lib/Common/DataStructures/BaseDictionary.h
obastemur: IIRC this is not an issue with clang as a part of MS compatibility and we already use it project wide.

lib/Runtime/Language/ValueType.h
```diff
@@ -354,6 +354,21 @@ struct DefaultComparer<ValueType> : public ValueTypeComparer
 {
 };
 
+__forceinline ValueType ValueType::Merge(const ValueType other) const
```
obastemur: nit: please use `_ALWAYSINLINE` instead

digitalinfinity:
LGTM

lib/Runtime/Language/ValueType.h
curtisman: In Linux build, CommonPal.h already defined __forceinline:
`#if defined(__GNUC__) || defined(__clang__)
#define __forceinline inline __attribute__((always_inline))
#else
#define __forceinline inline
#endif`

_ALWAYSINLINE are for template implementation only, because clang don't like `inline` on them

PR #1052, opened 2016-05-26
Stack walker refactor

rajatd:


rajatd:
@dotnet-bot  test Windows x86_release please

rajatd:
@LouisLaf @abchatra , could you guys take a look?

lib/Runtime/Language/JavascriptStackWalker.cpp
```diff
@@ -298,147 +298,211 @@ namespace Js
...
+    bool JavascriptStackWalker::TryGetByteCodeOffsetFromInterpreterFrame(uint32& offset) const
+    {
+#if ENABLE_NATIVE_CODEGEN
+        if (this->lastInternalFrameInfo.codeAddress != nullptr)
+        {
+            return false;
+        }
+#endif
+        offset = this->interpreterFrame->GetReader()->GetCurrentOffset();
+        if (offset == 0)
```
farzonl: since there is nothing in this if condition maybe  make it offset != 0 and move the else block in?

lib/Runtime/Language/JavascriptStackWalker.cpp
```diff
@@ -606,15 +698,15 @@ namespace Js
...
+                    this->SetCachedInternalFrameInfo(InternalFrameType_LoopBody, InternalFrameType_LoopBody, true /*inlinedFramesOnStack*/);
                     isJavascriptFrame = false;
                 }
             }
 
             return true;
         }
 #endif
-
+        this->inlinedFramesOnStack = false;
```
farzonl: nit: you tend to alternate from adding is before a boolean to not doing it at all. I usually don't care about consistency issues like that, but you also do pointer compares in your conditional and assert statements. without the naming consistency  or equality checks in the conditionals\asserts it is hard to follow when something is a bool vs something else. I don't have the power of type inference over github so naming consistency would be very much appreciated.

lib/Runtime/Language/JavascriptStackWalker.cpp
rajatd: My intention was to keep the original comment about debugger and OOS for the case when offset is 0. Even by making the condition as offset!=0 I'd want to keep the comment

dilijev:
@dotnet-bot 

abchatra:
:shipit:

lib/Runtime/Language/JavascriptStackWalker.cpp
```diff
@@ -769,45 +861,33 @@ namespace Js
 
     bool JavascriptStackWalker::CheckJavascriptFrame(bool includeInlineFrames)
```
abchatra: I like the cleanup in this function.

lib/Runtime/Language/JavascriptStackWalker.cpp
rajatd: Which class member/local are you talking about? All the new members I have added don't have the 'is' prefix.

---
In reply to: [64985811](https://github.com/Microsoft/ChakraCore/pull/1052#discussion_r64985811) [](ancestors = 64985811)

rajatd: @farzonl , could you point me to the said inconsistency?

farzonl: in the section of code highlighted by github we have two bools isJavascriptFrame and inlinedFramesOnStack. Typically in the codebase if you want to let people know that your new variable is a boolean  you have prefix it with an is**name** or a has***condition*** . Some other code bases prefix with b. Anyways back on point, Some one else in this very function has established a precedent of using said is** naming convention. when you name a bool inlinedFramesOnStack you break the naming convention. my initial thought when reading that name was your new variable was a pointer until i saw you assign it to a bool, hence the nit. Really its not a big deal just thought I would point it out.

rajatd: I see, ok. Take a look at the latest commit.

farzonl: much better, thanks!

lib/Runtime/Language/JavascriptStackWalker.cpp
```diff
@@ -298,147 +298,212 @@ namespace Js
...
+
+    bool JavascriptStackWalker::InlinedFramesBeingWalked() const
+    {
+        if (lastInternalFrameInfo.codeAddress)
+        {
+            return false;
         }
 
-        return 0;
+        return inlinedFramesBeingWalked;
```
farzonl: did you miss a case here?

farzonl:
LGTM

dilijev:
@rajatd Linux builds now passing, thanks!

PR #1053, opened 2016-05-27
Minor simplification in BindPidRefsInScope

pleath:
De-templatize Parser::BindPidRefsInScope now that template parameter is unused. Delete the methods that dispatched to template functions. Merge the knopConstDecl case in Parser::BindPidRefs now that it is identical to knopVarDecl and knopLetDecl.


ianwjhalliday:
:+1:

PR #1054, opened 2016-05-27
Record in the profile that setting to __proto__ causes an implicit call

rajatd:
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/1054)


rajatd:
reopeing against release/1.2

PR #1055, opened 2016-05-27
Record in the profile that setting to __proto__ causes an implicit call

rajatd:
Setting to __proto__ property invokes a setter and changes the prototype.So, although PatchPut* populates the cache, the setter invalidates it (since it changes the prototype). PretendTrySetProperty looks at the inline cache type to update the cacheType on PropertyOperationInfo, which is used in populating the field info flags for this operation on the profile. Since the cache was invalidated, we dont get a match with either the type of the object with property or without it and the cacheType defaults to CacheType_None. This leads the profile info to say that this operation doesn't cause an accessor implicit call and JIT then doesn't kill live fields accross it and ends up putting a BailOutOnImplicitCalls
if there were live fields. This bailout always bails out.

Fixing it to also check the implicit call flags set by the set operation to indicate to the profile that this operation caused an implicit call or not.

### DIFF
--- a/lib/Runtime/Language/ProfilingHelpers.cpp
+++ b/lib/Runtime/Language/ProfilingHelpers.cpp
@@ -1076,6 +1076,10 @@ namespace Js
                     &operationInfo,
                     &propertyValueInfo))
             {
+                ThreadContext* threadContext = scriptContext->GetThreadContext();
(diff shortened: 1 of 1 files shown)


rajatd:
@pleath @LouisLaf  please review

rajatd:
Ping..

LouisLaf:
:shipit:

PR #1056, opened 2016-05-27
cross platform: add edge date cases support

obastemur:
This PR improves xplat date implementation with edge cases support.
Now test cases for Date passes on Ubuntu / Debian

 - tested against VSOFull and its test cases
 - no perf difference on Windows measured

Update:

Some details on what's been done with this update.

  - Linux implementation no longer converts from Windows SYSTEMTIME.
    ChakraCore's internal YMD structure was a good fit between tm and
    SYSTEMTIME. Besides, new implementation spends less time there.

  - Now xplat handles out of range year cases ( <1900 or > 2100) and
    Feb28/29 + 1 (see test file under Date)

  - ChakraCore's common date utilities consider the start year is epoch.
    Besides, system time and clock is based on epoch too. However, the C
    api that calculates the time zone offset, and DST in case of leap year
    depends on broken-out time starting from 1900.
    We handle the leap year differences.

  - On xplat we need exact date to grab right tm_zone information instead
    of caching approach on Windows. Windows API can bring back both DTS and
    STD abbrv with a single call.

### DIFF
--- a/lib/Runtime/PlatformAgnostic/Platform/Linux/DateTime.cpp
+++ b/lib/Runtime/PlatformAgnostic/Platform/Linux/DateTime.cpp
@@ -13,91 +13,142 @@ namespace PlatformAgnostic
 namespace DateTime
 {
 
-    #define updatePeriod 1000
-
-    const WCHAR *Utility::GetStandardName(size_t *nameLength) {
-        data.UpdateTimeZoneInfo();
-        *nameLength = data.standardZoneNameLength;
-        return data.standardZoneName;
-    }
-
-    const WCHAR *Utility::GetDaylightName(size_t *nameLength) {
-        // We have an abbreviated standard or daylight name
-        // based on the date and time zone we are in.
-        return GetStandardName(nameLength);
-    }
-
-    void UtilityPlatformData::UpdateTimeZoneInfo()
+    static inline bool IsLeap(const int year)
     {
-        uint32 tickCount = GetTickCount();
-        if (tickCount - lastTimeZoneUpdateTickCount > updatePeriod)
+        return (0 == (year & 3)) && (0 != (year % 100) || 0 == (year % 400));
+    }
+
+    // Windows DateTime implementation normalizes the year beyond <1900 >2100
+    // mktime etc. broken-out time bases 1900
+    static inline int NormalizeYMDYear(const int base_year)
+    {
+        if (base_year < -2100)
         {
-            time_t gtime = time(NULL);
-            struct tm ltime;
+            return 2100;
+        }
+        else if (base_year < -1900)
+        {
+            return base_year * -1;
+        }
 
-            tzset();
-            localtime_r(&gtime, &ltime);
+        return base_year;
+    }
 
-            standardZoneNameLength = strlen(ltime.tm_zone);
+    static inline int UpdateToYMDYear(const int base_year, const struct tm *time)
+    {
+        int year = time->tm_year;
 
-        #if defined(WCHAR_IS_CHAR16_T)
-            for(int i = 0; i < standardZoneNameLength; i++) {
-                standardZoneName[i] = (char16_t)ltime.tm_zone[i];
-            }
-        #elif defined(WCHAR_IS_WCHAR_T)
-            mbstowcs( (wchar_t*) standardZoneName, ltime.tm_zone,
-                      sizeof(standardZoneName) / sizeof(WCHAR));
-        #else
-        #error "WCHAR should be either wchar_t or char16_t"
-        #endif
+        if (base_year < -2100)
+        {
+            const int diff = year - 2100;
+            year = abs(base_year) - diff;
+        }
 
-            standardZoneName[standardZoneNameLength] = (WCHAR)0;
-            lastTimeZoneUpdateTickCount = tickCount;
+        if (base_year < -1900)
+        {
+            year *= -1;
+        }
+
+        return year;
+    }
+
+    static void YMD_TO_TM(const YMD *ymd, struct tm *time, bool *leap_added)
+    {
+        time->tm_year = NormalizeYMDYear(ymd->year);
+        time->tm_mon = ymd->mon;
+        time->tm_wday = ymd->wday;
+        time->tm_mday = ymd->mday;
+        int t = ymd->time;
+        t /= DateTimeTicks_PerSecond; // discard ms
+        time->tm_sec = t % 60;
+        t /= 60;
+        time->tm_min = t % 60;
+        t /= 60;
+        time->tm_hour = t;
+
+        // mktime etc. broken-out time accepts 1900 as a start year while epoch is 1970
+        // temporarily add a calendar day for leap pass
+        bool leap_year = IsLeap(time->tm_year);
+        *leap_added = false;
+        if (ymd->yday == 60 && leap_year) {
+            time->tm_mday++;
+            *leap_added = true;
         }
     }
 
-    // DateTimeHelper ******
-    static inline void TM_TO_SYSTIME(struct tm *time, SYSTEMTIME *sysTime)
+    static void TM_TO_YMD(const struct tm *time, YMD *ymd, const bool leap_added, const int base_year)
     {
-        sysTime->wYear = time->tm_year + 1900;
-        sysTime->wMonth = time->tm_mon + 1;
-        sysTime->wDayOfWeek = time->tm_wday;
-        sysTime->wDay = time->tm_mday;
-        sysTime->wHour = time->tm_hour;
-        sysTime->wMinute = time->tm_min;
+        ymd->year = UpdateToYMDYear(base_year, time);
+        ymd->mon = time->tm_mon;
+        ymd->mday = time->tm_mday;
 
-        // C99 tm_sec value is between 0 and 60. (1 leap second.)
-        // Discard leap second.
-        // In any case, this is a representation of a wallclock time.
-        // It can jump forwards and backwards.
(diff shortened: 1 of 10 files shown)


obastemur:
Ping @digitalinfinity @jianchun @curtisman

lib/Runtime/PlatformAgnostic/Platform/Linux/DateTime.cpp
```diff
@@ -13,155 +13,219 @@ namespace PlatformAgnostic
...
+    static inline bool IsLeap(const int year)
+    {
+        return (0 == (year & 3)) && (0 != (year % 100) || 0 == (year % 400));
     }
 
-    const WCHAR *Utility::GetDaylightName(size_t *nameLength) {
-        // We have an abbreviated standard or daylight name
-        // based on the date and time zone we are in.
-        return GetStandardName(nameLength);
+    static inline int NormalizeYMDYear(const int base_year)
```
digitalinfinity: Can you add some comments on what this method is for? Why 2100 or 1900?

lib/Runtime/PlatformAgnostic/Platform/Linux/DateTime.cpp
```diff
@@ -13,155 +13,219 @@ namespace PlatformAgnostic
...
 
-    #define updatePeriod 1000
-
-    const WCHAR *Utility::GetStandardName(size_t *nameLength) {
-        data.UpdateTimeZoneInfo();
-        *nameLength = data.standardZoneNameLength;
-        return data.standardZoneName;
+    static inline bool IsLeap(const int year)
+    {
+        return (0 == (year & 3)) && (0 != (year % 100) || 0 == (year % 400));
```
digitalinfinity: Instead of year & 3, can you do year % 4? It makes it more readable and the compiler should take care of optimizing it to & 3

lib/Runtime/PlatformAgnostic/Platform/Linux/DateTime.cpp
```diff
@@ -13,155 +13,219 @@ namespace PlatformAgnostic
...
 
-            tzset();
-            localtime_r(&gtime, &ltime);
+        return year;
+    }
 
-            standardZoneNameLength = strlen(ltime.tm_zone);
+    static void YMD_TO_TM(const YMD *ymd, struct tm *time, bool *leap_added)
+    {
+        time->tm_year = NormalizeYMDYear(ymd->year);
```
digitalinfinity: Based on the usage of this method, would YMDYearToTmYear (and the inverse) be better names for these methods?

lib/Runtime/PlatformAgnostic/Platform/Linux/DateTime.cpp
```diff
@@ -13,155 +13,219 @@ namespace PlatformAgnostic
...
-        // It can jump forwards and backwards.
-
-        sysTime->wSecond = time->tm_sec % 60;
+        *length = strlen(tm_zone);
+
+#if defined(WCHAR_IS_CHAR16_T)
+        for(int i = 0; i < *length; i++) {
+            wstr[i] = (char16_t)tm_zone[i];
+        }
+#elif defined(WCHAR_IS_WCHAR_T)
```
digitalinfinity: Can we get rid of this? Should be safe to assume that on non-Win32 platforms, wchar is char16?

lib/Runtime/PlatformAgnostic/Platform/Linux/DateTime.cpp
```diff
@@ -13,155 +13,219 @@ namespace PlatformAgnostic
...
+        wstr[*length] = (WCHAR)0;
     }
 
-    #define MIN_ZERO(value) value < 0 ? 0 : value
+    const WCHAR *Utility::GetStandardName(size_t *nameLength, const DateTime::YMD *ymd) {
+        AssertMsg(ymd != NULL, "xplat needs DateTime::YMD is defined for this call");
+        struct tm time_tm;
+        bool leap_added;
+        YMD_TO_TM(ymd, &time_tm, &leap_added);
+        mktime(&time_tm); // get zone name for the given date
```
digitalinfinity: Curious- does this use historic timezone information to get the timezone name? If you didn't pass in ymd, and instead called mktime with time_tm set to now, would you get the same answer as what the Windows version returns? Because it looks like the Windows version will return the timezone name based on the rules for the current time, whereas this might return it based on historic information for the passed in date?

lib/Runtime/PlatformAgnostic/Platform/Linux/DateTime.cpp
obastemur: YMD doesn't have a special year though. These methods normalize the abnormal dates temporarily for system calls

lib/Runtime/PlatformAgnostic/Platform/Linux/DateTime.cpp
obastemur: I couldn't see the same optimization with O2, Os or O3

lib/Runtime/PlatformAgnostic/Platform/Linux/DateTime.cpp
obastemur: What if the date you give to mktime is in standard time while the actual Date.toString within DST? There is no such issue in Windows.

lib/Runtime/PlatformAgnostic/Platform/Linux/DateTime.cpp
```diff
@@ -13,155 +13,219 @@ namespace PlatformAgnostic
...
+        int milliseconds = local->time % 1000;
 
         tzset();
+        int offset;
         time_t utime = timegm(&local_tm);
-        mktime(&local_tm); // we just want the gmt_off, don't care the result
-        utime -= local_tm.tm_gmtoff; // reverse UTC
+
+        // mktime min year 1900
+        if (local_tm.tm_year < 1900)
```
digitalinfinity: Nit: minor readability suggestion:
since local_tm is not used beyond this, you could probably just do the following:
if (local_tm.tm_year < 1900)
{
    local_tm.tm_year = 1900;
}

mktime(&local_tm);
offset = local_tm.tm_gmtoff;

What do you think?

lib/Runtime/PlatformAgnostic/Platform/Linux/DateTime.cpp
```diff
@@ -13,155 +13,219 @@ namespace PlatformAgnostic
...
-        time->tm_hour = sysTime->wHour;
-        time->tm_min = sysTime->wMinute;
-        time->tm_sec = sysTime->wSecond;
+    const WCHAR *Utility::GetDaylightName(size_t *nameLength, const DateTime::YMD *ymd) {
+        // xplat only gets the actual zone name for the given date
+        return GetStandardName(nameLength, ymd);
     }
 
-    static void SysLocalToUtc(SYSTEMTIME *local, SYSTEMTIME *utc)
+    static void SysLocalToUtc(YMD *local, YMD *utc)
```
digitalinfinity: Sys over here in SysLocalToUtc actually referes to SYSTEMTIME. Since you are now passing in YMD, you should change this to YmdLocalToUtc

obastemur:
@digitalinfinity thanks for the review. PR is updated.

lib/Runtime/PlatformAgnostic/Platform/Linux/DateTime.cpp
```diff
@@ -13,155 +13,219 @@ namespace PlatformAgnostic
...
         // put milliseconds back
-        local->wMilliseconds = utc->wMilliseconds;
-
-        isDaylightSavings = local_tm.tm_isdst;
+        local->time += milliseconds;
+
+        // ugly hack but;
+        // right in between dst pass
+        // we need mktime trick to get the correct dst
+        utc_tm.tm_isdst = 1;
```
digitalinfinity: Instead of doing this, can you just use the timezone variable that is set by tzset?

lib/Runtime/PlatformAgnostic/Platform/Linux/DateTime.cpp
digitalinfinity: Wow- that's really surprising- ok, leave it as is then but add the above information as a comment in the code so that no one accidentally removes it.

lib/Runtime/PlatformAgnostic/Platform/Linux/DateTime.cpp
obastemur: Actually we already get the above. However, that doesn't work with leap year dst pass and we don't know that moment until we trick the system. As mentioned with the comment, this is an ugly trick : (

lib/Runtime/PlatformAgnostic/Platform/Linux/DateTime.cpp
obastemur: Windows implementation FLeap does the same. I have carried this method from there.

digitalinfinity: Can we share the method then?

curtisman:
Overall it looks fine, but I didn't look at the logic too closely.

lib/Runtime/PlatformAgnostic/Platform/Linux/DateTime.cpp
obastemur: We could but keeping leap related arithmetic separate on xplat.

obastemur:
@digitalinfinity @curtisman Thanks for the review

PR #1061, opened 2016-05-28
xplat: switch to Utf8 Jsrt, implement -serialize

jianchun:
Implement 4 Utf8 variants of Jsrt Apis:
 - JsRunScriptUtf8
 - JsSerializeScriptUtf8
 - JsRunSerializedScriptUtf8
 - JsRunModuleUtf8

Change `ch` to use Utf8 script encoding and use the new Utf8 Jsrt Apis.

Refactor Narrow-Wide string converter and shared with runtime.
Refactor JsrtSourceHolder to support both wchar_t and utf8 loader.

Implement JavascriptFunction::DeferredDeserializeThunk for cross-platform.
Set major/minor version for xplat (needed by `-serialize`).

Turn on `-serialize` tests on cross-platform.

### DIFF
--- a/bin/ch/Helpers.cpp
+++ b/bin/ch/Helpers.cpp
@@ -3,79 +3,11 @@
 // Licensed under the MIT license. See LICENSE.txt file in the project root for full license information.
 //-------------------------------------------------------------------------------------------------------
 #include "stdafx.h"
-#include "Codex/Utf8Codex.h"
 
-///
-/// Use the codex library to encode a UTF16 string to UTF8.
-/// The caller is responsible for freeing the memory, which is allocated
-/// using malloc.
-/// The returned string is null terminated.
-///
-HRESULT Helpers::WideStringToNarrowDynamic(LPCWSTR sourceString, LPSTR* destStringPtr)
-{
-    size_t cchSourceString = wcslen(sourceString);
-    
-    if (cchSourceString >= MAXUINT32)
-    {
-        return E_OUTOFMEMORY;
-    }
-
-    size_t cbDestString = (cchSourceString + 1) * 3;
-
-    // Check for overflow- cbDestString should be >= cchSourceString
-    if (cbDestString < cchSourceString)
-    {
-        return E_OUTOFMEMORY;
-    }
-
-    utf8char_t* destString = (utf8char_t*)malloc(cbDestString);
-    if (destString == nullptr)
-    {
-        return E_OUTOFMEMORY;
-    }
-
-    size_t cbDecoded = utf8::EncodeIntoAndNullTerminate(destString, sourceString, (charcount_t) cchSourceString);
-    Assert(cbDecoded <= cbDestString);
-    static_assert(sizeof(utf8char_t) == sizeof(char), "Needs to be valid for cast");
-    *destStringPtr = (char*)destString;
-    return S_OK;
-}
-
-///
-/// Use the codex library to encode a UTF8 string to UTF16.
-/// The caller is responsible for freeing the memory, which is allocated
-/// using malloc.
-/// The returned string is null terminated.
-///
-HRESULT Helpers::NarrowStringToWideDynamic(LPCSTR sourceString, LPWSTR* destStringPtr)
-{
-    size_t cbSourceString = strlen(sourceString);
-    charcount_t cchDestString = utf8::ByteIndexIntoCharacterIndex((LPCUTF8) sourceString, cbSourceString);
-    size_t cbDestString = (cchDestString + 1) * sizeof(WCHAR);
-    
-    // Check for overflow- cbDestString should be >= cchSourceString
-    if (cbDestString < cchDestString)
-    {
-        return E_OUTOFMEMORY;
-    }
-
-    WCHAR* destString = (WCHAR*)malloc(cbDestString);
-    if (destString == nullptr)
-    {
-        return E_OUTOFMEMORY;
-    }
-
-    utf8::DecodeIntoAndNullTerminate(destString, (LPCUTF8) sourceString, cchDestString);
-    static_assert(sizeof(utf8char_t) == sizeof(char), "Needs to be valid for cast");
-    *destStringPtr = destString;
-    return S_OK;
-}
-
-HRESULT Helpers::LoadScriptFromFile(LPCSTR filename, LPCWSTR& contents, bool* isUtf8Out, LPCWSTR* contentsRawOut, UINT* lengthBytesOut, bool printFileOpenError)
+HRESULT Helpers::LoadScriptFromFile(LPCSTR filename, LPCSTR& contents, bool* isUtf8Out /*= nullptr*/, UINT* lengthBytesOut /*= nullptr*/)
 {
     HRESULT hr = S_OK;
-    LPCWSTR contentsRaw = nullptr;
-    byte * pRawBytes = nullptr;
+    BYTE * pRawBytes = nullptr;
     UINT lengthBytes = 0;
     bool isUtf8 = false;
     contents = nullptr;
@@ -87,34 +19,27 @@ HRESULT Helpers::LoadScriptFromFile(LPCSTR filename, LPCWSTR& contents, bool* is
     //
     if (fopen_s(&file, filename, "rb") != 0)
     {
-        if (printFileOpenError)
-        {
 #ifdef _WIN32
-            DWORD lastError = GetLastError();
-            char16 wszBuff[512];
-            fprintf(stderr, "Error in opening file '%s' ", filename);
-            wszBuff[0] = 0;
-            if (FormatMessage(FORMAT_MESSAGE_FROM_SYSTEM,
-                nullptr,
-                lastError,
-                0,
-                wszBuff,
-                _countof(wszBuff),
-                nullptr))
-            {
-                fwprintf(stderr, _u(": %s"), wszBuff);
-            }
-            fwprintf(stderr, _u("\n"));
-#elif defined(_POSIX_VERSION)
-            fprintf(stderr, "Error in opening file: ");
-            perror(filename);
-#endif            
-            IfFailGo(E_FAIL);
-        }
-        else
+        DWORD lastError = GetLastError();
+        char16 wszBuff[512];
+        fprintf(stderr, "Error in opening file '%s' ", filename);
+        wszBuff[0] = 0;
+        if (FormatMessage(FORMAT_MESSAGE_FROM_SYSTEM,
+            nullptr,
+            lastError,
+            0,
+            wszBuff,
+            _countof(wszBuff),
+            nullptr))
         {
-            return E_FAIL;
+            fwprintf(stderr, _u(": %s"), wszBuff);
         }
+        fwprintf(stderr, _u("\n"));
+#elif defined(_POSIX_VERSION)
+        fprintf(stderr, "Error in opening file: ");
+        perror(filename);
+#endif
+        IfFailGo(E_FAIL);
     }
 
     //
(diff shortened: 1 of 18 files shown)


jianchun:
@digitalinfinity @obastemur @curtisman @Yongqu Please help review, thanks!

~15 more tests pass (`-serialize` tests).

lib/Jsrt/Jsrt.cpp
```diff
@@ -2475,6 +2476,22 @@ JsErrorCode RunScriptCore(const wchar_t *script, JsSourceContext sourceContext,
...
+    utf8::NarrowToWide url((LPCSTR)sourceUrl);
+    if (!url)
+    {
+        return JsErrorOutOfMemory;
+    }
+
+    return RunScriptCore(reinterpret_cast<const byte*>(script), strlen(script), LoadScriptFlag_Utf8Source, sourceContext, url, parseOnly, parseAttributes, isSourceModule, result);
+}
+
+JsErrorCode RunScriptCore(const wchar_t *script, JsSourceContext sourceContext, const wchar_t *sourceUrl, bool parseOnly, JsParseScriptAttributes parseAttributes, bool isSourceModule, JsValueRef *result)
```
obastemur: wchar_t ?

lib/Jsrt/JsrtSourceHolder.cpp
```diff
@@ -7,60 +7,105 @@
...
-        *scriptLength = 0;
+    };
 
-        if (utf8AllocLength != nullptr)
+#ifdef _WIN32  // JsSerializedScriptLoadSourceCallback is WIN32 only
+    template <>
+    class JsrtSourceHolderPolicy<JsSerializedScriptLoadSourceCallback>
+    {
+    public:
+        typedef wchar_t TLoadCharType;
```
obastemur: wchar_t here again. Can we stop using wchar_t explicitly ?

bin/ch/Helpers.cpp
```diff
@@ -133,9 +58,9 @@ HRESULT Helpers::LoadScriptFromFile(LPCSTR filename, LPCWSTR& contents, bool* is
     //
     // Read the entire content as a binary block.
     //
-    fread((void*)contentsRaw, sizeof(char), lengthBytes, file);
+    fread(pRawBytes, sizeof(BYTE), lengthBytes, file);
     fclose(file);
-    *(WCHAR*)((byte*)contentsRaw + lengthBytes) = _u('\0'); // Null terminate it. Could be LPCWSTR.
+    *reinterpret_cast<WCHAR*>(pRawBytes + lengthBytes) = 0; // Null terminate it. Could be UTF16
```
obastemur: Couple of issues here. 1) char16_t size is not guaranteed to be X bytes. 2) Depending on the size of char16 and lengthBytes, this code behaves differently 3) reinterpret_cast, when especially used for aliasing, is suggested to be read only.

lib/Jsrt/Jsrt.cpp
jianchun: This is intentional, corresponding to public wchar_t version of JSRT APIs. We may need a cleanup to move all wchar_t versions under `#ifdef _WIN32`.

lib/Jsrt/JsrtSourceHolder.cpp
jianchun: Same as above.

bin/ch/Helpers.cpp
jianchun: Didn't understand these, could you elaborate more? 1) This doesn't depend on how many bytes char16_t is (however we only support up to WCHAR - 2 bytes). 2) What behaves differently, and why that matters? 3) Never heard of it, why is reinterpret_cast for read only?

I changed a variable name and cast style, otherwise this is same as old code. The purpose is to null terminate it no matter the loaded source is utf8 or utf16 (and we only support utf16 LE).

obastemur: I haven't tried yet but most probably compiler already knows `pRawBytes` is  `sizeof(BYTE)`!! x (lengthBytes) away from it's starting point. So it does the rest hence it wouldn't give trouble during the runtime. However comparing to old code, it is miss-leading. Adding `(BYTE*)` right before `pRawBytes` would make it more clear.

Old code has C style reinterpret_cast and yours C++. IMO, in both cases, I'm having trouble when I see char memory is being reinterpret cast to char16 (Endiannes ?) even though there are couple of general `harmless` usages like this. 

So it is `okay` to put null termination in a single line but IMO making it clear would give less trouble to reader.

> however we only support up to WCHAR - 2 bytes
I guess you mean (at least) 2 bytes ? Since, there is no guarantee that it will be 2 bytes.

lib/Jsrt/Jsrt.cpp
```diff
@@ -2746,3 +2762,47 @@ CHAKRA_API JsRunSerializedScriptWithCallback(_In_ JsSerializedScriptLoadSourceCa
     return RunSerializedScriptCore<JsSerializedScriptLoadSourceCallback, JsSerializedScriptUnloadCallback>(nullptr, scriptLoadCallback, scriptUnloadCallback, buffer, sourceContext, sourceUrl, false, result);
 }
 #endif // _WIN32
+
+
+CHAKRA_API JsRunScriptUtf8(
```
agarwal-sandeep: Since these are new APIs shouldn't we take length of source as well so that we can allow embedded nulls in script. I understand that these APIs are just equivalent of win32 wchar_t versions but these are any how new APIs and we should be careful in what is the final shape otherwise we will have to write new set of APIs for embedded null scenarios in future.

agarwal-sandeep:
>     m_jsApiHooks.pfJsrtRunScript = (JsAPIHooks::JsrtRunScriptPtr)GetChakraCoreSymbol(library, "JsRunScript");

Is this used anymore?

---
Refers to: bin/ch/ChakraRtInterface.cpp:93 in 89a30e9. [](commit_id = 89a30e9f48443de055603f73c9767262234bb6e6, deletion_comment = False)

bin/ch/Helpers.cpp
jianchun: > I haven't tried yet but most probably compiler already knows pRawBytes is  sizeof(BYTE)!! x (lengthBytes) away from it's starting point. So it does the rest hence it wouldn't give trouble during the runtime. However comparing to old code, it is miss-leading. Adding (BYTE*) right before pRawBytes would make it more clear.

With the new code `pRawBytes` is of type `BYTE*`, thus it does not need cast. Casting to its own type `(BYTE*)pRawBytes` is unnecessary and redundant. I'm not sure why you describe it as uncertain ("haven't tried...probably"), `pRawBytes + lengthBytes` means &pRawBytes[lengthBytes]` and yes `lengthBytes * sizeof(BYTE)`, or `lengthBytes` bytes away from start.

The old code needs casting because `contentsRaw` was declared as `LPCWSTR contentsRaw`.

> So it is okay to put null termination in a single line but IMO making it clear would give less trouble to reader.

What do you have in mind to make it clear? I agree with you this line is a bit tricky without understanding it is trying to null-terminate both Utf8 and Utf16 cases. I actually worried that if `lengthBytes` is odd and this WORD assigns to odd address and the instruction may fail, but tried out on Linux and it worked. (For context, this is test binary and the code isn't in product quality. As long as it works for our unit tests.)

> there is no guarantee that it will be 2 bytes.

You mean 'char16_t` isn't guaranteed to be 2-bytes? I can't believe... What's the use of `char16_t` if it doesn't guarantee to be `16 bits"?

lib/Jsrt/Jsrt.cpp
jianchun: Agreed, I thought of the same. There are other aspects of the APIs we may need to iterate on, e.g. lifetime contract of the script source.

jianchun:
> Is this used anymore?

Probably not. Not sure exactly how we are going to treat those supposedly WIN32 only APIs. We could delete these.

bin/ch/Helpers.cpp
obastemur: > You mean char16_t isn't guaranteed to be 2-bytes? I can't believe... What's the use of char16_t if it doesn't guarantee to be `16 bits"?

Logical answer: What if that particular platform min. 32 bits?

C++ International Standard: ...in particular, sizeof(bool), sizeof(char16_t),sizeof(char32_t), and sizeof(wchar_t) are implementation-defined. Guess why?

 http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2013/n3690.pdf

obastemur: > With the new code pRawBytes is of type BYTE*, thus it does not need cast. 

Right but that applies to many other things with `reinterpret_cast` (either they are okay or totally wrong)

> What do you have in mind to make it clear?

Although I'm always in favor of C++ style casts, old code is IMO more explicit. Either way, what's done there is nothing but trick**. When using tricks, IMO, it is nice to warn reader.

** i.e. there wouldn't be a reinterpret_cast from char16 string to char str. (vice versa)

jianchun: > Logical answer: What if that particular platform min. 32 bits?
> 
> C++ International Standard: ...in particular, sizeof(bool), sizeof(char16_t),sizeof(char32_t), and sizeof(wchar_t) are implementation-defined. Guess why?
> 
> http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2013/n3690.pdf

Thanks for the link, great to know!

lib/Runtime/Base/ScriptContext.cpp
```diff
@@ -1681,7 +1693,7 @@ if (!sourceList)
         }
 
         ParseNodePtr parseTree;
-        if ((loadScriptFlag & LoadScriptFlag_Utf8Source) == LoadScriptFlag_Utf8Source)
```
digitalinfinity: What happens to this flag now? Should we use it at least for assert purposes?

bin/ch/ch.cpp
```diff
@@ -199,31 +199,41 @@ static void CALLBACK PromiseContinuationCallback(JsValueRef task, void *callback
...
-HRESULT RunScript(const char* fileName, LPCWSTR fileContents, BYTE *bcBuffer, char *fullPath)
+static bool CHAKRA_CALLBACK DummyJsSerializedScriptLoadUtf8Source(_In_ JsSourceContext sourceContext, _Outptr_result_z_ const char** scriptBuffer)
+{
+    *scriptBuffer = reinterpret_cast<const char*>(sourceContext);
+    return true;
+}
+
+static void CHAKRA_CALLBACK DummyJsSerializedScriptUnload(_In_ JsSourceContext sourceContext)
+{
+    free(reinterpret_cast<void*>(sourceContext));
```
digitalinfinity: Can you add some comments on the lifetime management here? I understand what is going on but it'd be more clear with comments here to others (like, why we're calling free here when the Load callback counterpart is not doing any malloc etc).

lib/Runtime/Base/ScriptContext.cpp
jianchun: > What happens to this flag now? Should we use it at least for assert purposes?

I consolidated it into "utf8FromExternal" (which means it is utf8 and fromExternal). As discussed off line, this may be temporary, depending on how we are going to treat CESU8. (Can we drop CESU8?)

Yes I can add an assertion.

bin/ch/ch.cpp
jianchun: Sure. (I was also wondering what we should do with the byte code buffer...)

digitalinfinity:
:shipit:

obastemur:
LGTM

obastemur:
@jianchun One last thing before merging this PR. `es6/unicode_6_identifiers.js` test fails even after this PR is merged. Is this PR planned to be fixing that issue ?

```
xxx/Github/ChakraCore/BuildLinux/debug/ch -WERExceptionSupport -ExtendedErrorStackForTestHost -maxInterpretCount:1 -maxSimpleJitRunCount:1 -bgjit- -ES6Unicode xxx/Github/ChakraCore/test/es6/unicode_6_identifiers.js
Output: (at line 0)
----------------------------
SyntaxError: Invalid character
----------------------------
Expected Output:
----------------------------
3
----------------------------
```

jianchun:
> es6/unicode_6_identifiers.js test fails even after this PR is merged

Looked off line, there is a Windows/Linux diff in `bool IsIdStart(codepoint_t codepoint)` implementation. @digitalinfinity will investigate.

obastemur:
> Looked off line, there is a Windows/Linux diff in bool IsIdStart(codepoint_t codepoint) implementation. @digitalinfinity will investigate.

Okay, based on the result from the investigation I will update #1078

PR #1065, opened 2016-05-31
cross-platform: Implement locale number formatter

obastemur:
- Format number to locale string using user's locale

 - Now CC generates consistent output matching with the user's locale cross platform

 - Enabled number.toLocaleString basic test. Previously we weren't
   testing the output of this function due to differences on varying
   Windows editions or CI settings.

 - Tested against VSOFull. Benchmarks are flat.

 - Updated couple of other test files for cross platform compatibility


lib/Common/PlatformAgnostic/Numbers.h
```diff
@@ -0,0 +1,51 @@
...
+            inline bool  HasLocaleThousands()        { return localeThousands != 0; }
+            inline WCHAR GetLocaleDecimal()          { return localeDecimal; }
+            inline WCHAR GetLocaleNegativeSign()     { return localeNegativeSign; }
+
+            inline bool  IsDecimalPoint(const WCHAR wc)
+                                                     { return wc == defaultDecimalDot
+                                                           || wc == defaultDecimalComma; }
+        };
+
+        static NumbersLocale numbersLocale;
```
curtisman: Can the locale change in the middle of execution?

lib/Runtime/PlatformAgnostic/Platform/Common/NumbersUtility.cpp
```diff
@@ -0,0 +1,235 @@
...
+        localeDecimal = (WCHAR) *(locale_format->decimal_point);
+        localeNegativeSign = (WCHAR) *(locale_format->negative_sign);
+
+        // set process/LC_NUMERIC back
+        setlocale(LC_NUMERIC, old_locale);
+    }
+
+    Utility::NumbersLocale Utility::numbersLocale;
+
+    size_t Utility::NumberToDefaultLocaleString(const WCHAR *number_string,
```
curtisman: I don't think this covers all the cases the GetNumberFormatEx would support.
(e.g. I can switch the standard digits character to Chinese characters).

My guess is that ICU have support for richer locale specific number formatting as well?
For now, this custom implementation probably is sufficient for the case ICU support is not included.
But for windows, we should stick to GetNumberFormatEx so that we don't regress functionality.

lib/Common/PlatformAgnostic/Numbers.h
obastemur: Sure it can but I'm not sure if we should consider them during the same process lifecycle?

lib/Runtime/PlatformAgnostic/Platform/Common/NumbersUtility.cpp
obastemur: Indeed ICU support Chinese and others too, however we can't assume ICU is embedded by default. So, we should use GetNumberFormatEx on Windows by default and ICU on other platforms whenever enabled?

obastemur: > e.g. I can switch the standard digits character to Chinese characters

Even if you would do that, would we have `Chinese characters` from number.toString ?

obastemur:
@digitalinfinity @jianchun ping!

test/VT_DATE/getvardate.js
```diff
@@ -67,7 +67,15 @@ vdate = new Date(1234567890123).getVarDate();
 safeCall(function() { writeLine(vdate ? true : false); });
 safeCall(function() { writeLine([1, 2].indexOf(2, vdate)); }); // valid only in version 3
 safeCall(function() { writeLine(parseInt("1", vdate)); });
-safeCall(function() { writeLine([1, vdate, 2].toLocaleString()); });
+
+// normalize Old GetNumberFormatEx exception
+safeCall(function() {
+    var result = [1, vdate, 2].toLocaleString();
+    if (result == "1, [object Object], 2") {
```
digitalinfinity: Is this needed for the Linux case?

test/Array/toLocaleString.js
```diff
@@ -48,7 +48,15 @@ scenario("Array: normal");
 var o = [
     0, 1.23, NaN, true, "abc", {}, [], [0, 1, 2]
 ];
-echo(Array.prototype.toLocaleString.apply(o));
+
+// 0.toLocaleString() is supposed to be 0. However our baseline has an exception.
+// So, do not break it while supporting cross platform locale output
+var output = Array.prototype.toLocaleString.apply(o);
+if ( output == "0, 1.23, NaN, true, abc, [object Object], , 0, 1, 2" ) {
+    echo("0.00, 1.23, NaN, true, abc, [object Object], , 0.00, 1.00, 2.00");
```
digitalinfinity: Can you clarify what's going on here? If it's supposed to be 0, is there a bug in the current Windows implementation?

test/VT_DATE/getvardate.js
obastemur: Yes. That's correct output actually but we test against what we get from WinAPI

digitalinfinity: Interesting- do we need to fix the Windows case or are you leaving it here for compat? @abchatra- thoughts?

---
In reply to: [65751957](https://github.com/Microsoft/ChakraCore/pull/1065#discussion_r65751957) [](ancestors = 65751957)

lib/Runtime/PlatformAgnostic/Platform/Windows/NumbersUtility.cpp
```diff
@@ -0,0 +1,34 @@
...
+namespace PlatformAgnostic
+{
+namespace Numbers
+{
+    size_t Utility::NumberToDefaultLocaleString(const WCHAR *number_string,
+                                                const size_t length,
+                                                WCHAR *buffer,
+                                                const size_t pre_allocated_buffer_size)
+    {
+        size_t count = GetNumberFormatEx(LOCALE_NAME_USER_DEFAULT, 0,
```
digitalinfinity: GetNumberFormatEx returns an int, not a size_t. I'd change this to size_t and assert for count >= 0 before using it as a size_t. I dont know why we were checking for <= 0 before your change since the docs say its either 0 or greater.

lib/Runtime/Library/JavascriptNumber.cpp
```diff
@@ -1032,49 +1034,42 @@ namespace Js
...
             {
                 pszRes = pszToBeFreed = HeapNewArray(char16, count);
-            }
-            else
-            {
-                pszRes = szRes;
-            }
 
-            int newCount = GetNumberFormatEx(LOCALE_NAME_USER_DEFAULT, 0, szValue, NULL, pszRes, count);
+                count = Numbers::Utility::NumberToDefaultLocaleString(szValue, szLength, pszRes, bufSize);
```
digitalinfinity: Should you be calling this with count instead of bufSize since count > bufSize, and we just allocated a larger array.

Also, in the case where count > bufSize, before we did a max of two calls to GetNumberFormatEx. Now it looks like we could do 4? That is, the first call to NumberToDefaultString would do it with 0, and then bufSize, and then the second would again do it with 0, and then with count. Worth optimizing to match the earlier performance?

lib/Common/PlatformAgnostic/Numbers.h
```diff
@@ -0,0 +1,55 @@
...
+//-------------------------------------------------------------------------------------------------------
+
+#ifndef RUNTIME_PLATFORM_AGNOSTIC_COMMON_NUMBERS
+#define RUNTIME_PLATFORM_AGNOSTIC_COMMON_NUMBERS
+
+namespace PlatformAgnostic
+{
+namespace Numbers
+{
+    class Utility
```
digitalinfinity: Just a nit- why have this Utility class? PlatformAgnostic::Numbers::NumberToDefaultLocaleString seems fine

lib/Runtime/PlatformAgnostic/Platform/Linux/NumbersUtility.cpp
```diff
@@ -0,0 +1,267 @@
+//-------------------------------------------------------------------------------------------------------
+// Copyright (C) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE.txt file in the project root for full license information.
+//-------------------------------------------------------------------------------------------------------
+
+#include "Common.h"
+#include "ChakraPlatform.h"
+#include <locale.h>
+
+namespace PlatformAgnostic
```
digitalinfinity: This is the non-ICU case right? Should we wrap this in !ENABLE_GLOBALIZATION?

lib/Runtime/PlatformAgnostic/Platform/Linux/NumbersUtility.cpp
```diff
@@ -0,0 +1,267 @@
...
+{
+namespace Numbers
+{
+    #define BufSize 256
+
+    template<class T, class N=T>
```
digitalinfinity: When is N passed in? This looks like just memcpy, no? Can you use memcpy instead?

lib/Runtime/PlatformAgnostic/Platform/Linux/NumbersUtility.cpp
```diff
@@ -0,0 +1,267 @@
...
+    void ByteCopy(const T *from, N *to, const size_t max_length)
+    {
+        size_t length = 0;
+        for(;length < max_length && (to[length] = static_cast<N>(from[length])) != 0;
```
digitalinfinity: This is a little hard to read- can you rewrite this to be a normal loop (with the copy done in the body)?

lib/Runtime/PlatformAgnostic/Platform/Linux/NumbersUtility.cpp
```diff
@@ -0,0 +1,267 @@
...
+            length++);
+        if (length == max_length)
+        {
+            to[length-1] = N(0);
+        }
+    }
+
+    static const WCHAR number_zero        = WCHAR('0');
```
digitalinfinity: Nit: I thought our preference was to use char16 now?

lib/Runtime/PlatformAgnostic/Platform/Linux/NumbersUtility.cpp
```diff
@@ -0,0 +1,267 @@
...
+        defaultDecimalDot = WCHAR('.');
+        defaultDecimalComma = WCHAR(',');
+        localeNegativeSign = WCHAR('-');
+        
+        char buffer[BufSize];
+        char *old_locale = setlocale(LC_NUMERIC, NULL);
+
+        if(old_locale != NULL)
+        {
+            ByteCopy<char>(old_locale, buffer, BufSize);
```
digitalinfinity: Should be fine but would be nice to add an assert here that strlen(old_locale) < BufSize

lib/Runtime/PlatformAgnostic/Platform/Linux/NumbersUtility.cpp
```diff
@@ -0,0 +1,267 @@
...
+    }
+
+    static const WCHAR number_zero        = WCHAR('0');
+    static const WCHAR number_five        = WCHAR('5');
+    static const WCHAR number_nine        = WCHAR('9');
+
+    Utility::NumbersLocale::NumbersLocale()
+    {
+        maxDigitsAfterDecimals = 3;
+        defaultDecimalDot = WCHAR('.');
```
digitalinfinity: Nit: why not also set localeDecimal and localeThousands here, and then not worry about setting the default if localeconv returns null?

digitalinfinity:
:shipit:

lib/Runtime/PlatformAgnostic/Platform/Linux/NumbersUtility.cpp
obastemur: Could be but I think it is also clear this way.

lib/Runtime/PlatformAgnostic/Platform/Linux/NumbersUtility.cpp
obastemur: > Should be fine but would be nice to add an assert here that strlen(old_locale) < BufSize

No need since that's taken care of ByteCopy.

lib/Runtime/PlatformAgnostic/Platform/Linux/NumbersUtility.cpp
obastemur: Yes we do but everywhere else we use WCHAR.

lib/Runtime/PlatformAgnostic/Platform/Linux/NumbersUtility.cpp
obastemur: I prepared this method for later use since It's needed for ICU part. I had implemented that too but wanted to keep separate due to issues with ICU versions.

test/Array/toLocaleString.js
obastemur: I don't know what's the historical reason behind, didn't want to break the test. "0" for "0" is the right output.

test/VT_DATE/getvardate.js
obastemur: Left there for compat as mentioned above.

lib/Runtime/PlatformAgnostic/Platform/Windows/NumbersUtility.cpp
obastemur: I don't see a reason to carry an implementation without anything tangible. While deciding for the size_t return type, I have checked the cases and also documentation. On MSDN it clearly says the return value is >= 0.

lib/Runtime/Library/JavascriptNumber.cpp
obastemur: I'm not sure if I understood the 4 calls part. Perhaps you missed the if statement above. Agree on replacing bufSize to count and did that.

lib/Common/PlatformAgnostic/Numbers.h
obastemur: Sounds good.I had considered the possible number related methods after leaving PAL behind. We will need a utility interface anyway and this one was matching somehow. I don't have a strong opinion on this ?

lib/Runtime/PlatformAgnostic/Platform/Linux/NumbersUtility.cpp
obastemur: Yes and No. At the moment, we have some things from ICU, rest not. Although header file is common (hence the reason for ENABLE_GLOBALIZATION ifdef), I don't see a reason to compile non-ICU source file on ICU build.

PR #1066, opened 2016-05-31
Fixing bug when hoisting FromVar from the landing pad of an inner loop

rajatd:
When hoisting a FromVar out of a loop, we check whether it needs a bailout at its hoisted position. If it does and if doesn't already have a bailout, we insert the appropriate bailout on it, and then hoist it. If the FromVar was originally in the landing pad of an inner loop, we have to make sure that the bail out offset of the newly inserted bailout is that of the top of the loop.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/1066)


rajatd:
@pleath @LouisLaf, trivial change

pleath:
LGTM. Thanks.

PR #1067, opened 2016-05-31
Fix dev12 build

digitalinfinity:
The dev12 toolset was unhappy with the redeclaration of class static
constants, that was done to satisfy clang. Dev12 treats the second
declaration as a second definition and makes the linker unhappy.
Conditionally compiled out the redeclarations on dev12.

### DIFF
--- a/lib/Parser/CharSet.cpp
+++ b/lib/Parser/CharSet.cpp
@@ -1870,11 +1870,16 @@ namespace UnifiedRegex
     }
 #endif
 
-    const int  CharSetNode::directBits;
+    // The VS2013 linker treats this as a redefinition of an already
+    // defined constant and complains. So skip the declaration if we're compiling
+    // with VS2013 or below.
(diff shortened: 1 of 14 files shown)


lib/Parser/CharSet.cpp
```diff
@@ -1870,11 +1870,16 @@ namespace UnifiedRegex
     }
 #endif
 
-    const int  CharSetNode::directBits;
+	// The VS2013 linker treats this as a redefinition of an already
+	// defined constant and complains. So skip the declaration if we're compiling
+	// with VS2013 or below.
+#if !defined(_MSC_VER) || _MSC_VER >= 1900
+	const int  CharSetNode::directBits;
```
jianchun: tab?

jianchun:
LGTM (not-sure tab comment).

digitalinfinity:
Thanks! Fixed all the tabs in the comments

PR #1069, opened 2016-06-01
Fix implementation of GET_CURRENT_FRAME_ID on Linux

digitalinfinity:
GET_CURRENT_FRAME_ID is used by the stack walker on Linux to figure out
the start of the frame. However, the implementation was still using the
win32 x64 implementation which use the return address as the frame
id. This broke the stack walker on Linux which is based on the x86
one. Changed the x64 linux implementation to match the x86 implementation.

### DIFF
--- a/lib/Common/Common/GetCurrentFrameId.h
+++ b/lib/Common/Common/GetCurrentFrameId.h
@@ -21,8 +21,13 @@
 #define GET_CURRENT_FRAME_ID(f) \
     __asm { mov f, ebp }
 #elif defined(_M_X64)
+#ifdef _WIN32
 #define GET_CURRENT_FRAME_ID(f) \
     (f = _ReturnAddress())
+#else
+#define GET_CURRENT_FRAME_ID(f) \
+    (f = __builtin_frame_address(0))
+#endif
 #elif defined(_M_ARM)
 // ARM, like x86, uses the frame pointer rather than code address
 #define GET_CURRENT_FRAME_ID(f) \


obastemur:
:shipit:

lib/Common/Common/GetCurrentFrameId.h
```diff
@@ -21,8 +21,13 @@
 #define GET_CURRENT_FRAME_ID(f) \
     __asm { mov f, ebp }
 #elif defined(_M_X64)
+#ifdef _WIN32
 #define GET_CURRENT_FRAME_ID(f) \
     (f = _ReturnAddress())
```
obastemur: one small thing here. Can we fix _ReturnAddress instead?

jianchun: `_ReturnAddress` itself has nothing wrong. I'm guessing the issue here is it cannot be used as ...FRAME_ID... Not sure of the details. Maybe because we do comparisons on it, and return address doesn't work well (e.g. we don't yet have dynamic interpreter frame) to identify a frame?

lib/Common/Common/GetCurrentFrameId.h
```diff
@@ -21,8 +21,13 @@
...
+#else
+#define GET_CURRENT_FRAME_ID(f) \
+    __asm { mov f, rbp }
```
jianchun: Does `__builtin_frame_address(0)` work here, just curious?

jianchun:
LGTM

lib/Common/Common/GetCurrentFrameId.h
obastemur: > _ReturnAddress itself has nothing wrong.

Isn't it interesting that on one platform we use frame pointer and returnadress on another ?

obastemur: > Does __builtin_frame_address(0) work here, just curious?

Yes it does work.

jianchun: > Isn't it interesting that on one platform we use frame pointer and returnadress on another ?

The reason is currently on x64 cross-platform is not on parity with Windows yet. We are effectively using calling convention and stack walker implementation the same as Windows x86. We may change these when working on JIT.

digitalinfinity: Yup- JC is correct. There is no problem with the implementation of _ReturnAddress. The StackFrame API has two mechanisms with which the stack walker can initialize the frame- either by frame id or with the return address. The Windows AMD64 version of the stack frame initializes basedo n return address. However, we based the Linux stack walker on the Windows x86 one, where the frame id is not the return address but the value of ebp in that stack frame. We could change the stack frame implementation to use the return address instead but the simplest solution for now is to use the frame id. And JC's right, we might have to update this when we work on the JIT.

JC- I'll update to use __builtin_frame_address(0). Thanks for the review, guys!

PR #1070, opened 2016-06-01
MicroBuild v2: Enable CodeCoverage.

dilijev:
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/1070)


dilijev:
@Cellule @aneeshdk @ianwjhalliday @digitalinfinity @curtisman Please review?

Build/scripts/build.ps1
```diff
@@ -0,0 +1,93 @@
...
+    "/fl2 `"/flp2:logfile=${buildlogsPath}\build.${Env:BuildName}.err;errorsonly`"",
+    "/fl3 `"/flp3:logfile=${buildlogsPath}\build.${Env:BuildName}.wrn;warningsonly`"",
+    "/verbosity:normal"
+    ) -join " "
+
+$cleanTarget = "/t:`"Clean`""
+
+$subtypeParams = ""
+if (($subtype -eq "pogo") -and $pogocmd) {
+    # TODO
```
dilijev: Note to self: This hasn't been needed so far, but will probably be needed for changes to POGO builds. Should implement (not delete) this block.

Build/scripts/finalize_build.ps1
```diff
@@ -26,29 +22,96 @@ if (Test-Path $PogoFolder) {
...
+
+# Search for *.nuspec files and copy them to $metadataDir
+Get-ChildItem -Path (Join-Path $sourcesDir "Build") "*.nuspec" `
+    | % { Copy-Item -Verbose -Force $_.FullName $metadataDir }
+
+#
+# Copy POGO directory if present for this build
+#
+
+# TODO (doilij) REMOVE speculative POGO copy since all builds will be explicitly specified now
```
dilijev: Note to self: This is not exactly true. We should keep this block because it doesn't hurt.

Build/scripts/build.ps1
```diff
@@ -0,0 +1,93 @@
+#-------------------------------------------------------------------------------------------------------
+# Copyright (C) Microsoft. All rights reserved.
+# Licensed under the MIT license. See LICENSE.txt file in the project root for full license information.
+#-------------------------------------------------------------------------------------------------------
+
+# Build Script
```
dilijev: Note to self (and reviewers): I think it would be better to rename this file to something like "run_build.ps1" or "do_build.ps1" to reserve the name of "build.ps1" for a script which runs the entire build.

(After discussing offline with @Cellule it seems like it is more and more likely that a single script is going to be more or less where the MicroBuild definition goes - we're already converging to an entirely powershell-script-based build.)

Build/scripts/init_build.ps1
```diff
@@ -41,15 +99,7 @@ $CommitTime = Get-Date $CommitDateTime -Format yyMMdd.HHmm
 # TODO (doilij) refactor this into a method in a util script.
```
dilijev: Note to reviewers (and self): I've partially completed this refactoring but hadn't decided whether the block that extracts data from `$info` should also be refactored since that is done in two scripts as well.

Build/scripts/post_build.ps1
```diff
@@ -6,7 +6,7 @@
 # Post-Build Script
 #
 # Run this script after the build step to consume the artifacts produced by the build,
-# run tests, and process and deploy build logs.
+# run tests, and process and deploy build ogs.
```
dilijev: Whoops. I'll revert this.

Build/scripts/init_build.ps1
```diff
@@ -8,31 +8,89 @@
...
+        $BuildSubtype = $buildTypeSegments[2]
+    }
+
+    if ($BuildConfiguration -eq "codecoverage") {
+        $BuildConfiguration = "test" # codecoverage builds are actually "test" configuration
+        $BuildSubtype = "codecoverage" # keep information about codecoverage in the subtype
+        if (-not ($BuildSubtype -in @("pogo","codecoverage"))) {
+            Write-Error "Unsupported BuildSubtype: $BuildSubtype"
+        }
+    }
```
aneeshdk: Is this check really needed? You are assigning codecoverage to it just above.

aneeshdk:
LGTM. Thanks for implementing this.

Build/scripts/init_build.ps1
dilijev: Oh, thanks. I refactored this code and forgot to remove this check afterwards.

Cellule:
Reviewed 11 of 11 files at r1.
Review status: all files reviewed at latest revision, 8 unresolved discussions.

---

*[Build/scripts/build.ps1, line 87 \[r1\]](https://reviewable.io:443/reviews/microsoft/chakracore/1070#-KJCAxT7HawzXs73zt30:-KJCAxT7HawzXs73zt31:1393449180) ([raw file](https://github.com/microsoft/chakracore/blob/3d2a6cd98b0d57dd46edda52bc5ae27ef1fca031/Build/scripts/build.ps1#L87)):*
> ```PowerShell
> 
>     if ($clean) {
>         $cleanCommand = ". `"$msbuildExe`" $cleanTarget $defaultParams $loggingParams $subtypeParams"
> ```

I don't think you need to execute the clean command separately.
if you pass `/t:Clean` to the normal build command it should be fine no ?
If not, maybe `msbuild /t:Clean,Rebuild` ?

---

*[Build/scripts/finalize_build.ps1, line 55 \[r1\]](https://reviewable.io:443/reviews/microsoft/chakracore/1070#-KJCBln2JoriEeNRoKdP:-KJCBln3_eFQcBerkHpP:1688505626) ([raw file](https://github.com/microsoft/chakracore/blob/3d2a6cd98b0d57dd46edda52bc5ae27ef1fca031/Build/scripts/finalize_build.ps1#L55)):*
> ```PowerShell
> 
> $buildErrFile = Join-Path $buildLogsDropPath "build.${Env:BuildName}.err"
> $testSummaryFile = Join-Path $testLogsDropPath "summary.${Env:BuildName}.log"
> ```

Is it because we manually call msbuild that we need to do this check manually ?
Wouldn't the msbuild command have a different exit code if there was an error ?
If this approach works I don't mind, but it seems weird to rely on log content

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/microsoft/chakracore/1070)*

dilijev:
Review status: all files reviewed at latest revision, 8 unresolved discussions.

---

*[Build/scripts/build.ps1, line 87 \[r1\]](https://reviewable.io:443/reviews/microsoft/chakracore/1070#-KJCAxT7HawzXs73zt30:-KJCG0aznFwkSLjx2jzZ:-1897283101) ([raw file](https://github.com/microsoft/chakracore/blob/3d2a6cd98b0d57dd46edda52bc5ae27ef1fca031/Build/scripts/build.ps1#L87)):*
> I don't think you need to execute the clean command separately.
> if you pass `/t:Clean` to the normal build command it should be fine no ?
> If not, maybe `msbuild /t:Clean,Rebuild` ?
</details>
just doing /t:Clean only performs the Clean. Possibly, the /t:Clean,Rebuild would work but would that be somehow better than the current setup? Either way we need to capture the same logic in this file. Would the build be somehow faster if we only issue one msbuild command?

---

*[Build/scripts/finalize_build.ps1, line 55 \[r1\]](https://reviewable.io:443/reviews/microsoft/chakracore/1070#-KJCBln2JoriEeNRoKdP:-KJCGUQZRV5M_GRyfAV1:298232958) ([raw file](https://github.com/microsoft/chakracore/blob/3d2a6cd98b0d57dd46edda52bc5ae27ef1fca031/Build/scripts/finalize_build.ps1#L55)):*
> Is it because we manually call msbuild that we need to do this check manually ?
> Wouldn't the msbuild command have a different exit code if there was an error ?
> If this approach works I don't mind, but it seems weird to rely on log content
</details>
Because of the way that VSTS handles failures and continuation, I think it is more reliable to check the status of the log files, which gives more freedom to coerce VSTS to do the right thing.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/microsoft/chakracore/1070)*

Cellule:
Review status: all files reviewed at latest revision, 7 unresolved discussions.

---

*[Build/scripts/build.ps1, line 87 \[r1\]](https://reviewable.io:443/reviews/microsoft/chakracore/1070#-KJCAxT7HawzXs73zt30:-KJCHX15HWY2FPTOd0MF:-793030938) ([raw file](https://github.com/microsoft/chakracore/blob/3d2a6cd98b0d57dd46edda52bc5ae27ef1fca031/Build/scripts/build.ps1#L87)):*
<details><summary><i>Previously, dilijev (Doug Ilijev) wrote…</i></summary>
> just doing /t:Clean only performs the Clean. Possibly, the /t:Clean,Rebuild would work but would that be somehow better than the current setup? Either way we need to capture the same logic in this file. Would the build be somehow faster if we only issue one msbuild command?
</details>
We'd need to test to make sure it does the right thing. I would assume it would be slightly faster, but I could be wrong on this one.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/microsoft/chakracore/1070)*

dilijev:
Review status: all files reviewed at latest revision, 7 unresolved discussions.

---

*[Build/scripts/build.ps1, line 87 \[r1\]](https://reviewable.io:443/reviews/microsoft/chakracore/1070#-KJCAxT7HawzXs73zt30:-KJCHmQvODCwzDyU3NPg:-335068683) ([raw file](https://github.com/microsoft/chakracore/blob/3d2a6cd98b0d57dd46edda52bc5ae27ef1fca031/Build/scripts/build.ps1#L87)):*
> We'd need to test to make sure it does the right thing. I would assume it would be slightly faster, but I could be wrong on this one.
</details>
Most likely wouldn't be slower, could be faster. In any case, this seems like a good idea. I'll run the experiment.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/microsoft/chakracore/1070)*

dilijev:
Review status: all files reviewed at latest revision, 4 unresolved discussions.

---

*[Build/scripts/build.ps1, line 6 \[r1\]](https://reviewable.io:443/reviews/microsoft/chakracore/1070#-KJ92GoBtL61nJkHUOer-r1-6:-KJDBXGaoAsMe5HQ4oZG:-499146009) ([raw file](https://github.com/microsoft/chakracore/blob/3d2a6cd98b0d57dd46edda52bc5ae27ef1fca031/Build/scripts/build.ps1#L6)):*
<details><summary><i>Previously, dilijev (Doug Ilijev) wrote…</i></summary>
> Note to self (and reviewers): I think it would be better to rename this file to something like "run_build.ps1" or "do_build.ps1" to reserve the name of "build.ps1" for a script which runs the entire build (not yet implemented).
> 
> 
> 
> (After discussing offline with @Cellule it seems like it is more and more likely that a single script is going to be more or less where the MicroBuild definition goes - we're already converging to an entirely powershell-script-based build.)
</details>
Done.

---

*[Build/scripts/build.ps1, line 87 \[r1\]](https://reviewable.io:443/reviews/microsoft/chakracore/1070#-KJCAxT7HawzXs73zt30:-KJDBdy6FT9lkj9qGPzM:-499146009) ([raw file](https://github.com/microsoft/chakracore/blob/3d2a6cd98b0d57dd46edda52bc5ae27ef1fca031/Build/scripts/build.ps1#L87)):*
<details><summary><i>Previously, dilijev (Doug Ilijev) wrote…</i></summary>
> Most likely wouldn't be slower, could be faster. In any case, this seems like a good idea. I'll run the experiment.
</details>
Done.

---

*[Build/scripts/finalize_build.ps1, line 111 \[r1\]](https://reviewable.io:443/reviews/microsoft/chakracore/1070#-KJ92GoBtL61nJkHUOej-r1-111:-KJDBibwlSTkw2D7kIfF:-499146009) ([raw file](https://github.com/microsoft/chakracore/blob/3d2a6cd98b0d57dd46edda52bc5ae27ef1fca031/Build/scripts/finalize_build.ps1#L111)):*
<details><summary><i>Previously, dilijev (Doug Ilijev) wrote…</i></summary>
> Note to self: This is not exactly true. We should keep this block because it doesn't hurt.
</details>
Done.

---

*[Build/scripts/init_build.ps1, line 45 \[r1\]](https://reviewable.io:443/reviews/microsoft/chakracore/1070#-KJ92GoBtL61nJkHUOek-r1-45:-KJDCv8nvaiTXuujxfpG:19534896) ([raw file](https://github.com/microsoft/chakracore/blob/3d2a6cd98b0d57dd46edda52bc5ae27ef1fca031/Build/scripts/init_build.ps1#L45)):*
<details><summary><i>Previously, dilijev (Doug Ilijev) wrote…</i></summary>
> Oh, thanks. I refactored this code and forgot to remove this check afterwards.
</details>
I decided to changed the way I handle this logic and move it to a place where it will actually work. Done.

---

*[Build/scripts/post_build.ps1, line 9 \[r1\]](https://reviewable.io:443/reviews/microsoft/chakracore/1070#-KJ92GoBtL61nJkHUOel-r1-9:-KJDD8dxgaFmMDkrhrqj:-499146009) ([raw file](https://github.com/microsoft/chakracore/blob/3d2a6cd98b0d57dd46edda52bc5ae27ef1fca031/Build/scripts/post_build.ps1#L9)):*
<details><summary><i>Previously, dilijev (Doug Ilijev) wrote…</i></summary>
> Whoops. I'll revert this.
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/microsoft/chakracore/1070)*

dilijev:
Review status: all files reviewed at latest revision, 4 unresolved discussions.

---

*[Build/scripts/build.ps1, line 87 \[r1\]](https://reviewable.io:443/reviews/microsoft/chakracore/1070#-KJCAxT7HawzXs73zt30:-KJDH_-bbML2nol-KKfk:-299933888) ([raw file](https://github.com/microsoft/chakracore/blob/3d2a6cd98b0d57dd46edda52bc5ae27ef1fca031/Build/scripts/build.ps1#L87)):*
<details><summary><i>Previously, dilijev (Doug Ilijev) wrote…</i></summary>
> Done.
</details>
From experiment, it's basically the same speed. If there's any savings it's in not needing to start up the msbuild process again. I've made the change now so I'm keeping it :P

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/microsoft/chakracore/1070)*

Build/scripts/pogo_build.ps1
```diff
@@ -0,0 +1,143 @@
...
+#
+# Clean up
+#
+
+if (("$binpath" -ne "") -and (Test-Path $binpath)) {
+    # remove *.pgc, *.pgd, and pgort*
+    Get-ChildItem -Recurse -Path $binpath "*" `
+        | ? { $_.Name -match "(.*\.pg[cd]|pgort.*)" } `
+        | % { Remove-Item -Force $_.FullName }
+}
```
dilijev: @Cellule Could you review this file?

Cellule:
Reviewed 5 of 5 files at r2, 3 of 9 files at r3.
Review status: 8 of 14 files reviewed at latest revision, 4 unresolved discussions.

---

*[Build/scripts/pogo_build.ps1, line 143 \[r3\]](https://reviewable.io:443/reviews/microsoft/chakracore/1070#-KJISN-Sb_RLUYiO_pXo-r3-143:-KJIU4tarKaGAEWpehKA:-580240295) ([raw file](https://github.com/microsoft/chakracore/blob/c05e318684e8db989b1ac97aa604e7c16f3f5db6/Build/scripts/pogo_build.ps1#L143)):*
<details><summary><i>Previously, dilijev (Doug Ilijev) wrote…</i></summary>
> @Cellule Could you review this file?
</details>
It feels like there is a lot of copy paste from other scripts, from what I can see the reasons are 
 - to pass an extra param `/p:POGO_TYPE=...`
 - Because the buildName is not an environment variable
As for the rest, it feels pretty much the same as run_build.ps1.
Wouldn't it be possible to make these 2 things as parameters and just call that script instead ?

@thomsontan  for context

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/microsoft/chakracore/1070)*

dilijev:
Review status: 8 of 14 files reviewed at latest revision, 4 unresolved discussions.

---

*[Build/scripts/pogo_build.ps1, line 143 \[r3\]](https://reviewable.io:443/reviews/microsoft/chakracore/1070#-KJISN-Sb_RLUYiO_pXo-r3-143:-KJK3yS5uFg74YaDTnlQ:-1715867530) ([raw file](https://github.com/microsoft/chakracore/blob/c05e318684e8db989b1ac97aa604e7c16f3f5db6/Build/scripts/pogo_build.ps1#L143)):*
> It feels like there is a lot of copy paste from other scripts, from what I can see the reasons are 
>  - to pass an extra param `/p:POGO_TYPE=...`
>  - Because the buildName is not an environment variable
> As for the rest, it feels pretty much the same as run_build.ps1.
> Wouldn't it be possible to make these 2 things as parameters and just call that script instead ?
> 
> @thomsontan  for context
</details>
* Agree that parameters to the build script should be used to select between pogo and non-pogo.
* I'm not sure how to handle the semantics of an explicit pogo-only build while supporting non-pogo build followed by pogo on the same obj files (performed in the post-build script).
* Post-build script's pogo story also feels like a bit of repeated logic, but the pogocmd script param to post_build.ps1 has a different API and I wasn't sure how to unify those either. I think now that we have this proof of concept it would be worth meeting to hash out semantics and figure out how to refactor them to be better overall.
* I also want to do this with the goal in mind of creating a build.ps1 that could, in theory, be used by devs locally as well as usable by the build system.
* RE: Environment variables: My feeling is that these scripts will be FAR easier to use and maintain if we rely less on the environment and more on explicitly passing the appropriate parameters from call to call. The theorized build.ps1 is the only script that needs to have simple default parameters so that it is usable by a dev, the other scripts can require all the parameters necessary.
* Passing `/p:POGO_TYPE=...` is nice because it doesn't involve setting the environment. It eliminates the need for much of the logic in the `pgo/*.cmd` files and even would eliminate a few of those files entirely. But while we're at it, I think we could follow this general approach to replace the use-case of those scripts with a similar script (or ideally the same script pogo_build.ps1 or run_build.ps1 with the pogo build params you mentioned).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/microsoft/chakracore/1070)*

Cellule:
That sounds fair enough, overall it works so it's fine to merge like this and we'll iterate after
LGTM

dilijev:
@Cellule I fixed up the remaining actionable items (TODOs) and rebased to a single commit. The new build definitions are ready to be deployed when this is merged in. Any remaining concerns that are appropriate to address at this time (i.e. not involving the refactoring we discussed)?

PR #1072, opened 2016-06-01
cross-platform: Improve clang error reporting with build.sh

### LINKED ISSUE #954: cross-platform: Improve clang error reporting with build.sh
At the moment, ChakraCore linux support requires a certain minimum clang version. We have a couple of command line arguments to feed build script with the clang++ and clang. 

The output is not clear;
 - If the user only provides the path for cc. 
 - If the clang or clang++ from path are not matching the minimum required versions.

obastemur:
Fixes #954

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -110,7 +110,12 @@ if [[ ${#_VERBOSE} > 0 ]]; then
     echo ""
 fi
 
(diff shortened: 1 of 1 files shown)


digitalinfinity:
LGTM

PR #1073, opened 2016-06-01
Time travel debugging

mrkmarron:

To demonstrate the use of this functionality when have a version of Node-ChakraCore and VSCode that provide a reverse-step operation.


digitalinfinity:
@dotnet-bot 

PR #1077, opened 2016-06-02
Improve the perf of Array#filter a little bit by using specialized versions of GetItem

boingoing:
We were always going into the uint64 path for these helpers due to the way the logic of EntryFilter was written. Made a fix by factoring-out the bulk of this function into a templated version.

### DIFF
--- a/lib/Runtime/Library/JavascriptArray.cpp
+++ b/lib/Runtime/Library/JavascriptArray.cpp
@@ -8996,13 +8996,9 @@ Case0:
             JavascriptError::ThrowTypeError(scriptContext, JSERR_This_NullOrUndefined, _u("Array.prototype.filter"));
         }
 
-        RecyclableObject* newObj = nullptr;
-        JavascriptArray* newArr = nullptr;
         BigIndex length;
         JavascriptArray* pArr = nullptr;
         RecyclableObject* dynamicObject = nullptr;
-        RecyclableObject* callBackFn = nullptr;
-        Var thisArg = nullptr;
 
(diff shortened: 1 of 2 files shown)


boingoing:
This change is to address the performance concerns in https://microsoft.visualstudio.com/defaultcollection/web/wi.aspx?pcguid=cb55739e-4afe-46a3-970f-1b49d8ee7564&id=7763834

lib/Runtime/Library/JavascriptArray.cpp
```diff
@@ -9036,11 +9032,23 @@ namespace Js
             length = pArr->length;
         }
 
+        if (length.IsSmallIndex())
```
DigiTec: This method of factoring will invoke the large index version even for small indexes when the length itself is a large index. That might be a concern. This is, however, an edge case.

farzonl: Hi  Taylor, hope things are well. The link you posted doesn't really shed any light on the perf impact to people outside of redmond. could you quantitate the perf improvement and on what testing  you found the regressive behavior?

boingoing: Yes, that's true. If the length is a large index, though, we're going to walk across the entire 32-bit range and at least call HasItem for each one which would take hours. We could split the walk into two pieces - one for small index values and one for any big index values - but this is an edge case and I don't think it's warranted.

abchatra:
:shipit:

PR #1078, opened 2016-06-02
 cross-plaform: update tests for xplat

obastemur:
- `Error/RuntimeCompileStackOverflow.js` is disabled for xplat
  The expectation was 'Out of stack space'. If testing environment doesn't have
  a stack space similar to Windows, this test fails.

- `Strings/trim.js` and its baseline are updated to test unicode output cases
  internally


test/es6/rlexe.xml
```diff
@@ -1070,7 +1070,7 @@
   <test>
     <default>
       <files>bug_OS_2553885.js</files>
-      <tags>exclude_win7,BugFix</tags>
+      <tags>exclude_win7,BugFix,exclude_xplat</tags>
```
curtisman: Are these test just missing an UTF8 BOM?
Would adding it fixes the test?
Is there a reason we can't add it?  (e.g. the test requires that it not present?)

obastemur: Although UTF8 has no endinannes, No it doesn't miss UTF8 BOM. See below;
```
0000170: 2d0d 0a0d 0aef bbbf 6966 2028 27ed 939b  -.......if ('...
0000180: 272e 6c6f 6361 6c65 436f 6d70 6172 6528  '.localeCompare(
```

`\u1111\u1171\u11B6`
```
0000000: e184 91e1 85b1 e186 b60a
```

So the first unicode character is actually being compared to it's UCS2.

Mentioned one of this disabled tests on PR #1061 discussion.

jianchun: Copy over the comment:

es6/unicode_6_identifiers.js

Looked off line, there is a Windows/Linux diff in  bool IsIdStart(codepoint_t codepoint)  implementation. @digitalinfinity will investigate.

obastemur: Sure. As already mentioned from #1061 waiting for the investigation for disabled tests.

test/JSON/jx2.js
```diff
@@ -187,7 +187,7 @@ function repf(key, value) {
 
 var objectArray = [
 { obj: "SampleTest", desc: "String simple" },
-{ obj: "/test ze\0ro\vString\n_u4:\u0061_u2:\xbc_u1:\x0e_u2clean:\x8f", desc: "String complex" },
+{ obj: "/test ze\0ro\vString\n_u4:\u0061_u2:\x62_u1:\x0e_u2clean:\x50", desc: "String complex" },
```
goyakin: Was this the only test input with non-ASCII characters (i.e., above `127`) in the project being passed to `JSON.stringify`? If so, this might reduce test coverage.

obastemur: We were discussing offline and encoding related test updates are removed from the PR (apart from trim.js update). There will be an additional PR targeting this problem in general.

> Was this the only test input with non-ASCII characters (i.e., above 127) in the project being passed to JSON.stringify? If so, this might reduce test coverage.

There are couple of things to note with this test. This test has a platform dependency (timezone) hence it's already excluded from jenkins (already reduced coverage).

So, If upcoming PR may match Windows baseline for this file on xplat, we are good to go. If It doesn't, I would update this file back with ASCII and prepare a separate test file for non ASCII tests without baseline. At least we could test that part on jenkins.

obastemur:

digitalinfinity:
LGTM

PR #1080, opened 2016-06-02
Merge Time-Travel Debugging into master

mrkmarron:

To demonstrate the use of this functionality when have a version of Node-ChakraCore and VSCode that provide a reverse-step operation.


mrkmarron:
@dotnet-bot 
test Copyright Check please
test EOL Check please

bin/ch/ch.cpp
```diff
@@ -196,48 +207,536 @@ static void CALLBACK PromiseContinuationCallback(JsValueRef task, void *callback
     MessageQueue * messageQueue = (MessageQueue *)callbackState;
 
     WScriptJsrt::CallbackMessage *msg = new WScriptJsrt::CallbackMessage(0, task);
+
+#if ENABLE_TTD
+    ChakraRTInterface::JsTTDNotifyHostCallbackCreatedOrCanceled(true, false, false, task, msg->GetId());
```
agarwal-sandeep: Nit: comments for true/false arguments would improve readability like
JsTTDNotifyHostCallbackCreatedOrCanceled(/*isCreate*/ true, /*isCancel*/ false, /*isRepeating*/ false, task, msg->GetId())

bin/ch/ch.cpp
```diff
@@ -196,48 +207,536 @@ static void CALLBACK PromiseContinuationCallback(JsValueRef task, void *callback
...
+                if(FileInformation.dwFileAttributes & FILE_ATTRIBUTE_DIRECTORY)
+                {
+                    DeleteDirectory(strFilePath);
+                    ::RemoveDirectory(strFilePath);
+                }
+                else
+                {
+                    // Set file attributes
+                    ::SetFileAttributes(strFilePath, FILE_ATTRIBUTE_NORMAL);
+                    ::DeleteFile(strFilePath);
```
agarwal-sandeep: What happens if these fails?

bin/ch/ch.cpp
```diff
@@ -196,48 +207,536 @@ static void CALLBACK PromiseContinuationCallback(JsValueRef task, void *callback
...
+        //
+        AssertMsg(false, "Broken");
 
-    IfJsErrorFailLog(ChakraRTInterface::JsSetPromiseContinuationCallback(PromiseContinuationCallback, (void*)messageQueue));
+        Debugger* debugger = Debugger::GetDebugger(chRuntime);
+        debugger->StartDebugging(chRuntime);
+    }
+}
+
+void CreateDirectoryIfNeeded(const char16* path)
```
agarwal-sandeep: Please move helper functions which are generic to Helpers.cpp like CreateDirectoryIfNeeded, DeleteDirectory, GetFileFromURI

bin/ch/ch.cpp
```diff
@@ -196,48 +207,536 @@ static void CALLBACK PromiseContinuationCallback(JsValueRef task, void *callback
...
-        // because setTimeout can add scripts to execute.
-        do
+        GetDefaultTTDDirectory(fullTTDUri, uri + 1);
+    }
+#endif
+}
+
+static void CALLBACK TTInitializeForWriteLogStreamCallback(const char16* uri)
+{
+    //If the directory does not exist then we want to write it
```
agarwal-sandeep: nit: create it

bin/ch/ch.cpp
```diff
@@ -196,48 +207,536 @@ static void CALLBACK PromiseContinuationCallback(JsValueRef task, void *callback
...
+    CreateDirectoryIfNeeded(uri);
+
+    //Clear the logging directory so it is ready for us to write into
+    DeleteDirectory(uri);
+}
+
+static HANDLE TTOpenStream_Helper(const char16* uri, bool read, bool write)
```
agarwal-sandeep: un-used parameter?

bin/ch/ch.cpp
```diff
@@ -196,48 +207,536 @@ static void CALLBACK PromiseContinuationCallback(JsValueRef task, void *callback
...
+        LPTSTR pTemp = NULL;
+        FormatMessage(FORMAT_MESSAGE_ALLOCATE_BUFFER | FORMAT_MESSAGE_FROM_SYSTEM | FORMAT_MESSAGE_ARGUMENT_ARRAY, NULL, lastError, 0, (LPTSTR)&pTemp, 0, NULL);
+        fwprintf(stderr, _u(": %s\n"), pTemp);
+    }
+
+    return ok;
+#endif
+}
+
+static void CALLBACK TTFlushAndCloseStreamCallback(HANDLE strm, bool read, bool write)
```
agarwal-sandeep: Consider moving all static TTD specific functions to a separate .cpp file?

bin/ch/ch.cpp
```diff
@@ -196,48 +207,536 @@ static void CALLBACK PromiseContinuationCallback(JsValueRef task, void *callback
...
+                if(nextEventTime == -1)
+                {
+                    wprintf(_u("\nReached end of Execution -- Exiting.\n"));
+                    break;
+                }
+            }
+        }
+        catch(...)
+        {
+            wprintf(_u("Exception."));
```
agarwal-sandeep: Shouldn't reach here right? if so this is a fatal error, terminate the process?

bin/ch/ch.cpp
mrkmarron: Done + other callsites as well.

bin/ch/ch.cpp
```diff
@@ -196,48 +207,536 @@ static void CALLBACK PromiseContinuationCallback(JsValueRef task, void *callback
...
+        return;
+    }
+
+    BOOL success = CreateDirectory(fullpath, NULL);
+    if(!success)
+    {
+        DWORD lastError = GetLastError();
+        LPTSTR pTemp = NULL;
+        FormatMessage(FORMAT_MESSAGE_ALLOCATE_BUFFER | FORMAT_MESSAGE_FROM_SYSTEM | FORMAT_MESSAGE_ARGUMENT_ARRAY, NULL, lastError, 0, (LPTSTR)&pTemp, 0, NULL);
+        fwprintf(stderr, _u(": %s"), pTemp);
```
mrkmarron: Other failure messages in ch go to stderr (e.g. GenerateLibraryByteCodeHeader message). Not sure what the preferred style is.

bin/ch/ch.cpp
mrkmarron: Nothing bad. We will overwrite important files later but some useless files may remain. This is mainly to keep things clean.

bin/ch/ch.cpp
mrkmarron: Done.

bin/ch/ch.cpp
mrkmarron: For the current time this is true but we may want to allow rw opens later as well. I added an Assert to ensure current use is correct and make this explicit in the code.

bin/ch/ch.cpp
```diff
@@ -450,12 +1014,50 @@ unsigned int WINAPI StaticThreadProc(void *lpParam)
...
         PrintUsage();
         return EXIT_FAILURE;
     }
 
+    int cpos = 0;
+    for(int i = 0; i < argc; ++i)
+    {
+        if(wcsstr(argv[i], _u("-TTRecord:")) == argv[i])
+        {
+            doTTRecord = true;
```
agarwal-sandeep: Only one of doTTRecord and doTTDebug can be true at a time right? print error early if both are set?

bin/ch/ch.cpp
mrkmarron: Improved message + added terminate.

bin/ch/stdafx.h
```diff
@@ -86,6 +86,8 @@ if (!(exp)) \
 #include "ChakraDebug.h"
 
 typedef void * Var;
+#define IfJsrtErrorRet(expr) do { if((expr) != JsNoError) { return JS_INVALID_REFERENCE; } } while(0)
+#define IfJsrtErrorRetErrorCode(expr) do { JsErrorCode errorCode = (expr); if (errorCode != JsNoError) { return errorCode; } } while (0)
```
agarwal-sandeep: Nit: Not used anywhere?

bin/ch/ch.cpp
mrkmarron: Added check + exit in argument extraction code.

bin/ch/ch.cpp
mrkmarron: Done.

lib/Jsrt/ChakraCommon.h
```diff
@@ -2978,4 +2978,343 @@ typedef unsigned char* ChakraBytePtr;
         JsSetPromiseContinuationCallback(
             _In_ JsPromiseContinuationCallback promiseContinuationCallback,
             _In_opt_ void *callbackState);
+
+    /////////////////////
```
agarwal-sandeep: TTD specific APIs should be in ChakraDebug.h or in ChakraCore.h, ChakraCommon.h contain runtime API which are exposed both in ChakraCore.dll and Chakra.dll. Also TTD is debugging specific so better to have it in ChakraDebug.h

lib/Jsrt/ChakraCommon.h
```diff
@@ -2978,4 +2978,343 @@ typedef unsigned char* ChakraBytePtr;
...
+
+    /// <summary>
+    ///     TTD API -- may change in future versions:
```
agarwal-sandeep: We should use a <remarks>section just after <summary> to note that these APIs are experimental and may change in future. Or we put these in a new header file and only enable under a macro say USE_CHAKRA_TTD_EXPIREMENTAL_APIS

lib/Jsrt/ChakraCommon.h
```diff
@@ -2978,4 +2978,343 @@ typedef unsigned char* ChakraBytePtr;
...
+    ///     Given the the uri location specified for the TTD output data, which may be relative or contain other implcit information,
```
agarwal-sandeep: Repeated word "the"

lib/Jsrt/ChakraCommon.h
```diff
@@ -2978,4 +2978,343 @@ typedef unsigned char* ChakraBytePtr;
         JsSetPromiseContinuationCallback(
             _In_ JsPromiseContinuationCallback promiseContinuationCallback,
             _In_opt_ void *callbackState);
+
+    /////////////////////
+
+    /// <summary>
+    ///     TTD API -- may change in future versions:
+    ///     Given the the uri location specified for the TTD output data, which may be relative or contain other implcit information, 
+    ///     convert it into a fully normalized location descriptor. This fiully resolved location will be passed to the later callbacks
```
agarwal-sandeep: fully

lib/Jsrt/ChakraCommon.h
```diff
@@ -2978,4 +2978,343 @@ typedef unsigned char* ChakraBytePtr;
...
+    ///     such as JsTTDInitializeForWriteLogStreamCallback, JsTTDGetLogStreamCallback, and JsTTDGetSnapshotStreamCallback.
+    /// </summary>
+    /// <param name="uri">The uri the user provided for the output location of the TTD data.</param>
+    /// <param name="fullTTDUri">The fully resolved location for the TTD data output.</param>
+    typedef void (CALLBACK *JsTTDInitializeUriCallback)(_In_z_ const wchar_t* uri, _Out_ wchar_t** fullTTDUri);
+
+    /// <summary>
+    ///     TTD API -- may change in future versions:
+    ///     Ensure that the location specified for outputting the TTD data is clean. Specifically, ensure that any previous TTD 
+    ///     in the location has been removed.
```
agarwal-sandeep: TTD "information" in or TTD "data" in

lib/Jsrt/ChakraCommon.h
```diff
@@ -2978,4 +2978,343 @@ typedef unsigned char* ChakraBytePtr;
...
+    /// </summary>
+    /// <param name="fullTTDUri">The fully resolved location for the TTD data output as provied by JsTTDInitializeUriCallback.</param>
+    typedef void (CALLBACK *JsTTDInitializeForWriteLogStreamCallback)(_In_z_ const wchar_t* uri);
+
+    /// <summary>
+    ///     TTD API -- may change in future versions:
+    ///     Construct a HANDLE that will be used to read/write the event log portion of the TTD data based on the uri 
+    ///     provided by JsTTDInitializeUriCallback.
+    /// </summary>
+    /// <remarks><para>Exactly one of read or write will be set to true.</para></remarks>
```
agarwal-sandeep: <para> should be in next line

lib/Jsrt/ChakraCommon.h
```diff
@@ -2978,4 +2978,343 @@ typedef unsigned char* ChakraBytePtr;
...
+    /// <param name="uri">The fully resolved location for the TTD data as provied by JsTTDInitializeUriCallback.</param>
+    /// <param name="read">If the handle should be opened for reading.</param>
+    /// <param name="write">If the handle should be opened for writing.</param>
+    /// <returns>A HANDLE opened in read/write mode as specified.</returns>
```
agarwal-sandeep: Should be in next line, see other usage in header file

lib/Jsrt/ChakraCommon.h
```diff
@@ -2978,4 +2978,343 @@ typedef unsigned char* ChakraBytePtr;
...
+    typedef HANDLE (CALLBACK *JsTTDGetSnapshotStreamCallback)(_In_z_ const wchar_t* uri, _In_z_ const wchar_t* snapId, _In_ bool read, _In_ bool write);
+
+    /// <summary>
+    ///     TTD API -- may change in future versions:
+    ///     Construct a HANDLE that will be used to read/write information on source code loaded by the program.
+    /// </summary>
+    /// <remarks><para>Exactly one of read or write will be set to true.</para></remarks>
+    /// <param name="uri">The fully resolved root location for the TTD source code data.</param>
+    /// <param name="bodyCtrId">A unique string identifier for this source file.</param>
```
agarwal-sandeep: Something to consider: In JsTTDGetSrcCodeStreamCallback why can't we use the Utf8SourceInfo->m_sourceInfoId instead of bodyCtrId.

lib/Jsrt/ChakraCommon.h
```diff
@@ -2978,4 +2978,343 @@ typedef unsigned char* ChakraBytePtr;
...
+    /// <summary>
+    ///     TTD API -- may change in future versions:
+    ///     A callback for reading data from a handle.
+    /// </summary>
+    /// <param name="strm">The HANDLE to read the data from.</param>
+    /// <param name="buff">The buffer to place the data into.</param>
+    /// <param name="size">The max number of bytes that should be read.</param>
+    /// <param name="readCount">The actual number of bytes read and placed in the buffer.</param>
+    /// <returns>TRUE if the read was successful FALSE otherwise.</returns>
+    typedef BOOL(CALLBACK *JsTTDReadBytesFromStreamCallback)(_In_ HANDLE strm, _Out_writes_(size) BYTE* buff, _In_ DWORD size, _Out_ DWORD* readCount);
```
agarwal-sandeep: bool instead of BOOL?

lib/Jsrt/ChakraCommon.h
```diff
@@ -2978,4 +2978,343 @@ typedef unsigned char* ChakraBytePtr;
...
+    /// <param name="read">If the handle should be opened for reading.</param>
+    /// <param name="write">If the handle should be opened for writing.</param>
+    /// <returns>A HANDLE opened in read/write mode as specified.</returns>
+    typedef HANDLE (CALLBACK *JsTTDGetSrcCodeStreamCallback)(_In_z_ const wchar_t* uri, _In_z_ const wchar_t* bodyCtrId, _In_z_ const wchar_t* srcFileName, _In_ bool read, _In_ bool write);
+
+    /// <summary>
+    ///     TTD API -- may change in future versions:
+    ///     A callback for reading data from a handle.
+    /// </summary>
+    /// <param name="strm">The HANDLE to read the data from.</param>
```
agarwal-sandeep: Can we change it to - handle, dataHandle, fileHandle or fileStream. misspelled words are not intuitive.

lib/Jsrt/ChakraCommon.h
```diff
@@ -2978,4 +2978,343 @@ typedef unsigned char* ChakraBytePtr;
...
+    typedef void (CALLBACK *JsTTDFlushAndCloseStreamCallback)(_In_ HANDLE strm, _In_ bool read, _In_ bool write);
+
+    /// <summary>
+    ///     TTD API -- may change in future versions:
+    ///     Creates a new runtime in Record Mode.
+    /// </summary>
+    /// <param name="attributes">The attributes of the runtime to be created.</param>
+    /// <param name="infoUri">The uri where the recorded Time-Travel data should be stored.</param>
+    /// <param name="snapInterval">The interval to wait between snapshots (measured in millis).</param>
+    /// <param name="snapHistoryLength">The length of time to maintain before discadring (measured in number of snapshots).</param>
```
agarwal-sandeep: discarding. This is kind of ring buffer right? if so a remarks explaining this would be helpful

lib/Jsrt/ChakraCommon.h
```diff
@@ -2978,4 +2978,343 @@ typedef unsigned char* ChakraBytePtr;
...
+    /// <summary>
+    ///     TTD API -- may change in future versions:
+    ///     Creates a script context in Time Travel mode for running scripts.
+    /// </summary>
+    /// <remarks>See <c>JsCreateContext</c> for more information.</remarks>
+    /// <param name="runtime">The runtime the script context is being created in.</param>
+    /// <param name="newContext">The created script context.</param>
+    /// <returns>The code <c>JsNoError</c> if the operation succeeded, a failure code otherwise.</returns>
+    CHAKRA_API
+        JsTTDCreateContext(
```
agarwal-sandeep: Can we call JsCreateContext and set it as current context on a runtime created using JsTTDCreateDebugRuntime?

lib/Jsrt/ChakraCommon.h
```diff
@@ -2978,4 +2978,343 @@ typedef unsigned char* ChakraBytePtr;
...
+    CHAKRA_API
+        JsTTDSetDebuggerForReplay();
+
+    /// <summary>
+    ///     TTD API -- may change in future versions:
+    ///     Set the functions that the TTD system uses to write info out of main memory when needed.
+    /// </summary>
+    /// <param name="runtime">The runtime to set the functions for (must be created in debug mode).</param>
+    /// <param name="ttdInitializeTTDUriFunction">The <c>JsTTDInitializeUriCallback</c> function for converting the user provided location into an absolute location for reading/writing time travel recording data.</param>
+    /// <param name="writeInitializeFunction">The <c>JsTTDInitializeForWriteLogStreamCallback</c> function for performing any initializtion needed prepare uri for storing time travel recording data.</param>
```
agarwal-sandeep: initialization

agarwal-sandeep: consider rewording, looks incomplete

lib/Jsrt/ChakraCommon.h
```diff
@@ -2978,4 +2978,343 @@ typedef unsigned char* ChakraBytePtr;
...
+    ///     TTD API -- may change in future versions:
+    ///     Start Time-Travel Debugging.
+    /// </summary>
+    /// <returns>The code <c>JsNoError</c> if the operation succeeded, a failure code otherwise.</returns>
+    CHAKRA_API
+        JsTTDStartTimeTravelDebugging();
+
+    /// <summary>
+    ///     TTD API -- may change in future versions:
+    ///     Pause Time-Travel recording gefor executing code on behalf of debugger or other diagnostic/telemetry.
```
agarwal-sandeep: before

lib/Jsrt/ChakraCommon.h
```diff
@@ -2978,4 +2978,343 @@ typedef unsigned char* ChakraBytePtr;
...
+    /// </summary>
+    /// <returns>The code <c>JsNoError</c> if the operation succeeded, a failure code otherwise.</returns>
+    CHAKRA_API
+        JsTTDReStartTimeTravelAfterRuntimeOperation();
+
+    /// <summary>
+    ///     TTD API -- may change in future versions:
+    ///     Notify the Js runtime that the host as created/canceled a callback with the given function and id.
+    /// </summary>
+    /// <param name="isCancel">True if the action is to create the callback with the callbackId.</param>
```
agarwal-sandeep: isCreated

lib/Jsrt/ChakraCommon.h
```diff
@@ -2978,4 +2978,343 @@ typedef unsigned char* ChakraBytePtr;
...
+    /// <param name="isCancel">True if the action is to cancel the callback with the callbackId.</param>
+    /// <param name="isRepeating">True if the action is to create a repeating callback (e.g., setInterval).</param>
```
agarwal-sandeep: What is the use of these? when will the repeat happen?

lib/Jsrt/ChakraCommon.h
```diff
@@ -2978,4 +2978,343 @@ typedef unsigned char* ChakraBytePtr;
...
+            _In_ JsValueRef function, 
+            _In_ INT64 callbackId);
+
+    /// <summary>
+    ///     TTD API -- may change in future versions:
+    ///     Notify the Js runtime we are at a safe yield point in the event loop (i.e. no locals on the stack and we can proccess as desired).
+    /// </summary>
+    /// <returns>The code <c>JsNoError</c> if the operation succeeded, a failure code otherwise.</returns>
+    CHAKRA_API
+        JsTTDNotifyYield();
```
agarwal-sandeep: Does this mean no JavaScript execution on stack? what happens if we call this in between?

lib/Jsrt/ChakraCommon.h
```diff
@@ -2978,4 +2978,343 @@ typedef unsigned char* ChakraBytePtr;
...
+
+    /// <summary>
+    ///     TTD API -- may change in future versions:
+    ///     Execute from the current point in the log to the end returning the error code.
+    ///     If the debugger requested an abort the code is JsNoError -- rootEventTime is the target event time we need to move to and re-execute from.
+    ///     If we aborted at the end of the replay log the code is JsNoError -- rootEventTime is -1.
+    ///     If there was an unhandled script exception the code is JsErrorCategoryScript.
+    /// </summary>
+    /// <param name="rootEventTime">The event time that we should move to next or notification (-1) that replay has ended.</param>
+    /// <returns>See summary.</returns>
```
agarwal-sandeep: We should put the summary here instead of referring to summary section here.

lib/Jsrt/Jsrt.cpp
```diff
@@ -23,6 +23,17 @@
...
 #endif
 
+#if !ENABLE_TTD
+#define PERFORM_JSRT_TTD_RECORD_ACTION_CHECK(CTX) false
+
+#define PERFORM_JSRT_TTD_RECORD_ACTION_NORESULT(CTX, ACTION_CODE) 
+#define PERFORM_JSRT_TTD_RECORD_ACTION_WRESULT(CTX, ACTION_CODE) 
+#define PERFORM_JSRT_TTD_RECORD_ACTION_PROCESS_RESULT(RESULT) 
+
+//TODO: find and replace all of the occourences of this in jsrt.cpp
```
agarwal-sandeep: Is this comment still valid?

lib/Jsrt/Jsrt.cpp
```diff
@@ -69,12 +80,28 @@ JsErrorCode CheckContext(JsrtContext *currentContext, bool verifyRuntimeState, b
     return JsNoError;
 }
 
-CHAKRA_API JsCreateRuntime(_In_ JsRuntimeAttributes attributes, _In_opt_ JsThreadServiceCallback threadService, _Out_ JsRuntimeHandle *runtimeHandle)
+template <class Fn>
+bool CallbackWrapper(Fn fn)
```
agarwal-sandeep: unused code?

lib/Jsrt/Jsrt.cpp
```diff
@@ -69,12 +80,28 @@ JsErrorCode CheckContext(JsrtContext *currentContext, bool verifyRuntimeState, b
...
+{
+    return GlobalAPIWrapper([&]() -> JsErrorCode {
         PARAM_NOT_NULL(runtimeHandle);
         *runtimeHandle = nullptr;
 
+#if ENABLE_TTD
+        if(optRecordUri != nullptr || optDebugUri != nullptr)
+        {
+            attributes = static_cast<JsRuntimeAttributes>(attributes | JsRuntimeAttributeEnableExperimentalFeatures);
+        }
```
agarwal-sandeep: Why these? Doesn't look right to modify host provided attributes without notifying the host.

lib/Jsrt/Jsrt.cpp
```diff
@@ -90,7 +117,7 @@ CHAKRA_API JsCreateRuntime(_In_ JsRuntimeAttributes attributes, _In_opt_ JsThrea
         );
 
         Assert((attributes & ~JsRuntimeAttributesAll) == 0);
-        if ((attributes & ~JsRuntimeAttributesAll) != 0)
```
agarwal-sandeep: Is removing space after if intentional? if not we should keep a space after if keyword. At other places as well.

lib/Jsrt/Jsrt.cpp
```diff
@@ -156,6 +204,198 @@ CHAKRA_API JsCreateRuntime(_In_ JsRuntimeAttributes attributes, _In_opt_ JsThrea
...
+        {
+            AssertMsg(false, "Can't create a context under TT if runtime is not set to support TT!!!");
+            return JsErrorCategoryUsage;
+        }
+
+        if(createUnderTT & (threadContext->IsTTRecordRequested & threadContext->IsTTDebugRequested))
+        {
+            AssertMsg(false, "Can't set both of these at the same time!!!");
+            return JsErrorCategoryUsage;
+        }
```
agarwal-sandeep: Should we merge these 2 ifs first under createUnderTT and then check for !threadContext->IsTTRequested, (threadContext->IsTTRecordRequested & threadContext->IsTTDebugRequested)?
Similarly for below 2 ifs?

lib/Jsrt/Jsrt.cpp
```diff
@@ -156,6 +204,198 @@ CHAKRA_API JsCreateRuntime(_In_ JsRuntimeAttributes attributes, _In_opt_ JsThrea
...
+        {
+#if ENABLE_TTD_DEBUGGING
+            threadContext->SetThreadContextFlag(ThreadContextFlagNoJIT);
+#endif
+
+            if(threadContext->TTDInitializeTTDUriFunction == nullptr)
+            {
+                AssertMsg(false, "Must set this callback function first!!!");
+                return JsErrorCategoryUsage;
+            }
```
agarwal-sandeep: These check should be before setting ThreadContextFlagNoJIT

lib/Jsrt/Jsrt.cpp
```diff
@@ -156,6 +204,198 @@ CHAKRA_API JsCreateRuntime(_In_ JsRuntimeAttributes attributes, _In_opt_ JsThrea
...
+
+            wchar_t* ttdlogStr = nullptr;
+            threadContext->TTDInitializeTTDUriFunction(threadContext->TTDUri, &ttdlogStr);
+
+            threadContext->InitTimeTravel(ttdlogStr, threadContext->IsTTRecordRequested, threadContext->IsTTDebugRequested, threadContext->TTSnapInterval, threadContext->TTSnapHistoryLength);
+
+            CoTaskMemFree(ttdlogStr);
```
agarwal-sandeep: Is it guaranteed that these memory is allocated by host using CoTaskMemAlloc? The JsTTDInitializeUriCallback callback doesn't mention that.

lib/Jsrt/Jsrt.cpp
```diff
@@ -156,6 +204,198 @@ CHAKRA_API JsCreateRuntime(_In_ JsRuntimeAttributes attributes, _In_opt_ JsThrea
...
+            HostScriptContextCallbackFunctor callbackFunctor(context, &JsrtContext::OnScriptLoad_TTDCallback);
+            threadContext->BeginCtxTimeTravel(context->GetScriptContext(), callbackFunctor);
+
+            if(threadContext->IsTTRecordRequested)
+            {
+                //
+                //TODO: We currently force this into debug mode in record as well to make sure parsing/bytecode generation is same as during replay.
+                //      Later we will want to be clever during inflate and not do this.
+                //
+                context->GetScriptContext()->GetDebugContext()->SetDebuggerMode(Js::DebuggerMode::Debugging);
```
agarwal-sandeep: Call InitializeDebugging() instead

lib/Jsrt/Jsrt.cpp
```diff
@@ -156,6 +204,198 @@ CHAKRA_API JsCreateRuntime(_In_ JsRuntimeAttributes attributes, _In_opt_ JsThrea
...
+
+        return JsNoError;
+    });
+}
+
+/////////////////////
+
+CHAKRA_API JsCreateRuntime(_In_ JsRuntimeAttributes attributes, _In_opt_ JsThreadServiceCallback threadService, _Out_ JsRuntimeHandle *runtimeHandle)
+{
+    return CreateRuntimeCore(attributes, nullptr, nullptr, UINT_MAX, UINT_MAX, threadService, runtimeHandle);
```
agarwal-sandeep: Please add comments on what are these values for. Similarly for CreateContextCore

lib/Jsrt/Jsrt.cpp
```diff
@@ -853,12 +1135,17 @@ CHAKRA_API JsPointerToString(_In_reads_(stringLength) const wchar_t *stringValue
         PARAM_NOT_NULL(stringValue);
         PARAM_NOT_NULL(string);
 
+        PERFORM_JSRT_TTD_RECORD_ACTION_WRESULT(scriptContext, scriptContext->GetThreadContext()->TTDLog->RecordJsRTCreateString(scriptContext, stringValue, stringLength, &__ttd_resultPtr));
```
agarwal-sandeep: This should be after IsValidCharCount check?

lib/Jsrt/Jsrt.cpp
```diff
@@ -1332,6 +1689,23 @@ CHAKRA_API JsGetTypedArrayInfo(_In_ JsValueRef typedArray, _Out_opt_ JsTypedArra
             *byteLength = typedArrayBase->GetByteLength();
         }
     }
+
+#if ENABLE_TTD
```
agarwal-sandeep: In JsCreateTypedArray we have PERFORM_JSRT_TTD_RECORD_ACTION_NOT_IMPLEMENTED then why do we need these? Isn't the scenario involving TypedArray broken?

agarwal-sandeep:
>             VALIDATE_INCOMING_REFERENCE(description, scriptContext);

What if this fails?, PERFORM_JSRT_TTD_RECORD_ACTION_WRESULT will store null value as Var?

---
Refers to: lib/Jsrt/Jsrt.cpp:1831 in 290dcb0. [](commit_id = 290dcb0ea6344c0254bb996bc884f01836393f4d, deletion_comment = False)

lib/Jsrt/Jsrt.cpp
```diff
@@ -1781,38 +2170,9 @@ CHAKRA_API JsSetExternalData(_In_ JsValueRef object, _In_opt_ void *data)
 
 CHAKRA_API JsCallFunction(_In_ JsValueRef function, _In_reads_(cargs) JsValueRef *args, _In_ ushort cargs, _Out_opt_ JsValueRef *result)
 {
-    if (result != nullptr)
-    {
-        *result = nullptr;
+    return CallFunctionCore(-1, function, args, cargs, result);
     }
```
agarwal-sandeep: nit: extra spaces

lib/Jsrt/Jsrt.cpp
```diff
@@ -2067,6 +2457,16 @@ CHAKRA_API JsGetAndClearException(_Out_ JsValueRef *exception)
     {
         return JsErrorInvalidArgument;
     }
+
+#if ENABLE_TTD
+    BEGIN_JS_RUNTIME_CALL_NOT_SCRIPT(currentContext->GetScriptContext())
```
agarwal-sandeep: Use scriptContext instead

lib/Jsrt/Jsrt.cpp
```diff
@@ -2573,7 +3057,7 @@ JsErrorCode JsSerializeScriptCore(const wchar_t *script, BYTE *functionTable, in
         const Js::Utf8SourceInfo *sourceInfo = functionBody->GetUtf8SourceInfo();
         size_t cSourceCodeLength = sourceInfo->GetCbLength(_u("JsSerializeScript"));
 
-        // truncation of code length can lead to accessing random memory. Reject the call.
+        // trucation of code length can lead to accessing random memory. Reject the call.
```
agarwal-sandeep: truncation is correct?

lib/Jsrt/ChakraDebug.h
```diff
@@ -562,4 +562,342 @@
...
+            _In_ JsRuntimeHandle runtime,
+            _Out_ JsContextRef *newContext);
+
+    /// <summary>
+    ///     TTD API -- may change in future versions:
+    ///     Executes a script with additional Time-Travel causality tracking via the <c>hostCallbackId</c>.
+    /// </summary>
+    /// <remarks>See <c>JsRunScript</c> for more information.</remarks>
+    /// <param name="hostCallbackId">
+    ///     A unique id that specifies which callback execution caused this code to be registered for execution (e.g., the timeoutId from setTimeout).
```
agarwal-sandeep: It says unique but does it have any impact on internals of ChakraCore if same hostCallbackId is passed for 2 calls to JsTTDRunScript or JsTTDCallFunction?

agarwal-sandeep:
> 

Should be in ChakraCore.def and under #if ENABLE_TTD_DEBUGGING

---
Refers to: lib/Jsrt/JsrtCommonExports.inc:125 in 290dcb0. [](commit_id = 290dcb0ea6344c0254bb996bc884f01836393f4d, deletion_comment = False)

lib/Jsrt/Jsrt.cpp
```diff
@@ -2746,3 +3230,504 @@ CHAKRA_API JsRunSerializedScriptWithCallback(_In_ JsSerializedScriptLoadSourceCa
...
+CHAKRA_API JsTTDRunScript(_In_ INT64 hostCallbackId, _In_z_ const wchar_t *script, _In_ JsSourceContext sourceContext, _In_z_ const wchar_t *sourceUrl, _Out_ JsValueRef *result)
+{
+    return RunScriptCore(hostCallbackId, script, sourceContext, sourceUrl, false, JsParseScriptAttributeNone, false /*isModule*/, result);
+}
+
+CHAKRA_API JsTTDCallFunction(_In_ INT64 hostCallbackId, _In_ JsValueRef function, _In_reads_(argumentCount) JsValueRef *arguments, _In_ unsigned short argumentCount, _Out_opt_ JsValueRef *result)
+{
+    return CallFunctionCore(hostCallbackId, function, arguments, argumentCount, result);
+}
+
```
agarwal-sandeep: Should have #if !ENABLE_TTD check

lib/Jsrt/Jsrt.cpp
```diff
@@ -2746,3 +3230,504 @@ CHAKRA_API JsRunSerializedScriptWithCallback(_In_ JsSerializedScriptLoadSourceCa
...
+#else
+    JsrtContext *currentContext = JsrtContext::GetCurrent();
+    if(currentContext->GetRuntime()->GetThreadContext()->TTDLog == nullptr)
+    {
+        AssertMsg(false, "Need to create in TTD mode.");
+        return JsErrorCategoryUsage;
+    }
+
+    //Checks if already in debug mode internally so we don't need to check before calling.
+    currentContext->GetScriptContext()->GetDebugContext()->SetDebuggerMode(Js::DebuggerMode::Debugging);
```
agarwal-sandeep: Call ScriptContext::InitializeDebugging

lib/Jsrt/Jsrt.cpp
```diff
@@ -2746,3 +3230,504 @@ CHAKRA_API JsRunSerializedScriptWithCallback(_In_ JsSerializedScriptLoadSourceCa
...
+
```
agarwal-sandeep: CreateContextCore would have already put this ScriptContext in debug mode why are we resetting it? Can a context be created without debug mode under a runtime with TTD enabled so that we hit this code? If so what happens to the existing scripts/code which would have run in the context when it was not in debug mode?

lib/Jsrt/Jsrt.cpp
```diff
@@ -2746,3 +3230,504 @@ CHAKRA_API JsRunSerializedScriptWithCallback(_In_ JsSerializedScriptLoadSourceCa
...
+    return JsNoError;
+#endif
+}
+
+CHAKRA_API JsTTDSetIOCallbacks(_In_ JsRuntimeHandle runtime, _In_ JsTTDInitializeUriCallback ttdInitializeUriFunction, _In_ JsTTDInitializeForWriteLogStreamCallback writeInitializeFunction, _In_ JsTTDGetLogStreamCallback getLogStreamInfo, _In_ JsTTDGetSnapshotStreamCallback getSnapshotStreamInfo, _In_ JsTTDGetSrcCodeStreamCallback getSrcCodeStreamInfo, _In_ JsTTDReadBytesFromStreamCallback readBytesFromStream, _In_ JsTTDWriteBytesToStreamCallback writeBytesToStream, _In_ JsTTDFlushAndCloseStreamCallback flushAndCloseStream)
```
agarwal-sandeep: nit: That's a long list of parameters on a single line. Though we are not following up the code quality generally in the file but if possible can we have each parameter in a line so that it is easy to ready and see? Similarly for other new functions added.

lib/Jsrt/Jsrt.cpp
```diff
@@ -2746,3 +3230,504 @@ CHAKRA_API JsRunSerializedScriptWithCallback(_In_ JsSerializedScriptLoadSourceCa
...
+    {
+        AssertMsg(false, "Already stopped TTD.");
+        return JsErrorCategoryUsage;
+    }
+
+    if(!scriptContext->IsTTDActive())
+    {
+        AssertMsg(false, "TTD was never started.");
+        return JsErrorCategoryUsage;
+    }
```
agarwal-sandeep: Same checks/code in multiple APIs, can we put this in a macro similar to VALIDATE_RUNTIME_STATE_FOR_START_STOP_DEBUGGING

lib/Jsrt/Jsrt.cpp
```diff
@@ -2746,3 +3230,504 @@ CHAKRA_API JsRunSerializedScriptWithCallback(_In_ JsSerializedScriptLoadSourceCa
...
+}
+
+CHAKRA_API JsTTDPauseTimeTravelBeforeRuntimeOperation()
+{
+#if !ENABLE_TTD
+    return JsErrorCategoryUsage;
+#else
+    JsrtContext *currentContext = JsrtContext::GetCurrent();
+
+    if(currentContext->GetRuntime()->GetThreadContext()->TTDLog != nullptr)
```
agarwal-sandeep: JsErrorCategoryUsage error in case this is null?

lib/Jsrt/Jsrt.cpp
```diff
@@ -2746,3 +3230,504 @@ CHAKRA_API JsRunSerializedScriptWithCallback(_In_ JsSerializedScriptLoadSourceCa
...
+{
+#if !ENABLE_TTD
+    return JsErrorCategoryUsage;
+#else
+    JsrtContext *currentContext = JsrtContext::GetCurrent();
+    Js::ScriptContext* scriptContext = currentContext->GetScriptContext();
+
+    if(PERFORM_JSRT_TTD_RECORD_ACTION_CHECK(scriptContext))
+    {
+        Js::JavascriptFunction* jsFunction = (function != nullptr) ? Js::JavascriptFunction::FromVar(function) : nullptr;
```
agarwal-sandeep: Should have VALIDATE_INCOMING_FUNCTION check? what if this is not a JavaScript function?

lib/Jsrt/Jsrt.cpp
```diff
@@ -2746,3 +3230,504 @@ CHAKRA_API JsRunSerializedScriptWithCallback(_In_ JsSerializedScriptLoadSourceCa
...
+        return JsErrorCategoryUsage;
+    }
+
+    //Make sure we don't have any pending recorded exceptions
+    if(scriptContext->HasRecordedException())
+    {
+        scriptContext->GetAndClearRecordedException(nullptr);
+    }
+
+    //a special indicator to use the time from the argument flag or log diagnostic report
```
agarwal-sandeep: How does the host knows about these special value -2? don't see this in header file? Generally I don't like the idea of a special value if it is not explicit.

lib/Jsrt/Jsrt.cpp
```diff
@@ -2746,3 +3230,504 @@ CHAKRA_API JsRunSerializedScriptWithCallback(_In_ JsSerializedScriptLoadSourceCa
...
+    if(targetEventTime == -2)
+    {
+        targetEventTime = scriptContext->GetThreadContext()->TTDLog->GetKthEventTime(Js::Configuration::Global.flags.TTDStartEvent);
+    }
+
+    bool createFreshCtxs = false;
+    *targetStartSnapTime = scriptContext->GetThreadContext()->TTDLog->FindSnapTimeForEventTime(targetEventTime, &createFreshCtxs);
+
+    if(createFreshCtxs)
```
agarwal-sandeep: What is the most common case new context is created or not? if most of the time new context is created should this API return a bool specifying if a new context was created and ref to the new context?

lib/Jsrt/Jsrt.cpp
```diff
@@ -2746,3 +3230,504 @@ CHAKRA_API JsRunSerializedScriptWithCallback(_In_ JsSerializedScriptLoadSourceCa
...
+            context->GetScriptContext()->GetDebugContext()->GetProbeContainer()->InitializeInlineBreakEngine(runtime->GetJsrtDebugManager());
+            context->GetScriptContext()->GetDebugContext()->GetProbeContainer()->InitializeDebuggerScriptOptionCallback(runtime->GetJsrtDebugManager());
+            threadContext->GetDebugManager()->SetLocalsDisplayFlags(Js::DebugManager::LocalsDisplayFlags::LocalsDisplayFlags_NoGroupMethods);
+        }
+        catch(...)
+        {
+            //something went horribly wrong
+
+            *targetStartSnapTime = -1;
+        }
```
agarwal-sandeep: Is this fatal or recoverable? if fatal we should take down the process similar to Debugger_AttachDetach_fatal_error

lib/Jsrt/Jsrt.cpp
```diff
@@ -2746,3 +3230,504 @@ CHAKRA_API JsRunSerializedScriptWithCallback(_In_ JsSerializedScriptLoadSourceCa
...
+    }
+
+    bool createFreshCtxs = false;
+    *targetStartSnapTime = scriptContext->GetThreadContext()->TTDLog->FindSnapTimeForEventTime(targetEventTime, &createFreshCtxs);
+
+    if(createFreshCtxs)
+    {
+        try
+        {
+            AUTO_NESTED_HANDLED_EXCEPTION_TYPE((ExceptionType)(ExceptionType_OutOfMemory | ExceptionType_StackOverflow));
```
agarwal-sandeep: Why not use GlobalAPIWrapper?
Are other TTD APIs allowed when heap enum is in progress or we are in collect callback? if not the all TTD APIs should use ContextAPINoScriptWrapper so that CheckContext is executed.

lib/Jsrt/Jsrt.cpp
```diff
@@ -2746,3 +3230,504 @@ CHAKRA_API JsRunSerializedScriptWithCallback(_In_ JsSerializedScriptLoadSourceCa
...
+            threadContext->TTDLog->PushMode(TTD::TTDMode::ExcludedExecution);
+            context->GetScriptContext()->InitializeCoreImage_TTD();
+            threadContext->TTDLog->PopMode(TTD::TTDMode::ExcludedExecution);
+
+            context->GetScriptContext()->InitializeDebuggingActionsAsNeeded_TTD();
+
+            scriptContext->InitializeDebugging();
+            context->GetScriptContext()->GetDebugContext()->GetProbeContainer()->InitializeInlineBreakEngine(runtime->GetJsrtDebugManager());
+            context->GetScriptContext()->GetDebugContext()->GetProbeContainer()->InitializeDebuggerScriptOptionCallback(runtime->GetJsrtDebugManager());
+            threadContext->GetDebugManager()->SetLocalsDisplayFlags(Js::DebugManager::LocalsDisplayFlags::LocalsDisplayFlags_NoGroupMethods);
```
agarwal-sandeep: Here we set the inline break engine and script options but CreateContextCore and JsTTDSetDebuggerForReplay only sets debugger mode. Also when was JsrtDebugManager created?

lib/Jsrt/Jsrt.cpp
```diff
@@ -2746,3 +3230,504 @@ CHAKRA_API JsRunSerializedScriptWithCallback(_In_ JsSerializedScriptLoadSourceCa
...
+            context->GetScriptContext()->GetDebugContext()->SetDebuggerMode(Js::DebuggerMode::Debugging);
+
+            //initialize the core image but we need to disable debugging while this happens
+            threadContext->TTDLog->PushMode(TTD::TTDMode::ExcludedExecution);
+            context->GetScriptContext()->InitializeCoreImage_TTD();
+            threadContext->TTDLog->PopMode(TTD::TTDMode::ExcludedExecution);
+
+            context->GetScriptContext()->InitializeDebuggingActionsAsNeeded_TTD();
+
+            scriptContext->InitializeDebugging();
```
agarwal-sandeep: This is from oldContext and is gone?

lib/Jsrt/Jsrt.cpp
```diff
@@ -2746,3 +3230,504 @@ CHAKRA_API JsRunSerializedScriptWithCallback(_In_ JsSerializedScriptLoadSourceCa
...
+#if !ENABLE_TTD_DEBUGGING
+    return JsErrorCategoryUsage;
+#else
+    JsrtContext *currentContext = JsrtContext::GetCurrent();
+    Js::ScriptContext* scriptContext = currentContext->GetScriptContext();
+    if(scriptContext->GetThreadContext()->TTDLog == nullptr)
+    {
+        AssertMsg(false, "Should only happen in TT debugging mode.");
+        return JsErrorCategoryUsage;
+    }
```
agarwal-sandeep: This is at 4 places, better to move in a macro, improves code readability

lib/Jsrt/Jsrt.cpp
```diff
@@ -2746,3 +3230,504 @@ CHAKRA_API JsRunSerializedScriptWithCallback(_In_ JsSerializedScriptLoadSourceCa
...
+            TTD::TTDebuggerSourceLocation bpLocation;
+            scriptContext->GetThreadContext()->TTDLog->GetPendingTTDBPInfo(bpLocation);
+
+            Js::FunctionBody* body = bpLocation.ResolveAssociatedSourceInfo(scriptContext);
+            Js::Utf8SourceInfo* utf8SourceInfo = body->GetUtf8SourceInfo();
+
+            charcount_t charPosition;
+            charcount_t byteOffset;
+            utf8SourceInfo->GetCharPositionForLineInfo((charcount_t)bpLocation.GetLine(), &charPosition, &byteOffset);
+            long ibos = charPosition + bpLocation.GetColumn() + 1;
```
agarwal-sandeep: Are we guaranteed to have line character cache created?

lib/Jsrt/Jsrt.cpp
```diff
@@ -2746,3 +3230,504 @@ CHAKRA_API JsRunSerializedScriptWithCallback(_In_ JsSerializedScriptLoadSourceCa
...
+                    probe = debugDocument->SetBreakPoint(statement, BREAKPOINT_ENABLED);
+                    AssertMsg(probe != nullptr, "We have a bad line or something for setting a breakpoint.");
+                }
+                END_JS_RUNTIME_CALL(scriptContext);
+            }
+
+            scriptContext->GetThreadContext()->TTDLog->SetActiveBP(probe->GetId(), bpLocation);
+
+            //Finally clear the pending BP info so we don't get confused later
+            scriptContext->GetThreadContext()->TTDLog->ClearPendingTTDBPInfo();
```
agarwal-sandeep: Should only be if probe == nullptr and we do SetBreakPoint?

lib/Jsrt/Core/JsrtContextCore.h
```diff
@@ -138,7 +138,7 @@ class ChakraCoreHostScriptContext sealed : public HostScriptContext
 
     HRESULT SetDispatchInvoke(Js::JavascriptMethod dispatchInvoke) override
     {
-        AssertMsg(false, "no hostdispatch in jsrt");
+        // AssertMsg(false, "no hostdispatch in jsrt");
         return E_NOTIMPL;
```
agarwal-sandeep: Why this change? I had fixed a similar issue with my JsRTDebugging change where I initialized DispatchProfileInvoke in  ScriptContext and check for in ScriptContext::SetDispatchProfile.

lib/Jsrt/Jsrt.cpp
```diff
@@ -2746,3 +3230,504 @@ CHAKRA_API JsRunSerializedScriptWithCallback(_In_ JsSerializedScriptLoadSourceCa
...
+            elog->DoSnapshotInflate(snapshotTime);
+        }
+        END_JS_RUNTIME_CALL(scriptContext);
+        elog->PopMode(TTD::TTDMode::ExcludedExecution);
+
+        //If the log has a BP requested then we should set the actual bp here
+        if(scriptContext->GetThreadContext()->TTDLog->HasPendingTTDBP())
+        {
+            TTD::TTDebuggerSourceLocation bpLocation;
+            scriptContext->GetThreadContext()->TTDLog->GetPendingTTDBPInfo(bpLocation);
```
agarwal-sandeep: Who calls SetPendingTTDBPInfo? I don't see any caller. TTDebuggerSourceLocation constructor sets m_etime to -1 which mean invalid location

lib/Runtime/Base/FunctionBody.cpp
```diff
@@ -4373,6 +4372,13 @@ namespace Js
 /* END potentially reusable code */
 #endif /* IR_VIEWER */
 
+#if ENABLE_TTD
+    void FunctionBody::GetSourceLineFromStartOffset_TTD(const uint startOffset, ULONG* line, LONG* col)
```
agarwal-sandeep: Why a different function when it is not doing anything extra and just calling GetLineCharOffsetFromStartChar?

lib/Runtime/Debug/TTActionEvents.cpp
```diff
@@ -0,0 +1,986 @@
...
+            if(NSLogEvents::IsJsRTActionRootCall(evt))
+            {
+                const NSLogEvents::JsRTCallFunctionAction* rootEntry = NSLogEvents::GetInlineEventDataAs<NSLogEvents::JsRTCallFunctionAction, NSLogEvents::EventKind::CallExistingFunctionActionTag>(evt);
+
+                isRoot = true;
+                hasRtrSnap = rootEntry->AdditionalInfo->RtRSnap != nullptr;
+                return rootEntry->AdditionalInfo->CallEventTime;
+            }
+
+            return -10;
```
agarwal-sandeep: Magic number?

lib/Runtime/Debug/TTActionEvents.cpp
```diff
@@ -0,0 +1,986 @@
...
+            recordedException = ctx->GetAndClearRecordedException();
+            END_TRANSLATE_OOM_TO_HRESULT(hr)
+
+            if(hr == E_OUTOFMEMORY)
+            {
+                //
+                //TODO: we don't have support for OOM yet (and adding support will require a non-trivial amount of work)
+                //
+                AssertMsg(false, "OOM is not supported");
+            }
```
agarwal-sandeep: What in fretest build where assert is disabled?

lib/Runtime/Debug/TTActionEvents.cpp
```diff
@@ -0,0 +1,986 @@
...
+        }
+
+        void JsRTCallbackAction_UnloadEventMemory(EventLogEntry* evt, UnlinkableSlabAllocator& alloc)
+        {
+            JsRTCallbackAction* cbAction = GetInlineEventDataAs<JsRTCallbackAction, EventKind::CallbackOpActionTag>(evt);
+
+            if(cbAction->RegisterLocation != nullptr)
+            {
+                cbAction->RegisterLocation->Clear();
+                HeapDelete(cbAction->RegisterLocation);
```
agarwal-sandeep: Good to set it to nullptr after HeapDelete?

lib/Runtime/Debug/TTActionEvents.cpp
```diff
@@ -0,0 +1,986 @@
...
+            cfAction->ArgCount = argc + 1;
+
+            static_assert(sizeof(TTDVar) == sizeof(Js::Var), "These need to be the same size (and have same bit layout) for this to work!");
+
+            cfAction->ArgArray = alloc.SlabAllocateArray<TTDVar>(cfAction->ArgCount);
+            cfAction->ArgArray[0] = TTD_CONVERT_JSVAR_TO_TTDVAR(function);
+            js_memcpy_s(cfAction->ArgArray + 1, (cfAction->ArgCount -1) * sizeof(TTDVar), argv, argc * sizeof(Js::Var));
+
+            cfAction->AdditionalInfo->BeginTime = wallTime;
+            cfAction->AdditionalInfo->EndTime = -1.0;
```
agarwal-sandeep: All these special magic numbers should be a #define so that it is easy to understand what they mean.

agarwal-sandeep:
>             this->m_function->GetScriptContext()->TTDRootNestingCount--;

While decrementing the value here, it shouldn't be less then what we incremented in constructor right? should we save the original value in a debug only field and assert? Is that a concern at all?

---
Refers to: lib/Runtime/Debug/TTEventLog.cpp:53 in 290dcb0. [](commit_id = 290dcb0ea6344c0254bb996bc884f01836393f4d, deletion_comment = False)

agarwal-sandeep:
>         //Prepr the logging stream so it is ready for us to write into

Prepare

---
Refers to: lib/Runtime/Debug/TTEventLog.cpp:595 in 290dcb0. [](commit_id = 290dcb0ea6344c0254bb996bc884f01836393f4d, deletion_comment = False)

agarwal-sandeep:
>         ExcludedExecution = 0x20,  //Set when the system is executing code on behalf of the TTD/debugger (so we don't want to record/replay things for it)

Why not keep in order and have 0x10?

---
Refers to: lib/Runtime/Debug/TTSupport.h:110 in 290dcb0. [](commit_id = 290dcb0ea6344c0254bb996bc884f01836393f4d, deletion_comment = False)

bin/ch/ch.cpp
```diff
@@ -196,48 +207,364 @@ static void CALLBACK PromiseContinuationCallback(JsValueRef task, void *callback
...
-    WScriptJsrt::AddMessageQueue(messageQueue);
-    LPWSTR fullPathWide = nullptr;
+    if(dbgIPAddr == nullptr)
+    {
+        if(doTTDebug)
+        {
+            //we need to force the script context into dbg mode for replay even if we don't attach the debugger -- so do that here
+            ChakraRTInterface::JsTTDSetDebuggerForReplay();
+        }
+}
```
agarwal-sandeep: indentation

lib/Jsrt/ChakraCommon.h
mrkmarron: Done.

lib/Jsrt/ChakraCommon.h
mrkmarron: Done.

lib/Jsrt/ChakraCommon.h
mrkmarron: Done.

lib/Jsrt/ChakraCommon.h
mrkmarron: Done.

lib/Jsrt/ChakraCommon.h
mrkmarron: Done.

lib/Jsrt/ChakraCommon.h
mrkmarron: Done and checked layout in other comments.

lib/Jsrt/ChakraCommon.h
mrkmarron: Done.

lib/Jsrt/ChakraCommon.h
mrkmarron: Done.

lib/Jsrt/ChakraCommon.h
mrkmarron: Done.

lib/Jsrt/ChakraCommon.h
mrkmarron: Yes. That is ok but the other way around is not (and will fail with an error code).

lib/Jsrt/ChakraCommon.h
mrkmarron: Done.

lib/Jsrt/ChakraCommon.h
mrkmarron: Good catch, fixed.

lib/Jsrt/ChakraCommon.h
mrkmarron: If the host does a setInterval instead of setTimeout or a .on handler that may happen many times. We don't depend on this being set (so always setting it to false is ok) but if this info is provided we can do additional diagnostics.

lib/Jsrt/ChakraCommon.h
mrkmarron: If this is called at a non-yield point then we get an incomplete snapshot (and a hard failure). So, user must call this appropriately.

lib/Jsrt/ChakraCommon.h
mrkmarron: Done.

lib/Jsrt/Jsrt.cpp
mrkmarron: Yes, there are a few API's that we do not support yet.

lib/Jsrt/Jsrt.cpp
mrkmarron: It seems so, good spot.

lib/Jsrt/Jsrt.cpp
mrkmarron: Good call, I had this to ensure consistency with Node. Moved to Ch since that is where is should be.

lib/Jsrt/Jsrt.cpp
mrkmarron: Done.

lib/Jsrt/Jsrt.cpp
mrkmarron: Done. Nice simplification.

lib/Jsrt/Jsrt.cpp
mrkmarron: Done.

lib/Jsrt/Jsrt.cpp
mrkmarron: Added check with error code return.

lib/Jsrt/Jsrt.cpp
mrkmarron: Done.

lib/Jsrt/Jsrt.cpp
mrkmarron: Done.

lib/Jsrt/Jsrt.cpp
mrkmarron: No, we want to preserve the same error semantics.

lib/Jsrt/Jsrt.cpp
mrkmarron: Chakra can create a TypedArray internally and pass it to the host (which is what we have in some Node benchmarks) so we support this even though the explicit JsRT JsCreateTypedArray is not supported yet.

lib/Jsrt/Jsrt.cpp
mrkmarron: Fixed.

lib/Jsrt/Jsrt.cpp
mrkmarron: Good catch. Done.

lib/Jsrt/Jsrt.cpp
mrkmarron: Fixed typo.

lib/Jsrt/ChakraDebug.h
mrkmarron: No hard crash but then two call dependence chains will get merged. Need a unique id per callback event basically.

lib/Jsrt/Jsrt.cpp
mrkmarron: Done.

lib/Jsrt/Jsrt.cpp
mrkmarron: Done.

lib/Jsrt/Jsrt.cpp
mrkmarron: Done + checked other debug initialization paths.

lib/Jsrt/Jsrt.cpp
mrkmarron: Done.

lib/Jsrt/Jsrt.cpp
mrkmarron: Want the debugger to call this without needed other checks.

lib/Jsrt/Jsrt.cpp
mrkmarron: Done.

lib/Jsrt/Jsrt.cpp
mrkmarron: Creating new contexts is rare and needed to be managed by the host.

lib/Jsrt/Jsrt.cpp
mrkmarron: Done.

lib/Jsrt/Jsrt.cpp
mrkmarron: The debug host is expected to do this previously (but we don't have this fully unified yet so this is non-functional until we get the debugger hooked up again.)

lib/Jsrt/Jsrt.cpp
mrkmarron: Good catch, this is fixed.

lib/Jsrt/Jsrt.cpp
mrkmarron: Yes, only way to have a pending BP should be via TTLog reverse step and this should only be possible from the debugger.

lib/Jsrt/Core/JsrtContextCore.h
mrkmarron: Not sure what happened. I updated to match master.

lib/Jsrt/Jsrt.cpp
mrkmarron: This should happen in the JsRTDiag code. We need to finish merging in some of my debugger code there to full light up this scenario but currently blocked on there.

lib/Runtime/Base/FunctionBody.cpp
mrkmarron: We may want to adjust this API to a faster implementation in the future so I wanted to keep the uses separate.

lib/Runtime/Debug/TTActionEvents.cpp
mrkmarron: Switched to "natural" -1 value (any invalid value is ok as all callers check bool out params before use).

lib/Runtime/Debug/TTActionEvents.cpp
mrkmarron: Added fast return

lib/Runtime/Debug/TTActionEvents.cpp
mrkmarron: Agree, thanks.

bin/ChakraCore/ChakraCore.def
```diff
@@ -20,3 +20,21 @@ JsDiagSetBreakpoint
...
+JsTTDStopTimeTravelRecording
+JsTTDEmitTimeTravelRecording
+JsTTDStartTimeTravelDebugging
+JsTTDPauseTimeTravelBeforeRuntimeOperation
+JsTTDReStartTimeTravelAfterRuntimeOperation
+JsTTDNotifyHostCallbackCreatedOrCanceled
+JsTTDNotifyYield
+JsTTDPrepContextsForTopLevelEventMove
+JsTTDMoveToTopLevelEvent
+JsTTDReplayExecution
```
agarwal-sandeep: Better to have these under #if ENABLE_TTD?

bin/ch/ch.cpp
mrkmarron: Done.

bin/ch/ch.cpp
mrkmarron: Done.

mrkmarron:
Closing to ensure naming/history is clean (see PR #1160)

PR #1081, opened 2016-06-03
Clean up JSRT utf8 APIs

digitalinfinity:
1. Change utf8 APIs to only expect UTF8 output and not cesu output
2. Change ch to only support ANSI and utf8 files (not utf16-le)
3. Change existing tests which use utf16-le to use utf8
4. Fix bug in the scanner around decoding of utf8 codepoints which result in utf16 surrogate pairs.


digitalinfinity:
@dotnet-bot test this please

bin/ch/Helpers.cpp
```diff
@@ -42,81 +41,60 @@ HRESULT Helpers::LoadScriptFromFile(LPCSTR filename, LPCSTR& contents, bool* isU
...
-        {
-            // unicode LE
-            isUtf8 = false;
-        }
-        else
+
+        //
+        // Read the entire content as a binary block.
+        //
+        fread(pRawBytes, sizeof(BYTE), lengthBytes, file);
```
obastemur: `file` here is representing a `block device`. So, the assumption for `fread` will be successfully reading `lengthBytes` is open to failures and wrong.

bin/ch/Helpers.cpp
```diff
@@ -42,81 +41,60 @@ HRESULT Helpers::LoadScriptFromFile(LPCSTR filename, LPCSTR& contents, bool* isU
...
+        //
+        // Warning: The UNICODE buffer for parsing is supposed to be provided by the host.
+        // This is not a complete read of the encoding. Some encodings like UTF7, UTF1, EBCDIC, SCSU, BOCU could be
+        // wrongly classified as ANSI
+        //
         {
-            // Assume UTF8
-            isUtf8 = true;
+            LPCWSTR contentsRaw = reinterpret_cast<LPCWSTR>(pRawBytes);
+            if (0xFFFE == *contentsRaw || (0x0000 == *contentsRaw && 0xFEFF == *(contentsRaw + 1)) ||
```
obastemur: Does this have to be at the beginning of the content?

bin/ch/Helpers.cpp
```diff
@@ -42,81 +41,60 @@ HRESULT Helpers::LoadScriptFromFile(LPCSTR filename, LPCSTR& contents, bool* isU
...
-    {
-        contents = reinterpret_cast<LPCSTR>(pRawBytes);
-    }
-    else
-    {
-        LPSTR pNarrow = nullptr;
-        IfFailGo(WideStringToNarrowDynamic(reinterpret_cast<LPCWSTR>(pRawBytes), &pNarrow));
-        contents = pNarrow;
-    }
+    contents = reinterpret_cast<LPCSTR>(pRawBytes);
```
obastemur: There is also contentsRaw above. Can we use this there also?

bin/ch/Helpers.cpp
```diff
@@ -4,14 +4,13 @@
...
-HRESULT Helpers::LoadScriptFromFile(LPCSTR filename, LPCSTR& contents, bool* isUtf8Out /*= nullptr*/, UINT* lengthBytesOut /*= nullptr*/)
+HRESULT Helpers::LoadScriptFromFile(LPCSTR filename, LPCSTR& contents, UINT* lengthBytesOut /*= nullptr*/)
 {
     HRESULT hr = S_OK;
     BYTE * pRawBytes = nullptr;
     UINT lengthBytes = 0;
-    bool isUtf8 = false;
     contents = nullptr;
-    FILE * file;
+    FILE * file = nullptr;
```
obastemur: Reminder: Although nullptr and NULL are equivalent by the definition..`fopen` is from stdio.h and returns NULL in case of failure (c++11 standard doesn't have a guarantee for .h files). Besides, fopen_s is from PAL and assigns NULL to FILE.

bin/ch/Helpers.cpp
```diff
@@ -42,81 +41,60 @@ HRESULT Helpers::LoadScriptFromFile(LPCSTR filename, LPCSTR& contents, bool* isU
...
 
-    //
-    // Determine the file length, in bytes.
-    //
-    fseek(file, 0, SEEK_END);
-    lengthBytes = ftell(file);
-    fseek(file, 0, SEEK_SET);
-    pRawBytes = (LPBYTE)malloc(lengthBytes + sizeof(WCHAR));
-    if (nullptr == pRawBytes)
+    if (file != nullptr)
```
jianchun: nit: Is this check necessary? If yes, maybe can reorder the code a bit to make diff tool happy so it is easier to see what's changed.

bin/ch/Helpers.cpp
```diff
@@ -42,81 +41,60 @@ HRESULT Helpers::LoadScriptFromFile(LPCSTR filename, LPCSTR& contents, bool* isU
...
-            // unicode LE
-            isUtf8 = false;
-        }
-        else
+
+        //
+        // Read the entire content as a binary block.
+        //
+        fread(pRawBytes, sizeof(BYTE), lengthBytes, file);
+        *reinterpret_cast<WCHAR*>(pRawBytes + lengthBytes) = 0; // Null terminate it. Could be UTF16
```
jianchun: Now suppose we don't care about `WCHAR` null terminator any more. Can malloc +sizeof(BYTE) memory and only null terminate BYTE.

lib/Runtime/Base/ScriptContext.cpp
```diff
@@ -1591,7 +1591,10 @@ if (!sourceList)
 
         bool isLibraryCode = ((loadScriptFlag & LoadScriptFlag_LibraryCode) == LoadScriptFlag_LibraryCode);
 
-        bool utf8FromExternal = false;
+        // We assume that any buffer that wasn't passed in with LoadScriptFlag_Utf8Source is UTF16-LE
```
jianchun: I think you can just revert this file to the version prior to my change. That'll create less noise for future merging to master.

bin/ch/Helpers.cpp
jianchun: To be more robust (yes this was old code), consider adding check for `lengthBytes` (what if the file size is 0?), and maybe also avoid LPCWSTR and use a better 2-byte type (as @obastemur noted, WCHAR/char16_t can be more than 2 bytes...)

bin/ch/Helpers.cpp
digitalinfinity: Not sure if I understood this comment but updated it based on my interpretation of your comment 😄 Let me know if what I did was what you were suggesting?

bin/ch/Helpers.cpp
```diff
@@ -57,18 +57,25 @@ HRESULT Helpers::LoadScriptFromFile(LPCSTR filename, LPCSTR& contents, UINT* len
...
         //
-        fread(pRawBytes, sizeof(BYTE), lengthBytes, file);
-        *reinterpret_cast<WCHAR*>(pRawBytes + lengthBytes) = 0; // Null terminate it. Could be UTF16
+        size_t readBytes = fread(pRawBytes, sizeof(BYTE), lengthBytes, file);
+        if (readBytes == 0)
+        {
+            IfFailGo(E_FAIL);
+        }
 
+        *reinterpret_cast<BYTE*>(pRawBytes + lengthBytes) = 0; // Null terminate it. Could be UTF16
```
jianchun: `pRawBytes[lengthBytes] = 0`

bin/ch/Helpers.cpp
digitalinfinity: Yeah, I added that because static analysis complained about it 😞

digitalinfinity:
Thanks @obastemur @jianchun for the reviews- PR updated.

bin/ch/Helpers.cpp
jianchun: Not understand either -- I think this first version is a bit better than next version. `contentsRaw` is a temporary casted from `pRawBytes` (more straightforward than `contents`), return value `contents` is set in the last when successful (rather than in the middle).

jianchun:
LGTM (2 nit comments)

bin/ch/Helpers.cpp
```diff
@@ -42,81 +41,65 @@ HRESULT Helpers::LoadScriptFromFile(LPCSTR filename, LPCSTR& contents, bool* isU
...
-    *reinterpret_cast<WCHAR*>(pRawBytes + lengthBytes) = 0; // Null terminate it. Could be UTF16
-
-    //
-    // Read encoding, handling any conversion to Unicode.
-    //
-    // Warning: The UNICODE buffer for parsing is supposed to be provided by the host.
-    // This is not a complete read of the encoding. Some encodings like UTF7, UTF1, EBCDIC, SCSU, BOCU could be
-    // wrongly classified as ANSI
-    //
+    if (file != nullptr)
```
obastemur: NULL vs nullptr. How about 'if(file)' ?

bin/ch/Helpers.cpp
```diff
@@ -42,81 +41,65 @@ HRESULT Helpers::LoadScriptFromFile(LPCSTR filename, LPCSTR& contents, bool* isU
...
+            fwprintf(stderr, _u("out of memory"));
+            IfFailGo(E_OUTOFMEMORY);
         }
-        else
+
+        //
+        // Read the entire content as a binary block.
+        //
+        size_t readBytes = fread(pRawBytes, sizeof(BYTE), lengthBytes, file);
+        if (readBytes == 0)
```
obastemur: How about 'if( readBytes < lengthBytes * sizeof(BYTE))'

bin/ch/Helpers.cpp
```diff
@@ -42,81 +41,65 @@ HRESULT Helpers::LoadScriptFromFile(LPCSTR filename, LPCSTR& contents, bool* isU
...
-        {
-            *isUtf8Out = isUtf8;
-        }
         if (lengthBytesOut)
         {
             *lengthBytesOut = lengthBytes;
         }
     }
 
+    if (file != nullptr)
```
obastemur: NULL vs nullptr

bin/ch/Helpers.cpp
```diff
@@ -42,81 +41,65 @@ HRESULT Helpers::LoadScriptFromFile(LPCSTR filename, LPCSTR& contents, bool* isU
...
-        contents = pNarrow;
+        //
+        // Read encoding to make sure it's supported
+        //
+        // Warning: The UNICODE buffer for parsing is supposed to be provided by the host.
+        // This is not a complete read of the encoding. Some encodings like UTF7, UTF1, EBCDIC, SCSU, BOCU could be
+        // wrongly classified as ANSI
+        //
+        {
+            C_ASSERT(sizeof(WCHAR) == 2);
```
obastemur: sizeof(char16) is being bigger than 2 does not necessarily mean that we wouldn't support it.

Instead of casting from char to char16 below, you may compare the UTF-16 BOM using char type. At file level, char16 would never be more than 2 bytes. So the exception of char16 is being 32bits or even 64bits is independent from what it actually represents.

IMO converting from char16 to utf-8 is straightforward. Instead of marking it as unsupported, we can also add the conversion without requiring ICU etc.

> Some encodings like UTF7, UTF1, EBCDIC, SCSU, BOCU could be wrongly classified as ANSI

This mean that we support them or not ? We only check for utf-16 BOM below.

lib/Runtime/Base/ScriptContext.cpp
```diff
@@ -618,7 +618,7 @@ namespace Js
         // Stop profiling if present
         DeRegisterProfileProbe(S_OK, nullptr);
 #endif
-
+
```
obastemur: nit: extra spaces

lib/Runtime/Base/ScriptContext.cpp
```diff
@@ -3288,11 +3274,11 @@ if (!sourceList)
                     || entryPoint == DefaultDeferredDeserializeThunk || entryPoint == ProfileDeferredDeserializeThunk
                     || entryPoint == CrossSite::DefaultThunk || entryPoint == CrossSite::ProfileThunk);
 #else
-                Assert(entryPoint == DefaultDeferredParsingThunk
+                Assert(entryPoint == DefaultDeferredParsingThunk 
                     || entryPoint == DefaultDeferredDeserializeThunk
                     || entryPoint == CrossSite::DefaultThunk);
 #endif
-
+
```
obastemur: nit:extra spaces

obastemur:
Couple of places may require an update as given by the comments.

LGTM otherwise.

digitalinfinity:
Thanks @obastemur- made all the other updates except the ScriptContext spacing change since that files was just reverted to an earlier version of the file before @jianchun's commit, and didn't want to just have whitespace changes in that file.
Re UTF7/UTF1/other encodings- no, not supported, and we'll probably just crash if that's passed in. But again, ch is a test tool and so we can control the input encodings. For the question of whether Chakra should deal with any kind of encoding, we can track with https://github.com/Microsoft/ChakraCore/issues/1098

Thanks again for the review!

PR #1082, opened 2016-06-03
Rejit and rethunk issues

LouisLaf:
Fix a few rejits and rethunk issues: 
o  give up on bailOnNoProfile after 50 rejits (we can get in these situation on paths that always throw an exception) 
o  give up on bailOnNotNativeArray after 50 rejits (bailout does not fix the profile info for array types...) 
o  give up in general after 100 rejits...  Just Rethunk. 
o  for bailOnFailedFixedFieldType, the polymorphic cache state isn't affected if we just invalidated the guard.  Check if we are rethunking to the same JITed code, and rejit instead. 
o  for post-op bailOnImplicitCalls, the implicitCallFlags were not propagated correctly.  I've made a temporary fix to conservatively propagate it.

These were hitting on mobile.nytimes.com

For RS2, we should cleanup some of these with better fixes...

### DIFF
--- a/lib/Backend/BailOut.cpp
+++ b/lib/Backend/BailOut.cpp
@@ -2154,6 +2154,8 @@ void BailOutRecord::ScheduleFunctionCodeGen(Js::ScriptFunction * function, Js::S
         }
 
         Assert(!(rejitReason != RejitReason::None && reThunk));
+
+        profileInfo->IncRejitCount();
     }
 
     if(PHASE_FORCE(Js::ReJITPhase, executeFunction) && rejitReason == RejitReason::None)
--- a/lib/Backend/BailOut.cpp
+++ b/lib/Backend/BailOut.cpp
@@ -1050,6 +1050,15 @@ Js::Var BailOutRecord::BailOut(BailOutRecord const * bailOutRecord)
 #endif
 
     Js::JavascriptCallStackLayout *const layout = bailOutRecord->GetStackLayout();
+    Js::ScriptFunction * function = (Js::ScriptFunction *)layout->functionObject;
+
+    if (bailOutRecord->bailOutKind == IR::BailOutOnImplicitCalls)
+    {
+        function->GetScriptContext()->GetThreadContext()->CheckAndResetImplicitCallAccessorFlag();
+    }
+
+    Js::ImplicitCallFlags savedImplicitCallFlags = function->GetScriptContext()->GetThreadContext()->GetImplicitCallFlags();
+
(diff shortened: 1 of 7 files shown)


LouisLaf:
@abchatra and @rajatd , please review.

lib/Backend/BailOut.cpp
```diff
@@ -1050,6 +1050,15 @@ Js::Var BailOutRecord::BailOut(BailOutRecord const * bailOutRecord)
 #endif
 
     Js::JavascriptCallStackLayout *const layout = bailOutRecord->GetStackLayout();
+    Js::ScriptFunction * function = (Js::ScriptFunction *)layout->functionObject;
+
+    if (bailOutRecord->bailOutKind == IR::BailOutOnImplicitCalls)
```
Cellule: Does this applies for ImplicitException also ?

rajatd:
:+1:

lib/Backend/BailOut.cpp
```diff
@@ -2161,6 +2181,21 @@ void BailOutRecord::ScheduleFunctionCodeGen(Js::ScriptFunction * function, Js::S
         rejitReason = RejitReason::Forced;
     }
 
+    // REVIEW: Temporary fix for RS1.  Disable Rejiting if it looks like it is not fixing the problem.
+    //         For RS2, turn this into an assert and let's fix all these issues.
+    if (!reThunk && rejitReason != RejitReason::None)
+    {
+        if (executeFunction->GetDynamicProfileInfo()->GetRejitCount() >= 100)
+        {
+            reThunk = true;
```
abchatra: It is better not to set this.

abchatra: rejitReason = RejitReason::None; will take care of not rejitting. 

---
In reply to: [65753177](https://github.com/Microsoft/ChakraCore/pull/1082#discussion_r65753177) [](ancestors = 65753177)

lib/Backend/BailOut.cpp
```diff
@@ -2161,6 +2181,21 @@ void BailOutRecord::ScheduleFunctionCodeGen(Js::ScriptFunction * function, Js::S
         rejitReason = RejitReason::Forced;
     }
 
+    // REVIEW: Temporary fix for RS1.  Disable Rejiting if it looks like it is not fixing the problem.
+    //         For RS2, turn this into an assert and let's fix all these issues.
+    if (!reThunk && rejitReason != RejitReason::None)
```
abchatra: Nit: Do you want to take care of -force:rejit scenario? Feel free to ignore the comment.

PR #1084, opened 2016-06-03
ch: force error message output

obastemur:
If thrown error is not an object (i.e. text) `ch` doesn't print anything.
Print the text with some extra details instead of failing silently

### DIFF
--- a/bin/ch/WScriptJsrt.cpp
+++ b/bin/ch/WScriptJsrt.cpp
@@ -686,13 +686,22 @@ bool WScriptJsrt::PrintException(LPCSTR fileName, JsErrorCode jsErrorCode)
                 LPCWSTR errorStack = nullptr;
                 size_t errorStackLength = 0;
 
-                IfJsrtErrorFail(ChakraRTInterface::JsGetPropertyIdFromName(_u("stack"), &stackPropertyId), false);
-                IfJsrtErrorFail(ChakraRTInterface::JsGetProperty(exception, stackPropertyId, &stackProperty), false);
+                JsErrorCode errorCode = ChakraRTInterface::JsGetPropertyIdFromName(_u("stack"), &stackPropertyId);
 
-                IfJsrtErrorFail(ChakraRTInterface::JsGetValueType(stackProperty, &propertyType), false);
-
-                if (propertyType == JsUndefined)
+                if (errorCode == JsErrorCode::JsNoError)
                 {
+                    errorCode = ChakraRTInterface::JsGetProperty(exception, stackPropertyId, &stackProperty);
+                    if (errorCode == JsErrorCode::JsNoError)
+                    {
+                        errorCode = ChakraRTInterface::JsGetValueType(stackProperty, &propertyType);
+                    }
+                }
+
+                if (errorCode != JsErrorCode::JsNoError || propertyType == JsUndefined)
+                {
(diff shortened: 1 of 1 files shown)


bin/ch/WScriptJsrt.cpp
```diff
@@ -686,13 +686,21 @@ bool WScriptJsrt::PrintException(LPCSTR fileName, JsErrorCode jsErrorCode)
...
+                    if (errorCode == JsErrorCode::JsNoError)
+                    {
+                        errorCode = ChakraRTInterface::JsGetValueType(stackProperty, &propertyType);
+                    }
+                }
 
-                if (propertyType == JsUndefined)
+                if (errorCode != JsErrorCode::JsNoError || propertyType == JsUndefined)
                 {
+                    const char *fName = fileName != nullptr ? fileName : "(Unkown)";
```
akroshg: typo : **Unkown** to **unknown**

bin/ch/WScriptJsrt.cpp
```diff
@@ -686,13 +686,21 @@ bool WScriptJsrt::PrintException(LPCSTR fileName, JsErrorCode jsErrorCode)
...
+                    fprintf(stderr, "thrown at %s:\n^\n", fName);
```
akroshg: change fprintf to fwprintf (for consistency in the file) ?

akroshg:
Looks good to me otherwise.

bin/ch/WScriptJsrt.cpp
obastemur: Opps. Thanks!

bin/ch/WScriptJsrt.cpp
obastemur: I would like to keep this one as is since filename is not wchar and cross platform mixed wchar/char output fails. However gonna add comment there to make this point clear!

akroshg: sounds good.

obastemur:
@akroshg thanks for the review!

@digitalinfinity @jianchun anything to add?

bin/ch/WScriptJsrt.cpp
```diff
@@ -686,13 +686,22 @@ bool WScriptJsrt::PrintException(LPCSTR fileName, JsErrorCode jsErrorCode)
...
+                        errorCode = ChakraRTInterface::JsGetValueType(stackProperty, &propertyType);
+                    }
+                }
 
-                if (propertyType == JsUndefined)
+                if (errorCode != JsErrorCode::JsNoError || propertyType == JsUndefined)
                 {
+                    const char *fName = fileName != nullptr ? fileName : "(unknown)";
+                    // do not mix char/wchar. print them separately
+                    fprintf(stderr, "thrown at %s:\n^\n", fName);
```
digitalinfinity: You can do fwprintf with "%s" to print out a narrow string- should work with fwprintf since that's PAL provided I think, not sure if it works fine with fprintf though

digitalinfinity:
Just that one comment- LGTM otherwise

bin/ch/WScriptJsrt.cpp
obastemur: Yes I know that but it doesnt work.

obastemur:
@digitalinfinity thanks for review too!

PR #1085, opened 2016-06-03
cross-platform: Enable compiler optimization for non debug builds

obastemur:


digitalinfinity:
What does -O3 passed into the linker do? Also, this doesn't enable LTO for ChakraCore, right? AFAIK, we might need the LLVM gold plugin to do that: http://llvm.org/docs/GoldPlugin.html

obastemur:
> What does -O3 passed into the linker do?

Clang Compiler optimization for O3 -> http://clang.llvm.org/docs/CommandGuide/clang.html#code-generation-options . I see the wording in the PR confuses hence changed to compiler optimization.

> Also, this doesn't enable LTO for ChakraCore, right? AFAIK, we might need the LLVM gold plugin to do that: http://llvm.org/docs/GoldPlugin.html

No. IMO LTO / llvm gold plugins are no go for a project like this. Consider we would require that plugin is present for preparing a release build. I'm not sure the gain would suffice to take such a dependency.

obastemur:
I guess this is good to go?

digitalinfinity:
Thanks @obastemur for the documentation pointer. I understand the use of CMAKE_CXX_FLAGS getting -O3- my question was more around the passing -O3 to CMAKE_LINK_FLAGS. The docs are talking about the compiler, not the linker, yes? Sorry, I'm still missing something.
Re LTO, it would be worth doing to experiment to see what how much we gain if we use LTO (if the plugin is available on the system)- want to add that to your list 😉 ?

obastemur:
> Thanks @obastemur for the documentation pointer. I understand the use of CMAKE_CXX_FLAGS getting -O3- my question was more around the passing -O3 to CMAKE_LINK_FLAGS. 

Oops. So many copy paste there :) @digitalinfinity Thanks for the review!

PR #1086, opened 2016-06-03
Fix AllocationPolicyManager and enable a real memory limit on Posix/Unix

obastemur:
Story Behind:

 Posix malloc may not return NULL. Instead, process receives an address that is not actually available.
 Finally, kernel sends SIGKILL in case of the process uses that memory space. (see Posix malloc man)

 ArrayBuffer::Finalize was reporting that the external memory is already freed on Finalize.
 However, It wasn't really freed until garbage collection disposes the object behind.

Changes:

 - moved recycler reporting into ArrayBuffer::Dispose (after free call)
 - enabled a real memory limit on Posix/Unix (under ch)
 - Instead of throwing outofmemoryexception straight after `ReportExternalMemoryAllocation`
   initiating a manual garbage collection and trying it again. In case of failure, throwing an
   oom.

### DIFF
--- a/lib/Runtime/Library/ArrayBuffer.cpp
+++ b/lib/Runtime/Library/ArrayBuffer.cpp
@@ -447,22 +447,33 @@ namespace Js
                 buffer = (BYTE*)allocator(length);
                 if (buffer == nullptr)
                 {
-                    recycler->CollectNow<CollectOnTypedArrayAllocation>();
+                    recycler->ReportExternalMemoryFree(length);
+                }
+            }
+
+            if (buffer == nullptr)
+            {
+                recycler->CollectNow<CollectOnTypedArrayAllocation>();
+
+                if (recycler->ReportExternalMemoryAllocation(length))
+                {
                     buffer = (BYTE*)allocator(length);
                     if (buffer == nullptr)
                     {
                         recycler->ReportExternalMemoryFailure(length);
                     }
                 }
+                else
+                {
+                    JavascriptError::ThrowOutOfMemoryError(GetScriptContext());
+                }
             }
 
-            if (buffer == nullptr)
+            if (buffer != nullptr)
             {
-                JavascriptError::ThrowOutOfMemoryError(GetScriptContext());
+                bufferLength = length;
+                ZeroMemory(buffer, bufferLength);
             }
-
-            bufferLength = length;
-            ZeroMemory(buffer, bufferLength);
         }
     }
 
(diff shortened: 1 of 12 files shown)


curtisman:
Please do enough testing to make sure moving the freeing code from Dispose to Finalize is ok.
Otherwise LGTM.

obastemur:
> Please do enough testing to make sure moving the freeing code from Dispose to Finalize is ok.
Otherwise LGTM.

@curtisman @digitalinfinity Thanks for the reviews (we were discussing this over emails)
 
Although I went further than what we have on repo (tests), I'm not sure the enough testing part indeed. From what I've observed debugging all the way between Finalize and Dispose, I'm not sure what could be the problem. Actually, it was also a surprise to me that given memory wasn't freed on Finalize.

Do we have anything special to test ? or I'm going to merge as is.

lib/Runtime/Library/ArrayBuffer.cpp
```diff
@@ -445,24 +445,30 @@ namespace Js
             if (recycler->ReportExternalMemoryAllocation(length))
             {
                 buffer = (BYTE*)allocator(length);
-                if (buffer == nullptr)
+            }
+
+            if (buffer == nullptr)
+            {
+                recycler->CollectNow<CollectOnTypedArrayAllocation>();
+                if (recycler->ReportExternalMemoryAllocation(length))
```
Yongqu: you might be reporting the same allocation twice?

lib/Runtime/Library/ArrayBuffer.cpp
```diff
@@ -854,30 +860,34 @@ namespace Js
...
+            }
+            else
+            {
+              free(buffer);
+            }
+#else
+            free(buffer);
+#endif
+
             Recycler* recycler = GetType()->GetLibrary()->GetRecycler();
```
Yongqu: Is it because type was reclaimed at Dispose time instead of Recycler getting away? Agree Finalize should be a better time to free.

jianchun: Is it possible that the next object being finalized come and access this object? Access "buffer" that's freed?

Yongqu: This is arraybuffer. The virtual buffer (via heapalloc or CoTaskMemAlloc) can be accessed from one ArrayBuffer. You need to neuter it before give it to another thread.

jianchun: Does GC finalization allow reentrancy / script execution? If yes, suppose I have ArrayBuffer `a`, JS Object `obj` with `obj.a == a`, and an external object `x` created through `JsCreateExternalObject` with a finalizer, and `obj.x == x`. Could we finalize `a`, `x` in order, and `x` finalizer code access `a`, e.g. JsGetArrayBufferStorage?

lib/Runtime/Library/ArrayBuffer.cpp
obastemur: No. If it's bigger than the limit, it doesn't add up.

lib/Runtime/Library/ArrayBuffer.cpp
obastemur: How this scenario is different comparing to a garbage colllected variable?

jianchun: My question is more about the distinction between finalize and dispose. What state is an object supposed to be at after finalize? Should we set bufferLength to 0 and buffer to nullptr here?

lib/Runtime/Library/ArrayBuffer.cpp
jianchun: I also don't understand and still don't understand why you are not reporting twice. Once on line 445, then allocation failed, collect, and report again on line 453.

lib/Runtime/Library/ArrayBuffer.cpp
obastemur: > My question is more about the distinction between finalize and dispose. What state is an object supposed to be at after finalize? Should we set bufferLength to 0 and buffer to nullptr here?

IMO, its not needed. 

@curtisman ?

lib/Runtime/Library/ArrayBuffer.cpp
obastemur: You mean if we have nullptr from allocator? That makes sense.

lib/Runtime/Library/ArrayBuffer.cpp
curtisman: GC finalization does not allow reentrancy / script execution, but theoretically JSRT finalizers may be able to access those currently

Regardless, we should set those to 0 and nullptr for safety's sake.

obastemur:
Got the positive response from Crawler tests. Thanks for the reviews!

PR #1087, opened 2016-06-04
runtests.py: support python3

jianchun:
The script was hanging on python3 for some reason likely related to
byte data decoding.

Simply add `.decode('utf-8')` makes it work for both python3 and python2.

Changed baseline file reading to binary mode and `.decode('utf-8')`
manually. This bypasses python's builtin conversion and avoids resulted
`\r` / `\n` conversion difference.

### DIFF
--- a/test/runtests.py
+++ b/test/runtests.py
@@ -325,7 +325,7 @@ class TestVariant(object):
         start_time = datetime.now()
         try:
             timer.start()
-            js_output = proc.communicate()[0].replace('\r','')
+            js_output = proc.communicate()[0].decode('utf-8').replace('\r','')
             exit_code = proc.wait()
         finally:
             timer.cancel()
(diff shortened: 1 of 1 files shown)


test/runtests.py
```diff
@@ -356,16 +356,9 @@ def timeout_func(timeout_data):
...
-
-                # Cleanup carriage return
-                # todo: remove carriage return at the end of the line
-                #       or better fix ch to output same on all platforms
-                expected_output = re.sub('[\r]+\n', '\n', baseline_output)
-
-                # todo: implement wild cards support
+                with open(baseline, 'rb') as bs_file:
+                    expected_output = bs_file.read().decode('utf-8') \
+                                                    .replace('\r','')
```
obastemur: > This bypasses python's builtin conversion and avoids resulted \r / \n conversion difference.

Could you please elaborate? I didn't understand how it's different now ?

obastemur:
For many reasons, I disagree on supporting python3 going forward.

test/runtests.py
jianchun: > Could you please elaborate? I didn't understand how it's different now ?

The difference is from changing open file mode `r` to `rb`. (In mode `r` python may do end-of-line conversions.) For the same file we get `\r\n\n` and `\r\r\n` respectively in these two modes, resulting in `\n\n` and `\n` after our `replace`.

jianchun:
> For many reasons, I disagree on supporting python3 going forward.

I'm not to reason between python2 or python3. I look at python3 is because originally our internal tools requiring python3. So it'll be great if this also works with python3.

jianchun:
Please hold on reviewing this PR. I ran into hangs while testing and thought that was problem of my VM. Turned out it was due to utf8 decoding. I'll do further investigation.

jianchun:
Close for now, will reopen when I make it work.

PR #1090, opened 2016-06-06
xplat: re-enable GetThreadId assert in Recycler.cpp

xanlpz:
The method is implemented for Linux and POSIX/pthread now.

### DIFF
--- a/lib/Common/Memory/Recycler.cpp
+++ b/lib/Common/Memory/Recycler.cpp
@@ -1095,14 +1095,10 @@ bool Recycler::ExplicitFreeInternal(void* buffer, size_t size, size_t sizeCat)
 
 #if DBG || defined(RECYCLER_MEMORY_VERIFY) || defined(RECYCLER_PAGE_HEAP)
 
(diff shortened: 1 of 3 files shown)


xanlpz:
So obviously need to somehow include thread.hpp from PAL here. What's the best way to do that?

digitalinfinity:
@xanlpz It looks like GetCurrentThreadId is already implemented in pal/src/thread/thread.cpp (the one in thread.hpp that you mention implements CPalThread::GetCurrentThreadId). I'm not sure which implementation is better, I think the one in thread.cpp is slower. What do you think? If you decide to go with the one in thread.cpp, it's sufficient to just declare the API in ChakraPal.h.

xanlpz:
Unless I'm really confused the function in thread.cpp just calls one of the implementations in thread.hpp, so it's all just one thing. I'll declare it in ChakraPal.h then.

obastemur:
LGTM

digitalinfinity:
LGTM too- thanks @xanlpz!

PR #1091, opened 2016-06-06
Wasm 0xB format + Error handling

Cellule:
Includes @commonlisp changes for 0xB format and adds various fixes from running sexpr-wasm tests suite.
Closes #988 

This change also includes better error handling on Wasm Compilation error.
Instead of simply printing the error, it is now attached to the javascript error thrown so it can be caught and printed if desired by the javascript code.
Made a few improvements to the Lazy Trap scenario.

Added WasmOpCode `NYI` To discern Not Yet Implemented opcode vs unknown opcode in the error message.

@commonlisp @MikeHolman Can you guys review this please ?

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/1091)


Cellule:
This doesn't pass all tests in sexpr-wasm yet. There have been too many problem with different baseline between linux and Windows output.
I fixed most of those I could blame on the new format, but there is still investigation ongoing.

bin/ch/ChakraRtInterface.cpp
```diff
@@ -77,6 +77,8 @@ HINSTANCE ChakraRTInterface::LoadChakraDll(ArgInfo& argInfo)
     m_jsApiHooks.pfJsrtDoubleToNumber = (JsAPIHooks::JsrtDoubleToNumberPtr)GetProcAddress(library, "JsDoubleToNumber");
     m_jsApiHooks.pfJsrtGetExternalData = (JsAPIHooks::JsrtGetExternalDataPtr)GetProcAddress(library, "JsGetExternalData");
     m_jsApiHooks.pfJsrtCreateArray = (JsAPIHooks::JsrtCreateArrayPtr)GetProcAddress(library, "JsCreateArray");
+    m_jsApiHooks.pfJsrtCreateArrayBuffer = (JsAPIHooks::JsrtCreateArrayBufferPtr)GetProcAddress(library, "JsCreateArrayBuffer");
```
commonlisp: Yeah, this got clobbered during one of the merges. Do we want to restore it here or in a commit before the post-order?

Cellule: Yeah I could do it, but I doubt it's going to make a difference, unless someone tries to build directly one of my first 2 commits.

commonlisp:
Review status: 0 of 23 files reviewed at latest revision, 6 unresolved discussions.

---

*[bin/ch/ChakraRtInterface.cpp, line 80 \[r3\]](https://reviewable.io:443/reviews/microsoft/chakracore/1091#-KJajMoG5ZcdOrXH3jBT-r3-80:-KJfKxWdanD_Ar371Bqw:-834896264) ([raw file](https://github.com/microsoft/chakracore/blob/90169717f031e68960d9fa6e54bcd4af14461b0e/bin/ch/ChakraRtInterface.cpp#L80)):*
> Yeah I could do it, but I doubt it's going to make a difference, unless someone tries to build directly one of my first 2 commits.
</details>
The main issue is this. The current tip of WebAssembly is missing the JsCreateArrayBuffer initialization so one can't load WebAssembly (as in LoadBinaryFile) in the current tip of WebAssembly. It would be nice to separate this out so we have a commit with everything up to this point (i.e., mainly the result of the merge) except postorder and other 0xB format stuff for comparison. I can put this together if it sounds good to you.

---

*[lib/Parser/rterrors.h, line 278 \[r3\]](https://reviewable.io:443/reviews/microsoft/chakracore/1091#-KJfNSH9MHeOCpH_AH95:-KJfNSH9MHeOCpH_AH96:150585691) ([raw file](https://github.com/microsoft/chakracore/blob/90169717f031e68960d9fa6e54bcd4af14461b0e/lib/Parser/rterrors.h#L278)):*
> ```objc
> RT_ERROR_MSG(JSERR_AsmJsCompileError, 5148, "%s: Compiling asm.js failed.", "Compiling asm.js failed.", kjstError, 0)
> RT_ERROR_MSG(JSERR_ImmutablePrototypeSlot, 5149, "%s: Can't set the prototype of this object.", "Can't set the prototype of this object.", kjstTypeError, 0)
> RT_ERROR_MSG(JSERR_WasmCompileError, 5150, "Compiling wasm failed:  %s", "Compiling wasm failed.", kjstError, 0)
> ```

This does seem to depart from the format of the other error messages. Is rationale for handling the >= two-line errors?  I wonder if there is another way to format things.  There is also the matter of the assert_invalid error messages in the spec test-suite. Does it make sense to match those or at least keep our errors in a form that can be easily mapped to those?

---

*[lib/Runtime/Base/ScriptContext.cpp, line 1810 \[r3\]](https://reviewable.io:443/reviews/microsoft/chakracore/1091#-KJfRUM5xpqKf2yJYifn:-KJfRUM7QjVdSexbvvXb:-1652711298) ([raw file](https://github.com/microsoft/chakracore/blob/90169717f031e68960d9fa6e54bcd4af14461b0e/lib/Runtime/Base/ScriptContext.cpp#L1810)):*
> ```cpp
>         frameDisplay->SetItem(0, moduleMemoryPtr);
>         const auto createLazyTrap = [ctx](Wasm::WasmCompilationException* e) {
>             
> ```

Is this still needed for something?

---

*[lib/WasmReader/WasmByteCodeGenerator.cpp, line 473 \[r3\]](https://reviewable.io:443/reviews/microsoft/chakracore/1091#-KJkXsxSvDFSMS0hevmE:-KJkXsxT0Aam_LY8MEls:-1745343146) ([raw file](https://github.com/microsoft/chakracore/blob/90169717f031e68960d9fa6e54bcd4af14461b0e/lib/WasmReader/WasmByteCodeGenerator.cpp#L473)):*
> ```cpp
>         throw WasmCompilationException(_u("Operator NYI"));
>     default:
>         throw WasmCompilationException(_u("Unknown expression's op 0x%X"), op);
> ```

Should this be an Assert() now that you have the wnNYI case?

---

*[lib/WasmReader/WasmByteCodeGenerator.cpp, line 1317 \[r3\]](https://reviewable.io:443/reviews/microsoft/chakracore/1091#-KJkbZrSKLZzhpHgKtjl:-KJkbZrUJAlFc9jlVUQZ:-956875135) ([raw file](https://github.com/microsoft/chakracore/blob/90169717f031e68960d9fa6e54bcd4af14461b0e/lib/WasmReader/WasmByteCodeGenerator.cpp#L1317)):*
> ```cpp
> WasmBytecodeGenerator::PopEvalStack()
> {
>     return m_evalStack->Top()->Pop();
> ```

I have a follow-up commit that does explicit checks for empty here (also need it for below) since this should presumably be a compile error.

---

*[lib/WasmReader/WasmByteCodeGenerator.cpp, line 1348 \[r3\]](https://reviewable.io:443/reviews/microsoft/chakracore/1091#-KJkcVdAy9FNbhqwCenL:-KJkcVdCvG9PuS5FcmWf:-135114914) ([raw file](https://github.com/microsoft/chakracore/blob/90169717f031e68960d9fa6e54bcd4af14461b0e/lib/WasmReader/WasmByteCodeGenerator.cpp#L1348)):*
> ```cpp
>     if (size == -1)
>     {
>         size = 2048;
> ```

Is this used somewhere?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/microsoft/chakracore/1091)*

lib/WasmReader/WasmByteCodeGenerator.cpp
```diff
@@ -376,69 +384,96 @@ EmitInfo
...
+        break;
 #include "WasmKeywords.h"
-
+    case wnNYI:
+        // todo:: add the name of the operator that is NYI
+        throw WasmCompilationException(_u("Operator NYI"));
     default:
         throw WasmCompilationException(_u("Unknown expression's op 0x%X"), op);
     }
+    PushEvalStack(info);
```
MikeHolman: i am wondering if the expressiveness we allow is strictly greater than what the spec allows? it seems we would allow something like:
set_local a 1
set_local b 1
call hello
add

MikeHolman:
>     return EmitInfo();

looks like we still have issue with merging unreachable edges. not sure if that is something that is going to matter for angry bots though

---
Refers to: lib/WasmReader/WasmByteCodeGenerator.cpp:1101 in 9016971. [](commit_id = 90169717f031e68960d9fa6e54bcd4af14461b0e, deletion_comment = False)

lib/WasmReader/ModuleInfo.cpp
```diff
@@ -18,7 +18,7 @@ ModuleInfo::ModuleInfo(ArenaAllocator * alloc) :
     m_indirectFuncCount(0),
     m_exportCount(0),
     m_datasegCount(0),
-    m_startFunc(0) // XXX
+    m_startFunc((uint32)-1)
```
MikeHolman: nit: i would prefer we use something like Js::Constants::UninitializedValue for clarity

MikeHolman:
:shipit:

Cellule:
Review status: 0 of 23 files reviewed at latest revision, 7 unresolved discussions.

---

*[bin/ch/ChakraRtInterface.cpp, line 80 \[r3\]](https://reviewable.io:443/reviews/microsoft/chakracore/1091#-KJajMoG5ZcdOrXH3jBT-r3-80:-KJlOd1R6J-rPdmNNdpN:1095069592) ([raw file](https://github.com/microsoft/chakracore/blob/90169717f031e68960d9fa6e54bcd4af14461b0e/bin/ch/ChakraRtInterface.cpp#L80)):*
<details><summary><i>Previously, commonlisp (George Kuan) wrote…</i></summary>
> The main issue is this. The current tip of WebAssembly is missing the JsCreateArrayBuffer initialization so one can't load WebAssembly (as in LoadBinaryFile) in the current tip of WebAssembly. It would be nice to separate this out so we have a commit with everything up to this point (i.e., mainly the result of the merge) except postorder and other 0xB format stuff for comparison. I can put this together if it sounds good to you.
</details>
Well I can simply commit this change right now and then rebase my changes on top of it

---

*[lib/Parser/rterrors.h, line 278 \[r3\]](https://reviewable.io:443/reviews/microsoft/chakracore/1091#-KJfNSH9MHeOCpH_AH95:-KJlPF5IfYgMGM4EXKyl:-2063364014) ([raw file](https://github.com/microsoft/chakracore/blob/90169717f031e68960d9fa6e54bcd4af14461b0e/lib/Parser/rterrors.h#L278)):*
<details><summary><i>Previously, commonlisp (George Kuan) wrote…</i></summary>
> This does seem to depart from the format of the other error messages. Is rationale for handling the >= two-line errors?  I wonder if there is another way to format things.  There is also the matter of the assert_invalid error messages in the spec test-suite. Does it make sense to match those or at least keep our errors in a form that can be easily mapped to those?
</details>
I wanted to make something that is easier to read, mainly for debugging purposes.
It doesn't have to be the final format. I think we can revisit this later down the road.

---

*[lib/Runtime/Base/ScriptContext.cpp, line 1810 \[r3\]](https://reviewable.io:443/reviews/microsoft/chakracore/1091#-KJfRUM5xpqKf2yJYifn:-KJlPbWS2L6apL9yHGSa:-1291755328) ([raw file](https://github.com/microsoft/chakracore/blob/90169717f031e68960d9fa6e54bcd4af14461b0e/lib/Runtime/Base/ScriptContext.cpp#L1810)):*
<details><summary><i>Previously, commonlisp (George Kuan) wrote…</i></summary>
> Is this still needed for something?
</details>
I forgot to remove it, thanks

---

*[lib/WasmReader/WasmByteCodeGenerator.cpp, line 473 \[r3\]](https://reviewable.io:443/reviews/microsoft/chakracore/1091#-KJkXsxSvDFSMS0hevmE:-KJlQ1Cim8i_Lb-TUtTS:-591256058) ([raw file](https://github.com/microsoft/chakracore/blob/90169717f031e68960d9fa6e54bcd4af14461b0e/lib/WasmReader/WasmByteCodeGenerator.cpp#L473)):*
<details><summary><i>Previously, commonlisp (George Kuan) wrote…</i></summary>
> Should this be an Assert() now that you have the wnNYI case?
</details>
Since any user can pass anything as binary, it is not an assertion. It is very possible in real world to pass unknown operators.
Therefore it is an exception, not a potential problem with our code.

---

*[lib/WasmReader/WasmByteCodeGenerator.cpp, line 1317 \[r3\]](https://reviewable.io:443/reviews/microsoft/chakracore/1091#-KJkbZrSKLZzhpHgKtjl:-KJlQQsiOcFo1PyyIDZ5:-1790034715) ([raw file](https://github.com/microsoft/chakracore/blob/90169717f031e68960d9fa6e54bcd4af14461b0e/lib/WasmReader/WasmByteCodeGenerator.cpp#L1317)):*
<details><summary><i>Previously, commonlisp (George Kuan) wrote…</i></summary>
> I have a follow-up commit that does explicit checks for empty here (also need it for below) since this should presumably be a compile error.
</details>
I did see them, I will let you add this in a future PR

---

*[lib/WasmReader/WasmByteCodeGenerator.cpp, line 1348 \[r3\]](https://reviewable.io:443/reviews/microsoft/chakracore/1091#-KJkcVdAy9FNbhqwCenL:-KJlR0oigSaWgTVd_Lk-:581961955) ([raw file](https://github.com/microsoft/chakracore/blob/90169717f031e68960d9fa6e54bcd4af14461b0e/lib/WasmReader/WasmByteCodeGenerator.cpp#L1348)):*
<details><summary><i>Previously, commonlisp (George Kuan) wrote…</i></summary>
> Is this used somewhere?
</details>
copy/paste error, thanks

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/microsoft/chakracore/1091)*

PR #1092, opened 2016-06-06
MemOp Loop Memory allocation

Cellule:
Closes #1015 (retarget changes to master)

Allocate Memop memory only if needed.
I did some tracing on the benchmarks.
Before this change there were over 1000 occurrences of the memop memory to be allocated, but not used.
After this change there were 85 occurrences.

The main reason is that right now any loop with 2 blocks will have all the memop memory info allocated.
The downside of this change is that it adds a boolean to every loops.
I was thinking doing something funky on the pointer instead (like -1 means do not do memop), but I fear this might be confusing just to save 1 boolean.

This is not really fixing a bug, and there doesn't seems to be a perf impact, but I came across this as I was fixing something else and it bugged me.
I think it's worth changing.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/1092)


rajatd:
:+1:

PR #1094, opened 2016-06-06
Minor cleanup to closure capture of catch variables

pleath:
Clean up some long-standing oddness regarding closure capture of catch variables. We've been treating them as captured in certain unnecessary cases, such as catch inside with. This change gets rid of unnecessary bad CQ, gets rid of the duoplication of catch symbols in the body scope when they're not captured, and simplifies things a little for redeferral.


ianwjhalliday:
:shipit:

pleath:
@dotnet-bot please test

PR #1096, opened 2016-06-07
Fix arena access violation for -profile option

ThomsonTan:
Fixed -profile for NativeCodeGenerator, disabled it for GC phase as workaround.


lib/Common/Memory/Recycler.cpp
```diff
@@ -5240,6 +5240,7 @@ Recycler::FinishConcurrentCollect(CollectionFlags flags)
 
 #ifdef PROFILE_EXEC
     Js::Phase concurrentPhase = Js::ConcurrentCollectPhase;
+    static_cast<Js::Phase>(concurrentPhase);
```
curtisman: What is this line for?

lib/Common/Memory/Recycler.h
```diff
@@ -2372,7 +2375,8 @@ struct ForceLeafAllocator<RecyclerNonLeafAllocator>
     typedef RecyclerLeafAllocator AllocatorType;
 };
 
-#ifdef PROFILE_EXEC
+// TODO: enable -profile for GC phases
+#if defined(PROFILE_EXEC) && 0
```
curtisman: What is blocking from making this work? 
AFAIK, this used to work, but something might have broke it?

lib/Common/Memory/HeapInfo.cpp
```diff
@@ -881,6 +881,7 @@ HeapInfo::AddLargeHeapBlock(size_t size)
 void HeapInfo::SweepBuckets(RecyclerSweep& recyclerSweep, bool concurrent)
 {
     Recycler * recycler = recyclerSweep.GetRecycler();
+    static_cast<Recycler*>(recycler);
```
curtisman: What is this line for?

lib/Common/Memory/Recycler.h
```diff
@@ -979,6 +978,10 @@ class Recycler
     PageAllocator backgroundProfilerPageAllocator;
     DListBase<ArenaAllocator> backgroundProfilerArena;
 #endif
+
+    // destruct autoHeap after backgroundProfilerPageAllocator;
+    HeapInfo autoHeap;
```
curtisman: Can you explain why this needs to be after backgroundProfilerPageAllocator.  Those seems to me to be unrelated?

lib/Backend/NativeCodeGenerator.cpp
```diff
@@ -2802,25 +2810,40 @@ void
...
             codegenProfiler = codegenProfiler->next;
         }
     }
     else
     {
         //Merge all the codegenProfiler for single snapshot.
-        codegenProfiler = codegenProfiler->next;
-        while (codegenProfiler)
+        Js::ScriptContextProfiler* mergeToProfiler = codegenProfiler;
+        while (mergeToProfiler != nullptr && !mergeToProfiler->IsInitialized())
```
curtisman: Why an extra loop here to find the first initialized profiler?
The loop below should be able to handle that already.  
This loop seems redundant.

lib/Common/Memory/Recycler.cpp
ThomsonTan: As -profile is disabled for GC phases, this is workaround for the unreferenced local variable. I have plan to enable -profile for GC phases so this is temporary.

---
In reply to: [66280866](https://github.com/Microsoft/ChakraCore/pull/1096#discussion_r66280866) [](ancestors = 66280866)

lib/Common/Memory/Recycler.h
ThomsonTan: Background GC threads use the one profiler from the same recycler object. Profiler object is not thread safe (phase push/pop cannot match if accessed from multiple threads).

---
In reply to: [66281159](https://github.com/Microsoft/ChakraCore/pull/1096#discussion_r66281159) [](ancestors = 66281159)

curtisman:
I think this change should go to master instead of release/1.2?

lib/Common/Memory/HeapInfo.cpp
ThomsonTan: The same as above, a temporary workaround for unreferenced local variable.

---
In reply to: [66281536](https://github.com/Microsoft/ChakraCore/pull/1096#discussion_r66281536) [](ancestors = 66281536)

lib/Backend/NativeCodeGenerator.cpp
```diff
@@ -2802,25 +2810,40 @@ void
...
 {
     Js::ScriptContextProfiler *codegenProfiler = this->backgroundCodeGenProfiler;
+
     if (Js::Configuration::Global.flags.Verbose)
     {
         //Print individual CodegenProfiler information in verbose mode
         while (codegenProfiler)
         {
-            codegenProfiler->ProfilePrint(Js::Configuration::Global.flags.Profile.GetFirstPhase());
+            if (codegenProfiler->IsInitialized())
```
curtisman: This is already checked in ScriptContextProfiler::ProfilePrint.  No need to add a check here again

lib/Common/Memory/Recycler.h
ThomsonTan: backgroundProfilerArena allocated heap block from autoHeap. If autoHeap destructed before backgroundProfilerArena, the destructor for the latter will run on some freed heap blocks.

---
In reply to: [66281817](https://github.com/Microsoft/ChakraCore/pull/1096#discussion_r66281817) [](ancestors = 66281817)

lib/Backend/NativeCodeGenerator.cpp
ThomsonTan: Below loop doesn't ensure the first profiler (mergeToProfiler) is initialized.

---
In reply to: [66282264](https://github.com/Microsoft/ChakraCore/pull/1096#discussion_r66282264) [](ancestors = 66282264)

ThomsonTan:
Thanks for the reminder. I am experimenting JIT throughput based on release/1.2 so I try to push to this branch. Is there any other reason this should go to master?

---
In reply to: [224635925](https://github.com/Microsoft/ChakraCore/pull/1096#issuecomment-224635925) [](ancestors = 224635925)

lib/Backend/NativeCodeGenerator.cpp
ThomsonTan: You are right, this check is redundant. No merge action here so the check in ScriptContextProfiler::ProfilePrint is enough.

---
In reply to: [66284121](https://github.com/Microsoft/ChakraCore/pull/1096#discussion_r66284121) [](ancestors = 66284121)

lib/Common/Memory/Recycler.h
curtisman: Arena doesn't allocate stuff in recycler's autoHeap.

PR #1097, opened 2016-06-07
JsDiagGetStackTrace should only give frames which were valid when halt was triggered.

agarwal-sandeep:
JsrtDebugStackFrames::StackFrames should only get frame pointers if the ScriptContext/ProbeContainer had valid stack when halt was triggered. Since user can execute JavaScript code after halt any frames after halt are transient so they should be skipped

### DIFF
--- a/lib/Jsrt/JsrtDebugManager.cpp
+++ b/lib/Jsrt/JsrtDebugManager.cpp
@@ -201,7 +201,7 @@ void JsrtDebugManager::ReportScriptCompile(Js::JavascriptFunction* scriptFunctio
             jsDiagDebugEvent = JsDiagDebugEventSourceCompile;
         }
 
-        this->CallDebugEventCallback(jsDiagDebugEvent, eventDataObject, scriptContext);
+        this->CallDebugEventCallback(jsDiagDebugEvent, eventDataObject, scriptContext, false /*isBreak*/);
     }
 }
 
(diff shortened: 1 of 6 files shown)


agarwal-sandeep:
@akroshg please take a look

lib/Runtime/Debug/ProbeContainer.h
```diff
@@ -91,6 +91,7 @@ namespace Js
         void Initialize(ScriptContext* pScriptContext);
         void Close();
 
+        bool HaveFramePointers() const { return this->framePointers != nullptr; }
```
akroshg: nit: HasFramePointers

lib/Jsrt/JsrtDebuggerObject.cpp
```diff
@@ -801,24 +801,30 @@ Js::JavascriptArray * JsrtDebugStackFrames::StackFrames(Js::ScriptContext * scri
     tempScriptContext != nullptr && tempScriptContext->IsScriptContextInDebugMode();
         tempScriptContext = tempScriptContext->next)
     {
-        Js::WeakDiagStack * framePointers = tempScriptContext->GetDebugContext()->GetProbeContainer()->GetFramePointers();
-        if (framePointers != nullptr)
+        Js::ProbeContainer* probeContainer = tempScriptContext->GetDebugContext()->GetProbeContainer();
+
+        // Only give frames which were preset when break was triggered as we might have other user (non-library) scripts executed after break
+        if (probeContainer->HaveFramePointers())
```
akroshg: I am not sure if I follow this. We create the stackframe for other scriptcontext only when they were asked for. So when do the other scriptcontext will have their frame pointers populated?

lib/Jsrt/JsrtDebugUtils.cpp
```diff
@@ -27,9 +27,7 @@ void JsrtDebugUtils::AddFileNameOrScriptTypeToObject(Js::DynamicObject* object,
 
         Js::FunctionBody* anyFunctionBody = utf8SourceInfo->GetAnyParsedFunction();
 
-        Assert(anyFunctionBody != nullptr);
-
-        LPCWSTR sourceName = anyFunctionBody->GetSourceName();
+        LPCWSTR sourceName = (anyFunctionBody != nullptr) ? anyFunctionBody->GetSourceName() : Js::Constants::UnknownScriptCode;
```
akroshg: Is it an unrelated fix? Is there a possibility that we have utf8sourcecontext without a single function? it should atleast have global function.

lib/Jsrt/JsrtDebugManager.cpp
```diff
@@ -364,8 +364,21 @@ void JsrtDebugManager::CallDebugEventCallback(JsDiagDebugEvent debugEvent, Js::D
...
-        this->debugEventCallback(debugEvent, eventDataObject, this->callbackState);
+        if (isBreak)
+        {
+            // If we are reporting break we should clear all objects after call returns
+            AutoClear autoClear(this);
+
+            // Save the frame address, when asking for stack we will only give stack which is under this address
+            // because host can execute javascript after break which should not be part of stack.
+            void *frameAddress = _AddressOfReturnAddress();
+            scriptContext->GetThreadContext()->GetDebugManager()->SetDispatchHaltFrameAddress(frameAddress);
```
akroshg: It would be good idea to do SetDispatchHaltFrameAddress(nullptr) after the debugEventCallback call below, so that threadContext->GetDebugManager()->GetDispatchHaltFrameAddress() can return null once the break is over.

akroshg:
Looks good to me

PR #1099, opened 2016-06-08
xplat: runtests.py improvements

jianchun:
1. Add error handling

 Python multiprocessing parallel processes do not print anything if an
 unhandled exception occurs. This causes the script to "hang" silently.
 Added 2 levels of try/except so that the script no longer hangs on
 exception. Script exceptions and backtraces will be reported and logged.

2. Switch to python3 `bytes` for output/baseline diff

 This has no effect on python 2, but enables the script on python 3.
 Capture output as bytes (subprocess communicate() returned type), load
 baseline files in mode 'rb' and diff `bytes` type directly. This avoids
 issues caused by encoding/end-of-line conversions.

3. Refactor, notably move 'folder' data to an attribute of test so that
we don't need to pass folder around.

4. Add a tag "xplatslow" to 2 tests (temporarily, so can be excluded on
command line: `--not-tag xplatslow`).

### DIFF
--- a/test/runtests.py
+++ b/test/runtests.py
@@ -11,6 +11,7 @@ from threading import Timer
 import sys
 import os
 import subprocess as SP
+import traceback
 import argparse
 import xml.etree.ElementTree as ET
 import re
@@ -112,6 +113,15 @@ if sys.platform != 'win32':
 not_compile_flags = set(['-simdjs']) \
     if sys.platform != 'win32' else None
 
+# use tags/not_tags/not_compile_flags as case-insensitive
+def lower_set(s):
+    return set([x.lower() for x in s] if s else [])
+
+tags = lower_set(tags)
+not_tags = lower_set(not_tags)
+not_compile_flags = lower_set(not_compile_flags)
+
+
 class LogFile(object):
     def __init__(self, log_file_path = None):
         self.file = None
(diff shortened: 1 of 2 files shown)


test/runtests.py
```diff
@@ -347,65 +381,60 @@ def timeout_func(timeout_data):
...
-                    baseline_output = bs_file.read()
-
-                # Cleanup carriage return
-                # todo: remove carriage return at the end of the line
-                #       or better fix ch to output same on all platforms
-                expected_output = re.sub('[\r]+\n', '\n', baseline_output)
-
-                # todo: implement wild cards support
+                with open(baseline, 'rb') as bs_file:
+                    expected_output = bs_file.read().replace(b'\r', b'')
```
obastemur: Old implementation only replaces CR(+)LF to LF, however the new one removes all the CR from the string. What would happen, if we had a test string with CR inside?

test/typedarray/rlexe.xml
```diff
@@ -215,13 +215,13 @@ Below test fails with difference in space. Investigate the cause and re-enable t
   <test>
     <default>
       <files>allocation.js</files>
-      <tags>typedarray,exclude_arm</tags>
+      <tags>typedarray,exclude_arm,xplatslow</tags>
```
obastemur: The reason these tests were slow in fact due to OOM fixed by https://github.com/Microsoft/ChakraCore/pull/1086 .At this point I don't see a reason to disable these.

obastemur:
> ...This avoids issues caused by encoding/end-of-line conversions.

Can you please elaborate how this actually works? IMO, binary mode doesn't make a difference in terms of utf-16 BOM / endiannes. For end-of-line left a comment in place.

jianchun:
> Can you please elaborate how this actually works? IMO, binary mode doesn't make a difference in terms of utf-16 BOM / endiannes. For end-of-line left a comment in place.

I mean diffing bytes rather than diffing decoded Unicode strings. I sent and closed a PR last week in which I `decode('utf-8')` and then diff the strings. That didn't work as some corner cases were not in utf-8.

obastemur:
>> I mean diffing bytes rather than diffing decoded Unicode strings.

>> load baseline files in mode 'rb' and diff bytes type directly.

@jianchun , I must be missing that part somehow. Could you please point me out where actually you do that?

And eventually how it's going to help with saved baseline files ? Especially, considering the fact utf-16 BOM defines the order there. (let's forget about utf-8 BOM)

test/runtests.py
jianchun: One reason I changed this is to be consistent with what we do when capturing ch output (`proc.communicate()[0].replace`). Testing CR with baseline is corner case / problematic and we should not do.

jianchun:
> @jianchun , I must be missing that part somehow. Could you please point me out where actually you do that?

All the added `b`.

> And eventually how it's going to help with saved baseline files ? Especially, considering the fact utf-16 BOM defines the order there. (let's forget about utf-8 BOM)

It does not help with the problem of baseline files in different encodings. Existing ch output and baseline files are not in certain encoding. We could fix that in future and then we can decode here; but currently decoding (ch output) runs into errors.

test/typedarray/rlexe.xml
jianchun: That's great! I'll hold on this / revert after your fix.

obastemur:
:shipit:

jianchun:
@dotnet-bot test this please

PR #1101, opened 2016-06-09
Remove sexpr parser

Cellule:
It has been a while now that the Sexpr parser was not working in our engine.
I do not think we will try to make it work anytime soon, nor would we do it this way.
I think a better future proof solution would be to have a sexpr->wasm converter in our engine and then process the resulting binary.

This way we can remove the abstraction for the reader and simply use the binary reader.
I am sure there are other places where we had some weird logic to support both parser that I didn't catch, feel free to point them to me.

This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/1101)


Cellule:
@commonlisp @MikeHolman Can you guys review this please ?

lib/WasmReader/WasmByteCodeGenerator.cpp
```diff
@@ -552,34 +552,18 @@ WasmBytecodeGenerator::EmitBlock()
...
-            EmitInfo info = EmitExpr(op);
-            ReleaseLocation(&info);
-            op = m_reader->ReadFromBlock();
-        } while (op != wnLIMIT);
+        nonEmpty = true;
+        tmpInfo = EmitExpr(op);
     }
+    ExitEvalStackScope();
+    // block yields last value
+    if (nonEmpty) blockInfo = tmpInfo;
```
MikeHolman: this was existing code, but it seems like it doesn't do anything?

MikeHolman:
:shipit:

lib/WasmReader/WasmByteCodeGenerator.cpp
Cellule: Well it is used to yield the last value of the block, but it could be refactored to simply use blockInfo directly and an empty block wouldn't get an assignment

PR #1107, opened 2016-06-09
work for iteratorclose

akroshg:
This changeset enables IteratorClose scenario for destructuring, for..of
loop and built-ins (Promise, Array.from, Map, WeakMap, Set and WeakSet).
Closing the iterator means calling the return function of the iterator. The implementation for destructuring and for..of is done in the Bytecode level.
Destructuring : In the array destructuring pattern the try..finally
bytecodes are emitted in order to capture the abrupt completion of the
destructuring elements assignment.
For..of : The loop is surrounded by try..catch..finally. The catch block
is to tackle the abrupt completion due to exception and finally block is
to tackle the scenarios which involves break, return, continue.
Added unittest.

### DIFF
--- a/lib/Runtime/ByteCode/ByteCodeEmitter.cpp
+++ b/lib/Runtime/ByteCode/ByteCodeEmitter.cpp
@@ -158,10 +158,10 @@ bool IsArguments(ParseNode *pnode)
             pnode = pnode->sxTri.pnode3;
             break;
 
-        //
-        // Cases where we don't check for "arguments" yet.
-        // Assume that they might have it. Disable the optimization is such scenarios
-        //
+            //
+            // Cases where we don't check for "arguments" yet.
+            // Assume that they might have it. Disable the optimization is such scenarios
+            //
         case knopList:
         case knopObject:
         case knopVarDecl:
@@ -245,7 +245,7 @@ static const Js::OpCode nopToCMOp[knopLim] =
 
 Js::OpCode ByteCodeGenerator::ToChkUndeclOp(Js::OpCode op) const
 {
-    switch(op)
+    switch (op)
     {
     case Js::OpCode::StLocalSlot:
         return Js::OpCode::StLocalSlotChkUndecl;
@@ -2043,12 +2043,12 @@ void ByteCodeGenerator::LoadAllConstants(FuncInfo *funcInfo)
                 {
                     scopeLocation = funcInfo->funcExprScope->GetLocation();
                     this->m_writer.Property(Js::OpCode::StFuncExpr, sym->GetLocation(), scopeLocation,
-                                            funcInfo->FindOrAddReferencedPropertyId(sym->GetPosition()));
+                        funcInfo->FindOrAddReferencedPropertyId(sym->GetPosition()));
                 }
                 else
                 {
                     this->m_writer.ElementU(Js::OpCode::StLocalFuncExpr, sym->GetLocation(),
-                                            funcInfo->FindOrAddReferencedPropertyId(sym->GetPosition()));
+                        funcInfo->FindOrAddReferencedPropertyId(sym->GetPosition()));
                 }
             }
             else if (ShouldTrackDebuggerMetadata())
@@ -2437,10 +2437,10 @@ void ByteCodeGenerator::EmitInternalScopedSlotLoad(FuncInfo *funcInfo, Scope *sc
 
     if (envIndex != -1)
     {
-        this->m_writer.SlotI2(opcode, symbolRegister, envIndex + Js::FrameDisplay::GetOffsetOfScopes()/sizeof(Js::Var), slot, profileId);
+        this->m_writer.SlotI2(opcode, symbolRegister, envIndex + Js::FrameDisplay::GetOffsetOfScopes() / sizeof(Js::Var), slot, profileId);
     }
     else if (scopeLocation != Js::Constants::NoRegister &&
-             (scopeLocation == funcInfo->frameSlotsRegister || scopeLocation == funcInfo->frameObjRegister))
+        (scopeLocation == funcInfo->frameSlotsRegister || scopeLocation == funcInfo->frameObjRegister))
     {
         this->m_writer.SlotI1(opcode, symbolRegister, slot, profileId);
     }
@@ -2478,10 +2478,10 @@ void ByteCodeGenerator::EmitInternalScopedSlotStore(FuncInfo *funcInfo, Js::RegS
 
     if (envIndex != -1)
     {
-        this->m_writer.SlotI2(opcode, symbolRegister, envIndex + Js::FrameDisplay::GetOffsetOfScopes()/sizeof(Js::Var), slot);
+        this->m_writer.SlotI2(opcode, symbolRegister, envIndex + Js::FrameDisplay::GetOffsetOfScopes() / sizeof(Js::Var), slot);
     }
     else if (scopeLocation != Js::Constants::NoRegister &&
-             (scopeLocation == funcInfo->frameSlotsRegister || scopeLocation == funcInfo->frameObjRegister))
+        (scopeLocation == funcInfo->frameSlotsRegister || scopeLocation == funcInfo->frameObjRegister))
     {
         this->m_writer.SlotI1(opcode, symbolRegister, slot);
     }
@@ -3544,7 +3544,7 @@ void ByteCodeGenerator::EmitScopeList(ParseNode *pnode, ParseNode *breakOnBodySc
     {
         switch (pnode->nop)
         {
-            case knopFncDecl:
+        case knopFncDecl:
 #ifndef TEMP_DISABLE_ASMJS
             if (pnode->sxFnc.GetAsmjsMode())
             {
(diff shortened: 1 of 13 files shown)


akroshg:
Perf seems to be flat in general. However the pdfjs.js shows regression if I ran the whole octane suite. Though if I run the same test individual it does not show any regression. That test does not have any usage of for..of to have it regressed.

@abchatra @ianwjhalliday @pleath Please have a look.

lib/Runtime/ByteCode/ByteCodeEmitter.cpp
```diff
@@ -158,10 +158,10 @@ bool IsArguments(ParseNode *pnode)
...
             break;
 
-        //
-        // Cases where we don't check for "arguments" yet.
-        // Assume that they might have it. Disable the optimization is such scenarios
-        //
+            //
+            // Cases where we don't check for "arguments" yet.
+            // Assume that they might have it. Disable the optimization is such scenarios
+            //
```
akroshg: I didn't mean some indentation changes - VS have done that for me. If you have concerns regarding that let me know.

test/es6/iteratorclose.js
```diff
@@ -0,0 +1,950 @@
...
+            returnCount = 0;
+            shouldNextThrow = true;
+            assert.throws(function () { var [...a] = iterable; }, Error, "Assigning to rest head can throw", "Exception from next function");
+            assert.throws(function () { [...a] = iterable; }, Error, "Assigning to rest head can throw", "Exception from next function");
+            shouldNextThrow = false;
+            assert.areEqual(returnCount, 0);
+		}
+	},    
+    {
+		name : "For..of - validation of calling .return function on abrupt loop break",
```
akroshg: test262 has some good test cases for the iteratorclose. I'll report once I get the result on them.

akroshg:
Will trigger some internal test runs for validation.

lib/Runtime/ByteCode/ByteCodeEmitter.cpp
```diff
@@ -8540,6 +8676,36 @@ void EmitIteratorNext(Js::RegSlot itemLocation, Js::RegSlot iteratorLocation, Js
     byteCodeGenerator->Writer()->MarkLabel(skipThrow);
 }
 
+// Generating
+// if (hasReturnFunction) {
+//     value = Call Retrun;
```
ianwjhalliday: typo

lib/Runtime/ByteCode/ByteCodeEmitter.cpp
```diff
@@ -6321,8 +6343,122 @@ void EmitDestructuredArray(
...
+
+    byteCodeGenerator->SetHasFinally(true);
+    byteCodeGenerator->SetHasTry(true);
+    byteCodeGenerator->TopFuncInfo()->byteCodeFunction->SetDontInline(true);
+
+    // Insert try node here 
+    Js::ByteCodeLabel finallyLabel = byteCodeGenerator->Writer()->DefineLabel();
+    byteCodeGenerator->Writer()->RecordCrossFrameEntryExitRecord(true);
+
+    byteCodeGenerator->Writer()->Br(Js::OpCode::TryFinally, finallyLabel);
```
akroshg: Note to self : Generator case needs to be handled in the destructuring case as well. I have made those changes in my local branch.

lib/Runtime/ByteCode/ByteCodeEmitter.cpp
akroshg: will do.

lib/Runtime/Language/JavascriptOperators.cpp
```diff
@@ -10443,6 +10443,27 @@ namespace Js
         return RecyclableObject::FromVar(iterator);
     }
 
+    void JavascriptOperators::IteratorClose(RecyclableObject* iterator, ScriptContext* scriptContext)
+    {
+        try
+        {
+            Var func = JavascriptOperators::GetProperty(iterator, PropertyIds::return_, scriptContext);
+
+            if (func != nullptr && JavascriptConversion::IsCallable(func))
```
ianwjhalliday: `GetProperty` can't return `nullptr`, can it?

lib/Runtime/Library/JavascriptPromise.cpp
```diff
@@ -201,16 +201,18 @@ namespace Js
...
+        bool shouldCallReturn = false;
         try
         {
             // 4. Let iterator be GetIterator(iterable).
-            RecyclableObject* iterator = JavascriptOperators::GetIterator(iterable, scriptContext);
+            iterator = JavascriptOperators::GetIterator(iterable, scriptContext);
             Var next;
             values = library->CreateArray(0);
 
             while (JavascriptOperators::IteratorStepAndValue(iterator, scriptContext, &next))
```
ianwjhalliday: Is there a reason not to use `JavascriptOperators::DoIteratorStepAndValue` here instead of doing it manually?  This feels error prone.  Same with `EntryRace` below.

lib/Runtime/Library/JavascriptSet.cpp
```diff
@@ -88,12 +88,9 @@ namespace Js
 
         if (iter != nullptr)
         {
-            Var nextItem;
-
-            while (JavascriptOperators::IteratorStepAndValue(iter, scriptContext, &nextItem))
```
ianwjhalliday: Is `IteratorStepAndValue` used anywhere anymore?  Can we delete it and rename `DoIteratorStepAndValue` to `IteratorStepAndValue`?

lib/Runtime/Library/JavascriptPromise.cpp
akroshg: try- catch was already present. Intentionally left as is. Same for EntryRace

lib/Runtime/Language/JavascriptOperators.inl
```diff
@@ -31,6 +31,32 @@ namespace Js
         }
     }
 
+    // A helper function which will do the IteratorStep and fetch value - however in the event of an exception it will perform the IteratorClose as well.
+    template <typename THandler>
+    void JavascriptOperators::DoIteratorStepAndValue(RecyclableObject* iterator, ScriptContext* scriptContext, THandler handler)
```
ianwjhalliday: I'm very happy to see this.  Thanks!

lib/Runtime/Language/JavascriptOperators.cpp
akroshg: ah, I think it may return `GetMissingPropertyResult`. Will fix this.

lib/Runtime/Library/JavascriptPromise.cpp
ianwjhalliday: I think it would be fragile to leave it as is.  I see the exception if there was one is used afterwards.  Either the while loop can be changed within the try/catch here and it should work the same, or `DoIteratorStepAndValue` can be modified to optionally return the exception instead of rethrowing.  Either way I think that would be cleaner and safer than leaving it as is.

---
In reply to: [66875446](https://github.com/Microsoft/ChakraCore/pull/1107#discussion_r66875446) [](ancestors = 66875446)

test/es6/iteratorclose.js
```diff
@@ -0,0 +1,950 @@
...
+                },
+                return: returnFnc
+              };
+            };
+            
+            var a;
+            [a] = iterable;
+            assert.areEqual(a, 20);
+            
+            returnFnc = null;
```
ianwjhalliday: This doesn't affect the object pointed to by `iterable` at all. You need to set `iterable.return` to null directly

test/es6/iteratorclose.js
```diff
@@ -0,0 +1,950 @@
...
+						return {
+							value : 10,
+							done : false
+						};
+					}
+				};
+			};
+
+			var [a] = iterable;
+			assert.areEqual(a, 10);
```
ianwjhalliday: All of these assertions are missing descriptions.
It is very slow to read tests and follow what they are doing without helpful descriptions.

For instance I don't clearly see the difference between the two assertions here.  What is the significance of repeating it with a different variable?

test/es6/iteratorclose.js
ianwjhalliday: Oh my bad, I misread the function above.  It does work, I forgot there are two layers of indirection with iterables.

---
In reply to: [66877366](https://github.com/Microsoft/ChakraCore/pull/1107#discussion_r66877366) [](ancestors = 66877366)

ianwjhalliday:
Implementation looks great.  I would like to see description messages on all the assertions in the tests.  For now I've skimmed lightly through the tests and like the comprehensiveness that appears to be there, but descriptions would help me read through with more depth and hopefully catch bugs or missing cases if there are any.

`yield*` and spread are missing from this implementation.  I'm not sure if spread is spec'ed to do iterator closing, I couldn't quickly find it in the spec.  `yield*` definitely has it though.

lib/Runtime/Library/JavascriptSet.cpp
akroshg: Yes it used in the object enumerator and typed array.

akroshg:
@ianwjhalliday  thanks for looking at it. I have updated the PR by addressing the comments.
I have added yield* test as well. I could not find spread related work on the spec. I'll find that out later and see if that is needed.

ianwjhalliday:
yield* itself needs to do iterator closing in its own loop, see runtime semantics for yield* here https://tc39.github.io/ecma262/#sec-generator-function-definitions-runtime-semantics-evaluation

test/es6/iteratorclose.js
```diff
@@ -0,0 +1,982 @@
...
+            };
+            assert.throws(function () { for (i of gf2()) { break; } }, Error, "return function throws", "Exception from return function");
+            returnCount = 0;
+            assert.throws(function () { for (obj.prop of gf2()) { } }, Error, "Exception at destructuring element wins", "From setter");
+            assert.areEqual(returnCount, 1, "Exception in loop should call the return function");
+            
+		}
+	},    
+    {
+		name : "Iterator close with yield *",
```
ianwjhalliday: This test is actually verifying that yield* calls the inner generator with the same api call (next, throw, or return) that was used to invoke itself.  This is behavior we already had prior to your change and I believe we have tests for in the generator test suite.  These tests here are redundant.

Instead you'll need to verify that a yield* expression itself calls return() upon a throwing inner iterable.  E.g. something like

```javascript
function* test() {
    try {
        yield* iterableWithReturnMethodThatThrowsInNextMethod;
    } catch (e) {
    }
    // verify here that the return method was called
}
var g = test();
g.next();
g.next();
g.next(); // however many times to make the throw occur
```

ianwjhalliday:
For spread I found the runtime semantics.

Calls: https://tc39.github.io/ecma262/#sec-argument-lists-runtime-semantics-argumentlistevaluation
Array initializers: https://tc39.github.io/ecma262/#sec-runtime-semantics-arrayaccumulation

They are not spec'ed to perform iterator closing which feels like an oversight. @bterlson do you know if this is intentional or a spec bug?

test/es6/iteratorclose.js
akroshg: Thanks I'll take a look at that. I'll probably work on that in my next PR.

akroshg:
I'll follow up things in my next pr. I am going to merge this in.

akroshg:
thanks @ianwjhalliday

PR #1108, opened 2016-06-09
xplat: fix empty GetResourceString simulation

jianchun:
On Windows the resource string loading code returns NULL if a resource
string content is empty (""). Simulate the same behavior on xplat.

es6/ES6PromiseAsync.js now passes.


digitalinfinity:
LGTM

PR #1113, opened 2016-06-09
xplat: implement wmemcmp

jianchun:
We were using `wcsncmp` in place of `wmemcmp` on xplat. They are not
inter-changable. `wmemcmp` does not stop on null-terminator. Implements
wmemcmp for xplat, modified from wcsncmp code.

es6/ES6StringAPIs.js now passes.


obastemur:
:shipit:

digitalinfinity:
LGTM

PR #1114, opened 2016-06-09
Remove bad assert.

LouisLaf:
Fix for 7832185. My rejit changes added and assert which I then made invalid.
The assert said the code path can only hit once which isn't true anymore.


LouisLaf:
@rajatd please review

rajatd:
LGTM

PR #1115, opened 2016-06-09
Fix some more unicode related unit tests on Linux

digitalinfinity:
- Removed es5\jx2.js since it was exactly the same as JSON\jx2.js
- Changed JSON\jx2.js to not print out Unicode characters
- Disabled es6\bug_OS_2553885.js since it needs Intl
- Fixed bugs in NormalizeString, IsIdStart and IsIdContinue
- Fixed an unrelated break in disable-jit builds


jianchun:
LGTM

lib/Runtime/PlatformAgnostic/Platform/Linux/UnicodeText.ICU.cpp
```diff
@@ -324,7 +324,9 @@ namespace PlatformAgnostic
             //  * C1_SPACE corresponds to the Unicode Zs category.
             //  * C1_BLANK corresponds to a hardcoded list thats ill-defined.
             // We'll skip that compatibility here and just check for Zs.
-            if ((charTypeMask & U_GC_ZS_MASK) != 0)
+            // We explicitly check for 0xFEFF to satisfy the unit test in es5/Lex_u3.js   
+            if ((charTypeMask & U_GC_ZS_MASK) != 0 ||
+                character == 0xFEFF)
```
obastemur: is it possible to have also `0xFFFE` ?

lib/Runtime/PlatformAgnostic/Platform/Linux/UnicodeText.ICU.cpp
```diff
@@ -42,6 +46,47 @@ namespace PlatformAgnostic
...
+            // converted to utf8. If it can't, see if the error is that
+            // the string is malformed.
+
+            Assert(invalidIndex != nullptr);
+            *invalidIndex = -1;
+            int32_t length = u_strlen(str);
+            int32_t i = 0;
+
+            for (;;)
+            {
```
obastemur: 1 - what happens here if str doesn't have 0 etc. and we end up with an undefined behavior?
2 - can we use size_t/unsigned instead of `int` - `int32_t` ? for `length`, and `i`, also could be nice for invalidIndex ( and reporting invalid index without -1, I don't think we need -1 there though )

digitalinfinity:
Thanks @obastemur @Jianchun. PR updated

PR #1116, opened 2016-06-10
Wasm Block yielding

### LINKED ISSUE #1110: Wasm block yielding
Capture what is the next tmp register for all types when entering the block (do not acquire)
At the end of the block, if we have to yield something, move it to the corresponding location, release everything then yield that location.

Cellule:
Closes #1110
Blocks now move the yielded value to a tmp register after releasing all the tmp register used in the block.
Exiting an eval stack now releases the location of tmp registers.
Added some debugging utils

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/1116)


Cellule:
ping

MikeHolman:
lgtm

PR #1117, opened 2016-06-10
Wasm single eval stack

### LINKED ISSUE #1109: Wasm single eval stack
Instead of having a stack of stack for the eval stack, change it to a single stack with delimiter between nested blocks

Cellule:
Closes #1109
Use a single stack for the eval stack instead of a stack of stacks

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/1117)

### DIFF
--- a/lib/WasmReader/WasmByteCodeGenerator.cpp
+++ b/lib/WasmReader/WasmByteCodeGenerator.cpp
@@ -23,7 +23,8 @@ WasmBytecodeGenerator::WasmBytecodeGenerator(Js::ScriptContext * scriptContext, 
     m_f32RegSlots(nullptr),
     m_f64RegSlots(nullptr),
     m_i32RegSlots(nullptr),
-    m_currentFunc(nullptr)
+    m_currentFunc(nullptr),
+    m_evalStack(&m_alloc)
 {
     m_writer.Create();
 
@@ -32,7 +33,6 @@ WasmBytecodeGenerator::WasmBytecodeGenerator(Js::ScriptContext * scriptContext, 
     m_writer.InitData(&m_alloc, astSize);
 
     m_labels = Anew(&m_alloc, SListCounted<Js::ByteCodeLabel>, &m_alloc);
-    m_evalStack = Anew(&m_alloc, SList<evalStackScope*>, &m_alloc);
 
     // Initialize maps needed by binary reader
     Binary::WasmBinaryReader::Init(scriptContext);
@@ -179,6 +179,10 @@ WasmBytecodeGenerator::GenerateFunction()
     m_currentFunc->wasmInfo = m_funcInfo;
     m_nestedIfLevel = 0;
     m_maxArgOutDepth = 0;
+    
+    Assert(m_evalStack.Empty());
+    // The stack should always be empty when starting a new function, make sure by clearing it in case we missed something
+    m_evalStack.Clear();
 
     // TODO: fix these bools
     m_writer.Begin(m_currentFunc->body, &m_alloc, true, true, false);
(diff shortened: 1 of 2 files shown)


lib/WasmReader/WasmByteCodeGenerator.cpp
```diff
@@ -1282,26 +1283,38 @@ WasmBytecodeGenerator::GetRegisterSpace(WasmTypes::WasmType type) const
...
-    m_evalStack->Top()->Push(info);
+    Assert(!m_evalStack.Empty());
+    m_evalStack.Push(info);
 }
 
 void
 WasmBytecodeGenerator::EnterEvalStackScope()
 {
-    m_evalStack->Push(Anew(&m_alloc, evalStackScope, &m_alloc));
+    m_evalStack.Push(EmitInfo(WasmTypes::Limit));
```
commonlisp: Ah, nice. I was considering introducing a separate scope marker in my first pass instead of a stack of stacks but ended up not wanting to introducing some variation of EmitInfo. EmitInfo(WasmTypes::Limit) is a nice choice.

lib/WasmReader/WasmByteCodeGenerator.cpp
```diff
@@ -173,6 +173,7 @@ WasmBytecodeGenerator::GenerateFunction()
     m_currentFunc->wasmInfo = m_funcInfo;
     m_nestedIfLevel = 0;
     m_maxArgOutDepth = 0;
+    m_evalStack.Clear();
```
commonlisp: Hmm, I'm thinking about how this is going to work out. If the scopes are properly formed and we ExitEvalStackScope all the way, m_evalStack should be empty anyway. If we didn't do that, would be we leaving the possibility of unreleased locations? But we really shouldn't be in the situation, right? Perhaps this is taken care of elsewhere.

Cellule: This can happen if another function throws an exception and we continue to compile other functions.
I could do the cleanup in the exception case and have an assert here. What do you think ?

MikeHolman: Maybe this should be under the WasmLazyTrapPhase flag?

---
In reply to: [66656352](https://github.com/Microsoft/ChakraCore/pull/1117#discussion_r66656352) [](ancestors = 66656352)

Cellule:
ping

lib/WasmReader/WasmByteCodeGenerator.cpp
```diff
@@ -323,6 +327,9 @@ WasmBytecodeGenerator::GenerateFunction()
         {
             throw WasmCompilationException(_u("%s\n  Function %s"), ex.GetErrorMessage(), functionName);
         }
+
+        // Since we will continue compilation for other functions, clear the eval stack
+        m_evalStack.Clear();
```
commonlisp: Will this trigger the right cleanup?

lib/WasmReader/WasmByteCodeGenerator.cpp
commonlisp: Yeah, cleanup in the exception would be good.

lib/WasmReader/WasmByteCodeGenerator.cpp
Cellule: We create new register space for every functions, therefore, it is not needed to release the location of a function that failed.
We only need to clear that stack to make sure we have it fresh for the next function
I would like to move this stack to a more appropriate structure that would be per Function instead of on the ByteCodeGenerator itself. (will be in an effort to test deferred parse)

MikeHolman:
:shipit:

PR #1118, opened 2016-06-10
apply -BaselineMode for -regexDebug dump

jianchun:
-regexDebug tests dump instruction offsets which are based on C++
`sizeof(struct ...)`. The results vary on different arch/compiler, thus
do not suit baseline based tests.

Add a common switch `-BaselineMode` to test harnesses. Change Regex code to
not dump real instruction offsets in this mode in order to do baseline
diff.


obastemur:
LGTM

PR #1119, opened 2016-06-10
Add override support to runtests.py

obastemur:
Takes care of typedarray/nan.js test


jianchun:
:shipit:

obastemur:

PR #1120, opened 2016-06-10
[wasm] copysign compatibility

commonlisp:
Use older CRT _copysign for compatibility

### DIFF
--- a/lib/Runtime/Math/WasmMath.inl
+++ b/lib/Runtime/Math/WasmMath.inl
@@ -27,9 +27,7 @@ WasmMath::Eqz(int value)
 template<typename T>
 inline T WasmMath::Copysign(T aLeft, T aRight)
 {
-    //return copysign(aLeft, aRight);
-    //Todo:: fix copysign for ChakraFull
-    return 0;
+    return (T)_copysign(aLeft, aRight);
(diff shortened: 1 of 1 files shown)


Cellule:
It's my fault the build is failing, I have a buddy build running with the fix, will checkin once it passes then you can rebase

lib/Runtime/Math/WasmMath.inl
```diff
@@ -27,9 +27,7 @@ WasmMath::Eqz(int value)
 template<typename T>
 inline T WasmMath::Copysign(T aLeft, T aRight)
 {
-    //return copysign(aLeft, aRight);
-    //Todo:: fix copysign for ChakraFull
-    return 0;
+    return (T)_copysign(aLeft, aRight);
```
MikeHolman: I think for float we probably want to do _copysignf, then you don't need to cast

commonlisp: Unfortunately, it appears the Full CRT doesn't have that either.

MikeHolman: ah ok

MikeHolman:
lgtm

PR #1121, opened 2016-06-10
Fix JavascriptMethod is not being called with the correct arguments

obastemur:


jianchun:
:shipit:

obastemur:

PR #1123, opened 2016-06-10
Return JsErrorScriptException from JsDiagEvaluate if script throws exception. Out parameter evalResult still contains the error object (ResolvedObject)

agarwal-sandeep:


lib/Jsrt/JsrtDebuggerObject.cpp
```diff
@@ -484,26 +487,36 @@ Js::DynamicObject* JsrtDebuggerStackFrame::Evaluate(Js::ScriptContext* scriptCon
...
+                // len can be big, if we failed just have empty string
+                resolvedObject.name = _u("");
             }
-            wcscpy_s((WCHAR*)resolvedObject.name, len + 1, source);
 
             resolvedObject.typeId = Js::JavascriptOperators::GetTypeId(resolvedObject.obj);
             JsrtDebuggerObjectBase::CreateDebuggerObject<JsrtDebuggerObjectProperty>(this->debuggerObjectsManager, resolvedObject, scriptContext, [&](Js::Var marshaledObj)
             {
-                evalResult = (Js::DynamicObject*)marshaledObj;
+                *evalResult = (Js::DynamicObject*)marshaledObj;
```
akroshg: it would be better to put success to true here, no?

lib/Jsrt/JsrtDiag.cpp
```diff
@@ -658,14 +658,15 @@ CHAKRA_API JsDiagEvaluate(
             return JsErrorInvalidArgument;
         }
 
-        Js::DynamicObject* result = debuggerStackFrame->Evaluate(scriptContext, expression, static_cast<int>(len), false);
+        Js::DynamicObject* result = nullptr;
+        bool success = debuggerStackFrame->Evaluate(scriptContext, expression, static_cast<int>(len), false, &result);
 
         if (result != nullptr)
```
akroshg: check for success here ?

akroshg:
looks good to me otherwise.

lib/Jsrt/JsrtDiag.cpp
```diff
@@ -658,14 +658,15 @@ CHAKRA_API JsDiagEvaluate(
             return JsErrorInvalidArgument;
         }
 
-        Js::DynamicObject* result = debuggerStackFrame->Evaluate(scriptContext, expression, static_cast<int>(len), false);
+        Js::DynamicObject* result = nullptr;
```
akroshg: nit: you don't need to use this extra variable *result*, as after the Evaluate api, if the result is null then *evalResult is already be null.

lib/Jsrt/JsrtDebuggerObject.cpp
agarwal-sandeep: No because here we can reach from both an script exception and successful execution.

lib/Jsrt/JsrtDiag.cpp
agarwal-sandeep: If success is false (script exception) we want to return the result object which will contain exception details so only checking for result. Below based on success we are returning no error or script exception error.

akroshg:
:shipit:

PR #1124, opened 2016-06-10
[wasm] Fix debugger test baseline

commonlisp:
Get the tests passing again after merge


MikeHolman:
lgtm

akroshg:
how it was broken before? was it due to bad merge job or some functionality go changed?

commonlisp:
I think it was a little of both. Parts of a Jsrt object was changed and there was way too much time since the last merge so some things did break.

commonlisp:
@Cellule So this is all covered with your commit, right? 
Closing

Cellule:
It was partly a bad merge.
There were stuff changed in WscriptJsrt and I had to rewrite some things added for wasm.
Which made the baseline slightly different.
It is indeed fixed in 95ff95ff3b336615ff7a8776f34771714180fb37

PR #1126, opened 2016-06-11
xplat: skip _fullPath conversion on dummy file name

jianchun:
In "WScript.LoadScript" we use dummy a file name "script.js". However we
later convert that to absolute path with `_fullpath`. PAL's implementation
on Linux has problem though. It calls `realpath` to get absolute path. This
fails on Linux.

Slightly change the code to avoid `_fullpath` conversion.


digitalinfinity:
LGTM

PR #1130, opened 2016-06-13
cross-platform: Improve O1+ stack probing for test builds

obastemur:
(Test/Debug Builds) keep Clang from:
- Inlining functions those are not assigned for..
- Optimizing sibling calls (tail call elimination)

Besides:
- Remove `__forceinline` from PAL / use it from CommonPal
- Replace `__declspec(noinline)` with new macro `_NOINLINE` and use attribute on xplat instead
- Make Error/stack.js pushing harder with stack limits

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -106,12 +106,16 @@ if(CLR_CMAKE_PLATFORM_UNIX)
         # -Wno-switch # switch values not handled
 
     # CXX COMPILER FLAGS
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} \
-        -fdelayed-template-parsing"
-    )
 
+    # xplat-todo for release build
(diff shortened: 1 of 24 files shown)


digitalinfinity:
Overall, change looks fine. I am concerned about the disabling of inlining in release builds though- seems like a high perf penalty no? It looks like it's a todo item so it's fine to tackle later but please open an issue to track in that case?

obastemur:
@digitalinfinity  Indeed that is a big perf penalty. However, they are not particularly disabled for any type of build. We don't have that separation set in terms of optimization switches but `O3` and O3 defined separately there.

Besides, we don't have a release build yet. Once we have a release build, we should indeed clear out that todo part and decide how to approach all these switches.

Saw @jianchun 's thumb there, So thank you both for the reviews

PR #1131, opened 2016-06-13
cross-platform: Make amd64 asm files compile on OSX

tadeuzagallo:
The assembly files fail to assemble due to unsupported `@plt` modifier and fail to link on OS X due to missing label prefix on function names.

### DIFF
--- a/lib/Runtime/Library/amd64/JavascriptFunctionA.S
+++ b/lib/Runtime/Library/amd64/JavascriptFunctionA.S
@@ -6,11 +6,11 @@
 .intel_syntax noprefix
 #include "unixasmmacros.inc"
 
-.extern _ZN2Js18JavascriptFunction13DeferredParseEPPNS_14ScriptFunctionE
-.extern _ZN2Js18JavascriptFunction19DeferredDeserializeEPNS_14ScriptFunctionE
+.extern C_FUNC(_ZN2Js18JavascriptFunction13DeferredParseEPPNS_14ScriptFunctionE)
+.extern C_FUNC(_ZN2Js18JavascriptFunction19DeferredDeserializeEPNS_14ScriptFunctionE)
 
-.global _ZN2Js18JavascriptFunction20DeferredParsingThunkEPNS_16RecyclableObjectENS_8CallInfoEz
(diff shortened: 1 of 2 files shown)


lib/Runtime/Library/amd64/JavascriptFunctionA.S
```diff
@@ -6,15 +6,15 @@
...
 
-.global _ZN2Js18JavascriptFunction20DeferredParsingThunkEPNS_16RecyclableObjectENS_8CallInfoEz
-.global _ZN2Js18JavascriptFunction24DeferredDeserializeThunkEPNS_16RecyclableObjectENS_8CallInfoEz
+.global C_FUNC(_ZN2Js18JavascriptFunction20DeferredParsingThunkEPNS_16RecyclableObjectENS_8CallInfoEz)
+.global C_FUNC(_ZN2Js18JavascriptFunction24DeferredDeserializeThunkEPNS_16RecyclableObjectENS_8CallInfoEz)
 
 #ifndef __APPLE__
-.type _ZN2Js18JavascriptFunction20DeferredParsingThunkEPNS_16RecyclableObjectENS_8CallInfoEz, @function
-.type _ZN2Js18JavascriptFunction24DeferredDeserializeThunkEPNS_16RecyclableObjectENS_8CallInfoEz, @function
+.type C_FUNC(_ZN2Js18JavascriptFunction20DeferredParsingThunkEPNS_16RecyclableObjectENS_8CallInfoEz), @function
```
obastemur: C_FUNC is not needed here

tadeuzagallo: I know it's not currently necessary, but it relies on an assumption of how the compiler mangles the symbols, so I usually prefer to keep it for consistency. Of course I can remove it though.

obastemur: @tadeuzagallo pseudo-op `.type` is not supported by OSX target anyways.

obastemur:
@tadeuzagallo IMO these files need a bit more change to be OSX compatible or to be able to pass the linker.

However,  I'm okay with the changes overall. Please update the title from `compatible with` to `compiles on`, and squash your changes.

Thanks for the PR!

obastemur:
LGTM. @tadeuzagallo Thanks for the PR again

PR #1134, opened 2016-06-14
cross-platform: ChakraCore on OSX

obastemur:
Disclaimer: Currently, ChakraCore is not officially supported on OSX.

This PR targets the individuals interested with experimenting ChakraCore on
OSX. After seeing #1131, spent couple of hours to fix the compile, and link 
time issues on OSX target.

Steps to compile :
```
brew install icu4c
./build.sh --debug -j:2 --icu=/usr/local/Cellar/icu4c/54.1/include/
```

Status:
CMAKE has an issue with shared bundles on OSX (a.k.a. so) and currently CH does not
play well with dylib.


tadeuzagallo:
Thanks for jumping into it so quickly. I tested it locally and I saw it already compiles, that's great. Let me know if there's anything I can help with. I'd also be more than happy to start looking into iOS after that.

bin/ChakraCore/CMakeLists.txt
```diff
@@ -3,10 +3,6 @@ add_library (ChakraCore SHARED
     ConfigParserExternals.cpp
     TestHooks.cpp
 )
-
-# By default, don't export any symbols from this library
-# We'll manually export the relevant individual functions
-set_target_properties(ChakraCore PROPERTIES CXX_VISIBILITY_PRESET hidden)
```
jianchun: Is this removed intentionally even on Linux, the comments not true?

build.sh
```diff
@@ -177,4 +177,3 @@ cmake $CMAKE_GEN $CC_PREFIX $ICU_PATH -DCMAKE_BUILD_TYPE=$BUILD_TYPE ../..
 
 $MAKE $MULTICORE_BUILD 2>&1 | tee build.log
 popd > /dev/null
-popd > /dev/null
```
jianchun: 2 pushd, 1 popd?

build.sh
```diff
@@ -165,7 +165,7 @@ fi
 
 pushd BuildLinux > /dev/null
 
-build_directory="${BUILD_TYPE,,}"
+build_directory="${BUILD_TYPE:0}"
```
jianchun: Actually what was this line doing... why not `build_directory=$BUILD_TYPE`?

lib/Runtime/Library/amd64/JavascriptFunctionA.S
```diff
@@ -71,9 +71,9 @@ NESTED_ENTRY amd64_CallFunction, _TEXT, NoHandler
 
 
         mov r10, rcx    // Store argc in r10
-        add r10, 3      // function, callInfo, ..., + 1 for alignment
+        add r10,  3      // function, callInfo, ..., + 1 for alignment
         and r10, -2     // Mask off the lower bit to 16 byte align the stack
-        shl r10, 3      // Calculate space for remaining args (* sizeof(Var*))
+        shl r10,  3      // Calculate space for remaining args (* sizeof(Var*))
```
jianchun: Why add spaces here? No longer aligned.

bin/ChakraCore/CMakeLists.txt
obastemur: This is a false setting..So, it doesn't affect statically inherited libraries, hence linker gives warnings.

pal/inc/pal_char16.h
```diff
@@ -28,26 +28,22 @@ This file is used to define the wchar_t type as a 16-bit type on Unix.
 // vc++, for whom wchar_t is already a typedef instead of a built-in.
 
 #if !defined(PAL_STDCPP_COMPAT) || defined(USING_PAL_STDLIB)
-#if defined (PLATFORM_UNIX) && defined(__GNUC__)
-#undef wchar_t
-#define wchar_t __wchar_16_cpp__
```
jianchun: What's the reason to make changes in this file? Original code was to redefine wchar_t. What doesn't work?

pal/inc/pal_char16.h
```diff
@@ -28,26 +28,22 @@ This file is used to define the wchar_t type as a 16-bit type on Unix.
...
 
 #if !defined(_WCHAR_T_DEFINED) || !defined(_MSC_VER)
 #if defined (PLATFORM_UNIX)
 #if defined(__cplusplus)
+#undef wchar_t
+#define wchar_t __wchar_16_cpp__
 typedef char16_t wchar_t;
 #else
-typedef unsigned short wchar_t;
+typedef unsigned short char16_t;
```
jianchun: Why the original code doesn't work for C? Why instead of defining wchar_t, you changed to define char16_t for C?

pal/inc/pal_mstypes.h
```diff
@@ -612,16 +620,9 @@ typedef unsigned long int uintptr_t;
...
-
-typedef char16_t WCHAR;
-
-#define WCHAR_IS_CHAR16_T 1
  
 #else // !PAL_STDCPP_COMPAT
- 
-#define WCHAR_IS_WCHAR_T 1
 
-typedef wchar_t WCHAR;
```
jianchun: Similarly don't understand the reason for these changes.

pal/src/include/pal/palinternal.h
```diff
@@ -311,7 +311,7 @@ function_name() to call the system's implementation
 #undef va_copy
 #endif
 
-#define wchar_t char16_t
+#define char16_t char16_t
```
jianchun: This looks weird.

build.sh
jianchun: Tested and original code converts to lower case... Your change is equivalent to `build_directory=$BUILD_TYPE`? Was your goal to change the dir name case?

bin/ChakraCore/CMakeLists.txt
digitalinfinity: Can you provide some more context here please? If we wanted a dynamic library on Linux, is that still not needed?

obastemur: This SO inherits other static libraries with visibility=default and expect to be a proxy. If we want a particular interface is hidden, this is not the place in our case.

build.sh
obastemur: `BuildLinux/debug` vs `BuildLinux/Debug` also runtests etc are updated

pal/inc/pal_char16.h
obastemur: @jianchun  I will try to answer all related questions here;

We did redefine `wchar_t` to `2 bytes` or however the char16_t is defined. Besides, ensured that for the rest of the library that we use `char16_t`

The problem is that unix has wchar is 32 bits, while char16_t is 2 bytes. As a result, we were redefining that 4bytes wchar back to char16_t. Somehow in the logic we were doing it before including `type_traits`. So, as a result, we were ending up redefining the same interface again and again.

IMO this PR fixes the inconsistencies around wchar_t. So we do expect sizeof(wchar_t) == sizeof(char16_t)

obastemur:
PR is updated. @tadeuzagallo @jianchun @digitalinfinity Thanks for the reviews so far.

bin/ChakraCore/CMakeLists.txt
digitalinfinity: Should this line then be moved to the root CMakeLists.txt if that's the behavior we want?

bin/ChakraCore/CMakeLists.txt
```diff
@@ -32,42 +28,82 @@ target_include_directories (
...
+      -Wl,--version-script=${CMAKE_CURRENT_SOURCE_DIR}/libChakraCoreLib.version
+    )
+    #
+    # Post build step to copy the built shared library
+    # to BuildLinux (or whatever the CMakeBuildDir is)
+    add_custom_command(TARGET ChakraCore POST_BUILD
+      COMMAND ${CMAKE_COMMAND} -E copy_if_different
+      "${CHAKRACORE_BINARY_DIR}/bin/ChakraCore/libChakraCore.so"
+       ${CHAKRACORE_BINARY_DIR}/)
+elseif(CMAKE_SYSTEM_NAME STREQUAL Darwin)
```
digitalinfinity: One thing is that its a little hard to read what the difference is between OSX and Linux- can we take this list, assign it to a CMakeList variable and then append just the diffs in the OSX/Linux cases (and then assign the customized list to target_link_libraries). If not, a comment on what the differences are in this particular case between OSX and Linux would make this easier to read in the future.

bin/ChakraCore/CMakeLists.txt
obastemur: See `-fvisibility=hidden` under root/CMakeLists.txt

obastemur:
Hmm. CI is down I guess?

Issue #1083, opened 2016-06-03
Checking arguments.length slow


iskiselev:
Checking arguments.length slow with Chakra. Here is sample page:
```
<html>
<body>
<script>
var start = Date.now();

function one(a){
}

function two(a,b){
}

function route(a,b){
   switch (arguments.length) {
   case 1:
     return one(a);
   case 2:
     return two(a,b);
   }
}

for (var i = 0; i<1000000; i++) {
  route(1,2);
}

document.write(Date.now() - start);
</script>
</body>
</html>
```
My results:
- Edge: ~380ms
- Chrome: ~20ms
- Firefox: ~3ms

kg:
Maybe do something to make sure the loop isn't being optimized out? Like return a result from ```one``` and ```two```, then sum the results up and check them at the end. That way the times are at least comparable.

iskiselev:
Updated test:
```
<html>
<body>
<script>
var start = Date.now();

function one(a){
  return a;
}

function two(a,b){
  return b;
}

function route(a,b){
   var argc = arguments.length | 0;
   switch (argc) {
   case 1:
     return one(a);
   case 2:
     return two(a,b);
   }
}

var sum = 0;
for (var i = 0; i<1000000; i++) {
  var r = Math.random();
  if (Math.random() > .5) {
    sum += route(r,-r);
  } else {
    sum += route(r);
  }
}

document.write("sum: " + sum + "<br>time: " + (Date.now() - start));
</script>
</body>
</html>
```
My results:

Edge: ~400ms
Chrome: ~80ms or ~270ms
Firefox: ~17ms

bterlson:
Curious how adding a "use strict" to the top of your script will impact this. Strict mode arguments object is much more lightweight.

iskiselev:
@bterlson, thank you. It was last piece of issue in sq/JSIL#978. In strict mode it is totally awful:
Edge: __~1380ms__
Chrome: ~65ms or ~250ms
Firefox: ~17ms

satheeshravi:
This issue will be fixed in https://github.com/Microsoft/ChakraCore/pull/1023. This is a WIP item.

ianwjhalliday:
Curious that strict mode is even worse

akroshg:
@ianwjhalliday it seems we create the actual arguments object in a normal path and later we unmap them. Look for `ConvertToUnmappedArguments`

pleath:
[...]

PR #1135, opened 2016-06-14
Stack Arguments Optimization with presence of Formals

satheeshravi:
**Premise:**
Whenever there is a use of "arguments" object in a function, we are supposed to create a Heap arguments object on the stack - so that both the formals and arguments element access can access a common memory.

The creation of this "heap arguments" object on the stack is currently optimized(not created) away in the jitted code, when there are no formals present in the function (but not in the presence of formals).

This change enables "heap arguments" object to be optimized away during the presence of formals. 

**This PR contains the following changes**
 - Dead Store of ArgIns (when the formals are not being used in the body of the function )
 - PropertyId Array is moved from the Bytecode's auxiliary data to the function body itself.
     ○ Layout change of all opcode  that currently explicitly uses Property Id Array as one of its source operand.
     ○ Byte code cache has been regenerated for Intl and Promise.
     ○ Refactoring of all code that reads property id array of formals from auxiliary data.
 
 - LdPropId is no more accompanied with LdHeapArguments instruction.
 - Removal of Heap arguments creation in the jitted code, when the function has formals.
 - Removal of Scope object creation upon heap arguments optimization.
 - During javascript stack walking, we no more use the existing heap arguments object on the frame. Instead we create a copy of the heap arguments object and give it to the .caller. This makes us not to worry about other functions changing the formals value, interrupting the stack arguments optimization.
       -TEST instr at the beginning of the LdElem Fast path is also removed.
- Field copy prop is disabled for arguments property (to avoid foo.arguments being field copy-proped)

**Functioning of Heap Arguments Optimization with Formals.**
 - We track several Instructions to facilitate this optimization in the forward and backward pass, along with the help from the front end.
Parse:
 - We disable this optimization, when there are any write to formals.
 - We also disable this optimization, when we have any non-local references inside nested functions, deferred nested functions, and presence of any nested functions.

_ForwardPass:_
 - Track Scope object in LdHeapArgs (all versions of this opcode)
 - Track Formals array in LdSlotArr
 - Track Formals in LdSlot.
 - Trace Stack Sym for formals in inlinee.
 - Insert BailOnStackArgsOutOfActualsRange on LdElemI_A, when optimization is turned on.

_DeadStorePass:_
 - We are sure at this point, whether we will be doing the Stack arguments optimization for formals or not.
 - Insert ArgIns for formals at the beginning of the function.
 - Replace LdSlot with Ld_A with the tracked stack syms  for formals.
 - Dead store all scope object related instrs  (we are sure that only formals use scope object).
 - Remove BailOnStackArgsOutOfActualsRange, depending on whether the optimization is still turned on or not.

**Perf Impact:**

- This change improves node.js - Specifically _acme-benchmark(~16.3%)_ and _tcpSlowPerf (~18.36%)_.
- Speedometer benchmark - gets an _overall 12 % improvement_ (_Specifically 30% in react_ test)
- This also has impact in real-world websites.
- IE Perf Lab: http://ieperf-web-01/perfresults/Comparison?testRunId=419859&baseRunId=419858&lab=IEPERF&getRelatedRunData=false&priorities=0,1&categories=failures,warnings,improvements,no-change&search=
- _More than 95% websites_ show that they get benefited from this optimization [Trace was collected using web crawler].
- Other benchmarks look flat.


satheeshravi:
Fixes https://github.com/Microsoft/ChakraCore/issues/313 and https://github.com/Microsoft/ChakraCore/issues/1083

akroshg:
would you squash all commits to one? It is easier to review in that case.

satheeshravi:
@akroshg : If you click on the latest commit tab in CR, it should show the entire change in all the commits.

lib/Backend/GlobOpt.h
```diff
@@ -1186,7 +1186,7 @@ class GlobOpt
     SparseArray<Value>       *  byteCodeConstantValueArray;
     // Global bitvectors
     BVSparse<JitArenaAllocator> * byteCodeConstantValueNumbersBv;
-
+
```
rajatd: nit: extra line

lib/Backend/GlobOpt.cpp
```diff
@@ -3825,32 +3826,157 @@ GlobOpt::TestAnyArgumentsSym()
...
+                //Skipping extra actuals passed
+                for (Js::ArgSlot i = 0; i < extraActuals; i++)
+                {
+                    argOutInstr = argOutInstr->GetSrc2()->GetStackSym()->GetInstrDef();
+                }
+            }
+
+            StackSym * undefinedSym = nullptr;
+
+            for (Js::ArgSlot param = formalsCount; param > 0; param--)
```
rajatd: Don't you want to iterate till param == 0 since you already subtracted 1 from functionBody->m_inParamsCount to get the 'formalsCount'?

lib/Backend/GlobOpt.cpp
```diff
@@ -3825,32 +3826,157 @@ GlobOpt::TestAnyArgumentsSym()
...
+                {
+                    if (argOutInstr->GetSrc1()->IsRegOpnd())
+                    {
+                        argOutSym = argOutInstr->GetSrc1()->GetStackSym();
+                    }
+                    else
+                    {
+                        //TODO: saravind: Can I just use the same source operand like this?
+                        // We will always have ArgOut instr - so the source operand will not be removed.
+                        argOutSym = StackSym::New(func);
```
rajatd: you should use the inliner's func here and in the creation of dstOpnd and assignInstr as you are inserting the Ld_A before InlineeStart

lib/Backend/GlobOpt.cpp
```diff
@@ -3825,32 +3826,157 @@ GlobOpt::TestAnyArgumentsSym()
...
+                    argOutInstr = argOutInstr->GetSrc2()->GetStackSym()->GetInstrDef();
+                }
+                else
+                {
+                    /*When param is out of range of actuals count, load undefined*/
+                    // TODO: saravind: This will insert undefined for each of the param not having an actual. - Clean up this by having a sym for undefined on func ?
+                    Assert(formalsCount > actualsCount);
+                    if (undefinedSym == nullptr)
+                    {
+                        undefinedSym = StackSym::New(func);
```
rajatd: Inliner's func

lib/Backend/GlobOpt.cpp
```diff
@@ -3825,32 +3826,157 @@ GlobOpt::TestAnyArgumentsSym()
...
+                if (param <= actualsCount)
+                {
+                    Assert(argOutSym);
+                    func->TrackStackSymForFormalIndex(param - 1, argOutSym);
+                    argOutInstr = argOutInstr->GetSrc2()->GetStackSym()->GetInstrDef();
+                }
+                else
+                {
+                    /*When param is out of range of actuals count, load undefined*/
+                    // TODO: saravind: This will insert undefined for each of the param not having an actual. - Clean up this by having a sym for undefined on func ?
```
rajatd: Sounds reasonable.

rajatd: nit: rename to assignUndefined?

lib/Backend/GlobOpt.cpp
```diff
@@ -4663,6 +4789,13 @@ GlobOpt::OptInstr(IR::Instr *&instr, bool* isInstrRemoved)
 
     this->OptArguments(instr);
 
+    //StackArguments Optimization - We bail out if the index is out of range of actuals.
+    if (instr->m_opcode == Js::OpCode::LdElemI_A && instr->DoStackArgsOpt(this->func) && !this->IsLoopPrePass())
+    {
+        AssertMsg(!instr->HasBailOutInfo(), "There should not be any existing bail out on LdElem instr, at this time.");
```
rajatd: Is there anything functionally wrong if it does have a bailout? In other words, what does this assert protect against?

lib/Backend/BackwardPass.cpp
```diff
@@ -1818,6 +1908,15 @@ BackwardPass::ProcessBailOutInfo(IR::Instr * instr)
             BVSparse<JitArenaAllocator> * byteCodeUpwardExposedUsed = byteCodeUsesInstr->byteCodeUpwardExposedUsed;
             if (byteCodeUpwardExposedUsed != nullptr)
             {
+                //Stack Args for Formals optimization.
```
rajatd: Might want to mention here that if stack args opt is enabled, then the only thing that may require scope object downstream is formals access (which is why it is ok to clear the bit in byteCodeUpwardExposedUsed)

lib/Backend/BackwardPass.cpp
```diff
@@ -2735,6 +2846,168 @@ BackwardPass::ProcessBlock(BasicBlock * block)
...
+    IR::Instr * instr = this->currentInstr;
+    Func * currFunc = instr->m_func;
+
+    if (this->tag == Js::DeadStorePhase && instr->IsStackArgsEnabled() && !IsPrePass())
+    {
+        switch (instr->m_opcode)
+        {
+            /*
+            *   This LdSlot loads the formal from the formals array. We replace this a Ld_A <ArgInSym>.
+            *   ArgInSym is inserted at the beginning of the function during the start of the deadstore pass- for the inliner.
```
rajatd: change to 'top func'

lib/Backend/Lower.cpp
```diff
@@ -265,7 +253,10 @@ Lowerer::LowerRange(IR::Instr *instrStart, IR::Instr *instrEnd, bool defaultDoFa
         case Js::OpCode::InvalCachedScope:
             this->LowerBinaryHelper(instr, IR::HelperOP_InvalidateCachedScope);
             break;
-
+
```
rajatd: nit: extra line

lib/Backend/Lower.cpp
```diff
@@ -7692,6 +7684,17 @@ Lowerer::LowerUnaryHelperMemWithBoolReference(IR::Instr *instr, IR::JnHelperMeth
     return this->LowerUnaryHelperMem(instr, helperMethod, useBoolForBailout ? boolOpnd : nullptr);
 }
 
+IR::Instr *
+Lowerer::LowerInitCachedScope(IR::Instr* instr)
+{
+    instr->m_opcode = Js::OpCode::CallHelper;
```
rajatd: This needs to be lowered.

lib/Backend/Lower.cpp
```diff
@@ -14977,7 +15012,7 @@ Lowerer::GenerateFastStringLdElem(IR::Instr * ldElem, IR::LabelInstr * labelHelp
 }
 
 bool
-Lowerer::GenerateFastLdElemI(IR::Instr *& ldElem, bool *instrIsInHelperBlockRef)
+Lowerer::GenerateFastLdElemI(IR::Instr *& ldElem, bool *instrIsInHelperBlockRef, IR::LabelInstr ** bailOutLabelInstr)
```
rajatd: Unused?

test/Function/StackArgsWithFormals.js
```diff
@@ -0,0 +1,141 @@
+//Test1 : Tests Optimization with bailout.
```
akroshg: copyright header.

test/Function/StackArgsWithFormals.js
```diff
@@ -0,0 +1,141 @@
+//Test1 : Tests Optimization with bailout.
+var shouldBailout = false;
```
akroshg: convert this test to use the unittestframework. You can see other test in the es6 folder to get the idea.

test/Function/StackArgsWithFormals.js
```diff
@@ -0,0 +1,141 @@
...
+print("TEST 7 : ");
+shouldBailout = false;
+test7();
+shouldBailout = true;
+test7();
+
+//Test 8 : for-in statement
+function test8(a,b)
+{
+	for(a in [10,20,30])
```
akroshg: add test case : for ([a] in [[10]]) ...

test/Function/StackArgsWithFormals.js
```diff
@@ -0,0 +1,141 @@
...
+	}
+}
+print("TEST 8 : ");
+test8(1,2);
+test8('x','y');
+
+//Test 9 : Destructuring.
+function test9(a,b)
+{
+	[a,b] = [10,20];
```
akroshg: note : you should object destructuring test as well.

lib/Backend/Lower.cpp
```diff
@@ -8181,7 +8184,7 @@ Lowerer::LowerStElemI(IR::Instr * instr, Js::PropertyOperationFlags flags, bool
 ///
 ///----------------------------------------------------------------------------
 IR::Instr *
-Lowerer::LowerLdElemI(IR::Instr * instr, IR::JnHelperMethod helperMethod, bool isHelper)
+Lowerer::LowerLdElemI(IR::Instr * instr, IR::JnHelperMethod helperMethod, bool isHelper, IR::LabelInstr * labelBailOut)
```
rajatd: Not required, I think.

lib/Backend/Lower.cpp
```diff
@@ -8313,7 +8320,7 @@ Lowerer::LowerLdElemI(IR::Instr * instr, IR::JnHelperMethod helperMethod, bool i
 
     src1->Free(this->m_func);
 
-    m_lowererMD.ChangeToHelperCall(instr, helperMethod, nullptr, nullptr, nullptr, isHelper);
+    m_lowererMD.ChangeToHelperCall(instr, helperMethod, labelBailOut, nullptr, nullptr, isHelper);
```
rajatd: same here

lib/Runtime/Library/ArgumentsObject.cpp
```diff
@@ -85,20 +85,17 @@ namespace Js
...
-            args = JavascriptOperators::LoadHeapArguments(
-                funcCaller,
-                paramCount - 1,
-                walker->GetJavascriptArgs(),
-                scriptContext->GetLibrary()->GetNull(),
-                scriptContext->GetLibrary()->GetNull(),
-                scriptContext,
-                /* formalsAreLetDecls */ false);
-
-            walker->SetPermanentArguments(args);
```
akroshg: Are these functions (Set/GetParmanentArgumetns) used anywhere else? You could remove them from the stack-walker if they are not.

lib/Backend/Func.h
```diff
@@ -52,6 +52,36 @@ struct Cloner
...
 };
 
+/*
+* This class keeps track of various information required for Stack Arguments optimization with formals.
+*/
+class StackArgWithFormalsTracker
+{
+private:
+    BVSparse<JitArenaAllocator> *  formalsArraySyms;    //Tracks Formal parameter Array - Is this Bv required explicitly?
+    StackSym**                     formalsIndexToSymIdMap; //Tracks the stack sym for each formal
```
rajatd: This is more of a formals index to sym (instead of sym id) map, right? Consider renaming it as such.

lib/Runtime/Library/ArgumentsObject.cpp
rajatd: @akroshg , also used in JavascriptFunction::GetArgumentsProperty

---
In reply to: [67767669](https://github.com/Microsoft/ChakraCore/pull/1135#discussion_r67767669) [](ancestors = 67767669)

rajatd:
@satheeshravi, reviewed backend changes.

lib/Runtime/Library/ArgumentsObject.cpp
akroshg: It is removed from there is well, no?

---
In reply to: [67768600](https://github.com/Microsoft/ChakraCore/pull/1135#discussion_r67768600) [](ancestors = 67768600,67767669)

rajatd: Oh, I might have an out of date local copy

---
In reply to: [67770818](https://github.com/Microsoft/ChakraCore/pull/1135#discussion_r67770818) [](ancestors = 67770818,67768600,67767669)

lib/Runtime/Library/ArgumentsObject.h
```diff
@@ -122,7 +127,9 @@ namespace Js
         BOOL IsFormalArgument(PropertyId propertyId);
         BOOL IsFormalArgument(PropertyId propertyId, uint32* pIndex);    // Checks whether property is numeric, and on success sets that index.
         BOOL IsArgumentDeleted(uint32 index) const;
+    public:
         const ActivationObject* const GetFrameObject() { return frameObject; }
+        void SetFrameObject(ActivationObject * value) { frameObject = value; }
```
akroshg: is it good to assert that frameObject was null before?

lib/Backend/Lower.cpp
```diff
@@ -8197,8 +8200,12 @@ Lowerer::LowerLdElemI(IR::Instr * instr, IR::JnHelperMethod helperMethod, bool i
     {
```
rajatd: You should probably add an assert that the only bailout kind on the instruction is BailOnStackArgsOutOfActualsRange

lib/Runtime/Language/JavascriptOperators.cpp
```diff
@@ -6838,8 +6779,14 @@ namespace Js
         AssertMsg(actualsCount != (uint32)-1 && formalsCount != (uint32)-1,
                   "Loading the arguments object in the global function?");
 
-        // Create and initialize the Arguments object.
         HeapArgumentsObject *argsObj = JavascriptOperators::CreateHeapArguments(funcCallee, actualsCount, formalsCount, frameObj, scriptContext);
+        
+        return FillScopeObject(funcCallee, actualsCount, formalsCount, frameObj, paramAddr, nullptr, argsObj, scriptContext, nonSimpleParamList, true);
+    }
+
+    Var JavascriptOperators::FillScopeObject(JavascriptFunction *funcCallee, uint32 actualsCount, uint32 formalsCount, Var frameObj, Var * paramAddr, Js::PropertyIdArray *propIds, HeapArgumentsObject * argsObj, ScriptContext * scriptContext, bool nonSimpleParamList, bool useCachedScope)
```
akroshg: nit: break parameters in multiple lines - it is hard to read.

akroshg: is this private function?

akroshg: never mind

---
In reply to: [67771852](https://github.com/Microsoft/ChakraCore/pull/1135#discussion_r67771852) [](ancestors = 67771852)

lib/Runtime/ByteCode/ByteCodeSerializer.cpp
```diff
@@ -3298,6 +3327,63 @@ class ByteCodeBufferReader
         return current;
     }
 
+    const byte * ReadPropertyIdOfFormals(const byte * current, FunctionBody * function)
+    {
+#ifdef BYTE_CODE_MAGIC_CONSTANTS
+        int constant;
```
akroshg: initialize this variable. Same for variable count below.

lib/Parser/Parse.cpp
```diff
@@ -11763,6 +11818,10 @@ ParseNodePtr Parser::ConvertArrayToArrayPattern(ParseNodePtr pnode)
         {
             *itemRef = ConvertObjectToObjectPattern(item);
         }
+        else if (item->nop == knopName)
+        {
+            TrackAssignment<true>(item, nullptr, 0, 0);
```
akroshg: why not use the item's ichmin and ichLim instead of passing 0.

lib/Runtime/Base/FunctionBody.h
```diff
@@ -1862,6 +1862,7 @@ namespace Js
         WriteBarrierPtr<void*> inlineCaches;
         InlineCachePointerArray<PolymorphicInlineCache> polymorphicInlineCaches; // Contains the latest polymorphic inline caches
         WriteBarrierPtr<PropertyId> cacheIdToPropertyIdMap;
+        WriteBarrierPtr<PropertyIdArray> formalsPropIdArray;
```
akroshg: This will increase the size of the function body, even though the function will not have eval or arguments usage. Could you check with @leirocks for alternate way to fix this?

lib/Runtime/ByteCode/ByteCodeGenerator.cpp
```diff
@@ -2333,9 +2333,31 @@ FuncInfo* PreVisitFunction(ParseNode* pnode, ByteCodeGenerator* byteCodeGenerato
...
+#ifdef PERF_HINT
+                if (PHASE_TRACE1(Js::PerfHintPhase) && !doStackArgsOpt)
+                {
+                    WritePerfHint(PerfHints::HeapArgumentsDueToWriteToFormals, funcInfo->GetParsedFunctionBody(), 0);
+                }
+#endif
+                //Go conservative if it has any nested functions, or any non-local references.
+                //With statements - need scope object to be present.
+                //Nested funtions - need scope object to be present - LdEnv/LdFrameDisplay needs it.
+                if ((pnode->sxFnc.funcInfo->GetParamScope()->Count() > 1) && (pnode->sxFnc.funcInfo->hasLocalInClosure || pnode->sxFnc.funcInfo->HasDeferredChild() || pnode->sxFnc.nestedCount > 0 ||
```
akroshg: hasLocalInClosure be ever be ever true if the nestedCount is 0 ?

akroshg: should you check first if the doStackArgsOpt is true, before checking all this conditions?

test/Function/rlexe.xml
```diff
@@ -361,6 +361,13 @@
   </test>
   <test>
     <default>
+      <files>StackArgsWithFormals.js</files>
+      <baseline>StackArgsWithFormals.baseline</baseline>
```
akroshg: could you add the test case with forceserialized as well?

lib/Runtime/Language/InterpreterStackFrame.cpp
```diff
@@ -8330,6 +8331,98 @@ const byte * InterpreterStackFrame::OP_ProfiledLoopBodyStart(const byte * ip)
...
+    HeapArgumentsObject * InterpreterStackFrame::CreateEmptyHeapArgumentsObject(ScriptContext* scriptContext)
+    {
+        HeapArgumentsObject * args = JavascriptOperators::CreateHeapArguments(this->function->GetRealFunctionObject(), this->m_inSlotsCount - 1, 0, nullptr, scriptContext);
+        this->m_arguments = args;
+        return args;
+    }
+
+    void InterpreterStackFrame::TrySetFrameObjectInHeapArgObj(ScriptContext * scriptContext, bool hasNonSimpleParams)
+    {
+        if (PHASE_OFF1(Js::StackArgFormalsOptPhase))
```
akroshg: As I understand this function is called when the stackargopt is enabled, so can this phase be off?

lib/Parser/Parse.cpp
```diff
@@ -8437,6 +8450,46 @@ ParseNodePtr Parser::ParseExpr(int oplMin,
...
+            ref->TrackAssignment(ichMin, ichLim);
+        }
+    }
+}
+
+void PidRefStack::TrackAssignment(charcount_t ichMin, charcount_t ichLim)
+{
+    if (this->isAsg)
+    {
+        if (this->GetIchMin() <= ichMin)
```
akroshg: what is the use of this code under the `if (this->isAsg)`. I understand this was old code reverted. But if you don;t need this code could you just set `this->isAsg = true` in this function.

lib/Runtime/Base/FunctionBody.h
```diff
@@ -2690,6 +2693,23 @@ namespace Js
         bool GetHasRestParameter() const { return (flags & Flags_HasRestParameter) != 0; }
         void SetHasRestParameter() { SetFlags(true, Flags_HasRestParameter); }
 
+        bool NeedScopeObjectForArguments(bool hasNonSimpleParams)
+        {
+            Assert(HasReferenceableBuiltInArguments());
+
```
aneeshdk: nit

lib/Runtime/ByteCode/ByteCodeGenerator.cpp
```diff
@@ -2333,9 +2333,31 @@ FuncInfo* PreVisitFunction(ParseNode* pnode, ByteCodeGenerator* byteCodeGenerato
...
+                    pnode->sxFnc.HasWithStmt() || byteCodeGenerator->IsInDebugMode() || PHASE_OFF1(Js::StackArgFormalsOptPhase) || PHASE_OFF1(Js::StackArgOptPhase)))
+                {
+                    doStackArgsOpt = false;
+#ifdef PERF_HINT
+                    if (PHASE_TRACE1(Js::PerfHintPhase) && !doStackArgsOpt)
```
aneeshdk: This is not needed.

lib/Runtime/Language/DynamicProfileInfo.h
```diff
@@ -500,6 +500,7 @@ namespace Js
             bool disableObjTypeSpec_jitLoopBody : 1;
             bool disablePowIntIntTypeSpec : 1;
             bool disableLoopImplicitCallInfo : 1;
+			bool disableStackArgOpt : 1;
```
aneeshdk: nit

lib/Runtime/Language/InterpreterStackFrame.cpp
```diff
@@ -8330,6 +8331,98 @@ const byte * InterpreterStackFrame::OP_ProfiledLoopBodyStart(const byte * ip)
...
+                heapArgObj = (Js::HeapArgumentsObject*)m_arguments;
+            }
+            else
+            {
+                Assert(false);
+                return;
+            }
+        }
+
+        bool isCachedScope = false;
```
aneeshdk: Move it inside the if condition.

lib/Runtime/Language/InterpreterStackFrame.cpp
```diff
@@ -8330,6 +8331,98 @@ const byte * InterpreterStackFrame::OP_ProfiledLoopBodyStart(const byte * ip)
...
+
+        //For Non-simple params, we don't have a scope object created.
+        if (this->m_functionBody->NeedScopeObjectForArguments(hasNonSimpleParams))
+        {
+                isCachedScope = m_functionBody->HasCachedScopePropIds();
+                if (isCachedScope)
+                {
+                    Js::DynamicType *literalType = nullptr;
+
+                    propIds = this->m_functionBody->GetFormalsPropIdArray();
```
aneeshdk: This can be moved outside

lib/Runtime/Language/ReadOnlyDynamicProfileInfo.h
```diff
@@ -50,6 +50,7 @@ namespace Js
             isObjTypeSpecDisabled_jitLoopBody(false),
             isPowIntIntTypeSpecDisabled(false),
             isLoopImplicitCallInfoDisabled(false),
+			isStackArgOptDisabled(false),
```
aneeshdk: nit

lib/Backend/GlobOpt.cpp
satheeshravi: done.

---
In reply to: [67736799](https://github.com/Microsoft/ChakraCore/pull/1135#discussion_r67736799) [](ancestors = 67736799)

lib/Backend/GlobOpt.cpp
satheeshravi: Good catch. Thanks

---
In reply to: [67736248](https://github.com/Microsoft/ChakraCore/pull/1135#discussion_r67736248) [](ancestors = 67736248)

lib/Backend/GlobOpt.cpp
satheeshravi: Done.

---
In reply to: [67736377](https://github.com/Microsoft/ChakraCore/pull/1135#discussion_r67736377) [](ancestors = 67736377)

lib/Backend/GlobOpt.cpp
satheeshravi: I think I will remove this. But this Assert is anyways in GenerateBailAtOperation API.

---
In reply to: [67737983](https://github.com/Microsoft/ChakraCore/pull/1135#discussion_r67737983) [](ancestors = 67737983)

lib/Backend/BackwardPass.cpp
satheeshravi: Done.

---
In reply to: [67740116](https://github.com/Microsoft/ChakraCore/pull/1135#discussion_r67740116) [](ancestors = 67740116)

lib/Backend/BackwardPass.cpp
satheeshravi: ArgIns insertion is already using top Func. And the funcs used here should be the func from the operating instr. Please clarify.

---
In reply to: [67741254](https://github.com/Microsoft/ChakraCore/pull/1135#discussion_r67741254) [](ancestors = 67741254)

lib/Backend/Lower.cpp
satheeshravi: This will be lowerer during the subsequent iteration of Instructions.

---
In reply to: [67748513](https://github.com/Microsoft/ChakraCore/pull/1135#discussion_r67748513) [](ancestors = 67748513)

test/Function/StackArgsWithFormals.js
satheeshravi: Done.

---
In reply to: [67765651](https://github.com/Microsoft/ChakraCore/pull/1135#discussion_r67765651) [](ancestors = 67765651)

test/Function/StackArgsWithFormals.js
satheeshravi: Done.

---
In reply to: [67766996](https://github.com/Microsoft/ChakraCore/pull/1135#discussion_r67766996) [](ancestors = 67766996)

lib/Backend/Lower.cpp
satheeshravi: Done.

---
In reply to: [67767463](https://github.com/Microsoft/ChakraCore/pull/1135#discussion_r67767463) [](ancestors = 67767463)

lib/Backend/Lower.cpp
satheeshravi: Done.

---
In reply to: [67767447](https://github.com/Microsoft/ChakraCore/pull/1135#discussion_r67767447) [](ancestors = 67767447)

lib/Runtime/Library/ArgumentsObject.cpp
satheeshravi: Removed the Setter API. Getter is used in HeapArgumentsObject::AdvanceWalkerToArgsFrame(JavascriptStackWalker *walker)

---
In reply to: [67770975](https://github.com/Microsoft/ChakraCore/pull/1135#discussion_r67770975) [](ancestors = 67770975,67770818,67768600,67767669)

lib/Backend/Func.h
satheeshravi: Done.

---
In reply to: [67768147](https://github.com/Microsoft/ChakraCore/pull/1135#discussion_r67768147) [](ancestors = 67768147)

lib/Runtime/Library/ArgumentsObject.h
satheeshravi: Sure. will try that out.

---
In reply to: [67771245](https://github.com/Microsoft/ChakraCore/pull/1135#discussion_r67771245) [](ancestors = 67771245)

lib/Runtime/Language/JavascriptOperators.cpp
satheeshravi: Done.

---
In reply to: [67771761](https://github.com/Microsoft/ChakraCore/pull/1135#discussion_r67771761) [](ancestors = 67771761)

lib/Runtime/ByteCode/ByteCodeSerializer.cpp
satheeshravi: Done.

---
In reply to: [67773061](https://github.com/Microsoft/ChakraCore/pull/1135#discussion_r67773061) [](ancestors = 67773061)

lib/Parser/Parse.cpp
satheeshravi: They are not used in buildAst code path. So just pass 0 ?

---
In reply to: [67775587](https://github.com/Microsoft/ChakraCore/pull/1135#discussion_r67775587) [](ancestors = 67775587)

lib/Runtime/Language/ReadOnlyDynamicProfileInfo.h
satheeshravi: Done.

---
In reply to: [67796330](https://github.com/Microsoft/ChakraCore/pull/1135#discussion_r67796330) [](ancestors = 67796330)

lib/Runtime/Language/InterpreterStackFrame.cpp
satheeshravi: this is used in the last API call in this function.

---
In reply to: [67794725](https://github.com/Microsoft/ChakraCore/pull/1135#discussion_r67794725) [](ancestors = 67794725)

lib/Runtime/Language/InterpreterStackFrame.cpp
satheeshravi: Done.

---
In reply to: [67794959](https://github.com/Microsoft/ChakraCore/pull/1135#discussion_r67794959) [](ancestors = 67794959)

lib/Runtime/Language/DynamicProfileInfo.h
satheeshravi: Done.

---
In reply to: [67793943](https://github.com/Microsoft/ChakraCore/pull/1135#discussion_r67793943) [](ancestors = 67793943)

lib/Runtime/ByteCode/ByteCodeGenerator.cpp
satheeshravi: Done.

---
In reply to: [67793884](https://github.com/Microsoft/ChakraCore/pull/1135#discussion_r67793884) [](ancestors = 67793884)

lib/Runtime/ByteCode/ByteCodeGenerator.cpp
satheeshravi: Done.

---
In reply to: [67784322](https://github.com/Microsoft/ChakraCore/pull/1135#discussion_r67784322) [](ancestors = 67784322)

lib/Runtime/Base/FunctionBody.h
satheeshravi: Done.

---
In reply to: [67793826](https://github.com/Microsoft/ChakraCore/pull/1135#discussion_r67793826) [](ancestors = 67793826)

lib/Runtime/Language/InterpreterStackFrame.cpp
satheeshravi: Removed. Left overs.

---
In reply to: [67786183](https://github.com/Microsoft/ChakraCore/pull/1135#discussion_r67786183) [](ancestors = 67786183)

test/Function/rlexe.xml
satheeshravi: Done.

---
In reply to: [67785847](https://github.com/Microsoft/ChakraCore/pull/1135#discussion_r67785847) [](ancestors = 67785847)

lib/Runtime/ByteCode/ByteCodeGenerator.cpp
satheeshravi: Done.

---
In reply to: [67783889](https://github.com/Microsoft/ChakraCore/pull/1135#discussion_r67783889) [](ancestors = 67783889)

lib/Backend/Lower.cpp
satheeshravi: Done.

---
In reply to: [67771530](https://github.com/Microsoft/ChakraCore/pull/1135#discussion_r67771530) [](ancestors = 67771530)

test/Function/StackArgsWithFormals.js
satheeshravi: Done.

---
In reply to: [67766232](https://github.com/Microsoft/ChakraCore/pull/1135#discussion_r67766232) [](ancestors = 67766232)

test/Function/StackArgsWithFormals.js
satheeshravi: At first look, it doesn't seem to me that i can run each function twice, for jit to kick in ?

---
In reply to: [67765923](https://github.com/Microsoft/ChakraCore/pull/1135#discussion_r67765923) [](ancestors = 67765923)

lib/Backend/Lower.cpp
satheeshravi: Removed. Done

---
In reply to: [67765060](https://github.com/Microsoft/ChakraCore/pull/1135#discussion_r67765060) [](ancestors = 67765060)

lib/Runtime/Base/FunctionBody.h
satheeshravi: Done.

---
In reply to: [67777247](https://github.com/Microsoft/ChakraCore/pull/1135#discussion_r67777247) [](ancestors = 67777247)

lib/Parser/Parse.cpp
satheeshravi: Sure. WIll have a word with Paul and make necessary changes.

---
In reply to: [67786544](https://github.com/Microsoft/ChakraCore/pull/1135#discussion_r67786544) [](ancestors = 67786544)

obastemur:
nit: Once this PR is done. It could be so nice to merge smaller commits into bigger ones. See also https://github.com/Microsoft/ChakraCore/pull/1160#issuecomment-227320749

lib/Backend/BackwardPass.cpp
rajatd: I meant in the comment, change 'inliner' to 'top func' since an inliner may not be top func

---
In reply to: [67800979](https://github.com/Microsoft/ChakraCore/pull/1135#discussion_r67800979) [](ancestors = 67800979,67741254)

lib/Backend/Lower.cpp
rajatd: Yes, sorry I missed that.

---
In reply to: [67801018](https://github.com/Microsoft/ChakraCore/pull/1135#discussion_r67801018) [](ancestors = 67801018,67748513)

lib/Backend/GlobOpt.cpp
```diff
@@ -3825,32 +3826,156 @@ GlobOpt::TestAnyArgumentsSym()
...
+            }
+        }
+        else if (instr->m_opcode == Js::OpCode::InlineeStart)
+        {
+            Assert(instr->m_func->IsInlined());
+            Js::ArgSlot actualsCount = instr->m_func->actualCount - 1;
+            Js::ArgSlot formalsCount = instr->m_func->GetJnFunction()->GetInParamsCount() - 1;
+
+            Func * func = instr->m_func;
+            Func * topFunc = this->func;
```
rajatd: This should be instr's func's immediate parent func.

test/Function/StackArgsWithFormals.js
```diff
@@ -0,0 +1,168 @@
...
+{
+    if(shouldBailout)
+    {
+        arguments[0] = 4;
+        for(var o in arguments)
+        {
+            print(o);
+        }
+    }
+    WScript.Echo(argMath3);
```
akroshg: nit:print ?

test/Function/StackArgsWithFormals.js
akroshg: Discussed this in person - we should be tracing if the optimization is enabled or not. 
Also instead of printing values in the baseline, it is better if is expected and actual values comparisons in the test file itself and print only if that fails. 

---
In reply to: [67812487](https://github.com/Microsoft/ChakraCore/pull/1135#discussion_r67812487) [](ancestors = 67812487,67765923)

lib/Backend/GlobOpt.cpp
satheeshravi: Yes. My bad :( Fixed.

---
In reply to: [67909136](https://github.com/Microsoft/ChakraCore/pull/1135#discussion_r67909136) [](ancestors = 67909136)

lib/Backend/GlobOpt.cpp
satheeshravi: Fixed

---
In reply to: [67799824](https://github.com/Microsoft/ChakraCore/pull/1135#discussion_r67799824) [](ancestors = 67799824,67736248)

lib/Backend/BackwardPass.cpp
satheeshravi: Sure. I actually meant that but wrote it wrongly. Done

---
In reply to: [67908395](https://github.com/Microsoft/ChakraCore/pull/1135#discussion_r67908395) [](ancestors = 67908395,67800979,67741254)

lib/Backend/BackwardPass.cpp
```diff
@@ -252,6 +252,84 @@ BackwardPass::CleanupBackwardPassInfoInFlowGraph()
...
+void
+BackwardPass::InsertArgInsForFormals()
+{
+    if (func->IsStackArgsEnabled() && !func->GetJnFunction()->GetHasImplicitArgIns())
+    {
+        IR::Instr * insertAfterInstr = func->m_headInstr->m_next;
+        AssertMsg(insertAfterInstr->IsLabelInstr(), "First Instr of the first block should always have a label");
+
+        Js::ArgSlot paramsCount = insertAfterInstr->m_func->GetJnFunction()->GetInParamsCount() - 1;
+        IR::Instr *     argIninstr = nullptr;
```
ThomsonTan: argInInstr would be more readable?

test/Function/StackArgsWithFormals.js
satheeshravi: Done.

---
In reply to: [67927884](https://github.com/Microsoft/ChakraCore/pull/1135#discussion_r67927884) [](ancestors = 67927884)

lib/Backend/BackwardPass.cpp
satheeshravi: Done.

---
In reply to: [67976968](https://github.com/Microsoft/ChakraCore/pull/1135#discussion_r67976968) [](ancestors = 67976968)

lib/Backend/BackwardPass.cpp
```diff
@@ -2735,6 +2848,168 @@ BackwardPass::ProcessBlock(BasicBlock * block)
...
+                if (src1 && src1->IsSymOpnd())
+                {
+                    Sym * sym = src1->AsSymOpnd()->m_sym;
+                    Assert(sym);
+                    if (IsFormalParamSym(currFunc, sym))
+                    {
+                        AssertMsg(!currFunc->GetJnFunction()->GetHasImplicitArgIns(), "We don't have mappings between named formals and arguments object here");
+
+                        instr->m_opcode = Js::OpCode::Ld_A;
+                        PropertySym * propSym = sym->AsPropertySym();
```
ThomsonTan: Could it be a non-property sym here?

lib/Backend/BackwardPass.cpp
```diff
@@ -252,6 +252,84 @@ BackwardPass::CleanupBackwardPassInfoInFlowGraph()
...
+        {
+            AssertMsg(instr->GetBailOutKind() & IR::BailOnStackArgsOutOfActualsRange, "Stack args bail out is not set when the optimization is turned on? ");
+            if (instr->GetBailOutKind() & ~IR::BailOnStackArgsOutOfActualsRange)
+            {
+                Assert(instr->GetBailOutKind() == (IR::BailOnStackArgsOutOfActualsRange | IR::BailOutOnImplicitCallsPreOp));
+                //We are sure at this point, that we will not have any implicit calls as we don't have any call to helper.
+                instr->SetBailOutKind(IR::BailOnStackArgsOutOfActualsRange);
+            }
+        }
+        else if (instr->HasBailOutInfo() && (instr->GetBailOutKind() & IR::BailOnStackArgsOutOfActualsRange))
```
pleath: Can you take out this check on HasBailOutInfo, since you checked it above?

lib/Backend/BackwardPass.cpp
```diff
@@ -2735,6 +2848,168 @@ BackwardPass::ProcessBlock(BasicBlock * block)
...
+    if (tag == Js::DeadStorePhase && this->currentInstr->IsStackArgsEnabled())
+    {
+        Func * currFunc = this->currentInstr->m_func;
+        switch (this->currentInstr->m_opcode)
+        {
+            case Js::OpCode::LdHeapArguments:
+            case Js::OpCode::LdHeapArgsCached:
+            case Js::OpCode::LdLetHeapArguments:
+            case Js::OpCode::LdLetHeapArgsCached:
+            {
```
pleath: I didn't see cases for the LdHeapArguments family in the method below that does the dead store. Is that deliberate?

lib/Runtime/ByteCode/ByteCodeGenerator.cpp
```diff
@@ -2333,9 +2333,31 @@ FuncInfo* PreVisitFunction(ParseNode* pnode, ByteCodeGenerator* byteCodeGenerato
...
+                if (PHASE_TRACE1(Js::PerfHintPhase) && !doStackArgsOpt)
+                {
+                    WritePerfHint(PerfHints::HeapArgumentsDueToWriteToFormals, funcInfo->GetParsedFunctionBody(), 0);
+                }
+#endif
+                //Go conservative if it has any nested functions, or any non-local references.
+                //With statements - need scope object to be present.
+                //Nested funtions - need scope object to be present - LdEnv/LdFrameDisplay needs it.
+                if ((doStackArgsOpt && pnode->sxFnc.funcInfo->GetParamScope()->Count() > 1) && (pnode->sxFnc.funcInfo->HasDeferredChild() || pnode->sxFnc.nestedCount > 0 ||
+                    pnode->sxFnc.HasWithStmt() || byteCodeGenerator->IsInDebugMode() || PHASE_OFF1(Js::StackArgFormalsOptPhase) || PHASE_OFF1(Js::StackArgOptPhase)))
```
pleath: This seems too conservative. We don't need to disable stack args in all nested function cases, only when there's a nested reference to a formal. To detect that, you can walk the formals looking at sym->hasNonLocalReference. Can you explain why you disable in the presence of "with"? (I assume you disable in debug mode because of console eval.)

lib/Parser/Parse.cpp
```diff
@@ -8437,6 +8448,36 @@ ParseNodePtr Parser::ParseExpr(int oplMin,
...
+        if (pToken->tk == tkID)
+        {
+            PidRefStack *ref = pToken->pid->GetTopRef();
+            Assert(ref);
+            ref->TrackAssignment();
+        }
+    }
+}
+
+void PidRefStack::TrackAssignment()
```
akroshg: this is not needed now, all remains in this function is a single line.

lib/Parser/Parse.cpp
```diff
@@ -8437,6 +8448,36 @@ ParseNodePtr Parser::ParseExpr(int oplMin,
     return pnode;
 }
 
+template<bool buildAST>
+void Parser::TrackAssignment(ParseNodePtr pnodeT, IdentToken* pToken)
+{
+    if (buildAST)
+    {
+        Assert(pnodeT != NULL);
```
akroshg: nit: change this to nullptr.

lib/Runtime/Base/FunctionBody.h
```diff
@@ -2690,6 +2693,22 @@ namespace Js
...
+            // We can avoid creating a scope object with arguments present if:
+            bool dontNeedScopeObject =
+                // Either we are in strict mode, or have strict mode formal semantics from a non-simple parameter list, and
+                (GetIsStrictMode() || hasNonSimpleParams)
+                // Neither of the scopes are objects
+                && !HasScopeObject();
+            
+            return 
+                // Regardless of the conditions above, we won't need a scope object if there aren't any formals.
+                (GetInParamsCount() > 1 || GetHasRestParameter())
```
akroshg: Rest parameter makes param to non-simple param already. Do you have redundant check here?

lib/Runtime/Base/FunctionBody.cpp
```diff
@@ -4730,6 +4758,7 @@ namespace Js
         this->SetScopeInfo(nullptr);
         this->SetCodeGenRuntimeData(nullptr);
         this->cacheIdToPropertyIdMap = nullptr;
+        this->SetFormalsPropIdArray(nullptr);
```
akroshg: do you need to do anything on the FunctionBody::CloneByteCodeInto function?

lib/Runtime/Library/ArgumentsObject.h
akroshg: nit : it is more readable if you do frameObject == nullptr

---
In reply to: [67801971](https://github.com/Microsoft/ChakraCore/pull/1135#discussion_r67801971) [](ancestors = 67801971,67771245)

lib/Runtime/ByteCode/ByteCodeSerializer.cpp
```diff
@@ -3298,6 +3327,63 @@ class ByteCodeBufferReader
...
+
+        bool has__proto__;
+        current = ReadBool(current, &has__proto__);
+        propIds->has__proto__ = has__proto__;
+
+        bool hasNonSimpleParams;
+        current = ReadBool(current, &hasNonSimpleParams);
+        propIds->hasNonSimpleParams = hasNonSimpleParams;
+
+        int id;
```
akroshg: initialize this as well.

lib/Backend/Func.cpp
```diff
@@ -1552,6 +1555,128 @@ Func::EnsureFuncEndLabel()
     return m_funcEndLabel;
 }
 
+StackArgWithFormalsTracker *
```
ThomsonTan: nit: no need to return it out?

lib/Backend/Func.h
```diff
@@ -52,6 +52,36 @@ struct Cloner
...
 
+/*
+* This class keeps track of various information required for Stack Arguments optimization with formals.
+*/
+class StackArgWithFormalsTracker
+{
+private:
+    BVSparse<JitArenaAllocator> *  formalsArraySyms;    //Tracks Formal parameter Array - Is this Bv required explicitly?
+    StackSym**                     formalsIndexToStackSymMap; //Tracks the stack sym for each formal
+    StackSym*                      m_scopeObjSym;   // Track the stack sym for the scope object that is created.
```
ThomsonTan: Tracks? Just keep consistent.

satheeshravi: Done.

---
In reply to: [68144068](https://github.com/Microsoft/ChakraCore/pull/1135#discussion_r68144068) [](ancestors = 68144068)

test/Function/StackArgsWithFormals.js
satheeshravi: Done.

---
In reply to: [67928173](https://github.com/Microsoft/ChakraCore/pull/1135#discussion_r67928173) [](ancestors = 67928173,67812487,67765923)

lib/Runtime/Library/ArgumentsObject.h
satheeshravi: Done.

---
In reply to: [68136675](https://github.com/Microsoft/ChakraCore/pull/1135#discussion_r68136675) [](ancestors = 68136675,67801971,67771245)

lib/Backend/BackwardPass.cpp
satheeshravi: We can't have a non-property sym with LdSlot.

---
In reply to: [68091293](https://github.com/Microsoft/ChakraCore/pull/1135#discussion_r68091293) [](ancestors = 68091293)

satheeshravi: If it has an invalid source sym, it will fail when performing the casting.

---
In reply to: [68147586](https://github.com/Microsoft/ChakraCore/pull/1135#discussion_r68147586) [](ancestors = 68147586,68091293)

lib/Backend/BackwardPass.cpp
satheeshravi: oh sure. Done

---
In reply to: [68112935](https://github.com/Microsoft/ChakraCore/pull/1135#discussion_r68112935) [](ancestors = 68112935)

lib/Backend/BackwardPass.cpp
satheeshravi: Actually the instr itself doesn't get dead stored, but it helps for not marking the sym of the LdHeapArguments sym from getting marked as SymUsed

---
In reply to: [68114689](https://github.com/Microsoft/ChakraCore/pull/1135#discussion_r68114689) [](ancestors = 68114689)

satheeshravi: We retain LdHeapArgXXX instrs and later lower them to MOV NULL in lowerer.

---
In reply to: [68148183](https://github.com/Microsoft/ChakraCore/pull/1135#discussion_r68148183) [](ancestors = 68148183,68114689)

lib/Runtime/ByteCode/ByteCodeGenerator.cpp
satheeshravi: For nested functions, we still require scope object creation and my current implementation doesn't support partial optimization (eliminating heap arguments creation but retaining scope obj creation)- This is on my plate for post-check-in work.
The case with "with: - It created a new scope in the scope chain. So it is creating an alias for scope object and working with it. Again, here we can do partial optimization of retaining scope obj creation and eliminating heap arguments optimization.

---
In reply to: [68118003](https://github.com/Microsoft/ChakraCore/pull/1135#discussion_r68118003) [](ancestors = 68118003)

lib/Backend/BackwardPass.cpp
pleath: Okay. Thanks. Would you mind adding a comment to that effect? The asymmetry seems a little weird otherwise.

lib/Runtime/ByteCode/ByteCodeGenerator.cpp
pleath: We only require scope object creation if there are actually non-local references to the scope object contents. You can check that by looking at pnode->sxFnc.funcInfo->hasLocalInClosure instead of looking at nested functions. I assume calls to eval already disable the optimization? As for "with", I don't understand. Are you talking about the code we generate when there's a call to a local function from inside "with"?

lib/Backend/IRBuilder.cpp
```diff
@@ -1698,35 +1732,6 @@ IRBuilder::BuildReg2(Js::OpCode newOpcode, uint32 offset, Js::RegSlot R0, Js::Re
 
     switch (newOpcode)
     {
-    case Js::OpCode::LdHeapArguments:
```
ThomsonTan: Why move this to BuildReg1?

satheeshravi: Bytecode layout of this instruction is changed from Two Source Reg to One Source Reg.
Earlier it was taking property id array as one of the operands - Now we can retrieve the property id array of the formals from the FunctionBody directly.

---
In reply to: [68152909](https://github.com/Microsoft/ChakraCore/pull/1135#discussion_r68152909) [](ancestors = 68152909)

lib/Backend/BackwardPass.cpp
satheeshravi: Sure. Done. Added a detailed comment

---
In reply to: [68150708](https://github.com/Microsoft/ChakraCore/pull/1135#discussion_r68150708) [](ancestors = 68150708)

lib/Runtime/ByteCode/ByteCodeSerializer.cpp
satheeshravi: Done.

---
In reply to: [68136835](https://github.com/Microsoft/ChakraCore/pull/1135#discussion_r68136835) [](ancestors = 68136835)

lib/Backend/Func.cpp
satheeshravi: Done.

---
In reply to: [68142569](https://github.com/Microsoft/ChakraCore/pull/1135#discussion_r68142569) [](ancestors = 68142569)

lib/Runtime/ByteCode/OpCodes.h
```diff
@@ -318,7 +318,7 @@ MACRO_EXTEND_WMS(       InitClassMemberSetComputedName,ElementI,    OpSideEffect
 MACRO_EXTEND_WMS(       InitClassMemberGetComputedName,ElementI,    OpSideEffect|OpHasImplicitCall|OpPostOpDbgBailOut)                  // Class member in get syntax with computed property name
 MACRO_EXTEND_WMS(       BrOnClassConstructor,       BrReg1,         None)               // Branch if argument is a class constructor
 
-MACRO_BACKEND_ONLY(     ArgIn_A,                    Empty,          OpSideEffect)       // Copy from "in slot" to "local slot", unchecked
```
ThomsonTan: Is it safe to remove OpSideEffect?

lib/Runtime/Base/FunctionBody.cpp
satheeshravi: Oh yeah. Added it now. Do you now, in what scenarios we clone a function body ?

lib/Runtime/ByteCode/OpCodes.h
satheeshravi: @ThomsonTan : Yes. It is safe to remove OpSideEffect for ArgIn. We talked about it during design to make sure that it really can't have any side effect. The reason for removing this attribute is to enable Dead store for ArgIn insturctions (where some formals that are not being used will be dead stored.)

lib/Runtime/Base/FunctionBody.cpp
pleath: HTC is the only one I'm aware of.

lib/Parser/Parse.cpp
satheeshravi: Done.

lib/Parser/Parse.cpp
satheeshravi: Done.

---
In reply to: [68135024](https://github.com/Microsoft/ChakraCore/pull/1135#discussion_r68135024) [](ancestors = 68135024)

lib/Runtime/Base/FunctionBody.h
satheeshravi: Rest parameter requires scope object, but non-simple parameters(without rest) doesn't require scope object. So the second check is required. I want to maintain the same semantics of what we have in ByteCodeGenerator::NeedScopeObjectForArguments(FuncInfo *funcInfo, ParseNode *pnodeFnc) 

---
In reply to: [68135396](https://github.com/Microsoft/ChakraCore/pull/1135#discussion_r68135396) [](ancestors = 68135396)

satheeshravi:
@ThomsonTan , @pleath , @akroshg - Have addressed all new comments. Can you please verify and close the comments in Codeflow ?

lib/Runtime/Base/FunctionBody.cpp
```diff
@@ -6013,6 +6043,14 @@ namespace Js
     }
 #endif
 
+    PropertyIdArray * FunctionBody::AllocatePropertyIdArrayForFormals(uint32 size, uint32 count)
+    {
+        //TODO: saravind: Should the allocation be a Leaf Allocation?
```
satheeshravi: @digitalinfinity : Can you please comment on this ?

akroshg:
:shipit:

pleath:
So what is the plan for dealing with the overly-conservative disabling of the optimization?

--Paul

satheeshravi:
@pleath : I am going to work on it next week along with other left overs.

pleath:
Okay. LGTM, then.

--Paul

rajatd:
:shipit:

lib/Runtime/Language/InterpreterStackFrame.cpp
```diff
@@ -8330,6 +8331,92 @@ const byte * InterpreterStackFrame::OP_ProfiledLoopBodyStart(const byte * ip)
...
+            //We always set the Frame object to nullptr in BailOutRecord::EnsureArguments for stack args optimization.
+            if (((Js::HeapArgumentsObject*)(m_arguments))->GetFrameObject() == nullptr)
+            {
+                heapArgObj = (Js::HeapArgumentsObject*)m_arguments;
+            }
+            else
+            {
+                Assert(false);
+                return;
+            }
```
satheeshravi: I have removed this else block. Was added for debugging purpose.

rajatd:
You'll probably want to regenerate no-jit intl bytecodes as well.

satheeshravi:
Yes. Done.

---
In reply to: [228410808](https://github.com/Microsoft/ChakraCore/pull/1135#issuecomment-228410808) [](ancestors = 228410808)

lib/Backend/BackwardPass.cpp
LouisLaf: So it is guaranteed not to be used by another instr?  A better implementation would have been to set the bit because of this instr in the first place.  Now you are turning it off even if someone else had a valid use.

---
In reply to: [67800813](https://github.com/Microsoft/ChakraCore/pull/1135#discussion_r67800813) [](ancestors = 67800813,67740116)

LouisLaf: Seems like you already do this in ProcessSymUse...  So why is it still needed here?

---
In reply to: [68475602](https://github.com/Microsoft/ChakraCore/pull/1135#discussion_r68475602) [](ancestors = 68475602,67800813,67740116)

satheeshravi: I guess, you are right. 
Now that I am thinking back, this piece is not required. Let me remove this and add back the tracking for scope object removal - So that i can use this information to decide whether to recreate the scope object or use the existing one in the bail out path.

Currently, I kinda rely on the fact that the scope object will not be restored for sure, because i am removing the bit here - and will always recreate the scope object in the bail out path.

---
In reply to: [68476633](https://github.com/Microsoft/ChakraCore/pull/1135#discussion_r68476633) [](ancestors = 68476633,68475602,67800813,67740116)

satheeshravi: Actually, I remember why this code is required. There will be some paths which restores scope obj and some path which doesn't restore. So while creating heap arguments object -> we need do not know if the scope object is being restored or not in RestoreValues().  I had two options to tackle this -> either restore scope object in all paths or just recreate it always during the heap arguments creation  (i implemented the second way). 
So it is not guaranteed that there won't be other uses, but it is guaranteed that it will be restored correctly. 
Also, until we complete the forward pass, we do not if we are going to do the optimization or not. That is the reason, why i had to this in the dead store pass.
That's why I always clear the bit here.

I didn't understand your other proposal though.

Let me know if this is unclear.

---
In reply to: [68478772](https://github.com/Microsoft/ChakraCore/pull/1135#discussion_r68478772) [](ancestors = 68478772,68476633,68475602,67800813,67740116)

LouisLaf:
:shipit:

PR #1137, opened 2016-06-14
During reparse script context can get closed and destroy debug context, cache script context to check closed status

agarwal-sandeep:
During reparse script context can get closed and destroy debug context, cache script context to check closed status
Only hard fail for reparse failure if scriptcontext is not closed state

### DIFF
--- a/lib/Runtime/Base/ScriptContext.cpp
+++ b/lib/Runtime/Base/ScriptContext.cpp
@@ -2743,62 +2743,64 @@ if (!sourceList)
 
         hr = this->GetDebugContext()->RundownSourcesAndReparse(shouldPerformSourceRundown, /*shouldReparseFunctions*/ true);
 
+        if (this->IsClosed())
+        {
+            return hr;
+        }
+
         // Debugger attach/detach failure is catastrophic, take down the process
         DEBUGGER_ATTACHDETACH_FATAL_ERROR_IF_FAILED(hr);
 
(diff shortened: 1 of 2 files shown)


lib/Runtime/Base/ScriptContext.cpp
```diff
@@ -2743,11 +2743,11 @@ if (!sourceList)
 
         hr = this->GetDebugContext()->RundownSourcesAndReparse(shouldPerformSourceRundown, /*shouldReparseFunctions*/ true);
 
-        // Debugger attach/detach failure is catastrophic, take down the process
-        DEBUGGER_ATTACHDETACH_FATAL_ERROR_IF_FAILED(hr);
-
         if (!this->IsClosed())
```
akroshg: I am okay to move the FATAL error inside, but why we are proceeding when the this is closed? why not bail out from this function in that case?

akroshg: Oh - we are not proceeding. I go confused that there is something after this 'if'.

akroshg:
LGTM.

lib/Runtime/Base/ScriptContext.cpp
agarwal-sandeep: Everything is inside the !closed check. I will change the check to return hr if closed and move rest of code out of if.

PR #1138, opened 2016-06-14
Fix SAL annotation.

dilijev:
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/1138)

### DIFF
--- a/lib/Runtime/Library/TypedArray.cpp
+++ b/lib/Runtime/Library/TypedArray.cpp
@@ -2844,7 +2844,7 @@ namespace Js
         return false;
     }
 
-    BOOL TypedArrayBase::ValidateIndexAndDirectSetItem(__in Js::Var index, __in Js::Var value, __in bool * isNumericIndex)
+    _Use_decl_annotations_ BOOL TypedArrayBase::ValidateIndexAndDirectSetItem(Js::Var index, Js::Var value, bool * isNumericIndex)
     {
         bool skipSetItem = false;
         uint32 indexToSet = ValidateAndReturnIndex(index, &skipSetItem, isNumericIndex);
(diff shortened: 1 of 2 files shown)


dilijev:
Thanks @Yongqu for pointing this out.

lib/Runtime/Library/TypedArray.cpp
```diff
@@ -2844,7 +2844,7 @@ namespace Js
         return false;
     }
 
-    BOOL TypedArrayBase::ValidateIndexAndDirectSetItem(__in Js::Var index, __in Js::Var value, __in bool * isNumericIndex)
+    BOOL TypedArrayBase::ValidateIndexAndDirectSetItem(__in Js::Var index, __in Js::Var value, __out bool * isNumericIndex)
```
boingoing: You can use _Use_decl_annotations_ here so that you don't have to annotate both function declarations and definitions.

boingoing:
Looks fine to me. I'd prefer to use `_Use_decl_annotations_` at the point of function definition and leave the annotations in the header.

lib/Runtime/Library/TypedArray.cpp
dilijev: Sounds like a good idea - but AFAICT we haven't done that anywhere in our code base so far. Wonder why?

dilijev:
@boingoing look good?

boingoing:
LGTM

lib/Runtime/Library/TypedArray.cpp
boingoing: This macro is in SAL v2, which we haven't been using until recently.

dilijev:
Thanks @boingoing and @Yongqu

PR #1139, opened 2016-06-14
Fixing an issue with symbol capturing in function defined in catch block's param

aneeshdk:
Fixing an issue with symbol capturing in function defined in catch block's
param

When catch block's param has a function definition, its node should be added to the list after the catch block. Right now we were adding the catch block to the list only after the params are parsed. Also during the visit the pnodescopes should be visited separate, not along with the params.

### DIFF
--- a/lib/Parser/Parse.cpp
+++ b/lib/Parser/Parse.cpp
@@ -8876,6 +8876,31 @@ ParseNodePtr Parser::ParseCatch()
 
         pnodeCatchScope = StartParseBlock<buildAST>(PnodeBlockType::Regular, isPattern ? ScopeType_CatchParamPattern : ScopeType_Catch);
 
+        if (buildAST)
+        {
+            // Add this catch to the current scope list.
+
+            if (m_ppnodeExprScope)
+            {
+                Assert(*m_ppnodeExprScope == nullptr);
+                *m_ppnodeExprScope = pnode;
+                m_ppnodeExprScope = &pnode->sxCatch.pnodeNext;
+            }
+            else
+            {
+                Assert(m_ppnodeScope);
+                Assert(*m_ppnodeScope == nullptr);
+                *m_ppnodeScope = pnode;
+                m_ppnodeScope = &pnode->sxCatch.pnodeNext;
+            }
+
+            // Keep a list of function expressions (not declarations) at this scope.
+
+            ppnodeExprScopeSave = m_ppnodeExprScope;
+            m_ppnodeExprScope = &pnode->sxCatch.pnodeScopes;
+            pnode->sxCatch.pnodeScopes = nullptr;
+        }
+
         if (isPattern)
         {
             ParseNodePtr pnodePattern = ParseDestructuredLiteral<buildAST>(tkLET, true /*isDecl*/, true /*topLevel*/, DIC_ForceErrorOnInitializer);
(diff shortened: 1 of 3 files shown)


lib/Parser/Parse.cpp
```diff
@@ -8876,6 +8876,31 @@ ParseNodePtr Parser::ParseCatch()
 
         pnodeCatchScope = StartParseBlock<buildAST>(PnodeBlockType::Regular, isPattern ? ScopeType_CatchParamPattern : ScopeType_Catch);
 
+        if (buildAST)
```
akroshg: do we need to take care of defer parse case?

lib/Runtime/ByteCode/ByteCodeGenerator.cpp
```diff
@@ -3117,14 +3117,24 @@ void VisitNestedScopes(ParseNode* pnodeScopeList, ParseNode* pnodeParent, ByteCo
...
             PreVisitCatch(pnodeScope, byteCodeGenerator);
 
-            Visit(pnodeScope->sxCatch.pnodeParam, byteCodeGenerator, prefix, postfix);
             if (pnodeScope->sxCatch.pnodeParam->nop == knopParamPattern)
             {
+                Parser::MapBindIdentifier(pnodeScope->sxCatch.pnodeParam->sxParamPattern.pnode1, [byteCodeGenerator](ParseNodePtr pnode)
+                {
+                    Assert(pnode->nop == knopLetDecl);
+                    pnode->sxVar.sym->SetLocation(byteCodeGenerator->NextVarRegister());
+                });
```
akroshg: don't you have to call
Visit(pnodeScope->sxCatch.pnodeParam, byteCodeGenerator, prefix, postfix);
after this in the param pattern case?

test/es6/destructuring_catch.js
```diff
@@ -189,6 +189,93 @@ var tests = [
...
+            (function() {
+                try {
+                    throw ['inside', {}];
+                } catch ([x, { y = () => arguments[0] }]) {
+                    assert.areEqual(y(), 10, "Function should be able to capture the arguments symbol from the parent function");
+                    assert.areEqual(x, 'inside', "Function should be able to capture symbols from try's body properly");
+                } 
+            })(10);
+
+            (function(a = 1, b = () => a) {
```
akroshg: good to repeat few tests with 'eval' in it (in the try block, catch block and eval in the function on param pattern).

lib/Parser/Parse.cpp
akroshg: ah - you just moved this up. never mind.

---
In reply to: [67058501](https://github.com/Microsoft/ChakraCore/pull/1139#discussion_r67058501) [](ancestors = 67058501)

akroshg:
Looks good to me otherwise

test/es6/destructuring_catch.js
aneeshdk: Done

test/es6/destructuring_catch.js
```diff
@@ -189,6 +189,164 @@ var tests = [
...
+                } catch ([x = 10, y = function() { return x; }]) {
+                    eval("");
+                    assert.areEqual(y(), 10, "Function should be able to capture symbols from catch's param");
+                } 
+            })();
+
+            (function() {
+                try {
+                    throw {};
+                } catch ({x = 10, y = function() { return x; }}) {
```
akroshg: what happens if you have eval in the function, which is defined at the catch param ?

aneeshdk: Line 307 has a test case with eval in the function

akroshg: great.

akroshg:
Looks good to me.

PR #1141, opened 2016-06-15
xplat: fix win32 JsParseSerializedScript/JsRunSerializedScript

jianchun:
Previous implementation has some TODOs (passing nullptr for "script"
parameter).

The "script" parameter was removed in Utf8 version of the APIs. But we need
to support it for existing win32 APIs.

For win32 version, use dummy load/unload callbacks with extra internal
"scriptLoadSourceContext". That sourceContext is stored by
JsrtSourceHolder and only used to call the loadSourceCallback. Thus I can
store a dummy sourceContext for the dummy callbacks, without affecting
the real sourceContext.


jianchun:
@agarwal-sandeep @digitalinfinity Figured out an easy way to fix the win32 version.

PR #1142, opened 2016-06-15
Wasm nyi & errors

Cellule:
Mark wasm operators that are not fully working as NYI 
Simplify the Compilation error message format 
Output the name of the operator that is NYI  in the error message

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/1142)

### DIFF
--- a/lib/WasmReader/WasmByteCodeGenerator.cpp
+++ b/lib/WasmReader/WasmByteCodeGenerator.cpp
@@ -336,10 +336,10 @@ WasmBytecodeGenerator::GenerateFunction()
     {
         if (!PHASE_ON1(Js::WasmLazyTrapPhase))
         {
-            throw WasmCompilationException(_u("%s\n  Function %s"), ex.GetErrorMessage(), functionName);
+            throw WasmCompilationException(_u("function %s: %s"), functionName, ex.GetErrorMessage());
         }
         Assert(m_module->lazyTraps != nullptr);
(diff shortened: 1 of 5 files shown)


lib/WasmReader/WasmByteCodeGenerator.cpp
```diff
@@ -690,8 +698,10 @@ WasmBytecodeGenerator::EmitCall()
         switch (info.type)
         {
         case WasmTypes::F32:
-            Assert(wasmOp != wnCALL_IMPORT);
-            // REVIEW: support FFI call with f32 params?
+            if (wasmOp == wnCALL_IMPORT) 
+            {
+                throw WasmCompilationException(_u("External calls with float argument NYI"));
```
MikeHolman: I'm not sure that the JS ABI supports this. Does it?

MikeHolman: I guess there's no real reason to forbid, but we would still convert to double (when creating the js number) before calling.

MikeHolman:
lgtm

lib/WasmReader/WasmByteCodeGenerator.cpp
Cellule: The main reason I put this is because there is a test in the WebAssembly/testsuite that hit this assert. Therefore, I assume it is possible to do in Wasm.
I haven't thought on how to do it yet. This is to convert an assert into a runtime error.

PR #1143, opened 2016-06-15
xplat: build.sh improvements

jianchun:
1. Enforce standard `--option=VALUE` syntax (previously accepts
--option?VALUE).

2. Use standard `-j [N], --jobs[=N]` syntax.

 Users used to `make` syntax may run `-j N`, which was handled as `-j`
 without arg and considered infinity jobs by `make` previously. This
 hangs Ubuntu VM.

 Fix the script to accept full `-j [N], --jobs[=N]` syntax. When option is
 specified without N, use appropriate default to avoid VM hang.

3. Error on unknown command line options (previously silently ignores).

4. Allow current working directory to be a directory other than
chakracore repo root.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -17,26 +17,27 @@ PRINT_USAGE() {
     echo ""
     echo "[ChakraCore Build Script Help]"
     echo ""
-    echo "  build.sh [options]"
+    echo "build.sh [options]"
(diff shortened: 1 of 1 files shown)


digitalinfinity:
LGTM

PR #1144, opened 2016-06-15
fix debugging asm.js when closure is named function expression

MikeHolman:
FunctionBody wants named function to be in slotArray even though it isn't actually closure captured, because of our temporary deferral of inner functions

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/1144)


MikeHolman:
@aneeshdk @tcare @agarwal-sandeep: can you take a look?

aneeshdk:
LGTM

tcare:
We don't expect this outside of debugger attach, right? Should we assert as much?

Otherwise LGTM.

MikeHolman:
@tcare This will be the case always, because we always temporarily defer asm.js functions.

tcare:
Ah, makes sense.

PR #1145, opened 2016-06-15
Wasm br table immediate

Cellule:
Since Wasm doesn't have a constant table, the IR for switch cases wouldn't build properly with the added Ld_I4 before every case. 
Therefore, I added a bytecode with the case value as an immediate instead of a register. 
This also generates a smaller bytecode for switch statements
I had to do some refactoring in CaseNode.h to properly support `IntConstOpnd` and also minimize code duplication

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/1145)


Cellule:
@MikeHolman @commonlisp  Ping

MikeHolman:
lgtm

PR #1146, opened 2016-06-16
BUG 7853164: Check for nullptr before searching prototype chain; if propertyDescriptor is null then it's prototype will be nullptr.

dilijev:
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/1146)


curtisman:
LGTM

dilijev:
Thanks @curtisman

PR #1147, opened 2016-06-16
Calls to overridable iteration methods should go through ExecuteImplicitCall so that side-effects are tracked if we wind up executing user code.

pleath:


pleath:
@ianwjhalliday @suwc

pleath:
@akroshg @aneeshdk @tcare @boingoing @kunalspathak This is related to yesterday's discussion about implicit calls.

lib/Runtime/Language/JavascriptOperators.cpp
```diff
@@ -10449,18 +10459,30 @@ namespace Js
...
+                return scriptContext->GetLibrary()->GetUndefined();
+            }
             JavascriptError::ThrowTypeError(scriptContext, JSERR_NeedFunction);
         }
 
         RecyclableObject* callable = RecyclableObject::FromVar(func);
         Js::Var args[] = { iterator, value };
         Js::CallInfo callInfo(Js::CallFlags_Value, _countof(args) + (value == nullptr ? -1 : 0));
-        Var result = JavascriptFunction::CallFunction<true>(callable, callable->GetEntryPoint(), Js::Arguments(callInfo, args));
+        Var result = threadContext->ExecuteImplicitCall(callable, ImplicitCall_Accessor, [&]() -> Var
```
akroshg: I am implementing the IteratorClose, which is similar to IteratorNext. This reminds me to changes that as well. Thanks,

lib/Runtime/Language/JavascriptOperators.cpp
```diff
@@ -10449,18 +10459,30 @@ namespace Js
...
+            {
```
akroshg: Also in this file - I have seen few other places we have ExecuteImplicitCall, but the whole function is wrapped under ExecuteImplicitCall, as opposed to what you have done (just the callfunction). Was that oversight?

akroshg:
Looks good to me - is there a way to add a unittest for this?

suwc:
:shipit:

lib/Runtime/Language/JavascriptOperators.cpp
pleath: No, this is intended. The C++ code passed to ExecuteImplicitCall should be the whole action you want to take if implicit calls are permitted, or avoid doing if they’re not. In some cases, there’s some checking/setup that only makes sense to do if you’re going to go ahead and call the user method, but here there isn’t much. (Though I suppose the one CallInfo creation could be moved into the ExecuteImplicitCall block.)

--Paul

pleath:
There already is, actually. The issue was exposed in a couple of UT’s when I was doing redeferral work.

--Paul

PR #1148, opened 2016-06-16
xplat: re-enable compile assert for BVSparseNode

xanlpz:
It's indeed 16 bytes on GNU/Linux 64bit Clang 3.8. Since it seems
important to avoid regressions here we should re-enable it.


obastemur:
p.s. We don't have back-end is enabled yet though.

digitalinfinity:
LGTM

PR #1150, opened 2016-06-17
cross-platform: Fix clang issues

obastemur:
- remove `-fno-inline-functions` from build as it breaks CC / Clang3.8+
   ATM we have all the tests are passing. So, we no longer need this compiler
   switch in place.

 - enable `Werror` to prevent commits are breaking clang support

 - enable both debug and test-build tests on CI

 - fix clang 3.8/3.9 template issues

### DIFF
--- a/lib/Runtime/Types/NullTypeHandler.h
+++ b/lib/Runtime/Types/NullTypeHandler.h
@@ -91,6 +91,6 @@ namespace Js
         static NullTypeHandler defaultInstance;
 
     public:
-        static NullTypeHandler * GetDefaultInstance() { return &defaultInstance; }
+        static NullTypeHandler * GetDefaultInstance();
     };
 }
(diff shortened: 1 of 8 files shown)


lib/Runtime/Types/NullTypeHandler.h
```diff
@@ -91,6 +91,6 @@ namespace Js
         static NullTypeHandler defaultInstance;
 
     public:
-        static NullTypeHandler * GetDefaultInstance() { return &defaultInstance; }
```
jianchun: This doesn't work with clang?

jianchun:
LGTM

lib/Runtime/Types/NullTypeHandler.h
obastemur: nope, none of these work. It asks for explicit declaration.

obastemur:

Issue #1128, opened 2016-06-12
function.name doesn't return name if function is not defer parsed


kunalspathak:
```js
var obj = {};
obj.foo = function () {
    print(obj.foo.name);
}
obj.foo();
```
 
Above test case prints function name only if function is defer parsed.
 
```cmd
ch.exe –Es6Experimental –forceDeferParse test.js
Prints “foo”

ch.exe –Es6Experimental test.js
Prints nothing.
 ```

kunalspathak:
FYI - @pleath , @akroshg , @abchatra

akroshg:
-forcedeferparse is incorrect. https://tc39.github.io/ecma262/#sec-assignment-operators-runtime-semantics-evaluation
We should assign the name when the LHS is identifier ref.

akroshg:
Fixed by #1151

PR #1151, opened 2016-06-17
Function.name issue when the LHS is not identifier ref.

akroshg:
The variable isNameIdentifierRef is used to guide the logic to give the function.name. That flag will be set when the LHS is reference identifier (say foo = function() {}), that works correct in the no-deferparse case. For example foo.bar =function() {}, the variable isNameIdentifier should be false. Which works correct till the function is deferred. However when we undefer that function, we parse the function after assignment (foo.bar information is lost). in that case the isNameIdentifier was not set correctly. Fixed that by setting back that flag to new function body from the old parseablefunctioninfo, after the parsing is done.
UT passed. Added one test.

### DIFF
--- a/lib/Runtime/Base/FunctionBody.cpp
+++ b/lib/Runtime/Base/FunctionBody.cpp
@@ -1912,6 +1912,9 @@ namespace Js
 
         if (fParsed == TRUE)
         {
+            // Restore if the function has nameIdentifier reference, as that name on the left side will not be parsed again while deferparse.
+            funcBody->SetIsNameIdentifierRef(this->GetIsNameIdentifierRef());
+
             this->UpdateFunctionBodyImpl(funcBody);
             this->m_hasBeenParsed = true;
         }
--- a/test/es6/function.name.js
+++ b/test/es6/function.name.js
@@ -761,6 +761,17 @@ var tests = [
             assert.areEqual("get", desc.get.name);
             assert.areEqual("set", desc.set.name);
         }
+    },
+	{
+        name: "Function name will be set only when LHS is identifier reference.",
+        body: function()
+        {
+            var obj = {};
+            obj.foo = function() {
+                assert.areEqual(obj.foo.name, "");
+            }
+            obj.foo();
+        }
     }
 
 ];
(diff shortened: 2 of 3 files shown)


akroshg:
Fixes #1128 
@kunalspathak have a look.

test/es6/function.name.js
```diff
@@ -761,6 +761,17 @@ var tests = [
...
             assert.areEqual("set", desc.set.name);
         }
+    },
+	{
+        name: "Function name will be set only when LHS is identifier reference.",
+        body: function()
+        {
+            var obj = {};
+            obj.foo = function() {
+                assert.areEqual(obj.foo.name, "");
```
kunalspathak: what happens when we have something like 
```js
var a = function b() {}. 
```
What does `a.name` return? Can we have a test case for that as well as for `function b() {}; b.name` case?

akroshg: Those are covered. at line 631?

kunalspathak:
:shipit:

lib/Runtime/Base/FunctionBody.cpp
```diff
@@ -1912,6 +1912,9 @@ namespace Js
 
         if (fParsed == TRUE)
         {
+            // Restore if the function has nameIdentifier reference, as that name on the left side will not be parsed again while deferparse.
+            funcBody->SetIsNameIdentifierRef(this->GetIsNameIdentifierRef());
```
farzonl: the bug is LHS is not an identifier ref. while I'm sure this solves the problem, is it appropriate to set name is identifier ref?  Does this bool need to be rename  to maintain spec logic in the code base?

akroshg: I need to set that as the parser above will change the state. imagine the code is 
`a.b = function() { /*blah*/}`
however in the defer parse we see the code as 
        ` function() { /*blah*/ } `
So I am restoring the previous state.
If you are talking about renaming, I am not sure if this is needed.

akroshg:
@dotnet-bot test this please

PR #1152, opened 2016-06-17
Capture bailout symbols incrementally in one block

ThomsonTan:
Capture constant and copy prop symbols for bailout takes lots of time in regexp benchmark from Octane one some low end machine, this change caches the first capture values for bailout, and reuse it when capturing value in the following bailout in the same block.

The new bailout is constructed according to the previous bailout and changed symbol value map since then.

### DIFF
--- a/lib/Backend/GlobOpt.cpp
+++ b/lib/Backend/GlobOpt.cpp
@@ -1061,6 +1061,9 @@ GlobOpt::MergePredBlocksValueMaps(BasicBlock *block)
         this->KillAllObjectTypes();
     }
 
+    this->blockData.capturedArgs = JitAnew(this->alloc, BVSparse<JitArenaAllocator>, this->alloc);
+    this->blockData.changedSyms = JitAnew(this->alloc, BVSparse<JitArenaAllocator>, this->alloc);
+
     this->CopyBlockData(&block->globOptData, &this->blockData);
 
     if (this->IsLoopPrePass())
@@ -1509,6 +1512,10 @@ GlobOpt::NulloutBlockData(GlobOptBlockData *data)
 
     data->stackLiteralInitFldDataMap = nullptr;
 
+    data->capturedValues = nullptr;
+    data->capturedArgs = nullptr;
+    data->changedSyms = nullptr;
+
     data->OnDataUnreferenced();
 }
 
@@ -1603,6 +1610,12 @@ GlobOpt::ReuseBlockData(GlobOptBlockData *toData, GlobOptBlockData *fromData)
 
     toData->stackLiteralInitFldDataMap = fromData->stackLiteralInitFldDataMap;
 
+    // toData->capturedValues = fromData->capturedValues;
+    toData->capturedArgs = fromData->capturedArgs;
+    toData->changedSyms = fromData->changedSyms;
+    toData->capturedArgs->ClearAll();
+    toData->changedSyms->ClearAll();
+
     toData->OnDataReused(fromData);
 }
 
@@ -1639,8 +1652,12 @@ GlobOpt::CopyBlockData(GlobOptBlockData *toData, GlobOptBlockData *fromData)
     toData->inlinedArgOutCount = fromData->inlinedArgOutCount;
     toData->hasCSECandidates = fromData->hasCSECandidates;
 
-    toData->stackLiteralInitFldDataMap = fromData->stackLiteralInitFldDataMap;
+    toData->capturedArgs = fromData->capturedArgs;
+    toData->changedSyms = fromData->changedSyms;
+    toData->capturedArgs->ClearAll();
+    toData->changedSyms->ClearAll();
 
+    toData->stackLiteralInitFldDataMap = fromData->stackLiteralInitFldDataMap;
     toData->OnDataReused(fromData);
 }
 
@@ -3188,7 +3205,7 @@ GlobOpt::MergeValues(
     // Set symStore if same on both paths.
     if (toDataValue->GetValueInfo()->GetSymStore() == fromDataValue->GetValueInfo()->GetSymStore())
     {
-        newValueInfo->SetSymStore(toDataValue->GetValueInfo()->GetSymStore());
+        this->SetSymStoreDirect(newValueInfo, toDataValue->GetValueInfo()->GetSymStore());
     }
 
     return newValue;
@@ -5059,6 +5076,14 @@ GlobOpt::OptInstr(IR::Instr *&instr, bool* isInstrRemoved)
         }
     }
 
+    if (instr->HasBailOutInfo() && !this->IsLoopPrePass())
+    {
+        this->currentBlock->globOptData.capturedArgs->ClearAll();
+        this->currentBlock->globOptData.changedSyms->ClearAll();
+        this->currentBlock->globOptData.capturedValues = 
+            this->currentBlock->globOptData.capturedValuesCandicate;
+    }
+
     return instrNext;
 }
 
@@ -5356,7 +5381,7 @@ GlobOpt::OptDst(
                         }
                     }
 
-                    dstVal->GetValueInfo()->SetSymStore(dstVarSym);
+                    this->SetSymStoreDirect(dstVal->GetValueInfo(), dstVarSym);
                 } while(false);
             }
         }
@@ -6365,7 +6390,7 @@ GlobOpt::CopyProp(IR::Opnd *opnd, IR::Instr *instr, Value *val, IR::IndirOpnd *p
 
     if (PHASE_OFF(Js::CopyPropPhase, this->func))
     {
-        valueInfo->SetSymStore(opndSym);
+        this->SetSymStoreDirect(valueInfo, opndSym);
         return opnd;
     }
 
@@ -6389,7 +6414,7 @@ GlobOpt::CopyProp(IR::Opnd *opnd, IR::Instr *instr, Value *val, IR::IndirOpnd *p
             //   fieldHoistSym = Ld_A t1    <- we're looking at t1 now, but want to copy-prop fieldHoistSym forward
             return opnd;
         }
-        valueInfo->SetSymStore(opndSym);
+        this->SetSymStoreDirect(valueInfo, opndSym);
     }
     return opnd;
 }
@@ -7057,10 +7082,23 @@ GlobOpt::SetSymStore(ValueInfo *valueInfo, Sym *sym)
     }
     if (valueInfo->GetSymStore() == nullptr || valueInfo->GetSymStore()->IsPropertySym())
     {
-        valueInfo->SetSymStore(sym);
+        SetSymStoreDirect(valueInfo, sym);
     }
 
     return sym;
+}
+
+void
+GlobOpt::SetSymStoreDirect(ValueInfo * valueInfo, Sym * sym)
+{
+    Sym * prevSymStore = valueInfo->GetSymStore();
+    if (prevSymStore && prevSymStore->IsStackSym() &&
+        prevSymStore->AsStackSym()->HasByteCodeRegSlot() &&
+        this->blockData.changedSyms)
+    {
+        this->blockData.changedSyms->Set(prevSymStore->m_id);
+    }
+    valueInfo->SetSymStore(sym);
 }
 
 void
@@ -7069,8 +7107,9 @@ GlobOpt::SetValue(GlobOptBlockData *blockData, Value *val, Sym * sym)
     ValueInfo *valueInfo = val->GetValueInfo();
 
     sym = this->SetSymStore(valueInfo, sym);
+    bool isStackSym = sym->IsStackSym();
 
-    if (sym->IsStackSym() && sym->AsStackSym()->IsFromByteCodeConstantTable())
+    if (isStackSym && sym->AsStackSym()->IsFromByteCodeConstantTable())
     {
         // Put the constants in a global array. This will minimize the per-block info.
         this->byteCodeConstantValueArray->Set(sym->m_id, val);
@@ -7079,6 +7118,10 @@ GlobOpt::SetValue(GlobOptBlockData *blockData, Value *val, Sym * sym)
     else
     {
         SetValueToHashTable(blockData->symToValueMap, val, sym);
+        if (blockData->changedSyms && isStackSym && sym->AsStackSym()->HasByteCodeRegSlot())
+        {
+            blockData->changedSyms->Set(sym->m_id);
+        }
     }
 }
 
@@ -7344,13 +7387,13 @@ GlobOpt::ValueNumberDst(IR::Instr **pInstr, Value *src1Val, Value *src2Val)
             // Update symStore for field hoisting
             if (loop != nullptr && (dstValueInfo != nullptr))
             {
-                dstValueInfo->SetSymStore(fieldHoistSym);
+                this->SetSymStoreDirect(dstValueInfo, fieldHoistSym);
             }
             // Update symStore if it isn't a stackSym
             if (dstVal && (!dstValueInfo->GetSymStore() || !dstValueInfo->GetSymStore()->IsStackSym()))
             {
                 Assert(dst->IsRegOpnd());
-                dstValueInfo->SetSymStore(dst->AsRegOpnd()->m_sym);
+                this->SetSymStoreDirect(dstValueInfo, dst->AsRegOpnd()->m_sym);
             }
             if (src1Val != dstVal)
             {
@@ -14318,7 +14361,7 @@ GlobOpt::ToTypeSpecUse(IR::Instr *instr, IR::Opnd *opnd, BasicBlock *block, Valu
                 if(!varSym || !IsFloat64TypeSpecialized(varSym, block))
                 {
                     // Clear the symstore to ensure it's set below to this new symbol
-                    val->GetValueInfo()->SetSymStore(nullptr);
+                    this->SetSymStoreDirect(val->GetValueInfo(), nullptr);
                     varSym = StackSym::New(TyVar, instr->m_func);
                     newFloatSym = true;
                 }
@@ -16253,7 +16296,7 @@ GlobOpt::OptArraySrc(IR::Instr * *const instrRef)
 
             // SetValue above would have set the sym store to newLengthSym. This sym won't be used for copy-prop though, so
             // remove it as the sym store.
-            lengthValue->GetValueInfo()->SetSymStore(nullptr);
+            this->SetSymStoreDirect(lengthValue->GetValueInfo(), nullptr);
 
             // length = [array + offsetOf(length)]
             IR::Instr *const loadLength =
@@ -16308,7 +16351,7 @@ GlobOpt::OptArraySrc(IR::Instr * *const instrRef)
                     Assert(!FindValue(block->globOptData.symToValueMap, newLengthSym));
                     Value *const lengthValueCopy = CopyValue(lengthValue, lengthValue->GetValueNumber());
                     SetValue(&block->globOptData, lengthValueCopy, newLengthSym);
-                    lengthValueCopy->GetValueInfo()->SetSymStore(nullptr);
+                    this->SetSymStoreDirect(lengthValueCopy->GetValueInfo(), nullptr);
                 }
             }
             else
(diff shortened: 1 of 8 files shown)


ThomsonTan:
@LouisLaf , could you please take a look?

lib/Backend/GlobOpt.h
```diff
@@ -957,6 +957,10 @@ class GlobOptBlockData
         argOutCount(0),
         totalOutParamCount(0),
         callSequence(nullptr),
+        capturedValuesCandicate(nullptr),
```
rajatd: typo: should be capturedValuesCandidate

lib/Backend/GlobOpt.h
```diff
@@ -1007,6 +1011,11 @@ class GlobOptBlockData
     SListBase<IR::Opnd *> *                 callSequence;
     StackLiteralInitFldDataMap *            stackLiteralInitFldDataMap;
 
+    CapturedValues *                        capturedValuesCandicate;
```
rajatd: typo: should be capturedValuesCandidate

ThomsonTan: Right, will fix this one.

---
In reply to: [67772925](https://github.com/Microsoft/ChakraCore/pull/1152#discussion_r67772925) [](ancestors = 67772925)

lib/Backend/GlobOptFields.cpp
```diff
@@ -2934,6 +2933,7 @@ GlobOpt::SetObjectTypeFromTypeSym(StackSym *typeSym, const Js::Type *type, Js::E
     {
         JsTypeValueInfo* valueInfo = JsTypeValueInfo::New(this->alloc, type, typeSet);
         valueInfo->SetSymStore(typeSym);
+        this->SetSymStoreDirect(valueInfo, typeSym);
```
rajatd: why not replace SetSymStore with SetSymStoreDirect as you've done at other places?

ThomsonTan: Good catch, the old one should be removed.

---
In reply to: [67777712](https://github.com/Microsoft/ChakraCore/pull/1152#discussion_r67777712) [](ancestors = 67777712)

lib/Backend/GlobOpt.cpp
```diff
@@ -6935,20 +6960,34 @@ GlobOpt::SetSymStore(ValueInfo *valueInfo, Sym *sym)
...
     return sym;
 }
 
 void
+GlobOpt::SetSymStoreDirect(ValueInfo * valueInfo, Sym * sym)
+{
+    Sym * prevSymStore = valueInfo->GetSymStore();
+    if (prevSymStore && prevSymStore->IsStackSym() &&
+        prevSymStore->AsStackSym()->HasByteCodeRegSlot() &&
+        this->blockData.changedSyms)
```
satheeshravi: Should this be an assert instead of a pre-condition ?

ThomsonTan: The thought is SetSymStore doesn't depend on whether changedSyms is available or not, so it just checks here. I will consider change it to assert or not in later change.

---
In reply to: [68113107](https://github.com/Microsoft/ChakraCore/pull/1152#discussion_r68113107) [](ancestors = 68113107)

satheeshravi: Sure. I was suggesting to assert inside the if condition and not outside.

---
In reply to: [68113857](https://github.com/Microsoft/ChakraCore/pull/1152#discussion_r68113857) [](ancestors = 68113857,68113107)

lib/Backend/GlobOpt.cpp
```diff
@@ -1602,6 +1609,12 @@ GlobOpt::ReuseBlockData(GlobOptBlockData *toData, GlobOptBlockData *fromData)
 
     toData->stackLiteralInitFldDataMap = fromData->stackLiteralInitFldDataMap;
 
+    // toData->capturedValues = fromData->capturedValues;
```
satheeshravi: nit

lib/Backend/GlobOpt.cpp
```diff
@@ -6935,20 +6959,34 @@ GlobOpt::SetSymStore(ValueInfo *valueInfo, Sym *sym)
...
     {
-        valueInfo->SetSymStore(sym);
+        SetSymStoreDirect(valueInfo, sym);
     }
 
     return sym;
 }
 
 void
+GlobOpt::SetSymStoreDirect(ValueInfo * valueInfo, Sym * sym)
```
LouisLaf: Why Direct instead of just SetSymStore?  The direct part made me believe at first that this version didn't set the changed bit.

ThomsonTan: Direct here is no type check in the original SetSymStore, but all sym store need check changed bit. Cannot overload SetSymStore since both have identical input parameter types.

---
In reply to: [68327056](https://github.com/Microsoft/ChakraCore/pull/1152#discussion_r68327056) [](ancestors = 68327056)

lib/Backend/GlobOpt.cpp
```diff
@@ -1060,6 +1060,9 @@ GlobOpt::MergePredBlocksValueMaps(BasicBlock *block)
         this->KillAllObjectTypes();
     }
 
+    this->blockData.capturedArgs = JitAnew(this->alloc, BVSparse<JitArenaAllocator>, this->alloc);
+    this->blockData.changedSyms = JitAnew(this->alloc, BVSparse<JitArenaAllocator>, this->alloc);
```
LouisLaf: This should go in InitBlockData().  The Clone and Reuse path will have valid BVs and you are allocating new ones unnecessarily.

lib/Backend/GlobOpt.cpp
LouisLaf: Oh I get it.  I miss the fact there already was a SetSymStore on GLobOpt.

---
In reply to: [68327517](https://github.com/Microsoft/ChakraCore/pull/1152#discussion_r68327517) [](ancestors = 68327517,68327056)

lib/Backend/GlobOpt.cpp
LouisLaf: I don't see a path where changedSyms would be NULL.  You should remove the NULL checks.

---
In reply to: [68132431](https://github.com/Microsoft/ChakraCore/pull/1152#discussion_r68132431) [](ancestors = 68132431,68113857,68113107)

lib/Backend/GlobOpt.cpp
```diff
@@ -6957,6 +6995,10 @@ GlobOpt::SetValue(GlobOptBlockData *blockData, Value *val, Sym * sym)
     else
     {
         SetValueToHashTable(blockData->symToValueMap, val, sym);
+        if (blockData->changedSyms && isStackSym && sym->AsStackSym()->HasByteCodeRegSlot())
```
LouisLaf: Here too.

lib/Backend/GlobOpt.cpp
ThomsonTan: Yes, reuse is good. Will take this change in the cross block version.

---
In reply to: [68328456](https://github.com/Microsoft/ChakraCore/pull/1152#discussion_r68328456) [](ancestors = 68328456)

lib/Backend/GlobOpt.cpp
ThomsonTan: Will try remove it when propagating values across blocks, it should be fine if its construction is moved to InitBlockData.

---
In reply to: [68328835](https://github.com/Microsoft/ChakraCore/pull/1152#discussion_r68328835) [](ancestors = 68328835,68132431,68113857,68113107)

lib/Backend/GlobOpt.cpp
ThomsonTan: Will do it.

---
In reply to: [68328865](https://github.com/Microsoft/ChakraCore/pull/1152#discussion_r68328865) [](ancestors = 68328865)

lib/Backend/GlobOptBailOut.cpp
```diff
@@ -15,52 +15,258 @@ GlobOpt::CaptureValue(BasicBlock *block, StackSym * stackSym, Value * value, Bai
...
+            Sym * sym = hasConstValue ? iterConst.Data().Key() : nullptr;
+            Value * val = nullptr;
+            HashBucket<Sym *, Value *> * symIdBucket = nullptr;
+            bool symIdHandled = false;
+
+            // copy unchanged sym to new capturedValues
+            while (sym && sym->m_id < symId)
+            {
+                Assert(sym->IsStackSym());
+                if (!sym->AsStackSym()->HasArgSlotNum() || block->globOptData.capturedArgs->Test(sym->m_id))
```
LouisLaf: This should always be false, otherwise the sym would be in the changedSym bit vector.

lib/Backend/GlobOptBailOut.cpp
```diff
@@ -15,52 +15,258 @@ GlobOpt::CaptureValue(BasicBlock *block, StackSym * stackSym, Value * value, Bai
...
+                    continue;
+                }
+
+                sym = symIdBucket->value;
+                Assert(sym->IsStackSym() && (sym->AsStackSym()->HasByteCodeRegSlot() || sym->AsStackSym()->HasArgSlotNum()));
+
+                val =  block->globOptData.capturedArgs->Test(symId) ? FindValue(sym) : symIdBucket->element;
+                ValueInfo* valueInfo = val->GetValueInfo();
+
+                if (valueInfo->GetSymStore() != nullptr)
```
LouisLaf: Is this necessary?

lib/Backend/GlobOptBailOut.cpp
```diff
@@ -15,52 +15,258 @@ GlobOpt::CaptureValue(BasicBlock *block, StackSym * stackSym, Value * value, Bai
...
+                {
+                    int32 intConstValue;
+                    BailoutConstantValue constValue;
+
+                    if (valueInfo->TryGetIntConstantValue(&intConstValue))
+                    {
+                        constValue.InitIntConstValue(intConstValue);
+                        bailOutConstValuesIter.InsertNodeBefore(this->func->m_alloc, sym->AsStackSym(), constValue);
+                        symIdHandled = true;
```
LouisLaf: Get rid of symIdHandled and just put continue here.

lib/Backend/GlobOptBailOut.cpp
```diff
@@ -15,52 +15,258 @@ GlobOpt::CaptureValue(BasicBlock *block, StackSym * stackSym, Value * value, Bai
...
+                StackSym * copyPropSym = (sym != nullptr) ? iterCopyPropSym.Data().Value() : nullptr;
+                Sym * mergeSym = nullptr;
+                bool mergeSymId = false;
+
+                // copy unchanged copy prop sym
+                if (sym && !tempBv->Test(copyPropSym->m_id) && sym->m_id < symId)
+                {
+                    Assert(sym->IsStackSym());
+
+                    if (!sym->AsStackSym()->HasArgSlotNum() || block->globOptData.capturedArgs->Test(sym->m_id))
```
LouisLaf: Again this should always be false.

lib/Backend/GlobOptBailOut.cpp
```diff
@@ -15,52 +15,258 @@ GlobOpt::CaptureValue(BasicBlock *block, StackSym * stackSym, Value * value, Bai
...
+            sym = hasCopyPropSym ? iterCopyPropSym.Data().Key() : nullptr;
+
+            while (true)
+            {
+                StackSym * copyPropSym = (sym != nullptr) ? iterCopyPropSym.Data().Value() : nullptr;
+                Sym * mergeSym = nullptr;
+                bool mergeSymId = false;
+
+                // copy unchanged copy prop sym
+                if (sym && !tempBv->Test(copyPropSym->m_id) && sym->m_id < symId)
```
LouisLaf: Put the Bv test at the end to avoid it when sym->m_id is >= to symId.

lib/Backend/GlobOptBailOut.cpp
```diff
@@ -15,52 +15,258 @@ GlobOpt::CaptureValue(BasicBlock *block, StackSym * stackSym, Value * value, Bai
...
+            }
+
+            if (symIdHandled)
+            {
+                continue;
+            }
+
+            sym = hasCopyPropSym ? iterCopyPropSym.Data().Key() : nullptr;
+
+            while (true)
```
LouisLaf: This code is really hard to follow.  Why didn't you lay it out like the constant case above:
1. loop over the unchanged syms
2. captureValue() on the symId
3. skip to next if sym->m_id == symId

That while(true) should only loop over the unchanged syms.  Put the rest of the code out of that loop.

lib/Backend/GlobOptFields.cpp
```diff
@@ -689,8 +689,7 @@ GlobOpt::PreparePrepassFieldHoisting(Loop * loop)
 
                 // Set object type live on prepass so we can track if it got killed in the loop. (see FinishOptHoistedPropOps)
                 JsTypeValueInfo* typeValueInfo = JsTypeValueInfo::New(this->alloc, nullptr, nullptr);
-                typeValueInfo->SetSymStore(typeSym);
-                typeValueInfo->SetIsShared();
```
LouisLaf: Why did you get rid of SetIsShared()?

lib/Backend/GlobOptBailOut.cpp
ThomsonTan: Yes, this is useless code.

---
In reply to: [68345673](https://github.com/Microsoft/ChakraCore/pull/1152#discussion_r68345673) [](ancestors = 68345673)

lib/Backend/GlobOptBailOut.cpp
ThomsonTan: This check is inherited from the previous version.

---
In reply to: [68345902](https://github.com/Microsoft/ChakraCore/pull/1152#discussion_r68345902) [](ancestors = 68345902)

lib/Backend/GlobOptBailOut.cpp
ThomsonTan: Will remove it.

---
In reply to: [68346060](https://github.com/Microsoft/ChakraCore/pull/1152#discussion_r68346060) [](ancestors = 68346060)

lib/Backend/GlobOptBailOut.cpp
ThomsonTan: Will update the order, my previous thought is put condition with more selectivity (not almost always true) in front.

---
In reply to: [68346499](https://github.com/Microsoft/ChakraCore/pull/1152#discussion_r68346499) [](ancestors = 68346499)

LouisLaf: The Bv test is much more expensive.

---
In reply to: [68348778](https://github.com/Microsoft/ChakraCore/pull/1152#discussion_r68348778) [](ancestors = 68348778,68346499)

lib/Backend/GlobOptBailOut.cpp
ThomsonTan: Will remove this.

---
In reply to: [68346272](https://github.com/Microsoft/ChakraCore/pull/1152#discussion_r68346272) [](ancestors = 68346272)

lib/Backend/GlobOptBailOut.cpp
ThomsonTan: The structure is similar, but this loop need to handle more things because the change in copy sym will also trigger recapture path before reaching changed symId.

---
In reply to: [68347583](https://github.com/Microsoft/ChakraCore/pull/1152#discussion_r68347583) [](ancestors = 68347583)

lib/Backend/GlobOptFields.cpp
ThomsonTan: Oh, this is not intended, will add it back. Thanks for the catch.

---
In reply to: [68347957](https://github.com/Microsoft/ChakraCore/pull/1152#discussion_r68347957) [](ancestors = 68347957)

lib/Backend/GlobOptBailOut.cpp
LouisLaf: If copy-sym changed but the sym itself didn't, then all you have to do is not add sym to the CopySyms list in the loop copying the unchanged syms.  

---
In reply to: [68350810](https://github.com/Microsoft/ChakraCore/pull/1152#discussion_r68350810) [](ancestors = 68350810,68347583)

lib/Backend/GlobOpt.cpp
ThomsonTan: InitDataBlockData is called when there is no predecessor. Still need to create new BV when have predecessor so InitDataBlockData doesn't meet the requirement.

---
In reply to: [68328885](https://github.com/Microsoft/ChakraCore/pull/1152#discussion_r68328885) [](ancestors = 68328885,68328456)

LouisLaf: Oh, I though you already had code in CloneBlockData but it is in CopyBlockData.
What I mean is that you should put this in InitBlockData/CloneBlockData/ReuseBlockData like all the other data structures.  This will allow better reuse.

---
In reply to: [68686479](https://github.com/Microsoft/ChakraCore/pull/1152#discussion_r68686479) [](ancestors = 68686479,68328885,68328456)

lib/Backend/GlobOptBailOut.cpp
```diff
@@ -5,62 +5,247 @@
...
+
+    CapturedValues capturedValues;
+    SListBase<ConstantStackSymValue>::EditingIterator bailOutConstValuesIter(&capturedValues.constantValues);
+    SListBase<CopyPropSyms>::EditingIterator bailOutCopySymsIter(&capturedValues.copyPropSyms);
+
+    bailOutConstValuesIter.Next();
+    bailOutCopySymsIter.Next();
+
+    // tempBv has both changed syms with bytecode reg slot or arg slot
+    BVSparse<JitArenaAllocator> * tempBv = JitAnew(this->tempAlloc, BVSparse<JitArenaAllocator>, this->tempAlloc);
```
LouisLaf: Rename to something like changedSymsAndArgs.

lib/Backend/GlobOptBailOut.cpp
```diff
@@ -5,62 +5,247 @@
...
+
+    if (!block->globOptData.capturedValues)
     {
```
LouisLaf: Factor the code into:
if ()
{
    CaptureValueFull()
} else {
   foreach_bits() {
        ProcessCaptureVarConstValue()
        ProcessCaptureCopoySymValue()
    }
}

lib/Backend/GlobOptBailOut.cpp
```diff
@@ -5,62 +5,247 @@
...
+                    }
+                }
+
+                hasCopyPropSym = iterCopyPropSym.Next();
+                sym = hasCopyPropSym ? iterCopyPropSym.Data().Key() : nullptr;
+            }
+            if (sym && sym->m_id == symId)
+            {
+                hasCopyPropSym = iterCopyPropSym.Next();
+            }
```
LouisLaf: Moved to end of loop as it prepares the next iteration.

lib/Backend/GlobOptBailOut.cpp
```diff
@@ -5,62 +5,247 @@
...
+                }
+
+                hasConstValue = iterConst.Next();
+                sym = hasConstValue ? iterConst.Data().Key() : nullptr;
+            }
+            if (sym && sym->m_id == symId)
+            {
+                hasConstValue = iterConst.Next();
+            }
+            if (symId != Js::Constants::InvalidSymID)
```
LouisLaf: Moved to end of loop as it prepares the next iteration.

lib/Backend/GlobOptBailOut.cpp
```diff
@@ -5,62 +5,247 @@
...
+            {
+                hasCopyPropSym = iterCopyPropSym.Next();
+            }
+            if (symId != Js::Constants::InvalidSymID)
+            {
+                // recapture changed copy prop sym
+                symIdBucket = block->globOptData.symToValueMap->GetBucket(symId);
+                if (symIdBucket != nullptr)
+                {
+                    sym = symIdBucket->value;
```
LouisLaf: Rename as this 'sym' is not at all the same 'sym' as before.

LouisLaf:
:shipit:

Issue #1064, opened 2016-05-31
Expose JSRT Apis to control ES6 experimental features


kunalspathak:
Currently, [JsRuntimeAttributeEnableExperimentalFeatures](https://github.com/Microsoft/ChakraCore/blob/release/1.2/lib/Jsrt/ChakraCommon.h#L311) flag is the only way to turn ON all the ES6 experimental features. Instead have a JSRT API / enum to expose more flags that user can use to control which experimental features should be turned ON/OFF (similar to `--harmony-*` that v8 has).

kunalspathak:
@liminzhu - FYI

ianwjhalliday:
Such an API would require us to forever support feature flags.  Something to carefully consider since this sounds like a fine-grained doc modes feature and could potentially burden us at some point in the future.

Are there clear merits to having an API with such fine grained control?

jianchun:
What if we accept string based hints? Discard unrecognized ones. Less robust but possible to maintain.

liminzhu:
Interesting idea @kunalspathak. I'd like to add on @ianwjhalliday 's comment though and ask about the merits of having the ability to handpick experimental features to turn on. Seems to me that in most cases you would want none/all of the experimental features on.

kunalspathak:
@ianwjhalliday , I agree that this will be similar to fine-grained doc modes, but in scenarios that I recently faced, we need to have some handling to turn off specific experimental ES6 feature based on behavior of host. I have often seen node scenarios expose ES6 features like `Proxy`, `SIMD` on `global` object only if specific flags are passed to it. 

We can have an ability using some kind of string based hints that @jianchun  pointed which will let host control the features exposed by engine.

liminzhu:
Looks like we are approaching the node scenario with compile flags rather than an JSRT API. Do we still need an API for this?

kunalspathak:
@liminzhu , don't think we need it for now.

liminzhu:
Ok @kunalspathak. I'm going to close the issue and we can re-open this if one day we see a use case.

PR #1154, opened 2016-06-18
Disable ES6 experimental features through compile flag

agarwal-sandeep:
JsRuntimeAttributeEnableExperimentalFeatures attribute to JsCreateRuntime
enables all experimental ES6 features together. Selective features can be
disabled using compile time options. When a feature is disabled config
takes precedence and it can be enabled.

The scenario is to disable certain feature when chakracore is compiled
with node but still being able to use test/chk build of
Microsoft/ChakraCore to work with node-chakracore for testing.


agarwal-sandeep:
@curtisman @kunalspathak please take a look. This is to resolve issue https://github.com/Microsoft/ChakraCore/issues/1064

kunalspathak:
LGTM.

PR #1155, opened 2016-06-20
xplat: enable Runtime/Base/Entropy.cpp

xanlpz:
Only needs an implementation of GetProcessIoCounters.

### DIFF
--- a/pal/src/thread/process.cpp
+++ b/pal/src/thread/process.cpp
@@ -1276,6 +1276,46 @@ TerminateProcess(
 
 /*++
 Function:
+  GetProcessIoCounters
+
+Note:
+  hProcess is a handle on the current process.
+
+See MSDN doc.
+--*/
+
+BOOL
+PALAPI
+GetProcessIoCounters(
+    IN  HANDLE       hProcess,
+    OUT PIO_COUNTERS lpIoCounters
+)
+{
+    BOOL ret = FALSE;
+    struct rusage usage;
+
+    // xplat-todo: getrusage cannot query an arbitrary process, so for
+    // now we basically have to ignore the hProcess argument.
+    if (getrusage(RUSAGE_SELF, &usage) == 0)
+    {
+        ret = TRUE;
+
+        lpIoCounters->ReadOperationCount = usage.ru_inblock;
+        lpIoCounters->WriteOperationCount = usage.ru_oublock;
+        // xplat-todo: unclear how to figure out the 'Other'
+        // operations. The exact number of bytes written/read is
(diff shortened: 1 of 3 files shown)


xanlpz:
This is the best I think we can do with standard APIs. If we parse /proc/<id>/io we can fill in most data save for the 'Other' stuff. Also for now the #define for the feature is still turned off.

pal/src/thread/process.cpp
```diff
@@ -1276,6 +1276,42 @@ TerminateProcess(
...
+    BOOL ret = FALSE;
+    struct rusage usage;
+
+    // xplat-todo: getrusage cannot query an arbitrary process, so for
+    // now we basically have to ignore the hProcess argument.
+    if (getrusage(RUSAGE_SELF, &usage) == 0)
+    {
+        ret = TRUE;
+
+        // xplat-todo: unclear how to figure out the 'Other'
```
jianchun: Should clear all other fields to avoid returning unknown value?

xanlpz:
Sure. I was thinking it would not really matter since we use it as a source of entropy, but we should pretend we are doing the API for real.

obastemur:
IMO, we should implement Entropy from scratch for xplat. 

- Using Current PR we may not enable "ENABLE_CUSTOM_ENTROPY" IMO, it doesn't do the task
- Parsing any stream also may not be a good idea because of the perf cost and other issues on the way.

xanlpz:
What would it mean to reimplement from scratch in this context? Perhaps we can add some extra code to compensate for the lack of data in the IO counters implementation. For example using some of the memory statistics available as a replacement for the number of bytes written/read.

obastemur:
> For example using some of the memory statistics available as a replacement for the number of bytes written/read.

@xanlpz can you please elaborate on this? i.e. exactly which source ?

I meant we need much more than two IO numbers to create our entropy. BTW; We have already /dev/random on linux available that is actually based on the entropy from several system devices.

xanlpz:
For example these two:

       ru_minflt
              The  number  of page faults serviced without any I/O activity; here I/O activity is avoided by “reclaiming” a page frame  from the list of pages awaiting reallocation.

       ru_majflt
              The number of page faults serviced that required I/O activity.

Or we could use /dev/random too (or exclusively), yep.

rmrewak:
> BTW; We have already /dev/random on linux available that is actually based on the entropy from several system devices.

> Or we could use /dev/random too (or exclusively), yep.

+1

xanlpz:
Created new PR here: https://github.com/Microsoft/ChakraCore/pull/1174

PR #1158, opened 2016-06-20
cross-platform: ChakraCore StaticLibrary

obastemur:
- Implements static library support on xplat
 - Fixes thread cleanup issue xplat - shared lib

### DIFF
--- a/lib/Jsrt/ChakraCommon.h
+++ b/lib/Jsrt/ChakraCommon.h
@@ -34,12 +34,6 @@
 typedef DWORD_PTR ChakraCookie;
 typedef BYTE* ChakraBytePtr;
 #else // Non-Windows VC++
-#include <stdint.h>     // Needed for uintptr_t
-#ifndef __cplusplus
-#include <stdbool.h>    // Needed fo bool
-#include <stdio.h>      // for NULL
-#define nullptr NULL
-#endif
 
 // SAL compat
 #define _Return_type_success_(x)
@@ -1023,6 +1017,12 @@ typedef unsigned char* ChakraBytePtr;
             _Out_opt_ unsigned int *nextIdleTick);
 
 #ifdef _WIN32
+#define CSTR_TYPE wchar_t
+#else // !_WIN32
+// non-Windows ChakraCore overrides wchar_t to char16_t
+#define CSTR_TYPE char16_t
+#endif // _WIN32
+
     /// <summary>
     ///     Parses a script and returns a function representing the script.
     /// </summary>
@@ -1040,9 +1040,9 @@ typedef unsigned char* ChakraBytePtr;
     /// </returns>
     CHAKRA_API
         JsParseScript(
-            _In_z_ const wchar_t *script,
+            _In_z_ const CSTR_TYPE *script,
             _In_ JsSourceContext sourceContext,
-            _In_z_ const wchar_t *sourceUrl,
+            _In_z_ const CSTR_TYPE *sourceUrl,
             _Out_ JsValueRef *result);
 
     /// <summary>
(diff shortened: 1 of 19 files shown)


lib/Jsrt/ChakraCommon.h
```diff
@@ -1040,9 +1040,9 @@ typedef unsigned char* ChakraBytePtr;
     /// </returns>
     CHAKRA_API
         JsParseScript(
-            _In_z_ const wchar_t *script,
+            _In_z_ const CSTR_TYPE *script,
```
jianchun: Not sure if this affects MSDN documentation.

Are these APIs Win32 only? If so keeping wchar_t looks fine? + @digitalinfinity

lib/Jsrt/ChakraDebug.h
```diff
@@ -538,7 +538,6 @@
             _In_ unsigned int objectHandle,
             _Out_ JsValueRef *handleObject);
 
-#ifdef _WIN32
```
jianchun: Why removing this? We may need to define a Utf8 version for xplat instead?

lib/Jsrt/ChakraCommon.h
digitalinfinity: Yes, those APIs should be Win32 only

obastemur: They are already wchar_t for Windows.
```
#ifdef _WIN32
#define CSTR_TYPE wchar_t
#else // !_WIN32
..
```

lib/Jsrt/ChakraDebug.h
obastemur: `ch` references this method whether it is used or not. Option 1) disable from CH for nonWindows. Option 2) assert for non-Windows.

lib/Jsrt/ChakraCommon.h
jianchun: Since they are win32 only, they should be #ifdef out for non-Win32, not typedef CSTR_TYPE and appearing cross-platform.

lib/Jsrt/ChakraDebug.h
jianchun: What would be non-Windows debugging story? Should `ch` use a Utf8 version? + @agarwal-sandeep

agarwal-sandeep: I had done the underlying work for JsrtDebuggerStackFrame::Evaluate to take char16* and length. Please create a utf8 version and have ch use that?

obastemur: For now just disabling it for nonWindows with a todo note on utf8 implementation.

lib/Jsrt/ChakraCommon.h
obastemur: If we have a set of Jsrt API we are no longer planning to support on xplat, IMO we should mark them separately. This PR does not intend to mark what we may or may not support on xplat.

For the CSTR_TYPE macro. See the comment on its definition. We should warn the user that we do not use same type cross platform. Either that particular method is supported or not.

lib/Jsrt/JsrtHelper.cpp
```diff
@@ -49,3 +49,121 @@ void JsrtCallbackState::ObjectBeforeCallectCallbackWrapper(JsObjectBeforeCollect
...
+    }
+
+    _NOINLINE bool InitializeProcess()
+    {
+#if !defined(_WIN32)
+        pthread_key_create(&s_threadLocalDummy, DISPOSE_CHAKRA_CORE_THREAD);
+#endif
+
+#if defined(CHAKRA_STATIC_LIBRARY)
+#ifndef _WIN32
```
digitalinfinity: A lot of this looks copy-pasted from AttachProcess in ChakraCoreDllFunc.cpp, yes? Can you add a comment in both places indicating that updating one should update an other, and a TODO comment to consolidate?

pal/src/init/pal.cpp
```diff
@@ -610,22 +616,16 @@ PAL_InitializeChakraCore(int argc, char** argv)
...
     {
         PAL_Enter(PAL_BoundaryTop);
+        pal_was_initialized = true;
         return ERROR_SUCCESS;
     }
 
-    // Now that the PAL is initialized it's safe to call the initialization methods for the code that used to
-    // be dynamically loaded libraries but is now statically linked into CoreCLR just like the PAL, i.e. the
-    // PAL RT and mscorwks.
-    if (!LOADInitializeChakraCoreModule())
```
digitalinfinity: @jianchun looks like this is what called DllMain before?

digitalinfinity:
:shipit:

obastemur:
@curtisman @jianchun @digitalinfinity Thanks for the reviews!

obastemur:

PR #1159, opened 2016-06-21
xplat: add 4 more JSRT Utf8 APIs implementation

jianchun:
Implement `JsPointerToStringUtf8`, `JsStringToPointerUtf8`,
`JsGetPropertyIdFromNameUtf8`, `JsParseScriptUtf8`.

Some APIs need further work, marked xplat-todos.

### DIFF
--- a/lib/Common/Codex/Utf8Helper.h
+++ b/lib/Common/Codex/Utf8Helper.h
@@ -14,9 +14,11 @@ namespace utf8
     /// The returned string is null terminated.
     ///
     template <class Allocator>
-    HRESULT WideStringToNarrow(LPCWSTR sourceString, LPSTR* destStringPtr)
+    HRESULT WideStringToNarrow(
+        LPCWSTR sourceString, size_t sourceCount,
+        LPSTR* destStringPtr, size_t* destCount)
     {
-        size_t cchSourceString = wcslen(sourceString);
+        size_t cchSourceString = sourceCount;
 
         if (cchSourceString >= MAXUINT32)
         {
@@ -37,9 +39,11 @@ namespace utf8
             return E_OUTOFMEMORY;
         }
 
-        utf8::EncodeIntoAndNullTerminate(destString, sourceString, (charcount_t) cchSourceString);
+        size_t cbEncoded = utf8::EncodeIntoAndNullTerminate(destString, sourceString, (charcount_t) cchSourceString);
+        Assert(cbEncoded <= cbDestString);
         static_assert(sizeof(utf8char_t) == sizeof(char), "Needs to be valid for cast");
         *destStringPtr = (char*)destString;
+        *destCount = cbEncoded;
         return S_OK;
     }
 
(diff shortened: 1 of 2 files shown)


lib/Common/Codex/Utf8Helper.h
```diff
@@ -69,8 +75,10 @@ namespace utf8
         }
 
         utf8::DecodeIntoAndNullTerminate(destString, (LPCUTF8) sourceString, cchDestString);
+        Assert(destString[cchDestString] == 0);
```
obastemur: This looks like redundant. `DecodeIntoAndNullTerminate` puts `0` there no matter what.

lib/Common/Codex/Utf8Helper.h
```diff
@@ -83,52 +91,82 @@ namespace utf8
...
 
     template <class Allocator>
     class NarrowWideStringConverter<Allocator, LPCSTR, LPWSTR>
     {
     public:
-        static HRESULT Convert(LPCSTR sourceString, LPWSTR* destStringPtr)
+        static size_t Length(LPCSTR src)
         {
-            return NarrowStringToWide<Allocator>(sourceString, destStringPtr);
+            return strlen(src);
```
obastemur: How about using `strnlen` here with INT_MAX for the ceiling?

lib/Common/Codex/Utf8Helper.h
```diff
@@ -83,52 +91,82 @@ namespace utf8
...
         }
     };
 
     template <class Allocator>
     class NarrowWideStringConverter<Allocator, LPCWSTR, LPSTR>
     {
     public:
-        static HRESULT Convert(LPCWSTR sourceString, LPSTR* destStringPtr)
+        static size_t Length(LPCWSTR src)
+        {
```
obastemur: Can we add a todo here? We use PAL_wcslen for measuring the length of char16_t (on xplat) and it has its own problems and I'm not sure the result would match to `wchar_t`/wcslen

lib/Jsrt/Jsrt.cpp
```diff
@@ -889,6 +901,35 @@ CHAKRA_API JsStringToPointer(_In_ JsValueRef stringValue, _Outptr_result_buffer_
...
+    size_t wstrLen;
+    JsErrorCode err = JsStringToPointer(stringValue, &wstr, &wstrLen);
+    if (err == JsNoError)
+    {
+        PARAM_NOT_NULL(stringPtr);
+        PARAM_NOT_NULL(stringLength);
+        *stringPtr = nullptr;
+        *stringLength = 0;
+
+        // xplat-todo: fix lifetime. caller allocate space, or caller free?
```
obastemur: As of now this leaks right? Can we say `JsStringToPointerUtf8Copy` and caller is responsible from freeing it? but that would be adding a new interface..

obastemur:
LGTM

lib/Common/Codex/Utf8Helper.h
jianchun: I am trying to do a minimum validation of result WCHAR string length because this returns the length now. `cchDestString` was calculated by another API. This confirms a bit if `DecodeIntoAndNullTerminate` did decode these many characters.

lib/Jsrt/Jsrt.cpp
jianchun: Yes this leaks unless caller `free *stringPtr`. This is new API and its signature or requirement/documentation can be changed. Need more work here.

lib/Common/Codex/Utf8Helper.h
jianchun: Don't know what todo you meant exactly. We have wcslen calls elsewhere. I'll add a comment -- Here it is used rarely; Typically we'd ask caller to provide input string length. This is a fallback.

jianchun:
@obastemur Thanks for review!

PR #1160, opened 2016-06-21
Merge Time-Travel Debugging

mrkmarron:
See PR #1080 for comments -- earlier PR was closed + this PR opened to ensure naming/history is clean.


dilijev:
Hi Mark, before merging this, could you use interactive rebase to loop some of your quick fixes and other small commits into larger commits. Please ensure that each commit in this PR will build and pass tests.

dilijev:
Better yet, unless there are distinct logical commits you want to have here, you could merge --squash this whole set of commits into a single commit on top of the latest master branch.

PR #1161, opened 2016-06-21
Update linux-ci to latest linux branch

obastemur:


dilijev:
@obastemur this is fine but since this branch is non-critical and nominally will track the `linux` branch, if there's no merge conflicts (and there shouldn't be) you can just fast-forward linux-ci and push it directly.

dilijev:
To be clear, what I mean is that since `linux-ci` should, at most points in time, just act as a shadow for `linux` and `linux-ci` is not a protected branch (intended for use in testing changes to the CI config), it should be okay to do the following:

```
git checkout -b linux-ci github/linux-ci
git merge --ff-only linux
git push github linux-ci -u -f
```

obastemur:
@dilijev indeed but this branch will (eventually) end up having couple of extra experimental commits anyway. I'm not sure if we want to force / reset etc. this branch.

dilijev:
@obastemur Once we merge the changes from this branch, linux and linux-ci will be in sync again, no need to force or reset.

obastemur:
@dilijev What's the point of experimenting here then? and what we are going to do with problematic commits in the history?

dilijev:
The goal of this PR, if I'm not mistaken, is to first make `linux` and `linux-ci` exactly match up before making any code changes.

I'd rather not do that by introducing a merge commit because it makes future work on this branch more confusing than it needs to be.

obastemur:
The goal of this PR is to make sure test-build will pass the tests ;)

dilijev:
You won't be able to see anything to effect until after that specific change is merged. All that has happened here is you're updating linux-ci to linux via a merge commit. Other work would presumably happen in a different PR.

obastemur:
@dilijev feel free to merge the commits then. What I need here is the final state for linux branch is available here. Thanks!

dilijev:
Pushed to update linux-ci just now.

PR #1163, opened 2016-06-21
CI: add test-build testing

obastemur:


netci.groovy
```diff
@@ -115,18 +115,18 @@ def CreateBuildTasks = { machine, configTag, buildExtra, testExtra, runCodeAnaly
 
 def CreateLinuxBuildTasks = { machine, configTag, linuxBranch, nonDefaultTaskSetup ->
     [true, false].each { isPR ->
-        ['debug', 'release'].each { buildType ->
+        ['debug', 'release', 'test-build'].each { buildType ->
```
dilijev: We previously called this configuration "test", but for Linux I think this makes sense since this simplifies the logic around selecting the buildFlag below.

Do we expect those flags to be stable into the future?

I meant to comment on this before but does it bother anyone else that the only build-flavor-related flag that has a `-build` suffix is `--test-build`? I understand that the disambiguation is important but considering that we've only got one flag to "run tests" wouldn't it make sense to call that "--run-tests" and keep the flags for build configurations simple?

cc @digitalinfinity @jianchun

dilijev:
LGTM!

netci.groovy
obastemur: Hmm. naming difference can be handled below though. I mean that part can be 'test' and if 'test' then use 'test-build' flag etc. IMO also better. Let me update the PR.

obastemur:
@dotnet-bot test ci please

obastemur:
@dilijev Thanks!

dilijev:
@obastemur no problem

PR #1165, opened 2016-06-21
Bug 7877542: Simplify EntryNormalize for IsNormalizedString case to avoid LiteralString not allowed assert.

dilijev:
pThis is a propertyString. We can't use the buffer to create the same string if the propertyString's buffer is from a PropertyRecord, which would result in am interior pointer.

Since we are just returning the same string in this case, we should just return pThis.

Fixes Bug [7877542](https://microsoft.visualstudio.com/defaultcollection/web/wi.aspx?pcguid=cb55739e-4afe-46a3-970f-1b49d8ee7564&id=7877542)


dilijev:
@curtisman CR please?

dilijev:

curtisman:
LGTM

PR #1167, opened 2016-06-21
Precise closure-capture of local variables in the presence of deferred nested functions. This is a good thing to do anyway (improves CQ and solves some odd memory leak cases) and is especially important for redeferral and it allows consistency of byte code regardless of deferral. (Byte code will not necessarily be identical, but it will be functionally the same.) The parser is now responsible for detecting and marking non-local references to user variables, and we no longer force closure-capture of all locals in the presence of a deferred child function.

pleath:
Precise closure-capture of local variables in the presence of deferred
nested functions. This is a good thing to do anyway (improves CQ and
solves some odd memory leak cases) and is especially important for
redeferral and it allows consistency of byte code regardless of
deferral. (Byte code will not necessarily be identical, but it will be
functionally the same.) The parser is now responsible for detecting and
marking non-local references to user variables, and we no longer force
closure-capture of all locals in the presence of a deferred child
function.


pleath:
@ianwjhalliday @aneeshdk

lib/Backend/amd64/LowererMDArch.cpp
```diff
@@ -414,7 +414,7 @@ LowererMDArch::LoadFuncExpression(IR::Instr *instrFuncExpr)
         paramOpnd = IR::SymOpnd::New(paramSym, TyMachReg, this->m_func);
     }
 
-    if (this->m_func->GetJnFunction()->IsGenerator())
+    if (instrFuncExpr->m_func->GetJnFunction()->IsGenerator())
```
ianwjhalliday: Is this a drive by bug fix?

pleath: Yes and no. It’s not directly related to the feature, but there are a few things that fix bugs the feature exposed in the UT’s, mainly by enabling more optimization.

--Paul

lib/Parser/Parse.cpp
```diff
@@ -1701,6 +1703,12 @@ void Parser::BindPidRefsInScope(IdentPtr pid, Symbol *sym, int blockId, uint max
             sym->SetIsModuleExportStorage(true);
         }
 
+        if (ref->GetFuncScopeId() != funcId && !sym->GetIsGlobal())
+        {
+            Assert(ref->GetFuncScopeId() > funcId);
```
ianwjhalliday: So this means that the ref here is coming from a nested function, correct?

I'm not following what this is always the case, versus a ref coming from an enclosing scope.  I will admit my understanding of the PidRef model is not good.

lib/Parser/Parse.cpp
```diff
@@ -4505,48 +4486,21 @@ ParseNodePtr Parser::ParseFncDecl(ushort flags, LPCOLESTR pNameHint, const bool
...
-        m_currentNodeDeferredFunc = pnodeFncSave;
-        if (m_currentNodeFunc && pnodeFnc->sxFnc.HasWithStmt())
-        {
-            GetCurrentFunctionNode()->sxFnc.SetHasWithStmt(true);
-        }
-    }
+    // Lambdas do not have "arguments" and instead capture their parent's
+    // binding of "arguments.  To ensure the arguments object of the enclosing
+    // non-lambda function is loaded propagate the UsesArguments flag up to
+    // the parent function
```
ianwjhalliday: Why did this comment move up here?  It doesn't mention eval at all, and this code here isn't specific to lambdas

lib/Parser/Parse.cpp
```diff
@@ -4713,15 +4671,72 @@ bool Parser::ParseFncDeclHelper(ParseNodePtr pnodeFnc, ParseNodePtr pnodeFncPare
...
+        {
+            AnalysisAssert(pnodeFnc);
+
+            if (!fLambda)
+            {
+                m_currentNodeNonLambdaDeferredFunc = pnodeFnc;
+            }
+
+            m_currentNodeDeferredFunc = pnodeFnc;
+        }
```
ianwjhalliday: This code looks the same as above, could be refactored into a helper function

lib/Parser/Parse.cpp
```diff
@@ -5243,6 +5258,36 @@ bool Parser::ParseFncDeclHelper(ParseNodePtr pnodeFnc, ParseNodePtr pnodeFncPare
...
+        if (!fLambda)
+        {
+            Assert(pnodeFnc == m_currentNodeNonLambdaDeferredFunc);
+            m_currentNodeNonLambdaDeferredFunc = pnodeFncSaveNonLambda;
+        }
+        m_currentNodeDeferredFunc = pnodeFncSave;
+        if (m_currentNodeFunc && pnodeFnc->sxFnc.HasWithStmt())
+        {
+            GetCurrentFunctionNode()->sxFnc.SetHasWithStmt(true);
+        }
```
ianwjhalliday: Huh, why is this only propagated in the deferred parse case?

Also seems out of place here, as opposed to to where we propagate callseval and hasarguments flags above, at the call site to ParseFncDeclHelper

lib/Runtime/ByteCode/ByteCodeEmitter.cpp
```diff
@@ -2536,7 +2541,7 @@ void ByteCodeGenerator::GetEnclosingNonLambdaScope(FuncInfo *funcInfo, Scope * &
         {
             envIndex++;
         }
-        if (((scope == scope->GetFunc()->GetBodyScope() || scope == scope->GetFunc()->GetParamScope()) && !scope->GetFunc()->IsLambda()) || scope->IsGlobalEvalBlockScope())
+        if ((((scope == scope->GetFunc()->GetBodyScope() && (!scope->GetFunc()->GetParamScope() || scope->GetFunc()->GetParamScope()->GetCanMergeWithBodyScope())) || scope == scope->GetFunc()->GetParamScope()) && !scope->GetFunc()->IsLambda()) || scope->IsGlobalEvalBlockScope())
```
ianwjhalliday: Is this an unrelated bug fix or was this broken by the precise closure-capture somehow?

lib/Runtime/ByteCode/ByteCodeGenerator.cpp
```diff
@@ -2695,6 +2691,12 @@ FuncInfo* PostVisitFunction(ParseNode* pnode, ByteCodeGenerator* byteCodeGenerat
         return top;
     }
 
+    if (top->paramScope && top->paramScope->GetCanMergeWithBodyScope())
+    {
+        Scope::MergeParamAndBodyScopes(pnode);
+        Scope::RemoveParamScope(pnode);
+    }
```
ianwjhalliday: Is this a change to how we handle the param scope that was necessary for precise capture?

I'm concerned that this may have bugs and if it is not specific to precise capture, would be better to have in a separate change.

lib/Backend/amd64/LowererMDArch.cpp
ianwjhalliday: Gotcha

ianwjhalliday:
Awesome to see our closure capture becoming precise.

Mostly I had questions rather than comments. Overall looks good to me.

ianwjhalliday:
One of @Cellule @MikeHolman should take a quick look at the asmjs change.

MikeHolman:
asm.js change lgtm

lib/Parser/Parse.cpp
```diff
@@ -1679,6 +1680,7 @@ void Parser::BindPidRefs(BlockInfoStack *blockInfo, uint maxBlockId)
 void Parser::BindPidRefsInScope(IdentPtr pid, Symbol *sym, int blockId, uint maxBlockId)
 {
     PidRefStack *ref, *nextRef, *lastRef = nullptr;
+    Js::LocalFunctionId funcId = GetCurrentFunctionNode()->sxFnc.functionId;
     Assert(sym);
 
     for (ref = pid->GetTopRef(); ref && ref->GetScopeId() >= blockId; ref = nextRef)
```
aneeshdk: Not related to change. Do we need both >= condition here and the equality check at the end of the loop body to break out of the loop?

lib/Parser/Parse.cpp
```diff
@@ -4693,8 +4647,12 @@ void Parser::AppendFunctionToScopeList(bool fDeclaration, ParseNodePtr pnodeFnc)
 Parse a function definition.
 ***************************************************************************/
 template<bool buildAST>
-bool Parser::ParseFncDeclHelper(ParseNodePtr pnodeFnc, ParseNodePtr pnodeFncParent, LPCOLESTR pNameHint, ushort flags, bool *pHasName, bool fUnaryOrParen, bool noStmtContext, bool *pNeedScanRCurly)
+bool Parser::ParseFncDeclHelper(ParseNodePtr pnodeFnc, LPCOLESTR pNameHint, ushort flags, bool *pHasName, bool fUnaryOrParen, bool noStmtContext, bool *pNeedScanRCurly)
 {
+    ParseNodePtr pnodeFncSave = buildAST ? m_currentNodeFunc : m_currentNodeDeferredFunc;
```
aneeshdk: Will calling GetCurrentFunctionNode() work?

aneeshdk: Similarly GetCurrentNonLamdaFunctionNode() below?

---
In reply to: [68703551](https://github.com/Microsoft/ChakraCore/pull/1167#discussion_r68703551) [](ancestors = 68703551)

lib/Parser/Parse.cpp
pleath: Yes, that's what it means. The parser is now tracking func ID's and scope ID's separately for each name reference. A func ID != the declaration's func ID indicates a reference in a nested function. (And since a function's ID is < all its child ID's, we can assert that the reference ID is > the declaration's ID.)

lib/Parser/Parse.cpp
pleath: True. I put the comment in the right place.

lib/Parser/Parse.cpp
pleath: True. It was a little more painful to do that the way I originally structured the change, but it's easy now, so I'll do it.

lib/Parser/Parse.cpp
pleath: This is (supposed to be) the code from Parser::ParseFncDecl, moved in here because we now update the current function inside ParseFncDeclHelper.

lib/Runtime/ByteCode/ByteCodeEmitter.cpp
pleath: It didn't matter before, because references did get bound properly, but we weren't retrieving the right scope. It does matter now, though.

lib/Runtime/ByteCode/ByteCodeGenerator.cpp
pleath: This is necessary for the change, yes. We need to do the merge earlier so we can propagate the has-local-in-closure property to the body scope before we process it.

lib/Runtime/ByteCode/ByteCodeEmitter.cpp
```diff
@@ -3858,7 +3860,7 @@ void ByteCodeGenerator::StartEmitFunction(ParseNode *pnodeFnc)
                 // outer function's arguments directly.
                 sym = funcInfo->GetArgumentsSymbol();
                 Assert(sym);
-                if (sym->GetHasNonLocalReference() || autoCapturesAllScope.OldCapturesAll())
+                if (sym->NeedsSlotAlloc(funcInfo))
                 {
                     sym->EnsureScopeSlot(funcInfo);
```
aneeshdk: EnsureScopeSlot() also does a NeedsSlotAlloc check.

aneeshdk:
>                 top->SetFuncExprNameReference(true);

Shouldn't this be already true?

---
Refers to: lib/Runtime/ByteCode/ByteCodeGenerator.cpp:2455 in db13fe8. [](commit_id = db13fe85a2a4226e5774ba57f86e893841f00c78, deletion_comment = False)

PR #1168, opened 2016-06-22
Added console.log to ch.exe as a convenient alias for WScript.Echo

dilijev:
This enables us to easily run examples written for the web browser that rely on console.log because they will now work in ch.exe without modification.

---

Short of merging this PR, here is a shim to work around this issue:

```js
if (typeof(console) === "undefined" &&
    typeof(WScript) != "undefined" &&
    typeof(WScript.Echo) != "undefined") {
    var console = {
        log: WScript.Echo
    }
}
```

If you include that in a test case (and you assume that you will never have a `console` object that doesn't have the `log` method defined, and you only use the `WScript.Echo` and `console.log` methods in mutually compatible ways) you can reuse the entire text of the test between browser hosts and the `ch.exe` host.

### DIFF
--- a/test/Debugger/JsDiagGetStackProperties.js.dbg.baseline
+++ b/test/Debugger/JsDiagGetStackProperties.js.dbg.baseline
@@ -3,6 +3,7 @@
     "this": "Object {...}",
     "{exception}": "number 1",
     "locals": {
+      "console": "Object {...}",
       "FuncLevel1": "function <large string>",
       "outerFunc1": "function <large string>",
       "globalVar": "undefined undefined"
@@ -12,6 +13,7 @@
     "this": "Object {...}",
     "{exception}": "Error Caught Error",
     "locals": {
+      "console": "Object {...}",
       "FuncLevel1": "function <large string>",
       "outerFunc1": "function <large string>",
       "globalVar": "undefined undefined"
@@ -20,6 +22,7 @@
   {
     "this": {
       "telemetryLog": "function <large string>",
+      "console": "Object {...}",
       "FuncLevel1": "function <large string>",
       "outerFunc1": "function <large string>",
       "globalVar": "Object {...}"
@@ -46,6 +49,7 @@
     "globals": {
       "WScript": "Object {...}",
       "print": "function print",
+      "console": "Object {...}",
       "FuncLevel1": "function <large string>",
       "outerFunc1": "function <large string>",
       "globalVar": "Object {...}"
@@ -54,6 +58,7 @@
   {
     "this": {
       "telemetryLog": "function <large string>",
+      "console": "Object {...}",
       "FuncLevel1": "function <large string>",
       "outerFunc1": "function <large string>",
       "globalVar": "Object {...}"
@@ -101,6 +106,7 @@
     "globals": {
       "WScript": "Object {...}",
       "print": "function print",
+      "console": "Object {...}",
       "FuncLevel1": "function <large string>",
       "outerFunc1": "function <large string>",
       "globalVar": "Object {...}"
--- a/bin/ch/WScriptJsrt.cpp
+++ b/bin/ch/WScriptJsrt.cpp
@@ -322,7 +322,7 @@ JsErrorCode WScriptJsrt::LoadModuleFromString(LPCSTR fileName, LPCSTR fileConten
     JsValueRef errorObject = JS_INVALID_REFERENCE;
 
     // ParseModuleSource is sync, while additional fetch & evaluation are async.
-    errorCode = ChakraRTInterface::JsParseModuleSource(requestModule, dwSourceCookie, (LPBYTE)fileContent, 
+    errorCode = ChakraRTInterface::JsParseModuleSource(requestModule, dwSourceCookie, (LPBYTE)fileContent,
         (unsigned int)strlen(fileContent), JsParseModuleSourceFlags_DataIsUTF8, &errorObject);
     if ((errorCode != JsNoError) && errorObject != JS_INVALID_REFERENCE)
     {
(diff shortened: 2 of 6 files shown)


akroshg:
nice to see that. lgtm

digitalinfinity:
+1 👍 :shipit:

obastemur:
It could be nice to have `console.error` under the same PR too.

dilijev:
@obastemur WScript doesn't already support something of that variety AFAICT. `console.log` is a big improvement and didn't take much effort to add so I'll stick to that for now.

See:

`for (a in WScript) console.log(a)`

```
Echo
Quit
LoadScriptFile
LoadModuleFile
LoadScript
LoadModule
SetTimeout
ClearTimeout
Attach
Detach
DumpFunctionPosition
RequestAsyncBreak
Edit
Arguments
```

obastemur:
> WScript doesn't already support an equivalent of console.log

This is another part I was also about to comment. This part also breaks the initial reason of this PR. It is also not that complicated to walk through args and print them IMO. Same goes for console.error. Only difference is use stderr not stdout.

I agree that adding `console.log/error` is a good idea. I'm just against to custom implementation of such a common / known interface.

dilijev:
@obastemur - I mistyped. We already had EchoCallback for WScript.Echo which is equivalent to `console.log`. I understand it wouldn't be hard but this was a quick quality of life fix and not a priority for me to make a complete implementation of the methods on `console` for `ch.exe`.

obastemur:
> We already had EchoCallback for WScript.Echo which is equivalent to console.log

Opps. Mixed with `print`. Just checked, good to go.

obastemur:
Hmm. print is also doing the same, somehow remembered it wasn't supporting additional args. We have too many alias for a basic logging task. Anyways.

dilijev:
*shrug* @Cellule added `print` because it is in webasm and it was convenient to have it. Unfortunately print in browsers (at least in Chrome) opens the print dialog. `console.log` seems like a more appropriate name for console logging in any case. I don't see a problem with having multiple convenience methods and retaining backwards compat via `WScript.Echo`.

dilijev:
Opened https://github.com/Microsoft/ChakraCore/issues/1175 to track possibly adding others. I'll merge this.

test/Debugger/JsDiagGetStackProperties.js.dbg.baseline
```diff
@@ -3,6 +3,7 @@
     "this": "Object {...}",
     "{exception}": "number 1",
     "locals": {
+      "console": "Object {...}",
```
ianwjhalliday: Are you adding this change to jshost too?  These tests will fail with jshost otherwise.

dilijev: Yep, that's why I haven't merged yet. There are some other failures I can't figure out, too.

ThomsonTan:
console.log supports substitution strings like console.log("%d", 1) which is not supported by WScript.Echo. It might cause some subtle issues if we select the name but implements different/partial functionality.

dilijev:
@ThomsonTan That's a good point, which is another reason I've been hesitant to finish up and merge this PR...

dilijev:
@ThomsonTan my thought on this is that we're defining a host-specific implementation of a logging API called `console.log` which in some circumstances (printing a single string) is mutually compatible with `WScript.Echo`. Again, it's just for convenience, so different in implementation is not that concerning. I'd welcome feedback on this logic before merging this.

bterlson:
Another option is to use [eshost](https://GitHub.com/bterlson/eshost-cli) and use print everywhere (including browsers) :-P

dilijev:
@bterlson Yep, that's a good option too (and thanks for the link, I was wondering how to set that up...)

ianwjhalliday:
Is there a strong use case for having console.log?  In my experience the occasional repro has console.log, but that's easily replaced with print as needed.

If there is uncertainty about merging this change I wonder if it is worth spending the time thinking about it.

dilijev:
@ianwjhalliday at this point, I agree. I don't want to spend more time on this. There are plenty of alternatives and any easy implementation of console.log feels incomplete. I'll close this now.

jianchun:
Having `console.log` really makes life easier to test snippets across `ch/node/browser`. 👍 +1 on this change.

ThomsonTan:
@jianchun , we'll need to consider its compatibility with other implementation if we want it to be used for Node. We probably could make a full implementation if it deserves.

jianchun:
@ThomsonTan I am not looking for any advanced features (99% of time that is irrelevant). Just need a common mechanism to output text, so that snippets can be tested across different hosts without any modification. That looks super useful to me. Moreover, migrating to `console.log` would make our code familiar to external developers.

ThomsonTan:
@jianchun I understand console.log is very useful for xplat, I wanted to have to it when running Octane, but printf style formatting would not be an advanced feature for common format specifier which exists in print for most programming languages. I'd like to vote for a full implementation.

dilijev:
@jianchun @ThomsonTan FWIW this implementation would not be used in Node or any other host. This is code specific to `ch.exe` which simply makes the `console.log` function point to `WScript.Echo`. In fact in that way it is basically identical to doing the shim I listed above:

```js
var console = {
    log: WScript.Echo
}
```

No checking required because within `ch.exe` we already know `console` object doesn't exist and `WScript.Echo` does.

dilijev:
One reason to merge this in would be that some people genuinely feel it would be helpful, even when we acknowledge that it would be an incomplete implementation. It doesn't harm anything to have the `console.log` API available.

dilijev:
(Reopened because of ongoing discussion.)

jianchun:
Yes, shim works, but mostly wasting our time. To compare our engine with others (or ch vs. Edge), tests written with "WScript.Echo/print" need to either ReplaceAll or shim "WScript.Echo/print" to "console.log". To try an external script written with "console.log", need to ReplaceAll or shim the other way.

@ThomsonTan We can consider that when needed.

dilijev:
@jianchun @ThomsonTan Added the task to add a complete implementation of `console.log` when/as needed at a later time, so we can move forward with the simple convenience of having `console.log(str)`

tcare:
I am happy to merge this and we can add the additional functionality if it is requested later. The typical use case is enough that we will get a lot of benefit out of merging this.

dilijev:
@tcare Agreed. I'll fix up the rest of what is needed internally and then merge this in.

dilijev:

PR #1169, opened 2016-06-22
Enabling eval in default param initializers

aneeshdk:
This change list is to enable eval in default params. The change will be
still under the experimental flag, not enabled by default. When eval
occurs in the param scope the cannot merge flag is set for the param
scope thus avoiding merging of param and body scope.

### DIFF
--- a/lib/Parser/Parse.cpp
+++ b/lib/Parser/Parse.cpp
@@ -5054,9 +5054,27 @@ bool Parser::ParseFncDeclHelper(ParseNodePtr pnodeFnc, ParseNodePtr pnodeFncPare
             {
                 OUTPUT_TRACE_DEBUGONLY(Js::ParsePhase, _u("The param and body scope of the function %s cannot be merged\n"), pnodeFnc->sxFnc.pnodeName ? pnodeFnc->sxFnc.pnodeName->sxVar.pid->Psz() : _u("Anonymous function"));
                 // Add a new symbol reference for each formal in the param scope to the body scope.
-                paramScope->ForEachSymbol([this](Symbol* param) {
+                // While inserting symbols into the symbol list we always insert at the front, so while traversing the list we will be visiting the last added
+                // formals first. Normal insertion of those into the body will reverse the order of symbols, which will eventually result in different order
+                // for scope slots allocation for the corresponding symbol in both param and body scope. Inserting them in the opposite order will help us
+                // have the same sequence for scope slots allocation in both scopes. This makes it easy to read the bytecode and may help in some optimization
+                // later.
+                paramScope->ForEachSymbol([this, pnodeFnc](Symbol* param) {
                     OUTPUT_TRACE_DEBUGONLY(Js::ParsePhase, _u("Creating a duplicate symbol for the parameter %s in the body scope\n"), param->GetPid()->Psz());
-                    ParseNodePtr paramNode = this->CreateVarDeclNode(param->GetPid(), STVariable, false, nullptr, false);
+
+                    ParseNodePtr paramNode = nullptr;
+                    if (this->m_ppnodeVar != &pnodeFnc->sxFnc.pnodeVars)
+                    {
+                        ParseNodePtr *const ppnodeVarSave = m_ppnodeVar;
+                        m_ppnodeVar = &pnodeFnc->sxFnc.pnodeVars;
+                        paramNode = this->CreateVarDeclNode(param->GetPid(), STVariable, false, nullptr, false);
+                        m_ppnodeVar = ppnodeVarSave;
+                    }
+                    else
+                    {
+                        paramNode = this->CreateVarDeclNode(param->GetPid(), STVariable, false, nullptr, false);
+                    }
+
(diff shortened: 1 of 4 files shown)


akroshg:
Cool to see that you are wrapping up this item quickly. I'll take a look at the change shortly.

aneeshdk:
@pleath @akroshg @tcare Can you please take a look?

lib/Parser/Parse.cpp
```diff
@@ -5054,9 +5054,23 @@ bool Parser::ParseFncDeclHelper(ParseNodePtr pnodeFnc, ParseNodePtr pnodeFncPare
             {
                 OUTPUT_TRACE_DEBUGONLY(Js::ParsePhase, _u("The param and body scope of the function %s cannot be merged\n"), pnodeFnc->sxFnc.pnodeName ? pnodeFnc->sxFnc.pnodeName->sxVar.pid->Psz() : _u("Anonymous function"));
                 // Add a new symbol reference for each formal in the param scope to the body scope.
-                paramScope->ForEachSymbol([this](Symbol* param) {
+                // Retain the same position of formals in the body variables too.
+                paramScope->ForEachSymbol([this, pnodeFnc](Symbol* param) {
```
akroshg: It would be nice if the ForEachSymbol provides vars in the order itself.

akroshg:
Looks good to me.

lib/Runtime/ByteCode/ByteCodeEmitter.cpp
```diff
@@ -3809,6 +3826,7 @@ void ByteCodeGenerator::StartEmitFunction(ParseNode *pnodeFnc)
                 !funcInfo->Escapes() &&
                 funcInfo->frameObjRegister != Js::Constants::NoRegister &&
                 !ApplyEnclosesArgs(pnodeFnc, this) &&
+                funcInfo->paramScope->GetCanMergeWithBodyScope() && // There is eval in the param scope
```
tcare: Do we want to explicitly check that this case triggers because there is an eval?

tcare: Actually, on second thought, that might be a bit hard because we have the notion of CallsEval but not at the scope level. Nevermind.

test/es6/default-splitscope.js
```diff
@@ -777,18 +785,54 @@ var tests = [
...
+            assert.areEqual(10, eval("a"), "In the body initial value of the symbol should be same as the final value from param scope");
+            a = 20;
+            assert.areEqual(20, eval("a"), "In the body after assignment the symbol value is updated");
+            assert.areEqual(10, b(), "Eval in the param scope captures the symbol from the param scope");
+        }
+        f1();
+
+        function f2(a = 10, b = () => eval("a")) {
+            a = 20;
+            assert.areEqual(10, b(), "Eval in the param scope captures the symbol from the param scope evan when there is no eval in the body");
```
tcare: even* (below too)

tcare:
Looks good

lib/Parser/Parse.cpp
```diff
@@ -5054,9 +5054,23 @@ bool Parser::ParseFncDeclHelper(ParseNodePtr pnodeFnc, ParseNodePtr pnodeFncPare
...
                     OUTPUT_TRACE_DEBUGONLY(Js::ParsePhase, _u("Creating a duplicate symbol for the parameter %s in the body scope\n"), param->GetPid()->Psz());
-                    ParseNodePtr paramNode = this->CreateVarDeclNode(param->GetPid(), STVariable, false, nullptr, false);
+
+                    ParseNodePtr paramNode = nullptr;
+                    if (this->m_ppnodeVar != &pnodeFnc->sxFnc.pnodeVars)
```
pleath: Can you describe the case that you're covering here? (And add a comment, I guess, since it's not immediately clear?)

lib/Runtime/ByteCode/ByteCodeEmitter.cpp
```diff
@@ -1822,22 +1830,24 @@ void ByteCodeGenerator::InitScopeSlotArray(FuncInfo * funcInfo)
 
         funcInfo->GetBodyScope()->ForEachSymbol(setPropIdsForScopeSlotArray);
 
-        if (funcInfo->thisScopeSlot != Js::Constants::NoRegister)
+        bool isSplitScope = funcInfo->paramScope && !funcInfo->paramScope->GetCanMergeWithBodyScope();
+
```
pleath: Can the if statements that follow this line be grouped so that each one doesn't have to check !isSplitScope?

lib/Parser/Parse.cpp
aneeshdk: While inserting the symbols to the symbol list we always insert at the front as don't have the tail pointer. Changing this behavior will require changing the symbol list implementation. I will skip that for now.

test/es6/default-splitscope.js
aneeshdk: Done

lib/Parser/Parse.cpp

lib/Runtime/ByteCode/ByteCodeEmitter.cpp

PR #1170, opened 2016-06-22
Refactor includes and fix unguarded path in record/replay.

mrkmarron:


lib/Runtime/Library/JavascriptLibrary.cpp
```diff
@@ -4707,10 +4707,13 @@ namespace Js
         else
         {
 #if ENABLE_TTD
-            //
-            //TODO: need to implement support for this path
-            //
-            AssertMsg(false, "Path not implemented in TTD!!!");
+            if(this->scriptContext->ShouldPerformRecordAction() | this->scriptContext->ShouldPerformDebugAction())
```
dilijev: Is that supposed to be `||` or is this masking a bit vector?

Also I like having a space between if and the open paren.

dilijev:
Otherwise LGTM

PR #1172, opened 2016-06-22
Added tags to unit TTD unit test definitions.

mrkmarron:


test/TTBasic/rlexe.xml
```diff
@@ -5,471 +5,471 @@
       <files>accessor.js</files>
       <compile-flags>-TTRecord:!accessorTest -TTSnapInterval:0</compile-flags>
       <baseline>accessorRecord.baseline</baseline>
-      <tags>exclude_dynapogo</tags>
+      <tags>exclude_dynapogo,exclude_jshost,exclude_snap,exclude_serialized</tags>
```
agarwal-sandeep: Instead of adding the tag to each test, add the tag to exclude the TTBasic and TTExecuteBasic directory in  test\rlexedirs.xml

satheeshravi: +1

---
In reply to: [68102924](https://github.com/Microsoft/ChakraCore/pull/1172#discussion_r68102924) [](ancestors = 68102924)

PR #1174, opened 2016-06-22
xplat: implement custom entropy for GNU/Linux

xanlpz:
We'll just use /dev/random since it's not really possible to implement
GetProcessIoCounters in a good way.

### DIFF
--- a/lib/Runtime/Base/Entropy.cpp
+++ b/lib/Runtime/Base/Entropy.cpp
@@ -4,6 +4,12 @@
 //-------------------------------------------------------------------------------------------------------
 #include "RuntimeBasePch.h"
 
+#ifdef __LINUX__
+#include <linux/random.h>
+#include <sys/syscall.h>
+#include <unistd.h>
+#endif
+
 const uint32 Entropy::kInitIterationCount = 3;
 
 void Entropy::BeginAdd()
@@ -54,6 +60,7 @@ void Entropy::Add(const char *buffer, size_t size)
 
 void Entropy::AddIoCounters()
 {
+#ifdef _WIN32
     IO_COUNTERS ioc = {0};
     if (GetProcessIoCounters(GetCurrentProcess(), &ioc))
     {
@@ -64,6 +71,21 @@ void Entropy::AddIoCounters()
         Add((char *)&ioc.WriteTransferCount,  sizeof(ioc.WriteTransferCount));
         Add((char *)&ioc.OtherTransferCount,  sizeof(ioc.OtherTransferCount));
     }
+#elif defined(__LINUX__)
+    // It's difficult to implement GetProcessIoCounters on Linux in a
+    // satisfactory way. We'll just use /dev/random to generate an
+    // equivalent amount of random bytes (6 times 64 bytes).
+#define ENTROPY_BYTES 64 * 6
+    int ret;
+    char buffer[ENTROPY_BYTES];
+    size_t size = sizeof(buffer);
+
+    ret = syscall(SYS_getrandom, buffer, size, 0);
+    if (ret == size)
+    {
+        Add(buffer, size);
+    }
+#endif // __LINUX__
 
     AddCurrentTime();
 }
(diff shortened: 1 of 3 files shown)


lib/Runtime/Base/Entropy.cpp
```diff
@@ -64,6 +71,21 @@ void Entropy::AddIoCounters()
...
+    // equivalent amount of random bytes (6 times 64 bytes).
+#define ENTROPY_BYTES 64 * 6
+    int ret;
+    char buffer[ENTROPY_BYTES];
+    size_t size = sizeof(buffer);
+
+    ret = syscall(SYS_getrandom, buffer, size, 0);
+    if (ret == size)
+    {
+        Add(buffer, size);
```
obastemur: What do you think if we can use the implementation under pal/src/misc/random.cpp? We might hold a bit bigger cache there though.

obastemur:
@xanlpz We have `dev/urandom` in place which is (AFAIK) less entropy juiced comparing to `dev/random`. It is okay to reuse the code we have there and fallback to `dev/random` (this is what our entropy for) when (if ever) `dev/urandom` fails. The only problem I can see that `dev/random` is a blocking device but it's a backup anyways.

So we keep `rand_s` as is and use `dev/random` for `ENABLE_CUSTOM_ENTROPY`, however it could be nice to have the `dev/random` related part is under `pal/src/misc/random.cpp`

Sounds good?

xanlpz:
It's possible to use /dev/random without blocking with the right flag, and I think we have no choice because we cannot block the whole thing until the system generates enough entropy right? I'm just reading that /dev/random is basically meant for strong cryptography, which I'd say is not our case (correct me if I'm wrong). So for our use case /dev/urandom should be enough.

I think the deeper issue here is simply that in Linux since we implement rand_s with that too we'd basically have no alternative fallback, but we should be good since /dev/urandom already generates pseudo-random numbers when it does not have enough entropy from its IO sources, etc. So I think my proposal here is to skip the custom entropy fallback and document our rationale. What do you think?

obastemur:
@xanlpz agreed. As I mentioned, I would not expect urandom is failing ever in a way that the current implementation expects

Either way works.

xanlpz:
Just to summarize where we are at:

Current code uses rand_s as first option. On Linux this means we use /dev/urandom. We:

* cannot use /dev/urandom as a fallback, because we'd use the same codepath for our first shot and the fallback.
*cannot use /dev/random because the default way of accessing it is blocking. It's also really meant for really strong cyptography, which is not our use case. My first patch in this PR says /dev/random but the patch is really using /dev/urandom, so just ignore that confusion.

As fallback we cannot really implement GetProcessIoCounters properly either, because Linux lacks the proper APIs to do so (no API for exact number of bytes written/read, for example). We could use a mixture of data from getrusage(), but it's not clear whether that's good enough or not.

obastemur:
We were discussion this on gitter. One more reason for not using dev/random etc. is that the attempt may fail for the same reason as it failed on GetRandom. We wouldn't be actually having a real fallback there.

It may not be a perfect solution though but just for discussion. How about `rand_r` for the fallback case on posix?

dilijev:
@xanlpz @obastemur any update on this PR?

xanlpz:
I think we really did not advance much further than my last summary comment, where I suggested we skip a fallback implementation on Linux and document why. But maybe I'm forgetting something. Should I send a patch doing that?

obastemur:
@xanlpz closing this one for now. Thanks for the effort!

dilijev:
@xanlpz Please feel free to reopen this PR against master with a revised implementation after the design is finalized.

obastemur:
@dilijev from your comment, I felt like we finally got us a solution here! :)

Future reference for anyone into solving this problem we have here;

Please read the discussion above. 

The problem we have is summarized below;

- Alternative approach uses the same `random` source, so not a good candidate to backup the original one
- Missing number of entropy sources or perf penalties
- We don't have an external API for embedder to provide a custom one [not yet discussed]

@xanlpz has a comment above for some more details.

Lets keep this part `not implemented` unless we have a solution for it.

Again @xanlpz thanks for the effort!

PR #1178, opened 2016-06-22
cross-platform: Fix global initialisation ordering issues on OSX

tadeuzagallo:
Neither `init_sec` nor `init_priority` work on Darwin (`init_priority` only works within the compilation unit), so Chakra was crashing due to out of order initialisation.

This diff uses construct on first use for `PAL` and a nifty counter for `AutoSystemInfo` to ensure the initialisation order.

I've also removed the calls to `PAL_InitializeChakraCore`. which I'm not very sure about, but they seemed redundant, since `Initialize` will already be called by `AutoSystemInfo`, and the second call crashing due to the assertion on `pal/src/loader/module.cpp`:

```
BOOL LOADInitializeModules()
{
    _ASSERTE(exe_module.prev == nullptr);
```

With these changes I managed to compile and execute Chakra on OSX, using the following build command:

```
./build.sh --icu=/usr/local/opt/icu4c/include --debug --jobs=4 --static
```

### DIFF
--- a/lib/Common/Core/SysInfo.cpp
+++ b/lib/Common/Core/SysInfo.cpp
@@ -29,6 +29,7 @@
 EXTERN_C IMAGE_DOS_HEADER __ImageBase;
 #endif
 
+#ifdef CHAKRA_STATIC_LIBRARY
 static int AutoSystemInfoInitializer_counter;
 static typename std::aligned_storage<sizeof(AutoSystemInfo), alignof(AutoSystemInfo)>::type AutoSystemInfoInitializer_buffer;
 AutoSystemInfo &AutoSystemInfo::Data = reinterpret_cast<AutoSystemInfo&>(AutoSystemInfoInitializer_buffer);
@@ -44,6 +45,9 @@ AutoSystemInfoInitializer::~AutoSystemInfoInitializer() {
     (&AutoSystemInfo::Data)->~AutoSystemInfo();
   }
 }
+#else
+AutoSystemInfo AutoSystemInfo::Data INIT_PRIORITY(300);
+#endif
 
 #if DBG
 bool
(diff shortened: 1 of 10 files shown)


lib/Common/Core/SysInfo.cpp
```diff
@@ -8,6 +8,7 @@
 #endif
 #include <wincrypt.h>
 #include <VersionHelpers.h>
+#include <type_traits>
```
obastemur: Fails for cross platform.

I guess the `AutoSystemInfoInitializer` is process wide. Instead of dealing with counter, can we use the same static initialization approach (which is also thread safe) ? 

- Is there a reason why don't we host this under PAL ? 

- Why we do limit this only for static library?

tadeuzagallo: >   Instead of dealing with counter, can we use the same static initialization approach (which is also thread safe) ?

You mean using construction on first use? If so, the only reason I didn't do that was because it's used everywhere in the codebase, and I was trying to keep the diff small.

> Is there a reason why don't we host this under PAL ?

You tell me, I'm happy to move it if you think it's better.

> Why we do limit this only for static library?

 For dynamic libraries, it fails to link, since `ch` includes the header, but doesn't link against the implementation.

obastemur:
@tadeuzagallo can you try signing CLA again?

tadeuzagallo:
Unfortunately I can't sign it myself, I pinged the FB team and they are looking into it. Feel free to close it if you fix it before they sort it out though.

obastemur:
@tadeuzagallo thanks for the effort. #1234 && #1235 are taking care of OSX. Closing this one. 

Hope to see this CLA situation is fixed and see you around.

tadeuzagallo:
Thanks for fixing this! I'll test your diffs.

And yeah, I'm still waiting on the CLA situation, hopefully I can come back to contributing soon 😊

PR #1182, opened 2016-06-23
Inlining in jitted loop bodies

rajatd:
This PR enables script function inlining in jitted loop bodies. 
- Small changes in the stack walker to support this scenario.
- Tweaks to inlining heuristics so that we don't inline as aggressively in loop bodies as we do in functions.

This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/1182)


rajatd:
@LouisLaf please review
FYI, @pleath @abchatra

LouisLaf:
:shipit:

PR #1183, opened 2016-06-23
X64 call arg MOV 0

Cellule:
When optimizing MOV 0, take into accounts registers used for arguments.
It was possible to reuse a register already used for an argument to do the `XOR reg, reg; MOV arg,reg` optimization.
The result would be to change the value of 1 argument to 0.

This bug is hitting is WebAssembly and probably in asm.js as well.
This can not hit with javascript user code, because it it specific to `MOV arg, 0` pattern and we never have const 0 (at least a tagged int 0 -> 1).
It could have been a problem with helper calls, but I guess we got lucky.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/1183)


rajatd:
LGTM.

LouisLaf:
:shipit:

PR #1187, opened 2016-06-23
Set Release and Pre-Release macros to 0 in master.

dilijev:
Set CHAKRA_CORE_VERSION_RELEASE, CHAKRA_CORE_VERSION_PRERELEASE, and CHAKRA_VERSION_RELEASE to 0.

We do not need to update the GUID because by making these changes we are switching from release configuration to engineering configuration by making these changes. See ByteCodeSerializer.cpp for details on this.

This will also enable the logic which allows the build number to be formatted into the x and y fields of the file version format 1.2.x.y.


dilijev:
@curtisman @digitalinfinity review please?

PR #1188, opened 2016-06-23
improve changePrototype scenario

kunalspathak:
Whenever we change prototype of an object, we invalidate all proto inline caches of all the properties present in all of the object's prototype chain. However this can be avoided if the object is not a prototype of any other object since it would have never got registered in proto inline cache as a prototype object. So, skip invalidation if the object is not a prototype of any object.
The only thing that should be done in that case is to change the type of object so that if it's properties were using proto inline caches, they will be invalidated by type mismatch and we will update their proto inline cache on next property access.

Gives approx. 20% gain in AcmeAir benchmarks of node.

Test: UnitTest pass, Internal test runs in progress.

### DIFF
--- a/lib/Runtime/Library/JavascriptObject.cpp
+++ b/lib/Runtime/Library/JavascriptObject.cpp
@@ -158,6 +158,7 @@ namespace Js
             CrossSite::ForceCrossSiteThunkOnPrototypeChain(newPrototype);
             return proxy->SetPrototypeTrap(newPrototype, shouldThrow);
         }
+        
         // 2.   Let extensible be the value of the [[Extensible]] internal data property of O.
         // 3.   Let current be the value of the [[Prototype]] internal data property of O.
         // 4.   If SameValue(V, current), then return true.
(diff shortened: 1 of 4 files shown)


kunalspathak:
@pleath , @curtisman , please review.

lib/Runtime/Library/JavascriptObject.cpp
```diff
@@ -205,45 +205,62 @@ namespace Js
...
-            // The HaveOnlyWritableDataProperties cache is cleared when a property is added or changed,
-            // but only for types in the same script context. Therefore, if the prototype is in another
-            // context, the object's cache won't be cleared when a property is added or changed on the prototype.
-            // Moreover, an object is added to the cache only when its whole prototype chain is in the same
-            // context.
-            //
-            // Since we don't have a way to find out which objects have a certain object as their prototype,
-            // we clear the cache here instead.
+            // Notify old prototypes that they are being removed from a prototype chain. This triggers invalidating protocache, etc.
+            if (!JavascriptProxy::Is(object))
```
pleath: Why should this not be done for proxies?

kunalspathak: This is the old code and I have just moved it under `if (isInvalidationOfInlineCacheNeeded) {}`. 
To answer your question, if this is proxy object, we call Trap and exit from this function in the [beginning](https://github.com/Microsoft/ChakraCore/pull/1188/files#diff-ea542b055c5ff44ee26e3c8131164bf2R155). I will remove this redundant if check for proxy.

pleath: Great. Thanks. LGTM, then.

PR #1189, opened 2016-06-23
Few iterator close issues - array destructuring.

akroshg:
Array destructuring was wrapped with try..finally as it was assumed that we will get abrupt completion due to throw. But 'yield' makes it possible to abrupt completion with 'return'. Fixed that by extend that to try.catch.finally.
Also we were emitting reference of the LHS after emitting of RHS. However that is supposed to happen first. Fixed that by Emit reference of LHS first.
Added unittest.

### DIFF
--- a/lib/Parser/Parse.cpp
+++ b/lib/Parser/Parse.cpp
@@ -12197,11 +12197,6 @@ ParseNodePtr Parser::ParseDestructuredArrayLiteral(tokens declarationType, bool 
     {
         while (true)
         {
-            if (seenRest) // Rest must be in the last position.
-            {
-                Error(ERRDestructRestLast);
-            }
-
             ParseNodePtr pnodeElem = ParseDestructuredVarDecl<buildAST>(declarationType, isDecl, &seenRest, topLevel);
             if (buildAST)
             {
@@ -12222,6 +12217,11 @@ ParseNodePtr Parser::ParseDestructuredArrayLiteral(tokens declarationType, bool 
             if (m_token.tk != tkComma)
             {
                 Error(ERRDestructNoOper);
+            }
+
+            if (seenRest) // Rest must be in the last position.
+            {
+                Error(ERRDestructRestLast);
             }
 
             m_pscan->Scan();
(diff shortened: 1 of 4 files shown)


akroshg:
@tcare @ianwjhalliday please have a look at this.

test/es6/iteratorclose.js
```diff
@@ -398,6 +398,100 @@ var tests = [
...
+            iter.next();
+            assert.throws(function () {
+                iter.return();
+            }, Error, "calling return on the outer generator will call return on inner generator", 'Exception from return function');
+		}
+	},
+    {
+		name : "Destructuring - yield in the pattern and both caller and inner return throws",
+		body : function () {
+            function *innerGen () { yield undefined;}
```
tcare: spacing

tcare:
I'm surprised the LHS was emitted first. I seem to remember there being a reason, perhaps because we used to do the same for events.

How did we find the yield case?

Review looks good.

akroshg:
yield was found from test262. Now 40 more tests are passing.

tcare:
Awesome!

lib/Parser/Parse.cpp
```diff
@@ -12224,6 +12219,11 @@ ParseNodePtr Parser::ParseDestructuredArrayLiteral(tokens declarationType, bool
                 Error(ERRDestructNoOper);
             }
 
+            if (seenRest) // Rest must be in the last position.
+            {
+                Error(ERRDestructRestLast);
+            }
+
```
ianwjhalliday: What was the issue that required this change?

akroshg: see the changes in the test\destructuring.js in this PR.

test/es6/iteratorclose.js
```diff
@@ -398,6 +398,100 @@ var tests = [
...
+                ([...{}[yield]] = innerGen());
+                assert.fail('Unreachable code');
+            }());
+
+            iter.next();
+            iter.return();
+            assert.areEqual(returnCalled, 1, "ensuring that return function is called");
+		}
+	},
+    {
```
ianwjhalliday: In all these tests the inner generator code either throws or is unreachable. How about testing the scenarios that do execute the inner generator without error to validate that return is not called in those case?

PR #1190, opened 2016-06-23
cross platform: fix release/static lib build

obastemur:
At the moment, Release build doesn't have the test hooks are enabled. Although we force define `ENABLE_TEST_HOOKS` from ch/stdafx.h, underlying shared lib doesn't include the tooling for release build. Following the same approach for static builds, fixes release/static library builds on xplat.

### DIFF
--- a/bin/ch/ch.cpp
+++ b/bin/ch/ch.cpp
@@ -521,7 +521,7 @@ int _cdecl wmain(int argc, __in_ecount(argc) LPWSTR argv[])
     HINSTANCE chakraLibrary = nullptr;
     bool success = ChakraRTInterface::LoadChakraDll(&argInfo, &chakraLibrary);
 
-#if defined(CHAKRA_STATIC_LIBRARY)
+#if defined(CHAKRA_STATIC_LIBRARY) && !defined(NDEBUG)
     // handle command line flags
     OnChakraCoreLoaded();
 #endif


bin/ch/ch.cpp
```diff
@@ -521,7 +521,7 @@ int _cdecl wmain(int argc, __in_ecount(argc) LPWSTR argv[])
     HINSTANCE chakraLibrary = nullptr;
     bool success = ChakraRTInterface::LoadChakraDll(&argInfo, &chakraLibrary);
 
-#if defined(CHAKRA_STATIC_LIBRARY)
+#if defined(CHAKRA_STATIC_LIBRARY) && !defined(NDEBUG)
```
jianchun: Prefer `&& ENABLE_DEBUG_CONFIG_OPTIONS`. That's used everywhere to represent DEBUG and TEST build types.

obastemur: > That's used everywhere to represent DEBUG and TEST build types.

That wouldn't work. It is also defined for Release build under `ch`. CH leaks `ENABLE_TEST_HOOKS` definition. As a result CommonDefines.h defines `ENABLE_DEBUG_CONFIG_OPTIONS` too.

NDEBUG is explicitly defined by common build props.

jianchun: You are right. LGTM.

digitalinfinity:
:shipit:

obastemur:
@jianchun @digitalinfinity Thanks for the review

PR #1191, opened 2016-06-23
Intl: removed invalid assert

kunalspathak:
`WindowsCreateStringReference` sets string as null if length of input string is 0.
Removed the assert where we were checking that output string should never be null.

Fixes: (7953593)[https://microsoft.visualstudio.com//defaultcollection/OS/_workItems?_a=edit&id=7953593&triage=true]

### DIFF
--- a/lib/Runtime/Base/WindowsGlobalizationAdapter.cpp
+++ b/lib/Runtime/Base/WindowsGlobalizationAdapter.cpp
@@ -385,7 +385,6 @@ namespace Js
 
         // Construct HSTRING of timeZoneId passed
         IfFailThrowHr(GetWindowsGlobalizationLibrary(scriptContext)->WindowsCreateStringReference(timeZoneId, static_cast<UINT32>(wcslen(timeZoneId)), &timeZoneHeader, &timeZone));
-        Assert(timeZone);
 
         // The warning is timeZone could be '0'. This is valid scenario and in that case, ChangeTimeZone() would
         // return error HR in which case we will throw.


kunalspathak:
@tcare , please review.

lib/Runtime/Base/WindowsGlobalizationAdapter.cpp
```diff
@@ -385,7 +385,6 @@ namespace Js
 
         // Construct HSTRING of timeZoneId passed
         IfFailThrowHr(GetWindowsGlobalizationLibrary(scriptContext)->WindowsCreateStringReference(timeZoneId, static_cast<UINT32>(wcslen(timeZoneId)), &timeZoneHeader, &timeZone));
-        Assert(timeZone);
```
akroshg: why not return here? Also in this file many places we were asserting on this HSTRING for being there, could we just remove them?

kunalspathak: Validity of timeZone value is decided by `windows.globalization.dll` and would like to stick to that. If this timeZone was invalid because of empty string or invalid value, `hr = timeZoneCalendar->ChangeTimeZone(timeZone);`  would return error hr in which case we will return.
However, I can go back and check other places where we might remove similar assert.

Yongqu: LGTM

kunalspathak: Alright, I went ahead and checked and seems all other asserts are valid i.e. if they are hit, there is a problem.

tcare:
Looks good

PR #1192, opened 2016-06-24
Fix WebAssembly tests

Cellule:
Made -forcenative to prejit asm.js and wasm so rl.exe's second pass will run with prejit. 

Fix bug in asmjs bytecode basic dumper.

Update misc.js test to take into account the NYI operators.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/1192)


Cellule:
@MikeHolman @commonlisp Can you review this please ?

MikeHolman:
I think some of the asm.js UTs use -maic:0. We should remove those cases in favor of prejit.

Cellule:
I removed most -maic:0 from asm.js tests. Which means that those test will now have a step with interpreter and 1 with forcenative.
I didn't remove the tests that had -nonative, because the first run of RL is not exactly -nonative (it actually uses -mic:1 -simplejit:1)
There could still be some more cleanup to be done, but it won't hurt to have some duplicate testing.

commonlisp:
LGTM

PR #1195, opened 2016-06-24
Enable ES6IsConcatSpreadable under experimental

suwc:
Enable ES6IsConcatSpreadable under experimental flag.
Optimize fastpath for cases where [@@isConcatSpreadable] is undefined by caching in ThreadContext.
Add cache invalidations when users add [@@isConcatSpreadable] properties.
Add unit tests to cover more extended usage scenarios.


suwc:
@Microsoft/chakra-es could you take a look pls?

lib/Runtime/Language/InlineCache.h
```diff
@@ -964,6 +964,70 @@ namespace Js
...
+    class IsConcatSpreadableCache
+    {
+        Type *type0, *type1;
+        int lastAccess;
+        BOOL result0, result1;
+
+    public:
+        IsConcatSpreadableCache()
+        {
+            type0 = nullptr;
```
akroshg: Initialize these as initialization list.

lib/Runtime/Language/JavascriptOperators.cpp
```diff
@@ -1357,17 +1357,50 @@ namespace Js
...
+        if (!PHASE_OFF1(IsConcatSpreadableCachePhase))
+        {
+            BOOL retVal = FALSE;
+            Type *instanceType = instance->GetType();
+            IsConcatSpreadableCache *isConcatSpreadableCache = scriptContext->GetThreadContext()->GetIsConcatSpreadableCache();
+
+            if (isConcatSpreadableCache->TryGetIsConcatSpreadable(instanceType, &retVal))
+            {
+#ifdef ENABLE_DEBUG_CONFIG_OPTIONS
+                if (Js::Configuration::Global.flags.Trace.IsEnabled(Phase::IsConcatSpreadableCachePhase))
```
akroshg: nit : change this to OUTPUT_TRACE, that macro checks for IsEnabled as well also you don't need to put #ifdef also.

lib/Runtime/Library/JavascriptProxy.cpp
```diff
@@ -321,6 +321,9 @@ namespace Js
                 return FALSE;
             JavascriptError::ThrowTypeError(scriptContext, JSERR_ErrorOnRevokedProxy, _u("get"));
         }
+
+        RecyclableObject *target = this->target;
```
akroshg: why?

lib/Runtime/Language/InlineCache.h
```diff
@@ -964,6 +964,70 @@ namespace Js
...
+            type1 = nullptr;
+            result0 = FALSE;
+            result1 = FALSE;
+            lastAccess = 0;
+        }
+
+        BOOL TryGetIsConcatSpreadable(Type *type, _Out_ BOOL *result)
```
akroshg: nit: change this to bool.

lib/Runtime/Language/JavascriptOperators.cpp
```diff
@@ -1357,17 +1357,50 @@ namespace Js
...
+                {
+                    Output::Print(_u("IsConcatSpreadableCache hit: %p\n"), instanceType);
+                }
+#endif
+                return retVal;
+            }
+
+            Var spreadable = JavascriptOperators::GetProperty(instance, PropertyIds::_symbolIsConcatSpreadable, scriptContext);
+
+            if (spreadable != scriptContext->GetLibrary()->GetUndefined())
```
pleath: Does the spec say that the value of the spreadable property should be converted to Boolean? If so, is it right to treat undefined as a special case?

lib/Runtime/Language/JavascriptOperators.cpp
```diff
@@ -1357,17 +1357,50 @@ namespace Js
...
+            }
+#endif
+            retVal = JavascriptOperators::IsArray(instance);
+            isConcatSpreadableCache->CacheIsConcatSpreadable(instanceType, retVal);
+            return retVal;
+        }
+
         Var spreadable = JavascriptOperators::GetProperty(instance, PropertyIds::_symbolIsConcatSpreadable, scriptContext);
+
         if (spreadable != scriptContext->GetLibrary()->GetUndefined())
```
pleath: I see the same logic I was questioning above is already there...

lib/Runtime/Library/JavascriptArray.cpp
```diff
@@ -3133,18 +3138,25 @@ namespace Js
...
                     idxDest = idxDest + pIntArray->length;
                 }
-                else
+                else if (JavascriptNativeFloatArray::Is(aItem))
                 {
                     JavascriptNativeFloatArray* pItemArray = JavascriptNativeFloatArray::FromVar(aItem);
                     converted = CopyNativeFloatArrayElements(pDestArray, idxDest, pItemArray);
                     idxDest = idxDest + pItemArray->length;
                 }
+                else
```
pleath: The caller (EntryConcat) confirms that we only enter this method if the source array contents are float or convertible to float.

lib/Runtime/Library/JavascriptObject.cpp
```diff
@@ -2123,6 +2123,8 @@ namespace Js
             scriptContext->GetLibrary()->SetArrayObjectHasUserDefinedSpecies(true);
         }
 
+        JavascriptLibrary::CheckAndInvalidateIsConcatSpreadableCache(propId, scriptContext);
```
pleath: Is it necessary to catch this case in this method if we're already catching it in the type handler code?

pleath:
I don't see the part of the change that clears the cache if there are side-effects of reading isConcatSpreadable. Is that coming later? Or did I miss it?

lib/Runtime/Library/JavascriptProxy.cpp
suwc: According to ECMA262 9.5.5, [[GetOwnProperty]](P) must assign proxy target to a local variable 'target' and do business with the same.


---
In reply to: [68428440](https://github.com/Microsoft/ChakraCore/pull/1195#discussion_r68428440) [](ancestors = 68428440)

lib/Runtime/Language/JavascriptOperators.cpp
suwc: Good catch. Will change in next rev.

---
In reply to: [68426232](https://github.com/Microsoft/ChakraCore/pull/1195#discussion_r68426232) [](ancestors = 68426232)

lib/Runtime/Language/InlineCache.h
suwc: Will change in next rev.

---
In reply to: [68425816](https://github.com/Microsoft/ChakraCore/pull/1195#discussion_r68425816) [](ancestors = 68425816)

lib/Runtime/Language/JavascriptOperators.cpp
suwc: Yes, the spec (22.1.3.1.1) says:

3.If spreadable is not undefined, return ToBoolean(spreadable).
4.Return ? IsArray(O). 


---
In reply to: [68430889](https://github.com/Microsoft/ChakraCore/pull/1195#discussion_r68430889) [](ancestors = 68430889)

lib/Runtime/Library/JavascriptArray.cpp
suwc: With -es6isconcatspreadable on, JavascriptOperators::IsConcatSpreadable() call may run user code that potentially alters the array type. Search "getter altering array type" in es6IsConcatSpreadable.js

---
In reply to: [68432695](https://github.com/Microsoft/ChakraCore/pull/1195#discussion_r68432695) [](ancestors = 68432695)

lib/Runtime/Language/JavascriptOperators.cpp
pleath: Thanks. That seems odd. It doesn’t distinguish between a non-existent property and a property whose value is “undefined”. But probably not worth raising as a spec issue.

--Paul

pleath:
Okay. LGTM.

PR #1197, opened 2016-06-24
Wasm br/br_table yield 

Cellule:
Added a landing pad for Wasm loops. Since loops don't loop by default, jump over the pad. Used instead of jumping directly to loop header. This was problematic when nested loops wanted to break to the loop head of an outer loop.

Acquire a tmp register when entering a block. When yielding to a block used the previously save register. This allows br and br_table from anywhere inside the block to yield a value before breaking.
If any path yields a value conflicting type with another path, change the yield type to void and do not use the value (Like the yield never happened).

Introduced the type `unreachable` to differentiate paths that are unreachable. Currently this is used like an `any` type to not change block yield type to void.
For instance `(block (block (br 1 (i32.const 0))`. The second block will have type `unreachable` and the first block will merge `i32` and `unreachable` => `i32`.

There is still some work to do with `unreachable` validate things like `i32.ctz (br 0)`. This is supposed to be valid, but shouldn't produce any bytecode.
See https://github.com/WebAssembly/design/issues/707 for details about this.

This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/1197)

### DIFF
--- a/lib/WasmReader/WasmByteCodeGenerator.cpp
+++ b/lib/WasmReader/WasmByteCodeGenerator.cpp
@@ -24,15 +24,14 @@ WasmBytecodeGenerator::WasmBytecodeGenerator(Js::ScriptContext * scriptContext, 
     m_f64RegSlots(nullptr),
     m_i32RegSlots(nullptr),
     m_currentFunc(nullptr),
-    m_evalStack(&m_alloc)
+    m_evalStack(&m_alloc),
+    m_blockInfos(&m_alloc)
 {
     m_writer.Create();
 
     // TODO (michhol): try to make this more accurate?
     const long astSize = 0;
     m_writer.InitData(&m_alloc, astSize);
-
-    m_labels = Anew(&m_alloc, SListCounted<Js::ByteCodeLabel>, &m_alloc);
 
     // Initialize maps needed by binary reader
     Binary::WasmBinaryReader::Init(scriptContext);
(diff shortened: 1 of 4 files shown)


Cellule:
@MikeHolman @commonlisp Review please

lib/WasmReader/WasmParseTree.h
```diff
@@ -15,8 +15,10 @@ namespace Wasm
             Void,
 #define WASM_LOCALTYPE(token, name) token,
 #include "WasmKeywords.h"
-            Limit
+            Limit,
+            Unreachable
```
MikeHolman: Seems like Limit should be after unreachable?

lib/WasmReader/WasmByteCodeGenerator.cpp
```diff
@@ -572,42 +566,38 @@ WasmBytecodeGenerator::EmitConst()
...
-    if (blockInfo.type != WasmTypes::Void)
+    if (isEndUnreachable) 
     {
-        Js::RegSlot loc = GetRegisterSpace(blockInfo.type)->AcquireTmpRegister();
-        if (loc != blockInfo.location) 
-        {
-            m_writer.AsmReg2(GetLoadOp(blockInfo.type), loc, blockInfo.location);
-            blockInfo.location = loc;
-        }
+        // In case there were other expr after the unreachable one
```
MikeHolman: I don't think this holds

lib/WasmReader/WasmParseTree.h
MikeHolman: I guess I see why you did this way, but in a lot of cases Limit == count of the enum, and there is risk of someone misusing like that. Might be good to rename to avoid confusion.

MikeHolman:
lgtm. Though as you note there is still more to be done with some of the stranger type checking semantics.

lib/WasmReader/WasmParseTree.h
Cellule: Yeah, I think there's going to need a good revamp of that enum.
I will try to do some cleanup of this enum in a later change

lib/WasmReader/WasmByteCodeGenerator.cpp
Cellule: Yeah I think you're right. This is so confusing

Cellule:
I will merge this and do another pass on the types.
I will try to follow the approach others are doing for types in Wasm.
Also some cleanup regarding Limit

PR #1199, opened 2016-06-25
Reducing lambda usage to improve code coverage attribution

EdMaurer:
The least covered methods at present are lambdas or template functions parameterized by lambdas. Replace those lambdas with nominal methods to make it easier to locate the source referred to in the code coverage report.

### DIFF
--- a/lib/Runtime/ByteCode/ByteCodeGenerator.cpp
+++ b/lib/Runtime/ByteCode/ByteCodeGenerator.cpp
@@ -17,12 +17,38 @@ bool IsCallOfConstants(ParseNode *pnode)
 template <class PrefixFn, class PostfixFn>
 void Visit(ParseNode *pnode, ByteCodeGenerator* byteCodeGenerator, PrefixFn prefix, PostfixFn postfix, ParseNode * pnodeParent = nullptr);
 
+//the only point of this type (as opposed to using a lambda) is to provide a named type in code coverage
+template <typename TContext> class ParseNodeVisitor
+{
+    TContext* m_context;
+    void(*m_fn)(ParseNode* pnode, ByteCodeGenerator* byteCodeGenerator, TContext* context);
+public:
+
+    ParseNodeVisitor(TContext* ctx, void(*prefixParam)(ParseNode* pnode, ByteCodeGenerator* byteCodeGenerator, TContext* context)) :
+        m_context(ctx), m_fn(prefixParam)
+    {
+    }
+
+    void operator () (ParseNode* pnode, ByteCodeGenerator* byteCodeGenerator)
+    {
+        if (m_fn)
+        {
+            m_fn(pnode, byteCodeGenerator, m_context);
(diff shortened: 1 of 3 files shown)


dilijev:
I understand the reasons for this, but doesn't modifying our code just to get better code coverage attribution seem like the wrong approach to this? Shouldn't we figure out a change in tooling which would prevent this kind of change?

Going forward, should we avoid using lambdas at all?

In any case, in this case, I don't think readability changes in any meaningful way as a result of this change, so...

LGTM!

lib/Runtime/ByteCode/ByteCodeGenerator.cpp
```diff
@@ -17,24 +17,36 @@ bool IsCallOfConstants(ParseNode *pnode)
 template <class PrefixFn, class PostfixFn>
 void Visit(ParseNode *pnode, ByteCodeGenerator* byteCodeGenerator, PrefixFn prefix, PostfixFn postfix, ParseNode * pnodeParent = nullptr);
 
+//the only point of this type (as opposed to using a lambda) is to provide a named type in code coverage
```
dilijev: I think we can improve this comment to have a more positive feel. Instead of "the only point of" -- just say what it is for. "This type is used instead of a lambda to provide a named type in code coverage."

lib/Runtime/Language/InterpreterLoop.inl
```diff
@@ -25,6 +25,126 @@
 #define INTERPRETERPROFILE 0
 #define PROFILEDOP(prof, unprof) unprof
 #endif
+
+//two layers of macros are necessary to get arguments to the invocation of the top level macro expanded.
```
satheeshravi: Didn't quite understand the reason for two layers. Isn't one macro showing up in the call stack ?

obastemur: In general, this expansion trick is not needed. However, in case the entry level macro receives another macro as an input, single layer wouldn't work. In other words, initial macro call would expand over caller. That's the reason for multiple layers.

EdMaurer:
Yes, I would have preferred to have a solution that allowed us to build a friendly report without avoiding lambdas, but it was not clear how to do so. When template types and methods are instantiated with lambda types, I'm at a loss as to how to clarify which exact type and instantiation is being dealt with. For the moment, we understand these to be the biggest uncovered functions. Let's just remove the lambdas from here and see if any further changes are needed after we improve the coverage in these.

---
In reply to: [228499050](https://github.com/Microsoft/ChakraCore/pull/1199#issuecomment-228499050) [](ancestors = 228499050)

dilijev:
Unfortunate that this is needed, but it makes sense.

ianwjhalliday:
:shipit:

aneeshdk:
LGTM

tcare:
Looks good. Interested to see the results.

PR #1202, opened 2016-06-26
Implementation of Object.getOwnPropertyDescriptors.

### LINKED ISSUE #254: Implement Object.getOwnPropertyDescriptors
[`Object.getOwnPropertyDescriptors`](https://github.com/tc39/proposal-object-getownpropertydescriptors) is stage 3. At this time, Webkit has implemented it and v8 has an open patch.

Would be a good first PR!

jordonwii:
Fixes #254 

This is a rough implementation based on a combination of getOwnPropertyDescriptor as a reference, and what I could intuit from a day of digging. 

I'd appreciate as much feedback as possible on how I did this. I used existing functions I was aware of when possible, but if there are better choices of helper functions here, please do let me know. I also did the tests all in javascript (similar to what I could find on getOwnPropertyDescriptor), and tried to make them as thorough as possible.

### DIFF
--- a/lib/Common/ConfigFlagsList.h
+++ b/lib/Common/ConfigFlagsList.h
@@ -575,7 +575,7 @@ PHASE(All)
 #define DEFAULT_CONFIG_ES7ExponentionOperator  (true)
 #define DEFAULT_CONFIG_ES7TrailingComma        (true)
 #define DEFAULT_CONFIG_ES7ValuesEntries        (true)
-
+#define DEFAULT_CONFIG_ESObjectGetOwnPropertyDescriptors (true)
 #define DEFAULT_CONFIG_ES6Verbose              (false)
 #define DEFAULT_CONFIG_ES6All                  (false)
 // ES6 DEFAULT BEHAVIOR
@@ -1012,6 +1012,9 @@ FLAGPR           (Boolean, ES6, ES6Verbose             , "Enable ES6 verbose tra
     #define COMPILE_DISABLE_ArrayBufferTransfer 0
 #endif
 FLAGPR_REGOVR_EXP(Boolean, ES6, ArrayBufferTransfer    , "Enable ArrayBuffer.transfer"                              , DEFAULT_CONFIG_ArrayBufferTransfer)
+
+FLAGPR           (Boolean, ES6, ESObjectGetOwnPropertyDescriptors, "Enable Object.getOwnPropertyDescriptors"              , DEFAULT_CONFIG_ESObjectGetOwnPropertyDescriptors)
+
 // /ES6 (BLUE+1) features/flags
 
 #ifdef ENABLE_PROJECTION
--- a/lib/Runtime/Library/JavascriptLibrary.cpp
+++ b/lib/Runtime/Library/JavascriptLibrary.cpp
@@ -3781,6 +3781,11 @@ namespace Js
             propertyCount += 2;
         }
 
+        if (scriptContext->GetConfig()->IsESObjectGetOwnPropertyDescriptorsEnabled())
+        {
+            propertyCount++;
+        }
+
         typeHandler->Convert(objectConstructor, mode, propertyCount);
 
         library->AddMember(objectConstructor, PropertyIds::length, TaggedInt::ToVarUnchecked(1), PropertyNone);
@@ -3794,6 +3799,11 @@ namespace Js
             library->AddFunctionToLibraryObject(objectConstructor, PropertyIds::defineProperty, &JavascriptObject::EntryInfo::DefineProperty, 3));
         scriptContext->SetBuiltInLibraryFunction(JavascriptObject::EntryInfo::GetOwnPropertyDescriptor.GetOriginalEntryPoint(),
             library->AddFunctionToLibraryObject(objectConstructor, PropertyIds::getOwnPropertyDescriptor, &JavascriptObject::EntryInfo::GetOwnPropertyDescriptor, 2));
+        if (scriptContext->GetConfig()->IsESObjectGetOwnPropertyDescriptorsEnabled())
+        {
+            scriptContext->SetBuiltInLibraryFunction(JavascriptObject::EntryInfo::GetOwnPropertyDescriptors.GetOriginalEntryPoint(),
+                library->AddFunctionToLibraryObject(objectConstructor, PropertyIds::getOwnPropertyDescriptors, &JavascriptObject::EntryInfo::GetOwnPropertyDescriptors, 1));
+        }
         scriptContext->SetBuiltInLibraryFunction(JavascriptObject::EntryInfo::DefineProperties.GetOriginalEntryPoint(),
             library->AddFunctionToLibraryObject(objectConstructor, PropertyIds::defineProperties, &JavascriptObject::EntryInfo::DefineProperties, 2));
         library->AddFunctionToLibraryObject(objectConstructor, PropertyIds::create, &JavascriptObject::EntryInfo::Create, 2);
@@ -6810,6 +6820,10 @@ namespace Js
 
         REG_OBJECTS_LIB_FUNC(defineProperty, JavascriptObject::EntryDefineProperty);
         REG_OBJECTS_LIB_FUNC(getOwnPropertyDescriptor, JavascriptObject::EntryGetOwnPropertyDescriptor);
+        if (scriptContext->GetConfig()->IsESObjectGetOwnPropertyDescriptorsEnabled())
+        {
+            REG_OBJECTS_LIB_FUNC(getOwnPropertyDescriptors, JavascriptObject::EntryGetOwnPropertyDescriptors);
+        }
 
         REG_OBJECTS_LIB_FUNC(defineProperties, JavascriptObject::EntryDefineProperties);
         REG_OBJECTS_LIB_FUNC(create, JavascriptObject::EntryCreate);
--- a/lib/Runtime/Library/JavascriptObject.cpp
+++ b/lib/Runtime/Library/JavascriptObject.cpp
@@ -837,6 +837,73 @@ namespace Js
         return isPropertyDescriptorDefined;
     }
 
+    Var JavascriptObject::EntryGetOwnPropertyDescriptors(RecyclableObject* function, CallInfo callInfo, ...)
+    {
+        PROBE_STACK(function->GetScriptContext(), Js::Constants::MinStackDefault);
+
+        ARGUMENTS(args, callInfo);
+        ScriptContext* scriptContext = function->GetScriptContext();
+
+        Assert(!(callInfo.Flags & CallFlags_New));
+
+        RecyclableObject* obj = nullptr;
+
+        if (args.Info.Count < 2)
+        {
+            obj = RecyclableObject::FromVar(JavascriptOperators::ToObject(scriptContext->GetLibrary()->GetUndefined(), scriptContext));
+        }
+        else
+        {
+            // Convert the argument to object first
+            obj = RecyclableObject::FromVar(JavascriptOperators::ToObject(args[1], scriptContext));
+        }
+
+        // If the object is HostDispatch try to invoke the operation remotely
+        if (obj->GetTypeId() == TypeIds_HostDispatch)
+        {
+            Var result;
+            if (obj->InvokeBuiltInOperationRemotely(EntryGetOwnPropertyDescriptors, args, &result))
+            {
+                return result;
+            }
+        }
+
+        Var ownPropertyKeys = JavascriptOperators::GetOwnPropertyKeys(obj, scriptContext);
+        Assert(JavascriptArray::Is(ownPropertyKeys));
+
+        if (!JavascriptArray::Is(ownPropertyKeys))
+        {
+            ownPropertyKeys = scriptContext->GetLibrary()->CreateArray(0);
+        }
+
+
+        JavascriptArray *ownPropsArray = JavascriptArray::FromVar(ownPropertyKeys);
+
+        RecyclableObject* resultObj = scriptContext->GetLibrary()->CreateObject(true, (Js::PropertyIndex) ownPropsArray->GetLength());
+        
+        PropertyDescriptor propDesc;
+        Var propKey = nullptr;
+
+        for (uint i = 0; i < ownPropsArray->GetLength(); i++)
+        {
+            BOOL getPropResult = ownPropsArray->DirectGetItemAt(i, &propKey);
+            Assert(getPropResult);
+
+            if (!getPropResult)
+            {
+                continue;
+            }
+            
+            PropertyRecord const * propertyRecord;
+            JavascriptConversion::ToPropertyKey(propKey, scriptContext, &propertyRecord);
+
+            Var newDescriptor = JavascriptObject::GetOwnPropertyDescriptorHelper(obj, propKey, scriptContext);
+            resultObj->SetProperty(propertyRecord->GetPropertyId(), newDescriptor, PropertyOperation_None, nullptr);
+        }
+
+        return resultObj;
+    }
+
     Var JavascriptObject::EntryGetPrototypeOf(RecyclableObject* function, CallInfo callInfo, ...)
     {
         PROBE_STACK(function->GetScriptContext(), Js::Constants::MinStackDefault);
--- a/test/Object/getOwnPropertyDescriptors.js
+++ b/test/Object/getOwnPropertyDescriptors.js
@@ -1,0 +1,159 @@
+//-------------------------------------------------------------------------------------------------------
+// Copyright (C) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE.txt file in the project root for full license information.
+//-------------------------------------------------------------------------------------------------------
+
+if (this.WScript && this.WScript.LoadScriptFile) { // Check for running in ch
+    this.WScript.LoadScriptFile("..\\UnitTestFramework\\UnitTestFramework.js");
+}
+
+function verifyObjectDescriptors(descriptors, allTruePropName, allFalsePropName) {
+    var allProperties = Object.getOwnPropertyNames(descriptors).concat(Object.getOwnPropertySymbols(descriptors));
+
+    assert.areEqual([allTruePropName, allFalsePropName], allProperties, "Result should have one descriptor for each own property");
+
+    assert.isTrue(descriptors.hasOwnProperty(allTruePropName), "Result should contain all own properties");
+    assert.isTrue(descriptors.hasOwnProperty(allFalsePropName), "Result should contain all own properties");
+    assert.areEqual(descriptors[allTruePropName].value, "fooAllTrue", "Result value attribute should match the value set by defineProperties");
+    assert.areEqual(descriptors[allFalsePropName].value, "fooAllFalse", "Result value attribute should match the value set by defineProperties");
+
+    var expectedProps = ['configurable', 'writable', 'enumerable'];
+    for (var i in expectedProps) {
+        assert.isTrue(descriptors[allTruePropName][expectedProps[i]], "Result value attribute should match the value set by defineProperties");
+        assert.isFalse(descriptors[allFalsePropName][expectedProps[i]], "Result value attribute should match the value set by defineProperties");
+    }
+}
+
+var tests = [
+    {
+        name: "Object has getOwnPropertyDescriptors method",
+            body: function() {
+                assert.isTrue(Object.hasOwnProperty("getOwnPropertyDescriptors"), 'Object should have getOwnPropertyDescriptors method');
+
+                assert.isFalse(Object.hasOwnProperty({}, "getOwnPropertyDescriptors"), 'New objects should have a property getOwnPropertyDescriptors');
+                assert.isUndefined(Object.getOwnPropertyDescriptor({}, "getOwnPropertyDescriptors"), 'Object.getOwnPropertyDescriptor({}, "getOwnPropertyDescriptors") should be undefined');
+
+                for (var p in {}) {
+                    assert.isTrue(p != "getOwnPropertyDescriptors", "getOwnPropertyDescriptors should not be enumerable on new objects");
+                }
+
+                assert.areEqual(1, Object.getOwnPropertyDescriptors.length, "Object.getOwnPropertyDescriptors requires exactly one parameter.");
+            }
+    },
+    {
+        name: "Correctly handles bad parameters.",
+        body: function() {
+            assert.throws(function() {
+                Object.getOwnPropertyDescriptors();
+            }, TypeError, "Missing first parameter should cause a TypeError.", "Object expected");
+
+            assert.throws(function() {
+                Object.getOwnPropertyDescriptors(null);
+            }, TypeError, "Null first parameter should cause a TypeError", "Object expected");
+        }
+    },
+    {
+        name: "The resulting get and set are identical with the original get and set.",
+        body: function() {
+            // This test is adapted from https://github.com/tc39/proposal-object-getownpropertydescriptors/blob/master/test/built-ins/Object/getOwnPropertyDescriptors/has-accessors.js
+            var a = {
+                get a() {},
+                set a(value) {}
+            };
+            var b = Object.getOwnPropertyDescriptors(a);
+
+            assert.isTrue(b.a.get === Object.getOwnPropertyDescriptor(a, 'a').get);
+            assert.isTrue(b.a.set === Object.getOwnPropertyDescriptor(a, 'a').set);
+        }
+    },
+    {
+        name: "For properties with string names, the list of property descriptors includes all own properties with correct descriptors",
+        body: function() {
+            var foo = {}
+
+            Object.defineProperties(foo, {
+                "fooAllTrue": {
+                    configurable: true,
+                    enumerable: true,
+                    value: "fooAllTrue",
+                    writable: true
+                },
+                "fooAllFalse": {
+                    configurable: false,
+                    enumerable: false,
+                    value: "fooAllFalse",
+                    writable: false
+                }
+            });
+
+            var desc = Object.getOwnPropertyDescriptors(foo);
+            assert.isTrue(desc instanceof Object, "Result must be an object");
+
+            verifyObjectDescriptors(desc, "fooAllTrue", "fooAllFalse");
+        }
+    },
+    {
+        name: "For properties with number names, the list of property descriptors includes all own properties with correct descriptors",
+        body: function() {
+            var foo = {}
+
+            var allTrueNum = 1234;
+            var allFalseNum = 5678;
+
+            Object.defineProperties(foo, {
+                [allTrueNum]: {
+                    configurable: true,
+                    enumerable: true,
+                    value: "fooAllTrue",
+                    writable: true
+                },
+                [allFalseNum]: {
+                    configurable: false,
+                    enumerable: false,
+                    value: "fooAllFalse",
+                    writable: false
+                }
+            });
+
+            var desc = Object.getOwnPropertyDescriptors(foo);
+            assert.isTrue(desc instanceof Object, "Result must be an object");
+
+            verifyObjectDescriptors(desc, allTrueNum.toString(), allFalseNum.toString());
+
+            // Also verify that the properties are accessible as numbers
+            assert.areEqual(desc[allTrueNum].value, "fooAllTrue", "For properties with number names, resulting properties should be accessible with numeric names.")
+            assert.areEqual(desc[allFalseNum].value, "fooAllFalse", "For properties with number names, resulting properties should be accessible with numeric names.")
+        }
+    },
+    {
+        name: "For properties with symbol names, the list of property descriptors includes all own properties with correct descriptors",
+        body: function() {
+            var foo = {}
+
+            var allTrueSymbol = Symbol("allTrue");
+            var allFalseSymbol = Symbol("allFalse");
+
+            Object.defineProperties(foo, {
+                [allTrueSymbol]: {
+                    configurable: true,
+                    enumerable: true,
+                    value: "fooAllTrue",
+                    writable: true
+                },
+                [allFalseSymbol]: {
+                    configurable: false,
+                    enumerable: false,
+                    value: "fooAllFalse",
+                    writable: false
+                }
+            });
+
+            var desc = Object.getOwnPropertyDescriptors(foo);
+            assert.isTrue(desc instanceof Object, "Result must be an object");
+
+            verifyObjectDescriptors(desc, allTrueSymbol, allFalseSymbol);
+        }
+    }
+]
+
+testRunner.runTests(tests, { verbose: WScript.Arguments[0] != "summary" });
--- a/test/Object/rlexe.xml
+++ b/test/Object/rlexe.xml
@@ -202,6 +202,12 @@
   </test>
   <test>
     <default>
+      <files>getOwnPropertyDescriptors.js</files>
+      <compile-flags>-args summary -endargs -ESObjectGetOwnPropertyDescriptors</compile-flags>
+    </default>
+  </test>
+  <test>
+    <default>
       <files>objectCreationOptimizations.js</files>
       <baseline>objectCreationOptimizations.baseline</baseline>
     </default>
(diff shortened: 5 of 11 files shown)


test/Object/getOwnPropertyDescriptors.js
```diff
@@ -0,0 +1,99 @@
...
+        WScript.Echo(Object.getOwnProperties(null));
+    });
+}
+
+function ExpectException(f) {
+    try {
+        f();
+    }
+    catch(e) {
+        WScript.Echo("PASS");
```
ianwjhalliday: It would be better to verify what type of error is thrown, since the spec specifies this.  E.g. TypeError or ReferenceError.

Further, I always recommend that the error string is also compared, to verify that we are getting the expected error rather than some other error of the same type (TypeError is very common).

This is made easier with the UnitTestFramework's assert.throws() method.  Look for other examples of usage of UTF and assert.throws().

test/Object/getOwnPropertyDescriptors.js
```diff
@@ -0,0 +1,99 @@
...
+
+tests = [
+    TestNoParameter,
+    TestNullParameter,
+    TestSameGettersAndSetters,
+    TestGetOwnPropertyDescriptors
+]
+
+for (var i in tests) {
+    CatchAndWriteExceptions(tests[i]);
```
ianwjhalliday: This sort of run all tests regardless of exceptions behavior is also made easy by using UnitTestFramework

jordonwii:
Apologies for the broken push - I didn't expect that branch to show up here before I explicitly did something somewhere to update the PR. 

In any case, I rewrote my existing tests in the UTF (which is *much* better). The only spec I could find was the one linked to in the issue, and that had one test which I included a slightly modified copy of in my tests.

Also, from the explanation of the property terminology in #254, I can definitely improve that code. It looks like propertyRecord is unused, so I can remove that, and the "propName" variable should be called "propKey".

test/Object/rlexe.xml
```diff
@@ -195,6 +195,11 @@
   </test>
   <test>
     <default>
+      <files>getOwnPropertyDescriptors.js</files>
```
tcare: Make sure to pass "-args summary -endargs" as the flags. That way we can do away with the baseline altogether.

lib/Runtime/Library/JavascriptObject.cpp
```diff
@@ -817,6 +817,64 @@ namespace Js
...
+
+        // If the object is HostDispatch try to invoke the operation remotely
+        if (obj->GetTypeId() == TypeIds_HostDispatch)
+        {
+            Var result;
+            if (obj->InvokeBuiltInOperationRemotely(EntryGetOwnPropertyDescriptors, args, &result))
+            {
+                return result;
+            }
+        }
```
ianwjhalliday: Not sure if all our built-ins need to do this.  Feels like either they all would or not though.  @Yongqu is this necessary and correct?

lib/Runtime/Library/JavascriptObject.cpp
```diff
@@ -817,6 +817,64 @@ namespace Js
...
+            PropertyRecord const * propertyRecord;
+            JavascriptConversion::ToPropertyKey(propName, scriptContext, &propertyRecord);
+
+            PropertyDescriptor newPropertyDescriptor = PropertyDescriptor();
+            newPropertyDescriptor.SetConfigurable(true);
+            newPropertyDescriptor.SetEnumerable(true);
+            newPropertyDescriptor.SetWritable(true);
+            newPropertyDescriptor.SetValue(JavascriptObject::GetOwnPropertyDescriptorHelper(obj, propName, scriptContext));
+
+            JavascriptOperators::DefineOwnPropertyDescriptor(resultObj, propertyRecord->GetPropertyId(), newPropertyDescriptor, true, scriptContext);
```
ianwjhalliday: I believe you can use `SetProperty` instead, to avoid creating a descriptor, since the attributes are all true (their defaults).  Not sure off the top of my head if you should use `resultObj->SetProperty(...)` or `JavascriptOperators::SetProperty(...)` though.  Take a look at what the JavascriptOperators API does to see if it is necessary.  I think it isn't.

lib/Runtime/Library/JavascriptObject.cpp
```diff
@@ -817,6 +817,64 @@ namespace Js
...
+            if (obj->InvokeBuiltInOperationRemotely(EntryGetOwnPropertyDescriptors, args, &result))
+            {
+                return result;
+            }
+        }
+
+        Var ownPropertyKeys = JavascriptOperators::GetOwnPropertyKeys(obj, scriptContext);
+        JavascriptArray *ownPropsArray = JavascriptArray::FromVar(ownPropertyKeys);
+
+        RecyclableObject* resultObj = scriptContext->GetLibrary()->CreateObject(true);
```
ianwjhalliday: Since we know how many keys there are up front, we could optimize this object by specifying how many properties it will have during creation.  This would then add slots for those properties inline in the allocation for the object (instead of requiring an auxiliary slot array as a second allocation).

We can probably also give the type up front too but I'm not clear on whether we can always share the type from the source object.  E.g. the source object might have special internal properties on its type which shouldn't be on the result object.

Something to make a note of but probably not a priority here.

lib/Runtime/Library/JavascriptObject.cpp
jordonwii: It didn't seem necessary to me, so I did use `resultObj->SetProperty(...)`.

lib/Runtime/Library/JavascriptObject.cpp
jordonwii: I passed in the length of `ownPropsArray` as the `requestedSlots` parameter - is that the correct way to make that optimization? If we do this, what happens if the user then tries to define a property on resultObj after we return it?

I'm guessing the way to specify the type would be the `prototype` parameter to CreateObject?  What would the type be, here?

jordonwii:
I incorporated all of the comments left on the previous diff (except the remote dispatch one, since that's unresolved). 

Is the ARM failure normal, or is it because of my change? The only error message in the Jenkins log I can find isn't clear.

lib/Runtime/Library/JavascriptObject.cpp
```diff
@@ -817,6 +817,60 @@ namespace Js
...
+
+        for (uint i = 0; i < ownPropsArray->GetLength(); i++)
+        {
+            ownPropsArray->DirectGetItemAt(i, &propKey);
+            
+            PropertyRecord const * propertyRecord;
+            JavascriptConversion::ToPropertyKey(propKey, scriptContext, &propertyRecord);
+
+			Var newDescriptor = JavascriptObject::GetOwnPropertyDescriptorHelper(obj, propKey, scriptContext);
+			resultObj->SetProperty(propertyRecord->GetPropertyId(), newDescriptor, PropertyOperation_None, nullptr);
```
ianwjhalliday: Just noticed there are tabs here.  Please convert tabs to spaces.  Sorry, a nitpick convention we maintain.

lib/Runtime/Library/JavascriptObject.cpp
```diff
@@ -817,6 +817,60 @@ namespace Js
...
+            }
+        }
+
+        Var ownPropertyKeys = JavascriptOperators::GetOwnPropertyKeys(obj, scriptContext);
+        JavascriptArray *ownPropsArray = JavascriptArray::FromVar(ownPropertyKeys);
+
+        RecyclableObject* resultObj = scriptContext->GetLibrary()->CreateObject(true, (Js::PropertyIndex) ownPropsArray->GetLength());
+        
+        PropertyDescriptor propDesc;
+        Var propKey;
```
ianwjhalliday: The arm failures appear to be due to extra static analysis.  There is a warning, and warnings are errors in our build.  The compiler believes it is possible for propKey to be used while uninitialized.

It might be the case that DirectGetItemAt doesn't always set propKey to a value.  Should be in this case, but the compiler probably can't see that.

Initializing propKey to nullptr here will probably silence the warning.

I don't know why this is only showing up on the ARM test and release builds.

test/Object/getOwnPropertyDescriptors.js
```diff
@@ -0,0 +1,89 @@
...
+if (this.WScript && this.WScript.LoadScriptFile) { // Check for running in ch
+    this.WScript.LoadScriptFile("..\\UnitTestFramework\\UnitTestFramework.js");
+}
+
+var tests = [
+    {
+        name: "Object has getOwnPropertyDescriptors method",
+            body: function() {
+                assert.isTrue(Object.hasOwnProperty("getOwnPropertyDescriptors"), 'Object.hasOwnProperty("getOwnPropertyDescriptors")');
+
```
ianwjhalliday: Need more tests here for the spec details of the property and function object, i.e. the property attributes are set correctly and the length property of the function is correct.  There are examples of such tests in some of the other es6 tests.

test/Object/getOwnPropertyDescriptors.js
```diff
@@ -0,0 +1,89 @@
...
+            }
+    },
+    {
+        name: "Correctly handles bad parameters.",
+        body: function() {
+            assert.throws(function() {
+                WScript.Echo(Object.getOwnProperties());
+            }, TypeError, "");
```
ianwjhalliday: Please fill in the error message for the assertion to verify that we are seeing the expected error rather than some other TypeError.

test/Object/getOwnPropertyDescriptors.js
```diff
@@ -0,0 +1,89 @@
...
+                    enumerable: true,
+                    value: "fooAllTrue",
+                    writable: true
+                },
+                "fooAllFalse": {
+                    configurable: false,
+                    enumerable: false,
+                    value: "fooAllFalse",
+                    writable: false
+                }
```
ianwjhalliday: Should also have some properties with symbol names and with integer names, since these property names are handled differently than string names.

ianwjhalliday:
Implementation looks great.  I'd like to see more tests before we merge it.  I apologize for not having thoroughly looked through the tests until now.

lib/Runtime/Library/JavascriptObject.cpp
ianwjhalliday: `requestedSlots` is correct.  This means we will allocate the object with its header and its initial slots in one allocation.  If later more properties are added to it they will move to an auxiliary slot array.  Hmm, that could be bad if there are a lot of properties.  Let me think about this more.  It will be fine to take this as is and change it later if this is a problem.

The type is a lot more complicated.  A type is the mapping of property names to property slot indexes in the object's slot array.  Every object has a pointer to its type in its header data.  However, it is non-trivial to know if a type can be shared.  Deep knowledge of the type system would be best to decide how to handle this.  I'll ask around.  I think it would be fine to not bother with the type, but if it is safe and cheap to do it would be best to share.

lib/Runtime/Library/JavascriptObject.cpp
ianwjhalliday: Great!

lib/Runtime/Library/JavascriptObject.cpp
jordonwii: Woops. Didn't realize those made it in.

lib/Runtime/Library/JavascriptObject.cpp
jordonwii: Makes sense - is there a way for me to verify that the build is working locally? I don't seem to be able to build the ARM target.

jordonwii:
Thanks for the comments. Assuming the ARM build now works, I think I addressed all of them in the latest commit. The tests now cover properties of the method as well as Symbol and Number names for properties.

lib/Runtime/Library/JavascriptObject.cpp
ianwjhalliday: You need a particular SDK installed to get the ARM toolset.  I'm not familiar with which SDK though.  Looks like it builds now, so I wouldn't worry about it.

lib/Runtime/Library/JavascriptObject.cpp
jordonwii: FWIW, that code is a copy of the code used by `EntryGetOwnPropertyDescriptor` (the only difference is what operation is passed to `InvokeBuiltInOperationRemotely`).

ianwjhalliday:
@jordonwii quick update, this change is ready to go imo.  I just kicked off internal tests.  Last thing I want to consider before merging is whether this should be on by default or should be guarded behind a flag.

I don't see any functional reason to guard it, but there may be other reasons to do so.  Once I have an answer for that I will merge.

ljharb:
@ianwjhalliday it's stage 4, so the only reason to put it behind a flag is if there's serious web compatibility concern. MS TC39 reps stated there was no such concern with this method, so I hope it won't end up behind a flag.

test/Object/getOwnPropertyDescriptors.js
```diff
@@ -1,89 +1,166 @@
...
-    }
-]
-
-
+//-------------------------------------------------------------------------------------------------------
+// Copyright (C) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE.txt file in the project root for full license information.
+//-------------------------------------------------------------------------------------------------------
+
+
```
dilijev: nit: blank line

test/Object/getOwnPropertyDescriptors.js
```diff
@@ -1,89 +1,166 @@
...
+            verifyObjectDescriptors(desc, allTrueSymbol, allFalseSymbol);
+        }
+    }
+
+
+]
+
+
 testRunner.runTests(tests, { verbose: WScript.Arguments[0] != "summary" });
```
dilijev: nit: newline at end of file

lib/Runtime/Library/JavascriptObject.cpp
```diff
@@ -855,7 +855,7 @@ namespace Js
         RecyclableObject* resultObj = scriptContext->GetLibrary()->CreateObject(true, (Js::PropertyIndex) ownPropsArray->GetLength());
         
         PropertyDescriptor propDesc;
-        Var propKey;
+        Var propKey = nullptr;
```
dilijev: Thanks. I like to see this. :)

lib/Runtime/Library/JavascriptObject.cpp
```diff
@@ -817,6 +817,60 @@ namespace Js
...
+        PROBE_STACK(function->GetScriptContext(), Js::Constants::MinStackDefault);
+
+        ARGUMENTS(args, callInfo);
+        ScriptContext* scriptContext = function->GetScriptContext();
+
+        Assert(!(callInfo.Flags & CallFlags_New));
+
+        RecyclableObject* obj = nullptr;
+
+
```
dilijev: nit: remove extra blank line

test/Object/rlexe.xml
```diff
@@ -195,6 +195,13 @@
...
+        <compile-flags>-args summary -endargs</compile-flags>
+        <tags></tags>
```
dilijev: You can omit the empty <tags>...</tags> line as it has the reasonable default of being empty. :)

dilijev:
@jordonwii Thanks for this. Before merging, we prefer history to be compressed via `git rebase -i` or `git merge --squash` to a single commit, unless there is future value in having multiple commits in a pull request. (Especially when some of those commits failed tests.) We like to keep the history as clean as possible.

jordonwii:
I can do that. Would that cause the previous comments on specific diff lines to lose context, though, or does GH preserve the diffs if I rewrite history?

dilijev:
@jordonwii It will fold the comments and say they are on an "outdated diff" so you'll lose context on any open discussions IF the lines changed, but hopefully that means you resolved the discussions anyway.

test/Object/getOwnPropertyDescriptors.js
```diff
@@ -0,0 +1,159 @@
...
+
+            assert.throws(function() {
+                Object.getOwnPropertyDescriptors(null);
+            }, TypeError, "Null first parameter should cause a TypeError", "Object expected");
+        }
+    },
+    {
+        name: "The resulting get and set are identical with the original get and set.",
+        body: function() {
+            // This test is taken from https://github.com/tc39/proposal-object-getownpropertydescriptors/blob/master/test/built-ins/Object/getOwnPropertyDescriptors/has-accessors.js
```
dilijev: We should probably say "adapted from" since the asserts use our UTF.

dilijev:
LGTM!

lib/Runtime/Library/JavascriptObject.cpp
```diff
@@ -817,6 +817,59 @@ namespace Js
...
+        {
+            Var result;
+            if (obj->InvokeBuiltInOperationRemotely(EntryGetOwnPropertyDescriptors, args, &result))
+            {
+                return result;
+            }
+        }
+
+        Var ownPropertyKeys = JavascriptOperators::GetOwnPropertyKeys(obj, scriptContext);
+        JavascriptArray *ownPropsArray = JavascriptArray::FromVar(ownPropertyKeys);
```
Yongqu: Can you add a JavascriptArray::Is(ownPropertyKeys) check? I am a bit afraid of potential change here in the future.

lib/Runtime/Library/JavascriptObject.cpp
```diff
@@ -817,6 +817,59 @@ namespace Js
...
+
+        RecyclableObject* resultObj = scriptContext->GetLibrary()->CreateObject(true, (Js::PropertyIndex) ownPropsArray->GetLength());
+        
+        PropertyDescriptor propDesc;
+        Var propKey = nullptr;
+
+        for (uint i = 0; i < ownPropsArray->GetLength(); i++)
+        {
+            ownPropsArray->DirectGetItemAt(i, &propKey);
```
Yongqu: check return value.

jordonwii:
@Yongqu I added Asserts to handle the failure cases in the implementation. Should any of those *not* be asserts?

dilijev:
@jordonwii You just need to ask yourself for each case, since Asserts are not present in Release builds:

* Should we crash the process (unrecoverable state results from this code path)? Then we should FailFast instead of Assert
* Was this simply an unexpected code path that we want to capture in debug builds: Then we should Assert
  * Remember that the Assert will not be present in a Release build, so the rest of the code should handle that case in some reasonable way.

lib/Runtime/Library/JavascriptObject.cpp
```diff
@@ -858,10 +859,12 @@ namespace Js
 
         for (uint i = 0; i < ownPropsArray->GetLength(); i++)
         {
-            ownPropsArray->DirectGetItemAt(i, &propKey);
+            BOOL getPropResult = ownPropsArray->DirectGetItemAt(i, &propKey);
+            Assert(getPropResult);
```
dilijev: How should we handle this case in Release builds (when the Assert is compiled out)? Is `getPropResult == false` something that should **never** happen if the code is written correctly?

ianwjhalliday: Good question.  I was assuming that `JavascriptOperators::GetOwnPropertyKeys` always returns a simple JavascriptArray, created by our runtime, so we can make assumptions about it.  However I have not verified this.  @jordonwii is that what you found?

Also I do not know about the assert below for `propertyKeyResult`.  Does `JavascriptConversion::ToPropertyKey` always succeed with a result from `JavascriptOperators::GetOwnPropertyKeys`?

lib/Runtime/Library/JavascriptObject.cpp
ianwjhalliday: Interesting.  If `GetOwnPropertyKeys` is always returning a `JavascriptArray*` we should ideally change the return type to enforce strong typing and break the callsites if the return type changes in the future.

No reason to make this a runtime check if it can be a compile time enforcement.

ianwjhalliday:
@ljharb indeed it should remain enabled so long as the spec accepts it.  We decided that it should be controlled with a flag so that it can be disabled if a scenario warrants it (e.g. website bug reduction).

@jordonwii can you add a flag to control the inclusion of this API on the Object constructor.  See `ConfigFlagsList.h` and `JavascriptLibrary.cpp` for examples of how we do this for other APIs.  In this case please add a new flag specifically for this one API.  We may consolidate it with other flags in the future but for now a specific flag for this one API is good.

ianwjhalliday:
@jordonwii the linux branch merged into master and pending PRs need to be rebased to include commit 606c442.  Could you also rebase to include the latest master please.  You can do this when feedback has settled and you're done making changes and we're ready to merge.

lib/Runtime/Library/JavascriptObject.cpp
jordonwii: Assuming the object isn't a `Proxy`, `JavascriptOperators::GetOwnPropertyKeys` just returns the result of `JavascriptObject::CreateKeysHelper`, which is always a `JavascriptArray`.

I understood @Yongqu to be concerned about catching a change in something along the chain of calls after `JavascriptOperators::GetOwnPropertyKeys` that such that it *doesn't* ultimately return an Array (hence an assert being appropriate here). I just noticed that `JavascriptArray::FromVar` would already catch such a case, though, since `FromVar` includes the assert:
`AssertMsg(Is(aValue), "Ensure var is actually a 'JavascriptArray'");`

So the one I added is redundant, it looks like. 

Both `JavascriptConversion::ToPropertyKey` and `ownPropsArray->DirectGetItemAt` return a BOOL indicating success/failure, although AFAICT the implementation of `JavascriptConversion::ToPropertyKey` *always* succeeds. That doesn't mean it always will, though, and `JavascriptObject::GetValuesOrEntries` currently includes an Assert identical to the one I added (see https://github.com/Microsoft/ChakraCore/blob/master/lib/Runtime/Library/JavascriptObject.cpp#L1202)

lib/Runtime/Library/JavascriptObject.cpp
Yongqu: reflect.ownKeys (26.1.10) has return CreateArrayFromList(keys) which creates a builtin normal array, so from spec perspective it should be a valid normal JavascriptArray. However we don't know if this could change in the future an assert might not be able to catch all the scenario, especially people run retail on the web. I think we should have a runtime check. 

---
In reply to: [69818695](https://github.com/Microsoft/ChakraCore/pull/1202#discussion_r69818695) [](ancestors = 69818695)

ianwjhalliday: Sure but this is controlled by our runtime and so if currently our runtime always returns a JavascriptArray*, we can simply be more specific in the return type, to avoid being confused about whether the result is always an array or not at the call sites.

The FromVar conversion would then be unnecessary.

Also I just remembered that FromVar performs the assert check already, so the assert added here is actually redundant as it stands.

tcare: Unless we change the contract to be JavascriptArray*, a Var from JavascriptOperators should have an ::Is check. Most JavascriptOperator methods return Var though.

Usually I would do this for anything stronger than a RecyclableObject* assumption. It's quite important for anything we don't end up catching from asserts, especially if it would require a different error path.

lib/Runtime/Library/JavascriptObject.cpp
ianwjhalliday: This opened up a good discussion in the office :)  @Yongqu @tcare and I got together and discussed the design of the APIs we have.

1. Currently the array you're calling `DirectGetItemAt` is guaranteed by our implementation to have items in order up to its length.  However we cannot convey this guarantee in the type system and so have no compile time guarantee.  In general this may not be true of JavascriptArray, elements can be deleted causing `DirectGetItemAt` to return `FALSE` for example.

 Because we cannot make a compile time guarantee here, we would like to see the return value handled in release builds.  Probably the best thing to do is Assert if it is false, and then skip the loop body and go to the next element in the array, since this shouldn't happen, but we don't want to use `propKey` if it wasn't guaranteed to be set to anything legitimate.

2. `JavascriptConversion::ToPropertyKey` should return `void`.  It is wasting time by having a confusing bool return type when it always succeeds.  This was poor API design but we can address it.  Not in this PR.  I'll open a separate issue.  I took a look and almost all callers of this API ignore the return value.  I think you should do the same and we'll change the return type to void in another change soon.

3. I verified that `JavascriptOperators::GetOwnPropertyKeys` always returns a `JavascriptArray*` regardless of which control path it takes.  `JavascriptOperators` APIs have traditionally been `Var` or `BOOL` returning, high level operations matching closely with the spec.  Returning `Var` makes sense particularly when the JIT leverages these APIs as helper calls.  However `GetOwnPropertyKeys` is used only internally to our runtime (not used by JIT) and all call sites immediately convert its result back to `JavascriptArray*`.  There's no reason we should conver to `Var` and back to `JavascriptArray*`, C++ is strongly typed, we should take advantage of that.

 I will open an issue to track making this improvement to our code base.  Let's leave it out of scope for your PR.  Since we're leaving it out though, your PR should be resilient, so you should have a code path that handles `JavascriptArray::Is(ownPropertyKeys)` returning false.  In that case you could just set ownPropertyKeys to a new zero length array via `scriptContext->GetLibrary()->CreateArray(0)`.

jordonwii: Great. That all makes sense to me. Glad this discussion was useful!

jordonwii:
@ianwjhalliday Minor question re: the flag. What I've read about TC39 says that this feature isn't strictly ES{N}, but do we want it disabled when ES6 disabled, anyway, as ES6 and ES7 features are now (i.e. do we want ES6 do be a parent flag of the new one in ConfigFlagsList.h)?

ianwjhalliday:
@jordonwii that is an orthogonal issue we can leave alone for now.  That is, go with the status quo, let it be controlled by the ES6 master switch.

Figuring out how to go forward with switch names in the face of the new way TC39 is adding features to the spec is something we haven't done yet.  We have had some discussions but nothing we like has come about yet.

jordonwii:
Alright, I added the flag and incorporated the decisions wrt the implementation. I didn't add any automated test for the flag, but verified manually that the flag being false removes getOwnPropertyDescriptors from Object, and passing in -ObjectGetOwnPropertyDescriptors makes all tests pass.

ljharb:
@ianwjhalliday @jordonwii this feature is strictly ES2017 / ES8, since it's stage 4 after ES2016 was finalized. Certainly if "ES6" is disabled, then it makes sense to disable everything that comes after that - but I also can't conceive of any scenario where you could realistically disable ES6+ going forward.

tcare:
@ljharb ES6 flag control was introduced at a time where compat modes were still a thing in IE. When Edge came along and we no longer had compat modes it just became the default. It just became a catch all for new ES features.

Often 'ES6' in Chakra is misused, including the language team who still calls themselves ES6 :)

As for the flag set itself, we don't have a need to disable an already released standard. The flag set is just used to group things. Individual flags are what we care about, where they will get fed into automation and used for debugging and reducing.

We should really review our usage of 'es6' naming to reflect current standards conventions.

lib/Common/ConfigFlagsList.h
```diff
@@ -903,6 +903,9 @@ FLAGPR_REGOVR_EXP(Boolean, ES6, ES6RegExSymbols        , "Enable ES6 RegExp symb
 FLAGPR_REGOVR_EXP(Boolean, ES6, ES6HasInstance         , "Enable ES6 @@hasInstance symbol"                          , DEFAULT_CONFIG_ES6HasInstanceOf)
 FLAGPR           (Boolean, ES6, ES6Verbose             , "Enable ES6 verbose trace"                                 , DEFAULT_CONFIG_ES6Verbose)
 FLAGPR_REGOVR_EXP(Boolean, ES6, ArrayBufferTransfer    , "Enable ArrayBuffer.transfer"                              , DEFAULT_CONFIG_ArrayBufferTransfer)
+
+FLAGPR           (Boolean, ES6, ObjectGetOwnPropertyDescriptors, "Enable Object.getOwnPropertyDescriptors"              , DEFAULT_CONFIG_ObjectGetOwnPropertyDescriptors)
```
tcare: At the very least I would prefix the flag with ES - it's handy to be able to see that this is a new language feature from a distance.

@ianwjhalliday thoughts?

test/Object/rlexe.xml
```diff
@@ -195,6 +195,12 @@
   </test>
   <test>
     <default>
+      <files>getOwnPropertyDescriptors.js</files>
+      <compile-flags>-args summary -endargs</compile-flags>
```
tcare: Passing the flag in here is a good idea too - that way unit tests still pass if you mess with what is on by default.

jordonwii:
Problem was that the "dynapogo" tests passed, but the full suite didn't, and I only noticed the results of the former. I fixed that, and added -ObjectGetOwnPropertyDescriptors to rlexe.xml.

jordonwii:
The latest push rebased off of master, and includes the Linux merge. If there isn't any more feedback, then I think we just need a verdict on whether to change the name of the flag, and then I can `rebase -i` to squash all my commits together.

dilijev:
@jordonwii If you want to avoid the rebase -i interface, you can also use `merge --squash`:

```
git fetch origin
git branch -M issue254_getownpropertydescriptors i254
git checkout -b issue254_getownpropertydescriptors origin/master
git merge --squash i254
git push issue254_getownpropertydescriptors
```

dilijev:
Yeah I think that all of those commits probably fit logically into a single commit. Definitely rebase -i or squash IMO

lib/Common/ConfigFlagsList.h
ianwjhalliday: Agree with ES prefix to easily differentiate ES feature flags at a distance.

I think we can just use ES going forward and not include the version since the features won't always clearly make it into a particular ES publication in the new yearly release train model.

lib/Runtime/Library/JavascriptLibrary.cpp
```diff
@@ -6810,6 +6815,7 @@ namespace Js
 
         REG_OBJECTS_LIB_FUNC(defineProperty, JavascriptObject::EntryDefineProperty);
         REG_OBJECTS_LIB_FUNC(getOwnPropertyDescriptor, JavascriptObject::EntryGetOwnPropertyDescriptor);
+        REG_OBJECTS_LIB_FUNC(getOwnPropertyDescriptors, JavascriptObject::EntryGetOwnPropertyDescriptors);
```
ianwjhalliday: This should be controlled by the flag too

lib/Runtime/Library/JavascriptLibrary.cpp
```diff
@@ -3794,6 +3794,11 @@ namespace Js
             library->AddFunctionToLibraryObject(objectConstructor, PropertyIds::defineProperty, &JavascriptObject::EntryInfo::DefineProperty, 3));
         scriptContext->SetBuiltInLibraryFunction(JavascriptObject::EntryInfo::GetOwnPropertyDescriptor.GetOriginalEntryPoint(),
             library->AddFunctionToLibraryObject(objectConstructor, PropertyIds::getOwnPropertyDescriptor, &JavascriptObject::EntryInfo::GetOwnPropertyDescriptor, 2));
+        if (scriptContext->GetConfig()->IsObjectGetOwnPropertyDescriptorsEnabled())
+        {
+            scriptContext->SetBuiltInLibraryFunction(JavascriptObject::EntryInfo::GetOwnPropertyDescriptors.GetOriginalEntryPoint(),
```
ianwjhalliday: propertyCount above needs to be incremented.  It is a hint for the number of properties the object is going to have.  I believe we round up to the nearest 4 or 8, but regardless, the number we supply should be exact or at least exactly the maximum possible (with all features enabled)

ianwjhalliday:
Few minor things, ES prefix on the flag name and a couple minor details in JavascriptLibrary.cpp

After fixing those I'd say good to squash

jordonwii:
@ianwjhalliday Addressed all of your comments and squashed the commits. Looks like everything built successfully. Let me know if anything else needs changing.

dilijev:
LGTM

tcare:
Looks good!

ianwjhalliday:
:+1: :shipit: I'll get this merged.  Some internal test baselines need to change so might take me a moment to get it together and validate it.

ianwjhalliday:
Huzzah, it's in! Thank you @jordonwii for all your work here, much appreciated.

tcare:
Congrats @jordonwii! Thanks for your contribution.

jordonwii:
Awesome. Thank you all for your help, feedback, and explanations, @Yongqu @ianwjhalliday @tcare @dilijev!

PR #1203, opened 2016-06-27
cross-platform: fix AutoSystemInfo Initialization

obastemur:
AutoSystemInfo class was being initialized before its cross dependencies are able to initialize.

This is a compiler/linker dependent issue, requires that AutoSystemInfo is initialized
at the end of the initializion phase.

Although the problem at hand  was not necessarily requiring a thread safe solution, the one proposed here (static function type of initializer) is thread safe and incredibly liteweight. It does the job by
passing the first `__cxx_global_var_init` (or similar) phase for its host module.

### DIFF
--- a/lib/Common/Core/SysInfo.cpp
+++ b/lib/Common/Core/SysInfo.cpp
@@ -28,7 +28,7 @@
 EXTERN_C IMAGE_DOS_HEADER __ImageBase;
 #endif
 
-AutoSystemInfo AutoSystemInfo::Data INIT_PRIORITY(300);
+AutoSystemInfo AutoSystemInfo::Data;
 
 #if DBG
 bool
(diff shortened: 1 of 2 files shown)


digitalinfinity:
Overall, looks ok- little confused though, I don't see any use of `__cxx_global_var_init `- can you explain this part some more please?

obastemur:
> I don't see any use of __cxx_global_var_init- can you explain this part some more please?

That refers to i.e. a static object is being initialized at runtime by the order defined by linker/compiler. So you don't directly use it but can observe if you assert/break on an initializing i.e. global variable.

digitalinfinity:
Got it- I thought you were saying that you'd explicitly annotated the static somewhere, understood now.
One thing- the change doesn't guarantee the order of static initialization (the original priority was there basically to make sure that the PAL statics were initialized before AutoSystemInfo). Your solution is to lazily initialize it, which I think should work, but the caveat is, no other static can directly access fields from AutoSystemInfo::Data. On windows, we do explicitly control static variable initialization- might want to leave a comment to make sure that no field should directly be accessed, and every getter should call InitializeOnDemand.

curtisman:
Can we check if there are perf impact?  We may have assumed that these are super lite weight with just a data load, and now there is an extra compare/branch.

obastemur:
Closing this one as #1234 fixes this issue without altering the behavior of AutoSystemInfo

PR #1204, opened 2016-06-27
Debugging: Multiple context stack should be in order and sorted by return address.

agarwal-sandeep:
When getting frame pointers only create DiagFrame for frames which are
below dispatch address.

### DIFF
--- a/lib/Jsrt/JsrtDebuggerObject.cpp
+++ b/lib/Jsrt/JsrtDebuggerObject.cpp
@@ -795,56 +795,85 @@ JsrtDebugStackFrames::~JsrtDebugStackFrames()
     }
 }
 
+static int __cdecl DiagStackFrameSorter(void * dispatchHaltFrameAddress, const void * diagStackFrame1, const void * diagStackFrame2)
+{
+    const DWORD_PTR *p1 = reinterpret_cast<const DWORD_PTR*>(diagStackFrame1);
+    const DWORD_PTR *p2 = reinterpret_cast<const DWORD_PTR*>(diagStackFrame2);
+
+    Js::DiagStackFrame * pStackFrame1 = (Js::DiagStackFrame *)(*p1);
+    Js::DiagStackFrame * pStackFrame2 = (Js::DiagStackFrame *)(*p2);
+
+    DWORD_PTR stackAddress1 = pStackFrame1->GetStackAddress();
+    DWORD_PTR stackAddress2 = pStackFrame2->GetStackAddress();
+
+    return stackAddress1 > stackAddress2 ? 1 : -1;
+}
+
 Js::JavascriptArray * JsrtDebugStackFrames::StackFrames(Js::ScriptContext * scriptContext)
 {
-    Js::JavascriptArray* stackTraceArray = scriptContext->GetLibrary()->CreateArray();
-
-    if (this->framesDictionary == nullptr)
-    {
-        this->framesDictionary = Anew(this->jsrtDebugManager->GetDebugObjectArena(), FramesDictionary, this->jsrtDebugManager->GetDebugObjectArena(), 10);
-    }
-    else
-    {
-        this->framesDictionary->Clear();
-    }
-
-    uint frameCount = 0;
+    Js::JavascriptArray* stackTraceArray = nullptr;
 
     ThreadContext* threadContext = scriptContext->GetThreadContext();
 
     DWORD_PTR dispatchHaltFrameAddress = threadContext->GetDebugManager()->GetDispatchHaltFrameAddress();
-    AssertMsg(dispatchHaltFrameAddress, "Didn't set the dispatchHaltFrameAddress at time of break?");
+    AssertMsg(dispatchHaltFrameAddress > 0, "Didn't set the dispatchHaltFrameAddress at time of break?");
 
-    for (Js::ScriptContext *tempScriptContext = threadContext->GetScriptContextList();
-    tempScriptContext != nullptr && tempScriptContext->IsScriptContextInDebugMode();
-        tempScriptContext = tempScriptContext->next)
+    if (dispatchHaltFrameAddress != 0)
     {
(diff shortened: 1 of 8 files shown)


agarwal-sandeep:
@akroshg Please take a look

lib/Runtime/Debug/ProbeContainer.h
```diff
@@ -91,7 +91,7 @@ namespace Js
         void Initialize(ScriptContext* pScriptContext);
         void Close();
 
-        WeakDiagStack* GetFramePointers();
+        WeakDiagStack* GetFramePointers(DWORD_PTR dispatchHaltFrameAddress = 0);
```
akroshg: I don't understand the usage of dispatchHaltFrameAddress? when this can be 0? Is this for legacy debugger support?

lib/Jsrt/JsrtDebuggerObject.cpp
```diff
@@ -808,43 +822,51 @@ Js::JavascriptArray * JsrtDebugStackFrames::StackFrames(Js::ScriptContext * scri
...
-
     ThreadContext* threadContext = scriptContext->GetThreadContext();
 
     DWORD_PTR dispatchHaltFrameAddress = threadContext->GetDebugManager()->GetDispatchHaltFrameAddress();
     AssertMsg(dispatchHaltFrameAddress, "Didn't set the dispatchHaltFrameAddress at time of break?");
 
-    for (Js::ScriptContext *tempScriptContext = threadContext->GetScriptContextList();
-    tempScriptContext != nullptr && tempScriptContext->IsScriptContextInDebugMode();
-        tempScriptContext = tempScriptContext->next)
+    if (dispatchHaltFrameAddress != 0)
```
akroshg: When can dispatchHaltFrameAddress be 0 ?

lib/Runtime/Debug/ProbeContainer.h
agarwal-sandeep: Yes for pdm based debugger it will be 0.

lib/Jsrt/JsrtDebuggerObject.cpp
```diff
@@ -808,43 +822,51 @@ Js::JavascriptArray * JsrtDebugStackFrames::StackFrames(Js::ScriptContext * scri
...
         }
+
+        // Frames can be from multiple contexts, sort them based on stack address
+        stackList->Sort(DiagStackFrameSorter, (void*)dispatchHaltFrameAddress);
+
+        stackList->Map([&](int index, Js::DiagStackFrame* stackFrame)
+        {
+            Js::DynamicObject* stackTraceObject = this->GetStackFrame(stackFrame, index);
+            Js::Var marshaledObj = Js::CrossSite::MarshalVar(scriptContext, stackTraceObject);
+            Js::JavascriptOperators::OP_SetElementI((Js::Var)stackTraceArray, Js::JavascriptNumber::ToVar(index, scriptContext), marshaledObj, scriptContext);
```
akroshg: nit: we could create the stackTraceArray just before this Map function? and also CreateArray takes the length as well, as we know by this time the actual length we could create the array the actual length.

lib/Jsrt/JsrtDebuggerObject.cpp
```diff
@@ -808,43 +822,51 @@ Js::JavascriptArray * JsrtDebugStackFrames::StackFrames(Js::ScriptContext * scri
...
-                    if (stackAddress >= dispatchHaltFrameAddress)
+                    int count = stackFrames->Count();
+                    for (int frameIndex = 0; frameIndex < count; ++frameIndex)
                     {
-                        Js::DynamicObject* stackTraceObject = this->GetStackFrame(stackFrame, frameCount);
-
-                        Js::Var marshaledObj = Js::CrossSite::MarshalVar(scriptContext, stackTraceObject);
-                        Js::JavascriptOperators::OP_SetElementI((Js::Var)stackTraceArray, Js::JavascriptNumber::ToVar(frameCount++, scriptContext), marshaledObj, scriptContext);
+                        Js::DiagStackFrame* stackFrame = stackFrames->Peek(frameIndex);
+                        stackList->Add(stackFrame);
```
akroshg: as you have remove the stackAddress >= dispatchHaltFrameAddress check, as I assume that this api will be called just after the break. Is this assumption will be valid all the time? Did you move this check somewhere? could we have assert just in case if it is not anywhere?

lib/Jsrt/JsrtDebuggerObject.cpp
agarwal-sandeep: It can't be 0 for chakracore debugging, but in case it is 0 (or not set, a BUG) we won't AV and return a empty stack array.

---
In reply to: [68851362](https://github.com/Microsoft/ChakraCore/pull/1204#discussion_r68851362) [](ancestors = 68851362)

lib/Jsrt/JsrtDebuggerObject.cpp
akroshg: Instead of OP_SetElementI, we can use DirectSetItemAt provided on JavascriptArray.

lib/Jsrt/JsrtDebuggerObject.cpp
akroshg: fair enough. Could you change above assert expression to dispatchHaltFrameAddress > 0, just to make it more readable? Also framesDictionary can be moved under this check in that case.

---
In reply to: [68853053](https://github.com/Microsoft/ChakraCore/pull/1204#discussion_r68853053) [](ancestors = 68853053,68851362)

akroshg:
Other than few comments - Looks good to me.

lib/Jsrt/JsrtDebuggerObject.cpp
agarwal-sandeep: This API can be called any time after break, when every it is called we have the dispatchHaltFrameAddress saved at time of break and we will only take frames which after below that address. That check is now in GetFramePointers/UpdateFramePointers

---
In reply to: [68852558](https://github.com/Microsoft/ChakraCore/pull/1204#discussion_r68852558) [](ancestors = 68852558)

PR #1207, opened 2016-06-27
Merge from master into linux

digitalinfinity:
Two commits in this PR- 
* The merge commit
* A commit that re-enables TTD after the merge (936b724)

In the merge commit, conflicts fixed in following files:
```        bin/ch/ChakraRtInterface.cpp
        bin/ch/ChakraRtInterface.h
        bin/ch/WScriptJsrt.cpp
        bin/ch/ch.cpp
        lib/Common/DataStructures/BaseDictionary.h
        lib/Jsrt/ChakraCommon.h
        lib/Jsrt/ChakraDebug.h
        lib/Jsrt/Jsrt.cpp
        lib/Runtime/Base/ScriptContext.cpp
        lib/Runtime/Types/NullTypeHandler.h
```


bin/ch/ChakraRtInterface.h
```diff
@@ -74,7 +74,7 @@ struct JsAPIHooks
     typedef JsErrorCode(WINAPI *JsrtTTDCreateRecordRuntimePtr)(JsRuntimeAttributes attributes, char16* infoUri, UINT32 snapInterval, UINT32 snapHistoryLength, JsThreadServiceCallback threadService, JsRuntimeHandle *runtime);
     typedef JsErrorCode(WINAPI *JsrtTTDCreateDebugRuntimePtr)(JsRuntimeAttributes attributes, char16* infoUri, JsThreadServiceCallback threadService, JsRuntimeHandle *runtime);
```
jianchun: Should `infoUrl` use `char*`?

lib/Common/CommonDefines.h
```diff
@@ -325,7 +325,7 @@
 
 ////////
 //Time Travel flags
-#define ENABLE_TTD 0
+#define ENABLE_TTD 1
```
jianchun: Don't know the context -- intentionally turn on?

lib/Jsrt/Jsrt.cpp
```diff
@@ -2870,6 +2870,9 @@ JsErrorCode RunScriptCore(INT64 hostCallbackId, const byte *script, size_t cb, L
     uint64 bodyCtrId = 0;
 #endif
 
+    LPWSTR ttdWideSourceString = nullptr;
+    AutoMallocPtr<WCHAR> ttdSourceStringGuard(nullptr);
```
jianchun: This works fine. There is an existing mechanism without need for a 2nd guard:

```
utf8::NarrowToWide ttdWideSource((LPCSTR)script);
if (!ttdWideSource) { return JsErrorOutOfMemory; }
// use ttdWideSource and ttdWideSource.Length()...
// ttdWideSource free at destructor
```

jianchun: 👍  A few minor questions/comments.

lib/Common/CommonDefines.h
digitalinfinity: Yeah, so I turned it off during the merge to get the merge to build cleanly- and then turned it back on to fix up TTD specific conflicts.

obastemur:
Commented one of the methods with char16/wchar_t argument. We shouldn't expose any wchar Jsrt method to xplat.

bin/ch/ChakraRtInterface.h
```diff
@@ -288,6 +332,28 @@ class ChakraRTInterface
     static JsErrorCode WINAPI JsRunModuleUtf8(const char *script, JsSourceContext sourceContext, const char *sourceUrl, JsValueRef *result) { return HOOK_JS_API(ExperimentalApiRunModuleUtf8(script, sourceContext, sourceUrl, result)); }
     static JsErrorCode WINAPI JsPointerToStringUtf8(const char *stringValue, size_t length, JsValueRef *value) { return HOOK_JS_API(PointerToStringUtf8(stringValue, length, value)); }
     static JsErrorCode WINAPI JsStringFree(char *stringValue) { return HOOK_JS_API(StringFree(stringValue)); }
+    static JsErrorCode WINAPI JsTTDCreateRecordRuntime(JsRuntimeAttributes attributes, char16* infoUri, UINT32 snapInterval, UINT32 snapHistoryLength, JsThreadServiceCallback threadService, JsRuntimeHandle *runtime) { return m_jsApiHooks.pfJsrtTTDCreateRecordRuntime(attributes, infoUri, snapInterval, snapHistoryLength, threadService, runtime); }
```
obastemur: Use `HOOK_JS_API` instead of m_jsApiHooks. Same for below ones

bin/ch/Helpers.cpp
```diff
@@ -211,3 +211,382 @@ void Helpers::LogError(__in __nullterminated const char16 *msg, ...)
...
+{
+    if(!ok)
+    {
+#ifdef _WIN32
+        DWORD lastError = GetLastError();
+        LPTSTR pTemp = NULL;
+        FormatMessage(FORMAT_MESSAGE_ALLOCATE_BUFFER | FORMAT_MESSAGE_FROM_SYSTEM | FORMAT_MESSAGE_ARGUMENT_ARRAY, NULL, lastError, 0, (LPTSTR)&pTemp, 0, NULL);
+        fwprintf(stderr, _u(": %s"), pTemp);
+#endif
+        fprintf(stderr, "msg is: %s", msg);
```
obastemur: if the message is utf8 and a baseline depends on an output from this interface, the output won't match cross platform.

bin/ch/Helpers.cpp
```diff
@@ -211,3 +211,382 @@ void Helpers::LogError(__in __nullterminated const char16 *msg, ...)
...
+        AssertMsg(false, "IO Error!!!");
+    }
+}
+
+void Helpers::CreateDirectoryIfNeeded(const char16* path)
+{
+#ifndef _WIN32
+    AssertMsg(false, "Not XPLAT yet.");
+#else
+    bool isPathDirName = (path[wcslen(path) - 1] == _u('\\'));
```
obastemur: what if wcslen(path) == 0 ?

obastemur: what if path separator is `/` ?

bin/ch/Helpers.cpp
```diff
@@ -211,3 +211,382 @@ void Helpers::LogError(__in __nullterminated const char16 *msg, ...)
...
+
+void Helpers::DeleteDirectory(const char16* path)
+{
+#ifndef _WIN32
+    AssertMsg(false, "Not XPLAT yet.");
+#else
+    HANDLE hFile;
+    WIN32_FIND_DATA FileInformation;
+
+    bool isPathDirName = (path[wcslen(path) - 1] == _u('\\'));
```
obastemur: same as line 237

bin/ch/Helpers.cpp
```diff
@@ -211,3 +211,382 @@ void Helpers::LogError(__in __nullterminated const char16 *msg, ...)
...
+
+HANDLE CALLBACK Helpers::TTGetLogStreamCallback(const char16* uri, bool read, bool write)
+{
+#ifndef _WIN32
+    AssertMsg(false, "Not XPLAT yet.");
+    return 0;
+#else
+    AssertMsg((read | write) & !(read & write), "Should be either read or write and at least one.");
+
+    size_t rlength = (wcslen(uri) + wcslen(_u("ttdlog.log")) + 1);
```
obastemur: uri always assumed to be ending with `/` or `\\` ?

obastemur:
BTW;  @digitalinfinity I just realized that some of the comments I'm adding not much to do with your changes but the ones merging from master. Looks like we will be dealing with them separately ?

obastemur:
:shipit:

digitalinfinity:
@obastemur Thanks for the review! Updated the PR (i'll rebase the last two commits before merging, kept them separate for ease of review).

You are correct, I'm not verifying the stability/performance of the TTD API in this PR- @mrkmarron and I can deal with that separately in a subs. I'm not fixing the comments around TTD code that you made- most of them are in win32 specific code or in functions which have a todo for xplat. We can deal with them in a separate commit, but useful feedback nonetheless.

bin/ch/ch.cpp
```diff
@@ -769,6 +769,11 @@ int _cdecl wmain(int argc, __in_ecount(argc) LPWSTR argv[])
         ChakraRTInterface::UnloadChakraDll(chakraLibrary);
     }
 
+    if (ttUri != nullptr)
+    {
+        free(ttUri);
```
obastemur: I've also used Utf8.Detach and then free in couple of places. However on our discussion with @curtisman heard that Windows DLL allocation map might differ then the calling process. We should introduce a free mechanism for Utf8 or use new JSStringFree here. IIRC `.so` shouldn't have a problem with it.

lib/Jsrt/Jsrt.cpp
```diff
@@ -2924,13 +2939,16 @@ JsErrorCode RunScriptCore(INT64 hostCallbackId, const byte *script, size_t cb, L
...
+
+            if (length == (size_t) -1)
             {
                 return JsErrorOutOfMemory;
             }
 
-            ttdSourceStringGuard = ttdWideSourceString;
+            Assert(length + 1 < INT_MAX);
+            ttdWideSourceString.Set(wideSource.Detach(), (int) (length + 1) /* include null terminator */);
             Assert(length < UINT32_MAX);
```
obastemur: The assert above and this. One should be redundant?

bin/ch/ch.cpp
digitalinfinity: Sorry, not sure I follow- the allocation and freeing of memory is done by the ch module using a function in codex which is always linked as static lib with ch so it should be fine. I think Curtis' point is regarding ChakraCore dynamic library (that is, we should treat the ChakraCore static lib case the same as if it was a dynamic library). Codex is always static so should be fine.

obastemur: >  Codex is always static so should be fine.

That makes perfect sense.

PR #1210, opened 2016-06-28
Return a value from pre_pgi.cmd to imply check passed

ThomsonTan:
Pre_pgi.cmd doesn't exit with 0 explicitly after all check passed. The caller (pogo.git.bat) will get incorrect %errorlevel% if it was set previously.


ianwjhalliday:
:shipit:

dilijev:
LGTM

PR #1214, opened 2016-06-29
Add any preprocessor definitions passed using msbuild enviornment to list of PreprocessorDefinitions

agarwal-sandeep:
This is needed for passing preprocessors to msbuild for node build https://github.com/Microsoft/ChakraCore/commit/62fe85a08bb12244626a51456222113544c706d7

### DIFF
--- a/Build/Common.Build.props
+++ b/Build/Common.Build.props
@@ -9,6 +9,9 @@
 
       <PreprocessorDefinitions Condition="'$(RuntimeLib)'=='static_library'">%(PreprocessorDefinitions);USE_STATIC_RUNTIMELIB</PreprocessorDefinitions>
 
+      <!-- Add any preprocessor definitions passed using msbuild environment -->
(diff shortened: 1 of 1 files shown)


agarwal-sandeep:
@dilijev please take a look

dilijev:
So the goal here is to use this to temporarily add definitions to all invocations of CL.exe via MsBuild by the following?

```
set AdditionalPreprocessorDefinitions=DEFINED_THING=1;ANOTHER_THING=2
msbuild ...
```

and/or

```
msbuild ... "/p:AdditionalPreprocessorDefinitions=DEFINED_THING=1;ANOTHER_THING=2" ...
```

I think either would work (although if both were used I think the `/p:...` would clobber the ENV var.

dilijev:
Besides being interested in that clarification, this LGTM.

@digitalinfinity what do you think?

agarwal-sandeep:
The idea is to use pass AdditionalPreprocessorDefinitions in msbuild switch for diabling ES6 features.
```
msbuild ... "/p:AdditionalPreprocessorDefinitions=DEFINED_THING=1;ANOTHER_THING=2" ...
```

dilijev:
Sounds good. LGTM

kunalspathak:
LGTM

PR #1216, opened 2016-06-29
Fix a with-related precise-capture bug

pleath:
Fix a precise-capture bug exposed in the unreleased branch. Also make our conservative with-statement closure-capture code immune to deferred nested functions.

### DIFF
--- a/lib/Parser/Parse.cpp
+++ b/lib/Parser/Parse.cpp
@@ -5265,10 +5265,11 @@ bool Parser::ParseFncDeclHelper(ParseNodePtr pnodeFnc, LPCOLESTR pNameHint, usho
             m_currentNodeNonLambdaDeferredFunc = pnodeFncSaveNonLambda;
         }
         m_currentNodeDeferredFunc = pnodeFncSave;
-        if (m_currentNodeFunc && pnodeFnc->sxFnc.HasWithStmt())
-        {
-            GetCurrentFunctionNode()->sxFnc.SetHasWithStmt(true);
-        }
+    }
+
+    if (m_currentNodeFunc && pnodeFnc->sxFnc.HasWithStmt())
+    {
+        GetCurrentFunctionNode()->sxFnc.SetHasWithStmt(true);
     }
 
     return true;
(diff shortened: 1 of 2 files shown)


pleath:
@aneeshdk This is the issue we were looking at yesterday evening.

lib/Parser/Parse.cpp
```diff
@@ -5265,10 +5265,11 @@ bool Parser::ParseFncDeclHelper(ParseNodePtr pnodeFnc, LPCOLESTR pNameHint, usho
             m_currentNodeNonLambdaDeferredFunc = pnodeFncSaveNonLambda;
         }
         m_currentNodeDeferredFunc = pnodeFncSave;
-        if (m_currentNodeFunc && pnodeFnc->sxFnc.HasWithStmt())
-        {
-            GetCurrentFunctionNode()->sxFnc.SetHasWithStmt(true);
-        }
+    }
```
pleath: @ianwjhalliday I think you tried to call this out in your initial review, but I missed the problem with the way I moved the code around.

ianwjhalliday:
What was the bug?

lib/Parser/Parse.cpp
ianwjhalliday: Yeah, I just was about to say :)

The other thing I noted was that we propagated other information similarly but in the caller of `ParseFncDeclHelper`, instead of here.  Would it make sense to consolidate all these flag propagations together in the same place?

pleath:
@ianwjhalliday We missed marking a func decl parse node as containing a with statement, so we missed marking the function as needing a scope object. But a nested function reference was generated expecting a scope object.

ianwjhalliday:
Gotcha.  Worth adding a regression test?

lib/Parser/Parse.cpp
pleath: This particular propagation relates to the parent node, so it has to happen here, after the function stack has been restored with the parent on the top.

pleath:
A UT will come over from unreleased/rs2.

lib/Parser/Parse.cpp
ianwjhalliday: I believe it is the same for the others?  Not super important though.

PR #1220, opened 2016-06-30
xplat: some JSRT fixes

jianchun:
- Add std includes to make "ChakraCore.h" self-contained (otherwise
  missing defitions for `size_t`, `uintptr_t`).
- Fix ChakraDebug.h to work on xplat.
  - `charcount_t` type isn't available in JSRT. Changed to `size_t`.
  - Add `_Out_writes_` definition.
  - Add a bunch of Windows types.
- Add missing JsParseScriptUtf8 export def.
- Fix non-Win32 shared library InitProcess.

### DIFF
--- a/lib/Jsrt/ChakraDebug.h
+++ b/lib/Jsrt/ChakraDebug.h
@@ -22,6 +22,14 @@
 
 #include "ChakraCommon.h"
 
+#if !defined(_WIN32) || !defined(_MSC_VER)
+typedef uint32_t UINT32;
+typedef int64_t INT64;
+typedef void* HANDLE;
+typedef unsigned char BYTE;
+typedef UINT32 DWORD;
+#endif
+
     /// <summary>
     ///     Debug events reported from ChakraCore engine.
     /// </summary>
@@ -629,25 +637,25 @@
     /////////////////////
     /// <summary>
     ///     TTD API -- may change in future versions:
-    ///     Given the uri location specified for the TTD output data, which may be relative or contain other implcit information, 
-    ///     convert it into a fully normalized location descriptor. This fully resolved location will be passed to the later callbacks 
+    ///     Given the uri location specified for the TTD output data, which may be relative or contain other implcit information,
+    ///     convert it into a fully normalized location descriptor. This fully resolved location will be passed to the later callbacks
     ///     such as JsTTDInitializeForWriteLogStreamCallback, JsTTDGetLogStreamCallback, and JsTTDGetSnapshotStreamCallback.
     /// </summary>
     /// <param name="uri">The uri the user provided for the output location of the TTD data.</param>
     /// <param name="fullTTDUri">The fully resolved location for the TTD data output.</param>
-    typedef void (CALLBACK *JsTTDInitializeUriCallback)(_In_z_ const wchar_t* uri, _Out_ wchar_t** fullTTDUri);
+    typedef void (CHAKRA_CALLBACK *JsTTDInitializeUriCallback)(_In_z_ const wchar_t* uri, _Out_ wchar_t** fullTTDUri);
 
     /// <summary>
     ///     TTD API -- may change in future versions:
-    ///     Ensure that the location specified for outputting the TTD data is clean. Specifically, ensure that any previous TTD 
+    ///     Ensure that the location specified for outputting the TTD data is clean. Specifically, ensure that any previous TTD
     ///     in the location has been removed.
     /// </summary>
     /// <param name="fullTTDUri">The fully resolved location for the TTD data output as provied by JsTTDInitializeUriCallback.</param>
-    typedef void (CALLBACK *JsTTDInitializeForWriteLogStreamCallback)(_In_z_ const wchar_t* uri);
+    typedef void (CHAKRA_CALLBACK *JsTTDInitializeForWriteLogStreamCallback)(_In_z_ const wchar_t* uri);
 
     /// <summary>
     ///     TTD API -- may change in future versions:
-    ///     Construct a HANDLE that will be used to read/write the event log portion of the TTD data based on the uri 
+    ///     Construct a HANDLE that will be used to read/write the event log portion of the TTD data based on the uri
     ///     provided by JsTTDInitializeUriCallback.
     /// </summary>
     /// <remarks>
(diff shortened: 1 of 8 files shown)


jianchun:
@digitalinfinity @agarwal-sandeep @Yongqu @obastemur  please review, thanks!

lib/Jsrt/ChakraDebug.h
```diff
@@ -22,6 +22,14 @@
 
 #include "ChakraCommon.h"
 
+#if !defined(_WIN32) || !defined(_MSC_VER)
```
obastemur: This is for clang only right? how about `__clang__` ?

lib/Jsrt/JsrtHelper.cpp
```diff
@@ -89,7 +89,7 @@ void JsrtCallbackState::ObjectBeforeCallectCallbackWrapper(JsObjectBeforeCollect
         pthread_key_create(&s_threadLocalDummy, DISPOSE_CHAKRA_CORE_THREAD);
 #endif
 
-#if defined(CHAKRA_STATIC_LIBRARY)
+#if defined(CHAKRA_STATIC_LIBRARY) || !defined(_WIN32)
```
obastemur: What is the intention here?

obastemur:
A suggestion and a question. LGTM otherwise.

obastemur:
BTW; Saw additional TTD work is coming to master. Possibly it may break this part again.

lib/Jsrt/ChakraDebug.h
jianchun: Copied this from "ChakraCommon.h" -- non-Win32, or non-VS (clang?) on Win32. Better than __clang__?

lib/Jsrt/JsrtHelper.cpp
jianchun: Basically it was failing (other host with libChakraCore.so) until I enabled this part. In the case of ch, we have manual PAL_Init... call in wmain. Other hosts only call JSRT. This code was excluded and InitializeProcess() was no-op. PAL_Init... was never called, and `DllMain` was not called either.

jianchun:
> BTW; Saw additional TTD work is coming to master. Possibly it may break this part again.

Thanks for the warning. Will fix accordingly before we have some guard in place...

lib/Jsrt/JsrtHelper.cpp
jianchun: For `ch` case I was talking about https://github.com/Microsoft/ChakraCore/blob/linux/bin/ch/ch.cpp#L659. I was thinking of removing that call and see if `ch` works.

lib/Jsrt/ChakraDebug.h
obastemur: I would go with `__clang__` instead. For someone not familiar with the intention, it is confusing. I had commented on that usage earlier and the reason given was not having an alternative. IMO `__clang__` is an alternative here.

lib/Jsrt/JsrtHelper.cpp
obastemur: > Basically it was failing (other host with libChakraCore.so) until I enabled this part. 

We shouldn't try to fix shared library issues by enabling static lib initialization.

> In the case of ch, we have manual PAL_Init... call in wmain. Other hosts only call JSRT.

Actually, my plan was to remove PAL_Init.. call entirely. Similar to SysInfo update, we will initialize PAL part lazily.  My guess is that we will have WinBash is also working after that. However, for now, the only option we have is to use static lib on other host until we cleanup that part or initialize PAL as we do with ch. Enabling static lib features on non-Windows no matter what doesn't sound like a good idea.

jianchun: Are you ok with this workaround before you fix that? The current situation is all the initializations in this function are missing.

Reason I'm still working with shared lib is because static lib is unstable (hangs) and need more time. I need to ensure something working for coming events.

obastemur: > Reason I'm still working with shared lib is because static lib is unstable (hangs)

You are referring this for using from outside right? I will dedicate some personal time if you share me a repo/branch at that particular state. Currently static part is the only part is working cross platform (osx etc.). Whatever the problem is there, I'm not sure it is because of the CC side. Actually by enabling this part; you are using the static lib interface for shared lib. So, it is interesting that it doesn't hang.

Please feel free to enable this trick until I introduce lazy init to PAL though. I really appreciate if you can add a todo note to me there.

jianchun: > Actually by enabling this part; you are using the static lib interface for shared lib. So, it is interesting that it doesn't hang.

From the code, you had following effective on static lib *and* xplat shared lib:
```
#if defined(CHAKRA_STATIC_LIBRARY) || !defined(_WIN32)
_NOINLINE void VALIDATE_ENTER_CURRENT_THREAD();
```
VALIDATE_ENTER_CURRENT_THREAD ensures 2 things lazily -- InitProcess and InitThread. Previously InitThread is done for xplat shared lib -- not sure if you did that intentionally or accidentally. Any way, this change turns on InitProcess, and it works...



> Please feel free to enable this trick until I introduce lazy init to PAL though. I really appreciate if you can add a todo note to me there.

Sure

obastemur: @jianchun if you read the code and comments carefully, that is used in order to make sure we clean the things for the thread on shared library (linux) Before that code, it was leaking..

As for the PAL thing. I would like step a bit back. Currently the only reason for enabling this trick is to enable PAL lazily from outside right? Eventually, what I understood from both the messages here and email I've got that you do this to make some unit tests passing temporarily on other project right? However, I'm not sure if introducing this unintended change to main code is a really good approach.

Before merging this, it could be nice to have some details on what is actually failing and how etc.

jianchun: No, it was not for some unit tests passing. Without this change, (from outside) shared lib completely broke. Nobody does global initializations, ThreadContextTLSEntry/ThreadContext/...

lib/Jsrt/ChakraDebug.h
jianchun: I don't know what the best is, but don't want to deviate from existing usage. If we are to fix, fix them all. P.S. This entire piece might go away when TTD resolves the types issue.

jianchun:
Rebased, added a TODO comment for Oguz, disabled a failing test. @obastemur @digitalinfinity

obastemur:
> No, it was not for some unit tests passing. Without this change, (from outside) shared lib completely broke.

@jianchun My reference was for static library. I'm testing CC static library with our Jsrt sample (without any additional PAL etc call) and it just works fine. AFAIK the priority there was to use static library on xplat.

digitalinfinity:
LGTM

jianchun:
> My reference was for static library. I'm testing CC static library with our Jsrt sample (without any additional PAL etc call) and it just works fine. AFAIK the priority there was to use static library on xplat.

I see. As mentioned I need shared lib working before figuring out what's wrong with static lib. Will make sure this TODO be addressed.

jianchun:
Thanks @obastemur @digitalinfinity

PR #1224, opened 2016-06-30
A native application for running native unittests.

akroshg:
This application is based upon the Catch framework (Open source). Most our internal tests are transformed from the TE to adopt the catch framework.


bin/nativetests/JsRTApiTest.cpp
```diff
@@ -0,0 +1,1603 @@
...
+// Copyright (C) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE.txt file in the project root for full license information.
+//-------------------------------------------------------------------------------------------------------
+#include "stdafx.h"
+#include "catch.hpp"
+#include <process.h>
+
+#pragma warning(disable:4100) // unreferenced formal parameter
+
+namespace JsRTApiTest
```
akroshg: These tests are transformed from our full repo. So the intent of the tests have not been changed.

akroshg:
@digitalinfinity @dilijev @ianwjhalliday @EdMaurer  have a look.

jianchun:
Looks LoadScriptFile conflicts with changes in Linux branch

akroshg:
thanks @Jianchun, I'll have a look at those. Also please have a look at the changes wrt to cross plat. I haven't done anything specific to enable that. That would be next stage.

dilijev:
@akroshg We will need to discuss whether it is appropriate to add the Microsoft copyright notice to `bin/External/catch.hpp` or to add an exclusion for that file or the entire directory.

If that code is licensed under a different license we might need to update `LICENSE.txt`

/cc @EdMaurer @liminzhu on that decision.

akroshg:
Sure @dilijev, I have initiated offline discussion for this.

akroshg:
Strange, *_release branch failures do not show up in the our buddy build system. @dilijev do you know if there are difference in between those?

dilijev:
@akroshg I think you mean the PreFAST warnings? The commands run in buddy builds are slightly different from the Jenkins scripts. We should probably add PreFAST analysis there. PreFAST analysis is only performed in release builds so that's why it only shows up there.

You can see `netci.groovy` to see what commands are run so that you can reproduce the issues locally. Sent some more info offline.

To repro, you'll need to run:

```
call .\jenkins\buildone.cmd x64 release "/p:runcodeanalysis=true"  
powershell .\Build\scripts\check_prefast_error.ps1 . CodeAnalysis.err 
```

bin/fileloadhelpers/FileLoadHelpers.cpp
```diff
@@ -0,0 +1,140 @@
...
+    if (nullptr == contentsRaw)
+    {
+        fwprintf(stderr, _u("out of memory"));
+        IfFailGo(E_OUTOFMEMORY);
+    }
+
+    //
+    // Read the entire content as a binary block.
+    //
+    fread((void*)contentsRaw, sizeof(char), lengthBytes, file);
```
obastemur: the assumption `fread` has read entire content is wrong

bin/fileloadhelpers/FileLoadHelpers.cpp
```diff
@@ -0,0 +1,140 @@
...
+    fclose(file);
+    *(WCHAR*)((byte*)contentsRaw + lengthBytes) = _u('\0'); // Null terminate it. Could be LPCWSTR.
```
obastemur: `fread` may not read lengthBytes. Its return value tells the total number of bytes and it can be lower than `lengthBytes`

bin/fileloadhelpers/FileLoadHelpers.cpp
```diff
@@ -0,0 +1,140 @@
...
+    if (fopen_s(&file, filename, "rb") != 0)
+    {
+        if (printFileOpenError)
+        {
+#ifdef _WIN32
+            DWORD lastError = GetLastError();
+            char16 wszBuff[512];
+            fprintf(stderr, "Error in opening file '%s' ", filename);
+            wszBuff[0] = 0;
+            if (FormatMessage(FORMAT_MESSAGE_FROM_SYSTEM,
```
obastemur: do we really need this? can we output something here cross platform?

akroshg: @obastemur  This code refactored from the ch\helpers.cpp. Nothing new is added here than it was before.

ianwjhalliday:
This looks good to me at first glance.  Some checkin gates have failed but overall I don't see anything I'd change about the approach taken.

bin/fileloadhelpers/FileLoadHelpers.cpp
obastemur: @akroshg Is there a reason that you haven't change this one before merging? 
RE: `the assumption fread has read entire content is wrong`

bin/fileloadhelpers/FileLoadHelpers.cpp
obastemur: @akroshg  Same goes to here. 
RE: `fread may not read lengthBytes. Its return value tells the total number of bytes and it can be lower than lengthBytes` 

The usages of `fread` are wrong as commented on this PR. See also https://msdn.microsoft.com/en-us/library/kt0etdcs.aspx or http://man7.org/linux/man-pages/man3/fread.3.html

akroshg: @obastemur  The reason I left as-is, as the tests under nativetests are testing non-utf8 version. So I didn't utilize the one in the ch\helpers.cpp.
Nativetests is not cross plat yet, I have to see when I can work on that.
However the usage of fread, I can fix that.

PR #1226, opened 2016-06-30
xplat: revert temp shared lib init change

jianchun:
The previous temp shared lib init change breaks ch. ch already calls
PAL_init... manually even with shared lib. Cannot init again in
InitProcess.

### DIFF
--- a/lib/Jsrt/JsrtHelper.cpp
+++ b/lib/Jsrt/JsrtHelper.cpp
@@ -89,9 +89,7 @@ void JsrtCallbackState::ObjectBeforeCallectCallbackWrapper(JsObjectBeforeCollect
         pthread_key_create(&s_threadLocalDummy, DISPOSE_CHAKRA_CORE_THREAD);
 #endif
 
-// xplat-todo: Oguz: revert " || !defined(_WIN32)". This was supposed for
-// static lib only. Temporary workaround for shared lib.
-#if defined(CHAKRA_STATIC_LIBRARY) || !defined(_WIN32)
+#if defined(CHAKRA_STATIC_LIBRARY)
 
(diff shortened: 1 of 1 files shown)


jianchun:
@obastemur @digitalinfinity Reverting the temp shared lib change as it broke ch. Will deal with it in host project instead.

digitalinfinity:

digitalinfinity:
Thanks @Jianchun. To be clear, I don't think the problem here was that ch called PAL_Init manually- ch loads the ChakraCore shared lib in ChakraRTInterface::LoadChakraDll- this uses PAL's LoadLibrary to load the module, and LoadLibrary will call DllMain in the shared library. This does the same things that InitializeProcess does in JsrtHelper, so we end up double-initializing when JsCreateRuntime is called.

jianchun:
> ch loads the ChakraCore shared lib in ChakraRTInterface::LoadChakraDll

Thanks for the correction, will include this in merge commit message.

obastemur:
RE: https://github.com/Microsoft/ChakraCore/pull/1220#discussion_r69186505 

> @obastemur @digitalinfinity Reverting the temp shared lib change as it broke ch. Will deal with it in host project instead.

@jianchun As I mentioned on the previous comment; It could be nice to have a repro etc. details. I'm using Jsrt sample for this purpose. Currently IMO you might better focus on making static lib is working with other project and I am ready to help there. reminder: Shared library option won't be much useful on some other platforms. I'm not sure if the current DllMain approach also works properly anyways.

jianchun:
@obastemur I have completed updating to latest changes with shared lib for now. Will look at static lib and update you soon.

PR #1229, opened 2016-07-01
xplat: support disabling experimental features

jianchun:
Enable disabling experimental features in parallel of msbuild support.

```
  build.sh --without=Simdjs
```

Translates to `-DCOMPILE_DISABLE_Simdjs=1`.

Note:
- Experimental feature name is case sensitive.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -31,6 +31,9 @@ PRINT_USAGE() {
     echo "  -t, --test-build    Test build (by default Release build)"
     echo "      --static        Build as static library (by default shared library)"
     echo "  -v, --verbose       Display verbose output including all options"
+    echo "      --without=FEATURE,FEATURE,..."
+    echo "                      Disable FEATUREs from JSRT experimental"
+    echo "                      features."
     echo ""
     echo "example:"
     echo "  ./build.sh --cxx=/path/to/clang++ --cc=/path/to/clang -j"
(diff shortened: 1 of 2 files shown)


jianchun:
@digitalinfinity @agarwal-sandeep Please review, thanks

build.sh
```diff
@@ -116,6 +120,16 @@ while [[ $# -gt 0 ]]; do
...
         ;;
 
+    --without=*)
+        FEATURE=$1
+        if [[ "$WITHOUT_FEATURES" == "" ]]; then
+            WITHOUT_FEATURES="-DWITHOUT_FEATURES="
+        else
+            WITHOUT_FEATURES="$WITHOUT_FEATURES;"
+        fi
+        WITHOUT_FEATURES="${WITHOUT_FEATURES}-DCOMPILE_DISABLE_${FEATURE:10}=1"
```
agarwal-sandeep: Maybe a comment here explaining why 10. Rest looks good to me

obastemur:
@jianchun are without options case sensitive?

build.sh
jianchun: Agree it is not straight forward. This is the same approach used through out the script, need some effort to understand.

jianchun:
> are without options case sensitive?

Yes, as mentioned in description. Explored case-insensitive names but that'd make code very ugly and inconsistent.

obastemur:
>  build.sh --without=Simdjs --without=...

This is a very unusual approach. I read this as the second `--without` alters the first one. Why don't we use comma separated option here?

jianchun:
> Why don't we use comma separated option here?

Sounds better, changed to use comma separated list.

obastemur:
:shipit:

jianchun:
@agarwal-sandeep @obastemur Thanks!

PR #1232, opened 2016-07-01
cross platform: fix broken unix build

obastemur:
Fixes broken Unix build after #1220

### DIFF
--- a/lib/Jsrt/ChakraDebug.h
+++ b/lib/Jsrt/ChakraDebug.h
@@ -22,7 +22,7 @@
 
 #include "ChakraCommon.h"
 
-#if !defined(_WIN32) || !defined(_MSC_VER)
+#if (!defined(_WIN32) || !defined(_MSC_VER)) && !defined(__MSTYPES_DEFINED)
 typedef uint32_t UINT32;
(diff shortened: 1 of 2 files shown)


lib/Jsrt/ChakraDebug.h
```diff
@@ -22,7 +22,7 @@
 
 #include "ChakraCommon.h"
 
-#if !defined(_WIN32) || !defined(_MSC_VER)
+#if (!defined(_WIN32) || !defined(_MSC_VER)) && !defined(PAL_DEFINES_MSTYPES)
```
jianchun: Since this is public JSRT header, `PAL_...` term here sounds a bit strange to external user. Maybe be more generic? (Expect most of defs here would go away.)

jianchun:
LGTM, one comment.

lib/Jsrt/ChakraDebug.h
obastemur: makes sense but alternative `MSTYPES_WERE_ALREADY_DEFINED` feel so long. Any idea?

obastemur:
@jianchun Thanks for the review.

PR #1233, opened 2016-07-02
Fix #36: class name in extends clause should be ReferenceError (use before declaration).

### LINKED ISSUE #36: Class name in extends clause should be ReferenceError
Adapted from test262
https://github.com/tc39/test262/blob/master/test/language/statements/class/name-binding/in-extends-expression-assigned.js

```
var x = (class x extends x {}); // expected: ReferenceError, actual: TypeError
```

The error seen looks like the following:

```
TypeError: Function is not a constructor
        at Global code (E:\dev\debugging\class\repro.js:1:1)
```

Details about the expected behavior can be found in [ECMAScript 2015 14.5.14](http://www.ecma-international.org/ecma-262/6.0/index.html#sec-runtime-semantics-classdefinitionevaluation).

**UPDATE:** See [ES 2016 14.5.14](https://tc39.github.io/ecma262/2016/#sec-runtime-semantics-classdefinitionevaluation) and [ES 2017 14.5.13](https://tc39.github.io/ecma262/#sec-runtime-semantics-bindingclassdeclarationevaluation).

There is a scoping issue with the extends clause that is causing x to appear to be of an invalid type, rather than causing an error because of the attempt to reference the type which is being defined.

The following should also produce a `ReferenceError` when `eval("x")` is evaluated as the `extends` clause.

```
var x = (class x extends eval("x") {}); // expected: ReferenceError, actual: TypeError
```

See bug [4310932](https://microsoft.visualstudio.com/web/wi.aspx?pcguid=cb55739e-4afe-46a3-970f-1b49d8ee7564&id=4310932)

dilijev:
Previously would be TypeError or ReferenceError with a different message in many cases.

The spec changed at some point to make it so that the name of the class should effectively live in a temporary deadzone when the extends clause expression is being evaluated, but we were treating it as part of the scope in which the class statement or expression is being evaluated, which means that it could be undefined (in the case of `var x = (class x extends x`)) or not defined (in the case of `let x = (class x extends x)`), and similar behavior. In the case of `let x = ...` we would emit the correct error

`eval('x')` and other expressions using eval in the extends clause should also have the same semantics for binding `x` as long as we have the scopes organized correctly.

What this essentially meant was that we needed to move the parsing of the extends clause inside the class declaration's scope, and update the bytecode generator to reflect the new semantics.

Added a variety of tests to capture different cases to cover all the possibilities that produced different error cases. Also fixed up some tests that were no longer valid under the new semantics.

Thanks @ianwjhalliday for helping me understand the parser and bytecode semantics.
Thanks @bterlson for helping me understand the spec semantics.

Fixes #36.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/1233)

### DIFF
--- a/lib/Runtime/ByteCode/ByteCodeGenerator.cpp
+++ b/lib/Runtime/ByteCode/ByteCodeGenerator.cpp
@@ -286,11 +286,13 @@ void Visit(ParseNode *pnode, ByteCodeGenerator* byteCodeGenerator, PrefixFn pref
     }
     case knopClassDecl:
     {
-        // Visit the extends expression first, since it's bound outside the scope containing the class name.
-        Visit(pnode->sxClass.pnodeExtends, byteCodeGenerator, prefix, postfix);
         Visit(pnode->sxClass.pnodeDeclName, byteCodeGenerator, prefix, postfix);
         // Now visit the class name and methods.
         BeginVisitBlock(pnode->sxClass.pnodeBlock, byteCodeGenerator);
+        // The extends clause is bound to the scope which contains the class name
+        // (and the class name identifier is in a TDZ when the extends clause is evaluated).
+        // See ES 2017 14.5.13 Runtime Semantics: ClassDefinitionEvaluation.
+        Visit(pnode->sxClass.pnodeExtends, byteCodeGenerator, prefix, postfix);
         Visit(pnode->sxClass.pnodeName, byteCodeGenerator, prefix, postfix);
         Visit(pnode->sxClass.pnodeStaticMembers, byteCodeGenerator, prefix, postfix);
         Visit(pnode->sxClass.pnodeConstructor, byteCodeGenerator, prefix, postfix);
--- a/test/es6/classes.js
+++ b/test/es6/classes.js
@@ -696,17 +696,9 @@ var tests = [
   {
     name: "Immutable binding within class body, declarations also have normal let binding in enclosing context",
     body: function() {
-
-      // Class expression with a name, eval should bind against enclosing context, not context containing
-      // class name.
-      class a{};
-
-      this.k = a;
-      var c = class k extends eval('k') {
-          k() { k(); }
+      var c = class k {
           reassign() { eval('k = 0; WScript.Echo(k);'); }
-      }
-      assert.areEqual(Object.getPrototypeOf(c.prototype), this.k.prototype, "Extends calling eval");
+      };
 
       // Class name is immutable within class body.
       var obj1 = new c();
@@ -714,8 +706,9 @@ var tests = [
 
       // Class name is also immutable within body of class declaration statement
       class Q extends c {
-          reassign() { eval('Q = 0;') }
+          reassign() { eval('Q = 0;'); }
       };
+
       var obj2 = new Q();
       assert.throws(function() { obj2.reassign() }, TypeError);
       // Class name binding in enclosing context is mutable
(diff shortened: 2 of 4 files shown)


dilijev:
Please review? @ianwjhalliday 
@suwc  who worked previously in the area of the test I fixed up (relating to reassigning the class name inside the body of the class
@digitalinfinity @curtisman FYI

test/es6/classes.js
```diff
@@ -1140,6 +1133,55 @@ var tests = [
...
+
+            // Assigning the result to a different identifier
+            assert_referr("var y = class x extends x {}");
+            assert_referr("let y = class x extends x {}");
+
+            // Define `y` after the class to use default initialization (var y) or temporary deadzone (let y).
+            assert.throws(`
+                class x extends y {}; // y == undefined
+                var y = function() {};
+            `, TypeError);
```
ianwjhalliday: Is this right? assert.throws expects a function as its first argument.  It might be throwing TypeError because you're passing a string instead.

dilijev: Whoops. You're right.

dilijev: @ianwjhalliday Fixed. It's supposed to be a TypeError but you're right that it was throwing TypeError for the wrong reason in this case.

lib/Runtime/ByteCode/ByteCodeGenerator.cpp
```diff
@@ -286,11 +286,10 @@ void Visit(ParseNode *pnode, ByteCodeGenerator* byteCodeGenerator, PrefixFn pref
     }
     case knopClassDecl:
     {
-        // Visit the extends expression first, since it's bound outside the scope containing the class name.
-        Visit(pnode->sxClass.pnodeExtends, byteCodeGenerator, prefix, postfix);
         Visit(pnode->sxClass.pnodeDeclName, byteCodeGenerator, prefix, postfix);
         // Now visit the class name and methods.
         BeginVisitBlock(pnode->sxClass.pnodeBlock, byteCodeGenerator);
+        Visit(pnode->sxClass.pnodeExtends, byteCodeGenerator, prefix, postfix);
```
tcare: We might want to add a comment about the reason for the visitation order since we removed the old comment.

dilijev: I'll add a comment, but I think that comment was there because it was unusual/unexpected that the extends expression was bound outside the scope containing the class name. It seems that was true in the spec at some point but it was updated to the current behavior which has the extends clause within the scope containing the class name (where inside the extends clause the class name is an unbound reference)

tcare: Fair enough. I've just become used to weird behavior I think. :)

lib/Runtime/ByteCode/ByteCodeGenerator.cpp
```diff
@@ -286,11 +286,13 @@ void Visit(ParseNode *pnode, ByteCodeGenerator* byteCodeGenerator, PrefixFn pref
...
     {
-        // Visit the extends expression first, since it's bound outside the scope containing the class name.
-        Visit(pnode->sxClass.pnodeExtends, byteCodeGenerator, prefix, postfix);
         Visit(pnode->sxClass.pnodeDeclName, byteCodeGenerator, prefix, postfix);
         // Now visit the class name and methods.
         BeginVisitBlock(pnode->sxClass.pnodeBlock, byteCodeGenerator);
+        // The extends clause is bound to the scope which contains the class name
+        // (and the class name identifier is in a TDZ when the extends clause is evaluated).
+        // See ES 2017 14.5.13 Runtime Semantics: ClassDefinitionEvaluation.
+        Visit(pnode->sxClass.pnodeExtends, byteCodeGenerator, prefix, postfix);
```
dilijev: @tcare How's this?

tcare: Very useful, thank you!

tcare:
Looks good

dilijev:

dilijev:
@ianwjhalliday could you sign off when ready?

suwc:
LGTM.

suwc:
:shipit:

ianwjhalliday:
:shipit:

dilijev:
Thanks everyone! Merging.

PR #1236, opened 2016-07-05
Remove unused methods

obastemur:
~~I saw these methods couple of hours ago and~~ 
~~wanted to open a separate PR for discussion / suggestions~~

Update:  
Problem: Some methods were calling theirselves.

Solution: Discussion #1236 suggests the removal of these methods.

### DIFF
--- a/lib/Parser/CharSet.h
+++ b/lib/Parser/CharSet.h
@@ -477,12 +477,6 @@ namespace UnifiedRegex
         }
 
         //Replaces an existing character with a new value
-        inline void ReplaceCompactChar(uint index, Char value)
-        {
-            ReplaceCompactChar(index, (Char)(value));
-        }
-
-        //Replaces an existing character with a new value
         inline void ReplaceCompactCharU(uint index, uint value)
         {
             Assert(index < this->GetCompactLength());
--- a/lib/Runtime/Language/ModuleNamespace.h
+++ b/lib/Runtime/Language/ModuleNamespace.h
@@ -20,8 +20,6 @@ namespace Js
     public:
         static ModuleNamespace* GetModuleNamespace(ModuleRecordBase* moduleRecord);
 
-        static bool Is(Var ns) { return ModuleNamespace::Is(ns); }
-
         //virtual int GetPropertyCount() override;
         //virtual PropertyId GetPropertyId(PropertyIndex index) override;
         //virtual PropertyId GetPropertyId(BigPropertyIndex index) override;


lib/Parser/CharSet.h
```diff
@@ -479,7 +479,8 @@ namespace UnifiedRegex
         //Replaces an existing character with a new value
         inline void ReplaceCompactChar(uint index, Char value)
```
agarwal-sandeep: Seems like the function is not used anywhere, lets remove it instead.

dilijev: I also confirmed it's unused. Agree it should be removed.

lib/Runtime/Language/ModuleNamespace.h
```diff
@@ -20,7 +20,11 @@ namespace Js
     public:
         static ModuleNamespace* GetModuleNamespace(ModuleRecordBase* moduleRecord);
 
-        static bool Is(Var ns) { return ModuleNamespace::Is(ns); }
+        static bool Is(Var ns)
+        {
+            // This function calls itself ?
+            return false; // ModuleNamespace::Is(ns);
```
dilijev: I changed the body of this function to 

```c++
            AssertMsg(false, "ModuleNamespace::Is");
            return false; // ModuleNamespace::Is(ns);
```

Observed no failures when running the unit test suite.

I surmise that this function is likely not called under normal circumstances. I'm not necessarily sure that returning `false` is the right thing to do, but it's better than letting it pass through to DynamicObject::Is which could return `true` even if `ModuleNamespace::Is` should be false.

Maybe it's worth making this change and seeing how a Crawler run goes.

dilijev:
@aneeshdk Does code coverage tell us anything about these methods?

aneeshdk:
No they are not covered by any test

lib/Runtime/Language/ModuleNamespace.h
agarwal-sandeep: Would be good to add an assert(false) to know when this is called. Crawler will probably catch this.

dilijev: Should we change to

```c++
            AssertMsg(false, "ModuleNamespace::Is is not expected to be called");
            return false; // ModuleNamespace::Is(ns);
```

And check that in and check that we get no hits on it for a while before removing it altogether?

obastemur:
If anything would use any of these methods, I would expect a sort of infinite loop anyways (as long as compiler is not doing some interesting things there)

So be it. I'm okay with with putting `assert` to the second one and removing the first.

@agarwal-sandeep @dilijev @aneeshdk Thanks for the reviews. Gonna merge after CI tests.

dilijev:
@obastemur Would you mind updating the commit message and title of this pull request to indicate the change instead of the issue-style title?

ianwjhalliday:
Does the build fail if `ModuleNamespace::Is` is removed?

dilijev:
@ianwjhalliday it does not. No code directly calls it (only polymorphically) and no tests hit that code path.

ianwjhalliday:
Actually it is not a polymorphic method since it is static and non-virtual.

I would either remove it or leave it undefined (but declared) so that it becomes a build time error if someone tries to use it.

dilijev:
@ianwjhalliday oh, my mistake. I thought it was virtual. Yes, agreed it seems that this method could be removed as long as no host like Edge is trying to call it, since there are no references in our code.

dilijev:
@ianwjhalliday @obastemur I like the declaration-only -> build-time error approach.

ianwjhalliday:
It is not part of the internal public API.  I also verified that `ModuleNamespace` is only referenced within ChakraCore via wincode

dilijev:
@ianwjhalliday sounds good to remove entirely, then

obastemur:
Don't know the interface or what it does. Not rushing on merge. I also prefer removing that method in case there is no consumer. Build error approach also sounds good. No strong opinion as long as we take care of these.

ianwjhalliday:
I suggest the build error approach because anything that inherits from `DynamicObject` (or also `RecyclableObject`) inherits a static `Is` method.

If it was removed you could still call `MethodNamespace::Is()` but it would just be an alias for `DynamicObject::Is()`.

I've caught myself making this mistake before, thinking I was calling an Is method testing for a derived class type, but it was just doing a DynamicObject::Is().

This code is IE/Edge specific and looks like it won't be touched or used much at all anymore, so it doesn't really matter.

obastemur:
@agarwal-sandeep @dilijev @aneeshdk @ianwjhalliday Thanks for the discussion. Removing both methods as discussed.

dilijev:
LGTM

ianwjhalliday:
:shipit:

PR #1238, opened 2016-07-05
JSRT context lifetime management change

Yongqu:
In JSRT we are using JSRTContext to control the lifetime of all the recycler objects in the scriptcontext.  Due to GC object finalize and sweep behavior, so non-finalizable object could be swept potentially before  some finalizer function access those fields. The change is to remove the codepath where we touch those  non-finalizable fields in javascriptlibrary in finalizer functions.

### DIFF
--- a/lib/Runtime/Base/ScriptContext.cpp
+++ b/lib/Runtime/Base/ScriptContext.cpp
@@ -81,6 +81,7 @@ namespace Js
         sourceSize(0),
         deferredBody(false),
         isScriptContextActuallyClosed(false),
+        isFinalized(false),
         isInvalidatedForHostObjects(false),
         fastDOMenabled(false),
         directHostTypeId(TypeIds_GlobalObject),
@@ -359,6 +360,7 @@ namespace Js
 
     ScriptContext::~ScriptContext()
     {
+        Assert(isFinalized);
         // Take etw rundown lock on this thread context. We are going to change/destroy this scriptContext.
         AutoCriticalSection autocs(GetThreadContext()->GetEtwRundownCriticalSection());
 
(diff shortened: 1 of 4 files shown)


Yongqu:
@digitalinfinity can you take a look? This is an old change porting to release/1.2 branch

lib/Runtime/Base/ScriptContext.cpp
```diff
@@ -377,18 +379,18 @@ namespace Js
             // clear out all inline caches to remove our proto inline caches from the thread context
             threadContext->ClearInlineCaches();
 
+            ClearInlineCaches();
```
kunalspathak: we were not clearing scriptcontext level inline caches till now?

lib/Runtime/Base/ScriptContext.cpp
```diff
@@ -377,18 +379,18 @@ namespace Js
...
 
         if (this->hasIsInstInlineCache)
         {
             // clear out all inline caches to remove our proto inline caches from the thread context
             threadContext->ClearIsInstInlineCaches();
+            ClearIsInstInlineCaches();
             Assert(!this->hasIsInstInlineCache);
         }
 
-        threadContext->UnregisterScriptContext(this);
```
kunalspathak: Is this intended?

lib/Runtime/Base/ScriptContext.cpp
Yongqu: The loop in threadContext->ClearInlineCaches will loop through and call scriptContext->ClearInlineCaches; I'm just moving the particular call out as we have removed itself from the linklist.

kunalspathak:
:shipit:

digitalinfinity:
:shipit:

PR #1239, opened 2016-07-05
Cleanup empty function and related functions

meg-gupta:
This change does the following :
- Eliminates call count profile when there is no bailout

- Eliminates stack probe for leaf functions with small stack footprint

- Eliminates arg saves on stack for a leaf function with no arg usage

### DIFF
--- a/lib/Backend/Encoder.cpp
+++ b/lib/Backend/Encoder.cpp
@@ -642,6 +642,12 @@ Encoder::ShortenBranchesAndLabelAlign(BYTE **codeStart, ptrdiff_t *codeSize)
     BYTE* buffEnd = buffStart + *codeSize;
     ptrdiff_t newCodeSize = *codeSize;
 
+    RelocList* relocList = m_encoderMD.GetRelocList();
+
+    if (relocList == nullptr)
+    {
+        return false;
+    }
 
 #if DBG
     // Sanity check
@@ -661,8 +667,6 @@ Encoder::ShortenBranchesAndLabelAlign(BYTE **codeStart, ptrdiff_t *codeSize)
         , &m_origPragmaInstrToRecordOffset
         , &m_origOffsetBuffer );
 
-    RelocList* relocList = m_encoderMD.GetRelocList();
-    Assert(relocList != nullptr);
     // Here we mark BRs to be shortened and adjust Labels and relocList entries offsets.
     uint32 offsetBuffIndex = 0, pragmaInstToRecordOffsetIndex = 0, inlineeFrameRecordsIndex = 0, inlineeFrameMapIndex = 0;
     int32 totalBytesSaved = 0;
--- a/lib/Backend/Func.cpp
+++ b/lib/Backend/Func.cpp
@@ -79,6 +79,8 @@ Func::Func(JitArenaAllocator *alloc, CodeGenWorkItem* workItem, const Js::Functi
     hasInlinee(false),
     thisOrParentInlinerHasArguments(false),
     hasStackArgs(false),
+    hasImplicitParamLoad(false),
+    hasThrow(false),
     hasNonSimpleParams(false),
     hasUnoptimizedArgumentsAcccess(false),
     hasApplyTargetInlining(false),
--- a/lib/Backend/Func.h
+++ b/lib/Backend/Func.h
@@ -263,6 +263,12 @@ static const unsigned __int64 c_debugFillPattern8 = 0xcececececececece;
         Assert(this->m_jnFunction);     // For now we always have a function body
         return this->m_jnFunction->GetHasFinally();
     }
+    bool HasThis() const
+    {
+        Assert(this->IsTopFunc());
+        Assert(this->m_jnFunction);     // For now we always have a function body
+        return this->m_jnFunction->GetHasThis();
+    }
     Js::ArgSlot GetInParamsCount() const
     {
         Assert(this->IsTopFunc());
@@ -275,7 +281,22 @@ static const unsigned __int64 c_debugFillPattern8 = 0xcececececececece;
         Assert(this->m_jnFunction);     // For now we always have a function body
         return this->m_jnFunction->GetIsGlobalFunc();
     }
-
+    bool IsGeneratorFunc() const
+    {
+        Assert(this->IsTopFunc());
+        Assert(this->m_jnFunction);     // For now we always have a function body
+        return this->m_jnFunction->IsGenerator();
+    }
+    bool IsLambda() const
+    {
+        Assert(this->IsTopFunc());
+        Assert(this->m_jnFunction);     // For now we always have a function body
+        return this->m_jnFunction->IsLambda();
+    }
+    bool IsTrueLeaf() const
+    {
+        return !GetHasCalls() && !GetHasImplicitCalls();
+    }
     RecyclerWeakReference<Js::FunctionBody> *GetWeakFuncRef() const;
     Js::FunctionBody * GetJnFunction() const { return m_jnFunction; }
 
@@ -546,6 +567,8 @@ public:
     bool                hasBailout: 1;
     bool                hasBailoutInEHRegion : 1;
     bool                hasStackArgs: 1;
+    bool                hasImplicitParamLoad : 1; // True if there is a load of CallInfo, FunctionObject
+    bool                hasThrow : 1;
     bool                hasUnoptimizedArgumentsAcccess : 1; // True if there are any arguments access beyond the simple case of this.apply pattern
     bool                m_canDoInlineArgsOpt : 1;
     bool                hasApplyTargetInlining:1;
@@ -628,6 +651,12 @@ public:
                         }
                         return isStackArgsEnabled;
     }
+
+    bool                GetHasImplicitParamLoad() const { return this->hasImplicitParamLoad; }
+    void                SetHasImplicitParamLoad() { this->hasImplicitParamLoad = true; }
+
+    bool                GetHasThrow() const { return this->hasThrow; }
+    void                SetHasThrow() { this->hasThrow = true; }
 
     bool                GetHasUnoptimizedArgumentsAcccess() const { return this->hasUnoptimizedArgumentsAcccess; }
     void                SetHasUnoptimizedArgumentsAccess(bool args)
--- a/lib/Backend/Lower.cpp
+++ b/lib/Backend/Lower.cpp
@@ -10328,6 +10328,13 @@ Lowerer::HasSideEffects(IR::Instr *instr)
     return instr->HasAnySideEffects();
 }
 
+bool Lowerer::IsArgSaveRequired(Func *func) {
+    return (!func->IsTrueLeaf() || func->IsJitInDebugMode() ||
+        // GetHasImplicitParamLoad covers generators, asmjs,
+        // and other javascript functions that implicitly read from the arg stack slots
+        func->GetHasThrow() || func->GetHasImplicitParamLoad() || func->HasThis() || func->argInsCount > 0);
+}
+
 IR::Instr*
 Lowerer::GenerateFastInlineBuiltInMathRandom(IR::Instr* instr)
 {
@@ -10639,6 +10646,7 @@ Lowerer::LoadCallInfo(IR::Instr * instrInsert)
         IR::SymOpnd * generatorSymOpnd = IR::SymOpnd::New(generatorSym, TyMachPtr, func);
         IR::RegOpnd * generatorRegOpnd = IR::RegOpnd::New(TyMachPtr, func);
         LowererMD::CreateAssign(generatorRegOpnd, generatorSymOpnd, instrInsert);
+        func->SetHasImplicitParamLoad();
 
         IR::IndirOpnd * indirOpnd = IR::IndirOpnd::New(generatorRegOpnd, Js::JavascriptGenerator::GetCallInfoOffset(), TyMachPtr, func);
         IR::Instr * instr = LowererMD::CreateAssign(IR::RegOpnd::New(TyMachPtr, func), indirOpnd, instrInsert);
--- a/lib/Backend/LowerMDShared.cpp
+++ b/lib/Backend/LowerMDShared.cpp
@@ -2060,6 +2060,7 @@ LowererMD::LoadFunctionObjectOpnd(IR::Instr *instr, IR::Opnd *&functionObjOpnd)
         instr->InsertBefore(mov1);
         functionObjOpnd = mov1->GetDst()->AsRegOpnd();
         instrPrev = mov1;
+        instr->m_func->SetHasImplicitParamLoad();
     }
     else
     {
@@ -8050,6 +8051,7 @@ LowererMD::GetImplicitParamSlotSym(Js::ArgSlot argSlot, Func * func)
 
     StackSym * stackSym = StackSym::NewParamSlotSym(argSlot, func);
     func->SetArgOffset(stackSym, (2 + argSlot) * MachPtr);
+    func->SetHasImplicitParamLoad();
     return stackSym;
 }
 
@@ -8783,6 +8785,7 @@ void LowererMD::GenerateFastInlineBuiltInCall(IR::Instr* instr, IR::JnHelperMeth
             IR::Instr *floatCall = IR::Instr::New(Js::OpCode::CALL, floatCallDst, s1, this->m_func);
             instr->InsertBefore(floatCall);
 #endif
+            instr->m_func->SetHasCalls();
             // Save the result.
             instr->m_opcode = Js::OpCode::MOVSD;
             instr->SetSrc1(floatCall->GetDst());
--- a/lib/Backend/amd64/LowererMDArch.cpp
+++ b/lib/Backend/amd64/LowererMDArch.cpp
@@ -811,6 +811,9 @@ LowererMDArch::LowerCall(IR::Instr * callInstr, uint32 argCount)
     IR::Instr *retInstr = callInstr;
     callInstr->m_opcode = Js::OpCode::CALL;
 
+    // This is required here due to calls create during lowering
+    callInstr->m_func->SetHasCalls();
+
     if (callInstr->GetDst())
     {
         IR::Opnd *       dstOpnd;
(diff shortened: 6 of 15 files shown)


lib/Backend/Func.h
```diff
@@ -263,6 +263,12 @@ static const unsigned __int64 c_debugFillPattern8 = 0xcececececececece;
         Assert(this->m_jnFunction);     // For now we always have a function body
         return this->m_jnFunction->GetHasFinally();
     }
+    bool HasThis() const
```
obastemur: nit: It could be nice to keep an empty line among the method definitions

meg-gupta:
@satheeshravi , @LouisLaf Can you please review ?

lib/Backend/Encoder.cpp
```diff
@@ -642,6 +642,12 @@ Encoder::ShortenBranchesAndLabelAlign(BYTE **codeStart, ptrdiff_t *codeSize)
     BYTE* buffEnd = buffStart + *codeSize;
     ptrdiff_t newCodeSize = *codeSize;
 
+    RelocList* relocList = m_encoderMD.GetRelocList();
+
+    if (relocList == nullptr)
+    {
+        return false;
```
dilijev: It might still be nice to keep an `Assert(false)` here so it is easier to find and debug, unless this is considered normal behavior now.

lib/Runtime/Base/Constants.h
```diff
@@ -70,6 +70,9 @@ namespace Js
         // Minimum stack required to be able to execute a JITted Javascript function.
         static const unsigned MinStackJIT = 0x930 * WIN64_STACK_FACTOR;
 
+        // Maximum stack space allowed to avoid generating ProbeCurrentStack
+        static const unsigned MaxStackForNoProbe = 0x64;
```
satheeshravi: nit: MaxStackSizeForNoProbe

lib/Backend/amd64/LowererMDArch.cpp
```diff
@@ -1643,6 +1653,12 @@ LowererMDArch::GeneratePrologueStackProbe(IR::Instr *entryInstr, IntConstType fr
     // $done:
     //
 
+    if (this->m_func->m_isLeaf && !this->m_func->GetHasImplicitCalls() &&
```
satheeshravi: nit: Good to have it in a separate function, and add a comment of what it does.

lib/Backend/amd64/LowererMDArch.h
```diff
@@ -96,6 +96,7 @@ class LowererMDArch
 
     IR::Instr *         LoadCheckedFloat(IR::RegOpnd *opndOrig, IR::RegOpnd *opndFloat, IR::LabelInstr *labelInline, IR::LabelInstr *labelHelper, IR::Instr *instrInsert, const bool checkForNullInLoopBody = false);
 
+    static bool         IsArgSaveRequired(Func *func);
```
satheeshravi: Other archs ? - x86, arm32, arm64(just declaration)

lib/Backend/Encoder.cpp
meg-gupta: This change leaves applicable functions with zero labels, making relocList null, which is the new normal.

lib/Backend/Func.h
meg-gupta: Was sticking to the neighboring style in the source file.

lib/Backend/Encoder.cpp
dilijev: Sounds good :)

meg-gupta:
Spoke to @digitalinfinity offline about call count profile. We cannot eliminate it because of its use in expiring cold functions. Updated PR.

lib/Backend/Func.h
obastemur: > Was sticking to the neighboring style in the source file.

Saw that (that is why `nit` :) )

lib/Backend/amd64/LowererMDArch.cpp
```diff
@@ -11,6 +11,15 @@ const Js::OpCode LowererMD::MDExtend32Opcode = Js::OpCode::MOVSXD;
 
 extern const IRType RegTypes[RegNumCount];
 
+bool
+LowererMDArch::IsArgSaveRequired(Func *func) {
+    return (func->IsJitInDebugMode() || func->IsGeneratorFunc() ||
+        func->GetJnFunction()->GetIsAsmjsMode() || func->IsLambda() ||
```
LouisLaf: Is there a reason to not do this for asm.js?  I think that after my change, we would never need to emit them for asm.js (even non-leafs).

lib/Backend/amd64/LowererMDArch.cpp
```diff
@@ -1643,6 +1653,12 @@ LowererMDArch::GeneratePrologueStackProbe(IR::Instr *entryInstr, IntConstType fr
     // $done:
     //
 
+    if (this->m_func->IsTrueLeaf() &&
```
LouisLaf: I think we should be fine with IsLeaf here.  If a helper has an implicit call and calls JS code, it will probe again.

satheeshravi:
nit: Can you also add a before and after snapshot(codegen) for an empty function, in the PR description ? - Just to give an idea of what else is remaining and what is cleaned.

lib/Backend/amd64/LowererMDArch.cpp
satheeshravi: @LouisLaf : What if the calling JS code is also small and doesn't probe ?

---
In reply to: [69667255](https://github.com/Microsoft/ChakraCore/pull/1239#discussion_r69667255) [](ancestors = 69667255)

LouisLaf: The CallFunction() call invoking the JS code will have a probe.

satheeshravi: Sure.

---
In reply to: [69773959](https://github.com/Microsoft/ChakraCore/pull/1239#discussion_r69773959) [](ancestors = 69773959)

lib/Backend/amd64/LowererMDArch.cpp
```diff
@@ -1643,6 +1653,12 @@ LowererMDArch::GeneratePrologueStackProbe(IR::Instr *entryInstr, IntConstType fr
...
+        frameSize - Js::Constants::MinStackJIT < Js::Constants::MaxStackSizeForNoProbe)
+    {
+        return;
+    }
```
satheeshravi: Please add comment on why we decide to skip probing stack. Also, you can consider moving this code above the existing comments - because the existing comment is for the case where we generate ProbeCurrentStack

lib/Backend/amd64/LowererMDArch.cpp
meg-gupta: Asmjs functions seem to be reading from rcx and rdx saves on stack. GlobalObject and InlineCache references are passed via these registers.

LouisLaf: I see.  After my change, we can probably remove this as well then.

---
In reply to: [69825951](https://github.com/Microsoft/ChakraCore/pull/1239#discussion_r69825951) [](ancestors = 69825951)

lib/Backend/Func.h
```diff
@@ -275,7 +281,23 @@ static const unsigned __int64 c_debugFillPattern8 = 0xcececececececece;
...
+    }
+    bool IsLambda() const
+    {
+        Assert(this->IsTopFunc());
+        Assert(this->m_jnFunction);     // For now we always have a function body
+        return this->m_jnFunction->IsLambda();
+    }
+    bool IsTrueLeaf() const
+    {
+        if (m_isLeaf && !GetHasImplicitCalls()) return true;
```
satheeshravi: nit: you can just consider returning the expression without if.

lib/Backend/amd64/LowererMDArch.cpp
```diff
@@ -11,6 +11,16 @@ const Js::OpCode LowererMD::MDExtend32Opcode = Js::OpCode::MOVSXD;
 
 extern const IRType RegTypes[RegNumCount];
 
+bool
+LowererMDArch::IsArgSaveRequired(Func *func) {
+    return (func->IsJitInDebugMode() ||
+        func->GetJnFunction()->GetIsAsmjsMode() ||
+        func->IsGeneratorFunc() || func->IsLambda() ||
+        func->GetHasFunctionObjectLoad() ||
+        !func->IsTrueLeaf() || func->GetHasArgumentObject() ||
```
satheeshravi: Why is GetHasArgumentsObject() required ?

lib/Runtime/Language/InterpreterLoop.inl
```diff
@@ -166,7 +166,7 @@ Var Js::InterpreterStackFrame::INTERPRETERLOOPNAME()
     }
 
     Assert(this->returnAddress != nullptr);
-    AssertMsg(m_arguments == NULL || Js::ArgumentsObject::Is(m_arguments), "Bad arguments!");
```
satheeshravi: Instead of removing this assert, you should be checking if there is arguments usage in the function body, if so then check the validity of it. 
You can make use of m_functionBody to see if the function body has arguments object or not.

lib/Backend/amd64/LowererMDArch.cpp
meg-gupta: Its not required, my bad. Will update the diff.

lib/Runtime/Language/InterpreterLoop.inl
meg-gupta: I intended to do that. But looks like FunctionBody has no api for hasArgumentObject. I did not add a bool on FunctionBody as it is expensive.

meg-gupta: Looks like with @rajatd's Inlining in jitted loop bodies change, we do have this info on FunctionBody. will update the diff.

lib/Backend/LowerMDShared.cpp
```diff
@@ -2060,6 +2060,7 @@ LowererMD::LoadFunctionObjectOpnd(IR::Instr *instr, IR::Opnd *&functionObjOpnd)
         instr->InsertBefore(mov1);
         functionObjOpnd = mov1->GetDst()->AsRegOpnd();
         instrPrev = mov1;
+        instr->m_func->SetHasImplicitParamLoad();
```
satheeshravi: >SetHasImplicitParamLoad [](start = 23, length = 23)

You might want to add this setter in arm equivalent API for completeness, though you are not really optimizing for ARM

lib/Backend/amd64/LowererMDArch.cpp
```diff
@@ -1379,7 +1388,7 @@ LowererMDArch::LowerEntryInstr(IR::EntryInstr * entryInstr)
     //
     uint32 argSlotsForFunctionsCalled = this->m_func->m_argSlotsForFunctionsCalled;
     // Stack is always reserved for at least 4 parameters.
-    if (argSlotsForFunctionsCalled < 4)
+    if (argSlotsForFunctionsCalled < 4 && IsArgSaveRequired(this->m_func))
         argSlotsForFunctionsCalled = 4;
```
satheeshravi: >  argSlotsForFunctionsCalled = 4; [](start = 6, length = 33)

nit nit nit: since you touched this code, do you mind wrapping it with {}  :)

lib/Backend/Func.h
```diff
@@ -546,6 +567,7 @@ static const unsigned __int64 c_debugFillPattern8 = 0xcececececececece;
     bool                hasBailout: 1;
     bool                hasBailoutInEHRegion : 1;
     bool                hasStackArgs: 1;
+    bool                hasImplicitParamLoad : 1;
```
satheeshravi: >hasImplicitParamLoad [](start = 24, length = 20)

nit: hasImplicitParamLoadFromStack. You can add a comment, that is indicates the load of callInfo and FunctionObject from stack.

lib/Backend/amd64/LowererMDArch.cpp
```diff
@@ -1379,7 +1388,7 @@ LowererMDArch::LowerEntryInstr(IR::EntryInstr * entryInstr)
     //
     uint32 argSlotsForFunctionsCalled = this->m_func->m_argSlotsForFunctionsCalled;
     // Stack is always reserved for at least 4 parameters.
-    if (argSlotsForFunctionsCalled < 4)
+    if (argSlotsForFunctionsCalled < 4 && IsArgSaveRequired(this->m_func))
```
meg-gupta: The initialization value for argSlotsForFunctionCalled is read from the Func constructor which sets it to 0.

lib/Backend/amd64/LowererMDArch.cpp
```diff
@@ -822,6 +830,7 @@ LowererMDArch::LowerCall(IR::Instr * callInstr, uint32 argCount)
 
         if (callInstr->GetSrc1()->IsHelperCallOpnd())
         {
+            callInstr->m_func->SetHasJitCalls();
```
satheeshravi: > callInstr->m_func->SetHasJitCalls(); [](start = 11, length = 37)

You should move this to Lower.cpp. You can check for the CALL opcode and call the setter. This way, you can avoid duplicated code across architectures.

lib/Backend/Func.cpp
```diff
@@ -79,6 +79,9 @@ Func::Func(JitArenaAllocator *alloc, CodeGenWorkItem* workItem, const Js::Functi
     hasInlinee(false),
     thisOrParentInlinerHasArguments(false),
     hasStackArgs(false),
+    hasImplicitParamLoad(false),
+    hasThrow(false),
+    hasJitCalls(false),
```
satheeshravi: >hasJitCalls [](start = 4, length = 11)

if isLeaf is not being used anywhere beyond this phase, do you just want to use and modify the same bool ?

lib/Backend/Lower.cpp
```diff
@@ -7616,6 +7616,7 @@ Lowerer::LowerUnaryHelperMem(IR::Instr *instr, IR::JnHelperMethod helperMethod,
     IR::Instr *instrPrev;
 
     instrPrev = LoadScriptContext(instr);
+    instr->m_func->SetHasThrow();
```
satheeshravi: > instr->m_func->SetHasThrow(); [](start = 3, length = 30)

LowerUnaryHelperMem is a generic API which could be called for non-throw opcodes as well. 
You should have this under case::Throw:

lib/Backend/amd64/LowererMDArch.cpp
```diff
@@ -11,6 +11,14 @@ const Js::OpCode LowererMD::MDExtend32Opcode = Js::OpCode::MOVSXD;
 
 extern const IRType RegTypes[RegNumCount];
 
+bool
+LowererMDArch::IsArgSaveRequired(Func *func) {
+    return (!func->IsTrueLeaf() || func->GetHasJitCalls() ||
```
satheeshravi: >GetHasJitCalls [](start = 41, length = 14)

Should go inside IsTrueLeaf. But if you reuse isLeaf to take care of this, this API itself is not required then.

lib/Backend/arm/LowerMD.cpp
```diff
@@ -243,6 +243,11 @@ LowererMD::LowerCall(IR::Instr * callInstr, Js::ArgSlot argCount)
     IR::Opnd *targetOpnd = callInstr->GetSrc1();
     AssertMsg(targetOpnd, "Call without a target?");
 
+    if (callInstr->GetSrc1()->IsHelperCallOpnd())
```
satheeshravi: >if (callInstr->GetSrc1()->IsHelperCallOpnd()) [](start = 4, length = 45)

why do you need this ? why can't we always set the flag to true ?

lib/Backend/amd64/LowererMDArch.cpp
```diff
@@ -1283,6 +1293,7 @@ LowererMDArch::GenerateStackAllocation(IR::Instr *instr, uint32 size)
                 this->m_func);
 
             instr->InsertAfter(movHelperAddrInstr);
+            instr->m_func->SetHasCalls();
```
satheeshravi: >nstr->m_func->SetHasCalls(); [](start = 13, length = 28)

shouldn't LowerCall above already take care of this ?

lib/Backend/amd64/LowererMDArch.cpp
```diff
@@ -11,6 +11,13 @@ const Js::OpCode LowererMD::MDExtend32Opcode = Js::OpCode::MOVSXD;
 
 extern const IRType RegTypes[RegNumCount];
 
+bool
+LowererMDArch::IsArgSaveRequired(Func *func) {
```
satheeshravi: >IsArgSaveRequired [](start = 15, length = 17)

Actually there is nothing machine dependent in this API. So you can consider moving this to Lower.cpp

satheeshravi:
:shipit:

lib/Backend/amd64/LowererMDArch.cpp
meg-gupta: For getter functions, the CallGetter() doesn't probe the stack, we will need isTrueLeaf() due to this.

lib/Backend/Lower.cpp
```diff
@@ -2570,6 +2570,7 @@ Lowerer::LowerRange(IR::Instr *instrStart, IR::Instr *instrEnd, bool defaultDoFa
         case Js::OpCode::Throw:
         case Js::OpCode::InlineThrow:
         case Js::OpCode::EHThrow:
+            instr->m_func->SetHasThrow();
```
LouisLaf: It might be good to set these earlier in the JIT as other people might want to use these.  I'd set this one in the initial backwards phase, after FgBuild has eliminated the true unreachable code.

lib/Backend/Lower.cpp
```diff
@@ -10328,6 +10329,13 @@ Lowerer::HasSideEffects(IR::Instr *instr)
     return instr->HasAnySideEffects();
 }
 
+
+bool Lowerer::IsArgSaveRequired(Func *func) {
+    return (!func->IsTrueLeaf() || func->IsJitInDebugMode() ||
+        func->GetJnFunction()->GetIsAsmjsMode() || func->IsGeneratorFunc() || func->IsLambda() ||
```
LouisLaf: >GetIsAsmjsMode [](start = 31, length = 14)

Can't remember why this is needed for asm.js, but please add a comment to explain.  Do the same for generators/lambda/HasThrow as well.

lib/Backend/Lower.cpp
```diff
@@ -10665,6 +10673,7 @@ Lowerer::LoadCallInfo(IR::Instr * instrInsert)
         StackSym * srcSym = LowererMD::GetImplicitParamSlotSym(1, func);
         srcOpnd = IR::SymOpnd::New(srcSym, TyMachReg, func);
     }
+    func->SetHasImplicitParamLoad();
```
LouisLaf: Should this go in GetImplicitParamSlotSym?

LouisLaf:
:shipit:

PR #1241, opened 2016-07-05
Remove starting and ending newline character from function display string

agarwal-sandeep:
Chakra  "\nfunction(){\n  [native code]\n}\n"
Firefox "function (){\n}"
Chrome "function () {}"

For arguments.callee.__proto__.toString() (or any native function toString) Chakra shows form feed character (\012) at begining and end whereas Firefox and Chrome doesn't.

VSCode removes the newline after function but doesn't remove the starting so it shows up in locals display.

### DIFF
--- a/lib/Runtime/Library/JavascriptFunction.cpp
+++ b/lib/Runtime/Library/JavascriptFunction.cpp
@@ -144,9 +144,9 @@ namespace Js
     static char16 const genFuncName[] = _u("function* anonymous");
     static char16 const asyncFuncName[] = _u("async function anonymous");
     static char16 const openFormals[] = _u("(");
-    static char16 const closeFormals[] = _u("\012)");
+    static char16 const closeFormals[] = _u("\n)");
     static char16 const openFuncBody[] = _u(" {");
-    static char16 const closeFuncBody[] = _u("\012}");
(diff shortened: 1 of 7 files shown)


ljharb:
Thanks, this has long been the case for IE as well, and it'll be good to have some coherence across browsers :-)

akroshg:
:shipit:

jianchun:
Any reason why not use `\n` there?

agarwal-sandeep:
@jianchun This has been like this from very beginning and it seems we use \012 at other places as well like closeFormals. @bterlson to see if this is something related to spec though I couldn't find anything related

jianchun:
@agarwal-sandeep Thanks for look! I wondered why we used `\012` instead of `\n` every time I saw them.

dilijev:
@jianchun do you mean we render them as literal text `\n` or `\012` or do we render them as newlines?

jianchun:
@dilijev No problem for rendering. The source code uses `\012` where we'd usually use `\n`. I think they are equivalent, and `\012` is less apparent than `\n`.

dilijev:
@jianchun yes they are equivalent. I agree `\n` is more apparent.

PR #1243, opened 2016-07-06
Wasm: deferred parsing

Cellule:
Implement deferred parsing for Wasm functions under a flag `-on:wasmdeferred`
Added new class WasmModuleGenerator which reads only what is necessary to create the wasm module. 
Then WasmBytecodeGenerator only generates the bytecode for the function bodies.
Simplified logic for LazyTraps
Added instrumentation for `-profile` with Wasm

Took the opportunity to do some cleanup in the Wasm classes notably some unused methods and some rearrangement.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/1243)


Cellule:
This change is mostly related to https://github.com/WebAssembly/design/pull/719
Regardless of the outcome of that proposal, I think this refactoring will come in handy for the new js api

Cellule:
@MikeHolman @commonlisp Can you review this please?

Cellule:
ping

lib/Runtime/Base/ScriptContext.cpp
```diff
@@ -2011,6 +2035,72 @@ if (!sourceList)
...
+            push eax
+            call ScriptContext::WasmDeferredParseEntryPoint
+#ifdef _CONTROL_FLOW_GUARD
+            // verify that the call target is valid
+            mov  ecx, eax
+            call[__guard_check_icall_fptr]
+            mov eax, ecx
+#endif
+            pop ebp
+            // Although we don't restore ESP here on WinCE, this is fine because script profiler is not shipped for WinCE.
```
MikeHolman: what does this comment mean?

lib/WasmReader/WasmByteCodeGenerator.cpp
```diff
@@ -8,56 +8,101 @@
...
+
+typedef bool(*SectionProcessFunc)(WasmModuleGenerator*);
+typedef void(*AfterSectionCallback)(WasmModuleGenerator*);
+
+WasmModuleGenerator::WasmModuleGenerator(Js::ScriptContext * scriptContext, Js::Utf8SourceInfo * sourceInfo, byte* binaryBuffer, uint binaryBufferLength) :
+    m_sourceInfo(sourceInfo),
+    m_scriptContext(scriptContext),
+    m_alloc(scriptContext->GetThreadContext()->GetThreadAlloc())
+{
+    m_reader = Anew(m_alloc, Binary::WasmBinaryReader, m_alloc, binaryBuffer, binaryBufferLength);
```
MikeHolman: probably better to recycler allocate this to have same lifetime as AsmJsFunctionInfo (also recycler allocated)

MikeHolman: also, since this data structure is living forever now, we may want to think about merging some of this with asm.js info (e.g. parameters, return types, etc.), which is now redundantly stored

---
In reply to: [71214068](https://github.com/Microsoft/ChakraCore/pull/1243#discussion_r71214068) [](ancestors = 71214068)

MikeHolman:
right now this will have memory cost, so a bit hesitant about merging unless it will be used

lib/Runtime/Base/ScriptContext.cpp
Cellule: I got this from another thunk in this file, I don't really know what it means exactly either

lib/WasmReader/WasmByteCodeGenerator.cpp
```diff
@@ -8,56 +8,101 @@
...
+
+typedef bool(*SectionProcessFunc)(WasmModuleGenerator*);
+typedef void(*AfterSectionCallback)(WasmModuleGenerator*);
+
+WasmModuleGenerator::WasmModuleGenerator(Js::ScriptContext * scriptContext, Js::Utf8SourceInfo * sourceInfo, byte* binaryBuffer, uint binaryBufferLength) :
+    m_sourceInfo(sourceInfo),
+    m_scriptContext(scriptContext),
+    m_recycler(scriptContext->GetRecycler())
+{
+    m_reader = RecyclerNew(m_recycler, Binary::WasmBinaryReader, scriptContext, binaryBuffer, binaryBufferLength);
```
MikeHolman: I think this can be leaf.

MikeHolman:
:shipit:

lib/WasmReader/WasmByteCodeGenerator.cpp
```diff
@@ -237,7 +237,7 @@ Wasm::WasmFunction * WasmModuleGenerator::GenerateFunctionHeader(uint32 index)
     body->SetIsWasmFunction(true);
     body->GetAsmJsFunctionInfo()->SetIsHeapBufferConst(true);
 
-    WasmReaderInfo* readerInfo = RecyclerNew(m_recycler, WasmReaderInfo);
+    WasmReaderInfo* readerInfo = RecyclerNewLeaf(m_recycler, WasmReaderInfo);
```
MikeHolman: i think this can't be leaf

lib/WasmReader/WasmByteCodeGenerator.cpp
```diff
@@ -91,7 +91,7 @@ Wasm::WasmModule* WasmModuleGenerator::GenerateModule()
     m_sourceInfo->EnsureInitialized(0);
     m_sourceInfo->GetSrcInfo()->sourceContextInfo->EnsureInitialized();
 
-    m_module = RecyclerNew(m_recycler, WasmModule);
+    m_module = RecyclerNewLeaf(m_recycler, WasmModule);
```
MikeHolman: i think this can't be leaf

PR #1246, opened 2016-07-06
Fixing UT Failure in DisableJIT

satheeshravi:
Disabling nonative mode for the failing UT, which verifies the trace during the optimization in the backend.

### DIFF
--- a/test/Function/rlexe.xml
+++ b/test/Function/rlexe.xml
@@ -381,7 +381,7 @@
     <default>
       <files>StackArgsWithFormals.js</files>
       <compile-flags>-mic:1 -off:simpleJit -trace:stackargformalsopt</compile-flags>
-      <tags>exclude_dynapogo,exclude_ship, exclude_fre</tags>
+      <tags>exclude_dynapogo,exclude_ship,exclude_fre,exclude_nonative,require_backend</tags>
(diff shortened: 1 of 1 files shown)


dilijev:
@dotnet-bot test nojit tests please

test/Function/rlexe.xml
```diff
@@ -381,7 +381,7 @@
     <default>
       <files>StackArgsWithFormals.js</files>
       <compile-flags>-mic:1 -off:simpleJit -trace:stackargformalsopt</compile-flags>
-      <tags>exclude_dynapogo,exclude_ship, exclude_fre</tags>
+      <tags>exclude_dynapogo,exclude_ship, exclude_fre, exclude_nonative</tags>
```
dilijev: Actually I believe you cannot put spaces after commas in order for RL to pick up the tags correctly. The existing space would have been wrong as well. I ran into this a while back. It will just ignore anything after the spaces.

Additionally, in order to actually exclude this from DisableJIT mode you'll need to add `require_backend`.

satheeshravi: Done. Updated PR.

dilijev:
LGTM :shipit:

PR #1247, opened 2016-07-06
Capture values for bailout incrementally across blocks

ThomsonTan:
Capture values for bailout are built incrementally by selecting captured values in one of its predecessor block and mark all not uncommon symbols from all predecessor blocks as recapture when building new bailout.


lib/Backend/InlineeFrameInfo.h
```diff
@@ -14,6 +14,7 @@ struct BailoutConstantValue {
     };
     void InitVarConstValue(Js::Var value);
     void InitFloatConstValue(FloatConstType value) { this->type = TyFloat64; this->u.floatConst.value = value; }
+    bool operator==(const BailoutConstantValue & bailoutConstValue);
```
LouisLaf: I don't like operator overloading in this way as I hides the true cost and meaning of the operation.  An IsEqual() function accomplishes the same but I can easy browse to the definition now to see what the code is actually doing.

LouisLaf:
Sorry, there was no @LouisLaf here which is why it took me a while to get to it...

lib/Backend/GlobOptBailOut.cpp
```diff
@@ -951,7 +957,7 @@ GlobOpt::FillBailOutInfo(BasicBlock *block, BailOutInfo * bailOutInfo)
                 {
                     sym = opnd->GetStackSym();
                     Assert(FindValue(sym));
-                    this->blockData.capturedArgs->Set(sym->m_id);
+                    this->blockData.changedSyms->Set(sym->m_id);
```
LouisLaf: >sym [](start = 53, length = 3)

Why?  What is changing about sym here?  This at a minimum needs a comment.

ThomsonTan: Args need to be captured even if without setting, will add a comment.

---
In reply to: [71631271](https://github.com/Microsoft/ChakraCore/pull/1247#discussion_r71631271) [](ancestors = 71631271)

lib/Backend/GlobOpt.cpp
```diff
@@ -1829,6 +1826,58 @@ GlobOpt::CloneValues(BasicBlock *const toBlock, GlobOptBlockData *toData, GlobOp
...
 }
 
+template <typename CapturedList, typename CapturedItemsAreEqual>
+void
+GlobOpt::MergeCapturedValues(
+    GlobOptBlockData * toData,
+    SListBase<CapturedList> * toList,
+    SListBase<CapturedList> * fromList,
+    CapturedItemsAreEqual itemsAreEqual)
+{
```
LouisLaf: Add comment explaining how this code works and why you can't edit the final list.

LouisLaf:
:shipit:

PR #1248, opened 2016-07-06
Fixed memory leak in invalidation code of inlineCache

kunalspathak:
When we register proto/store field inline cache in thread context, we store it in a map of propertyId to inlineCacheList. When the prototype of object is changed and if object had more than 128 properties, we invalidate all the inline caches present in the map. However we simply reset the buckets and entries in the map. Instead we should also delete them and return them back to arena allocator.
Below sample code demonstrates memory leak for invalidation of proto inline cache. Before my change private bytes of process was approx. 200MB. After this change it comes down to 150MB.
```js
var noOfStickers = 1000;
function generateObject() {
  var obj = {};
  for (sticker = 0; sticker < noOfStickers; sticker++) {
    for (var i = 97; i < 123 ; i++) {
      obj[String.fromCharCode(i) + "_" + sticker] = i;
    }
  }
  return obj;
}

function loadProperties(o) {
  var x = 0;
  for (sticker = 0; sticker < noOfStickers; sticker++) {
    for (var i = 97; i < 123 ; i++) {
      var propName = String.fromCharCode(i);
      x += eval("o." + propName + "_" + sticker);
    }
  }
  return x;
}

for (var i = 0; i < 100; i++) {
  var obj = generateObject();
  var t1 = Object.create(obj);
  // Make sure t1 is prototype of something
  // else my changePrototype optimization would
  // kick in and we would skip invalidation, hiding 
  // the memory leak
  var r1 = Object.create(t1);
  loadProperties(t1);
  Object.setPrototypeOf(t1, {});
}
print('waiting...');
// Check private working set of process
while (true) {

}
```

### DIFF
--- a/lib/Runtime/Base/ThreadContext.cpp
+++ b/lib/Runtime/Base/ThreadContext.cpp
@@ -3223,7 +3223,7 @@ ThreadContext::InvalidateAllProtoInlineCaches()
     {
         InvalidateAndDeleteInlineCacheList(inlineCacheList);
     });
-    protoInlineCacheByPropId.ResetNoDelete();
+    protoInlineCacheByPropId.Reset();
 }
 
 #if DBG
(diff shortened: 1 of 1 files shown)


kunalspathak:
@Yongqu , @pleath - Please review!

Yongqu:
LGTM

Issue #85, opened 2016-01-14
ChakraCore NuGet package


ormico:
I didn't see this already so I hope it isn't a duplicate.
I also didn't see it in the roadmap.

I read the quick-start guide on embedding ChakraCore, but following this guidance will make it hard to keep an embedded version up to date. 
https://github.com/Microsoft/ChakraCore/wiki/Embedding-ChakraCore

A better solution would be if MS created a NuGet package for ChakraCore and auto-published it from its build server.

---

# Maintainers' Update

@dilijev (added 2016-04-26 / last updated 2016-01-31):

Wiki: https://github.com/Microsoft/ChakraCore/wiki/NuGet-Packages
NuGet feed: https://www.nuget.org/packages/Microsoft.ChakraCore
MyGet feed: https://www.myget.org/feed/chakracore-preview/package/nuget/Microsoft.ChakraCore

### Actionable tasks
- [x] Use build system to create and publish daily builds as NuGet packages to MyGet.org
- [x] Add link to MyGet feed to README and/or Wiki (@liminzhu) (#1253)
- [x] Create a NuGet.org feed to publish our official releases (once they are available)
- [x] _When available:_ Add NuGet.org feed link to README and/or Wiki (@liminzhu)
- [x] _When release 1.2 is finalized:_ Publish ChakraCore version 1.2 to NuGet.org
- [x] Include build number in preview package special version strings
  - Format for preview packages: `Microsoft.ChakraCore.<major>.<minor>.<patch>-preview-<build1>-<build2>` (e.g. `Microsoft.ChakraCore.1.3.0-preview-00008-12345`). Commit hash is, as always, encoded in binary metadata.
- ~~Preview builds of multiple branches: e.g. `release/*` and `master`~~ (Infrastructure now exists but there's no compelling reason to publish preview builds of multiple release branches -- patch updates to already-released minor versions like 1.2 and 1.3 will be extremely minor so there's nothing meaningful to preview.)
- [x] Reorganize package layout to enable Visual C++ projects to take a dependency on native NuGet packages and automatically integrate the ChakraCore binaries into a project. (see https://docs.nuget.org/consume/support-for-native-projects) (#2266 by @pre10der89)
- [x] NuGet package spec (nuspec) consumable by .NET projects (#2266 by @pre10der89)
- [x] Publish preview package for .NET for validation (https://github.com/Microsoft/ChakraCore/issues/85#issuecomment-276537774)
[...]

PR #1253, opened 2016-07-07
Add myget pre-built binaries info on README

liminzhu:
Add myget pre-built binaries info on README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,6 +85,8 @@ You can build ChakraCore on Windows 7 SP1 or above, and Windows Server 2008 R2 o
 
 More details in [Building ChakraCore](https://github.com/Microsoft/ChakraCore/wiki/Building-ChakraCore).
 
+Alternatively, you can get pre-built ChakraCore binaries from our [NuGet Packages](https://github.com/Microsoft/ChakraCore/wiki/NuGet-Packages).
+
 ## Using ChakraCore
 
 Once built, you have a few options for how you can use ChakraCore:
@@ -109,7 +111,7 @@ This project has adopted the [Microsoft Open Source Code of Conduct](https://ope
 
 Please refer to [Contribution guidelines](CONTRIBUTING.md) for more details.
 
-### License
+## License
 
 Code licensed under the [MIT License](https://github.com/Microsoft/ChakraCore/blob/master/LICENSE.txt).
 


liminzhu:
@dilijev

dilijev:
Looks fine to me, though I wonder if we can/should add more info about how to get the binaries from those packages. I assume most people don't know they are .zip files with a different extension.

@tcare How do we deal with NuGet packages as a means to distribute pre-built binaries, when they don't follow the NuGet package protocol.

Refer to #85 for work items to improve this.

README.md
```diff
@@ -77,6 +77,8 @@ You can build ChakraCore on Windows 7 SP1 or above, and Windows Server 2008 R2 o
 
 More details in [Building ChakraCore](https://github.com/Microsoft/ChakraCore/wiki/Building-ChakraCore).
 
+Alternatively, you can get pre-built ChakraCore binaries on [myget](https://www.myget.org/feed/chakracore-preview/package/nuget/Microsoft.ChakraCore).
```
dilijev: Change to `pre-built ChakraCore NuGet packages with our latest binaries from [MyGet]`?

tcare:
Agree with Doug on basic instructions to get out the binaries.

As for Doug's concern, I would put a note that it's not in the standard layout and we'll fix that soon.

dilijev:
Maybe we could add a Wiki page about NuGet packages and link to it from the README. Instead of specifically mentioning the MyGet host from the README, we can just say NuGet packages are available and put the links in the Wiki page

This way, we can reduce the amount of content and churn in the README for this purpose. In the short-term, we could explain the structure and how to extract the binaries from the NuGet packages hosted on MyGet. Then, when we fix up the directory structure of the packages we can provide more info on how to integrate those dependencies and develop against them. Later, when we have an official NuGet feed, we can link to that as well.

I'll draft a PR to the Wiki about this and then we can update this PR to point to the right page.

liminzhu:
SGTM :D.

README.md
dilijev: In light of https://github.com/Microsoft/ChakraCore-wiki/pull/3 recommend changing this to:

`Alternatively, you can get pre-built ChakraCore binaries from our [NuGet Packages](https://github.com/Microsoft/ChakraCore/wiki/NuGet-Packages).`

dilijev:
@liminzhu FYI I merged in https://github.com/Microsoft/ChakraCore-wiki/pull/3 so you can now link to   https://github.com/Microsoft/ChakraCore/wiki/NuGet-Packages

Issue #1228, opened 2016-07-01
default constructor creation for derived class is not spec compliant


kunalspathak:
As per [spec](https://tc39.github.io/ecma262/#sec-runtime-semantics-classdefinitionevaluation), if derived class doesn't have constructor, it should create a constructor something like this -
```js
constructor(... args){ super (...args);}
```
Instead we do something like this -
```js
constructor(){ super (...arguments);}
```
`...arguments` part loads argumentsObject on heap which is not performance efficient.

kunalspathak:
FYI - @akroshg , @suwc

ianwjhalliday:
Is the use of rest and spread optimized in this case or do they also add perf overhead that may be avoidable somehow?

kunalspathak:
> Is the use of rest and spread optimized in this case 

Probably. See #1230

PR #1255, opened 2016-07-08
Default constructor for derived classes

suwc:
Change default constructor to

constructor(...args) { super(...args); }

from (functionally equivalent pre-spread implementation)

constructor() { super(...arguments); }

Fixes https://github.com/Microsoft/ChakraCore/issues/1228

### DIFF
--- a/lib/Parser/Parse.cpp
+++ b/lib/Parser/Parse.cpp
@@ -6318,28 +6318,36 @@ ParseNodePtr Parser::GenerateEmptyConstructor(bool extends)
     ParseNodePtr pnodeFncSave = m_currentNodeFunc;
     m_currentNodeFunc = pnodeFnc;
 
+    ParseNodePtr argsId = nullptr;
+    ParseNodePtr *lastNodeRef = nullptr;
     ParseNodePtr pnodeBlock = StartParseBlock<buildAST>(PnodeBlockType::Parameter, ScopeType_Parameter);
+
+    if (extends)
+    {
+        ParseNodePtr *const ppnodeVarSave = m_ppnodeVar;
+        m_ppnodeVar = &pnodeFnc->sxFnc.pnodeVars;
+
+        IdentPtr pidargs = GenerateIdentPtr(_u("args"), 4);
+        ParseNodePtr pnodeT = CreateVarDeclNode(pidargs, STFormal);
+        pnodeT->sxVar.sym->SetIsNonSimpleParameter(true);
+        pnodeFnc->sxFnc.pnodeRest = pnodeT;
+        PidRefStack *ref = this->PushPidRef(pidargs);
+
+        argsId = CreateNameNode(pidargs, pnodeFnc->ichMin, pnodeFnc->ichLim);
+
+        argsId->sxPid.symRef = ref->GetSymRef();
+        m_ppnodeVar = ppnodeVarSave;
+    }
+
     ParseNodePtr pnodeInnerBlock = StartParseBlock<buildAST>(PnodeBlockType::Function, ScopeType_FunctionBody);
     pnodeBlock->sxBlock.pnodeScopes = pnodeInnerBlock;
     pnodeFnc->sxFnc.pnodeBodyScope = pnodeInnerBlock;
     pnodeFnc->sxFnc.pnodeScopes = pnodeBlock;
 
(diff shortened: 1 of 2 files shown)


suwc:
@Microsoft/chakra-es, @kunalspathak: could you take a look pls?

lib/Parser/Parse.cpp
```diff
@@ -6318,28 +6318,36 @@ ParseNodePtr Parser::GenerateEmptyConstructor(bool extends)
...
+    ParseNodePtr argsId = nullptr;
+    ParseNodePtr *lastNodeRef = nullptr;
     ParseNodePtr pnodeBlock = StartParseBlock<buildAST>(PnodeBlockType::Parameter, ScopeType_Parameter);
-    ParseNodePtr pnodeInnerBlock = StartParseBlock<buildAST>(PnodeBlockType::Function, ScopeType_FunctionBody);
-    pnodeBlock->sxBlock.pnodeScopes = pnodeInnerBlock;
-    pnodeFnc->sxFnc.pnodeBodyScope = pnodeInnerBlock;
-    pnodeFnc->sxFnc.pnodeScopes = pnodeBlock;
 
-    ParseNodePtr *lastNodeRef = nullptr;
     if (extends)
```
akroshg: good to add comment over here on what you want to achieve and what is expected code after this change.

lib/Parser/Parse.cpp
```diff
@@ -6318,28 +6318,36 @@ ParseNodePtr Parser::GenerateEmptyConstructor(bool extends)
...
-        ParseNodePtr argumentsId = CreateNameNode(wellKnownPropertyPids.arguments, pnodeFnc->ichMin, pnodeFnc->ichLim);
-        argumentsId->sxPid.symRef = ref->GetSymRef();
-        pnodeFnc->sxFnc.SetUsesArguments(true);
-        pnodeFnc->sxFnc.SetHasReferenceableBuiltInArguments(true);
-
         ParseNodePtr *const ppnodeVarSave = m_ppnodeVar;
         m_ppnodeVar = &pnodeFnc->sxFnc.pnodeVars;
-        CreateVarDeclNode(wellKnownPropertyPids.arguments, STVariable, true, pnodeFnc)->grfpn |= PNodeFlags::fpnArguments;
+
+        IdentPtr pidargs = GenerateIdentPtr(_u("args"), 4);
```
tcare: This is an ancient API.. we should remove it. Please use m_phtbl->PidHashNameLen

lib/Parser/Parse.cpp
tcare: +1, please keep the equivalent parse text.

akroshg:
did you verify how much the scenario got improved? what is the toString behavior on class after this change? I was expecting the debugger baseline will change, aren't they? Overall looks good to me.

tcare:
@akroshg We used to fake it in ScriptFunction at some point, but looks like we no longer do.. it is observable in spec so we should verify the toString is correct.

suwc:
Right - toString is being checked here:
es6/classes.js
name: "Default constructor"


---
In reply to: [231252379](https://github.com/Microsoft/ChakraCore/pull/1255#issuecomment-231252379) [](ancestors = 231252379)

suwc:
toString() behavior hasn't changed. And there's no debugger baseline update. This change is transparent.
@kunalspathak might be able to verify the scenario. Anyhow, use of 'arguments' is discouraged, so this is desirable.

---
In reply to: [231251525](https://github.com/Microsoft/ChakraCore/pull/1255#issuecomment-231251525) [](ancestors = 231251525)

lib/Parser/Parse.cpp
suwc: Added comment.

---
In reply to: [70012406](https://github.com/Microsoft/ChakraCore/pull/1255#discussion_r70012406) [](ancestors = 70012406)

lib/Parser/Parse.cpp
suwc: Thanks - I removed this API.

---
In reply to: [70013257](https://github.com/Microsoft/ChakraCore/pull/1255#discussion_r70013257) [](ancestors = 70013257)

akroshg:
:shipit:

PR #1256, opened 2016-07-08
Native tests fixes.

akroshg:
Added a check to return value of fread. Updated the copy-right. Added jenkins support.

Note : Addressing comments on the previous PR of adding native tests. Thanks for the feedback.

### DIFF
--- a/bin/NativeTests/FileLoadHelpers.cpp
+++ b/bin/NativeTests/FileLoadHelpers.cpp
@@ -67,7 +67,13 @@ HRESULT FileLoadHelpers::LoadScriptFromFile(LPCSTR filename, LPCWSTR& contents, 
     //
     // Read the entire content as a binary block.
     //
-    fread((void*)contentsRaw, sizeof(char), lengthBytes, file);
+    size_t readBytes = fread((void*)contentsRaw, sizeof(char), lengthBytes, file);
+    if (readBytes < lengthBytes)
+    {
+        fwprintf(stderr, _u("readBytes should be equal to lengthBytes"));
+        IfFailGo(E_FAIL);
+    }
+
     fclose(file);
     *(WCHAR*)((byte*)contentsRaw + lengthBytes) = _u('\0'); // Null terminate it. Could be LPCWSTR.
 
--- a/test/jenkins.testall.cmd
+++ b/test/jenkins.testall.cmd
@@ -53,6 +53,10 @@ set _HadFailures=0
   call :runTests x86test
   call :runTests x64debug
   call :runTests x64test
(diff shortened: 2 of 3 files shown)


akroshg:
@dilijev @obastemur thanks for the feedback. Please have a look at this.

bin/NativeTests/FileLoadHelpers.cpp
```diff
@@ -67,7 +67,13 @@ HRESULT FileLoadHelpers::LoadScriptFromFile(LPCSTR filename, LPCWSTR& contents,
     //
     // Read the entire content as a binary block.
     //
-    fread((void*)contentsRaw, sizeof(char), lengthBytes, file);
+    size_t readBytes = fread((void*)contentsRaw, sizeof(char), lengthBytes, file);
+    if (readBytes < lengthBytes)
```
obastemur: I see you took this from ch/helpers. For a big file sitting on a slow drive (i.e. a drive on network) this will be most likely failing. I'm adding these both to my list. Gonna send a PR for them. For now; this one at least fails in the right place.

obastemur:
LGTM (fread related change)

bin/NativeTests/Scripts/splay.js
```diff
@@ -2,7 +2,7 @@
 // base.js
 ////////////////////////////////////////////////////////////////////////////////
 
-// Copyright 2008 the V8 project authors. All rights reserved.
+// Copyright 2013 the V8 project authors. All rights reserved.
```
dilijev: Not sure if this copyright year should be updated, if in fact the version of the file you have here is from 2008. Maybe hold off on this until you can fix the test w.r.t. the version we already have in the repo?

dilijev:
:+1: for adding Jenkins support. Thanks!

Although this makes me wonder if we could have refactored the Jenkins stuff better...

akroshg:
for Jenkins, when do we use the Jenkins.testall.cmd?

dilijev:
@akroshg We don't; it's just convenient to have for starting a local test run of all flavors (and then e.g. going for lunch). I mostly used it while I was developing the Jenkins definitions. The Jenkins scripts are totally self-contained and you can run them locally.

akroshg:
hmm, may be testall can call testone.cmd for each combination.

akroshg:
ok I'll remove the splay changes and merge in.

dilijev:
@akroshg I want to say I tried that (since it's the obvious choice) and decided against it for whatever reason. I'll look into it again at some point in the future. There's no need for all the code to be duplicated; it just needs to be refactored correctly. :)

PR #1257, opened 2016-07-08
Enabling padStart and padEnd by default on.

akroshg:
Apart from enabling it, I found that spec for padStart/padEnd got changed where ,
     8. If filler is the empty String, return S. (https://tc39.github.io/ecma262/#sec-string.prototype.padstart)
fixed that and updated the test.
test262 is fully passed for these two APIs.


akroshg:
@ianwjhalliday (or @Microsoft/chakra-es) have a look at this.

ljharb:
Thank you!

akroshg:
@ljharb thanks for mentioning that.

bummer, I need to update few baselines (just ran es7 folder).

tcare:
Thanks for adding OOM scenario test. LGTM.

PR #1258, opened 2016-07-08
Caching the arguments type for strict mode on the library

satheeshravi:
**Premise:**
For strict mode, we start with a simple dictionary type for arguments object and transition to dictionary type upon adding setter/getter. This is costly.

**Fix:**
We, now, cache the dictionary type with accessors on the javascript library - so that we don't want to do the transition everytime.

_This improves the strict mode with arguments by 40% on micro benchmarks_. Note that this will be evident only when stack arg optimization is disabled (because with optimization - we will not be creating the heap arguments object in the first place).

**Perf tests:**
Surprisingly, this code change didn't show up evidently in Speedometer - especially given the fact that there are some functions that have unoptimized heap arguments object.
I am yet to try with acme-benchmark.

**IE PerfLab: (Few overall improvements in Elapsed/CPU Time)**
http://ieperf-web-01/perfresults/Comparison?testRunId=423952&baseRunId=423947&lab=IEPERF&getRelatedRunData=false&priorities=0,1&categories=failures,warnings,improvements,no-change&search=

**Note:**
_SetAccessor API is still hot as most of the time is spent in GetPropertyName()_ -> which looks up the propertyRecord from the Dictionary. I kinda gave up at this point on further making strict mode scenario faster as the returns wasn't really fruitful. Let me know if someone has any thoughts on this.

### DIFF
--- a/lib/Runtime/Types/DictionaryTypeHandler.cpp
+++ b/lib/Runtime/Types/DictionaryTypeHandler.cpp
@@ -13,6 +13,19 @@ namespace Js
     }
 
     template <typename T>
+    DictionaryTypeHandlerBase<T>* DictionaryTypeHandlerBase<T>::CreateTypeHandlerForArgumentsInStrictMode(Recycler * recycler, ScriptContext * scriptContext)
+    {
+        DictionaryTypeHandlerBase<T> * dictTypeHandler = New(recycler, 8, 0, 0);
+        
+        dictTypeHandler->Add(scriptContext->GetPropertyName(Js::PropertyIds::caller), PropertyWritable, scriptContext);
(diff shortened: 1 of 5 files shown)


satheeshravi:
@rajatd , @LouisLaf Please take a look.

lib/Runtime/Types/DictionaryTypeHandler.cpp
```diff
@@ -369,6 +383,12 @@ namespace Js
         descriptor.IsInitialized = isInitialized;
         descriptor.IsFixed = isFixed;
         descriptor.UsedAsFixed = usedAsFixed;
+
+        if (addGetterSetter)
+        {
+            descriptor.ConvertToGetterSetter(nextPropertyIndex);
```
rajatd: where was this (or the equivalent of this) operation being done before?

PR #1259, opened 2016-07-08
Zero pages with non-temporal write instead of memset to keep cache untouched

ThomsonTan:
This is for experimenting non-temporal write when zero-ing pages in background thread, so it doesn't evict cache lines other task is using.


rajatd:
:+1:

PR #1261, opened 2016-07-09
Adding typeof support to enable Stack arg optimization

satheeshravi:
**Premise:**

acme-benchmark uses arguments[i] with typeof keyword. We do not currently do Stack Arguments optimization for this scenario.

**Change:**

Added support in the Globopt to enable Stack arg optimization when the
function body has typeof keyword.

**Details:**

We optimistically insert (BailOnStackArgsOutOfActualsRange) for
TypeOfElem, if it is a candidate for optimization.
We decide during DeadStore phase, whether this bail out is required or not
and then change the instr accordingly(as shown below).

**After GlobOpt Phase:**
_Before:_
dst = TypeOfElem arguments[i]

_With this pack:_
tmpdst = LdElemI_A arguments[i] <(BailOnStackArgsOutOfActualsRange)>
dst = TypeOf tmpdst.

**Perf:**
This will help _acme-benchmark by 3%_ as pointed out by Kunal.
Yet to run other tests.

### DIFF
--- a/lib/Backend/BackwardPass.cpp
+++ b/lib/Backend/BackwardPass.cpp
@@ -300,7 +300,9 @@ BackwardPass::ProcessBailOnStackArgsOutOfActualsRange()
 {
     IR::Instr * instr = this->currentInstr;
 
-    if (tag == Js::DeadStorePhase && instr->m_opcode == Js::OpCode::LdElemI_A && instr->HasBailOutInfo() && !IsPrePass())
+    if (tag == Js::DeadStorePhase && 
+        (instr->m_opcode == Js::OpCode::LdElemI_A || instr->m_opcode == Js::OpCode::TypeofElem) && 
+        instr->HasBailOutInfo() && !IsPrePass())
     {
         if (instr->DoStackArgsOpt(this->func))
         {
@@ -308,7 +310,7 @@ BackwardPass::ProcessBailOnStackArgsOutOfActualsRange()
             if (instr->GetBailOutKind() & ~IR::BailOnStackArgsOutOfActualsRange)
             {
                 Assert(instr->GetBailOutKind() == (IR::BailOnStackArgsOutOfActualsRange | IR::BailOutOnImplicitCallsPreOp));
-                //We are sure at this point, that we will not have any implicit calls as we don't have any call to helper.
+                //We are sure at this point, that we will not have any implicit calls as we wouldn't have done this optimization in the first place.
                 instr->SetBailOutKind(IR::BailOnStackArgsOutOfActualsRange);
             }
         }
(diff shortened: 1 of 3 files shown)


satheeshravi:
@LouisLaf @rajatd Please take a look.
//cc @kunalspathak

lib/Backend/BackwardPass.cpp
```diff
@@ -2965,6 +2974,43 @@ BackwardPass::DeadStoreOrChangeInstrForScopeObjRemoval()
...
+                        tmpdst = LdElemI_A arguments[i] <(BailOnStackArgsOutOfActualsRange)>
+                        dst = TypeOf tmpdst
+                */
+
+                AssertMsg(instr->HasBailOutInfo() && (instr->GetBailOutKind() & IR::BailOutKind::BailOnStackArgsOutOfActualsRange), "Why is the bailout kind not set, when it is StackArgOptimized?");
+
+                instr->m_opcode = Js::OpCode::LdElemI_A;
+                IR::Opnd * dstOpnd = instr->UnlinkDst();
+
+                IR::RegOpnd * elementOpnd = IR::RegOpnd::New(StackSym::New(instr->m_func), IRType::TyVar, instr->m_func);
```
rajatd: >IRType::TyVar [](start = 91, length = 13)

can we get the type for this operand by looking at the type of the array?

lib/Backend/BackwardPass.cpp
```diff
@@ -2965,6 +2974,43 @@ BackwardPass::DeadStoreOrChangeInstrForScopeObjRemoval()
...
+    IR::Instr * instr = this->currentInstr;
+    if (instr->DoStackArgsOpt(this->func))
+    {
+        switch (instr->m_opcode)
+        {
+            case Js::OpCode::TypeofElem:
+            {
+                /*
+                    Before:
+                        dst = TypeOfElem arguments[i] <(BailOnStackArgsOutOfActualsRange)>
```
rajatd: >BailOnStackArgsOutOfActualsRange [](start = 56, length = 32)

nit: we wouldn't have this bailout today, right?

satheeshravi: Yes. You are right. But "before" here doesn't indicate today's code, but it indicates the current state of the code and how it gets transformed.

---
In reply to: [70345281](https://github.com/Microsoft/ChakraCore/pull/1261#discussion_r70345281) [](ancestors = 70345281)

rajatd: ok, makes sense


---
In reply to: [70507186](https://github.com/Microsoft/ChakraCore/pull/1261#discussion_r70507186) [](ancestors = 70507186,70345281)

lib/Backend/BackwardPass.cpp
rajatd: nevermind, this is only for loads from arguments array

---
In reply to: [70345214](https://github.com/Microsoft/ChakraCore/pull/1261#discussion_r70345214) [](ancestors = 70345214)

PR #1263, opened 2016-07-11
cross-platform: Compile into module blocks

obastemur:
- Combines static libraries from sub projects. (Linux, macOS)
   ChakraCore consumer now links to 3 libs instead of 18.

 - Defines AutoSystemInfo etc. initializers builtin for ease of embedding


pal/src/CMakeLists.txt
```diff
@@ -116,9 +116,9 @@ set(SOURCES
   exception/seh.cpp
   exception/signal.cpp
   file/directory.cpp
-  file/file.cpp
+  file/pal_file.cpp
```
obastemur: The reason for the changing the file names;

Weirdly, combiner was silently overwriting files with same naming under the PAL.

jianchun:
LGTM overall. One thing I don't understand -- can CH link to the bigger combined libs instead of the "Singular" ones? It'll be nice if we can skip building those Singular libs.

obastemur:
>  can CH link to the bigger combined libs instead of the "Singular" ones? It'll be nice if we can skip building those Singular libs.

The reason it doesn't use the combined lib for shared build is that, in such case CH would be already embedding everything available from shared. There we have to deal with AutoSystemInfo initializer etc....

Yongqu:
Looks like my newly introduced dependency of std::map is causing build break in static Linux build, and your change should be able to fix that. Can you retarget the change to master so I can merge your change in? thx

obastemur:
rebased to latest.

@Yongqu unfortunately this is not for master. However, since there is no big change since the last merge we must be able merge linux to master painlessly (especially before introducing something big to linux branch) /cc @digitalinfinity  ?

obastemur:
@jianchun Thanks for the review.

PR #1272, opened 2016-07-13
Fix missing color for items in BackgroundJobProcessor::DebugThreadNames

ThomsonTan:
The missing colon makes the debug name incorrect after it.


rajatd:
:shipit:

PR #1273, opened 2016-07-13
clang: cleanup all the compiler warnings and enable werror/xcode back

obastemur:


dilijev:
@obastemur Looks like there is unreachable code in Windows debug builds

lib/Runtime/ByteCode/OpCodeUtil.cpp
```diff
@@ -161,6 +162,7 @@ namespace Js
         {
             Assert(false);
             __assume(false);
```
dilijev: Unreachable code warnings are probably because this line makes it so that control flow can't reach the following line.

obastemur: Yes I'm aware. Thanks for following up anyways!

dilijev: @obastemur no problem :D

obastemur:
Any objection / thoughts on this PR?

dilijev:
Will have to take a look later when I have a bigger block of free time. Also you should probably wait for @digitalinfinity or @jianchun to sign off on this one.

bin/ch/WScriptJsrt.cpp
```diff
@@ -808,8 +808,10 @@ WScriptJsrt::CallbackMessage::~CallbackMessage()
     {
         WScriptJsrt::PrintException("", JsErrorScriptException);
     }
+CLANG_WNO_BEGIN("-Wunused-variable")
     JsErrorCode errorCode = ChakraRTInterface::JsRelease(m_function, nullptr);
```
tcare: Can we just change this to Assert(ChakraRTInterface::JsRelease(m_function, nullptr) == JsNoError) ?

lib/Parser/Scan.h
```diff
@@ -367,10 +367,13 @@ class Scanner : public IScanner, public EncodingPolicy
     {
         return HeapNewNoThrow(Scanner, parser, phtbl, ptoken, perr, scriptContext);
     }
+
+CLANG_WNO_BEGIN("-Wall") // -Wno-non-virtual-dtor
```
tcare: Curious - why not use the specific warning here?

pal/src/thread/process.cpp
```diff
@@ -1958,7 +1958,10 @@ CreateProcessModules(
         ssize_t read;
 
         char vmmapCommand[100];
+CLANG_WNO_BEGIN("-Wunused-variable")
```
tcare: Perhaps it's better to wrap this in #ifdef DEBUG? It's not used in release.

lib/Common/Core/ConfigFlagsTable.cpp
```diff
@@ -405,9 +405,11 @@ namespace Js
     {
         const auto VerifyExecutionModeLimits = [this]()
         {
+CLANG_WNO_BEGIN("-Wunused-variable");
```
tcare: Also thinking that these asserts and variables could be wrapped in #ifdef DEBUG instead. The warning could still be valid in debug build.

bin/ch/WScriptJsrt.cpp
jianchun: No we can't use Assert(...) here, which strips the code away on release build.

Why don't we use an "UNUSED" macro solution like here: https://github.com/nodejs/node-chakracore/pull/83/commits/ae1b9d49277dd58684b38c64fdb145874ecfac80#diff-6a427a4aebd067397b4135287236dae5R41

tcare: I would prefer disabling the warning in this case actually.

lib/Parser/Scan.h
obastemur: Unfortunately disabling `-Wnon-virtual-dtor` does not work..

jianchun: Add a virtual dtor, if that's what it means?

obastemur: > Add a virtual dtor, if that's what it means?

for `delete this` ?

jianchun: Just alternatives for suppressing the warning. Searching on the web shows options such as adding a public virtual dtor or a protected non virtual dtor. Either looks better than code suppressing warnings.

obastemur:

obastemur:
Closing this one for now. Our testing is not based on xcode proj. compile hence at the moment no way to keep code base from being introduced more warnings.

PR #1275, opened 2016-07-13
Memop bailout on negative length

Cellule:
Bailout on Memset/Memcopy if the length is <= 0. 
After the memop, we restore the induction variable to its expected value. 
If the memop didn't actually happen we shouldn't modify the induction variable.
This is unlikely to happen since we would normally bailout on no profile and not emit this memop.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/1275)

### DIFF
--- a/test/Array/bug8159763.js
+++ b/test/Array/bug8159763.js
@@ -1,0 +1,19 @@
+//-------------------------------------------------------------------------------------------------------
+// Copyright (C) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE.txt file in the project root for full license information.
+//-------------------------------------------------------------------------------------------------------
+
+function test0() {
+  var IntArr0 = [];
+  v5 = IntArr0.length;
+  for (var i = 10; i < v5; i++) {
+    IntArr0[i] = 0.5;
+  }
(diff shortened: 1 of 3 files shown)


Cellule:
@LouisLaf @rajatd @pleath @MikeHolman Could one of you review this please  ?

MikeHolman:
lgtm

satheeshravi:
:shipit:

test/Array/bug8159763.js
```diff
@@ -0,0 +1,19 @@
+//-------------------------------------------------------------------------------------------------------
+// Copyright (C) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE.txt file in the project root for full license information.
+//-------------------------------------------------------------------------------------------------------
+
+function test0() {
```
satheeshravi: do you want to add to one of the existing memop test cases - instead of adding a file with bug number ?

Cellule: Good idea, I check if I can get this done

Cellule: Actually I can't because it need additional host flags to test properly.

PR #1277, opened 2016-07-14
Fix crash when tracing EquivObjTypeSpec

MikeHolman:
failedPropertyIndex is uninitialized in NullTypeHandler. it is used for tracing EquivObjTypeSpec, and causes a crash

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/1277)


rajatd:
:shipit:

PR #1279, opened 2016-07-14
nativetests: fix static crt build

jianchun:
The ChakraCore solution can be built with static CRT
(`msbuild /p:Runtime=static_library ...`). NativeTests.vcxproj should not
specify its own `RuntimeLibrary` property. Should use imported props.


akroshg:
:shipit:

jianchun:
@akroshg That's shockingly fast! Thanks!

PR #1283, opened 2016-07-15
cross-platform: macOS use CoreFramework for DateTime

obastemur:
After this PR; 7 more tests are passing + 3 of them are excluded from OSX

Currently, only couple of ICU and stackoverflow tests are left on OSX.

Excluded tests will need a revisit later for xplat improvements.
i.e. DST offset for year 35000+ is not maching xplat..

Also revisited Linux DateTime implementation for other distros.
i.e. On some distros mktime return -1 for minus years while some others don't.

### DIFF
--- a/lib/Runtime/PlatformAgnostic/Platform/CMakeLists.txt
+++ b/lib/Runtime/PlatformAgnostic/Platform/CMakeLists.txt
@@ -13,7 +13,7 @@ elseif(CMAKE_SYSTEM_NAME STREQUAL Darwin)
 set(PL_SOURCE_FILES
   # xplat-todo: implement them for unix! or carry them into common
   Linux/UnicodeText.ICU.cpp
-  Linux/DateTime.cpp
+  Unix/DateTime.cpp
   Linux/HiResTimer.cpp
   Linux/NumbersUtility.cpp
   Unix/SystemInfo.cpp
--- a/lib/Runtime/PlatformAgnostic/Platform/Unix/DateTime.cpp
+++ b/lib/Runtime/PlatformAgnostic/Platform/Unix/DateTime.cpp
@@ -1,0 +1,97 @@
+//-------------------------------------------------------------------------------------------------------
+// Copyright (C) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE.txt file in the project root for full license information.
+//-------------------------------------------------------------------------------------------------------
+
+#ifndef __APPLE__
+// todo: for BSD consider moving this file into macOS folder
+#include "../Linux/DateTime.cpp"
+#else
+#include "Common.h"
+#include "ChakraPlatform.h"
+#include <CoreFoundation/CFDate.h>
+#include <CoreFoundation/CFTimeZone.h>
+
+namespace PlatformAgnostic
+{
+namespace DateTime
+{
+    const WCHAR *Utility::GetStandardName(size_t *nameLength, const DateTime::YMD *ymd)
+    {
+        AssertMsg(ymd != NULL, "xplat needs DateTime::YMD is defined for this call");
+        double tv = Js::DateUtilities::TvFromDate(ymd->year, ymd->mon, ymd->mday, ymd->time);
+        int64_t absoluteTime = tv / 1000;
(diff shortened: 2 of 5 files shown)


lib/Runtime/PlatformAgnostic/Platform/CMakeLists.txt
```diff
@@ -13,7 +13,7 @@ elseif(CMAKE_SYSTEM_NAME STREQUAL Darwin)
 set(PL_SOURCE_FILES
   # xplat-todo: implement them for unix! or carry them into common
   Linux/UnicodeText.ICU.cpp
-  Linux/DateTime.cpp
+  Unix/DateTime.cpp
```
digitalinfinity: Should this be OS X instead of Unix? Is CoreFramework available on other Unix platforms?

lib/Runtime/PlatformAgnostic/Platform/Unix/DateTime.cpp
```diff
@@ -0,0 +1,103 @@
...
+        absoluteTime -= kCFAbsoluteTimeIntervalSince1970;
+
+        CFTimeZoneRef timeZone = CFTimeZoneCopySystem();
+
+        int offset = (int)CFTimeZoneGetSecondsFromGMT(timeZone, (CFAbsoluteTime)absoluteTime);
+        absoluteTime -= offset;
+
+        char tz_name[128];
+        CFStringRef abbr = CFTimeZoneCopyAbbreviation(timeZone, absoluteTime);
+        CFStringGetCString(abbr, tz_name, sizeof(tz_name), kCFStringEncodingASCII);
```
digitalinfinity: Curious- is there a reason you preferred this over using kCFStringEncodingUTF16 and just copying the result directory to data.standardTime using wcscpy or something?

lib/Runtime/PlatformAgnostic/Platform/Unix/DateTime.cpp
```diff
@@ -0,0 +1,103 @@
...
+        double tv = Js::DateUtilities::TvFromDate(ymd->year, ymd->mon, ymd->mday, ymd->time);
+        int64_t absoluteTime = tv / 1000;
+        absoluteTime -= kCFAbsoluteTimeIntervalSince1970;
+
+        CFTimeZoneRef timeZone = CFTimeZoneCopySystem();
+
+        int offset = (int)CFTimeZoneGetSecondsFromGMT(timeZone, (CFAbsoluteTime)absoluteTime);
+        absoluteTime -= offset;
+
+        char tz_name[128];
```
digitalinfinity: Just noticed that UtilityPlatformData::standardName's length is 32 WCHARs- should this also be 32 (or use __CC_PA_TIMEZONE_ABVR_NAME_LENGTH?), or if not, should you do truncation in CopyTimeZoneName if *length > 32?

digitalinfinity:
:shipit:

lib/Runtime/PlatformAgnostic/Platform/CMakeLists.txt
obastemur: It is under elseif(Darwin) part. We may change this if we support BSD in the future but at the moment we are far from that. Yet, will include Linux/DateTime implicitly under Unix/DateTime if the OS is not macOS.

lib/Runtime/PlatformAgnostic/Platform/Unix/DateTime.cpp
obastemur: Our xplat expectation for tz_name is non-wide char from the source. Your suggestion may also work indeed.

lib/Runtime/PlatformAgnostic/Platform/Unix/DateTime.cpp
obastemur: in case the wide char was 4 bytes, 4 x 32 = 128

obastemur:
Handled the UTF16 + wcscpy part.

@digitalinfinity thanks for the review and suggestions. If there is no objection, going to merge after CI.

digitalinfinity:
Sure- LGTM

PR #1285, opened 2016-07-15
add testing for medium and large bytecode layout to nightly config

MikeHolman:
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/1285)

### DIFF
--- a/Build/Chakra.Build.props
+++ b/Build/Chakra.Build.props
@@ -24,6 +24,10 @@
         %(PreprocessorDefinitions);
         DISABLE_JIT=1
       </PreprocessorDefinitions>
+      <PreprocessorDefinitions Condition="'$(ENABLE_CODECOVERAGE)'=='true'">
+        %(PreprocessorDefinitions);
+        BYTECODE_TESTING=1
+      </PreprocessorDefinitions>
       <!-- REVIEW: These are warning are introduced when moving to VS2015 tools, may want to clean these up -->
       <DisableSpecificWarnings>
         %(DisableSpecificWarnings);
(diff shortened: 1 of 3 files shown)


MikeHolman:
@aneeshdk @boingoing @EdMaurer, can you take a look?

dilijev:
@MikeHolman just to confirm, do you want to add this to any variants besides nightlies (VSTS, ability to trigger a Jenkins build with the same like the slow tests?)

Do you have a dump of the logs from running these? (Maybe put on https://gist.github.com/ or pastebin?)

dilijev:
Otherwise LGTM

ianwjhalliday:
Both of these two new configs will run the entire test suite again adding significant time to test run completion.  For that reason I don't think we want to run them outside of nightlies.

Interesting that the medium and large layout switches are not available in test builds.  We should probably have those switches available in test to get coverage in optimized builds.

MikeHolman:
@ianwjhalliday I think the reason they aren't in test build is because adds extra 2 (predicted) branches while encoding each bytecode. This is pretty tight code, so could cause some small perf overhead.

However, I believe code coverage will have its own build configuration, and that config will `#define BYTECODE_TESTING`.

dilijev:
@ianwjhalliday @MikeHolman - I was suggesting maybe adding them to the slow tests configuration that is run daily in Jenkins, but not to the rolling tests. Still depending on how much extra time they take to run / value of running in Jenkins in addition to internal nightlies, there still may be no good reason to run them in Jenkins.

aneeshdk:
LGTM

ianwjhalliday:
@MikeHolman gotcha.  Sounds goods.

@dilijev yeah, sounds like the daily build on jenkins won't add any coverage that we don't already get with nightlies.

:shipit:

PR #1286, opened 2016-07-16
BugFix: StackArgs Optimization with formals

satheeshravi:
After we change LdSlot instruction to Ld_A, the sym wasn't calling
ProcessSymUse (in the DeadStore pass).
ProcessSymUse marks the src sym as a nonTemp (which was being missed in
this case - causing a cascade effect with boxing of number)

Test:
UTs pass. Started other harness runs


satheeshravi:
@rajatd , Please take a look.

rajatd:
:shipit:

PR #1287, opened 2016-07-16
Change JavascriptConversion::ToPropertyKey to have no return value.

### LINKED ISSUE #1251: JavascriptConversion::ToPropertyKey should return void instead of BOOL
See https://github.com/Microsoft/ChakraCore/pull/1202#discussion_r69828994

`ToPropertyKey` always returns `TRUE` making the `BOOL` return type confusing for callers about whether they need to check the return value in release builds.  If it always returns true, the return type should be void to remove that cognitive overhead when using this API.

jordonwii:
Fixes #1251 

It turned out the *only* place the return value of ToPropertyKey was used (that I could find) was where I saw it in #1202. I changed the function signature to void, and removed the variable and the assert in that one place it was used.


satheeshravi:
:shipit:

ianwjhalliday:
Thanks @jordonwii!

PR #1288, opened 2016-07-16
Update @@toStringTag ES6 compliance

tcare:
Our current implementation of @@toStringTag is based on an old spec
revision from 2015. This change updates the implementation to conform to
the current spec.

The current behavior for handling legacy and toStringTag feature switching
is fragile, and I decided to leave the majority of it in and rename
ToStringTagHelper to ToStringTagHelperOld. The legacy code relies on the
helper, and will call it even when the @@toStringTag feature is disabled.

Other changes of note:
- Slightly tweaked the ordering of ToStringHelper so we can take advantage
  of the existing HostDispatch and TypedArray Debugger handling behavior
  while still retaining legacy codepaths.
- TypedArray toString handler spec update is also guarded by a flag.
- Overhauled the tests for @@toStringTag, adding a lot more type coverage.

When the feature is turned on by default, we can remove the legacy
codepaths for simplicity.

### DIFF
--- a/lib/Runtime/Library/JavascriptObject.cpp
+++ b/lib/Runtime/Library/JavascriptObject.cpp
@@ -357,7 +357,119 @@ namespace Js
         return ToStringHelper(args[0], scriptContext);
     }
 
-    JavascriptString* JavascriptObject::ToStringTagHelper(Var thisArg, ScriptContext* scriptContext, TypeId type)
+    // ES2017 19.1.3.6 Object.prototype.toString()
+    JavascriptString* JavascriptObject::ToStringTagHelper(Var thisArg, ScriptContext *scriptContext, TypeId type)
+    {
+        JavascriptLibrary *library = scriptContext->GetLibrary();
+
+        // 1. If the this value is undefined, return "[object Undefined]".
+        if (type == TypeIds_Undefined)
+        {
+            return library->CreateStringFromCppLiteral(_u("[object Undefined]"));
+        }
+        // 2. If the this value is null, return "[object Null]".
+        if (type == TypeIds_Null)
+        {
+            return library->CreateStringFromCppLiteral(_u("[object Null]"));
+        }
+
+        // 3. Let O be ToObject(this value).
+        RecyclableObject *thisArgAsObject = RecyclableObject::FromVar(JavascriptOperators::ToObject(thisArg, scriptContext));
+
+        // 4. Let isArray be ? IsArray(O).
+        // There is an implicit check for a null proxy handler in IsArray, so use the operator.
+        BOOL isArray = JavascriptOperators::IsArray(thisArgAsObject);
+
+        // 15. Let tag be ? Get(O, @@toStringTag).
+        Var tag = JavascriptOperators::GetProperty(thisArgAsObject, PropertyIds::_symbolToStringTag, scriptContext); // Let tag be the result of Get(O, @@toStringTag).
+
+        // 17. Return the String that is the result of concatenating "[object ", tag, and "]". 
+        if (tag != nullptr && JavascriptString::Is(tag))
+        {
+            JavascriptString *tagStr = JavascriptString::FromVar(tag);
+
+            CompoundString::Builder<32> stringBuilder(scriptContext);
+
+            stringBuilder.AppendChars(_u("[object "));
+
+            stringBuilder.AppendChars(tagStr);
+            stringBuilder.AppendChars(_u(']'));
+
+            return stringBuilder.ToString();
+        }
+
+        // If we don't have a tag or it's not a string, use the 'built in tag'.
+        if (isArray)
+        {
+            // 5. If isArray is true, let builtinTag be "Array".
+            return library->CreateStringFromCppLiteral(_u("[object Array]"));
+        }
+
+        JavascriptString* builtInTag = nullptr;
+        switch (type)
+        {
+            // 6. Else if O is an exotic String object, let builtinTag be "String".
+        case TypeIds_String:
+        case TypeIds_StringObject:
+            builtInTag = library->CreateStringFromCppLiteral(_u("[object String]"));
+            break;
+
+            // 7. Else if O has an[[ParameterMap]] internal slot, let builtinTag be "Arguments".
+        case TypeIds_Arguments:
+            builtInTag = library->CreateStringFromCppLiteral(_u("[object Arguments]"));
+            break;
+
+            // 8. Else if O has a [[Call]] internal method, let builtinTag be "Function".
+        case TypeIds_Function:
+            builtInTag = library->CreateStringFromCppLiteral(_u("[object Function]"));
+            break;
+
+            // 9. Else if O has an [[ErrorData]] internal slot, let builtinTag be "Error".
+        case TypeIds_Error:
+            builtInTag = library->GetErrorDisplayString();
+            break;
+
+            // 10. Else if O has a [[BooleanData]] internal slot, let builtinTag be "Boolean".
+        case TypeIds_Boolean:
+        case TypeIds_BooleanObject:
+            builtInTag = library->CreateStringFromCppLiteral(_u("[object Boolean]"));
+            break;
+
+            // 11. Else if O has a [[NumberData]] internal slot, let builtinTag be "Number".
+        case TypeIds_Number:
+        case TypeIds_Int64Number:
+        case TypeIds_UInt64Number:
+        case TypeIds_Integer:
+        case TypeIds_NumberObject:
+            builtInTag = library->CreateStringFromCppLiteral(_u("[object Number]"));
+            break;
+
+            // 12. Else if O has a [[DateValue]] internal slot, let builtinTag be "Date".
+        case TypeIds_Date:
+        case TypeIds_WinRTDate:
+            builtInTag = library->CreateStringFromCppLiteral(_u("[object Date]"));
+            break;
+
+            // 13. Else if O has a [[RegExpMatcher]] internal slot, let builtinTag be "RegExp".
+        case TypeIds_RegEx:
+            builtInTag = library->CreateStringFromCppLiteral(_u("[object RegExp]"));
+            break;
+
+            // 14. Else, let builtinTag be "Object".
+        default:
+            builtInTag = library->GetObjectDisplayString(); // [object Object]
+            break;
+        }
+
+        Assert(builtInTag != nullptr);
+
+        return builtInTag;
+    }
+
+    // This is the previous implementation of @@toStringTag, which is used in some cases even when the feature is disabled.
+    // Leaving this here for now due to the fragility.
+    // TODO(tcare): Remove when @@toStringTag is enabled by default.
+    JavascriptString* JavascriptObject::ToStringTagHelperOld(Var thisArg, ScriptContext* scriptContext, TypeId type)
     {
         JavascriptString* tag = nullptr;
         bool addTilde = true;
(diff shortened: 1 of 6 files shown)


tcare:
@Microsoft/chakra-es please review

test/es6/toStringTag.js
```diff
@@ -5,299 +5,221 @@
...
     {
-       name: "sub class Function.toString",
-       body: function () {
-            class MyFunction extends Function {
-                constructor() {
-                   super();
-                }
+        name: "Proxy forwarding behavior",
+        body: function () {
+            for (var { tag, instance } of [...explicitTagBuiltIns]) {
```
dilijev: nit: you should be able to use just `of explicitTagBuiltIns` in this case, right?

test/es6/toStringTag.js
```diff
@@ -308,7 +230,7 @@ var tests = [
 
             assert.areEqual("[object before\0after]", o.toString(), "ToString implementation handles embedded null characters in @@toStringTag property value");
         }
-    },
+    }
```
dilijev: nit: keeping the trailing comma lets future additions not have an extra line of diff

akroshg:
toStringTag.js test case is disabled in the rlexe.xml. Has that changed?

lib/Runtime/Library/JavascriptObject.cpp
```diff
@@ -357,7 +357,127 @@ namespace Js
...
+        }
+
+        Assert(builtInTag != nullptr);
+
+        return builtInTag;
+    }
+
+    // This is the previous implementation of @@toStringTag, which is used in some cases even when the feature is disabled.
+    // Leaving this here for now due to the fragility.
+    // TODO(tcare): Remove when @@toStringTag is enabled by default.
```
suwc: >Remove when @@toStringTag is enabled by default. [](start = 20, length = 48)

Is @@toStringTag going to be enabled through experimental before enabled by default?

tcare: The previous implementation already is I believe - it shows up in Kangax.

test/es6/toStringTag.js
tcare: Yes, thanks. Used to be more there. :)

tcare:
@akroshg nice catch - yes it should be re-enabled.

lib/Runtime/Library/JavascriptObject.cpp
suwc: Was it enabled recently? Because my 7/8/16 clone is not showing ES6ToStringTag enabled through FLAGPR_REGOVR_EXP in ConfigFlagsList.h

---
In reply to: [71397520](https://github.com/Microsoft/ChakraCore/pull/1288#discussion_r71397520) [](ancestors = 71397520)

test/es6/toStringTag.js
```diff
@@ -5,299 +5,221 @@
...
+    {
+        name: "@@toStringTag properties with side effects",
+        body: function () {
+            var tagThrows = { get [Symbol.toStringTag]() { throw new Error(); } };
+            var threw = false;
             try {
-                a[Symbol.toStringTag];
-            } catch(e) {
-                assert.areEqual("[object ???]", Object.prototype.toString.call(a), "JS exceptions should have tag ???");
+                Object.prototype.toString.call(instance);
```
akroshg: nit: why not change this test to assert.throws()  instead of having the try..catch and a variable?

akroshg: what is an 'instance' over here?

lib/Runtime/Library/JavascriptObject.cpp
```diff
@@ -357,7 +357,127 @@ namespace Js
...
+        // There is an implicit check for a null proxy handler in IsArray, so use the operator.
+        BOOL isArray = JavascriptOperators::IsArray(thisArgAsObject);
+
+        Var tag = nullptr;
+        // 15. Let tag be ? Get(O, @@toStringTag).
+        try
+        {
+            tag = JavascriptOperators::GetProperty(thisArgAsObject, PropertyIds::_symbolToStringTag, scriptContext); // Let tag be the result of Get(O, @@toStringTag).
+        }
+        catch (JavascriptExceptionObject*)
```
suwc: >catch (JavascriptExceptionObject*) [](start = 8, length = 34)

Is this path exercised in unit tests?

test/es6/toStringTag.js
tcare: Copied it from the old one.. and it should be tagThrows.

test/es6/toStringTag.js
```diff
@@ -5,299 +5,221 @@
...
+            assert.isFalse(o.enumerable,   "Symbol @@toStringTag is not enumerable");
+            assert.isTrue(o.configurable,  "Symbol @@toStringTag is configurable");
 
-            assert.areEqual("WeakSet", (new WeakSet())[Symbol.toStringTag], "check String is WeakSet");
-            var o  = Object.getOwnPropertyDescriptor(WeakSet.prototype, Symbol.toStringTag);
-            assert.isFalse(o.writable,     "WeakSet @@toStringTag is not writable");
-            assert.isFalse(o.enumerable,   "WeakSet @@toStringTag is not enumerable");
-            assert.isTrue(o.configurable,  "WeakSet @@toStringTag is configurable");
+            function checkBuiltInPrototype(builtInName, prototype) {
+                WScript.Echo("Testing " + builtInName);
```
akroshg: why are we printing this?

lib/Runtime/Library/TypedArray.cpp
```diff
@@ -1404,13 +1405,25 @@ namespace Js
...
 
+        // 1. Let O be the this value.
+        // 2. If Type(O) is not Object, return undefined.
+        // 3. If O does not have a[[TypedArrayName]] internal slot, return undefined.
         if (args.Info.Count == 0 || !TypedArrayBase::Is(args[0]))
         {
-            JavascriptError::ThrowTypeError(scriptContext, JSERR_This_NeedTypedArray);
+            if (scriptContext->GetConfig()->IsES6ToStringTagEnabled())
+            {
+                return library->GetUndefinedDisplayString();
```
akroshg: >GetUndefinedDisplayString [](start = 32, length = 25)

should we return GetUndefined() here instead of GetUndfinedDisplayString ? the spec says to return undefined.

lib/Runtime/Library/JavascriptObject.cpp
```diff
@@ -357,7 +357,127 @@ namespace Js
...
+
+        // 4. Let isArray be ? IsArray(O).
+        // There is an implicit check for a null proxy handler in IsArray, so use the operator.
+        BOOL isArray = JavascriptOperators::IsArray(thisArgAsObject);
+
+        Var tag = nullptr;
+        // 15. Let tag be ? Get(O, @@toStringTag).
+        try
+        {
+            tag = JavascriptOperators::GetProperty(thisArgAsObject, PropertyIds::_symbolToStringTag, scriptContext); // Let tag be the result of Get(O, @@toStringTag).
```
akroshg: question mark in the 15 clause says ReturnIfAbrupt. So if the GetProperty throws an exception shouldn't we let it go instead of handling it here?

lib/Runtime/Library/JavascriptObject.cpp
```diff
@@ -357,7 +357,127 @@ namespace Js
...
+            break;
+
+            // 8. Else if O has a [[Call]] internal method, let builtinTag be "Function".
+        case TypeIds_Function:
+            builtInTag = library->CreateStringFromCppLiteral(_u("[object Function]"));
+            break;
+
+            // 9. Else if O has an [[ErrorData]] internal slot, let builtinTag be "Error".
+        case TypeIds_Error:
+            builtInTag = library->CreateStringFromCppLiteral(_u("[object Error]"));
```
akroshg: You can calll to GetErrorDisplayString.

lib/Runtime/Library/JavascriptObject.cpp
tcare: Yes you are correct. I need to update the test case.

test/es6/toStringTag.js
tcare: This is left over from debugging. I found it helped to diagnose some issues where the assert wouldn't fire and show the message, especially when we are testing multiple types of built ins. I can either leave it in and make it depend on WScript.Arguments[0] != "summary" or just take it out completely.

lib/Runtime/Library/JavascriptObject.cpp
```diff
@@ -535,49 +655,83 @@ namespace Js
...
         TypeId type = JavascriptOperators::GetTypeId(thisArg);
         JavascriptLibrary* library = scriptContext->GetLibrary();
+
+        if (type == TypeIds_HostDispatch)
+        {
+            RecyclableObject* hostDispatchObject = RecyclableObject::FromVar(thisArg);
+            DynamicObject* remoteObject = hostDispatchObject->GetRemoteObject();
+            if (!remoteObject)
+            {
+                Var result;
```
akroshg: initialize this var.

lib/Runtime/Library/JavascriptObject.cpp
```diff
@@ -535,49 +655,83 @@ namespace Js
...
+            return library->GetUndefined();
         }
     }
 
     Var JavascriptObject::ToStringHelper(Var thisArg, ScriptContext* scriptContext)
     {
         TypeId type = JavascriptOperators::GetTypeId(thisArg);
         JavascriptLibrary* library = scriptContext->GetLibrary();
+
+        if (type == TypeIds_HostDispatch)
```
akroshg: I don't understand this. Why are we handling this (I don;t see this in the spec). Shouldn't this be [object Object] ?

lib/Runtime/Library/JavascriptObject.cpp
```diff
@@ -535,49 +655,83 @@ namespace Js
...
+                values[0] = thisArg;
+                if (hostDispatchObject->InvokeBuiltInOperationRemotely(EntryToString, args, &result))
+                {
+                    return result;
+                }
+            }
+        }
+
+        // Dispatch to new @@toStringTag implemtation.
+        if (scriptContext->GetConfig()->IsES6ToStringTagEnabled()) {
```
akroshg: nit: brace on new line.

lib/Runtime/Library/JavascriptObject.cpp
```diff
@@ -535,49 +655,83 @@ namespace Js
...
+            if (type >= TypeIds_TypedArrayMin && type <= TypeIds_TypedArrayMax && !scriptContext->GetThreadContext()->IsScriptActive())
```
akroshg: when the script will not be active? I understand this will be exercise from the script only.

lib/Runtime/Library/JavascriptObject.cpp
akroshg: Have perf run with toStringTag enabled and see if we are okay.

tcare:
If the perf is okay, I'm tempted just to rip out the legacy code now and turn this on by default. Let me know what you think.

dilijev:
@tcare I'm all in favor of reducing complexity. This means we're turning it on by default and removing the ability to turn it off, right? (Is there usually a process or official decision on this sort of thing?) Has a perf run already been done?

tcare:
I'm in the process of doing the perf at the moment. If it does have an impact then Suwei's symbol inline cache that was used for isConcatSpreadable will be easily adapted to this use.

I'll discuss enabling it by default with the team in tomorrow's standup.

lib/Runtime/Library/JavascriptObject.cpp
tcare: It looks like that isn't used anywhere. Maybe I can remove it from JavascriptLibrary.

GetObjectDisplayString is similar. After this change, shouldn't it be argued that anything using this should have used EntryToString? We have a use in ToLocaleString on a VariantDate.

lib/Runtime/Library/JavascriptObject.cpp
tcare: As Akrosh pointed out elsewhere, this is not spec compliant. I'll remove it.

lib/Runtime/Library/JavascriptObject.cpp
tcare: Spoke with Akrosh offline - we hit this in some debugger test cases.

suwc:
I am all for ripping out "ToStringTagOld()". As to turning this on by default or not, I would ask the lead.

---
In reply to: [233802690](https://github.com/Microsoft/ChakraCore/pull/1288#issuecomment-233802690) [](ancestors = 233802690)

tcare:
Thanks for the review feedback. I have addressed most issues.

Some preliminary perf data:
35% slower in microbenchmark doing 1M iterations of Object.prototype.toString on all builtin types. This is expected due to the property lookup in the new spec.
Octane tests of interest are code-load and typescript. Perf is flat on my noisy machine, full perf results pending.

Not going to pursue perf at this time - however Suwei's @@isConcatSpreadable cache would easily be adapted to @@toStringTag as well.

We also decided that we will rip out the legacy code and enable this by default. This will be the next commit.

lib/Runtime/Library/JavascriptObject.cpp
```diff
@@ -535,49 +647,84 @@ namespace Js
...
+            RecyclableObject* hostDispatchObject = RecyclableObject::FromVar(thisArg);
+            DynamicObject* remoteObject = hostDispatchObject->GetRemoteObject();
+            if (!remoteObject)
+            {
+                Var result = nullptr;
+                Js::Var values[1];
+                Js::CallInfo info(Js::CallFlags_Value, 1);
+                Js::Arguments args(info, values);
+                values[0] = thisArg;
+                if (hostDispatchObject->InvokeBuiltInOperationRemotely(EntryToString, args, &result))
```
akroshg: I wonder if there is a way to test this scenario. @Yongqu do you know?

akroshg:
Overall looks good to me.

lib/Runtime/Library/JavascriptObject.cpp
```diff
@@ -357,7 +357,119 @@ namespace Js
...
+            stringBuilder.AppendChars(tagStr);
+            stringBuilder.AppendChars(_u(']'));
+
+            return stringBuilder.ToString();
+        }
+
+        // If we don't have a tag or it's not a string, use the 'built in tag'.
+        if (isArray)
+        {
+            // 5. If isArray is true, let builtinTag be "Array".
```
Yongqu: we don't need to worry about proxy override toString tag here?

lib/Runtime/Library/JavascriptObject.cpp
```diff
@@ -535,49 +647,84 @@ namespace Js
...
+        {
+            if (type >= TypeIds_TypedArrayMin && type <= TypeIds_TypedArrayMax && !scriptContext->GetThreadContext()->IsScriptActive())
+            {
+                Var toStringValue = nullptr;
+                BEGIN_JS_RUNTIME_CALL_EX(scriptContext, false);
+                toStringValue = ToStringTagHelper(thisArg, scriptContext, type);
+                END_JS_RUNTIME_CALL(scriptContext);
+                return toStringValue;
+            }
+            return ToStringTagHelper(thisArg, scriptContext, type);
```
tcare: We are missing a CrossSite::MarshalVar here, which was triggered by a JSRT debugger test case. Added in the next version.

lib/Runtime/Library/JavascriptObject.cpp
Yongqu: we should be able to use the LoadScriptFile(, "crossthread") to test test. Did we add EntryToString operation in the remote support in es5 timeframe?

lib/Runtime/Library/JavascriptObject.cpp
```diff
@@ -535,49 +647,84 @@ namespace Js
...
+                Js::Arguments args(info, values);
+                values[0] = thisArg;
+                if (hostDispatchObject->InvokeBuiltInOperationRemotely(EntryToString, args, &result))
+                {
+                    return result;
+                }
+            }
+        }
+
+        // Dispatch to new @@toStringTag implemtation.
```
Yongqu: typo :) implementation.

lib/Runtime/Library/JavascriptObject.cpp
tcare: Good catch. This should be setting the default tag, not returning it.

lib/Runtime/Library/JavascriptObject.cpp
```diff
@@ -535,49 +647,84 @@ namespace Js
...
+        if (scriptContext->GetConfig()->IsES6ToStringTagEnabled())
+        {
+            if (type >= TypeIds_TypedArrayMin && type <= TypeIds_TypedArrayMax && !scriptContext->GetThreadContext()->IsScriptActive())
```
Yongqu: hmm is this just for the winrt typed array types? when will this be called from non-scriptactive time?

tcare: This breaks PDM debugger TypedArray tests. I'm not exactly sure why.

lib/Runtime/Library/JavascriptObject.cpp
tcare: This should be handled by the GetProperty above. Is that what you meant?

tcare:
I fixed the crosssite issue, and am leaving the change off by default for now. As discussed offline, external objects need special handling to not break compat before we turn on this change by default. DOM still has some standards work that needs to be done.

I'll follow up in a different PR.

tcare:
Ping @akroshg @Yongqu

akroshg:
:shipit:

PR #1292, opened 2016-07-19
Fix ES2015 module redeclaration errors and root object pollution

boingoing:
Creates a function body to hold the module statements so var and block-scoped declarations don't leak into the root object. Previously, all modules were global code so declarations would leak between module code via the root object.

This resolves two bugs:
https://microsoft.visualstudio.com/web/wi.aspx?pcguid=cb55739e-4afe-46a3-970f-1b49d8ee7564&id=7657926
https://microsoft.visualstudio.com/web/wi.aspx?pcguid=cb55739e-4afe-46a3-970f-1b49d8ee7564&id=7717411

### DIFF
--- a/lib/Parser/Parse.cpp
+++ b/lib/Parser/Parse.cpp
@@ -1703,11 +1703,10 @@ void Parser::BindPidRefsInScope(IdentPtr pid, Symbol *sym, int blockId, uint max
 
         if (ref->IsModuleExport())
         {
-            Assert(sym->GetIsGlobal());
             sym->SetIsModuleExportStorage(true);
         }
 
-        if (ref->GetFuncScopeId() != funcId && !sym->GetIsGlobal())
+        if (ref->GetFuncScopeId() != funcId && !sym->GetIsGlobal() && !sym->GetIsModuleExportStorage())
         {
             Assert(ref->GetFuncScopeId() > funcId);
             sym->SetHasNonLocalReference();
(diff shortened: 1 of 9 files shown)


pleath:
Is it necessary to create a function body? These module-scoped variables seem analogous to lexical declarations at global scope. Could we handle them essentially the same way?

-- Paul

akroshg:
did you send the full code-review? I wanted to look at the cases you are fixing.

Yongqu:
having a test case would be nice. Maybe just import two files with same content but different filenames?

---
In reply to: [233803217](https://github.com/Microsoft/ChakraCore/pull/1292#issuecomment-233803217) [](ancestors = 233803217)

Yongqu:
>                 this->m_writer.Reg2Int1(Js::OpCode::LdThis, funcInfo->thisPointerRegister, funcInfo->nullConstantRegister, this->GetModuleID());

I'm wondering maybe this is the right time to get rid of all the moduleroot code and use the moduleID for modulerecord as well? 

---
Refers to: lib/Runtime/ByteCode/ByteCodeEmitter.cpp:2569 in 92bc0c3. [](commit_id = 92bc0c37afd35214d76cf4b6ca6f7635714d67da, deletion_comment = False)

lib/Parser/Parse.cpp
```diff
@@ -1703,11 +1703,10 @@ void Parser::BindPidRefsInScope(IdentPtr pid, Symbol *sym, int blockId, uint max
 
         if (ref->IsModuleExport())
         {
-            Assert(sym->GetIsGlobal());
```
pleath: I assume this is no longer true because the exports are now considered local to the top-level function body?

lib/Runtime/ByteCode/ByteCodeEmitter.cpp
```diff
@@ -1340,7 +1340,14 @@ void ByteCodeGenerator::DefineUserVars(FuncInfo *funcInfo)
 
             if (sym->GetSymbolType() == STVariable)
             {
-                if (fGlobal)
+                if (sym->GetIsModuleExportStorage())
```
pleath: Is it necessary to do this initialization of export memory at runtime? It seems like it could be done when we allocate the memory.

lib/Parser/Parse.cpp
boingoing: That's right. The export symbols are no longer global.

lib/Runtime/ByteCode/ByteCodeEmitter.cpp
boingoing: Yes, I suppose that's true. I removed this initialization.

pleath:
:shipit:

PR #1293, opened 2016-07-19
cross-platform: use default brew/icu path for macos

obastemur:


tcare:
Does macports have a similar path?

obastemur:
> Does macports have a similar path?

not it's not by default. However it installs to opt/include straight. Actually #1301 fixes the issues we have with icu on mac too. Closing this for now. Once #1301 is merged, we may improve the build script in a way to check whether macports/brew installed icu is matching our requirements or not.

PR #1294, opened 2016-07-19
cross platform: add deb package creation support

obastemur:


build.sh
```diff
@@ -116,6 +118,11 @@ while [[ $# -gt 0 ]]; do
         MAKE=0
         ;;
 
+	--create-deb=*)
```
dilijev: nit : indentation

build.sh
```diff
@@ -228,6 +235,34 @@ fi
...
+        cp $DEB_FOLDER/../ch $DEB_FOLDER/usr/local/bin/
+        if [[ $STATIC_LIBRARY == "" ]]; then
+            cp $DEB_FOLDER/../*.so $DEB_FOLDER/usr/local/bin/
+        fi
+        echo -e "Package: ChakraCore"\
+            "\nVersion: ${CREATE_DEB}"\
+            "\nSection: base"\
+            "\nPriority: optional"\
+            "\nArchitecture: amd64"\
+            "\nDepends: libc6 (>= 2.19)"\
```
digitalinfinity: While we get the ICU story straightened out- don't we need libicu here? Our build instructions also have libuuid/libunwind/liblttng-ust (maybe for the PAL?) should we add them here?

build.sh
```diff
@@ -228,6 +235,34 @@ fi
...
+            "\nMaintainer: ChakraCore <chakracore@microsoft.com>"\
+            "\nDescription: Microsoft ChakraCore"\
```
dilijev: Should probably include the ®: "Microsoft ® ChakraCore" or "Chakra Core" to be consistent with the metadata in the Windows binary.

build.sh
obastemur: no icu needed. As for the others, we should find a solution similar to icu though. That is just for the warning, it doesn't have much effect as dpkg has no control over repos.

build.sh
obastemur: isn't 'Microsoft ChakraCore' informative enough for description?

dilijev: I think the registered trademark symbol (®) with Microsoft might be a legal requirement or recommendation for any user-visible string which includes the company and product name.

As for ChakraCore versus Chakra Core I don't personally have a preference. Was just pointing out the inconsistency.

obastemur:
@dilijev @digitalinfinity thanks for the reviews.

build.sh
```diff
@@ -228,6 +235,34 @@ fi
...
+        fi
+        echo -e "Package: ChakraCore"\
+            "\nVersion: ${CREATE_DEB}"\
+            "\nSection: base"\
+            "\nPriority: optional"\
+            "\nArchitecture: amd64"\
+            "\nDepends: libc6 (>= 2.19), uuid-dev (>> 0), libunwind-dev (>> 0), libicu-dev (>> 0)"\
+            "\nMaintainer: ChakraCore <chakracore@microsoft.com>"\
+            "\nDescription: Chakra Core"\
+            "\n Open source Core of Chakra Javascript Engine"\
```
dilijev: @liminzhu @EdMaurer @bterlson Does this metadata look alright for a public release?

obastemur:
Merging this. We are not releasing this package yet. Necessary text additions can be made later.

PR #1295, opened 2016-07-19
Bug fixes

rajatd:
1. Inlinees in a loop body frame need not marked as candidates for boxing as we wouldn't have allocated nested functions on the stack for a jitted loop body frame (both for loop body and inlinees in it)

2. CheckObjType instruction shouldn't add a bytecode use of the symbol whose type it is checking. For this purpose, I'm marking the source operand as jit-optimized.

3. Bailout offset on InlineeEnd should be that of the next real instruction after the call to the inlinee.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/1295)

### DIFF
--- a/lib/Backend/GlobOptFields.cpp
+++ b/lib/Backend/GlobOptFields.cpp
@@ -2154,6 +2154,35 @@ GlobOpt::CheckIfPropOpEmitsTypeCheck(IR::Instr *instr, IR::PropertySymOpnd *opnd
     return CheckIfInstrInTypeCheckSeqEmitsTypeCheck(instr, opnd);
 }
 
+IR::PropertySymOpnd *
+GlobOpt::CreateOpndForTypeCheckOnly(IR::PropertySymOpnd* opnd, Func* func)
+{
+    // Used only for CheckObjType instruction today. Future users should make a call
+    // whether the new operand is jit optimized in their scenario or not.
+
+    Assert(!opnd->IsRootObjectNonConfigurableFieldLoad());
+    IR::PropertySymOpnd *newOpnd = opnd->CopyCommon(func);
+
+    newOpnd->SetObjTypeSpecFldInfo(opnd->GetObjTypeSpecInfo());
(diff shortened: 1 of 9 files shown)


rajatd:
@LouisLaf @pleath @satheeshravi please review.

lib/Backend/GlobOptFields.cpp
```diff
@@ -2154,6 +2154,35 @@ GlobOpt::CheckIfPropOpEmitsTypeCheck(IR::Instr *instr, IR::PropertySymOpnd *opnd
...
+    // only for the sake of downstream instructions, so the flags pertaining to this property access are
+    // irrelevant, because we don't do a property access here.
+    newOpnd->SetTypeCheckOnly(true);
+    newOpnd->usesFixedValue = false;
+
+    newOpnd->finalType = opnd->finalType;
+    newOpnd->guardedPropOps = opnd->guardedPropOps != nullptr ? opnd->guardedPropOps->CopyNew() : nullptr;
+    newOpnd->writeGuards = opnd->writeGuards != nullptr ? opnd->writeGuards->CopyNew() : nullptr;
+
+    newOpnd->SetIsJITOptimizedReg(true);
```
satheeshravi: >newOpnd->SetIsJITOptimizedReg(true); [](start = 4, length = 36)

nit: do you mind adding a comment here as why you set this to true always(the one that have in the PR desc should suffice)

satheeshravi:
Looks good. I am not aware of the change in StackScriptFunction.cpp. So sign off for rest of the changes.

---
In reply to: [233733814](https://github.com/Microsoft/ChakraCore/pull/1295#issuecomment-233733814) [](ancestors = 233733814)

satheeshravi:
:shipit:

lib/Backend/GlobOptFields.cpp
rajatd: There is a comment in the beginning of the function.

PR #1296, opened 2016-07-19
Fixing an assert in GlobOpt::TrackInstrsForScopeObjectRemoval

satheeshravi:
The assert under interest was assuming that the Src1 of ArgOut Instr will
always have a single def, which is false in some cases(the one which I
observed was with PRE introducing extra defintion with this src sym)


rajatd:
:shipit:

satheeshravi:
@dotnet-bot : please test x64_debug

satheeshravi:
@dotnet-bot : please test Windows x64_debug

PR #1297, opened 2016-07-19
SparseBitVector::TestAndClear should not allocate

pleath:
Fix a nit I found yesterday: SparseBitVector::TestAndClear will create a BV node if the BV doesn't have a node that contains the bit we're testing. Changed the API so it doesn't create and instead returns early if it doesn't have the node we're looking for.


pleath:
@LouisLaf

rajatd:
:shipit:

PR #1298, opened 2016-07-20
Add strong return types to some internal property enumerating functions

### LINKED ISSUE #1252: Use specific strong return types for JavascriptOperators::GetOwnPropertyKeys() et al
Upon reviewing #1202 (see https://github.com/Microsoft/ChakraCore/pull/1202#discussion_r69828994) @Yongqu @tcare and I decided we should strongly type the return types of APIs used exclusively internal to the runtime rather than using `Var`.

Motivation comes from `JavascriptOperators::GetOwnPropertyKeys`.  Code review raised concern about assuming the return value is always a `JavascriptArray*` and that callers should handle the case this isn't true.  It is always true however and so far most of the callers have made this assumption and merely Asserted it, rather than handling the alternative in release builds.

Clearly there is unnecessary cognitive overhead when working with `Var` return types on APIs we know to have guarantees on the type of their result values.  We should use specific return types for these APIs and only generalize to Var when necessary e.g. when the API is used as a JIT helper.

This issue recommends changing the following APIs and their dependencies to use `JavascriptArray*` for their return type rather than `Var`:

 - `JavascriptOperators::GetOwnPropertyNames`
 - `JavascriptOperators::GetOwnPropertySymbols`
 - `JavascriptOperators::GetOwnPropertyKeys`

Also #1251 `JavascriptConversion::ToPropertyKey` should return `void` instead of `BOOL`.

jordonwii:
Fixes #1252 

This adds strong return types (specifically, `JavascriptArray*`) to the following functions:

In JavascriptOperators:
1. `JavascriptOperators::GetOwnPropertyNames`
2. `JavascriptOperators::GetOwnPropertySymbols`
3. `JavascriptOperators::GetOwnPropertyKeys`
4. `JavascriptOperators::GetOwnEnumerablePropertyNames`
5. `JavascriptOperators::GetOwnEnumerablePropertyNamesSymbols`

The first 3 were mentioned in #1252, and 4 and 5 had essentially the same structure - all 5 just returned one of either JavascriptProxy::PropertyKeysTrap or one of the \*PropertiesHelper functions on JavascriptObject (listed below), so I went ahead and gave *all* of them a strong return type, as well.

In JavascriptObject (changed because the above all return the result of one of these):

1. `JavascriptObject::CreateOwnSymbolPropertiesHelper`
2. `JavascriptObject::CreateOwnStringPropertiesHelper`
3. `JavascriptObject::CreateOwnStringSymbolPropertiesHelper`
4. `JavascriptObject::CreateOwnEnumerableStringPropertiesHelper`
5. `JavascriptObject::CreateOwnEnumerableStringSymbolPropertiesHelper`
6. `JavascriptObject::CreateKeysHelper`

The first 5 all return the result of 6). 

I had to update the signature of `JavascriptProxy::PropertyKeysTrap` to return `JavascriptArray*`, as well. This seemed correct, because a). most uses of it were followed by an assert that the result was a JavascriptArray, anyway, and b). [MDN says ](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy/handler/ownKeys) the result of the ownKeys trap *must* be an array. 

The rest of the changes are removing asserts and simplifying and/or removing any code that checked the type of the result of calls to any of the above functions.

### DIFF
--- a/lib/Runtime/Debug/TTActionEvents.cpp
+++ b/lib/Runtime/Debug/TTActionEvents.cpp
@@ -338,7 +338,7 @@ namespace TTD
             const JsRTVarsArgumentAction* action = GetInlineEventDataAs<JsRTVarsArgumentAction, EventKind::GetOwnPropertyNamesInfoActionTag>(evt);
             Js::Var var = InflateVarInReplay(ctx, action->Var1);
 
-            Js::Var res = Js::JavascriptOperators::GetOwnPropertyNames(var, ctx);
+            Js::JavascriptArray* res = Js::JavascriptOperators::GetOwnPropertyNames(var, ctx);
 
             JsRTActionHandleResultForReplay<JsRTVarsArgumentAction, EventKind::GetOwnPropertyNamesInfoActionTag>(ctx, evt, res);
         }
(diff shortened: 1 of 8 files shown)


ianwjhalliday:
Yes! Excellent!

This is spot on, :shipit:

(I will merge shortly)

ianwjhalliday:
fyi @Yongqu @tcare

jordonwii:
Awesome. Thanks @ianwjhalliday!

PR #1299, opened 2016-07-20
runtests: set the PST timezone to run the tests

xanlpz:
This reduces the failing tests from 21 to 1 if the right timezone is not set system-wide. The failing test comes from using PST and not Pacific Daylight Time, but not sure how to magically get the daylight stuff in here. In any case I think this is already an improvement.

### DIFF
--- a/test/runtests.py
+++ b/test/runtests.py
@@ -15,6 +15,7 @@ import traceback
 import argparse
 import xml.etree.ElementTree as ET
 import re
+import time
 
 # handle command line args
 parser = argparse.ArgumentParser(
@@ -530,6 +531,10 @@ def main():
     if len(args.folders) == 0:
         files = (os.path.join(test_root, x) for x in os.listdir(test_root))
         args.folders = [f for f in sorted(files) if not os.path.isfile(f)]
+
+    # Set the right timezone, the tests need Pacific Standard Time
+    os.environ['TZ'] = 'US/Pacific'
+    time.tzset()
 
     # load all tests
     tests = []


obastemur:
Which timezone you are in ?

xanlpz:
I'm in Spain (Central European Summer Time)

obastemur:
I'm also in the same time zone yet don't see any fail after applying this.

test/runtests.py
```diff
@@ -531,6 +532,10 @@ def main():
         files = (os.path.join(test_root, x) for x in os.listdir(test_root))
         args.folders = [f for f in sorted(files) if not os.path.isfile(f)]
 
+    # Set the right timezone, the tests need Pacific Standard Time
+    os.environ['TZ'] = 'US/Pacific'
```
obastemur: `PDT` instead of `US/Pacific` ?

xanlpz:
Hum, weird. My failing test after the patch is:

[1203/1224 0.04] Failed -> Date/formatting.js                                           
/home/xan/git/ChakraCore/BuildLinux/Test/ch -WERExceptionSupport -ExtendedErrorStackForTestHost -BaselineMode -maxInterpretCount:1 -maxSimpleJitRunCount:1 -bgjit- -ForceOldDateAPI /home/xan/git/ChakraCore/test/Date/formatting.js
Output: (at line 352)
----------------------------
Sun Nov 04 2012 01:52:00 GMT-0700 (Pacific Standard Time)
----------------------------
Expected Output:
----------------------------
Sun Nov 04 2012 01:52:00 GMT-0700 (Pacific Daylight Time)

ianwjhalliday:
Does this change the host system's timezone permanently or does this only apply for the python process and child processes?  Does this work on Windows?

xanlpz:
> Does this change the host system's timezone permanently or does this only apply for the python process and child processes?

Only the test process and its children.

> Does this work on Windows?

I cannot test that but it really should, I'm using vanilla python code.

obastemur:
@xanlpz I'm wondering if you get exactly the same error on each run ?

xanlpz:
Yep, always the same error.

dilijev:
Related to https://github.com/Microsoft/ChakraCore/issues/319

Thanks for this! LGTM

obastemur:
I couldn't reproduce that error no matter what. IMO no point to keep this PR waiting any longer. 

Merging it. @xanlpz  Thanks!

PR #1303, opened 2016-07-20
Add support in our Unicode library to encode surrogate pairs as UTF8

digitalinfinity:


bin/NativeTests/CodexTests.cpp
```diff
@@ -0,0 +1,215 @@
...
+        CHECK(encodedBuffer[2] == 0xBD);
+    }
+
+    //
+    // Following test cases are based on the Utf-8 decoder tests 
+    // suggested by Markus Kuhn at https://www.cl.cam.ac.uk/~mgk25/ucs/examples/UTF-8-test.txt
+    //
+    TEST_CASE("CodexTest_EncodeTrueUtf8_SingleSurrogates", "[CodexTest]")
+    {
+        constexpr charcount_t charCount = 1;
```
jianchun: `constexpr` may be unavailable on VS 2013

lib/Common/Codex/Utf8Codex.cpp
```diff
@@ -360,7 +391,7 @@ namespace utf8
         }
     }
 
-    void DecodeIntoAndNullTerminate(__out_ecount(cch+1) __nullterminated char16 *buffer, LPCUTF8 ptr, size_t cch, DecodeOptions options)
+    void __cdecl DecodeIntoAndNullTerminate(__out_ecount(cch+1) __nullterminated char16 *buffer, LPCUTF8 ptr, size_t cch, DecodeOptions options)
```
jianchun: Some (maybe all) of these `__cdecl` are not needed. (I added some to my prototype because originally I attempted to use the Codex lib from external host. External host may be compiled with different default calling convention, so needed these to reconcile.) If you don't plan to make the lib available by its own, don't need these.

lib/Common/Codex/Utf8Helper.h
```diff
@@ -74,7 +74,11 @@ namespace utf8
             return E_OUTOFMEMORY;
         }
 
-        utf8::DecodeIntoAndNullTerminate(destString, (LPCUTF8) sourceString, cchDestString);
+        // Some node tests depend on the utf8 decoder swallowing invalid unicode characters
```
jianchun: Comment should be `*not* swallowing`?

digitalinfinity:
Thanks for the review @Jianchun. Removed __cdecls, left CodexAssert as is for now since it's pre-existing (will have to verify that Edge doesn't have a dependency on it).

bin/NativeTests/CodexTests.cpp
```diff
@@ -0,0 +1,215 @@
+//-------------------------------------------------------------------------------------------------------
+// Copyright (C) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE.txt file in the project root for full license information.
+//-------------------------------------------------------------------------------------------------------
+#include "stdafx.h"
+#include "catch.hpp"
+#include <process.h>
```
obastemur: Looks like this file is Windows only ?

lib/Common/Codex/Utf8Codex.h
```diff
@@ -173,6 +176,46 @@ namespace utf8
...
+                // Consume the low surrogate
+                *source = *source + 1;
+                *cch = *cch - 1;
+
+                return EncodeSurrogatePair(surrogateHigh, surrogateLow, ptr);
+            }
+        }
+
+        // Invalid input: insert the unicode replacement character instead
+        ptr[0] = 0xEF;
```
obastemur: do we need to check endiannes here?

obastemur:
single comment/question on BE/LE. Otherwise LGTM

bin/NativeTests/CodexTests.cpp
digitalinfinity: Yeah :( I want these tests to run on xplat too- @akroshg, is that on your radar?

lib/Common/Codex/Utf8Codex.h
digitalinfinity: Good question- historically, Windows does UTF-16 LE by default and it doesn't look like Codex handles UTF-16 BE today at all. Can you file a separate issue for that and we can prioritize accordingly if people have a need for big-endian UTF16 decoding?

digitalinfinity:
Thanks for the reviews and comments @obastemur @Jianchun. Merging

PR #1305, opened 2016-07-20
Wasm: Add TestSuite to CI

Cellule:
Simple node program to convert the wasm testsuite to binary C
an also create/update the baseline files Remove failing tests between interpreter and jit for the time being

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/1305)


Cellule:
@MikeHolman can you review this please ?

Cellule:
You can ignore files that are located in `WasmSpec/baselines` `WasmSpec/testsuite-bin`
The files in `WasmSpec/testsuite` comes from https://github.com/WebAssembly/testsuite (mirror source: https://github.com/WebAssembly/spec/tree/master/ml-proto/test)
I have commented out some of those tests because they produce different result between our interpreter and jit which makes creating a baseline difficult.
I have also excluded
```js
      [
        "f32",
        "f32_cmp",
        "f64",
        "f64_cmp",
        "float_exprs",
        "float_misc",
        "i32",
        "memory_redundancy",
        "nan-propagation",
        "resizing",
      ]
```
For the same reason. Most of the issue are because of `NaN` as we discussed offline.

This should give us some kind of coverage and way to verify that our implementations are correct.

Cellule:
For the time being, the dev needs to have `sexpr-wasm` locally to regenerate those binaries.
I will investigate if/how we can include `sexpr-wasm`'s third-party code directly into our source base.

THIRD-PARTY-NOTICES.txt
```diff
@@ -229,3 +229,11 @@ ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
...
 
 ---------------------------------------------------------------------------
+WebAssembly/testsuite
+
+URL of the license https://github.com/WebAssembly/spec/blob/master/LICENSE
+https://github.com/WebAssembly/testsuite is a mirror of a subfolder of
+https://github.com/WebAssembly/spec
+
+
+The files in test/WasmSpec/testsuite are licensed as follows unless otherwise stated in the file
```
MikeHolman: >as follows [](start = 50, length = 10)

as follows? nothing follows

MikeHolman:
:shipit:

PR #1309, opened 2016-07-22
Fix BoxStackInstance error

pleath:
Similar issue to the missiung-undef-initialization case I fixed last week, this is the case where we have a dominating initialization that nevertheless doesn't write to the variable's register, but a later delay-captured reference does write to the register. Conservatively emitting undef-init now; delay capture needs to be revisited in the wake of the stable closures change.


pleath:
@rajatd

rajatd:
:+1:

PR #1311, opened 2016-07-22
Zero free memory with non-temporal stores

ThomsonTan:
Cache misses looks servere on low-end machine with less processor cache. It is uncessary to zero free memory and load all of them on cache. Changing it to non-temporal store gives ~%5 improvement of Earley-boyer in Octane on some low-end machines.

### DIFF
--- a/lib/Common/Memory/PageAllocator.cpp
+++ b/lib/Common/Memory/PageAllocator.cpp
@@ -936,12 +936,7 @@ PageAllocatorBase<T>::FillFreePages(__in void * address, uint pageCount)
 #if defined(_M_IX86) || defined(_M_X64)
         if (CONFIG_FLAG(ZeroMemoryWithNonTemporalStore))
         {
-            size_t writeBytes = 0;
-            __m128i simdZero = _mm_setzero_si128();
-            for (__m128i * p = (__m128i *)address; writeBytes < pageCount * AutoSystemInfo::PageSize; p += 1, writeBytes += sizeof(__m128i))
-            {
-                _mm_stream_si128(p, simdZero);
-            }
+            js_memset_zero_nontemporal(address, AutoSystemInfo::PageSize * pageCount);
         }
         else
 #endif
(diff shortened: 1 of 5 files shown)


ThomsonTan:
@LouisLaf please take a look.

lib/Common/Memory/PageAllocator.cpp
```diff
@@ -929,7 +929,25 @@ PageAllocatorBase<T>::FillFreePages(__in void * address, uint pageCount)
...
+            for (__m128i * p = (__m128i *)address; writeBytes < pageCount * AutoSystemInfo::PageSize; p += 1, writeBytes += sizeof(__m128i))
+            {
+                _mm_stream_si128(p, simdZero);
+            }
+        }
+        else
+#endif
+        {
+            memset(address, 0, AutoSystemInfo::PageSize * pageCount);
+        }
```
LouisLaf: Can you add a NonTemporal_memset() routine for this?  Not sure where it should go... Api.h with js_memcpy_s?

ThomsonTan: This is good suggestion. Just try add js_memset_zero_nontemporal to api.cpp but got multiple definition of js_wmemcpy_s on Linux build. Not sure why add a reference to api.cpp in page allocator leads to such error, will revisit this later.

---
In reply to: [72547225](https://github.com/Microsoft/ChakraCore/pull/1311#discussion_r72547225) [](ancestors = 72547225)

LouisLaf: Somehow, api.cpp and api.h seem unrelated…  Try putting it in api.h.

ThomsonTan: api.cpp and api.h look related at least for js_memcpy_s and js_wmemcpy_s as declarations are in api.h and implementation in api.cpp. I followed the same convention. Do you mean put the implementation to api.h?

---
In reply to: [72552157](https://github.com/Microsoft/ChakraCore/pull/1311#discussion_r72552157) [](ancestors = 72552157)

LouisLaf: Maybe just put the definition in PageAllocator.cpp then.

ThomsonTan: PageAllcoator.cpp doesn't like the ideal place to put general library functions. I suggest merge the current change and I will figure it out that why we get multiple definition error in app.cpp which is not supposed to happen.

---
In reply to: [72664236](https://github.com/Microsoft/ChakraCore/pull/1311#discussion_r72664236) [](ancestors = 72664236)

LouisLaf: No rush to get this in.  Let’s find a good place for this function and check it in then.

ThomsonTan: Ok. Took some time to look at the link error and got the duplicated symbol definition in GCStress. Removed the duplicated one in GCStress.

---
In reply to: [72701078](https://github.com/Microsoft/ChakraCore/pull/1311#discussion_r72701078) [](ancestors = 72701078)

PR #1312, opened 2016-07-22
cross-platform: Fix GCC lib Fedora X64 compile time issues

obastemur:
Fedora x64, we have only 1 failing test at the moment: 'test/Regex/BoiHardFail.js'

Compile steps for Fedora 24

Install dependencies
```
sudo dnf group install -y "Development Tools" "C Development Tools and Libraries"
sudo dnf install -y git cmake clang gcc gcc-c++ kernel-devel python
sudo dnf install -y libuuid-devel lttng-ust-devel.x86_64  libunwind-devel.x86_64 libicu-devel.x86_64
```

Clone && Compile
```
> git clone https://github.com/Microsoft/ChakraCore
> cd ChakraCore
> ./build.sh
```

### DIFF
--- a/pal/src/include/pal/palinternal.h
+++ b/pal/src/include/pal/palinternal.h
@@ -677,4 +677,11 @@ inline T* InterlockedCompareExchangePointerT(
 
 #endif // __cplusplus
 
+#ifndef max
+#define max(a,b)            (((a) > (b)) ? (a) : (b))
+#endif
+#ifndef min
+#define min(a,b)            (((a) < (b)) ? (a) : (b))
+#endif
+
 #endif /* _PAL_INTERNAL_H_ */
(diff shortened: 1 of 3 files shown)


pal/src/include/pal/palinternal.h
```diff
@@ -677,4 +677,11 @@ inline T* InterlockedCompareExchangePointerT(
 
 #endif // __cplusplus
 
+#ifndef max
```
digitalinfinity: Curious, why is this needed for fedora and not needed for ubuntu? pal.h also defines min and max, so this seems unnecessary but I saw the same compile error as you. Can you explain why that definition is insufficient?

digitalinfinity:
Awesome. Thanks for fixing this @obastemur! LGTM, just had one question.

pal/src/include/pal/palinternal.h
obastemur: IIRC; NO_PAL_MINMAX was being defined in the chain by mistake.

obastemur:
@digitalinfinity @jianchun thanks for the reviews!

PR #1313, opened 2016-07-22
split ChakraCoreVersion.h file

jianchun:
Split version defs out from `CommonDefines.h` into a separate file.
Hosts that statically link to ChakraCore can then access ChakraCore
version info by `#include` this file at build time.


digitalinfinity:
LGTM. Is this file supposed to be redistributable with Jsrt?

jianchun:
> LGTM. Is this file supposed to be redistributable with Jsrt?

Thanks. Yes, I suppose so. Not needed for Windows/Chakra JSRT. Needed for ChakraCore.

PR #1315, opened 2016-07-22
destructuring issue : arrow function and shorthand object pattern on param is valid syntax.

akroshg:


akroshg:
@tcare - please have a look.

tcare:
Looks good.

akroshg:
thanks @tcare

PR #1316, opened 2016-07-22
Wasm: codegen for trunc and nearest with more tests

commonlisp:
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/1316)


lib/Backend/LowerMDShared.cpp
```diff
@@ -9151,17 +9143,30 @@ void LowererMD::GenerateFastInlineBuiltInCall(IR::Instr* instr, IR::JnHelperMeth
...
+            else if (instr->m_opcode == Js::OpCode::Nearest_A)
             {
-                roundMode = IR::IntConstOpnd::New(0x02, TyInt32, this->m_func);
+                roundMode = IR::IntConstOpnd::New(0x00, TyInt32, this->m_func);
             }
             else
+#endif //ENABLE_WASM
             {
-                roundMode = IR::IntConstOpnd::New(0x03, TyInt32, this->m_func);
+                if (isNotCeil)
```
MikeHolman: this is a bit confusing now. i think would be more clear what is going on if we have switch on m_opcode (and i'm pretty sure `|| instr->m_opcode != Js::OpCode::InlineMathFloor` will always be false)

lib/Backend/LowerMDShared.cpp
```diff
@@ -9153,32 +9153,27 @@ void LowererMD::GenerateFastInlineBuiltInCall(IR::Instr* instr, IR::JnHelperMeth
...
-                {
-                    roundMode = IR::IntConstOpnd::New(0x01, TyInt32, this->m_func);
-                }
-                else if (instr->GetDst()->IsInt32() || instr->m_opcode != Js::OpCode::InlineMathFloor)
-                {
-                    roundMode = IR::IntConstOpnd::New(0x02, TyInt32, this->m_func);
-                }
-                else
-                {
-                    roundMode = IR::IntConstOpnd::New(0x03, TyInt32, this->m_func);
```
MikeHolman: was this block dead before?

commonlisp: As far as I can tell. InlineMathRound and InlineMathFloor (i.e., != InlineMathCeil) are both handled by the first if-clause. Thus, the only remaining case is InlineMathCeil for which != InlineMathFloor always evaluates to true.

Cellule:
Reviewed 6 of 7 files at r1, 1 of 1 files at r2.
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---

*[lib/Backend/Lower.cpp, line 2989 \[r2\]](https://reviewable.io:443/reviews/microsoft/chakracore/1316#-KNsj5c8QjznsTEBjVF3:-KNsj5c8QjznsTEBjVF4:-268612229) ([raw file](https://github.com/microsoft/chakracore/blob/892694f50de056e17c598242d69774654bcfc002/lib/Backend/Lower.cpp#L2989)):*
> ```cpp
> 
>         case Js::OpCode::Trunc_A:
>             m_lowererMD.GenerateFastInlineBuiltInCall(instr, (IR::JnHelperMethod)0);
> ```

This requires SSE 4, it will crash or produce invalid code if the machine doesn't have sse 4.
You can test with `ch.exe misc.js -on:wasm -sse:2 -forcenative`
Either make a valid path for this case or make it NYI in this case like so
`WASM_UNARY__OPCODE(F32Trunc,          0x80, F_F , Trunc_Flt      , !AutoSystemInfo::Data.SSE4_1Available())`

---

*[lib/Backend/LowerMDShared.cpp, line 6136 \[r2\]](https://reviewable.io:443/reviews/microsoft/chakracore/1316#-KNsf3lkve9r-fLQ9ycn:-KNsf3lkve9r-fLQ9yco:1996805885) ([raw file](https://github.com/microsoft/chakracore/blob/b2fc06d44d9aac3e35fdbf342a80375b576fcbf5/lib/Backend/LowerMDShared.cpp#L6136)):*
> ```cpp
> 
> void
> LowererMD::GenerateTrunc(IR::Instr * instr)
> ```

Also remove the declaration of both functions in the header file

---

*[test/wasm/misc.wast, line 6 \[r2\]](https://reviewable.io:443/reviews/microsoft/chakracore/1316#-KNsjrutcZSyFLnuX7CK:-KNsjruugzVqe6XvIwHY:-1170309093) ([raw file](https://github.com/microsoft/chakracore/blob/892694f50de056e17c598242d69774654bcfc002/test/wasm/misc.wast#L6)):*
> ```
> ;;-------------------------------------------------------------------------------------------------------
> 
> (module
> ```

Add f64 tests for trunc and nearest. Also use named functions to avoid potential errors with the index when exporting

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/microsoft/chakracore/1316)*

lib/Backend/LowerMDShared.cpp
Cellule: I double checked and this is correct. The last block was never used.

MikeHolman:
*[lib/Backend/Lower.cpp, line 2989 \[r2\]](https://reviewable.io:443/reviews/microsoft/chakracore/1316#-KNsj5c8QjznsTEBjVF3:-KNsoccdeiBI-9Vdq1FD:-834542730) ([raw file](https://github.com/microsoft/chakracore/blob/892694f50de056e17c598242d69774654bcfc002/lib/Backend/Lower.cpp#L2989)):*
> This requires SSE 4, it will crash or produce invalid code if the machine doesn't have sse 4.
> You can test with `ch.exe misc.js -on:wasm -sse:2 -forcenative`
> Either make a valid path for this case or make it NYI in this case like so
> `WASM_UNARY__OPCODE(F32Trunc,          0x80, F_F , Trunc_Flt      , !AutoSystemInfo::Data.SSE4_1Available())`
</details>
Good catch, this definitely needs a code path for older machines. IMO we shouldn't have this defined conditional on SSE4.1 because that will mess up testing, e.g. in case the test VM has or does not have SSE4.1

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/microsoft/chakracore/1316)*

commonlisp:
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---

*[lib/Backend/Lower.cpp, line 2989 \[r2\]](https://reviewable.io:443/reviews/microsoft/chakracore/1316#-KNsj5c8QjznsTEBjVF3:-KO5wGbD8EXIKkfwOetz:-1607693070) ([raw file](https://github.com/microsoft/chakracore/blob/892694f50de056e17c598242d69774654bcfc002/lib/Backend/Lower.cpp#L2989)):*
> Good catch, this definitely needs a code path for older machines. IMO we shouldn't have this defined conditional on SSE4.1 because that will mess up testing, e.g. in case the test VM has or does not have SSE4.1
</details>
For now, using builtins for SSE2. This is consistent with how InlineMathCeil and Floor are being handled right now.

---

*[lib/Backend/LowerMDShared.cpp, line 6136 \[r2\]](https://reviewable.io:443/reviews/microsoft/chakracore/1316#-KNsf3lkve9r-fLQ9ycn:-KO5VwU6a9EnPdezefoK:-499146009) ([raw file](https://github.com/microsoft/chakracore/blob/b2fc06d44d9aac3e35fdbf342a80375b576fcbf5/lib/Backend/LowerMDShared.cpp#L6136)):*
> Also remove the declaration of both functions in the header file
</details>
Done.

---

*[test/wasm/misc.wast, line 6 \[r2\]](https://reviewable.io:443/reviews/microsoft/chakracore/1316#-KNsjrutcZSyFLnuX7CK:-KO5ytbvNSYvnVT23YCQ:-499146009) ([raw file](https://github.com/microsoft/chakracore/blob/892694f50de056e17c598242d69774654bcfc002/test/wasm/misc.wast#L6)):*
> Add f64 tests for trunc and nearest. Also use named functions to avoid potential errors with the index when exporting
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/microsoft/chakracore/1316)*

Cellule:
:lgtm:

---

Reviewed 9 of 9 files at r3.
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/microsoft/chakracore/1316#-:-KO6L1vlfPuLXrKCJF7g:1417088721)*

PR #1317, opened 2016-07-22
cross-platform: fix build break

obastemur:
include  `<string>` before PAL header file on non `_WIN32`

### DIFF
--- a/bin/ch/stdafx.h
+++ b/bin/ch/stdafx.h
@@ -26,6 +26,7 @@
 #define ENABLE_TEST_HOOKS 1
 #include "CommonDefines.h"
 #include <map>
+#include <string>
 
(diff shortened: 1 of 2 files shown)


obastemur:
updated the description, 'echo' failed to print inlines

bin/ch/stdafx.h
```diff
@@ -30,6 +30,7 @@
 #ifdef _WIN32
 #include <windows.h>
 #else
+#include <string>
```
jianchun: Consider move this to line 29 and share for all, and remove the `#include` from `WScriptJsrt.cpp`.

jianchun:
LGTM, one minor comment

obastemur:
@jianchun thanks for the instant review!

PR #1321, opened 2016-07-22
Handle certain Array.concat side-effects

pleath:
Make Array concat responsive to side-effects of calling @@species create to instantiate the destination array. Addresses CVE-2016-3214.


pleath:
@suwc

suwc:
:shipit:

PR #1322, opened 2016-07-23
Fix for StrictMode with args optimization

satheeshravi:
Dictionary type is not shareable with the current design and hence cannot be cached on the javascript library.
We can change this design if there is a compelling reason in the future. 
For now, Creating dictionary type everytime still works better than what we had before.


satheeshravi:
@pleath : Can you please take a look ?

pleath:
:shipit:

PR #1323, opened 2016-07-23
add missing serialization layouts; disable forceserialized from some UTs

MikeHolman:
There were some bytecode layouts missing serialization code. This change adds them.

Additionally, there are some unittests which give different output when -forceserialized flag is used, so I have disabled them for these configurations. These were all files which already had -serialized flag or a -trace flag.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/1323)

### DIFF
--- a/lib/Runtime/ByteCode/ByteCodeSerializer.cpp
+++ b/lib/Runtime/ByteCode/ByteCodeSerializer.cpp
@@ -790,7 +790,7 @@ public:
             opStart = reader.GetIP();
             opStart; // For prefast. It can't figure out that opStart is captured in saveBlock above.
             LayoutSize layoutSize;
-            OpCodeAsmJs op = (OpCodeAsmJs)reader.ReadOp(layoutSize);
+            OpCodeAsmJs op = reader.ReadAsmJsOp(layoutSize);
             if (op == OpCodeAsmJs::EndOfBlock)
             {
                 saveBlock();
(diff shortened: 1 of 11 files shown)


lib/Runtime/ByteCode/ByteCodeSerializer.cpp
```diff
@@ -1182,6 +1192,8 @@ class ByteCodeBufferBuilder
                 DEFAULT_LAYOUT_WITH_ONEBYTE(Class);
                 DEFAULT_LAYOUT_WITH_ONEBYTE(ElementU);
                 DEFAULT_LAYOUT_WITH_ONEBYTE(ElementRootU);
+                DEFAULT_LAYOUT_WITH_ONEBYTE(ElementScopedC);
+                DEFAULT_LAYOUT_WITH_ONEBYTE(ElementScopedC2);
```
MikeHolman: I think these 2 may not be necessary, as they are only for eval, which should not hit in real world (this was hitting under -force:EvalCompile). so maybe better fix is to disable this flag in case of serialization flag (or visa-versa).

MikeHolman:
@pleath Do you have any opinion on this?

pleath:
I think it's fine to include them. I suppose there's a possibility we could need to serialize those. If we were inventing something or deoptimizing the main path just to support an artificial scenario, I'd say no, But this seems harmless.

pleath:
LGTM.

PR #1325, opened 2016-07-23
cross platform: fix static CH release builds

obastemur:
ENABLE_TTD was being defined on consumer side unintentionally

@digitalinfinity @jianchun

### DIFF
--- a/bin/ch/ch.cpp
+++ b/bin/ch/ch.cpp
@@ -310,6 +310,12 @@ HRESULT RunScript(const char* fileName, LPCSTR fileContents, BYTE *bcBuffer, cha
             }
 
             runScript = ChakraRTInterface::JsTTDRunScript(-1, fileContents, WScriptJsrt::GetNextSourceContext(), fullPath, nullptr /*result*/);
+
+            if (runScript == JsErrorCategoryUsage)
+            {
+                wprintf(_u("FATAL ERROR: Core was compiled without ENABLE_TTD is defined. CH is trying to use TTD interface\n"));
+                abort();
+            }
 #else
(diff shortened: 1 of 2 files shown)


bin/ch/ch.cpp
```diff
@@ -310,6 +310,12 @@ HRESULT RunScript(const char* fileName, LPCSTR fileContents, BYTE *bcBuffer, cha
             }
 
             runScript = ChakraRTInterface::JsTTDRunScript(-1, fileContents, WScriptJsrt::GetNextSourceContext(), fullPath, nullptr /*result*/);
+
+            if (runScript == JsErrorCategoryUsage)
+            {
+                wprintf(_u("FATAL ERROR: Core was built without ENABLE_TTD.\n"));
```
dilijev: This is enclosed in `#if ENABLE_TTD` ... is this an accurate message?

obastemur: Yes.

dilijev: I don't understand what's going on then...

obastemur: That's the problem here. `ENABLE_TTD` can be triggered by the consumer. So, in that case abort and show this message because it is hard to process later.

jianchun:
LGTM. We'll figure out why CH turns on ENABLE_TEST_HOOKS and ENABLE_DEBUG_CONFIG_OPTIONS for release build. Ideally they should bother be off.

obastemur:
@dilijev @jianchun thanks for the reviews!

bin/ch/ch.cpp
dilijev: I understand now. Thanks.

dilijev:
:shipit:

obastemur:
@dotnet-bot test this please

PR #1326, opened 2016-07-24
Fix for preventing copy-prop of foo.arguments to support the creation of arguments object during each access.

satheeshravi:
Premise:
foo.arguments should get a new copy of the arguments object everytime it is accessed (this change was in effect from my recent Stack Args change).
We were doing this right only when the instruction is profiled - by adding a ImplicitAccessor flag in the fldInfoFlags.
But this doesn't cover the cases where the instruction is not profiled (as in cases equivalent to forcenative or forcejitloopbody).
The property sym for arguments fld can be copy proped to the subsequent uses of it, which is wrong.

Fix:
Added a direct check in the OptSrc to detect this case and to disable copy prop for this field.

Test: 
Will do a quick perf test and will start other tests.

### DIFF
--- a/lib/Backend/GlobOpt.cpp
+++ b/lib/Backend/GlobOpt.cpp
@@ -5690,6 +5690,13 @@ GlobOpt::OptSrc(IR::Opnd *opnd, IR::Instr * *pInstr, Value **indirIndexValRef, I
         }
         originalPropertySym = sym->AsPropertySym();
 
+        //Don't copy-prop for cases like foo.arguments
+        //We need new copies of the arguments object for accesses like foo.arguments - So copy-proping is not allowed.
+        if (instr->m_opcode == Js::OpCode::LdFld && originalPropertySym->m_propertyId == Js::PropertyIds::arguments)
+        {
+            return nullptr;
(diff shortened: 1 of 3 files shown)


lib/Backend/GlobOpt.cpp
```diff
@@ -5690,6 +5690,13 @@ GlobOpt::OptSrc(IR::Opnd *opnd, IR::Instr * *pInstr, Value **indirIndexValRef, I
         }
         originalPropertySym = sym->AsPropertySym();
 
+        //Don't copy-prop for cases like foo.arguments
```
rajatd: Can you check if you can use argumentsEquivBv on Globopt to solve your issue? That bit vector is there for these kind of issues only- maybe we need to expand the usage of it. Basically, right now we use the bv to disable copy-prop or arguments across inlinees

dilijev:
@satheeshravi Any update on this?

satheeshravi:
I have been working on some other high pri work. I will get back to this soon.

satheeshravi:
Will close this PR. There is a new fix for this.

PR #1327, opened 2016-07-24
cross-platform: use ICU libs from target --icu path instead

obastemur:
Searches for the libs from custom ICU path


digitalinfinity:
LGTM

obastemur:

PR #1328, opened 2016-07-24
cross platform: enable release binary basic testing

obastemur:
current xplat release binary was broken and CI could not detect that
This PR enables very basic testing to make sure the compiled binary is functioning

/cc @dilijev @digitalinfinity @jianchun

### DIFF
--- a/test/runtests.sh
+++ b/test/runtests.sh
@@ -19,6 +19,20 @@ if [[ $build_type != "-d" && $build_type != "-t" ]]; then
     elif [[ -f "$test_path/../BuildLinux/Test/ch" ]]; then
         echo "Warning: Test build was found"
         build_type="-t"
+    elif [[ -f "$test_path/../BuildLinux/Release/ch" ]]; then
+        # TEST flags are not enabled for release build
+        # however we would like to test if the compiled binary
+        # works or not
+        CH="$test_path/../BuildLinux/Release/ch"
+        echo "Warning: Release build was found"
+        RES=$(${CH} $test_path/test/basics/hello.js)
+        if [[ $RES =~ "Error :" ]]; then
+            echo "FAILED"
+            exit 1
+        else
+            echo "PASS"
+            exit 0
+        fi
     else
         echo 'Error: ch not found- exiting'
         exit 1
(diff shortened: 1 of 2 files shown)


obastemur:
note: Merge #1325 before this one.

test/runtests.sh
```diff
@@ -19,6 +19,20 @@ if [[ $build_type != "-d" && $build_type != "-t" ]]; then
     elif [[ -f "$test_path/../BuildLinux/Test/ch" ]]; then
         echo "Warning: Test build was found"
         build_type="-t"
+    elif [[ -f "$test_path/../BuildLinux/Release/ch" ]]; then
+        # TEST flags are not enabled for release build
+        # however we would like to test if the compiled binary
+        # works or not
+        CH="$test_path/../BuildLinux/Release/ch"
+        echo "Warning: Release build was found"
```
dilijev: Why is this a warning?

test/runtests.sh
```diff
@@ -19,6 +19,20 @@ if [[ $build_type != "-d" && $build_type != "-t" ]]; then
...
+        RES=$(${CH} $test_path/test/basics/hello.js)
```
dilijev: Is hello.js the only file we would be interested in testing? It seems even without test flags we could test a little more than the bare basics to make sure things are all OK.

test/runtests.sh
obastemur: similar to others. Attention, this is release testing etc.

test/runtests.sh
obastemur: We shouldn't bother testing others. `test-build` has same optimizations with flags are enabled.

test/runtests.sh
dilijev: I think in all cases, warning is a bit strong since there's no case where you won't get a warning. Maybe just change these to "Note: using Release build" etc.

test/runtests.sh
dilijev: So the point is just to make sure that it runs at all?

dilijev:
LGTM

PR #1331, opened 2016-07-25
Clean up PAL external dependencies

digitalinfinity:
Minor clean up of external PAL dependencies- we don't need uuid or liblttng-ust right now so we don't need to block the build on that.
We do need ICU so configure.cmake is updated to check for that. This part is temporary till we make the ICU optional.


digitalinfinity:
@jianchun please review

jianchun:
LGTM 👍

PR #1333, opened 2016-07-26
Fix #1318. Date.UTC should return NaN with < 2 arguments.

### LINKED ISSUE #1318: Date.UTC should return NaN with < 2 arguments
Currently, the spec says this behavior is implementation defined. We have decided that means to default these parameters to zero. All other implementations simply return NaN. I think we should align if possible and update the spec to mandate this behavior. Are there compelling reasons to keep our current behavior? It'd be a compat break for UWP/etc. but this behavior doesn't seem particularly useful so hopefully it's not used...

See also https://github.com/whatwg/javascript/issues/31

dilijev:
Fixes #1318


dilijev:
@bterlson I forgot to add unit tests but I think this should fix the issue.

bterlson:
Looks good. Let's not pull this in until we discuss in TC39, though!

bterlson:
Ok, so far we have consensus that the 0-arg version should return NaN. We will discuss whether the 1-arg form should return NaN or Date.UTC(x, 0) on Thursday.

dilijev:
@bterlson, sounds good. The failing test is a test whose behavior is broken by the new semantics. I'll wait to do the tests until concensus is reached.

PR #1337, opened 2016-07-26
Fix missing capture dest reg for bailout

ThomsonTan:
When tracking symbols for capturing bailout values incrementally, the dest register for the current instruction might be ignored because the set of changed symbols is cleaned up at the end of OptInstr. The fix captures all the symbols changed after bailout value capture and restore them to changed symbols set when attaching capture values to bailout finally.

### DIFF
--- a/lib/Backend/GlobOpt.cpp
+++ b/lib/Backend/GlobOpt.cpp
@@ -1511,6 +1511,7 @@ GlobOpt::NulloutBlockData(GlobOptBlockData *data)
 
     data->capturedValues = nullptr;
     data->changedSyms = nullptr;
+    data->changedSymsAfterIncBailoutCandidate = nullptr;
 
     data->OnDataUnreferenced();
 }
@@ -1559,6 +1560,7 @@ GlobOpt::InitBlockData()
     data->stackLiteralInitFldDataMap = nullptr;
 
     data->changedSyms = JitAnew(alloc, BVSparse<JitArenaAllocator>, alloc);
+    data->changedSymsAfterIncBailoutCandidate = JitAnew(alloc, BVSparse<JitArenaAllocator>, alloc);
 
     data->OnDataInitialized(alloc);
 }
@@ -1609,7 +1611,9 @@ GlobOpt::ReuseBlockData(GlobOptBlockData *toData, GlobOptBlockData *fromData)
     toData->stackLiteralInitFldDataMap = fromData->stackLiteralInitFldDataMap;
 
     toData->changedSyms = fromData->changedSyms;
+    toData->changedSymsAfterIncBailoutCandidate = fromData->changedSymsAfterIncBailoutCandidate;
     toData->changedSyms->ClearAll();
+    toData->changedSymsAfterIncBailoutCandidate->ClearAll();
 
     toData->OnDataReused(fromData);
 }
@@ -1648,7 +1652,7 @@ GlobOpt::CopyBlockData(GlobOptBlockData *toData, GlobOptBlockData *fromData)
     toData->hasCSECandidates = fromData->hasCSECandidates;
 
     toData->changedSyms = fromData->changedSyms;
-    toData->changedSyms->ClearAll();
+    toData->changedSymsAfterIncBailoutCandidate = fromData->changedSymsAfterIncBailoutCandidate;
 
     toData->stackLiteralInitFldDataMap = fromData->stackLiteralInitFldDataMap;
     toData->OnDataReused(fromData);
@@ -1764,7 +1768,9 @@ void GlobOpt::CloneBlockData(BasicBlock *const toBlock, GlobOptBlockData *const 
     }
 
     toData->changedSyms = JitAnew(alloc, BVSparse<JitArenaAllocator>, alloc);
+    toData->changedSymsAfterIncBailoutCandidate = JitAnew(alloc, BVSparse<JitArenaAllocator>, alloc);
     toData->changedSyms->Copy(fromData->changedSyms);
+    toData->changedSymsAfterIncBailoutCandidate->Copy(fromData->changedSymsAfterIncBailoutCandidate);
 
     Assert(fromData->HasData());
     toData->OnDataInitialized(alloc);
@@ -2354,6 +2360,9 @@ GlobOpt::DeleteBlockData(GlobOptBlockData *data)
     }
 
     JitAdelete(alloc, data->changedSyms);
+    data->changedSyms = nullptr;
+    JitAdelete(alloc, data->changedSymsAfterIncBailoutCandidate);
+    data->changedSymsAfterIncBailoutCandidate = nullptr;
 
     data->OnDataDeleted();
 }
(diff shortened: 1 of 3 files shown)


ThomsonTan:
@LouisLaf please help take a look.

lib/Backend/GlobOpt.h
```diff
@@ -1008,6 +1009,7 @@ class GlobOptBlockData
     CapturedValues *                        capturedValuesCandidate;
     CapturedValues *                        capturedValues;
     BVSparse<JitArenaAllocator> *           changedSyms;
+    BVSparse<JitArenaAllocator> *           changedSymsAfterIncBailoutCandidate;
```
LouisLaf: This doesn't need to be per-block does it?  One on the GlobOpt object should be enough.

ThomsonTan: Yes, moved it to be a member of GlobOpt.

---
In reply to: [72304117](https://github.com/Microsoft/ChakraCore/pull/1337#discussion_r72304117) [](ancestors = 72304117)

lib/Backend/GlobOpt.cpp
```diff
@@ -307,7 +308,12 @@ GlobOpt::Optimize()
         // Keep track of the last symbol for which we're guaranteed to have data.
         this->maxInitialSymID = this->func->m_symTable->GetMaxSymID();
         this->BackwardPass(Js::BackwardPhase);
+
+        // changedSymsAfterIncBailoutCandidate helps track building incremental bailout in ForwardPass
+        this->changedSymsAfterIncBailoutCandidate = JitAnew(alloc, BVSparse<JitArenaAllocator>, alloc);
```
rajatd: Can you initialize this bit vector along with the other members of GlobOpt in ForwardPass() ?

lib/Backend/GlobOpt.cpp
```diff
@@ -527,6 +532,9 @@ GlobOpt::ForwardPass()
     // Make sure we free most of them.
     Assert(freedCount >= spilledCount);
 
+    JitAdelete(alloc, this->changedSymsAfterIncBailoutCandidate);
```
LouisLaf: That Arena goes away right after returning from this function.  There is no need to free this.

lib/Backend/GlobOpt.cpp
```diff
@@ -5166,9 +5174,15 @@ GlobOpt::OptInstr(IR::Instr *&instr, bool* isInstrRemoved)
 
     if (instr->HasBailOutInfo() && !this->IsLoopPrePass())
     {
-        this->currentBlock->globOptData.changedSyms->ClearAll();
-        this->currentBlock->globOptData.capturedValues = 
-            this->currentBlock->globOptData.capturedValuesCandidate;
+        GlobOptBlockData * globOptData = &this->currentBlock->globOptData;
+        globOptData->changedSyms->ClearAll();
+        globOptData->changedSyms->Or(this->changedSymsAfterIncBailoutCandidate);
```
LouisLaf: Use Copy() instead?

lib/Backend/GlobOpt.cpp
ThomsonTan: Will this-alloc be freed in END_CODEGEN_PHASE, I don't see it is freed in it.

---
In reply to: [72335803](https://github.com/Microsoft/ChakraCore/pull/1337#discussion_r72335803) [](ancestors = 72335803)

lib/Backend/GlobOpt.cpp
ThomsonTan: Remove Or() and use swap instead to reduce the cost.

---
In reply to: [72336744](https://github.com/Microsoft/ChakraCore/pull/1337#discussion_r72336744) [](ancestors = 72336744)

lib/Backend/GlobOpt.cpp
```diff
@@ -7184,13 +7198,36 @@ GlobOpt::SetSymStoreDirect(ValueInfo * valueInfo, Sym * sym)
...
 
 void
+GlobOpt::SetChangedSym(SymID symId)
+{
+    GlobOptBlockData * globOptData = nullptr;
+    if (this->currentBlock->globOptData.changedSyms)
+    {
+        globOptData = &this->currentBlock->globOptData;
+    }
+    else
```
LouisLaf: This is a bit ugly...  Is the only call where this is needed from MergeValues()?  We could just directly set the valueInfo in this case, because your cross-block code already handle the merge here...

lib/Backend/GlobOpt.cpp
```diff
@@ -5166,9 +5174,22 @@ GlobOpt::OptInstr(IR::Instr *&instr, bool* isInstrRemoved)
 
     if (instr->HasBailOutInfo() && !this->IsLoopPrePass())
     {
-        this->currentBlock->globOptData.changedSyms->ClearAll();
-        this->currentBlock->globOptData.capturedValues = 
-            this->currentBlock->globOptData.capturedValuesCandidate;
+        GlobOptBlockData * globOptData = &this->currentBlock->globOptData;
+        globOptData->changedSyms->ClearAll();
```
LouisLaf: Also add comment here explaining what is going on.  The fact that the instr with the bailout might not have a bailout anymore, but if it does you need to use the changedSyms state to the point right after the CaptureValues call.

lib/Backend/GlobOpt.cpp
LouisLaf: It goes away in GlobOpt::Optimize(), right after the all to ForwardPass().

---
In reply to: [72353163](https://github.com/Microsoft/ChakraCore/pull/1337#discussion_r72353163) [](ancestors = 72353163,72335803)

lib/Backend/GlobOpt.cpp
ThomsonTan: Yes, this is only for MergeValues so far. MergeValues could create new value for merge, and merge changed syms for building bailout doesn't handle that. Do we need to be conservative to track these?

---
In reply to: [72355295](https://github.com/Microsoft/ChakraCore/pull/1337#discussion_r72355295) [](ancestors = 72355295)

LouisLaf: But MergeCapturedValues() should handle this no?

---
In reply to: [72358009](https://github.com/Microsoft/ChakraCore/pull/1337#discussion_r72358009) [](ancestors = 72358009,72355295)

ThomsonTan: This looks correct.

---
In reply to: [72362653](https://github.com/Microsoft/ChakraCore/pull/1337#discussion_r72362653) [](ancestors = 72362653,72358009,72355295)

lib/Backend/GlobOpt.cpp
```diff
@@ -7184,13 +7212,29 @@ GlobOpt::SetSymStoreDirect(ValueInfo * valueInfo, Sym * sym)
...
-        this->blockData.changedSyms->Set(prevSymStore->m_id);
+        this->SetChangedSym(prevSymStore->m_id);
     }
     valueInfo->SetSymStore(sym);
 }
 
 void
+GlobOpt::SetChangedSym(SymID symId)
+{
+    GlobOptBlockData * globOptData = &this->currentBlock->globOptData;
```
LouisLaf: >this->currentBlock [](start = 38, length = 18)

Let's just add a comment that currentBlock is the probably wrong block sometimes, but it just makes us overly conservative, not incorrect.

LouisLaf:
:shipit:

lib/Backend/GlobOpt.cpp
ThomsonTan: Seems the below comment in else statement serves the same purpose? globOptData in currentBlock is not initialized but that is handled by MergeCapturedValues?

---
In reply to: [72481209](https://github.com/Microsoft/ChakraCore/pull/1337#discussion_r72481209) [](ancestors = 72481209)

LouisLaf: It's not just MergeCaptureValues though.  The invariant code/PRE and others seem to call SetSymStore for values not in the current block.

dilijev:
@ThomsonTan any update on this PR?

ThomsonTan:
@dilijev , not sure why this one is still shown as open, it was merged. Let me close it manually.

    [MERGE #1337 @ThomsonTan] Fix missing capture dest reg for bailout
    
    Merge pull request #1337 from ThomsonTan:Fix_IncBailOut
    
    When tracking symbols for capturing bailout values incrementally, the dest register for the current instruction might be ignored because the set of changed symbols is cleaned up at the end of OptInstr. The fix captures all the symbols changed after bailout value capture and restore them to changed symbols set when attaching capture values to bailout finally.

dilijev:
@ThomsonTan Thanks!

PR #1343, opened 2016-07-26
Make `for (var x = 0 in ...` a syntax error in strict mode

### LINKED ISSUE #1340: For-in initializer should be a syntax error in strict mode
```js
'use strict';
for(var x = 0 in {}) { }
```

Expected: syntax error
Actual: x is initialized to 0

(This should remain working in sloppy mode. See also: https://github.com/tc39/ecma262/pull/614).

ianwjhalliday:
Fixes #1340

### DIFF
--- a/lib/Parser/Parse.cpp
+++ b/lib/Parser/Parse.cpp
@@ -8737,7 +8737,7 @@ ParseNodePtr Parser::ParseVariableDeclaration(
                 {
                     Error(ERRUnexpectedDefault);
                 }
-                if (pfForInOk && (declarationType == tkLET || declarationType == tkCONST))
+                if (pfForInOk && (declarationType == tkLET || declarationType == tkCONST || IsStrictMode()))
                 {
                     *pfForInOk = FALSE;
                 }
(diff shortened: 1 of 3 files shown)


ianwjhalliday:
@bterlson fyi
@Microsoft/chakra-es could some give this a quick review please

suwc:
:shipit:

tcare:
Looks good.

akroshg:
I didn't confirm with the spec. But assume that this is not valid of ```for (let x = 0``` or ```for (const x = 0```. Also ```for (x = 0 in ``` ?

ianwjhalliday:
@akroshg `let` and `const` are already invalid in this case.  I used the existing mechanism to handle `var` + `strict mode`.  `for (x = 0` form without the `var` is to remain unchanged.

akroshg:
@ianwjhalliday - alright, sounds good.

bterlson:
Yay, good stuff!

ianwjhalliday:
Thanks folks!

PR #1345, opened 2016-07-27
Only add defined properties to the result of getOwnPropertyDescriptors

### LINKED ISSUE #1342: Update `Object.getOwnPropertyDescriptors` with spec changes
Per https://github.com/tc39/ecma262/pull/593 - if the result of `[[GetOwnPropertyDescriptor]]` is undefined, it should *not* be stored on the returned descriptors object.

jordonwii:
Fixes #1342 

For the tests for this change, I adapted @ljharb's tests from the discussion he linked to in the issue and trimmed it down to test just this change.

### DIFF
--- a/test/Object/getOwnPropertyDescriptors.js
+++ b/test/Object/getOwnPropertyDescriptors.js
@@ -153,7 +153,29 @@ var tests = [
 
             verifyObjectDescriptors(desc, allTrueSymbol, allFalseSymbol);
         }
-    }
+    },
+    {
+        name:"For any property, if getOwnPropertyDescriptor(property) is undefined, that property should not be present on the result.",
+        body: function() {
+            // Adapted from: https://github.com/ljharb/test262/blob/c2eaa30b08fb1e041b7297e415b6bad8461f50dc/test/built-ins/Object/getOwnPropertyDescriptors/proxy-undefined-descriptor.js
+            var proxyHandler = {
+              getOwnPropertyDescriptor: function () {},
+            };
+
+            var key = "a";
+            var obj = {};
+            obj[key] = "value";
+
+            var proxy = new Proxy(obj, proxyHandler);
+
+            var descriptor = Object.getOwnPropertyDescriptor(proxy, key);
+            assert.areEqual(undefined, descriptor, "Descriptor matches result of [[GetOwnPropertyDescriptor]] trap");
+
+            var result = Object.getOwnPropertyDescriptors(proxy);
+            assert.isFalse(result.hasOwnProperty(key), "key should not be present in result");
+
+        }
+    },
 ]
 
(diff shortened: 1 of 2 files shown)


test/Object/getOwnPropertyDescriptors.js
```diff
@@ -153,6 +153,24 @@ var tests = [
 
             verifyObjectDescriptors(desc, allTrueSymbol, allFalseSymbol);
         }
+    },
+    {
+        name:"For any property, if getOwnPropertyDescriptor(property) is undefined, that property should not be present on the result.",
+        body: function() {
+            // Adapted from: https://github.com/ljharb/test262/blob/c2eaa30b08fb1e041b7297e415b6bad8461f50dc/test/built-ins/Object/getOwnPropertyDescriptors/proxy-undefined-descriptor.js
+            var proxyHandler = {
+              getOwnPropertyDescriptor: function () {},
```
ljharb: This bug is only truly testable if `ownKeys` returns a key, that `getOwnPropertyDescriptor` returns undefined for - so it's critical to define `ownKeys` to `return [key]`.

jordonwii: Got it. I misunderstood how that was being used, and removed the `ownKeys` bit as a last minute change. I'll add that back in.

akroshg:
Code wise looks good to me. Did you update the test accordingly?
 @ianwjhalliday should have the final review on this.

jordonwii:
@akroshg I didn't have a chance to update it yesterday, but I just did now. I also confirmed that the test fails without my change, and passes with it.

test/Object/getOwnPropertyDescriptors.js
```diff
@@ -153,7 +153,29 @@ var tests = [
...
+            var obj = {};
+            obj[key] = "value";
+
+            var proxy = new Proxy(obj, proxyHandler);
+
+            var descriptor = Object.getOwnPropertyDescriptor(proxy, key);
+            assert.areEqual(undefined, descriptor, "Descriptor matches result of [[GetOwnPropertyDescriptor]] trap");
+
+            var result = Object.getOwnPropertyDescriptors(proxy);
+            assert.isFalse(key in result, "key should not be present in result");
```
ianwjhalliday: The `in` operator also returns true for properties found in the prototype change.  Here we don't expect the prototype change of a normal plain object to have a property named 'a', but it would be better practice to use `obj.hasOwnProperty(key)` instead

ianwjhalliday:
Looks great.  One comment on the test that I think should be addressed, otherwise good to go.

test/Object/getOwnPropertyDescriptors.js
jordonwii: Right. Fixed!

ianwjhalliday:
Thanks @jordonwii

PR #1346, opened 2016-07-27
avoid ch ENABLE_TEST_HOOKS def side effect

jianchun:
Move down ch's ENABLE_TEST_HOOKS def after CommonDefines.h, so that it
does not accidentally turn on ENABLE_DEBUG_CONFIG_OPTIONS for release
build type.


akroshg:
:shipit:

jianchun:
Thanks @akroshg!

FYI @obastemur @curtisman

PR #1347, opened 2016-07-27
Fix a conditional check when determining whether we have an inlinee to box in a native frame

rajatd:
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/1347)


rajatd:
@pleath , please review.

pleath:
👍

PR #1349, opened 2016-07-27
Disable assignment of local registers to delay-captured vars

pleath:
Delay-capture optimization has been effectively disabled by the stable closures change, and assigning local registers to captured vars (as delay-capture needs to do) is causing persistent functional issues. Blocking the assignment of registers in such cases by default to stop the flow of such issues altogether (and restoring the original simple related logic in DefineUserVars). The harder work of re-enabling delay-capture will follow.


pleath:
@rajatd this is related to the couple of fixes you've looked at recently.

rajatd:
:+1:

PR #1350, opened 2016-07-27
Avoid calling ::pow(double, int) in CRT which has a imprecision bug

ThomsonTan:
::pow(double, int) causes significant imprecision for exponents which is not very small. This is fixed in Visual Studio 2015 update 2 but haven't reached us. No side effect of avoid calling it since we optimize integer exponent in our side.

### DIFF
--- a/lib/Runtime/Library/JavascriptNumber.cpp
+++ b/lib/Runtime/Library/JavascriptNumber.cpp
@@ -220,7 +220,8 @@ namespace Js
             }
         }
 
-        return ::pow(x, y);
+        // always call pow(double, double) in C runtime which has a bug to process pow(double, int).
+        return ::pow(x, static_cast<double>(y));
     }
 
 #if _M_IX86


ThomsonTan:
@LouisLaf @rajatd please help take a look.

LouisLaf:
Doesn't this cause large regressions on benchmarks?

ThomsonTan:
Do you mean perf? Should not since we optimization with small integer exponent in our side. The fixed ::pow in new CRT only handles 2 as exponent which is a subset of us, so this change is safe.

---
In reply to: [235738926](https://github.com/Microsoft/ChakraCore/pull/1350#issuecomment-235738926) [](ancestors = 235738926)

LouisLaf:
But why not call DirectPowDoubleInt() here to match the JIT?

ThomsonTan:
This is inside DirectPowDoubleInt.

---
In reply to: [235742358](https://github.com/Microsoft/ChakraCore/pull/1350#issuecomment-235742358) [](ancestors = 235742358)

LouisLaf:
Oh, duh.  I should have looked at the code with codeflow first instead of trying to make sense of it in GitHub.  That looks fine as that's what the code should have been all along.

Please make sure it does not regress the benchmarks though.

LouisLaf:
:shipit:

ThomsonTan:
Thanks for the reminder. Haven't seen obvious variance in Octane on desktop.

---
In reply to: [235743845](https://github.com/Microsoft/ChakraCore/pull/1350#issuecomment-235743845) [](ancestors = 235743845)

LouisLaf:
Try Kraken/Sunspider too as they rely on pow more than Octane.

ThomsonTan:
Checked both, same result, even for math-partial-sum.

---
In reply to: [235758417](https://github.com/Microsoft/ChakraCore/pull/1350#issuecomment-235758417) [](ancestors = 235758417)

PR #1354, opened 2016-07-28
xplat: add eh frame data to thunks

jianchun:
DeferredParsingThunk asm didn't have eh frame data. If an exception is
thrown during its execution (calling JavascriptFunction::DeferredParse),
process will crash because it cannot unwind the stack correctly.

Changed to use asm macroes which adds cfi directives. Applied the same
fix to DeferredDeserializeThunk.

### DIFF
--- a/lib/Runtime/Library/amd64/JavascriptFunctionA.S
+++ b/lib/Runtime/Library/amd64/JavascriptFunctionA.S
@@ -108,14 +108,13 @@ NESTED_END amd64_CallFunction, _TEXT
 
 
 .balign 16
-.text
-C_FUNC(_ZN2Js18JavascriptFunction20DeferredParsingThunkEPNS_16RecyclableObjectENS_8CallInfoEz):
-        push rbp
+NESTED_ENTRY _ZN2Js18JavascriptFunction20DeferredParsingThunkEPNS_16RecyclableObjectENS_8CallInfoEz, _TEXT, NoHandler
+        push_nonvol_reg rbp
         lea  rbp, [rsp]
 
         // save argument registers used by custom calling convention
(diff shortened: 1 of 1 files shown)


lib/Runtime/Library/amd64/JavascriptFunctionA.S
```diff
@@ -108,38 +108,38 @@ NESTED_END amd64_CallFunction, _TEXT
...
 
-        pop rsi
-        pop rdi
-        pop rbp
+        pop_register rsi
+        pop_register rdi
+        pop_nonvol_reg rbp
+NESTED_END _ZN2Js18JavascriptFunction20DeferredParsingThunkEPNS_16RecyclableObjectENS_8CallInfoEz, _TEXT
 
         jmp rax
```
obastemur: @jianchun this needs to be moved before NESTED_END. Same applies to the one below. After that, perhaps it should just work. I don't see anything else problematic here. Didn't check the codes but interesting clang/ld etc didn't bother linking this one on Ubuntu..

jianchun: @obastemur Thanks for looking! Do you mean moving `jmp rax`? (Was troubled by this one. Would look fine if it were a `call` and `ret` before NESTED_END. But it is a jmp.)

jianchun: Trying that, seems working fine after moved on Ubuntu

obastemur:
:shipit:

digitalinfinity:
lgtm

jianchun:
@obastemur @digitalinfinity Thanks!

PR #1355, opened 2016-07-28
Don't let parent function reference point to fake FunctionBody

pleath:
I've observed while investigating a class of site issues that the m_referenceInParentFunc field of the crashing FunctionBody in each case is pointing to the fake global FunctionBody shared by all deferred parses. I haven't been able to prove that this weirdness is responsible for the crashes, but it seems at the least like it's a good thing to fix.


boingoing:
👍

PR #1356, opened 2016-07-28
Let named function expression be eligible for IIFE heuristic

pleath:
Trivial change to our deferred-parsing heuristic for detecting IIFE's: also treat a function expression as an IIFE if it has a name, provided it is preceded by a unary operator or parenthesis. The name consideration dates from legacy mode, in which a function expression's name could be referred to outside the function.


boingoing:
LGTM

PR #1357, opened 2016-07-28
Fix two issues introduced by moving module code into a function body instead of being global code

boingoing:
1. Module export symbols do not get a register allocted and need to be
exempted from debugger symbol tracking by register. This was previously
working but broke after module export symbols moved to be non-global.

2. Some debug scenarios can cause the module body itself to be
defer-parsed. Previous code assumed that could never happen.

### DIFF
--- a/lib/Parser/Parse.cpp
+++ b/lib/Parser/Parse.cpp
@@ -10925,7 +10925,8 @@ ParseNodePtr Parser::Parse(LPCUTF8 pszSrc, size_t offset, size_t length, charcou
         }
     }
 
-    if (isModuleSource && !isDeferred)
+    // It's possible for the module global to be defer-parsed in debug scenarios.
+    if (isModuleSource && (!isDeferred || (isDeferred && grfscr & fscrGlobalCode)))
     {
         ParseNodePtr moduleFunction = GenerateModuleFunctionWrapper<true>();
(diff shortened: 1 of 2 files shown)


boingoing:
@pleath Would you mind taking a quick look?

lib/Parser/Parse.cpp
```diff
@@ -10925,7 +10925,8 @@ ParseNodePtr Parser::Parse(LPCUTF8 pszSrc, size_t offset, size_t length, charcou
         }
     }
 
-    if (isModuleSource && !isDeferred)
+    // It's possible for the module global to be defer-parsed in debug scenarios.
+    if (isModuleSource && (!isDeferred || (isDeferred && grfscr & fscrGlobalCode)))
```
pleath: Is it necessary that we allow the module global be deferred? (I assume this logic is just to make a synthetic scenario happy and that we would never do this in reality?)

boingoing: We hit this when we try to do source rundown and reparse for debug attach scenario. I think it might be a real scenario if user opens F12. Ordinary global functions work fine in that case but the module code needs to generate the function body wrapper or else import/export statements will be treated as syntax error.

boingoing:
@dotnet-bot test this please

PR #1358, opened 2016-07-29
Added shebang to build script

TorbenKoehn:
This avoids build problems when you're using another default shell like zsh.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -1,3 +1,4 @@
+#!/bin/sh
 #-------------------------------------------------------------------------------------------------------
 # Copyright (C) Microsoft. All rights reserved.
 # Licensed under the MIT license. See LICENSE.txt file in the project root for full license information.
--- a/build.sh
+++ b/build.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 #-------------------------------------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


jianchun:
Thanks for the change! Should we use /bin/bash? The script used to have problems with /bin/sh. I think previously we had trouble with the license header checker for this change. Seems that problem is gone. @dilijev

TorbenKoehn:
Absolutely. Changed it.

obastemur:
> Should we use /bin/bash? The script used to have problems with /bin/sh. 

bin/sh forces POSIX. So, the differences. However makes sure the experience due to strict bacward compatibility. On the other hand, for the range of platforms we target, 'bash' makes more sense for us.

obastemur:
LGTM

jianchun:
Thanks @TorbenKoehn, and @obastemur for confirmation. I'll merge this.

dilijev:
@jianchun yes, last time this came up I @obastemur and I fixed the copyright check to support shebangs as the first line of the file before the required copyright header.

jianchun:
@dilijev Thanks for confirm!

dilijev:
For reference, that logic is here: https://github.com/Microsoft/ChakraCore/blob/master/jenkins/check_copyright.py#L46

PR #1360, opened 2016-07-30
Wrong memop assert

Cellule:
Assert in PrepareSegmentForMemop was incorrectly using length to verify that we were at the end of the array.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/1360)

### DIFF
--- a/lib/Runtime/Library/JavascriptArray.inl
+++ b/lib/Runtime/Library/JavascriptArray.inl
@@ -843,7 +843,7 @@ SECOND_PASS:
             if (!endSeg)
             {
                 // end is beyond the length of the array
-                Assert(endIndex == this->length - 1);
+                Assert(endIndex == (current->left + current->length - 1));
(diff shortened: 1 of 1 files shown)


lib/Runtime/Library/JavascriptArray.inl
```diff
@@ -843,7 +843,7 @@ SECOND_PASS:
             if (!endSeg)
             {
                 // end is beyond the length of the array
-                Assert(endIndex == this->length - 1);
+                Assert(endIndex == (current->left + current->length - 1));
```
Cellule: `this->length` refers to the length field of the array, which can be anything really. What we mean to check here is that the end index points to the end of the current segment.

Cellule:
@LouisLaf @rajatd Please review

Cellule:

LouisLaf:
:shipit:

PR #1361, opened 2016-07-30
MemSet incorrect copy for native float array with -1

Cellule:
We used to check if the memset value is -1 to determine if we can do a byte level memset. 
This was valid ints but not for floats. 
Changed the check to really see if the bytes are all the same.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/1361)

### DIFF
--- a/lib/Backend/GlobOpt.cpp
+++ b/lib/Backend/GlobOpt.cpp
@@ -21446,7 +21446,7 @@ GlobOpt::EmitMemop(Loop * loop, LoopCount *loopCount, const MemOpEmitData* emitD
                 case TyInt16:
                 case TyInt32:
                 case TyInt64:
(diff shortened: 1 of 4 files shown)


Cellule:
@LouisLaf @rajatd @pleath Please review

Cellule:
I realized that there were other occurrences of this bug.
I made a function to gather gather them.
For the double type, I only check for 0 to do a `memset` as it is unlikely that all byte on a double be the same.
For Var, I don't even bother checking as it will most likely be a pointer or tagged value.
For Integer, I check if the byte are the same aka `0xabababab` is valid to do a memset.

rajatd:
>             }

remove this?

---
Refers to: lib/Runtime/Library/JavascriptArray.inl:1086 in 4faa24a. [](commit_id = 4faa24af1b57058d78e618687705d2d15d59c24c, deletion_comment = False)

rajatd:
:shipit:

PR #1362, opened 2016-07-30
BUG 8258993: Remove Assert that is causing many hits on the OOM path because of inconsistent finalization flags state.

dilijev:
Reviewed by @yongqu

### DIFF
--- a/lib/Runtime/Base/ScriptContext.cpp
+++ b/lib/Runtime/Base/ScriptContext.cpp
@@ -375,7 +375,6 @@ namespace Js
 
     ScriptContext::~ScriptContext()
     {
-        Assert(isFinalized || !isInitialized);
         // Take etw rundown lock on this thread context. We are going to change/destroy this scriptContext.
         AutoCriticalSection autocs(GetThreadContext()->GetEtwRundownCriticalSection());
 


dilijev:
@yongqu review please?

lib/Runtime/Base/ScriptContext.cpp
```diff
@@ -375,7 +375,7 @@ namespace Js
 
     ScriptContext::~ScriptContext()
     {
-        Assert(isFinalized || !isInitialized);
+        Assert(isFinalized);
```
Yongqu: shouldn't revert. We can get rid of the whole assert. 
I was afraid we might miss some scenario where we would delete scriptContext is wrong time. I think we cover all three now and the assert can be removed:
. finalize time when the scriptcontext is held by library
. half way during initialization of scriptcontext when we OOM; in both browser & JSRT scenario
. JSRT only, fail half way during JSRTContext initialization, but after scriptcontext is fully created.

dilijev: Logically `isFinalized || !isInitialized` seems like a reasonable assert though. Maybe we should fix the state of those flags along error paths so that we could keep the assert and ensure that we've really handled all of the cases...

Yongqu: then we need another state to know that we are half way in JsrtContext initialization code path...

dilijev: Ah I see. Maybe it's not worth doing that bookkeeping? In which case, deleting this assert seems like the right thing to do I guess.

Yongqu: yeah I think we should remove the assert at this time.

Yongqu:
LGTM

dilijev:

Issue #1271, opened 2016-07-13
No FastPath for new Array(constant)


kunalspathak:
Today we don't generate fast path for below case.
```js
var x = new Array(5); // argument should be constant between 0 and 8 inclusive.
```
That is because in [lowerer](https://github.com/Microsoft/ChakraCore/blob/master/lib/Backend/Lower.cpp#L4935) , we fetch the wrong source of instruction. I tried fixing it by extracting the right argument and based on that emit fast path. For below test case I saw 10% win. 
```js
var start = new Date().getTime();
function test() {
  var arr = [];
  for (var j = 0; j < 100; j++) {
    arr.push(new Array(5));
  }
  return arr;
}

for (var i = 0; i < 1000000; i++) {
  test();
}
var end = new Date().getTime();
var time = end - start;
print(time);
```

However, I later realized that `Array` ctor can take multiple parameters that are array elements and we need to emit code to initialize those elements in fast path. If we should/do not want to emit fast path if parameters are array elements (although less than or equal to 8), then my solution should work and I can submit the PR for this.

This was never caught because the only other code path that creates this fast path is `NewScObjArrayNoArg`.

I was also curious why we decided to emit fast path only for array length up to 8 and not beyond. On scanning through benchmarks, I noticed that this would improve Sunspider but other benchmarks like Octane, Jetstream, etc. won't be improved because they create `Array` from ctor of length around 1024 or greater.

kunalspathak:
FYI - @LouisLaf , @curtisman

kunalspathak:
On the second thought, I was wondering if we can also emit array creation fast path for below scenario
```js
new Array(variable); // 0 <= variable <= 8;
```
We can have bound checks on argument passed to `Array` ctor and call helper if argument is beyond specific range. In acme-air benchmark, below are the js functions that call `Array` ctor specifying size through variable instead of constant. 


![array_no_fastpath](https://cloud.githubusercontent.com/assets/12488060/16883000/aebc148c-4a76-11e6-972e-8f005eaa972d.PNG)

I traced the variable value passed to the `Array` ctor and found out all of them were less than 8. Optimizing this case should give us approx. 8% win.
* index.js!proto.handle = 3
* _stream_writable.js!onwrite = 8
* _stream_writable.js!clearBuffer = 2
* net.js!Socket.prototype._writeGeneric = 8
* _stream_readable.js!Readable.prototype.read - less than 5 (through inlining nextTick)
[...]

PR #1363, opened 2016-07-31
Generate fast path for "new Array(val)"

kunalspathak:
As pointed in #1271,  this is second part of generating fast path for `new Array(variable)` case where 0 <= variable <= 8.

Internal test : In progress.
Perf: Slight wins in Octane. In the sample test (modified for variable) mentioned in #1271, this change improves by 15%.

I will update this when I get results.

I have updated the PR based on bucketization logic I mentioned [below](https://github.com/Microsoft/ChakraCore/pull/1363#r74169891). 
Here is the bucketization I came up with for x86 and x64.
old = no. of elements we used to allocation given the array length passed.
new = no. of elements we will allocate given the array length passed.
e.g. In x64, for native int array, if user called `new Array(value); // value = 6`, and if it native float array, we were allocation space for 7 elements. After my change we would allocate space for 9 elements.

image

No regression in performance of benchmarks.

### DIFF
--- a/lib/Backend/Lower.cpp
+++ b/lib/Backend/Lower.cpp
@@ -3763,7 +3763,7 @@ IR::JnHelperMethod GetArrayAllocMemHelper<Js::JavascriptNativeFloatArray>()
 
 template <typename ArrayType>
 IR::RegOpnd *
-Lowerer::GenerateArrayAlloc(IR::Instr *instr, uint32 * psize, Js::ArrayCallSiteInfo * arrayInfo, bool * pIsHeadSegmentZeroed, bool containMissingValues /* = false */)
+Lowerer::GenerateArrayAlloc(IR::Instr *instr, uint32 * psize, Js::ArrayCallSiteInfo * arrayInfo, bool * pIsHeadSegmentZeroed, bool isArrayObjCtor /* = false */)
 {
     Func * func = this->m_func;
     IR::RegOpnd * dstOpnd = instr->GetDst()->AsRegOpnd();
@@ -3777,10 +3777,18 @@ Lowerer::GenerateArrayAlloc(IR::Instr *instr, uint32 * psize, Js::ArrayCallSiteI
     const IR::AutoReuseOpnd autoReuseHeadOpnd(headOpnd, func, false);
     IR::Instr * leaHeadInstr = nullptr;
     bool isHeadSegmentZeroed = false;
+
     if (ArrayType::HasInlineHeadSegment(count))
     {
-        uint32 allocCount = count == 0 ? Js::SparseArraySegmentBase::SMALL_CHUNK_SIZE : count;
-        arrayAllocSize = Js::JavascriptArray::DetermineAllocationSize<ArrayType, 0>(allocCount, nullptr, &alignedHeadSegmentSize);
+        if (isArrayObjCtor)
+        {
+            arrayAllocSize = Js::JavascriptArray::DetermineAllocationSizeForArrayObjects<ArrayType, 0>(count, nullptr, &alignedHeadSegmentSize);
+        }
+        else
+        {
+            uint32 allocCount = count == 0 ? Js::SparseArraySegmentBase::SMALL_CHUNK_SIZE : count;
+            arrayAllocSize = Js::JavascriptArray::DetermineAllocationSize<ArrayType, 0>(allocCount, nullptr, &alignedHeadSegmentSize);
+        }
         leaHeadInstr = IR::Instr::New(Js::OpCode::LEA, headOpnd,
             IR::IndirOpnd::New(dstOpnd, sizeof(ArrayType), TyMachPtr, func), func);
         isHeadSegmentZeroed = true;
@@ -3837,13 +3845,10 @@ Lowerer::GenerateArrayAlloc(IR::Instr *instr, uint32 * psize, Js::ArrayCallSiteI
     }
 #endif
 
-    // For "new Array(value)", we will create missing values. Set flags appropriately.
-    Js::DynamicObjectFlags arrayFlags = containMissingValues ? Js::DynamicObjectFlags::ObjectArrayFlagsTag : Js::DynamicObjectFlags::InitialArrayValue;
-
     // The same at this:
     //  GenerateMemInit(dstOpnd, ArrayType::GetOffsetOfArrayFlags(), (uint16)Js::DynamicObjectFlags::InitialArrayValue, instr, isZeroed);
     //  GenerateMemInit(dstOpnd, ArrayType::GetOffsetOfArrayCallSiteIndex(), arrayCallSiteIndex, instr, isZeroed);
-    GenerateMemInit(dstOpnd, ArrayType::GetOffsetOfArrayFlags(), (uint)arrayFlags | ((uint)arrayCallSiteIndex << 16), instr, isZeroed);
+    GenerateMemInit(dstOpnd, ArrayType::GetOffsetOfArrayFlags(), (uint)Js::DynamicObjectFlags::InitialArrayValue | ((uint)arrayCallSiteIndex << 16), instr, isZeroed);
     GenerateMemInit(dstOpnd, ArrayType::GetOffsetOfLength(), count, instr, isZeroed);
 
     if (leaHeadInstr != nullptr)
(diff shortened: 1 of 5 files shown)


kunalspathak:
FYI - @LouisLaf , @satheeshravi , @rajatd , @pleath

gorg84:
Server Error in '/' Application.

A potentially dangerous Request.Form value was detected from the client (ctl00$ctl00$ctl00$ctl00$cphRoot$cphSite$cphLC$cphC$txtCandidate_Notes_Web="...Maghsoudi <dralireza.maghsoudi...").
Description: Request Validation has detected a potentially dangerous client input value, and processing of the request has been aborted. This value may indicate an attempt to compromise the security of your application, such as a cross-site scripting attack. You can disable request validation by setting validateRequest=false in the Page directive or in the configuration section. However, it is strongly recommended that your application explicitly check all inputs in this case. 

Exception Details: System.Web.HttpRequestValidationException: A potentially dangerous Request.Form value was detected from the client (ctl00$ctl00$ctl00$ctl00$cphRoot$cphSite$cphLC$cphC$txtCandidate_Notes_Web="...Maghsoudi <dralireza.maghsoudi...").

Source Error: 

[No relevant source lines]

Source File: c:\Windows\Microsoft.NET\Framework64\v2.0.50727\Temporary ASP.NET Files\root\a24ecd5a\8d0ac35\App_Web_registerdetails.aspx.2587f01f.5ikejyg9.0.cs    Line: 0 

Stack Trace: 

[HttpRequestValidationException (0x80004005): A potentially dangerous Request.Form value was detected from the client (ctl00$ctl00$ctl00$ctl00$cphRoot$cphSite$cphLC$cphC$txtCandidate_Notes_Web="...Maghsoudi <dralireza.maghsoudi...").]
   System.Web.HttpRequest.ValidateString(String s, String valueName, String collectionName) +11188987
   System.Web.HttpRequest.ValidateNameValueCollection(NameValueCollection nvc, String collectionName) +71
   System.Web.HttpRequest.get_Form() +178
   System.Web.HttpRequest.get_HasForm() +11189223
   System.Web.UI.Page.GetCollectionBasedOnMethod(Boolean dontReturnNull) +124
   System.Web.UI.Page.DeterminePostBackMode() +83
   System.Web.UI.Page.ProcessRequestMain(Boolean includeStagesBeforeAsyncPoint, Boolean includeStagesAfterAsyncPoint) +11155303
   System.Web.UI.Page.ProcessRequest(Boolean includeStagesBeforeAsyncPoint, Boolean includeStagesAfterAsyncPoint) +11154842
   System.Web.UI.Page.ProcessRequest() +91
   System.Web.UI.Page.ProcessRequest(HttpContext context) +240
   ASP.candidate_register_registerdetails_aspx.ProcessRequest(HttpContext context) in c:\Windows\Microsoft.NET\Framework64\v2.0.50727\Temporary ASP.NET Files\root\a24ecd5a\8d0ac35\App_Web_registerdetails.aspx.2587f01f.5ikejyg9.0.cs:0
   System.Web.CallHandlerExecutionStep.System.Web.HttpApplication.IExecutionStep.Execute() +599
   System.Web.HttpApplication.ExecuteStep(IExecutionStep step, Boolean& completedSynchronously) +171


Version Information: Microsoft .NET Framework Version:2.0.50727.5485; ASP.NET Version:2.0.50727.5491

lib/Backend/Lower.cpp
```diff
@@ -3894,6 +4031,111 @@ Lowerer::GenerateProfiledNewScObjArrayFastPath(IR::Instr *instr, Js::ArrayCallSi
...
+    //        CMP s13,          alignedHeadSegSizeOpnd
+    //        JGE $done
+    //        MOV [s12+s14],    0x80000002                ; Save missing element
+    //        s14               = ADD s14, 4              ; headSegmentIndex += 4. NOTE: value 4 could be different for different type of arrays
+    //        s13               = ADD s13, 1              ; index ++
+    //        JMP $LoopTop
+    //
+    //$done
+
+    IR::RegOpnd * indexOpnd = IR::RegOpnd::New(TyUint32, func);
```
curtisman: If we are limiting this fast path to only <= 8 (above in generateRangeCheck), why do we need to generate general initialize code here.
Can't we generate the code as if the user has asked for a constant 8 element, but the only thing you need to do differently is to initialize the array's length to the user specified value?

kunalspathak: > If we are limiting this fast path to only <= 8 (above in generateRangeCheck), 

Once this change goes through, I was thinking of bumping the restriction to 16 instead of 8.

> Can't we generate the code as if the user has asked for a constant 8 element, 

There is a special case where if user specified value is 0, we would just initialize 4 elements (`Js::SparseArraySegmentBase::SMALL_CHUNK_SIZE`) instead of 8.

lib/Backend/Lower.cpp
```diff
@@ -3816,14 +3813,155 @@ Lowerer::GenerateArrayAlloc(IR::Instr *instr, uint32 * psize, Js::ArrayCallSiteI
...
+    IR::Opnd * objectSizeOpnd = IR::IntConstOpnd::New(Js::JavascriptArray::DetermineObjectSize<ArrayType, 0>(), TyUint32, func);
+
+    // Instead of performing (arraySize * sizeof(TElement), we would use shift operation instead.
+    // The multiplier operand calculates by how much we should shift arraySize to left to achieve the same result.
+    IR::Opnd * sizeOfMultiplierOpnd = IR::IntConstOpnd::New(Math::Log2(sizeof(typename ArrayType::TElement)), TyUint32, func);
+    Lowerer::InsertShift(Js::OpCode::Shl_A, false, arraySizeOpnd, sizeOpnd, sizeOfMultiplierOpnd, instr);
+    Lowerer::InsertAdd(false, tempOpnd1, objectSizeOpnd, arraySizeOpnd, instr);
+
+#if defined(_M_IX86_OR_ARM32)
+    Lowerer::InsertAdd(false, tempOpnd2, tempOpnd1, IR::IntConstOpnd::New(HeapConstants::ObjectGranularity - 1, TyUint32, func), instr);
```
satheeshravi: >HeapConstants::ObjectGranularity - 1, TyUint32 [](start = 74, length = 46)

You can use TyVar. and for ObjectGranularity -> you can cast it to IntConstType ?
So you don't need a ifdef here.

lib/Backend/Lower.cpp
```diff
@@ -3816,14 +3813,155 @@ Lowerer::GenerateArrayAlloc(IR::Instr *instr, uint32 * psize, Js::ArrayCallSiteI
...
+        ; ***** size = (arrayLen == 0) ? Js::SparseArraySegmentBase::HEAD_CHUNK_SIZE ? arrayLen *****
+        MOV sizeOpnd,           arrayLenOpnd                  ; Make copy so it doesn't get overwritten by HEAD_CHUNK_SIZE
+        CMP sizeOpnd,           0
+        LGE $L1
+        MOV sizeOpnd,           HEAD_CHUNK_SIZE;
+    $L1:
+        ; ***** calculate alignedArrayAllocationSize and alignedHeadSegSize *****
+        MOV objectSizeOpnd,     objectSize              ; objectSize = sizeof(ArrayType) + sizeof(SparseArraySegment<typename ArrayType::TElement>)
+        arraySize               = MUL sizeOpnd, elementSize     ; elementSize = sizeof(T::TElement) //TODO: Update this comment once get approver that SHL is ok todo.
+        SHL sizeOpnd,           N                               ; where 2^N = elementSize
```
satheeshravi: >= [](start = 76, length = 1)

please use ==
 comment is misleading, otherwise.

lib/Backend/Lower.cpp
```diff
@@ -3816,14 +3813,155 @@ Lowerer::GenerateArrayAlloc(IR::Instr *instr, uint32 * psize, Js::ArrayCallSiteI
...
+        temp1                   = ADD objectSizeOpnd, arraySizeOpnd
+        temp2                   = ADD temp1, (HeapConstants::ObjectGranularity - 1)
+        alignedArrayAllocSize   = AND temp2, ~(HeapConstants::ObjectGranularity - 1)
+        temp1                   = SUB alignedArrayAllocSize, objectSizeOpnd
+        alignedHeadSegSize      = DIV temp1, N                   ; 2^N = sizeof(typename T::TElement)
```
satheeshravi: >= [](start = 71, length = 1)

same here.

lib/Backend/Lower.cpp
```diff
@@ -3816,14 +3813,155 @@ Lowerer::GenerateArrayAlloc(IR::Instr *instr, uint32 * psize, Js::ArrayCallSiteI
...
+    //REVIEW: Assumption is that count will always in InlineHeadSegment because it will always be less than 64.
+    /*
+        ; ***** size = (arrayLen == 0) ? Js::SparseArraySegmentBase::HEAD_CHUNK_SIZE ? arrayLen *****
+        MOV sizeOpnd,           arrayLenOpnd                  ; Make copy so it doesn't get overwritten by HEAD_CHUNK_SIZE
+        CMP sizeOpnd,           0
+        LGE $L1
+        MOV sizeOpnd,           HEAD_CHUNK_SIZE;
+    $L1:
+        ; ***** calculate alignedArrayAllocationSize and alignedHeadSegSize *****
+        MOV objectSizeOpnd,     objectSize              ; objectSize = sizeof(ArrayType) + sizeof(SparseArraySegment<typename ArrayType::TElement>)
```
satheeshravi: > ; [](start = 55, length = 2)

nit: will be good to align all the comments.

lib/Backend/Lower.cpp
```diff
@@ -3816,14 +3813,155 @@ Lowerer::GenerateArrayAlloc(IR::Instr *instr, uint32 * psize, Js::ArrayCallSiteI
...
+    IR::RegOpnd * headOpnd = IR::RegOpnd::New(TyMachPtr, func);
+    const IR::AutoReuseOpnd autoReuseHeadOpnd(headOpnd, func, false);
+    IR::Instr * leaHeadInstr = nullptr;
+
+    //REVIEW: Assumption is that count will always in InlineHeadSegment because it will always be less than 64.
+    /*
+        ; ***** size = (arrayLen == 0) ? Js::SparseArraySegmentBase::HEAD_CHUNK_SIZE ? arrayLen *****
+        MOV sizeOpnd,           arrayLenOpnd                  ; Make copy so it doesn't get overwritten by HEAD_CHUNK_SIZE
+        CMP sizeOpnd,           0
+        LGE $L1
```
satheeshravi: >LGE [](start = 8, length = 3)

you mean JGE ? if so , you should rather use BrGt_A - in this file.

lib/Backend/Lower.cpp
```diff
@@ -3816,14 +3813,155 @@ Lowerer::GenerateArrayAlloc(IR::Instr *instr, uint32 * psize, Js::ArrayCallSiteI
...
+    {
+        Assert(arrayInfo == nullptr);
+    }
+#endif
+
+    // ***** Array object initialization *****
+    GenerateMemInit(dstOpnd, ArrayType::GetOffsetOfArrayFlags(), IR::IntConstOpnd::New((uint16)Js::DynamicObjectFlags::InitialArrayValue, TyUint16, func), instr, true);
+    GenerateMemInit(dstOpnd, ArrayType::GetOffsetOfLength(), sizeOpnd, instr, true);
+
+    if (leaHeadInstr != nullptr)
```
satheeshravi: >if (leaHeadInstr != nullptr) [](start = 4, length = 28)

This can't be false. There is a definition before this line.

lib/Backend/Lower.cpp
```diff
@@ -3816,14 +3813,155 @@ Lowerer::GenerateArrayAlloc(IR::Instr *instr, uint32 * psize, Js::ArrayCallSiteI
...
     GenerateMemInit(headOpnd, Js::SparseArraySegmentBase::GetOffsetOfSize(), alignedHeadSegmentSize, instr, isHeadSegmentZeroed);
     GenerateMemInitNull(headOpnd, Js::SparseArraySegmentBase::GetOffsetOfNext(), instr, isHeadSegmentZeroed);
 
     *pIsHeadSegmentZeroed = isHeadSegmentZeroed;
     return headOpnd;
 }
 
+template <typename ArrayType>
+IR::RegOpnd *
+Lowerer::GenerateArrayAlloc(IR::Instr *instr, IR::Opnd * arrayLenOpnd, Js::ArrayCallSiteInfo * arrayInfo, bool * pIsHeadSegmentZeroed, IR::Opnd ** alignedHeadSegSizeOpnd)
```
satheeshravi: >IR::Opnd ** alignedHeadSegSizeOpnd [](start = 135, length = 34)

do you wanna use SAL for this ?

lib/Backend/Lower.cpp
kunalspathak: I debugged through the code for constant 8 element and found out that for size <= 8, we allocate from 3 different buckets.

constant | allocation size | bucket
----------|-----------------|---------
0,3,4,5,6 | 0x70 bytes | 6 **
1,2 | 0x60 bytes | 5
7,8 | 0x80 bytes | 7

** size 0 falls in this bucket because we allocate `SMALL_CHUNK_SIZE` as mentioned above.

We hard-code the allocation size in JIT code, but if the value is `variable`, I have to calculate the allocation size in JIT. However there are couple of things that I can do here:
1. For `0` replace `SMALL_CHUNK_SIZE` with size that is used today by array of size `1` or `2`. So 0 elements array will get allocated from bucket 5. Alternatively, I could just use `SMALL_CHUNK_SIZE` for size `1` and `2` which will make them allocate from bucket 6. In that case for array size `<= 8 ` we will be just allocate from 2 buckets. 
2. Once that is done, I can replace the allocation size and aligned size calculate code in JIT with  conditional hard-coded allocation size code. Something like this.
```asm
CMP arrayLen, 2 ; assuming 0 is moved to bucket 5
JLE $bucket1
CMP arrayLen, 6
JLE $bucket2
MOV allocSize, 0x80
MOV alignedSize, 0xA
JMP $continue

$bucket1:
   MOV allocSize, 0x60 
   MOV alignedSize, 0x2
   JMP $continue
$bucket2:
   MOV allocSize, 0x70
   MOV alignedSize, 0x6

$continue:
...
```

Let me know your thoughts.

kunalspathak:
@satheeshravi , @curtisman , can you please review this?

curtisman:
Can we have the runtime have matching heuristics?

kunalspathak:
@curtisman , I will update the interpreter to match the same heuristics as JIT.

kunalspathak:
@curtisman , @satheeshravi , could one of you review the latest changes?

satheeshravi:
Why are the slot numbers different for x86 and x64 ? And why are the slot numbers for some cases odd number ?

kunalspathak:
@satheeshravi , if you refer the table slot numbers table (x86/x64) in PR comment, you will see the before/after slots calculation based on user input. That's the reason we have various version of slot numbers.

lib/Runtime/Library/JavascriptArray.h
```diff
@@ -922,6 +935,8 @@ namespace Js
 
         typedef int32 TElement;
 
+        static const uint8 AllocationBucketsCount = 3;
+        static uint allocationBuckets[][3];
```
satheeshravi: >allocationBuckets[][3]; [](start = 20, length = 23)

nit: Chk as discussed

lib/Runtime/Library/JavascriptArray.h
```diff
@@ -118,6 +123,12 @@ namespace Js
         static ushort const MergeSegmentsLengthHeuristics = 128; // If the length is less than MergeSegmentsLengthHeuristics then try to merge the segments
         static uint64 const FiftyThirdPowerOfTwoMinusOne = 0x1FFFFFFFFFFFFF;  // 2^53-1
 
+#if defined(_M_X64_OR_ARM64)
+        static const uint8 AllocationBucketsCount = 3;
+#else
+        static const uint8 AllocationBucketsCount = 2;
+#endif
+        static uint allocationBuckets[][3];
```
satheeshravi: >[] [](start = 37, length = 2)

pls specify allocation bucket size. (AllocationBucketsCount)

satheeshravi: >3 [](start = 40, length = 1)

#define

lib/Runtime/Library/JavascriptArray.h
```diff
@@ -6,6 +6,11 @@
 #pragma once
 
 #define ARRAY_CROSSOVER_FOR_VALIDATE 0
+#define ArrayAllocationBucket(ArrayType, index) ArrayType::allocationBuckets[index][0]
```
satheeshravi: How about using Getxxx and Setxxx  macros - instead of this way ?

---
In reply to: [82290881](https://github.com/Microsoft/ChakraCore/pull/1363#discussion_r82290881) [](ancestors = 82290881)

lib/Backend/Lower.cpp
kunalspathak: Not applicable anymore.

---
In reply to: [73918283](https://github.com/Microsoft/ChakraCore/pull/1363#discussion_r73918283) [](ancestors = 73918283)

lib/Backend/Lower.cpp
```diff
@@ -5006,30 +5222,52 @@ Lowerer::LowerNewScObjArray(IR::Instr *newObjInstr)
...
-            startCallInstr = linkSym->m_instrDef;
-            AssertMsg(startCallInstr->GetArgOutCount(false) == 2, "Generating ArrayFastPath for more than 1 parameter not allowed.");
+            bool emittedFastPath = true;
+            // 2a. If 1st paramter is a variable, emit fast path with checks
+            if (opndOfArrayCtor->IsRegOpnd())
+            {
+                // 3. GenerateFastPath
+                GenerateProfiledNewScObjArrayFastPath(newObjInstr, arrayInfo, arrayInfoAddr, weakFuncRef, helperLabel, labelDone, opndOfArrayCtor);
+            }
+            // 2b. If 1st paramter is a constant, it is in range 0 and upperBoundValue (inclusive)
```
satheeshravi: >1st paramter is a constant [](start = 22, length = 26)

can u assert for this ?

kunalspathak: `GetIntConstValue()` has the required assert `Assert(this->IsIntConst());`

---
In reply to: [82472519](https://github.com/Microsoft/ChakraCore/pull/1363#discussion_r82472519) [](ancestors = 82472519)

satheeshravi:
:shipit:

kunalspathak:
Thanks @curtisman  and @satheeshravi  for the review!

PR #1368, opened 2016-08-02
Fix bug in CheckIfTypeIsEquivalent helper

kunalspathak:
CheckIfTypeIsEquivalent helper is called from JITted code to perform equivalent type check.
It validates if the object's type is present in one of the 8 cached equivalent
types and if yes, updates the guard value and return back to JITted code.
If not, it takes slow path and verifies the prototype of object to that of cached
type. The cached type was always fetched from 0th index of array of cached equivalent types.
If this was nullptr, we were returning back to JITted code saying the type of object is not
equivalent to any cached types and then JITted code would bailout.
Some of the type caches can get collected in `PreSweepCallback` if types are not marked,
hence the assumption to fetch cached type from 0th index is not valid because it might have
got collected while other cached types in the array are still around. Hence we should
fetch a non null equivalent type from cached array of types and then do subsequent slow
path checks.

This change reduces the bailouts of kind `BailOutFailedEquivalentTypeCheck` significantly.

Perf: ~~In progress~~ Shows 2.5% win in acmeair. Others are flat.
Test : Unit test pass, internal tools in progress.

### DIFF
--- a/lib/Runtime/Base/FunctionBody.cpp
+++ b/lib/Runtime/Base/FunctionBody.cpp
@@ -8775,14 +8775,17 @@ namespace Js
             Type *type = reinterpret_cast<Type*>(this->guard->GetValue());
             if (!recycler->IsObjectMarked(type))
             {
-                this->guard->Invalidate();
+                this->guard->InvalidateDuringSweep();
             }
             else
             {
                 isAnyTypeLive = true;
             }
         }
-
+        uint16 nonNullIndex = 0;
+#if DBG
+        bool isGuardValuePresent = false;
+#endif
         for (int i = 0; i < EQUIVALENT_TYPE_CACHE_SIZE; i++)
         {
             Type *type = this->types[i];
@@ -8794,11 +8797,29 @@ namespace Js
                 }
                 else
                 {
+                    // compact the types array by moving non-null types
+                    // at the beginning.
+                    this->types[nonNullIndex++] = type;
                     isAnyTypeLive = true;
+#if DBG
+                    isGuardValuePresent = this->guard->GetValue() == reinterpret_cast<intptr_t>(type) ? true : isGuardValuePresent;
+#endif
                 }
             }
         }
+        if (nonNullIndex > 0)
+        {
+            memset((void*)(this->types + nonNullIndex), 0, sizeof(Js::Type*) * (EQUIVALENT_TYPE_CACHE_SIZE - nonNullIndex));
+        }
+        else if(guard->IsInvalidatedDuringSweep())
+        {
+            // just mark this as actual invalidated since there are no types
+            // present
+            guard->Invalidate();
+        }
 
+        // verify if guard value is valid, it is present in one of the types
+        AssertMsg(!this->guard->IsValid() || isGuardValuePresent, "After ClearUnusedTypes, valid guard value should be one of the cached equivalent types.");
         return isAnyTypeLive;
     }
 
--- a/lib/Runtime/Base/FunctionBody.h
+++ b/lib/Runtime/Base/FunctionBody.h
@@ -73,20 +73,53 @@ namespace Js
     {
         friend class PropertyGuardValidator;
     private:
+        enum GuardValue : intptr_t {
+            Invalidated = 0,
+            Uninitialized = 1,
+            Invalidated_DuringSweep = 2
+        };
         intptr_t value;
+#if DBG
+        bool wasReincarnated = false;
+#endif
     public:
         static PropertyGuard* New(Recycler* recycler) { return RecyclerNewLeaf(recycler, Js::PropertyGuard); }
-        PropertyGuard() : value(1) {}
-        PropertyGuard(intptr_t value) : value(value) { Assert(this->value != 0); }
+        PropertyGuard() : value(GuardValue::Uninitialized) {}
+        PropertyGuard(intptr_t value) : value(value) 
+        { 
+            // GuardValue::Invalidated and GuardValue::Invalidated_DuringSweeping can only be set using
+            // Invalidate() and InvalidatedDuringSweep() methods respectively.
+            Assert(this->value != GuardValue::Invalidated && this->value != GuardValue::Invalidated_DuringSweep); 
+        }
 
         inline static size_t const GetSizeOfValue() { return sizeof(((PropertyGuard*)0)->value); }
         inline static size_t const GetOffsetOfValue() { return offsetof(PropertyGuard, value); }
 
         intptr_t GetValue() const { return this->value; }
-        bool IsValid() { return this->value != 0; }
-        void SetValue(intptr_t value) { Assert(value != 0); this->value = value; }
+        bool IsValid()
+        {
+            return this->value != GuardValue::Invalidated && this->value != GuardValue::Invalidated_DuringSweep;
+        }
+        bool IsInvalidatedDuringSweep() { return this->value == GuardValue::Invalidated_DuringSweep; }
+        void SetValue(intptr_t value)
+        { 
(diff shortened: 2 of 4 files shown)


kunalspathak:
FYI - @LouisLaf , @pleath , @rajatd

satheeshravi:
>         if (cache->IsLoadedFromProto() && type->GetPrototype() != refType->GetPrototype())

@pleath, Why aren't we checking all non-null cached type in the slow path? Is it better off to bailout ? 

---
Refers to: lib/Runtime/Language/JavascriptOperators.cpp:8247 in 7ce8997. [](commit_id = 7ce8997c00764694c085e9dfefdbbedcdc894f18, deletion_comment = False)

pleath:
IIRC the type in slot 0 is the one that's checked by the online jitted code. If that's true it should always hold the last type we saw, and we should never allow it to become null. Is that not the case?

-- Paul

kunalspathak:
>         if (cache->IsLoadedFromProto() && type->GetPrototype() != refType->GetPrototype())

AFAIK, cached types would have same prototype/typeId (and thats why there are equivalent types). I can add an assert to ensure this is the case.

---
In reply to: [236767820](https://github.com/Microsoft/ChakraCore/pull/1368#issuecomment-236767820) [](ancestors = 236767820)

---
Refers to: lib/Runtime/Language/JavascriptOperators.cpp:8247 in 7ce8997. [](commit_id = 7ce8997c00764694c085e9dfefdbbedcdc894f18, deletion_comment = False)

rajatd:
I think we should look into why the caches are being zeroed. 

@kunalspathak , you mentioned the caches can get zeroed if the types weren't marked - that's weird because we pin all the types in an equivalent set at work item creation time. Right now the equivalent type cache can hold a maximum of 8 types, but we have been talking about making it growable. It would be nice to not have to loop through the cache for the first non-null entry.

@pleath - while you are correct that the jitted code checks against the first type in the equivalent set, but if I read the code correctly, it doesn't index into the equivalent type cache or set to get the first type. It loads it from the guard, and we do record the last seen type on the guard (if it was, in fact, equivalent to the other types).

kunalspathak:
@pleath , you are right, the guard value in online JITted code is same as type in slot[0]. So ideally `equivTypes[0] == guard->GetValue()`. However if you see below, in slow path, we will overwrite the guard value with newer type. At the same time, we knock off one of the cached type (for eg. at index 5) and replace it with current type. This might answer Rajat's question on why we collect the old type that was knocked off from cached type and now no longer exist as a guard value. 
Next time we come in `CheckIfTypeIsEquivalent`, the guard value is the one that was present at index 5, however we hard code and check the type at slot 0 which might have got collected now. Hence we should get the non-null type from cached types.

But after re-reading the code and I think instead of looping to fetch the `refType`, I could simply use the guard value as `refType` to check for prototype/typeid.


---
In reply to: [236775413](https://github.com/Microsoft/ChakraCore/pull/1368#issuecomment-236775413) [](ancestors = 236775413)

agarwal-sandeep:
>             return false;

Now if we reach here that means we don't have any type in equivTypes. Better to put an assert that all types from 1 to < EQUIVALENT_TYPE_CACHE_SIZE are nullptr. Will help in catching issue where compaction failed.

---
Refers to: lib/Runtime/Language/JavascriptOperators.cpp:8236 in d633d84. [](commit_id = d633d84a0971a52141cda8e72dc251e03dc21b89, deletion_comment = False)

lib/Runtime/Language/JavascriptOperators.cpp
```diff
@@ -8292,38 +8292,66 @@ namespace Js
...
+
+        // We have some empty slots, let us use those first
+        if (emptySlotIndex != -1)
+        {
+            if (PHASE_TRACE1(Js::EquivObjTypeSpecPhase))
             {
-                cache->nextEvictionVictim = 0;
+                Output::Print(_u("EquivObjTypeSpec: Saving type in unused slot of equiv types cache. \n"));
+                Output::Flush();
             }
```
agarwal-sandeep: Better to use just one line - OUTPUT_TRACE(Js::EquivObjTypeSpecPhase, _u("EquivObjTypeSpec: Saving type in unused slot of equiv types cache. \n"));

kunalspathak:
>             return false;

done.

---
In reply to: [237723552](https://github.com/Microsoft/ChakraCore/pull/1368#issuecomment-237723552) [](ancestors = 237723552)

---
Refers to: lib/Runtime/Language/JavascriptOperators.cpp:8236 in d633d84. [](commit_id = d633d84a0971a52141cda8e72dc251e03dc21b89, deletion_comment = False)

lib/Runtime/Language/JavascriptOperators.cpp
kunalspathak: Yeah, i will just follow the consistency here and keep it PHASE_TRACE1

---
In reply to: [73625408](https://github.com/Microsoft/ChakraCore/pull/1368#discussion_r73625408) [](ancestors = 73625408)

lib/Runtime/Base/FunctionBody.cpp
```diff
@@ -8794,11 +8797,29 @@ namespace Js
...
         }
+        if (nonNullIndex > 0)
+        {
+            memset((void*)(this->types + nonNullIndex), 0, sizeof(Js::Type*) * (EQUIVALENT_TYPE_CACHE_SIZE - nonNullIndex));
+        }
+        else if(guard->IsInvalidatedWhileSweeping())
+        {
+            // just mark this as actual invalidated since there are no types
+            // present
+            guard->Invalidate();
```
pleath: Correct me if this isn't true, but can InvalidateWhileSweeping() not be called when the current guard type is dead but other types in the equivalent cache are live? If that happens, do we really want to invalidate the guard?

kunalspathak: It was added more from perf perspective in `CheckIfTypeIsEquivalent` where first thing that is done is
```c++
if (guard->GetValue() == 0)
{
  return false;
}
```
So if the guard was invalidated while sweeping and none of equiv types are alive, we can return right away from `CheckIfTypeIsEquivalent`. Without this, we would go further and fetch `equivTypes[0]` and that would be `nullptr` and that is when we would have returned.

pleath: What I'm asking is, what prevents us from invalidating the guard and leaving it null while other equivalent types in the cache are alive?

lib/Runtime/Base/FunctionBody.h
```diff
@@ -73,20 +73,37 @@ namespace Js
...
+        bool IsInvalidatedWhileSweeping() { return this->value == 2; }
+        void SetValue(intptr_t value) { Assert(value != 0 || value != 2); this->value = value; }
         intptr_t const* GetAddressOfValue() { return &this->value; }
         void Invalidate() { this->value = 0; }
+        void InvalidateWhileSweeping()
+        { 
+#if DBG
+            wasReincarnated = true;
+#endif
+            this->value = 2;
```
pleath: Can you replace "2" with a defined constant?

lib/Runtime/Base/FunctionBody.cpp
pleath: @kunalspathak pointed out the place where we avoid invalidating if there's still a live type.

pleath:
:shipit:

kunalspathak:
@dotnet-bot  test OSX osx_osx_debug_static please

kunalspathak:

kunalspathak:
@pleath , please review fix for cross context

kunalspathak:

pleath:
👍

kunalspathak:

PR #1372, opened 2016-08-02
Check if script context is closed during RundownSourcesAndReparse

agarwal-sandeep:
Various calls can go out of engine during rundown and reparse and in between script context can get closed. Tighten the checks in RundownSourcesAndReparse

### DIFF
--- a/lib/Runtime/Debug/DebugContext.cpp
+++ b/lib/Runtime/Debug/DebugContext.cpp
@@ -148,11 +148,18 @@ namespace Js
             return hr;
         }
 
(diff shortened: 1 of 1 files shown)


agarwal-sandeep:
@akroshg please take a look

akroshg:
Looks good to me. On a side note - could we add a validation where we set a flag in the attach/detach and we raise fatal error (or assert) when the script context is getting closed when that flag (which we set in debug attach/detach) is still set ?

agarwal-sandeep:
In attach detach we handle close after coming out of reparse and return true if SC/SE is closed. Also now that we handle close in other cases assert/fatal error will hit in all those scenarios as well. Will go with current fix and I think we shouldn't see other issues.

PR #1374, opened 2016-08-02
Trivial save-scope-info assert fix

pleath:
Let an assert in saving scope info handle the case where the global eval block scope exists but is empty.


curtisman:
:shipit:

PR #1376, opened 2016-08-03
Abstract usage of _beginthreadex in PlatformAgnostic class

xanlpz:
PlatformAgnostic::Threads::Utility::BeginThread exports the
functionality we use consistently in Chakra. In particular the
"security" and "thread address" parameters are missing. See MSDN for
details.

The Windows implementation just calls _beginthreadex. On Linux we use
PAL's CreateThread, but in the future we would like to use pthread
directly. Future patches will enable the functionality on !Windows,
for now it remains disabled.

### DIFF
--- a/lib/Common/Common/Jobs.cpp
+++ b/lib/Common/Common/Jobs.cpp
@@ -531,7 +531,7 @@ namespace JsUtil
 
             this->parallelThreadData[i]->processor = this;
             // Make sure to create the thread suspended so the thread handle can be assigned before the thread starts running
-            this->parallelThreadData[i]->threadHandle = reinterpret_cast<HANDLE>(_beginthreadex(0, 0, &StaticThreadProc, this->parallelThreadData[i], CREATE_SUSPENDED, 0));
+            this->parallelThreadData[i]->threadHandle = reinterpret_cast<HANDLE>(PlatformAgnostic::Thread::Create(0, &StaticThreadProc, this->parallelThreadData[i], PlatformAgnostic::Thread::ThreadInitCreateSuspended));
             if (!this->parallelThreadData[i]->threadHandle)
             {
                 HeapDelete(parallelThreadData[i]);
--- a/lib/Common/CommonPal.h
+++ b/lib/Common/CommonPal.h
@@ -391,15 +391,7 @@ inline __int64 _abs64(__int64 n)
     return n < 0 ? -n : n;
 }
 
-// xplat-todo: implement these for JIT and Concurrent/Partial GC
-uintptr_t _beginthreadex(
-   void *security,
-   unsigned stack_size,
-   unsigned ( __stdcall *start_address )( void * ),
-   void *arglist,
-   unsigned initflag,
-   unsigned *thrdaddr);
-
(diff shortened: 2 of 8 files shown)


digitalinfinity:
Thanks @xanlpz! This is a very good start. Couple points of feedback:
1. Since it's a static function in a class that's already within a namespace, I prefer just making it a regular function- PlatformAgnostic::Threading::Create is my preferred name
2. For the init_flag parameter, I prefer we use an enum instead of an int- so that we have a platform agnostic type for the flags.  
3. Currently, you return uintptr_t for the return value- I think this is fine, I was just wondering if there is ever the case where a thread "handle" is not uintptr_t, but say a struct? Which is a roundabout way of asking, should we typedef the return value?
4. Minor stylistic point- generally, we prefer to be explicit about our types- so we prefer unsigned int to just regular unsigned
5. I suspect that the Windows builds are failing because ChakraRuntimePlatformAgnostic is not defined in our MSBuild props files. I think you're better off just adding a project reference (scroll towards the bottom of that file and see how Chakra.Common.Codex.vcxproj is defined). The bigger problem is that PlatformAgnostic today has a dependency on Runtime.Base 😞 - so I think for now, just let ch continue to use _beginthreadex since that code is already in #ifdef _WIN32. That particular behavior of ch is not crucial.

xanlpz:
(sorry for the delay, was on holidays!)

Thanks for the comments! New patch up. In detail:
1): done
2): done
3): to be honest I'm not entirely sure. I guess this also depends on what we want to do on Linux at some point? Maybe a typedef is safer for the future, but maybe we can just change it when we actually need to (I'm assuming it's fine to completely change these APIs at any moment)
4): done
5): ouch, yeah. So I reverted the code in all three binaries, because they were all giving similar problems (although rl and the tests are not even built on Linux atm). So just a couple of files changed in the project in the end.

obastemur:
1 )
> 3. Currently, you return uintptr_t for the return value- I think this is fine, I was just wondering if there is ever the case where a thread "handle" is not uintptr_t, but say a struct? Which is a roundabout way of asking, should we typedef the return value?

Agree with @digitalinfinity and we might use a typedef here 
i.e. How about `ThreadHandle` naming?

We could have other minor issues / discussions on this though, but 
this is a new start on Thread interface and we may improve it over time.

So; The rest LGTM. @xanlpz Thanks for the effort!

xanlpz:
Alright, so introduced a typedef for the return type (it's basically pointless atm but I guess that's to be expected) and cleaned up some class namespacing inside the methods while I was at it.

obastemur:
@xanlpz LGTM, could you please rebase ?

obastemur:
@xanlpz any progress on this? [squash + rebase]

xanlpz:
OK, I think it should be up to date now (somehow previous rebases were missing some commits from master)

obastemur:
@dotnet-bot test this please.

obastemur:
@dotnet-bot test this please.

obastemur:
@xanlpz Thanks for the effort!

PR #1379, opened 2016-08-03
Fixing stack pointer when destructuring at call expression.

akroshg:
The destructuring pattern is wrapped around try..finally, as it has to do iteratorclose when the pattern exhaust. However that pattern can appear at call expression. The finally block resetOut the stack pointer. This created the problem as we were middle of the call expression and the stack pointer (m_outSp and m_outParams) got reseted. In order to fix that we need to cache the m_outSp when we enter the try and restore that when we enter the finally block. We need to restore the m_outParams as well - so we are storing that information in the stack itself. There is no problem in the JIT (or bailout code path) so no fix required.
Added the tests for more clarity.


akroshg:
@pleath  could you review? Internal tools are running past 18 hours and seems to be fine.

pleath:
:shipit:

PR #1380, opened 2016-08-03
Fix bad catch variable bindings

pleath:
Enable binding logic for catch variables under !buildAST. Fixes cases where catch variables in deferred nested functions were mistakenly bound to same-named variables in enclosing functions.

### DIFF
--- a/test/EH/capture.js
+++ b/test/EH/capture.js
@@ -50,3 +50,19 @@ function f10(){
     f12();
 };
 f10();
+
+function outer(g) {
+    function inner() {
+        try {
+            throw 1;
+        }
+        catch(g) {
+            if (g !== 1) 
+                WScript.Echo('g === ' + g + ' in catch');
+        }
+    }
+    inner();
+    if (g !== 'g')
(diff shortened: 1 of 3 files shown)


test/EH/capture.js
```diff
@@ -50,3 +50,19 @@ function f10(){
     f12();
 };
 f10();
+
+function outer(g) {
+    function inner() {
+        try {
+            throw 1;
+        }
+        catch(g) {
```
akroshg: curious what happens when ```catch({g})``` ?

pleath: That's a syntax error. The catch has to take an identifier, not an expression.

pleath: Oh, I'm sorry. I saw double parens there, not {g}. With {g}, g is undefined.

pleath: (And we get the expected g === 1 if we throw {g:1}).

pleath:
@dotnet-bot test Windows arm_debug please

test/EH/capture.js
akroshg: great - looks good to me.

Issue #1062, opened 2016-05-28
Assertion failure in tests when changeHeap is not correctly implemented


Shiney:
In https://github.com/Shiney/ChakraCore/tree/AsmJsChangeHeapAssertionFailure

I have added a new version of the asmjs test relink.js with a small change to one of the constants to make it not match the expected change heap function.

https://github.com/Shiney/ChakraCore/commit/3d7ee3e40a5645b174cd902448e7cf62f258dfa3

https://github.com/Shiney/ChakraCore/blob/AsmJsChangeHeapAssertionFailure/test/AsmJs/relinkFail.js
https://github.com/Shiney/ChakraCore/blob/AsmJsChangeHeapAssertionFailure/test/AsmJs/relinkFail.baseline
https://github.com/Shiney/ChakraCore/blob/AsmJsChangeHeapAssertionFailure/test/AsmJs/rlexe.xml

Running the test with e.g.

runtests -x64debug -dirs asmjs 

Causes an assertion failure as shown below.

2>ASSERTION 40948: (D:\dev\ChakraCore\ChakraCore\lib\Runtime\Base\FunctionBody.c
pp, line 1485) this->GetFunctionProxy() == this
2> Failure: (this->GetFunctionProxy() == this)

Also getting the same issue when trying to implement #65

MikeHolman:
I just tried to repro on release/1.6 bits and seems that the bug is gone.

Issue #65, opened 2016-01-13
Remove change-heap support in asm.js 


abchatra:
See Luke Wagner's reasoning [here](http://discourse.wicg.io/t/request-for-comments-switching-resizing-heaps-in-asm-js/641/22). 

"While this feature works without issue or slowdown in FF/Edge, the change-heap function interferes with the generic closure analysis that allows V8/JSC to detect and optimize the immutability of the heap view variables [11]. In V8, this causes a drastic performance penalty (see issue 39075). If that was the end of the story, I think it'd make sense to hold out until optimizations improve (e.g., on V8 is planning to do full asm.js optimizations3). However, WebAssembly will have a simpler and more direct form of resizing6 along with cross-browser support. Due to the above V8 issue, I don't think many people are using asm.js heap resizing (ALLOW_MEMORY_GROWTH is off by default in Emscripten and recommended against), so to simplify engines, I'd like to remove support for the change-heap function from SpiderMonkey and have suggested to Chakra folks that they do the same. Of course, Emscripten can still support ALLOW_MEMORY_GROWTH by simply emitting almost-but-not-asm"

Shiney:
Is this still something that needs to be done? I'd like to give it a go if it is.

abchatra:
FYI - @MikeHolman 

Yes this still needs to be done. In first cut, you could add a configuration key in [configflagslist.h](https://github.com/Microsoft/ChakraCore/blob/master/lib/Common/ConfigFlagsList.h) to disable change heap api and return false from [here](https://github.com/Microsoft/ChakraCore/blob/master/lib/Runtime/Language/AsmJsModule.cpp#L1089) based on that key. By default this key will be off. 

First step is to disable this feature and second step is to completely remove it.

PR #1381, opened 2016-08-03
Disable change heap config option

Shiney:
This is a fix that creates the config flag for #65 it also removes some assertions that were firing in #1062 as I could not get the tests to work correctly with those assertions failing.

### DIFF
--- a/lib/Runtime/Base/FunctionBody.cpp
+++ b/lib/Runtime/Base/FunctionBody.cpp
@@ -1520,7 +1520,7 @@ namespace Js
 
     ScriptFunctionType * FunctionProxy::EnsureDeferredPrototypeType()
     {
-        Assert(this->GetFunctionProxy() == this);
+        //Assert(this->GetFunctionProxy() == this);
         return deferredPrototypeType != nullptr ?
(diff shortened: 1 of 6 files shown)


lib/Runtime/Base/FunctionBody.cpp
```diff
@@ -1520,7 +1520,7 @@ namespace Js
 
     ScriptFunctionType * FunctionProxy::EnsureDeferredPrototypeType()
     {
-        Assert(this->GetFunctionProxy() == this);
+        //Assert(this->GetFunctionProxy() == this);
```
MikeHolman: we need to fix the assert, we can't just disable it

Shiney: Okay.

I haven't really been able to work out what that assert is for, and it seems to be possible to make it happen without the other changes in this hotfix. (See #1062)

EdMaurer:
@Shiney should we close this out or do you want to keep going with it?

Shiney:
Sorry I can't work on this my current contract doesn't allow me to
contribute intellectual property except for my employer.

PR #1383, opened 2016-08-03
Enable @@toStringTag by default

tcare:
- Enabled @@toStringTag by default. Updated baselines.
- Removed the legacy helpers. We only have one special case now; any external object that would have resulted in [object Object] under new rules reverts back to the class name. This retains compat with both DOM and WinRT.
- Fixed the coverage for proxy which was outdated.
- Fixed OS 8331493, a nightly test that was failing.

### DIFF
--- a/lib/Runtime/Library/JavascriptObject.cpp
+++ b/lib/Runtime/Library/JavascriptObject.cpp
@@ -384,8 +384,7 @@ namespace Js
         Var tag = JavascriptOperators::GetProperty(thisArgAsObject, PropertyIds::_symbolToStringTag, scriptContext); // Let tag be the result of Get(O, @@toStringTag).
 
         // 17. Return the String that is the result of concatenating "[object ", tag, and "]". 
-        if (tag != nullptr && JavascriptString::Is(tag))
-        {
+        auto buildToString = [&scriptContext](Var tag) {
             JavascriptString *tagStr = JavascriptString::FromVar(tag);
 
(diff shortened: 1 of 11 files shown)


tcare:
Also fixes OS 8331493, a nightly test that was failing.

@akroshg @Yongqu please review

lib/Runtime/Library/JavascriptObject.cpp
```diff
@@ -384,8 +384,7 @@ namespace Js
         Var tag = JavascriptOperators::GetProperty(thisArgAsObject, PropertyIds::_symbolToStringTag, scriptContext); // Let tag be the result of Get(O, @@toStringTag).
 
         // 17. Return the String that is the result of concatenating "[object ", tag, and "]". 
-        if (tag != nullptr && JavascriptString::Is(tag))
-        {
+        auto buildToString = [&scriptContext](Var tag) {
```
Yongqu: curious what's the expected output of Object.getOwnPropertyNames(Math)? should we have Symbol.toString output as well?

dilijev:
> Also fixes OS 8331493, a nightly test that was failing.

nit: add this to the PR description so it will be automatically included in the merge commit description.

akroshg:
Looks good to me - it would be be good to remove -tostringtag flag from the rlexe.xml (makes the xml cleaner and when we get rid of this flag completely it will be easy transition).

tcare:
Sounds good.

Yongqu:
:shipit:

PR #1385, opened 2016-08-04
Move api.cpp from Chakra.Common.Common.lib to Chakra.Common.Core.lib

ThomsonTan:
The functions inside api.cpp (like js_memcpy_s) could be called when Chakra.Common.Common.lib should not be linked, so move it to Chakra.Common.Core.lib.


ThomsonTan:
@curtisman could you please help take a look?

curtisman:
:shipit:

PR #1386, opened 2016-08-04
Wasm: brtable entries are LEBs

commonlisp:
For binary format 0xc, brtable entries are now LEBs. This is consistent with the spec and current design. The spec test binaries need to be updated pending sexpr-wasm update.


MikeHolman:
:shipit:

Cellule:
LGTM

PR #1388, opened 2016-08-04
Simple fix to tracking of nested functions containing 'with'

pleath:
Call to a nested function from within a 'with' requires that the function's enclosing scope be represented as an object. We were missing a case where the function is located in a param scope that will later be merge with a body scope. Also adding asserts to verify the object-ness of a scope when we're generating byte code that requires it.

### DIFF
--- a/lib/Runtime/ByteCode/ByteCodeGenerator.cpp
+++ b/lib/Runtime/ByteCode/ByteCodeGenerator.cpp
@@ -2800,7 +2800,9 @@ FuncInfo* PostVisitFunction(ParseNode* pnode, ByteCodeGenerator* byteCodeGenerat
 
             parentFunc->SetChildHasWith();
 
-            if (parentFunc->GetBodyScope()->GetHasOwnLocalInClosure())
+            if (parentFunc->GetBodyScope()->GetHasOwnLocalInClosure() ||
+                (parentFunc->GetParamScope()->GetHasOwnLocalInClosure() &&
+                 parentFunc->GetParamScope()->GetCanMergeWithBodyScope()))
             {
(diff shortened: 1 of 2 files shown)


pleath:

pleath:

pleath:
@aneeshdk Related to merge-ability of param and body scopes.

lib/Runtime/ByteCode/ByteCodeGenerator.cpp
```diff
@@ -2800,7 +2800,9 @@ FuncInfo* PostVisitFunction(ParseNode* pnode, ByteCodeGenerator* byteCodeGenerat
 
             parentFunc->SetChildHasWith();
 
-            if (parentFunc->GetBodyScope()->GetHasOwnLocalInClosure())
+            if (parentFunc->GetBodyScope()->GetHasOwnLocalInClosure() ||
```
aneeshdk: Do we already have a test case for this?

lib/Runtime/ByteCode/ByteCodeGenerator.cpp
```diff
@@ -2800,7 +2800,9 @@ FuncInfo* PostVisitFunction(ParseNode* pnode, ByteCodeGenerator* byteCodeGenerat
...
+                (parentFunc->GetParamScope()->GetHasOwnLocalInClosure() &&
```
aneeshdk: >parentFunc->GetParamScope()->GetHasOwnLocalInClosure() [](start = 17, length = 54)

If the param scope has own local in closure then shouldn't it always be split? Else shouldn't we clear the HasOwnLocalInClosure flag when we merge both scopes?

aneeshdk:
:shipit:

PR #1390, opened 2016-08-05
Make -0 output "-0" when doing toString()

Cellule:
As I was doing some work related to `-0` I realized that we do the following
```
-0..toString() === "0"
0..toString() === "0"
```

Which is very confusing and different than what the other browsers do.

This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/1390)

### DIFF
--- a/lib/Runtime/Library/JavascriptNumber.cpp
+++ b/lib/Runtime/Library/JavascriptNumber.cpp
@@ -891,6 +891,11 @@ namespace Js
             return nanF;
         }
 
+        if (IsNegZero(value))
+        {
+            return scriptContext->GetLibrary()->CreateStringFromCppLiteral(_u("-0"));
+        }
+
         if (IsZero(value))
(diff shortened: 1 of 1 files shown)


agarwal-sandeep:
Sorry didn't see this before, will debugger show "-0" for such case as well?

Cellule:
There were too many problems with the change and I am not sure if it's worth it.
Namely, `parseInt(-0)` was returning `-0` after this change, which makes no sense because `-0` is not an int

bterlson:
This change is incorrect anyway. Spec is that toString of -0 is "0", and all implementations agree. Any differences you see here is differences in the toString behavior of the F12 tools: `-0..toString()` is `-(0..toString())` resulting `-0`. You can observe the differences in the various browser consoles simply by typing -0 in the console and hitting enter - we show "0", others show "-0".

PR #1394, opened 2016-08-05
Fix __declspec(guard(overflow)) for linux build

curtisman:
It is only available in VC compiler

### DIFF
--- a/lib/Common/CommonPal.h
+++ b/lib/Common/CommonPal.h
@@ -27,6 +27,13 @@
     #define __forceinline inline
 #endif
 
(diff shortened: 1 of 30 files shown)


digitalinfinity:
@dotnet-bot test this please

lib/Common/CommonPal.h
```diff
@@ -27,6 +27,13 @@
     #define __forceinline inline
 #endif
 
+// Only VC compiler support overflow guard
```
obastemur: How about gcc's 'stack-protector' compiler flag?

PR #1395, opened 2016-08-05
Add dtor for CapturedValues to make its SListBase members to be exception safe

ThomsonTan:
CapturedValues with 2 members of SListBase is used as stack variable in GlobOpt. The dtor of SListBase will complain if exception happened before destructing CapturedValues.  This fix added dtor to CapturedValues to Reset 2 SListBase member before destruct them.

### DIFF
--- a/lib/Backend/IR.h
+++ b/lib/Backend/IR.h
@@ -26,6 +26,14 @@ struct CapturedValues
     SListBase<ConstantStackSymValue> constantValues;           // Captured constant values during glob opt
     SListBase<CopyPropSyms> copyPropSyms;                      // Captured copy prop values during glob opt
     BVSparse<JitArenaAllocator> * argObjSyms;                  // Captured arg object symbols during glob opt
+
+    ~CapturedValues()
+    {
+        // Reset SListBase to be exception safe. Captured values are from GlobOpt->func->alloc
+        // in normal case the 2 SListBase are empty so no Clear needed, also no need to Clear in exception case
+        constantValues.Reset();
+        copyPropSyms.Reset();
+    }
 };
 
 class LoweredBasicBlock;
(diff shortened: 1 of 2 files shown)


ThomsonTan:
@curtisman could you please take a look?

rajatd:
:shipit:

ThomsonTan:
@dotnet-bot test this please.

lib/Backend/IR.h
```diff
@@ -26,6 +26,14 @@ struct CapturedValues
     SListBase<ConstantStackSymValue> constantValues;           // Captured constant values during glob opt
     SListBase<CopyPropSyms> copyPropSyms;                      // Captured copy prop values during glob opt
     BVSparse<JitArenaAllocator> * argObjSyms;                  // Captured arg object symbols during glob opt
+
+    ~CapturedValues()
+    {
+        // Reset SListBase to be exception safe. Captured values are from GlobOpt->func->alloc
+        // in normal case the 2 SListBase are empty so no Clear needed, also no need to Clear in exception case
+        constantValues.Reset();
```
curtisman: Use clear instead.  Especially if you are not using a local arena, but the Func arena, so the memorywill not be freed after Globopt. and be wasted until the Func finish JIT'ing.

ThomsonTan:
@dotnet-bot test this please.

lib/Backend/IR.h
ThomsonTan: @curtisman , in normal cases, these 2 local SListBase would have 0 elements because it just host the values temporarily, all the values are moved to bailoutInfo at the end of value capture which will be cleared at BailOutInfo::Clear(). The only case is when exception happened and the nodes have not been moved to bailoutInfo. But in this case the function will abort compilation so no need to free nodes to func->alloc aggresively. This idea looks similar to SListBase, the code who uses it need to ensure memory are freed, not the SListBase itself (dtor).

curtisman: Sounds good.

curtisman:
:shipit:

Issue #269, opened 2016-02-04
`WScript.LoadScriptFile` search relative to the script location


Cellule:
Right now, `WScript.LoadScriptFile` will search relative to the current working directory.
After the discussion in #244, I believe this was an oversight and this makes running test files without `rl` more difficult than it should be.
Unless someone as a valid use case to use current working directory, we should change it to use the current script's location.

agarwal-sandeep:
@Cellule   https://github.com/Microsoft/ChakraCore/commit/3c743fd807924004e0054b58fecde145033c6d60 implemented this.

//cc @kfarnung @dilijev

PR #1399, opened 2016-08-06
WScript.GetDirectory() & WScript.GetFileName()

Cellule:
Add WScript.GetFileName() and WScript.GetDirectory(), similar to __filename and __dirname in node.js.

Too often when trying to debug a file, you end up needing to either move to the script's directory or change your debugger to start in that script's directory.
Currently, there is no way to workaround this problem.

Will be able to do 

```
WScript.LoadScriptFile(WScript.GetDirectory() + "../TestHarness.js"); 
```

Then run that script from anywhere.

Additionally, we could append the script's current directory to `WScript.LoadScriptFile` `read` `readbuffer` if it uses relative paths.

---

This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/1399)

### DIFF
--- a/bin/ch/WScriptJsrt.h
+++ b/bin/ch/WScriptJsrt.h
@@ -8,7 +8,16 @@ class WScriptJsrt
 {
 public:
     static bool Initialize();
-
+    struct ContextData
+    {
+        char* scriptFullPath = nullptr;
+        ContextData(const char* fullPath) {
+            scriptFullPath = _strdup(fullPath);
+        }
+        ~ContextData() {
+            delete[] scriptFullPath;
+        }
+    };
     class CallbackMessage : public MessageBase
     {
         JsValueRef m_function;
@@ -91,6 +100,8 @@ private:
     static bool CreateNamedFunction(const char*, JsNativeFunction callback, JsValueRef* functionVar);
     static JsValueRef __stdcall EchoCallback(JsValueRef callee, bool isConstructCall, JsValueRef *arguments, unsigned short argumentCount, void *callbackState);
     static JsValueRef __stdcall QuitCallback(JsValueRef callee, bool isConstructCall, JsValueRef *arguments, unsigned short argumentCount, void *callbackState);
+    static JsValueRef __stdcall GetDirectoryCallback(JsValueRef callee, bool isConstructCall, JsValueRef *arguments, unsigned short argumentCount, void *callbackState);
+    static JsValueRef __stdcall GetFileNameCallback(JsValueRef callee, bool isConstructCall, JsValueRef *arguments, unsigned short argumentCount, void *callbackState);
     static JsValueRef __stdcall LoadScriptFileCallback(JsValueRef callee, bool isConstructCall, JsValueRef *arguments, unsigned short argumentCount, void *callbackState);
     static JsValueRef __stdcall LoadScriptCallback(JsValueRef callee, bool isConstructCall, JsValueRef *arguments, unsigned short argumentCount, void *callbackState);
     static JsValueRef __stdcall LoadModuleCallback(JsValueRef callee, bool isConstructCall, JsValueRef *arguments, unsigned short argumentCount, void *callbackState);
--- a/bin/ch/ch.cpp
+++ b/bin/ch/ch.cpp
@@ -397,6 +397,9 @@ HRESULT CreateAndRunSerializedScript(const char* fileName, LPCSTR fileContents, 
     JsContextRef context = JS_INVALID_REFERENCE, current = JS_INVALID_REFERENCE;
     BYTE *bcBuffer = nullptr;
     DWORD bcBufferSize = 0;
+    WScriptJsrt::ContextData* contextData = nullptr;
+    char fullPathLower[_MAX_PATH];
+    size_t len;
     IfFailGo(GetSerializedBuffer(fileContents, &bcBuffer, &bcBufferSize));
 
     // Bytecode buffer is created in one runtime and will be executed on different runtime.
@@ -408,13 +411,23 @@ HRESULT CreateAndRunSerializedScript(const char* fileName, LPCSTR fileContents, 
     IfJsErrorFailLog(ChakraRTInterface::JsGetCurrentContext(&current));
     IfJsErrorFailLog(ChakraRTInterface::JsSetCurrentContext(context));
 
+    // canonicalize that path name to lower case for the profile storage
+    // REVIEW: Assuming no utf8 characters here
+    len = strlen(fullPath) + 1;
+    for (size_t i = 0; i < len; i++)
+    {
+        fullPathLower[i] = (char)tolower(fullPath[i]);
+    }
+    contextData = new WScriptJsrt::ContextData(fullPath);
+    IfJsErrorFailLog(ChakraRTInterface::JsSetContextData(context, (void*)contextData));
+
     // Initialized the WScript object on the new context
     if (!WScriptJsrt::Initialize())
     {
         IfFailGo(E_FAIL);
     }
 
-    IfFailGo(RunScript(fileName, fileContents, bcBuffer, fullPath));
+    IfFailGo(RunScript(fileName, fileContents, bcBuffer, fullPathLower));
 
 Error:
     if (bcBuffer != nullptr)
@@ -471,6 +484,7 @@ HRESULT ExecuteTest(const char* fileName)
     {
         LPCOLESTR contentsRaw = nullptr;
 
+        char fullPathSrc[_MAX_PATH];
         char fullPath[_MAX_PATH];
         size_t len = 0;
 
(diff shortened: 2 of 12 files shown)


Cellule:
@dotnet-bot test this please

lib/Jsrt/JsrtCommonExports.inc
```diff
@@ -5,6 +5,7 @@
     JsAddRef
     JsRelease
     JsCreateContext
+    JsGetContextUrl
```
ianwjhalliday: Exposing a new JSRT API?  Is this intentional and API reviewed?

ianwjhalliday: This is the only way for a ChakraCore host to get the url I guess?  Probably want to run this by @curtisman if haven't already.

curtisman: Yes new API needs to be reviewed.

In this case, I don't think a new API is necessary. ch should be able to keep track of the script file name it is running, and get the filename and directory.
It is not the "current" script file, but it should be enough for what you need.

Cellule: Alright, I'll see what I can do. I just don't know what kind of information I can get in ch regarding the caller of a WScript method.
Since there can be multiple file after various `WScript.LoadScriptFile()`. Maybe I can make it that the methods returns the path of the initial script (ie from argv) and that might be good enough for the time being.

Cellule:
I used the existing APIs JsSetContextData and JsGetContextData instead.
Whenever we create a new jsrt context, we attach the path used of the script used for that context so we can refer to it later.
This will not work when using `WScript.LoadScriptFile("myfile.js", "self")` because we don't create a new context in this case.

Cellule:
I had to save the path before lowering the case because paths are case sensitive on Unix systems.

Cellule:
Ping

bin/ch/ch.cpp
```diff
@@ -402,13 +405,23 @@ HRESULT CreateAndRunSerializedScript(const char* fileName, LPCSTR fileContents,
     IfJsErrorFailLog(ChakraRTInterface::JsGetCurrentContext(&current));
     IfJsErrorFailLog(ChakraRTInterface::JsSetCurrentContext(context));
 
+    // canonicalize that path name to lower case for the profile storage
+    // REVIEW: Assuming no utf8 characters here
```
ianwjhalliday: how much effort would it be to handle unicode characters?  this seems like it could easily bite someone in the future, particularly folks in non-english speaking regions

ianwjhalliday: Oh I see this is already a review comment from below elsewhere in the code.  I suppose it hasn't been reported as a problem yet then and is okay to leave for now.

build.sh
```diff
@@ -29,6 +29,8 @@ PRINT_USAGE() {
     echo "  -j [N], --jobs[=N]   Multicore build, allow N jobs at once"
```
ianwjhalliday: Are these changes to build.sh meant to be part of this PR?

ianwjhalliday: Oh I see, it was an addon change.  Someone from the linux port team should take a look at this change.

test/host/scriptDirFile.js
```diff
@@ -0,0 +1,25 @@
...
+//-------------------------------------------------------------------------------------------------------
+const __filename = WScript.GetFileName();
+const __dirname = WScript.GetDirectory();
+const {__filename2, __dirname2} = WScript.LoadScriptFile(`${__dirname}\\scriptDirFile_2.js`, "samethread");
+WScript.LoadScriptFile(`${__dirname}\\scriptDirFile_3.js`, "self");
+const {__filename4, __dirname4} = WScript.LoadScriptFile(`${__dirname}\\nestedDir\\scriptDirFile_4.js`, "samethread");
+WScript.LoadScriptFile(`${__dirname}\\nestedDir\\scriptDirFile_5.js`, "self");
+
+if (
+  __dirname !== __dirname2 ||
```
ianwjhalliday: recommend inverting this if/else so the conditions are all showing the assertions instead of the negation of the assertions

test/host/scriptDirFile.js
```diff
@@ -0,0 +1,25 @@
...
+  __dirname !== __dirname3 ||
+  __dirname !== __dirname5 ||
+  __filename !== __filename3 ||
```
ianwjhalliday: also good to leave a note about why the filename is incorrect in the 3 and 5 case, due to the "self" limitation

build.sh
Cellule: Yeah I had to test my changes on linux and wanted some extra support. I can remove this part from the PR and open another one.

build.sh
```diff
@@ -152,6 +170,13 @@ while [[ $# -gt 0 ]]; do
     shift
 done
 
+if [ -z ${CUSTOM_MAKE+x} ]; then
+	CUSTOM_MAKE=
```
obastemur: extra spaces

build.sh
```diff
@@ -29,6 +29,8 @@ PRINT_USAGE() {
     echo "  -j [N], --jobs[=N]   Multicore build, allow N jobs at once"
     echo "  -n, --ninja          Build with ninja instead of make"
     echo "      --xcode          Generate XCode project"
+    echo "      --gen[=Gen]      Give a custom generator to cmake"
+    echo "      --make[=cmd]     Give a custom make command"
```
obastemur: Could you please explain why do we need this?

bin/ch/ch.cpp
```diff
@@ -402,13 +405,23 @@ HRESULT CreateAndRunSerializedScript(const char* fileName, LPCSTR fileContents,
     IfJsErrorFailLog(ChakraRTInterface::JsGetCurrentContext(&current));
     IfJsErrorFailLog(ChakraRTInterface::JsSetCurrentContext(context));
 
+    // canonicalize that path name to lower case for the profile storage
```
obastemur: I'm not sure if I understand this comment but Posix path names are case sensitive. Do you think that is a problem?

bin/ch/WScriptJsrt.h
```diff
@@ -8,7 +8,18 @@ class WScriptJsrt
 {
 public:
     static bool Initialize();
-
+    struct ContextData
+    {
+        char* scriptFullPath = nullptr;
+        ContextData(const char* fullPath) {
+            const size_t fullPathLen = strlen(fullPath) + 1;
```
obastemur: Why not strdup?

build.sh
Cellule: I wanted to use another cmake generator and it felt silly to just keep adding options for every generator.
Therefore, I figured it would be simpler to give the option to pass in anything you want.
in my case, I wanted CodeBlocks project file so I ran `./build.sh --gen=\"CodeBlocks - Unix Makefiles\""`

For the `--make` option, I saw that the make option is currently tied to the generator.
I didn't need it for my case as my generator creates project files, but I could easily see someone needing it for a different generator.

build.sh
```diff
@@ -222,15 +247,20 @@ fi
...
+eval $CMAKE_CMD
 
 _RET=$?
 if [[ $? == 0 ]]; then
     if [[ $MAKE != 0 ]]; then
         $MAKE $MULTICORE_BUILD 2>&1 | tee build.log
         _RET=${PIPESTATUS[0]}
     else
-        echo "Visit given folder above for xcode project file ----^"
+        echo "Visit ${build_directory} for ${CMAKE_GEN_NAME} project file ----^"
```
obastemur: Build directory and printed path may not be the same for xcode project file

build.sh
obastemur: '--gen' part I understood but not sure if we ever need '--make' argument. A bit confusing though

build.sh
ianwjhalliday: If they are two distinct changes I would opt for two separate PRs, but if there aren't any further comments at this point it would be okay to keep them together this time.

bin/ch/ch.cpp
Cellule: This is indeed a problem for this feature, which is why I had to capture the path before we lower it for profile purposes.

Cellule:
I rebased my change and remove the change for linux build script. When I'll have some more time I'll open another issue for that purpose.

bin/ch/ch.cpp
dilijev: Considering we want it to work cross-platform, maybe we should consider not lowering the case for profile storage at all. Could end up with collisions on Unix.

dilijev:
@Cellule any update on this issue?

Cellule:
It's still something I want to do, but I haven't had the time to finish the implementation in the private repo and it is low priority right now

dilijev:
Hi @Cellule, I just realized that this wasn't directly targeted to fix https://github.com/Microsoft/ChakraCore/issues/269 -- I assumed it was related. Is that true?

Cellule:
@dilijev Well it is related, it doesn't fix it in the sense it doesn't change WScript.LoadScriptFile, but it adds enough functionality to achieve the desired result, ie: load a script file with relative path by using current script's path

Cellule:
Closing this for now, there is no point really to keep it open until I find the time to fix what is left.

PR #1400, opened 2016-08-07
Cleanup JavascriptEnumerator interface

curtisman:
The JavascriptEnumerator function GetCurrentAndMoveNext is misleading because.  Because it move to the next one before return the next element that we moved to. Change to MoveAndGetNext.

Also remove MoveNext/GetCurrentIndex/GetCurrentPropertyId, which is equivalent to GetCurrentAndMoveNext. Move all usage of those API to  MoveAndGetNext.

### DIFF
--- a/lib/Runtime/Library/ForInObjectEnumerator.cpp
+++ b/lib/Runtime/Library/ForInObjectEnumerator.cpp
@@ -136,12 +136,7 @@ namespace Js
 
     Var ForInObjectEnumerator::GetCurrentIndex()
     {
-        if (currentIndex)
-        {
-            return currentIndex;
-        }
-        Assert(currentEnumerator != nullptr);
-        return currentEnumerator->GetCurrentIndex();
+        return currentIndex;
     }
 
     BOOL ForInObjectEnumerator::TestAndSetEnumerated(PropertyId propertyId)
@@ -155,11 +150,11 @@ namespace Js
     BOOL ForInObjectEnumerator::MoveNext()
     {
         PropertyId propertyId;
-        currentIndex = GetCurrentAndMoveNext(propertyId);
+        currentIndex = MoveAndGetNext(propertyId);
         return currentIndex != NULL;
     }
 
-    Var ForInObjectEnumerator::GetCurrentAndMoveNext(PropertyId& propertyId)
+    Var ForInObjectEnumerator::MoveAndGetNext(PropertyId& propertyId)
     {
         JavascriptEnumerator *pEnumerator = currentEnumerator;
         PropertyRecord const * propRecord;
@@ -168,7 +163,7 @@ namespace Js
         while (true)
         {
             propertyId = Constants::NoProperty;
-            currentIndex = pEnumerator->GetCurrentAndMoveNext(propertyId, &attributes);
+            currentIndex = pEnumerator->MoveAndGetNext(propertyId, &attributes);
 #if ENABLE_COPYONACCESS_ARRAY
             JavascriptLibrary::CheckAndConvertCopyOnAccessNativeIntArray<Var>(currentIndex);
 #endif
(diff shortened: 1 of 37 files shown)


lib/Runtime/Library/ForInObjectEnumerator.cpp
```diff
@@ -155,11 +150,11 @@ namespace Js
     BOOL ForInObjectEnumerator::MoveNext()
```
Yongqu: are we going to change this sometime as well?

Yongqu:
>                     propertyRecord = scriptContext->GetPropertyName(propertyId);

I'm always a bit confused here: should propertyId & propertyName be consistent at this point? should we assert at this point?

---
Refers to: lib/Runtime/Library/JavascriptObject.cpp:1152 in 871a891. [](commit_id = 871a8912d7fcd84c759f47fe5aaf1f887b807caa, deletion_comment = False)

curtisman:

Yongqu:
>         // Moves to the next element and gets the current value.

Can we update the comment here? I'll try to enumerate combinations of return value I know. Please let me know if I missed something:
. return nullptr. propertyId doesn't matter. we are done
. return undefined. propertyId is the real return value; we are not done yet. 
. return JavascriptString. propertyId should be consistent with JavascriptString. ie, the return value is scriptContext->GetPropertyString()
. we have a special case for es5array enumerator where return PropertyString but propertyId is NoProperty? any other special cases?

---
Refers to: lib/Runtime/Types/JavascriptEnumerator.h:32 in 871a891. [](commit_id = 871a8912d7fcd84c759f47fe5aaf1f887b807caa, deletion_comment = False)

lib/Runtime/Library/ForInObjectEnumerator.cpp
curtisman: This is still used in Chakra.dll

lib/Runtime/Library/JSONParser.cpp
```diff
@@ -131,16 +131,21 @@ namespace JSON
...
+
                         //NOTE: If testing key value call enumerator->GetCurrentValue() to confirm value is correct;
 
-                        AssertMsg(!Js::JavascriptOperators::IsUndefinedObject(propertyNameVar,  undefined) && Js::JavascriptString::Is(propertyNameVar) , "bad enumeration on a JSON Object");
+                        AssertMsg(Js::JavascriptString::Is(propertyNameVar) , "bad enumeration on a JSON Object");
 
-                        if (enumerator->GetCurrentPropertyId(&idMember))
+                        if (idMember != Js::Constants::NoProperty)
                         {
                             Js::Var newElement = Walk(Js::JavascriptString::FromVar(propertyNameVar), idMember, value);
```
Yongqu: Do we need to check for undefined?

lib/Runtime/Library/JSONParser.cpp
```diff
@@ -131,16 +131,21 @@ namespace JSON
...
                             if (Js::JavascriptOperators::IsUndefinedObject(newElement, undefined))
```
Yongqu: should we reverse the check order (IsUndefined first before casting to JavascriptString?

curtisman: huh?  We are checking for undefined on newElement, and we are not casting that to string.

curtisman:
>         // Moves to the next element and gets the current value.

I am planning to refactor is next and change the behavior.  So I will document it then.

---
In reply to: [238279890](https://github.com/Microsoft/ChakraCore/pull/1400#issuecomment-238279890) [](ancestors = 238279890)

---
Refers to: lib/Runtime/Types/JavascriptEnumerator.h:32 in 871a891. [](commit_id = 871a8912d7fcd84c759f47fe5aaf1f887b807caa, deletion_comment = False)

curtisman:
>                     propertyRecord = scriptContext->GetPropertyName(propertyId);

It is, but you need propertyRecord here.

---
In reply to: [238278090](https://github.com/Microsoft/ChakraCore/pull/1400#issuecomment-238278090) [](ancestors = 238278090)

---
Refers to: lib/Runtime/Library/JavascriptObject.cpp:1152 in 871a891. [](commit_id = 871a8912d7fcd84c759f47fe5aaf1f887b807caa, deletion_comment = False)

Yongqu:
:shipit:

PR #1401, opened 2016-08-08
ModuleNamespace Support

Yongqu:
Add support for module namespace. The basic syntax is
    export * as myNS from 'submodule.js'
    The namespace is a sealed object with all the exported properties from the currently
    module. The properties are not changable. They are also "reference property" in that
    it should look up the original property instead of caching the value of the property.
    
    There are three set of properties in the namespace object:
    . builtin Symbol properties, include Symbol.toStringTag & Symbol.iterator at this time.
    . local exports. As the local exports are allocated in the specific slot array.
    . nonlocal exports. We need to forward the call to other referencing module namespace object.
    
    Similarly, the enumerator needs to take care of the different kinds of properties. However, the
    iterator need to iterate through the property names in sorted order. We'll create the iterator
    for a sorted name list. CreateListIterator operation (7.4.8) is implemented as well.
    
    Add a set of unittests to cover the object behavior of namespace object.


lib/Runtime/Base/JnDirectFields.h
```diff
@@ -181,6 +181,7 @@ ENTRY(decodeURI)
 ENTRY(decodeURIComponent)
 ENTRY2(default_, _u("default"))
 ENTRY2(delete_, _u("delete")) // "delete" cannot be an identifier in C++ so using "delete_" instead
+ENTRY2(star, _u("*"))
```
boingoing: Should we indicate this is a symbol (change name to "_star" or similar)? It seems unlikely, but a built-in API might come along with the name of star (String.prototype.star maybe).

lib/Runtime/Language/ModuleNamespace.cpp
```diff
@@ -4,101 +4,133 @@
...
+                    (sourceTextModuleRecord->ResolveImport(localNameId, &importRecord)
+                    && importRecord == nullptr))
+                {
+                    BigPropertyIndex index = sourceTextModuleRecord->GetLocalExportSlotIndexByExportName(exportNameId);
+                    Assert((uint)index < sourceTextModuleRecord->GetLocalExportCount());
+                    SimpleDictionaryPropertyDescriptor<BigPropertyIndex> propertyDescriptor = { index, PropertyModuleNamespaceDefault };
+                    propertyMap->Add(exportNameRecord, propertyDescriptor);
+                }
+            });
         }
```
boingoing: This is pretty cool.

lib/Runtime/Language/ModuleNamespace.cpp
```diff
@@ -4,101 +4,133 @@
...
-        uint totalExportCount = (moduleRecord->GetExportedNames(nullptr) != nullptr) ? moduleRecord->GetExportedNames(nullptr)->Count() : 0;
-        uint localExportCount = moduleRecord->GetLocalExportCount();
-        uint unambiguousNonLocalCount = (nsObject->GetUnambiguousNonLocalExports() != nullptr) ? nsObject->GetUnambiguousNonLocalExports()->Count() : 0;
+        uint totalExportCount = exportedNames != nullptr ? exportedNames->Count() : 0;
+        uint unambiguousNonLocalCount = (this->GetUnambiguousNonLocalExports() != nullptr) ? this->GetUnambiguousNonLocalExports()->Count() : 0;
         Assert(totalExportCount == localExportCount + unambiguousNonLocalCount + unresolvableExportsCount);
 #endif
-        BOOL result = nsObject->PreventExtensions();
+        BOOL result = this->PreventExtensions();
         Assert(result);
```
boingoing: >Assert(result); [](start = 8, length = 15)

Should this throw on failure?

lib/Runtime/Language/ModuleNamespace.cpp
```diff
@@ -111,4 +143,220 @@ namespace Js
...
+        }
+        if (propertyMap != nullptr && propertyMap->TryGetValue(propertyRecord, &propertyDescriptor))
+        {
+            return TRUE;
+        }
+        if (unambiguousNonLocalExports != nullptr)
+        {
+            return unambiguousNonLocalExports->ContainsKey(propertyId);
+        }
+        return FALSE;
```
boingoing: >FALSE [](start = 15, length = 5)

NameSpace object has no prototype so all properties are own properties? Sounds right

lib/Runtime/Language/ModuleNamespace.cpp
```diff
@@ -111,4 +143,220 @@ namespace Js
...
     }
+
+    BOOL ModuleNamespace::HasProperty(PropertyId propertyId)
+    {
+        SimpleDictionaryPropertyDescriptor<BigPropertyIndex> propertyDescriptor;
+        const Js::PropertyRecord* propertyRecord = GetScriptContext()->GetThreadContext()->GetPropertyName(propertyId);
+        if (propertyRecord->IsSymbol())
+        {
+            return this->DynamicObject::HasProperty(propertyId);
+        }
```
boingoing: Is it possible to add properties to the namespace object? If not, this should always return false, right?

boingoing: Oh, this is for @@iterator and @@toStringTag? Nevermind, if so.

---
In reply to: [74163949](https://github.com/Microsoft/ChakraCore/pull/1401#discussion_r74163949) [](ancestors = 74163949)

lib/Runtime/Language/SourceTextModuleRecord.cpp
```diff
@@ -623,6 +644,7 @@ namespace Js
             });
         }
 
+        ModuleNamespace::GetModuleNamespace(this);
         Js::AutoDynamicCodeReference dynamicFunctionReference(scriptContext);
```
boingoing: It looks like this initializes the namespace object immediately? Can we delay that until we try to use resolve it? It might be that no one uses the namespace object for this module.

lib/Runtime/Library/JavascriptLibrary.cpp
```diff
@@ -1037,6 +1051,7 @@ namespace Js
         functionTypeDisplayString = CreateStringFromCppLiteral(_u("function"));
         booleanTypeDisplayString = CreateStringFromCppLiteral(_u("boolean"));
         numberTypeDisplayString = CreateStringFromCppLiteral(_u("number"));
+        moduleTypeDisplayString = CreateStringFromCppLiteral(_u("Module"));
```
boingoing: >Module [](start = 65, length = 6)

Just nitpick but should this also be lowercase to match other display strings?

lib/Runtime/Library/JavascriptListIterator.cpp
```diff
@@ -0,0 +1,70 @@
...
+        ListForListIterator* list = iterator->listForIterator;
+
+        if (list == nullptr)
+        {
+            return library->CreateIteratorResultObjectUndefinedTrue();
+        }
+
+        if (iterator->index >= iterator->count)
+        {
+            // Nulling out the m_string field is important so that the iterator
```
boingoing: >m_string [](start = 31, length = 8)

Is this comment accurate? looks like we are nulling out listForIterator. :)

boingoing:
> 

Extra lines here if you're feeling pedantic.

---
Refers to: lib/Runtime/Types/EdgeJavascriptTypeId.h:129 in 39d3005. [](commit_id = 39d3005fe6fd4423fc1fe81ecbaefdbac8380255, deletion_comment = False)

lib/Runtime/Types/EdgeJavascriptTypeId.h
```diff
@@ -120,6 +120,7 @@ enum TypeId
     TypeIds_ActivationObject = 73,
     TypeIds_SpreadArgument = 74,
     TypeIds_ModuleNamespace = 75,
```
boingoing: >TypeIds_ModuleNamespace [](start = 4, length = 23)

Is this object a "TrueJavascriptObjectType" ?

lib/Runtime/Types/EdgeJavascriptTypeId.h
```diff
@@ -120,6 +120,7 @@ enum TypeId
...
+    TypeIds_ListIterator = 76,
```
boingoing: >TypeIds_ListIterator [](start = 4, length = 20)

Should we group this with the other iterators? Seems like the ListIterator is a built-in dynamic object?

test/es6/module-namespace.js
```diff
@@ -0,0 +1,97 @@
...
+    },
+    // {
+        // name: "import namespace with verification",
+        // body: function () {
+           // // testModuleScript('import * as foo from "moduleExport1.js"; for (var i in foo) WScript.Echo(i + "=" + foo[i]);', '', false);
+           // // testModuleScript('import * as foo from "moduleExport1.js"; foo.verifyNamespace(foo);', '', false);
+           // // testModuleScript('import * as foo1 from "moduleExport1.js"; foo1.changeContent(); foo1.verifyNamespace(foo1);', '', false);
+        // }
+    // },
+    {
```
boingoing: This is disabled due to the module export symbol AV? Should be fixed by #1384 fyi

boingoing:
Cool stuff! 👍

lib/Runtime/Language/ModuleNamespace.cpp
Yongqu: This step is not strictly needed as we don't allow property change in the object operation level (we are not really using the dynamictype operations). More a contract visualization :)

---
In reply to: [74163600](https://github.com/Microsoft/ChakraCore/pull/1401#discussion_r74163600) [](ancestors = 74163600)

lib/Runtime/Language/ModuleNamespace.cpp
Yongqu: that's right :)

---
In reply to: [74168450](https://github.com/Microsoft/ChakraCore/pull/1401#discussion_r74168450) [](ancestors = 74168450,74163949)

lib/Runtime/Language/SourceTextModuleRecord.cpp
Yongqu: yeah. The logic gets a bit messy to do the resolve time initialization with the slot setup etc. One instance per namespace and I don't think the impact would be big, and the code is overall cleaner. We can change that if it becomes an issue.

---
In reply to: [74169052](https://github.com/Microsoft/ChakraCore/pull/1401#discussion_r74169052) [](ancestors = 74169052)

lib/Runtime/Library/JavascriptListIterator.cpp
Yongqu: oops. fixed.

---
In reply to: [74170621](https://github.com/Microsoft/ChakraCore/pull/1401#discussion_r74170621) [](ancestors = 74170621)

lib/Runtime/Types/EdgeJavascriptTypeId.h
Yongqu: yes, it's a Var can be seen in the code.

---
In reply to: [74170764](https://github.com/Microsoft/ChakraCore/pull/1401#discussion_r74170764) [](ancestors = 74170764)

lib/Runtime/Types/EdgeJavascriptTypeId.h
Yongqu: does it matter?? I am a bit hesitate to change order unless necessary...

---
In reply to: [74170867](https://github.com/Microsoft/ChakraCore/pull/1401#discussion_r74170867) [](ancestors = 74170867)

Yongqu:
> 

:) fixed

---
In reply to: [238736817](https://github.com/Microsoft/ChakraCore/pull/1401#issuecomment-238736817) [](ancestors = 238736817)

---
Refers to: lib/Runtime/Types/EdgeJavascriptTypeId.h:129 in 39d3005. [](commit_id = 39d3005fe6fd4423fc1fe81ecbaefdbac8380255, deletion_comment = False)

Yongqu:

PR #1405, opened 2016-08-09
Add OS X to the build status table.

dilijev:
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/1405)


dilijev:
@digitalinfinity review please?
Also, is it safe to add this to the daily build tables for `master` and `linux` branches in the wiki?

digitalinfinity:
:shipit: and yes, fine to add

dilijev:
@digitalinfinity FYI added. Also, I noticed Jenkins builds seem to be using OS X 10.11 - should I update the tables or is it fine to keep saying 10.9 which is the theoretical minimum requirement. (Are we actually testing that specific version anywhere?)

PR #1423, opened 2016-08-11
cross-platform: fix standard output bus error [pal] and format precision

obastemur:
issue: [slow running] JSON/simple.withLog test was randomly failing on xplat. [also crashing]

CC provides string length internally. However customized PAL interface expects null ending.

Apart from the fix, also removed unused PAL source files.


digitalinfinity:
:shipit:

obastemur:

PR #1433, opened 2016-08-12
MSFT 8152328 module initialization assert when one of the dependent was parsed earlier.

Yongqu:
Dependent module record notify the parent module that is is ready during DeclarationInitialization. We should
count the number of unfinished dependent module as those that are not done with DeclarationInitialization not Parse.
The test will be ported to core after we fix up ch test harness.


boingoing:
LGTM 🚢

PR #1437, opened 2016-08-13
cross-platform: protect stack on xplat

obastemur:
We have `__declspec(guard(overflow))` in place for Windows. Adding `-fstack-protector` flag
for a similar expectation on xplat


digitalinfinity:
Hi @obastemur - I don't think -fstack-protector is the same as __declspec(guard(overflow))- i think its more analogous to https://msdn.microsoft.com/en-us/library/8dbf701c.aspx. However, it's probably goodness to have anyway, so LGTM- would be nice if @curtisman or @LouisLaf take a look before merging too

curtisman:
Yes, -fstack-protector is more like /GS.  Please change the commit message.
Also, should we use -fstack-protector-strong instead?

obastemur:
We do not use Linux branch anymore.
Closing this one an going to send it again to master

PR #1440, opened 2016-08-15
Fixes build break on Chinese Windows 10 VS 2015

### LINKED ISSUE #1429: compilation failed for code page issues
chakra compilation failed on chinese version of windows 10 with vs2015 update3, following errors shown:

lib\Runtime\Library\JavascriptPromise.cpp : warning C4819: The file contains a character that cannot be represented in the current code page (936). Save the file in Unicode format to prevent data loss

this can be seen in:

https://github.com/Microsoft/ChakraCore/blob/master/lib/Runtime/Library/JavascriptPromise.cpp

line 54:

// 3. Let promise be ? OrdinaryCreateFromConstructor(NewTarget, "%PromisePrototype%", «[[PromiseState]], [[PromiseResult]], [[PromiseFulfillReactions]], [[PromiseRejectReactions]], [[PromiseIsHandled]] »).

the "»" symbol make the compilation failed

so you can save the file encoded in UTF-8 or just remove the  "»" symbol

ianwjhalliday:
Fixes #1429

Chevron characters appears in the source comments but the compiler on
Chinese Windows 10 believes the code page for the source file cannot
represent these characters.

Fix by converting them to '<<' and '>>'

### DIFF
--- a/lib/Runtime/Library/JavascriptPromise.cpp
+++ b/lib/Runtime/Library/JavascriptPromise.cpp
@@ -51,7 +51,7 @@ namespace Js
         }
         RecyclableObject* executor = RecyclableObject::FromVar(args[1]);
 
-        // 3. Let promise be ? OrdinaryCreateFromConstructor(NewTarget, "%PromisePrototype%", «[[PromiseState]], [[PromiseResult]], [[PromiseFulfillReactions]], [[PromiseRejectReactions]], [[PromiseIsHandled]] »).
+        // 3. Let promise be ? OrdinaryCreateFromConstructor(NewTarget, "%PromisePrototype%", <<[[PromiseState]], [[PromiseResult]], [[PromiseFulfillReactions]], [[PromiseRejectReactions]], [[PromiseIsHandled]] >>).
         JavascriptPromise* promise = library->CreatePromise();
         if (isCtorSuperCall)
         {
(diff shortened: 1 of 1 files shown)


ianwjhalliday:

bterlson:
Should also probably document this practice somewhere as pasting in spec text is fairly common... is it possible to teach the compiler on Chinese Windows 10 about utf-8?

ianwjhalliday:
@bterlson from what I've read it sounds like the compiler should handle utf-8, however I've also come across some stackoverflow posts where people still seemed to have trouble with utf-8 files fed to msvc.

Here I'm erring on the side of caution but you're right that this is a common practice.  I sent a psa mail to the dev team suggesting we be mindful when pasting text into comments.  Maybe we can go one step further and have a validation gate in jenkins.  I'll discuss with @dilijev.

dilijev:
@ianwjhalliday @bterlson If the file is properly encoded in utf8 with BOM would the msvc be able to handle it? I ask because our two options here are:

1. Use a gate to ban non-ASCII (I.e. 7-bit with leading 0) characters from the source
2. Convert all source files to UTF8 with BOM and use a gate to enforce that encoding

bterlson:
I would hardly call a properly encoded UTF-8 file one that includes a BOM. The BOM is useless for UTF-8 and the spec says usage of a BOM with UTF-8 encoded text is not recommended. But, option 2 is clearly the superior option here.

dilijev:
@bterlson I should clarify that I meant for "properly encoded in UTF8" (as opposed to extended ASCII)and "file with a BOM" as separate concerns. I just think the BOM is enough of a disambiguator and a convenient way to check encodings without making the gate scan every byte of the file.

PR #1444, opened 2016-08-16
Increase Extended OpCodes range

Cellule:
This change increases the range of possible opcodes in the interpreter.
Currently we are limited to 511 opcodes (0xFF + 0x100 when using extended prefix).
Now the extended opcodes are encoded using 2 bytes allowing the full range 16 bits range.
Therefore, the opcodes now looks as follow
```
Small Opcode (0x00 - 0xFF)
Extended OpCode (0x0100 - 0xFFFF)

Small OpCode : 1 byte
Small OpCode w/ Medium/Large layout: 1 byte for the prefix + 1 byte for the opcode = 2 bytes
Extended OpCode w/o Medium/Large Layout: 1 byte for the prefix + 2 bytes for the opcode = 3 bytes
```

In addition in this change I made some refactoring around the bytecode reader/writer to have only 1 implementation of the read/write of opcodes.
Increased Asm.js interpreter speed by removing useless branches in the interpreter loop (~80% for zlib interpreted)

This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/1444)

PR #1446, opened 2016-08-16
Wasm: Int64 Interpreter implementation

Cellule:
This change implements Int64 for WebAssembly in the interpreter only (-nonative required to test).
I am creating this review to get an initial review of the change as I am expecting tests failures due to reaching the end of the Extended OpCodes range (https://github.com/Microsoft/ChakraCore/pull/1444)

This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/1446)


lib/Runtime/Math/WasmMath.inl
```diff
@@ -6,9 +6,115 @@
...
+inline T WasmMath::ShrU( T aLeft, T aRight )
+{
+    return aLeft >> aRight;
+}
+
+#ifdef _M_IX86
+template<>
+inline int64 WasmMath::Shl( int64 aLeft, int64 aRight )
+{
+    // on x86 the compiler uses _allshl which returns 0 if aRight >= 64
```
MikeHolman: wasm shifts truncate to number of valid bits to shift, so instead of this can we do `aLeft << (aRight & (sizeof(T)*8-1))`? This should be shared with int32 shift, and it should also be applied to the other shifts as well

lib/Runtime/Math/WasmMath.inl
```diff
@@ -18,8 +124,77 @@ WasmMath::Ctz(int value)
...
+    AssertMsg(false, "WasmMath::PopCnt<int> not implemented for this architecture");
+    return 0;
+#endif
+}
+
+template<> 
+inline int64 WasmMath::PopCnt(int64 value)
+{
+#if _M_X64
+    return __popcnt64(value);
```
MikeHolman: popcnt is only available on very new hardware, can we do dynamic check and if pnpcnt isn't available do the slow algorithm for it?

lib/Runtime/Math/WasmMath.inl
```diff
@@ -81,16 +256,28 @@ inline T WasmMath::Nearest(T value)
...
 {
     return _rotl(aLeft, aRight);
 }
 
-inline int
-WasmMath::Ror(int aLeft, int aRight)
+template<>
+inline int64 WasmMath::Rol(int64 aLeft, int64 aRight)
+{
+    return _rotl64(aLeft, (int)aRight);
```
MikeHolman: what i said for shift count also applies to rotate count (mask with 5 or 6)

lib/WasmReader/WasmBinaryReader.cpp
```diff
@@ -789,22 +789,25 @@ WasmBinaryReader::ReadStartFunction()
...
+    uint maxReads = sizeof(MaxAllowedType) == 4 ? 5 : 10;
+    CompileAssert(sizeof(MaxAllowedType) == 4 || sizeof(MaxAllowedType) == 8);
 
     // LEB128 needs at least one byte
     CheckBytesLeft(1);
 
-    for (UINT i = 0; i < 5; i++, length++)
+    for (uint i = 0; i < maxReads; i++, length++)
     {
         // 5 bytes at most
```
MikeHolman: nit: comment is no longer correct

lib/Runtime/Language/WAsmjsUtils.cpp
```diff
@@ -1,49 +1,258 @@
 //-------------------------------------------------------------------------------------------------------
 // Copyright (C) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE.txt file in the project root for full license information.
+// Portions of this file are copyright 2014 Mozilla Foundation, available under the Apache 2.0 license.
 //-------------------------------------------------------------------------------------------------------
 
+#pragma once
```
MikeHolman: why #pragma once?

Cellule: bad copy/paste same for the Mozilla mention in the copyright

lib/Runtime/Language/InterpreterStackFrame.cpp
```diff
@@ -1543,29 +1536,30 @@ namespace Js
...
                 push layout;
                 call InterpreterStackFrame::GetDynamicRetType;
-                cmp eax, Void;
-                je end;
-                cmp eax, Signed;
-                je end;
-                cmp eax, Float;
-                jne skipFloat;
-                // float
+                mov ecx, eax;
```
MikeHolman: nit: this mov is unnecessary. can do `mov ecx, 1; shl ecx, al`

MikeHolman:
:shipit:

lib/Runtime/Language/InterpreterStackFrame.cpp
Cellule: http://x86.renejeschke.de/html/file_module_x86_id_285.html
`shl ecx, al` is not a legal operation. only `cl` is legal to do a shift using a register

MikeHolman: ah interesting, didn't realize that! had assumed it just needed to be a byteable register

---
In reply to: [75201861](https://github.com/Microsoft/ChakraCore/pull/1446#discussion_r75201861) [](ancestors = 75201861)

PR #1447, opened 2016-08-16
Wasm: Compilation exception message bugfix

Cellule:
Fix bug in the Wasm Exception error message memory handling.
The memory for the string was being freed to soon by the destructor of the c++ exception. Which caused the next SysAllocString to allocate a different string at the same pointer of the previous wasm message.
The string's memory needs to be handled manually.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/1447)


Cellule:
@MikeHolman @commonlisp Can you please review ?

MikeHolman:
:shipit:

PR #1449, opened 2016-08-16
Fastpath for hasOwnProperty == true.

dilijev:
Adds a fastpath for the case when `Object.hasOwnProperty` returns `true` (meaning the property was found on the current object).

This optimization was inspired by the observation that a very common pattern to use `Object.hasOwnProperty` in the following manner:

```
for (var x in obj) {
    if (obj.hasOwnProperty(x)) {
        // do something
    }
}
```

From our observations, most instances of these patterns are used with shallow objects (objects with most of their enumerable properties on the object itself, rather than in the prototype chain), so most of the time, `Object.hasOwnProperty` would return true when invoked via the above pattern. Additionally, we have already created a property cache containing the type that a property was enumerated from, so we can use that information to trivially know whether the `this` parameter of `hasOwnProperty` was the same object on which we looked up that property, and if so, we know the answer is `true`.

### DIFF
--- a/lib/Runtime/Base/ScriptContext.cpp
+++ b/lib/Runtime/Base/ScriptContext.cpp
@@ -1454,21 +1454,32 @@ if (!sourceList)
         return NULL;
     }
 
-
-    PropertyString* ScriptContext::GetPropertyString(PropertyId propertyId)
+    PropertyString* ScriptContext::TryGetPropertyString(PropertyId propertyId)
     {
         PropertyStringCacheMap* propertyStringMap = this->GetLibrary()->EnsurePropertyStringMap();
 
-        PropertyString *string;
         RecyclerWeakReference<PropertyString>* stringReference;
         if (propertyStringMap->TryGetValue(propertyId, &stringReference))
         {
-            string = stringReference->Get();
+            PropertyString *string = stringReference->Get();
             if (string != nullptr)
             {
                 return string;
             }
         }
+
+        return nullptr;
+    }
+
+    PropertyString* ScriptContext::GetPropertyString(PropertyId propertyId)
+    {
+        PropertyString *string = TryGetPropertyString(propertyId);
+        if (string != nullptr)
+        {
+            return string;
+        }
+
+        PropertyStringCacheMap* propertyStringMap = this->GetLibrary()->EnsurePropertyStringMap();
 
         const Js::PropertyRecord* propertyName = this->GetPropertyName(propertyId);
         string = this->GetLibrary()->CreatePropertyString(propertyName);
--- a/lib/Runtime/Language/JavascriptOperators.cpp
+++ b/lib/Runtime/Language/JavascriptOperators.cpp
@@ -1075,10 +1075,9 @@ CommonNumber:
 
     BOOL JavascriptOperators::HasOwnProperty(Var instance, PropertyId propertyId, ScriptContext *requestContext)
     {
-        BOOL result;
         if (TaggedNumber::Is(instance))
         {
-            result = false;
+            return FALSE;
         }
         else
         {
@@ -1091,10 +1090,22 @@ CommonNumber:
             }
             else
             {
+                PropertyString *propString = requestContext->TryGetPropertyString(propertyId);
+                if (propString != nullptr)
+                {
+                    const PropertyCache *propCache = propString->GetPropertyCache();
+                    if (object->GetType() == propCache->type)
+                    {
+                        // The type cached for the property was the same as the type of this object
+                        // (i.e. obj in obj.hasOwnProperty), so we know the answer is "true".
+                        Assert(TRUE == (object && object->HasOwnProperty(propertyId))); // sanity check on the fastpath result
+                        return TRUE;
+                    }
+                }
+
                 return object && object->HasOwnProperty(propertyId);
             }
         }
-        return result;
     }
 
     BOOL JavascriptOperators::GetOwnAccessors(Var instance, PropertyId propertyId, Var* getter, Var* setter, ScriptContext * requestContext)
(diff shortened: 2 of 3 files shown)


dilijev:
@curtisman @agarwal-sandeep please review?

dilijev:

lib/Runtime/Language/JavascriptOperators.cpp
```diff
@@ -1091,10 +1090,30 @@ namespace Js
...
             else
             {
-                return object && object->HasOwnProperty(propertyId);
+#ifdef DEBUG
+                // In debug builds, calculate expectedAnswer up-front so that
+                // we can Assert the fastpath gets the right result.
+                BOOL expectedAnswer = object && object->HasOwnProperty(propertyId);
+#endif
+
+                PropertyString *propString = requestContext->GetPropertyString(propertyId);
```
agarwal-sandeep: >GetPropertyString [](start = 61, length = 17)

This goes and create a PropertyString if one is not present in that case type below won't match, should we have a different function which returns only if PropertyString is present.

lib/Runtime/Language/JavascriptOperators.cpp
```diff
@@ -1091,10 +1090,30 @@ namespace Js
...
+                {
+                    // The type cached for the property was the same as the type of this object
+                    // (i.e. obj in obj.hasOwnProperty), so we know the answer is "true".
+                    Assert(expectedAnswer == TRUE); // sanity check on the fastpath result
+                    return TRUE;
+                }
+
+#ifndef DEBUG
+                // In release builds, wait until here to calculate the result.
+                BOOL expectedAnswer = object && object->HasOwnProperty(propertyId);
```
agarwal-sandeep: >BOOL expectedAnswer =  [](start = 16, length = 22)

No need to BOOL expectedAnswer, just return object && object->HasOwnProperty(propertyId);

ljharb:
I'm not sure I'm understanding what's going on in this PR, but could this also add a fast path for `Object.prototype.hasOwnProperty.call(obj, prop)`?

dilijev:
@ljharb This code is the builtin for the function Object.hasOwnProperty, so we will always get this advantage no matter how that function is invoked. (I'll add a test case just to be sure.)

I'll add some more information above about what this change accomplishes.

ljharb:
awesome, thanks :-)

lib/Runtime/Language/JavascriptOperators.cpp
dilijev: Great catch. #Resolved

lib/Runtime/Language/JavascriptOperators.cpp
dilijev: I did this because if I had the return line calculate that as well, then in DEBUG builds we will calculate it twice. The compiler will take care of removing the redundant variable assignment for the release build. #Resolved

lib/Runtime/Language/JavascriptOperators.cpp
dilijev: Great catch.

---
In reply to: [75036981](https://github.com/Microsoft/ChakraCore/pull/1449#discussion_r75036981) [](ancestors = 75036981)

lib/Runtime/Language/JavascriptOperators.cpp
dilijev: I did this because if I had the return line calculate that as well, then in DEBUG builds we will calculate it twice. The compiler will take care of removing the redundant variable assignment for the release build.

---
In reply to: [75037033](https://github.com/Microsoft/ChakraCore/pull/1449#discussion_r75037033) [](ancestors = 75037033)

agarwal-sandeep:
>             string = stringReference->Get();

Move PropertyString *string; from above to here, as we don't need it if propertyStringMap->TryGetValue fails

---
Refers to: lib/Runtime/Base/ScriptContext.cpp:1465 in 17986f4. [](commit_id = 17986f46cb104cb73a9eb37fedbd8bd025d961eb, deletion_comment = False)

lib/Runtime/Base/ScriptContext.cpp
```diff
@@ -1470,6 +1469,19 @@ if (!sourceList)
             }
         }
 
+        return nullptr;
+    }
+
+    PropertyString* ScriptContext::GetPropertyString(PropertyId propertyId)
+    {
+        PropertyString *string = TryGetPropertyString(propertyId);
+        if (string)
```
agarwal-sandeep: >if (string) [](start = 8, length = 11)

if (string != nullptr)

lib/Runtime/Language/JavascriptOperators.cpp
```diff
@@ -1091,10 +1090,32 @@ namespace Js
...
+
+                PropertyString *propString = requestContext->TryGetPropertyString(propertyId);
+                if (propString != nullptr)
+                {
+                    const PropertyCache *propCache = propString->GetPropertyCache();
+                    if (object->GetType() == propCache->type)
+                    {
+                        // The type cached for the property was the same as the type of this object
+                        // (i.e. obj in obj.hasOwnProperty), so we know the answer is "true".
+                        Assert(expectedAnswer == TRUE); // sanity check on the fastpath result
```
obastemur: Could be nicer to remove all `ifdef DEBUG` lines above and put `Assert(object && object->HasOwnProperty(propertyId))` here instead of `expectedAnswer == TRUE`

agarwal-sandeep:
:shipit:

lib/Runtime/Language/JavascriptOperators.cpp
dilijev: Great suggestion, thanks! Done :)

dilijev:

lib/Runtime/Base/ScriptContext.cpp
dilijev: Thanks, done!

PR #1450, opened 2016-08-16
cross-platform: fix memmove for overlapping memory cases

obastemur:
Also added a simple sort test to trigger hybrid_sort implementation

### DIFF
--- a/pal/src/safecrt/memmove_s.c
+++ b/pal/src/safecrt/memmove_s.c
@@ -6,48 +6,92 @@
 /***
 *memmove_s.c - contains memmove_s routine
 *
-
 *
 *Purpose:
 *       memmove_s() copies a source memory buffer to a destination buffer.
 *       Overlapping buffers are treated specially, to avoid propagation.
 *
 *Revision History:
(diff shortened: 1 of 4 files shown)


pal/src/safecrt/memmove_s.c
```diff
@@ -6,48 +6,92 @@
...
+    void * dst,
+    const void * src,
+    size_t count
+)
+{
+#if defined(__APPLE__) || defined(__FreeBSD__)
+    if (src <= dst && src + count > dst)
+    {
+        char *temp = (char*) malloc(count);
+        _VALIDATE_RETURN_ERRCODE(temp != NULL, NULL);
```
digitalinfinity: Did you mean to supply a return code for the second parameter?

digitalinfinity:
:shipit:

pal/src/safecrt/memmove_s.c
obastemur: I've just reused the definition while it was fitting

obastemur:

PR #1452, opened 2016-08-17
runtests.py: default timeout for slow running tests 180 secs on *nix

obastemur:
Some of the tests (slow running) randomly fails on default timeout.
Although timeout can be set from command line for different set of runs,
it is convenient to have a bigger timeout for slow running tests

Affects `--include-slow` runs on unix/posix

### DIFF
--- a/test/runtests.py
+++ b/test/runtests.py
@@ -34,6 +34,9 @@ test a single file:
     runtests.py Basics/hello.js
 ''')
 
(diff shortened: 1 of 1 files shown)


obastemur:

jianchun:
Consider put 180 into a named constant too.

Looks fine to me.

obastemur:
> Consider put 180 into a named constant too.

Good idea. did that too

PR #1456, opened 2016-08-17
Updating async function implementation to fix the symbol capturing issues

aneeshdk:
Updating async function implementation by desugaring to generators

This change removes the function wrapper implementation of async
functions. Now the async functions will behave similar to generator
functions. The actual function body will be in the sciptFunction field.
Symbols now need not be forced to slots. asyncspawn opcode and parse
node are no more needed.In the backend async functions are also
added to the generator JIT flag. await is desugared into yield.


aneeshdk:
@ianwjhalliday @pleath @akroshg Can you please take a look?

lib/Backend/Func.cpp
```diff
@@ -846,8 +846,8 @@ Func::AjustLocalVarSlotOffset()
 bool
 Func::DoGlobOptsForGeneratorFunc()
 {
-    // Disable GlobOpt optimizations for generators initially. Will visit and enable each one by one.
-    return !m_jnFunction->IsGenerator();
+    // Disable GlobOpt optimizations for generators and asyn functions initially. Will visit and enable each one by one.
```
ianwjhalliday: >asyn [](start = 56, length = 4)

typo

lib/Backend/Func.h
```diff
@@ -281,12 +281,6 @@ static const unsigned __int64 c_debugFillPattern8 = 0xcececececececece;
         Assert(this->m_jnFunction);     // For now we always have a function body
         return this->m_jnFunction->GetIsGlobalFunc();
     }
-    bool IsGeneratorFunc() const
```
ianwjhalliday: dead code?

lib/Backend/BailOut.cpp
```diff
@@ -1437,7 +1437,7 @@ BailOutRecord::BailOutHelper(Js::JavascriptCallStackLayout * layout, Js::ScriptF
     Js::InterpreterStackFrame* newInstance = nullptr;
     Js::Var* allocation = nullptr;
 
-    if (executeFunction->IsGenerator())
+    if (executeFunction->IsGenerator() || executeFunction->IsAsync())
```
ianwjhalliday: IsGenerator || IsAsync comes up a lot.

I recommend we combine these into a new IsCoroutine() method

lib/Runtime/ByteCode/ByteCodeEmitter.cpp
```diff
@@ -1227,11 +1227,11 @@ Js::RegSlot ByteCodeGenerator::DefineOneFunction(ParseNode *pnodeFnc, FuncInfo *
     // If we are in a parameter scope and it is not merged with body scope then we have to create the child function as an inner function
     if (regEnv == funcInfoParent->frameDisplayRegister || regEnv == funcInfoParent->GetEnvRegister())
     {
-        m_writer.NewFunction(pnodeFnc->location, pnodeFnc->sxFnc.nestedIndex, pnodeFnc->sxFnc.IsGenerator());
+        m_writer.NewFunction(pnodeFnc->location, pnodeFnc->sxFnc.nestedIndex, pnodeFnc->sxFnc.IsGenerator() || pnodeFnc->sxFnc.IsAsync());
```
ianwjhalliday: Again, IsCoroutine would be better than IsGenerator || IsAsync here

ianwjhalliday:
>                     sym->GetScope()->GetFunc()->root->sxFnc.IsAsync());

I think this IsAsync check should be removed from this assert now

---
Refers to: lib/Runtime/ByteCode/ByteCodeGenerator.cpp:2434 in 8727492. [](commit_id = 87274929d4e81b8a24e8e9e90f1b36f5f8119186, deletion_comment = False)

ianwjhalliday:
>                 if (parentFunc->root->sxFnc.IsAsync())

This should be removed, no longer true

---
Refers to: lib/Runtime/ByteCode/ByteCodeGenerator.cpp:5037 in 8727492. [](commit_id = 87274929d4e81b8a24e8e9e90f1b36f5f8119186, deletion_comment = False)

ianwjhalliday:
>     if (pnode->sxFnc.IsAsync())

Note for later, this may not be necessary with how we fix default parameter expression evaluation.  The try catch could probably go in the C++ code that does the initial next() call on the generator.

---
Refers to: lib/Runtime/ByteCode/ByteCodeEmitter.cpp:2897 in 8727492. [](commit_id = 87274929d4e81b8a24e8e9e90f1b36f5f8119186, deletion_comment = False)

lib/Runtime/Library/JavascriptGeneratorFunction.cpp
```diff
@@ -86,6 +129,63 @@ namespace Js
...
+    {
+        PROBE_STACK(function->GetScriptContext(), Js::Constants::MinStackDefault);
+        ARGUMENTS(stackArgs, callInfo);
+
+        ScriptContext* scriptContext = function->GetScriptContext();
+        JavascriptLibrary* library = scriptContext->GetLibrary();
+
+        Var* argsHeapCopy = RecyclerNewArray(scriptContext->GetRecycler(), Var, stackArgs.Info.Count);
+        js_memcpy_s(argsHeapCopy, sizeof(Var) * stackArgs.Info.Count, stackArgs.Values, sizeof(Var) * stackArgs.Info.Count);
+        Arguments heapArgs(callInfo, argsHeapCopy);
```
ianwjhalliday: Are we getting two heap copies of the arguments since we're doing another heap allocation and copy above in CreateGenerator?  We only need one heap allocated copy of the arguments.

Please check.

lib/Runtime/Library/JavascriptLibrary.cpp
```diff
@@ -835,6 +862,19 @@ namespace Js
...
+        if (scriptContext->GetConfig()->IsES6FunctionNameEnabled())
+        {
+            return JavascriptLibrary::GetDeferredAsyncFunctionTypeHandlerBase</*isNameAvailable*/ true>();
+        }
+        else
+        {
+            return JavascriptLibrary::GetDeferredAsyncFunctionTypeHandlerBase</*isNameAvailable*/ false>();
+        }
+
+    }
```
ianwjhalliday: nit: remove blank line above

lib/Runtime/Library/JavascriptPromise.cpp
```diff
@@ -965,7 +965,7 @@ namespace Js
         JavascriptPromiseAsyncSpawnStepArgumentExecutorFunction* asyncSpawnStepArgumentExecutorFunction = JavascriptPromiseAsyncSpawnStepArgumentExecutorFunction::FromVar(function);
         Var argument = asyncSpawnStepArgumentExecutorFunction->GetArgument();
 
-        JavascriptFunction* next = JavascriptFunction::FromVar(JavascriptOperators::GetProperty(asyncSpawnStepArgumentExecutorFunction->GetGenerator(), PropertyIds::next, function->GetScriptContext()));
+        JavascriptFunction* next = function->GetScriptContext()->GetLibrary()->EnsureGeneratorNextFunction();
```
ianwjhalliday: Just occurred to me that this is more than an optimization, it fixes a bug where the user could have changed async method behavior by changing the generator prototype next property.

Glad we are making this change to the implementation.

test/es7/asyncawait-functionality.js
```diff
@@ -675,6 +676,85 @@ var tests = [
             });  
         }
     },
+    {
```
ianwjhalliday: I forget, have you run these tests with force defer parse?  Curious to know if deferred parse is truly now a moot issue.

ianwjhalliday:
I found a few cases where there was old code specific to async functions that no longer applied, in bytecode generator.  I did not do a look through our code base though.  Please do a quick look through search results for IsAsync to see if there are any other special cases that should be removed.

ianwjhalliday:
nit: the title of this PR is incorrect.  We were desugaring before, now we are implementing it natively (taking advantage of heavy reuse of JavascriptGeneratorFunction).

lib/Backend/Func.cpp
aneeshdk: Fixed

lib/Backend/Func.h
aneeshdk: Yes

lib/Backend/BailOut.cpp
aneeshdk: Done

lib/Runtime/ByteCode/ByteCodeEmitter.cpp

lib/Runtime/Library/JavascriptGeneratorFunction.cpp
aneeshdk: Removed it.

lib/Runtime/Library/JavascriptLibrary.cpp
aneeshdk: Done

test/es7/asyncawait-functionality.js
aneeshdk: I did run all test cases with force:deferparse and all tests are passing fine.

lib/Parser/ptree.h
```diff
@@ -338,6 +338,7 @@ struct PnFnc
     bool IsDeclaration() const { return HasFlags(kFunctionDeclaration); }
     bool IsGeneratedDefault() const { return HasFlags(kFunctionIsGeneratedDefault); }
     bool IsGenerator() const { return HasFlags(kFunctionIsGenerator); }
+    bool IsCoroutine() const { return HasFlags(kFunctionIsGenerator) || HasFlags(kFunctionIsAsync); }
```
ianwjhalliday: HasFlags accepts multiple flags, so you could do `HasFlags(kFunctionIsGenerator | kFunctionIsAsync)` instead.  Doesn't matter though.

Or at least HasFlags _should_ support multiple flags, I'm not sure if that's ever been used and thus tested

lib/Runtime/Library/JavascriptGeneratorFunction.cpp
```diff
@@ -86,6 +129,50 @@ namespace Js
...
     }
 
+    Var JavascriptGeneratorFunction::EntryAsyncFunctionImplementation(RecyclableObject* function, CallInfo callInfo, ...)
+    {
+        PROBE_STACK(function->GetScriptContext(), Js::Constants::MinStackDefault);
+        ARGUMENTS(stackArgs, callInfo);
+
+        ScriptContext* scriptContext = function->GetScriptContext();
+        JavascriptLibrary* library = scriptContext->GetLibrary();
+        DynamicObject* generatorPrototype = scriptContext->GetLibrary()->CreateGeneratorConstructorPrototypeObject();
```
ianwjhalliday: I think this object is not accessible since it is meant for the .prototype property of the async/generator virtual script function object.  We might be able to get away with putting javascript null here for both async and generator implementations.  Otherwise we're getting wasted allocations for every invocation of generator and async functions.

I think, please confirm.

test/es7/asyncawait-functionality.js
akroshg: How about with force undo defer?

---
In reply to: [75369702](https://github.com/Microsoft/ChakraCore/pull/1456#discussion_r75369702) [](ancestors = 75369702)

lib/Runtime/Library/JavascriptLibrary.cpp
```diff
@@ -5869,6 +5934,15 @@ namespace Js
         return RecyclerNewEnumClass(this->GetRecycler(), EnumFunctionClass, JavascriptGeneratorFunction, type, scriptFunction);
     }
 
+    JavascriptAsyncFunction* JavascriptLibrary::CreateAsyncFunction(JavascriptMethod entryPoint, GeneratorVirtualScriptFunction* scriptFunction)
+    {
+        Assert(scriptContext->GetConfig()->IsES6GeneratorsEnabled());
```
akroshg: Is this the right assert here?

akroshg:
Overall looks good to me. Are you planning to add debugger tests as well?

lib/Runtime/Library/JavascriptLibrary.cpp
aneeshdk: Yes, we need the generator functionality all available for the async functions. But right now generators are enabled by default still I will just keep it there for now.

aneeshdk:
I will add debugger test cases in a separate change set

dilijev:

PR #1457, opened 2016-08-17
Fast paths for strings, likely boolean and likely objects for opcodes CmEq_A, CmNeq_A, CmSrEq_A and CmSrNeq_A

agarwal-sandeep:
We have fast paths for equivalent Br* opcodes so I refactored the code
and reused it for Cm* opcodes. Basically Cm* opcodes behave as branch
where we store success if equal otherwise false in destination.


agarwal-sandeep:
@LouisLaf please take a look

LouisLaf:
:shipit:

PR #1459, opened 2016-08-17
WebAssembly i32.popcnt + tests

Krovatkin:
interpreter and codegen implementations for i32.popcnt fyi: @commonlisp @arunetm 

@cellulle @mikeholman homeless PR looks for a nice review and branch to be adopted into


MikeHolman:
:shipit:

MikeHolman:
FYI @Cellule. You guys might want to work out how to handle merging this, since your int64 change has some of this already (implemented slightly differently)

arunetm:
@MikeHolman can you please help to sign off this change on the signing tool as well. 
@Cellule I can help with merging this change. Can this be merged directly to the WebAssembly branch as there could be conflicts with the int64 changes.

MikeHolman:
@arunetm Done.

test/wasm/rlexe.xml
```diff
@@ -35,4 +35,18 @@
...
+       <files>i32_popcnt.js</files>
+       <baseline>i32_popcnt.baseline</baseline>
+       <compile-flags>-on:Wasm</compile-flags>
+    </default>
+</test>
+<test>
+    <default>
+       <files>i32_popcnt.js</files>
+       <baseline>i32_popcnt.baseline</baseline>
+       <compile-flags>-on:Wasm -maic:0</compile-flags>
```
Cellule: this test is not needed as rl.exe will run a the other one with -forcenative

Cellule:
The int64 change will come later so I will take care of resolving conflicts, it should not be too bad as I hadn't done the codegen part.

Krovatkin:
@MikeHolman @Cellule, this is something i think i should prlly mention.
@commonlisp expressed a concern that moving `popcnt` to `Math` might interfere with inlining. 
I ran a small test to check if it gets inlined in the only caller of popcnt in Chakra `uint CharBitvec::Count() const` and it seems to be the case by just looking at disassembly for `CharBitvec::Count()`

test/wasm/rlexe.xml
Krovatkin: @Cellule right :-) let me update the PR

MikeHolman:
@Krovatkin I wouldn't worry about inlining, LTCG should take care of this just fine. But still good to hear confirmation.

test/wasm/rlexe.xml
Krovatkin: done

Cellule:
LGTM

PR #1462, opened 2016-08-18
Fix bug in assert in AddShadowedData()

ianwjhalliday:
DictionaryPropertyDescriptor<>::AddShadowedData() has an assert at the
end to verify the data property index is correct after the shadowing.
However GetDataPropertyIndex<>() relies on the PropertyLetConstGlobal
attribute being set on the Attributes field in order to return correct
results and this was being set after the call to AddShadowedData.

Fixed by moving the attribute flag set into AddShadowedData before the
assertion.

Found by internal web crawler, VSO bug [8482800](https://microsoft.visualstudio.com/OS/_workitems?id=8482800).

### DIFF
--- a/lib/Runtime/Types/DictionaryPropertyDescriptor.h
+++ b/lib/Runtime/Types/DictionaryPropertyDescriptor.h
@@ -164,6 +164,7 @@ namespace Js
         {
             this->Getter = nextPropertyIndex++;
         }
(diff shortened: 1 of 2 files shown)


ianwjhalliday:
@curtisman please review

curtisman:
:shipit:

ianwjhalliday:

PR #1463, opened 2016-08-18
Make eval("arguments") work inside of lambdas.

### LINKED ISSUE #1209: Unable to access arguments in eval inside an arrow method
``` JS
function f3(a) {
    b = () => eval("arguments[0]");
    print(b()); // Should print 1
}
f3(1);
```

This throws an error saying arguments is undefined

jordonwii:
Fixes #1209 

This is fairly close to the solution @aneeshdk @ianwjhalliday and I discussed. There's one minor difference: in order for the new location of `AddVarsToScope` to be as clear as possible, we planned to move it above _all_ of the paramScope code. Sadly, the call the `pushScope(paramScope)` needs to happen prior to `AddVarsToScope` to avoid an assertion failure in ByteCodeGenerator.cpp:1659 due to `currentScope` not being updated. 

I couldn't come up with a less repetitive way to test this than to just copy/paste the bulk of the "`arguments` in lambdas" test, and replace the use of `arguments` with `eval("arguments")`. Definitely open to suggestions here.


lib/Runtime/ByteCode/ByteCodeGenerator.cpp
```diff
@@ -3168,6 +3178,16 @@ void VisitNestedScopes(ParseNode* pnodeScopeList, ParseNode* pnodeParent, ByteCo
...
 
+                /*
+                We need this call to happen *before* visiting the param scope. If we have a split scope,
+                and this function has parameters that default to a lambda which uses the implicit "arguments" inside
+                of eval, we need this function's argumentsSymbol to be set already, so that we can mark it as
+                non-locally referenced inside of PostVisitFunction.
+
+                See Github issue #1209 for more.
+                */
+                AddVarsToScope(pnodeScope->sxFnc.pnodeVars, byteCodeGenerator);
```
pleath: These symbols have to be added to the body scope. Does that not mean that the call has to happen where it currently happens, after we've begun processing the body scope?

jordonwii: This works because `AddVarsToScope` explicitly adds vars to the current function's *body* scope. It uses `ByteCodeGenerator::AddSymbolToFunctionScope`, which in turn calls `ByteCodeGenerator:AddSymbolToScope` with `scope = currentScope->GetFunc()->GetBodyScope()`.

pleath: The real problem, I think, is that the processing of the arguments symbol is still being handled by a method that's specific to the body scope. How about if we move that logic to the processing of the param scope?

jordonwii: Good question. I don't have a solid answer to that. If you're referring to the setting of `argumentsSymbol` inside of `AddVarsToScope`, my thinking would be that we set it there because that already has to do a pass over all of the variables in the body scope, and doing it elsewhere would mean doing two passes. In the more common case where we don't have a split scope, we don't do any param scope processing logic, so we can't only set it there. 

@ianwjhalliday thoughts?

jordonwii:
It turns out that in my first push of the test, I mistakenly included an eval call inside of the new test body in `lambda1.js`, and that buried some failures which I'm currently trying to figure out. In the debug builds, there's an assertion failure I don't understand yet, and in the test builds, `arguments.callee` gets set to `assert.areEqual` rather than the lambda's non-lambda parent.

lib/Runtime/ByteCode/ByteCodeGenerator.cpp
pleath: We can restructure things so that two passes are not necessary, i.e., make the arguments symbol available without searching the body scope vars to find it. Going forward, we're going to create real symbols for this, super, and new.target to staunch the trickle of special-casey bugs involving those constructs, so whatever we invent to allow easy processing of "arguments" in the param scope will work for those other symbols as well.

ianwjhalliday: And indeed @aneeshdk and I discussed making the arguments symbol a normal symbol instead of a special case, to remove a lot of this extra work.  After some code review and discussion we concluded that there is no reason to have a specially handled arguments symbol and it should just work if we make it a normal symbol (with some special care to ensure it is available regardless whether the user script declared it or not).

ianwjhalliday: Out of scope for this PR of course.

ianwjhalliday: I am in favor of jiggering this code however needed to make it work for this PR.

Then we should very soon go and make the arguments symbol not a special case.

jordonwii:
I can't seem to reproduce either the test failure or the assertion failure outside of running `runtests.cmd`. The failures only happen in the "dynapogo" variant, and if I copy/paste the command it reports that it's running and run ch.exe, there are no failures. Any ideas as to why this might happen?

dilijev:
Contacted @jordonwii to follow up on this PR.

jordonwii:
Progress, at last. :)

It looks like the errors I mentioned in my last comment were the result of the jit trying to inline the outer function, when it shouldn't have been able to because the inner Lambda used eval. The code added in these last few commits fixes that by marking the outer function arguments object as referenceable. I haven't figured out how to verify that this isn't over compensating somehow, but right now it appears to work.

If this looks good, I'll flatten the commits tonight and add a couple comments, especially to the tests.

test/es6/lambda1.js
```diff
@@ -270,6 +270,52 @@ var tests = [
         }
     },
     {
+        name: "eval('arguments') works correctly within lambdas",
+        body: function () {
+            var arguments = 'not arguments object';
+            assert.areEqual(arguments, (() => eval("arguments"))(), "arguments in lambda should bind to outside arguments since it is not given its own arguments binding");
```
dilijev: To clarify this, should each instance of `arguments` in this line bind to the var called `arguments` or to an actual arguments object?

jordonwii: Both should refer to the variable ('not an arguments object'). The `var arguments` line shadows the default arguments object, so both the first param to `areEqual` and the `arguments` inside the lambda should refer to that.

dilijev:
@pleath: I contacted @jordonwii offline to discuss this PR and he said he was under the impression that this PR would be superceded by other refactoring work or another change. Is that the case? If so (regardless of whether that work is in progress or not), we should close this PR.

dilijev:
Per offline discussion with @pleath I'm closing this issue.

PR #1468, opened 2016-08-19
Fix call syntax after class expression syntax

### LINKED ISSUE #1465: Class expressions should be allowed in a member expression without parenthesizing
```js
print(class foo { }.toString());
```
Actual: Syntax Error
Expected: `"class foo { }"`

ianwjhalliday:
Call syntax was being disallowed after a call expression.  Appears to be
a mistake, perhaps copy and paste error.  Fix by simply remove line
disallowing call syntax in class expression parsing.

Fixes #1465

### DIFF
--- a/lib/Parser/Parse.cpp
+++ b/lib/Parser/Parse.cpp
@@ -3146,7 +3146,6 @@ LFunction :
     }
 
     case tkCLASS:
-        fAllowCall = FALSE;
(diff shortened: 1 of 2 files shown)


ianwjhalliday:
@Microsoft/chakra-es 
@tcare or @suwc

suwc:
:shipit:

tcare:
Looks good.

PR #1469, opened 2016-08-19
Change LICENSE.txt to use (C) instead of (c) to unify with source code headers.

dilijev:
Also has the benefit of better conforming with Microsoft recommendation and common practice.

https://en.wikipedia.org/wiki/Copyright_symbol#Digital_representation

> Because the © symbol has long been unavailable on typewriters and ASCII-based computer systems, it has been common to approximate this symbol with the characters (C).

### DIFF
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -1,7 +1,7 @@
 ChakraCore
 The MIT License (MIT)
 
-Copyright (c) Microsoft Corporation
+Copyright (C) Microsoft Corporation
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal


dilijev:
@dotnet-bot 
test Windows arm_debug
test Windows x64_debug
test Windows x86_release

dilijev:

dilijev:
@liminzhu review please?

liminzhu:
@dilijev Is there any semantic meaning in this change? Seems like other MS repos are using (c) ([example](https://github.com/PowerShell/PowerShell/blob/master/LICENSE.txt)).
@bterlson Can we make such change to our license?

dilijev:
No semantic meaning other than common practice, and that the C in the circle is upper-case, not lower-case. However, I seem to have misremembered the recommendation saying it should be a capital C, what I'm seeing now is that the copyright C symbol should be rendered in ASCII as `Copyright (c)`.

> Localization
> Replace © with “Copyright (c)” where © will be misrendered by language changes. Unicode will render the © correctly.

So now the remaining point is that the source code uses `Copyright (C)` with a capital C, which is in violation of the recommendation. Do we update all the source files, or leave it alone?

bterlson:
Casing of C doesn't matter at all AFAICT.

dilijev:
Okay I'm closing this PR. I don't think there's any meaningful action to take here.

PR #1476, opened 2016-08-22
Include release branches after release/1.2 in the set of branches on which we run xplat builds.

dilijev:
Also add `xplat` branch to the list of branches on which these builds will run.

### DIFF
--- a/netci.groovy
+++ b/netci.groovy
@@ -257,18 +257,35 @@ if (!branch.endsWith('-ci')) {
 CreateStyleCheckTasks('./jenkins/check_eol.sh', 'ubuntu_check_eol', 'EOL Check')
 CreateStyleCheckTasks('./jenkins/check_copyright.sh', 'ubuntu_check_copyright', 'Copyright Check')
 
+// --------------
+// XPLAT BRANCHES
+// --------------
+
+// Explicitly enumerate xplat-incompatible branches, because we don't anticipate any future incompatible branches
+def isXPlatCompatibleBranch = !(branch in ['release/1.1', 'release/1.1-ci', 'release/1.2', 'release/1.2-ci'])
+
+// Include these explicitly-named branches
+def isXPlatDailyBranch = branch in ['master', 'linux', 'xplat']
+// Include some release/* branches
+if (branch.startsWith('release')) {
+    // Allows all current and future release/* branches on which we should run daily builds of XPlat configs.
+    // RegEx matches e.g. release/1.1, release/1.2, release/1.3-ci, but not e.g. release/2.0, release/1.3, or release/1.10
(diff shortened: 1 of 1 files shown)


dilijev:
FYI @digitalinfinity @mmitche

dilijev:
@dotnet-bot test ci please

netci.groovy
```diff
@@ -257,18 +257,35 @@ if (!branch.endsWith('-ci')) {
...
+    releaseExclusion = branch =~ /^release\/((1\.[12](\D.*)?)|(\d+\.\d+\D.*))$/
+    isXPlatDailyBranch |= !releaseExclusion
+}
+
 // -----------------
 // LINUX BUILD TASKS
 // -----------------
 
-if (branch.startsWith('linux') || branch.startsWith('master')) {
+if (isXPlatCompatibleBranch) {
```
obastemur: Do we have (or planning to have) not compatible branch? IMO we shouldn't

dilijev: We do have incompatible branches: `release/1.1` and `release/1.2` -- builds from those branches can't build XPlat or run tests. The fancy RegEx I'm using here is primarily so that the theoretical 1.10 which comes after 1.1 won't be considered as "incompatible". We don't yet know when/whether we will version bump to 2.0, so we could in theory reach v1.10.

dilijev:
Will wait to merge this until https://github.com/dotnet/dotnet-ci/pull/419 is merged in so that we can run the CI test and so we will get builds once it is merged in.

netci.groovy
obastemur: So why don't we just keep those 2 out here instead of implementing `isXPlatCompatibleBranch`?  Logically, we shouldn't have any further `not compatible` branch hence no additional work here (in case we open another branch in the future)

dilijev: I was thinking about suffixes to those branches so matching seemed like a reasonable approach, but since we only configure `*` and `*-ci` for Jenkins CI builds anyway, I think it's fine to enumerate the incompatible branches explicitly. 

I'll make the change.

dilijev:
@dotnet-bot test this please

netci.groovy
dilijev: I made the change but I'm leaving it extracted to a variable because we use the result twice.

dilijev:
@dotnet-bot 
test OSX osx_osx_debug_static
test OSX osx_osx_release_static
test OSX osx_osx_test_static
test Ubuntu ubuntu_linux_debug
test Ubuntu ubuntu_linux_debug_static
test Ubuntu ubuntu_linux_release
test Ubuntu ubuntu_linux_release_static
test Ubuntu ubuntu_linux_test
test Ubuntu ubuntu_linux_test_static

dilijev:
@dotnet-bot test ci please

PR #1480, opened 2016-08-23
Changes to enable SemVer in binary and NuGet version numbers.

dilijev:
* Provide the ability to specify all four fields of the NuGet package version (by defining environment variables in the build environment with defaults specified in init_build.ps1).
* No longer pad binary version number segments.
* No longer use build number in the product version number (include that information in the file version string, along with branch, commit, and build date).


dilijev:
@curtisman Need to merge this so that we can build a NuGet package from specifically the `release/1.2` branch with the proper version number in the `.nuspec` as well as in the binaries themselves.

PR #1483, opened 2016-08-24
Adding more async function test cases

aneeshdk:
Missed updating the stack walker to treat async functions also similar to
generator functions. Added more test cases. Removed the catch class from
the functionality tests as they are not relevant for async functions.


aneeshdk:
@ianwjhalliday Can you please take a look?

akroshg:
curious, why are we dumping the validation to the baseline file?

aneeshdk:
The way our test framework executes does not account for any pending promise activity. Due to the async nature a sequential check for test success does not work for us.

ianwjhalliday:
👍

PR #1488, opened 2016-08-25
Fix for cleanup of empty functions in presence of -bailoutoneveryline and related flags

meg-gupta:
Flag -bailoutateveryline injects bailout code at every line. On a bailout, we assert if there is an argument object used in the function body, it should be either zero initialized or valid. We eliminate zero initialization of arg object when stack args is enabled because it is initialized again. But in the presence of bailout injection flag, we bailout much before the argument object is initialized. So in the presence of this flag, we should not eliminate zero initialization.

### DIFF
--- a/lib/Backend/amd64/LowererMDArch.cpp
+++ b/lib/Backend/amd64/LowererMDArch.cpp
@@ -1486,7 +1486,14 @@ LowererMDArch::LowerEntryInstr(IR::EntryInstr * entryInstr)
     if (this->m_func->HasArgumentSlot() && (this->m_func->IsStackArgsEnabled() ||
         this->m_func->IsJitInDebugMode() ||
         // disabling apply inlining leads to explicit load from the zero-inited slot
-        this->m_func->GetJnFunction()->IsInlineApplyDisabled()))
+        this->m_func->GetJnFunction()->IsInlineApplyDisabled())
+#ifdef BAILOUT_INJECTION
+        || Js::Configuration::Global.flags.IsEnabled(Js::BailOutFlag)
+        || Js::Configuration::Global.flags.IsEnabled(Js::BailOutAtEveryLineFlag)
+        || Js::Configuration::Global.flags.IsEnabled(Js::BailOutAtEveryByteCodeFlag)
+        || Js::Configuration::Global.flags.IsEnabled(Js::BailOutByteCodeFlag)
+#endif
+        )
     {
         // TODO: Support mov [rbp - n], IMM64
         raxOpnd = IR::RegOpnd::New(nullptr, RegRAX, TyUint32, this->m_func);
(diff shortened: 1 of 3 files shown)


meg-gupta:
@satheeshravi Can you please review ?

meg-gupta:
@rajatd Modified to eliminate zero-initialization in the presence of bailout injection flags only, and removed flag setters. Can you please review ?

lib/Backend/amd64/LowererMDArch.cpp
```diff
@@ -1486,7 +1486,14 @@ LowererMDArch::LowerEntryInstr(IR::EntryInstr * entryInstr)
     if (this->m_func->HasArgumentSlot() && (this->m_func->IsStackArgsEnabled() ||
         this->m_func->IsJitInDebugMode() ||
         // disabling apply inlining leads to explicit load from the zero-inited slot
-        this->m_func->GetJnFunction()->IsInlineApplyDisabled()))
+        this->m_func->GetJnFunction()->IsInlineApplyDisabled())
+#ifdef BAILOUT_INJECTION
+        || Js::Configuration::Global.flags.IsEnabled(Js::BailOutFlag)
+        || Js::Configuration::Global.flags.IsEnabled(Js::BailOutAtEveryLineFlag)
+        || Js::Configuration::Global.flags.IsEnabled(Js::BailOutAtEveryByteCodeFlag)
+        || Js::Configuration::Global.flags.IsEnabled(Js::BailOutByteCodeFlag)
```
satheeshravi: can this be just hasBailOut inside #ifdef BAILOUT_INJECTION ? 
We actually care about release build to have good code gen. So i think it is fine to use the bool directly.

satheeshravi:
nit: Also, now that you are fixing something related to empty-func, do you mind moving SetHasThrow(); to backwardPass, as suggested by Louis in your earlier PR ?

lib/Backend/amd64/LowererMDArch.cpp
rajatd: Sure, that works too.

---
In reply to: [76706901](https://github.com/Microsoft/ChakraCore/pull/1488#discussion_r76706901) [](ancestors = 76706901)

meg-gupta: Looks like BAILOUT_INJECTION  is enabled only in DEBUG build. Since most Func's have hasBailout flag set, we will differ a lot in eliminating zero-init between debug and test builds if we have hasBailout check instead of explicit flag check ?

satheeshravi: That's fine, as far as release release build doesn't generate unnecessary zero-init.

satheeshravi:
:shipit:

meg-gupta:
@dotnet-bot retest ubuntu_linux_test_static please

meg-gupta:

PR #1491, opened 2016-08-25
QuickSort

Cellule:
Remove our own implementation of QuickSort in favor of the stdlib version.

This implementation was used by SwitchBuilder which could run out of stack if we had a very long list of cases already sorted.
It is hard to hit in the release product has compiler optimization changes the number of cases needed to hit the issue and we don't jit functions past a certain bytecode size.
I was able to repro the issue with a switch of 16000 cases in debug.
This problem is more problematic for WebAssembly as we will jit very huge functions no matter what and cases are always sorted

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/1491)


Cellule:
@MikeHolman Can you review please ?

lib/Common/DataStructures/ImmutableList.h
```diff
@@ -729,7 +731,9 @@ namespace regex
                 return this;
             }
             auto arr = ToReferenceArrayInHeap(size);
-            QuickSort<T *, true>::Sort(arr, arr+(size-1), comparer);
+            qsort_s(arr, size, sizeof(T*), [](void* comparer, const void* a, const void* b) {
```
Cellule: I fail to see what makes this sort in reverse order, but this refactor shouldn't  change the current behavior.

dilijev:
LGTM

lib/Common/DataStructures/quicksort.h
```diff
@@ -1,103 +0,0 @@
...
-// Licensed under the MIT license. See LICENSE.txt file in the project root for full license information.
-//-------------------------------------------------------------------------------------------------------
-#pragma once
-
-namespace regex
-{
-    template <class T, bool partitionFromMiddle> class QuickSort
-    {
-    public:
-        static void Sort(T* low, T* high, Comparer<T> * comp)
```
ThomsonTan: I am wondering why no stack overflow for qsort_s.

Cellule: I don't know how qsort_s is implemented, but it is much smarter and faster than that basic quicksort implementation. I just did a quick test and even on release build this QuickSort crashes after ~22000 elements and it takes a few seconds to reach that point where as qsort_s terminates in no time at all

Issue #1492, opened 2016-08-25
[OS X] ChakraCore is built against different SDK than Rust which prevents linking ChakraCore into Rust projects


dilijev:
See https://github.com/Microsoft/ChakraCore/issues/1441#issuecomment-242277297

FYI @fishrock123

obastemur:
#1493 should take care of this.

dilijev:
@Fishrock123 Can you confirm whether this issue was fixed for you by #1493 ?

obastemur:
Closing this one for now. From the discussion at #1493 there should be other things to consider (no longer SDK difference) 

@Fishrock123 once you can share the steps, please open up an issue with the details so we may look into. Thanks!

obastemur:
Prepared the bash script below for anyone looking into calling ChakraCore from Rust

```
CHAKRA_BUILD_FOLDER=$1

if [ ! -d $CHAKRA_BUILD_FOLDER ]; then
  echo "usage: compile.sh <path to ChakraCore BuildLinux folder>"
  exit 0
fi

# UPDATE THIS to Rustlib path on your system
RUSTLIBS="/usr/local/Cellar/rust/1.13.0/lib/rustlib/x86_64-apple-darwin/lib"

cc -m64 -L $RUSTLIBS helloWorld.0.o -o helloWorld -Wl,-dead_strip \
-nodefaultlibs -L $RUSTLIBS \
 $RUSTLIBS/libstd-a4729905.rlib \
 $RUSTLIBS/libpanic_unwind-a4729905.rlib \
 $RUSTLIBS/libunwind-a4729905.rlib \
 $RUSTLIBS/librand-a4729905.rlib \
 $RUSTLIBS/libcollections-a4729905.rlib \
 $RUSTLIBS/librustc_unicode-a4729905.rlib \
 $RUSTLIBS/liballoc-a4729905.rlib \
 $RUSTLIBS/liballoc_jemalloc-a4729905.rlib \
 $RUSTLIBS/liblibc-a4729905.rlib \
 $RUSTLIBS/libcore-a4729905.rlib \
 $RUSTLIBS/libcompiler_builtins-a4729905.rlib \
 -Wl,-force_load \
 $CHAKRA_BUILD_FOLDER/pal/src/libChakra.Pal.a \
 $CHAKRA_BUILD_FOLDER/lib/Common/Core/libChakra.Common.Core.a \
[...]

PR #1493, opened 2016-08-25
macos: target min osx sdk is 10.7

obastemur:
Some of the embedders (i.e. node-chakracore targets an earlier version of SDK)


Fishrock123:
@obastemur I now get a slightly different error output, but it's still about the same length: https://gist.github.com/Fishrock123/c9dba8ff7d6c26e296d1fa7d6846b176

obastemur:
The rest of the errors looks like c++ exceptions related. Have you disabled c++ exceptions ? or do you have chance to use shared lib instead? (remove `--static` from build args)

Fishrock123:
> Have you disabled c++ exceptions ? 

Not sure if I can.. will follow up with Rust people

> or do you have chance to use shared lib instead?

Hmmm, so is that a dynamic lib then? Sorry I am still newer to some of this. Dx

Doesn't seem to help much though, even if I specify to link dynamically or statically without building with `--static`...

obastemur:
Okay, thanks for the details. 

This PR does what it supposed to do. For the rest; we can continue on #1492 
If you can share some details on how to repro things, I may take a look at closer.

pal/inc/safemath.h
```diff
@@ -156,717 +167,6 @@ inline bool DoubleFitsInIntType(double val)
...
-//      if( !ClrSafeInt<UINT32>::multiply( elementCount, sizeof(void*), tmp ) ||
-//              !ClrSafeInt<UINT32>::addition( tmp, headerSize, bufSize ) ) 
-//      { <overflow-error> }
-//      else { use bufSize }
-//      
-//-----------------------------------------------------------------------------
-// TODO: Any way to prevent unintended instantiations?  This is only designed to
-//  work with unsigned integral types (signed types will work but we probably 
-//  don't need signed support).
-template<typename T> class ClrSafeInt
```
dilijev: Was it necessary to remove ClrSafeInt because of SDK versions or was this a drive-by cleanup?

dilijev:
LGTM

pal/inc/safemath.h
obastemur: both cleanup and some newer SDK stuff used

obastemur:

dilijev:
Side note, the wiki currently says we target 10.9+ (https://github.com/Microsoft/ChakraCore/wiki/Building-ChakraCore#os-x). Should we revise to 10.7+?

obastemur:
@dilijev no we shouldn't. I'm not sure the compiler support etc. we need are available on 10.7.. SDK and actual OS are kinda separate things here.

dilijev:
@obastemur Got it, thanks.

PR #1495, opened 2016-08-26
Fix ByteCodeSerialize trace for encode values.

curtisman:
Need to include the count for the encoded sentinel


tcare:
Looks good

Issue #1478, opened 2016-08-23
ENABLE_BACKGROUND_PARSING should probably be undefined


EdMaurer:
It appears that all of the code to support parallel parsing is compiled in. Is it supposed to be? My guess is no.

Cellule:
It looks like we do
```
#if ENABLE_BACKGROUND_PARSING
```
and it is defined to either `0` or `1`.
I think the code is correctly removed for nojit build.

curtisman:
The parallel parsing code wasn't under #if before.    It was added for the xplat effort so that we don't have to port it to Linux for now.  I don't think the #if will disable all of the code associated with background parsing.

dilijev:
Looks like this question is answered, so I'm closing it. I think any work items related to this are already tracked in the xplat effort.

PR #1497, opened 2016-08-26
Fix incorrect parsing of yield/await in arrow functions

### LINKED ISSUE #333: yield should be reserved word in arrow functions within StrictFormalParameters
``` javascript
function* gf() {
    var a = (yield) => { }; // spec says error, ChakraCore gives no error
    var b = (x = yield) => { }; // spec says error, ChakraCore produces no error
}
```

This is an error by way of ArrowFunction passing the Yield parameter on to its StrictFormalParameters.  This is different from FunctionExpression which drops the Yield parameter entirely from its production.

https://tc39.github.io/ecma262/#sec-arrow-function-definitions
https://tc39.github.io/ecma262/#prod-FunctionExpression

ianwjhalliday:
The spec carries the Yield/Await parameter forward into the parameter
list grammar production for arrow functions specifically. We did not
adhere to this and erroneously treated these as identifiers in arrow
function parameter lists.

Fixes #333.  See it for more details.

### DIFF
--- a/lib/Parser/Parse.cpp
+++ b/lib/Parser/Parse.cpp
@@ -4739,6 +4739,8 @@ Parse a function definition.
 template<bool buildAST>
 bool Parser::ParseFncDeclHelper(ParseNodePtr pnodeFnc, LPCOLESTR pNameHint, ushort flags, bool *pHasName, bool fUnaryOrParen, bool noStmtContext, bool *pNeedScanRCurly, bool skipFormals)
 {
+    ParseNodePtr pnodeFncParent = GetCurrentFunctionNode();
+    // is the following correct? When buildAST is false, m_currentNodeDeferredFunc can be nullptr on transition to deferred parse from non-deferred
     ParseNodePtr pnodeFncSave = buildAST ? m_currentNodeFunc : m_currentNodeDeferredFunc;
     ParseNodePtr pnodeFncSaveNonLambda = buildAST ? m_currentNodeNonLambdaFunc : m_currentNodeNonLambdaDeferredFunc;
     int32* pAstSizeSave = m_pCurrentAstSize;
@@ -4968,7 +4970,7 @@ bool Parser::ParseFncDeclHelper(ParseNodePtr pnodeFnc, LPCOLESTR pNameHint, usho
 
         if (!skipFormals)
         {
-            this->ParseFncFormals<buildAST>(pnodeFnc, flags);
+            this->ParseFncFormals<buildAST>(pnodeFnc, pnodeFncParent, flags);
         }
 
         // Create function body scope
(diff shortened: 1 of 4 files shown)


lib/Parser/Parse.cpp
```diff
@@ -11130,6 +11156,9 @@ HRESULT Parser::ParseFunctionInBackground(ParseNodePtr pnodeFnc, ParseContext *p
     pnodeFnc->sxFnc.pnodeBody = nullptr;
     pnodeFnc->sxFnc.nestedCount = 0;
 
+    // getting the parent node this way might be incorrect
+    // todo check with Paul how to approach this
+    ParseNodePtr pnodeParentFnc = m_currentNodeFunc;
```
ianwjhalliday: @pleath is there a way to get the parent function node in parallel parse?  Is parallel parse going to be maintained into the future? #1478 suggests it is not used and could be excluded from build.

ianwjhalliday:
@pleath @Microsoft/chakra-es

ianwjhalliday:
@bterlson I made the same changes for `await`.  Is `await` intended to be treated the same way as `yield` in this regard?

bterlson:
@ianwjhalliday if you mean how `await` should not be allowed as a parameter name for an arrow in an async function, yes. It is parsed as an await expression but disallowed via static semantics IIRC.

test/es7/asyncawait-syntax.js
```diff
@@ -189,6 +189,19 @@ var tests = [
...
+            assert.throws(function () { eval("async function af() { var a = (x, await, y) => { }; }"); }, SyntaxError, "await cannot appear as a formal name within parenthesized arrow function parameter list (last formal)", "Syntax error");
+
+            assert.throws(function () { eval("async function af() { var a = (x = await 0) => { }; }"); }, SyntaxError, "await expression is disallowed within arrow function default parameter expression (single formal)", "'await' expression not allowed in this context");
+            assert.throws(function () { eval("async function af() { var a = (x, y = await 0, z = 0) => { }; }"); }, SyntaxError, "await expression is disallowed within arrow function default parameter expression (middle formal)", "'await' expression not allowed in this context");
+            assert.throws(function () { eval("async function af() { var a = (x, y, z = await 0) => { }; }"); }, SyntaxError, "await expression is disallowed within arrow function default parameter expression (last formal)", "'await' expression not allowed in this context");
```
dilijev: Is the `eval` essential for these tests?

pleath: The bad syntax has to be parsed dynamically, yes. If it's not contained in something like eval or "new Function", then the test will error out before executing anything.

lib/Parser/Parse.cpp
```diff
@@ -6033,13 +6035,22 @@ void Parser::UpdateOrCheckForDuplicateInFormals(IdentPtr pid, SList<IdentPtr> *f
 }
 
 template<bool buildAST>
-void Parser::ParseFncFormals(ParseNodePtr pnodeFnc, ushort flags)
+void Parser::ParseFncFormals(ParseNodePtr pnodeFnc, ParseNodePtr pnodeParentFnc, ushort flags)
```
pleath: Does this code work if we're parsing an arrow function contained in some number of other arrow functions? Does looking at the immediate parent tell us what we need to know?

aneeshdk:
This change is checked in by the pull request https://github.com/Microsoft/ChakraCore/pull/1785

PR #1498, opened 2016-08-26
Add a case for NetBSD in LOADLoadLibrary()

krytarowski:
The proper solution for NetBSD is to reuse libc.so as shortAsciiName.

The same solution should be promoted for FreeBSD.


krytarowski:
ping! the same code is in CoreCLR (I authored both)

obastemur:
@krytarowski Thanks for the contribution. Could you please squash the commits ?

krytarowski:
Rebased to master

PR #1500, opened 2016-08-26
Refactor Enumerators

curtisman:
Most object enumerate the property the same way, array indexes and then named properties.  They can share common data and instance to do those work. This change refactors how Chakra do Enumerator. Instead of create a enumerator for every enumeration,  the caller will supply a JavascriptStaticEnumerator to initialize. JavascritpStaticEnumerator first try to enumerate thru a prefix enuemrator, which the object can prepend any properties to the enumeration sequence. Then it will enumerate the array indexes.  Within JavascriptStaticEnumerator embeds the DynamicObjectPropertyEnumerator that will enumerate the name properties after the array index is done.

DynamicObjectEnumerator and its derived class is collapsed into DynamicObjectPropertyEnumerator.  And most "custom" enumerators are refactor to only do prefix or array index enumerations and leave the property enumeration to JavascriptStaticEnumerator.

Clarify the overload GetEnumerator that has the originalInstance as the first parameter by renamed to ArrayObject::GetIndexEnumerator, which only enumerate the index of an object.

Combines the 3 boolean parameter for enumerator into flags to specify whether it wants non-enumerable properties, symbol properties and snapshot semantics.

The fast path for GenerateFastBrBReturn that generate native code for for..in fastpath is converted into to platform agnostic lower code so they can share between x86/x64 and Arm.


curtisman:
@Yongqu Review please.

lib/Runtime/Types/JavascriptStaticEnumerator.cpp
```diff
@@ -0,0 +1,90 @@
...
+#include "Library/JavascriptSymbol.h"
+
+namespace Js
+{
+
+    bool JavascriptStaticEnumerator::Initialize(JavascriptEnumerator * prefixEnumerator, ArrayObject * arrayToEnumerate, 
+        DynamicObject * objectToEnumerate, EnumeratorFlags flags, ScriptContext * requestContext)
+    {
+        this->prefixEnumerator = prefixEnumerator;
+        this->arrayEnumerator = arrayToEnumerate ? arrayToEnumerate->GetIndexEnumerator(flags, requestContext) : nullptr;
```
Yongqu: nitpick: trailing white space :D

lib/Runtime/Types/JavascriptEnumerator.h
```diff
@@ -40,10 +39,6 @@ namespace Js {
         // seeing this comment please just consolidate the changes.
         virtual Var MoveAndGetNext(PropertyId& propertyId, PropertyAttributes* attributes = nullptr) = 0;
```
Yongqu: Can you update comment, especially on the relationship between return value & returned propertyId?

lib/Runtime/Types/JavascriptStaticEnumerator.cpp
```diff
@@ -0,0 +1,90 @@
...
+            }
+            this->currentEnumerator = (this->currentEnumerator == this->prefixEnumerator) ? this->arrayEnumerator : nullptr;
+        }
+
+        return nullptr;
+    }
+
+    Var JavascriptStaticEnumerator::MoveAndGetNext(PropertyId& propertyId, PropertyAttributes* attributes)
+    {
+        Var currentIndex = MoveAndGetNextFromEnumerator(propertyId, attributes);
```
Yongqu: curious why you want to separate out the nested enumerator cases into a different function?

curtisman: Just because it is cleaner :)

lib/Runtime/Types/JavascriptEnumerator.h
curtisman: I still want to change how that work with symbol with my next change.  I will defer the documentation.

lib/Runtime/Library/ForInObjectEnumerator.cpp
```diff
@@ -108,30 +100,18 @@ namespace Js
...
     }
 
-    BOOL ForInObjectEnumerator::GetCurrentEnumerator()
+    BOOL ForInObjectEnumerator::InitializeCurrentEnumerator()
     {
         Assert(object);
         ScriptContext* scriptContext = GetScriptContext();
-
+        EnumeratorFlags flags = EnumeratorFlags::EnumNonEnumerable | EnumeratorFlags::SnapShotSemantics | (enumSymbols ? EnumeratorFlags::EnumSymbols : EnumeratorFlags::None);
         if (VirtualTableInfo<DynamicObject>::HasVirtualTable(object))
```
Yongqu: Do we still need to have this code path for DynamicObject?

lib/Runtime/Types/JavascriptStaticEnumerator.cpp
```diff
@@ -0,0 +1,90 @@
...
+// Licensed under the MIT license. See LICENSE.txt file in the project root for full license information.
+//-------------------------------------------------------------------------------------------------------
+#include "RuntimeTypePch.h"
+#include "Library/JavascriptSymbol.h"
+
+namespace Js
+{
+
+    bool JavascriptStaticEnumerator::Initialize(JavascriptEnumerator * prefixEnumerator, ArrayObject * arrayToEnumerate, 
+        DynamicObject * objectToEnumerate, EnumeratorFlags flags, ScriptContext * requestContext)
```
Yongqu: Can we in debug build, save the requestContext, and assert the prefixEnumerator returned propertyString doesn't need marshalling?

lib/Runtime/Types/DynamicType.cpp
```diff
@@ -320,95 +321,23 @@ namespace Js
...
-        }
-        else
+        Js::ArrayObject * arrayObject = nullptr;
+        if (this->HasObjectArray())
         {
-            *enumerator = DynamicObjectEnumerator</*enumNonEnumerable*/false, /*enumSymbols*/false, /*snapShotSemantics*/false>::New(requestContext, this);
+            arrayObject = this->GetObjectArrayOrFlagsAsArray();
+            Assert(arrayObject->GetPropertyCount() == 0);
         }
+        return enumerator->Initialize(prefixEnumerator, arrayObject, this, flags, requestContext);
```
Yongqu: this is cool :)

lib/Runtime/Library/JavascriptReflect.cpp
```diff
@@ -4,9 +4,6 @@
 //-------------------------------------------------------------------------------------------------------
 #include "RuntimeLibraryPch.h"
 
-#include "Types/DynamicObjectEnumerator.h"
-#include "Types/DynamicObjectSnapshotEnumerator.h"
-#include "Types/DynamicObjectSnapshotEnumeratorWPCache.h"
 #include "Library/ForInObjectEnumerator.h"
```
Yongqu: nitpick: this is not needed either

lib/Runtime/Types/DynamicObjectPropertyEnumerator.cpp
```diff
@@ -0,0 +1,216 @@
...
+            objectIndex = Constants::NoBigSlot;
+            initialPropertyCount = GetSnapShotSemantics() ? this->object->GetPropertyCount() : Constants::NoBigSlot;
+            // Create the appropriate enumerator object.
+            if (GetSnapShotSemantics() && this->initialType->PrepareForTypeSnapshotEnumeration())
+            {
+                ScriptContext* scriptContext = this->object->GetScriptContext();
+                ThreadContext * threadContext = scriptContext->GetThreadContext();
+                CachedData * data = (CachedData *)threadContext->GetDynamicObjectEnumeratorCache(this->initialType);
+
+                if (data == nullptr || data->scriptContext != this->requestContext || data->enumNonEnumerable != GetEnumNonEnumerable() || data->enumSymbols != GetEnumSymbols())
```
Yongqu: this is cool

Yongqu:
:shipit:

lib/Runtime/Library/ForInObjectEnumerator.cpp
curtisman: For now, yes, to avoid the virtual call.  I am going to remove it once I do a JIT fast path for ForInObjectEnumerator

PR #1501, opened 2016-08-26
Fix incorrect error when calling non-existent symbol property

### LINKED ISSUE #1409: Incorrect error message for missing symbol methods
For example:

``` js
var obj = Object.create(null);
obj[Symbol.iterator]();
```

Actual: TypeError: Object doesn't support property or method 'ToString'
Expected: Object doesn't support property or method 'Symbol.iterator'

Or we could go with an error message similar to v8 and sm:

obj[Symbol.iterator] is not a function

ianwjhalliday:
We threw 'Object does not support ToString' when trying to call a symbol
property on an object when that property didn't exist.  This happened
because we try to do a toString conversion on the index var for the
error message string, but Symbol objects do not support toString via
JavascriptConversion::ToString().

Fix it by special casing symbol properties and directly calling the
static JavascriptSymbol::ToString method to get a useful string.

Fixes #1409

### DIFF
--- a/lib/Runtime/Language/JavascriptOperators.cpp
+++ b/lib/Runtime/Language/JavascriptOperators.cpp
@@ -4024,7 +4024,15 @@ CommonNumber:
 
         if (!hasProperty)
         {
-            JavascriptString* varName = JavascriptConversion::ToString(index, scriptContext);
+            JavascriptString* varName = nullptr;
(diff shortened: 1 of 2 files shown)


ianwjhalliday:
@tcare @boingoing @Microsoft/chakra-es cc @bterlson

tcare:
Looks good

tcare:
Checked in as 61e4635

PR #1502, opened 2016-08-26
Fixing stack balance under generator/yield

akroshg:
We were caching and restoring the sp in the processtryfinally in order to balance the stack when we are in the middle of call. However in the case of generator we can yield out of the call. In that case we actually haven't finished the try part.
So once we comeback after yield we don't need to cache the stack again. Introduced the flag to capture such state and fix the problem.

### DIFF
--- a/lib/Runtime/Language/InterpreterStackFrame.cpp
+++ b/lib/Runtime/Language/InterpreterStackFrame.cpp
@@ -1084,6 +1084,7 @@ namespace Js
         newInstance->m_flags        = InterpreterStackFrameFlags_None;
         newInstance->closureInitDone = false;
         newInstance->isParamScopeDone = false;
+        newInstance->shouldCacheSP = true;
 #if ENABLE_PROFILE_INFO
         newInstance->switchProfileMode = false;
         newInstance->isAutoProfiling = false;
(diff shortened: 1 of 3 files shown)


akroshg:
@aneeshdk @tcare @ThomsonTan please have a look.

tcare:
Looks good

lib/Runtime/Language/InterpreterStackFrame.cpp
```diff
@@ -6744,7 +6745,10 @@ const byte * InterpreterStackFrame::OP_ProfiledLoopBodyStart(const byte * ip)
             // mark the stackFrame as 'in try block'
             this->m_flags |= InterpreterStackFrameFlags_WithinTryBlock;
 
-            CacheSp();
+            if (shouldCacheSP)
```
ThomsonTan: >shouldCacheSP [](start = 16, length = 13)

Do we also need flip shouldCacheSP right after this if in case we re-enter ProcessTryFinally?

akroshg: We shouldn't do that, look at the second example in the test file we have two `try`s so we should not cache-sp in both cases. however say if you have two destructuring pattern in the call-expression we would have to cache sp on both occasion.

aneeshdk:
Can you also add one test case for async functions too?

lib/Runtime/Language/InterpreterStackFrame.cpp
ThomsonTan: I see. Yes, we need ignore cache SP for both ProcessTryFinally for your second test, but how about if there are some more ProcessTryFinally following  but which are not for the previous yield? Seems this fix will also ignore caching SP for them.

---
In reply to: [76500192](https://github.com/Microsoft/ChakraCore/pull/1502#discussion_r76500192) [](ancestors = 76500192)

PR #1503, opened 2016-08-26
Bug fix in memory profiler

kunalspathak:
Noticed that we were never printing summary of individual arenas because
of oversight in index variable used to track summaries.


kunalspathak:
@digitalinfinity , can you please review?

digitalinfinity:
:shipit:

satheeshravi:
:shipit:

PR #1505, opened 2016-08-27
Library bytecode verification unittests

tcare:
Reintroduces bytecode verification unittests for library code. We previously had this check in SD days, but it was complicated and called external scripts to work. The new version uses just RL to compare against the header file directly. A number of changes had to be made:

- Removed non-standard %REGRESS% psuedo environment variable. Instead, test files and baselines are relative paths based on the directory containing the rlexe.xml.
- Fixed the relative path behavior of runtests.py. Also added a guard to the timezone set so runtests.py works on Windows again.
- Fixed an issue where ch would generate different bytecode than jshost due to an implicit fscrNoAsmJs parser flag in ScriptContext. We don't need this flag on the initial entry, so it can be removed from the JSRT entrypoint for library bytecode generation.
- Disabled using the release version for serialized library bytecode, instead using a fixed version (all 0's). This is OK as long as we check the bytecode integrity in unit tests.
- Added an implicit disable_jit tag exclusion when not running disable JIT; this allows us to run some test only under disable JIT and not normally.
- Added a new bytecode directory with the xml for the tests.

Due to the changes to serialization all of the library bytecode headers needed to be changed.


tcare:
@curtisman @MikeHolman @ianwjhalliday please review

FYI @digitalinfinity @dilijev

lib/Runtime/ByteCode/ByteCodeSerializer.cpp
```diff
@@ -437,14 +437,24 @@ class ByteCodeBufferBuilder
...
             V4.value = guidDWORDs[3];
         }
+        else
+        {
+            // To keep consistent library code between Chakra.dll and ChakraCore.dll, use a fixed version.
+            // This goes hand in hand with the bytecode verification unit tests.
+            V1.value = 0;
+            V2.value = 0;
+            V3.value = 0;
+            V4.value = 0;
```
dilijev: For releases, we need to make sure we have the GUIDs in the ByteCode so that ReleaseVersioningScheme can be applied. In that scenario, can we also ensure consistent bytecode between Chakra and ChakraCore.dll

dilijev: Wait, I understand now. For library code, we don't want to do the GUID check? We still need to make sure the ByteCode is updated when it is supposed to be. How else can we specifically test this?

---
In reply to: [76506572](https://github.com/Microsoft/ChakraCore/pull/1505#discussion_r76506572) [](ancestors = 76506572)

dilijev:

Remove this file?

---
Refers to: test/out:1 in bdf71d1. [](commit_id = bdf71d14cf6f422efd64c7adc555d484030c78cc, deletion_comment = False)

tcare:
Oops thanks for the catch.

lib/Runtime/ByteCode/ByteCodeSerializer.cpp
tcare: For the library code it doesn't really matter, as long as the unit tests catch the incorrectness. We usually catch this in unit tests when we fail unit tests, but we don't always catch them right now. E.g. the jshost/ch inconsistency fixed in this change. The unit tests are guaranteed to catch the bytecode changes.

test/ByteCode/rlexe.xml
```diff
@@ -0,0 +1,35 @@
+<?xml version="1.0" encoding="utf-8"?>
+<regress-exe>
+  <test>
+    <default>
+      <tags>exclude_dynapogo,exclude_amd64,exclude_arm,require_backend</tags>
+      <compile-flags>-GenerateLibraryByteCodeHeader</compile-flags>
```
dilijev: ><compile-flags>-GenerateLibraryByteCodeHeader</compile-flags> [](start = 6, length = 61)

So the plan is to use a test to generate the entire bytecode and check for differences, and then alert the user to the fact that the bytecode needs to be updated?

Since we've already got the binaries handy, this is a good approach.

I think right now the way we regen bytecode depends on scripts. Since we're talking about eventually replacing rl.exe anyway, I don't want to suggest that we put that functionality into rl.exe, but we should define a unified way to generate bytecode.

tcare:
Looks like Jenkins is having trouble with release and crossplat. I'll investigate next week.

dilijev:
@tcare 
PreFAST warnings from release builds on Windows.
testrunner.py doesn't support `%VAR%` syntax so it can't parse the test definition xml file.

test/ByteCode/rlexe.xml
tcare: We used to use scripts in SD but never ported it over to git. There essentially doesn't need to be a script; it was doing what RL does. The only real change is that instead of writing to a file I write to stdout.

test/ByteCode/rlexe.xml
```diff
@@ -0,0 +1,35 @@
...
+      <files>%REGRESS%\..\lib\Runtime\Library\InJavascript\Intl.js</files>
```
dilijev: >%REGRESS% [](start = 13, length = 9)

I think paths are relative to the test location so you can just use ..\..\lib\

tcare: This doesn't work for using jshost to test core.

dilijev: Good point. Simplest solution is to make testrunner.py support `%VAR%`

tcare:
Closing for now until cross plat is resolved.

dilijev:
@tcare Why not keep it open to track it (do we think this will take a long time to resolve?)

tcare:
It'll be considerably different. I want to move away from %REGRESS%. Ideally I would move away from RL too ;)

tcare:
Reopened.

%REGRESS% is gone. We now use the test folder as the working path.

@jianchun can you please review the runtests.py changes?

tcare:
Looks like runtests.py needs more work. Unfortunately we don't have a good buddy build mechanism for Jenkins at the moment. I'll ping back when I update.

tcare:
Moving out serialization work to #1520

PR #1507, opened 2016-08-29
cross-platform: va_end may not be no-op

obastemur:
Standard requires a call to va_end due to its undefined behavior.


digitalinfinity:
LGTM

obastemur:

PR #1509, opened 2016-08-29
Use the error HR for the exception code for debug attach/detech fatal error

curtisman:


curtisman:
@Yongqu @agarwal-sandeep   Review please.

lib/Jsrt/JsrtDiag.cpp
```diff
@@ -129,9 +130,10 @@ CHAKRA_API JsDiagStopDebugging(
         {
             Assert(scriptContext->IsScriptContextInDebugMode());
 
-            if (FAILED(scriptContext->OnDebuggerDetached()))
+            HRESULT hr;
+            if (FAILED(hr = scriptContext->OnDebuggerDetached()))
             {
-                Debugger_AttachDetach_fatal_error(); // Inconsistent state, we can't continue from here
+                Debugger_AttachDetach_fatal_error(hr); // Inconsistent state, we can't continue from here
                 return JsErrorFatal;
```
agarwal-sandeep: >return JsErrorFatal; [](start = 16, length = 20)

We won't reach here ever, can be simplified as Debugger_AttachDetach_fatal_error(scriptContext->OnDebuggerDetached())?

agarwal-sandeep:
Otherwise LGTM

lib/Jsrt/JsrtDiag.cpp
curtisman: Well, Debugger_AttachDetach_fatal_error doesn't check hr, and it can be in one line.
I'd rather save the sorry by returning....

PR #1512, opened 2016-08-29
Fix bug in Cm* fastpath when dest is same as src (reg1=reg1==reg2)

agarwal-sandeep:
I was first moving true to dest and then comparing src1 with src2, that resulted in wrong result when dest and src1/src2 are same.


agarwal-sandeep:
@satheeshravi please take a look

rajatd:
:shipit:

PR #1513, opened 2016-08-30
Fix incorrect identifier recording of lambda parameter

ThomsonTan:
Single paramter of lambda function is tried to be parsed as identifier at first and could be added to PidRefStack of an identifier defined in outer scope. This causes non local reference to outer identifer by accident and leads to incorrect bytecode. (StInnerSlot is generated without NewInnerScope which causes access violation when setting the inner scope).


ThomsonTan:
@akroshg could you take a look?

ThomsonTan:
Close this since this is an incomplete fix.

Issue #1127, opened 2016-06-11
Wasm todo


Cellule:
This is a list of thing we need to implement for Wasm MVP
- [x] Br yielding values #1111
- [x] I32Popcnt
- [x] F32Min/Max #659
- [x] Trunc JIT
- [x] Nearest JIT
- [x] Int64 (@Cellule #1806)
- [x] Unreachable operator
- [x] Unreachable state
- [ ] Missing OpCodes
  - [x] current_memory (0x3b)
  - [ ] grow_memory (0x39) (@MikeHolman)
  - [x] f64.copysign (0x91)
  - [ ] i32.trunc_u/f32 (0x9f) (@Krovatkin #1812)
  - [ ] i32.trunc_u/f64 (0xa0) (@Krovatkin #1812)
  - [ ] f32.convert_u/i32 (0xa9) (@Krovatkin)
  - [x] f32.reinterpret/i32 (0xad) (@Krovatkin #1813)
  - [x] get_global (0xbb)
  - [x] set_global (0xbc)
  - [ ] Int64 Conversions
    - [ ] i64.trunc_s/f32
    - [ ] i64.trunc_s/f64
    - [ ] i64.trunc_u/f32
    - [ ] i64.trunc_u/f64
    - [ ] i64.extend_s/i32 (@meg-gupta)
    - [ ] i64.extend_u/i32 (@meg-gupta)
    - [ ] f32.convert_s/i64
    - [ ] f32.convert_u/i64
    - [ ] f64.convert_s/i64
    - [ ] f64.convert_u/i64
    - [ ] f64.reinterpret/i64 (@arunetm)
    - [ ] i64.reinterpret/f64 (@arunetm)
- [x] [Globals](https://github.com/WebAssembly/design/blob/binary_0xc/BinaryEncoding.md#global-section)
- [ ] [Init Expressions](https://github.com/WebAssembly/design/blob/binary_0xc/BinaryEncoding.md#init_expr) 
  - [ ] For element segments' [offset](https://github.com/WebAssembly/design/blob/binary_0xc/BinaryEncoding.md#element-section) (@arunetm)
  - [x] For data segments' [offset](https://github.com/WebAssembly/design/blob/binary_0xc/BinaryEncoding.md#data-section)
  - [x] For globals' [initial value](https://github.com/WebAssembly/design/blob/binary_0xc/BinaryEncoding.md#global-entry)
[...]

PR #1514, opened 2016-08-30
Wasm current memory

Krovatkin:
This PR  consists of two commits.
* The first one fixes an issue where wasm seems to violate calling conventions for internal asmjs routines  by providing arguments in the **reverse order** (calls to internal asmjs seems to be the only case when it happens). 
* memory_size + test fixes

### DIFF
--- a/lib/Backend/IRBuilderAsmJs.cpp
+++ b/lib/Backend/IRBuilderAsmJs.cpp
@@ -1897,23 +1897,36 @@ IRBuilderAsmJs::BuildAsmReg1(Js::OpCodeAsmJs newOpcode, uint32 offset)
 void
 IRBuilderAsmJs::BuildAsmReg1(Js::OpCodeAsmJs newOpcode, uint32 offset, Js::RegSlot dstReg)
 {
-    Assert(newOpcode == Js::OpCodeAsmJs::LdUndef);
+    Assert(newOpcode == Js::OpCodeAsmJs::LdUndef || newOpcode == Js::OpCodeAsmJs::CurrentMemory_Int);
 
-    Js::RegSlot dstRegSlot = GetRegSlotFromVarReg(dstReg);
-
-    IR::RegOpnd * dstOpnd = BuildDstOpnd(dstRegSlot, TyVar);
-
-    if (dstOpnd->m_sym->m_isSingleDef)
+    if (newOpcode == Js::OpCodeAsmJs::LdUndef)
     {
-        dstOpnd->m_sym->m_isConst = true;
-        dstOpnd->m_sym->m_isNotInt = true;
+
+        Js::RegSlot dstRegSlot = GetRegSlotFromVarReg(dstReg);
+        IR::RegOpnd * dstOpnd = BuildDstOpnd(dstRegSlot, TyVar);
+
+        if (dstOpnd->m_sym->m_isSingleDef)
+        {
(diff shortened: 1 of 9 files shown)


Krovatkin:
@MikeHolman @Cellule please take a look at current_memory impl (#1127) when you have a mnt 
The motivation for the first commit is in the PR description.

Krovatkin:
FYI: @arunetm @commonlisp

lib/Runtime/Language/InterpreterStackFrame.cpp
```diff
@@ -7655,6 +7655,12 @@ const byte * InterpreterStackFrame::OP_ProfiledLoopBodyStart(const byte * ip)
         m_localSimdSlots[localRegisterID] = bValue;
     }
 
+    int InterpreterStackFrame::OP_GetMemorySize()
+    {
+        JavascriptArrayBuffer* arr = *(JavascriptArrayBuffer**)GetNonVarReg(AsmJsFunctionMemory::ArrayBufferRegister);
+        return arr->GetByteLength() >> 16;
```
MikeHolman: should add nullptr check for arr

lib/Backend/IRBuilderAsmJs.cpp
```diff
@@ -1897,23 +1897,36 @@ IRBuilderAsmJs::BuildAsmReg1(Js::OpCodeAsmJs newOpcode, uint32 offset)
...
+        IR::Instr * instr = IR::Instr::New(Js::OpCode::Ld_A, dstOpnd, addrOpnd, m_func);
+        AddInstr(instr, offset);
+    }
+    else
     {
-        dstOpnd->m_sym->m_isConst = true;
-        dstOpnd->m_sym->m_isNotInt = true;
+        Js::RegSlot dstRegSlot = GetRegSlotFromIntReg(dstReg);
+        IR::RegOpnd * dstOpnd = BuildDstOpnd(dstRegSlot, TyInt32);
+        Assume(m_asmFuncInfo->UsesHeapBuffer());
```
MikeHolman: I don't think this can be assumed. However, I think we should be able to implement such that this is invariant (e.g. if max > 0, we always allocate at least one page). this way, we can have `if(UsesHeapBuffer) { AddInstr(dst = length >> 16)} else { AddInstr(dst = 0)}`

lib/Backend/IRBuilderAsmJs.cpp
```diff
@@ -2582,6 +2595,12 @@ IRBuilderAsmJs::BuildInt2(Js::OpCodeAsmJs newOpcode, uint32 offset, Js::RegSlot
         instr = IR::Instr::New(Js::OpCode::CmEq_I4, dstOpnd, srcOpnd, IR::IntConstOpnd::New(0, TyInt32, m_func), m_func);
         break;
 
+    case Js::OpCodeAsmJs::CurrentMemory_Int:
```
MikeHolman: I guess forgot to remove this?

lib/Backend/IRBuilderAsmJs.cpp
Krovatkin: @MikeHolman  arrgh, silly me :-(  
i didn't realize that memory section could be "optional."

fixed in the last commit

MikeHolman:
:shipit:

PR #1518, opened 2016-08-30
Move pdm based tests to core and run them with ch (jsrt debugging apis)

agarwal-sandeep:
This is first PR for moving couple of PDM based debugger tests to core so that same tests can be run with both PDM based debugging APIs and JsRT debugging APIs via ch. Support have been added to RL so that it can pick source files from a diffrent location than specified in rlexe.xml.


agarwal-sandeep:
@tcare please take a look at RL changes

tcare:
Looks good

PR #1519, opened 2016-08-31
WASM f32/f64 trunc and nearest to use WasmMath

arunetm:
VS13 CRT doesnt support round and trunc. Using CRT methods when SSE4.1 is
not available will break builds using VS13.

Modified trunc and nearest to use WasmMath library helper when SSE4.1 is not available.
Re enabling the feature and tests.


arunetm:
@MikeHolman @Cellule, please have a look. Thanks.

MikeHolman:
:shipit:

arunetm:
@MikeHolman Thanks. Can you please help to sign off the PR.

PR #1521, opened 2016-08-31
Fix a memory leak in JsRT debugging where we were not deleting weak ref of locals display

agarwal-sandeep:


bin/ch/DbgController.js
```diff
@@ -441,6 +441,8 @@ var controllerObj = (function () {
                     if (execStr && execStr.toString().search("removeExpr()") != -1) {
                         _onasyncbreakCommands = undefined;
                     }
+                } else if (id === "debuggerStatement") {
+                    execStr = "dumpBreak();locals();stack();"
```
akroshg: is this part of memory leak fix, or this is functionality gap fix?

lib/Jsrt/JsrtDebuggerObject.cpp
```diff
@@ -295,18 +289,13 @@ Js::DynamicObject * JsrtDebuggerStackFrame::GetLocalsObject(Js::ScriptContext* s
...
-    }
-
     ReferencedArenaAdapter* pRefArena = scriptContext->GetThreadContext()->GetDebugManager()->GetDiagnosticArena();
     Js::IDiagObjectModelDisplay* pLocalsDisplay = Anew(pRefArena->Arena(), Js::LocalsDisplay, this->stackFrame);
-    this->pObjectModelWalker = pLocalsDisplay->CreateWalker();
+    WeakArenaReference<Js::IDiagObjectModelWalkerBase>* objectModelWalker = pLocalsDisplay->CreateWalker();
 
-    if (this->pObjectModelWalker != nullptr)
+    if (objectModelWalker != nullptr)
     {
```
akroshg: curious when can this be null? could we add assert in the else case?

bin/ch/DbgController.js
agarwal-sandeep: This is a functionality gap and fix in Evaluate is for an assert.

lib/Jsrt/JsrtDebuggerObject.cpp
```diff
@@ -826,7 +814,7 @@ Js::JavascriptArray * JsrtDebugStackFrames::StackFrames(Js::ScriptContext * scri
         }
         else
         {
-            this->framesDictionary->Clear();
+            this->ClearFrameDictionary();
```
akroshg: do you have to set frameDictionary to be null here?

lib/Jsrt/JsrtDebuggerObject.cpp
agarwal-sandeep: Only in case of OOM, no harm in having the check.

---
In reply to: [77227996](https://github.com/Microsoft/ChakraCore/pull/1521#discussion_r77227996) [](ancestors = 77227996)

lib/Jsrt/JsrtDebuggerObject.cpp
agarwal-sandeep: No, its allocated just once (the first time). deleted and removed in JsrtDebugStackFrames destructor. 

---
In reply to: [77228263](https://github.com/Microsoft/ChakraCore/pull/1521#discussion_r77228263) [](ancestors = 77228263)

akroshg:
Looks good to me.

PR #1526, opened 2016-09-01
Linux to Master

obastemur:


obastemur:
@agarwal-sandeep please review last commit

From this PR forward we won't be using `linux` branch.

/cc @liminzhu we may want to update Wiki and other places?
/cc @kunalspathak master is ready

/cc @digitalinfinity @jianchun

test/DebuggerCommon/step_in_from_interpreted_function_attach.js.dbg.baseline
```diff
@@ -2,19 +2,19 @@
   {
     "callStack": [
       {
-        "line": 27,
+        "line": 29,
```
agarwal-sandeep: Since you changed the lines in file you will also need to update full repro baseline at [FullRepro]\unittest\DebuggerCommon\step_in_from_interpreted_function_attach.js.dbg.baseline. And since you need to update full baseline you may not do delete WScript.Platform and have Platform show in dbg baseline

obastemur: WScript.Platform is specific to `ch`

obastemur: Another solution is enabling WScript.Platform only when a specific ch flag is used though. This way we won't have similar issues going forward

liminzhu:
@obastemur gotcha. After it's merged I'll point docs to master branch instead of Linux.

test/DebuggerCommon/step_in_from_interpreted_function_attach.js.dbg.baseline
agarwal-sandeep: If WScript.Platform is not needed for majority of the cases then exposing it under a switch is ideal otherwise update the core baseline without modifying the test.

jianchun: I wish we could arbitrarily ignore certain objects from all our baselines: WScript, Debug, ... (may we attach and examine a special property, e.g. hide object data completely if it has property __hide_baseline__?)

agarwal-sandeep: We do but only at top level under locals. Maybe we can revisit and fix ignoring at all levels.

obastemur:
@agarwal-sandeep Thanks for the review

PR #1527, opened 2016-09-01
Disable loading ArgIns from regs

LouisLaf:
Disable loading ArgIns from regs because it breaks unittests for ChakraFull.  I'm working on a better fix.


LouisLaf:
@ThomsonTan  and @satheeshravi , please review

lib/Backend/SccLiveness.cpp
```diff
@@ -379,7 +379,7 @@ SCCLiveness::ProcessSrc(IR::Opnd *src, IR::Instr *instr)
 		IR::SymOpnd *symOpnd = src->AsSymOpnd();
 		RegNum reg = LinearScanMD::GetParamReg(symOpnd, this->func);
 
-		if (reg != RegNOREG && !PHASE_OFF(Js::LivenessPhase, this->func))
+		if (reg != RegNOREG && PHASE_ON(Js::RegParamsPhase, this->func))
```
ThomsonTan: >PHASE_ON(Js::RegParamsPhase, this->func) [](start = 25, length = 40)

Move check to above else if to avoid calling to GetParamReg?

ThomsonTan: >		 [](start = 0, length = 2)

nit: seems tab is used in this new code block (else if block) instead of 4 spaces.

dilijev:
Thanks @LouisLaf

satheeshravi:
:shipit:

LouisLaf:
Thanks Tom.  I'll do the cleanup when I checkin the final fix.

ThomsonTan:
:shipit:

PR #1529, opened 2016-09-02
fix validation bug with implicit void returns in asm.js

MikeHolman:
We weren't failing validation in some cases, which led to functional difference where we should print undefined but return 0 instead.


MikeHolman:
@abchatra, @Cellule: can you take a look?

Cellule:
Are we always guarantied to have a return statement as the last statement when the return type is not void ?
Don't we set the return type to be void by default ?

MikeHolman:
Yes, last statement must be return for non-void function.

We don't have any default state. Call-site coercions are required, so you won't ever be in position where you call a function and don't know if it is void or has some type.

Cellule:
LGTM

PR #1530, opened 2016-09-02
Address deref issue

akroshg:
During the forward global optimizer pass, given a property store that causes an object layout to go from object-header-inlined to
non-object-header-inlined,  kill all type syms with object-header-inlined types to protect against aliasing.

### DIFF
--- a/lib/Backend/GlobOptFields.cpp
+++ b/lib/Backend/GlobOptFields.cpp
@@ -2170,7 +2170,7 @@ GlobOpt::FinishOptPropOp(IR::Instr *instr, IR::PropertySymOpnd *opnd, BasicBlock
         isObjTypeSpecialized = ProcessPropOpInTypeCheckSeq<true>(instr, opnd, block, updateExistingValue, emitsTypeCheckOut, changesTypeValueOut, &isObjTypeChecked);
     }
 
-    if (opnd == instr->GetDst() && this->objectTypeSyms && !isObjTypeChecked)
+    if (opnd == instr->GetDst() && this->objectTypeSyms)
     {
         if (block == nullptr)
         {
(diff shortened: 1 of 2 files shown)


akroshg:
@pleath @LouisLaf @curtisman @rajatd - any of you take a look at this? thanks.

akroshg:
@dotnet-bot test this please.

dilijev:
@dotnet-bot test this please

lib/Backend/GlobOptFields.cpp
```diff
@@ -2170,7 +2170,7 @@ GlobOpt::FinishOptPropOp(IR::Instr *instr, IR::PropertySymOpnd *opnd, BasicBlock
         isObjTypeSpecialized = ProcessPropOpInTypeCheckSeq<true>(instr, opnd, block, updateExistingValue, emitsTypeCheckOut, changesTypeValueOut, &isObjTypeChecked);
     }
 
-    if (opnd == instr->GetDst() && this->objectTypeSyms && !isObjTypeChecked)
+    if (opnd == instr->GetDst() && this->objectTypeSyms)
```
pleath: So the change you're making here is that we kill the current type sym even if its type has been checked upstream? But if the type has been checked, and if the store causes us to transition to a type with a different layout, then we can update the type sym's value in the value table, and that should cause downstream dereferences to be correct. Is that not happening?

akroshg: Wait, I am porting the change you have made. Was there a problem in the porting?

pleath: That's pretty funny. I obviously have vacation brain. I'll look again later.

-- Paul

akroshg:
Since it was just a porting I went ahead and merged it. You can refine this later.

PR #1531, opened 2016-09-02
cross-platform: ChakraCore compiles against iOS amd64

### LINKED ISSUE #1058: Fix use of IsInAssert on Linux
IsInAssert is supposed to be a thread static variable but on Linux it's marked as a global static variable. It's use needs to be examined and fixed up accordingly.

obastemur:
Was fixing an open issue, combined that with a matching part of a side branch
Important Notice: Unless otherwise specified, iOS is not supported

This PR fixes #1058 

- Also removed Desktop alert boxes as they may not play well with server type of apps


obastemur:
@dotnet-bot test this please.

lib/Common/CommonDefines.h
```diff
@@ -693,3 +693,19 @@
...
+    #if defined(_MSC_VER) && _MSC_VER <= 1800 // VS2013?
+        #define THREAD_LOCAL __declspec(thread)
+    #else // VS2015+, linux Clang etc.
+        #define THREAD_LOCAL thread_local
+    #endif // VS2013?
+#else // __APPLE__
+    #ifndef __IOS__
+        #define THREAD_LOCAL _Thread_local
+    #else
+        #define THREAD_LOCAL
```
digitalinfinity: Is this intended? THREAD_LOCAL has no effect on iOS?

lib/Common/Core/Assertions.h
```diff
@@ -15,9 +15,11 @@ _declspec(selectany) int AssertsToConsole = false;
 
 #if _WIN32
 _declspec(thread, selectany) int IsInAssert = false;
+#elif !defined(__IOS__)
+__declspec(thread, selectany) int IsInAssert = false;
```
digitalinfinity: Oh, I guess thread-local is not yet supported on iOS? Ok. Curious, what is the difference between _declspec and __declspec? If they're the same, can we just combine the _WIN32 and !defined(__IOS__) cases?

digitalinfinity:
LGTM

lib/Common/Core/Assertions.h
obastemur: Yes. It's not supported on iOS. There is no difference between single / double underscore. Single one is old MS style. It could be combined though but left it mostly because of an explicit separation.

obastemur:

PR #1536, opened 2016-09-03
Bug fix to not remove GetNewScObject in the loop prepass

rajatd:
We should wait for the main pass of a loop to make changes to the IR based on value types of operands. In this case GetNewScObject instruction was being removed in the prepass and this led to functional differences from nonative runs.


rajatd:
@LouisLaf @pleath , could you review, please?

rajatd:
ping :)

pleath:
:shipit:

PR #1540, opened 2016-09-05
cross-platform: x86 support

obastemur:
ChakraCore compiles to x86 target

- Add xplat x86 Support [~~not tested - linker problems - see below~~]
- Improvements to existing amd64 implementation
- ~~This PR needs 'ICU/etc static linkage' support for linker pass~~
- Additional work will be needed on posix (At the moment, the main focus for this x86 PR is iOS simulator / Android targets)

Important: Unless otherwise specified, x86 is not officially supported.


obastemur:
At the moment, this PR is capable of running javascript on x86 arch / unix. There are couple of missing parts (i.e. function call  stack failures)

Since this is not an official support and this work is beyond compiling step now; (also the improvements to 64 bit side)
@digitalinfinity @jianchun how about merging this work as is (apart from the review issues) and handling the rest in smaller iterations ?

lib/Common/Common/NumberUtilities.inl
```diff
@@ -72,7 +72,7 @@ namespace Js
     NUMBER_UTIL_INLINE INT64 NumberUtilities::TryToInt64(double T1)
     {
         INT64 T4_64;
-#if defined(_M_IX86)
+#if defined(_M_IX86) && defined(_WIN32) // TODO: xplat support
```
digitalinfinity: Why not do the same thing you did for AutoSystemInfo::Data::SSE2Available, and just have that function return false with a todo

lib/Common/Memory/HeapBlock.cpp
```diff
@@ -1590,7 +1590,7 @@ void SmallHeapBlockT<TBlockAttributes>::FillFreeMemory(__in_bcount(size) void *
     }
 
     // REVIEW: Do DbgMemFill on debug build?
-#if defined(_M_IX86)
+#if defined(_M_IX86) && defined(_WIN32) // TODO: xplat support
```
digitalinfinity: Why is this _WIN32 specific?

lib/Common/Common/NumberUtilities.inl
obastemur: Because the instructions are not available on xplat. So, it doesn't compile

digitalinfinity: I'm confused- FLD and FISTTP are not understood by other x86 assemblers? Or is it that we need to pass in a switch to get them to understand x87 instructions? Because FLD is a very old instruction. FISTTP is relatively newer- I think -msse3 should support both. Maybe it should get wrapped in a different ifdef (something like _M_CHAKRA_SSE3 ifdef that the makefile can pass in along with -msse3 to the compiler?)

---
In reply to: [78863793](https://github.com/Microsoft/ChakraCore/pull/1540#discussion_r78863793) [](ancestors = 78863793)

lib/Runtime/Library/JavascriptFunction.cpp
```diff
@@ -1126,8 +1126,24 @@ namespace Js
...
+{
+    // todo: fix this!!!
+    __asm
+    {
+        push    ecx
+        lea     ecx, [esp + 8]
+        sub     ecx, eax
+        and     ecx, (16 - 1)
+        add     eax, ecx
+        ret
```
digitalinfinity: Confused by what this is doing but I guess that's why there is a todo. I'm assuming the fix involves changing esp. Right now, it looks like this would leave the stack unaligned since you do a push ecx but no pop?

lib/Runtime/Library/JavascriptNumber.cpp
```diff
@@ -230,6 +230,14 @@ namespace Js
 
     static const double d1_0 = 1.0;
 
+#ifdef __clang__
+    // todo : make this fast!
+    // Clang doesn't support so many things under the fast version below
```
digitalinfinity: Weird- are we using -msse3?

lib/Runtime/Library/MathLibrary.cpp
```diff
@@ -147,7 +147,7 @@ namespace Js
     {
         double result;
 
-#if defined(_M_IX86)
+#if defined(_M_IX86) && defined(_WIN32) // TODO: xplat support
```
digitalinfinity: Yeah, I'd do this under a different ifdef (assuming -msse3 doesn't fix it)

pal/inc/pal.h
```diff
@@ -2026,7 +2026,7 @@ QueueUserAPC(
          IN HANDLE hThread,
          IN ULONG_PTR dwData);
 
-#ifdef _X86_
+#ifdef __i686__
```
digitalinfinity: What's the reason for this change?

pal/inc/pal_mstypes.h
```diff
@@ -42,22 +42,11 @@ extern "C" {
...
-#if !defined(__cdecl)
-#define __cdecl        __attribute__((cdecl))
-#endif
-#if !defined(_cdecl)
-#define _cdecl         __cdecl
-#endif
-#if !defined(CDECL)
-#define CDECL          __cdecl
-#endif
+#define __stdcall
```
digitalinfinity: Curious- does clang not support stdcall?

pal/src/arch/i386/activationhandlerwrapper.S
```diff
@@ -28,4 +28,23 @@ LOCAL_LABEL(ActivationHandlerReturn):
     ret
 NESTED_END ActivationHandlerWrapper, _TEXT
 
+#else
+
+// Offset of the return address from the ActivationHandler in the ActivationHandlerWrapper
+.globl C_FUNC(ActivationHandlerReturnOffset)
```
digitalinfinity: Who needs this?

pal/src/debug/debug.cpp
```diff
@@ -422,7 +422,12 @@ DebugBreak(
 BOOL
 IsInDebugBreak(void *addr)
 {
+#ifdef _M_IX86
+    // todo: implement this
+    return FALSE;
```
digitalinfinity: Hmm- why does the original implementation not work?

digitalinfinity:
:shipit:

pal/src/debug/debug.cpp
obastemur: Depending PAL part fails. Disabled temporarily

pal/src/arch/i386/activationhandlerwrapper.S
obastemur: machexception

pal/inc/pal_mstypes.h
obastemur: we already disabled it for x64. It is not because it doesn't support that. It is because of our usage of templates.

lib/Runtime/Library/JavascriptFunction.cpp
obastemur: close. there is much more to do there (i.e. we use more than 0x1000 ??)

lib/Runtime/Library/MathLibrary.cpp
obastemur: @digitalinfinity `__libm_sse2_acos` (msvcrt) is the problem there.

lib/Runtime/Library/JavascriptNumber.cpp
obastemur: It is not about msse3 etc. Clang inline assembly does not support non-asm in naked function.

lib/Runtime/Library/JavascriptFunction.cpp
curtisman: So this doesn't really work at all? since it doesn't move ESP?
Should we just do "int 3" here so we are guarantee to crash?

obastemur: `int 3` is a good idea

lib/Runtime/Library/JavascriptNumber.cpp
curtisman: I think this is only called by the JIT.  And this change will break that.
Might want to put it under #ifdef ENABLE_NATIVE_CODEGEN instead?  So we will hit this and fix it property when we work on x86 JIT?

pal/inc/pal_mstypes.h
curtisman: So template doesn't support stdcall?

obastemur: Surely supports but we do violate some things clang is pretty strict

pal/src/debug/debug.cpp
curtisman: Put more detail in the comment please.

lib/Runtime/Library/MathLibrary.cpp
curtisman: Yeah... this depends on whether the xplat CRT has a direct entry point to the SSE2 version of these trig functions.....  My guess is probably not :(  Even if they have it is probably under a different name...  defined(_WIN32) is probably fine.
A better comment would be great.

obastemur: @curtisman  Thought `lib_sse2_acos` can be implemented. So no need ?

curtisman: That probably isn't necessary. This is really for perf.  Functionally, this is not necessary.
If the xplat CRT implementation already support SSE2, then we get most of the perf already.
It just avoid the is SSE2 support check at the beginning of these functions.

I'd rather not maintain our copy of these trig functions.

obastemur: updated accordingly

obastemur:
@digitalinfinity @curtisman Thanks for the review

PR #1541, opened 2016-09-06
cross-platform: --no-icu support

obastemur:
All the none-unicode tests pass

Added bunch of 'todo' comments for unicode improvements without ICU


lib/Parser/CharClassifier.cpp
```diff
@@ -419,11 +419,12 @@ Js::CharClassifier::CharClassifier(ScriptContext * scriptContext)
     }
 #endif
 
-#if ENABLE_UNICODE_API
     // If we're in ES6 mode, and we have full support for Unicode character classification
     // from an external library, then use the ES6/Surrogate pair supported versions of the functions
     // Otherwise, fallback to the ES5 versions which don't need an external library
+#if ENABLE_UNICODE_API
```
digitalinfinity: Curious, why did you chose to use the ES6 functions instead of the default functions if ENABLE_UNICODE_API is not enabled?

digitalinfinity:
LGTM

lib/Parser/CharClassifier.cpp
obastemur: IMHO; ENABLE_UNICODE_API is there temporarily (for now) In order to complete no-icu support properly (i.e. current unicode test cases passing) I thought, it could be nice to take the ES6 path and list of TODOs. 

We may want to redefine 'ENABLE_UNICODE_API' for Intl support only.

obastemur:

PR #1542, opened 2016-09-06
Delete property optimizations

rajatd:
A couple of optimizations for deleting a property on an object.

1. When we delete a property, we change the type of the object to invalidate any caches which may have cached its type. We don't have to change the type if it wasn't cached anywhere.

2. For objects that have SimpleDictionary or Dictionary Type Handlers, we have the capability of looking up a property descriptor from the property name itself. However, currently we first lookup the property record for the property in the property map and use the record to look up the descriptor. Getting rid of the intermediate step of property record lookup if we have the property name (delete obj[prop])

### DIFF
--- a/lib/Runtime/Language/JavascriptOperators.cpp
+++ b/lib/Runtime/Language/JavascriptOperators.cpp
@@ -2700,6 +2700,40 @@ CommonNumber:
     {
         return DeleteProperty_Impl<false>(instance, propertyId, propertyOperationFlags);
     }
+
+    bool JavascriptOperators::ShouldTryDeleteProperty(RecyclableObject* instance, JavascriptString *propertyNameString, PropertyRecord const **pPropertyRecord)
+    {
+        PropertyRecord const *propertyRecord = nullptr;
+        if (!JavascriptOperators::CanShortcutOnUnknownPropertyName(instance))
+        {
+            instance->GetScriptContext()->GetOrAddPropertyRecord(propertyNameString->GetString(), propertyNameString->GetLength(), &propertyRecord);
+        }
+        else
+        {
+            instance->GetScriptContext()->FindPropertyRecord(propertyNameString, &propertyRecord);
+        }
+
+        if (propertyRecord == nullptr)
+        {
+            return false;
+        }
+        *pPropertyRecord = propertyRecord;
+        return true;
+    }
+
+    BOOL JavascriptOperators::DeleteProperty(RecyclableObject* instance, JavascriptString *propertyNameString, PropertyOperationFlags propertyOperationFlags)
+    {
+#ifdef ENABLE_MUTATION_BREAKPOINT
+        ScriptContext *scriptContext = instance->GetScriptContext();
+        if (MutationBreakpoint::IsFeatureEnabled(scriptContext)
+            && scriptContext->HasMutationBreakpoints())
+        {
+            MutationBreakpoint::HandleDeleteProperty(scriptContext, instance, propertyNameString);
+        }
+#endif
+        return instance->DeleteProperty(propertyNameString, propertyOperationFlags);
+    }
+
     BOOL JavascriptOperators::DeletePropertyUnscopables(RecyclableObject* instance, PropertyId propertyId, PropertyOperationFlags propertyOperationFlags)
     {
         return DeleteProperty_Impl<true>(instance, propertyId, propertyOperationFlags);
@@ -2707,7 +2741,6 @@ CommonNumber:
     template<bool unscopables>
     BOOL JavascriptOperators::DeleteProperty_Impl(RecyclableObject* instance, PropertyId propertyId, PropertyOperationFlags propertyOperationFlags)
     {
-
         if (unscopables && JavascriptOperators::IsPropertyUnscopable(instance, propertyId))
         {
             return false;
@@ -4846,12 +4879,17 @@ CommonNumber:
 
         uint32 indexVal;
         PropertyRecord const * propertyRecord;
+        JavascriptString * propertyNameString = nullptr;
         BOOL result = TRUE;
-        IndexType indexType = GetIndexType(index, scriptContext, &indexVal, &propertyRecord, false);
+        IndexType indexType = GetIndexType(index, scriptContext, &indexVal, &propertyRecord, &propertyNameString, false, true);
 
         if (indexType == IndexType_Number)
         {
             result = JavascriptOperators::DeleteItem(object, indexVal, propertyOperationFlags);
+        }
+        else if (indexType == IndexType_JavascriptString)
+        {
+            result = JavascriptOperators::DeleteProperty(object, propertyNameString, propertyOperationFlags);
         }
         else
         {
@@ -8375,7 +8413,7 @@ CommonNumber:
                 // Else, set it to next element after current nextEvictionVictim index
                 cache->nextEvictionVictim = (cache->nextEvictionVictim + 1) & (EQUIVALENT_TYPE_CACHE_SIZE - 1);
             }
-
+   
             if (PHASE_TRACE1(Js::EquivObjTypeSpecPhase))
             {
                 Output::Print(_u("EquivObjTypeSpec: Saving type in used slot of equiv types cache at index = %d. NextEvictionVictim = %d. \n"), index, cache->nextEvictionVictim);
@@ -8383,7 +8421,7 @@ CommonNumber:
             }
             Assert(index < EQUIVALENT_TYPE_CACHE_SIZE);
             __analysis_assume(index < EQUIVALENT_TYPE_CACHE_SIZE);
-            equivTypes[index] = type;
+            equivTypes[index] = type;   
         }
         
         // Fixed field checks allow us to assume a specific type ID, but the assumption is only
@@ -8398,6 +8436,7 @@ CommonNumber:
             }
         }
 
+        type->SetHasBeenCached();
         guard->SetTypeAddr((intptr_t)type);
         return true;
     }
--- a/lib/Runtime/Library/GlobalObject.cpp
+++ b/lib/Runtime/Library/GlobalObject.cpp
@@ -1576,7 +1576,6 @@ LHexError:
         Assert(!(callInfo.Flags & CallFlags_New));
 
         ScriptContext* scriptContext = function->GetScriptContext();
-
         if (!scriptContext->GetConfig()->IsCollectGarbageEnabled()
 #ifdef ENABLE_PROJECTION
             && scriptContext->GetConfig()->GetHostType() != HostType::HostTypeApplication
@@ -2115,6 +2114,18 @@ LHexError:
         }
 
         // Non-existent property
+        return TRUE;
+    }
+
+    BOOL GlobalObject::DeleteProperty(JavascriptString *propertyNameString, PropertyOperationFlags flags)
+    {
+        PropertyRecord const *propertyRecord = nullptr;
+        if (JavascriptOperators::ShouldTryDeleteProperty(this, propertyNameString, &propertyRecord))
+        {
+            Assert(propertyRecord);
+            return DeleteProperty(propertyRecord->GetPropertyId(), flags);
+        }
+
         return TRUE;
     }
 
--- a/lib/Runtime/Library/JavascriptStringObject.cpp
+++ b/lib/Runtime/Library/JavascriptStringObject.cpp
@@ -331,6 +331,18 @@ namespace Js
         return DynamicObject::DeleteProperty(propertyId, flags);
     }
 
+    BOOL JavascriptStringObject::DeleteProperty(JavascriptString *propertyNameString, PropertyOperationFlags propertyOperationFlags)
+    {
+        JsUtil::CharacterBuffer<WCHAR> propertyName(propertyNameString->GetString(), propertyNameString->GetLength());
+        if (BuiltInPropertyRecords::length.Equals(propertyName))
+        {
+            JavascriptError::ThrowCantDeleteIfStrictMode(propertyOperationFlags, this->GetScriptContext(), propertyNameString->GetString());
+
+            return FALSE;
+        }
+        return DynamicObject::DeleteProperty(propertyNameString, propertyOperationFlags);
+    }
+
     BOOL JavascriptStringObject::HasItem(uint32 index)
     {
         if (this->InternalUnwrap()->HasItem(index))
--- a/lib/Runtime/Types/DictionaryTypeHandler.cpp
+++ b/lib/Runtime/Types/DictionaryTypeHandler.cpp
@@ -869,6 +869,81 @@ namespace Js
     }
 
     template <typename T>
+    BOOL DictionaryTypeHandlerBase<T>::DeleteProperty(DynamicObject *instance, JavascriptString *propertyNameString, PropertyOperationFlags propertyOperationFlags)
+    {
+        AssertMsg(!PropertyRecord::IsPropertyNameNumeric(propertyNameString->GetString(), propertyNameString->GetLength()),
+            "Numeric property names should have been converted to uint or PropertyRecord* ");
+
+        ScriptContext* scriptContext = instance->GetScriptContext();
+        DictionaryPropertyDescriptor<T>* descriptor;
+        JsUtil::CharacterBuffer<WCHAR> propertyName(propertyNameString->GetString(), propertyNameString->GetLength());
+
+        if (propertyMap->TryGetReference(propertyName, &descriptor))
+        {
+            Assert(descriptor->SanityCheckFixedBits());
+
+            if (descriptor->Attributes & PropertyDeleted)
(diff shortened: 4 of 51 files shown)


rajatd:
@curtisman, @LouisLaf , @pleath, @kunalspathak  could you guys please review this?

rajatd:
Hi, could I get some eyes on this?

dilijev:
@rajatd Don't check in JavascriptOperators.cpp.orig

rajatd:
Oh, thanks for noticing that, @dilijev . Will make sure of that

lib/Runtime/Language/JavascriptOperators.cpp
```diff
@@ -8310,6 +8327,16 @@ namespace Js
             return false;
         }
 
+     	bool evictCachedType = true;
+        if (DynamicType::Is(type->GetTypeId()))
+        {
+            DynamicType* dynamicType = static_cast<DynamicType*>(type);
+            if ((dynamicType->GetTypeHandler()->IsSimpleDictionaryTypeHandler() || dynamicType->GetTypeHandler()->IsDictionaryTypeHandler()) &&
```
kunalspathak: >(dynamicType->GetTypeHandler()->IsSimpleDictionaryTypeHandler() || dynamicType->GetTypeHandler()->IsDictionaryTypeHandler()) [](start = 16, length = 124)

Why not just have condition for non-shared type. BTW, I have similar changes in [https://github.com/Microsoft/ChakraCore/pull/1578](https://github.com/Microsoft/ChakraCore/pull/1578), so you might want to hold off on this?

lib/Runtime/Library/GlobalObject.cpp
```diff
@@ -2118,6 +2117,24 @@ namespace Js
         return TRUE;
     }
 
+    BOOL GlobalObject::DeleteProperty(JavascriptString *propertyNameString, PropertyOperationFlags flags)
+    {
+        PropertyRecord const *propertyRecord = nullptr;
```
kunalspathak: >PropertyRecord const *propertyRecord = nullptr; [](start = 8, length = 47)

This code is repeated at lot of places. Can it go in a static method that takes "this" as parameter and then you call "DeleteProperty" on this?

lib/Runtime/Library/JavascriptStringObject.cpp
```diff
@@ -331,6 +331,18 @@ namespace Js
         return DynamicObject::DeleteProperty(propertyId, flags);
     }
 
+    BOOL JavascriptStringObject::DeleteProperty(JavascriptString *propertyNameString, PropertyOperationFlags propertyOperationFlags)
+    {
+        JsUtil::CharacterBuffer<WCHAR> propertyName(propertyNameString->GetString(), propertyNameString->GetLength());
+        if (BuiltInPropertyRecords::length.Equals(propertyName))
```
kunalspathak: >BuiltInPropertyRecords::length.E [](start = 12, length = 32)

likewise this code is repeated at few other places.

lib/Runtime/Language/JavascriptOperators.cpp
kunalspathak: Probably worth adding an assert that all equivTypes has cached flag set, equivTypes[i]->HasBeenCached()

---
In reply to: [78785136](https://github.com/Microsoft/ChakraCore/pull/1542#discussion_r78785136) [](ancestors = 78785136)

lib/Runtime/Types/Type.h
```diff
@@ -62,6 +63,14 @@ namespace Js
         inline BOOL SkipsPrototype() const { return (this->flags & TypeFlagMask_SkipsPrototype) != 0 ; }
         inline BOOL CanHaveInterceptors() const { return (this->flags & TypeFlagMask_CanHaveInterceptors) != 0; }
         inline BOOL IsFalsy() const { return flags & TypeFlagMask_IsFalsy; }
+        inline BOOL HasBeenCached() const { return flags & TypeFlagMask_HasBeenCached; }
+        inline void SetHasBeenCached(const bool hasBeenCached)
+        {
+            // Once set, this flag should never be reset.
```
kunalspathak: >// Once set, this flag should never be reset. [](start = 12, length = 45)

what happens if we cached a type and later invalidated it and it is no longer cached. Will this flag still remain ON?

lib/Runtime/Types/DictionaryTypeHandler.cpp
```diff
@@ -869,6 +869,81 @@ namespace Js
     }
 
     template <typename T>
+    BOOL DictionaryTypeHandlerBase<T>::DeleteProperty(DynamicObject *instance, JavascriptString *propertyNameString, PropertyOperationFlags propertyOperationFlags)
```
kunalspathak: >DeleteProperty [](start = 39, length = 14)

nit: can you move this method next to DeleteProperty_Internal() as it is similar to that?

lib/Runtime/Types/DictionaryTypeHandler.cpp
```diff
@@ -869,6 +869,81 @@ namespace Js
...
+        DictionaryPropertyDescriptor<T>* descriptor;
+        JsUtil::CharacterBuffer<WCHAR> propertyName(propertyNameString->GetString(), propertyNameString->GetLength());
+
+        if (propertyMap->TryGetReference(propertyName, &descriptor))
+        {
+            Assert(descriptor->SanityCheckFixedBits());
+
+            if (descriptor->Attributes & PropertyDeleted)
+            {
+                return true;
```
kunalspathak: >return true; [](start = 16, length = 12)

If i compare the code from DeleteProperty_Internal() below, there is some code missing here. Is that intentional?
```c++
 // If PropertyDeleted and PropertyLetConstGlobal are set then we have both
                // a deleted global property and let/const variable in this descriptor.
                // If allowLetConstGlobal is true then the let/const shadows the property
                // and we should return false for a failed delete by going into the else
                // if branch below.
                if (allowLetConstGlobal && (descriptor->Attributes & PropertyLetConstGlobal))
                {
                    JavascriptError::ThrowCantDeleteIfStrictMode(propertyOperationFlags, scriptContext, propertyRecord->GetBuffer());

                    return false;
                }
```

lib/Runtime/Language/JavascriptOperators.cpp
rajatd: Because a type with a path type handler may start off as non-shared but is definitely shareable. Didn't want to evict such types

---
In reply to: [78785136](https://github.com/Microsoft/ChakraCore/pull/1542#discussion_r78785136) [](ancestors = 78785136)

lib/Runtime/Types/Type.h
rajatd: Yes, it will. In that case, we just won't get the optimization. 

---
In reply to: [78790053](https://github.com/Microsoft/ChakraCore/pull/1542#discussion_r78790053) [](ancestors = 78790053)

lib/Runtime/Types/DictionaryTypeHandler.cpp
rajatd: Well, it is really the counterpart of DeleteProperty which takes PropertyId as parameter.

---
In reply to: [78791124](https://github.com/Microsoft/ChakraCore/pull/1542#discussion_r78791124) [](ancestors = 78791124)

lib/Runtime/Types/DictionaryTypeHandler.cpp
rajatd: Yes, it is intentional. allowLetConstGlobal is true only when dealing with a property on the root object. However, if we are coming to this function, the object is not a root object. Does that make sense? I was a little skeptical of this part, so I appreciate a second set of eyes.

---
In reply to: [78791249](https://github.com/Microsoft/ChakraCore/pull/1542#discussion_r78791249) [](ancestors = 78791249)

lib/Runtime/Types/Type.h
kunalspathak: When we delete a property, we change the type of the object to invalidate any caches which may have cached its type. We don't have to change the type if it wasn't cached anywhere.
But won't it mean deleteProperty on invalidated type change the type even though it shouldn't because the type is no longer cached?

---
In reply to: [78816821](https://github.com/Microsoft/ChakraCore/pull/1542#discussion_r78816821) [](ancestors = 78816821,78790053)

lib/Runtime/Types/DictionaryTypeHandler.cpp
kunalspathak: Sure, what I mean is it will be good to have those 2 methods next to each other.

---
In reply to: [78817010](https://github.com/Microsoft/ChakraCore/pull/1542#discussion_r78817010) [](ancestors = 78817010,78791124)

kunalspathak: I am also worried about have "counterpart" code you added in all the files. In future, people might make changes in one of the method and forget making similar change in its counterpart. Instead is there a way to extract out common part in a method that both properytNameString and propertyId would call?

---
In reply to: [78818261](https://github.com/Microsoft/ChakraCore/pull/1542#discussion_r78818261) [](ancestors = 78818261,78817010,78791124)

lib/Runtime/Types/Type.h
rajatd: Not changing the type on a delete property if the type wasn't cached is an optimization, not a functional change. So, in the case that you brought up, we'll just behave the same as we do now.

---
In reply to: [78818005](https://github.com/Microsoft/ChakraCore/pull/1542#discussion_r78818005) [](ancestors = 78818005,78816821,78790053)

kunalspathak: > is an optimization, not a functional change

Sure, I get that. But your change is optimizing that very case. Correct me if I am wrong, but won't it be more complete to optimize for cases where types are invalidated (which happens a lot all the time). 

---
In reply to: [78828446](https://github.com/Microsoft/ChakraCore/pull/1542#discussion_r78828446) [](ancestors = 78828446,78818005,78816821,78790053)

lib/Runtime/Types/DictionaryTypeHandler.cpp
rajatd: Yes, let me look into refactoring that.

---
In reply to: [78818835](https://github.com/Microsoft/ChakraCore/pull/1542#discussion_r78818835) [](ancestors = 78818835,78818261,78817010,78791124)

lib/Runtime/Types/Type.h
rajatd: Types are invalidated? Do you mean caches are invalidated? If we want to optimize those cases, the tracking becomes much harder - we'll basically have to somehow track all the caches in which a type has been cached, then reset this flag once all of those caches have been invalidated. (When an inline cache gets invalidated, there may still be others which hold on to the type)

---
In reply to: [78846360](https://github.com/Microsoft/ChakraCore/pull/1542#discussion_r78846360) [](ancestors = 78846360,78828446,78818005,78816821,78790053)

lib/Runtime/Types/Type.h
```diff
@@ -62,6 +63,14 @@ namespace Js
         inline BOOL SkipsPrototype() const { return (this->flags & TypeFlagMask_SkipsPrototype) != 0 ; }
         inline BOOL CanHaveInterceptors() const { return (this->flags & TypeFlagMask_CanHaveInterceptors) != 0; }
         inline BOOL IsFalsy() const { return flags & TypeFlagMask_IsFalsy; }
+        inline BOOL HasBeenCached() const { return flags & TypeFlagMask_HasBeenCached; }
+        inline void SetHasBeenCached(const bool hasBeenCached)
```
kunalspathak: >SetHasBeenCached [](start = 20, length = 16)

why to even have parameter `hasBeenCached`? Just assume that caller of this API wants to set the `TypeFlagMask_HasBeenCached` flag.

lib/Runtime/Types/Type.h
kunalspathak: > Do you mean caches are invalidated?

Yes, I meant inline caches holding those types get invalidated.

Yeah, I see what you mean. Probably having refcounting would help where every time you cache a type, you increase its refcount and we remove it, you decrease its refcount. If refcount == 0 means type is not cached. I am not sure how much benefit it will give but in big benchmarks like acmeair, lot of invalidation do happen and technically for those types, it is okay not to create new type for delete property.

---
In reply to: [78863703](https://github.com/Microsoft/ChakraCore/pull/1542#discussion_r78863703) [](ancestors = 78863703,78846360,78828446,78818005,78816821,78790053)

lib/Runtime/Types/Type.h
rajatd: I had the parameter to guard against illegal use of this API

---
In reply to: [79017104](https://github.com/Microsoft/ChakraCore/pull/1542#discussion_r79017104) [](ancestors = 79017104)

lib/Runtime/Types/DictionaryTypeHandler.cpp
rajatd: @kunalspathak, whether propertyString and propertyId versions call any common code depends on the type of the object and the kind of the type handler. As you might have noticed, there are cases in which the propertyString version falls back to the propertyId after doing a few checks. Those checks are only required in the propertyString version. 
Type handlers that dont support deleting a property by name already call the propertyString version of the base class TypeHandler.

---
In reply to: [78863297](https://github.com/Microsoft/ChakraCore/pull/1542#discussion_r78863297) [](ancestors = 78863297,78818835,78818261,78817010,78791124)

lib/Runtime/Types/DictionaryTypeHandler.cpp
```diff
@@ -2460,7 +2538,8 @@ namespace Js
     }
 
     template <typename T>
-    void DictionaryTypeHandlerBase<T>::InvalidateFixedField(DynamicObject* instance, PropertyId propertyId, DictionaryPropertyDescriptor<T>* descriptor)
+    template <typename TPropertyKey>
```
kunalspathak: >typename TPropertyKey [](start = 14, length = 21)

Dumb question. You have renamed template variable T to TPropertyKey. How does it work for DictionaryPropertyDescriptor<T>* descriptor in the signature?

rajatd: T is still there :) one line above TPropertyKey

---
In reply to: [79702610](https://github.com/Microsoft/ChakraCore/pull/1542#discussion_r79702610) [](ancestors = 79702610)

kunalspathak: Ah, yes. Of course. :)

---
In reply to: [79732431](https://github.com/Microsoft/ChakraCore/pull/1542#discussion_r79732431) [](ancestors = 79732431,79702610)

dilijev: Why not `template <typename T, typename TPropertyKey>`?

kunalspathak:
:shipit:

lib/Runtime/Types/DictionaryTypeHandler.cpp
rajatd: T is for the class DictionaryTypeHandlerBase, TPropertyKey is for this function only. Similar to the way it is written for other methods such as TryGetFixedProperty in this file and others in SimpleDictionaryTypeHandler.cpp

---
In reply to: [79738900](https://github.com/Microsoft/ChakraCore/pull/1542#discussion_r79738900) [](ancestors = 79738900)

dilijev:
@rajatd Once you rebase onto master or merge master into your branch, the Tab Check will pass. Right now, the script is missing from your branch.

rajatd:

lib/Runtime/Language/JavascriptOperators.cpp
```diff
@@ -8375,15 +8413,15 @@ namespace Js
                 // Else, set it to next element after current nextEvictionVictim index
                 cache->nextEvictionVictim = (cache->nextEvictionVictim + 1) & (EQUIVALENT_TYPE_CACHE_SIZE - 1);
             }
-
+   
             if (PHASE_TRACE1(Js::EquivObjTypeSpecPhase))
             {
-                Output::Print(_u("EquivObjTypeSpec: Saving type in used slot of equiv types cache at index = %d. NextEvictionVictim = %d. \n"), index, cache->nextEvictionVictim);
+                Output::Print(_u("Evicted a type from equiv type cache because of a non shared dictionary type"));
```
kunalspathak: >Output::Print(_u("Evicted a type from equiv type cache because of a non shared dictionary type")); [](start = 16, length = 98)

why did you change this?

rajatd: Sorry, bad merge.

rajatd:

dilijev:
@dotnet-bot 
test Windows x64_test
test Windows x86_debug
test Windows x86_test

PR #1544, opened 2016-09-06
need marshal while adding item to proxy ownkeys trap snapshot array

leirocks:


leirocks:
@Yongqu  can you review?

Yongqu:
:shipit:

PR #1546, opened 2016-09-06
Move WinRT date code to (full) file in which it is used.

EdMaurer:


lib/Common/Common/DateUtilities.h
```diff
@@ -17,20 +17,19 @@ namespace Js
...
-        static const INT64 jsEpochMilliseconds;
 
     public:
-        static HRESULT WinRTDateToES5Date(INT64 winrtDate, __out double* pResult);
+
+        static const INT64 ticksPerMillisecond;
+        static const double ticksPerMillisecondDouble;
+        static const INT64 jsEpochMilliseconds;
+
         static HRESULT ES5DateToWinRTDate(double es5Date, __out INT64* pResult);
```
agarwal-sandeep: >ES5DateToWinRTDate [](start = 23, length = 18)

This also seems WinRT specific, do we plan to remove this in next iteration?

agarwal-sandeep:
:shipit:

lib/Common/Common/DateUtilities.h
EdMaurer: ES5DateToWinRTDate is wound into the Intl code and can't obviously move out.

PR #1547, opened 2016-09-07
Stack nested functions in jitted loop bodies

pleath:
Allow jitted loop bodies to make use of stack nested functions if indicated by the byte code. The prologue code will initialize a pointer to the interpreter instance's stack nested functions, and each NewScFunc that allocates on the stack will make use of the pointer. The IRBuilder inserts an instruction that refers to the closure environment and the stack nested function pointer and feeds the NewScFunc instruction, since NewScFunc already takes 2 source operands in some cases.

### DIFF
--- a/lib/Backend/Func.cpp
+++ b/lib/Backend/Func.cpp
@@ -147,25 +147,20 @@ Func::Func(JitArenaAllocator *alloc, CodeGenWorkItem* workItem, const Js::Functi
         m_workItem->GetEntryPoint()->SetHasJittedStackClosure();
     }
 
-    if (m_workItem->Type() == JsFunctionType)
+    if (m_workItem->Type() == JsFunctionType &&
+        m_jnFunction->GetDoBackendArgumentsOptimization() && 
+        !m_jnFunction->GetHasTry())
     {
-        if (m_jnFunction->GetDoBackendArgumentsOptimization() && !m_jnFunction->GetHasTry())
-        {
-            // doBackendArgumentsOptimization bit is set when there is no eval inside a function
-            // as determined by the bytecode generator.
-            SetHasStackArgs(true);
-        }
-        if (doStackNestedFunc && m_jnFunction->GetNestedCount() != 0 &&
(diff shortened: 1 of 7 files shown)


lib/Backend/Func.cpp
```diff
@@ -147,25 +147,18 @@ Func::Func(JitArenaAllocator *alloc, CodeGenWorkItem* workItem, const Js::Functi
...
-            stackNestedFunc = true;
-            this->GetTopFunc()->hasAnyStackNestedFunc = true;
-        }
+        // doBackendArgumentsOptimization bit is set when there is no eval inside a function
+        // as determined by the bytecode generator.
+        SetHasStackArgs(true);
     }
-    else
+    if (doStackNestedFunc && m_jnFunction->GetNestedCount() != 0 &&
+        (this->IsTopFunc() || this->GetTopFunc()->m_workItem->Type() != JsLoopBodyWorkItemType)) // make sure none of the functions inlined in a jitted loop body allocate nested functions on the stack
```
rajatd: >this->GetTopFunc()->m_workItem->Type() != JsLoopBodyWorkItemType [](start = 30, length = 64)

Can we now get rid of this condition for inlinees in a jitted loop body?

rajatd:
:shipit:

lib/Backend/Func.cpp
pleath: I don't think so. An inlinee can't make use of the functions allocated by the interpreter instance, of course. Are we allocating nested functions on the parent's frame if the top-level function is a jitted loop body? (If we're not, we could, but I didn't want to do that as part of this change.)

rajatd: Right now, we are not allocating nested functions in an inlinee on the stack at all if the inlinee was in a jitted loop body. I wanted to know if we could do that now. Agreed that it is out of scope of this change.

---
In reply to: [77870462](https://github.com/Microsoft/ChakraCore/pull/1547#discussion_r77870462) [](ancestors = 77870462)

PR #1554, opened 2016-09-08
Fix memory leak around inlineCache invalidationList

kunalspathak:
When inlineCaches are created, they are registered in `threadContext` in an invalidation list and maintained using `registeredInlineCacheCount` in `threadContext`. When these caches are removed from invalidation list, we record that number as well using `unregisteredInlineCacheCount` in `threadContext`. If the ratio of `registeredInlineCacheCount` to `unregisteredInlineCacheCount` is less than 4, we perform compaction of the invalidation list by deleting `unregisteredInlineCacheCount` nodes from the list that doesn't hold inlineCache and return their memory back to the arena. 

* Ideally, `unregisteredInlineCacheCount` should always match no. of inlineCaches that were removed from invalidation list (aka `cachesRemoved`). However today, `cachesRemoved > unregisteredInlineCacheCount` most of the time. Because of this, we were not returning `cachesRemoved - unregisteredInlineCacheCount` nodes of invalidation list back to arena and the memory taken up by these nodes kept piling leading to memory leak. The reason for `cachesRemoved > unregisteredInlineCacheCount` was because there were couple of places where we were not recording the removal of inlineCaches in `unregisteredInlineCacheCount`. Also, we didn't update `unregisteredInlineCacheCount` when we bulk deleted the invalidation list. Fixing these 2 places, the `cachesRemoved == unregisteredInlineCacheCount` holds true.

* `registeredInlineCacheCount` was updated (reduced that count) when we bulk delete entire invalidation lists. However we never updated (again reduced that count) when we performed compaction of invalidation list. Because of this, `registeredInlineCacheCount` kept growing and it became harder and harder to meet the condition of compaction. This leads to nodes having invalid inline cache still holding memory causing leak.

Also did minor code cleanup and added asserts.

### DIFF
--- a/lib/Runtime/Library/ScriptFunction.cpp
+++ b/lib/Runtime/Library/ScriptFunction.cpp
@@ -753,8 +753,9 @@ namespace Js
                 }
             }
 
-            if (!isShutdown && unregisteredInlineCacheCount > 0 && !scriptContext->IsClosed())
+            if (unregisteredInlineCacheCount > 0)
             {
+                AssertMsg(!isShutdown && !scriptContext->IsClosed(), "Unregistration of inlineCache should only be done if this is not shutdown or scriptContext closing.");
                 scriptContext->GetThreadContext()->NotifyInlineCacheBatchUnregistered(unregisteredInlineCacheCount);
             }
         }
--- a/lib/Runtime/Base/FunctionBody.cpp
+++ b/lib/Runtime/Base/FunctionBody.cpp
@@ -7170,13 +7170,11 @@ namespace Js
             }
         }
 
-        if (!IsScriptContextShutdown)
+        if (unregisteredInlineCacheCount > 0)
         {
+            AssertMsg(!IsScriptContextShutdown, "Unregistration of inlineCache should only be done if this is not scriptContext shutdown.");
             ThreadContext* threadContext = this->m_scriptContext->GetThreadContext();
-            if (unregisteredInlineCacheCount > 0)
-            {
-                threadContext->NotifyInlineCacheBatchUnregistered(unregisteredInlineCacheCount);
-            }
+            threadContext->NotifyInlineCacheBatchUnregistered(unregisteredInlineCacheCount);
         }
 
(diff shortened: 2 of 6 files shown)


kunalspathak:
@pleath, please review.

lib/Runtime/Language/InlineCache.h
```diff
@@ -513,10 +513,6 @@ namespace Js
         template<bool CheckLocal, bool CheckProto, bool CheckAccessor>
         void CloneInlineCacheToEmptySlotInCollision(Type *const type, uint index);
 
-#ifdef CLONE_INLINECACHE_TO_EMPTYSLOT
```
rajatd: >CLONE_INLINECACHE_TO_EMPTYSLOT [](start = 7, length = 30)

Please dont delete this code. There has been an instance where we ended up using some code under this ifdef. The code under this ifdef is still good to keep around for experimenting with it.

kunalspathak: Alright. Thanks for letting me know.

lib/Runtime/Library/ScriptFunction.cpp
```diff
@@ -753,7 +753,7 @@ namespace Js
                 }
             }
 
-            if (!isShutdown && unregisteredInlineCacheCount > 0 && !scriptContext->IsClosed())
+            if (unregisteredInlineCacheCount > 0)
```
rajatd: How is it okay to remove the shutdown and scriptContext closed conditions?

kunalspathak: If you see above, we call `RemoveFromInvalidationList()` only if `!isShutdown` and `!scriptContext->IsClosed` which means `unregisteredInlineCacheCount  > 0` only if this 2 conditions are met. So the check here was redundant.

pleath: Can it be replaced with an assert?

kunalspathak: Replaced.

PR #1555, opened 2016-09-08
fix issue where passing global object fails asm.js validation

MikeHolman:
We were doing lookup to see if the library object that the user passed has SIMD, and then we look at the SIMD properties which are used. So even if SIMD is disabled and unused, we still do that one property lookup of SIMD itself. Since SIMD is not on the global object (this is behind a flag), the property lookup fails and we need to go to the host to complete the lookup. Unfortunately, when this happens in Edge, we mark the property lookup as potentially having side effects and abort asm.js validation (as side effecting operations may cause observable semantic differences when running in js vs. asm.js).

This change prevents doing a property lookup while linking unless the property is actually used.

### DIFF
--- a/lib/Runtime/Language/AsmJsLink.cpp
+++ b/lib/Runtime/Language/AsmJsLink.cpp
@@ -94,7 +94,7 @@ namespace Js{
         }
 
         Js::JavascriptLibrary* library = scriptContext->GetLibrary();
-        if (mathBuiltinUsed.Test(AsmJSMathBuiltinFunction::AsmJSMathBuiltin_infinity))
+        if (mathBuiltinUsed.TestAndClear(AsmJSMathBuiltinFunction::AsmJSMathBuiltin_infinity))
         {
             Var asmInfinityObj = JavascriptOperators::OP_GetProperty(stdlib, PropertyIds::Infinity, scriptContext);
             if (!JavascriptConversion::SameValue(asmInfinityObj, library->GetPositiveInfinite()))
@@ -103,7 +103,7 @@ namespace Js{
                 return false;
             }
         }
-        if (mathBuiltinUsed.Test(AsmJSMathBuiltinFunction::AsmJSMathBuiltin_nan))
+        if (mathBuiltinUsed.TestAndClear(AsmJSMathBuiltinFunction::AsmJSMathBuiltin_nan))
         {
             Var asmNaNObj = JavascriptOperators::OP_GetProperty(stdlib, PropertyIds::NaN, scriptContext);
             if (!JavascriptConversion::SameValue(asmNaNObj, library->GetNaN()))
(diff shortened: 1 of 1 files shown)


MikeHolman:
@abchatra, @Cellule: can you please take a look?

lib/Runtime/Language/AsmJsLink.cpp
```diff
@@ -143,18 +146,21 @@ namespace Js{
         }
 
 #ifdef ENABLE_SIMDJS
-        Var asmSimdObject = JavascriptOperators::OP_GetProperty(stdlib, PropertyIds::SIMD, scriptContext);
-        for (int i = 0; i < AsmJsSIMDBuiltinFunction::AsmJsSIMDBuiltin_COUNT; i++)
+        if (!simdBuiltinUsed.IsAllClear())
         {
```
abchatra: Change looks good. Just curious, 

>>Since SIMD is not on the global object (this is behind a flag), the property lookup fails and we need to go to the host to complete the lookup. Unfortunately, when this happens in Edge, we mark the property lookup as potentially having side effects and abort asm.js validation (as side effecting operations may cause observable semantic >> differences when running in js vs. asm.js).

So every asm.js validation failed without this change?

lib/Runtime/Language/AsmJsLink.cpp
```diff
@@ -143,18 +146,21 @@ namespace Js{
         }
 
 #ifdef ENABLE_SIMDJS
-        Var asmSimdObject = JavascriptOperators::OP_GetProperty(stdlib, PropertyIds::SIMD, scriptContext);
```
Cellule: So if I understand correctly, the issue is here where we get the SIMD even if none are used right ?

lib/Runtime/Language/AsmJsLink.cpp
```diff
@@ -112,19 +112,22 @@ namespace Js{
...
             {
-                continue;
-            }
-            AsmJSMathBuiltinFunction mathBuiltinFunc = (AsmJSMathBuiltinFunction)i;
-            if (!CheckMathLibraryMethod(scriptContext, asmMathObject, mathBuiltinFunc))
-            {
-                AsmJSCompiler::OutputError(scriptContext, _u("Asm.js Runtime Error : Math builtin function is invalid"));
-                return false;
+                //check if bit is set
+                if (!mathBuiltinUsed.Test(i) || i == AsmJSMathBuiltinFunction::AsmJSMathBuiltin_infinity || i == AsmJSMathBuiltinFunction::AsmJSMathBuiltin_nan)
```
Cellule: Since you've previously cleared nan and infinity, shouldn't `mathBuiltinUsed.Test(i)` always be false for those 2, thus not requiring to check explicitly for them

lib/Runtime/Language/AsmJsLink.cpp
MikeHolman: In Edge if the global object is used for stdlib, yes. Emscripten doesn't do this, so none of our benchmarks/large browser tests ran into this.

lib/Runtime/Language/AsmJsLink.cpp
MikeHolman: Exactly.

lib/Runtime/Language/AsmJsLink.cpp
MikeHolman: Good catch, didn't notice we we're doing this.

Cellule:
LGTM

MikeHolman:

PR #1564, opened 2016-09-09
Create PlatformAgnostic::Event

xanlpz:
Largely based on the existig lib/Common/Common/Event.\* class, but
moved to PlatformAgnostic for a future Linux/Unix version. Changed
Jobs and Recycler to use the new API, and took the opportunity to do
some slight cleanups (e.g. avoid the constant checking of event !=
nullptr, etc.).

Remove the Event API declarations from PAL, since they should be
unused now and were unimplemented anyway.

### DIFF
--- a/lib/Common/Memory/Recycler.cpp
+++ b/lib/Common/Memory/Recycler.cpp
@@ -3,9 +3,6 @@
 // Licensed under the MIT license. See LICENSE.txt file in the project root for full license information.
 //-------------------------------------------------------------------------------------------------------
 #include "CommonMemoryPch.h"
-#if ENABLE_CONCURRENT_GC
-#include <process.h>
-#endif
 
 #ifdef _M_AMD64
 #include "amd64.h"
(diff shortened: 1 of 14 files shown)


lib/Common/Memory/Recycler.cpp
```diff
@@ -4340,46 +4337,25 @@ Recycler::InitializeConcurrent(JsUtil::ThreadService *threadService)
...
     }
     catch (Js::OutOfMemoryException)
     {
-        Assert(concurrentWorkReadyEvent == nullptr);
-        if (concurrentWorkDoneEvent)
-        {
-            CloseHandle(concurrentWorkDoneEvent);
-            concurrentWorkDoneEvent = nullptr;
-        }
+        Assert(concurrentWorkReadyEvent);
```
jianchun: `Assert(!concurrentWorkReadyEvent)`?

xanlpz: Yep, good catch!

obastemur:
@xanlpz Closing this for now. The code base is largely changed and we had some of this stuff covered during the JIT implementation. Thanks for the effort!

PR #1565, opened 2016-09-09
Bug fix when proxy is from different context

kunalspathak:
We set `target` only if proxy is from same context, but while asserting we were not taking cross-context into account and hitting assert.


curtisman:
LGTM

Yongqu:
:shipit:

PR #1567, opened 2016-09-10
Fix unsafe lowering of fast path for String.charAt

pleath:
We optimistically load a cached value into the result operand. But if the cached value is null, we'll have to call a helper to complete the operation, and if the result operand and string source have the same symbol, then we have overwritten the string we pass to the helper with null. Use a temp in such a case and copy it to the result operand only if it is non-null.


rajatd:
:shipit:

PR #1568, opened 2016-09-12
cross-platform: Embed ICU statically

obastemur:
Download ICU 57 [ if user accepts the license ] and build
Link to ChakraCore statically

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -14,6 +14,13 @@ SAFE_RUN() {
     echo $SF_RETURN_VALUE
 }
 
+ERROR_EXIT() {
+    if [[ $? != 0 ]]; then
+        echo $($1 2>&1)
+        exit 1;
+    fi
+}
+
 PRINT_USAGE() {
     echo ""
     echo "[ChakraCore Build Script Help]"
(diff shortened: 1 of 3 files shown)


build.sh
```diff
@@ -83,6 +105,47 @@ while [[ $# -gt 0 ]]; do
         BUILD_TYPE="Debug"
         ;;
 
+    --embed-icu)
+        if [ ! -d "${CHAKRACORE_DIR}/deps/icu/source/output" ]; then
+            ICU_URL="http://source.icu-project.org/repos/icu/icu/tags/release-57-1"
+            echo "----------------------------------------------------------------"
```
dapont: Can we replace with the following:

----------------------------------------------------------------
This script will download ICU-LIB from:
http://source.icu-project.org/repos/icu/icu/tags/release-57-1
It is licensed to you by its publisher, not Microsoft.
Microsoft is not responsible for the software.
Your installation and use of ICU-LIB is subject to the publisher’s terms available here: 
http://www.unicode.org/copyright.html#License
----------------------------------------------------------------
If you don't agree, press Ctrl+C to terminate
Hit ENTER to continue (or wait 10 seconds)

obastemur:
@dapont @digitalinfinity Thanks for the review. I did update the text accordingly.

PR #1570, opened 2016-09-12
Removed `Output::Print()`

kunalspathak:
The trace is only printed if host has `idleGC` implemented, so haven't noticed so far in `ch.exe` but was showing up in node+chakracore after my idleGC changes.

This will fix  https://github.com/nodejs/node-chakracore/issues/115


kunalspathak:
@digitalinfinity, can you take a look?

digitalinfinity:
LGTM

PR #1571, opened 2016-09-12
Fix assertion on recovering the protection of unusable pages when failed to add it to block map (8398465)

leirocks:


leirocks:
@digitalinfinity @curtisman can you review.

PR #1574, opened 2016-09-13
Invertloop issue.

akroshg:
When we invertloop in the bytecode gen time, we didn't copy the call node correctly. That triggerred the assert. Fixed the copynode code for the call node.


akroshg:
@LouisLaf - please have a look at this.

LouisLaf:
:shipit:

PR #1575, opened 2016-09-13
Revise code around record/replay of ScriptContext enter/exit.

mrkmarron:
- Refactored code to ensure script context enter/exit is captured accurately in record.
- Refactored replay code to ensure we enter/exit contexts exactly as in record.
- Updated exception propagation code to ensure replay fidelity and needed interaction with time-travel operations.


mrkmarron:
@agarwal-sandeep please take a look.

lib/Jsrt/Jsrt.cpp
```diff
@@ -2152,19 +2146,16 @@ CHAKRA_API JsCallFunction(_In_ JsValueRef function, _In_reads_(cargs) JsValueRef
         Js::Arguments jsArgs(callInfo, reinterpret_cast<Js::Var *>(args));
 
 #if ENABLE_TTD
-        double ttdStartTime = -1.0;
-
+        TTD::TTDJsRTActionResultAutoRecorder _actionEntryPopper(TTD::ContextWrapperEnterExitStatus::EnterContextAPIWrapper);
         if(PERFORM_JSRT_TTD_RECORD_ACTION_CHECK(scriptContext))
```
agarwal-sandeep: Should this be inside the if?

lib/Jsrt/Jsrt.cpp
```diff
@@ -3491,36 +3468,44 @@ CHAKRA_API JsTTDStartTimeTravelRecording()
...
-        if (threadContext->TTDLog == nullptr)
-        {
-            AssertMsg(false, "Need to create in TTD mode.");
-            return JsErrorCategoryUsage;
-        }
+    JsrtContext *currentContext = JsrtContext::GetCurrent();
+    JsErrorCode cCheck = CheckContext(currentContext, true);
+    if(cCheck != JsNoError)
+    {
+        AssertMsg(false, "This shouldn't happen!!!");
```
agarwal-sandeep: >AssertMsg(false, "This shouldn't happen!!!"); [](start = 8, length = 45)

nit: Seems like not a very useful assert given that this is external API and can be called when this is supposed to fail like when context is not set.

agarwal-sandeep:
:clock1:

Issue #1496, opened 2016-08-26
ES6 constructor returns class instead of object


eebbing:
The code example below should log the numbers from 0 up to 500. Instead, the code breaks after about 100 iterations. The issue is that after about 100 iterations the constructor of class `B` returns the class itself instead of an instance of `B`. 

The fact that this issue only occurs after multiple iterations suggests to me that the bug is somewhere in the JIT engine.

By the way, the error disappears if the method body of `A.toB()` is replaced by `return new B(this);`. 

I tested the following code in Microsoft Edge 38.14393.0.0 (Microsoft EdgeHTML 14.14393)

	var Test = {};

	class A {
		constructor(foo) { this.foo = foo; }
		toB() { return new Test.B(this); }
	}

	class B {
		constructor(bar) { this.bar = bar; }
	}
	
	Test.B = B;
	
	for (let i=0; i<500; i++)
	{
		const a = new A(i);
		const b = a.toB();

		try
		{
			console.log(b.bar.foo);
		}
		catch (e)
		{
			console.log(e); // e.description -> "Unable to get property 'foo' of undefined or null reference"
			console.log(b); // b -> class B { ... }
			break;
		}
	}

eebbing:
For what it's worth, I decided to look a little bit further into it, and was able to reproduce the error with the latest source code. I ran the script below via `ch.exe` (x64 debug build) with the following arguments:

    ch.exe -forceNative -MaxinterpretCount:10 -MaxSimpleJITRunCount:10 E:\path-to\test.js

The script is:

	var Test = {};

	class A {
		constructor(foo) { this.foo = foo; }
		toB() { return new Test.B(this); }
	}

	class B {
		constructor(bar) { this.bar = bar; }
	}

	Test.B = B;

	for (let i=0; i<20; i++)
	{
		const a = new A(i);
		const b = a.toB();

		try
		{
		WScript.Echo(b.bar.foo);        
		}
		catch (e)
		{
			WScript.Echo(e);
			WScript.Echo(b);
			break;
		}
	}

The output of the script is:

	0
	1
	2
	3
	4
	5
	6
	7
	8
	9
	TypeError: Unable to get property 'foo' of undefined or null reference
	class B {
		constructor(bar) { this.bar = bar; }
	}

ianwjhalliday:
Awesome! Thanks @eebbing, having a simple repro that we can run with `ch.exe` makes it a lot easier to debug.  We'll look into this soon.

[...]

PR #1577, opened 2016-09-13
ES6 constructor returns class instead of object upon constructorCache hit.

suwc:
ES6 constructor returns class instead of object upon constructorCache hit.
constructorCache->ctorHasNoExplicitReturnValue set to 'true' for ES6 class
constructors, therefore opcode 'GetNewScObject' is optimized away in
inliner. Fix by de-asserting Flags_HasNoExplicitReturnValue for ES6 class
constructors.


suwc:
https://github.com/Microsoft/ChakraCore/issues/1496

suwc:
@Microsoft/chakra-es @eebbing @Krovatkin could you take a look pls?

lib/Backend/GlobOptFields.cpp
```diff
@@ -2700,7 +2700,7 @@ GlobOpt::OptNewScObject(IR::Instr** instrPtr, Value* srcVal)
         instr->m_func->GetConstructorCache(static_cast<Js::ProfileId>(instr->AsProfiledInstr()->u.profileId)) : nullptr;
 
     Assert(ctorCache == nullptr || srcVal->GetValueInfo()->IsVarConstant() && Js::JavascriptFunction::Is(srcVal->GetValueInfo()->AsVarConstant()->VarValue()));
-    Assert(ctorCache == nullptr || !ctorCache->typeIsFinal || ctorCache->ctorHasNoExplicitReturnValue);
+    Assert(ctorCache == nullptr || !ctorCache->typeIsFinal || ctorCache->ctorHasNoExplicitReturnValue || isCtorInlined);
```
rajatd: Shouldn't this be a check for a class constructor? We emit NewScObjectNoCtor in cases other than class ctor too.

lib/Backend/Lower.cpp
```diff
@@ -4577,7 +4577,7 @@ bool Lowerer::TryLowerNewScObjectWithFixedCtorCache(IR::Instr* newObjInstr, IR::
         ctorCache = newObjInstr->m_func->GetConstructorCache(static_cast<Js::ProfileId>(newObjInstr->AsProfiledInstr()->u.profileId));
         Assert(ctorCache != nullptr);
         Assert(!ctorCache->skipNewScObject);
-        Assert(!ctorCache->typeIsFinal || ctorCache->ctorHasNoExplicitReturnValue);
+        Assert(!ctorCache->typeIsFinal || ctorCache->ctorHasNoExplicitReturnValue || newObjInstr->m_opcode == Js::OpCode::NewScObjectNoCtor);
```
rajatd: >newObjInstr->m_opcode == Js::OpCode::NewScObjectNoCtor) [](start = 85, length = 55)

same as in GloboptFields.cpp

lib/Runtime/ByteCode/ByteCodeEmitter.cpp
```diff
@@ -2727,6 +2727,11 @@ void ByteCodeGenerator::EmitFunctionBody(FuncInfo *funcInfo)
     }
     Assert(!pnode->CapturesSyms());
     EmitTopLevelStatement(pnode, funcInfo, false);
+
+    if (funcInfo->IsClassConstructor())
+    {
+        funcInfo->GetParsedFunctionBody()->SetHasNoExplicitReturnValue(false);
```
ianwjhalliday: Is this the right place to set this flag?  Seems like this is something that could be set earlier when the function body object is created.

suwc: Possibly. Note other calls to SetHasNoExplicitReturnValue() won't happen till inside EmitTopLevelStatement(). I think it's more consistent to set the flag here.

---
In reply to: [78957170](https://github.com/Microsoft/ChakraCore/pull/1577#discussion_r78957170) [](ancestors = 78957170)

suwc:
Any other comments or questions?

test/es6/rlexe.xml
```diff
@@ -1331,4 +1331,12 @@
     <compile-flags>-force:deferparse -args summary -endargs</compile-flags>
   </default>
 </test>
+<test>
+  <default>
+    <files>bug_issue_1496.js</files>
+    <compile-flags>-mic:1 -maxsimplejitruncount:2</compile-flags>
```
tcare: Why does dynapogo not work here? Because of the simplejit runcount?

suwc: No, dynapogo works for the test case, but it didn't repro the bug in the first place. And given the flags running dynapogo here has little value.

tcare: Looks like other tests that include -mic:1 allow dynapogo. I don't see the harm in having it.

tcare:
Looks good

test/es6/rlexe.xml
suwc: No harm - it's value vs. (a little extra) run time.
Let's get 2nd opinion.
@ianwjhalliday, what do you think?

lib/Runtime/ByteCode/ByteCodeEmitter.cpp
pleath: I'm guessing that other cases set the flag when they visit the return statement node. I think for class ctors we should either do as Ian suggests and set the flag in StartBindFunction, which is a friendlier place, or have the parser emit an explicit return node.

test/es6/rlexe.xml
pleath: Let's have dynapogo run the test. More aggressive jitting in such a case can expose issues with inlining, etc., down the road.

PR #1578, opened 2016-09-14
CheckIfTypeIsEquivalent improvement

kunalspathak:
* Always emit a bailout instead of helper if Equivalent Type check fails
* If equivalent type check is full, start evicting non-shared types.

### DIFF
--- a/lib/Backend/BackwardPass.cpp
+++ b/lib/Backend/BackwardPass.cpp
@@ -2107,6 +2107,13 @@ BackwardPass::DeadStoreTypeCheckBailOut(IR::Instr * instr)
         return;
     }
 
+    // If bailOutKind is equivTypeCheck then leave alone the bailout
+    if (bailOutKind == IR::BailOutFailedEquivalentTypeCheck ||
+        bailOutKind == IR::BailOutFailedEquivalentFixedFieldTypeCheck)
+    {
+        return;
+    }
+
     // We're not checking for polymorphism, so don't let the bailout indicate that we
     // detected polymorphism.
     instr->GetBailOutInfo()->polymorphicCacheIndex = (uint)-1;
--- a/lib/Backend/Lower.cpp
+++ b/lib/Backend/Lower.cpp
@@ -6825,7 +6825,7 @@ Lowerer::GenerateStFldWithCachedType(IR::Instr *instrStFld, bool* continueAsHelp
 
     if (hasTypeCheckBailout)
     {
-        AssertMsg(PHASE_ON1(Js::ObjTypeSpecIsolatedFldOpsWithBailOutPhase) || !propertySymOpnd->IsTypeDead(),
+        AssertMsg(PHASE_ON1(Js::ObjTypeSpecIsolatedFldOpsWithBailOutPhase) || !propertySymOpnd->IsTypeDead(), 
             "Why does a field store have a type check bailout, if its type is dead?");
 
         if (instrStFld->GetBailOutInfo()->bailOutInstr != instrStFld)
--- a/lib/Runtime/Language/JavascriptOperators.cpp
+++ b/lib/Runtime/Language/JavascriptOperators.cpp
@@ -8243,6 +8243,19 @@ CommonNumber:
             return true;
         }
 
+        // Since the equivTypes cache is always compacted, if last entry is non-nullptr, it means the cache is full.
+        DynamicType * dynamicType = (type && DynamicType::Is(type->GetTypeId())) ? static_cast<DynamicType*>(type) : nullptr;
+        bool isEquivTypesCacheFull = equivTypes[EQUIVALENT_TYPE_CACHE_SIZE - 1] != nullptr;
+        if (dynamicType != nullptr)
+        {
+            if (!dynamicType->GetIsShared() && isEquivTypesCacheFull)
+            {
+                // If this is non-shared type and we don't have space to cache it, just consider
+                // it as non-equivalent so we will eventually bailout
+                return false;
+            }
+        }
+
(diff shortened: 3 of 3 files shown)


kunalspathak:
@pleath , @LouisLaf  , please review!

lib/Backend/BackwardPass.cpp
```diff
@@ -2107,6 +2107,16 @@ BackwardPass::DeadStoreTypeCheckBailOut(IR::Instr * instr)
         return;
     }
 
+    // If oldBailoutKind is equivTypeCheck and appropriate opnd doesn't have initial/final type
+    // then leave alone the bailout
+    if (oldBailOutKind == IR::BailOutFailedEquivalentTypeCheck || oldBailOutKind == IR::BailOutFailedEquivalentFixedFieldTypeCheck)
+    {
+        if (!(propertySymOpnd->HasInitialType() || propertySymOpnd->HasFinalType()))
```
pleath: Is this part necessary? If we're really emitting an equivalent bailout, then final type optimization isn't involved, right?

lib/Runtime/Language/JavascriptOperators.cpp
```diff
@@ -8243,6 +8243,20 @@ namespace Js
             return true;
         }
 
+        // Since the equivTypes cache is always compacted, if last entry is non-nullptr, it means
+        // the cache is full.
+        DynamicType * dynamicType = DynamicType::Is(type->GetTypeId()) ? static_cast<DynamicType*>(type) : nullptr;
+        bool isEquivTypesCacheIsFull = equivTypes[EQUIVALENT_TYPE_CACHE_SIZE - 1] != nullptr;
```
pleath: Nit: remove one "is" from the name?

lib/Backend/BackwardPass.cpp
kunalspathak: That's what I thought, but we do end up here with non-null initialType and finalType and if I don't have this, we hit assert in Lower where that if it is not equivTypeCheck then there shouldn't be equiv bailouts.
```c++
Assert(doEquivTypeCheck || !instrChk->HasEquivalentTypeCheckBailOut());
```

lib/Runtime/Language/JavascriptOperators.cpp
kunalspathak: Ah yes. Will do.

lib/Backend/BackwardPass.cpp
pleath: If that's the case, then can you find out how we got here with an equiv bailout kind?

kunalspathak: Sure.

kunalspathak: So I was mistakenly checking the `bailoutKind` that we have in forward pass instead of checking that for backward pass. With that, I don't need the check for initial/final type.

lib/Backend/Lower.cpp
```diff
@@ -5710,9 +5710,6 @@ Lowerer::GenerateLdFldWithCachedType(IR::Instr * instrLdFld, bool* continueAsHel
 
     if (hasTypeCheckBailout)
     {
-        AssertMsg(PHASE_ON1(Js::ObjTypeSpecIsolatedFldOpsWithBailOutPhase) || !propertySymOpnd->IsTypeDead(),
```
pleath: Can you leave the asserts but assert that either the type is dead or the bailout kind is equivalent?

kunalspathak: Turned out that it is fine to leave assert here and the one below because we will only come this code path if there is no typeCheckBailout.
```c++
if(hasTypeCheckBailout) {

```

lib/Backend/BackwardPass.cpp
```diff
@@ -2073,8 +2073,8 @@ BackwardPass::DeadStoreTypeCheckBailOut(IR::Instr * instr)
             preOpBailOutInstrToProcess = nullptr;
         }
         return;
-    }
-
+    } 
+
```
dilijev: nit: trailing whitespace

lib/Backend/Lower.cpp
```diff
@@ -6825,8 +6824,7 @@ Lowerer::GenerateStFldWithCachedType(IR::Instr *instrStFld, bool* continueAsHelp
...
-            "Why does a field store have a type check bailout, if its type is dead?");
+        AssertMsg(PHASE_ON1(Js::ObjTypeSpecIsolatedFldOpsWithBailOutPhase) || !propertySymOpnd->IsTypeDead(), "Why does a field load have a type check bailout, if its type is dead?");
```
dilijev: In this line and above, it looks like there's no change. Can you restore to the original version? (splitting into 2 lines will be slightly easier to read)

lib/Runtime/Language/JavascriptOperators.cpp
```diff
@@ -8243,6 +8243,20 @@ namespace Js
             return true;
         }
 
+        // Since the equivTypes cache is always compacted, if last entry is non-nullptr, it means
+        // the cache is full.
```
dilijev: nit: no need to break this comment to multiple lines since the code following is longer.

lib/Backend/Lower.cpp
dilijev: Correction -- the one above says "load" and this one originally said "store" but now says "load" -- was that change intentional?

---
In reply to: [79080126](https://github.com/Microsoft/ChakraCore/pull/1578#discussion_r79080126) [](ancestors = 79080126)

PR #1579, opened 2016-09-14
Regex backtracking improvements

rajatd:
This PR improves the algorithm for backtracking through a set with a * or a + wildcard. 

When we get an unsuccessul match after such a set, we used to backtrack one character at a time over the matched characters of the set. The optimization being made in this change is that if we know at compile time what needs to follow the set for a successful match, we could record the offsets of that (those) character(s) while matching the set, and then backtrack only to those offsets.


rajatd:
@microsoft/chakra-es, @abchatra , could you please review this?
cc: @LouisLaf @curtisman

suwc:
@Microsoft/chakra-es

rajatd:
Call for review, @curtisman 
FYI, @dilijev

tcare:
Ping @Microsoft/chakra-es on behalf of Rajat

lib/Parser/RegexCompileTime.cpp
```diff
@@ -3460,6 +3460,12 @@ namespace UnifiedRegex
         innerFollow->UnionInPlace(compiler.ctAllocator, *accumFollow);
         innerFollow->UnionInPlace(compiler.ctAllocator, *body->firstSet);
 
+        if (followSet->IsSingleton())
```
akroshg: is it possible that followSet is empty?

akroshg: I meant to comment on the GetCompactChar line.

---
In reply to: [81023203](https://github.com/Microsoft/ChakraCore/pull/1579#discussion_r81023203) [](ancestors = 81023203)

rajatd: If it is singleton, it means that it has 1 element.

---
In reply to: [81023296](https://github.com/Microsoft/ChakraCore/pull/1579#discussion_r81023296) [](ancestors = 81023296,81023203)

lib/Parser/RegexRuntime.cpp
```diff
@@ -3052,6 +3052,83 @@ namespace UnifiedRegex
...
+            : inputOffset + static_cast<CharCount>(repeatsUpper);
+#if ENABLE_REGEX_CONFIG_OPTIONS
+        matcher.CompStats();
+#endif
+        while (inputOffset < inputEndOffset && matchSet.Get(input[inputOffset]))
+        {
+#if ENABLE_REGEX_CONFIG_OPTIONS
+            matcher.CompStats();
+#endif
+            if (!PHASE_OFF1(Js::RegexOptBTPhase) && input[inputOffset] == this->followFirst)
```
akroshg: should we move this phasecheck outside of the loop? also why do we check this here, why not have phase detection at the RegexCompileTime while we create the LoopSetWithFollowFirstInst instance?

lib/Parser/RegexRuntime.cpp
```diff
@@ -3052,6 +3052,83 @@ namespace UnifiedRegex
...
+            {
+                loopInfo->EnsureOffsetsOfFollowFirst(matcher);
```
akroshg: >EnsureOffsetsOfFollowFirst [](start = 26, length = 26)

nit : we could move the EnsureOffsetsOfFollowFirst outside of the loop. considering this loop could be hot.

akroshg:
few nits, overall looks good to me.

lib/Parser/RegexRuntime.cpp
rajatd: Yes, makes sense to do the phase check at compile time.

lib/Parser/RegexRuntime.cpp
rajatd: I want to create the offsetsOfFollowFirst list only if I get at least one match of the followFirst. And, I think it is small enough that it'll be inlined

---
In reply to: [81027735](https://github.com/Microsoft/ChakraCore/pull/1579#discussion_r81027735) [](ancestors = 81027735)

PR #1581, opened 2016-09-14
Minor false positive fixes in ChakraCore

digitalinfinity:
1. BitVectors were not allocated as leaf, and this could cause the GC to interpret it's contents as pointer references. Switched to leaf
2. Temporary guest-arenas that were being cached could still get scanned, even if they weren't actively being used. Added a mechanism to unregister said arena when it wasn't used.

### DIFF
--- a/lib/Backend/NativeCodeData.h
+++ b/lib/Backend/NativeCodeData.h
@@ -38,6 +38,8 @@ public:
 
         char * Alloc(DECLSPEC_GUARD_OVERFLOW size_t requestedBytes);
         char * AllocZero(DECLSPEC_GUARD_OVERFLOW size_t requestedBytes);
+        char * AllocLeaf(__declspec(guard(overflow)) size_t requestedBytes);
+
         NativeCodeData * Finalize();
         void Free(void * buffer, size_t byteSize);
 
--- a/lib/Runtime/Base/TempArenaAllocatorObject.cpp
+++ b/lib/Runtime/Base/TempArenaAllocatorObject.cpp
@@ -43,6 +43,31 @@ namespace Js
         Assert(allocator.AllocatedSize() == 0);
     }
 
+    template <bool isGuestArena>
+    void TempArenaAllocatorWrapper<isGuestArena>::AdviseInUse()
+    {
+        if (isGuestArena)
+        {
+            if (externalGuestArenaRef == nullptr)
+            {
+                externalGuestArenaRef = this->recycler->RegisterExternalGuestArena(this->GetAllocator());
+            }
+        }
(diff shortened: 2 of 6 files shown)


lib/Runtime/Base/TempArenaAllocatorObject.cpp
```diff
@@ -43,6 +43,31 @@ namespace Js
         Assert(allocator.AllocatedSize() == 0);
     }
 
+    template <bool isGuestArena>
+    void TempArenaAllocatorWrapper<isGuestArena>::AdviseInUse()
+    {
+        if (isGuestArena)
+        {
+            if (externalGuestArenaRef == nullptr)
```
curtisman: If we assert that externalGuestArenaRef != nullptr in

curtisman: If we assert that externalGuestArenaRef != nullptr in AdviseNotInUse, should this be an assert externalGuestArenaRef == nullptr too?

lib/Runtime/Base/TempArenaAllocatorObject.h
```diff
@@ -18,6 +18,8 @@ namespace Js
 
     public:
 
+        void AdviseInUse();
+        void AdviseNotInUse();
```
dilijev: Were the implementations of these already present?

digitalinfinity: No? Newly added

dilijev: Derp; I didn't see them above

lib/Backend/NativeCodeData.h
```diff
@@ -38,6 +38,8 @@ class NativeCodeData
 
         char * Alloc(DECLSPEC_GUARD_OVERFLOW size_t requestedBytes);
         char * AllocZero(DECLSPEC_GUARD_OVERFLOW size_t requestedBytes);
+        char * AllocLeaf(__declspec(guard(overflow)) size_t requestedBytes);
```
jianchun: @digitalinfinity Strange, why CI did not complain on this? Should use DECLSPEC_GUARD_OVERFLOW

digitalinfinity: derp- I don't know? @dilijev any ideas? @jainchun, i'll make the change- sorry, had ported the change from Threshold and didn't notice the macro conversion 😞

dilijev: JIT is still disabled on Linux so this whole block is probably compiled out on Linux

PR #1584, opened 2016-09-15
Fix legacy builds

obastemur:
VS2013 doesn't support uint32_t as is. Replacing it to ~~uint32_t~~ unsigned
@tcare Thanks for figuring this out and review please.

### DIFF
--- a/bin/ch/WScriptJsrt.cpp
+++ b/bin/ch/WScriptJsrt.cpp
@@ -725,7 +725,7 @@ bool WScriptJsrt::InstallObjectsOnObject(JsValueRef object, const char* name, Js
     memcpy(path, msg, (size_t)str_len);          \
     path[str_len] = char(0)
 
-void GetBinaryLocation(char *path, const uint32_t size)
+void GetBinaryLocation(char *path, const unsigned size)
 {
     AssertMsg(size >= 512 && path != nullptr, "Min path buffer size 512 and path can not be nullptr");
     AssertMsg(size < INT_MAX, "Isn't it too big for a path buffer?");
@@ -754,13 +754,13 @@ void GetBinaryLocation(char *path, const uint32_t size)
         return;
     }
 
-    if ((uint32_t)str_len > size - 1)
+    if ((unsigned)str_len > size - 1)
     {
         str_len = (int) size - 1;
     }
     path[str_len] = char(0);
 #elif defined(__APPLE__)
-    uint32_t path_size = size;
+    uint32_t path_size = (uint32_t)size;
     char *tmp = nullptr;
     int str_len;
     if (_NSGetExecutablePath(path, &path_size))


dilijev:

bin/ch/WScriptJsrt.cpp
```diff
@@ -725,7 +725,7 @@ bool WScriptJsrt::InstallObjectsOnObject(JsValueRef object, const char* name, Js
     memcpy(path, msg, (size_t)str_len);          \
     path[str_len] = char(0)
 
-void GetBinaryLocation(char *path, const uint32_t size)
+void GetBinaryLocation(char *path, const unsigned size)
```
dilijev: Can we be explicit and say `unsigned int`? Is there no aliased type macro for this?

bin/ch/WScriptJsrt.cpp
```diff
@@ -754,13 +754,13 @@ void GetBinaryLocation(char *path, const uint32_t size)
...
 
-    if ((uint32_t)str_len > size - 1)
+    if ((unsigned)str_len > size - 1)
     {
         str_len = (int) size - 1;
     }
     path[str_len] = char(0);
 #elif defined(__APPLE__)
-    uint32_t path_size = size;
+    uint32_t path_size = (uint32_t)size;
```
dilijev: Ew. But okay. Is this really the best option?

dilijev:
I'm wondering if any of the other occurrences of `uint32_t` would be a problem?

```
$ git grep uint32_t
bin/ch/WScriptJsrt.cpp:728:void GetBinaryLocation(char *path, const uint32_t size)
bin/ch/WScriptJsrt.cpp:757:    if ((uint32_t)str_len > size - 1)
bin/ch/WScriptJsrt.cpp:763:    uint32_t path_size = size;
lib/Common/CommonPal.h:418:#define MAXUINT32   ((uint32_t)~((uint32_t)0))
lib/Jsrt/ChakraCommon.h:83:typedef uint32_t UINT32;
lib/Runtime/Language/AsmJsTypes.h:29:    typedef uint32 uint32_t;
lib/Runtime/Language/AsmJsTypes.h:594:            return (uint32_t)( reg - mNbConst ) < (uint32_t)( mRegisterCount - mNbConst );
pal/inc/pal_mstypes.h:223:typedef unsigned __int32 uint32_t;
pal/src/init/pal.cpp:1373:    uint32_t bufsize = sizeof(exec_path);
```

bin/ch/WScriptJsrt.cpp
tcare: DWORD? @digitalinfinity opinion?

tcare: Preferably since we pass a size_t in I'd prefer that but this may be easier.

obastemur:
@dilijev @tcare Thanks for the review

bin/ch/WScriptJsrt.cpp
obastemur: Actually it could be even `int` though. It is `unsigned` due to being `size` but not `index`. What this function does; I don't see any advantage of using `unsigned int` explicitly. @tcare already pointed out uint64_t part.

dilijev: `unsigned int` is just more clear, but since it's equivalent, there's no actual problem or advantage/disadvantage from a functionality perspective. We can table that.

I'm more concerned about the size_t versus unsigned and the bitness of the type of this param and the value passed in on various systems.

Since the value passed in is a size_t what happens if that value is larger than 32 bits?

obastemur: @dilijev  what do you mean by `Since the value passed in is a size_t` ? You mean `sizeof` returns `size_t` ?

Reminder, that param is used only for max buffer size that is actually filled in by a system call.

dilijev: Yes, sizeof returns a size_t. If the buffer size is guaranteed to be 32 bits or smaller on all platforms then there's no problem.

obastemur: >Yes, sizeof returns a size_t.

So, let's hope that we don't have a variable / stack array that is actually bigger than `4,294,967,295`

dilijev: I'm all for "hope" but I've been paranoid these days about what happens when we make wrong assumptions :P (@digitalinfinity will probably agree)

obastemur: No worries @dilijev , luckily it is `unsigned` ;) Unsigned_MAX_Value + 1 --> 0 .. So we wouldn't use that buffer. Still, why we would ever have/need that big buffer for a file path / or how the system call itself can handle that nice piece of memory / is yet another PR discussion indeed :) Kudos to @digitalinfinity

PR #1585, opened 2016-09-15
Redefer function bodies

pleath:
Redefer function bodies that are not currently being executed and are eligible for deferred parsing (e.g., not arrow functions, not functions-in-block). Define a 'force' mode in which all eligible functions are redeferred on GC, as well as a 'stress' mode in which all candidates are redeferred on each stack probe.


pleath:
Merged with OOP JIT. That was fun. :)

pleath:
Fully synched.

lib/Backend/BailOut.cpp
```diff
@@ -1763,7 +1763,7 @@ void BailOutRecord::ScheduleFunctionCodeGen(Js::ScriptFunction * function, Js::S
     bailOutRecordNotConst->bailOutCount++;
 
     Js::FunctionEntryPointInfo *entryPointInfo = function->GetFunctionEntryPointInfo();
-    uint8 callsCount = entryPointInfo->callsCount;
+    uint32 callsCount = entryPointInfo->callsCount;
```
rajatd: As @LouisLaf mentioned in #1757, we should have callsCount be MIN of entryPointInfo->callsCount and 255

pleath:
Reverted to treating call counts as saturating uint8's in bailout code. (Thanks, @rajatd.)

lib/Runtime/Base/ThreadContext.h
```diff
@@ -650,6 +652,22 @@ class ThreadContext sealed :
...
+    {
+        InitialRedeferralState,
+        StartupRedeferralState,
+        MainRedeferralState
+    };
+    RedeferralState redeferralState;
+    uint gcSinceLastRedeferral;
+    uint gcSinceCallCountsCollected;
+
+    static const uint InitialRedeferralDelay = 5;
```
rajatd: >InitialRedeferralDelay [](start = 22, length = 22)

should we have these as configurable values so we can tune redeferral in the future if need be?

pleath: That's a good idea.

lib/Runtime/Base/ThreadContext.cpp
```diff
@@ -2062,19 +2080,6 @@ ThreadContext::ExecuteRecyclerCollectionFunction(Recycler * recycler, Collection
 
     BOOL ret = FALSE;
```
rajatd: >RedeferFunctionBodies [](start = 23, length = 21)

Correct me if I'm wrong, but I think we should also pass the result of this->GetRedeferralalCollectionInterval() to ScriptContext::RedeferFunctionBodies. Right now, the result of ThreadContext::GetRedeferralInactiveThreshold is being used in FunctionBody::DoRedeferFunction to determine if a function has remained inactive for long enough that it could be redeferred.

lib/Runtime/Base/FunctionBody.cpp
```diff
@@ -570,6 +577,291 @@ namespace Js
...
+        {
+            return false;
+        }
+
+        if (!PHASE_FORCE(Js::RedeferralPhase, this) && !PHASE_STRESS(Js::RedeferralPhase, this))
+        {
+            uint inactiveCount;
+            auto fn = [&](){ inactiveCount = 0xFFFFFFFF; };
+            inactiveCount = UInt32Math::Mul(this->GetInactiveCount(), this->GetCompileCount(), fn);
+            if (inactiveCount < inactiveThreshold)
```
rajatd: >inactiveCount < inactiveThreshold [](start = 16, length = 33)

The idea here is that if a function was parsed after having been redeferred, make it less likely to be deferred again, right? If so, shouldn't we multiply the threshold by the compileCount?

lib/Runtime/Base/FunctionBody.cpp
```diff
@@ -1705,25 +2085,8 @@ namespace Js
...
-                this->GetNestedCount(),
-                this->GetUtf8SourceInfo(),
-                this->m_functionNumber,
-                this->GetUtf8SourceInfo()->GetSrcInfo()->sourceContextInfo->sourceContextId, /* script id */
-                this->functionId, /* function id */
-                propertyRecordList,
-                (Attributes)(this->GetAttributes() & ~(Attributes::DeferredDeserialize | Attributes::DeferredParse)),
-                Js::FunctionBody::FunctionBodyFlags::Flags_HasNoExplicitReturnValue
-#ifdef PERF_COUNTERS
-                , false /* is function from deferred deserialized proxy */
```
rajatd: Is this assert no longer valid? If so, can we just remove it?

lib/Runtime/Base/FunctionBody.cpp
```diff
@@ -6637,11 +7015,11 @@ namespace Js
...
 
         // Simple JIT counts down and transitions on overflow
-        const uint8 callCount = simpleJitEntryPointInfo->callsCount;
+        const uint32 callCount = simpleJitEntryPointInfo->callsCount;
         Assert(simpleJitLimit == 0 ? callCount == 0 : simpleJitLimit > callCount);
         return callCount == 0 ?
             static_cast<uint16>(simpleJitLimit) :
-            static_cast<uint16>(simpleJitLimit) - callCount - 1;
+            static_cast<uint16>(simpleJitLimit) - static_cast<uint16>(callCount) - 1;
     }
```
rajatd: Is this just a bug fix?

lib/Common/Memory/CustomHeap.cpp
```diff
@@ -251,6 +251,10 @@ Allocation* Heap::Alloc(size_t bytes, ushort pdataCount, ushort xdataSize, bool
         {
             page = AllocNewPage(bucket, canAllocInPreReservedHeapPageSegment, isAnyJittedCode, isAllJITCodeInPreReservedRegion);
         }
+        else if (!canAllocInPreReservedHeapPageSegment && isAnyJittedCode)
+        {
+            *isAllJITCodeInPreReservedRegion = false;
+        }
```
rajatd: not part of your change, right?

rajatd:
LGTM apart from the comments.

pleath:
@dotnet-bot please test Windows x86_release

lib/Runtime/Base/FunctionBody.cpp
pleath: Just silencing any possible complaint about overflow.

lib/Common/Memory/CustomHeap.cpp
pleath: Well, it's a real issue that my change exposed, probably by causing more re-jitting.

PR #1587, opened 2016-09-16
Defer LdLen instruction handling until globopt so that we can leave it as LdFld for non-array-like objects.

Penguinwizzard:
This changset makes us handle situations better where the .length field is used on objects other than arrays/strings/arguments, and makes a few optimizations a little more robust.

### DIFF
--- a/lib/Backend/GlobOpt.cpp
+++ b/lib/Backend/GlobOpt.cpp
@@ -4877,6 +4877,41 @@ GlobOpt::OptInstr(IR::Instr *&instr, bool* isInstrRemoved)
         this->KillStateForGeneratorYield();
     }
 
+    // Change LdFld on arrays, strings, and 'arguments' to LdLen when we're accessing the .length field
+    if (instr->GetSrc1() && instr->GetSrc1()->IsSymOpnd() && instr->m_opcode == Js::OpCode::ProfiledLdFld || instr->m_opcode == Js::OpCode::LdFld || instr->m_opcode == Js::OpCode::ScopedLdFld)
+    {
+        IR::Opnd * opnd = instr->GetSrc1();
+        Sym *sym = opnd->AsSymOpnd()->m_sym;
+        if (sym->IsPropertySym())
+        {
+            PropertySym *originalPropertySym = sym->AsPropertySym();
+            IR::RegOpnd* newopnd = IR::RegOpnd::New(originalPropertySym->m_stackSym, IRType::TyVar, instr->m_func);
+            // only on .length
+            if (this->lengthEquivBv != nullptr && this->lengthEquivBv->Test(originalPropertySym->m_id))
+            {
+                Value *const objectValue = FindValue(originalPropertySym->m_stackSym);
+                // Only for things we'd emit a fast path for
+                if (
+                    objectValue->GetValueInfo()->IsLikelyAnyArray() ||
+                    objectValue->GetValueInfo()->HasHadStringTag() ||
+                    objectValue->GetValueInfo()->IsLikelyString() ||
+                    newopnd->IsArgumentsObject() ||
+                    (this->blockData.argObjSyms && IsArgumentsOpnd(newopnd))
+                   )
+                {
+                    IR::Instr *newinstr = IR::Instr::New(Js::OpCode::LdLen_A, instr->m_func);
+                    instr->TransferTo(newinstr);
+                    newinstr->UnlinkSrc1();
+                    newinstr->SetSrc1(newopnd);
+                    newinstr->GetSrc1()->SetValueType(objectValue->GetValueInfo()->Type());
+                    instr->InsertAfter(newinstr);
+                    instr->Remove();
+                    instr = newinstr;
+                }
+            }
+        }
+    }
+
     // Consider: Do we ever get post-op bailout here, and if so is the FillBailOutInfo call in the right place?
     if (instr->HasBailOutInfo() && !this->IsLoopPrePass())
     {
--- a/lib/Backend/IRBuilder.cpp
+++ b/lib/Backend/IRBuilder.cpp
@@ -6074,7 +6074,16 @@ IRBuilder::BuildProfiledCallI(Js::OpCode opcode, uint32 offset, Js::RegSlot retu
         newOpcode = opcode;
         Js::OpCodeUtil::ConvertNonCallOpToNonProfiled(newOpcode);
         Assert(newOpcode == Js::OpCode::NewScObject || newOpcode == Js::OpCode::NewScObjectSpread);
-        returnType = ValueType::GetObject(ObjectType::UninitializedObject);
+        if (newOpcode == Js::OpCode::NewScObjectSpread || !this->m_func->HasProfileInfo())
+        {
+            // Testing required before enabling this for newscobjectspread
+            returnType = ValueType::GetObject(ObjectType::UninitializedObject);
+        }
+        else
+        {
+            // If we have profile data, make use of it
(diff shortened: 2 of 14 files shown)


lib/Backend/IRBuilder.cpp
```diff
@@ -6074,7 +6074,24 @@ IRBuilder::BuildProfiledCallI(Js::OpCode opcode, uint32 offset, Js::RegSlot retu
...
+            // If we have profile data, make use of it
+            ValueType knownReturnType = this->m_func->GetReadOnlyProfileInfo()->GetReturnType(opcode, profileId);
+            if (knownReturnType != ValueType::GetObject(ObjectType::UninitializedObject))
+            {
+                returnType = knownReturnType;
+            }
+            else
+            {
+                returnType = ValueType::GetObject(ObjectType::UninitializedObject);
+            }
```
rajatd: Can we just have "returnType = knownReturnType" or "returnType = this->m_func->GetReadOnlyProfileInfo()->GetReturnType(opcode, profileId)" instead of the if-else?

lib/Backend/GlobOpt.cpp
```diff
@@ -4877,6 +4877,40 @@ GlobOpt::OptInstr(IR::Instr *&instr, bool* isInstrRemoved)
         this->KillStateForGeneratorYield();
     }
 
+    // Change LdFld on arrays, strings, and 'arguments' to LdLen when we're accessing the .length field
+    if (instr->GetSrc1() && instr->GetSrc1()->GetKind() == IR::OpndKindSym && instr->m_opcode == Js::OpCode::ProfiledLdFld || instr->m_opcode == Js::OpCode::LdFld || instr->m_opcode == Js::OpCode::ScopedLdFld)
```
rajatd: >instr->GetSrc1()->GetKind() == IR::OpndKindSym [](start = 28, length = 46)

Replace this by a call to IsSymOpnd()

lib/Backend/GlobOpt.cpp
```diff
@@ -4877,6 +4877,40 @@ GlobOpt::OptInstr(IR::Instr *&instr, bool* isInstrRemoved)
...
+    {
+        IR::Opnd * opnd = instr->GetSrc1();
+        Sym *sym = opnd->AsSymOpnd()->m_sym;
+        if (sym->IsPropertySym()) {
```
rajatd: nit: '{' on next line

lib/Backend/GlobOpt.cpp
```diff
@@ -4877,6 +4877,40 @@ GlobOpt::OptInstr(IR::Instr *&instr, bool* isInstrRemoved)
...
+            IR::RegOpnd* newopnd = IR::RegOpnd::New(originalPropertySym->m_stackSym, IRType::TyVar, instr->m_func);
+            // only on .length
+            if (this->lengthEquivBv != nullptr && this->lengthEquivBv->Test(originalPropertySym->m_id)) {
+                Value *const objectValue = FindValue(originalPropertySym->m_stackSym);
+                // Only for things we'd emit a fast path for
+                if (
+                    objectValue->GetValueInfo()->IsLikelyAnyArray() ||
+                    objectValue->GetValueInfo()->HasHadStringTag() ||
+                    objectValue->GetValueInfo()->IsLikelyString() ||
+                    newopnd->IsArgumentsObject() ||
```
rajatd: >IsArgumentsObject [](start = 29, length = 17)

The IsArgumentsOpnd() check below is a more precise check for arguments object. You could get rid of this IsArgumentsObject check, unless you kept it for short circuiting

Penguinwizzard: Yeah, that's the idea; the IsArgumentsObject call is much simpler and short-circuits-out needing to run the more complicated arguments check.

lib/Backend/GlobOpt.cpp
```diff
@@ -4877,6 +4877,40 @@ GlobOpt::OptInstr(IR::Instr *&instr, bool* isInstrRemoved)
...
+                    (this->blockData.argObjSyms && IsArgumentsOpnd(newopnd))
+                   )
+                {
```
rajatd: You could just change the current LdFld instruction instead of creating a new instr.

lib/Backend/Lower.cpp
```diff
@@ -14961,6 +14965,19 @@ Lowerer::GenerateFastElemIIntIndexCommon(
...
 
+        // This is a common enough case that we want to go through this path instead of the simpler one, since doing it this way is faster for preallocated but un-filled arrays.
+        if (!!(bailOutKind & IR::BailOutOnInvalidatedArrayLength))
+        {
+            // If we'd increase the array length, go to the helper
+            indirOpnd = IR::IndirOpnd::New(arrayOpnd, Js::JavascriptArray::GetOffsetOfLength(), TyUint32, this->m_func);
+            InsertCompareBranch(
+                newLengthOpnd,
+                indirOpnd,
+                Js::OpCode::BrGe_A,
```
rajatd: >BrGe_A [](start = 28, length = 6)

should this be BrGt_A ?

rajatd:
LGTM except the comments.

lib/Backend/Lower.cpp
Penguinwizzard: Thanks, you are correct.

lib/Backend/IRBuilder.cpp
```diff
@@ -6074,7 +6074,16 @@ IRBuilder::BuildProfiledCallI(Js::OpCode opcode, uint32 offset, Js::RegSlot retu
         newOpcode = opcode;
         Js::OpCodeUtil::ConvertNonCallOpToNonProfiled(newOpcode);
         Assert(newOpcode == Js::OpCode::NewScObject || newOpcode == Js::OpCode::NewScObjectSpread);
-        returnType = ValueType::GetObject(ObjectType::UninitializedObject);
+        if (newOpcode == Js::OpCode::NewScObjectSpread || !this->m_func->HasProfileInfo())
```
satheeshravi: >newOpcode == Js::OpCode::NewScObjectSpread [](start = 12, length = 42)

Why is it special for this?

lib/Backend/GlobOpt.cpp
satheeshravi: +1

---
In reply to: [80305959](https://github.com/Microsoft/ChakraCore/pull/1587#discussion_r80305959) [](ancestors = 80305959)

lib/Runtime/Language/ValueType.h
```diff
@@ -150,6 +150,7 @@ class ValueType
 
     bool HasBeenString() const;
     bool IsString() const;
+    bool HasHadStringTag() const; // Note: this is likely not what you're looking for; this is a short-cut for LdLen optimization.
```
satheeshravi: >Note: this is likely not what you're looking for; this is a short-cut for LdLen optimization. [](start = 37, length = 93)

can you elaborate on this comment here? will be useful for future read.

Penguinwizzard: I've expanded the comment now and moved it so that intellisense can properly pick it up.

lib/Backend/IRBuilder.cpp
Penguinwizzard: I hadn't properly tested enabling it for NewScObjectSpread; the concern was that I would see similar performance degradations as with NewScObject, which would then require additional patches and expand the scope somewhat. I'm currently running tests on performance and functionality with profile data properly enabled for NewScObjectSpread return values.

Penguinwizzard: After a couple of related fixes, NewScObjectSpread profile information use is now enabled in this PR as well.

Issue #963, opened 2016-05-10
xplat: examine _WIN64, should it be _M_X64?


jianchun:
e.g. CallInfo.h, is following "unused" patch applicable to xplat x64?

```
        CallInfo(CallFlags flags, ushort count)
            : Flags(flags)
            , Count(count)
#ifdef _WIN64
            , unused(0)
#endif
        {
        }

        // Assumes big-endian layout
        // If the size of the count is changed, change should happen at following places also
        //  - scriptdirect.idl
        //  - LowererMDArch::LoadInputParamCount
        //
        unsigned  Count : 24;
        CallFlags Flags : 8;
#ifdef _WIN64
        unsigned unused : 32;
#endif
```

obastemur:
How about `_M_X64_OR_ARM64` instead of `_M_X64` ? Especially for the sample you've shared.

jianchun:
Yes, that may be better. Or maybe a new symbol `BIT64`.

LouisLaf:
I don't like _M_X64 of _M_X64_OR_ARM64 as these are specific CPUs is the defines are just about 64 targets.  BIT64 is better but not that descriptive.  How about TARGET64 or HOST64?

jianchun:
> I don't like _M_X64 of _M_X64_OR_ARM64 as these are specific CPUs is the defines are just about 64 targets.  BIT64 is better but not that descriptive.  How about TARGET64 or HOST64?

+1 TARGET64

jianchun:
We now have `TARGET_64`. But lots of code (including new code) still uses `_WIN64`. Needs to search for all and evaluate which should change to `TARGET_64`.

PR #1591, opened 2016-09-16
JIT: enable JIT on Linux

jianchun:
Splitted into 4 commits.

JIT: enable JIT on Linux

    Primary changes to JIT on Linux.

    Custom calling convention
    -------------------------

    Calls to native helpers conform to Sys V AMD64 ABI.
        - Change all code to use correct arg registers.

    Calls to JsMethod and native helpers that depend on
    ARGUMENTS/RUNTIME_ARGUMENTS use custom calling convention:
        - Caller home arg registers onto stack.

    Implemented all needed assembly thunks.

    Unwind data
    -----------

    Emit .eh_frame data for dynamic interpreter thunks and all jit functions.
        - For less noise, the new code hides behind existing
            PrologEncoder/PData/XData.

    Misc
    ----

    ARGUMENTS/RUNTIME_ARGUMENTS:
        - Extended to support all runtime usages.
        - Limitation: known arg list must end with "callInfo".

    Int32Math:
        - Use builtins to test signed int overflow. Original code does not work
            due to clang optimization.

    TODO
        - AsmJs: missing ArrayBuffer out of bound access recovery.
        - Simd.

JIT: to compile on Linux

    Noisy changes to compile native code gen code on Linux with clang.
    Most changes are:
        - Fix missing parentheses around `&&` within `||` expressions
        - Fix `char16/_u` issues
        - Fix incorrect `inline`s

JIT: PAL related changes

    PAL related changes to build JIT on Linux.
        - pal.h: bug fixes in BitScan.
        - Add missing APIs, enable Event.

JIT: build and test changes

    Build:
        - Add backend files.
        - Enable --no-jit option, and as default on OSX (temporarily).

    Test:
        - Add `dynapogo` test variant. Change `interpreted` test variant to
            create dynamic profiles and input to `dynapogo` test variant.
            To ensure the same profile file is used for these 2 variants,
            TestLoader (new) assigns each test case a unique test id, and uses
            the id to generate dynamic profile file name.
        - Add 'disable_jit` test variant and as default on OSX (temporarily).
        - Exclude 2 failing tests that need further work.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -37,7 +37,8 @@ PRINT_USAGE() {
     echo "      --icu=PATH       Path to ICU include folder (see example below)"
     echo "  -j [N], --jobs[=N]   Multicore build, allow N jobs at once"
     echo "  -n, --ninja          Build with ninja instead of make"
-    echo "  --no-icu             Compile without unicode/icu support"
+    echo "      --no-icu         Compile without unicode/icu support"
+    echo "      --no-jit         Disable JIT"
     echo "      --xcode          Generate XCode project"
     echo "  -t, --test-build     Test build (by default Release build)"
     echo "      --static         Build as static library (by default shared library)"
@@ -62,6 +63,7 @@ BUILD_TYPE="Release"
 CMAKE_GEN=
 MAKE=make
 MULTICORE_BUILD=""
+NO_JIT=
 ICU_PATH="-DICU_SETTINGS_RESET=1"
 STATIC_LIBRARY="-DSHARED_LIBRARY_SH=1"
 WITHOUT_FEATURES=""
@@ -191,6 +193,10 @@ while [[ $# -gt 0 ]]; do
         ICU_PATH="-DNO_ICU_PATH_GIVEN_SH=1"
         ;;
 
+    --no-jit)
+        NO_JIT="-DNO_JIT=1"
+        ;;
+
     --xcode)
         CMAKE_GEN="-G Xcode -DCC_XCODE_PROJECT=1"
         MAKE=0
@@ -306,7 +312,7 @@ else
 fi
 
 echo Generating $BUILD_TYPE makefiles
-cmake $CMAKE_GEN $CC_PREFIX $ICU_PATH $STATIC_LIBRARY $ARCH -DCMAKE_BUILD_TYPE=$BUILD_TYPE $WITHOUT_FEATURES ../..
+cmake $CMAKE_GEN $CC_PREFIX $ICU_PATH $STATIC_LIBRARY $ARCH -DCMAKE_BUILD_TYPE=$BUILD_TYPE $NO_JIT $WITHOUT_FEATURES ../..
 
 _RET=$?
 if [[ $? == 0 ]]; then
--- a/build.sh
+++ b/build.sh
@@ -42,6 +42,8 @@ PRINT_USAGE() {
     echo "      --xcode          Generate XCode project"
     echo "  -t, --test-build     Test build (by default Release build)"
     echo "      --static         Build as static library (by default shared library)"
+    echo "      --sanitize=CHECKS Build with clang -fsanitize checks,"
+    echo "                       e.g. undefined,signed-integer-overflow"
     echo "  -v, --verbose        Display verbose output including all options"
     echo "      --create-deb=V   Create .deb package with given V version"
     echo "      --without=FEATURE,FEATURE,..."
@@ -66,6 +68,7 @@ MULTICORE_BUILD=""
 NO_JIT=
 ICU_PATH="-DICU_SETTINGS_RESET=1"
 STATIC_LIBRARY="-DSHARED_LIBRARY_SH=1"
+SANITIZE=
 WITHOUT_FEATURES=""
 CREATE_DEB=0
 ARCH="-DCC_TARGETS_AMD64_SH=1"
@@ -194,7 +197,7 @@ while [[ $# -gt 0 ]]; do
         ;;
 
     --no-jit)
-        NO_JIT="-DNO_JIT=1"
+        NO_JIT="-DNO_JIT_SH=1"
         ;;
 
     --xcode)
@@ -211,13 +214,19 @@ while [[ $# -gt 0 ]]; do
         STATIC_LIBRARY="-DSTATIC_LIBRARY_SH=1"
         ;;
 
+    --sanitize=*)
+        SANITIZE=$1
+        SANITIZE=${SANITIZE:11}    # value after --sanitize=
+        SANITIZE="-DCLANG_SANITIZE_SH=${SANITIZE}"
+        ;;
+
     --without=*)
         FEATURES=$1
         FEATURES=${FEATURES:10}    # value after --without=
         for x in ${FEATURES//,/ }  # replace comma with space then split
         do
             if [[ "$WITHOUT_FEATURES" == "" ]]; then
-                WITHOUT_FEATURES="-DWITHOUT_FEATURES="
+                WITHOUT_FEATURES="-DWITHOUT_FEATURES_SH="
             else
                 WITHOUT_FEATURES="$WITHOUT_FEATURES;"
             fi
@@ -312,7 +321,8 @@ else
 fi
 
 echo Generating $BUILD_TYPE makefiles
-cmake $CMAKE_GEN $CC_PREFIX $ICU_PATH $STATIC_LIBRARY $ARCH -DCMAKE_BUILD_TYPE=$BUILD_TYPE $NO_JIT $WITHOUT_FEATURES ../..
+cmake $CMAKE_GEN $CC_PREFIX $ICU_PATH $STATIC_LIBRARY $ARCH \
+    -DCMAKE_BUILD_TYPE=$BUILD_TYPE $SANITIZE $NO_JIT $WITHOUT_FEATURES ../..
 
 _RET=$?
 if [[ $? == 0 ]]; then
--- a/lib/Backend/EhFrame.cpp
+++ b/lib/Backend/EhFrame.cpp
@@ -1,0 +1,226 @@
+//-------------------------------------------------------------------------------------------------------
+// Copyright (C) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE.txt file in the project root for full license information.
+//-------------------------------------------------------------------------------------------------------
+#include "Backend.h"
+#include "EhFrame.h"
+
+// AMD64 ABI -- DWARF register number mapping
+static const ubyte DWARF_RegNum[] =
+{
+    // Exactly same order as RegList.h!
+    -1, // NOREG,
+    0,  // RAX,
+    2,  // RCX,
+    1,  // RDX,
+    3,  // RBX,
+    7,  // RSP,
+    6,  // RBP,
+    4,  // RSI,
+    5,  // RDI,
+    8,  // R8,
+    9,  // R9,
+    10, // R10,
+    11, // R11,
+    12, // R12,
+    13, // R13,
+    14, // R14,
+    15, // R15,
+    17,  // XMM0,
+    18,  // XMM1,
+    19,  // XMM2,
+    20,  // XMM3,
+    21,  // XMM4,
+    22,  // XMM5,
+    23,  // XMM6,
+    24,  // XMM7,
+    25,  // XMM8,
+    26,  // XMM9,
+    27,  // XMM10,
+    28,  // XMM11,
+    29,  // XMM12,
+    30,  // XMM13,
+    31,  // XMM14,
+    32,  // XMM15,
+};
+
+static const ubyte DWARF_RegRA = 16;
+
+ubyte GetDwarfRegNum(ubyte regNum)
+{
+    return DWARF_RegNum[regNum];
+}
+
+// Enocde into ULEB128 (Unsigned Little Endian Base 128)
+BYTE* EmitLEB128(BYTE* pc, unsigned value)
+{
+    do
+    {
+        BYTE b = value & 0x7F; // low order 7 bits
+        value >>= 7;
+
+        if (value)  // more bytes to come
+        {
+            b |= 0x80;
+        }
+
+        *pc++ = b;
+    }
+    while (value != 0);
+
+    return pc;
+}
+
+// Encode into signed LEB128 (Signed Little Endian Base 128)
+BYTE* EmitLEB128(BYTE* pc, int value)
+{
+    static const int size = sizeof(value) * 8;
+    static const bool isLogicShift = (-1 >> 1) != -1;
+
+    const bool signExtend = isLogicShift && value < 0;
+
+    bool more = true;
+    while (more)
+    {
+        BYTE b = value & 0x7F; // low order 7 bits
+        value >>= 7;
+
+        if (signExtend)
+        {
+            value |= - (1 << (size - 7)); // sign extend
+        }
+
+        const bool signBit = (b & 0x40) != 0;
+        if ((value == 0 && !signBit) || (value == -1 && signBit))
+        {
+            more = false;
+        }
+        else
+        {
+            b |= 0x80;
+        }
+
+        *pc++ = b;
+    }
+
+    return pc;
+}
+
+
+void EhFrame::Entry::Begin()
+{
+    Assert(beginOffset == -1);
+    beginOffset = writer->Count();
+
+    // Write Length place holder
+    const uword length = 0;
+    writer->Write(length);
+}
+
+void EhFrame::Entry::End()
+{
+    // padding
+    size_t padding = (MachPtr - writer->Count() % MachPtr) % MachPtr;
+    for (size_t i = 0; i < padding; i++)
+    {
+        cfi_nop();
+    }
+
+    // update length record
+    uword length = writer->Count() - beginOffset
+                    - sizeof(length);  // exclude length itself
+    writer->Write(beginOffset, length);
+}
+
+void EhFrame::Entry::cfi_advance(uword advance)
+{
+    if (advance <= 0x3F)        // 6-bits
+    {
+        cfi_advance_loc(static_cast<ubyte>(advance));
+    }
+    else if (advance <= 0xFF)   // 1-byte
+    {
+        cfi_advance_loc1(static_cast<ubyte>(advance));
+    }
+    else if (advance <= 0xFFFF) // 2-byte
+    {
+        cfi_advance_loc2(static_cast<uword>(advance));
+    }
+    else                        // 4-byte
+    {
+        cfi_advance_loc4(advance);
+    }
+}
+
+void EhFrame::CIE::Begin()
+{
+    Assert(writer->Count() == 0);
+    Entry::Begin();
+
+    const uword cie_id = 0;
+    Emit(cie_id);
+
+    const ubyte version = 1;
+    Emit(version);
+
+    const ubyte augmentationString = 0; // none
+    Emit(augmentationString);
+
+    const ULEB128 codeAlignmentFactor = 1;
+    Emit(codeAlignmentFactor);
+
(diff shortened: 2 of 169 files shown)


jianchun:
@LouisLaf @curtisman @digitalinfinity @obastemur Please CR, thanks!

jianchun:
@dotnet-bot Windows x64_test please

jianchun:

lib/Backend/EhFrame.cpp
```diff
@@ -0,0 +1,226 @@
...
+{
+    // padding
+    size_t padding = (MachPtr - writer->Count() % MachPtr) % MachPtr;
+    for (size_t i = 0; i < padding; i++)
+    {
+        cfi_nop();
+    }
+
+    // update length record
+    uword length = writer->Count() - beginOffset
```
obastemur: `Assert(sizeof(length) + beginOffset <= writer->Count())` ?

lib/Backend/amd64/LinearScanMdA.S
```diff
@@ -0,0 +1,139 @@
...
+// Copyright (C) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE.txt file in the project root for full license information.
+//-------------------------------------------------------------------------------------------------------
+
+.intel_syntax noprefix
+#include "unixasmmacros.inc"
+
+
+#ifndef __APPLE__
+// BailOutRecord::BailOut(BailOutRecord const * bailOutRecord)
```
obastemur: Is this intentionally left?

lib/Backend/amd64/LowererMDArch.cpp
```diff
@@ -857,22 +881,85 @@ LowererMDArch::LowerCall(IR::Instr * callInstr, uint32 argCount)
...
     }
 
+#ifndef _WIN32
+    // Manually home args
+    if (argCount > 0)
+    {
+        static const RegNum s_argRegs[IntArgRegsCount] = {
+    #define REG_INT_ARG(Index, Name)  Reg ## Name,
+    #include "RegList.h"
+        };
```
obastemur: undef REG_INT_ARG ?

lib/Backend/amd64/LowererMDArch.cpp
```diff
@@ -934,51 +1021,38 @@ LowererMDArch::GetArgSlotOpnd(uint16 index, StackSym * argSym)
...
-                break;
-            default:
-                Assume(UNREACHED);
-            }
+#define REG_INT_ARG(Index, Name)    \
+        case ((Index) + 1):         \
+            reg = Reg ## Name;      \
+            break;
+#include "RegList.h"
+        default:
```
obastemur: undef REG_INT_ARG?

lib/Backend/amd64/LowererMDArch.cpp
```diff
@@ -934,51 +1021,38 @@ LowererMDArch::GetArgSlotOpnd(uint16 index, StackSym * argSym)
...
-                break;
-            default:
-                Assume(UNREACHED);
-            }
+#define REG_XMM_ARG(Index, Name)    \
+        case ((Index) + 1):         \
+            reg = Reg ## Name;      \
+            break;
+#include "RegList.h"
+        default:
```
obastemur: undef REG_XMM_ARG ?

lib/Backend/amd64/PeepsMD.cpp
```diff
@@ -18,19 +18,12 @@ PeepsMD::ProcessImplicitRegs(IR::Instr *instr)
...
-        this->peeps->ClearReg(RegXMM2);
-        this->peeps->ClearReg(RegXMM3);
-        this->peeps->ClearReg(RegXMM4);
-        this->peeps->ClearReg(RegXMM5);
+#define REGDAT(Name, Listing, Encode, Type, BitVec) \
+        if (!((BitVec) & (RA_CALLEESAVE | RA_DONTALLOCATE))) \
+        { \
+            this->peeps->ClearReg(Reg ## Name); \
+        }
+#include "RegList.h"
```
obastemur: undef REGDAT?

lib/Common/Common/Int32Math.cpp
```diff
@@ -176,17 +176,31 @@ Int32Math::Not(int32 val, int32 *pResult)
...
 {
+#ifdef _MSC_VER
     *pResult = val + 1;
     // Overflow if result ends up less than input
     return *pResult <= val;
+#elif defined(__APPLE__)
+    *pResult = val + 1;
+    return val == INT32_MAX; // Overflow if val was int max
+#else
+    return __builtin_add_overflow(val, 1, pResult);
```
obastemur: Is there a reason why we haven't use this for `__APPLE__` too ?

lib/Common/Common/Int32Math.cpp
```diff
@@ -176,17 +176,31 @@ Int32Math::Not(int32 val, int32 *pResult)
...
 {
+#ifdef _MSC_VER
     *pResult = val - 1;
     // Overflow if result ends up greater than input
     return *pResult >= val;
+#elif defined(__APPLE__)
+    *pResult = val - 1;
+    return val == INT32_MIN; // Overflow if val was int min
+#else
+    return __builtin_sub_overflow(val, 1, pResult);
```
obastemur: Is there a reason why we haven't use this for __APPLE__ too ?

lib/Runtime/Language/Arguments.h
```diff
@@ -4,17 +4,58 @@
...
 #pragma once
 
+// To extract variadic args array after known args list:
+//      argx, callInfo, ...
+// NOTE: The last known arg name is hard-coded to "callInfo".
 #ifdef _WIN32
-#define VA_LIST_TO_VARARRAY(vl, va, callInfo) Js::Var* va = (Js::Var*) vl;
+#define DECLARE_ARGS_VARARRAY(va, ...)                              \
+    va_list _vl;                                                    \
+    va_start(_vl, callInfo);                                        \
```
obastemur: We better call `va_end` once we are done with it. Previously, I had cleaned up the va leaks. See #1507

lib/Runtime/Language/amd64/amd64_Thunks.S
```diff
@@ -0,0 +1,382 @@
...
+        ret
+NESTED_END _ZN2Js21InterpreterStackFrame19InterpreterAsmThunkEPNS_20AsmJsCallStackLayoutE, _TEXT
+
+//============================================================================================================
+// Js::AsmJsExternalEntryPoint
+//============================================================================================================
+
+//extrn ?GetStackSizeForAsmJsUnboxing@Js@@YAHPEAVScriptFunction@1@@Z: PROC
+//extrn ?UnboxAsmJsArguments@Js@@YAPEAXPEAVScriptFunction@1@PEAPEAXPEADUCallInfo@1@@Z : PROC
+// extrn ?BoxAsmJsReturnValue@Js@@YAPEAXPEAVScriptFunction@1@HNM@Z : PROC
```
obastemur: nit: extra space

pal/inc/rt/palrt.h
```diff
@@ -890,10 +890,10 @@ Remember to fix the errcode defintion in safecrt.h.
 #define _vscprintf _vscprintf_unsafe
 #define _vscwprintf _vscwprintf_unsafe
 
-#define sprintf_s _snprintf
-#define swprintf_s _snwprintf
-#define vsprintf_s _vsnprintf
-#define vswprintf_s _vsnwprintf
+// #define sprintf_s _snprintf
```
obastemur: We might better remove them?

build.sh
```diff
@@ -184,6 +186,10 @@ while [[ $# -gt 0 ]]; do
         ICU_PATH="-DNO_ICU_PATH_GIVEN_SH=1"
         ;;
 
+    --no-jit)
+        NO_JIT="-DNO_JIT=1"
```
jianchun: Will try this, thanks!

lib/Backend/EhFrame.cpp
jianchun: I don't see this assert to be very useful. Once Begin() was called this would hardly fail as the same uword type is written again at beginOffset. Maybe I can assert beginOffset != -1 to ensure Begin() was called?

lib/Backend/amd64/LinearScanMdA.S
jianchun: These 2 symbols are used as JMP targets in this file. Don't know why it works with these commented out. Not sure if OSX needs them. Can cleanup when enabling this for OSX.

lib/Backend/amd64/LowererMDArch.cpp
jianchun: Because this file is included many times and the file has default `#define`s, I added `#undef`s in the file itself.

lib/Common/Common/Int32Math.cpp
jianchun: AppleClang does not have this builtin (reported by CI).

lib/Runtime/Language/Arguments.h
jianchun: This is Windows x64 only and the macro does not know when we are done with it (we cast `_vl` to `Var*` and using it).

lib/Backend/amd64/LinearScanMdA.S
obastemur: But that part is `ifndef __APPLE__` I don't think it has anything to do with OSX

jianchun: Yes, originally I thought I needed them on Linux that's why they are under `ifndef __APPLE__`. I can remove `ifndef` and only keep the commented `.extern`s.

lib/Common/Common/Int32Math.cpp
obastemur: Actually AppleClang has that. Don't know why CI complained.

Instead of making it `ifdef __APPLE__` we better feature detect from clang.

```
#ifdef __has_builtin
#if __has_builtin(__builtin_add_overflow)
```
Same applies to other builtin comments below

jianchun: Thanks, will do that. Does this work on gcc?

obastemur: > Does this work on gcc?

Hopefully for GCC 6.0. We don't support GCC though. There are so many problems with GCC.

DemiMarie: @obastemur what are those problems?

obastemur: @DemiMarie not a small list. i.e. Clang has some useful flags for C/C++/ASM code that is initially developed for VC/VC++

build.sh
jianchun: This works, thanks! I'm using `NO_JIT` in place of `NO_JIT_SH` as `NO_JIT` is not needed afterwards.

lib/Backend/amd64/LinearScanMdA.S
jianchun: Removed `ifndef __APPLE__`.

build.sh
obastemur: nit:`_SH` suffix is useful for remembering the source of the variable though. The rest of the definitions from build.sh use the same suffix.

jianchun:

jianchun:
@dotnet-bot test all please

jianchun:
@dotnet-bot run tests please

jianchun:

jianchun:
@dotnet-bot test Windows arm_debug please

jianchun:
@rajatd @obastemur @curtisman Updated, please review, thanks!

build.sh
jianchun: @obastemur Agreed, changed 3 variables (all added by me) to `_sh` suffix in latest update.

obastemur: @jianchun Thanks

lib/Backend/Opnd.cpp
```diff
@@ -3141,7 +3141,7 @@ Opnd::DumpOpndKindMemRef(bool AsmDumpMode, Func *func)
 void
 Opnd::WriteToBuffer(_Outptr_result_buffer_(*count) char16 **buffer, size_t *count, const char16 *fmt, ...)
 {
-    va_list argptr = nullptr;
```
rajatd: why remove the initialization?

lib/Common/CommonDefines.h
```diff
@@ -183,6 +183,12 @@
 #endif
 #endif
 
+#if ENABLE_NATIVE_CODEGEN
+#ifdef _WIN32
```
rajatd: || with _WIN64?

rajatd: >_WIN32 [](start = 7, length = 6)

|| with _WIN64?

jianchun: _WIN32 contains _WIN64 (https://msdn.microsoft.com/en-us/library/b0084kay.aspx):

```
_WIN32 Defined as 1 when the compilation target is 32-bit ARM, 64-bit ARM, x86, or x64. Otherwise, undefined.

_WIN64 Defined as 1 when the compilation target is 64-bit ARM or x64. Otherwise, undefined. 
```

lib/Backend/amd64/LowererMDArch.cpp
```diff
@@ -857,22 +881,85 @@ LowererMDArch::LowerCall(IR::Instr * callInstr, uint32 argCount)
...
+    // Manually home args
+    if (argCount > 0)
+    {
+        static const RegNum s_argRegs[IntArgRegsCount] = {
+    #define REG_INT_ARG(Index, Name)  Reg ## Name,
+    #include "RegList.h"
+        };
+
+        const int callArgCount = static_cast<int>(argCount) + this->helperCallArgsCount;
+        const int argRegs = min(callArgCount, static_cast<int>(IntArgRegsCount));
```
rajatd: >min(callArgCount, static_cast<int>(IntArgRegsCount)); [](start = 28, length = 53)

XmmArgRegsCount is 8. If there were, in fact, 8 float arguments, wouldn't we be missing to home 2 of them in the below loop?

jianchun: This actually only homes the int registers, corresponding to our runtime usage (mostly function, callInfo, Js::Var... which don't use float registers). I'm not sure of AsmJs/Simd requirements. This PR has some basic thunk assembly implementations targeting AsmJs unit tests. Simd not looked at yet.

rajatd:
> #ifdef _WIN64

change this to TARGET_64?

---
Refers to: lib/Backend/Encoder.cpp:689 in ded0397. [](commit_id = ded03978554fa07785becd66eb02caa6fc74a40a, deletion_comment = False)

rajatd:
> #if _WIN64

change this to TARGET_64?

---
Refers to: lib/Backend/BackwardPass.cpp:163 in ded0397. [](commit_id = ded03978554fa07785becd66eb02caa6fc74a40a, deletion_comment = False)

jianchun:
> change this to TARGET_64?

Sure. Maybe I can search for all _WIN64 usage -- likely they should all be TARGET_64?

lib/Backend/amd64/LowererMDArch.cpp
```diff
@@ -857,22 +881,85 @@ LowererMDArch::LowerCall(IR::Instr * callInstr, uint32 argCount)
...
+#ifndef _WIN32
+    // Manually home args
+    if (argCount > 0)
+    {
+        static const RegNum s_argRegs[IntArgRegsCount] = {
+    #define REG_INT_ARG(Index, Name)  Reg ## Name,
+    #include "RegList.h"
+        };
+
+        const int callArgCount = static_cast<int>(argCount) + this->helperCallArgsCount;
```
rajatd: >helperCallArgsCount [](start = 68, length = 19)

Can helperCallArgsCount be non-zero here? If yes, then the loop below could potentially overwrite some of the stack slots written to in the above while(argsLeft>0) loop. Please verify this and add asserts to make sure that only one of these loops run. Some comments about which loop is used for what kind of call would be nice too.

jianchun: Yes, will test that. Thanks!

jianchun: @rajatd I looked at this issue. Yes helperCallArgsCount can be non-zero. The reason is that many lowering code uses LowerCallArgs() to lower variadic JS runtime args, and LoadHelperArgument() to lower other "known args". So the full args length is actually their sum, while `argCount > 0` is an indicator that we need variadic args support so we need to manually home args.

Your concern of manual homing args here overwriting above helperCallArgs stack slots could potentially happen... e.g. `helperCallArgs == 7` so it has an arg at stack[0]. I can fix this by changing above `/*isHelper*/true` to `argCount == 0` so that their assumptions are consistent.

jianchun:
> Maybe I can search for all _WIN64 usage -- likely they should all be TARGET_64?

To avoid more noise in this PR, I'll follow up afterwards. We already have #963.

lib/Backend/amd64/LowererMDArch.cpp
rajatd: Yes, using "argCount == 0" as the argument to GetArgSlotOpnd should ensure correctness.

---
In reply to: [81197219](https://github.com/Microsoft/ChakraCore/pull/1591#discussion_r81197219) [](ancestors = 81197219)

lib/Runtime/Language/Arguments.h
```diff
@@ -4,17 +4,58 @@
...
+#define DECLARE_ARGS_VARARRAY_N(va, n)                              \
+    Js::Var* va = _get_va(_AddressOfReturnAddress(), n);            \
+    Assert(*reinterpret_cast<Js::CallInfo*>(va - 1) == callInfo)
+
+#define DECLARE_ARGS_VARARRAY(va, ...)                              \
+    DECLARE_ARGS_VARARRAY_N(va, _count_args(__VA_ARGS__))
+
+inline Js::Var* _get_va(void* addrOfReturnAddress, int n)
+{
+    Js::Var* pArgs = reinterpret_cast<Js::Var*>(addrOfReturnAddress) + 1;
```
rajatd: >1 [](start = 71, length = 1)

A comment about what this '1' is for, would be good. (saved frame pointer, right?)

jianchun: No, RA (ReturnAddress). Args are after RA.

rajatd: Oh right, my bad.

---
In reply to: [81249310](https://github.com/Microsoft/ChakraCore/pull/1591#discussion_r81249310) [](ancestors = 81249310)

jianchun: Will add a comment, thanks!

lib/Runtime/Language/Arguments.h
```diff
@@ -4,17 +4,58 @@
...
+    return pArgs + n;
+}
+
+inline int _count_args(Js::CallInfo callInfo)
+{
+    return 2;  // for typical JsMethod with 2 known args "function, callInfo"
```
rajatd: Is it possible to somehow enforce/assert here that the other argument is a JavascriptFunction?

rajatd:
Reviewed most of the backend changes and left relevant comments. Haven't reviewed PDataManager, PrologEncoder, EhFrame and XDataAllocator as I dont know much about that area. Maybe @curtisman @pleath or @LouisLaf can help with that.

lib/Runtime/Language/Arguments.h
jianchun: This is to support the majority usages of `ARGUMENTS(args, callInfo)`. I can't think of a good way to check that. If it is used incorrectly runtime likely to fail. Hope tests can guard that.

curtisman:
:shipit:

jianchun:
Many thanks to @rajatd @obastemur @curtisman for reviewing this huge PR!
@pleath Please let me know if you have comments, I'll address later.

Issue #1415, opened 2016-08-10
Function.prototype.call returns the function itself instead of the real return value


Tragetaschen:
I have convinced myself, that I am looking at an optimization issue in EdgeHTML 14. The most annoying part is, that it doesn't reproduce when the F12 tools are open.

I'm building an application including the new d3 version 4, angular2, and core-js' es6 modules (amongst other things). Everything is bundled with webpack. I have tried to reduce the problem by editing the webpack-generated JavaScript and removing some of the moving parts, but that seems to fix it. I'll try to describe my environment, some code snippets etc., but I wouldn't expect this to be easy to reproduce.

The problem appears in [the range function](https://github.com/d3/d3-scale/blob/master/src/continuous.js#L97), more exactly in the `range = slice.call(_)` assignment in the first branch. `slice` is imported from `array.js` and is just `Array.prototype.slice`.

The first three times this function gets called (when `arguments.length === 1`, an array with two elements), the `.call(_)` returns the expected copy of that array. The fourth time, `range` gets `function slice() { [native code] }` assigned instead.

With webpack in the picture, here is what the browser actually executes:

``` js
/* harmony import */ var __WEBPACK_IMPORTED_MODULE_2__array__ = __webpack_require__(22);

// […]

  scale.range = function(_) {
    return arguments.length ? (range = __WEBPACK_IMPORTED_MODULE_2__array__["a" /* slice */].call(_), rescale()) : range.slice();
  };
```

or my debug version:

``` js
  scale.range = function (_) {
      var result;
      if (arguments.length) {
          console.log('change!');
          console.log(range);
          range = __WEBPACK_IMPORTED_MODULE_2__array__["a" /* slice */].call(_);
          console.log(range);
          result = rescale();
      }
      else {
          result = range.slice();
      }
      return result;
  };
```

I have rerouted `console.log()` and `console.error()` to append the logs to a DOM element, so I have a chance to see what happened when I open the F12 tools after things blew up. The log is:

```
Angular 2 is running in the development mode. Call enableProdMode() to enable the production mode.
change!
0,1
278.3333333333333,0
change!
0,1
0,885
change!
0,1
278.3333333333333,0
change!
0,1
function slice() { [native code] }
```

Continuing the program executing then runs into an exception when the `range` function is called without arguments:

```
TypeError: Object doesn't support property or method 'slice'
  at scale.range (…)
  at …
```

I have verified, that the webpack imported `slice` function is always the same.
[...]

PR #1598, opened 2016-09-18
Fix bug with .call and getter inlining

rajatd:
Fixing an issue when the target of a .call/.apply is the result of a property load that has a getter. We were ending up inlining the getter instead of the function returned by the getter.


rajatd:
Fixes #1415

rajatd:
@LouisLaf could you review please?

LouisLaf:
LGTM

PR #1599, opened 2016-09-18
OSX SDK 10.12 build break

obastemur:
Just saw this after updating one of my dev machines.
CC is not compiling on a machine with Xcode 8.0+ installed due to changes to latest clang.

`clang: warning: libstdc++ is deprecated; move to libc++ with a minimum deployment target of OS X 10.9`

This warning breaks the CMake feature detection scripts.
We could update to libc++. However, that would break node-chakracore.

In order not to break node-chakracore builds, I'm turning off the CXX `-werror` for osx until node-chakracore is updated to libc++

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -125,6 +125,11 @@ if(CMAKE_SYSTEM_NAME STREQUAL Linux)
     endif()
 
     set(CLR_CMAKE_PLATFORM_LINUX 1)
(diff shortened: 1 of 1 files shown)


digitalinfinity:
LGTM- are you tracking re-enabling with an issue or should we add a todo here?

obastemur:
Sierra is out. So, merging this now.

@digitalinfinity Thanks for the review. We might open an issue on this but lets see where the offline discussion will go first.

PR #1601, opened 2016-09-19
Fix update of likely int bounds

meg-gupta:
The function GlobOpt::GetIntBoundsToUpdate should not return bounds for a
likely int. The bounds returned by this function are adjusted based on
whether the conditional branch was less than, less than equal to, etc.
Updating bound on a likely int, can result in incorrect results for edge
cases.

### DIFF
--- a/lib/Backend/GlobOptIntBounds.cpp
+++ b/lib/Backend/GlobOptIntBounds.cpp
@@ -773,9 +773,10 @@ IntBounds *GlobOpt::GetIntBoundsToUpdate(
             // New relative bounds are not being set, will use IntRangeValueInfo instead
             return nullptr;
         }
+        return IntBounds::New(constantBounds, false, alloc);
     }
 
-    return IntBounds::New(constantBounds, false, alloc);
+    return nullptr;
 }
 
 ValueInfo *GlobOpt::UpdateIntBoundsForEqual(
--- a/test/Bugs/rlexe.xml
+++ b/test/Bugs/rlexe.xml
@@ -330,4 +330,10 @@
       <tags>exclude_dynapogo</tags>
     </default>
   </test>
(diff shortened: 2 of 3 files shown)


meg-gupta:
@LouisLaf Please take a look. Thanks!

rajatd:
No regressions anywhere?

meg-gupta:
No perf regressions, exprgen is still running

rajatd:
LGTM. Its a risky change though.

meg-gupta:
Thanks @rajatd. Will do one more round of buddy runs, before checking in.

lib/Backend/GlobOptIntBounds.cpp
```diff
@@ -773,9 +773,10 @@ IntBounds *GlobOpt::GetIntBoundsToUpdate(
             // New relative bounds are not being set, will use IntRangeValueInfo instead
             return nullptr;
         }
+        return IntBounds::New(constantBounds, false, alloc);
     }
 
-    return IntBounds::New(constantBounds, false, alloc);
+    return nullptr;
```
satheeshravi: >nullptr [](start = 11, length = 7)

Is there a failing unit test? If so, can you add one ?

meg-gupta: Added unit test

test/Bugs/rlexe.xml
```diff
@@ -330,4 +330,11 @@
       <tags>exclude_dynapogo</tags>
     </default>
   </test>
+  <test>
+    <default>
+      <files>typespec_bug.js</files>
+      <tags>Slow</tags>
```
satheeshravi: Slow is not necessary unless the test case is actually slow.

PR #1603, opened 2016-09-19
stop inlining decider selecting asm.js functions with -force:inline

MikeHolman:
This caused issues with oop jit when trying to build inline info for the asm.js closure, which is an invalid scenario


rajatd:
:shipit:

PR #1605, opened 2016-09-20
fix oop jit bugs with reclaimed property records

MikeHolman:
when property records were reclaimed we should remove them from list of pending jit property records.
when reclaimed property later has pid reused, we should first delete the old record on the server map before adding new one.

### DIFF
--- a/lib/Runtime/Base/ThreadContext.cpp
+++ b/lib/Runtime/Base/ThreadContext.cpp
@@ -167,6 +167,7 @@ ThreadContext::ThreadContext(AllocationPolicyManager * allocationPolicyManager, 
 #if ENABLE_NATIVE_CODEGEN
     codeGenNumberThreadAllocator(nullptr),
     m_pendingJITProperties(nullptr),
+    m_reclaimedJITProperties(nullptr),
     xProcNumberPageSegmentManager(nullptr),
 #if DYNAMIC_INTERPRETER_THUNK || defined(ASMJS_PLAT)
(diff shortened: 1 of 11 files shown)


lib/Runtime/Base/ThreadContext.cpp
```diff
@@ -1277,9 +1277,10 @@ bool ThreadContext::IsActivePropertyId(Js::PropertyId pid)
 void ThreadContext::InvalidatePropertyRecord(const Js::PropertyRecord * propertyRecord)
 {
     InternalInvalidateProtoTypePropertyCaches(propertyRecord->GetPropertyId());     // use the internal version so we don't check for active property id
-
+#if ENABLE_NATIVE_CODEGEN
+    m_pendingJITProperties->Remove(propertyRecord);
```
leirocks: Need to update the server that the property record is removed? maybe maintain a removed list of propIds, do incremental removing similar to the incremental adding?

MikeHolman: This was something I considered doing, but it seemed like a memory optimization (reclaiming in the runtime keeps the size bounded: on server we'll never hold more than the highest propertyid number of records), and I was a bit worried about increasing complexity/perf cost.

That said, upon reflection I think you're right. It's not much more effort, and it gives us more complete story for property records. I'll add the removal support in tomorrow morning.

MikeHolman: @leirocks I've added support for JIT reclaiming property records

leirocks:
:+1:

PR #1606, opened 2016-09-20
fix arm xdata for oopjit

MikeHolman:
xdata registration was not correct for arm. we heap allocate xdata on arm, so we should put it in native code data buffer when we are oop


leirocks:
:shipit:

PR #1607, opened 2016-09-20
move cfg registration oop

MikeHolman:
moves cfg registration to server side


satheeshravi:
:shipit:

dilijev:
When you rebase on master, the check_tab.sh script will be present and the Tab Check should pass.

PR #1612, opened 2016-09-20
wasm 0xc misc

commonlisp:
- end opcode
- removal of arities from return and call
- updated unit tests to conform


commonlisp:
@Cellule @MikeHolman It is unfortunate that the check_tabs.sh fixes were lumped together here (and likely introducing future merge conflicts), but this PR does get WebAssembly checking again and updated with a few 0xc changes.

dilijev:
@commonlisp @MikeHolman @Cellule Sorry about the disruption the Tab Check caused. I meant to enable it only on master. Anyway, looks like you're all fixed up. I'm not too worried about future merge conflicts, but we'll see.

Cellule:
:lgtm:

---

Reviewed 10 of 11 files at r1, 1 of 1 files at r2, 1 of 2 files at r3, 5 of 5 files at r4.
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---

*[lib/WasmReader/WasmBinaryReader.cpp, line 314 at r4](https://reviewable.io:443/reviews/microsoft/chakracore/1612#-KSE2AgzkhmcJbGCibsT:-KSE2AgzkhmcJbGCibsU:bsueqch) ([raw file](https://github.com/microsoft/chakracore/blob/e87405b973f27316fa45c1cc06ebd4d2f4398898/lib/WasmReader/WasmBinaryReader.cpp#L314)):*
> ```cpp
>     {
>         // end of AST
>         if ((WasmOp)*m_pc++ != wbEnd)
> ```

Since we will read that op no matter what, how about we move `WasmOp op = (WasmOp)*m_pc++;` before `if(EndOfFunc())` then just check with `if(op != wbEnd)`
We could actually move up also, `++m_funcState.count;` and leave the previous behavior of `EndOfFunc`.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/microsoft/chakracore/1612#-:-KSEAxB9_lEzAnXfCbmG:bnfp4nl)*

MikeHolman:
>     ++m_funcState.count;

agree with @Cellule on moving this before the `EndOfFunc` check. Otherwise lgtm

---
Refers to: lib/WasmReader/WasmBinaryReader.cpp:322 in 5547e84. [](commit_id = 5547e847713b3e21d11c6c45411a631785cdea57, deletion_comment = False)

MikeHolman:
:shipit:

commonlisp:
Yeah, good point, I've moved them up and restored the old EndOfFunc().

Cellule:
I signed your change, you can merge whenever the jenkins task are done :)

PR #1617, opened 2016-09-21
It is pointless to check program name itself (argv[0]) against flags.

navidR:
AFAIK in every platform first parameter in argv will be program name itself, so it is pointless to check program name itself to flags.

### DIFF
--- a/bin/ch/ch.cpp
+++ b/bin/ch/ch.cpp
@@ -755,8 +755,8 @@ int _cdecl wmain(int argc, __in_ecount(argc) LPWSTR argv[])
     }
 #endif
 
-    int cpos = 0;
-    for(int i = 0; i < argc; ++i)
+    int cpos = 1;
+    for(int i = 1; i < argc; ++i)
     {
         if(wcsstr(argv[i], _u("-TTRecord=")) == argv[i])
         {


dilijev:
The reason this broke all those tests is that it's actually important in this context to start from 0. This block of code is pre-processing certain flags and then rewriting the argc and argv values with those flags removed from argv. Since argv[0] is expected to be the program name when it is processed later, and your code now effectively removes the program name from argv, the tests don't run as expected.

bin/ch/ch.cpp
```diff
@@ -756,7 +756,7 @@ int _cdecl wmain(int argc, __in_ecount(argc) LPWSTR argv[])
 #endif
 
     int cpos = 0;
```
obastemur: setting `cpos` to 1 should fix the issues.

obastemur:
@dilijev I guess @navidR missed updating `cpos` to 1 also. Other than that it's a nice effort to improve the things around. @navidR Thanks for the PR.

Please update `cpos` under the same commit and try not to add another one :)

dilijev:
Although I'm wondering now why we @mrkmarron decided to parse these flags ahead of time and in this manner instead of via the same flag parsing mechanisms used for other flags. Maybe the right thing to do here is to unify the argument parsing.

obastemur:
@dilijev I asked that question myself too :) However, we consume TTD flags and others a bit differently. That might be the reason.

navidR:
> The reason this broke all those tests is that it's actually important in this context to start from 0. This block of code is pre-processing certain flags and then rewriting the argc and argv values with those flags removed from argv. 

Good to know, I was wandering for an hour what went wrong there :) 

> Please update cpos under the same commit and try not to add another one :) i.e. squash or force push

of course .

obastemur:
@navidR both `cpos` and `i` should be 1

mrkmarron:
Hi, as obastemur, noted the TTD flags are intended for the host instead of the Chakra engine. Since the purpose and destinations are different I opted to break the action out into separate code. That said, it is not the clearest arrangement and I am definitely open to proposals to clean this up.

@navidR thanks for the contribution and thanks for working through the confusion around my TTD specific argument parsing.

dilijev:
Looks good. We can take this PR as-is and work on other ways of cleaning this up at a later time if it is necessary.

PR #1618, opened 2016-09-21
Few TypedArray's methods fix.

akroshg:
There are few subtle difference in few typedarray methods when you compare to spec. Namely,
1. In filter method the TypedArraySpeciesCreate should happen after we go through each element.
2. In join method separator should be evaluated first, before looking for typedarray objects' length.
3. In join method for 'undefined' we use ',' as separator.
4. keys/values/entries should work only with TypedArray object only.

### DIFF
--- a/lib/Runtime/Library/TypedArray.cpp
+++ b/lib/Runtime/Library/TypedArray.cpp
@@ -1730,6 +1730,24 @@ namespace Js
         return JavascriptArray::CopyWithinHelper(nullptr, typedArrayBase, typedArrayBase, length, args, scriptContext);
     }
 
+    Var TypedArrayBase::GetKeysEntriesValuesHelper(Arguments& args, ScriptContext *scriptContext, LPCWSTR apiName, JavascriptArrayIteratorKind kind)
+    {
+        if (args.Info.Count == 0 || !TypedArrayBase::Is(args[0]))
+        {
+            JavascriptError::ThrowTypeError(scriptContext, JSERR_This_NeedTypedArray);
+        }
+
+        RecyclableObject* thisObj = TypedArrayBase::FromVar(args[0])->ToObject(scriptContext);
+        Assert(thisObj != nullptr);
+
+        if (TypedArrayBase::IsDetachedTypedArray(thisObj))
+        {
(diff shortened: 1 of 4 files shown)


akroshg:
@boingoing @suwc please have a look at this.

lib/Runtime/Library/TypedArray.cpp
```diff
@@ -1730,6 +1730,24 @@ namespace Js
         return JavascriptArray::CopyWithinHelper(nullptr, typedArrayBase, typedArrayBase, length, args, scriptContext);
     }
 
+    Var TypedArrayBase::GetKeysEntriesValuesHelper(Arguments& args, ScriptContext *scriptContext, LPCWSTR apiName, JavascriptArrayIteratorKind kind)
```
suwc: >Arguments&  [](start = 51, length = 11)

My impression is pointer type is now preferred over reference type...

suwc: >GetKeysEntriesValuesHelper [](start = 24, length = 26)

TypeArrayBase::ValidateTypedArray() exists for the same "ValidateTypedArray()" operation.
Do you want to use it or change it (and use it)?

akroshg: it is for consistency in the file. Look for other usage in the .h file.

---
In reply to: [79927525](https://github.com/Microsoft/ChakraCore/pull/1618#discussion_r79927525) [](ancestors = 79927525)

suwc: Sure.

---
In reply to: [79928194](https://github.com/Microsoft/ChakraCore/pull/1618#discussion_r79928194) [](ancestors = 79928194,79927525)

akroshg: I am inclined to leave this as is. It does few other operations than generic check. I am working on second iteration (which has some other fixes) which has some other changes, I can make few adjustment to that function itself. Is that okay?

---
In reply to: [79927530](https://github.com/Microsoft/ChakraCore/pull/1618#discussion_r79927530) [](ancestors = 79927530)

suwc: Sure.

suwc:
:shipit:

PR #1624, opened 2016-09-21
JsRT debugging, shim code can execute after break in user code and it may not be the top frame when stack walker happens. Top frame should be the frame on which break happened and set accordingly.

agarwal-sandeep:


agarwal-sandeep:
@akroshg please take a look

lib/Runtime/Debug/ProbeContainer.cpp
```diff
@@ -194,6 +201,7 @@ namespace Js
 
         if (pHaltState->framePointers->Count() > 0)
         {
+            pHaltState->framePointers->Peek(0)->SetIsTopFrame();
```
akroshg: instead of peeking twice, you could do pHaltState->topFrame->SetIsTopFrame();

lib/Runtime/Debug/DiagStackFrame.cpp
```diff
@@ -394,8 +398,8 @@ namespace Js
         return GetNonVarRegValue(location);
     }
 
-    DiagInterpreterStackFrame::DiagInterpreterStackFrame(InterpreterStackFrame* frame, int frameIndex) :
-        DiagStackFrame(frameIndex),
+    DiagInterpreterStackFrame::DiagInterpreterStackFrame(InterpreterStackFrame* frame) :
+        DiagStackFrame(),
```
akroshg: no explicit need of calling base class if that is default ctor, no?

akroshg:
looks good to me.

agarwal-sandeep:
Opened #1634 retargeting to master for having broader test coverage. Once the fix is validated will take in release/1.3

PR #1626, opened 2016-09-22
fix arm64 buildbreak after oopjit

MikeHolman:


lib/Common/Memory/Chakra.Common.Memory.vcxproj
```diff
@@ -45,6 +45,11 @@
       <!-- Since there are more then one XDataAllocator.cpp, we need to set them output into different directory, even when they are ExcludedFromBuild -->
       <ObjectFileName Condition="'$(Platform)'!='ARM'">$(IntDir)\arm</ObjectFileName>
     </ClCompile>
+    <ClCompile Include="$(MSBuildThisFileDirectory)arm64\XDataAllocator.cpp">
```
tcare: @curtisman does this look OK to you?

tcare: Ah this should have been for the output directory part below

rajatd:
:clock1:

satheeshravi:
:shipit:

lib/Common/Memory/arm64/XDataAllocator.cpp
```diff
@@ -3,87 +3,58 @@
...
-        Assert(pdata->UnwindData != 0);
-        Assert(pdata->BeginAddress != 0);
-        pdata->BeginAddress = (DWORD)((ULONG_PTR)pdata->BeginAddress - functionStart);
-        if(pdata->Flag != 1) // if it is not packed unwind data
-        {
-            pdata->UnwindData = (DWORD)((ULONG_PTR)pdata->UnwindData - functionStart);
-        }
-    }
-    Assert(allocation.functionTable == null);
+    __debugbreak();
```
dilijev: Why are we replacing all the implementations with a no-op/break depending on configuration?

MikeHolman: the implementations were just copy/pasted from arm32, and they would not work with arm64.

lib/Common/Memory/Chakra.Common.Memory.vcxproj
curtisman: Yeah... this looks ok.

PR #1635, opened 2016-09-23
fix xProc number allocator

leirocks:
- some fix needed because of the IDL adjustment(numberSegment changing to pointer and Arena allocated)
- delay allocate the segment only when necessary(when there's number need to be allocated in JIT time)
- added test switch to make the integration and segment full code path easier to test in oop JIT (currently 0x20 pages hold 0x2000 numbers which is extremely hard to fullfil the segment)
- make recycler memory verify work for cross process number allocator

### DIFF
--- a/lib/Backend/CodeGenNumberAllocator.cpp
+++ b/lib/Backend/CodeGenNumberAllocator.cpp
@@ -4,7 +4,7 @@
 //-------------------------------------------------------------------------------------------------------
 #include "Backend.h"
 
-
+#if !FLOATVAR
 CodeGenNumberThreadAllocator::CodeGenNumberThreadAllocator(Recycler * recycler)
     : recycler(recycler), currentNumberSegment(nullptr), currentChunkSegment(nullptr),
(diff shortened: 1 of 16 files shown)


MikeHolman:
>                             this->topFunc->GetNumberAllocator(),

can you have ifdef only for this line?

---
Refers to: lib/Backend/Inline.cpp:1190 in ecfe189. [](commit_id = ecfe189c7df07c506b073da0ff7af430384d0f42, deletion_comment = False)

leirocks:
>                             this->topFunc->GetNumberAllocator(),

No, this is in a Macro (JitAnew),  '#if' will be treated as a Macro input argument

---
In reply to: [249614624](https://github.com/Microsoft/ChakraCore/pull/1635#issuecomment-249614624) [](ancestors = 249614624)

---
Refers to: lib/Backend/Inline.cpp:1190 in ecfe189. [](commit_id = ecfe189c7df07c506b073da0ff7af430384d0f42, deletion_comment = False)

MikeHolman:
:shipit:

Issue #1627, opened 2016-09-22
Null pointer dereference issue with unsigned compare peep


Penguinwizzard:
Bug was initially externally reported. LouisL identified the issue as not being a security bug, and as being one with the unsigned compare peep:
" 
We emit:

    s13.var         =  ShrU_A         s12.var, s2.var                         #000a
    s6.var          =  ShrU_A         s4.var, s2.var                          #000e
                       ByteCodeUses   s6, s13                                 #0012
    s14.var         =  CmUnGt_A       s12.var, s4.var                         #0012

And I believe we should emit:

    s13.var         =  ShrU_A         s12.var, s2.var                         #000a
    s6.var          =  ShrU_A         s4.var, s2.var                          #000e
    s14.var         =  CmUnGt_A       s12.var, s4.var                         #0012
                       ByteCodeUses   s6, s13                                 #0012
"

Steps to reproduce:
Run the following code:

function t(f,i){
    for(var j=0;i.length;++j) {
        for(var k=0;i.length;++k){
            try{
                (f([j],[k]))
            }catch(e){}
        }
    }
}
m=function(x,y){
    (
     ((Math.co0>>>0)>(y>>>0))=ddd
    );
};
t(m,[[]])

Penguinwizzard:
Fixed in https://github.com/Microsoft/ChakraCore/commit/a27e567cc75f48db4557ada882c295aea7fd900c.

ThomsonTan:
This is an interesting case. Seems Chrome report it as ReferenceError about the left-hand side of assignment operator which is uncaught and abort run, but we only generate exception for the right-hand side about un-declared variable which could be caught.

PR #1636, opened 2016-09-23
Reorder the ByteCodeUse and Cm/Br(Sr)?[GL][te] instructions in the unsigned compare peep

Penguinwizzard:
This helps avoid situations in which the arguments to the compare op
would get optimized out too aggressively, and are then unavailable in
exceptional cases.


Penguinwizzard:
Fixes https://github.com/Microsoft/ChakraCore/issues/1627. @rajatd please review.

rajatd:
:shipit:

Issue #1549, opened 2016-09-07
[cross-platform] Improve '--no-icu' option


obastemur:
We now have `--no-icu` option to compile ChakraCore without `icu` dependency.

At the moment, ChakraCore `no-icu` option fails on unicode tests and the expectation is to make them passing.

Steps to reproduce:

```
./build.sh --no-icu --debug
test/runtests.py -d
```

obastemur:
Remaining test cases;

```
es6/unicode_idDeferParseFunctions_utf8.js
es6/classes_bugfixes.js [ 2 fails ]
es6/unicode_6_identifiers_utf8.js
es6/normalize.js
Strings/unicode_toUpperCase_toLowerCase.js
```

obastemur:
#4285 fixes `Strings/unicode_toUpperCase_toLowerCase.js` and makes `no-icu` testable.

jackhorton:
Closing this since we have --no-icu in CI now.

PR #1638, opened 2016-09-24
Implement none-unicode replacements for no-icu

obastemur:
With no-icu option, total 15 tests were failing. Now we have only 6 tests left.

Updating #1549 with the remaining tests.


digitalinfinity:
LGTM

obastemur:

PR #1641, opened 2016-09-26
Optimizating ForInObjectEnumerator overhead

curtisman:
The overhead for ForInObjectEnumerator can become significant if we enumerate a lot of object with small number of properties.

This change improves Sunspider by about 3.5ms (or 1.6%) and Speedometer by about 60ms (or 0.6%) on my machine.
No change on Octane or Kraken.

1.  Allocate ForInObjectEnumerator on the stack instead of dynamically
  - Keep track of the maximum for..in loop depth at byte code generation.
  - Change the for..in byte code to InitForInEnumerator and BrOnEmpty to refer to the for..in loop depth
  - Allocate stack space for the ForInObjectEnumerator for interpreter and JIT'ed code.
  - When bailing out of JIT'ed code, the bailout interpreter frame will not allocate new stack space for the ForInObjectEnumerator, but instead reuse the space allocated on the JIT'ed frame.
  - When we are in JIT loop body, it also won't allocate new ForInObjectEnumerator. the loop body code will instead refer to the ForInObjectEnumerator allocated in the interpreter frame.
  - In JIT inlining, we need to keep track of inlinee for..in loop depth as well to allocate enough space for all of them.

2. Reduce the size of ForInObjectEnumerator
  - Remove scriptContext and flags fields in ForInObjectEnumerator by reusing the same fields in DynamicObjectPropertyEnumerator and always initializing those.
  - Move all the fields needed to keep track of properties shadowing in prototypes into a dynamically allocated structure.  We already needed to allocate the sparse bit vector before already, so this is not a new allocation, but it will minimize stack usage for cases that doesn't have enumerable properties on the prototype.

3. Add inline cache that maps the object type to DynamicObjectPropertyEnumerator::CacheData on InitForInEnumerator
  - The cache is pass to all GetEnumerator API to be given to DynamicObjectPropertyEnumerator to avoid the dictionary look up
  - Since we clear the type to cache data map on every GC, we also clear the inline cache on every GC as well.

4. Implemented JIT fast path for OP_InitForInEnumerator and improved JIT fast path for BrOnEmpty
  - Implemented the OP_InitForInEnumerator when we have an inline cache match, with a plain dynamic object, doesn't have enumerable property on the prototype and doesn't index properties (object array)
  - Instead of checking multiple conditions, initialize a single bool on ForInObjectEnumerator canUseJITFastPath to indicate whether we can use the fast path, that combines the check for no prefix/array enumerator, no enumerable property on the prototype and have cache data
  - Avoid going to the helper before we exit the loop by adding a loop exit check for whether the cache is complete at the end of the enumeration
  - If prototype doesn't have any enumerable property, we don't have to keep track of shadowing.  So just ask the object enumerator to return enumerable property only, so that the for in object enumerator and the jit fast path doesn't have to check whether the property is enumerable or not.

5. Minor cleanup
  - Delete JavascriptEnumeratorIterator, which is no longer used.
  - Refactor code in the backend to add loop in lower


lib/Backend/Func.cpp
```diff
@@ -9,6 +9,8 @@
 #include "Base/VTuneChakraProfile.h"
 #endif
 
+#include "Library\ForInObjectEnumerator.h"
```
agarwal-sandeep: >\ [](start = 17, length = 1)

Forward slash, build failure on linux?

dilijev:
Tab Check failure is attributed to @LouisLaf. @Penguinwizzard is already fixing it in https://github.com/Microsoft/ChakraCore/pull/1636. If that is merged in first and then this PR is rebased the problem will be resolved.

Penguinwizzard:
Said tab check failure fix is now merged in.

lib/Backend/Func.cpp
curtisman: Fixed.

lib/Runtime/Library/JavascriptLibraryBase.h
```diff
@@ -168,6 +167,7 @@ namespace Js
         RuntimeFunction* uriErrorConstructor;
         RuntimeFunction* proxyConstructor;
         RuntimeFunction* promiseConstructor;
+        void * unused;
```
akroshg: why?

lib/Runtime/Types/JavascriptStaticEnumerator.cpp
```diff
@@ -49,6 +49,11 @@ namespace Js
         return this->prefixEnumerator == nullptr && this->arrayEnumerator == nullptr && this->propertyEnumerator.IsNullEnumerator();
     }
 
+    bool JavascriptStaticEnumerator::CanUseJITFastPath() const
+    {
+        return this->propertyEnumerator.CanUseJITFastPath() && this->currentEnumerator == nullptr;
```
akroshg: nit : should you check the this->currentEnumerator check first?

lib/Runtime/Base/ThreadContext.cpp
```diff
@@ -2894,6 +2896,7 @@ ThreadContext::ClearInlineCaches()
     registeredInlineCacheCount = 0;
     unregisteredInlineCacheCount = 0;
 }
+#endif //PERSISTENT_INLINE_CACHES
```
rajatd: Should we just move this #endif to above ThreadContext::ClearInvalidatedUniqueGuards ?

lib/Runtime/Language/InterpreterHandler.inl
```diff
@@ -322,8 +322,8 @@ EXDEF2    (EMPTY,                   ChkNewCallFlag,             OP_ChkNewCallFla
 EXDEF3_WMS(CUSTOM_L_Value,          NewInnerStackScFunc,        OP_NewInnerStackScFunc, ElementSlot)
 EXDEF2_WMS(GET_ELEM_SLOT_FB,        NewInnerScFunc,             ScriptFunction::OP_NewScFunc)
 EXDEF2_WMS(GET_ELEM_SLOT_FB,        NewInnerScGenFunc,          JavascriptGeneratorFunction::OP_NewScGenFunc)
-  DEF2_WMS(A1toA1MemNonVar,         GetForInEnumerator,         JavascriptOperators::OP_GetForInEnumerator)
-  DEF3_WMS(A1toXXMemNonVar,         ReleaseForInEnumerator,     JavascriptOperators::OP_ReleaseForInEnumerator, ForInObjectEnumerator *)
+  DEF2_WMS(A1U1toXX,                InitForInEnumerator,        OP_InitForInEnumerator)
+  DEF2_WMS(A1U1toXXWitCache,        ProfiledInitForInEnumerator,OP_InitForInEnumeratorWithCache)
```
rajatd: >A1U1toXXWitCache [](start = 11, length = 16)

change Wit to With.

lib/Runtime/Types/DynamicObjectPropertyEnumerator.cpp
```diff
@@ -203,12 +266,12 @@ namespace Js
 
     Var DynamicObjectPropertyEnumerator::MoveAndGetNext(PropertyId& propertyId, PropertyAttributes * attributes)
     {
+        if (this->cachedData && this->initialType == this->object->GetDynamicType())
```
rajatd: Did this->initialType always use to have the same value as this->cachedDataType?

lib/Runtime/Language/InterpreterStackFrame.cpp
```diff
@@ -8438,6 +8461,23 @@ const byte * InterpreterStackFrame::OP_ProfiledLoopBodyStart(const byte * ip)
         return m_inParams + 1;
     }
 
+    ForInObjectEnumerator * InterpreterStackFrame::GetForInEnumerator(uint index)
```
rajatd: >index [](start = 75, length = 5)

This argument really means the for..in loop level, right? If so, we might be better off calling it level instead of index (less confusion to readers). (would also have to change names of params in other methods from forInLoopId to forInLoopLevel)

lib/Backend/IRBuilder.cpp
```diff
@@ -2820,16 +2860,8 @@ IRBuilder::BuildProfiledReg1Unsigned1(Js::OpCode newOpcode, uint32 offset, Js::R
     IR::IntConstOpnd * srcOpnd;
 
     srcOpnd = IR::IntConstOpnd::New(value, TyInt32, m_func);
-    if (m_func->DoSimpleJitDynamicProfile())
-    {
-        instr = IR::JitProfilingInstr::New(newOpcode, dstOpnd, srcOpnd, m_func);
-        instr->AsJitProfilingInstr()->profileId = profileId;
-    }
-    else
```
rajatd: Why did you remove this code?

lib/Backend/IR.h
```diff
@@ -625,6 +625,11 @@ class LabelInstr : public Instr
     BYTE                    isOpHelper : 1;
     BYTE                    m_hasNonBranchRef : 1;
     BYTE                    m_isDataLabel : 1;
+
+    // Indicate whether the label is the target of a for in loop exist (BrOnEmpty or BrOnNotEmpty)
```
rajatd: >exist [](start = 65, length = 5)

change exist to exit

lib/Runtime/SerializableFunctionFields.h
```diff
@@ -71,6 +71,7 @@ PRIVATE_FIELDS
     DECLARE_SERIALIZABLE_FIELD(bool, m_hasFinally, Bool);
     DECLARE_SERIALIZABLE_FIELD(bool, hasScopeObject, Bool);
     DECLARE_SERIALIZABLE_FIELD(bool, hasCachedScopePropIds, Bool);
+    DECLARE_SERIALIZABLE_ACCESSOR_FIELD(uint, ForInLoopDepth, UInt32);
```
rajatd: >UInt32 [](start = 62, length = 6)

ForInLoopDepth will be smaller than or equal to ProfiledForInLoopCount. Why are we giving a higher type e to ForInLoopDepth?

lib/Backend/BailOut.cpp
```diff
@@ -252,10 +251,19 @@ BailOutInfo::FinalizeBailOutRecord(Func * func)
             }
             // The locals size contains the inlined-arg-area size, so remove the inlined-arg-area size from the
             // adjustment for normal locals whose offsets are relative to the start of the locals area.
-            offset -= (localsSize - inlineeArgStackSize);
+            offset -= (inlinedArgSlotAdjust - inlineeArgStackSize);
```
rajatd: >inlinedArgSlotAdjust [](start = 23, length = 20)

inlinedArgSlotAdjust may not have the same value as localsSize for arm

lib/Backend/Peeps.cpp
```diff
@@ -303,7 +303,7 @@ Peeps::PeepAssign(IR::Instr *assign)
     IR::Instr *instrNext = assign->m_next;
 
     // MOV reg, sym
-    if (src->IsSymOpnd())
+    if (src->IsSymOpnd() && src->AsSymOpnd()->m_offset == 0)
```
rajatd: >src->AsSymOpnd()->m_offset == 0 [](start = 28, length = 31)

I didn't get what this check is for.

rajatd:
@curtisman , done looking at the changes. Let me know what you think of the comments/questions

lib/Runtime/Types/JavascriptStaticEnumerator.cpp
curtisman: currentEnumerator is no less likely to be non-null, so I don't think it matters.

lib/Runtime/Library/JavascriptLibraryBase.h
curtisman: It was only there for testing.  I will remove it.

lib/Runtime/Language/InterpreterStackFrame.cpp
curtisman: Changed.

lib/Backend/IR.h
curtisman: Fixed

lib/Backend/Lower.cpp
```diff
@@ -23001,6 +22972,242 @@ Lowerer::LoadIndexFromLikelyFloat(
...
+    InsertMove(typeOpnd, IR::IndirOpnd::New(objectOpnd, Js::DynamicObject::GetOffsetOfType(), TyMachPtr, func), instr);
+    InsertCompareBranch(IR::MemRefOpnd::New(&forInCache->type, TyMachPtr, func, IR::AddrOpndKindForInCacheType), typeOpnd, Js::OpCode::BrNeq_A, helperLabel, instr);
+    
+    // Check forInCacheData->EnumNonEnumerable == false  
+    //
+    //   MOV forInCacheDataOpnd, [&forInCache->data]
+    //   CMP forInCacheDataOpnd->enumNonEnumerable, 0
+    //   JNE $helper
+
+    IR::RegOpnd * forInCacheDataOpnd = IR::RegOpnd::New(TyMachPtr, func);
```
curtisman: No..  FI the forInCache->type manages the object's type, that means the type can't be null, therefore, the data can't be null either.

lib/Backend/Lower.cpp
```diff
@@ -23001,6 +22972,242 @@ Lowerer::LoadIndexFromLikelyFloat(
...
+        
+    // Check has object array
+    GenerateHasObjectArrayCheck(objectOpnd, typeOpnd, helperLabel, instr);
+
+    // Check first prototype with enumerable properties
+    //
+    //   MOV prototypeObjectOpnd, [type + offset(prototype)]
+    //   MOV prototypeTypeOpnd, [prototypeObjectOpnd + offset(type)]
+    //   CMP [prototypeTypeOpnd + offset(typeId)], TypeIds_Null
+    //   JEQ $noPrototypeWithEnumerablePropertiesLabel
```
curtisman: No.. we want to check if there are any prototype that has enumerable property.  If it has, then we need to go to the helper.  That label is for when we found no prototype that has enumerable property.

lib/Backend/Peeps.cpp
curtisman: A sym operand can refer to different offset of a stack sym.   The peep can only track one offset because the reg is store on the stack sym itself.  So we don't only peep if the sym operand referencing the beginning of the stack sym.

lib/Backend/BailOut.cpp
curtisman: Well  this is in the MD_GROW_LOCALS_AREA_UP
`uint inlinedArgSlotAdjust = (func->m_localStackHeight + func->m_ArgumentsOffset);`
`int localsSize = func->m_localStackHeight + func->m_ArgumentsOffset;`

Why can't they be substituted?

lib/Runtime/SerializableFunctionFields.h
curtisman: ForInLoopDepth is the real depth.  ProfiledForInLoopCount is restrict to uint16 depth.   We will stop generating profiled byte code for depth > uint16.

lib/Backend/Lower.cpp
rajatd: Ok, got it.

---
In reply to: [81577863](https://github.com/Microsoft/ChakraCore/pull/1641#discussion_r81577863) [](ancestors = 81577863)

lib/Backend/Lower.cpp
rajatd: Got it.

---
In reply to: [81578312](https://github.com/Microsoft/ChakraCore/pull/1641#discussion_r81578312) [](ancestors = 81578312)

dilijev:
nit: Title of PR / Commit message "optimizating" -> "optimizing"

lib/Backend/BailOut.cpp
rajatd: Sorry, my bad. I missed that macro.

---
In reply to: [81579468](https://github.com/Microsoft/ChakraCore/pull/1641#discussion_r81579468) [](ancestors = 81579468)

lib/Runtime/SerializableFunctionFields.h
rajatd: Oh ok. Thanks

---
In reply to: [81579788](https://github.com/Microsoft/ChakraCore/pull/1641#discussion_r81579788) [](ancestors = 81579788)

lib/Backend/IRBuilder.cpp
curtisman: I have reverted this change.

lib/Runtime/Language/InterpreterHandler.inl
curtisman: Fixed.

rajatd:
:shipit:

curtisman:

curtisman:
@dotnet-bot Ubuntu ubuntu_linux_test

curtisman:

curtisman:

PR #1642, opened 2016-09-26
cross-platform: Enable intrinsics for Math.ceil

obastemur:
We no longer / do not have any reason to keep this part disabled on xplat.


obastemur:

PR #1643, opened 2016-09-26
use target inlinee runtime info when building JIT data

MikeHolman:
I was assuming that structure of inlinee JIT data and runtime data would be same at codegen time. However, runtime data can become more polymorphic between WI creation and codegen time, so we should select the runtime data corresponding to what we built at WI creation.


rajatd:
:shipit:

PR #1646, opened 2016-09-27
Passing proper property operation flag during delete operation.

akroshg:
On many Array methods, as per spec, we need to throw when we try to delete
the non-configurable property. Earlier we were passing wrong flag
PropertyOperation_ThrowIfNotExtensible. Fixed that by passing the correct
flag and throw when that happens.
Added few tests for reference.


akroshg:
@pleath @rajatd @suwc - please have a look.

pleath:
LGTM. Thanks.

akroshg:
Thanks @pleath

PR #1647, opened 2016-09-27
Remove CSE for BailOnNotStackArgs

ThomsonTan:
Lowerer relies on LoadHeapArguments to be followed by BailOnNotStackArgs to ensure stackargs option is consistent between inliner and inlinee. When the same function inlined twice in the same inliner (because it is called twice in sequence), CSE will eliminate the second BailOnNotStackArgs which breaks the assumption for lowerer. This fix disable CSE for BailOnNotStackArgs like the workaround for hoist case.

### DIFF
--- a/lib/Backend/Inline.cpp
+++ b/lib/Backend/Inline.cpp
@@ -2307,7 +2307,7 @@ IR::Instr * Inline::InlineApplyWithArgumentsObject(IR::Instr * callInstr, IR::In
     }
 
     // set src1 to avoid CSE on BailOnNotStackArgs for different arguments object
-    bailOutOnNotStackArgs->SetSrc1(ldHeapArguments->GetDst());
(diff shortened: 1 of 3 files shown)


ThomsonTan:
@satheeshravi @rajatd please take a look.

rajatd:
Oh it is cse'd between two call sites of the same inlinee. I think we can do better than disabling cse for BailOnNotStackArgs. There's already code that kills live "arguments" and "caller" fields to prevent copy prop of inlinee.arguments and inlinee.caller between two call sites of an inlinee. We should look at doing something similar to prevent this buggy cse

ThomsonTan:
That's right. Added an arguments related object to src1 to prevent buggy CSE from happening.

---
In reply to: [249762936](https://github.com/Microsoft/ChakraCore/pull/1647#issuecomment-249762936) [](ancestors = 249762936)

satheeshravi:
:shipit:

PR #1649, opened 2016-09-27
Bug fix for paramSlot change (louisl)

satheeshravi:
Bug fix for Louis's paramSlot change

Issue:
Param Slot numbers were changed for generator, yield objects and other
implicit params for x86 and x64. Apparently some of the changes are in
Lower.cpp which also demands changes in arm32(which is not the case
currently).

So the param slot numbers in arm32 are not reflecting the new numbering
logic. This results in overlap of param slot numbers for different
params.

Fix:
Actual fix is to change the numbering pattern in arm. But since Louis had
other fixes pending for his original change. I am just disabling ARM to
use the new API that uses new numbering pattern - to relax Exprgen from
getting high bug hits.


satheeshravi:
@rajatd can you please review? 
//cc: @LouisLaf

rajatd:
LGTM.

PR #1650, opened 2016-09-27
fix xdata issues on arm

MikeHolman:
We were masking the code address for thumb mode before registering the unwind data with the OS, which causes unwinding to fail.

Also, for OOP JIT specifically, when we have expanded unwind data we need to fixup the .xdata pointer. Also fixed an issue I recently introduced which was causing bad alignment for everything in NativeDataBuffer (necessary here as .xdata is required to be 4 byte aligned).

### DIFF
--- a/lib/Backend/Encoder.cpp
+++ b/lib/Backend/Encoder.cpp
@@ -323,8 +323,6 @@ Encoder::Encode()
         m_func->GetThreadContextInfo()->GetProcessHandle());
 #elif _M_ARM
     m_func->m_unwindInfo.EmitUnwindInfo(m_func->GetJITOutput(), alloc);
-    m_func->GetJITOutput()->SetCodeAddress(m_func->GetJITOutput()->GetCodeAddress() | 0x1); // Set thumb mode
-
     if (m_func->IsOOPJIT())
     {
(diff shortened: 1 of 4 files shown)


pleath:
:shipit:

satheeshravi:
:shipit:

lib/Backend/Encoder.cpp
```diff
@@ -337,6 +335,8 @@ Encoder::Encode()
         XDataAllocator::Register(&alloc->allocation->xdata, m_func->GetJITOutput()->GetCodeAddress(), m_func->GetJITOutput()->GetCodeSize());
         m_func->GetInProcJITEntryPointInfo()->SetXDataInfo(&alloc->allocation->xdata);
     }
+
+    m_func->GetJITOutput()->SetCodeAddress(m_func->GetJITOutput()->GetCodeAddress() | 0x1); // Set thumb mode
```
ThomsonTan: Why do we need set it after RecordXData?

MikeHolman: Sorry, my initial description was not quite right. The issue is not with the xdata itself, but with registration of the xdata with the OS. The registration API doesn't want the thumb encoded address.

MikeHolman:
This fixes the in-proc case, but turns out that oop case still needs a bit more.

MikeHolman:
I added some code here for doing fixup when there is expanded unwind data. @leirocks, can you take another look at that code?

PR #1651, opened 2016-09-28
IEEE754 Perf Improvement

obastemur:
We don't know whether BHL or BLH is the correct number
When nDigits is set(>=0), that means the consumer is not
interested with the entire value.
That means, we may not just roundup to the higher bound.
Doing so would break IEEE754
Imagine LowerBound is 4999... while the UpperBoud is 50001..
We need to know where actually the number is instead of picking
either Lower or UpperBoud
In this case, we should skip and fail
and let FDblToRgbPrecise do the math.

Perf Improvement [new]:
Exception for a number that has smaller amount of fractional part
than the number of digits are being asked.


obastemur:

obastemur:

PR #1652, opened 2016-09-28
cross-platform: Improve compiler options

obastemur:
- make PAL specific flags project wide (i.e. fstack-protector-all)
- force Wc++11-narrowing (similar to VC++)
- disable warning reporting from system headers -wsystem-headers [node-chakracore issue]
- set -gdwarf-4 instead of gdwarf-3 as it is the currently supported version on GCC targets
- set -gdwarf-4 only for debug builds

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -208,6 +208,8 @@ if(CLR_CMAKE_PLATFORM_XPLAT)
 
     # CXX WARNING FLAGS
     set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} \
+        -Wno-system-headers\
+        -Wc++11-narrowing\
         -Wno-ignored-attributes\
         -Wno-deprecated-declarations\
         -Wno-parentheses-equality\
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -240,6 +242,9 @@ if(CLR_CMAKE_PLATFORM_XPLAT)
         # -Wno-return-type # switch unreachable code
         # -Wno-switch # switch values not handled
 
+    add_compile_options(-fexceptions)
+    add_compile_options(-fstack-protector-all)
```
curtisman: Do we really want all? or should -fstack-protector sufficient?
What's the perf impact with -fstack-protector-all?

obastemur: > What's the perf impact with -fstack-protector-all?

Comparing to before ? Almost, none. PAL was enabling that and we depend on PAL. In general ? i.e. can we disable it to get more perf ? We `maybe` could (since this is the version with -all flag, it will be adding on everywhere) but this is already being done in an optimized way. 

What we already have + [add more space for stack][1/2] + check/compare[1] + cleanup[3] + stack_guard?

My guess, up to 5/8 additional instructions per call (worst case), 5/8 additional inst. vs an average CC function, it is less than 1 percent ?

My earlier PR,  I had enabled only `-fstack-protector` for similar reasons, but IMO after your comment to that PR `all` makes more sense for CC. I am not sure if we have lots of critical functions under the threshold of `should we include this too`?

curtisman: Ok... when we are ready to look at perf, this is an area to measure to see if it actually make a difference or not.

obastemur: > Ok... when we are ready to look at perf, this is an area to measure to see if it actually make a difference or not.

will add a todo item there

PR #1653, opened 2016-09-28
cross-platform: Fix heap overflow - MiscGetenv

obastemur:
strlen(palEnvironment[i]) can be smaller than name

This is fixed on coreclr repo

### DIFF
--- a/pal/src/cruntime/misc.cpp
+++ b/pal/src/cruntime/misc.cpp
@@ -412,6 +412,18 @@ void MiscCleanup(void)
      InternalDeleteCriticalSection(&gcsEnvironment);
 }
 
+int compare(const char *left, const char *right, int *last_index)
+{
+    int i;
+    int result = 0;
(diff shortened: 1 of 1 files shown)


pal/src/cruntime/misc.cpp
```diff
@@ -412,6 +412,18 @@ void MiscCleanup(void)
      InternalDeleteCriticalSection(&gcsEnvironment);
 }
 
+int compare(const char *left, const char *right, int *last_index)
+{
+    int i;
+    int result = 0;
+    for(i = 0; left[i] != 0 && result == 0 && i < INT_MAX; i++)
+    {
+        result = (left[i] == right[i]) ? 0 : (left[i] < right[i] ? -1 : 1);
```
curtisman: Can't you exit early if  they are not equal?

obastemur:

PR #1654, opened 2016-09-28
tell legalize we're post-reg alloc on LinearScanMD::InsertLea for arm

MikeHolman:
Sometimes offset would be hoisted into generic reg opnd (now it will be hoisted into scratch reg)


rajatd:
:shipit:

pleath:
LGTM.

PR #1655, opened 2016-09-28
TTD -- Bug fix FunctionBody lookup and Assert condition.

mrkmarron:


mrkmarron:
@agarwal-sandeep please have a look.

PR #1657, opened 2016-09-28
Constrain conv_bool optimization to improve correctness.

Penguinwizzard:
Due to my recent change to the unsigned compare peep, the conv_bool
optimization was causing instructions to be generated in a way that
the some registers could have two overlapping lifetimes. This fixes
the issue by checking for these ByteCodeUses instructions, and then
placing the conv_bool after them.

### DIFF
--- a/lib/Backend/GlobOpt.cpp
+++ b/lib/Backend/GlobOpt.cpp
@@ -12257,7 +12257,19 @@ LOutsideSwitch:
 
                 varDst = IR::RegOpnd::New(intDst->m_sym->GetVarEquivSym(this->func), TyVar, this->func);
                 IR::Instr *convBoolInstr = IR::Instr::New(Js::OpCode::Conv_Bool, varDst, intDst, this->func);
-                instr->InsertAfter(convBoolInstr);
+                // In some cases (e.g. unsigned compare peep code), a comparison will use variables
+                // other than the ones initially intended for it, if we can determine that we would
+                // arrive at the same result. This means that we get a ByteCodeUses operation after
+                // the actual comparison. Since Inserting the Conv_bool just after the compare, and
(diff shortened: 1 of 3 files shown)


Penguinwizzard:
@dilijev could you please review?

lib/Backend/GlobOpt.cpp
```diff
@@ -12257,7 +12257,19 @@ GlobOpt::TypeSpecializeBinary(IR::Instr **pInstr, Value **pSrc1Val, Value **pSrc
...
                 IR::Instr *convBoolInstr = IR::Instr::New(Js::OpCode::Conv_Bool, varDst, intDst, this->func);
-                instr->InsertAfter(convBoolInstr);
+                // In some cases (e.g. unsigned compare peep code), a comparison will use variables
+                // other than the ones initially intended for it, if we can determine that we would
+                // arrive at the same result. This means that we get a ByteCodeUses operation after
+                // the actual comparison. Since Inserting the Conv_bool just after the compare, and
+                // just before the ByteCodeUses, would cause issues later on with register lifetime
+                // calculation, we want to insert the Conv_bool after the whole compare instruction
+                // block.
+                IR::Instr *putafter = instr;
```
dilijev: >putafter [](start = 27, length = 8)

nit: camelCase

dilijev:
Had some discussion offline to fully understand this change. AFAICT it does what's advertised. I'd run it by someone who is more familiar with GlobOpt to make sure there are no other cases you might be missing.

dilijev:
If there was a bug you are fixing with this change, distill it into a minimal repro test case and add it to the tests -- otherwise add a test if you can.

dilijev:
LGTM :shipit:

Penguinwizzard:

PR #1659, opened 2016-09-29
Fix buffer overflow of a global variable

obastemur:
`memcmp` was reaching beyond the global variable ~~heap~~ buffer.

`cch` does not include `sizeof(char16)` while `pszs->cch` includes that. See https://github.com/Microsoft/ChakraCore/blob/master/lib/Runtime/Library/DateImplementation.cpp#L1199-L1200

### DIFF
--- a/lib/Runtime/Library/DateImplementation.cpp
+++ b/lib/Runtime/Library/DateImplementation.cpp
@@ -47,7 +47,7 @@ namespace Js {
 
     const static SZS g_rgszs[] =
     {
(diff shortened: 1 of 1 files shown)


akroshg:
do you have repro for this?

obastemur:

obastemur:
> do you have repro for this?

@akroshg I was reading the code + making some analysis and reached out to this issue. I'm not sure giving a clean path to a repro is a good idea :)

obastemur:
@akroshg @jianchun Thanks for review

PR #1660, opened 2016-09-29
For-of fix for ToObject

akroshg:
The expression evaluated on for..of head should be an Object, so it should
follow the ToObject rule, where it will throw when the object is
undefined/null.

### DIFF
--- a/lib/Runtime/ByteCode/ByteCodeEmitter.cpp
+++ b/lib/Runtime/ByteCode/ByteCodeEmitter.cpp
@@ -9219,6 +9219,11 @@ void EmitForInOrForOf(ParseNode *loopNode, ByteCodeGenerator *byteCodeGenerator,
         return;
     }
 
+    Js::ByteCodeLabel skipThrow = byteCodeGenerator->Writer()->DefineLabel();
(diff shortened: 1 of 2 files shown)


akroshg:
@suwc @tcare - please have a look.

lib/Runtime/ByteCode/ByteCodeEmitter.cpp
```diff
@@ -9219,6 +9219,11 @@ void EmitForInOrForOf(ParseNode *loopNode, ByteCodeGenerator *byteCodeGenerator,
         return;
     }
 
+    Js::ByteCodeLabel skipThrow = byteCodeGenerator->Writer()->DefineLabel();
+    byteCodeGenerator->Writer()->BrReg2(Js::OpCode::BrNeq_A, skipThrow, loopNode->sxForInOrForOf.pnodeObj->location, funcInfo->undefinedConstantRegister);
```
tcare: So we can skip the null check because the pnodeObj will be a object?

akroshg: `BrNeq_A` is not strict check. So it will check both undefined and null. Is that what you asked?

tcare: Yep ok makes sense :)

tcare:
Looks good

akroshg:
@dotnet-bot test this please

akroshg:
Thanks @tcare.

PR #1661, opened 2016-09-29
Fix regression in gameboy

kunalspathak:
Fix regression in gameboy

My earlier change to ChangeIfTypeIsEquivalent introduced regression in gameboy benchmark. Earlier if equivalent type cache was full, we evicted existing type from cache to store new type we saw. In https://github.com/Microsoft/ChakraCore/pull/1578 I changed it to say "if i see non-shared type and cache is full, then treat the type as non-equivalent and perform helper path to Ld/St operation". This leads to lot of repeating non-shared types to perform operation using helper path or we bailout. Changed the heuristics to evict non-shared type (if present) if we see non-shared type. This fixes the regression.

### DIFF
--- a/lib/Runtime/Language/JavascriptOperators.cpp
+++ b/lib/Runtime/Language/JavascriptOperators.cpp
@@ -8277,19 +8277,6 @@ CommonNumber:
             return true;
         }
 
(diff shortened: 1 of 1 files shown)


rajatd:
:shipit:

kunalspathak:
Thanks @rajatd  for the review!

PR #1664, opened 2016-09-29
Fixing a bug in FreeLoopBodyJobManager::GetJob

rajatd:
`FreeLoopBodyJobManager::GetJob` first checks whether `FreeLoopBodyJobManager::waitingForStackJob` is true, if yes, then it returns the passed in job if `FreeLoopBodyJobManager::stackJobProcessed` is false, otherwise return `nullptr`. If `waitingForStackJob` is false, the method just returns the passed in job. However, waitingForStackJob is set to false in
`FreeLoopBodyJobManager::JobProcessed`, which makes the first check in `GetJob` fail and it returns
the job. We later try to move the job to the beginning of the jobs queue but since the job was already processed, it is not present in the queue.

Fix is to have `GetJob` check for a stack allocated job instead of checking for `waitingForStackJob`.


rajatd:
@MikeHolman could you review?

MikeHolman:
:shipit:

PR #1670, opened 2016-09-30
Do not reuse by resetting the JavascriptStaticEnumerator in the ForInObjectEnumerator

curtisman:
Even if the type match, the object may have change to have array index, producing wrong result.
Since we now have an instance of the DynamicObjectPropertyEnumerator in the JavascriptStaticEnumerator already, reset doesn't really gain us much.  So just remove it.


curtisman:
@rajatd  please review.

agarwal-sandeep:
:shipit:

PR #1673, opened 2016-09-30
Fix issue with unsigned compare peep in dead code.

Penguinwizzard:
The unsigned compare peep optimization, after being changed to have
the bytecodeuses instruction emitted after the compare instead of before
it, had issues with lowering of code hidden behind a bailonnoprofile, in
that it would have conflicts on lifetimes due to the loss of atomicity
of the bytecodereg use and generation of the compare. This patch attempts
to restore that by aggregating the bytecodeuse information for a single
instruction into once place.

### DIFF
--- a/lib/Backend/GlobOptBailOut.cpp
+++ b/lib/Backend/GlobOptBailOut.cpp
@@ -1085,6 +1085,43 @@ GlobOpt::ConvertToByteCodeUses(IR::Instr * instr)
     this->CaptureByteCodeSymUses(instr);
     IR::ByteCodeUsesInstr * byteCodeUsesInstr = this->InsertByteCodeUses(instr, true);
     instr->Remove();
+    // If possible, we want to aggregate with subsequent ByteCodeUses Instructions, so
+    // that we can do some optimizations in other places where we can simplify args in
+    // a compare, but still need to generate them for bailouts. Without this, we cause
+    // problems because we end up with an instruction losing atomicity in terms of its
+    // bytecode use and generation lifetimes.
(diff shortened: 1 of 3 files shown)


rajatd:
@LouisLaf

Penguinwizzard:

lib/Backend/GlobOptBailOut.cpp
```diff
@@ -1085,6 +1085,43 @@ GlobOpt::ConvertToByteCodeUses(IR::Instr * instr)
...
+                    byteCodeUsesInstr->byteCodeUpwardExposedUsed->Or(nextinstr->byteCodeUpwardExposedUsed);
+                    JitAdelete(nextinstr->byteCodeUpwardExposedUsed->GetAllocator(), nextinstr->byteCodeUpwardExposedUsed);
+                    nextinstr->byteCodeUpwardExposedUsed = nullptr;
+                }
+                else
+                {
+                    byteCodeUsesInstr->byteCodeUpwardExposedUsed = nextinstr->byteCodeUpwardExposedUsed;
+                    nextinstr->byteCodeUpwardExposedUsed = nullptr;
+                }
+            }
```
LouisLaf: Why not delete nextinstr at this point?

Penguinwizzard: The code in RemoveCodeAfterNoFallthroughInstr has already referenced the next instruction by the time that we do this. Doing nextinstr->Remove() causes an access violation a couple frames up the call stack if you remove the instruction here.

PR #1674, opened 2016-09-30
Initial redeferral heuristic

pleath:
Implement a basic heuristic for redeferral, based on experimental work that Rajat did a while back. The heuristic uses 5 constants: an initial delay, which is the number of GC's we perform before we start checking for redeferral candidates; the number of GC's to wait between redeferral checks while we're in the startup phase; the number of GC's that must have passed since a function was executed before we redefer it in the startup phase; and the same wait and inactivity constants for the main phase, to which we transition after the startup phase ends. Todo: emit jitted code to reset a function's inactivity count when the function is executed, and/or avoid redeferring a jitted function if we think it may be executed again.


rajatd:
@pleath you might want to change your base branch to master

pleath:
Out-of-date PR, incorporated into https://github.com/Microsoft/ChakraCore/pull/1585.

PR #1675, opened 2016-10-01
fix bug with trying to use full JS strings from JIT

MikeHolman:
For OOP JIT, we only marshal across the necessary bits we need from RecyclableObjects, in particular for JavascriptStrings we only marshal the typeId, buffer, and length.
We then need to make sure to only use the bits we have available.

There was a place where we do FromVar, which under an assert calls type->CanHaveInterceptors(). And that is data we don't have, so it reads some invalid memory from the RPC buffer.

To make this all much safer (and fix the assert) I have added a new JITJavascriptString class which contains only the fields we have, and replaced all backend string usage with this. This also gave me the opportunity to enable a couple of places where we were being cautious and disabling string optimizations for OOP JIT (these optimizations were pretty minor, so OOP JIT benchmarks are flat).

### DIFF
--- a/lib/Backend/JITRecyclableObject.h
+++ b/lib/Backend/JITRecyclableObject.h
@@ -46,19 +46,13 @@ public:
 
     static JITJavascriptString * JITJavascriptString::FromVar(Js::Var var)
     {
-        JITJavascriptString * jitString = reinterpret_cast<JITJavascriptString*>(var);
-#if DBG
-        Assert(Is(var));
-        Assert(jitString->GetTypeId() == Js::TypeIds_String);
-        if (!JITManager::GetJITManager()->IsOOPJITEnabled())
-        {
-            Js::JavascriptString * fullString = Js::JavascriptString::FromVar(var);
-            // ensure layouts are same
-            Assert(fullString->GetLength() == jitString->GetLength());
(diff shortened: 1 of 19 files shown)


lib/Backend/JITRecyclableObject.h
```diff
@@ -0,0 +1,84 @@
...
+        JITJavascriptString * jitString = reinterpret_cast<JITJavascriptString*>(var);
+#if DBG
+        Assert(Is(var));
+        Assert(jitString->GetTypeId() == Js::TypeIds_String);
+        if (!JITManager::GetJITManager()->IsOOPJITEnabled())
+        {
+            Js::JavascriptString * fullString = Js::JavascriptString::FromVar(var);
+            // ensure layouts are same
+            Assert(fullString->GetLength() == jitString->GetLength());
+            Assert(wmemcmp(fullString->GetString(), jitString->GetString(), jitString->GetLength()) == 0);
```
leirocks: I don't think this necessary, it's just comparing the same memory, it slows down chk build. if you want you can add static assert for the layout concern.

MikeHolman: True, this should be static assert.

PR #1676, opened 2016-10-01
Minor fix for prefast warnings

curtisman:
Fix wrong sal annotation
Change to use explicit buffer size.

### DIFF
--- a/lib/Runtime/Types/DictionaryTypeHandler.cpp
+++ b/lib/Runtime/Types/DictionaryTypeHandler.cpp
@@ -2742,7 +2742,7 @@ namespace Js
             //
 
             Js::PropertyId pid = iter.CurrentKey()->GetPropertyId();
-            if(!DynamicTypeHandler::ShouldMarkPropertyId_TTD(pid) | (!descriptor.IsInitialized) | (descriptor.Attributes & PropertyDeleted))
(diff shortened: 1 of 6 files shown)


lib/Runtime/Types/DictionaryTypeHandler.cpp
```diff
@@ -2742,7 +2742,7 @@ namespace Js
             //
 
             Js::PropertyId pid = iter.CurrentKey()->GetPropertyId();
-            if(!DynamicTypeHandler::ShouldMarkPropertyId_TTD(pid) | (!descriptor.IsInitialized) | (descriptor.Attributes & PropertyDeleted))
+            if(!DynamicTypeHandler::ShouldMarkPropertyId_TTD(pid) || (!descriptor.IsInitialized) || (descriptor.Attributes & PropertyDeleted))
```
dilijev: @mrkmarron Had some rationale for using `|` instead of `||`. I wonder if he has any comments on this change.

mrkmarron: This is a very hot loop for the TTD snapshot (we run this over every object in the heap on every snapshot). So, I wanted to avoid the extra branching required by "||" and let the compiler optimize the bit masking/logic ops with "|" as straight-line code.

curtisman: 'k.  I will revert it back and just put paren around the first condition.

PR #1677, opened 2016-10-01
remove invalid assert for jit page state

MikeHolman:
This assert is outside the lock, so if another JIT is happening at same time, the page can become read/write/execute


MikeHolman:

digitalinfinity:
LGTM to me although if the assert was moved within the lock, would it then be legitimate?

MikeHolman:
@digitalinfinity there actually already is an assert for this inside the lock.

digitalinfinity:
Ah excellent- thanks for checking!

PR #1681, opened 2016-10-03
If we have no user crash handler we should generate additional info.

Penguinwizzard:
The previous way we were doing things, we'd only generate the additional
backtrace information about arguments and the like when an exception was
_totally_ uncaught. This interfered with our implicit autocatchhandlers,
which were added for some instructions. This caused divergences with the
nonative version of the debug output, where we wouldn't add these, which
caused us to only generate the additional backtrace information in a few
of the test configurations, but not all of them. By allowing creation of
the additional backtrace information when we have a catch handler but no
user catch handler, we handle these cases more uniformly.

This does, however, mean that our information on whether or not there is
a user catch handler; the existing code does this to some degree, but it
is polluted by a bunch of implied catch handlers in both interpreter and
in helper code. I added an argument to autocatchhandler to specify if we
can mark a particular instance of the autocatchhandler as user-made; the
values that I've chosen for it (i.e. where it is true) should be checked
and verified. Note that if the argument is true, that only means that it
_can_ be a user catch handler; if it's not in user code (e.g. in chakra-
generated JS) then it still won't be marked as a user catch handler.

### DIFF
--- a/lib/Runtime/Language/JavascriptExceptionOperators.cpp
+++ b/lib/Runtime/Language/JavascriptExceptionOperators.cpp
@@ -17,14 +17,13 @@ namespace Js
     {
         Assert(scriptContext);
 
-        bool fFound = false;
         // If the outer try catch was already in the user code, no need to go any further.
         if (!m_previousCatchHandlerToUserCodeStatus)
         {
             if (!isUserExceptionHandling)
             {
-                fFound = true;
                 m_threadContext->SetIsUserCode(false);
+                return;
             }
             else
             {
@@ -35,17 +34,14 @@ namespace Js
                     if (caller != NULL && (funcBody = caller->GetFunctionBody()) != NULL)
                     {
                         m_threadContext->SetIsUserCode(funcBody->IsNonUserCode() == false);
-                        fFound = true;
+                        return;
                     }
                 }
             }
         }
 
-        if (!fFound)
-        {
-            // If not successfully able to find the caller, set this catch handler belongs to the user code.
-            m_threadContext->SetIsUserCode(true);
-        }
+        // If not successfully able to find the caller, set this catch handler belongs to the user code.
+        m_threadContext->SetIsUserCode(true);
     }
 
     JavascriptExceptionOperators::AutoCatchHandlerExists::AutoCatchHandlerExists(ScriptContext* scriptContext, bool isUserExceptionHandling)
--- a/lib/Runtime/Language/JavascriptExceptionOperators.cpp
+++ b/lib/Runtime/Language/JavascriptExceptionOperators.cpp
@@ -13,7 +13,7 @@ extern "C" PVOID __guard_check_icall_fptr;
 
 namespace Js
 {
-    void JavascriptExceptionOperators::AutoCatchHandlerExists::FetchNonUserCodeStatus(ScriptContext * scriptContext)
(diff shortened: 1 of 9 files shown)


lib/Runtime/Language/JavascriptExceptionOperators.cpp
```diff
@@ -13,22 +13,30 @@ extern "C" PVOID __guard_check_icall_fptr;
...
         // If the outer try catch was already in the user code, no need to go any further.
         if (!m_previousCatchHandlerToUserCodeStatus)
         {
-            Js::JavascriptFunction* caller;
-            if (JavascriptStackWalker::GetCaller(&caller, scriptContext))
+            if (!isUserExceptionHandling)
             {
-                Js::FunctionBody *funcBody = NULL;
-                if (caller != NULL && (funcBody = caller->GetFunctionBody()) != NULL)
+                fFound = true;
```
agarwal-sandeep: >fFound = true; [](start = 16, length = 14)

nit: We actually don't need fFound and can just return.

lib/Runtime/Language/JavascriptExceptionOperators.cpp
```diff
@@ -62,7 +67,11 @@ namespace Js
 
     bool JavascriptExceptionOperators::CrawlStackForWER(Js::ScriptContext& scriptContext)
     {
-        return Js::Configuration::Global.flags.WERExceptionSupport && !scriptContext.GetThreadContext()->HasCatchHandler();
+        // If we have exception handling, but no user exception handling, we should make
+        // the additional backtrace information available, as that increases consistency
+        // between native and nonative approaches (native code has some automatic error-
+        // handling structures that get added)
```
agarwal-sandeep: Can we add a test? None of our baselines gets effected?

lib/Runtime/Language/JavascriptExceptionOperators.h
```diff
@@ -36,10 +36,10 @@ namespace Js
             bool m_previousCatchHandlerExists;
             bool m_previousCatchHandlerToUserCodeStatus;
             ThreadContext* m_threadContext;
-            void FetchNonUserCodeStatus(ScriptContext *scriptContext);
+            void FetchNonUserCodeStatus(ScriptContext *scriptContext, bool isUserExecptionHandling);
```
digitalinfinity: Nit: spelling

lib/Runtime/Language/JavascriptExceptionOperators.cpp
Penguinwizzard: I've added a test in the new commit. The reason that none of our baselines were affected is that they pretty much all catch their exceptions in user code; this bug relates to errors _not_ caught in user code.

lib/Runtime/Language/JavascriptExceptionOperators.h
Penguinwizzard: Fixed in new iteration, thanks.

lib/Runtime/Language/JavascriptExceptionOperators.cpp
Penguinwizzard: Thanks, I've removed all instances of fFound and used returns to replace the related logic in the new commit on the PR.

Penguinwizzard:

PR #1630, opened 2016-09-23
Replace AddrOpnd with IntConstOpnd for 64-bit immediates

meg-gupta:
AddrOpnd was used with AddrOpndKindConstant for 64bit immediates, and then
legalized. Replace all such uses with IntConstOpnd, and add code in
legalizer to handle them.

PR #1682, opened 2016-10-03
Improve performance on some post-minifier code.

Penguinwizzard:
Running a minifier on some common benchmarks revealed areas where we could improve our performance.

### DIFF
--- a/lib/Backend/GlobOpt.cpp
+++ b/lib/Backend/GlobOpt.cpp
@@ -9212,6 +9212,14 @@ GlobOpt::GetConstantVar(IR::Opnd *opnd, Value *val)
     return nullptr;
 }
 
+bool BoolAndIntStaticAndTypeMismatch(Value* src1Val, Value* src2Val, Js::Var src1Var, Js::Var src2Var)
+{
+    ValueInfo *src1ValInfo = src1Val->GetValueInfo();
+    ValueInfo *src2ValInfo = src2Val->GetValueInfo();
+    return (src1ValInfo->IsNumber() && src1Var && src2ValInfo->IsBoolean() && src1Var != Js::TaggedInt::ToVarUnchecked(0) && src1Var != Js::TaggedInt::ToVarUnchecked(1)) ||
+        (src2ValInfo->IsNumber() && src2Var && src1ValInfo->IsBoolean() && src2Var != Js::TaggedInt::ToVarUnchecked(0) && src2Var != Js::TaggedInt::ToVarUnchecked(1));
+}
+
 bool
 GlobOpt::OptConstFoldBranch(IR::Instr *instr, Value *src1Val, Value*src2Val, Value **pDstVal)
 {
@@ -9247,12 +9255,7 @@ GlobOpt::OptConstFoldBranch(IR::Instr *instr, Value *src1Val, Value*src2Val, Val
     case Js::OpCode::BrNotNeq_A:
         if (!src1Var || !src2Var)
         {
-            ValueInfo *src1ValInfo = src1Val->GetValueInfo();
-            ValueInfo *src2ValInfo = src2Val->GetValueInfo();
-            if (
-                (src1ValInfo->IsNumber() && src1Var && src2ValInfo->IsBoolean() && src1Var != Js::TaggedInt::ToVarUnchecked(0) && src1Var != Js::TaggedInt::ToVarUnchecked(1)) ||
-                (src2ValInfo->IsNumber() && src2Var && src1ValInfo->IsBoolean() && src2Var != Js::TaggedInt::ToVarUnchecked(0) && src2Var != Js::TaggedInt::ToVarUnchecked(1))
-               )
+            if (BoolAndIntStaticAndTypeMismatch(src1Val, src2Val, src1Var, src2Var))
             {
                     result = false;
             }
@@ -9275,12 +9278,7 @@ GlobOpt::OptConstFoldBranch(IR::Instr *instr, Value *src1Val, Value*src2Val, Val
     case Js::OpCode::BrNotEq_A:
         if (!src1Var || !src2Var)
         {
-            ValueInfo *src1ValInfo = src1Val->GetValueInfo();
-            ValueInfo *src2ValInfo = src2Val->GetValueInfo();
-            if (
-                (src1ValInfo->IsNumber() && src1Var && src2ValInfo->IsBoolean() && src1Var != Js::TaggedInt::ToVarUnchecked(0) && src1Var != Js::TaggedInt::ToVarUnchecked(1)) ||
-                (src2ValInfo->IsNumber() && src2Var && src1ValInfo->IsBoolean() && src2Var != Js::TaggedInt::ToVarUnchecked(0) && src2Var != Js::TaggedInt::ToVarUnchecked(1))
-                )
+            if (BoolAndIntStaticAndTypeMismatch(src1Val, src2Val, src1Var, src2Var))
             {
                 result = true;
             }
--- a/lib/Backend/Lower.cpp
+++ b/lib/Backend/Lower.cpp
@@ -20253,10 +20253,6 @@ bool Lowerer::GenerateFastEqBoolInt(IR::Instr * instr, bool *pNeedHelper)
         targetInstr = instrBranch->GetTarget();
         labelFallthrough = instrBranch->GetOrCreateContinueLabel();
     }
-    else
-    {
-        return false;
-    }
 
     // Assume we need the helper until we can show otherwise.
     *pNeedHelper = true;
@@ -20334,7 +20330,6 @@ bool Lowerer::GenerateFastEqBoolInt(IR::Instr * instr, bool *pNeedHelper)
     }
     if (srcBool->IsIntConstOpnd())
     {
-        Assert(srcBool->IsIntConstOpnd());
         IR::IntConstOpnd * constSrcBool = srcBool->AsIntConstOpnd();
         IntConstType constIntVal = constSrcBool->GetValue();
         srcBoolConst = true;
@@ -20415,32 +20410,31 @@ bool Lowerer::GenerateFastEqBoolInt(IR::Instr * instr, bool *pNeedHelper)
     else if (srcIntConst && srcBoolConst)
     {
         // If both arguments are constant, we can statically determine the result.
-        bool sameval = srcIntConstVal == srcBoolConstVal;
+        bool sameVal = srcIntConstVal == srcBoolConstVal;
         if (isBranchNotCompare)
         {
             // For constant branches, branch to the target
             Assert(instr);
-            IR::LabelInstr * target = sameval && srcIntIsBoolable ? equalResultTarget : inequalResultTarget;
+            IR::LabelInstr * target = sameVal && srcIntIsBoolable ? equalResultTarget : inequalResultTarget;
             instr->InsertBefore(IR::BranchInstr::New(LowererMD::MDUncondBranchOpcode, target, this->m_func));
 #if DBG
             // Since we're not making a non-helper path to one of the branches, we need to tell
             // DbCheckPostLower that we are going to have a non-helper label without non-helper
             // branches.
             // Note: this following line isn't good practice in general
-            (sameval && srcIntIsBoolable ? inequalResultTarget : equalResultTarget)->m_noHelperAssert = true;
+            (sameVal && srcIntIsBoolable ? inequalResultTarget : equalResultTarget)->m_noHelperAssert = true;
 #endif
         }
         else
         {
             // For constant compares, load the constant result
-            LowererMD::CreateAssign(instr->GetDst(), this->LoadLibraryValueOpnd(instr, sameval && srcIntIsBoolable ? equalResultValue : inequalResultValue), instr);
+            LowererMD::CreateAssign(instr->GetDst(), this->LoadLibraryValueOpnd(instr, sameVal && srcIntIsBoolable ? equalResultValue : inequalResultValue), instr);
             instr->InsertBefore(IR::BranchInstr::New(LowererMD::MDUncondBranchOpcode, labelDone, this->m_func));
         }
     }
     else if (!srcIntConst && !srcBoolConst)
     {
         // If neither is constant, we can still do a bit better than loading the helper
-        IR::LabelInstr * firstTrue = IR::LabelInstr::New(Js::OpCode::Label, this->m_func);
         IR::LabelInstr * firstFalse = IR::LabelInstr::New(Js::OpCode::Label, this->m_func);
         IR::LabelInstr * forceInequal = IR::LabelInstr::New(Js::OpCode::Label, this->m_func);
         // We branch based on the zero-ness of the integer argument to two checks against the boolean argument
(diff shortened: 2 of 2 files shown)


lib/Backend/Lower.cpp
```diff
@@ -2009,6 +2009,13 @@ Lowerer::LowerRange(IR::Instr *instrStart, IR::Instr *instrEnd, bool defaultDoFa
                         this->LowerBrCMem(instr, IR::HelperOp_Equal, false);
                     }
                 }
+                else if (this->GenerateFastBrEqBoolInt(instr->AsBranchInstr(), &needHelper))
```
agarwal-sandeep: >GenerateFastBrEqBoolInt [](start = 31, length = 23)

Should we also have this for CmEq_A and CmNeq_A as most of the code is reusable?

rajatd:
What kind of perf improvements did you see?

Penguinwizzard:
The score of post-minifier octane navier-stokes.js improved ~30%, post-minifier mandreel.js improved ~22%, post-minifier splay.js improved ~24%, and a number of other tests posted smaller improvements. This doesn't impact the score on a non-minified version of octane though.

Penguinwizzard:

lib/Backend/Lower.cpp
```diff
@@ -20035,26 +20035,37 @@ Lowerer::TryGenerateFastBrOrCmTypeOf(IR::Instr *instr, IR::Instr **prev, bool *p
...
             {
-                if (!instrSrc1->m_isTempLastUse)
+                typeOpnd = instrSrc1;
+                idOpnd = instrSrc2;
+            }
+            else if (instrSrc2->m_sym == typeOfDst->m_sym) {
+                typeOpnd = instrSrc2;
+                idOpnd = instrSrc1;
+            }
+            if (typeOpnd && idOpnd)
```
rajatd: null check for instrSrc1 and instrSrc2 above makes this check redundant.

rajatd:
@pleath @akroshg  for bytecode changes

lib/Backend/GlobOpt.cpp
```diff
@@ -4268,16 +4268,29 @@ GlobOpt::MarkArgumentsUsedForBranch(IR::Instr * instr)
     if (instr->IsBranchInstr() && !instr->AsBranchInstr()->IsUnconditional())
     {
         IR::BranchInstr * bInstr = instr->AsBranchInstr();
+        IR::Opnd *src1 = bInstr->GetSrc1();
         IR::Opnd *src2 = bInstr->GetSrc2();
-        if (((!src2 && (instr->m_opcode == Js::OpCode::BrFalse_A || instr->m_opcode == Js::OpCode::BrTrue_A))
-            || (src2 &&  src2->IsConstOpnd()))
-            && bInstr->GetSrc1()->IsRegOpnd())
+        // These are used because we don't want to rely on src1 or src2 to always be the register/constant
+        IR::RegOpnd *regopnd = nullptr;
```
rajatd: >regopnd [](start = 21, length = 7)

nit: please use camelCase for variable names.

lib/Backend/GlobOpt.cpp
```diff
@@ -4268,16 +4268,29 @@ GlobOpt::MarkArgumentsUsedForBranch(IR::Instr * instr)
...
+        IR::Opnd *constopnd = nullptr;
+        if (!src2 && (instr->m_opcode == Js::OpCode::BrFalse_A || instr->m_opcode == Js::OpCode::BrTrue_A) && src1->IsRegOpnd()) {
```
rajatd: nit: opening brace on the next line, as has been done in the rest of the file

lib/Backend/GlobOpt.cpp
```diff
@@ -4268,16 +4268,29 @@ GlobOpt::MarkArgumentsUsedForBranch(IR::Instr * instr)
...
+        }
+        // We need to check for (0===arg) and (arg===0); this is especially important since some minifiers
+        // change all instances of one to the other.
+        else if (src2 && src2->IsConstOpnd() && src1->IsRegOpnd()) {
+            regopnd = src1->AsRegOpnd();
+            constopnd = src2;
+        }
+        else if (src2 && src2->IsRegOpnd() && src1->IsConstOpnd()) {
+            regopnd = src2->AsRegOpnd();
+            constopnd = src1;
```
rajatd: >constopnd [](start = 12, length = 9)

what is constOpnd used for?

lib/Backend/GlobOpt.cpp
```diff
@@ -4268,16 +4268,29 @@ GlobOpt::MarkArgumentsUsedForBranch(IR::Instr * instr)
...
     {
         IR::BranchInstr * bInstr = instr->AsBranchInstr();
+        IR::Opnd *src1 = bInstr->GetSrc1();
         IR::Opnd *src2 = bInstr->GetSrc2();
-        if (((!src2 && (instr->m_opcode == Js::OpCode::BrFalse_A || instr->m_opcode == Js::OpCode::BrTrue_A))
-            || (src2 &&  src2->IsConstOpnd()))
-            && bInstr->GetSrc1()->IsRegOpnd())
+        // These are used because we don't want to rely on src1 or src2 to always be the register/constant
+        IR::RegOpnd *regopnd = nullptr;
+        IR::Opnd *constopnd = nullptr;
```
rajatd: >constopnd [](start = 18, length = 9)

what is constOpnd used for?

lib/Backend/BackwardPass.cpp
```diff
@@ -6988,19 +6988,31 @@ BackwardPass::ProcessBailOnNoProfile(IR::Instr *instr, BasicBlock *block)
...
+    IR::Instr *instrNope = nullptr;
     if (curNext->m_opcode == Js::OpCode::Ld_A && curNext->GetDst()->IsRegOpnd() && curNext->GetDst()->AsRegOpnd()->m_fgPeepTmp)
     {
-        block->RemoveInstr(instr);
-        return true;
+        // Instead of just giving up, we can be a little trickier. We can instead treat the tmp declaration(s) as a
+        // part of the block prefix, and put the bailonnoprofile immediately after them. This has the added benefit
+        // that we can still merge up blocks beginning with bailonnoprofile, even if they would otherwise not allow
+        // us to, due to the fact that these tmp declarations would be pre-empted by the higher-level bailout.
+        while (curNext->m_opcode == Js::OpCode::Ld_A && curNext->GetDst()->IsRegOpnd() && curNext->GetDst()->AsRegOpnd()->m_fgPeepTmp)
```
LouisLaf: This can be cleaned up a bit it seems.  The while condition is the same as the if, making the if unnecessary.

lib/Backend/BackwardPass.cpp
```diff
@@ -6988,19 +6988,31 @@ BackwardPass::ProcessBailOnNoProfile(IR::Instr *instr, BasicBlock *block)
...
+            curNext = curNext->GetNextRealInstrOrLabel();
+        }
+        instrNope = curNext;
     }
 
     curInstr = instr->m_prev;
 
     // Move to top of block.
-    while(!curInstr->StartsBasicBlock())
+    while(!curInstr->StartsBasicBlock() && curInstr != instrNope)
```
LouisLaf: >nstrNope [](start = 56, length = 8)

The way I read this, if instrNope is set, it will always be == curInstr...

lib/Backend/BackwardPass.cpp
Penguinwizzard: I've restructured the code to be a single while loop, removing the extra copy of the condition.

lib/Backend/BackwardPass.cpp
Penguinwizzard: Tracing the code seems to indicate that it has the desired effect; while they will be equal in cases where the BailOnNoProfile is already just under the Ld_A fgPeepTmps, in many cases there will be other instructions in between. In that case, instrNope serves as a second "start of block", at which the search back towards the start of the block should stop.

lib/Backend/GlobOpt.cpp
Penguinwizzard: I've removed that in the new revision, thanks; it was an artefact of how I ambidextrized the check.

lib/Backend/GlobOpt.cpp
Penguinwizzard: Fixed.

lib/Backend/FlowGraph.h
```diff
@@ -347,6 +347,7 @@ class BasicBlock
     uint8                hasCall:1;
     uint8                isVisited:1;
     uint8                isAirLockCompensationBlock:1;
+    uint8                beginsBailOnNoProfile:1;
```
rajatd: >beginsBailOnNoProfile [](start = 25, length = 21)

nit: consider changing the name to beginsWithBailOnNoProfile

lib/Backend/GlobOpt.cpp
```diff
@@ -9230,47 +9243,77 @@ GlobOpt::OptConstFoldBranch(IR::Instr *instr, Value *src1Val, Value*src2Val, Val
     case Js::OpCode::BrNotNeq_A:
         if (!src1Var || !src2Var)
         {
-            return false;
+            ValueInfo *src1ValInfo = src1Val->GetValueInfo();
+            ValueInfo *src2ValInfo = src2Val->GetValueInfo();
+            if (
+                (src1ValInfo->IsNumber() && src1Var && src2ValInfo->IsBoolean() && src1Var != Js::TaggedInt::ToVarUnchecked(0) && src1Var != Js::TaggedInt::ToVarUnchecked(1)) ||
```
rajatd: Can we also fold a branch when we get a non-null constant var for the src which is definitely boolean and then compare the constant boolean with the constant number value of the other src?

lib/Backend/GlobOpt.cpp
```diff
@@ -9230,47 +9243,77 @@ GlobOpt::OptConstFoldBranch(IR::Instr *instr, Value *src1Val, Value*src2Val, Val
...
     case Js::OpCode::BrNeq_A:
     case Js::OpCode::BrNotEq_A:
         if (!src1Var || !src2Var)
         {
-            return false;
+            ValueInfo *src1ValInfo = src1Val->GetValueInfo();
+            ValueInfo *src2ValInfo = src2Val->GetValueInfo();
+            if (
+                (src1ValInfo->IsNumber() && src1Var && src2ValInfo->IsBoolean() && src1Var != Js::TaggedInt::ToVarUnchecked(0) && src1Var != Js::TaggedInt::ToVarUnchecked(1)) ||
+                (src2ValInfo->IsNumber() && src2Var && src1ValInfo->IsBoolean() && src2Var != Js::TaggedInt::ToVarUnchecked(0) && src2Var != Js::TaggedInt::ToVarUnchecked(1))
```
rajatd: nit: I would refactor this condition (and the one above for BrEq_A) into a method.

lib/Backend/Lower.cpp
```diff
@@ -19995,6 +20006,422 @@ Lowerer::GenerateIsBuiltinRecyclableObject(IR::RegOpnd *regOpnd, IR::Instr *inse
...
+    case Js::OpCode::BrNeq_A:
+    case Js::OpCode::BrNotEq_A:
+    case Js::OpCode::CmNeq_A:
+        isNegOp = true;
+        break;
+    default:
+        // This opcode is not one of the ones that should be handled here.
+        return false;
+        break;
+    }
```
rajatd: Could make this part cleaner. Example,
switch (instr->m_opcode)
    {
    case Js::OpCode::BrSrEq_A:
    case Js::OpCode::BrSrNotNeq_A:
    case Js::OpCode::CmSrEq_A:
        isStrict = true;
        break;

    case Js::OpCode::BrSrNeq_A:
    case Js::OpCode::BrSrNotEq_A:
    case Js::OpCode::CmSrNeq_A:
        isStrict = true;
    case Js::OpCode::BrNeq_A:
    case Js::OpCode::BrNotEq_A:
    case Js::OpCode::CmNeq_A:
        isNeqOp = true;
        break;
    }

And I would put an assert in the default case for unexpected opcode.

lib/Backend/Lower.cpp
```diff
@@ -19995,6 +20006,422 @@ Lowerer::GenerateIsBuiltinRecyclableObject(IR::RegOpnd *regOpnd, IR::Instr *inse
...
+    {
+        srcInt = src1;
+        srcBool = src2;
+    }
+    else
+    {
+        return false;
+    }
+
+    // If either instruction is constant, we can simplify the check. If both are constant, we can eliminate it
```
rajatd: >instruction [](start = 17, length = 11)

nit: source

lib/Backend/Lower.cpp
```diff
@@ -19995,6 +20006,422 @@ Lowerer::GenerateIsBuiltinRecyclableObject(IR::RegOpnd *regOpnd, IR::Instr *inse
...
+            instr->InsertBefore(IR::BranchInstr::New(LowererMD::MDUncondBranchOpcode, labelDone, this->m_func));
+            instr->InsertBefore(isZero);
+            LowererMD::CreateAssign(instr->GetDst(), this->LoadLibraryValueOpnd(instr, !srcBoolConstVal ? equalResultValue : inequalResultValue), instr);
+            instr->InsertBefore(IR::BranchInstr::New(LowererMD::MDUncondBranchOpcode, labelDone, this->m_func));
+        }
+    }
+    if (*pNeedHelper)
+    {
+        instr->InsertBefore(labelHelper);
+    }
```
rajatd: I would rather have a wrapper for this function, have that wrapper pass the helper label to this function, insert the helper label when this function returns, and replace all callsites of this function with that helper. It would make using this function in other places easier.

lib/Backend/Lower.cpp
```diff
@@ -19995,6 +20006,422 @@ Lowerer::GenerateIsBuiltinRecyclableObject(IR::RegOpnd *regOpnd, IR::Instr *inse
...
+    else if (srcBool->IsAddrOpnd())
+    {
+        IR::AddrOpnd * addrSrcBool = srcInt->AsAddrOpnd();
+        if (!(addrSrcBool && addrSrcBool->IsVar() && Js::TaggedInt::Is(addrSrcBool->m_address)))
+        {
+            return false;
+        }
+        int32 value = Js::TaggedInt::ToInt32(addrSrcBool->m_address);
+        srcBoolConst = true;
+        srcBoolConstVal = value != 0;
```
rajatd: This block shouldn't be required after @meg-gupta 's #1630

lib/Backend/Lower.cpp
```diff
@@ -19995,6 +20006,422 @@ Lowerer::GenerateIsBuiltinRecyclableObject(IR::RegOpnd *regOpnd, IR::Instr *inse
...
+        }
+    }
+    else if (srcInt->IsConstOpnd())
+    {
+        // Not handled yet
+        return false;
+    }
+    if (srcBool->IsIntConstOpnd())
+    {
+        Assert(srcBool->IsIntConstOpnd());
```
rajatd: >Assert(srcBool->IsIntConstOpnd()); [](start = 8, length = 34)

redundant?

PR #1683, opened 2016-10-03
fix oop jit issue with undefined const this

MikeHolman:


leirocks:
:shipit:

lib/Backend/Inline.cpp
```diff
@@ -4799,7 +4799,26 @@ Inline::MapFormals(Func *inlinee,
...
+                        // with OOP JIT we may create const Opnds for library vars without materializing a JITRecyclableObject
+                        IR::Opnd * thisConstOpnd = thisConstSym->GetConstOpnd();
+                        if (thisConstOpnd->GetValueType().IsUndefined())
+                        {
+                            typeId = Js::TypeIds_Undefined;
+                        }
+                        else if (thisConstOpnd->GetValueType().IsNull())
+                        {
+                            typeId = Js::TypeIds_Null;
+                        }
```
MikeHolman: I thought there might be a way for it to be some other library value that we wouldn't optimize. Since condition below has checks that would exclude other library values, I thought it should be ok not to check for them.

PR #1684, opened 2016-10-04
line/column difference on emitting index on boolean expression.

akroshg:
The index expression was forgotton to have start/end source maps. That causes the error reported on the wrong line/column. Fixed that.


akroshg:
@rajatd have a look

akroshg:
thanks @agarwal-sandeep

PR #1686, opened 2016-10-04
require min of VS2015 Update 3 for constexpr

MikeHolman:


dilijev:
Good thing I just updated to VS 2015 Update 3 :P

MikeHolman:
This will not impact VS2013 support, as all constexpr code is behind flag/not used for impacting runtime behavior. I had originally set min version where we support constexpr to VS2015 RTM (used for some static asserts, behind the HAS_CONSTEXPR flag). Unfortunately, VS2015 RTM had some bugs with constexpr, in our case causing the assert condition to not be considered static.

I think this was probably fixed in Update 1 since VC team put out a whole [blog post](https://blogs.msdn.microsoft.com/vcblog/2015/12/02/constexpr-in-vs2015-update-1/) talking about fixing/improving constexpr. But I have Update 3, which I know works, so figured we may as well be conservative.

jianchun:
How about clang? Try turn it on for non MSC?

MikeHolman:
@jianchun I thought about it but didn't know what exactly the condition should be and this seemed sufficiently inconsequential that whoever has some particular want can add it in. But if you know the appropriate macro condition I'll add it.

dilijev:
@MikeHolman is there a significant runtime performance gain for using constexpr? If so, I'll make sure all of our build infra that we use for testing has Dev14 U3.

MikeHolman:
Ok, by popular demand, I've loosened the restriction so it will be defined for clang build as well.

@dilijev Right now the only thing it is being used for is as a condition in CompileAsserts, so it's useful to have in CI, but not going to see any perf benefit (and I think in general C++ backends are good enough at this type of code that it wouldn't be significant win).

PR #1688, opened 2016-10-04
explicitly copy profile data for background in-proc jit

MikeHolman:
I changed how the profile data is passed for OOP JIT (which will be copied via RPC), however I didn't consider in-proc JIT at the time. This adds back code to copy ld/st elem info for in-proc background jit.


satheeshravi:
:shipit:

MikeHolman:

PR #1689, opened 2016-10-04
OOP JIT server side error handling

leirocks:
adding a new type of exception to indicate system call failed(mostly because content process is terminated/crash and VM is deleted, the JIT server is still trying to do the VM operations, which fails)
wrap all server call in exception handling
make relationship between scriptcontext and threadcontext in server side

### DIFF
--- a/lib/Backend/CodeGenNumberAllocator.cpp
+++ b/lib/Backend/CodeGenNumberAllocator.cpp
@@ -345,9 +345,13 @@ Js::JavascriptNumber* XProcNumberPageSegmentImpl::AllocateNumber(Func* func, dou
         {
             Assert((unsigned int)((char*)tail->GetEndAddress() - (char*)tail->GetCommitEndAddress()) >= BlockSize);
             // TODO: implement guard pages (still necessary for OOP JIT?)
-            auto ret = ::VirtualAllocEx(hProcess, tail->GetCommitEndAddress(), BlockSize, MEM_COMMIT, PAGE_READWRITE);
-            if (!ret)
+            LPVOID addr = ::VirtualAllocEx(hProcess, tail->GetCommitEndAddress(), BlockSize, MEM_COMMIT, PAGE_READWRITE);
+            if (addr == nullptr)
             {
+                if (hProcess != GetCurrentProcess())
+                {
+                    Js::Throw::CheckAndThrowJITOperationFailed();
+                }
                 Js::Throw::OutOfMemory();
             }
             tail->committedEnd += BlockSize;
@@ -359,6 +363,10 @@ Js::JavascriptNumber* XProcNumberPageSegmentImpl::AllocateNumber(Func* func, dou
     void* pages = ::VirtualAllocEx(hProcess, nullptr, PageCount * AutoSystemInfo::PageSize, MEM_RESERVE, PAGE_READWRITE);
     if (pages == nullptr)
     {
+        if (hProcess != GetCurrentProcess())
+        {
+            Js::Throw::CheckAndThrowJITOperationFailed();
+        }
         Js::Throw::OutOfMemory();
     }
 
--- a/lib/Backend/CodeGenNumberAllocator.cpp
+++ b/lib/Backend/CodeGenNumberAllocator.cpp
@@ -333,6 +333,7 @@ Js::JavascriptNumber* XProcNumberPageSegmentImpl::AllocateNumber(Func* func, dou
             if (!WriteProcessMemory(hProcess, (void*)number, pLocalNumber, sizeCat, &bytesWritten)
                 || bytesWritten != sizeCat)
             {
+                Js::Throw::CheckAndThrowJITOperationFailed();
                 Output::Print(_u("FATAL ERROR: WriteProcessMemory failed, GLE: %d\n"), GetLastError());
                 Js::Throw::FatalInternalError(); // TODO: don't bring down whole server process, but pass the last error to main process
             }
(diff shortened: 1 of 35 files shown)


lib/Backend/CodeGenNumberAllocator.cpp
```diff
@@ -345,9 +346,13 @@ Js::JavascriptNumber* XProcNumberPageSegmentImpl::AllocateNumber(Func* func, dou
         {
             Assert((unsigned int)((char*)tail->GetEndAddress() - (char*)tail->GetCommitEndAddress()) >= BlockSize);
             // TODO: implement guard pages (still necessary for OOP JIT?)
-            auto ret = ::VirtualAllocEx(hProcess, tail->GetCommitEndAddress(), BlockSize, MEM_COMMIT, PAGE_READWRITE);
-            if (!ret)
+            LPVOID addr = ::VirtualAllocEx(hProcess, tail->GetCommitEndAddress(), BlockSize, MEM_COMMIT, PAGE_READWRITE);
+            if (addr == nullptr)
             {
+                if (hProcess != GetCurrentProcess())
```
MikeHolman: maybe we still should throw oom here?

MikeHolman: i'm thinking -- the WriteProcessMemory failing is more of a fatal error like AV (and we should at least disable JIT after that). This, it could be that proc legitimately is OOM.

---
In reply to: [81823255](https://github.com/Microsoft/ChakraCore/pull/1689#discussion_r81823255) [](ancestors = 81823255)

lib/Backend/NativeCodeGenerator.cpp
```diff
@@ -909,6 +909,9 @@ NativeCodeGenerator::CodeGen(PageAllocator * pageAllocator, CodeGenWorkItem* wor
             Js::Throw::OutOfMemory();
         case VBSERR_OutOfStack:
             throw Js::StackOverflowException();
+        case E_ACCESSDENIED:
+            // OOP JIT TODO: if server side can't handle request any more, use better error code and turn off JIT
+            throw Js::JITOperationFailedException(workItem->codeGenResult);
```
MikeHolman: wonder if we want to assert against this or even throw fatal error? i think it should only happen in case this process is already dead

lib/JITServer/JITServer.cpp
```diff
@@ -131,7 +131,13 @@ ServerInitializeThreadContext(
 {
     AUTO_NESTED_HANDLED_EXCEPTION_TYPE(static_cast<ExceptionType>(ExceptionType_OutOfMemory | ExceptionType_StackOverflow));
 
-    ServerThreadContext * contextInfo = HeapNew(ServerThreadContext, threadContextData);
+    ServerThreadContext * contextInfo = HeapNewNoThrow(ServerThreadContext, threadContextData);
```
MikeHolman: i'm pretty sure rpc handles the OOM and returns E_OUTOFMEMORY. but if we don't want to use that, maybe we should remove the AUTO_NESTED_HANDLED_EXCEPTION_TYPE?

leirocks: btw, looks we don't need AUTO_NESTED_HANDLED_EXCEPTION_TYPE anywhere in this file, that looks like mostly for run time in script to prevent unknown exception get out of scope. the Assertion inside is always true here in this file (see !JsUtil::ExternalApi::IsScriptActiveOnCurrentThreadContext() )

---
In reply to: [81828161](https://github.com/Microsoft/ChakraCore/pull/1689#discussion_r81828161) [](ancestors = 81828161,81825163)

MikeHolman:
>         [in] CHAKRA_PTR threadContextInfoAddress,

i guess now we don't need this anymore

---
Refers to: lib/JITIDL/ChakraJIT.idl:84 in b64f838. [](commit_id = b64f838c9d9f33aa9af97aec2818a2b52562cff9, deletion_comment = False)

lib/Backend/NativeCodeGenerator.cpp
leirocks: I'm removing this, for client side just use fatal error for now. we can change to turning off JIT in some cases(like JIT server is dead) in the future. 

---
In reply to: [81824509](https://github.com/Microsoft/ChakraCore/pull/1689#discussion_r81824509) [](ancestors = 81824509)

lib/Backend/CodeGenNumberAllocator.cpp
```diff
@@ -333,6 +333,7 @@ Js::JavascriptNumber* XProcNumberPageSegmentImpl::AllocateNumber(Func* func, dou
             if (!WriteProcessMemory(hProcess, (void*)number, pLocalNumber, sizeCat, &bytesWritten)
                 || bytesWritten != sizeCat)
             {
+                Js::Throw::CheckAndThrowJITOperationFailed();
                 Output::Print(_u("FATAL ERROR: WriteProcessMemory failed, GLE: %d\n"), GetLastError());
```
leirocks: done

---
In reply to: [81823110](https://github.com/Microsoft/ChakraCore/pull/1689#discussion_r81823110) [](ancestors = 81823110)

lib/JITServer/JITServer.cpp
MikeHolman: are you sure? I thought i have seen it fire before. but i guess my point was we should probably have try/catch or use no throw allocator

---
In reply to: [81830381](https://github.com/Microsoft/ChakraCore/pull/1689#discussion_r81830381) [](ancestors = 81830381,81828161,81825163)

leirocks:
>         [in] CHAKRA_PTR threadContextInfoAddress,

ah yes, can be queried from scriptContext, let me remove this

---
In reply to: [251483111](https://github.com/Microsoft/ChakraCore/pull/1689#issuecomment-251483111) [](ancestors = 251483111)

---
Refers to: lib/JITIDL/ChakraJIT.idl:84 in b64f838. [](commit_id = b64f838c9d9f33aa9af97aec2818a2b52562cff9, deletion_comment = False)

MikeHolman:
:shipit:

leirocks:
@dotnet-bot test Copyright Check  please

PR #1692, opened 2016-10-04
xplat: fix some clang undefined sanitizer issues

jianchun:
Utf8Codex.cpp, Alloc.cpp, TypePath.h: fix data alignment.

HeapBlock: refactor to remove "this == nullptr" scenario.

PreReservedPageAllocator: refactor to use a single instance for
VirtualAllocWrapper and remove "this == nullptr" scenario.

Others: fix bad casts.

### DIFF
--- a/lib/Common/Codex/Utf8Codex.cpp
+++ b/lib/Common/Codex/Utf8Codex.cpp
@@ -37,7 +37,7 @@ namespace utf8
 
     inline bool ShouldFastPath(LPCUTF8 pb, LPCOLESTR pch)
     {
-        return (reinterpret_cast<size_t>(pb) & mAlignmentMask) == 0 || (reinterpret_cast<size_t>(pch) & mAlignmentMask) == 0;
+        return (reinterpret_cast<size_t>(pb) & mAlignmentMask) == 0 && (reinterpret_cast<size_t>(pch) & mAlignmentMask) == 0;
     }
 
     inline size_t EncodedBytes(char16 prefix)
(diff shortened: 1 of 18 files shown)


jianchun:
@curtisman @digitalinfinity @satheeshravi Could you please help CR? Thanks

lib/Common/Codex/Utf8Codex.cpp
```diff
@@ -37,7 +37,7 @@ namespace utf8
 
     inline bool ShouldFastPath(LPCUTF8 pb, LPCOLESTR pch)
     {
-        return (reinterpret_cast<size_t>(pb) & mAlignmentMask) == 0 || (reinterpret_cast<size_t>(pch) & mAlignmentMask) == 0;
+        return (reinterpret_cast<size_t>(pb) & mAlignmentMask) == 0 && (reinterpret_cast<size_t>(pch) & mAlignmentMask) == 0;
```
digitalinfinity: Still don't quite understand how/when this fast path is taken but it's not blocking for this PR. The correctness hasn't changed with your change, so that should be ok. We should separately look at the effectiveness of ShouldFastPath...

digitalinfinity:
:shipit:

lib/Common/Codex/Utf8Codex.cpp
jianchun: Agree. (This change does not affect Scanner/Parser, only used in utf8~utf16 conversion routines, thus impact is relatively low.) Thanks for review!

PR #1694, opened 2016-10-04
Insert StatementBoundary at the right position for polymorphic inlining

rajatd:
Since polymorphic inlining using fixed methods emits a helper block in the inliner itself, the inliner should emit a StatementBoundary for the helper block so that it stays with the block if/when the block gets moved around in the flowgraph builder and/or globopt.


akroshg:
could you add a repro case for this?

rajatd:
@LouisLaf @akroshg @abchatra  mind taking a look?

akroshg:
:shipit:

LouisLaf:
:shipit:

PR #1695, opened 2016-10-05
Set debugContext to null before calling close and HeapDelete

agarwal-sandeep:
In multiple context scenarios (Slate) with chk builds I see debugContext
is not null and is garbage, ScriptContext is in closed state which means
debugContext is in the process of being deleted and has not be set to null.
I can put critical section to the particular AV but it’s better to hold
debugContext in a temporary variable set scriptContext->debugContext to
null and then HeapDelete memory pointed by temporary variable.

I have kept the autoDebugContextCloseCS for PDM case though it should
not be needed (I think so)

### DIFF
--- a/lib/Runtime/Base/ScriptContext.cpp
+++ b/lib/Runtime/Base/ScriptContext.cpp
@@ -670,9 +670,10 @@ namespace Js
             // Guard the closing and deleting of DebugContext as in meantime PDM might
             // call OnBreakFlagChange
             AutoCriticalSection autoDebugContextCloseCS(&debugContextCloseCS);
-            this->debugContext->Close();
-            HeapDelete(this->debugContext);
+            DebugContext* tempDebugContext = this->debugContext;
             this->debugContext = nullptr;
+            tempDebugContext->Close();
+            HeapDelete(tempDebugContext);
         }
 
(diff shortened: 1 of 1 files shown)


lib/Runtime/Base/ScriptContext.cpp
```diff
@@ -670,9 +670,10 @@ namespace Js
             // Guard the closing and deleting of DebugContext as in meantime PDM might
             // call OnBreakFlagChange
             AutoCriticalSection autoDebugContextCloseCS(&debugContextCloseCS);
-            this->debugContext->Close();
-            HeapDelete(this->debugContext);
+            DebugContext* tempDebugContext = this->debugContext;
```
akroshg: you can use interlockedexchangepointer in order to make this operation atomic.

akroshg:
:shipit:

lib/Runtime/Base/ScriptContext.cpp
agarwal-sandeep: I think that is not necessary as context switch between these 2 instructions won't cause an issue as InternalClose is non-entrant and no where else scriptContext->debugContext is changed.

PR #1698, opened 2016-10-05
fix crash when using jscript ETW provider

MikeHolman:


agarwal-sandeep:
> #ifdef ENABLE_BASIC_TELEMETRY

nit: Just for consistency can we move NativeCodeGenerator::LogCodeGenDone(workItem, &start_time); here?

---
Refers to: lib/Backend/NativeCodeGenerator.cpp:1154 in 44b6b1d. [](commit_id = 44b6b1d1a2733236aaf066691f1f5a65a5854d5b, deletion_comment = False)

lib/Backend/CodeGenWorkItem.h
```diff
@@ -42,7 +42,7 @@ struct CodeGenWorkItem : public JsUtil::Job
 
     uint GetFunctionNumber() const
     {
-        return this->jitData.jitData->bodyData->funcNumber;
+        return this->functionBody->GetFunctionNumber();
```
ThomsonTan: >this->functionBody [](start = 15, length = 18)

What the difference between this->functionBody and this->jitData.jitData->bodyData? Also wondering why we could get AV in the latter case.

MikeHolman: this->jitData.jitData is not initialized until NativeCodeGenerator::CodeGen, whereas functionBody is passed as argument to the constructor of CodeGenWorkItem so is there from the beginning.

This area definitely got a bit messy with OOP JIT, and I'm planning on doing bigger cleanup of CodeGenWorkItem to address this sort of strangeness.

ThomsonTan:
:shipit:

PR #1700, opened 2016-10-06
refactor, avoid extra copying of ObjTypeSpecInfo

MikeHolman:
After OOP JIT change, we were doing extra copying for ObjTypeSpecInfo, but ObjTypeSpecInfo is only used for JIT and lifetime is that of the workitem, so we don't need to copy this and can lay it out such that RPC can directly marshal it..

This is giving me about 2% perf win in OOP JIT Typescript, everything else is within noise range.


dilijev:
@MikeHolman Any update here?

MikeHolman:
@dilijev I still want to merge this. But I never got a review.

rajatd:
:shipit:

MikeHolman:
@dotnet-bot test  Windows x64_release

PR #1701, opened 2016-10-06
FindNextProperty needs to return string in the request context, not of it's own context.

curtisman:


curtisman:
@rajatd please review

lib/Runtime/Types/DictionaryTypeHandler.cpp
```diff
@@ -199,7 +199,7 @@ namespace Js
                 }
 
                 *propertyId = propertyRecord->GetPropertyId();
-                *propertyStringName = type->GetScriptContext()->GetPropertyString(*propertyId);
+                *propertyStringName = scriptContext->GetPropertyString(*propertyId);
```
agarwal-sandeep: >GetPropertyString [](start = 53, length = 17)

Can we add a test as well?

PR #1703, opened 2016-10-06
xplat: enforce TypedArray bound check in AsmJs

jianchun:
For xplat, always do bound check for Ld/St TypedArray in AsmJs. Currently
we don't support out of bound access violation recovery on non _WIN32.

### DIFF
--- a/lib/Backend/amd64/LowererMDArch.cpp
+++ b/lib/Backend/amd64/LowererMDArch.cpp
@@ -1113,13 +1113,19 @@ LowererMDArch::LowerAsmJsLdElemHelper(IR::Instr * instr, bool isSimdLoad /*= fal
 {
     IR::Instr* done;
     IR::Opnd * src1 = instr->UnlinkSrc1();
+    IRType type = src1->GetType();
     IR::RegOpnd * indexOpnd = src1->AsIndirOpnd()->GetIndexOpnd();
     const uint8 dataWidth = instr->dataWidth;
 
     Assert(isSimdLoad == false || dataWidth == 4 || dataWidth == 8 || dataWidth == 12 || dataWidth == 16);
 
+#ifdef _WIN32
     // For x64, bound checks are required only for SIMD loads.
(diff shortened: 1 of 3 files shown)


jianchun:
@MikeHolman Could you please help review?

lib/Backend/AsmJsJITInfo.cpp
```diff
@@ -159,6 +159,11 @@ AsmJsJITInfo::UsesHeapBuffer() const
 bool
 AsmJsJITInfo::AccessNeedsBoundCheck(uint offset) const
 {
+#ifdef _WIN32
     // Normally, heap has min size of 0x10000, but if you use ChangeHeap, min heap size is increased to 0x1000000
     return offset >= 0x1000000 || (IsHeapBufferConst() && offset >= 0x10000);
```
MikeHolman: There is runtime guarantee that any access less than this is in bounds, so don't need to exclude this from xplat

lib/Backend/amd64/LowererMDArch.cpp
```diff
@@ -1152,7 +1157,10 @@ LowererMDArch::LowerAsmJsLdElemHelper(IR::Instr * instr, bool isSimdLoad /*= fal
         }
         Lowerer::InsertBranch(Js::OpCode::Br, loadLabel, helperLabel);
 
-        lowererMD->m_lowerer->GenerateRuntimeError(loadLabel, JSERR_ArgumentOutOfRange, IR::HelperOp_RuntimeRangeError);
+        if (isSimdLoad)
```
MikeHolman: else, you should move 0 into dst.

lib/Backend/AsmJsJITInfo.cpp
jianchun: Will revert this. (Looks I originally added this when my problem was the other piece not emitting bound check.)

jianchun:
@MikeHolman Updated, thanks!

PR #1706, opened 2016-10-06
Correct a flag added recently, fixing formatting of exceptions thrown in catch blocks.

Penguinwizzard:
I recently added a parameter to AutoCatchHandlers to help detection
of whether or not an exception can be caught in user code, to match
the -nonative output better. Unfortunately, I set it incorrectly in
one place; namely, the case where we're in a "catch" statement made
by the user. Flipping the flag the other way fixes the issue.

### DIFF
--- a/lib/Runtime/Language/JavascriptExceptionOperators.cpp
+++ b/lib/Runtime/Language/JavascriptExceptionOperators.cpp
@@ -91,7 +91,7 @@ namespace Js
 
         try
         {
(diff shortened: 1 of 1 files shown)


akroshg:
why this change in the x64 case only? I see x86 and others are still set to `true`.

Penguinwizzard:
The issue is actually a bit more complicated than I first thought, on top of having to fix it appropriately for other architectures. There's apparently some inaccuracies with the scoping of the autocatchhandler in this case that is the true cause of the inaccuracy.

PR #1707, opened 2016-10-06
move dynamic interpreter thunk generation OOP

MikeHolman:


lib/JITServer/JITServer.cpp
```diff
@@ -396,6 +396,109 @@ ServerCleanupScriptContext(
...
+#if PDATA_ENABLED
+            &pdataStart,
+            &epilogEnd,
+#endif
+            &thunkCount
+        );
+
+        if (!emitBufferManager->ProtectBufferWithExecuteReadWriteForInterpreter(allocation))
+        {
+            Js::Throw::JITOperationFailed(GetLastError());
```
leirocks: >Js::Throw::JITOperationFailed(GetLastError()); [](start = 12, length = 46)

don't need this. VirtualProtectEx call failure has been checked right after the call. if it's not system call failure, there's fail fast, also the last error will be meaningless

leirocks: I think this call should never never return false, if we reach here and get a false return, I think it means we missed a path that we should fail fast. you can double check with Satheesh

---
In reply to: [82306816](https://github.com/Microsoft/ChakraCore/pull/1707#discussion_r82306816) [](ancestors = 82306816)

leirocks:
>         HeapDeleteArray(sizeInBytes, writeBuffer);

can you fix the memory leak here, I missed this in my change
or, maybe we can use stack memory instead? at this point the stack usage should be very low for OOP JIT and background JIT, right? 
otherwise maybe we can use stack to repeatedly do the copying (like 1KB by 1KB)

---
Refers to: lib/Common/Memory/MemUtils.cpp:45 in 2863255. [](commit_id = 286325525afb3e3cfb548704b0f8de3e29d4e355, deletion_comment = False)

lib/JITServer/JITServer.cpp
```diff
@@ -396,6 +396,107 @@ ServerCleanupScriptContext(
...
+
+        bool success = emitBufferManager->ProtectBufferWithExecuteReadWriteForInterpreter(allocation);
+        Assert(success);
+
+        if (!WriteProcessMemory(threadContext->GetProcessHandle(), remoteBuffer, localBuffer, bufferSize, nullptr))
+        {
+            Js::Throw::JITOperationFailed(GetLastError());
+        }
+
+        success = emitBufferManager->CommitReadWriteBufferForInterpreter(allocation, remoteBuffer, bufferSize);
```
satheeshravi: >CommitReadWriteBufferForInterpreter [](start = 37, length = 35)

you might want to add JITOperationFailed in the false path for this API. It will return false, when the protection is unsuccessful - which can happen with the CFG around -

satheeshravi:
Looks good, otherwise.

satheeshravi:
:shipit:

lib/Common/Memory/CustomHeap.cpp
```diff
@@ -1056,7 +1056,30 @@ void FillDebugBreak(_In_ BYTE* buffer, __in size_t byteCount, HANDLE processHand
...
+        {
+            for (size_t i = 0; i < byteCount; i+= sizeof(char16))
+            {
+                if (!WriteProcessMemory(processHandle, buffer + i, &pattern, sizeof(char16), NULL))
+                {
+                    Js::Throw::CheckAndThrowJITOperationFailed();
+                    Js::Throw::FatalInternalError();
+                }
+            }
+            if (byteCount % 2 != 0)
```
MikeHolman: i changed this to use different algorithm instead that doesn't need to use heap.

MikeHolman:
@dotnet-bot test this please

PR #1708, opened 2016-10-06
xplat: keep uncaught exception object alive

jianchun:
On xplat when throwing a JS exception, the exception object pointer
is not on stack and could be collected by GC prematually (triggered
by destructors). Manually track it in threadContext data temporarily.


jianchun:
@digitalinfinity @obastemur @curtisman Please help CR

jianchun:
@curtisman suggested I tight up this further. Will update.

PR #1709, opened 2016-10-06
Fix for AddrOpnd to IntConstOpnd tests break

meg-gupta:
When a IntConstOpnd fits in the opnd field, but the instruction doesnt
accept an int constant as its operand, then the immediate should be
hoisted into a register and then used in the instruction. Fix the conditions so
that this case is handled correctly.


meg-gupta:
@satheeshravi @rajatd Can you please review ?

satheeshravi:
:shipit:

PR #1711, opened 2016-10-07
fix segment IsInPreReservedHeapPageAllocator regression

jianchun:
My recent change breaks segment->IsInPreReservedHeapPageAllocator check
(when CFG is enabled). Refactor related SegmentBase/PageAllocatorBase
to move relevant data to super classes so that we can perform the check on
super class similar to before.


jianchun:
@curtisman @satheeshravi @agarwal-sandeep Please help review, thanks

curtisman:
:shipit:

jianchun:
Thanks @curtisman

PR #1712, opened 2016-10-07
oop jit error handler fix

leirocks:
should not throw in page allocator, record the first LastError(which is the most interested error) and send back to caller


MikeHolman:
:shipit:

lib/JITServer/JITServer.cpp
```diff
@@ -633,9 +633,10 @@ void ServerContextManager::UnRegisterThreadContext(ServerThreadContext* threadCo
     auto iter = scriptContexts.GetIteratorWithRemovalSupport();
     while (iter.IsValid())
     {
-        if (iter.Current().Key()->GetThreadContext() == threadContext)
-        {
-            iter.Current().Key()->Close();
+        ServerScriptContext* scriptContext = iter.Current().Key();
+        if (scriptContext->GetThreadContext() == threadContext)
```
MikeHolman: ah, was this a timing issue where wrong thread context would be closed?

MikeHolman: wait, actually that doesn't seem possible. what was this change for?

---
In reply to: [82904007](https://github.com/Microsoft/ChakraCore/pull/1712#discussion_r82904007) [](ancestors = 82904007)

PR #1715, opened 2016-10-07
xplat: keep uncaught exception object alive

jianchun:
On xplat when throwing a JS exception, the exception object pointer
is not on stack and could be collected by GC prematually (triggered
by destructors).

To ensure the exception object is kept alive before caught, manually
track it in threadContext data. Throw a C++ wrapper object instead,
and the catch handler must retrieve and clear the exception object
from the wrapper.

Changed all throw and catch code related to previous
JavascriptExceptionObject* to use C++ wrapper JavascriptException.


jianchun:
@curtisman @digitalinfinity @obastemur Please help CR, updated after Curtis' comment.

digitalinfinity:
:shipit:

curtisman:
:shipit:

jianchun:
Thanks @digitalinfinity @curtisman !

PR #1719, opened 2016-10-10
ForInCacheArray may not be allocated before cleanup.  Need a null check.

curtisman:


lib/Runtime/Base/FunctionBody.cpp
```diff
@@ -5355,14 +5355,17 @@ namespace Js
             return;
         }
         ForInCache * forInCacheArray = (ForInCache *)this->GetAuxPtr(AuxPointerType::ForInCacheArray);
-        if (isShutdown)
+        if (forInCacheArray)
```
dilijev: I think in most other places, we are explicit about saying `x != nullptr`. Same here?

PR #1722, opened 2016-10-11
Fix prefast warning for JsDiagEvaluateUtf8

agarwal-sandeep:
JsDiagEvaluateUtf8 calls strlen to get length of passed string and need to
specify that it is null terminated


digitalinfinity:
>     /// <param name="expression">Expression to evaluate.</param>

Consider adding a comment here indicating that expression is null-terminated.

---
Refers to: lib/Jsrt/ChakraDebug.h:599 in 0c344df. [](commit_id = 0c344df908f7cf8968a6480e374b9ae51f8c9e99, deletion_comment = False)

digitalinfinity:
:shipit:

PR #1725, opened 2016-10-11
ByteCodeWriter offset reasoning bug

Cellule:
Revert part of the extended opcode change #1444. There was an error in reasoning about bytecode offset before EncodeT. 
I thought the next thing to be encoded was always the opcode, this is not true if there is a branch Island created there.


Cellule:
@dilijev @curtisman Can you review this please ?

PR #1727, opened 2016-10-12
WASM Xplat: Linux

arunetm:
Enabling Webassembly to build with Linux

### DIFF
--- a/lib/Runtime/ByteCode/OpLayoutsAsmJs.h
+++ b/lib/Runtime/ByteCode/OpLayoutsAsmJs.h
@@ -283,165 +283,167 @@ namespace Js
 #define Uint32x4LayoutType typename SizePolicy::RegSlotType
 #define Uint16x8LayoutType typename SizePolicy::RegSlotType
 #define Uint8x16LayoutType typename SizePolicy::RegSlotType
+#define LAYOUT_FIELDS_HELPER(x, y) x ## y
+#define LAYOUT_FIELDS_DEF(x, y) LAYOUT_FIELDS_HELPER(x, y)
 #define LAYOUT_TYPE_WMS_REG2(layout, t0, t1) \
     template <typename SizePolicy> struct OpLayoutT_##layout\
     {\
-        t0##LayoutType LAYOUT_PREFIX_##t0()0;\
-        t1##LayoutType LAYOUT_PREFIX_##t1()1;\
+        t0##LayoutType LAYOUT_FIELDS_DEF(LAYOUT_PREFIX_##t0(), 0);\
+        t1##LayoutType LAYOUT_FIELDS_DEF(LAYOUT_PREFIX_##t1(), 1);\
     };
 #define LAYOUT_TYPE_WMS_REG3(layout, t0, t1, t2) \
     template <typename SizePolicy> struct OpLayoutT_##layout\
     {\
-        t0##LayoutType LAYOUT_PREFIX_##t0()0;\
-        t1##LayoutType LAYOUT_PREFIX_##t1()1;\
-        t2##LayoutType LAYOUT_PREFIX_##t2()2;\
+        t0##LayoutType LAYOUT_FIELDS_DEF(LAYOUT_PREFIX_##t0(), 0);\
+        t1##LayoutType LAYOUT_FIELDS_DEF(LAYOUT_PREFIX_##t1(), 1);\
+        t2##LayoutType LAYOUT_FIELDS_DEF(LAYOUT_PREFIX_##t2(), 2);\
     };
 
 #define LAYOUT_TYPE_WMS_REG4(layout, t0, t1, t2, t3)\
     template <typename SizePolicy> struct OpLayoutT_##layout\
     {\
-        t0##LayoutType LAYOUT_PREFIX_##t0()0;\
-        t1##LayoutType LAYOUT_PREFIX_##t1()1;\
-        t2##LayoutType LAYOUT_PREFIX_##t2()2;\
-        t3##LayoutType LAYOUT_PREFIX_##t3()3;\
+        t0##LayoutType LAYOUT_FIELDS_DEF(LAYOUT_PREFIX_##t0(), 0);\
+        t1##LayoutType LAYOUT_FIELDS_DEF(LAYOUT_PREFIX_##t1(), 1);\
+        t2##LayoutType LAYOUT_FIELDS_DEF(LAYOUT_PREFIX_##t2(), 2);\
+        t3##LayoutType LAYOUT_FIELDS_DEF(LAYOUT_PREFIX_##t3(), 3);\
     };
 #define LAYOUT_TYPE_WMS_REG5(layout, t0, t1, t2, t3, t4)\
     template <typename SizePolicy> struct OpLayoutT_##layout\
     {\
-        t0##LayoutType LAYOUT_PREFIX_##t0()0;\
-        t1##LayoutType LAYOUT_PREFIX_##t1()1;\
-        t2##LayoutType LAYOUT_PREFIX_##t2()2;\
-        t3##LayoutType LAYOUT_PREFIX_##t3()3;\
-        t4##LayoutType LAYOUT_PREFIX_##t4()4;\
+        t0##LayoutType LAYOUT_FIELDS_DEF(LAYOUT_PREFIX_##t0(), 0);\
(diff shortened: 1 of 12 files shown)


arunetm:
@Cellule @MikeHolman please review.

lib/WasmReader/WasmBinaryReader.cpp
```diff
@@ -174,11 +173,7 @@ WasmBinaryReader::ReadSectionHeader()
 
 #if ENABLE_DEBUG_CONFIG_OPTIONS
     Assert(idSize < 64);
-    char16 buf[64];
-    size_t convertedChars = 0;
-    mbstowcs_s(&convertedChars, buf, idSize + 1, sectionName, _TRUNCATE);
-    buf[idSize] = 0;
-    TRACE_WASM_SECTION(_u("Section Header: %s, length = %u (0x%x)"), buf, sectionSize, sectionSize);
+    TRACE_WASM_SECTION(_u("Section Header: %s, length = %u (0x%x)"), sectionName, sectionSize, sectionSize);
```
Cellule: `sectionName` is a `char*` does it work in with a printf function using a char16 format string ?

lib/WasmReader/WasmSection.h
```diff
@@ -27,8 +27,8 @@ namespace Wasm
     {
         SectionFlag flag;
         SectionCode precedent;
-        char16* name;
-        char* id;
+        const wchar_t* name;
```
Cellule: I thought the point of `char16` was for cross plat

Cellule:
@dotnet-bot test linux tests please

Cellule:
@obastemur  I'm sorry for my ignorance, but what does it mean for a module to be part of the resulting binary or not?

lib/WasmReader/WasmSection.h
jianchun: +1. Please try avoid wchar_t.

---
In reply to: [83068286](https://github.com/Microsoft/ChakraCore/pull/1727#discussion_r83068286) [](ancestors = 83068286)

lib/WasmReader/CMakeLists.txt
```diff
@@ -0,0 +1,20 @@
+add_library (Chakra.WasmReader OBJECT
+
```
jianchun: nit: unneeded extra blank line

lib/Runtime/ByteCode/OpLayoutsAsmJs.h
```diff
@@ -283,165 +283,167 @@ namespace Js
...
 #define Uint16x8LayoutType typename SizePolicy::RegSlotType
 #define Uint8x16LayoutType typename SizePolicy::RegSlotType
+#define LAYOUT_FIELDS_HELPER(x, y) x ## y
+#define LAYOUT_FIELDS_DEF(x, y) LAYOUT_FIELDS_HELPER(x, y)
 #define LAYOUT_TYPE_WMS_REG2(layout, t0, t1) \
     template <typename SizePolicy> struct OpLayoutT_##layout\
     {\
-        t0##LayoutType LAYOUT_PREFIX_##t0()0;\
-        t1##LayoutType LAYOUT_PREFIX_##t1()1;\
+        t0##LayoutType LAYOUT_FIELDS_DEF(LAYOUT_PREFIX_##t0(), 0);\
```
jianchun: >LAYOUT_FIELDS_DEF(LAYOUT_PREFIX_##t0(), 0);\ [](start = 23, length = 44)

Is this to have `()` in a field name (weird, why do we want that)? From the change I assume directly `##0` doesn't work, right?

jianchun:
> #ifdef ENABLE_WASM

Is `ENABLE_WASM` turned on Linux?

---
Refers to: lib/WasmReader/WasmSection.h:8 in de4503e. [](commit_id = de4503ef51fc7bc69496608fd54449d6e72e3ffb, deletion_comment = False)

jianchun:
>         if (!memcmp(SectionInfo::All[i].id, sectionName, idSize))

Does "sectionId" sound a better name for this variable (I was confused and thought this was the same as sectionInfo.name)?

---
Refers to: lib/WasmReader/WasmBinaryReader.cpp:167 in de4503e. [](commit_id = de4503ef51fc7bc69496608fd54449d6e72e3ffb, deletion_comment = False)

jianchun:
> @obastemur  I'm sorry for my ignorance, but what does it mean for a module to be part of the resulting binary or not?

The new Chakra.WasmReader lib is not linked to anything. Kind of related to my other question -- is it enabled on Linux? Suppose it would link to ch and Jsrt.lib?

lib/WasmReader/WasmSection.h
arunetm: Will change this back. The use of "mbstowcs_s" was giving pain in Linux, will look for a workaround.

lib/Runtime/ByteCode/OpLayoutsAsmJs.h
arunetm: () is part of the macro name and it is not going to be part of the actual field name. The tricky part was to have the number to be appended to the end of the filed name to avoid naming conflicts. 
##0 does not work as ## expects strings as operands.

lib/WasmReader/WasmBinaryReader.cpp
jianchun: I'm not sure if `%S` will work here to print `char` string. If that doesn't, try use utf8::NarrowToWide

lib/WasmReader/CMakeLists.txt
arunetm: will remove.

arunetm:
@jianchun, ENABLE_WASM is turned on for Linux. Its on for all debug/test builds. 
@Cellule, can you please verify whether this is the right thing to do?

arunetm:
@obastemur, @jianchun Forgive my ignorance, I missed the step to link Cahkra.WasmReader lib for Linux. Will make this change and update the PR. 
Thanks!

Cellule:
@arunetm We don't need to enable WebAssembly on Linux right now. I don't see the harm, but regardless it is intended to be enable only in debug and test builds at the moment since it is still a WIP

MikeHolman:
@Cellule I think it is useful to have on Linux, if for no other reason than to be able to fuzz with AFL.

Cellule:
@MikeHolman Yes I believe it would be good to have, but what I am wondering is, does it have to be a blocker to merge WebAssembly to master ?

Cellule:
Any update on this ?

arunetm:
@Cellule Currently fixing a linker error and looking for a workaround for mbstowcs_s.  
Will update the PR soon I get to  a solution.

arunetm:
The linker fails to find WasmLibrary::WasmDeferredParseExternalThunk  defined in Language/amd64/amd64_Thunks.asm (called from Library/WasmLibrary.cpp). CMakeLists.txt includes only Language/amd64/amd64_Thunks.S and we don't have the WasmDeferredParseExternalThunk defined in it (hence this issue)
Forgive my ignorance of cmake, but adding  amd64_Thunks.asm to CMakeLists.txt doesn't seem to recognize the format and it leads to many failures.

@jianchun. This is a long shot question but, is it possible to allow CMake to recognize the .asm format and include Language/amd64/amd64_Thunks.asm for the build? 

@Cellule Is there a reason why we have WasmDeferredParseExternalThunk defined in Languages/amd64/amd64_Thunks.asm instead of Library/amd64/JavaScriptFunction.asm since its being called from Library/WasmLibrary.cpp only?  Should we move it?
Also to have amd64_Thunks.S (gnu compliant) implement WasmDeferredParseExternalThunk, does it need to be written again  from scratch? 
thanks in advance.

jianchun:
@arunetm WasmDeferredParseExternalThunk needs to be rewritten in amd64_Thunks.S. You can see that by comparing amd64_Thunks.asm and amd64_Thunks.S.

arunetm:
@Cellule @MikeHolman. I'm starting to rewrite the  WasmDeferredParseExternalThunk inside  amd64_Thunks.S (or any other possible missing pieces) to make it available for xplat builds. Is this path beneficial to do this at this point or should we disable WebAssembly in Linux in preparation to the master integration?

MikeHolman:
@arunetm I think it's beneficial to do, but for the merge to master let's just disable on x-plat.

PR #1728, opened 2016-10-12
handle race condition in OOP JIT

leirocks:
CloseScriptContext call can happen before or in the middle of CodeGen call.
In JIT server call, between detecting contexts alive and increasing the refcount, there can be close and initialize new context call, the new context can get same address from HeapAlloctor and causes codegen succeeded with wrong scriptContext.
Also, there are two calls for cleaning up script context and both can cause server scriptContext delete thus new scriptContext get same address as the deleted one.
Move the context closing stack recording to seperate location, so the context itself can be deleted normally

### DIFF
--- a/lib/JITServer/JITServer.h
+++ b/lib/JITServer/JITServer.h
@@ -8,16 +8,85 @@ class ServerContextManager
 public:
     static void RegisterThreadContext(ServerThreadContext* threadContext);
     static void UnRegisterThreadContext(ServerThreadContext* threadContext);
-    static bool IsThreadContextAlive(ServerThreadContext* threadContext);
 
     static void RegisterScriptContext(ServerScriptContext* scriptContext);
     static void UnRegisterScriptContext(ServerScriptContext* scriptContext);    
-    static bool IsScriptContextAlive(ServerScriptContext* scriptContext);
+
+    static bool CheckLivenessAndAddref(ServerScriptContext* context);
+    static bool CheckLivenessAndAddref(ServerThreadContext* context);
+
 private:
     static JsUtil::BaseHashSet<ServerThreadContext*, HeapAllocator> threadContexts;
     static JsUtil::BaseHashSet<ServerScriptContext*, HeapAllocator> scriptContexts;
     static CriticalSection cs;
+
+#ifdef STACK_BACK_TRACE
+public:
+    template<class T>
+    struct ClosedContextEntry
+    {
+        __declspec(noinline)
+        ClosedContextEntry(T* context)
+            :context(context)
+        {
+            stack = StackBackTrace::Capture(&NoThrowHeapAllocator::Instance, 2);
+        }
+        ~ClosedContextEntry()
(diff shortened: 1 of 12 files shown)


lib/JITServer/JITServer.h
```diff
@@ -8,29 +8,113 @@ class ServerContextManager
...
+
+    static void Shutdown()
+    {
+#ifdef STACK_BACK_TRACE
+        while (!ClosedThreadContextList.Empty())
+        {
+            auto record = ClosedThreadContextList.Pop();
+            if (record)
+            {
+                delete record;
```
curtisman: HeapDelete?

lib/JITServer/JITServer.h
```diff
@@ -8,29 +8,113 @@ class ServerContextManager
...
+            }
+        }
+        while (!ClosedScriptContextList.Empty())
+        {
+            auto record = ClosedScriptContextList.Pop();
+            if (record)
+            {
+                delete record;
```
curtisman: HeapDelete?

curtisman:
:shipit:

lib/Runtime/Base/ScriptContext.cpp
```diff
@@ -1267,12 +1267,6 @@ if (!sourceList)
     void ScriptContext::SetIsClosed()
     {
         this->isClosed = true;
-#if ENABLE_NATIVE_CODEGEN
-        if (m_remoteScriptContextAddr)
-        {
-            JITManager::GetJITManager()->CloseScriptContext(m_remoteScriptContextAddr);
```
MikeHolman: Having thought more about this, I think we should have Close call still. With cleanup only, I think it may prevent us from short circuiting JIT on scripts that are closing (I think the JIT work item will keep the ScriptContext alive?)

However, Close should never actually free the ServerScriptContext (to prevent the races we see now).

leirocks: what do you want the server to do if the scriptContext is closed but not cleaned up, rejecting the call? if yes we can just move current cleanup call to client scriptContext::Close(). if you allow the server call after scriptContext is closed in client side, current implementation is already reflecting that. also, because of the racing, neither rejecting nor allow can be strictly enforced, unless the client side can change to not sending other calls after scriptContext is closed or pending close.

---
In reply to: [83082897](https://github.com/Microsoft/ChakraCore/pull/1728#discussion_r83082897) [](ancestors = 83082897)

MikeHolman: If it is closed but not cleaned up, we don't need to reject the call. It should really only be for telling codegen that it can give up on any current JIT calls (we have periodic checks for IsClosed during codegen).

---
In reply to: [83088845](https://github.com/Microsoft/ChakraCore/pull/1728#discussion_r83088845) [](ancestors = 83088845,83082897)

MikeHolman:
:shipit:

PR #1729, opened 2016-10-12
Remove constexpr from GetOffsetOfpszValue() and GetOffsetOfcharLength()

Nostracodus:
The Microsoft C++ compiler defines offsetof() as a macro that uses
reinterpret_cast to calculate the offset of a member. reinterpret_cast
cannot be used in constant expressions. MSVC has been lax in this regard
in the past, but using offsetof() in a constexpr function definition is
illegal with the current implementation of offsetof().

This change adheres to the letter of the law by making GetOffsetOfpszValue()
and GetOffsetOfcharLength() not constexpr and converting the static asserts
that utilize these methods into runtime asserts. In the future, if the
implementation of offsetof() changes in MSVC, these can be converted back
into static_assert.

### DIFF
--- a/lib/Backend/JITRecyclableObject.h
+++ b/lib/Backend/JITRecyclableObject.h
@@ -55,10 +55,8 @@ public:
 
     static JITJavascriptString * FromVar(Js::Var var)
     {
-#ifdef HAS_CONSTEXPR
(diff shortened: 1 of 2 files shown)


MikeHolman:
lgtm

PR #1731, opened 2016-10-13
Revert change to confirm perf regression

Penguinwizzard:
Revert "Add test, fix nits."

This reverts commit cbef9d84c5751f61c5d264169062e9b353ef5016.

Revert "If we have no user crash handler we should generate additional info."

This reverts commit e8b715d37227834f404e7c6caecfdbe694559280.


Penguinwizzard:
@kunalspathak please review

kunalspathak:
:shipit:

PR #1710, opened 2016-10-06
Wasm: spec tests + unreachable state + bug fixes

Cellule:
Update wasm testsuite files from spec repo.
Removed unreachable type and use Unreachable state instead. Simplify how we handle unreachable to be more spec compliant.
Trap if we want to access the memory without a memory available

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/1710)

PR #1734, opened 2016-10-13
WASM Globals implementation

Krovatkin:
This commit implements WASM globals as outlined in wasm spec. Namely, we support 
* init_exprs (consts and get_globals)
* import globals
* export globals

@Cellule @MikeHolman please take a look when you have a mnt. This version should be aligned with the way you guys were envisioning it.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/1734)


Cellule:
I just merged #1710, I suspect you will have to rebase and fix some baselines in the spec test suite

Krovatkin:
@dotnet-bot test this please

Krovatkin:
@dotnet-bot test this please

Krovatkin:
Rebased against Cellule's latest changes. Getting some weird I/O issues with build jobs. Restarted the tests.

Krovatkin:
Rebased once more, fixed the tests.

Cellule:
Reviewed 23 of 23 files at r1.
Review status: all files reviewed at latest revision, 17 unresolved discussions.

---

*[lib/Runtime/Library/WasmLibrary.cpp, line 289 at r1](https://reviewable.io:443/reviews/microsoft/chakracore/1734#-KU4gIYt68FFDaE6tbZU:-KU4gIYuq4vPuQENxIM8:b-l1d4pq) ([raw file](https://github.com/microsoft/chakracore/blob/9dafd078fcdde64d5ea70bc6ea9045e76157ffe1/lib/Runtime/Library/WasmLibrary.cpp#L289)):*
> ```cpp
>                     case Wasm::WasmTypes::I64:
>                     default:
>                         Assert(UNREACHED);
> ```

Could you set obj to undefined in this case so we don't have undetermined behavior in release/test builds

---

*[lib/Runtime/Library/WasmLibrary.cpp, line 306 at r1](https://reviewable.io:443/reviews/microsoft/chakracore/1734#-KU4gnxlBMEyASewIBiW:-KU4gnxlBMEyASewIBiX:b-5wn6j9) ([raw file](https://github.com/microsoft/chakracore/blob/9dafd078fcdde64d5ea70bc6ea9045e76157ffe1/lib/Runtime/Library/WasmLibrary.cpp#L306)):*
> ```cpp
>         Var modProp = JavascriptOperators::OP_GetProperty(ffi, modPropertyRecord->GetPropertyId(), ctx);
> 
>         PropertyRecord const * propertyRecord = nullptr;
> ```

nit: This is used only in `nameLen > 0` case, we can move the definition there

---

*[lib/Runtime/Library/WasmLibrary.cpp, line 362 at r1](https://reviewable.io:443/reviews/microsoft/chakracore/1734#-KU4hWj655ppEaOSUjVO:-KU4hWj655ppEaOSUjVP:b-d5og0x) ([raw file](https://github.com/microsoft/chakracore/blob/9dafd078fcdde64d5ea70bc6ea9045e76157ffe1/lib/Runtime/Library/WasmLibrary.cpp#L362)):*
> ```cpp
>             global->SetReferenceType(Wasm::WasmGlobal::Const);
> ```

nit: double newline

---

*[lib/Runtime/Library/WasmLibrary.cpp, line 373 at r1](https://reviewable.io:443/reviews/microsoft/chakracore/1734#-KU4ijSJBG9oJs5njWZO:-KU4ijSJBG9oJs5njWZP:b-s9tqnb) ([raw file](https://github.com/microsoft/chakracore/blob/9dafd078fcdde64d5ea70bc6ea9045e76157ffe1/lib/Runtime/Library/WasmLibrary.cpp#L373)):*
> ```cpp
>             {
>                 int val;
>                 bool isInt32;
> ```

`isInt32` is only used/initialized in not taggedInt path, you should move it there

---

*[lib/Runtime/Library/WasmLibrary.cpp, line 382 at r1](https://reviewable.io:443/reviews/microsoft/chakracore/1734#-KU4j3AR7O6jzCjkCTYN:-KU4j3AR7O6jzCjkCTYO:brc2ism) ([raw file](https://github.com/microsoft/chakracore/blob/9dafd078fcdde64d5ea70bc6ea9045e76157ffe1/lib/Runtime/Library/WasmLibrary.cpp#L382)):*
> ```cpp
>                 {
>                     bool isInt = JavascriptNumber::TryGetInt32OrUInt32Value(JavascriptNumber::GetValue(prop), &val, &isInt32);
>                     if (!isInt)
> ```

Does it really need to be an int32 or can we just convert it to int32 ?

---

*[lib/Runtime/Library/WasmLibrary.cpp, line 399 at r1](https://reviewable.io:443/reviews/microsoft/chakracore/1734#-KU4jGBSCRG-XPvR6p8t:-KU4jGBSCRG-XPvR6p8u:b-wlurzb) ([raw file](https://github.com/microsoft/chakracore/blob/9dafd078fcdde64d5ea70bc6ea9045e76157ffe1/lib/Runtime/Library/WasmLibrary.cpp#L399)):*
> ```cpp
>             case Wasm::WasmTypes::F64:
>             {
>                 double val = (double)JavascriptNumber::GetValue(prop);
> ```

nit: useless cast

---

*[lib/Runtime/Library/WasmLibrary.cpp, line 433 at r1](https://reviewable.io:443/reviews/microsoft/chakracore/1734#-KU4jmXvd4BpLVpc_PRS:-KU4jmXvd4BpLVpc_PRT:b-isi981) ([raw file](https://github.com/microsoft/chakracore/blob/9dafd078fcdde64d5ea70bc6ea9045e76157ffe1/lib/Runtime/Library/WasmLibrary.cpp#L433)):*
> ```cpp
>                 global->cnst = sourceGlobal->cnst;
> 
>                 Assert(sourceGlobal->GetReferenceType() == Wasm::WasmGlobal::Const);
> ```

What if sourceGlobal is also a LocalReference that will get resolved later ? Shouldn't we check and throw on this case ?

---

*[lib/WasmReader/WasmBinaryReader.cpp, line 843 at r1](https://reviewable.io:443/reviews/microsoft/chakracore/1734#-KU44ImouIo9k3bXNR4P:-KU44ImouIo9k3bXNR4Q:b-65enfj) ([raw file](https://github.com/microsoft/chakracore/blob/9dafd078fcdde64d5ea70bc6ea9045e76157ffe1/lib/WasmReader/WasmBinaryReader.cpp#L843)):*
> ```cpp
>         m_funcState.size = (UINT)(m_end - m_pc);
> 
>         WasmOp op = ReadExpr();
> ```

There's a method that was added recently called `ReadInitExpr` to do this part. You'll still have to check the node returned by the method though.

---

*[lib/WasmReader/WasmBinaryReader.cpp, line 939 at r1](https://reviewable.io:443/reviews/microsoft/chakracore/1734#-KU48iGpCGTjSWus__nR:-KU48iGpCGTjSWus__nS:b-q91b9t) ([raw file](https://github.com/microsoft/chakracore/blob/9dafd078fcdde64d5ea70bc6ea9045e76157ffe1/lib/WasmReader/WasmBinaryReader.cpp#L939)):*
> ```cpp
>             importedGlobal->importVar = wi;
>             importedGlobal->SetReferenceType(WasmGlobal::ImportedReference);
>             m_module->globals.Add(importedGlobal);
> ```

Could you encapsulate this operation in the module instead of calling `globals.Add` directly

---

*[lib/WasmReader/WasmBinaryReader.cpp, line 953 at r1](https://reviewable.io:443/reviews/microsoft/chakracore/1734#-KU47Ob5obEhKPC9Ok4t:-KU47Ob5obEhKPC9Ok4u:bsv94ra) ([raw file](https://github.com/microsoft/chakracore/blob/9dafd078fcdde64d5ea70bc6ea9045e76157ffe1/lib/WasmReader/WasmBinaryReader.cpp#L953)):*
> ```cpp
>     }
>     m_module->SetImportGlobalCount(importGlobalCount);
>     m_module->SetImportCount(importFunctionCount);
> ```

This is incorrect. 
Take the following example, you import 1 global and 1 function. The global is first.
Then you will call `m_module->SetFunctionImport(1, ...)` for the function, but afterward, you will set the number of imports to 1.
The problem is, later we'll think there is only 1 imported function and it should be at index 0, however, at index 0 we have nullptr.

We should use `importFunctionCount` for the index where we set the function import in the module

---

*[lib/WasmReader/WasmByteCodeGenerator.cpp, line 1492 at r1](https://reviewable.io:443/reviews/microsoft/chakracore/1734#-KU4ktBSGsR3_NVwsxom:-KU4ktBT90L89iPr6rUh:bpr1113) ([raw file](https://github.com/microsoft/chakracore/blob/9dafd078fcdde64d5ea70bc6ea9045e76157ffe1/lib/WasmReader/WasmByteCodeGenerator.cpp#L1492)):*
> ```cpp
>         blockInfo.yieldInfo->didYield = true;
>         m_writer.AsmReg2(GetLoadOp(expr.type), yieldInfo.location, expr.location);
>         //ReleaseLocation(&expr);
> ```

nit: forgot to remove this ?

---

*[lib/WasmReader/WasmModule.cpp, line 360 at r1](https://reviewable.io:443/reviews/microsoft/chakracore/1734#-KU4AABjqAngjoMtTBSg:-KU4AABjqAngjoMtTBSh:bsfolg9) ([raw file](https://github.com/microsoft/chakracore/blob/9dafd078fcdde64d5ea70bc6ea9045e76157ffe1/lib/WasmReader/WasmModule.cpp#L360)):*
> ```cpp
>     WasmTypes::WasmType type = global->GetType();
> 
>     for (uint i = 0; i < (uint) type; i++)
> ```

nit: add braces

---

*[lib/WasmReader/WasmModule.h, line 13 at r1](https://reviewable.io:443/reviews/microsoft/chakracore/1734#-KU49ANFAbtf3Z9vA7mU:-KU49ANFAbtf3Z9vA7mV:b4d8rsq) ([raw file](https://github.com/microsoft/chakracore/blob/9dafd078fcdde64d5ea70bc6ea9045e76157ffe1/lib/WasmReader/WasmModule.h#L13)):*
> ```objc
>     class WasmBinaryReader;
> 
>     static const uint TYPES_COUNT = 5;
> ```

use `WasmTypes::Limit` instead

---

*[lib/WasmReader/WasmModule.h, line 59 at r1](https://reviewable.io:443/reviews/microsoft/chakracore/1734#-KU49LmioCLo2DgxAi3C:-KU49LmioCLo2DgxAi3D:b-uvlq60) ([raw file](https://github.com/microsoft/chakracore/blob/9dafd078fcdde64d5ea70bc6ea9045e76157ffe1/lib/WasmReader/WasmModule.h#L59)):*
> ```objc
>         void AllocateFunctionExports(uint32 entries);
>         uint GetExportCount() const { return m_exportCount; }
>         void SetFunctionExport(uint32 iExport, uint32 funcIndex, char16* exportName, uint32 nameLength, ImportKinds::ImportKind kind);
> ```

Should rename this to `SetExport` instead since it is no longer exclusive to functions

---

*[lib/WasmReader/WasmParseTree.h, line 25 at r1](https://reviewable.io:443/reviews/microsoft/chakracore/1734#-KU4Bc00LP8RufKdXU7A:-KU4Bc00LP8RufKdXU7B:bnl316q) ([raw file](https://github.com/microsoft/chakracore/blob/9dafd078fcdde64d5ea70bc6ea9045e76157ffe1/lib/WasmReader/WasmParseTree.h#L25)):*
> ```objc
>     }
> 
>     namespace ImportKinds
> ```

I think I was wrong to call this type `ImportKinds` if I recall, you initially named that `ExternalKinds` maybe we should rename this to match how the spec calls it.

---

*[test/wasm/global.js, line 1 at r1](https://reviewable.io:443/reviews/microsoft/chakracore/1734#-KU4Cx4wQ8G0xdmT_X3Y:-KU4Cx4xIvbUFouMM1-E:b7tatk) ([raw file](https://github.com/microsoft/chakracore/blob/9dafd078fcdde64d5ea70bc6ea9045e76157ffe1/test/wasm/global.js#L1)):*
> ```JavaScript
> ```

Does this test bring anything more than what the spec tests does ? it looks pretty much the same.
I did realize that I incorrectly regenerated the spec testsuite otherwise you would have had to update those baselines.

---

*[test/wasm/global.js, line 7 at r1](https://reviewable.io:443/reviews/microsoft/chakracore/1734#-KU4Bn_ZKqQ0bvtFSQ_p:-KU4Bn_ZKqQ0bvtFSQ_q:biig4ma) ([raw file](https://github.com/microsoft/chakracore/blob/9dafd078fcdde64d5ea70bc6ea9045e76157ffe1/test/wasm/global.js#L7)):*
> ```JavaScript
> 
> function printAll (a) {
>     
> ```

nit: trailing whitespace and useless newline (same for `printExportedGlobals`)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/microsoft/chakracore/1734)*

Krovatkin:
Review status: 15 of 23 files reviewed at latest revision, 17 unresolved discussions.

---

*[lib/Runtime/Library/WasmLibrary.cpp, line 382 at r1](https://reviewable.io:443/reviews/microsoft/chakracore/1734#-KU4j3AR7O6jzCjkCTYN:-KUIglAl_vSTKCiwx-h8:b-sljbgu) ([raw file](https://github.com/microsoft/chakracore/blob/9dafd078fcdde64d5ea70bc6ea9045e76157ffe1/lib/Runtime/Library/WasmLibrary.cpp#L382)):*
> Does it really need to be an int32 or can we just convert it to int32 ?
</details>
I'm not sure I completely understand the question  ¯\_(ツ)_/¯ 

---

*[lib/WasmReader/WasmBinaryReader.cpp, line 843 at r1](https://reviewable.io:443/reviews/microsoft/chakracore/1734#-KU44ImouIo9k3bXNR4P:-KUIfhAamrW00Ug3KdsG:b-qpv3fk) ([raw file](https://github.com/microsoft/chakracore/blob/9dafd078fcdde64d5ea70bc6ea9045e76157ffe1/lib/WasmReader/WasmBinaryReader.cpp#L843)):*
> There's a method that was added recently called `ReadInitExpr` to do this part. You'll still have to check the node returned by the method though.
</details>
Specialized your `ReadInitExpr` to specifically handle reading int consts. Refactored mine into a separate method.  They don't share that much in common 

---

*[lib/WasmReader/WasmModule.h, line 59 at r1](https://reviewable.io:443/reviews/microsoft/chakracore/1734#-KU49LmioCLo2DgxAi3C:-KUIgZOxiBPe9ZoZpcD9:b-jc7h27) ([raw file](https://github.com/microsoft/chakracore/blob/9dafd078fcdde64d5ea70bc6ea9045e76157ffe1/lib/WasmReader/WasmModule.h#L59)):*
> Should rename this to `SetExport` instead since it is no longer exclusive to functions
</details>
Oops. Missed that one... Will address in another commit 

---

*[test/wasm/global.js, line 1 at r1](https://reviewable.io:443/reviews/microsoft/chakracore/1734#-KU4Cx4wQ8G0xdmT_X3Y:-KUIgKvGUYO5Hl33ky6T:b-qgchmp) ([raw file](https://github.com/microsoft/chakracore/blob/9dafd078fcdde64d5ea70bc6ea9045e76157ffe1/test/wasm/global.js#L1)):*
> Does this test bring anything more than what the spec tests does ? it looks pretty much the same.
> I did realize that I incorrectly regenerated the spec testsuite otherwise you would have had to update those baselines.
</details>
I think so. Import tests in the spec test suite are rather simple. Mine import at least a couple of variables of different types to make sure we are getting the offset computations right :-) 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/microsoft/chakracore/1734)*

Krovatkin:
@Cellule plz take one more look when you have a mnt. thanks! fyi @arunetm

Cellule:
Reviewed 8 of 8 files at r2.
Review status: all files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[lib/Runtime/Library/WasmLibrary.cpp, line 382 at r1](https://reviewable.io:443/reviews/microsoft/chakracore/1734#-KU4j3AR7O6jzCjkCTYN:-KUKC-qcplRtCArdEPEM:b-q2lq51) ([raw file](https://github.com/microsoft/chakracore/blob/9dafd078fcdde64d5ea70bc6ea9045e76157ffe1/lib/Runtime/Library/WasmLibrary.cpp#L382)):*
<details><summary><i>Previously, Krovatkin (Nick Korovaiko) wrote…</i></summary>
> I'm not sure I completely understand the question  ¯\_(ツ)_/¯ 
</details>
If the imported value is `3.14` it is not an int32. However, is it legal (by wasm spec) to import this into an int32 and coerce it to `3` ?

---

*[lib/WasmReader/WasmBinaryReader.cpp, line 843 at r1](https://reviewable.io:443/reviews/microsoft/chakracore/1734#-KU44ImouIo9k3bXNR4P:-KUKF77k6U7JJ9AAQASS:b8wpt8a) ([raw file](https://github.com/microsoft/chakracore/blob/9dafd078fcdde64d5ea70bc6ea9045e76157ffe1/lib/WasmReader/WasmBinaryReader.cpp#L843)):*
<details><summary><i>Previously, Krovatkin (Nick Korovaiko) wrote…</i></summary>
> Specialized your `ReadInitExpr` to specifically handle reading int consts. Refactored mine into a separate method.  They don't share that much in common 
</details>
This is not going to be a normal case. The check for int32 const is not by spec, it was just Not Yet Implemented.
Data Segments, Table Elements and Globals can be initialized using `init_expr` and they can all be initialized by both globals and consts.

That's why the check for int32.const wasn't in ReadInitExpr() as it shouldn't be there.

There should be 1 function called ReadInitExpr that returns details about the init_expr then the caller consumes it accordingly.

---

*[lib/WasmReader/WasmModule.h, line 89 at r2](https://reviewable.io:443/reviews/microsoft/chakracore/1734#-KUKEuMN5QxergJvaVlD:-KUKEuMOhbCsbTvLZLBp:b-suq4i7) ([raw file](https://github.com/microsoft/chakracore/blob/48f48b0fc0466f7093e1f98561507340be172b7b/lib/WasmReader/WasmModule.h#L89)):*
> ```objc
>         uint GetOffsetForGlobal(WasmGlobal* global);
>         void SetImportGlobalCount(uint count) { importGlobalCount = count;  }
>         uint GetImportGlobalCount() { return importGlobalCount; }
> ```

return `globals.Count()` here and drop `SetImportGlobalCount` method  and `importGlobalCount` field

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/microsoft/chakracore/1734)*

Krovatkin:
Review status: all files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[lib/Runtime/Library/WasmLibrary.cpp, line 382 at r1](https://reviewable.io:443/reviews/microsoft/chakracore/1734#-KU4j3AR7O6jzCjkCTYN:-KUSiRwHrjxAp2dHLdp3:b-yduzl7) ([raw file](https://github.com/microsoft/chakracore/blob/9dafd078fcdde64d5ea70bc6ea9045e76157ffe1/lib/Runtime/Library/WasmLibrary.cpp#L382)):*
> If the imported value is `3.14` it is not an int32. However, is it legal (by wasm spec) to import this into an int32 and coerce it to `3` ?
</details>
This seems to be a very difficult question to answer :-)

**wabt** doesn't implement importing globals just yet. Additionally, the reference implementation (**ml-proto**) gives me some weird results: if I run this

```
(module
  (global (export "global") f32 (f32.const 44.13))
)

(register "test")

  (module (import "test" "global" (global i32))
  (global $g2 i32 (get_global 0))
   (func (export "get-0") (result i32) (get_global 0))
   (func (export "get-1") (result i32) (get_global $g2))
  )
  
  (invoke "get-0")
  (invoke "get-1")
```

prints out the original floating value in both cases.

Given that WASM enforces strong typing everywhere else I'd say we should probably avoid coercion. Even if we turned out to be wrong, an error will make things pretty obvious. In contrast, coercion errors might be more insidious to catch later on.  

---

*[lib/WasmReader/WasmBinaryReader.cpp, line 843 at r1](https://reviewable.io:443/reviews/microsoft/chakracore/1734#-KU44ImouIo9k3bXNR4P:-KUSmOnP8w3hws3eG8wK:bdoyezv) ([raw file](https://github.com/microsoft/chakracore/blob/9dafd078fcdde64d5ea70bc6ea9045e76157ffe1/lib/WasmReader/WasmBinaryReader.cpp#L843)):*
> This is not going to be a normal case. The check for int32 const is not by spec, it was just Not Yet Implemented.
> Data Segments, Table Elements and Globals can be initialized using `init_expr` and they can all be initialized by both globals and consts.
> 
> That's why the check for int32.const wasn't in ReadInitExpr() as it shouldn't be there.
> 
> There should be 1 function called ReadInitExpr that returns details about the init_expr then the caller consumes it accordingly.
</details>
Got ya. Reversed the change and I'm now using `ReadInitExpr` in `ReadGlobalsSection`

---

*[lib/WasmReader/WasmBinaryReader.cpp, line 953 at r1](https://reviewable.io:443/reviews/microsoft/chakracore/1734#-KU47Ob5obEhKPC9Ok4t:-KUSoEY09RHQwdZX0AbX:b-7ql1td) ([raw file](https://github.com/microsoft/chakracore/blob/9dafd078fcdde64d5ea70bc6ea9045e76157ffe1/lib/WasmReader/WasmBinaryReader.cpp#L953)):*
> This is incorrect. 
> Take the following example, you import 1 global and 1 function. The global is first.
> Then you will call `m_module->SetFunctionImport(1, ...)` for the function, but afterward, you will set the number of imports to 1.
> The problem is, later we'll think there is only 1 imported function and it should be at index 0, however, at index 0 we have nullptr.
> 
> We should use `importFunctionCount` for the index where we set the function import in the module
</details>
fixed

---

*[lib/WasmReader/WasmModule.h, line 89 at r2](https://reviewable.io:443/reviews/microsoft/chakracore/1734#-KUKEuMN5QxergJvaVlD:-KUSmbB3xanDnU1TaoXV:b-yaldc5) ([raw file](https://github.com/microsoft/chakracore/blob/48f48b0fc0466f7093e1f98561507340be172b7b/lib/WasmReader/WasmModule.h#L89)):*
> return `globals.Count()` here and drop `SetImportGlobalCount` method  and `importGlobalCount` field
</details>
That might *not* necessarily work. I actually need *the number of imported global variables" to simplify the iteration logic. `globals.Count()` after processing of the global section will return *the total number of globals*.  Strictly speaking, we can get by even w/o that count, but the iteration logic will look much uglier. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/microsoft/chakracore/1734)*

Krovatkin:
Review status: 16 of 23 files reviewed at latest revision, 4 unresolved discussions.

---

*[lib/Runtime/Library/WasmLibrary.cpp, line 382 at r1](https://reviewable.io:443/reviews/microsoft/chakracore/1734#-KU4j3AR7O6jzCjkCTYN:-KUSv1w2dFj_xOMLREPP:b-7izkds) ([raw file](https://github.com/microsoft/chakracore/blob/9dafd078fcdde64d5ea70bc6ea9045e76157ffe1/lib/Runtime/Library/WasmLibrary.cpp#L382)):*
<details><summary><i>Previously, Krovatkin (Nick K) wrote…</i></summary>
> This seems to be a very difficult question to answer :-)
> 
> **wabt** doesn't implement importing globals just yet. Additionally, the reference implementation (**ml-proto**) gives me some weird results: if I run this
> 
> ```
> (module
>   (global (export "global") f32 (f32.const 44.13))
> )
> 
> (register "test")
> 
>   (module (import "test" "global" (global i32))
>   (global $g2 i32 (get_global 0))
>    (func (export "get-0") (result i32) (get_global 0))
>    (func (export "get-1") (result i32) (get_global $g2))
>   )
>   
>   (invoke "get-0")
>   (invoke "get-1")
> ```
> 
> prints out the original floating value in both cases.
> 
> Given that WASM enforces strong typing everywhere else I'd say we should probably avoid coercion. Even if we turned out to be wrong, an error will make things pretty obvious. In contrast, coercion errors might be more insidious to catch later on.  
</details>
On the second thought, maybe we should allow coercion. Otherwise, it's hard to tell how this entire thing is supposed to interface with JS :-)))))))))))) 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/microsoft/chakracore/1734)*

Krovatkin:
@Cellule rebased against your latest changes.
On the bright side, 
* we have a very basic support for `get_global` in **data segments**
* **importGlobalCount** is gone now 

On the rainy side,
* still got no idea what to do about coercion of imported globals 
* int64 seems to render globals test in spec suite useless :-(

MikeHolman:
@Krovatkin What do you mean about imported globals?

Cellule:
:lgtm:

---

Reviewed 4 of 7 files at r3, 9 of 10 files at r4, 1 of 1 files at r5.
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

*[lib/WasmReader/WasmDataSegment.cpp, line 22 at r5](https://reviewable.io:443/reviews/microsoft/chakracore/1734#-KU_YZmbxiN-Mf3ObiFh:-KU_YZmbxiN-Mf3ObiFi:btaummo) ([raw file](https://github.com/microsoft/chakracore/blob/63fccf3f5892a3ee6d254532bd8cc4a3f7cf925a/lib/WasmReader/WasmDataSegment.cpp#L22)):*
> ```cpp
> 
> uint32
> WasmDataSegment::getDestAddr(WasmModule* module) const
> ```

Nice addition to the PR

---

*[test/WasmSpec/baselines/globals.baseline, line 2 at r5](https://reviewable.io:443/reviews/microsoft/chakracore/1734#-KU_YyYXuLAJS30XiFhi:-KU_YyYXuLAJS30XiFhj:b-yj7ou5) ([raw file](https://github.com/microsoft/chakracore/blob/63fccf3f5892a3ee6d254532bd8cc4a3f7cf925a/test/WasmSpec/baselines/globals.baseline#L2)):*
> ```
> Unexpected Error while compiling globals.0.wasm
> Error: Compiling wasm failed: i64 globals NYI
> ```

Yeah this is annoying, hopefully we'll land int64 soon

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/microsoft/chakracore/1734#-:-KU_ZBdbk-5a1QYliTRy:bnfp4nl)*

arunetm:
@Cellule, I am preparing to merge this change. Thanks!

PR #1736, opened 2016-10-13
Functional improvements and fixes for redeferral. 

pleath:
1) Re-use existing FunctionBody/ParseableFunctionInfo's in the case where we re-compile a function that encloses already-compiled functions. This is necessary to avoid having multiple FunctionInfo structures around for a single FunctionBody, not all of which get updated when things change. It also reduces the cost of re-compiling. 2) Change the FunctionBody's nested function array to hold FunctionInfo's instead of FunctionProxy's. This fixes a class of functional issues involving the need to update the nested array as nested functions are un-deferred and re-deferred. The FunctionInfo pointers don't change and thus don't need updating. 3) Eliminate the m_referenceInParentFunction member of FunctionProxy, which was there to enable the nested array updates that we no longer need. 4) Re-use ScopeInfo in the case where a parent of the current function has been re-deferred.


pleath:
Out-of-date PR, incorporated into https://github.com/Microsoft/ChakraCore/pull/1585.

PR #1737, opened 2016-10-13
Changes addressing CVE_2016-3382, CVE-2016-3386, CVE-2016-3389, CVE-2016-3390, CVE-2016-7189, and a mitigation of a CFG bypass.

pleath:
Details:

Calls that target the external thunks should decrement the callinfo param count if the extra-param flag is set.

Don't optimize spread operation in a parameter list if the array we're spreading may have gaps. Accessing an element in the prototype chain may have side-effects that invalidate the optimization.

Port disabling of UT that times out

Type confusion in JavascriptArray
Type confusion in JavascriptArray::TemplatedGetItem()
    Array.prototype.join()
    Array.prototype.indexOf()
    Array.prototype.lastIndexOf()

Type confusion in JavascriptArray::TemplatedGetItem()
    Function.prototype.apply()

Type confusion in JavascriptArray::MapHelper()
    Array.prototype.map()

CRC computation and validation for the encoder buffer

Premise:
Encoder phase takes longer time for a relatively larger function. So the buffer to which we write the encoded bytes will be RWX all the while till it completes the encoding.
This time is big enough for the main thread to write in this region.
We then transfer the data to the final buffer and execute the code in the buffer(which, now, also contains the modified code).

Mitigation:
We can check the integrity of the buffer data using CRC32(Cyclic Redundancy Check) at suitable spots.
Following is the mechanism for validation:
- Start with a random CRC seed.
- Compute the CRC1 during the encoding phase.
- Validate the CRC1 during branch shortening.
- Compute CRC2 during branch shortening.
- Validate CRC2 (or CRC1, if branch shortening didn't happen) after copying the entire buffer to the final RX buffer.
- Finally, register the entry point as a valid CFG target.

CRC32 Intrinsic instruction is available only on SSE4 and above. Hence for other cases, CRC32 algorithm is implemented.
We were storing LabelInstr* directly in the encoded bytes - Moved it to be stored in a property. - To enable CRC calculation.

Perf results:
No visible changes in console benchmark run (desktop and low-memory device).

Fixes to use-after-free in Globopt, Lowering.

Tail duplication consists of the following code :

 branchEntry->ReplaceTarget(mergeLabel, tailBranch->GetTarget());
 instr = branchEntry;

branchEntry is a reference to a SList node that can get deleted within
ReplaceTarget function. Subsequent use of the same reference is referring
to a freed value. Fix by caching branchEntry before ReplaceTarget.

Lowering floor builtin code creates a 'zero' MemRefOpnd, which gets passed
through Legalizer, which can delete the Opnd. Subsequent uses of the
MemRefOpnd in Lowering refers to a freed value. This is fixed by
AutoReuseOpnd which will avoid this scenario.

Correcting the version check for SSE4


pleath:
@meg-gupta , @satheeshravi , @suwc , please review your individual changes. (Thanks.)

satheeshravi:
LGTM

meg-gupta:
LGTM

suwc:
:shipit:

suwc:
LGTM

PR #1739, opened 2016-10-13
Wasm: Bug fix in the unreachable state

Cellule:
We need to reset the Unreachable state before emitting the Else block in an if/else expression.
Otherwise we won't emit the code the the else block thinking we are in unreachable code

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/1739)


Cellule:
@MikeHolman @Krovatkin @arunetm @commonlisp Can you guys review this please ?

Krovatkin:
**LGTM**, though i need to catch up on the exact implementation details of `unreachable`

PR #1744, opened 2016-10-14
osx: JIT support

obastemur:
This PR;
- Enables JIT on macOS
- Fixes / improvements to xplat exception handling

### DIFF
--- a/lib/Common/Memory/amd64/XDataAllocator.cpp
+++ b/lib/Common/Memory/amd64/XDataAllocator.cpp
@@ -12,6 +12,10 @@ CompileAssert(false)
 #include "XDataAllocator.h"
 #include "Core/DelayLoadLibrary.h"
 
+#ifndef _WIN32
+#include "PlatformAgnostic/AssemblyCommon.h" // __REGISTER_FRAME / __DEREGISTER_FRAME
+#endif
+
 XDataAllocator::XDataAllocator(BYTE* address, uint size, HANDLE processHandle) :
     freeList(nullptr),
     start(address),
@@ -155,7 +159,7 @@ void XDataAllocator::Register(XDataAllocation * xdataInfo, ULONG_PTR functionSta
 
 #else  // !_WIN32
     Assert(ReadHead(xdataInfo->address));  // should be non-empty .eh_frame
-    __register_frame(xdataInfo->address);
+    __REGISTER_FRAME(xdataInfo->address);
 #endif
 }
 
@@ -176,6 +180,6 @@ void XDataAllocator::Unregister(XDataAllocation * xdataInfo)
 
 #else  // !_WIN32
     Assert(ReadHead(xdataInfo->address));  // should be non-empty .eh_frame
-    __deregister_frame(xdataInfo->address);
+    __DEREGISTER_FRAME(xdataInfo->address);
 #endif
 }
--- a/lib/Common/PlatformAgnostic/AssemblyCommon.h
+++ b/lib/Common/PlatformAgnostic/AssemblyCommon.h
@@ -1,0 +1,29 @@
+//-------------------------------------------------------------------------------------------------------
+// Copyright (C) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE.txt file in the project root for full license information.
+//-------------------------------------------------------------------------------------------------------
+#pragma once
+
+#ifndef _WIN32
+
+extern "C" void __register_frame(const void* ehframe);
+extern "C" void __deregister_frame(const void* ehframe);
+
+#if defined(_AMD64_) && !defined(DISABLE_JIT)
+#if defined(__APPLE__) // no multi fde support
+#ifndef __IOS__
+typedef void(*mac_fde_reg_op)(const void*addr);
+void mac_fde_wrapper(const char *dataStart, mac_fde_reg_op reg_op);
+#define __REGISTER_FRAME(addr) mac_fde_wrapper((const char*)addr, __register_frame)
+#define __DEREGISTER_FRAME(addr) mac_fde_wrapper((const char*)addr, __deregister_frame)
+#else // __IOS__ // no JIT
+#define __REGISTER_FRAME(addr)
+#define __DEREGISTER_FRAME(addr)
+#endif // !__IOS__
+#else // multi FDE support
+#define __REGISTER_FRAME(addr) __register_frame(addr)
+#define __DEREGISTER_FRAME(addr) __deregister_frame(addr)
+#endif // __APPLE__
+#endif // _AMD64_ && !DISABLE_JIT
+
+#endif // !_WIN32
--- a/lib/Runtime/ByteCode/ByteCodeWriter.h
+++ b/lib/Runtime/ByteCode/ByteCodeWriter.h
@@ -96,6 +96,7 @@ namespace Js
 
             template <LayoutSize layoutSize> void EncodeOpCode(uint16 op, ByteCodeWriter* writer);
             template <> void EncodeOpCode<SmallLayout>(uint16 op, ByteCodeWriter* writer);
+
             // EncodeT functions return the global offset where the opcode has been encoded
             template <LayoutSize layoutSize> uint EncodeT(OpCode op, ByteCodeWriter* writer);
             template <LayoutSize layoutSize> uint EncodeT(OpCode op, const void * rawData, int byteSize, ByteCodeWriter* writer);
--- a/lib/Runtime/Language/DynamicProfileInfo.cpp
+++ b/lib/Runtime/Language/DynamicProfileInfo.cpp
@@ -889,10 +889,11 @@ namespace Js
 
     // We are overloading the value types to store whether it is a mod by power of 2.
     // TaggedInt:
-    inline void DynamicProfileInfo::RecordModulusOpType(FunctionBody* body, ProfileId profileId, bool isModByPowerOf2)
+    void DynamicProfileInfo::RecordModulusOpType(FunctionBody* body,
+                                 ProfileId profileId, bool isModByPowerOf2)
     {
         Assert(profileId < body->GetProfiledDivOrRemCount());
-        // allow one op of the modulus to be optimized - anyway
+        /* allow one op of the modulus to be optimized - anyway */
         if (divideTypeInfo[profileId].IsUninitialized())
         {
             divideTypeInfo[profileId] = ValueType::GetInt(true);
@@ -901,11 +902,13 @@ namespace Js
         {
             if (isModByPowerOf2)
             {
-                divideTypeInfo[profileId] = divideTypeInfo[profileId].Merge(ValueType::GetInt(true));
+                divideTypeInfo[profileId] = divideTypeInfo[profileId]
+                                           .Merge(ValueType::GetInt(true));
             }
             else
             {
-                divideTypeInfo[profileId] = divideTypeInfo[profileId].Merge(ValueType::Float);
+                divideTypeInfo[profileId] = divideTypeInfo[profileId]
+                                           .Merge(ValueType::Float);
             }
         }
     }
--- a/lib/Runtime/Library/amd64/JavascriptFunctionA.S
+++ b/lib/Runtime/Library/amd64/JavascriptFunctionA.S
@@ -113,31 +113,6 @@ NESTED_END amd64_CallFunction, _TEXT
 //extrn ?GetStackSizeForAsmJsUnboxing@Js@@YAHPEAVScriptFunction@1@@Z: PROC
 //extrn ?GetArgsSizesArray@Js@@YAPEAIPEAVScriptFunction@1@@Z : PROC
 
-// int64 CallAsmJsFunction<int64>(RecyclableObject *function, JavascriptMethod entryPoint, uint argc, Var *argv);
-.balign 16
-LEAF_ENTRY _ZN2Js18JavascriptFunction17CallAsmJsFunctionIlEET_PNS_16RecyclableObjectEPFPvS4_NS_8CallInfoEzEjPS5_, _TEXT
-    jmp _ZN2Js18JavascriptFunction17CallAsmJsFunctionIiEET_PNS_16RecyclableObjectEPFPvS4_NS_8CallInfoEzEjPS5_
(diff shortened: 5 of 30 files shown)


obastemur:
Couple of CI / xcode ver. issues. (i.e. std::nullptr_t is not available...... )

obastemur:

obastemur:

obastemur:
@jianchun @curtisman review please.

lib/Runtime/Language/DynamicProfileInfo.cpp
```diff
@@ -887,28 +887,7 @@ namespace Js
...
-            {
-                divideTypeInfo[profileId] = divideTypeInfo[profileId].Merge(ValueType::GetInt(true));
-            }
-            else
-            {
-                divideTypeInfo[profileId] = divideTypeInfo[profileId].Merge(ValueType::Float);
-            }
-        }
-    }
+    DEFINE_DYNAMICPROFILEINFO_RECORDMODULUSOPTYPE_FNC
```
jianchun: This approach looks weird. How about removing the "inline" keyword from old implementation instead?

lib/Common/Memory/amd64/XDataAllocator.cpp
```diff
@@ -12,6 +12,10 @@ CompileAssert(false)
 #include "XDataAllocator.h"
 #include "Core/DelayLoadLibrary.h"
 
+#ifndef _WIN32
```
jianchun: nit: `#ifndef ...` not needed.

pal/src/exception/machexception.cpp
```diff
@@ -15,16 +13,16 @@ Module Name:
 
     Implementation of MACH exception API functions.
 
-
-
 --*/
 
+#include "pal/dbgmsg.h"
+SET_DEFAULT_DEBUG_CHANNEL(EXCEPT); // some headers have code with asserts, so do this first
+
```
jianchun: I know nothing about these big changes in PAL. Did you work on these changes, or did you sync them to dotnet coreCLR latest?

lib/Runtime/PlatformAgnostic/Platform/Unix/AssemblyCommon.cpp
```diff
@@ -0,0 +1,32 @@
...
+// Licensed under the MIT license. See LICENSE.txt file in the project root for full license information.
+//-------------------------------------------------------------------------------------------------------
+
+#include "Common.h"
+#include "PlatformAgnostic/AssemblyCommon.h"
+
+extern void mac_fde_wrapper(const char *dataStart, mac_fde_reg_op reg_op)
+{
+    for(const char *head = dataStart, *last = dataStart + CC_FDE_SIZE;
+        head != last;
```
jianchun: I suggest avoid the `last` check. FDE is not of fixed size. The check may mislead some to think it is. The CIE/FDE data is guaranteed to terminate from the `break` in the loop body below. If for any reason the data is wrong and does not, it makes no difference whether it hangs here or crashes later on. Hanging here is better in my opinion because it is closer to the source of problem.

lib/Runtime/PlatformAgnostic/Platform/Unix/AssemblyCommon.cpp
```diff
@@ -0,0 +1,32 @@
...
+        /* noop */)
+    {
+        const char *op = head;
+        // Get Length of record [ Read 4 bytes ]
+        const uint32_t* op_data = (const uint32_t*)op;
```
jianchun: >const uint32_t* op_data [](start = 8, length = 23)

Why not `const uint32_t length = ...`?

lib/Runtime/PlatformAgnostic/Platform/Unix/AssemblyCommon.cpp
```diff
@@ -0,0 +1,32 @@
...
+        if(*op_data == 0)
+        {
+            break;
+        }
+        // if it's not 0xffffffff,
+        // there we have the length of the CIE or FDE record.
```
jianchun: nit: Comment is not very clear (as code not checking 0xFFFFFFFF). We actually ignored 0xFFFFFFFF which identifies extended length. Our records are small so we never need extended length.

lib/Runtime/PlatformAgnostic/Platform/Unix/AssemblyCommon.cpp
```diff
@@ -0,0 +1,32 @@
...
+        head += 4; // get next
```
jianchun: >head += 4; // get next [](start = 8, length = 22)

nit: Please comment it is to read next 4 bytes and non zero value identifies FDE (zero is CIE).

lib/Common/PlatformAgnostic/AssemblyCommon.h
```diff
@@ -0,0 +1,30 @@
...
+// Licensed under the MIT license. See LICENSE.txt file in the project root for full license information.
+//-------------------------------------------------------------------------------------------------------
+#pragma once
+
+#ifndef _WIN32
+
+extern "C" void __register_frame(const void* ehframe);
+extern "C" void __deregister_frame(const void* ehframe);
+
+#if defined(_AMD64_) && !defined(DISABLE_JIT)
```
jianchun: >!defined(DISABLE_JIT) [](start = 24, length = 21)

Please use `ENABLE_NATIVE_CODEGEN` instead of `!defined(DISABLE_JIT)`. The former is used more.

lib/Common/PlatformAgnostic/AssemblyCommon.h
```diff
@@ -0,0 +1,30 @@
...
+#if defined(__APPLE__) // no multi fde support
+#ifndef __IOS__
+#define CC_FDE_SIZE 64
+typedef void(*mac_fde_reg_op)(const void*addr);
+void mac_fde_wrapper(const char *dataStart, mac_fde_reg_op reg_op);
+#define __REGISTER_FRAME(addr) mac_fde_wrapper((char*)addr, __register_frame)
+#define __DEREGISTER_FRAME(addr) mac_fde_wrapper((char*)addr, __deregister_frame)
```
jianchun: nit: use `(const char*)addr` to be more consistent to signature.

jianchun: nit: Use `(const char*)addr` to be consistent with signature.

lib/Runtime/ByteCode/ByteCodeWriter.h
```diff
@@ -96,6 +96,28 @@ namespace Js
...
+                }                                                                  \
+                else                                                               \
+                {                                                                  \
+                    byte byteop = (byte)Js::OpCode::ExtendedOpcodePrefix;          \
+                    Write(&byteop, sizeof(byte));                                  \
+                    Write(&op, sizeof(uint16));                                    \
+                }                                                                  \
+                Assert(OpCodeUtil::EncodedSize((Js::OpCode)op, SmallLayout)        \
+                       == (currentOffset - offset));                               \
+            }
```
obastemur: Indeed weird. Tried other things [i.e. remove inline / use it from header ].. Older Clang behaves strange :/ Actually, we already define those in multiple places (using `include` )

pal/src/exception/machexception.cpp
obastemur: Some I had to update + sync to PAL.

lib/Runtime/PlatformAgnostic/Platform/Unix/AssemblyCommon.cpp
obastemur: There is no guarantee here for a meaningful hang or crash. Indeed FDE is not a fixed size and shouldn't be bigger than the defined size above. + if it is bigger, see assert below

lib/Runtime/PlatformAgnostic/Platform/Unix/AssemblyCommon.cpp
obastemur: could be... op -> cast to -> op_data?

lib/Common/PlatformAgnostic/AssemblyCommon.h
obastemur: Is there a reason we define `DISABLE_JIT` in that case?

obastemur:
@dotnet-bot test Ubuntu

obastemur:

obastemur:
@dotnet-bot test Ubuntu ubuntu_linux_static

obastemur:

obastemur:

lib/Common/Memory/amd64/XDataAllocator.cpp
jianchun: I seem to recall you had this same include elsewhere without `#ifndef _WIN32`.

lib/Runtime/PlatformAgnostic/Platform/Unix/AssemblyCommon.cpp
jianchun: My biggest complain is about the misleading `for` loop. It reads as if it would loop some fixed size and end at `last`. Actually no, it ends at a terminator and break from inside the loop body.

lib/Runtime/PlatformAgnostic/Platform/Unix/AssemblyCommon.cpp
jianchun: I meant the pointer `op_data` was not as straightforward. Just do `const uint32_t length = *(const uint32_t*)op;` and use `length`.

lib/Common/PlatformAgnostic/AssemblyCommon.h
jianchun: `ENABLE_NATIVE_CODEGEN` existed from the beginning of this project and used everywhere. `DISABLE_JIT` used rarely (haven't checked so not sure about history, might be added recently, maybe we should not have added it). I prefer the former for overall consistency.

obastemur: I recall `DISABLE_JIT` is used while enabling JIT on linux. Since we have it and naming explains the intention, I prefer keeping it. If we want to remove them all, that's something else.

lib/Runtime/PlatformAgnostic/Platform/Unix/AssemblyCommon.cpp
jianchun: I prefer self-explanatory code instead of comment. `head != last` is meaningless. It is almost always true.

lib/Common/PlatformAgnostic/AssemblyCommon.h
jianchun: I'm ok if you insist. Just want to note that another reason I prefer `ENABLE_NATIVE_CODEGEN` is it is positive. Easier to read and understand than double negative `!defined(DISABLE_JIT)`.

digitalinfinity: Just noticed this part. ENABLE_NATIVE_CODEGEN is a subset of DISABLE_JIT. ENABLE_NATIVE_CODEGEN is generally meant for JIT-related code directly. DISABLE_JIT was added as part of doing the Interpreter-only version of ChakraCore last year- it implies !ENABLE_NATIVE_CODEGEN but also disables other things like dynamic interpreter thunks, interpreter profiling etc.

jianchun: Thanks @digitalinfinity for explanation, now I understand. `!defined(DISABLE_JIT)` here is fine to me.

lib/Runtime/Language/DynamicProfileInfo.cpp
```diff
@@ -901,11 +902,13 @@ namespace Js
...
-                divideTypeInfo[profileId] = divideTypeInfo[profileId].Merge(ValueType::GetInt(true));
+                divideTypeInfo[profileId] = divideTypeInfo[profileId]
+                                           .Merge(ValueType::GetInt(true));
             }
             else
             {
-                divideTypeInfo[profileId] = divideTypeInfo[profileId].Merge(ValueType::Float);
+                divideTypeInfo[profileId] = divideTypeInfo[profileId]
+                                           .Merge(ValueType::Float);
             }
```
jianchun: nit: DynamicProfileInfo.cpp contains some formatting changes that are unnecessary (changes when you originally put them into a macro?).

lib/Runtime/Library/amd64/JavascriptFunctionA.S
```diff
@@ -113,31 +113,6 @@ NESTED_END amd64_CallFunction, _TEXT
...
-    jmp _ZN2Js18JavascriptFunction17CallAsmJsFunctionIiEET_PNS_16RecyclableObjectEPFPvS4_NS_8CallInfoEzEjPS5_
-LEAF_END _ZN2Js18JavascriptFunction17CallAsmJsFunctionIdEET_PNS_16RecyclableObjectEPFPvS4_NS_8CallInfoEzEjPS5_, _TEXT
-
-// __m128 JavascriptFunction::CallAsmJsFunction(RecyclableObject * function, void* entryPoint, uint argc, Var * argv);
-//.balign 16
-//??$CallAsmJsFunction@T__m128@@@JavascriptFunction@Js@@SA?AT__m128@@PEAVRecyclableObject@1@PEAXIPEAPEAX@Z PROC FRAME
-//    jmp _ZN2Js18JavascriptFunction17CallAsmJsFunctionIiEET_PNS_16RecyclableObjectEPFPvS4_NS_8CallInfoEzEjPS5_
-//??$CallAsmJsFunction@T__m128@@@JavascriptFunction@Js@@SA?AT__m128@@PEAVRecyclableObject@1@PEAXIPEAPEAX@Z ENDP
-
-
```
jianchun: nit: Why moving this section... here they are closer to next <int> implementation.
Also you deleted __m128 version. Commented out currently, maybe needed in future.

lib/Runtime/Language/DynamicProfileInfo.cpp
obastemur: 80 chars. since it is updated, left it.

obastemur:

PR #1749, opened 2016-10-15
Wasm: Update testsuite + indirect call with equivalent signature

Cellule:
I realized that I didn't update every spec test in my last pass #1710
There were some failures in the indirect calls.
It is valid to do an indirect call with an equivalent signature type.

Take the following example
```
  (func $empty-sig-1) 
  (type $empty-sig-duplicate (func))
  (table anyfunc (elem $empty-sig-1))
  (func (export "signature-explicit-reused")
    (call_indirect $empty-sig-duplicate (i32.const 0))
  )
```

The function `$empty-sig-1` has signature id 0
Therefore, in the table, the element at index 0 has the signature id 0.
However, in the indirect call, we specify the signature `$empty-sig-duplicate` which has the signature id 1.
Both signatures are equivalent, thus this indirect call is valid.

I disabled 4 tests in func.wast because of a bug in jitted code. I believe I already fixed that bug in my Int64 branch so I'm leaving them disabled for the time being.

This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/1749)


Cellule:
@MikeHolman @Krovatkin @arunetm @commonlisp Can you guys review this please ?

lib/Runtime/Library/WasmLibrary.cpp
```diff
@@ -307,6 +307,42 @@ namespace Js
 
     void WasmLibrary::WasmLoadIndirectFunctionTables(Wasm::WasmModule * wasmModule, ScriptContext* ctx, Var** indirectFunctionTables, Var* localModuleFunctions, Var* importFunctions)
     {
+        ArenaAllocator alloc(_u("WasmTempIndirectTable"), ctx->GetThreadContext()->GetPageAllocator(), Js::Throw::OutOfMemory);
```
MikeHolman: I wonder if we should coalesce equivalent signatures earlier on (i.e. before we even emit the indirect calls)?

Cellule: Yes, I did think about something like that too. I am wondering if it would be better/cleaner.
I can give it a try

Cellule: I decided to implement it and I think in the end it's better because we don't have to fill additional spaces in the table.
When dealing with indirect table (call or filling it) we have to map the signature id first, that way we always deal with only a set of unique signatures

arunetm:
LGTM

PR #1755, opened 2016-10-17
fix issue with for-in cache using wrong proc address under oop jit

MikeHolman:


MikeHolman:
Ok, I was a bit hasty in attributing the bug. This is not the issue.

PR #1757, opened 2016-10-17
Make EntryPointInfo::callsCount usable by redeferral heuristic. 

pleath:
Change the uint8's toi uint32's, which in turn allows us to dispense with the full-JIT code that saturates these values at 255.


lib/Runtime/Base/FunctionBody.cpp
```diff
@@ -6591,11 +6591,11 @@ namespace Js
         }
 
         // Simple JIT counts down and transitions on overflow
-        const uint8 callCount = simpleJitEntryPointInfo->callsCount;
+        const uint32 callCount = simpleJitEntryPointInfo->callsCount;
         Assert(simpleJitLimit == 0 ? callCount == 0 : simpleJitLimit > callCount);
         return callCount == 0 ?
             static_cast<uint16>(simpleJitLimit) :
-            static_cast<uint16>(simpleJitLimit) - callCount - 1;
+            static_cast<uint16>(simpleJitLimit) - static_cast<uint8>(callCount) - 1;
```
pleath: The call count can't be > 255 while we're decrementing toward the simple JIT limit. We could probably do without the static cast, but I didn't want any static tools complaining about overflow.

rajatd:
IIRC, the reason for saturating the call count at 255 was that if it were allowed to grow to uin32_max, and we started bailing out, it will take a long time for the call count to get to 0 and, in turn, for us to rejit the function. Is that not a concern anymore?

pleath:
I think @LouisLaf had an idea for mitigating this. Should be easy to add it later.

lib/Backend/Lower.cpp
```diff
@@ -14506,22 +14506,65 @@ IR::Instr *Lowerer::InsertConvertFloat64ToFloat32(
...
+    IR::LabelInstr *const overflowLabel = Lowerer::InsertLabel(false, insertBeforeInstr);
+
+    //     subs temp, src, 1
+    IR::RegOpnd *const tempOpnd = IR::RegOpnd::New(StackSym::New(TyUint32, func), TyUint8, func);
+    const IR::AutoReuseOpnd autoReuseTempOpnd(tempOpnd, func);
+    Lowerer::InsertSub(true, tempOpnd, src, IR::IntConstOpnd::New(1, TyUint32, func, true), overflowLabel);
+
+    //     bcs $overflow
+    Lowerer::InsertBranch(Js::OpCode::BrLt_A, true, overflowLabel, overflowLabel);
+
```
LouisLaf: With int32, I think we don't even care about the overflow.  Plus, with saturation, it doesn't fix the retiring problem, it just delays it.

lib/Backend/BailOut.cpp
```diff
@@ -1763,7 +1763,7 @@ void BailOutRecord::ScheduleFunctionCodeGen(Js::ScriptFunction * function, Js::S
     bailOutRecordNotConst->bailOutCount++;
 
     Js::FunctionEntryPointInfo *entryPointInfo = function->GetFunctionEntryPointInfo();
-    uint8 callsCount = entryPointInfo->callsCount;
+    uint32 callsCount = entryPointInfo->callsCount;
```
LouisLaf: You can just change this to:
uint32 callsCount = MIN(entryPointInfo->callsCount, 255);
to retain the previous bailout behavior we had with saturation.

PR #1760, opened 2016-10-18
fix issue when polymorphic inline info layouts differ in JIT and runtime data

MikeHolman:
Exact order of next pointers for polymorphic CodeGen JIT and CodeGen Runtime data may differ. So, while walking CodeGen JIT data during BuildJITTimeData data we should pass along the head CodeGen Runtime pointer, otherwise inner calls will not be able to find inlinees who appear earlier in the list.


MikeHolman:

MikeHolman:

rajatd:
:shipit:

PR #1761, opened 2016-10-18
Enabling Stack Args when there is presence of nested functions.

satheeshravi:
**Premise:**
The initial implementation of Stack Args was disabled when there is any nested function.
We were missing a bunch of opportunities in the real world web pages - where nested function is a common thing.
This PR enables this optimization for nested function.

**Summary:**
Stack Args will work partially when nested functions are present. We will attempt to remove only Heap Arguments Object creation instr and retain scope object creation instr (whenever required).

**Details:**
**ByteCodeGeneration:**
- We track if there are any params being closure captured by any nested functions -we give up the optimization if this is the case.
- We track if there are any non-local references inside the nested function - we do Stack Args partially here - i.e. we attempt to remove Heap Arguments Object creation but still create scope object - as scope object is required for FrameDisplay.

**IRBuilder:**
- Scope object tracking has been moved from Forward pass (Globopt) to this phase.
- This enables us to mark the scope obj sym to be live during the first backward pass.

**Backward pass:**
- Scope Object Sym is kept alive in all code paths.
- This is to facilitate Bailout to record the live Scope object Sym, whenever required.
- Reason for doing is this because - Scope object has to be implicitly live whenever Heap Arguments object is live.
- When we restore HeapArguments object in the bail out path, it expects the scope object also to be restored - if one was created.
- We do not know detailed information about Heap arguments obj syms(aliasing etc.) until we complete Forward Pass.
- And we want to avoid dead sym clean up (in this case, scope object though not explicitly live, it is live implicitly) during Block merging in the forward pass.

**Globopt pass:**
- Extra tracking for instrs having arguments object is done.

**Dead Store pass:**
- Scope obj creation instruction is not dead stored - It will be a MOV NULL instr in the Lowerer.
- All instrs related cached scope obj is deadstored. Reason is to dead store the cached scope object creation instruction itself.
- InitCachedFuncs is deadstored.
- GetCachedFunc is replaced with NewScFunc instr.

**Lowerer pass:**
- If Stack Args optimization is still turned on and the parser has hinted that Scope obj creation is not required, then scope object creation instruction will be turned to a MOV NULL.
       - We retain the instruction and not remove it because we want the value of the scope object to be restored
- If Stack Args opt is turned on and the parser has hinted that Scope obj creation is required, then we do a helper call to call scope object and do necessary type transitions for the formals.
- That's when we can always rely on the bail out path to restore some valid value (NULL or the scope object itself).

**BailOut path:**
- We use the restored scope object value (we track this by using a flag on the bailoutInfo) and copy the param values to it.
- If no scope object was restored, then we re-create the scope object with param values.

**Perf Results:**
- No visible improvements in Speedometer/ acme-air upon intial runs. But all the nested function cases are being successfully handled now.
- I am waiting on IEPerfLab.
- Pending: Other benchmark runs. Will update when it is done (I don't expect any wins here).
- IE PerfLab Results: http://ieperf-web-01/perfresults/Comparison?testRunId=443743&baseRunId=443732&lab=IEPERF&getRelatedRunData=false&priorities=0,1&categories=failures,warnings,improvements,no-change&search=
  - The above link shows some overall improvement in few web sites.

**Functional Test:**
- Looks fine.


lib/JITIDL/JITTypes.h
```diff
@@ -479,6 +479,7 @@ typedef struct FunctionBodyDataIDL
     boolean isParamAndBodyScopeMerged;
     boolean hasFinally;
     boolean usesArgumentsObject;
+    boolean doScopeObjectCreation;
```
MikeHolman: I know this is annoying, but midl hates implicit padding (for some reason) and gives a really heavy perf penalty for it. we need another 1 byte of padding here, and then another 2 bytes, and then another 4 bytes on x64 to 8 byte align the pointers.

satheeshravi:
@rajatd @aneeshdk @pleath @LouisLaf @MikeHolman : Can you please review?

lib/Runtime/ByteCode/ByteCodeGenerator.cpp
```diff
@@ -2918,6 +2918,55 @@ FuncInfo* PostVisitFunction(ParseNode* pnode, ByteCodeGenerator* byteCodeGenerat
...
+
+                if (currentFormalNode->sxParamPattern.pnodeNext == nullptr)
+                {
+                    currentFormalNode = nullptr;
+                }
+                else
+                {
+                    currentFormalNode = currentFormalNode->sxParamPattern.pnodeNext;
+                }
+            }
```
aneeshdk: Can't we just do paramScope->GetHasOwnLocalInClosure() here instead of going through the list? In merge case also we are not resetting this flag.

lib/Runtime/ByteCode/ByteCodeGenerator.cpp
```diff
@@ -2918,6 +2918,55 @@ FuncInfo* PostVisitFunction(ParseNode* pnode, ByteCodeGenerator* byteCodeGenerat
...
+            {
+                if (currentFormalNode->sxPid.sym->GetHasNonLocalReference())
+                {
+                    hasAnyParamInClosure = true;
+                    break;
+                }
+
+                if (currentFormalNode->sxParamPattern.pnodeNext == nullptr)
+                {
+                    currentFormalNode = nullptr;
```
kunalspathak: >currentFormalNode = nullptr; [](start = 20, length = 28)

nit: can just `break;` here.

lib/Backend/BackwardPass.cpp
```diff
@@ -2996,6 +3011,60 @@ BackwardPass::DeadStoreOrChangeInstrForScopeObjRemoval()
...
+                                instrDef = instrDef->GetSrc2()->GetStackSym()->GetInstrDef();
+                                Assert(instrDef->m_opcode == Js::OpCode::ArgOut_A);
+                            }
+                            else
+                            {
+                                instrDef = nullptr;
+                            }
+                            instrToDelete->Remove();
+                        }
+                        if (nextInstr != nullptr)
```
rajatd: can nextInstr ever be nullptr?

lib/Backend/Lower.cpp
```diff
@@ -249,12 +249,80 @@ Lowerer::LowerRange(IR::Instr *instrStart, IR::Instr *instrEnd, bool defaultDoFa
...
 
         case Js::OpCode::InitCachedScope:
-            instrPrev = this->LowerInitCachedScope(instr);
+            if (instr->m_func->GetJITFunctionBody()->GetDoScopeObjectCreation() || !instr->m_func->IsStackArgsEnabled())
+            {
+                instrPrev = this->LowerInitCachedScope(instr);
+            }
+            else
+            {
+                instr->UnlinkSrc1();
```
rajatd: this is just leaking m_src1. Use Instr::ReplaceSrc1 for setting the src to null

lib/Runtime/Language/JavascriptOperators.cpp
```diff
@@ -6804,7 +6804,7 @@ namespace Js
                 // and handlers created here to be marked as shared up-front. This is to ensure we don't get any fixed fields and that the handler
                 // is ready for storing values directly to slots.
                 DynamicType* newType = PathTypeHandlerBase::CreateNewScopeObject(scriptContext, frameObject->GetDynamicType(), propIds, nonSimpleParamList ? PropertyLetDefaults : PropertyNone);
-
+
```
rajatd: nit: extra line

rajatd:
Backend changes looked code, barring the comments

aneeshdk:
LGTM

lib/Backend/BailOut.h
```diff
@@ -471,6 +471,7 @@ struct GlobalBailOutRecordDataTable
     bool isLoopBody;
     bool hasNonSimpleParams;
     bool hasStackArgOpt;
+    bool isScopeObjRestored;
```
LouisLaf: Might be worth it to change these bools to a bitfield at this point.

LouisLaf:
:shipit:

PR #1762, opened 2016-10-18
Use the right target for Object's internal array get item

curtisman:
When a built in is trying fill in index property from the prototype, we need to pass in the original object to get the value of the element if the prototype is an object with and internal es5 array, so that accessors will get the right target instead of the internal es5 array.

### DIFF
--- a/test/es5/es5array_objproto_builtin.js
+++ b/test/es5/es5array_objproto_builtin.js
@@ -1,0 +1,17 @@
+//-------------------------------------------------------------------------------------------------------
+// Copyright (C) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE.txt file in the project root for full license information.
+//-------------------------------------------------------------------------------------------------------
+
+var ary = Array(1);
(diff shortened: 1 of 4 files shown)


curtisman:
@jianchun please review.

test/es5/es5array_objproto_builtin.js
```diff
@@ -0,0 +1,17 @@
+//-------------------------------------------------------------------------------------------------------
+// Copyright (C) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE.txt file in the project root for full license information.
+//-------------------------------------------------------------------------------------------------------
+
+var ary = Array(1);
+ary.prop = "Got array propert. Failed";
```
akroshg: nit: property.

akroshg:
:shipit:

test/es5/es5array_objproto_builtin.js
curtisman: Fixed.

PR #1767, opened 2016-10-18
initialize oop interpreter thunk emitter

MikeHolman:
After rebasing I messed up the initialization code.

### DIFF
--- a/lib/JITServer/JITServer.cpp
+++ b/lib/JITServer/JITServer.cpp
@@ -444,7 +444,6 @@ ServerNewInterpreterThunkBlock(
         thunkInfo->pdataTableStart = (intptr_t)pdataStart;
         thunkInfo->epilogEndAddr = epilogEnd;
 #endif
-        arena->Clear();
 
         return S_OK;
     });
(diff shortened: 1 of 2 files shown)


lib/JITServer/JITServer.cpp
```diff
@@ -397,7 +397,7 @@ ServerNewInterpreterThunkBlock(
 
         // REVIEW: OOP JIT should we clear arena at end?
         ArenaAllocator * arena = scriptContext->GetSourceCodeArena();
-        BYTE * localBuffer = AnewArray(arena, BYTE, bufferSize);
+        BYTE * localBuffer = HeapNewArray(BYTE, bufferSize);
```
leirocks: >localBuffer [](start = 15, length = 11)

how big is the buffer? can use stack? otherwise you need to fix the memory leak for following OOM throw

leirocks:
:shipit:

lib/JITServer/JITServer.cpp
MikeHolman: it's just a page. i'll leave in arena memory though... wasn't really reason to change that

PR #1769, opened 2016-10-18
Wasm: Deferred parsing On by default

Cellule:
Make Wasm Deferred parsing on by default

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/1769)

### DIFF
--- a/lib/Runtime/Library/WasmLibrary.cpp
+++ b/lib/Runtime/Library/WasmLibrary.cpp
@@ -214,7 +214,7 @@ namespace Js
             funcObj->SetEnvironment(frameDisplay);
             localModuleFunctions[i] = funcObj;
 
-            if (PHASE_ON(WasmDeferredPhase, funcObj->GetFunctionBody()))
+            if (!PHASE_OFF(WasmDeferredPhase, funcObj->GetFunctionBody()))
             {
                 funcObj->GetDynamicType()->SetEntryPoint(WasmLibrary::WasmDeferredParseExternalThunk);
                 entypointInfo->jsMethod = WasmLibrary::WasmDeferredParseInternalThunk;


Cellule:
@MikeHolman @arunetm @Krovatkin Can you guys review this please ?

Krovatkin:
out of curiosity 
* why are we switching to deferred parsing? 
   * i could only think of savings in loading times :-( 
* will the switch make lazy traps "less" useful?

MikeHolman:
@Krovatkin Yes, the point is to improve load times.

I'm wondering if we need lazy traps since we have deferred parse? Seems lazy trap is subset of defer parse.

lib/Runtime/Library/WasmLibrary.cpp
```diff
@@ -214,7 +214,7 @@ namespace Js
             funcObj->SetEnvironment(frameDisplay);
             localModuleFunctions[i] = funcObj;
 
-            if (PHASE_ON(WasmDeferredPhase, funcObj->GetFunctionBody()))
+            if (!PHASE_OFF(WasmDeferredPhase, funcObj->GetFunctionBody()))
```
Krovatkin: not even a nit. i understand some people typically try to avoid double negations. In this particular case **false** and **true** blocks pretty simple anyways, so it might make sense to get rid of double negation and switch the **false** arm with **true**

```cpp
            if (PHASE_OFF(WasmDeferredPhase, funcObj->GetFunctionBody()))
            {
   WasmFunctionGenerateBytecode(funcObj, !PHASE_ON(WasmLazyTrapPhase, funcObj->GetFunctionBody()));

            }
            else
            {
                funcObj->GetDynamicType()->SetEntryPoint(WasmLibrary::WasmDeferredParseExternalThunk);
                entypointInfo->jsMethod = WasmLibrary::WasmDeferredParseInternalThunk;             
            }
```

arunetm:
LGTM

PR #1770, opened 2016-10-18
fix issue with caching global this object incorrectly

MikeHolman:
I misunderstood meaning of FunctionCodeGenJitTimeData::isInlined and was trying to use that to save global this address for topfunc only, but that field has a different meaning. Instead, just always store it when creating FunctionCodeGenJitTimeData for full FunctionBody.


MikeHolman:

PR #1771, opened 2016-10-18
Put OOP CFG registration under a flag in test builds

MikeHolman:
To help with downlevel testing, only enable OOP CFG registration on release build


LouisLaf:
:shipit:

PR #1773, opened 2016-10-18
Add delay free listing for func arenas in debug builds

meg-gupta:
Freed memory on func arenas are returned to a delay free list instead of
making them available immediately for allocation. For Opnd frees,
isDeleted bit is set, this enables detecting erroneous code that can use-afer-free of Opnd
memory. At the end of a phase, the delay free list is merged into the free
list making it available for subsequent allocations.

### DIFF
--- a/lib/Backend/Opnd.inl
+++ b/lib/Backend/Opnd.inl
@@ -17,6 +17,7 @@ namespace IR {
 inline Opnd *
 Opnd::Use(Func *func)
 {
+    AssertMsg(!isDeleted, "Using deleted operand");
     if (!m_inUse)
     {
         m_inUse = true;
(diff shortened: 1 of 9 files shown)


meg-gupta:
@LouisLaf Can you please review ?

lib/Backend/Opnd.inl
```diff
@@ -19,6 +19,7 @@ Opnd::Use(Func *func)
 {
     if (!m_inUse)
     {
+        AssertMsg(!isDeleted, "Using deleted operand");
```
LouisLaf: >      AssertMsg(!isDeleted, "Using deleted operand"); [](start = 2, length = 53)

This should go outside the if().  Free might have memset m_inUse to true.

LouisLaf:
:shipit:

PR #1774, opened 2016-10-18
use stack memory for some of the allocation in interpreter thunk emitter

leirocks:
looks these allocation is small and can just use stack memory, otherwise we need use a seperater arena allocator or heap allocator to prevent leak.


lib/JITServer/JITServer.cpp
```diff
@@ -403,19 +403,12 @@ ServerNewInterpreterThunkBlock(
...
-        BYTE * localBuffer = AnewArray(arena, BYTE, bufferSize);
-
         BYTE* remoteBuffer;
         EmitBufferAllocation * allocation = emitBufferManager->AllocateBuffer(bufferSize, &remoteBuffer);
-        if (!allocation)
-        {
-            Js::Throw::OutOfMemory();
-        }
 
+        Assert(bufferSize <= 0x1000); // in case this is changed some day we might switch to use other allocator
```
MikeHolman: CompileAssert?

PR #1778, opened 2016-10-19
SAL/TVS Fixes

tcare:
lib/Common/Codex/Utf8Helper.h:
- UTF8Codex has outdated SAL annotations, but without fixing them we can let SAL know that the pointers are just passed through for the Narrow <-> Wide string conversions.

lib/Common/Common/NumberUtilities_strtod.cpp:
- SAL false positive mitigations left us with some more breaks to do with loop invariants.
- Removing our forced analysis asserts and assumes fixes the analysis!
- Re-enabled a disabled SAL warning.
- Updated SAL to SAL2.

### DIFF
--- a/lib/Common/Codex/Utf8Helper.h
+++ b/lib/Common/Codex/Utf8Helper.h
@@ -14,9 +14,7 @@ namespace utf8
     /// The returned string is null terminated.
     ///
     template <class Allocator>
-    HRESULT WideStringToNarrow(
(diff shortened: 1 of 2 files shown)


tcare:
@curtisman

tcare:
Small update to disable a warning that only occurs in Jenkins prefast.

tcare:
@dotnet-bot test this please

PR #1782, opened 2016-10-19
fix server script context initialize

leirocks:


leirocks:
looks @MikeHolman has fixed this for me. Thanks!

Issue #1671, opened 2016-09-30
Perf: Too many bailouts/rejits for FailedTagCheck


kunalspathak:
For below test case, we keep bailing out and rejitting for same reason.

``` js
function test(object) {
    var value = object.x;
    if (value && value.test) { // bailout happens here 
        value.test();
    }
    return value;
}

for (var i = 0; i < 40000; i++) {
    var obj = {};
    if (i == 0) {
        obj.x = {};
    } else {
        obj.x = 1;
    }
    test(obj);
}
print('done');
```

``` batch
$ ch e:\temp\test.js -mic:1 -off:Simplejit -stats:rejit
done
Bailout Reason,                           Count
BailOutOnTaggedValue,                     39999
TOTAL,                                    39999


Rejit Reason,                             Count
FailedTagCheck,                             100
TOTAL,                                      100

```

kunalspathak:
//cc - @rajatd , @satheeshravi , @LouisLaf , @pleath

rajatd:
The problem here is that when we bailout to a LdFld, we only update the value type of the result of LdFld. In the case of BailOutOnTaggedValue, we are bailing out because the object from which we tried to load a property, is a tagged value. So, we bail out, execute LdFld in the interpreter but fail to update the bad profile data that caused the bailout in the first place.

PR #1784, opened 2016-10-19
Disable tag checks optimization for BailOutOnTaggedValue

ThomsonTan:
For BailOutOnTaggedValue, we could not have a chance to update profile when we get tagged value because profile will only be produced when defining the value, not using them. So even rejit after some bailouts will not help because of missing profile on the value type.  This change disable tag checks for rejit when BailOutOnTaggedValue is detected because disable it doesn't hurt perf when the value type is not fixed.


ThomsonTan:
@LouisLaf @kunalspathak please have a look.

pleath:
LGTM. Thanks.

kunalspathak:
:shipit:

kunalspathak:
Fix #1671

PR #1785, opened 2016-10-19
Fix incorrect parsing of yield/await in arrow functions

### LINKED ISSUE #333: yield should be reserved word in arrow functions within StrictFormalParameters

aneeshdk:
The spec carries the Yield/Await parameter forward into the parameter
list grammar production for arrow functions specifically. We did not
adhere to this and erroneously treated these as identifiers in arrow
function parameter lists.

Fixes #333.  See it for more details.

### DIFF
--- a/lib/Parser/Parse.cpp
+++ b/lib/Parser/Parse.cpp
@@ -2850,22 +2850,26 @@ ParseNodePtr Parser::ParseTerm(BOOL fAllowCall,
         ichMin = m_pscan->IchMinTok();
         iecpMin  = m_pscan->IecpMinTok();
 
+        if (pid == wellKnownPropertyPids.async &&
+            m_scriptContext->GetConfig()->IsES7AsyncAndAwaitEnabled())
+        {
+            isAsyncExpr = true;
+        }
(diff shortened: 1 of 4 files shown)


aneeshdk:
This is an update to the Pull Request https://github.com/Microsoft/ChakraCore/pull/1497/files#r76477843 from Ian

aneeshdk:
@dotnet-bot test please

pleath:
Need to look at the enclosing non-arrow function, not necessarily the immediate parent, since the current function may be enclosed in any number of arrow functions.

PR #1786, opened 2016-10-19
Revise redeferral heuristic

pleath:
Revise redeferral heuristic to track execution of full-jitted functions without generating new code. We make use of the per-entry-point calls count for jitted entry points as well as the interpreted count for the FunctionBody. Incorporates the jitted code changes to stop saturating the callsCount field.


pleath:
Out of date PR incorporated into https://github.com/Microsoft/ChakraCore/pull/1585.

PR #1789, opened 2016-10-20
Fix SIMDCheckUint32Number when input value is not in range

meg-gupta:
JavascriptConversion::ToNumber on a string can result in -NaN, static_cast
on this is not valid. A RangeError should be thrown. Handle this scenario
in SIMDCheckUint32Number.


meg-gupta:
@arunetm Can you please review ?

arunetm:
lgtm

PR #1795, opened 2016-10-21
pal : Cherry-pick fix for not blocking signals in ubuntu

kunalspathak:
Behavior : I was investigating a hang in node-cc on Ubuntu. The repro was that parent process spawns child process and waits (_epoll in libuv) to receive `SIGCHLD` signal from child process when it terminates. Parent process never receives this signal. Looking at `/proc/<pid>/status` says that `SIGCHLD` is in `SigBlk` (signal blocked) and `ShdPnd` (shared pending signal) state.

Problem: In UNIX, parent process can mask certain signals from receiving for brief period of time so that signal requests that comes to the parent process doesn’t interfere what it is processing currently. When we create `NativeCodeGenerator`, it calls PAL’s `InitializePreCreate()` which blocks `SIGCHLD` from receiving. Later when libuv waits for `SIGCHLD` it never gets it because it was masked (or blocked) by PAL. Searching on PAL's repo, realized they already fixed the issue. So cherry picking https://github.com/dotnet/coreclr/pull/4863.

### DIFF
--- a/pal/src/exception/signal.cpp
+++ b/pal/src/exception/signal.cpp
@@ -155,8 +155,6 @@ void SEHCleanupSignals()
 {
     TRACE("Restoring default signal handlers\n");
 
-    // Do not remove handlers for SIGUSR1 and SIGUSR2. They must remain so threads can be suspended
-    // during cleanup after this function has been called.
     restore_signal(SIGILL, &g_previous_sigill);
     restore_signal(SIGTRAP, &g_previous_sigtrap);
     restore_signal(SIGFPE, &g_previous_sigfpe);
--- a/pal/src/include/pal/seh.hpp
+++ b/pal/src/include/pal/seh.hpp
@@ -25,8 +25,7 @@ Abstract:
 #include "pal/palinternal.h"
 #include "pal/corunix.hpp"
 
-// Uncomment this define to turn off the signal handling thread.
-// #define DO_NOT_USE_SIGNAL_HANDLING_THREAD
+#define DO_NOT_USE_SIGNAL_HANDLING_THREAD
 
 /*++
 Function :
--- a/pal/src/include/pal/threadsusp.hpp
+++ b/pal/src/include/pal/threadsusp.hpp
@@ -361,11 +361,6 @@ namespace CorUnix
                 DWORD *pdwSuspendCount
             );
 
-#if !HAVE_MACH_EXCEPTIONS
-            static 
-            VOID InitializeSignalSets();
-#endif // !HAVE_MACH_EXCEPTIONS
-
             VOID InitializeSuspensionLock();
 
             void SetBlockingPipe(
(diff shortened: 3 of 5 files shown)


kunalspathak:
@jianchun , @digitalinfinity  - Please review.

digitalinfinity:
LGTM- thanks for investigating and fixing!

jianchun:
> LGTM- thanks for investigating and fixing!

Ditto, thanks!

obastemur:
Good work! However I have concerns with the shape of the PR.

I believe all these signal handling has no much place in our code. After the current PR, I think some pieces of the PAL will be obsolete. Any plan to take care of them?

Please think of un-commenting https://github.com/Microsoft/ChakraCore/blob/master/pal/src/include/pal/seh.hpp#L29 instead of removing the signal handling partially. OR lets remove it for good.

kunalspathak:
@dotnet-bot test please

kunalspathak:
@dotnet-bot test Pre-Merge signing tool  please

kunalspathak:
@dotnet-bot  test Ubuntu Ubuntu_linux_release please

kunalspathak:
@dotnet-bot test Ubuntu Ubuntu_linux_test_static please

kunalspathak:
@dotnet-bot test Ubuntu ubuntu_linux_test  please

obastemur:
@kunalspathak Can we just uncomment that one and leave the rest as is? Upstream changed a lot and this update in the middle might leave some dead code behind. If you just uncomment that line, the effect should be the same. 

Later when we update, it will be much easier.

kunalspathak:
> Can we just uncomment that one

I just looked into [latest pal src](https://github.com/dotnet/coreclr/search?utf8=%E2%9C%93&q=DO_NOT_USE_SIGNAL_HANDLING_THREAD) and doesn't seem it is getting used, so safe to delete it out. We can add it back if pal adds and use it.

obastemur:
I agree, they don't have that either. But there are other changes too. We have an interface behind that is also not available on latest PAL. It's like we sync to something in the middle. I agree on the signal issue and appreciate the fix yet we should not cherry pick this one as is.

I've been updating the mach_exceptions and saw how the codebase is different now. 

We may not have to match with PAL 100% though. We are also not matching on `mach` side as is.. They have to manage the whole process life time while we are not responsible from process..

kunalspathak:
> I've been updating the mach_exceptions and saw how the codebase is different now.

I see, then you know more. As you suggested, i have uncommented the `#define`

obastemur:
Thanks @kunalspathak

kunalspathak:
@dotnet-bot  test please

kunalspathak:
Thanks @digitalinfinity , @jianchun , @obastemur  for review!

PR #1797, opened 2016-10-21
swb: annotate BaseDictionary (partial)

jianchun:
Expand `Pointer` macro to choose `WriteBarrierPtr/NoWriteBarrierPtr`
based on optional `Allocator` type using `WriteBarrierPtrTraits`.
(E.g. pointer fields but using with Arena, choose NoWriteBarrierPtr.)

Add `WriteBarrier` function to trigger write barrier on array content
when needed.

Added `CopyArray` function to copy array data. Triggers write barrier
on the array content when needed.

By default only pointer types, WriteBarrierPtr wrapper, or explicit
_write_barrier_policy trigger write barrier (when using Recycler).
Users can also use template instantiation to alter custom allocator/type
write barrier policy.

This change does not handle individual `Item` write barrier yet.

### DIFF
--- a/lib/Common/DataStructures/BaseDictionary.h
+++ b/lib/Common/DataStructures/BaseDictionary.h
@@ -90,17 +90,17 @@ namespace JsUtil
         friend class Js::RemoteDictionary<BaseDictionary>;
         template <typename ValueOrKey> struct ComparerType { typedef Comparer<ValueOrKey> Type; }; // Used by diagnostics to access Comparer type
 
-        int* buckets;
-        EntryType* entries;
-        AllocatorType* alloc;
-        int size;
-        uint bucketCount;
-        int count;
-        int freeList;
-        int freeCount;
+        Pointer(int, TAllocator) buckets;
+        Pointer(EntryType, TAllocator) entries;
+        PointerNoBarrier(AllocatorType) alloc;
+        Field(int) size;
+        Field(uint) bucketCount;
+        Field(int) count;
+        Field(int) freeList;
+        Field(int) freeCount;
 
 #if PROFILE_DICTIONARY
-        DictionaryStats *stats;
+        PointerNoBarrier(DictionaryStats) stats;
 #endif
         enum InsertOperations
         {
@@ -179,11 +179,8 @@ namespace JsUtil
             freeList = other.freeList;
             freeCount = other.freeCount;
 
-            size_t copySize = bucketCount * sizeof(buckets[0]);
-            js_memcpy_s(buckets, copySize, other.buckets, copySize);
-
-            copySize = size * sizeof(entries[0]);
-            js_memcpy_s(entries, copySize, other.entries, copySize);
+            CopyArray<TAllocator>(buckets, bucketCount, other.buckets, bucketCount);
+            CopyArray<TAllocator, EntryType, ValueType>(entries, size, other.entries, size);
 
 #if PROFILE_DICTIONARY
             stats = DictionaryStats::Create(typeid(this).name(), size);
@@ -698,11 +695,8 @@ namespace JsUtil
             freeList = other->freeList;
             freeCount = other->freeCount;
 
-            size_t copySize = bucketCount * sizeof(buckets[0]);
-            js_memcpy_s(buckets, copySize, other->buckets, copySize);
-
-            copySize = size * sizeof(entries[0]);
-            js_memcpy_s(entries, copySize, other->entries, copySize);
+            CopyArray<TAllocator>(buckets, bucketCount, other->buckets, bucketCount);
+            CopyArray<TAllocator, EntryType, ValueType>(entries, size, other->entries, size);
 
 #if PROFILE_DICTIONARY
             stats = DictionaryStats::Create(typeid(this).name(), size);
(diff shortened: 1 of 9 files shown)


jianchun:

jianchun:
Please hold on review -- Curtis suggested a better approach. Will update.

jianchun:

jianchun:
@leirocks @curtisman @digitalinfinity Please CR, thanks!

lib/Common/DataStructures/BaseDictionary.h
```diff
@@ -90,9 +90,9 @@ namespace JsUtil
         friend class Js::RemoteDictionary<BaseDictionary>;
         template <typename ValueOrKey> struct ComparerType { typedef Comparer<ValueOrKey> Type; }; // Used by diagnostics to access Comparer type
 
-        int* buckets;
-        EntryType* entries;
-        AllocatorType* alloc;
+        Pointer(int, TAllocator) buckets;
+        Pointer(EntryType, TAllocator) entries;
+        PointerNoBarrier(AllocatorType) alloc;
         int size;
```
digitalinfinity: For consistency, it's better if we annotate all the fields (so annotate the non-pointer fields with Field)

digitalinfinity:
>             js_memcpy_s(newEntries, sizeof(EntryType) * newSize, entries, sizeof(EntryType) * count);

This probably needs the write barrier but I think you already noted that you haven't set barriers on the items themselves

---
Refers to: lib/Common/DataStructures/BaseDictionary.h:1020 in 9958e2f. [](commit_id = 9958e2fe274083a3a724264de789b61d399db20f, deletion_comment = False)

digitalinfinity:
:shipit:

lib/Common/DataStructures/BaseDictionary.h
jianchun: Yes, will do (paid too much attention to buckets and entries).

lib/Common/DataStructures/BaseDictionary.h
```diff
@@ -860,9 +860,14 @@ namespace JsUtil
...
             AssertMsg(initBucketCount > 0, "Size returned by policy should be greater than 0");
-            Allocate(&buckets, &entries, initBucketCount, initSize);
 
-            // Allocation can throw - assign the size only after allocation has succeeded.
+            int* newBuckets = nullptr;
+            EntryType* newEntries = nullptr;
+            Allocate(&newBuckets, &newEntries, initBucketCount, initSize);
+
+            // Allocation can throw - assign only after allocation has succeeded.
+            SetArray<TAllocator>(this->buckets, newBuckets, initBucketCount);
```
jianchun: The write barrier on elements scenario is hard to understand (I'm still not too clear after asking Lei and Curtis.) `this->buckets = newBuckets` is fine here, but `this->entries = newEntries` is not. I was thinking of how we prevent accidental error. In the case of pointer field, the `operator=` function takes care of all scenarios, be it raw pointer, NoWriteBarrierPtr, or WriteBarrierPtr. So in comparison, I was thinking maybe we require all array assignment through this `SetArray` function. `SetArray` determines if we need write barrier or not. Possibly we can track that in clang plugin (not trivial...).

jianchun:
>             js_memcpy_s(newEntries, sizeof(EntryType) * newSize, entries, sizeof(EntryType) * count);

I was hoping line 1046 below will take care of that. I was unsure if it would be a problem around line 1039 where we `DeleteEntries(entries, ...)`. But line 1046 will write barrier the whole array content. Would it work?

---
In reply to: [255494454](https://github.com/Microsoft/ChakraCore/pull/1797#issuecomment-255494454) [](ancestors = 255494454)

---
Refers to: lib/Common/DataStructures/BaseDictionary.h:1020 in 9958e2f. [](commit_id = 9958e2fe274083a3a724264de789b61d399db20f, deletion_comment = False)

lib/Common/Memory/RecyclerPointers.h
```diff
@@ -6,6 +6,130 @@
 
 namespace Memory
 {
+class Recycler;
+class RecyclerNonLeafAllocator;
+
+// Dummy tag classes to mark yes/no write barrier policy
```
jianchun: Agree, looks quite complex to myself...

lib/Common/DataStructures/BaseDictionary.h
digitalinfinity: Why is `this->entries = newEntries` not fine? this->entries and this->buckets are both pointer types- that is, setting them goes through the barrier on the Dictionary instance itself, not on the entries. If you assign `this->entries = blah`, you don't need to see all the bits corresponding to the range of memory symbolized by the entries array, you need to only set the bit corresponding to the dictionary itself, since that's the memory that has changed. That being said, you'll need to do what SetArray does when you do something like memcpy or set individual elements in the array.

jianchun: Yes, `this->entries = newEntries` is fine when `newEntries` already takes care of write barrier on the array elements. So either I do it here with `SetArray`, or I do something similar ahead replacing `js_memcpy_s` with a version of `CopyArray`. Either approach has the concern that it is hard to know we missed one. Do you think `CopyArray` is safer in terms of write barrier correctness? e.g. is it possible GC collects in between `js_memcpy_s` and my current `SetArray`, and `SetArray` write barrier is too late?

jianchun: Making this change now. Revert `SetArray`, prefer earlier with `CopyArray`.

jianchun:

jianchun:

digitalinfinity:
LGTM

jianchun:

jianchun:
Thanks @digitalinfinity !

PR #1801, opened 2016-10-21
Implementing i32 truncation instructions

Krovatkin:
A stab at implementing:

* i32.trunc_u/f32 (0x9f)
* i32.trunc_u/f64 (0xa0)
* i32.trunc_s/f64 (0xa0) //needs to throw an overflow exception
* i32.trunc_s/f64 (0xa0) //needs to throw an overflow exception


Krovatkin:
fyi : @Cellule @MikeHolman I'll rebase this onto master before I ask you guys to review it.

Issue #1800, opened 2016-10-21
C++ language conformance changes (msvc /permissive-)


Rastaban:
There is a new compiler switch in the next release of MSVC that makes the compiler less permissive (/permissive-)  There are three types of language conformance issues that are exposed in the Chakra codebase when building with /permissive-

1) Strict string conversion (this is also covered by the /Zc:strictStrings flag)

  The 'const' is not implicitly dropped by string literals, which means the following is not allowed:

  `char *str = "const string literal"; //error: cannot convert a 'const char*' to a 'char*'`

  This fix to to make str 'const char*'.  (This can have a domino effect depending on where str is used)

2) Fully qualified inline declarations members inside class

```
   struct A {
      void A::f() { } // Error: illegal qualified name in member declaration, remove redundant 'A::' to fix
   };
```

3) MSVC by default will allows name lookup in a dependent base.  This is disabled by /permissive-

```
    template <class T> struct B {
      void f();
    };

    template <class T> struct D
        : public B<T> //B is a dependent base because its type depends on the type of T in D<T>.
    {
        //One possible fix is to uncomment the following line.  If this were a type we should have 'using typename'...
        //using B<T>::f;

        void g() {
          f(); //Error: identifier not found, one possible fix is change it to 'this->f();'
        }
    };

    void h()
    {
       D<int> d;
       d.g();
    }
```

The MSVC compiler team uses the ChakraCore build as part of their tests, so we have already made the necessary changes in our own branch and are interested in submitting a pull request.

dilijev:
I've encountered the third issue a lot when writing code that needed to compile in both GCC and MSVC, I didn't realize that was an issue of permissivity. It's good to see this change come to ChakraCore, it should help with portability of our code.

dilijev:
Resolved by https://github.com/Microsoft/ChakraCore/pull/1802

PR #1802, opened 2016-10-21
C++ conformance. (building with /permissive-)

Rastaban:
This addresses issue #1800

These issues were found when building with the /permissive- flag in the latest version of MSVC.  No tests were added/modified because this does not change any behavior.

There are three types of language conformance issues fixed in this:

1) Strict string conversion (this is also covered by the /Zc:strictStrings flag)

  The 'const' is not implicitly dropped by string literals, which means the following is not allowed:

```
char *str = "const string literal"; //error: cannot convert a 'const char*' to a 'char*'
```

  This fix to to make str 'const char*'.  (This can have a domino effect depending on where str is used)

2) Fully qualified inline declarations members inside class

```
struct A {
    void A::f() { } // Error: illegal qualified name in member declaration, remove redundant 'A::' to fix
};
```

3) MSVC by default will allows name lookup in a dependent base.  This is disabled by /permissive-

```
template <class T> struct B {
  void f();
};

template <class T> struct D
    : public B<T> //B is a dependent base because its type depends on the type of T in D<T>.
{
    //One possible fix is to uncomment the following line.  If this were a type we should have 'using typename'...
    //using B<T>::f;

    void g() {
      f(); //Error: identifier not found, one possible fix is change it to 'this->f();'
    }
};

void h()
{
   D<int> d;
   d.g();
}
```


Rastaban:
I am looking into the failures now

dilijev:
To prevent regressions against `/permissive-` we should probably add this flag to our various build invocations @boingoing

Rastaban:
It would make my life easier if you added it to the build.  The /permissive- switch is brand new, so you will only be able to use it in the upcoming 'VS 2017' RC.  Also we have not had time to get all of the windows SDK headers clean, so you would not be able to add it to the parts of your project that use WRL.

dilijev:
@Rastaban Since we're doing all of our testing with released products we expect the community to be able to use, we would want to make sure that including that flag when the compiler doesn't understand it won't cause any problems.

Then, we will probably leave it out of build automation (e.g. Jenkins CI) which uses only released products we expect the community to be able to easily use.

We could add it to the VS project such that anyone using the VS 2017 RC would have that flag automatically enabled, but it might be a strange experience to see additional errors when building with that toolchain.

Rastaban:
@dilijev That's fine if you don't want to use /permissive- in your build yet.  Hopefully you don't mind an occasional pull request like this one with /permissive- related updates.  You may consider using the /Zc:strictStrings flag where possible though.

ThomsonTan:
@Rastaban , "/Zc:strictStrings" doesn't sound like a new switch. Is this enabled for varies products? Our project is built under a few configuration, like internal and external. It will be better to keep them consistent.

Rastaban:
@ThomsonTan "/Zc:strictStrings" had been around for awhile and can be added to your build.  /permissive- is new and will imply /Zc:strictStrings (https://blogs.msdn.microsoft.com/vcblog/2016/11/16/permissive-switch/)

Rastaban:
This pull request does not enable /permissive-, but it does fix all the places your build would fail if /permissive- were enabled.  This includes all the places where the code is not yet /Zc:strictStrings compliant.  So if you enable /Zc:strictStrings, I recommend taking this pull request to save you from duplicating those fixes.

dilijev:
Assigned myself so I don't forget about this PR.

dilijev:

bin/rl/rl.cpp
```diff
@@ -4167,8 +4170,8 @@ ProcessConfig
 
       // Walk the condition nodes looking for applicability.
 
-      ConditionNodeList * conditionNodeList = NULL;
-      ConditionNodeList * conditionNodeLast = NULL;
+      conditionNodeList = NULL;
+      conditionNodeLast = NULL;
```
dilijev: Hoisting these declarations above the loop is not useful. They are already scoped to the loop. Additionally, this change is redundant as pertains to the assignment of NULL at initialization time. What problem is this trying to solve that the compiler cannot figure out (e.g. hoisting of declarations)?

bin/rl/rlrun.cpp
```diff
@@ -1016,17 +1016,17 @@ BOOL
...
-        fFailed = TRUE;
-        goto logFailure;
+    {
+        int retval = ExecuteCommand(pDir->GetDirectoryPath(), buf, millisecTimeout, envFlags);
+        if (retval == WAIT_TIMEOUT) {
+            ASSERT(millisecTimeout != INFINITE);
+            LogOut("ERROR: Test timed out after %ul seconds", millisecTimeout / 1000);
+            fFailed = TRUE;
+            goto logFailure;
+        }
```
dilijev: Maybe there's a code cleanliness issue here, but is *NOT* reducing the scope of `retval` going to cause an error with `/permissive-`?

lib/Common/Core/DbgHelpSymbolManager.cpp
```diff
@@ -53,7 +54,7 @@ DbgHelpSymbolManager::Initialize()
 
     // Let's make sure the directory where chakra.dll is, is on the symbol path.
 
-    char16 const * wszModule = AutoSystemInfo::GetJscriptDllFileName();
+    wszModule = AutoSystemInfo::GetJscriptDllFileName();
```
dilijev: So the reason for hoisting this one to the top of the function is because it would allow the variable to be uninitialized in case of `return` or `goto end` in the lines above? Curious, what problem does this cause in real-world code? If it's unused, why require initialization on penalty of compile error?

dilijev:
@tcare Thoughts on these changes?

bin/rl/rl.cpp
Rastaban: The 'goto Label_Skip' a few lines above will skip the initialization, which is not allowed by the C++ standard.  There are a few options of ways to fix this.  I can remove the redundant "=NULL" at the old location, or I can move the initialization to a separate line and keep it in the original location, which would look like this:

ConditionNodeList * conditionNodeLast;
conditionNodeLast = NULL

bin/rl/rlrun.cpp
Rastaban: The 'goto logFailure' up above will skip the initialization of retval.  By adding the {} it will cause retval to go out of scope before the logFailure label.

lib/Common/Core/DbgHelpSymbolManager.cpp
Rastaban: The problem is not the `return`, it is just the `goto end.` that is skipping the declaration with an initializer.  Here is the applicable paragraph from the C++11 standard 
[stmt.dcl]/3
It is possible to transfer into a block, but not in a way that bypasses declarations with initialization. A program that jumps90 from a point where a variable with automatic storage duration is not in scope to a point where it is in scope is ill-formed unless the variable has scalar type, class type with a trivial default constructor and a trivial destructor, a cv-qualified version of one of these types, or an array of one of the preceding types **and is declared without an initializer** (8.6).

dilijev: @Rastaban Thanks for clarifying, I see the problem now.

bin/rl/rlrun.cpp
dilijev: @Rastaban got it, thanks.

tcare:
Looks good. CC @EdMaurer for any updates to style guide we may need to make.

bin/rl/rl.cpp
dilijev: I see why they have to be broken up. I personally think that variables should be explicitly initialized when they are declared, so the code as-is seems fine. 

There are sometimes performance reasons that we want to avoid explicit initialization. This wouldn't apply in this case, but for example, declaring a pointer and then passing a pointer to that pointer to a function as an outparam -- the compiler wouldn't elide an explicit initialization because it doesn't know about the semantics of that value when passed in to the function.

I just want to be sure that using `/permissive-` wouldn't prevent us from making that decision.

Rastaban: The following is valid, which means it is allowed under /permissive-:

    SomeType* ptr; //declared without an initializer, can skip over with a goto
    initializeSomeType(ptr);

dilijev: Thanks for clarifying!

dilijev:
@Rastaban I'll bring this in to run tests internally and then merge it. No logic has changed so this should be straightforward.

PR #1803, opened 2016-10-21
Fixing on reentrant parsemodulesource.

akroshg:
When we get the ParseModuleSource while we fetch module - we go into a state where we don't notify the host about the error or ready state. That makes the module hung.
Fixed that by setting parent and child relation in the InitializeModuleRecord. That way when the ParseModuleSource re-entrancy happens and we can bubble up the error/ready state to parent properly.
Added native unittest for forcing the re-entrant parsemodulesource.

### DIFF
--- a/lib/Runtime/Language/SourceTextModuleRecord.cpp
+++ b/lib/Runtime/Language/SourceTextModuleRecord.cpp
@@ -136,6 +136,15 @@ namespace Js
             if (SUCCEEDED(hr))
             {
                 hr = PostParseProcess();
+                if (hr == S_OK && this->errorObject != nullptr && this->hadNotifyHostReady)
+                {
+                    // This would be the case where the child module got error and current module has notified error already.
+                    if (*exceptionVar == nullptr)
+                    {
+                        *exceptionVar = this->errorObject;
+                    }
+                    return E_FAIL;
+                }
             }
         }
         if (FAILED(hr))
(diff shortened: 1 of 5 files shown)


akroshg:
@boingoing please have a look

akroshg:
@boingoing ping.

AtOMiCNebula:
Sorry to get bogged down on this...some other stuff has come up, and I may not be able to get back to this until later in the week.  I'll try to get some tests written, though. :)

lib/Runtime/Language/SourceTextModuleRecord.cpp
```diff
@@ -547,10 +555,36 @@ namespace Js
...
+        if (!parentRecord->childrenModuleSet->ContainsKey(moduleName))
+        {
+            parentRecord->childrenModuleSet->AddNew(moduleName, this);
+
+            if (this->parentModuleList == nullptr)
+            {
+                Recycler* recycler = GetScriptContext()->GetRecycler();
+                this->parentModuleList = RecyclerNew(recycler, ModuleRecordList, recycler);
+            }
+            bool contains = this->parentModuleList->Contains(parentRecord);
```
boingoing: Nitpick: We don't need to see if parentRecord is in the list if we just constructed the list.

boingoing:
LGTM 🚢

akroshg:
rebasing this pr on release/1.4
#2110 
Closing this PR.

PR #1806, opened 2016-10-22
WebAssembly Int64

Cellule:
This add support for Int64 in WebAssembly.
Also contains some refactoring to more easily support different asm.js/wasm types

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/1806)


lib/Backend/Lower.cpp
```diff
@@ -8824,7 +8851,16 @@ Lowerer::LowerLdArrViewElem(IR::Instr * instr)
         }
         done = instr;
     }
-    InsertMove(dst, src1, done);
+    if (dst->IsInt64())
```
MikeHolman: IMO it would be nice if InsertMove took care of this rather than doing the check at all the call sites (same with ChangeToAssign)

Cellule: I agree, the reason I had to do this is because we need an instance of Lowerer in order to lower int64 assign, ChangeToAssign and InsertMove are static.
I tried changing them to be methods, but if I recall there are place in LinearScan where we call these methods to handle move instructions without a Lowerer.

MikeHolman:
:shipit:

Cellule:
I fixed the serializer issues I was seeing in full. @MikeHolman Could review the last 3 commits please ?

PR #1812, opened 2016-10-24
Implementation of i32 truncation instructions

### LINKED ISSUE #1857: WASM - i32.trunc_u/f32
Implement operator i32.trunc_u/f32

### LINKED ISSUE #1858: WASM - i32.trunc_u/f64
Implement operator i32.trunc_u/f64

Krovatkin:
A stab at implementing:
- i32.trunc_u/f32 (0x9f)
- i32.trunc_u/f64 (0xa0)
- i32.trunc_s/f64 (0xa0) //needs to throw an overflow exception
- i32.trunc_s/f64 (0xa0) //needs to throw an overflow exception

Closes #1857 
Closes #1858 

---

This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/1812)


Krovatkin:
@Cellule Could you please take a look whenever you have a mnt? 
fyi : @commonlisp @arunetm

Cellule:
Reviewed 19 of 19 files at r1.
Review status: all files reviewed at latest revision, 6 unresolved discussions.

---

*[lib/Backend/LowerMDShared.cpp, line 5827 at r1](https://reviewable.io:443/reviews/microsoft/chakracore/1812#-KUsYw5xyrS1mP8Z9jJy:-KUsYw5ydFhBxRHyP-Bo:bwrnm8e) ([raw file](https://github.com/microsoft/chakracore/blob/f571eb99132b044fe651ed93a0d21d7f63575b01/lib/Backend/LowerMDShared.cpp#L5827)):*
> ```cpp
> IR::RegOpnd* LowererMD::MaterializeDoubleConstFromInt(int intConst, IR::Instr* instr)
> {
>     IR::Opnd * intReg = IR::RegOpnd::New(TyMachPtr, m_func);
> ```

Use `TyMachReg` instead

---

*[lib/Backend/LowerMDShared.cpp, line 5828 at r1](https://reviewable.io:443/reviews/microsoft/chakracore/1812#-KUsL4EdQSLTwr1UrCaO:-KUsL4Ee3D24EHPog1Fs:b1hukj4) ([raw file](https://github.com/microsoft/chakracore/blob/f571eb99132b044fe651ed93a0d21d7f63575b01/lib/Backend/LowerMDShared.cpp#L5828)):*
> ```cpp
> {
>     IR::Opnd * intReg = IR::RegOpnd::New(TyMachPtr, m_func);
> #if FLOATVAR
> ```

You should use `IntConstOpnd` with `TyMachReg` here. We avoid using `AddrOpnd` when dealing with integer.

---

*[lib/Backend/LowerMDShared.cpp, line 5867 at r1](https://reviewable.io:443/reviews/microsoft/chakracore/1812#-KUsYyl-nuTd1nPRCRBa:-KUsYyl-nuTd1nPRCRBb:bjqdolf) ([raw file](https://github.com/microsoft/chakracore/blob/f571eb99132b044fe651ed93a0d21d7f63575b01/lib/Backend/LowerMDShared.cpp#L5867)):*
> ```cpp
>     if (instr->GetDst()->IsUInt32())
>     {
>         min = 0;
> ```

In this case, we don't need to convert 0 to double, we should simply `xor` the xmm register and that's it

---

*[lib/Backend/LowerMDShared.cpp, line 5868 at r1](https://reviewable.io:443/reviews/microsoft/chakracore/1812#-KUsZ3EApejRz73m9xVE:-KUsZ3EApejRz73m9xVF:bnip7m4) ([raw file](https://github.com/microsoft/chakracore/blob/f571eb99132b044fe651ed93a0d21d7f63575b01/lib/Backend/LowerMDShared.cpp#L5868)):*
> ```cpp
>     {
>         min = 0;
>         max = INT_MAX; //we will + 2^31 below
> ```

It should be possible on x64 to simply convert `UINT_MAX` directly to double as VC is doing that too.

---

*[lib/Backend/LowerMDShared.cpp, line 5880 at r1](https://reviewable.io:443/reviews/microsoft/chakracore/1812#-KUsZDmCAFB01jbWzKTL:-KUsZDmCAFB01jbWzKTM:b9ag8o0) ([raw file](https://github.com/microsoft/chakracore/blob/f571eb99132b044fe651ed93a0d21d7f63575b01/lib/Backend/LowerMDShared.cpp#L5880)):*
> ```cpp
> 
>     GenerateCheck(Js::OpCode::COMISD, src64, limitReg, Js::OpCode::JB, throwLabel, instr); // < min
>     IR::Opnd * twoTo31Float = MaterializeConstFromBits(_mm_extract_epi32(X86_TWO_31_F4.m128i_value, 0), TyFloat32, instr);
> ```

Move this opnd declaration to its condition as it is only used in the unsigned x86 case.

---

*[lib/Backend/LowerMDShared.cpp, line 5953 at r1](https://reviewable.io:443/reviews/microsoft/chakracore/1812#-KUs7r7APJfuGztZOhQi:-KUs7r7APJfuGztZOhQj:botgs2l) ([raw file](https://github.com/microsoft/chakracore/blob/f571eb99132b044fe651ed93a0d21d7f63575b01/lib/Backend/LowerMDShared.cpp#L5953)):*
>  
>  

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/microsoft/chakracore/1812)*

Cellule:
@MikeHolman could you take a look also please, mostly LowerMDShared.cpp

lib/Backend/LowerMDShared.cpp
```diff
@@ -5812,6 +5812,165 @@ LowererMD::GenerateCtz(IR::Instr * instr)
...
+LowererMD::GenerateTruncWithCheck(IR::Instr * instr)
+{
+    Assert(AutoSystemInfo::Data.SSE2Available());
+
+    GenerateTruncChecks(instr); //converts src to double and checks if  MIN <= src <= MAX
+
+    IR::Opnd* src1 = instr->GetSrc1();
+    IR::Opnd* dst = instr->GetDst();
+
+    IR::Opnd * twoTo31Float = MaterializeConstFromBits(_mm_extract_epi32(X86_TWO_31_F4.m128i_value, 0), TyFloat32, instr);
```
MikeHolman: this intrinsic requires SSE4. please just define a constant for this

lib/Backend/LowerMDShared.cpp
```diff
@@ -5812,6 +5812,165 @@ LowererMD::GenerateCtz(IR::Instr * instr)
...
+
+    return tmpReg;
+}
+
+IR::Opnd*
+LowererMD::TruncLower31BitsDb(IR::Instr* instr, IR::Opnd* src1, IR::Opnd* dst, IR::Opnd* twoTo31Float)
+{
+    IR::RegOpnd * tmpReg = IR::RegOpnd::New(TyFloat64, m_func);
+    IR::Opnd * result = IR::RegOpnd::New(TyFloat64, m_func);
+    EmitFloat32ToFloat64(result, twoTo31Float, instr);
```
MikeHolman: we shouldn't add runtime conversion. at jit time just emit load of double value...

lib/Backend/LowerMDShared.cpp
```diff
@@ -5812,6 +5812,165 @@ LowererMD::GenerateCtz(IR::Instr * instr)
...
+    instr->InsertBefore(IR::Instr::New(Js::OpCode::SUBPD, result, result, tmpReg, m_func));
+    instr->InsertBefore(IR::Instr::New(Js::OpCode::CVTTSD2SI, dst, result, m_func));
+
+    instr->InsertBefore(IR::Instr::New(Js::OpCode::CVTSD2SS, twoTo31Float, tmpReg, m_func));
+
+    return twoTo31Float;
+}
+
+void
+LowererMD::GenerateTruncWithCheck(IR::Instr * instr)
```
MikeHolman: Note that you might want to make use of InsertCompareBranch

---
In reply to: [84817095](https://github.com/Microsoft/ChakraCore/pull/1812#discussion_r84817095) [](ancestors = 84817095)

lib/Backend/LowerMDShared.cpp
Krovatkin: arrgh, right... missed it :-(

lib/Backend/LowerMDShared.cpp
Krovatkin: i was trying to avoid inserting more branches then necessary... did it get too unwieldy?

MikeHolman: Doesn't your code have more branches? Mine only has 1 (not taken) branch in the main case (helper block will be moved past ret in layout phase so main path will just have cvtss2si dst, src1; cmp src1, 80000000H; je helper; )

---
In reply to: [84940996](https://github.com/Microsoft/ChakraCore/pull/1812#discussion_r84940996) [](ancestors = 84940996)

lib/Backend/LowerMDShared.cpp
```diff
@@ -5812,6 +5815,114 @@ LowererMD::GenerateCtz(IR::Instr * instr)
...
+    instr->InsertBefore(conversion);
+    return src64;
+}
+
+void
+LowererMD::GenerateTruncWithCheck(IR::Instr * instr)
+{
+    Assert(AutoSystemInfo::Data.SSE2Available());
+
+    IR::Opnd* src64 = GenerateTruncChecks(instr); //converts src to double and checks if  MIN <= src <= MAX
```
Krovatkin: @MikeHolman 

Agreed. For the unsigned case, we would still have **3** checks in the worst case scenario and **2** checks in the best case?  
 
We check for an overflow `80000000H`  (1st check) 
* No overflow
  * Check if the source is positive (2 checks)
* Overflow 
  * Check if the source is positive, 2 checks (possibly **MIN_INT**-1)
  * Check if the source is less than **MAX_UINT**, 3 checks

Does this sound right?

MikeHolman: Not exactly. For unsigned I think we can do the following:

1. Shift the range so conversion will not overflow (cvtss2si considers anything outside range [-2^31,2^31-1] as overflowing). This means, subtract 2^31 from the source so all valid uints are now within that range, where the conversion ops will not overflow.
2. Do the conversion.
3. Check for overflow.
    * a. On overflow, check if the initial value was (double)MAX_UINT, and if so we are done. If not, trap.
    * b. If not overflow, ADD 2^31 to the result to shift the range back.

For signed, it is even simpler:

1. Do the conversion.
2. Check for overflow.
    * a. On overflow, check if the initial value was (double)MAX_INT, and if so we are done. If not, trap.
    * b. Otherwise, we are done.

Am I missing something?

Krovatkin:
@MikeHolman darn, i didn't realize that subtracting 2^31 results in a more efficient sequence. My bad :-( .  Are we positive we need to compare against `(double)MAX_INT` and **not** against `(double)MIN_INT` (in signed case) and against `(double)MAX_UINT` and **not** `0` in (unsigned case)?

MikeHolman:
Oh yeah I think you're right about that

Krovatkin:
@arunetm @commonlisp could you please take a look whenever you have a mnt? I incorporated Mike's latest feedback. I want to make sure I didn't introduce any obvious goofs before I ask him to look at it again.

lib/Backend/Lower.cpp
```diff
@@ -1863,8 +1863,20 @@ Lowerer::LowerRange(IR::Instr *instrStart, IR::Instr *instrEnd, bool defaultDoFa
...
-
         case Js::OpCode::Conv_Prim:
+        {
+
+            IR::RegOpnd* src1 = instr->GetSrc1()->IsRegOpnd() ? instr->GetSrc1()->AsRegOpnd() : nullptr;
+            if (src1 && src1->m_sym->IsSingleDef() && src1->m_sym->GetInstrDef()->m_opcode == Js::OpCode::OverflowCheck)
+            {
+                //TODO : overflow checks shouldn't be CSE'ed because of side effects.
+                //DbCheck if it's safe to remove mov?
+                src1->m_sym->GetInstrDef()->m_opcode = Js::OpCode::MOV;
```
Krovatkin: i'm actually thinking if i should just add a case to a switch in `Lowerer::LowerRange`. It will prlly look cleaner

lib/Backend/Lower.cpp
```diff
@@ -1863,8 +1863,19 @@ Lowerer::LowerRange(IR::Instr *instrStart, IR::Instr *instrEnd, bool defaultDoFa
...
+        case Js::OpCode::OverflowCheck:
+            instr->m_opcode = Js::OpCode::MOV;
+            break;
         case Js::OpCode::Conv_Prim:
+        {
+
+            IR::RegOpnd* src1 = instr->GetSrc1()->IsRegOpnd() ? instr->GetSrc1()->AsRegOpnd() : nullptr;
+            if (src1 && src1->m_sym->IsSingleDef() && src1->m_sym->GetInstrDef()->m_opcode == Js::OpCode::OverflowCheck)
+            {
+                GenerateTruncWithCheck(instr);
```
commonlisp: Should instr still be removed in this case?

lib/Backend/Lower.cpp
```diff
@@ -1863,8 +1863,19 @@ Lowerer::LowerRange(IR::Instr *instrStart, IR::Instr *instrEnd, bool defaultDoFa
...
+                break;
+            }
+
             if (instr->GetDst()->IsFloat())
```
commonlisp: Wonder if we might as well factor out instr->GetDst() and instr->GetSrc1() since they are everyone here.

lib/Runtime/Math/WasmMath.h
```diff
@@ -17,6 +17,8 @@ class WasmMath
     template<typename T> static T Nearest(T aLeft);
     static int Rol(int aLeft, int aRight);
     static int Ror(int aLeft, int aRight);
+    template <typename T, T MIN, T MAX> static bool IsInValidRange (double aLeft);
```
commonlisp: Hmm, wondering if this is in the right place. IsInValidRange isn't actually exposed through Wasm.

lib/Runtime/Language/InterpreterHandlerAsmJs.inl
```diff
@@ -180,6 +180,10 @@ EXDEF2_WMS( F1toF1Mem        , Nearest_Flt      , Wasm::WasmMath::Nearest<float>
 EXDEF2_WMS( D1toD1Mem        , Trunc_Db         , Wasm::WasmMath::Trunc<double>                      )
 EXDEF2_WMS( D1toD1Mem        , Nearest_Db       , Wasm::WasmMath::Nearest<double>                    )
 EXDEF2_WMS( I1toI1Mem        , PopCnt_Int       , ::Math::PopCnt32                                   )
+EXDEF2_WMS( D1toI1Scr        , Conv_Check_DTI   , (OP_TruncWithCheck<int,INT_MIN,INT_MAX>)           )
```
commonlisp: Does the parser require the enclosing parens for the TruncWithCheck?

lib/Backend/LowerMDShared.cpp
```diff
@@ -5812,6 +5814,107 @@ LowererMD::GenerateCtz(IR::Instr * instr)
...
+{
+    IR::Opnd * regBits = IR::RegOpnd::New(TyInt32, m_func);
+    this->m_lowerer->InsertMove(regBits, IR::IntConstOpnd::New(bits, TyInt32, m_func), instr);
+    IR::RegOpnd * regConst = IR::RegOpnd::New(type, m_func);
+    instr->InsertBefore(IR::Instr::New(Js::OpCode::MOVD, regConst, regBits, m_func));
+    return regConst;
+}
+
+IR::Opnd*
+LowererMD::Subtract2To31Db(IR::Opnd* src1, IR::Opnd* intMinFP, IR::Instr* instr)
```
Krovatkin: fixed

lib/Runtime/Language/InterpreterHandlerAsmJs.inl
Krovatkin: yes since it contains a ",". Initially, I used `#define COMMA ,`  but it doesn't work with clang.

Krovatkin:
@MikeHolman Please take a look whenever you have a moment.

MikeHolman:
:shipit:

lib/Backend/LowerMDShared.cpp
```diff
@@ -5928,6 +5930,98 @@ LowererMD::GenerateCtz(IR::Instr * instr)
...
+    IR::Opnd * regBits = IR::RegOpnd::New(TyInt32, m_func);
+    this->m_lowerer->InsertMove(regBits, IR::IntConstOpnd::New(bits, TyInt32, m_func), instr);
+    IR::RegOpnd * regConst = IR::RegOpnd::New(type, m_func);
+    instr->InsertBefore(IR::Instr::New(Js::OpCode::MOVD, regConst, regBits, m_func));
+    return regConst;
+}
+
+IR::Opnd* LowererMD::Subtract2To31(IR::Opnd* src1, IR::Opnd* intMinFP, IRType type, IR::Instr* instr)
+{
+    Js::OpCode op = (type == TyFloat32) ? Js::OpCode::SUBPS : Js::OpCode::SUBPD;
```
MikeHolman: nit: i think scalar op should probably use SUBSD/SUBSS?

Krovatkin:
@Cellule @MikeHolman Guys would it be okay if I close this PR and consolidate all trap and truncation-related stuff in this PR https://github.com/Microsoft/ChakraCore/pull/1848? My changes appear to be pretty tightly coupled and if I were to commit them in the wrong order or rebase individually I would be in the world of pain :-(

Cellule:
Fine by me

Krovatkin:
This item will be tracked in PR #1848

PR #1815, opened 2016-10-24
Fix a functional redeferral issue involving inlinees. 

pleath:
A function that has been inlined in a non-deferred function cannot itself be deferred. This means that we need two passes over the set of active FunctionBody's to determine the set of legal candidates. The first pass closes over the set of candidates: the functional and heuristic rules are applied, and for each function that will not be redeferred, we (recursively) mark all its inlinees as non-candidates. The second pass redefers all the functions that have not be marked as non-candidates.


pleath:

PR #1819, opened 2016-10-25
swb: annotate BaseDictionary entry items

jianchun:
BaseDictionary Key or Value can be of GC pointer type, potentially needs
write barrier on each entry item.

Use macro and template type traits to replace underlying entry Key/Value
type with WriteBarrierPtr when Key or Value is pointer type and used with
recycler allocators.

BaseDictionary methods still use original Key/Value types. Only the
underlying EntryType may contain swapped type. Most methods are not
affected. However, when WriteBarrierPtr is used, only immutable
TryGetReference should be used. (We cannot get a mutable reference and
possibly change the pointer underneath WriteBarrierPtr.)

Allow WriteBarrierPtr to take immutable address to support TryGetReference
(used a lot). Fixed a bunch of incorrect `const` usage.

Update: Apply template traits to field types to simplify write barrier field
annotations. Now most annotations are Field or FieldNoBarrier (for pointers
known not needing write barrier).

### DIFF
--- a/lib/Common/Memory/RecyclerPointers.h
+++ b/lib/Common/Memory/RecyclerPointers.h
@@ -77,6 +77,27 @@ template <class T,
           class Policy = typename AllocatorWriteBarrierPolicy<Allocator, T*>::Policy>
 struct WriteBarrierPtrTraits { typedef typename _WriteBarrierPtrPolicy<T, Policy>::Ptr Ptr; };
 
+// Choose WriteBarrierPtr type if Allocator is recycler type and element type T
+// is a pointer type, otherwise use type T unchanged.
+//
+// Used to wrap array item type when write barrier is needed (wraps pointer
+// item type with WriteBarrierPtr).
+//
+template <class T, class Policy>
+struct _ArrayItemTypeTraits
+{
+    typedef T Type;
+};
+template <class T>
+struct _ArrayItemTypeTraits<T*, _write_barrier_policy>
+{
+    typedef WriteBarrierPtr<T> Type;
+};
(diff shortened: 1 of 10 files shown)


jianchun:

jianchun:
@dotnet-bot test Ubuntu daily_ubuntu_linux_release please

jianchun:
@leirocks @digitalinfinity @curtisman Please help CR, thanks!

jianchun:

lib/Common/Memory/RecyclerPointers.h
```diff
@@ -216,21 +214,24 @@ class WriteBarrierPtr
...
+    T* const * AddressOf() const { return &ptr; }
     T** AddressOf() { return &ptr; }
 
-    T const** operator&() const
-    {
-        static_assert(false, "Might need to set barrier for this operation, and use AddressOf instead.");
-        return &ptr;
-    }
-    T** operator&()
+    // Taking immutable address is ok
```
leirocks: >Taking immutable address is ok [](start = 7, length = 30)

we might need const_cast check in our tool? if so please add a TODO here

leirocks:
:shipit:

lib/Common/Memory/RecyclerPointers.h
jianchun: Talked offline. @leirocks's idea is to examine and baseline our usage of const_cast to prevent future incorrect use related to this code. That is doable. Will add a note here.

jianchun:
Thanks @leirocks !

PR #1823, opened 2016-10-25
Add out of bound error for wasm array access in interpreter

### LINKED ISSUE #1877: WASM - Array access out of bounds
Throw trap when doing out of bounds access

meg-gupta:
Closes #1877


meg-gupta:
@Cellule Can you please take a look ?

test/wasm/array.js
```diff
@@ -0,0 +1,6 @@
+const blob = WScript.LoadBinaryFile('array.wasm');
+const moduleBytesView = new Uint8Array(blob);
+var a = Wasm.instantiateModule(moduleBytesView, {}).exports;
+print(a["goodload"](0));
+//print(a["badload"](0));
```
Cellule: You should do a try/catch here and print `failed` if we don't trap

test/wasm/array.wast
```diff
@@ -0,0 +1,7 @@
+(module
```
Cellule: Missing copyright header, check other .wast files for example

test/wasm/array.js
```diff
@@ -0,0 +1,6 @@
+const blob = WScript.LoadBinaryFile('array.wasm');
```
Cellule: Missing copyright header

Cellule: You should add this test to `rlexe.xml` you will need to add a baseline file for it too.

lib/Runtime/Language/InterpreterStackFrame.cpp
```diff
@@ -8208,6 +8208,12 @@ const byte * InterpreterStackFrame::OP_ProfiledLoopBodyStart(const byte * ip)
         {
             BYTE* buffer = arr->GetBuffer();
             *(T2*)(buffer + index) = (T2)GetRegRaw<T2>(value);
+            return;
+        }
+        if (this->m_functionBody->IsWasmFunction())
```
Cellule: You should do the check and trapping in OP_StArrWasm and OP_LdArrWasm, we don't want to add checks for normal asm.js

lib/Runtime/Language/InterpreterStackFrame.cpp
```diff
@@ -8208,6 +8208,12 @@ const byte * InterpreterStackFrame::OP_ProfiledLoopBodyStart(const byte * ip)
...
+        {
+            //MGTODO : Change this to throw WebAssembly.RuntimeError once implemented
+            Js::Throw::FatalInternalError();
```
Cellule: You should throw a normal javascript runtime error for the time being so we can try/catch it in tests

lib/Backend/amd64/LowererMDArch.cpp
```diff
@@ -1177,6 +1177,28 @@ LowererMDArch::LowerAsmJsLdElemHelper(IR::Instr * instr, bool isSimdLoad /*= fal
         Lowerer::InsertBranch(Js::OpCode::Br, doneLabel, loadLabel);
         done = doneLabel;
     }
+    else if (instr->m_func->GetJITFunctionBody()->IsWasmFunction())
```
Cellule: I haven't looked yet at this file, but since this Pull Request is for Interpreter support, I would do this part later, also because I would prefer having x64 and x86 support at the same time

meg-gupta: I updated the PR with a new commit to handle x86 as well.

Cellule:
Can you rebase this change on top of master and also run `ch spec.js -args testsuite-bin\address.json -endargs -on:wasm` in `test\wasmspec`.
You will have to edit `spec.js` to 
```js
         case 'assert_trap':
            assertTrap(m, command);
            break;
```
I tried real quick and it crashes in interpreter and doesn't throw with `-maic:0` (aka -forcenative for asm.js code)

Cellule:
I also found a case in `WasmByteCodeGenerator.cpp` where we emit a trap by default if there is no memory in the module
```
    // todo:: check for imported memory
    if (m_module->GetMemory() == nullptr)
    {
        // todo:: make that an out of bounds trap
        m_writer.EmptyAsm(Js::OpCodeAsmJs::Unreachable_Void);
    }
```
You should fix that as part of this change.

meg-gupta:
Updated the commit to handle the case when index + sizeof the memop exceeds the heapsize. There are pending failures that need to be handled for the spec.
- Right now JIT is deadstoring an operator like ( drop ( load *)), but the baseline expects a trap on this load.
- The bytecode is generated such a way that when index + offset overflows, the load sees a wrapped around index and there is no trap

Cellule:
Reviewed 3 of 9 files at r2, 4 of 6 files at r3, 4 of 4 files at r4.
Review status: all files reviewed at latest revision, 6 unresolved discussions, some commit checks failed.

---

*[lib/Backend/LowerMDShared.cpp, line 291 at r4](https://reviewable.io:443/reviews/microsoft/chakracore/1823#-KVCaaMFbPKFptRAVhzO:-KVCaaMFbPKFptRAVhzP:ba1djkf) ([raw file](https://github.com/microsoft/chakracore/blob/e00165d45c8acebefed5e62899ccbc55d8b02bc3/lib/Backend/LowerMDShared.cpp#L291)):*
> ```cpp
>     {
>         Lowerer::InsertAdd(true, IR::RegOpnd::New(TyMachReg, m_func), indexOpnd, IR::IntConstOpnd::New(addrOpnd->GetSize() - 1, TyMachReg, m_func), helperLabel);
>         Lowerer::InsertBranch(Js::OpCode::JO, helperLabel, helperLabel);
> ```

On x64, since the index opnd is 32 bits and TyMachReg is 64bits is impossible to overflow this addition.
You could do the following check
```cpp
        if (indexOpnd->GetSize() >= TySize[TyMachReg])
        {
            // Check if the addition overflowed
            Lowerer::InsertBranch(Js::OpCode::JO, helperLabel, helperLabel);
        }
```

---

*[lib/Backend/LowerMDShared.cpp, line 293 at r4](https://reviewable.io:443/reviews/microsoft/chakracore/1823#-KVC_m9MMpUennn9Zl8g:-KVC_m9MMpUennn9Zl8h:byz8k3q) ([raw file](https://github.com/microsoft/chakracore/blob/e00165d45c8acebefed5e62899ccbc55d8b02bc3/lib/Backend/LowerMDShared.cpp#L293)):*
> ```cpp
>         Lowerer::InsertBranch(Js::OpCode::JO, helperLabel, helperLabel);
>     }
>     m_lowerer->InsertCompareBranch(indexOpnd, arrayLenOpnd, Js::OpCode::BrGe_A, true, helperLabel, helperLabel);
> ```

You have to use the result of the addition here

---

*[lib/Runtime/Language/InterpreterStackFrame.cpp, line 8291 at r4](https://reviewable.io:443/reviews/microsoft/chakracore/1823#-KVCNd_eDdnvDHV629dH:-KVCNd_eDdnvDHV629dI:b7wemfv) ([raw file](https://github.com/microsoft/chakracore/blob/e00165d45c8acebefed5e62899ccbc55d8b02bc3/lib/Runtime/Language/InterpreterStackFrame.cpp#L8291)):*
> ```cpp
>             BYTE* buffer = arr->GetBuffer();
>             *(ArrayType*)(buffer + index) = (ArrayType)GetRegRaw<RegType>(regSlot);
>             return;
> ```

remove this return

---

*[lib/Runtime/Language/InterpreterStackFrame.cpp, line 8365 at r4](https://reviewable.io:443/reviews/microsoft/chakracore/1823#-KVCW_nY-4pYhKUM8wvN:-KVCW_nY-4pYhKUM8wvO:bpwyxqb) ([raw file](https://github.com/microsoft/chakracore/blob/e00165d45c8acebefed5e62899ccbc55d8b02bc3/lib/Runtime/Language/InterpreterStackFrame.cpp#L8365)):*
> ```cpp
>         const uint32 index = (uint32)GetRegRawInt(playout->SlotIndex);
>         JavascriptArrayBuffer* arr = *(JavascriptArrayBuffer**)GetNonVarReg(AsmJsFunctionMemory::ArrayBufferRegister);
>         if (index >= arr->GetByteLength() || index + TypeToSizeMap[playout->ViewType] > arr->GetByteLength())
> ```

@MikeHolman Do you think there is a better way to check this case.
We need to trap if you are trying to access `mem[65533]` for a int32 with the max length being `65536`.
However, I would really to find a way to avoid 2 checks.

The only thing I can think of is
`if (index > arr->GetByteLength() - TypeToSizeMap[playout->ViewType])`
Since we should trap on every memory access if there is no memory, then `arr->GetByteLength() > 0` in here.

---

*[lib/Runtime/Language/InterpreterStackFrame.cpp, line 8369 at r4](https://reviewable.io:443/reviews/microsoft/chakracore/1823#-KVCQQpYllc2qzgN82ra:-KVCQQpYllc2qzgN82rb:b-g0rdly) ([raw file](https://github.com/microsoft/chakracore/blob/e00165d45c8acebefed5e62899ccbc55d8b02bc3/lib/Runtime/Language/InterpreterStackFrame.cpp#L8369)):*
> ```cpp
>             JavascriptError::ThrowRangeError(scriptContext, JSERR_InvalidTypedArrayIndex);
>         }
>         (this->*LdArrFunc[playout->ViewType])(index, playout->Value);
> ```

The call to this function will do an additional check. If you can find a clean way to avoid the additional check it would be nice.
Otherwise we'll leave it as a todo for perf optimizations

---

*[test/wasm/array.js, line 9 at r4](https://reviewable.io:443/reviews/microsoft/chakracore/1823#-KVCNFUdYgdQFntmOQdr:-KVCNFUdYgdQFntmOQds:bua5inj) ([raw file](https://github.com/microsoft/chakracore/blob/e00165d45c8acebefed5e62899ccbc55d8b02bc3/test/wasm/array.js#L9)):*
> ```JavaScript
> const moduleBytesView = new Uint8Array(blob);
> var a = Wasm.instantiateModule(moduleBytesView, {}).exports;
> print(a["goodload"](0));
> ```

Could you please test the following cases
`a.goodload(65535) == 0`
`a.goodload(65536) => trap`
// Add gootstore method
`a.goodstore(0) == 0`
`a.goodstore(65535) == 0`
`a.goodstore(65536) => trap`

Could you also add a few cases with 32bits load/store

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/microsoft/chakracore/1823)*

MikeHolman:
>         // any access below 0x10000 is safe

this is not true in wasm. WebAssembly memory can be any number of pages.

---
Refers to: lib/Backend/Lower.cpp:8847 in e15956a. [](commit_id = e15956ad0da1db58e97d95d55155bdf31bae42fe, deletion_comment = False)

lib/Backend/LowerMDShared.cpp
```diff
@@ -268,6 +268,39 @@ LowererMD::LowerAsmJsCallE(IR::Instr * callInstr)
...
+    IR::LabelInstr * doneLabel = Lowerer::InsertLabel(false, instr);
+
+    // Find array buffer length
+    IR::RegOpnd * indexOpnd = addrOpnd->AsIndirOpnd()->GetIndexOpnd();
+    IR::RegOpnd * baseOpnd = addrOpnd->AsIndirOpnd()->GetBaseOpnd();
+    Assert(baseOpnd->m_sym->IsSingleDef());
+    IR::Instr *defInstr = baseOpnd->m_sym->m_instrDef;
+    IR::RegOpnd *arrayBuffer = defInstr->GetSrc1()->AsIndirOpnd()->GetBaseOpnd();
+    IR::Opnd *srcOpnd = IR::IndirOpnd::New(arrayBuffer, Js::ArrayBuffer::GetByteLengthOffset(), TyMachReg, m_func);
+    IR::RegOpnd *arrayLenOpnd = IR::RegOpnd::New(TyMachReg, m_func);
```
MikeHolman: buffer length is already in src2

Cellule:
I think we should make `AsmJsJITInfo::AccessNeedsBoundCheck` return return true for WebAssembly so we don't fall in that case

---

Reviewed 3 of 3 files at r5.
Review status: all files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---

*[lib/WasmReader/WasmByteCodeGenerator.cpp, line 1129 at r5](https://reviewable.io:443/reviews/microsoft/chakracore/1823#-KVCgrrS1xvJjOJOU4LF:-KVCgrrS1xvJjOJOU4LG:b-buf591) ([raw file](https://github.com/microsoft/chakracore/blob/e15956ad0da1db58e97d95d55155bdf31bae42fe/lib/WasmReader/WasmByteCodeGenerator.cpp#L1129)):*
> ```cpp
>     {
>         // todo:: make that an out of bounds trap
>         m_writer.EmptyAsm(Js::OpCodeAsmJs::Unreachable_Void);
> ```

We should add `Js::OpCodeAsmJs::MemOutOfBoundsAccess` and emit this instead.

---

*[lib/WasmReader/WasmByteCodeGenerator.cpp, line 1146 at r5](https://reviewable.io:443/reviews/microsoft/chakracore/1823#-KVCgCHY9jKT5ixb_BPg:-KVCgCHY9jKT5ixb_BPh:b8h5qh2) ([raw file](https://github.com/microsoft/chakracore/blob/e15956ad0da1db58e97d95d55155bdf31bae42fe/lib/WasmReader/WasmByteCodeGenerator.cpp#L1146)):*
> ```cpp
>     if (offset >= m_module->GetMemory()->maxSize)
>     {
>         throw WasmCompilationException(_u("Index is out of bounds"));
> ```

I don't think this is by spec. If the previous instruction have side effect we should still trap only when we reach this load/store.
Additionally, `maxSize` is a page count, thus if you want to add this, you need to multiply by `PAGE_SIZE`.
Be careful with overflow if you do multiply by page size however .

Put this check with `if (m_module->GetMemory()->minSize == 0)` above 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/microsoft/chakracore/1823#-:-KVCfr4ay9-tcjxoGWFq:b-90sp3d)*

Cellule:
Reviewed 3 of 3 files at r6.
Review status: all files reviewed at latest revision, 9 unresolved discussions, some commit checks failed.

---

*[lib/WasmReader/WasmByteCodeGenerator.cpp, line 1128 at r6](https://reviewable.io:443/reviews/microsoft/chakracore/1823#-KVChmf3ks1ZN-RBdUmT:-KVChmf3ks1ZN-RBdUmU:b-q3fyse) ([raw file](https://github.com/microsoft/chakracore/blob/89835b09b093b90eef1e5fa2fa991876c43551ae/lib/WasmReader/WasmByteCodeGenerator.cpp#L1128)):*
> ```cpp
>     if (m_module->GetMemory()->minSize == 0)
>     {
>         throw WasmCompilationException(_u("Index is out of bounds"));
> ```

This is wrong, the function is not invalid, but we need to trap accordingly if this happens.
This check is an optimization so we should emit an opcode that will trap instead.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/microsoft/chakracore/1823)*

lib/WasmReader/WasmByteCodeGenerator.cpp
```diff
@@ -1140,6 +1139,12 @@ WasmBytecodeGenerator::EmitMemAccess(WasmOp wasmOp, const WasmTypes::WasmType* s
     {
         throw WasmCompilationException(_u("Index expression must be of type I32"));
     }
+
+    if (offset >= m_module->GetMemory()->maxSize)
+    {
+        throw WasmCompilationException(_u("Index is out of bounds"));
```
MikeHolman: we shouldn't add this check. it isn't a syntax error

MikeHolman:
Review status: all files reviewed at latest revision, 10 unresolved discussions, some commit checks failed.

---

*[lib/Runtime/Language/InterpreterStackFrame.cpp, line 8365 at r4](https://reviewable.io:443/reviews/microsoft/chakracore/1823#-KVCW_nY-4pYhKUM8wvN:-KVCjIIpRhcg3FRkdHLU:b-494l9s) ([raw file](https://github.com/microsoft/chakracore/blob/e00165d45c8acebefed5e62899ccbc55d8b02bc3/lib/Runtime/Language/InterpreterStackFrame.cpp#L8365)):*
> @MikeHolman Do you think there is a better way to check this case.
> We need to trap if you are trying to access `mem[65533]` for a int32 with the max length being `65536`.
> However, I would really to find a way to avoid 2 checks.
> 
> The only thing I can think of is
> `if (index > arr->GetByteLength() - TypeToSizeMap[playout->ViewType])`
> Since we should trap on every memory access if there is no memory, then `arr->GetByteLength() > 0` in here.
> 
</details>
Should just be able to do check `if(index + TypeToSizeMap[playout->ViewType] > arr->GetByteLength())`, right?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/microsoft/chakracore/1823)*

lib/WasmReader/WasmByteCodeGenerator.cpp
meg-gupta: We are generating this bytecode sequence : 
    017d   Ld_IntConst        I3  int:-1
    0183   Add_Int              I2  I2  I3 
    0187   LdArrWasm        I2 = HEAP32[I2]

When index + offset overflow, by the time we execute LdArrWasm, the value would be wrapped around. And we cant trap. We might need a different sequence here.

Adding the above check allowed me to catch such cases during bytecode generation. 
Should we generate a different sequence instead ?

MikeHolman: yeah, we probably need to emit different code here. if index+offset overflow uint32, then we need to trap at runtime. this check doesn't prevent that from happening. For example, even with small offset you might have some large index that causes index+offset to overflow

---
In reply to: [85627647](https://github.com/Microsoft/ChakraCore/pull/1823#discussion_r85627647) [](ancestors = 85627647)

MikeHolman:
*[lib/Runtime/Language/InterpreterStackFrame.cpp, line 8365 at r4](https://reviewable.io:443/reviews/microsoft/chakracore/1823#-KVCW_nY-4pYhKUM8wvN:-KVCm0MfJ1nTnQV80BRo:b-pzh4if) ([raw file](https://github.com/microsoft/chakracore/blob/e00165d45c8acebefed5e62899ccbc55d8b02bc3/lib/Runtime/Language/InterpreterStackFrame.cpp#L8365)):*
> Should just be able to do check `if(index + TypeToSizeMap[playout->ViewType] > arr->GetByteLength())`, right?
</details>
Eh, maybe I'm wrong. I guess we need to consider case where that overflows uint32. (existing code doesn't work for that either though). as for length 0, we will have to check at runtime, as you can start with no memory or memory of size 0 and using grow memory or have a memory import. therefore, this needs to be a runtime check and cannot be assumed from time of bytecode gen

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/microsoft/chakracore/1823)*

lib/Backend/LowerMDShared.cpp
meg-gupta: I don't understand, src2 of which instruction ? I am reading it from the Js::ArrayBuffer::GetByteLengthOffset() right now.

MikeHolman: src2 of instr

meg-gupta: s38.i32         =  LdArrViewElem  [s3.var+s37.i32].u8 has no src2 ? And s3/s37 don't have the length of the array buffer ?

lib/WasmReader/WasmByteCodeGenerator.cpp
Cellule: We should have the check but emit an opcode that traps, that way it is a runtime trap not compile error, see my previous comment about this.

Cellule:
Review status: all files reviewed at latest revision, 10 unresolved discussions, some commit checks failed.

---

*[lib/Runtime/Language/InterpreterStackFrame.cpp, line 8365 at r4](https://reviewable.io:443/reviews/microsoft/chakracore/1823#-KVCW_nY-4pYhKUM8wvN:-KVDOIcRVmk-2F1C0fid:bqdzzht) ([raw file](https://github.com/microsoft/chakracore/blob/e00165d45c8acebefed5e62899ccbc55d8b02bc3/lib/Runtime/Language/InterpreterStackFrame.cpp#L8365)):*
> Eh, maybe I'm wrong. I guess we need to consider case where that overflows uint32. (existing code doesn't work for that either though). as for length 0, we will have to check at runtime, as you can start with no memory or memory of size 0 and using grow memory or have a memory import. therefore, this needs to be a runtime check and cannot be assumed from time of bytecode gen
</details>
yeah you're right for length 0.
the current check is correct since we check first if the index >= buffer length. We know the buffer cannot reach max uint32 (we would throw out of memory way before that) so it is safe to add a small value for the second check. 
I was just hoping to avoid having 2 checks.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/microsoft/chakracore/1823)*

lib/WasmReader/WasmByteCodeGenerator.cpp
MikeHolman: No, we can't know this at compile time. Offset+index might overflow where offset alone doesn't. Besides this doesn't work for import or grow memory

lib/Backend/IRBuilderAsmJs.cpp
```diff
@@ -3108,6 +3121,14 @@ IRBuilderAsmJs::BuildLong1Int1(Js::OpCodeAsmJs newOpcode, uint32 offset, Js::Reg
         srcOpnd = BuildSrcOpnd(srcRegSlot, TyUint32);
```
Cellule: > Js::OpCodeAsmJs::Conv_ITD & Js::OpCodeAsmJs::Conv_UTD

This looks like a bad copy paste on my part, could you remove it while you're there

lib/Backend/i386/LowererMDArch.cpp
```diff
@@ -960,6 +960,45 @@ LowererMDArch::LowerAsmJsCallI(IR::Instr * callInstr)
...
+        Lowerer::InsertBranch(Js::OpCode::JO, helperLabel, helperLabel);
+        // check low bits + size is within heapsize
+        lowererMD->m_lowerer->InsertCompareBranch(cmpOpnd, arrayLenOpnd, Js::OpCode::BrGe_A, true, helperLabel, helperLabel);
+    }
+
+    // MGTODO : call RuntimeError once implemented
+    lowererMD->m_lowerer->GenerateRuntimeError(loadLabel, JSERR_InvalidTypedArrayIndex, IR::HelperOp_RuntimeRangeError);
+    Lowerer::InsertBranch(Js::OpCode::Br, loadLabel, helperLabel);
+
+    addrOpnd->AsIndirOpnd()->SetIndexOpnd(indexPair.low->AsRegOpnd());
```
Cellule: Right now it is true that `indexPair.low` is a RegOpnd, but once we implement const folding for int64, the index could be a constant opnd.
Since we have unit tests hitting that path, just put an assert for now and it will be easier to fix once we get there.

Cellule:
LGTM after minor fixes and rebase on master

meg-gupta:
Done. Thanks @Cellule  and @MikeHolman.

PR #1824, opened 2016-10-25
Skip GC during Script Context close

digitalinfinity:
This change adds support to the Recycler to check for an additional heuristic while deciding whether to collect. Clients can indicate that a script context was closed, and therefore the other GC heuristics need to be overridden. This override happens only if the GC gets triggered at callrootlevel = 0- if not, we'll continue to use the existing heuristics.

### DIFF
--- a/lib/Common/Memory/Recycler.cpp
+++ b/lib/Common/Memory/Recycler.cpp
@@ -3346,54 +3346,65 @@ Recycler::CollectWithHeuristic()
 {
     // CollectHeuristic_Never flag should only be used with exhaustive candidate
     Assert((flags & CollectHeuristic_Never) == 0);
-
+    
+    BOOL isScriptContextCloseGCPending = FALSE; 
     const BOOL allocSize = flags & CollectHeuristic_AllocSize;
     const BOOL timedIfScriptActive = flags & CollectHeuristic_TimeIfScriptActive;
     const BOOL timedIfInScript = flags & CollectHeuristic_TimeIfInScript;
     const BOOL timed = (timedIfScriptActive && isScriptActive) || (timedIfInScript && isInScript) || (flags & CollectHeuristic_Time);
 
+    if ((flags & CollectOverride_CheckScriptContextClose) != 0)
+    {
+        isScriptContextCloseGCPending = this->collectionWrapper->GetIsScriptContextCloseGCPending();
+    }
+
+    // If there is a script context close GC pending, we need to do a GC regardless
+    // Otherwise, we should check the heuristics to see if a GC is necessary
+    if (!isScriptContextCloseGCPending)
+    {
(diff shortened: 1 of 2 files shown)


lib/Common/Memory/Recycler.cpp
```diff
@@ -3342,54 +3342,60 @@ Recycler::CollectWithHeuristic()
...
     const BOOL allocSize = flags & CollectHeuristic_AllocSize;
     const BOOL timedIfScriptActive = flags & CollectHeuristic_TimeIfScriptActive;
     const BOOL timedIfInScript = flags & CollectHeuristic_TimeIfInScript;
     const BOOL timed = (timedIfScriptActive && isScriptActive) || (timedIfInScript && isInScript) || (flags & CollectHeuristic_Time);
 
-#if ENABLE_PARTIAL_GC
-    if (GetPartialFlag<flags>())
+    // If there is a script context close GC pending, we need to do a GC regardless
+    // Otherwise, we should check the heuristics to see if a GC is necessary
+    if (!isScriptContextCloseGCPending)
```
agarwal-sandeep: >isScriptContextCloseGCPending [](start = 9, length = 29)

CollectWithHeuristic function gets indirectly called from CollectNow at many places and by default the value of isScriptContextCloseGCPending is true does that mean this if will be skipped most of the time?

digitalinfinity:
Reviving this PR- @agarwal-sandeep updated with your comments

digitalinfinity:

agarwal-sandeep:
:shipit:

curtisman:
>     SetupPostCollectionFlags<flags>();

Shouldn't we still clear the flag if we triggered a GC thru the heuristic and callRootLevel != 0

---
Refers to: lib/Common/Memory/Recycler.cpp:3423 in 8e4a74a. [](commit_id = 8e4a74a718295bb017ead75b188c24ef5076d68f, deletion_comment = False)

lib/Runtime/Base/ThreadContext.cpp
```diff
@@ -4105,6 +4106,20 @@ uint ThreadContext::GetRandomNumber()
 #endif
 }
 
+BOOL ThreadContext::GetAndClearIsScriptContextCloseGCPending()
+{
+    // Pretend that there is no GC pending if script code is still executing
+    // In that case, let the regular GC heuristics kick in
+    if (this->callRootLevel > 0)
```
curtisman: Instead of callRootLevel == 0 check. change it to add a CollectionFlags CollectOverride_CheckScriptcontextClose to CollectOnScriptExit to indicate whether we need to check this flag?

digitalinfinity:
@curtisman thanks for the review! Updated with your feedback

curtisman:
:shipit:

PR #1827, opened 2016-10-26
Fix for ArrayIterator object.

akroshg:
typedarray.keys/values/entries should iterate over its internal length slot, instead of doing get('length'). Fixed that.
Also we should not be calling GetOwnItem (as per the spec). Fixed that.

### DIFF
--- a/lib/Runtime/Library/JavascriptArray.cpp
+++ b/lib/Runtime/Library/JavascriptArray.cpp
@@ -8118,6 +8118,9 @@ Case0:
             JavascriptError::ThrowTypeError(scriptContext, JSERR_This_NullOrUndefined, _u("Array.prototype.values"));
         }
 
+#if ENABLE_COPYONACCESS_ARRAY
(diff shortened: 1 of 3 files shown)


akroshg:
@boingoing @suwc please have a look

caitp:
why not leave opinions on https://github.com/tc39/ecma262/issues/713 about this? :)

suwc:
:shipit:

suwc:
LGTM

akroshg:
@caitp I didn't know there is an active discussion going over for this. I stumble upon on a spec gap wrt chakra code and fixed that. I'll hold this one till @bterlson responds.

bterlson:
@caitp @akroshg I support calling ValidateTypedArray in %ArrayIteratorPrototype%.next (so it throws on detached buffers). I do not support calling [[get]] of length. I think this change is fine, but do you want to implement the ValidateTypedArray call as proposed in tc39/ecma262#713 as well as part of this PR?

akroshg:
@bterlson yes, call to ValidateTypedArray before getting length would be fairly easy to add. Will do that in this PR.

PR #1828, opened 2016-10-26
reduce native code data memory usage for in-proc jit

leirocks:
In OOP JIT when allocating native code data we record some information about the allocation. At the end of a JIT work item, the allocation information is used to make the pointer fixing up table which then used to fixup the internal pointers after passing native code data to runtime process. After a JIT work item is done, all the native code data including the allocation information are freed in JIT server, only the fixup table and aggregated data buffer is passed to runtime process. The fixup table is freed right after doing fixup in runntime process after the codegen call returned.

When doing in-proc jit we don't need these fixup information, and we don't do pointer fixing up and data aggregating/copying, the the fixup information has same life time as the native code data themselves, thus wasting memory.

This change bring back the same data structure which before the OOP JIT change and use it for in-proc JIT

Note in OOP JIT it uses one pointer size less memory than in-proc JIT because even the next pointer on DataChunk is not kept after data aggregating.

### DIFF
--- a/lib/Backend/BackendApi.cpp
+++ b/lib/Backend/BackendApi.cpp
@@ -195,7 +195,7 @@ SetProfilerFromNativeCodeGen(NativeCodeGenerator * toNativeCodeGen, NativeCodeGe
 }
 #endif
 
-void DeleteNativeCodeData(NativeCodeData * data)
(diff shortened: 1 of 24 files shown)


leirocks:

leirocks:

leirocks:
replaced with #1829

PR #1830, opened 2016-10-26
Jsrt: new string API

obastemur:
- JsCreateString
- JsCreateStringUtf8
- JsCreateStringUtf16
- JsCopyString
- JsCopyStringUtf8
- JsCopyStringUtf16
- JsParse
- JsRun
- JsSerialize
- JsParseSerialized
- JsRunSerialized
- JsCreatePropertyIdUtf8
- JsCopyPropertyIdUtf8

New: `ExternalArrayBuffer` can be used as script source.
In case the ExternalArrayBuffer source is Utf8, CC uses it as is.
no additional copy/encode/decode..

p.s. [ Create/Copy String credits goes to @jianchun ]


bin/ch/ChakraRtInterface.h
```diff
@@ -25,7 +25,6 @@ struct JsAPIHooks
     typedef JsErrorCode (WINAPI *JsrtConvertValueToBooleanPtr)(JsValueRef value, JsValueRef *booleanValue);
     typedef JsErrorCode (WINAPI *JsrtStringToPointerUtf8CopyPtr)(JsValueRef value, char **stringValue, size_t *length);
```
jianchun: > typedef JsErrorCode (WINAPI *JsrtStringToPointerUtf8CopyPtr)(JsValueRef value, char **stringValue, size_t *length); [](start = 3, length = 116)

JsrtStringToPointerUtf8CopyPtr not needed?

bin/ch/stdafx.h
```diff
@@ -152,32 +152,113 @@ do { \
...
+        }
+        else
+        {
+            strValue = value;
+        }
+        if (errorCode == JsNoError)
         {
-            return;
+            size_t len = 0;
+            errorCode = ChakraRTInterface::JsWriteStringUtf8(strValue, nullptr, 0, &len);
```
jianchun: This is fine, though we have JsWriteStringUtf16. In case the user instantiate this class only to GetWideString(), should we consider delay reading string content and also exercise JsWriteStringUtf16? May be a good chance to test out that API.

lib/Jsrt/ChakraCommon.h
```diff
@@ -500,7 +500,12 @@ typedef UINT32 DWORD;
         /// <summary>
         ///     Specified script is internal and non-user code. Hidden from debugger
         /// </summary>
-        JsParseScriptAttributeLibraryCode = 0x1
+        JsParseScriptAttributeLibraryCode = 0x1,
+        /// <summary>
+        ///     ChakraCore assumes ExternalArrayBuffer is Utf8 by default.
+        ///     This one needs to be set for Utf16
+        /// </summary>
+        JsParseScriptAttributeUtf16Encoding = 0x2,
```
jianchun: >JsParseScriptAttributeUtf16Encoding [](start = 8, length = 35)

The comment is clear, but users may possibly be confused by the name and think this flag always apply? Should we explicitly have "ArrayBuffer" in the flag name? Or not?

lib/Jsrt/ChakraCore.h
```diff
@@ -163,4 +163,346 @@ JsGetModuleHostInfo(
...
+///     This callback is only supported by the Win32 version of the API
+/// </summary>
+/// <param name="sourceContext">The context passed to Js[Parse|Run]SerializedScriptWithCallback</param>
+/// <param name="script">The script returned.</param>
+/// <returns>
+///     true if the operation succeeded, false otherwise.
+/// </returns>
+typedef bool (CHAKRA_CALLBACK * JsSerializedLoadScriptCallback)
+    (JsSourceContext sourceContext, JsValueRef *value,
+    JsParseScriptAttributes *parseAttributes);
```
jianchun: Missing SAL annotations

lib/Jsrt/ChakraCore.h
```diff
@@ -163,4 +163,346 @@ JsGetModuleHostInfo(
...
+/// <summary>
+///     Create String Variable from C String
+/// </summary>
+/// <remarks>
+///     <para>
+///     Experimental. We may update the name or behavior until it is stable.
+///     </para>
+/// </remarks>
+/// <param name="content">Pointer to string memory.</param>
+/// <param name="length">Length of the string.</param>
```
jianchun: Please explain more details of "length". Does it include null terminator? Is it required or optional.

lib/Jsrt/ChakraCore.h
```diff
@@ -163,4 +163,346 @@ JsGetModuleHostInfo(
...
+///     The code <c>JsNoError</c> if the operation succeeded, a failure code otherwise.
+/// </returns>
+CHAKRA_API
+    JsCreateStringUtf16(
+        const uint16_t *content,
+        size_t length,
+        _Out_ JsValueRef *value);
+
+/// <summary>
+///     Write string value into C string buffer
```
jianchun: >Write string value into C string buffer [](start = 8, length = 39)

Please describe more details of this API. What if start/length out of range. What if buffer is null. What assumption we have on buffer size.

lib/Jsrt/Jsrt.cpp
```diff
@@ -1439,7 +1399,7 @@ CHAKRA_API JsDefineProperty(_In_ JsValueRef object, _In_ JsPropertyIdRef propert
 
 CHAKRA_API JsCreateArray(_In_ unsigned int length, _Out_ JsValueRef *result)
 {
-    return ContextAPIWrapper<true>([&] (Js::ScriptContext *scriptContext) -> JsErrorCode {
+    return ContextAPINoScriptWrapper([&](Js::ScriptContext *scriptContext) -> JsErrorCode {
```
jianchun: >ContextAPINoScriptWrapper([&](J [](start = 11, length = 31)

Some changes unrelated to this PR? Need to evaluate carefully if these are safe.

lib/Jsrt/JsrtSourceHolder.h
```diff
@@ -17,6 +17,9 @@ namespace Js
         TUnloadCallback scriptUnloadCallback;
         JsSourceContext sourceContext;
 
+#ifndef NTBUILD
```
jianchun: Will this be build break on Full? Maybe ok to left out this `#ifndef`.

lib/Jsrt/Jsrt.cpp
obastemur: No, it is not unrelated. JsScriptHost onload event may have called when the script execution is active. Updated these 2 similar to String interface.

bin/ch/ch.cpp
```diff
@@ -86,16 +87,22 @@ HRESULT GetSerializedBuffer(LPCSTR fileContents, __out BYTE **byteCodeBuffer, __
 
     unsigned int bcBufferSize = 0;
     unsigned int newBcBufferSize = 0;
-    IfJsErrorFailLog(ChakraRTInterface::JsSerializeScriptUtf8(fileContents, bcBuffer, &bcBufferSize));
-    // Above call will return the size of the buffer only, once succeed we need to allocate memory of that much and call it again.
+    JsValueRef scriptSource;
+    ChakraRTInterface::JsCreateExternalArrayBuffer((void*)fileContents, (unsigned int)strlen(fileContents),
```
agarwal-sandeep: >strlen [](start = 86, length = 6)

This means we won't support null character in file when serializing buffers?

lib/Jsrt/ChakraCommon.h
agarwal-sandeep: Being explicit is better,  though that will make this a long name but we should make it JsParseScriptAttributeArrayBufferIsUtf16Encoded

---
In reply to: [85450903](https://github.com/Microsoft/ChakraCore/pull/1830#discussion_r85450903) [](ancestors = 85450903)

bin/ch/Debugger.cpp
```diff
@@ -175,10 +175,9 @@ JsValueRef Debugger::Evaluate(JsValueRef callee, bool isConstructCall, JsValueRe
     {
         IfJsErrorFailLogAndRet(ChakraRTInterface::JsNumberToInt(arguments[1], &stackFrameIndex));
 
-        AutoString argstr;
-        size_t length;
-        IfJsErrorFailLogAndRet(ChakraRTInterface::JsValueToCharCopy(arguments[2], &argstr, &length));
-        ChakraRTInterface::JsDiagEvaluateUtf8(*argstr, stackFrameIndex, &result);
+        AutoString argstr(arguments[2]);
+        IfJsErrorFailLogAndRet(argstr.GetError());
+        ChakraRTInterface::JsDiagEvaluateUtf8(argstr.GetString(), stackFrameIndex, &result);
```
agarwal-sandeep: >JsDiagEvaluateUtf8 [](start = 27, length = 18)

Shouldn't we replace this Utf8 version API with JsValueRef api as well?

lib/Jsrt/ChakraCore.h
```diff
@@ -163,4 +163,346 @@ JsGetModuleHostInfo(
     _In_ JsModuleHostInfoKind moduleHostInfo,
     _Outptr_result_maybenull_ void** hostInfo);
 
+/// <summary>
+///     Called by the runtime to load the source code of the serialized script.
+///     The caller must keep the script buffer valid until the JsSerializedScriptUnloadCallback.
+///     This callback is only supported by the Win32 version of the API
```
agarwal-sandeep: >This callback is only supported by the Win32 version of the API [](start = 8, length = 63)

What does it mean by Win32 version of the API? we only have one version of JsCompileSerialized/JsRunSerialized.

lib/Jsrt/ChakraCore.h
```diff
@@ -163,4 +163,346 @@ JsGetModuleHostInfo(
...
+///     Create String Variable from C String
+/// </summary>
+/// <remarks>
+///     <para>
+///     Experimental. We may update the name or behavior until it is stable.
+///     </para>
+/// </remarks>
+/// <param name="content">Pointer to string memory.</param>
+/// <param name="length">Length of the string.</param>
+/// <param name="value">.</param>
```
agarwal-sandeep: >value [](start = 17, length = 5)

Please explain the value will contain the internal representation of string which can be used in other APIs like ....

lib/Runtime/Library/WasmLibrary.cpp
```diff
@@ -523,7 +523,12 @@ namespace Js
             *ppSourceInfo = nullptr;
             Wasm::WasmModule * wasmModule = nullptr;
 
-            *ppSourceInfo = Utf8SourceInfo::New(scriptContext, (LPCUTF8)script, lengthBytes / sizeof(char16), lengthBytes, pSrcInfo, false);
+            *ppSourceInfo = Utf8SourceInfo::New(scriptContext, (LPCUTF8)script, lengthBytes / sizeof(char16), lengthBytes, pSrcInfo, false
+#ifndef NTBUILD
+                            , nullptr);
+#else
+                            );
+#endif
```
jianchun: nit: The same change in GlobalObject.cpp looks better than this one.

lib/Jsrt/ChakraCore.h
```diff
@@ -163,4 +163,346 @@ JsGetModuleHostInfo(
...
+/// <param name="buffer">Pointer to buffer.</param>
+/// <param name="bufferSize">Buffer size.</param>
+/// <param name="written">Total number of characters written</param>
+/// <returns>
+///     The code <c>JsNoError</c> if the operation succeeded, a failure code otherwise.
+/// </returns>
+CHAKRA_API
+    JsWriteStringUtf8(
+        JsValueRef value,
+        uint8_t* buffer,
```
agarwal-sandeep: >uint8_t* buffer, [](start = 8, length = 16)

Which this api doesn't take start and length?

lib/Runtime/Library/WasmLibrary.cpp
jianchun: Similarly, TTSanpValues.cpp and TTActionEvents.cpp can use GlobalObject.cpp style.

---
In reply to: [85626412](https://github.com/Microsoft/ChakraCore/pull/1830#discussion_r85626412) [](ancestors = 85626412)

lib/Jsrt/ChakraCore.h
```diff
@@ -163,4 +163,346 @@ JsGetModuleHostInfo(
...
+///     Parses a serialized script and returns a function representing the script.
+///     Provides the ability to lazy load the script source only if/when it is needed.
+/// </summary>
+/// <remarks>
+///     <para>
+///     Requires an active script context.
+///     </para>
+///     <para>
+///     The runtime will hold on to the buffer until all instances of any functions created from
+///     the buffer are garbage collected.  It will then call scriptUnloadCallback to inform the
```
agarwal-sandeep: >scriptUnloadCallback [](start = 61, length = 20)

Where is this set?

lib/Jsrt/ChakraCore.h
obastemur: This by design draft. @jianchun?

lib/Jsrt/Jsrt.cpp
```diff
@@ -4266,3 +4161,431 @@ CHAKRA_API JsTTDReplayExecution(_Inout_ JsTTDMoveMode* moveMode, _Inout_ int64_t
...
+    size_t strLength = 0;
+    JsErrorCode errorCode = JsStringToPointer(value, &str, &strLength);
+    if (errorCode != JsNoError)
+    {
+        return errorCode;
+    }
+
+    if (start < 0 || (size_t)start > strLength)
+    {
+        return JsNoError;  // start out of range, no chars written
```
agarwal-sandeep: >JsNoError [](start = 15, length = 9)

Should such errors be JsErrorInvalidArgument instead of JsNoError

lib/Jsrt/ChakraCore.h
jianchun: Yes, I hate it too, these 3 APIs are indeed inconsistent but required to implement compatible nodecc chakrashim. You can examine their usage in nodecc/xplat. Not sure if we can find a way to make them consistent and satisfy chakrashim.

lib/Jsrt/Jsrt.cpp
```diff
@@ -4266,3 +4161,431 @@ CHAKRA_API JsTTDReplayExecution(_Inout_ JsTTDMoveMode* moveMode, _Inout_ int64_t
...
+    {
+        scriptFlag = (LoadScriptFlag) (LoadScriptFlag_Utf8Source);
+    }
+    else
+    {
+        scriptFlag = LoadScriptFlag_None;
+    }
+
+    return JsSerializeScriptCore(script, cb, scriptFlag, nullptr,
+        0, buffer, bufferSize, scriptVal);
```
agarwal-sandeep: Some of JsParseScriptAttributes are not transferred to LoadScriptFlag like LbraryCode.

lib/Jsrt/JsrtSourceHolder.cpp
```diff
@@ -118,20 +235,61 @@ namespace Js
...
+        ThreadContext *threadContext = ThreadContext::GetContextForCurrentThread();
+        bool isActive = threadContext->IsScriptActive();
+        // check if we are still in script land
+        if (isActive)
+        {
+            threadContext->SetIsScriptActive(false);
+            threadContext->GetRecycler()->SetIsScriptActive(false);
+        }
+
+        if (!scriptLoadCallback(sourceContext, &scriptVal, &attributes))
```
agarwal-sandeep: Should scriptLoadCallback be guarded by normal BEGIN_LEAVE_SCRIPT macro because scriptLoadCallback is external function and it can call any other JsRT api before returning

lib/Runtime/Base/Utf8SourceInfo.h
```diff
@@ -389,7 +406,15 @@ namespace Js
         ULONG parseFlags;
         ULONG byteCodeGenerationFlags;
 
-        Utf8SourceInfo(ISourceHolder *sourceHolder, int32 cchLength, SRCINFO const* srcInfo, DWORD_PTR secondaryHostSourceContext, ScriptContext* scriptContext, bool isLibraryCode);
+        Utf8SourceInfo(ISourceHolder *sourceHolder, int32 cchLength, SRCINFO const* srcInfo,
+            DWORD_PTR secondaryHostSourceContext, ScriptContext* scriptContext, bool isLibraryCode
+#ifdef NTBUILD
+            );
+#else
+            , Js::Var scriptSource);
```
agarwal-sandeep: So many #ifdef doesn't look good, we should make it a required parameter. 

---
In reply to: [85627248](https://github.com/Microsoft/ChakraCore/pull/1830#discussion_r85627248) [](ancestors = 85627248)

lib/Jsrt/JsrtSourceHolder.cpp
obastemur: Makes sense. One question; We have other callbacks and they don't follow this pattern. Any idea why?

bin/ch/stdafx.h
obastemur: We have test/native-test for this purpose. Prefer a sample there

bin/ch/Debugger.cpp
obastemur: Good idea but we may need ParseScriptArguments there too.

lib/Jsrt/Jsrt.cpp
obastemur: Agree. @jianchun, FYI (for chakrashim)

lib/Jsrt/JsrtSourceHolder.cpp
agarwal-sandeep: PromiseContinuationCallback follows the same pattern, see JavascriptLibrary::EnqueueTask. I think it depends from where the callback is called. We may have missed some places where it should be guarded by begin/end

lib/Jsrt/ChakraCommon.h
obastemur: Okay two vs one. Updating to `JsParseScriptAttributeArrayBufferIsUtf16Encoded`

bin/ch/stdafx.h
obastemur: see new `native-test` `test-char16`.

/cc @liminzhu . We need to update Chakra-Samples repo parallel to this PR.

lib/Jsrt/ChakraCore.h
```diff
@@ -163,4 +163,346 @@ JsGetModuleHostInfo(
...
+/// <param name="parseAttributes">Encoding for the script.</param>
+/// <returns>
+///     The code <c>JsNoError</c> if the operation succeeded, a failure code otherwise.
+/// </returns>
+CHAKRA_API
+    JsSerialize(
+        JsValueRef script,
+        BYTE *buffer,
+        unsigned int *bufferSize,
+        JsParseScriptAttributes parseAttributes);
```
liminzhu: Suggest putting the parseAttributesinput before the outputs (buffer and buffersize)

lib/Jsrt/ChakraCore.h
```diff
@@ -163,4 +163,346 @@ JsGetModuleHostInfo(
...
+/// </param>
+/// <param name="sourceUrl">The location the script came from.</param>
+/// <param name="result">A function representing the script code.</param>
+/// <returns>
+///     The code <c>JsNoError</c> if the operation succeeded, a failure code otherwise.
+/// </returns>
+CHAKRA_API
+    JsCompileSerialized(
+        BYTE *buffer,
+        JsSerializedLoadScriptCallback scriptLoadCallback,
```
liminzhu: Nitpick - put scriptLoadCallback as the first arg similar as [JsParseSerializedScriptWithCallback](https://github.com/Microsoft/ChakraCore/wiki/JsParseSerializedScriptWithCallback)

lib/Jsrt/ChakraCore.h
```diff
@@ -163,4 +163,346 @@ JsGetModuleHostInfo(
...
+        uint16_t* buffer,
+        _Out_opt_ size_t* written);
+
+/// <summary>
+///     Parses a script and returns a function representing the script.
+/// </summary>
+/// <remarks>
+///     Requires an active script context.
+/// </remarks>
+/// <param name="script">The script to run.</param>
```
liminzhu: May want to add something like 'may accept string or external arraybuffer' to `JsValueRef script` args here and elsewhere for clarity.

lib/Jsrt/ChakraCore.h
```diff
@@ -163,4 +163,346 @@ JsGetModuleHostInfo(
...
+///     <para>
+///     Consumer is responsible from calling JsStringFree to free allocated memory.
+///
+///     Experimental. We may update the name or behavior until it is stable.
+///     </para>
+/// </remarks>
+/// <param name="propertyId">The property ID to get the name of.</param>
+/// <param name="buffer">The name associated with the property ID, encoded as utf8.</param>
+/// <param name="bufferSize">Size of the buffer.</param>
+/// <param name="written">Total number of characters written or to be written</param>
```
liminzhu: What `written`? Should check if there's incorrect comment elsewhere.

lib/Jsrt/ChakraCore.h
```diff
@@ -163,4 +163,346 @@ JsGetModuleHostInfo(
...
+///     Requires an active script context.
+///     </para>
+///     <para>
+///     Consumer is responsible from calling JsStringFree to free allocated memory.
+///
+///     Experimental. We may update the name or behavior until it is stable.
+///     </para>
+/// </remarks>
+/// <param name="propertyId">The property ID to get the name of.</param>
+/// <param name="buffer">The name associated with the property ID, encoded as utf8.</param>
```
liminzhu: The buffer holding the name associated with the property ID, encoded as utf8

liminzhu:
Late to the design party, but the design overall looks great to me. May I suggest,

- Rename a few new APIs so that the resemblance with the old ones are clear? 

// this way new APIs names except propertyId ones are just old names dropping _script_ in the end
`JsCompile` -> `JsParse`
`JsCompileSerialized` -> `JsParseSerialized`

// IIRC these are utf8 versions of existing APIs?
`JsCreatePropertyIdUtf8` -> `JsGetPropertyIdFromNameUtf8`
`JsWritePropertyIdUtf8` -> `JsGetPropertyNameFromIdUtf8`

- What happened to SAL in the new APIs? Do they not work on xplat? 

- Some inline nitpicks above.  

Once the PR is merged, I will update samples we have.

lib/Jsrt/ChakraCore.h
```diff
@@ -163,4 +163,346 @@ JsGetModuleHostInfo(
...
+
+/// <summary>
+///     Gets the name associated with the property ID.
+/// </summary>
+/// <remarks>
+///     <para>
+///     Requires an active script context.
+///     </para>
+///     <para>
+///     Consumer is responsible from calling JsStringFree to free allocated memory.
```
liminzhu: Is there still `JsStringFree`? Correct me if I'm wrong, I thought ChakraCore copies propertyid string and manages its lifetime.

lib/Jsrt/ChakraCore.h
obastemur: Indeed but how about the ones we enable to Windows?

lib/Jsrt/JsrtSourceHolder.h
obastemur: simplified other `ifdefs`  this one remains to keep unused variable from NTBUILD

lib/Jsrt/Jsrt.cpp
obastemur: `JsSerializeScriptCore` doesn't take JsParseScriptAttributes

lib/Jsrt/ChakraCore.h
obastemur: This is by design template

lib/Jsrt/ChakraCore.h
obastemur: same. this is by design template.

lib/Jsrt/ChakraCore.h
```diff
@@ -163,4 +163,346 @@ JsGetModuleHostInfo(
...
+/// </param>
+/// <param name="sourceUrl">The location the script came from.</param>
+/// <param name="result">
+///     The result of running the script, if any. This parameter can be null.
+/// </param>
+/// <returns>
+///     The code <c>JsNoError</c> if the operation succeeded, a failure code otherwise.
+/// </returns>
+CHAKRA_API
+    JsRunSerialized(
```
obastemur: same. this is by design template.

lib/Jsrt/ChakraCore.h
```diff
@@ -163,4 +163,346 @@ JsGetModuleHostInfo(
...
+CHAKRA_API
+    JsWriteStringUtf16(
+        JsValueRef value,
+        int start,
+        int length,
+        uint16_t* buffer,
+        _Out_opt_ size_t* written);
+
+/// <summary>
+///     Parses a script and returns a function representing the script.
```
obastemur: see remarks

lib/Jsrt/Jsrt.cpp
agarwal-sandeep: Yes, but if user passes librarycode flag in parseAttributes to JsSerialize, he expects it to be treated as library code but it won't so there is inconsistency between JsSerialize and JsSerializeScriptCore, which should be either fixed or JsSerialize should mention that it only owners JsParseScriptAttributeUtf16Encoding

lib/Jsrt/JsrtSourceHolder.h
jianchun: This will cause build error on Full. See line 51.

---
In reply to: [85743215](https://github.com/Microsoft/ChakraCore/pull/1830#discussion_r85743215) [](ancestors = 85743215)

lib/Jsrt/ChakraCore.h
```diff
@@ -163,4 +163,394 @@ JsGetModuleHostInfo(
...
+
+/// <summary>
+///     Write JavascriptString value into C string buffer
+/// </summary>
+/// <remarks>
+///     <para>
+///         When size of the `buffer` is unknown,
+///         `buffer` argument can be nullptr.
+///         `written` argument will return the size needed.
+///     </para>
```
jianchun: Documentation of JsWriteString/JsWriteStringUtf16 are incorrect (do not match code). These 2 do not return "needed buffer size".

lib/Jsrt/JsrtSourceHolder.h
```diff
@@ -45,6 +48,7 @@ namespace Js
             scriptLoadCallback(scriptLoadCallback),
             scriptUnloadCallback(scriptUnloadCallback),
             sourceContext(sourceContext),
+            mappedScriptValue(nullptr),
```
jianchun: (add a new comment instead) This line will build break on Full as field not available on NTBUILD.

lib/Jsrt/ChakraCore.h
obastemur: See updated code. They actually return needed buffer size in bytes now.

bin/ch/Debugger.cpp
obastemur: @agarwal-sandeep This method; we use on CH temporarily and one of the methods we have left from the experimental phase. While we are enabling Debug on xplat, we should take care of this part too though.

lib/Jsrt/ChakraCommon.h
```diff
@@ -500,7 +500,12 @@ typedef UINT32 DWORD;
         /// <summary>
         ///     Specified script is internal and non-user code. Hidden from debugger
         /// </summary>
-        JsParseScriptAttributeLibraryCode = 0x1
+        JsParseScriptAttributeLibraryCode = 0x1,
+        /// <summary>
+        ///     ChakraCore assumes ExternalArrayBuffer is Utf8 by default.
+        ///     This one needs to be set for Utf16
```
jianchun: nit: description may need to mention parse script context, otherwise reading by itself is confusing (why ExternalArrayBuffer has anything to do with utf8).

lib/Jsrt/ChakraCore.h
```diff
@@ -163,4 +163,396 @@ JsGetModuleHostInfo(
     _In_ JsModuleHostInfoKind moduleHostInfo,
     _Outptr_result_maybenull_ void** hostInfo);
 
+#ifndef NTBUILD
+/// <summary>
+///     Called by the runtime to load the source code of the serialized script.
+///     The caller must keep the script buffer valid until the JsSerializedScriptUnloadCallback.
+///     This callback is only supported by the Win32 version of the API
+/// </summary>
+/// <param name="sourceContext">The context passed to Js[Parse|Run]SerializedScriptWithCallback</param>
```
jianchun: Comment outdated (passed to "API names" changed)

lib/Jsrt/ChakraCore.h
```diff
@@ -163,4 +163,396 @@ JsGetModuleHostInfo(
...
+/// <summary>
+///     Create JavascriptString variable from C string
+/// </summary>
+/// <remarks>
+///     <para>
+///         C string is expected to be ASCII
+///     </para>
+/// </remarks>
+/// <param name="content">Pointer to string memory.</param>
+/// <param name="length">Number of bytes within the string</param>
```
jianchun: Consider "number of characters"?

lib/Jsrt/ChakraCore.h
```diff
@@ -163,4 +163,396 @@ JsGetModuleHostInfo(
...
+/// <summary>
+///     Create JavascriptString variable from Utf16 string
+/// </summary>
+/// <remarks>
+///     <para>
+///         Expects Utf16 string
+///     </para>
+/// </remarks>
+/// <param name="content">Pointer to string memory.</param>
+/// <param name="length">Number of bytes within the string</param>
```
obastemur: It is number of bytes not chars.

lib/Jsrt/ChakraCore.h
obastemur: This is also number of bytes and actually same thing for this one

jianchun:
LGTM. Left some comments on API descriptions. Maybe @liminzhu could go through "ChakraCore.h" and help with the API docs?

lib/Jsrt/ChakraCore.h
jianchun: No, I think it is number of characters.

obastemur:
Did some extra cleanup to old comments. 

> Maybe @liminzhu could go through "ChakraCore.h" and help with the API docs?

We may better decide on remaining types etc. part first (before updating anything again). Changing design again and again on the PR is a huge pain.

lib/Jsrt/ChakraCore.h
obastemur: On the email it shows `JsCreateStringUtf8` on top... followed the line from here and what you mean here is Utf16... Yes for Utf16, it is the length of string..

obastemur:
@jianchun @agarwal-sandeep @liminzhu Thanks for the review and discussion.

obastemur:

kphillisjr:
This pull request is incomplete. The ChakraCore Samples needs updated. See this Issue: https://github.com/Microsoft/Chakra-Samples/issues/47

obastemur:
@kphillisjr https://github.com/Microsoft/Chakra-Samples/pull/48

PR #1831, opened 2016-10-26
Add fast path for x86 to fix inconsistent issue

ThomsonTan:
Math.pow uses inline assembly __libm_sse2_pow instead of pow from CRT, which makes it not handled the same way as other architectures where SSE2 instructions have already been used in pow from CRT. This change adds fast path for Math.pow call from interpreter to make it consistent with JIT code which goes to fast path already.


ThomsonTan:
@rajatd please help review.

rajatd:
:shipit:

PR #1835, opened 2016-10-27
Wasm Globals offset bugix

arunetm:
Fixing crash caused by globals being emitted to incorrect slots


Cellule:
I fixed that issue in #1806 with proper alignment calculation for globals

arunetm:
@Cellule, ah thanks. We need this fix to test global initexprs for elements. 
will close this PR and pull in your fix.

Cellule:
It should be merged in today

PR #1838, opened 2016-10-27
Augmenting .apply inlining

rajatd:
This PR augments apply inlining when only 1 argument is passed to the apply call. We can now inline the call to .apply and the apply target if it is a script function. Inlining a built-in target will come in a later PR

---

This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/1838)

### DIFF
--- a/lib/Backend/Inline.cpp
+++ b/lib/Backend/Inline.cpp
@@ -1335,7 +1335,7 @@ Inline::TryOptimizeCallInstrWithFixedMethod(IR::Instr *callInstr, const Function
         JITTimeFunctionBody * calleeFunctionBody = inlineeInfo != nullptr && inlineeInfo->HasBody() ? inlineeInfo->GetBody() : nullptr;
         const char16* calleeName = calleeFunctionBody != nullptr ? calleeFunctionBody->GetDisplayName() : _u("<unknown>");
 
-        Output::Print(_u("FixedFields: function %s (%s): considering method <unknown> (%s %s): polymorphic = %d, built-in = %d, ctor = %d, inlined = %d, functionInfo = %p.\n"),
+        Output::Print(_u("FixedFields: function %s (%s): considering method (%s %s): polymorphic = %d, built-in = %d, ctor = %d, inlined = %d, functionInfo = %p.\n"),
             callInstr->m_func->GetJITFunctionBody()->GetDisplayName(), callInstr->m_func->GetDebugNumberSet(debugStringBuffer), calleeName,
             calleeFunctionBody ? inlineeInfo->GetDebugNumberSet(debugStringBuffer2) : _u("(null)"),
             isPolymorphic, isBuiltIn, isCtor, isInlined, inlineeInfo ? inlineeInfo->GetFunctionInfoAddr() : 0);
@@ -1357,7 +1357,7 @@ Inline::TryOptimizeCallInstrWithFixedMethod(IR::Instr *callInstr, const Function
             JITTimeFunctionBody* calleeFunctionBody = inlineeInfo != nullptr && inlineeInfo->HasBody() ? inlineeInfo->GetBody() : nullptr;
             const char16* calleeName = calleeFunctionBody != nullptr ? calleeFunctionBody->GetDisplayName() : _u("<unknown>");
 
-            Output::Print(_u("FixedFields: function %s (%s): %s non-fixed method <unknown> (%s %s), because callee is not single def.\n"),
+            Output::Print(_u("FixedFields: function %s (%s): %s non-fixed method (%s %s), because callee is not single def.\n"),
                 callInstr->m_func->GetJITFunctionBody()->GetDisplayName(), callInstr->m_func->GetDebugNumberSet(debugStringBuffer),
                 inlineeInfo != nullptr ? _u("inlining") : _u("calling"), calleeName,
                 calleeFunctionBody ? inlineeInfo->GetDebugNumberSet(debugStringBuffer2) : _u("(null)"));
@@ -1368,6 +1368,28 @@ Inline::TryOptimizeCallInstrWithFixedMethod(IR::Instr *callInstr, const Function
     }
 
     IR::Instr* ldMethodFldInstr = methodValueSym->GetInstrDef();
+    while (ldMethodFldInstr->m_opcode == Js::OpCode::BytecodeArgOutCapture)
+    {
+        StackSym* sym = ldMethodFldInstr->GetSrc1()->GetStackSym();
+        if (!sym->IsSingleDef())
+        {
+#if TRACE_FIXED_FIELDS
+            if (printFixedFieldsTrace)
+            {
+                JITTimeFunctionBody* calleeFunctionBody = inlineeInfo != nullptr && inlineeInfo->HasBody() ? inlineeInfo->GetBody() : nullptr;
+                const char16* calleeName = calleeFunctionBody != nullptr ? calleeFunctionBody->GetDisplayName() : _u("<unknown>");
+
+                Output::Print(_u("FixedFields: function %s (%s): %s non-fixed method (%s %s), because callee is not single def.\n"),
+                    callInstr->m_func->GetJITFunctionBody()->GetDisplayName(), callInstr->m_func->GetDebugNumberSet(debugStringBuffer),
+                    inlineeInfo != nullptr ? _u("inlining") : _u("calling"), calleeName,
+                    calleeFunctionBody ? inlineeInfo->GetDebugNumberSet(debugStringBuffer2) : _u("(null)"));
+                Output::Flush();
+            }
+#endif
+            return false;
+        }
+        ldMethodFldInstr = sym->GetInstrDef();
+    }
     if (ldMethodFldInstr->m_opcode != Js::OpCode::ScopedLdMethodFld
         && ldMethodFldInstr->m_opcode != Js::OpCode::LdRootMethodFld
         && ldMethodFldInstr->m_opcode != Js::OpCode::LdMethodFld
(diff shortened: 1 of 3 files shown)


rajatd:
@satheeshravi please take a look

lib/Backend/Inline.cpp
```diff
@@ -1368,6 +1368,28 @@ Inline::TryOptimizeCallInstrWithFixedMethod(IR::Instr *callInstr, const Function
...
+        StackSym* sym = ldMethodFldInstr->GetSrc1()->GetStackSym();
+        if (!sym->IsSingleDef())
+        {
+#if TRACE_FIXED_FIELDS
+            if (printFixedFieldsTrace)
+            {
+                JITTimeFunctionBody* calleeFunctionBody = inlineeInfo != nullptr && inlineeInfo->HasBody() ? inlineeInfo->GetBody() : nullptr;
+                const char16* calleeName = calleeFunctionBody != nullptr ? calleeFunctionBody->GetDisplayName() : _u("<unknown>");
+
+                Output::Print(_u("FixedFields: function %s (%s): %s non-fixed method <unknown> (%s %s), because callee is not single def.\n"),
```
satheeshravi: ><unknown> [](start = 85, length = 9)

pls remove this

satheeshravi:
>             Output::Print(_u("FixedFields: function %s (%s): %s non-fixed method <unknown> (%s %s), because callee is not single def.\n"),

same here

---
Refers to: lib/Backend/Inline.cpp:1360 in 6087d29. [](commit_id = 6087d29576158850a81a9f99895cc2e5c70ca81c, deletion_comment = False)

lib/Backend/Inline.cpp
```diff
@@ -1928,7 +1950,7 @@ Inline::InlineBuiltInFunction(IR::Instr *callInstr, const FunctionJITTimeInfo *
     IR::Instr *inlineBuiltInEndInstr = nullptr;
     if (inlineCallOpCode ==  Js::OpCode::InlineFunctionApply)
     {
-       inlineBuiltInEndInstr = InlineApply(callInstr, inlineeData, inlinerData, symCallerThis, pIsInlined, profileId, recursiveInlineDepth);
+       inlineBuiltInEndInstr = InlineApply(callInstr, inlineeData, inlinerData, symCallerThis, pIsInlined, profileId, recursiveInlineDepth, inlineCallArgCount - 1);
```
satheeshravi: >inlineCallArgCount [](start = 140, length = 18)

why is 1 less here ? 
can inlineCallArgCount be 0 ? - as indicated by the assert above. If so, the caller takes an uint and will result in incorrect functionality ?

lib/Backend/Inline.cpp
```diff
@@ -2294,16 +2321,32 @@ IR::Instr* Inline::InlineApply(IR::Instr *callInstr, const FunctionJITTimeInfo *
...
             *pIsInlined = true;
+
+#if defined(ENABLE_DEBUG_CONFIG_OPTIONS)
+            TraceInlining(inlinerData, Js::JavascriptLibrary::GetNameForBuiltIn(builtInId),
+                nullptr, 0, this->topFunc->GetWorkItem()->GetJITTimeInfo(), 0, nullptr, callSiteId, callInstr->m_func->GetTopFunc()->IsLoopBody(), builtInId);
+#endif
+
             // TODO: OOP JIT enable assert (readprocessmemory?)
             //Assert((inlineeData->GetFunctionInfo()->GetAttributes() & Js::FunctionInfo::Attributes::BuiltInInlinableAsLdFldInlinee) != 0);
             return InlineApplyWithArray(callInstr, applyData, Js::JavascriptLibrary::GetBuiltInForFuncInfo(inlineeData->GetFunctionInfoAddr(), this->topFunc->GetThreadContextInfo()));
```
satheeshravi: >InlineApplyWithArray [](start = 19, length = 20)

nit: can you be consistent with the naming like InlineApplyWithArrayArgument

lib/Backend/Inline.cpp
```diff
@@ -2506,8 +2537,97 @@ IR::Instr * Inline::InlineApplyWithArray(IR::Instr * callInstr, const FunctionJI
...
+    if (TryOptimizeCallInstrWithFixedMethod(callInstr, applyTargetInfo, false /*isPolymorphic*/, false /*isBuiltIn*/, false /*isCtor*/, true /*isInlined*/, safeThis /*unused here*/))
+    {
+        Assert(callInstr->m_opcode == Js::OpCode::CallIFixed);
+        callInstr->InsertBefore(IR::ByteCodeUsesInstr::New(callInstr, originalCallTargetStackSym->m_id));
+    }
+
+    return callInstr;
+}
+
+void Inline::GetArgInstrsForCallAndApply(IR::Instr* callInstr, IR::Instr** implicitThisArgOut, IR::Instr** explicitThisArgOut, IR::Instr** argumentsOrArrayArgOut, uint &argOutCount)
```
satheeshravi: >& [](start = 168, length = 1)

shld this just be a copy and not a ref?

lib/Backend/Inline.cpp
```diff
@@ -2385,12 +2425,12 @@ IR::Instr * Inline::InlineApplyWithArgumentsObject(IR::Instr * callInstr, IR::In
...
 
     // Move argouts close to call
-    IR::Instr* argInsertInstr = builtInStartInstr;
+    /*IR::Instr* argInsertInstr = builtInStartInstr;
     builtInStartInstr->IterateArgInstrs([&](IR::Instr* argInstr) {
         argInstr->Move(argInsertInstr);
         argInsertInstr = argInstr;
         return false;
-    });
+    });*/
```
satheeshravi: >);*/ [](start = 5, length = 4)

Remove?

lib/Backend/Inline.cpp
```diff
@@ -2506,8 +2537,97 @@ IR::Instr * Inline::InlineApplyWithArray(IR::Instr * callInstr, const FunctionJI
...
+        callInstr->InsertBefore(useCallTargetInstr);
+    }
+
+    InsertInlineeBuiltInStartEndTags(callInstr, 2); // 2 args (implicit this + explicit this)
+
+    IR::Instr * startCall = IR::Instr::New(Js::OpCode::StartCall,
+        IR::RegOpnd::New(TyVar, callInstr->m_func),
+        IR::IntConstOpnd::New(1, TyInt32, callInstr->m_func),
+        callInstr->m_func);
+    callInstr->InsertBefore(startCall);
```
satheeshravi: at prima facie, I see that this piece of the code is same as its counterpart API. Do you wanna a common API which forks to 2 different flavors ?

lib/Backend/Inline.cpp
```diff
@@ -2506,8 +2537,97 @@ IR::Instr * Inline::InlineApplyWithArray(IR::Instr * callInstr, const FunctionJI
...
+
+        linkOpnd = argInstr->GetSrc2();
+        return false;
+    });
+    linkOpnd->AsRegOpnd()->m_sym->m_isInlinedArgSlot = true;
+}
+
+/* 
+This method only inlines targets which are script functions, under the
+condition that the second argument (if any) passed to apply is arguments object.
```
satheeshravi: > passed to apply is arguments object. [](start = 43, length = 37)

why is this? - what about for other array objects?

lib/Backend/Inline.cpp
rajatd: Because I want to pass in the actual number of arguments passed to the apply call. inlineeCallArgCount takes into account the this pointer as well for a call to apply.

---
In reply to: [85817969](https://github.com/Microsoft/ChakraCore/pull/1838#discussion_r85817969) [](ancestors = 85817969)

lib/Backend/Inline.cpp
rajatd: I'm all ears for better names. However, InlineApplyWithArrayArgument seems to imply just the presence of the array argument; this API is for when the array argument is actually an array and not the arguments object. 

---
In reply to: [85818453](https://github.com/Microsoft/ChakraCore/pull/1838#discussion_r85818453) [](ancestors = 85818453)

lib/Backend/Inline.cpp
rajatd: this is passed as a ref for call target inlining.

---
In reply to: [85820234](https://github.com/Microsoft/ChakraCore/pull/1838#discussion_r85820234) [](ancestors = 85820234)

lib/Backend/Inline.cpp
rajatd: yes, thanks

---
In reply to: [85820369](https://github.com/Microsoft/ChakraCore/pull/1838#discussion_r85820369) [](ancestors = 85820369)

lib/Backend/Inline.cpp
rajatd: A common API for just inserting the StartCall?

---
In reply to: [85822614](https://github.com/Microsoft/ChakraCore/pull/1838#discussion_r85822614) [](ancestors = 85822614)

lib/Backend/Inline.cpp
rajatd: No, it doesn't. We dont yet support script function target inlining when apply is called with an array (which is not the arguments object)

---
In reply to: [85823107](https://github.com/Microsoft/ChakraCore/pull/1838#discussion_r85823107) [](ancestors = 85823107)

lib/Backend/Inline.cpp
```diff
@@ -2548,7 +2668,7 @@ bool Inline::InlineApplyTarget(IR::Instr *callInstr, const FunctionJITTimeInfo*
     const auto inlineCacheIndex = applyTargetLdOpnd->AsPropertySymOpnd()->m_inlineCacheIndex;
     const auto inlineeData = inlinerData->GetLdFldInlinee(inlineCacheIndex);
 
-    if (!isArrayOpndArgumentsObject || SkipCallApplyTargetInlining_Shared(callInstr, inlinerData, inlineeData, /*isApplyTarget*/ true, /*isCallTarget*/ false))
+    if ((!isArrayOpndArgumentsObject && (argsCount == 2)) || SkipCallApplyScriptTargetInlining_Shared(callInstr, inlinerData, inlineeData, /*isApplyTarget*/ true, /*isCallTarget*/ false))
```
rajatd: The caller does the argsCount == 1 checked after the call to this API. I'm calling this API to at least get the inlineeData for the target

---
In reply to: [85824508](https://github.com/Microsoft/ChakraCore/pull/1838#discussion_r85824508) [](ancestors = 85824508)

lib/Backend/Inline.cpp
satheeshravi: So should the above assert be changed to inlineCallArgCount >=1 ?

---
In reply to: [85823517](https://github.com/Microsoft/ChakraCore/pull/1838#discussion_r85823517) [](ancestors = 85823517,85817969)

lib/Backend/Inline.cpp
satheeshravi: Somehow the highlighting got messed up. :(

---
In reply to: [85857121](https://github.com/Microsoft/ChakraCore/pull/1838#discussion_r85857121) [](ancestors = 85857121,85824174,85822614)

lib/Backend/Inline.cpp
rajatd: Not really. If you look at the calculation of inlineCallArgCount, it is obtained by subtracting 1 from actualCount for built-ins that dont use the this pointer. So, it can be zero. For apply I know that it uses this pointer, so I know that inlineCallArgCount takes into account the 'this' argument. Since, I want to pass the actual arguments passed to .apply, I discount the implicit 'this' argument.

---
In reply to: [85857022](https://github.com/Microsoft/ChakraCore/pull/1838#discussion_r85857022) [](ancestors = 85857022,85823517,85817969)

rajatd:
Thanks @satheeshravi

PR #1843, opened 2016-10-28
Update TLS Slot to ThreadLocal

obastemur:
This PR aims to improve performance on XPLAT
and simplified initialization process for ChakraCore.

XPLAT, we don't have a fast implementation of TLS and the fastest
we can get is ThreadLocal variable.

Using TLS over ThreadLocal also adds some additional layer to JsrtContext
and ThreadContextTlsEntry interfaces.


obastemur:
Also removed all TLS related stuff from PAL.

obastemur:

PR #1844, opened 2016-10-28
Don't hoist InlineeStart in peeps

ThomsonTan:
InlineeStart is hoisted in peeps for branching (the same instruction in branch follower and target is hoisted before branch). The owner function of InlineeStart is not checked for instruction equivalence (actually cannot hoist even for the same function, because the inline ranges in inliner function are different). Hoist InlineeStart causes encoder not generating InlineeFrameRecord for some inlinee which will corrupt baiout and stack walk in such inlinee.


ThomsonTan:

rajatd:
:shipit:

PR #1845, opened 2016-10-28
Redeferral allocation savings

pleath:
Avoid creating a useless extra FunctionInfo when we transition from ParseableFunctionInfo to FunctionBody on un-deferral.


pleath:

PR #1846, opened 2016-10-28
decommit interpreter thunk emitter

leirocks:
previously because the decommit call is after the scriptContext closing call so the decommit call is rejected.
moving the decommitting to script context cleaning up call and save the decommit call.


lib/Runtime/Base/ScriptContext.cpp
```diff
@@ -1279,6 +1275,11 @@ if (!sourceList)
 #endif
     }
 
+    void ScriptContext::SetIsClosed()
+    {
+        this->isClosed = true;
```
MikeHolman: I think with this we might not call: ServerContextManager::UnRegisterScriptContext(scriptContextInfo);

Maybe we should have check on server during cleanup if ScriptContext is not closed then close/unregister it first

lib/Backend/ServerScriptContext.cpp
```diff
@@ -63,11 +63,15 @@ ServerScriptContext::~ServerScriptContext()
 #endif
     if (m_asmJsInterpreterThunkBufferManager)
     {
+        m_asmJsInterpreterThunkBufferManager->Decommit();
```
MikeHolman: do you think there is benefit to decommit first? we are just going to free right after this so seems like maybe not necessary

leirocks: the following delete seems not decommiting the page
~EmitBufferManager()->Clear()->FreeAllocations(true);  // true means release

    if (release)
    {
        this->allocations = nullptr;
    }
    else
    {
        this->allocationHeap.DecommitAll();
    }

weird design though


---
In reply to: [85839883](https://github.com/Microsoft/ChakraCore/pull/1846#discussion_r85839883) [](ancestors = 85839883)

leirocks: just double checked, the delete will free first and then fill debugBreak, then decommit. however at this stage we don't need to free the heap and fill debugbreak, so having decommit first should have a little better perf (save a WriteProcessMemory call)

---
In reply to: [85840452](https://github.com/Microsoft/ChakraCore/pull/1846#discussion_r85840452) [](ancestors = 85840452,85839883)

PR #1849, opened 2016-10-31
use RPC rundown to do the server cleanup

leirocks:
instead of polling all the client processes exit code on every call to determin client is alive or not, implement the context handles which will be cleaned up in case the client disconnect or terminated abnormally.


leirocks:

PR #1850, opened 2016-10-31
ScriptContext close during codegen should not cause failfast

MikeHolman:


lib/Backend/NativeCodeGenerator.cpp
```diff
@@ -901,6 +901,11 @@ NativeCodeGenerator::CodeGen(PageAllocator * pageAllocator, CodeGenWorkItem* wor
             workItem->GetJITData(),
             scriptContext->GetRemoteScriptAddr(),
             &jitWriteData);
+        if (hr == E_ACCESSDENIED && body->GetScriptContext()->IsClosed())
```
agarwal-sandeep: Do we need to restrict this with hr == E_ACCESSDENIED? Seems like what ever the error code is if SC is closed it should be aborted. Today the error code might be E_ACCESSDENIED but in future if any new error code is added this will break?

MikeHolman: There are some cases where we will want to failfast here. For example, in case WriteProcessMemory fails we throw some different hresult and we do want to failfast (as that would be an AV for in-proc JIT).

E_ACCESSDENIED is only thrown in case script context was closed. But maybe we should move to use custom hresults for this sort of thing which are more meaningful for us. At least that makes more clear what code is intended to do.

agarwal-sandeep: Can we add a comment here for future reference

leirocks: agree we might need a custom hresult for it. currently WriteProcessMemory can also return E_ACCESSDENIED, however with scriptContext->IsClosed() check we can differentiate it and fail fast. but a custom hresult may be more clear

PR #1851, opened 2016-10-31
xplat: Better Perf + PAL cleanup

obastemur:
On xplat, measured ~20% better perf with Octane. Didn't check others.

Details;
- Cleanup RtlXXXX stuff from PAL and ch. FaultInjection is not yet active on Xplat.
- inline couple of one liner methods (they are in a very hot zone)
- prefered updating array index in place instead of a seperate memset call (in a very hot zone)


obastemur:
Will split the cleanup and perf

PR #1852, opened 2016-10-31
Asm.js revert backend optimization

Cellule:
Reverts an optimization done in #1806.

We can't calculate const math in asm.js and wasm the same way we do for normal javascript.
The following case is erroneous.
```js
// flags: -maic:1
var asmHeap = new ArrayBuffer(1 << 20);
var m = function (stdlib, foreign, heap) {
  'use asm';
  function f(d0, i1) {
    d0 = +d0;
    i1 = i1 | 0;
    return 248749990 / 4294967295 | 0;
  }
  return f;
}({}, {}, asmHeap);
print(m());
print(m());
// Expected Output:
// 0
// 0
// Actual Output:
// 0
// -248749990
```

I am adding this work item in the wasm todo list to revisit this at a later time.
Most likely most operators can do it the same way as javascript, but it needs deeper investigation to find and deal with the special cases.

This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/1852)

### DIFF
--- a/lib/Backend/IR.cpp
+++ b/lib/Backend/IR.cpp
@@ -3618,7 +3618,6 @@ bool Instr::BinaryCalculator(IntConstType src1Const, IntConstType src2Const, Int
 
     switch (this->m_opcode)
     {
-    case Js::OpCode::Add_I4:
     case Js::OpCode::Add_A:
         if (IntConstMath::Add(src1Const, src2Const, &value))
         {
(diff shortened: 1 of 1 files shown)


Cellule:

MikeHolman:
Ah I guess it gets considered as -1?

MikeHolman:
lgtm

Cellule:
Yes, these path probably don't check the IRType to do unsigned math.
Also there are cases where we have to trap (wasm) or return 0 (asm.js)

PR #1853, opened 2016-10-31
xplat: Better Performance

obastemur:
On xplat, measured ~8-12% better perf with Octane. Didn't check others.
This should also affect Windows performance numbers.

- replaced `memset` call with an in-place assignment. `memset` was taking the most expensive path.

less importantly;
- inline couple of one liner methods (they are in a very hot zone)

EDIT: Revising ~20% to ~8-12% . Tested on a cleaner system.

### DIFF
--- a/lib/Runtime/Base/FunctionBody.cpp
+++ b/lib/Runtime/Base/FunctionBody.cpp
@@ -8688,31 +8688,34 @@ namespace Js
             Type *type = this->types[i];
             if (type != nullptr)
             {
-                if (!recycler->IsObjectMarked(type))
-                {
-                    this->types[i] = nullptr;
-                }
-                else
+                this->types[i] = nullptr;
+                if (recycler->IsObjectMarked(type))
                 {
                     // compact the types array by moving non-null types
                     // at the beginning.
                     this->types[nonNullIndex++] = type;
-                    isAnyTypeLive = true;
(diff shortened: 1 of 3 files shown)


lib/Runtime/Base/FunctionBody.cpp
```diff
@@ -8697,22 +8697,29 @@ namespace Js
                     // compact the types array by moving non-null types
                     // at the beginning.
                     this->types[nonNullIndex++] = type;
-                    isAnyTypeLive = true;
 #if DBG
                     isGuardValuePresent = this->guard->GetValue() == reinterpret_cast<intptr_t>(type) ? true : isGuardValuePresent;
 #endif
+                    this->types[i] = nullptr;
```
curtisman: What if nonNullIndex (before++) == i? You will null out something that we just assigned.

obastemur: i >= nonNullIndex. Carrying this->types[i] above should take care of the task since we already have `type` is looking to that index. Worst case i == nonNullIndex.

curtisman:
Good find. 
@kunalspathak  FYI.

kunalspathak:
LGTM

obastemur:
@curtisman @kunalspathak Thanks for the review

PR #1854, opened 2016-10-31
Module hangs if it imports an errored module

AtOMiCNebula:
ResolveExternalModuleDependencies has no problem today resolving against a module that has already been determined to be in an error state.  In this case, it has already notified its dependents at the time of its failure, and no further notification will occur.  As a result, the module hangs if we do not handle this scenario.

This fix addresses the issue by not proceeding if the module being imported is already in an error state.  The error is passed upward.

### DIFF
--- a/lib/Runtime/Language/SourceTextModuleRecord.cpp
+++ b/lib/Runtime/Language/SourceTextModuleRecord.cpp
@@ -559,6 +559,7 @@ namespace Js
                 ArenaAllocator* allocator = scriptContext->GeneralAllocator();
                 childrenModuleSet = (ChildModuleRecordSet*)AllocatorNew(ArenaAllocator, allocator, ChildModuleRecordSet, allocator);
             }
+            Var error = nullptr;
             requestedModuleList->MapUntil([&](IdentPtr specifier) {
                 ModuleRecordBase* moduleRecordBase = nullptr;
                 SourceTextModuleRecord* moduleRecord = nullptr;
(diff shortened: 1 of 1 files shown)


AtOMiCNebula:
@boingoing @akroshg Baby's first ChakraCore change! ;)

akroshg:
Could you create a native unittest for this?
this PR  #1803 has added native unittests to craft such scenario.

Yongqu:
LGTM

lib/Runtime/Language/SourceTextModuleRecord.cpp
```diff
@@ -587,8 +594,12 @@ namespace Js
             });
             if (FAILED(hr))
             {
-                JavascriptError *error = scriptContext->GetLibrary()->CreateError();
-                JavascriptError::SetErrorMessageProperties(error, hr, _u("fetch import module failed"), scriptContext);
+                if (error == nullptr)
+                {
```
AtOMiCNebula: I want to propagate the inner error outward, so I only create the "fetch import module failed" error if we don't have a better one to use.

dilijev:
@boingoing @AtOMiCNebula Is this PR waiting on anything? Can we get that test added and merge this in? (At this point this should probably go into release/1.4 so the PR will need to be retargeted.)

AtOMiCNebula:
I let this one get super stale, sorry!  I hope to be able to revisit this in the not too distant future; I'll close this PR for now, and open a new PR after rebasing the change onto a more recent master. :)

dilijev:
Thanks @AtOMiCNebula! Looking forward to the update!

PR #1890, opened 2016-10-31
Bugfix: Initialization of allocationBuckets in multi-thread env

kunalspathak:
In #1363, I added a static AllocationBuckets 2D matrix for JavascriptArray that will be initialize at runtime. However it can be simultaneously accessed by 2 threads (runtime and code-gen) and one thread might use old information of this matrix. Fixed it by adding a boolean flag to indicate when the initialization of matrix is done.
For ARM, since order of reads/writes are not guaranteed, added `MemoryBarrier` before reading and after writing to this boolean variable.

Test: Unit test passes.

### DIFF
--- a/lib/Runtime/Library/JavascriptArray.inl
+++ b/lib/Runtime/Library/JavascriptArray.inl
@@ -1670,18 +1670,12 @@ SECOND_PASS:
     template<class ArrayType> 
     void JavascriptArray::EnsureCalculationOfAllocationBuckets()
     {
-        // If allocation size for current ArrayType has already calculated, no need to recalculate it.
-        bool determineAllocationSize = ArrayType::allocationBuckets[0][MissingElementsCountIndex] == 0;
-        if (determineAllocationSize)
+        uint temp;
+        for (uint8 i = 0;i < ArrayType::AllocationBucketsCount;i++)
         {
-            uint temp;
-            for (uint8 i = 0;i < ArrayType::AllocationBucketsCount;i++)
-            {
-                ArrayType::allocationBuckets[i][AllocationSizeIndex] = (uint)DetermineAllocationSize<ArrayType, 0>(ArrayType::allocationBuckets[i][AllocationBucketIndex], nullptr, &temp);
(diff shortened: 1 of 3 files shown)


lib/Runtime/Library/JavascriptArray.h
```diff
@@ -417,6 +419,7 @@ namespace Js
         void TryAddToSegmentMap(Recycler* recycler, SparseArraySegmentBase* seg);
 
     private:
+
```
kunalspathak: I will undo the white lines I added accidentally.

kunalspathak:
@digitalinfinity , can you please help review?

kunalspathak:
@dotnet-bot test ubuntu_linux_debug_static  please

lib/Runtime/Library/JavascriptArray.inl
```diff
@@ -1671,15 +1671,24 @@ SECOND_PASS:
...
         // If allocation size for current ArrayType has already calculated, no need to recalculate it.
-        bool determineAllocationSize = ArrayType::allocationBuckets[0][MissingElementsCountIndex] == 0;
-        if (determineAllocationSize)
+#if defined(_M_ARM32_OR_ARM64)
+        // On ARM machines, order of writes is not guaranteed. So we need to have one memory barrier
+        // before reading the IsArrayAllocationBucketsInitialized and another one after writing
+        // to the IsArrayAllocationBucketsInitialized.
+        MemoryBarrier();
+#endif
+        if (!ArrayType::IsArrayAllocationBucketsInitialized)
```
curtisman: Can't we just use a global ctor to initialize this?

kunalspathak: Yeah, i think thats much better and cleaner. I was deferring the initialization till we actually have to use those values. But doesn't hurt if we initialize one-time in `JavascriptLibrary` for all array types. I will update PR. Thanks for the feedback.

satheeshravi:
:shipit:

kunalspathak:
Thanks @satheeshravi  and @curtisman  for the review.

kunalspathak:
Landed in 8983758

PR #1892, opened 2016-11-01
fix re-entrance issue in PageAllocatorPool

leirocks:
introduced in the RPC server rundown change by mistake

### DIFF
--- a/lib/Backend/PageAllocatorPool.cpp
+++ b/lib/Backend/PageAllocatorPool.cpp
@@ -21,7 +21,8 @@ PageAllocatorPool::PageAllocatorPool()
 
 PageAllocatorPool::~PageAllocatorPool()
 {
-    Shutdown();
+    AutoCriticalSection autoCS(&cs);
+    RemoveAll();
 }
 
 void PageAllocatorPool::Initialize()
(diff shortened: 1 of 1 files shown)


akroshg:
Is there a repro we can add just to prevent this issues?

lib/Backend/PageAllocatorPool.cpp
```diff
@@ -37,8 +38,10 @@ void PageAllocatorPool::Shutdown()
     AutoCriticalSection autoCS(&cs);
     if (Instance)
     {
-        CloseHandle(Instance->idleCleanupTimer);
-        Instance->RemoveAll();
+        if (Instance->idleCleanupTimer)
+        {
+            CloseHandle(Instance->idleCleanupTimer);
```
leirocks: yes I think, it should not enter here again though. or might be better to keep the data for debugging so we know the timer is initialized correctly

akroshg: What is the guarantee that it will not re-enter again (in future changes) ?

Other practice you can do that 
tempInstance = Instance
Instance = nullptr;
HeapDelete(tempInstance).
that way if the Shutdown calls itself - we will be safe.



---
In reply to: [86002065](https://github.com/Microsoft/ChakraCore/pull/1892#discussion_r86002065) [](ancestors = 86002065)

leirocks: I mean the function should not re-enter. if it re-enters there's a bug in caller which we should fix. I understand you approach which make sure it is safe but lost the opportunity to catch the bug in caller. I can use your approach and add assert though

---
In reply to: [86002826](https://github.com/Microsoft/ChakraCore/pull/1892#discussion_r86002826) [](ancestors = 86002826,86002065)

akroshg:
:shipit:

PR #1894, opened 2016-11-01
fix dead lock in SetLoopBodyEntryPoint

leirocks:
should not aquire FunctionProxy global lock while out of collecting code gen data phase. adding assertion to enforce that

### DIFF
--- a/lib/Runtime/Base/FunctionBody.cpp
+++ b/lib/Runtime/Base/FunctionBody.cpp
@@ -104,6 +104,7 @@ namespace Js
         {
             return nullptr;
         }
+        Assert(!this->GetScriptContext()->GetThreadContext()->GetJobProcessor()->GetCriticalSection()->IsLocked());
         AutoCriticalSection autoCS(&GlobalLock);
         return AuxPtrsT::GetAuxPtr(this, e);
     }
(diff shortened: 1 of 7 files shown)


lib/Runtime/Base/FunctionBody.cpp
```diff
@@ -8305,30 +8305,45 @@ namespace Js
...
-        if (!this->sharedPropertyGuards->TryGetValue(propertyId, &guard))
+        if (!localGuards->TryGetValue(propertyId, &guard))
         {
             ThreadContext* threadContext = scriptContext->GetThreadContext();
             guard = threadContext->RegisterSharedPropertyGuard(propertyId);
-            this->sharedPropertyGuards->Add(propertyId, guard);
+
+            struct AutoResizeLock
+            {
+                AutoResizeLock(SharedPropertyGuardDictionary* d) :d(d) { d->LockResize(); }
```
leirocks: you are correct. this is recycler allocated, should not worry the old buffer being deleted

lib/Runtime/Base/FunctionBody.cpp
```diff
@@ -8336,14 +8329,6 @@ namespace Js
         if (sharedPropertyGuards != nullptr)
         {
             auto guards = AnewArray(alloc, Js::PropertyId, sharedPropertyGuards->Count());
-
-            struct AutoResizeLock
-            {
-                AutoResizeLock(SharedPropertyGuardDictionary* d) :d(d) { d->LockResize(); }
-                ~AutoResizeLock() { d->UnlockResize(); }
-                SharedPropertyGuardDictionary* d;
-            } autoResizeLock(sharedPropertyGuards);
```
MikeHolman: I think this one we still need resize lock

leirocks: just took a closer look, we might still need lock on both location. not for old buffer being deleted, but for the count and the iterator to be consistent. while adding to the dictionary, it increase the count first, and then write the buffer, if background thread finished a get operation in between, it'll get an count-1 entries

MikeHolman:
:shipit:

Issue #1713, opened 2016-10-07
Optimize deleting last property on an object


kunalspathak:
Today if we are in `PathTypeHandler` and we delete a property we go to `SimpleDictionaryTypeHandler`. Having a type in `SimpleDictionaryTypeHandler` is expensive both performance and memory wise when it comes to adding properties. Most of the time, it evolves new type when properties are added and it is less likely to get shared. Optimize so that if we are deleting last property then transition to previous type of `PathTypeHandler`.

agarwal-sandeep:
Closing since this is complete and merged

kunalspathak:
Fixes https://github.com/Microsoft/ChakraCore/pull/1897

PR #1897, opened 2016-11-01
Optimize deleting last property on an object

agarwal-sandeep:
Today if we are in PathTypeHandler and we delete a property we go to
SimpleDictionaryTypeHandler. Having a type in SimpleDictionaryTypeHandler
is expensive both performance and memory wise when it comes to adding
properties. Most of the time, it evolves new type when properties are
added and it is less likely to get shared. Optimize so that if we are
deleting last property then transition to previous type of PathTypeHandler

Address issue #1713


agarwal-sandeep:
@pleath @curtisman Please take a look

lib/Runtime/Types/DynamicObject.cpp
```diff
@@ -295,6 +295,24 @@ namespace Js
...
 
+    void DynamicObject::ReplaceTypeToPredecessorType(DynamicType * predecessorType)
+    {
+        Assert(this->GetTypeHandler()->IsPathTypeHandler());
+        Assert(((PathTypeHandlerBase*)this->GetTypeHandler())->GetPredecessorType()->GetTypeHandler()->IsPathTypeHandler());
+
+        Assert(((PathTypeHandlerBase*)this->GetTypeHandler())->GetPredecessorType() == predecessorType);
+
+        Assert(!predecessorType->isLocked || predecessorType->GetTypeHandler()->GetIsLocked());
+        Assert(!predecessorType->isShared || predecessorType->GetTypeHandler()->GetIsShared());
```
kunalspathak: >isShared [](start = 33, length = 8)

nit: use GetIsShared(), GetIsLocked() instead.

lib/Runtime/Types/DynamicObject.cpp
```diff
@@ -295,6 +295,24 @@ namespace Js
...
+
+        PathTypeHandlerBase* currentPathTypeHandler = (PathTypeHandlerBase*)this->GetTypeHandler();
+        PathTypeHandlerBase* predecessorPathTypeHandler = (PathTypeHandlerBase*)predecessorType;
```
kunalspathak: >predecessorType [](start = 80, length = 15)

This type-cast doesn't make sense. Did you mean to do `(PathTypeHandlerBase*)predecessorType->GetTypeHandler()`

lib/Runtime/Types/DynamicObject.cpp
```diff
@@ -295,6 +295,24 @@ namespace Js
...
+
+        Assert(predecessorPathTypeHandler->GetInlineSlotCapacity() >= currentPathTypeHandler->GetInlineSlotCapacity());
```
kunalspathak: >decessorPathTypeHandler->GetInlineSlotCapacity() >= currentPathTypeHandler [](start = 18, length = 74)

Can we never have case where currentPathType has more inlineSlotCapacity than predecessor except the shrink slot case?

lib/Runtime/Types/PathTypeHandler.cpp
```diff
@@ -299,11 +299,90 @@ namespace Js
...
+
+    BOOL PathTypeHandlerBase::DeleteLastProperty(DynamicObject *const object)
+    {
+        // We are deleting the last property from PathTypeHandler
+        DynamicType* predecessorType = this->GetPredecessorType();
+
+        // -----------------------------------------------------------------------------------------
+        //     Current Type  | Predecessor Type |                  Action
+        // -----------------------------------------------------------------------------------------
+        //          OHI      |      OHI         | No movement needed
```
kunalspathak: >OHI [](start = 20, length = 3)

nit: In the comment, can you also write OHI = ObjectHeaderInlined

lib/Runtime/Types/PathTypeHandler.cpp
```diff
@@ -299,11 +299,90 @@ namespace Js
...
+        {
+            Assert(predecessorTypeHandler->GetInlineSlotCapacity() == (this->GetUnusedBytesValue() - 1));
+            this->MoveAuxSlotsToObjectHeader(object);
+        }
+
+#if DBG
+        if ((isCurrentTypeOHI && isPredecessorTypeOHI) || (!isCurrentTypeOHI && !isPredecessorTypeOHI))
+        {
+            AssertMsg(this->GetInlineSlotCapacity() == predecessorTypeHandler->GetInlineSlotCapacity(), "InlineSlotCapacity of types should match");
+        }
```
kunalspathak: can you instead have assert `DynamicObject::IsTypeHandlerCompatibleForObjectHeaderInlining(oldtypeHandler, newTypeHandler)` . It validates much more than just inlineSlotCapacity.

lib/Runtime/Types/PathTypeHandler.cpp
```diff
@@ -299,11 +299,90 @@ namespace Js
...
+
+        // Another type (this) reached the old (predecessorType) type so share it.
+        // ShareType will take care of invalidating fixed fields and removing singleton object from predecessorType
+        predecessorType->ShareType();
+
+        this->typePath->ClearSingletonInstanceIfSame(object);
+
+        object->ReplaceTypeToPredecessorType(predecessorType);
+
+        return TRUE;
```
kunalspathak: > return TRUE [](start = 7, length = 12)

should this always be `TRUE`. Can you refer ES6 spec to double check? I remember there are scenarios when `delete` returns false.

lib/Runtime/Types/PathTypeHandler.cpp
```diff
@@ -299,11 +299,90 @@ namespace Js
...
+        {
+            newInlineSlots[i] = oldInlineSlots[i];
+            i++;
+        }
+
+        // auxSlot should only have 2 entry, move that to inlineSlot
+        newInlineSlots[i] = auxSlotZero;
+        newInlineSlots[i + 1] = auxSlotOne;
+
+        Assert(i == this->GetInlineSlotCapacity());
```
kunalspathak: >Assert(i == this->GetInlineSlotCapacity()); [](start = 8, length = 43)

won't this always true because `i++` is driven by this->GetInlineSlotCapacity() ?

lib/Runtime/Types/DynamicObject.cpp
```diff
@@ -295,6 +295,24 @@ namespace Js
         this->type = type;
     }
 
+    void DynamicObject::ReplaceTypeToPredecessorType(DynamicType * predecessorType)
```
satheeshravi: >To [](start = 35, length = 2)

nit: with

lib/Runtime/Types/PathTypeHandler.cpp
```diff
@@ -299,11 +299,90 @@ namespace Js
...
+        AssertMsg(this->GetUnusedBytesValue() - this->GetInlineSlotCapacity() == 3, "Should have only 3 values in auxSlot");
+
+        // Get the auxSlot[0] and auxSlot[1] value as we will over write it
+        Var auxSlotZero = object->GetAuxSlot(0);
+        Var auxSlotOne = object->GetAuxSlot(1);
+
+        // Move all current inline slots up to object header inline offset
+        Var *const oldInlineSlots = reinterpret_cast<Var *>(reinterpret_cast<uintptr_t>(object) + this->GetOffsetOfInlineSlots());
+        Var *const newInlineSlots = reinterpret_cast<Var *>(reinterpret_cast<uintptr_t>(object) + this->GetOffsetOfObjectHeaderInlineSlots());
+        PropertyIndex i = 0;
```
satheeshravi: >i [](start = 22, length = 1)

nit:rename

lib/Runtime/Types/PathTypeHandler.cpp
```diff
@@ -299,11 +299,90 @@ namespace Js
...
 
-    BOOL PathTypeHandlerBase::DeleteProperty(DynamicObject* instance, PropertyId propertyId, PropertyOperationFlags flags)
+    void PathTypeHandlerBase::MoveAuxSlotsToObjectHeader(DynamicObject *const object)
     {
-#ifdef PROFILE_TYPES
-        instance->GetScriptContext()->convertPathToDictionaryCount2++;
+        Assert(object);
+        AssertMsg(!this->IsObjectHeaderInlinedTypeHandler(), "Already ObjectHeaderInlined?");
+        AssertMsg(!object->HasObjectArray(), "Can't move auxSlots to inline when we have ObjectArray");
+        AssertMsg(this->GetUnusedBytesValue() - this->GetInlineSlotCapacity() == 3, "Should have only 3 values in auxSlot");
```
satheeshravi: >"Should have only 3  [](start = 84, length = 20)

nit: it will be a good idea to add comments here stating - why the number of properties should be 3 in auxSlots.

lib/Runtime/Types/PathTypeHandler.cpp
satheeshravi: @kunalspathak : So we don't delete in those scenarios ?

---
In reply to: [86039231](https://github.com/Microsoft/ChakraCore/pull/1897#discussion_r86039231) [](ancestors = 86039231)

lib/Runtime/Types/PathTypeHandler.cpp
```diff
@@ -299,11 +299,90 @@ namespace Js
...
+        Var auxSlotZero = object->GetAuxSlot(0);
+        Var auxSlotOne = object->GetAuxSlot(1);
+
+        // Move all current inline slots up to object header inline offset
+        Var *const oldInlineSlots = reinterpret_cast<Var *>(reinterpret_cast<uintptr_t>(object) + this->GetOffsetOfInlineSlots());
+        Var *const newInlineSlots = reinterpret_cast<Var *>(reinterpret_cast<uintptr_t>(object) + this->GetOffsetOfObjectHeaderInlineSlots());
+        PropertyIndex i = 0;
+        while (i < this->GetInlineSlotCapacity())
+        {
+            newInlineSlots[i] = oldInlineSlots[i];
```
agarwal-sandeep: We don't explicitly delete auxSlots even when we resize (DynamicTypeHandler::AdjustSlots, EXPLICIT_FREE_SLOTS is undefined)

lib/Runtime/Types/PathTypeHandler.cpp
```diff
@@ -299,11 +299,90 @@ namespace Js
         return PathTypeHandlerBase::AddPropertyInternal(instance, propertyId, value, info, flags, possibleSideEffects);
     }
 
-    BOOL PathTypeHandlerBase::DeleteProperty(DynamicObject* instance, PropertyId propertyId, PropertyOperationFlags flags)
+    void PathTypeHandlerBase::MoveAuxSlotsToObjectHeader(DynamicObject *const object)
```
satheeshravi: >MoveAuxSlotsToObjectHeader [](start = 30, length = 26)

pls consider adding traces for this scenario. one of them could be ObjectHeaderInlining phase.

kunalspathak: I was about to suggest this. Will be good for debugging as well.

---
In reply to: [86214204](https://github.com/Microsoft/ChakraCore/pull/1897#discussion_r86214204) [](ancestors = 86214204)

test/TypeSystem/rlexe.xml
```diff
@@ -0,0 +1,9 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<regress-exe>
+  <test>
+    <default>
+      <files>PathTypeDeleteLastProperty.js</files>
+      <compile-flags>-mic:1 -msjrc:1</compile-flags>
```
satheeshravi: >-msjrc:1 [](start = 28, length = 8)

do you need msjrc?

lib/Runtime/Types/PathTypeHandler.cpp
kunalspathak: 1 scenario was if the property doesn't exist, but will be good to check other scenarios from spec.

---
In reply to: [86213097](https://github.com/Microsoft/ChakraCore/pull/1897#discussion_r86213097) [](ancestors = 86213097,86039231)

test/TypeSystem/rlexe.xml
```diff
@@ -0,0 +1,9 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<regress-exe>
+  <test>
+    <default>
+      <files>PathTypeDeleteLastProperty.js</files>
```
satheeshravi: >PathTypeDeleteLastProperty [](start = 13, length = 26)

Do you wanna add this under Object folder?

lib/Runtime/Types/PathTypeHandler.cpp
```diff
@@ -312,6 +391,18 @@ namespace Js
             return PathTypeHandlerBase::DeleteItem(instance, indexVal, flags);
         }
 
+        PropertyIndex index = PathTypeHandlerBase::GetPropertyIndex(propertyId);
+        uint16 pathLength = GetPathLength();
+        if ((index + 1) == pathLength &&
+            this->GetPredecessorType() != nullptr &&
+            scriptContext->GetThreadContext()->GetDynamicObjectEnumeratorCache(instance->GetDynamicType()) == nullptr)
```
kunalspathak: >GetDynamicObjectEnumeratorCache [](start = 47, length = 31)

Can you add comment for this check?

lib/Runtime/Types/PathTypeHandler.cpp
kunalspathak: Could you add explicitly delete auxslots under `#ifdef EXPLICIT_FREE_SLOTS` to have it consistent and in case we enable it in future?

---
In reply to: [86213167](https://github.com/Microsoft/ChakraCore/pull/1897#discussion_r86213167) [](ancestors = 86213167)

lib/Runtime/Types/PathTypeHandler.cpp
agarwal-sandeep: Even for non-existing properties delete on PathType is returning true (before my change as well) so we are good here. As @rajatd mentioned I will add a early exit if index is NoSlot which will save call to convert to simple dictionary. Also I will add an assert after ConvertToSimpleDictionaryType to see if ever it returns false in PathTypeHandler

lib/Runtime/Types/PathTypeHandler.cpp
```diff
@@ -299,11 +299,103 @@ namespace Js
...
 
-    BOOL PathTypeHandlerBase::DeleteProperty(DynamicObject* instance, PropertyId propertyId, PropertyOperationFlags flags)
+    void PathTypeHandlerBase::MoveAuxSlotsToObjectHeader(DynamicObject *const object)
     {
-#ifdef PROFILE_TYPES
-        instance->GetScriptContext()->convertPathToDictionaryCount2++;
+        Assert(object);
+        AssertMsg(!this->IsObjectHeaderInlinedTypeHandler(), "Already ObjectHeaderInlined?");
+        AssertMsg(!object->HasObjectArray(), "Can't move auxSlots to inline when we have ObjectArray");
+
```
pleath: Can this be written in a way that doesn't assume that there will be exactly 2 inline properties? I ask because we ran into issues with hard-coded assumptions about capacity when we expanded object header inlining to handle more cases, so I'd rather not add more such assumptions now.

agarwal-sandeep: When we reach here there can be any number of inline properties but only 3 properties in auxSlots. Even if we change inlineSlotCapacity this won't get effected until we change DynamicObject to have anything other than auxSlots and objectArrays.

lib/Runtime/Types/PathTypeHandler.cpp
```diff
@@ -299,11 +299,103 @@ namespace Js
...
+        // auxSlot should only have 2 entry, move that to inlineSlot
+        newInlineSlots[propertyIndex++] = auxSlotZero;
+        newInlineSlots[propertyIndex++] = auxSlotOne;
+
+#ifdef EXPLICIT_FREE_SLOTS
+        object->GetRecycler()->ExplicitFreeNonLeaf(auxSlots, auxSlotsCapacity * sizeof(Var));
 #endif
+
+        Assert(this->GetPredecessorType()->GetTypeHandler()->IsPathTypeHandler());
```
kunalspathak: > Assert(this->GetPredecessorType()->GetTypeHandler()->IsPathTypeHandler()); [](start = 7, length = 75)

you have already asserted this in `DeleteLastProperty`.

PR #1902, opened 2016-11-02
fix case where reused property record isn't updated correctly on server

MikeHolman:
There was a synchronization issue between numeric property map on server and runtime.

The following is the problematic sequence:
Add numeric property id
Update server map
Reclaim that property id
Add same numeric property id (this incorrectly causes to be put on pendingPropsList)
Remove that property (since it is in pendingPropsList, we don't think we need to update server)
Add same non-numeric property id
update server map (does nothing; should remove propId)
Codegen (with that propid used in an InitFld).

The server still has this propId in numeric props list, which breaks some assumptions in globopt.


leirocks:
Nice find!
:shipit:

PR #1905, opened 2016-11-02
Couple of redeferral fixes

pleath:
1. Fix an off-by-one error in the heuristic that was causing us not to redefer with default switches. 2. Fix a check on the entry point info's state that was preventing us from redeferring functions that have not been jitted.


pleath:

PR #1906, opened 2016-11-03
WASM - Imported wasm functions are used as is

### LINKED ISSUE #1887: WASM - Int64/Float32 support in call import
Add support for int64 and float32 arguments when calling imports
Depends on #1895

### LINKED ISSUE #1895: WASM - Treat imported WebAssembly functions as internal functions
When importing a WebAssembly function into another module, that function needs to be called without boxing/unboxing.
The main reason is to support types that are not supported by javascript.

The solution we came up with is to create a wasm thunk (similar to #1807) that will then call the import and put this in the function array of the module's environment.
If the function we import is a WebAssembly function, we will replace the thunk by that function.
Then all calls will be internal calls.
This also solves #1873
Depends on #1821

Cellule:
A thunk is created for every imports. That thunk will do the conversion from primitive types to javascript types.
If the imported function is a WebAssembly function, the thunk is not used in that instance.
This also simplifies the internal model of the module because the function index now matches the representation the spec has.

Closes #1895 
Closes #1887

This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/1906)


lib/WasmReader/WasmBinaryReader.cpp
```diff
@@ -790,7 +791,7 @@ WasmBinaryReader::ReadElementSection()
             }
             if (funcType == FunctionIndexTypes::Import)
             {
-                ThrowDecodingError(_u("Import functions in the table NYI"));
+                ThrowDecodingError(_u("Imported functions are not allowed in the table"));
```
MikeHolman: is this a possible error? i'd think this indicates bug in our thunk generation?

Cellule:
Review status: 0 of 46 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[lib/WasmReader/WasmBinaryReader.cpp, line 794 at r1](https://reviewable.io:443/reviews/microsoft/chakracore/1906#-KVbnJ6NN9W14jB26mJn-r1-794:-KVkluD_RL3uO8BuIwNN:b5feiai) ([raw file](https://github.com/microsoft/chakracore/blob/4e0138fc2e9dcc1f4d03f3fae04a47562a1016d9/lib/WasmReader/WasmBinaryReader.cpp#L794)):*
> is this a possible error? i'd think this indicates bug in our thunk generation?
</details>
Yes this is would indeed be a bug since `m_module->GetFunctionIndexType` cannot return this type anymore.
I will leave the exception to be safe but I will also put an assert to help us catch this if it happens in the future

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/microsoft/chakracore/1906)*

MikeHolman:
:shipit:

PR #1910, opened 2016-11-03
String.prototype.lastIndexOf behaves incorrectly when string contains high order ASCII characters

tcare:
Fixes OS 9271625.

Our implementation of String.prototype.lastIndexOf has been incorrect when the string contains ASCII values >0x7F. I believe this has been broken since IE9.

If the string contains ASCII >0x7F, we avoid using the Boyer-Moore shift table and revert back to a brute force search. This search was refactored out incorrectly for edge cases.

e.g. "ÄÜÖABC".lastIndexOf("ÄÜÖABC", 0) returns -1 instead of 0.

Changes:
- Re-implemented the brute-force.
- Fixed up a lot of the naming so it is more clear.
- Switched back to wmemcmp.
- Added a bunch of test cases.

Really we should consider using Boyer-Moore for more than ASCII >0x7F, and possibly change it to TextbookBoyerMoore along with indexOf. There may be potential perf improvements here even for ASCII <0x7F which I will outline in another issue.


tcare:
@curtisman

tcare:
Ping..

tcare:
@akroshg @aneeshdk

dilijev:
Since it's all single-byte code points, it probably doesn't matter in this case, but I thought it would be worth mentioning that 8-bit (Extended) ASCII could actually represent several different encodings depending on system language and context (https://en.wikipedia.org/wiki/Extended_ASCII#ISO_8859_and_proprietary_adaptations)

tcare:
Right. I think we need to move away from this implementation. But for the moment it should just be brought up to spec.

tcare:
@akroshg or @curtisman would you mind giving a 2nd signoff?

lib/Runtime/Library/JavascriptString.cpp
```diff
@@ -1149,83 +1142,98 @@ namespace Js
...
+        if (searchLen > inputLen)
         {
             return JavascriptNumber::ToVar(-1, scriptContext);
         }
 
-        if(args.Info.Count > 2)
+        if (args.Info.Count > 2)
         {
             double pos = JavascriptConversion::ToNumber(args[2], scriptContext);
             if (!JavascriptNumber::IsNan(pos))
```
akroshg: >(! [](start = 15, length = 2)

curious what happens on positive or negative infinity

tcare: Positive infinity is the normal upper bound for non-specified position (per spec) and negative infinity will get normalized to 0 when doing max(pos,0).

Would be good to add tests for that though.

akroshg: alright, LGTM

---
In reply to: [87507596](https://github.com/Microsoft/ChakraCore/pull/1910#discussion_r87507596) [](ancestors = 87507596)

PR #1911, opened 2016-11-03
Enabling the well-known symbols IsConcatSpreadable, ToPrimitive and HasInstance by default

aneeshdk:
Enabling the well-known symbols IsConcatSpreadable, ToPrimitive and HasInstance by default

### DIFF
--- a/lib/Runtime/Language/JavascriptConversion.cpp
+++ b/lib/Runtime/Language/JavascriptConversion.cpp
@@ -321,6 +321,7 @@ CommonNumber:
     //    Boolean:  The result equals the input argument (no conversion).
     //    Number:   The result equals the input argument (no conversion).
     //    String:   The result equals the input argument (no conversion).
+    //    VariantDate:Returns the value for variant date by calling ToPrimitve directly.
     //    Object:   Return a default value for the Object.
     //              The default value of an object is retrieved by calling the [[DefaultValue]]
     //              internal method of the object, passing the optional hint PreferredType.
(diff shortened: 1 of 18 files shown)


aneeshdk:
@akroshg @suwc @curtisman Can you please take a look?

lib/Runtime/Language/JavascriptConversion.cpp
```diff
@@ -329,6 +329,8 @@ namespace Js
     //----------------------------------------------------------------------------
     Var JavascriptConversion::ToPrimitive(Var aValue, JavascriptHint hint, ScriptContext * requestContext)
     {
+        Var result;
```
suwc: >Var result; [](start = 8, length = 11)

Move this into the block?

lib/Runtime/Language/JavascriptConversion.cpp
```diff
@@ -340,6 +342,15 @@ namespace Js
         case TypeIds_Symbol:
             return aValue;
 
+        case TypeIds_VariantDate:
```
suwc: Add to description?

suwc:
:shipit:

lib/Runtime/Language/JavascriptConversion.cpp
akroshg: Also should initialize to nullptr.

---
In reply to: [86454973](https://github.com/Microsoft/ChakraCore/pull/1911#discussion_r86454973) [](ancestors = 86454973)

curtisman:
:shipit:

pleath:
Change to call ExecuteImplicitCall LGTM.

PR #1914, opened 2016-11-03
Bugfix: Retain slot capacity while type sharing

kunalspathak:
See [9020511](https://microsoft.visualstudio.com//defaultcollection/OS/_workItems?_a=edit&id=9020511&triage=true)

If we are sharing types, we should make sure the old type and new type has same slotCapacity. Added additional assert.

Also add a `DBG` only flag on `DynamicType` that will indicate if the type was cached for ChangePrototype feature. This will make debugging easy especially when one of the cached types are involved in failure in OOP codegen.


kunalspathak:
@pleath, can you please help with the review?

pleath:
LGTM. Thanks.

PR #1915, opened 2016-11-04
Redeferral fixes

pleath:
1. Prevent redeferring during undeferral. 2. Clear inline caches on ScriptFunctionWithInlineCache after FunctionBody has been redeferred. 3. Fix asserts that appeared after Lei's deadlock fixes.


pleath:

PR #1916, opened 2016-11-04
Implement ThreadContextTLSEntry Win+Xplat separately

obastemur:
ChakraDiag depends on the private TLS member of ThreadContextTLSEntry.
Previous implementation was fast for Win32 but affecting XPLAT.

This PR implements ThreadLocal for XPLAT only.

### DIFF
--- a/bin/ChakraCore/ChakraCoreDllFunc.cpp
+++ b/bin/ChakraCore/ChakraCoreDllFunc.cpp
@@ -39,6 +39,11 @@ static ATOM  lockedDll = 0;
 
 static BOOL AttachProcess(HANDLE hmod)
 {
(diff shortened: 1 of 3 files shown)


curtisman:
Yuck.   Can we just change ChakraDiag instead?

obastemur:
> Yuck. Can we just change ChakraDiag instead?

The interface there expects to reach a slot item from cross thread / index + exporting thread_local and reading its value cross thread may not work as good as TLS

curtisman:
Fine for now.  Probably need to figure out a way to access thread_local with DAC or debugger extension in the future.

obastemur:

PR #1920, opened 2016-11-04
swb: strictly check write barrier bit

leirocks:
strictly check where write barrier bit is set for none write barrier pages, in order to check the unnecessary annotation

### DIFF
--- a/lib/Common/ConfigFlagsList.h
+++ b/lib/Common/ConfigFlagsList.h
@@ -1475,6 +1475,9 @@ FLAGNR(Boolean, CFG, "Force enable CFG on jshost. version in the jshost's manife
 
 FLAGR(Number, JITServerIdleTimeout, "Idle timeout in seconds to do the cleanup in JIT server", 10)
 FLAGR(Number, JITServerMaxInactivePageAllocatorCount, "Max inactive page allocators to keep before schedule a cleanup", 10)
+
+FLAGNR(Boolean, StrictWriteBarrierCheck, "Check write barrier setting on none write barrier pages", false)
+
 #undef FLAG_REGOVR_EXP
 #undef FLAG_REGOVR_ASMJS
 
--- a/lib/Common/Exceptions/ReportError.cpp
+++ b/lib/Common/Exceptions/ReportError.cpp
@@ -19,9 +19,6 @@ void ReportFatalException(
     if (IsDebuggerPresent())
     {
         DebugBreak();
-#if DBG
-        return;
-#endif
     }
 
 #ifdef DISABLE_SEH
--- a/lib/Common/Exceptions/ReportError.cpp
+++ b/lib/Common/Exceptions/ReportError.cpp
@@ -19,6 +19,9 @@ void ReportFatalException(
     if (IsDebuggerPresent())
     {
         DebugBreak();
+#if DBG
+        return;
+#endif
     }
 
 #ifdef DISABLE_SEH
--- a/lib/Common/Memory/PageAllocator.cpp
+++ b/lib/Common/Memory/PageAllocator.cpp
@@ -58,20 +58,13 @@ SegmentBase<T>::~SegmentBase()
         GetAllocator()->ReportFree(this->segmentPageCount * AutoSystemInfo::PageSize); //Note: We reported the guard pages free when we decommitted them during segment initialization
 #if defined(_M_X64_OR_ARM64) && defined(RECYCLER_WRITE_BARRIER_BYTE)
 #if ENABLE_DEBUG_CONFIG_OPTIONS
-        if (Js::Configuration::Global.flags.StrictWriteBarrierCheck)
+        if (Js::Configuration::Global.flags.StrictWriteBarrierCheck && this->isWriteBarrierEnabled)
         {
-            if (this->isWriteBarrierEnabled)
-            {
-                RecyclerWriteBarrierManager::OnSegmentFree(this->address, this->segmentPageCount);
-                RecyclerWriteBarrierManager::ToggleBarrier(this->address, this->segmentPageCount * AutoSystemInfo::PageSize, false);
-            }
-        }
-        else
-#endif
-        {
-            RecyclerWriteBarrierManager::OnSegmentFree(this->address, this->segmentPageCount);
+            RecyclerWriteBarrierManager::ToggleBarrier(this->address, this->segmentPageCount * AutoSystemInfo::PageSize, false);
         }
 #endif
+#endif
+        RecyclerWriteBarrierManager::OnSegmentFree(this->address, this->segmentPageCount);
     }
 }
 
--- a/lib/Common/Memory/PageAllocator.cpp
+++ b/lib/Common/Memory/PageAllocator.cpp
@@ -23,7 +23,7 @@ bool SegmentBaseCommon::IsInPreReservedHeapPageAllocator() const
 }
 
 template<typename T>
-SegmentBase<T>::SegmentBase(PageAllocatorBase<T> * allocator, size_t pageCount) :
+SegmentBase<T>::SegmentBase(PageAllocatorBase<T> * allocator, size_t pageCount, bool enableWriteBarrier) :
     SegmentBaseCommon(allocator),
     address(nullptr),
     trailingGuardPageCount(0),
@@ -32,6 +32,7 @@ SegmentBase<T>::SegmentBase(PageAllocatorBase<T> * allocator, size_t pageCount) 
     secondaryAllocator(nullptr)
 #if defined(_M_X64_OR_ARM64) && defined(RECYCLER_WRITE_BARRIER)
     , isWriteBarrierAllowed(false)
+    , isWriteBarrierEnabled(enableWriteBarrier)
 #endif
 {
     this->segmentPageCount = pageCount + secondaryAllocPageCount;
@@ -56,7 +57,20 @@ SegmentBase<T>::~SegmentBase()
         GetAllocator()->GetVirtualAllocator()->Free(originalAddress, GetPageCount() * AutoSystemInfo::PageSize, MEM_RELEASE, this->GetAllocator()->processHandle);
         GetAllocator()->ReportFree(this->segmentPageCount * AutoSystemInfo::PageSize); //Note: We reported the guard pages free when we decommitted them during segment initialization
 #if defined(_M_X64_OR_ARM64) && defined(RECYCLER_WRITE_BARRIER_BYTE)
-        RecyclerWriteBarrierManager::OnSegmentFree(this->address, this->segmentPageCount);
+#if ENABLE_DEBUG_CONFIG_OPTIONS
+        if (Js::Configuration::Global.flags.StrictWriteBarrierCheck)
+        {
+            if (this->isWriteBarrierEnabled)
+            {
+                RecyclerWriteBarrierManager::OnSegmentFree(this->address, this->segmentPageCount);
+                RecyclerWriteBarrierManager::ToggleBarrier(this->address, this->segmentPageCount * AutoSystemInfo::PageSize, false);
+            }
+        }
+        else
+#endif
+        {
+            RecyclerWriteBarrierManager::OnSegmentFree(this->address, this->segmentPageCount);
+        }
 #endif
     }
 }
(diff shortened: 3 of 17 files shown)


lib/Common/ConfigFlagsList.h
```diff
@@ -1492,6 +1492,9 @@ FLAGNR(Boolean, CFG, "Force enable CFG on jshost. version in the jshost's manife
 
 FLAGR(Number, JITServerIdleTimeout, "Idle timeout in seconds to do the cleanup in JIT server", 10)
 FLAGR(Number, JITServerMaxInactivePageAllocatorCount, "Max inactive page allocators to keep before schedule a cleanup", 10)
+
+FLAGNR(Boolean, StrictWriteBarrierCheck, "Check write barrier setting on none write barrier pages", false)
```
jianchun: Do we want this flag available on "Test" build? If yes, change caller RecyclerWriteBarrierManager to use it in "Test" build (DBG -> ENABLE_DEBUG_CONFIG_OPTIONS).

lib/Common/Memory/RecyclerWriteBarrierManager.cpp
```diff
@@ -30,6 +33,7 @@ BYTE* RecyclerWriteBarrierManager::cardTable = RecyclerWriteBarrierManager::x64C
 #else
 // Each byte in the card table covers 4096 bytes so the range covered by the table is 4GB
 BYTE RecyclerWriteBarrierManager::cardTable[1 * 1024 * 1024];
+bool dummy = RecyclerWriteBarrierManager::Initialize();
```
digitalinfinity: Why is this needed?

digitalinfinity:
@leirocks can you provide more details on the purpose of this change? What exactly does "strict barrier check" mean?

lib/Common/Memory/RecyclerWriteBarrierManager.cpp
jianchun: I was looking at this too; I read that it was to initialize the "g_..." function for the other "CopyArray" code to verify "no barrier needed" when the other no barrier variants were used. But this looks heavy weighted and incur penalty for general case -- looks the other code always call this "g_..." function no matter what. I think you should be able to (1) only do the verification in debug/test builds, and (2) forward declare a global function instead of creating this global variable.

---
In reply to: [86642266](https://github.com/Microsoft/ChakraCore/pull/1920#discussion_r86642266) [](ancestors = 86642266)

lib/Common/Memory/Recycler.h
```diff
@@ -2353,6 +2385,7 @@ class ListTypeAllocatorFunc<Recycler, isLeaf>
 // Dummy class to choose the allocation function
 class RecyclerLeafAllocator;
 class RecyclerNonLeafAllocator;
+class RecyclerWriteBarrierAllocator;
```
jianchun: >RecyclerWriteBarrierAllocator [](start = 6, length = 29)

Could you explain why this new allocator type is needed and where it is used? This makes choosing a Recycler allocator more complex.

lib/Common/Memory/RecyclerPointers.h
```diff
@@ -58,8 +59,23 @@ struct AllocatorWriteBarrierPolicy
...
+
 template <class T>
-struct AllocatorWriteBarrierPolicy<RecyclerNonLeafAllocator, T> { typedef _write_barrier_policy Policy; };
+struct AllocatorWriteBarrierPolicy<RecyclerNonLeafAllocator, T> 
+{
+#if GLOBAL_FORCE_USE_WRITE_BARRIER
+    typedef _write_barrier_policy Policy;
+#else
+    typedef _no_write_barrier_policy Policy; 
+#endif
```
jianchun: As offline conversation, a cleaner way might be introducing a 3rd policy variable -- default global policy, which is on by default xplat; off by default Windows,on if globally force. The effective policy would be determined by: allocator type, data type, default global policy (I could work on that after my current annotation batch.)

leirocks:
JC  wants to use the tool to annotate all possible fields, but some of them are not necessary, because some of the struct is never recycler allocated with write barrier. this change will Assert on those address which not allocated with barrier but we are setting barrier bit, so with this we can find out which struct does not need annotate.  this can save a lot of work to annotate the structures instead of understanding usage of each structure and field.

this can also tell us which structure has annotation allocated both with or without barrier , we may specialize it to have both annotated version and non-annotated version



---
In reply to: [258561828](https://github.com/Microsoft/ChakraCore/pull/1920#issuecomment-258561828) [](ancestors = 258561828)

lib/Common/Memory/Recycler.h
leirocks: because not all NonLeaf are allocated with software write barrier, on windows, swb is only enabled in a few class

---
In reply to: [86644009](https://github.com/Microsoft/ChakraCore/pull/1920#discussion_r86644009) [](ancestors = 86644009)

lib/Common/Memory/RecyclerWriteBarrierManager.cpp
leirocks: because of the initialization order, before the card table is allocated, we are already allocating dictionary etc with heap allocator. and in "struct _ArrayWriteBarrier" RecyclerWriteBarrierManager is not availble yet, otherwise we need to change the include and initialization order
the g_ function should only be called when the flag is enabled, I'll fix that part.

---
In reply to: [86643214](https://github.com/Microsoft/ChakraCore/pull/1920#discussion_r86643214) [](ancestors = 86643214,86642266)

lib/Common/ConfigFlagsList.h
leirocks: maybe yes, not sure if there are many difference on the data structures. let me enable it on test build

---
In reply to: [86637769](https://github.com/Microsoft/ChakraCore/pull/1920#discussion_r86637769) [](ancestors = 86637769)

lib/Common/Memory/RecyclerPointers.h
```diff
@@ -109,23 +135,33 @@ void WriteBarrier(T * address, size_t count)
...
+        return CopyArray<Allocator, T, PolicyType>((T*)dst, dstCount, (const T*)src, srcCount);
+    }
+    static void CopyArray(WriteBarrierPtr<T>& dst, size_t dstCount, const WriteBarrierPtr<T>& src, size_t srcCount)
+    {
+        return CopyArray<Allocator, T, PolicyType>((T*)dst, dstCount, (const T*)src, srcCount);
+    }
+};
 
+template <class T, class PolicyType>
+struct WriteBarrierUtil<Recycler, T, PolicyType>
```
leirocks: because function template can't be partial specialized

---
In reply to: [86643736](https://github.com/Microsoft/ChakraCore/pull/1920#discussion_r86643736) [](ancestors = 86643736)

lib/Common/Memory/RecyclerWriteBarrierManager.h
```diff
@@ -124,6 +145,32 @@ class RecyclerWriteBarrierManager
 public:
     static void WriteBarrier(void * address);
     static void WriteBarrier(void * address, size_t bytes);
+    static void ToggleBarrier(void * address, size_t bytes, bool enable);
+    static BOOL IsBarrierAddress(void * address);
+    static BOOL IsBarrierAddress(uintptr_t index);
+    static void VerifyIsBarrierAddress(void * address, size_t bytes);
+    static void VerifyIsNotBarrierAddress(void * address, size_t bytes);
+    static void VerifyIsBarrierAddress(void * address);
+    static bool Initialize();
```
leirocks: the linker should drop them, but yes, should make them only applicable for test build

---
In reply to: [86644302](https://github.com/Microsoft/ChakraCore/pull/1920#discussion_r86644302) [](ancestors = 86644302)

lib/Common/Exceptions/ReportError.cpp
```diff
@@ -19,6 +19,9 @@ void ReportFatalException(
     if (IsDebuggerPresent())
     {
         DebugBreak();
+#if DBG
+        return;
+#endif
```
jianchun: Curious what this change is for? Is it to enable continue execution in debugger after an assertion?

leirocks: that's just easier to debug, forgot to remove this after rebase/merge with my local change

lib/Common/Memory/PageAllocator.cpp
```diff
@@ -56,7 +57,20 @@ SegmentBase<T>::~SegmentBase()
...
+            {
+                RecyclerWriteBarrierManager::OnSegmentFree(this->address, this->segmentPageCount);
+                RecyclerWriteBarrierManager::ToggleBarrier(this->address, this->segmentPageCount * AutoSystemInfo::PageSize, false);
+            }
+        }
+        else
+#endif
+        {
+            RecyclerWriteBarrierManager::OnSegmentFree(this->address, this->segmentPageCount);
+        }
```
jianchun: Looks this fragment can be simplified a bit. Always call "OnSegmentFree". After that, do "#if...if... call ToggleBarrier". Avoids duplicate code "OnSegmentFree".

lib/Common/Memory/PageAllocator.cpp
```diff
@@ -138,17 +152,39 @@ SegmentBase<T>::Initialize(DWORD allocFlags, bool excludeGuardPages)
...
+            // only commit card table for write barrier pages for strict check
+            // we can do this in free build if all write barrier annotated struct only allocate with write barrier pages
+            registerBarrierResult = RecyclerWriteBarrierManager::OnSegmentAlloc(this->address, this->segmentPageCount);
+        }
+    }
+    else
+#endif
+    {
+        registerBarrierResult = RecyclerWriteBarrierManager::OnSegmentAlloc(this->address, this->segmentPageCount);
+    }
```
jianchun: Why "if" and "else" body looks the same?

lib/Common/Memory/PageAllocator.cpp
```diff
@@ -138,17 +152,39 @@ SegmentBase<T>::Initialize(DWORD allocFlags, bool excludeGuardPages)
...
         GetAllocator()->GetVirtualAllocator()->Free(originalAddress, GetPageCount() * AutoSystemInfo::PageSize, MEM_RELEASE, this->GetAllocator()->processHandle);
         this->GetAllocator()->ReportFailure(GetPageCount() * AutoSystemInfo::PageSize);
         this->address = nullptr;
         return false;
     }
-    else
+#endif
+
+    this->isWriteBarrierAllowed = true;
+#if DBG
```
jianchun: >DBG [](start = 4, length = 3)

Should this be ENABLE_DEBUG_CONFIG_OPTIONS?

lib/Common/Memory/Recycler.h
```diff
@@ -1355,6 +1355,7 @@ class Recycler
     DEFINE_RECYCLER_ALLOC(AllocTrackedWithBarrier, ClientTrackableObjectWithBarrierBits);
     DEFINE_RECYCLER_ALLOC(AllocFinalizedClientTrackedWithBarrier, ClientFinalizableObjectWithBarrierBits);
 #endif
+
```
jianchun: nit: spaces

lib/Common/Memory/RecyclerPointers.h
```diff
@@ -220,6 +235,9 @@ class NoWriteBarrierField
     // Setters
     NoWriteBarrierField& operator=(T const& value)
     {
+#if defined(RECYCLER_WRITE_BARRIER)
+        RecyclerWriteBarrierManager::VerifyIsBarrierAddress(this);
```
jianchun: Should this be VerifyIs"Not"? Should this be under #if defined(RECYCLER_WRITE_BARRIER) && ENABLE_DEBUG_CONFIG_OPTIONS, && commandline options?

lib/Common/Memory/RecyclerWriteBarrierManager.cpp
```diff
@@ -20,7 +20,10 @@
 #pragma init_seg(".CRT$XCAU")
 
 #ifdef RECYCLER_WRITE_BARRIER
-
+namespace Memory 
+{
+    FN_VerifyIsNotBarrierAddress* g_verifyIsNotBarrierAddress = nullptr;
+}
```
jianchun: Wrap under #if defined(RECYCLER_WRITE_BARRIER) && ENABLE_DEBUG_CONFIG_OPTIONS

lib/Common/Memory/PageAllocator.cpp
leirocks: yes, since OnSegmentFree do nothing. otherwise if it decommit card table we might still need current pattern, need toggle barrier first though

---
In reply to: [87499422](https://github.com/Microsoft/ChakraCore/pull/1920#discussion_r87499422) [](ancestors = 87499422)

lib/Common/Memory/RecyclerWriteBarrierManager.cpp
```diff
@@ -218,6 +226,10 @@ X64WriteBarrierCardTableManager::Initialize()
 {
     AutoCriticalSection critSec(&_cardTableInitCriticalSection);
 
+#if ENABLE_DEBUG_CONFIG_OPTIONS
+    RecyclerWriteBarrierManager::Initialize();
+#endif
+
```
jianchun: Is there any potential concurrent concern? If so maybe move this after _cardTable initialized?

lib/Common/Memory/RecyclerWriteBarrierManager.h
```diff
@@ -63,12 +76,20 @@ class X64WriteBarrierCardTableManager
...
     BYTE * GetAddressOfCardTable() { return _cardTable; }
 
-#if DBG
+#if ENABLE_DEBUG_CONFIG_OPTIONS
     void AssertWriteToAddress(_In_ void* address)
     {
         Assert(_cardTable);
         Assert(committedSections.Test(GetSectionIndex(address)));
     }
+    BOOLEAN IsCardTableCommited(_In_ uintptr_t index)
```
jianchun: >BOOLEAN [](start = 4, length = 7)

Any reason not to use "bool"? Same applies to others in this file.

leirocks: need to cast then, prefast warns perf impact, it's debug code though, let me change it

---
In reply to: [87501538](https://github.com/Microsoft/ChakraCore/pull/1920#discussion_r87501538) [](ancestors = 87501538)

jianchun:
:shipit:

PR #1923, opened 2016-11-05
WASM - Int64 testing api

### LINKED ISSUE #1886: WASM - assert when printing functions
The following asserts at this time.
```js
print(Wasm.instantiateModule(buffer, ffi).exports.a);
```
Investigate and fix the bug

Cellule:
Added an method on the WebAssembly object to test int64 by passing value as string and converting result as string.
Fixes #1886

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/1923)


Cellule:
I will break down this change in multiple Pull Requests

PR #1924, opened 2016-11-05
Typeof fast path

rajatd:
Adding fast paths for typeof and for branches and value producing compares that use the result of typeof.

1. Setting the value type of the result of typeof as definitely string.

2. Created on the javascriptLibrary, a map of typeid to the string that would be the result of typeof on an object of that typeid and nullptr for typeids for which the result of typeof is not unique or not defined.

3. Lowerer: Opcode Typeof is lowered to emit code to load the resultant string directly from the above map.
    There are object types such as proxies, the result of typeof for which may vary. For such types, I have null entries in the map and the helper for typeof is called in these cases.

4. Since the result of typeof is always a pre-created JavascriptString from the library, branch and compare fast paths can just compare pointers to those strings if both the operands of the branch/compare were results of typeof.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/1924)

### DIFF
--- a/lib/Backend/Lower.cpp
+++ b/lib/Backend/Lower.cpp
@@ -1043,7 +1043,7 @@ Lowerer::LowerRange(IR::Instr *instrStart, IR::Instr *instrEnd, bool defaultDoFa
             break;
 
         case Js::OpCode::Typeof:
-            this->LowerUnaryHelperMem(instr, IR::HelperOp_Typeof);
+            m_lowererMD.LowerTypeof(instr);
             break;
 
         case Js::OpCode::TypeofElem:
@@ -1364,7 +1364,7 @@ Lowerer::LowerRange(IR::Instr *instrStart, IR::Instr *instrEnd, bool defaultDoFa
                 Assert(instr->GetSrc1()->GetType() == instr->GetSrc2()->GetType());
                 this->m_lowererMD.GenerateFastCmXxR8(instr);
             }
-            else if (!PHASE_OFF(Js::BranchFastPathPhase, this->m_func) && !noMathFastPath && this->TryGenerateFastBrOrCmTypeOf(instr, &instrPrev, &fNoLower))
+            else if (!PHASE_OFF(Js::BranchFastPathPhase, this->m_func) && !noMathFastPath && this->TryGenerateFastBrOrCmTypeOf(instr, &instrPrev, false /*isNeqOp*/, &fNoLower))
             {
                 if (!fNoLower)
                 {
@@ -1396,7 +1396,7 @@ Lowerer::LowerRange(IR::Instr *instrStart, IR::Instr *instrEnd, bool defaultDoFa
                 Assert(instr->GetSrc1()->GetType() == instr->GetSrc2()->GetType());
                 this->m_lowererMD.GenerateFastCmXxR8(instr);
             }
-            else if (!PHASE_OFF(Js::BranchFastPathPhase, this->m_func) && !noMathFastPath && this->TryGenerateFastBrOrCmTypeOf(instr, &instrPrev, &fNoLower))
+            else if (!PHASE_OFF(Js::BranchFastPathPhase, this->m_func) && !noMathFastPath && this->TryGenerateFastBrOrCmTypeOf(instr, &instrPrev, true /*isNeqOp*/, &fNoLower))
             {
                 if (!fNoLower)
                 {
@@ -1423,7 +1423,7 @@ Lowerer::LowerRange(IR::Instr *instrStart, IR::Instr *instrEnd, bool defaultDoFa
         case Js::OpCode::CmSrEq_A:
         {
             bool needHelper = true;
-            if (!PHASE_OFF(Js::BranchFastPathPhase, this->m_func) && !noMathFastPath && this->TryGenerateFastBrOrCmTypeOf(instr, &instrPrev, &fNoLower))
+            if (!PHASE_OFF(Js::BranchFastPathPhase, this->m_func) && !noMathFastPath && this->TryGenerateFastBrOrCmTypeOf(instr, &instrPrev, false /*isNeqOp*/, &fNoLower))
             {
                 if (!fNoLower)
                 {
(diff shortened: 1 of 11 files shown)


rajatd:
@LouisLaf @MikeHolman could you guys take a look, please?

lib/Runtime/Library/JavascriptLibrary.cpp
```diff
@@ -3388,6 +3389,92 @@ namespace Js
...
+                break;
+            case TypeIds_SIMDBool8x16:
+                typeDisplayStrings[typeId] = simdBool8x16DisplayString;
+                break;
+#endif
+            case TypeIds_Enumerator:
+            case TypeIds_HostDispatch:
+            case TypeIds_WithScopeObject:
+            case TypeIds_UndeclBlockVar:
+            case TypeIds_Proxy:
```
kunalspathak: >TypeIds_Proxy: [](start = 17, length = 14)

shouldn't "typeof Proxy" === "object" ?

lib/Backend/LowerMDShared.cpp
```diff
@@ -9748,3 +9755,96 @@ LowererMD::LowerDivI4AndBailOnReminder(IR::Instr * instr, IR::LabelInstr * bailO
...
+    GenerateObjectTest(src1, typeOfInstr, taggedIntLabel);
+
+    // MOV typeId, TypeIds_Object
+    // MOV typeRegOpnd, [src1 + offset(Type)]
+    // MOV objTypeId, [typeRegOpnd + offsetof(typeId)]
+    // CMP objTypeId, TypeIds_Limit                                      /*external object test*/
+    // CMOVB typeId, objTypeId
+    // TEST [typeRegOpnd + offsetof(flags)], TypeFlagMask_IsFalsy        /*test for falsy*/
+    // CMOVE typeId, TypeIds_Undefined
+    // MOV dst, typeDisplayStrings[typeId]
```
kunalspathak: >typeDisplayStrings [](start = 16, length = 18)

nit : typeDisplayStringsArray

lib/Runtime/Library/JavascriptLibrary.cpp
rajatd: No, it can be either "object" or "function". [https://github.com/Microsoft/ChakraCore/blob/master/lib/Runtime/Library/JavascriptProxy.cpp#L1544](https://github.com/Microsoft/ChakraCore/blob/master/lib/Runtime/Library/JavascriptProxy.cpp#L1544)

---
In reply to: [86930470](https://github.com/Microsoft/ChakraCore/pull/1924#discussion_r86930470) [](ancestors = 86930470)

lib/Backend/LowerMDShared.cpp
rajatd: Ah, thanks. I had used the field name on the library

---
In reply to: [86930874](https://github.com/Microsoft/ChakraCore/pull/1924#discussion_r86930874) [](ancestors = 86930874)

rajatd:
Also tagging @pleath if he could review this.

lib/Backend/Lower.h
```diff
@@ -474,10 +474,11 @@ class Lowerer
 
     void            RelocateCallDirectToHelperPath(IR::Instr* argoutInlineSpecialized, IR::LabelInstr* labelHelper);
 
-    bool            TryGenerateFastBrOrCmTypeOf(IR::Instr *instr, IR::Instr **prev, bool *pfNoLower);
-    void            GenerateFastBrTypeOf(IR::Instr *branch, IR::RegOpnd *object, IR::IntConstOpnd *typeIdOpnd, IR::Instr *typeOf, bool *pfNoLower);
-    void            GenerateFastCmTypeOf(IR::Instr *compare, IR::RegOpnd *object, IR::IntConstOpnd *typeIdOpnd, IR::Instr *typeOf, bool *pfNoLower);
-    void            GenerateFalsyObjectTest(IR::Instr *insertInstr, IR::RegOpnd *TypeOpnd, Js::TypeId typeIdToCheck, IR::LabelInstr* target, IR::LabelInstr* done, bool isNeqOp);
+    bool            TryGenerateFastBrOrCmTypeOf(IR::Instr *instr, IR::Instr **prev, bool isNeqOp, bool *pfNoLower);
+    void            GenerateFastBrTypeOf(IR::Instr *branch, IR::RegOpnd *object, IR::IntConstOpnd *typeIdOpnd, IR::Instr *typeOf, bool isNeqOp, bool *pfNoLower);
```
kunalspathak: >bool isNeqOp, [](start = 130, length = 13)

nit: move it to last to have it consistent with 
`void            GenerateFalsyObjectTest(IR::Instr *insertInstr, IR::RegOpnd *TypeOpnd, Js::TypeId typeIdToCheck, IR::LabelInstr* target, IR::LabelInstr* done, bool isNeqOp);` ?

lib/Backend/Lower.cpp
```diff
@@ -20570,31 +20570,110 @@ Lowerer::TryGenerateFastBrOrCmTypeOf(IR::Instr *instr, IR::Instr **prev, bool *p
...
+                    InsertMove(instr->GetDst(), LoadLibraryValueOpnd(instr, LibraryValue::ValueTrue), instr);
+
+#if defined(_M_ARM32_OR_ARM64)
+                    IR::LabelInstr * doneLabel = IR::LabelInstr::New(Js::OpCode::Label, instr->m_func);
+                    InsertCompareBranch(instrSrc1, instrSrc2, isNeqOp ? Js::OpCode::BrNeq_A : Js::OpCode::BrEq_A, doneLabel, instr);
+                    InsertMove(instr->GetDst(), LoadLibraryValueOpnd(instr, LibraryValue::ValueFalse), instr);
+                    instr->InsertBefore(doneLabel);
+#else
+                    InsertCompare(instrSrc1, instrSrc2, instr);
+                    m_lowererMD.InsertCmovCC(isNeqOp ? Js::OpCode::CMOVE : Js::OpCode::CMOVNE, instr->GetDst(), LoadLibraryValueOpnd(instr, LibraryValue::ValueFalse), instr);
```
kunalspathak: can you fix the alignment here?

rajatd: >InsertCmovCC [](start = 32, length = 12)

FYI reviewers, I didn't use CMOV for arm because it is more costly than a branch on arm.

lib/Backend/LowerMDShared.cpp
```diff
@@ -9748,3 +9755,96 @@ LowererMD::LowerDivI4AndBailOnReminder(IR::Instr * instr, IR::LabelInstr * bailO
...
+
+    GenerateObjectTest(src1, typeOfInstr, taggedIntLabel);
+
+    // MOV typeId, TypeIds_Object
+    // MOV typeRegOpnd, [src1 + offset(Type)]
+    // MOV objTypeId, [typeRegOpnd + offsetof(typeId)]
+    // CMP objTypeId, TypeIds_Limit                                      /*external object test*/
+    // CMOVB typeId, objTypeId
+    // TEST [typeRegOpnd + offsetof(flags)], TypeFlagMask_IsFalsy        /*test for falsy*/
+    // CMOVE typeId, TypeIds_Undefined
```
kunalspathak: >CMOVE [](start = 7, length = 5)

CMOVNE?

kunalspathak:
:shipit:

PR #1925, opened 2016-11-05
Short circuit IsNativeAddress for Chakra.dll functions when OOP

MikeHolman:
In case all code is not in prereserved code segment or on platforms where prereserved code segment is not available, stack walking requires RPC for each frame to check if an address is JIT code.

If an address is in Chakra, we know it isn't JIT code. A large majority of functions on the stack at any time are interpreter frames, so by checking this we dramatically reduce the frequency of these calls.

Improves Octane CodeLoad perf by 20% on ARM, bringing it back to par with in-proc perf.


satheeshravi:
:shipit:

PR #1928, opened 2016-11-06
Re-enable apply inlining with 1 argument

rajatd:
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/1928)


rajatd:
@satheeshravi please review

test/inlining/applyBugs.baseline
```diff
@@ -0,0 +1,12 @@
...
+PASSED
+
+PASSED
+
+PASSED
+
+PASSED
+1
+1
+1
```
satheeshravi: nit: can you check this value array inside the script itself? - and avoid this baseline file ?

satheeshravi:
:shipit:

PR #1929, opened 2016-11-06
Unpin redeferred functions, etc.

pleath:
1. On redeferral, remove FunctionBody's from the dictionary on the source info. That dictionary is non-leaf and pins the contents; changing it to leaf seems feasible and is recommended by a comment, but it seems best to do it outside this change. 2. Handle re-use of FunctionBody for which we've already allocated literal regexes. 3. Add some rudimentary redeferral tracing, to which I'll add this week.


pleath:

PR #1938, opened 2016-11-07
Enabling StackArgs optimization for strict mode with writes to formals

satheeshravi:
- Strict mode doesn't maintain mapping between named formals and arguments
  object elements.
- All named formals will have ArgIn instructions and are local registers -
  Hence we can let write to formals to happen.

### DIFF
--- a/lib/Runtime/ByteCode/ByteCodeGenerator.cpp
+++ b/lib/Runtime/ByteCode/ByteCodeGenerator.cpp
@@ -2345,16 +2345,15 @@ FuncInfo* PreVisitFunction(ParseNode* pnode, ByteCodeGenerator* byteCodeGenerato
             funcInfo->GetParsedFunctionBody()->SetUsesArgumentsObject(true);
             if (pnode->sxFnc.HasHeapArguments())
             {
-                bool doStackArgsOpt = !pnode->sxFnc.HasAnyWriteToFormals();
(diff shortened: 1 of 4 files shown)


satheeshravi:
@rajatd, @aneeshdk  , can you please take a look ?
//cc: @LouisLaf

rajatd:
:shipit:

lib/Runtime/ByteCode/ByteCodeGenerator.cpp
```diff
@@ -2345,16 +2345,15 @@ FuncInfo* PreVisitFunction(ParseNode* pnode, ByteCodeGenerator* byteCodeGenerato
             funcInfo->GetParsedFunctionBody()->SetUsesArgumentsObject(true);
             if (pnode->sxFnc.HasHeapArguments())
             {
-                bool doStackArgsOpt = !pnode->sxFnc.HasAnyWriteToFormals();
+                bool doStackArgsOpt = (!pnode->sxFnc.HasAnyWriteToFormals() || funcInfo->GetIsStrictMode());
```
kunalspathak: >GetIsStrictMode [](start = 89, length = 15)

Thanks for getting this in. Should see some benefits in node scenarios. BTW, do you mind adding couple of unit test?

kunalspathak:
:shipit:

satheeshravi:
Thanks for reviewing. This pull request got merged, but is still open for some reason.

PR #1940, opened 2016-11-07
WASM- Int64 testing api

Cellule:
Added an method on the WebAssembly object to test int64 by passing value as string and converting result as string.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/1940)


MikeHolman:
:shipit:

PR #1944, opened 2016-11-08
Disabling StackArgs when the source is inside eval expression

satheeshravi:
This is a bug fix - where one of the assumptions was failing as an
assertion failure - This is true only when the source is inside an eval
expression.

Though the fix for this bug is really straight forward, I have decided to
disable the optimization when the source is inside an eval expression - to
minimize risk.

### DIFF
--- a/lib/Runtime/ByteCode/ByteCodeGenerator.cpp
+++ b/lib/Runtime/ByteCode/ByteCodeGenerator.cpp
@@ -2355,14 +2355,22 @@ FuncInfo* PreVisitFunction(ParseNode* pnode, ByteCodeGenerator* byteCodeGenerato
                 //Go conservative if it has any nested functions, or any non-local references.
                 //With statements - need scope object to be present.
                 //Nested funtions - need scope object to be present - LdEnv/LdFrameDisplay needs it.
-                if ((doStackArgsOpt && pnode->sxFnc.funcInfo->GetParamScope()->Count() > 1) && (pnode->sxFnc.funcInfo->HasDeferredChild() ||
+                if ((doStackArgsOpt && pnode->sxFnc.funcInfo->GetParamScope()->Count() > 1) && (pnode->sxFnc.funcInfo->HasDeferredChild() || (byteCodeGenerator->GetFlags() & fscrEval) ||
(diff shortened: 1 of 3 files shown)


satheeshravi:
@aneeshdk @pleath : Can you please take a look ?

pleath:
LGTM. BTW, I assume that we don't attempt stack args when "arguments" appears at global scope in eval?

lib/Runtime/ByteCode/ByteCodeGenerator.cpp
```diff
@@ -2355,14 +2355,22 @@ FuncInfo* PreVisitFunction(ParseNode* pnode, ByteCodeGenerator* byteCodeGenerato
                 //Go conservative if it has any nested functions, or any non-local references.
                 //With statements - need scope object to be present.
                 //Nested funtions - need scope object to be present - LdEnv/LdFrameDisplay needs it.
-                if ((doStackArgsOpt && pnode->sxFnc.funcInfo->GetParamScope()->Count() > 1) && (pnode->sxFnc.funcInfo->HasDeferredChild() ||
+                if ((doStackArgsOpt && pnode->sxFnc.funcInfo->GetParamScope()->Count() > 1) && (pnode->sxFnc.funcInfo->HasDeferredChild() || (byteCodeGenerator->GetFlags() & fscrEval) ||
```
kunalspathak: >byteCodeGenerator [](start = 142, length = 17)

were we hitting the assertion in unit test? If not, can we have a small unit test ?

kunalspathak:
:shipit:

aneeshdk:
LGTM

satheeshravi:
@pleath : Yeah. Just verified. Looks like we are good in the scenario that you mentioned.

lib/Runtime/ByteCode/ByteCodeGenerator.cpp
satheeshravi: Nope. Bug from exprgen. Okay I will add a unit test.

satheeshravi:
@dotnet-bot please test osx_osx_test

satheeshravi:
@dotnet-bot : please test OSX

PR #1945, opened 2016-11-08
Change the bytecodeuse apis to better check for correctness going forward.

Penguinwizzard:
There are a few cases where ByteCodeUses instructions can cause us to fail asserts due to being emitted with operands that are already jitoptimized away. A couple of changes to the api, as well as a few additional asserts, should help identify issues with inserting these much more locally to the issue.

### DIFF
--- a/lib/Backend/IR.cpp
+++ b/lib/Backend/IR.cpp
@@ -892,32 +892,87 @@ ByteCodeUsesInstr::New(Func * func, uint32 offset)
     return byteCodeUses;
 }
 
-void ByteCodeUsesInstr::Set(bool isJITOptimizedReg, uint symId)
+const BVSparse<JitArenaAllocator> * ByteCodeUsesInstr::GetByteCodeUpwardExposedUsed() const
 {
-    Assert(!isJITOptimizedReg);
-    // Although this shouldn't happen, we might as well handle it properly
-    if (isJITOptimizedReg)
-    {
-        return;
-    }
-    if(!byteCodeUpwardExposedUsed)
-    {
-        byteCodeUpwardExposedUsed = JitAnew(m_func->m_alloc, BVSparse<JitArenaAllocator>, m_func->m_alloc);
-    }
-    byteCodeUpwardExposedUsed->Set(symId);
+    return this->byteCodeUpwardExposedUsed;
 }
 
+// In the case of instances where you would like to add a ByteCodeUses to some sym,
+// which doesn't have an operand associated with it (like a block closure sym), use
+// this to set it without needing to pass the check for JIT-Optimized registers.
+void ByteCodeUsesInstr::SetNonOpndSymbol(uint symId)
+{
+    if (!this->byteCodeUpwardExposedUsed)
+    {
+        this->byteCodeUpwardExposedUsed = JitAnew(m_func->m_alloc, BVSparse<JitArenaAllocator>, m_func->m_alloc);
+    }
+    this->byteCodeUpwardExposedUsed->Set(symId);
+}
+
+// In cases where the operand you're working on may be changed between when you get
+// access to it and when you determine that you can set it in the ByteCodeUsesInstr
+// set method, cache the values and use this caller.
+void ByteCodeUsesInstr::SetRemovedOpndSymbol(bool isJITOptimizedReg, uint symId)
+{
+    if (isJITOptimizedReg)
+    {
+        AssertMsg(false, "Tried to add a jit-optimized register to a ByteCodeUses instruction!");
+        // Although we assert on debug builds, we should actually be ok with release builds
+        // if we ignore the operand; not ignoring it, however, can cause us to introduce an
+        // inconsistency in bytecode register lifetimes.
+        return;
+    }
+    if(!this->byteCodeUpwardExposedUsed)
+    {
+        this->byteCodeUpwardExposedUsed = JitAnew(m_func->m_alloc, BVSparse<JitArenaAllocator>, m_func->m_alloc);
+    }
+    this->byteCodeUpwardExposedUsed->Set(symId);
+}
+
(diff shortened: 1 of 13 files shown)


Penguinwizzard:
@satheeshravi please review

lib/Backend/IR.h
```diff
@@ -503,13 +504,19 @@ class Instr
...
-    static ByteCodeUsesInstr * New(Func * func);
-    static ByteCodeUsesInstr* New(IR::Instr* originalBytecodeInstr, SymID symid);
+private:
     BVSparse<JitArenaAllocator> * byteCodeUpwardExposedUsed;
-    PropertySym *              propertySymUse;
+    
+public:
+    static ByteCodeUsesInstr * New(IR::Instr * originalBytecodeInstr);
+    static ByteCodeUsesInstr * New(Func * originalBytecodeInstr, uint32 offset);
+    const BVSparse<JitArenaAllocator> * getByteCodeUpwardExposedUsed() const { return byteCodeUpwardExposedUsed; }
```
satheeshravi: >getByteCodeUpwardExposedUsed [](start = 40, length = 28)

nit: Getxxx

lib/Backend/IR.h
```diff
@@ -503,13 +504,19 @@ class Instr
...
 
-    void Set(uint symId);
+    PropertySym *              propertySymUse;
+    void Set(bool isjitoptimizedreg, uint symId);
+    void Clear(uint symId) { Assert(byteCodeUpwardExposedUsed != nullptr); byteCodeUpwardExposedUsed->Clear(symId); };
```
satheeshravi: >{ Assert(byteCodeUpwardExposedUsed != nullptr); byteCodeUpwardExposedUsed->Clear(symId); } [](start = 27, length = 90)

nit: definitions in cpp ?

lib/Backend/IR.h
```diff
@@ -503,13 +504,19 @@ class Instr
...
-public:
-    static ByteCodeUsesInstr * New(Func * func);
-    static ByteCodeUsesInstr* New(IR::Instr* originalBytecodeInstr, SymID symid);
+private:
     BVSparse<JitArenaAllocator> * byteCodeUpwardExposedUsed;
-    PropertySym *              propertySymUse;
+    
+public:
+    static ByteCodeUsesInstr * New(IR::Instr * originalBytecodeInstr);
+    static ByteCodeUsesInstr * New(Func * originalBytecodeInstr, uint32 offset);
```
satheeshravi: >originalBytecodeInstr [](start = 42, length = 21)

func?

lib/Backend/IR.cpp
```diff
@@ -870,37 +870,73 @@ ProfiledInstr::CopyProfiledInstr() const
...
+        {
+            if (this->byteCodeUpwardExposedUsed)
+            {
+                this->byteCodeUpwardExposedUsed->Or(target->byteCodeUpwardExposedUsed);
+                JitAdelete(target->byteCodeUpwardExposedUsed->GetAllocator(), target->byteCodeUpwardExposedUsed);
+                target->byteCodeUpwardExposedUsed = nullptr;
+            }
+            else
+            {
+                Assert(this->m_func == target->m_func);
```
satheeshravi: >Assert(this->m_func == target->m_func); [](start = 16, length = 39)

you can do this assert outside if-else

lib/Backend/IR.cpp
```diff
@@ -870,37 +870,73 @@ ProfiledInstr::CopyProfiledInstr() const
...
     byteCodeUses->propertySymUse = nullptr;
+    byteCodeUses->SetByteCodeOffset(offset);
     return byteCodeUses;
 }
 
-void ByteCodeUsesInstr::Set(uint symId)
+void ByteCodeUsesInstr::Set(bool isJITOptimizedReg, uint symId)
 {
+    Assert(!isJITOptimizedReg);
+    // Although this shouldn't happen, we might as well handle it properly
```
satheeshravi: >Although this shouldn't happen, we might as well handle it properly [](start = 7, length = 67)

this is cryptic - can you pls add some more comments.

lib/Backend/IR.h
```diff
@@ -503,13 +504,19 @@ class Instr
...
+    const BVSparse<JitArenaAllocator> * getByteCodeUpwardExposedUsed() const { return byteCodeUpwardExposedUsed; }
 
-    void Set(uint symId);
+    PropertySym *              propertySymUse;
+    void Set(bool isjitoptimizedreg, uint symId);
```
satheeshravi: >bool isjitoptimizedreg, uint symId [](start = 13, length = 34)

nit: you can actually consider getting just the Opnd* as argument - and then you can look up the symId and the isJitOptReg fields.

satheeshravi:
:shipit:

lib/Backend/IR.h
Penguinwizzard: Fixed.

lib/Backend/IR.h
Penguinwizzard: I've moved it to the cpp.

lib/Backend/IR.h
Penguinwizzard: Fixed.

lib/Backend/IR.cpp
Penguinwizzard: I've moved it outside of both layers of ifs.

lib/Backend/IR.h
Penguinwizzard: I've added two additional set methods; one takes an Opnd* as an argument, and the other just takes a symId (for cases where I was previously just passing "false" statically into the set method). I've gone through and used these where applicable, and added comments such that intellisense/other autocompletion will provide additional information.

PR #1956, opened 2016-11-08
Fixing async function instance intialization issue

aneeshdk:
When doing a type conversion for async function we were passing the wrong
type handler.


aneeshdk:
@akroshg @tcare @pleath Can you please take a look?

akroshg:
:shipit:

tcare:
Looks good

PR #1960, opened 2016-11-09
Implementing f32.convert_u/i32

### LINKED ISSUE #1967: WASM - f32.convert_u/i32 bugs
The following tests are failing in jitted code
`test/WasmSpec/testsuite/conversions.wast` lines 266 & 268
`test/WasmSpec/testsuite/float_exprs.wast` lines 493 & 698-700

Krovatkin:
This PR implements f32.convert_u/i32 for x86-32/64

Fixes #1967


Krovatkin:
@MikeHolman Encoder doesn't seem to respect a source operand's type for `CVTSI2SS` (i.e. it uses 32-bit registers no matter what). Do we want to extend Encoder or would be okay with this sequence?

MikeHolman:
Let's fix the encoder.

Krovatkin:
@MikeHolman or would you prefer we  

* either add a property `REXSRC` to OpcodesMD.h and add this extra check to `EmitRexByte(prexByte, rexByte, skipRexByte || (instrSize < 8 && !(opdope & REXSRC)), reservedRexByte)`
* introduce another function which takes an opcode and tells us if we should override `skipRexByte`? I.e. `EmitRexByte(prexByte, rexByte, skipRexByte || (instrSize < 8 && skipRexByteIfSizeLt8(instr->m_opcode)), reservedRexByte)`

It is a bit difficult to figure out which one is better since Encoder is already heavily special-cased :-)

Cellule:
I think it would be simpler to do
```cpp
  //  lib\Backend\amd64\LowererMDArch.cpp LowererMDArch::EmitUIntToFloat

    // CVTSI2SD dst, tempReg.i64  (Use the tempreg as if it is 64 bit without sign extension)
    instrInsert->InsertBefore(IR::Instr::New(dst->IsFloat64() ? Js::OpCode::CVTSI2SD : Js::OpCode::CVTSI2SS, dst->UseWithNewType(TyFloat64, m_func),
        tempReg->UseWithNewType(TyInt64, this->m_func), this->m_func));
```
This is tricking our encoder in thinking this use of `CVTSI2SS` is 64bits (which is technically true) and avoid a special case in EncoderMD

Cellule:
Also please re-enable the tests disabled from #1967. 
This will help see that it fixes the bug also.
You will probably have to rebase on master before that

MikeHolman:
Discussed this a little earlier with @Cellule. I think best thing is to add flag to opdope to tell us that we should look at src1 to know if we need rex prefix, and then in the switch where we do `rexByte    |= this->GetRexByte(this->REXR, opr1);`, pass opr2 in case the flag is set.

I think flag should be added for CVTSI2SD and CVTSI2SS.

Krovatkin:
@Cellule 

> This is tricking our encoder in thinking this use of CVTSI2SS is 64bits (which is technically true) and avoid a special case in EncoderMD

This is exactly what the original sequence was using. Unfortunately, Encoder bluntly disregards `src1`'s size and skips a rex byte. We **can**  trick Encoder into doing the **Right Thing™** by running `CVTSI2SD` and then `CVTSD2SS`, but @MikeHolman indicated he would rather have this issue fixed in Encoder -- a sentiment I can easily relate to, since it is only a matter of time before somebody else introduces another use of CVTSD2SS with a 64-bit source operand and Encoder won't do the **Right Thing™*  🤕  

@MikeHolman 

> Discussed this a little earlier with @Cellule. I think best thing is to add flag to opdope to tell us that we should look at src1 to know if we need rex prefix, and then in the switch where we do rexByte |= this->GetRexByte(this->REXR, opr1);, pass opr2 in case the flag is set.

>I think flag should be added for CVTSI2SD and CVTSI2SS.

Nice! This is a very elegant crystallization of my morning ramblings that we should somehow use a new flag to solve this issue :-) 

 @Cellule 
 > Also please re-enable the tests disabled from #1967. 
This will help see that it fixes the bug also.
You will probably have to rebase on master before that

Sir, yes, sir!!!

Krovatkin:
@MikeHolman Unfortunately, I couldn't enable a new flag for `CVTSI2SD` yet, since I ran into a bunch of test failures. We could start with delivering `CVTSI2SS` and I'll enable `CVTSI2SD` as soon as I fix the test failures?

MikeHolman:
That sounds good to me.

Cellule:
Reviewed 6 of 7 files at r1, 1 of 2 files at r2, 18 of 24 files at r3.
Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[lib/Backend/amd64/EncoderMD.cpp, line 862 at r3](https://reviewable.io:443/reviews/microsoft/chakracore/1960#-KWG2syuAMQDDg5uzRFf:-KWG2syuAMQDDg5uzRFg:b-9ex55l) ([raw file](https://github.com/microsoft/chakracore/blob/ccfa6581bd724b82389fe22b6aa8a0362c78a3cb/lib/Backend/amd64/EncoderMD.cpp#L862)):*
> ```cpp
>             else if (opr1->IsRegOpnd())
>             {
>                 rexByte |= this->GetRexByte(this->REXR, opdope & DREXSRC ? opr2 : opr1);
> ```

it is possible for opr2 is not a RegOpnd is it really needed here if we already checked the instrSize above ?

---

*[lib/Backend/amd64/MdOpCodes.h, line 76 at r3](https://reviewable.io:443/reviews/microsoft/chakracore/1960#-KWG33WktAz9JvhYp7YZ:-KWG33WktAz9JvhYp7Y_:btb2gc2) ([raw file](https://github.com/microsoft/chakracore/blob/ccfa6581bd724b82389fe22b6aa8a0362c78a3cb/lib/Backend/amd64/MdOpCodes.h#L76)):*
> CVTSI2SD

This needs the flag as well, we don't need it now, but might as well fix it now

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/microsoft/chakracore/1960)*

Krovatkin:
@Cellule 

> This needs the flag as well, we don't need it now, but might as well fix it now

Unfortunately, adding the flag to `CVTSI2SD` breaks quite a few tests. We are a little bit short on time to fix all the failures.

> it is possible for opr2 is not a RegOpnd is it really needed here if we already checked the instrSize above ?

Yes, good catch! @MikeHolman Could you please advise why you were suggesting to add this `rexByte |= this->GetRexByte(this->REXR, opr1);` ? It looks like we seem to be doing the right thing already. If `opr1` is a register **GetRexByte** adds **REX.R** and if `opr2` is a register too **EmitModRM** adds **REX.B**  That piece of logic seems to focus on whether `R` or `B` bytes are needed rather than whether the REX prefix is needed. I'm a bit fuzzy about the case where instruction size is 4 bytes in which case Encoder will **skip** the REX byte, even though one of the operands **can** be an extended registers (e.g. r8 to r15)? Though, my change isn't making this case any worse

Krovatkin:

Cellule:
I think we should remove the change to ` rexByte |= this->GetRexByte(this->REXR, opdope & DREXSRC ? opr2 : opr1);`.
The first part should be enough.

Krovatkin:
@Cellule Yup, I rolled back that line. I wonder what @MikeHolman thinks; after all he suggested that particular change 😖

Krovatkin:
@Cellule Correction. I *thought* I had rolled back that particular line 😖 Let me update the PR

Cellule:
@Krovatkin I still see it in the latest version
I talked with @MikeHolman and we determined that we should remove it. I was the one to initially suggest that to Michael and then realized after that it wasn't an optimal solution and it wasn't enough

Krovatkin:

Krovatkin:
@dotnet-bot test this please

Cellule:
:lgtm:

---

Reviewed 2 of 3 files at r4, 1 of 1 files at r5.
Review status: all files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/microsoft/chakracore/1960#-:-KWKRfw_uFXRPEMJci_X:bnfp4nl)*

Cellule:
We seem to have a problem with jenkins, I just kicked off a test build on VSO. If it comes back green I will merge this

Krovatkin:
@Cellule Thanks a lot!

PR #1962, opened 2016-11-09
update jenkins test names

obastemur:
Fixes https://github.com/Microsoft/ChakraCore/issues/1933 and `no-jit` test names

### DIFF
--- a/netci.groovy
+++ b/netci.groovy
@@ -121,10 +121,11 @@ def CreateXPlatBuildTask = { isPR, buildType, staticBuild, machine, platform, co
     def numConcurrentCommand = (platform == "osx" ? "sysctl -n hw.logicalcpu" : "nproc")
 
     config = (configTag == null) ? config : "${configTag}_${config}"
-    config = staticBuild ? "${config}_static" : config
+    config = staticBuild ? "static_${config}" : "shared_${config}"
+    config = customOption ? customOption.replaceAll(/[-]+/, "_") + "_" + config : config
 
     // params: Project, BaseTaskName, IsPullRequest (appends '_prtest')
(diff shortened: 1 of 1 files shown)


netci.groovy
```diff
@@ -121,10 +121,11 @@ def CreateXPlatBuildTask = { isPR, buildType, staticBuild, machine, platform, co
     def numConcurrentCommand = (platform == "osx" ? "sysctl -n hw.logicalcpu" : "nproc")
 
     config = (configTag == null) ? config : "${configTag}_${config}"
-    config = staticBuild ? "${config}_static" : config
+    config = staticBuild ? "static_${config}" : "shared_${config}"
+    config = customOption ? customOption.replaceAll(/[-]+/, "_") + "_" + config : config
 
     // params: Project, BaseTaskName, IsPullRequest (appends '_prtest')
-    def jobName = Utilities.getFullJobName(project, config, isPR) + customOption.replaceAll(/[-]+/, "_")
+    def jobName = Utilities.getFullJobName(project, config, isPR)
```
dilijev: I like this change, it will make the names so that `_prtest` always appears at the end. Are we running any such tests and/or reporting them in the wiki right now?

obastemur:
@dilijev dont we have to update README too ? or we should update that in a separate PR?

dilijev:
Yes, you'll need to update the readme and the wiki. If you have the time to do that now, go ahead. Keep in mind only to update the README in the affected branches and only the Build Status tables in the wiki that are affected. Older branches will not be affected by this change.

dilijev:
@obastemur ping, you can merge this when ready

dilijev:
@obastemur ping

dilijev:
@obastemur Any update here? Should we retarget this to latest master and get it merged in?

obastemur:

PR #1963, opened 2016-11-09
xplat improvements

obastemur:
- remove unused functions and exclude an SEH only function
- enable stack test on osx


jianchun:
:shipit:

obastemur:

PR #1964, opened 2016-11-09
xplat: enable optimize sibling calls and omit**frame-pointer + force O3

obastemur:
additional ~4% performance gain (octane and load test)

The reason for some of the flags there because of old stability issues.

Besides, instead of disabling tail call optimization globally,
disabled only for functions with noinline attribute. Older Clang doesn't support this.


jianchun:
:shipit:

obastemur:

PR #1965, opened 2016-11-09
Add wasm wrap support

meg-gupta:


meg-gupta:
Investigating a unit test failure where extend and wrap are folded violating what spec expects.

meg-gupta:
@Cellule Can you please review ?

Cellule:
Reviewed 19 of 19 files at r1.
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[lib/Backend/i386/LowererMDArch.cpp, line 2593 at r1](https://reviewable.io:443/reviews/microsoft/chakracore/1965#-KWAmbHqizkl_X21BIua:-KWAmbHqizkl_X21BIub:bgyorxe) ([raw file](https://github.com/microsoft/chakracore/blob/c08cc16b86b32e94bbe917395968faef6489e23c/lib/Backend/i386/LowererMDArch.cpp#L2593)):*
> ```cpp
> 
>     Int64RegPair srcPair = lowererMD->m_lowerer->FindOrCreateInt64Pair(src);
>     instrInsert->InsertBefore(IR::Instr::New(Js::OpCode::MOV, dst, srcPair.low, this->m_func));
> ```

Please add `mov high, 0`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/microsoft/chakracore/1965)*

meg-gupta:
I thought about this, moving 0 to srcPair.high can destroy the upper bits, which may be needed in  a subsequent instruction using the entire 64bit value ?

Cellule:
Well that's the point, if we use it afterward it needs to be zero'd.
If we actually want to use the original int64 non wrapped value, then it will be a different operand anyway.

meg-gupta:
FindOrCreateInt64Pair will return the same regpair for subsequent use of the operand as 64bit.
Something like the below test will print wrong result for the second print, because we thrash the value of upper bits.
  (func (export "test") (param $x i64)
    (call $print32 (i32.wrap/i64 (get_local $x)))
    (call $print64 (get_local $x))

Or are we expected to thrash the upper 32 bits ?

PR #1966, opened 2016-11-09
BailOnNoProfile improvements

rajatd:
1. Record in the profile that we loaded a HostDispatch object from a LdFld or LdSlot. This is to not let the IRBuilder think that the respective LdFld or LdSlot was non-profiled (IRBuilder would then insert a BailOnNoProfile before the operation and we'll bailout but not collect any new information, and we'll put the bailout in the next rejit too, and so on). I'm not recording anything specific for HostDispatch, but just that we've seen an object at that point.

2. While hoisting BailOnNoProfile from successors to predecessors, dont remove the bailouts from the successors as we may not end up moving the bailout all the way to the top of the predecessor and would just remove it then. If the bailouts were removed from the successors, we wouldn't have BailOnNoProfile on paths that should have it.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/1966)

### DIFF
--- a/lib/Backend/BackwardPass.cpp
+++ b/lib/Backend/BackwardPass.cpp
@@ -7055,7 +7055,7 @@ BackwardPass::ProcessBailOnNoProfile(IR::Instr *instr, BasicBlock *block)
         curInstr = curInstr->m_prev;
     }
 
-    // Didn't get to the top of the block, delete this BailOnNoProfile...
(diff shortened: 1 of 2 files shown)


rajatd:
@LouisLaf , @pleath , @satheeshravi could either of you please review this?

satheeshravi:
:shipit:

rajatd:
Thanks @satheeshravi

Issue #757, opened 2016-04-07
Array.of should throw when it sets a non-configurable property


nicolo-ribaudo:
``` js
function Arr() {
    Object.defineProperty(this, "0", {
        configurable: false
    });
}

try {
    const arr = Array.of.call(Arr, "a");
    print(JSON.stringify(arr));
} catch (error) {
    print(error);
}
```

Expected result:

```
TypeError: Cannot modify non-writable property '0'
```

Actual result:

```
{"length":1}
```

Array.of should throw a TypeError when it tries to set Arr#0, because it isn't configurable.

Specification:

```
22.1.2.3 Array.of ( ...items )
    ...
    6. Let k be 0.
    7. Repeat, while k < len
        a. Let kValue be items[k].
        b. Let Pk be ! ToString(k).
        c. Perform ? CreateDataPropertyOrThrow(A, Pk, kValue). // 'A' is 'new this'
        d. Increase k by 1.
    ...

7.3.6 CreateDataPropertyOrThrow (O, P, V)
    ...
    3. Let success be ? CreateDataProperty(O, P, V).
    4. If success is false, throw a TypeError exception.
    5. Return success.

7.3.4 CreateDataProperty (O, P, V)
    ...
    3. Let newDesc be the PropertyDescriptor{[[Value]]: V, [[Writable]]: true, [[Enumerable]]: true, [[Configurable]]: true}.
    4. Return ? O.[[DefineOwnProperty]](P, newDesc).

9.1.16 [[DefineOwnProperty]] (P, Desc)
    1. Return ? OrdinaryDefineOwnProperty(O, P, Desc).

9.1.6.1 OrdinaryDefineOwnProperty (O, P, Desc)
    1. Let current be ? O.[[GetOwnProperty]](P).
    2. Let extensible be the value of the [[Extensible]] internal slot of O.
    3. Return ValidateAndApplyPropertyDescriptor(O, P, extensible, Desc, current).

9.1.6.3 ValidateAndApplyPropertyDescriptor (O, P, extensible, Desc, current)
    1. Assert: If O is not undefined, then IsPropertyKey(P) is true.
    2. If current is undefined, then
        ...
    3. Return true, if every field in Desc is absent.
    4. Return true, if every field in Desc also occurs in current and the value of every field in Desc is the same value as the corresponding field in current when compared using the SameValue algorithm.
    5. If the [[Configurable]] field of current is false, then
        a. Return false, if the [[Configurable]] field of Desc is true.
[...]

PR #1969, opened 2016-11-09
Creating property on non-config slot should throw type error.

akroshg:
Based on the spec, when we try to set property on the non-configurable property slot should throw type error. Fixed that multiple places.

### DIFF
--- a/lib/Parser/rterrors.h
+++ b/lib/Parser/rterrors.h
@@ -380,3 +380,4 @@ RT_ERROR_MSG(WASMERR_NeedWebAssemblyFunc, 5679, "%s is not a WebAssembly exporte
 RT_ERROR_MSG(WASMERR_SignatureMismatch, 5680, "%s called with invalid signature", "Function called with invalid signature", kjstWebAssemblyRuntimeError, 0)
 RT_ERROR_MSG(WASMERR_ElementSegOutOfRange, 5681, "", "Element segment is out of range", kjstTypeError, 0)
 RT_ERROR_MSG(WASMERR_TableIndexOutOfRange, 5682, "", "Table index is out of range", kjstWebAssemblyRuntimeError, 0)
+RT_ERROR_MSG(JSERR_CantRedefineProp, 5683, "Cannot redefine property '%s'", "Cannot redefine property", kjstTypeError, 0)
--- a/lib/Runtime/Library/JavascriptArray.cpp
+++ b/lib/Runtime/Library/JavascriptArray.cpp
@@ -2901,6 +2901,23 @@ namespace Js
         return  JavascriptNumber::ToVar(idxDest, scriptContext);
     }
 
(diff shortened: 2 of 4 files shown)


akroshg:
@suwc - please have a look.

akroshg:
fixes #757.

suwc:
:shipit:

suwc:
LGTM.

lib/Parser/rterrors.h
```diff
@@ -374,3 +374,4 @@ RT_ERROR_MSG(WASMERR_InvalidGlobalRef, 5673, "", "Global initialization does not
 RT_ERROR_MSG(WASMERR_NeedMemoryObject, 5674, "%s is not a WebAssembly.Memory", "WebAssembly.Memory object expected", kjstTypeError, 0)
 RT_ERROR_MSG(WASMERR_InvalidTypeConversion, 5675, "Invalid WebAssembly type conversion %s to %s", "Invalid WebAssembly type conversion", kjstTypeError, 0)
 RT_ERROR_MSG(WASMERR_DivideByZero, 5676,    "",  "Division by zero", kjstError, 0)
+RT_ERROR_MSG(JSERR_CantRedefineProp, 5677, "Cannot redefine property '%s'", "Cannot redefine property", kjstTypeError, 0)
```
agarwal-sandeep: >JSERR_CantRedefineProp [](start = 13, length = 22)

How is this different from error message JSERR_DefineProperty_NotConfigurable (Cannot redefine non-configurable property). Are there cases when we hit this error and the property is configurable?

lib/Runtime/Library/JavascriptArray.cpp
```diff
@@ -3032,7 +3040,7 @@ namespace Js
                             }
                             else
                             {
-                                SetArrayLikeObjects(pDestObj, idxDest, subItem);
+                                ThrowErrorOnFailure(SetArrayLikeObjects(pDestObj, idxDest, subItem), scriptContext, idxSubItem);
```
agarwal-sandeep: >idxSubItem [](start = 116, length = 10)

Should this be idxDest? if not can we put a comment explaining why its idxSubItem as we are setting the element at idxDest.

lib/Parser/rterrors.h
akroshg: That was my first choice as well, but I am trying to match the wording with other browser. Also SetItem can fail for other reasons, so it is better to be not-specific to configurable only.

---
In reply to: [87445471](https://github.com/Microsoft/ChakraCore/pull/1969#discussion_r87445471) [](ancestors = 87445471)

lib/Runtime/Library/JavascriptArray.cpp
akroshg: It should be idxDest, nice catch. will fix that.

---
In reply to: [87446064](https://github.com/Microsoft/ChakraCore/pull/1969#discussion_r87446064) [](ancestors = 87446064)

PR #1970, opened 2016-11-09
fix incorrect calculation in IsChakraAddress

MikeHolman:
Previous code was incorrectly doing comparison of addr < base + chakraSize * sizeof(void*)

When debugging, I noticed that we already have code to do this, so my fix is to remove my methods and use the existing ones.


satheeshravi:
:shipit:

PR #1971, opened 2016-11-09
Memory leak fixes

kunalspathak:
This PR includes some memory leak fixes.
* Change `FunctionBodyDictionary` and `DeferredFunctionsDictionary` on `Utf8Sourceinfo` to use `LeafValueDictonary`. This will guarantee that references of functionbody stored in these dictionaries does not hold back those function bodies from getting collected.
* Change `TypeTransitionMap` used in type sharing for ChangePrototype to use `WeakReferenceDictionary` so the `DynamicType` references stored in the map does not hold back those types from getting collected. Also added heuristic to periodically remove entries from this map for types that are already collected.


lib/Runtime/Types/PathTypeHandler.cpp
```diff
@@ -1537,24 +1538,52 @@ namespace Js
             Assert(oldTypeToPromotedTypeMap && (Js::Var)oldTypeToPromotedTypeMap != scriptContext->GetLibrary()->GetUndefined());
             oldTypeToPromotedTypeMap = reinterpret_cast<TypeTransitionMap*>(oldTypeToPromotedTypeMap);
 
-            if (oldTypeToPromotedTypeMap->TryGetValue(reinterpret_cast<uintptr_t>(oldType), &cachedDynamicType))
+            RecyclerWeakReference<DynamicType>* weakRefCachedtype;
+            
+            if (oldTypeToPromotedTypeMap->TryGetValue(taggedOldType, &weakRefCachedtype))
             {
+                cachedDynamicType = weakRefCachedtype->Get();
```
digitalinfinity: Can there ever be a case where cachedDynamicType got reclaimed, and then got reallocated, and we now go down this codepath thinking the type is there when in fact its not valid?

kunalspathak: No. When `cachedDynamictype` gets reclaimed, the `strongRef` field of weakReferenceHandle will be set to `nullptr` and will not be reset with reallocation. `Get()` will return `nullptr` if that is the case.
However I made a repro where other way round could happen. `oldType` is cached in dictionary (with tag) and it gets collected and re-allocated with new type. We come down this path and see that `oldType` is present in the map and extracts the `cachedDynamicType` which is wrong.

lib/Runtime/Types/PathTypeHandler.cpp
```diff
@@ -1521,6 +1521,7 @@ namespace Js
         DynamicType* cachedDynamicType = nullptr;
         BOOL isJsrtExternalType = instance->GetType()->IsJsrtExternal();
         DynamicType* oldType = isJsrtExternalType ? 0 : instance->GetDynamicType();
```
kunalspathak: `oldType` will not get collected because `instance` is holding on to it. We replace it with `cachedDynamicType` at the end of this method. But as I mentioned in the other comment, `oldType` can get reclaimed and re-allocated after this function in which case we will get wrong `cachedDynamicType`.

kunalspathak:
@digitalinfinity - Updated to have a weak reference on oldType and strong reference on newType. `WeaklyReferencedKeyDictionary` takes care of cleaning up the entries if keys got collected so reverted the heuristics I added previously.

cc :  @pleath

lib/Runtime/Types/PathTypeHandler.cpp
```diff
@@ -1520,7 +1520,11 @@ namespace Js
         ScriptContext* scriptContext = instance->GetScriptContext();
         DynamicType* cachedDynamicType = nullptr;
         BOOL isJsrtExternalType = instance->GetType()->IsJsrtExternal();
-        DynamicType* oldType = isJsrtExternalType ? 0 : instance->GetDynamicType();
+        // Share the type of all JsrtExternal objects which will have 'newPrototype' as their new prototype.
+        // TypeTransitionMap is a map with key as weakReference and to store it, key need to have a valid heap block.
+        // So instead of using some constant, use newPrototype as the key for saving newType for such objects
+        // in TypeTransitionMap.
+        DynamicType* oldType = isJsrtExternalType ? reinterpret_cast<DynamicType*>(newPrototype) : instance->GetDynamicType();
```
digitalinfinity: I don't understand this bit but the rest of the change is fine. Why can't you use newPrototype->GetType() here to use a real type instead of pretending like an object is a type- seems dangerous. But I'll defer to @curtisman/@pleath here

kunalspathak: > Why can't you use newPrototype->GetType() 

I thought of using `newPrototype->GetType()`, but type of `newPrototype` can change and we will end up creating new entry in the map.

kunalspathak:
@dotnet-bot test Ubuntu ubuntu_linux_release_static please.

lib/Runtime/Types/PathTypeHandler.cpp
curtisman: Can you explain how you can share the same type for all JsrtExternal objects?  what if the current JsrtExternalObject has some properties already? should the newType with the newPrototype be necessary be different then one that is for JsrtExternalObject that has no property?

---
In reply to: [87488710](https://github.com/Microsoft/ChakraCore/pull/1971#discussion_r87488710) [](ancestors = 87488710)

curtisman: Can you explain how you can share the same type for all JsrtExternal objects?  what if the current JsrtExternalObject has some properties already? should the newType with the newPrototype be necessary be different then one that is for JsrtExternalObject that has no property?

kunalspathak: I had this concern when I initially implemented it but I didn't find a reason to have separate types for scenarios I was optimizing it for. I will revisit this and see if I can come up with better fix, if needed. Note that this PR doesn't change the original design though except that the key that was being used for `JsrtExternalObject` has been changed  from `0`  to `newPrototype`.

curtisman: That would be a bug and needs to fix, no?

kunalspathak: True. I have already started working on it.
BTW, we always create a new `JsrtExternalType` for a given `JsrtExternalObject`. If I start caching it in the map of `newPrototype`, we will end up having lot of entries in the map for these types. I am experimenting something for `JsrtExternalType` on similar lines as caching in `CreateObjectType()`. I will update this PR with the fix.

kunalspathak:

kunalspathak:
@dotnet-bot please test Windows arm_release

pleath:
Can you merge with yesterday's redeferral fixes and do a new crawler/exprgen run?

pleath:
LGTM, though you probably want to involve a Jsrt expert as well.

kunalspathak:
> Can you merge with yesterday's redeferral fixes and do a new crawler/exprgen run?

That's the plan. I have started building binaries and once they are ready, will start the run.

kunalspathak:
@jianchun , @digitalinfinity , @curtisman - can you help review the code?

digitalinfinity:
:shipit:

PR #1975, opened 2016-11-10
A rejit/rethunk change

rajatd:
If we are about to rejit a function body and the last rejit was caused by a bailout at the same bytecode offset as the current one, then its a fair assumption that if there is a newer entrypoint available, the problem that is causing the rejit would have been fixed in that entry point. So, rethunk to that entry point instead of rejitting.

This specially helps with bailouts that don't disable an optimization on rejit but rely on updated profile info to fix the problem.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/1975)


rajatd:
@LouisLaf @pleath @curtisman could one of you please review this?

LouisLaf:
:shipit:

PR #1978, opened 2016-11-10
WASM - import thunk name fix

Cellule:
Correctly set the buffer length when generating the Wasm Import's thunk name + some error handling

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/1978)


MikeHolman:
lgtm

PR #1979, opened 2016-11-10
Change to address CVE-2016-7207

pleath:
Fix incorrect identification of non-local-references when parsing arrow function parameter lists. Since we can't identify such syntax on the initial scan and have to backtrack and re-parse, our stacks of identifier references enter an inconsistent state that causes us to bind non-existent references and treat them as closure-captured. This in turn leads to bad byte code generation. The problem is easy to solve in the case of a single arrow function parameter not enclosed in parentheses. To handle the harder case of a list enclosed in parens, advance the current block ID when we begin parsing a parenthetical expression. Identifier references will be created that are amenable to the arrow function case, and if we do wind up with an arrow function, we only need to correct the function ID's on the references we've already made.

### DIFF
--- a/lib/Parser/Parse.cpp
+++ b/lib/Parser/Parse.cpp
@@ -109,6 +109,7 @@ Parser::Parser(Js::ScriptContext* scriptContext, BOOL strictMode, PageAllocator 
     m_nextFunctionId = nullptr;
     m_errorCallback = nullptr;
     m_uncertainStructure = FALSE;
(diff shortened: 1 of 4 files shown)


pleath:
@aneeshdk This is related to the arrow function parameter fix you reviewed previously.

suwc:
:shipit:

suwc:
LGTM

PR #1980, opened 2016-11-10
Fix arm64 build failure

meg-gupta:
Some recent wasm lower function definitions were not added in arm64/LowerMD.h


meg-gupta:
@tcare  Can you please review ?

PR #1982, opened 2016-11-11
Change to address CVE-2016-7200,CVE-2016-7201,CVE-2016-7202,CVE-2016-7203,CVE-2016-7208,CVE-2016-7240,CVE-2016-7241,CVE-2016-7242,CVE-2016-7243

suwc:
Type confusion in Array.prototype.filter
Type confusion due to reentrancy can cause a Var to be written into a native int array.
Fix by making sure type-specialized code path is used only when ArraySpeciesCreate() invokes
built-in Array constructor.

Heap overflow in Array.prototype.splice
In Array.prototype.splice, array length is cached before ArraySpeciesCreate() is invoked.
Side-effect from ArraySpeciesCreate() can change array length and result in inconsistent states
and possibly heap overflow.
Fix by adding length check to keep cases with side effects out of fast path with pre-calculated length.
Also tweak logic in ArraySpeciesCreate() to flag a non-built-in constructor with missing [@species] property.

Type confusion in FillFromPrototypes
In ForEachOwnMissingArrayIndexOfObject(), existing array enumeration logic assumes Var array.
A native array from caller can cause type confusion and leak.
Fix by converting incoming native arrays to Var arrays.

Parameter type confusion in eval
Extra argument signified by CallFlags_ExtraArg shall be cast to FrameDisplay unless the extra argument
is used for new.target, in which case CallFlags_NewTarget is be set. Type confusion and AV occur because
existing logic in eval() does not check if CallFlags_NewTarget is cleared before using extra argument as
FrameDisplay.
Fix by adding CallFlags_NewTarget check to eval() before cast to FrameDisplay.

Type confusion in JSON.parse
Non-native array is expected in JSONParser::Walk(). A native array from caller
can cause type confusion and heap overflow
Fix by converting native arrays to Var arrays.

Type confusion in Array.prototype.concat and .splice
Array newly created by ArraySpeciesCreate is not being checked if it is a
JavascriptCopyOnAccessNativeIntArray, causing near-nullptr AVs.
Fix by adding check-and-convert against JavascriptCopyOnAccessNativeIntArray in
affected built-ins.

### DIFF
--- a/lib/Runtime/Base/CallInfo.h
+++ b/lib/Runtime/Base/CallInfo.h
@@ -65,6 +65,16 @@ namespace Js
         static const ushort ksizeofCount;
         static const ushort ksizeofCallFlags;
         static const uint kMaxCountArgs;
+
+        static bool isDirectEvalCall(CallFlags flags)
+        {
+            // This was recognized as an eval call at compile time. The last one or two args are internal to us.
+            // Argcount will be one of the following when called from global code
+            //  - eval("...")     : argcount 3 : this, evalString, frameDisplay
+            //  - eval.call("..."): argcount 2 : this(which is string) , frameDisplay
+
+            return (flags & (CallFlags_ExtraArg | CallFlags_NewTarget)) == CallFlags_ExtraArg;  // ExtraArg == 1 && NewTarget == 0
+        }
     };
 
     struct InlineeCallInfo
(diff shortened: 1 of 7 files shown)


suwc:
@pleath could you review/approve this pls?

pleath:
LGTM. Thanks.

PR #1988, opened 2016-11-11
Fix JsParse/JsRun for JavascriptString argument

agarwal-sandeep:
JavascriptString is internally wide, so pass correct length and flag in
CompileRun.

GetSz for JavascriptString can allocate memory and need to handle OOM so
need API wrapper.

We are using the wrapper twice but refactoring is not simple as
RunScriptCore is called at multiple, for now lets keep two wrappers

### DIFF
--- a/lib/Jsrt/Jsrt.cpp
+++ b/lib/Jsrt/Jsrt.cpp
@@ -4277,11 +4277,28 @@ _ALWAYSINLINE JsErrorCode CompileRun(
 {
     PARAM_NOT_NULL(scriptVal);
     VALIDATE_JSREF(scriptVal);
+    PARAM_NOT_NULL(sourceUrl);
 
     bool isExternalArray = Js::ExternalArrayBuffer::Is(scriptVal),
          isString = false;
     bool isUtf8   = !(parseAttributes & JsParseScriptAttributeArrayBufferIsUtf16Encoded);
-    if (!isExternalArray)
+
+    LoadScriptFlag scriptFlag = LoadScriptFlag_None;
+    const byte* script;
+    size_t cb;
+    const wchar_t *url;
+
+    if (isExternalArray)
+    {
+        script = ((Js::ExternalArrayBuffer*)(scriptVal))->GetBuffer();
+
+        cb = ((Js::ExternalArrayBuffer*)(scriptVal))->GetByteLength();
+
+        scriptFlag = (LoadScriptFlag)(isUtf8 ?
+            LoadScriptFlag_ExternalArrayBuffer | LoadScriptFlag_Utf8Source :
+            LoadScriptFlag_ExternalArrayBuffer);
+    }
+    else
     {
         isString = Js::JavascriptString::Is(scriptVal);
         if (!isString)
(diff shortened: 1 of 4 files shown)


agarwal-sandeep:
@obastemur @jianchun please take a look

agarwal-sandeep:
@dotnet-bot Ubuntu ubuntu_linux_test_static

agarwal-sandeep:

lib/Jsrt/Jsrt.cpp
```diff
@@ -4293,11 +4293,28 @@ _ALWAYSINLINE JsErrorCode CompileRun(
 {
     PARAM_NOT_NULL(scriptVal);
     VALIDATE_JSREF(scriptVal);
+    PARAM_NOT_NULL(sourceUrl);
```
jianchun: >PARAM_NOT_NULL(sourceUrl); [](start = 4, length = 26)

Not sure if we have convention anywhere else or not -- do we want to enable passing nullptr(JsInvalidReference) to mean user does not care source url?

agarwal-sandeep: We don't expect source url to be null at all places, RunScriptCore have PARAM_NOT_NULL(sourceUrl);

lib/Jsrt/Jsrt.cpp
```diff
@@ -4306,36 +4323,34 @@ _ALWAYSINLINE JsErrorCode CompileRun(
...
     }
 
-    LoadScriptFlag scriptFlag;
-    const byte* script = isExternalArray ?
-        ((Js::ExternalArrayBuffer*)(scriptVal))->GetBuffer() :
-        (const byte*)((Js::JavascriptString*)(scriptVal))->GetSz();
-    const size_t cb = isExternalArray ?
-        ((Js::ExternalArrayBuffer*)(scriptVal))->GetByteLength() :
-        ((Js::JavascriptString*)(scriptVal))->GetLength();
+    JsErrorCode error = ContextAPINoScriptWrapper_NoRecord([&](Js::ScriptContext *scriptContext) -> JsErrorCode {
```
jianchun: >ContextAPINoScriptWrapper_NoRecord [](start = 24, length = 34)

If only care about OOM, GlobalAPIWrapper seems better?

lib/Jsrt/Jsrt.cpp
```diff
@@ -4306,36 +4323,34 @@ _ALWAYSINLINE JsErrorCode CompileRun(
...
-    {
-        scriptFlag = LoadScriptFlag_None;
-    }
+        if (!Js::JavascriptString::Is(sourceUrl))
+        {
+            return JsErrorInvalidArgument;
+        }
 
-    const wchar_t *url;
+        url = ((Js::JavascriptString*)(sourceUrl))->GetSz();
```
jianchun: >((Js::JavascriptString*)(sourceUrl)) [](start = 14, length = 36)

JavascriptString::FromVar?

lib/Jsrt/Jsrt.cpp
jianchun: This is a convenience matter we talked about offline. Previous Windows APIs use wchar_t* for sourceUrl and easy for caller to construct. Now it is harder. If we allow nullptr/JsInvalidRef and treat them as `""`, more convenient for many scenarios. Maybe discuss with limin.

---
In reply to: [87932689](https://github.com/Microsoft/ChakraCore/pull/1988#discussion_r87932689) [](ancestors = 87932689)

lib/Jsrt/Jsrt.cpp
```diff
@@ -4293,11 +4293,28 @@ _ALWAYSINLINE JsErrorCode CompileRun(
...
-    if (!isExternalArray)
+
+    LoadScriptFlag scriptFlag = LoadScriptFlag_None;
+    const byte* script;
+    size_t cb;
+    const wchar_t *url;
+
+    if (isExternalArray)
+    {
+        script = ((Js::ExternalArrayBuffer*)(scriptVal))->GetBuffer();
```
obastemur: @jianchun Correct, it may also resolve ArrayBuffer here as if it is an External Array Buffer. Not sure what difference that would make?

jianchun: @obastemur +@liminzhu  I don't see why we can't support general ArrayBuffer. I merely noticed the problem that everywhere in the docs and code we claim support external ArrayBuffer, but actually the code works for general ArrayBuffer. If we decide general ArrayBuffer is fine, should update the docs and code.

One point might worth further investigation is if ArrayBuffer transfer/detach has any impact on this.

lib/Jsrt/Jsrt.cpp
agarwal-sandeep: This needs to be fixed in all other APIs as well. Lets take that as a separate change.

lib/Jsrt/Jsrt.cpp
```diff
@@ -4290,36 +4307,30 @@ _ALWAYSINLINE JsErrorCode CompileRun(
...
-    else if (isUtf8)
-    {
-        scriptFlag = (LoadScriptFlag) (LoadScriptFlag_Utf8Source);
-    }
-    else
-    {
-        scriptFlag = LoadScriptFlag_None;
-    }
+            // JavascriptString is 2 bytes (wchar_t/char16)
+            cb = jsString->GetLength() * sizeof(wchar_t);
```
agarwal-sandeep: >cb = jsString->GetLength() * sizeof(wchar_t); [](start = 12, length = 45)

This doesn't work if the scriptVal was created using JsCreateStringUtf16. That means if we pass a scriptVal created using JsCreateStringUtf16 we should pass JsParseScriptAttributeArrayBufferIsUtf16Encoded and then not do this math but that doesn't look right.

lib/Jsrt/Jsrt.cpp
obastemur: >  I don't see why we can't support general ArrayBuffer.

@jianchun I'm not sure if I meant that, actually I meant totally otherwise above. Indeed ArrayBuffer should also make the job but the notes on docs are mostly about perf suggestion, not a limitation

liminzhu: Does this mean passing a _JavaScript_ arraybuffer to script arguments would also work?

jianchun: @liminzhu Yes, that's what I meant by "general ArrayBuffer".

liminzhu: 👍

PR #1998, opened 2016-11-12
fix negZero min/max

meg-gupta:
Interpreter did not handle negative zero case for min/max.
Jit did not convert float to fouble before calling the helper IsNegZero
which accepts double.


meg-gupta:
@MikeHolman Can you please review ?

Cellule:
These implementation are share for asm.js and wasm. Is the spec the same for both, as in did we have a bug before in asm.js ?

lib/Backend/LowerMDShared.cpp
```diff
@@ -9690,15 +9690,19 @@ void LowererMD::GenerateFastInlineBuiltInCall(IR::Instr* instr, IR::JnHelperMeth
 IR::Opnd* LowererMD::IsOpndNegZero(IR::Opnd* opnd, IR::Instr* instr)
 {
     IR::Opnd * isNegZero = IR::RegOpnd::New(TyInt32, this->m_func);
+    IR::Opnd *src = opnd;
+
+    if (opnd->IsFloat32())
+    {
+        src = IR::RegOpnd::New(TyFloat64, this->m_func);
+        instr->InsertBefore(IR::Instr::New(LowererMD::MDConvertFloat32ToFloat64Opcode, src, opnd, this->m_func));
```
Cellule: For now let's just fix the bug, I've open an issue #2041 to track this so we'll remember to fix it later

Cellule:
LGTM

meg-gupta:
@dotnet-bot test Windows x64_test

PR #2001, opened 2016-11-14
Fixes Jsrt API C99/C++98 compatability

jcoffland:
``ChakraCore.h`` appears to desire compatibility with C99 however this is currently broken due to the use of ``nullptr``, ``bool`` and  C++11 style enums.  Fortunately these problems are easily remedied.

### DIFF
--- a/lib/Jsrt/ChakraCommon.h
+++ b/lib/Jsrt/ChakraCommon.h
@@ -71,6 +71,7 @@ typedef BYTE* ChakraBytePtr;
 #define CHAKRA_API extern "C" SET_API_VISIBILITY JsErrorCode
 #else
 #define CHAKRA_API extern     SET_API_VISIBILITY JsErrorCode
+#include <stdbool.h>
 #endif
 
 #include <stddef.h>  // for size_t
@@ -321,7 +322,7 @@ typedef UINT32 DWORD;
     /// <summary>
     ///     An invalid runtime handle.
     /// </summary>
-    const JsRuntimeHandle JS_INVALID_RUNTIME_HANDLE = nullptr;
+    const JsRuntimeHandle JS_INVALID_RUNTIME_HANDLE = 0;
 
     /// <summary>
     ///     A reference to an object owned by the Chakra garbage collector.
@@ -338,7 +339,7 @@ typedef UINT32 DWORD;
     /// <summary>
     ///     An invalid reference.
     /// </summary>
-    const JsRef JS_INVALID_REFERENCE = nullptr;
+    const JsRef JS_INVALID_REFERENCE = 0;
 
     /// <summary>
     ///     A reference to a script context.
--- a/bin/ch/ChakraRtInterface.h
+++ b/bin/ch/ChakraRtInterface.h
@@ -92,7 +92,7 @@ struct JsAPIHooks
     typedef JsErrorCode(WINAPI *JsrtTTDNotifyYieldPtr)();
     typedef JsErrorCode(WINAPI *JsrtTTDHostExitPtr)(int statusCode);
 
-    typedef JsErrorCode(WINAPI *JsrtTTDGetSnapTimeTopLevelEventMovePtr)(JsRuntimeHandle runtimeHandle, JsTTDMoveMode moveMode, int64_t* targetEventTime, int64_t* targetStartSnapTime, int64_t* targetEndSnapTime);
+  typedef JsErrorCode(WINAPI *JsrtTTDGetSnapTimeTopLevelEventMovePtr)(JsRuntimeHandle runtimeHandle, JsTTDMoveMode moveMode, uint32_t kthEvent, int64_t* targetEventTime, int64_t* targetStartSnapTime, int64_t* targetEndSnapTime);
     typedef JsErrorCode(WINAPI *JsrtTTDMoveToTopLevelEventPtr)(JsRuntimeHandle runtimeHandle, JsTTDMoveMode moveMode, int64_t snapshotStartTime, int64_t eventTime);
     typedef JsErrorCode(WINAPI *JsrtTTDReplayExecutionPtr)(JsTTDMoveMode* moveMode, int64_t* rootEventTime);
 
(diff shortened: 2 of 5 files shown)


jianchun:
@dotnet-bot test this please

jianchun:
@dotnet-bot test this please

obastemur:
@jcoffland could you please squash the commits ? That should also trigger CI

lib/Jsrt/ChakraCommon.h
```diff
@@ -71,6 +71,10 @@ typedef BYTE* ChakraBytePtr;
 #define CHAKRA_API extern "C" SET_API_VISIBILITY JsErrorCode
 #else
 #define CHAKRA_API extern     SET_API_VISIBILITY JsErrorCode
+#include <stdbool.h>
+#endif
+#if !defined(__cplusplus) || (__cplusplus <= 199711L)
```
jianchun: nit pick: Could you please insert a blank line before this `#if` for readability?

jcoffland:
Ok, I've added the whitespace and squashed the commits.

jcoffland:
Any chance this will get merged?  It's a very basic fix and I followed all the guidelines.  Will someone please take responsibility for merging this PR?

jianchun:
@jcoffland CI has been failing on this PR weirdly. Could you please try squash (still seeing 3 commits) and rebase onto latest master and force push your PR? Thanks!

jcoffland:
Ok, I believe I've sorted it out.  I had already squashed my commits but then updated my fork by doing a merge which added extra commits to this PR.  I fixed this by redoing the fork update as a rebase.

obastemur:
@dotnet-bot test this please

jianchun:
Did a search and found

```
JsTTDMoveMode moveMode = (JsTTDMoveMode)(JsTTDMoveMode::JsTTDMoveKthEvent | ((int64) startEventCount) << 32);
```
Looks current `JsTTDMoveMode` needs 64 bits.

@mrkmarron

jcoffland:
@jianchun good catch.  I found the line you quoted above https://github.com/Microsoft/ChakraCore/blob/master/bin/ch/ch.cpp#L275.  I did a little investigating and found that ``JsTTDMoveMode`` was hijacked to also pass in an event count.  You can see it being used at https://github.com/Microsoft/ChakraCore/blob/master/lib/Jsrt/Jsrt.cpp#L3781.  It appears that ``JsTTDGetSnapTimeTopLevelEventMove()`` is the only place ``JsTTDMoveMode`` is used in this way and https://github.com/Microsoft/ChakraCore/blob/master/bin/ch/ch.cpp#L275 is the only place it's ever called like this.  In fact, this is the only place ``JsTTDGetSnapTimeTopLevelEventMove()`` is called.

Abusing ``JsTTDMoveMode`` like this seems like a bad idea since it is very likely to cause confusion.  It's not at all obvious from looking at the definition of ``JsTTDMoveMode`` that it is going to be used in this way.  The easiest solution is to add a ``kthEvent`` parameter to ``JsTTDGetSnapTimeTopLevelEventMove()``.  I've implemented this change.

mrkmarron:
Hi @jcoffland, you are 100% correct that I was doing some parameter hijacking. I agree with you that this is suboptimal and converting the API to use a second parameter instead is more appropriate (thanks for your help cleaning this up). 

One comment I have is that kthEvent is an optional parameter so we should probably make that clear in the SAL annotation/name. 

@jianchun we will need to update the use of this API in the node/xplat branch when we merge later as well.

jcoffland:
@mrkmarron right.  I've now changed the parameter annotation from ``_In_`` to ``_In_opt``.

dilijev:
Is it possible to add a sort of test suite which exercises the use of our API when embedding ChakraCore in C99 and C++98 projects?

On that note, @liminzhu, I remember someone mentioning that one or more samples had broken when we updated the API. Should we think about including a set of tests that include the samples?

Should this fix be back-ported to release/1.4, release/1.3, and/or release/1.2 because we still support 1.2 and 1.3, and we had intended this to work?

dilijev:
On another note:

Should this fix be back-ported to release/1.4, release/1.3, and/or release/1.2 because we still support 1.2 and 1.3, and we had intended this to work?

jcoffland:
It wasn't an issue in 1.3 when I tried it.

dilijev:
@jcoffland Thanks for confirming that. In any case, I think this PR should probably be re-targeted to 1.4 (because I'm pretty sure this issue would affect that release). But, hold off on that until we confirm. Worst-case we can just merge to master and then back-port the change to 1.4.

obastemur:
> I remember someone mentioning that one or more samples had broken when we updated the API. Should we think about including a set of tests that include the samples to ensure we don't unexpectedly break anything? 

@dilijev We already test that for some time. See https://github.com/Microsoft/ChakraCore/tree/master/test/native-tests

jcoffland:
@obastemur Do those tests check that the samples in https://github.com/Microsoft/Chakra-Samples work?

obastemur:
@jcoffland AFAIK we do check a bit more here. However, we should add more tests there indeed.

jcoffland:
It would make sense to trigger CI builds on https://github.com/Microsoft/Chakra-Samples to make sure they build.  It would be best to have CI build badges on that project's README.md.  The reason being that the samples are one of the first places new users start with ChakraCore.  If it doesn't work they are likely to go elsewhere.

obastemur:
> It would make sense to trigger CI builds on https://github.com/Microsoft/Chakra-Samples to make sure they build.

I am not sure if I was being clear. We already build and test the Jsrt usage (with the same xpalt compile scripts and same code) on ChakraCore CI for each PR.

If CI here is not failing, it is unlikely that we break Chakra-Samples. That is what https://github.com/Microsoft/ChakraCore/tree/master/test/native-tests for.

jcoffland:
@obastemur What keeps native-tests in sync with Chakra-Samples?  My experience when I tried building Chakra-Samples was that they would not compile against ChakraCore ``master``.  This makes for a disheartening introduction.

Anyway, I've gone off topic.  Can this PR be merged now?

obastemur:
> What keeps native-tests in sync with Chakra-Samples?

We do. Anything is failing on native tests, should alert devs to update Chakra Samples. IMHO it is similar to updating documentation.

obastemur:
> Can this PR be merged now?

Sorry for keeping this PR open for longer than it should. @dilijev any update?

In the mean time #2136 to make sure we don't break it again.

dilijev:
> should alert devs to update Chakra Samples

@obastemur while this seems true in principle, I think it is easy to miss, and since, unlike documentation, Samples are something which we can actually break, maybe we should seriously think about adding some automated test mechanism to ensure they all continue to work.

dilijev:
Assigning myself. I'll test and merge.

jcoffland:
It should be noted that there are additional compile problems using the Jsrt API with the MSVC compiler in C mode that are not addressed by this PR.

Thanks for the merge!

dilijev:
@jcoffland Thanks for the heads-up. Could you open an issue so we can track that?

PR #2007, opened 2016-11-14
Fix regression from Bool-Int equality fast-path GenerateFastEqBoolInt

curtisman:
1. XOR requires the dst and src1 to be the same.  Just use the dst instead of a temp
2. The true ^ false is not a AddrOpndKindDynamicVar, change it to AddrOpndKindDynamicMisc


curtisman:
@Penguinwizzard please review.

LouisLaf:
:shipit:

PR #2008, opened 2016-11-14
fixing an assert in bytecodedump while printing int64 consts for asm.js function 

Krovatkin:
-dump:bytecode hits an assertion when dumping bytecodes for an asm.js function due to the fact we are trying to print int64 consts.

### DIFF
--- a/lib/Runtime/ByteCode/AsmJsByteCodeDumper.cpp
+++ b/lib/Runtime/ByteCode/AsmJsByteCodeDumper.cpp
@@ -227,6 +227,10 @@ namespace Js
         for (int i = 0; i < WAsmJs::LIMIT; ++i)
         {
             WAsmJs::Types type = (WAsmJs::Types)i;
+            if (func->GetTypedRegisterAllocator().IsTypeExcluded(type))
+            {
+                continue;
+            }
             uint constCount = func->GetTypedRegisterAllocator().GetRegisterSpace(type)->GetConstCount();
             if (constCount > 0)
             {


Krovatkin:
@Cellule this seems to be the only spot which is not guarded by `IsTypeExcluded` fyi @MikeHolman

Cellule:
In this case, the const count should be 0. What are you trying to dump that is hitting a problem ?

Krovatkin:
@Cellule 

> What are you trying to dump that is hitting a problem ?

My guess is that adding `-dump:bytecode` to dump bytecodes for **any** asm.js function will hit this assertion. I was running this test specifically

```javascript
function AsmModule(stdlib,foreign,buffer) {
    "use asm";
var HEAP32 =new stdlib.Int32Array(buffer);
function f1(f){
    
    f = f|0;
    var a=0,b=2,c=3,d=6,e=4,i=0,j=0,k=0,x=0,y=0;

    a=HEAP32[f+4>>2]|0;
    b=HEAP32[f+8>>2]|0;
    d=HEAP32[f+12>>2]|0;


return (a+b+d)|0;
}
return f1;
}

var global = {Int32Array:Int32Array}
var env = {}
var buffer = new ArrayBuffer(1<<20);

var asmModule = AsmModule(global,env,buffer);
for(var i=0;i<10000;i++)
asmModule(8);
```

> In this case, the const count should be 0.

`func->GetTypedRegisterAllocator().GetRegisterSpace(type)->GetConstCount()`
To get the count, we call `GetRegisterSpace(type)` first, `GetRegisterSpace(type)` triggers the assert.
It looks like the assert is there to stop one from using register space for an excluded type **at all** even if the **use** is as innocuous as getting the number of the consts of that particular type.

Cellule:
Oh you are right, when I wrote it I thought the `constCount > 0` was enough, but I had forgottent we would assert before the check.
LGTM

Cellule:

Krovatkin:

Krovatkin:
@Cellule 
Lol, should we bump up count to a bigger number? Looks like CI guys got all new mac books :-)))) `200000` doesn't overflow anymore 

```
  var count = 200000; // Keep this unrealistic number as we do (osx)
                        // and do not limit stack memory to a particular capacity
```

Krovatkin:
@Cellule Hooorraaaaah!!! It passseeeeeeeeeeedddddddd!! Could you please re-approve and help me merge this?

PR #2012, opened 2016-11-15
[swb] bring back quick sort

leirocks:
we need this when sorting an array from write barrier memory

### DIFF
--- a/lib/Common/DataStructures/QuickSort.h
+++ b/lib/Common/DataStructures/QuickSort.h
@@ -1,0 +1,53 @@
+//-------------------------------------------------------------------------------------------------------
+// Copyright (C) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE.txt file in the project root for full license information.
+//-------------------------------------------------------------------------------------------------------
+
+#pragma once
+namespace JsUtil
+{
+    template <class T, typename Comparer>
+    class QuickSort
+    {
+    public:
+        static void Sort(T* low, T* high, Comparer comparer, void* context)
+        {
+            if ((low == NULL)||(high == NULL))
+            {
+                return;
+            }
+            if (high > low)
+            {
+                T* pivot = Partition(low, high, comparer, context);
+                Sort(low, pivot-1, comparer, context);
+                Sort(pivot+1, high, comparer, context);
+            }
+        }
+
+    private:
+        static T* Partition(T* l, T* r, Comparer comparer, void* context)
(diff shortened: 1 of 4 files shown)


leirocks:

lib/Common/Memory/RecyclerPointers.h
```diff
@@ -375,6 +375,22 @@ class WriteBarrierPtr
...
+    WriteBarrierPtr& operator--()  // prefix --
+    {
+        --ptr;
+#ifdef RECYCLER_WRITE_BARRIER
+        RecyclerWriteBarrierManager::WriteBarrier(this);
+#endif
+        return *this;
+    }
+
+    WriteBarrierPtr operator--(int)  // postfix --
```
jianchun: Just to confirm, you added these operators for some other usages, not because of quick sort code added here right? Quicksort here only operates on `T*`, not `WriteBarrierPtr`.

lib/Common/DataStructures/QuickSort.h
```diff
@@ -0,0 +1,53 @@
...
+            if (high > low)
+            {
+                T* pivot = Partition(low, high, comparer, context);
+                Sort(low, pivot-1, comparer, context);
+                Sort(pivot+1, high, comparer, context);
+            }
+        }
+
+    private:
+        static T* Partition(T* l, T* r, Comparer comparer, void* context)
```
jianchun: This quicksort implementation performs poorly on already ordered array or array with large amount of equal elements. Good for now, will need to consider future improvements.

jianchun:
LGTM

lib/Common/Memory/RecyclerPointers.h
leirocks: that was because of quick sort code added here, but after I change the implementation it's not necessary. I kept it just for the completion - in case there's usage in the future. I can remove if you think it should not be here

lib/Common/DataStructures/QuickSort.h
leirocks: this was just bring back the one we removed previously. let's revisit this while doing perf investigation

leirocks:
@dotnet-bot test this please

lib/Common/Memory/RecyclerPointers.h
jianchun: This is fine, thanks for confirmation.

leirocks:

PR #2013, opened 2016-11-15
use bv for jit numeric property records

MikeHolman:
This improves perf in case you have large number of numeric properties. This also resolves some corner cases where OOM could lead to incorrect behavior.

I'm seeing a 1.5-2% improvement in splay. Perf looks flat other than this.


leirocks:
Nice idea! :shipit:

Issue #2009, opened 2016-11-14
jit: finally block is not executed after try/catch throws exception


xanlpz:
The Error/stack.js test fails because all the "finally" blocks are not executed (as of git master bbf46243132). I can reduce the issue to just this code
```javascript
try {
    try {
        throw 42;
    } finally {
        print("In finally");
    }
}
catch (e) {
    print("In catch", e.description);
}
```

If the test is run with -forceNative we get:
localhost:\~/git/ChakraCore%/home/xan/git/ChakraCore/BuildLinux/Test/ch -forceNative foo.js
In catch undefined
Without:
localhost:\~/git/ChakraCore%/home/xan/git/ChakraCore/BuildLinux/Test/ch foo.js             
In finally
In catch undefined

The tests expect the latter behavior.

jianchun:
Just tried this in an older branch and also latest master on Ubuntu, no repro...

xanlpz:
This is a test build, compile target amd64, built with clang 3.9.1 on Fedora 25. I can easily try an older compiler just in case, not sure what else might be relevant.

xanlpz:
Compiling with clang 3.8.1 fixes the bug. It also fixes the other failures I was seeing. I'll try to figure out whether this is a bug in clang or something a newer clang has uncovered in Chakra if that seems OK.

jianchun:
I also reproduced with clang 3.9 on Ubuntu. Thanks for reporting, and yes please help investigate!

obastemur:
I don't recall this issue while I was testing ArchLinux / Ubuntu with clang 3.9 couple of weeks ago

obastemur:
Could you please try this one with the latest master? #1964 disables tail call optimization only for noinline functions + clang 3.9 build fix is also merged in.

xanlpz:
Same results as of commit 526fbe4fb39. I tried to test with clang 4.0.0 (master) but Chakra does not compile.

xanlpz:
I have bisected this and not surprisingly the "bad" commit is the one enabling JIT on Linux. Still as I said above it remains to be seen whether there's a bug in there or in newer clangs.

obastemur:
@xanlpz Thanks for pushing on this. I can also confirm the bug with Clang 3.9

obastemur:
BTW; I don't think this is a Clang bug.

Krovatkin:
[...]

PR #2016, opened 2016-11-15
Fix incorrect TypedArray constructor overloading

### LINKED ISSUE #2014: TypedArray constructor throws TypeError for undefined argument
Effectively, 'new Uint8Array(0)' and 'new Uint8Array()' and 'new Uint8Array(undefined)' should have the same behavior.

Converting from an internal issue.

This issue is not a spec deviation from ES 6.0. It's a spec change in ES 2017. In ES6, we were supposed to (and are correctly) throwing TypeError if length is undefined. 

ES 6.0 spec text:

22.2.1.2 %TypedArray% ( length )

This description applies only if the %TypedArray% function is called with at least one argument and the Type of the first argument is not Object.

%TypedArray% called with argument length performs the following steps:

Assert: Type(length) is not Object. 
1. If NewTarget is undefined, throw a TypeError exception. 
1. If length is undefined, throw a TypeError exception. 
1. Let numberLength be ToNumber(length). 
1. Let elementLength be ToLength(numberLength). 
1. ReturnIfAbrupt(elementLength). 
1. If SameValueZero(numberLength, elementLength) is false, throw a RangeError exception. 
1. Return AllocateTypedArray(NewTarget, elementLength).

Test case:

```js
new Uint8Array(0)
new Uint8Array()
new Uint8Array(undefined) // TypeError
```

ianwjhalliday:
Fixes #2014.

Since TypedArray was originally implemented the spec has changed and the
constructors have changed how they handle arguments.  Explicit undefined
is no longer a TypeError, instead it converts to zero.  Likewise bools,
null, strings, symbols, and numbers are converted via the ToIndex
algorithm.  Objects are treated as array-likes if not iterables, nor
array buffers nor typed arrays.

### DIFF
--- a/lib/Runtime/Library/TypedArray.cpp
+++ b/lib/Runtime/Library/TypedArray.cpp
@@ -400,39 +400,38 @@ namespace Js
         if (args.Info.Count > 1)
         {
             Var firstArgument = args[1];
-            if (TypedArrayBase::Is(firstArgument))
-            {
-                // Constructor(TypedArray array)
-                typedArraySource = static_cast<TypedArrayBase*>(firstArgument);
-                if (typedArraySource->IsDetachedBuffer())
-                {
-                    JavascriptError::ThrowTypeError(scriptContext, JSERR_DetachedTypedArray, _u("[TypedArray]"));
-                }
-
-                elementCount = typedArraySource->GetLength();
-                if (elementCount >= ArrayBuffer::MaxArrayBufferLength/elementSize)
-                {
-                    JavascriptError::ThrowRangeError(scriptContext, JSERR_InvalidTypedArrayLength);
-                }
-            }
-            else if (ArrayBufferBase::Is(firstArgument))
-            {
-                // Constructor(ArrayBuffer buffer,
-                //  optional uint32 byteOffset, optional uint32 length)
-                arrayBuffer = ArrayBufferBase::FromVar(firstArgument);
-                if (arrayBuffer->IsDetached())
-                {
-                    JavascriptError::ThrowTypeError(scriptContext, JSERR_DetachedTypedArray, _u("[TypedArray]"));
-                }
-            }
-            else if (JavascriptNumber::Is(firstArgument) ||
-                    TaggedInt::Is(firstArgument))
+            if (!Js::JavascriptOperators::IsObject(firstArgument))
             {
                 elementCount = ArrayBuffer::ToIndex(firstArgument, JSERR_InvalidTypedArrayLength, scriptContext, ArrayBuffer::MaxArrayBufferLength / elementSize);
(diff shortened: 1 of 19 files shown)


ianwjhalliday:
Please review @boingoing @dilijev. (The diff is easier to read if you ignore whitespace changes)

I might want to change this PR to submit to release/1.3 because this fix is needed to move node-chakracore ahead to more recent node source.

dilijev:
@ianwjhalliday If you do retarget this PR to branch `release/1.3` GitHub now lets you modify the target of the PR, but in order to make sure your PR consists of just this change, you'll also need to cherry-pick this change onto release/1.3 on your end and then force-push the branch that corresponds to this PR.

lib/Runtime/Library/TypedArray.cpp
```diff
@@ -448,47 +447,44 @@ namespace Js
...
-                                JavascriptError::ThrowTypeError(
-                                    scriptContext, JSERR_InvalidTypedArray_Constructor);
-                            }
-
-                            elementCount = ArrayBuffer::ToIndex(lengthVar, JSERR_InvalidTypedArrayLength, scriptContext, ArrayBuffer::MaxArrayBufferLength / elementSize);
-                        }
+                        // REVIEW: unclear why this JavascriptConversion::ToObject() call is being made.
+                        // It ends up calling RecyclableObject::ToObject which at least Proxy objects can
+                        // hit with non-trivial behavior.
+                        Js::Throw::FatalInternalError();
```
akroshg: I did not understand this change - you already checked IsObject above, so this shouldn't be false? If the firstargument is proxy, so we should be returning proxy.

akroshg:
changes LGTM.

lib/Runtime/Library/TypedArray.cpp
ianwjhalliday: Yes, but `JavascriptConversion::ToObject` makes a dynamic call to `RecyclableObject::ToObject` on the `Var` which has side effects when the object is a proxy.  There may be other objects that override this method and do something special too.

I didn't want to spend time figuring this out so I just maintained it and assert that it does indeed always return true as expected (fatal error if not).

boingoing:
🚢

ianwjhalliday:

ianwjhalliday:

ianwjhalliday:

ianwjhalliday:
this is incredibly frustrating

akroshg:
This task failed on my job as well. Is there a bug tracking this?

ianwjhalliday:
Turns out this bug was brought up a few days ago in #2009.  I'll raise priority on that issue since it is slowing down our check-in rate.

PR #2017, opened 2016-11-15
ENABLE_SEH instead of DISABLE_SEH

obastemur:
I was enabling debugger on xplat. Saw DISABLE_SEH and notes on ENABLE_SEH.
Instead of adding another DISABLE_SEH, introducing couple of changes.

This PR also fixes an exit code issue with `ch` and SEH related build issues with Windows/Clang

New:

- Both `__try` and `__finally` are no-op on xplat.
- `ifdef` out only the `__except` logic.

### DIFF
--- a/lib/Common/PlatformAgnostic/Exception.h
+++ b/lib/Common/PlatformAgnostic/Exception.h
@@ -1,0 +1,15 @@
+//-------------------------------------------------------------------------------------------------------
+// Copyright (C) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE.txt file in the project root for full license information.
+//-------------------------------------------------------------------------------------------------------
+
+#ifdef __clang__
+    #undef __try
+    #undef __finally
+    #define __try
+    #define __finally
+#else
+    #ifndef ENABLE_SEH
+        #define ENABLE_SEH 1
+    #endif
+#endif
--- a/lib/Runtime/Library/JavascriptFunction.cpp
+++ b/lib/Runtime/Library/JavascriptFunction.cpp
@@ -664,12 +664,6 @@ namespace Js
         }
 #endif
 
-#ifdef DISABLE_SEH
-        // xplat: JavascriptArrayBuffer::AllocWrapper is disabled on cross-platform
-        // (IsValidVirtualBufferLength always returns false).
-        // SEH and ResumeForOutOfBoundsArrayRefs are not needed.
-        ret = CallRootFunctionInternal(args, scriptContext, inScript);
-#else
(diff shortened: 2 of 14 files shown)


obastemur:
@jianchun @agarwal-sandeep review please

lib/Common/PlatformAgnostic/Exception.h
```diff
@@ -0,0 +1,15 @@
+//-------------------------------------------------------------------------------------------------------
+// Copyright (C) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE.txt file in the project root for full license information.
+//-------------------------------------------------------------------------------------------------------
+
+#ifdef __clang__
+    #undef __try
+    #undef __finally
+    #define __try
+    #define __finally
```
jianchun: This looks overkill to me and brings more problems than good. We discourage SEH use because that's not cross platform. Better not to have them available than to define them.

lib/Runtime/Library/RuntimeLibraryPch.h
```diff
@@ -99,4 +99,9 @@
 #define JS_DIAG_INLINE inline
 #else
 #define JS_DIAG_INLINE
-#endif
\ No newline at end of file
+#endif
+
+// Include this last. GNU redefines __try as `try`
```
jianchun: One more reason not to mess with `__try`. I think the previous `#if ... #else ...` pattern is clear enough. Or we could `#ifdef` out the `__try/__except` part individually. I'd support leaving it inconvenient to discourage more SEH use.

lib/Common/PlatformAgnostic/Exception.h
```diff
@@ -0,0 +1,15 @@
+//-------------------------------------------------------------------------------------------------------
+// Copyright (C) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE.txt file in the project root for full license information.
+//-------------------------------------------------------------------------------------------------------
+
+#ifdef __clang__
```
jianchun: BTW, wondering if I misunderstand `__clang__`. Why use this compiler symbol to determine platform feature? Does clang on Windows support SEH?

lib/Runtime/Library/JavascriptFunction.cpp
```diff
@@ -697,7 +696,7 @@ namespace Js
                 UnexpectedExceptionHandling_fatal_error(&exceptionInfo);
             }
```
jianchun: This is one example the dummy `__try/__finally` approach brings confusion. This finally block is meaningless and not applicable. I think the before-change code is clearer.

obastemur:
Necessary changes made under Debugger PR

PR #2021, opened 2016-11-15
Fix TVS false positive by using a failfast

tcare:
Fixes OS 9022474.

PREFast warns of a potential overflow when writing halfway through a struct. We don't think this will overflow halfway, but change to a failfast just in case.

### DIFF
--- a/lib/Common/Core/Assertions.h
+++ b/lib/Common/Core/Assertions.h
@@ -75,6 +75,18 @@ __declspec(selectany) int IsInAssert = false;
 #define AnalysisAssert(x)               Assert(x); __analysis_assume(x)
 #define AnalysisAssertMsg(x, comment)   AssertMsg(x, comment); __analysis_assume(x)
 
+#ifdef DBG
+#define AssertOrFailFast(x)                 Assert(x)
+#define AssertOrFailFastMsg(x, msg)         AssertMsg(x, msg)
+#define AnalysisAssertOrFailFast(x)         AnalysisAssert(x)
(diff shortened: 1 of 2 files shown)


tcare:
@curtisman @ianwjhalliday 

@suwc I know you had something similar in the works. I went with 'or' instead of 'and' because we already use 'and' elsewhere.

lib/Common/Core/Assertions.h
```diff
@@ -75,6 +75,18 @@ __declspec(selectany) int IsInAssert = false;
...
+#ifdef DBG
+#define AssertOrFailFast(x)                 Assert(x)
+#define AssertOrFailFastMsg(x, msg)         AssertMsg(x, msg)
+#define AnalysisAssertOrFailFast(x)         AnalysisAssert(x)
+#define AnalysisAssertOrFailFastMsg(x, msg) AnalysisAssertMsg(x, msg)
+#else
+#define AssertOrFailFast(x)                 do { if (!(x)) { Js::Throw::FatalInternalError(); } } while (false)
+#define AssertOrFailFastMsg(x, msg)         AssertOrFailFast(x)
+#define AnalysisAssertOrFailFast(x)         AssertOrFailFast(x)
+#define AnalysisAssertOrFailFastMsg(x, msg) AssertOrFailFast(x)
```
dilijev: Since these are all semantically FailFast in the non-DBG case, maybe you could add a macro FailFast(x) which does what AssertOrFailFast (non-DBG) does (in line 84) and then define all of line 84 - line 87 as FailFast(x)?

PR #2022, opened 2016-11-16
Fixing the case when caller property points to a generator function

aneeshdk:
While calling another function from a generator function, the caller
property of that function should point to the generator function.
Currently we were walking the stack and returning the inner virtual function
which is not equivalent to the generator function.

### DIFF
--- a/lib/Runtime/Library/JavascriptFunction.cpp
+++ b/lib/Runtime/Library/JavascriptFunction.cpp
@@ -2622,6 +2622,11 @@ LABEL1:
                 // Caller of this function expects nullValue from the requestContext.
                 funcCaller = nullValue;
             }
+            if (ScriptFunction::Is(funcCaller))
+            {
+                // Is this is the internal function of a generator function then return the original generator function
+                funcCaller = ScriptFunction::FromVar(funcCaller)->GetRealFunctionObject();
+            }
         }
 
         return StackScriptFunction::EnsureBoxed(BOX_PARAM(funcCaller, nullptr, _u("caller")));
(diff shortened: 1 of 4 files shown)


aneeshdk:
@ianwjhalliday @akroshg Can you please take a look?

akroshg:
are you planning to add the cross-site scenarios as well?

akroshg:
:shipit:

test/es6/generators-functionality.js
```diff
@@ -1766,6 +1766,23 @@ var tests = [
...
+            }
+            function* gf(flag, value) {
+                if (!flag) {
+                    yield func();
+                }
+                yield value * value;
+            }
+            var callergf = gf().next().value;
+            assert.areEqual(callergf, gf, "Generator function returned through the caller property should be the same as the original generator function");
+            assert.areEqual(callergf(true, 10).next().value, 100, "Generator returned through the caller property should behave the same as the original generator function");
```
ianwjhalliday: Expected and actual are reversed in this assert.

Should be `assert.areEqual(100, callergf(...`

Same for the above.

lib/Runtime/Library/JavascriptFunction.cpp
```diff
@@ -2622,6 +2622,11 @@ void __cdecl _alloca_probe_16()
                 // Caller of this function expects nullValue from the requestContext.
                 funcCaller = nullValue;
             }
+            if (GeneratorVirtualScriptFunction::Is(funcCaller))
```
ianwjhalliday: You don't need to create a `GeneratorVirtualScriptFunction::Is()` method, instead you can just test that `funcCaller` is a `ScriptFunction` via `funcCaller->IsScriptFunction()`.  `GetRealFunctionObject` is provided on `ScriptFunction` and returns itself if it is not a `GeneratorVirtualScriptFunction`.

PR #2023, opened 2016-11-16
Add option to delay full JIT small functions

ThomsonTan:
Add a new option -delayFullJITSmallFunc. This is supposed to help JIT through but not turned on this time.

### DIFF
--- a/lib/Runtime/Base/FunctionBody.cpp
+++ b/lib/Runtime/Base/FunctionBody.cpp
@@ -6834,9 +6834,9 @@ namespace Js
                     - If the size is >= 100, scale by 1.6
             */
             const uint loopPercentage = GetByteCodeInLoopCount() * 100 / max(1u, GetByteCodeCount());
-            const int byteCodeSizeThresholdForInlineCandidate = CONFIG_FLAG(InlineThreshold) + CONFIG_FLAG(InlineThresholdAdjustCountInSmallFunction);
+            const int byteCodeSizeThresholdForInlineCandidate = CONFIG_FLAG(LoopInlineThreshold);
             bool delayFullJITThisFunc =
-                (CONFIG_FLAG(DelayFullJITSmallFunc) > 0) && (byteCodeSizeThresholdForInlineCandidate < 0 || this->GetByteCodeWithoutLDACount() > (uint)byteCodeSizeThresholdForInlineCandidate);
+                (CONFIG_FLAG(DelayFullJITSmallFunc) > 0) && (this->GetByteCodeWithoutLDACount() <= (uint)byteCodeSizeThresholdForInlineCandidate);
 
             if(loopPercentage <= 50 || delayFullJITThisFunc)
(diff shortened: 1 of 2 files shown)


ThomsonTan:

lib/Runtime/Base/FunctionBody.cpp
```diff
@@ -6834,11 +6834,19 @@ namespace Js
                     - If the size is >= 100, scale by 1.6
             */
             const uint loopPercentage = GetByteCodeInLoopCount() * 100 / max(1u, GetByteCodeCount());
-            if(loopPercentage <= 50)
+            const int byteCodeSizeThresholdForInlineCandidate = CONFIG_FLAG(InlineThreshold) + CONFIG_FLAG(InlineThresholdAdjustCountInSmallFunction);
```
rajatd: You should use CONFIG_FLAG(LoopInlineThreshold) here because, today, that is the largest bytecode size for which we will always inline.

lib/Runtime/Base/FunctionBody.cpp
```diff
@@ -6834,11 +6834,19 @@ namespace Js
...
+            bool delayFullJITThisFunc =
+                (CONFIG_FLAG(DelayFullJITSmallFunc) > 0) && (byteCodeSizeThresholdForInlineCandidate < 0 || this->GetByteCodeWithoutLDACount() > (uint)byteCodeSizeThresholdForInlineCandidate);
```
rajatd: Shouldn't the byte code size of current function be less than the byteCodeSizeThresholdForInlineCandidate for you to be sure that this function will get inlined?

lib/Runtime/Base/FunctionBody.cpp
ThomsonTan: Changed threshold to CONFIG_FLAG(LoopInlineThreshold).

lib/Runtime/Base/FunctionBody.cpp
ThomsonTan: Good catch, fixed it in the new commit.

lib/Runtime/Base/FunctionBody.cpp
```diff
@@ -6834,11 +6834,19 @@ namespace Js
                     - If the size is >= 100, scale by 1.6
             */
             const uint loopPercentage = GetByteCodeInLoopCount() * 100 / max(1u, GetByteCodeCount());
-            if(loopPercentage <= 50)
+            const int byteCodeSizeThresholdForInlineCandidate = CONFIG_FLAG(LoopInlineThreshold);
```
rajatd: >CONFIG_FLAG(LoopInlineThreshold) [](start = 64, length = 32)

We could probably call a function here that would return the minimum of all the inline thresholds that we have. Just something to keep in mind when you enable this.

rajatd:
:shipit:

PR #2024, opened 2016-11-16
Two bug fixes

rajatd:
1. Set the paramClosure from the field on InterpreterStackFrame to its reserved bytecode register before calling out to a jitted loop body, and set the register back to nullptr after returning from loop body.

2. Change an assert in JITObjTypeSpecFldInfo into a condition check as we now have a valid case which fails the assert.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/2024)


rajatd:
@aneeshdk @pleath please review.

pleath:
LGTM.

rajatd:
@dotnet-bot test Windows x64_test
@dotnet-bot test OSX osx_osx_debug_static

PR #2025, opened 2016-11-16
Avoid returning global function on .caller

akroshg:
We return null for arguments.callee.caller is global function. But in the case of exception we fetch the last exception and return top function where the exception caused (window.onerror scenario). That exception object can be global function. So added check to not do that.

### DIFF
--- a/lib/Runtime/Library/JavascriptFunction.cpp
+++ b/lib/Runtime/Library/JavascriptFunction.cpp
@@ -2630,6 +2630,7 @@ LABEL1:
     BOOL JavascriptFunction::GetCallerProperty(Var originalInstance, Var* value, ScriptContext* requestContext)
     {
         ScriptContext* scriptContext = this->GetScriptContext();
+        *value = nullptr;
 
         if (this->IsStrictMode())
         {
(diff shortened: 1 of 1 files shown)


akroshg:
the repro will be 
```js
    <script>
    window.onerror = function() {
        arguments.callee.caller.toString();
    }
    </script>
    <script>
    document.domain = 'foobar.com';
    </script>
```

akroshg:
@leirocks please have a look.

lib/Runtime/Library/JavascriptFunction.cpp
```diff
@@ -2672,7 +2672,10 @@ void __cdecl _alloca_probe_16()
             {
                 JavascriptFunction* exceptionFunction = unhandledExceptionObject->GetFunction();
                 // This is for getcaller in window.onError. The behavior is different in different browsers
-                if (exceptionFunction && scriptContext == exceptionFunction->GetScriptContext())
+                if (exceptionFunction 
+                    && scriptContext == exceptionFunction->GetScriptContext()
+                    && exceptionFunction->IsScriptFunction()
+                    && !exceptionFunction->GetFunctionBody()->GetIsGlobalFunc())
                 {
                     *value = exceptionFunction;
```
agarwal-sandeep: Can we assign the value to null on top of this function. GetCallerProperty is called from GetPropertyBuiltIns which is used in JavascriptFunction::GetProperty without checking the return value.

akroshg:
@dotnet-bot  test Ubuntu ubuntu_linux_test  please

akroshg:
@agarwal-sandeep @leirocks  please have a look.

agarwal-sandeep:
Any impact on what is seen under F12? otherwise LGTM

akroshg:
I am fixing one of the edgecrawler crash itself. Also our behavior is aligning more with other browsers.
if the .caller is global function, we return null. However in the exception case we don't do that. So I fixed that.
I don't see any impact on F12. During debugging we actually intercept the exception before OnScriptError.

PR #2028, opened 2016-11-16
Fix copysign code generation in jit to use the right sign mask for doubles

meg-gupta:


meg-gupta:
@dotnet-bot test OSX osx_osx_debug_static

lib/Runtime/Math/WasmMath.inl
```diff
@@ -127,10 +127,16 @@ inline int WasmMath::Eqz(T value)
...
+inline double WasmMath::Copysign(double aLeft, double aRight)
+{
+    return copysign(aLeft, aRight);
+}
+
+template<>
+inline float WasmMath::Copysign(float aLeft, float aRight)
 {
-    return (T)_copysign(aLeft, aRight);
+    return copysignf(aLeft, aRight);
```
Cellule: use `_copysignf` as `copysignf` is not available in ChakraFull's compiler

lib/Runtime/Math/WasmMath.inl
```diff
@@ -127,10 +127,16 @@ inline int WasmMath::Eqz(T value)
     return value == 0;
 }
 
-template<typename T>
-inline T WasmMath::Copysign(T aLeft, T aRight)
+template<>
+inline double WasmMath::Copysign(double aLeft, double aRight)
+{
+    return copysign(aLeft, aRight);
```
Cellule: use `_copysign` as `copysign` is not available in ChakraFull's compiler

lib/Backend/LowerMDShared.cpp
```diff
@@ -6255,11 +6255,13 @@ LowererMD::GenerateCopysign(IR::Instr * instr)
 
     // Copy sign from src2 to src1
     IR::Opnd* src1 = instr->GetSrc1();
+    IR::Opnd* src2 = instr->GetSrc2();
     Assert(src1->IsFloat32() || src1->IsFloat64());
     GenerateFloatAbs(src1->AsRegOpnd(), instr);
 
     IR::Instr* t2 = IR::Instr::New(Js::OpCode::ANDPS, instr->GetSrc2(), instr->GetSrc2(),
-        IR::MemRefOpnd::New(this->m_func->GetThreadContextInfo()->GetSgnBitCst(), src1->GetType(), this->m_func, src1->IsFloat32() ? IR::AddrOpndKindDynamicFloatRef : IR::AddrOpndKindDynamicDoubleRef),
+        IR::MemRefOpnd::New(src2->IsFloat32() ? this->m_func->GetThreadContextInfo()->GetSgnFloatBitCst() : this->m_func->GetThreadContextInfo()->GetSgnDoubleBitCst(),
```
Cellule: nit: this is getting long. Could you extract the 2 ternary in variables?

lib/Runtime/Math/WasmMath.inl
meg-gupta: The Linux builds are complaining about _copysignf. Do you know of an alternative ?

Cellule:
I think you will have to implement `_copysignf` in `pal\src\cruntime\finite.cpp` and declare it in `pal\inc\pal.h`
You can search for `_copysign`.
The implementation should be to simply call `copysignf` and return the result

PR #2029, opened 2016-11-16
Redeferral fix for recursive inlinees

pleath:
Detect cycles in the FunctionCodeGenRuntimeData tree when redeferral is looking for inlinee FunctionBody's that need to be kept alive. Stop descending at that point, since we've already processed the parent data that points to itself.

### DIFF
--- a/lib/Runtime/Base/FunctionBody.cpp
+++ b/lib/Runtime/Base/FunctionBody.cpp
@@ -747,14 +747,18 @@ namespace Js
         return !!pActiveFuncs->TestAndSet(this->GetFunctionNumber());
     }
 
-    void FunctionBody::UpdateActiveFunctionsForOneDataSet(ActiveFunctionSet *pActiveFuncs, FunctionCodeGenRuntimeData **dataSet, uint count) const
+    void FunctionBody::UpdateActiveFunctionsForOneDataSet(ActiveFunctionSet *pActiveFuncs, FunctionCodeGenRuntimeData *parentData, FunctionCodeGenRuntimeData **dataSet, uint count) const
     {
         FunctionCodeGenRuntimeData *inlineeData;
         for (uint i = 0; i < count; i++)
         {
             for (inlineeData = dataSet[i]; inlineeData; inlineeData = inlineeData->GetNext())
             {
-                inlineeData->GetFunctionBody()->UpdateActiveFunctionSet(pActiveFuncs, inlineeData);
+                // inlineeData == parentData indicates a cycle in the structure. We've already processed parentData, so don't descend.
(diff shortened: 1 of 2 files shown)


pleath:
@rajatd -- This is the recursion issue we were just discussing.
@kunalspathak -- This fixes the web issue you ran into.

lib/Runtime/Base/FunctionBody.cpp
```diff
@@ -783,12 +787,12 @@ namespace Js
         FunctionCodeGenRuntimeData **data = this->GetCodeGenRuntimeData();
         if (data != nullptr)
         {
-            this->UpdateActiveFunctionsForOneDataSet(pActiveFuncs, data, this->GetProfiledCallSiteCount());
+            this->UpdateActiveFunctionsForOneDataSet(pActiveFuncs, nullptr, data, this->GetProfiledCallSiteCount());
```
kunalspathak: >nullptr [](start = 67, length = 7)

I don't have much insights in this code, but won't this always go inside the if-condition you have above `inlineeData != parentData`?

lib/Runtime/Base/FunctionBody.cpp
```diff
@@ -747,14 +747,18 @@ namespace Js
...
+    void FunctionBody::UpdateActiveFunctionsForOneDataSet(ActiveFunctionSet *pActiveFuncs, FunctionCodeGenRuntimeData *parentData, FunctionCodeGenRuntimeData **dataSet, uint count) const
     {
         FunctionCodeGenRuntimeData *inlineeData;
         for (uint i = 0; i < count; i++)
         {
             for (inlineeData = dataSet[i]; inlineeData; inlineeData = inlineeData->GetNext())
             {
-                inlineeData->GetFunctionBody()->UpdateActiveFunctionSet(pActiveFuncs, inlineeData);
+                // inlineeData == parentData indicates a cycle in the structure. We've already processed parentData, so don't descend.
+                if (inlineeData != parentData)
```
kunalspathak: >inlineeData [](start = 20, length = 11)

can you add a unit test for this?

lib/Runtime/Base/FunctionBody.cpp
pleath: We make recursive calls as we descend the structure, so we can still wind up in the looping case.

lib/Runtime/Base/FunctionBody.cpp
pleath: I have crash dumps, but I don't have a console repro that actually fails when I run it locally.

pleath:

PR #2031, opened 2016-11-17
Fix flakiness in serialized tests in ch

digitalinfinity:
Serialized bytecode tests started seeing flakiness after my ETW change.
This is because that change exposed a bug in ch- ch was freeing the serialized bytecode before disposing the runtime. We just happened to
not need any of the serialized data during our test runs before- with the ETW change, we started logging the name of the function body being collected and this would AV if the serialized bytecode data got decommited before
we called JsDisposeRuntime.

The simplest fix for now is to call JsDisposeRuntime before freeing the bytecode data

### DIFF
--- a/bin/ch/ch.cpp
+++ b/bin/ch/ch.cpp
@@ -456,11 +456,6 @@ HRESULT CreateAndRunSerializedScript(const char* fileName, LPCSTR fileContents, 
     IfFailGo(RunScript(fileName, fileContents, bcBuffer, fullPath));
 
 Error:
-    if (bcBuffer != nullptr)
-    {
-        delete[] bcBuffer;
-    }
-
     if (current != JS_INVALID_REFERENCE)
     {
         ChakraRTInterface::JsSetCurrentContext(current);
@@ -470,6 +465,12 @@ Error:
     {
         ChakraRTInterface::JsDisposeRuntime(runtime);
     }
+
+    if (bcBuffer != nullptr)
+    {
+        delete[] bcBuffer;
+    }
+
     return hr;
 }
 


digitalinfinity:
@akroshg @curtisman please review

akroshg:
LGTM, however if someone writing an App using jsrt apis, they need to understand this ordering correctly. Is it something documented?

digitalinfinity:
Thanks @akroshg. Looks like at least one doc alludes to this: https://msdn.microsoft.com/en-us/library/dn249594(v=vs.94).aspx. However, even in that doc, it's not clear that it needs to be kept alive till the runtime is disposed. @liminzhu can we make sure that the documentation makes it clear that if the host is managing serialized bytecode, the host needs to keep it alive until the runtime using it is disposed?

curtisman:
@obastemur Should JsRunSerialized support unload call back?

obastemur:
> @obastemur Should JsRunSerialized support unload call back similar to JsRunSerializedScriptWithCallback?

@curtisman definitely should and remember raising inconsistency concerns about this. The experimental Utf8 version of it was already including that. https://github.com/Microsoft/ChakraCore/blob/73751c2461787c0afef3f456d20778682f44bcb1/lib/Jsrt/Jsrt.cpp#L3358

curtisman:
> >@obastemur Should JsRunSerialized support unload call back similar to JsRunSerializedScriptWithCallback?
> @curtisman definitely should and remember raising inconsistency concerns about this. The experimental Utf8 version of it was already including that. https://github.com/Microsoft/ChakraCore/blob/73751c2461787c0afef3f456d20778682f44bcb1/lib/Jsrt/Jsrt.cpp#L3358
To elaborate: On JsRunSerializedScriptWithCallback, the unload call back would signal both the source and byte code can be unloaded.  We can have support the same behavior in JsRunSerialized.  

With the new API, all the "unload" callback are done via ExternalArrayBuffer, and JsRunSerialized can do the same and accept a Var for the byte code buffer and get the unload call back?

digitalinfinity:
Yeah, I support a callback variant for JsRunSerialized. That's why I described this fix as the simplest fix 😄 - wanted to get a quick fix out there since the VC++ team is hitting this while running our tests.

liminzhu:
An unload callback sounds more straightforward to use. 

@digitalinfinity  to your earlier point about documentation, I have [a bunch of things](https://github.com/Microsoft/ChakraCore-wiki/issues/5#issuecomment-238417862) I want to improve including serialization b/c they come up fairly common. Haven't got any cycles to actually write them so far, but I probably will tackle them in December.

obastemur:
> With the new API, all the "unload" callback are done via ExternalArrayBuffer, and JsRunSerialized can do the same and accept a Var for the byte code buffer and get the unload call back?

- Both JsRunSerialized and JsParseSerialized depend on `JsSerialize` and JsSerialize actually works like JsStringCopy.. 

- had a proposition for an additional encoding enum but we ended up using parseAttributes and that didn't fit to others.

- I'm a 1% pro on Limin's side on `An unload callback sounds more straightforward to use. ` however, if we are into making it convenient with others, that is also fine. However, this approach will add couple of extra calls.

PR #2032, opened 2016-11-17
WASM - api testing + bugfixes

### LINKED ISSUE #2033: WASM - js api bugs
Fix WebAssembly.instantiate, Module.exports and Module.imports

### LINKED ISSUE #2034: WASM - WebAssembly.Module.imports keep order
WebAssembly.Module.imports resulting array should be in the same order as it was initially describe in the wasm section

Cellule:
This is a pretty big change as I had to do some refactoring to fix some issues with the `WebAssembly.Module.imports` api and a bug in reusing the same module for multiple instance with imported globals.
I took the opportunity to cleanup the global/imports code as well as the module environment in the instance.
Fixes #2033

Overview of the changes I made:
- Throw a different error message if the element in the function table is null on call.
- Throw a `WebAssembly.RuntimeError ` for exception thrown from the jitted code for Wasm functions
- Throw in `WebAssembly::ReadBufferSource` if the buffer is empty and/or null since we assume the buffer is valid after that
- Created the class `WebAssemblyEnvironment` to handle get/set elements in the module environment. This allows to do additional checks to avoid writing past the buffer limit.
- The `WasmGlobal` objects are now immutable. This allows to re-instantiate the module with imported globals. The globals needs to be read from the `WebAssemblyEnvironment` for init_expr and local references
- Some of the apis we use in `WebAssemblyInstance::CreateInstance` can throw a `WasmCompilationException`, for now I converted that exception to a TypeError. Related #2035
- Insert all the imports type in the imports array in `WebAssemblyModule`. Treat all imports similarly to how we do exports in instantiate. 
  Fixes #2034

This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/2032)


Cellule:
Reviewed 44 of 44 files at r1.
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[test/wasm/baselines/api.baseline, line 1 at r1](https://reviewable.io:443/reviews/microsoft/chakracore/2032#-KWk9vfAju4slptEYd9d:-KWk9vfB2niAmjfpY6UJ:b67mss0) ([raw file](https://github.com/microsoft/chakracore/blob/81bdc6d60150629dfc4d7e5d9f3534286a7a117c/test/wasm/baselines/api.baseline#L1)):*
> ```
> 
> ```

There are some unexpected results here that I hadn't seen before, I will reinvestigate that

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/microsoft/chakracore/2032)*

Cellule:

lib/Runtime/Library/WebAssembly.cpp
```diff
@@ -212,25 +212,26 @@ WebAssembly::ReadBufferSource(Var val, ScriptContext * ctx, _Out_ BYTE** buffer,
...
     }
-    else
+    else if (isArrayBuffer)
     {
         Js::ArrayBuffer* arrayBuffer = Js::ArrayBuffer::FromVar(val);
         *buffer = arrayBuffer->GetBuffer();
         *byteLength = arrayBuffer->GetByteLength();
     }
+
+    if (*buffer == nullptr || *byteLength == 0)
```
MikeHolman: is it invalid for bytelength to be 0? probably unimportant, but i wonder if this case should be a WebAssemblyCompileError?

lib/Runtime/Library/WebAssemblyInstance.cpp
```diff
@@ -83,60 +114,57 @@ WebAssemblyInstance::CreateInstance(WebAssemblyModule * module, Var importObject
...
+        uint32 startFuncIdx = module->GetStartFunction();
+        if (startFuncIdx != Js::Constants::UninitializedValue)
+        {
+            AsmJsScriptFunction* start = environment.GetWasmFunction(startFuncIdx);
+            Js::CallInfo info(Js::CallFlags_New, 1);
+            Js::Arguments startArg(info, (Var*)&start);
+            Js::JavascriptFunction::CallFunction<true>(start, start->GetEntryPoint(), startArg);
+        }
+    }
+    catch (Wasm::WasmCompilationException e)
```
MikeHolman: should we assert here? compile error shouldn't escape to here

lib/Runtime/Library/WebAssemblyInstance.cpp
```diff
@@ -260,257 +284,198 @@ void WebAssemblyInstance::BuildObject(WebAssemblyModule * wasmModule, ScriptCont
...
+            env->SetGlobalValue(global, cnst);
             break;
         }
-        case Wasm::WasmTypes::I64:
-            Js::JavascriptError::ThrowTypeError(ctx, WASMERR_InvalidTypeConversion);
         default:
             Assert(UNREACHED);
-            break;
-
+            JavascriptError::ThrowTypeError(ctx, VBSERR_InternalError);
```
MikeHolman: You can do Js::ThrowInternalError(), which will assert on chk build and throw error on test/release build

MikeHolman:
:shipit:

lib/Runtime/Library/WebAssembly.cpp
Cellule: It will at least not have the magic number. It might not be the same error, but it will be an error nonetheless

lib/Runtime/Library/WebAssemblyInstance.cpp
Cellule: The problem is, we use this exception in some of the api that we use for instantiation.
Thus a `WasmCompilationException` can be considered an Instantiation error.

Cellule:

PR #2039, opened 2016-11-17
Do not mix enum types in IsWriteBarrierTriggerableValue

maksqwe:
IsAddrOpnd() is checking GetKind() == OpndKindAddr;
And next check AsAddrOpnd()->GetKind() == AddrOpndKindDynamicVar would be always false.


dilijev:
@dotnet-bot test this please

dilijev:
Hi @maksqwe there was a test failure before. I've triggered a new CI build but it looks like this PR is out of date enough that there are conflicts so the CI will likely fail. Could you update this PR and see if you can resolve the failure? Also, if you can construct a test case which would reveal the bad behavior before your change and add that to the test suite, we'd really appreciate it.

dilijev:
@MikeHolman assigning you since it looks like you're adopting this one

PR #2042, opened 2016-11-17
Disable test/Error/stack.js on osx re #2009

ianwjhalliday:
test/Error/stack.js sometimes fails, prominently in the OSX
osx_debug_static CI test.  See #2009

Disabling the test on mac for now to reduce pain on check-ins.

### DIFF
--- a/test/Error/rlexe.xml
+++ b/test/Error/rlexe.xml
@@ -30,6 +30,7 @@
     <default>
       <files>stack.js</files>
       <baseline>stack.baseline</baseline>
+      <tags>exclude_mac</tags>
     </default>
   </test>
   <test>


ianwjhalliday:
@jianchun @obastemur @akroshg @xanlpz 

Someone needs to fix that test but in the meantime lets disable it so folks don't have to experience delays in their legitimate check-ins like Akrosh and I experienced yesterday.

obastemur:
Just seeing this one.

@ianwjhalliday This issue can be CI related. It is referenced on #2009 due to interesting similarity with the source.

On #2009 this file fails on Linux machine not OSX.

ianwjhalliday:
@obastemur my guess was that we may have code that exhibits undefined behavior, and, depending on build configuration and environment, charkacore.dll is less or more likely to behave incorrectly due to this undefined behavior.

But just a guess

PR #2043, opened 2016-11-17
Provide more information when test scripts fail.

dilijev:
This patch was helpful in identifying the cause of a build error (differentiating between unit tests and native tests) when both steps seemed to indicate nothing was wrong.

### DIFF
--- a/test/runcitests.cmd
+++ b/test/runcitests.cmd
@@ -116,7 +116,10 @@ set _HadFailures=0
 
   call :do %_TestDir%\runtests.cmd -%1%2 -quiet -cleanupall -binDir %_StagingDir%\bin
 
-  if ERRORLEVEL 1 set _HadFailures=3
+  if %ERRORLEVEL% NEQ 0 (
+    echo -- runcitests.cmd ^>^> runtests.cmd failed
+    set _HadFailures=3
+  )
 
   goto :eof
 
@@ -127,7 +130,10 @@ set _HadFailures=0
 
   call :do %_TestDir%\runnativetests.cmd -%1%2 > %_TestDir%\logs\%1_%2\nativetests.log 2>&1
 
-  if ERRORLEVEL 1 set _HadFailures=4
+  if %ERRORLEVEL% NEQ 0 (
+    echo -- runcitests.cmd ^>^> runnativetests.cmd failed
+    set _HadFailures=4
+  )
 
   goto :eof
 
--- a/test/runcitests.cmd
+++ b/test/runcitests.cmd
@@ -128,11 +128,12 @@ set _HadFailures=0
 :: ============================================================================
 :runNativeTests
 
-  call :do %_TestDir%\runnativetests.cmd -%1%2 > %_TestDir%\logs\%1_%2\nativetests.log 2>&1
+  set _LogFile=%_TestDir%\logs\%1_%2\nativetests.log
+  call :do %_TestDir%\runnativetests.cmd -%1%2 > %_LogFile% 2>&1
 
   if %ERRORLEVEL% NEQ 0 (
-    echo -- runcitests.cmd ^>^> runnativetests.cmd failed (printing nativetests.log below)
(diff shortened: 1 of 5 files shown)


dilijev:
@kunalspathak @boingoing - I'm targeting this to release/1.3 because that's the farthest back that we run Native Unit tests.

test/runnativetests.cmd
```diff
@@ -179,7 +179,8 @@ goto :main
 :runtest
   pushd %_TestTempDir%
   call :do nativetests.exe
-  if ERRORLEVEL 1 set _HadFailures=1
+  if %ERRORLEVEL% NEQ 0 set _HadFailures=1
+  echo -- runnativetests.cmd ^>^> nativetests.exe exiting with exit code %_HadFailures%
```
kunalspathak: Is there a way to access test logs of `nativetests.exe`. Currently I don't see a log for these test.

dilijev: /cc @akroshg

dilijev:
The exit code by itself is an improvement but it only allows us to differentiate whether it was unit tests or native tests that failed. Unfortunately it still requires someone to look at the logs to find out which is which. It would be better to also print to stdout the information we have about the issue. I'll make this update.

test/runnativetests.cmd
dilijev: Apparently it produces a summary if the command completes successfully. This could be improved to help make it more clear when something went wrong while executing nativetests.exe but that's out of scope of this change.

The log from native tests is printed to the standard output of the build and the nativetests.log file is copied to the build outputs.

kunalspathak:
LGTM

test/runcitests.cmd
```diff
@@ -127,10 +130,13 @@ set _HadFailures=0
 
   call :do %_TestDir%\runnativetests.cmd -%1%2 > %_TestDir%\logs\%1_%2\nativetests.log 2>&1
 
-  if ERRORLEVEL 1 set _HadFailures=1
+  if %ERRORLEVEL% NEQ 0 (
+    echo -- runcitests.cmd ^>^> runnativetests.cmd failed
```
akroshg: could we echo that look for nativetests.log for more information ?

akroshg:
:shipit:

test/runcitests.cmd
dilijev: AFAIK right now it doesn't contain any information that isn't already echoed to stdout.

Also, we already print out that we're xcopying E:\A\_work\5\s\test\logs\x86_test\nativetests.log (among other things) to the testlogs folder (which is then copied to the drop path). Making the message more specific would probably get too specific to the VSO build infra (which is bad since this can be used for local builds as well).

I think this is probably unnecessary.

dilijev: Oh I guess it actually does contain information that isn't echoed to stdout. Maybe we can echo the log file to stdout to solve this problem?

akroshg: entire log would not be necessary, however point to the location of the log would be good help. 

---
In reply to: [88581337](https://github.com/Microsoft/ChakraCore/pull/2043#discussion_r88581337) [](ancestors = 88581337)

dilijev: Updating to display nativetests.log (with a call-out that we are printing the contents of that file).

akroshg: awesome

---
In reply to: [88587091](https://github.com/Microsoft/ChakraCore/pull/2043#discussion_r88587091) [](ancestors = 88587091)

dilijev:
@dotnet-bot
test Windows x64_debug
test Windows x64_test
test Windows x86_debug
test Windows x86_test

dilijev:
@dotnet-bot test this please

dilijev:
@dotnet-bot test this please

dilijev:

dilijev:
Will replace with a new PR when ready.

PR #2044, opened 2016-11-17
OS8050875:Direct super calls in a base class constructor

suwc:
Direct super calls in a base class constructor should be SyntaxError. Define new flags to signify where direct super calls are syntactically correct and pass these flags into eval() and multiple nested cases comprised of a mix of eval and lambda.

### DIFF
--- a/lib/Parser/Parse.cpp
+++ b/lib/Parser/Parse.cpp
@@ -362,8 +362,7 @@ HRESULT Parser::ParseSourceInternal(
 
         if ((grfscr & fscrEvalCode) != 0)
         {
-            // This makes the parser to believe when eval() is called, it accept any super access in global scope.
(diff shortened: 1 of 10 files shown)


suwc:
@aneeshdk @LouisLaf @pleath @Microsoft/chakra-es, could you review pls?

suwc:
@aneeshdk @LouisLaf @pleath @Microsoft/chakra-es could you review pls?

suwc:
@aneeshdk @LouisLaf @pleath @Microsoft/chakra-es, changed to lazy stack-walking to avoid special treatment of extra argument (unit test added to es6/rest.js). Perf flat. Review pls?

suwc:
Moving to release/1.4

PR #2047, opened 2016-11-18
Fix reuse of redeferred function's child

pleath:
If a function is redeferred, we need to reuse any child FunctionInfo's that have already been created. Handle the case where a child we want to be re-use has also been redeferred but we went ahead and created an AST for it. For now, just discard the AST and treat it as deferred. Consider transitioning to full FunctionBody on the spot if this hits more than rarely (which I doubt).


pleath:
@rajatd, can you look? Nice little fix.

rajatd:
:shipit:

PR #2053, opened 2016-11-18
Fix call to Math.pow x86 from JIT code

ThomsonTan:
When x86 JIT code assumes both parameters for Math.pow are double according to its profile data, the function (DirectPow) could still get small integer within our fast path. Need to check that before going to CRT function for result consistency of Math.pow.


ThomsonTan:
@rajatd please help review.

rajatd:
:shipit:

PR #2057, opened 2016-11-18
Redeferral fix for cross-site functions

pleath:
Don't let redeferral update the entry point on a function object's type if the entry point is already a cross-site thunk.


rajatd:
:shipit:

PR #2059, opened 2016-11-19
Fix the eqboolint lowering more

Penguinwizzard:
The issue here was that I had forgotten to add a jump in the firstTrue
case to the lower block of code; this resulted in negative operators only
runnign the first (copy) option.


Penguinwizzard:
@rajatd  please review

rajatd:
Add the test case; looks good otherwise

---
In reply to: [261678553](https://github.com/Microsoft/ChakraCore/pull/2059#issuecomment-261678553) [](ancestors = 261678553)

PR #2060, opened 2016-11-19
Fix build break in x86 for Full build

meg-gupta:


meg-gupta:
@Cellule can you please review ?

Cellule:
LGTM

PR #2063, opened 2016-11-19
RegAlloc fold spilled store

Cellule:
Fold a spilled store only if the source and dest have the same size.
This was causing problems in WebAssembly where we were doing a conversion from int32 to int64 and the int64 dest was getting spilled.
It was using the source's size to determine the spill size.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/2063)

### DIFF
--- a/lib/Backend/LinearScan.cpp
+++ b/lib/Backend/LinearScan.cpp
@@ -870,7 +870,7 @@ LinearScan::SetDstReg(IR::Instr *instr)
                         RecordDef(lifetime, instr, 0);
                     }
                 }
-                if (LowererMD::IsAssign(instr) && instr->GetSrc1()->IsRegOpnd())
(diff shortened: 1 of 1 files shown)


Cellule:
@MikeHolman @LouisLaf can you review this please?
I am not entirely sure if this can impact the codegen for the rest of javascript. I doubt this pattern even exists, but I will be running Exprgen over the weekend.
This solves one of the issue we've been having for WebAssembly with big tests.
I haven't been able to create a small repro as x64 has too many register and I couldn't get it to spill the right one

MikeHolman:
@Cellule Nice find, this sounds like the bug we've been looking for!

Cellule:
@MikeHolman well it's a bug, but after fixing this one I am hitting another one on the same test

Cellule:
@LouisLaf You are right, I forgot about MOV_TRUNC. I tested and it works and has way less chance of affecting the rest of javascript.
Thanks

Cellule:
I created #2067 to fix this issue the right way

PR #2066, opened 2016-11-19
WASM - source context id

Cellule:
Create a SourceContext for each WebAssembly Module, this helps debugging as they won't share the `-1` context id

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/2066)


Cellule:
@MikeHolman can you review this please ?

MikeHolman:
Definitely something we want, but I wonder if there could be side effects. @pleath, what do you think?

Cellule:
Should I merge this in release/1.4 or leave it targeted for master ?
I don't think we need this for the release, so maybe we might as well leave it in master only

PR #2067, opened 2016-11-19
WASM - uint to int64 bug fix

Cellule:
Use MOV_TRUNC instead of MOV when converting an unsigned int to int64

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/2067)

### DIFF
--- a/lib/Backend/amd64/LowererMDArch.cpp
+++ b/lib/Backend/amd64/LowererMDArch.cpp
@@ -2569,13 +2569,7 @@ LowererMDArch::EmitIntToLong(IR::Opnd *dst, IR::Opnd *src, IR::Instr *instrInser
     Assert(dst->IsRegOpnd() && dst->IsInt64());
     Assert(src->IsInt32());
 
-    if (src->IsIntConstOpnd())
-    {
-        Lowerer::InsertMove(dst, src, instrInsert);
-        return;
-    }
-    Assert(src->IsRegOpnd());
-    instrInsert->InsertBefore(IR::Instr::New(Js::OpCode::MOVSXD, dst, src, this->m_func));
+    Lowerer::InsertMove(dst, src, instrInsert);
 }
 
 void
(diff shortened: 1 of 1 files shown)


Cellule:
@MikeHolman @meg-gupta Can you review this please ?

MikeHolman:
I think we should do MOVZX instead.

Cellule:
I didn't see a version in the instruction references that supports `MOVZX r64, r32` It seems to be used only for `r8` or `r16` for sources

MikeHolman:
Ah yeah it looks like there isn't. I don't really like the name of MOV_TRUNC for this, since it implies the opposite. You should call Legalize though, otherwise high bytes might have garbage.

Cellule:
I tried replacing the content of that function to `Lowerer::InsertMove(dst, src, instrInsert);` and it seems to work, but I don't understand why.
The original instr is `Conv_Prim i64, u32`
Legalizer converts it to `MOV_TRUNC u32, u32`
I am wondering if this work only because we're not using indexes past i32

MikeHolman:
That definitely doesn't sound right to me.

Sorry I was unclear though. I meant that you should call Legalize on what you had (I saw code for MOV_TRUNC that zeros out the register before using it).

Cellule:
I think calling InsertMove is still the best option as it handles thing way better than we do and it calls Legalize.
I tried on my case and everything seems fine

Buggy code on the left, with this fix on the Right
image

This is way better because we can use the register `r10` since it is available and we will correctly spill 64bits so the load near the end will be correct

MikeHolman:
@Cellule, what is value of rsi coming into this block? Is it possible that upper bits are garbage? Do we initialize with `xor rsi, rsi` or `xor esi, esi`?

Cellule:
Since `rsi` is u32 only `esi` is used.
we will convert this to `mov r10d, esi` which zero out the high bits of r10 and completely ignore the high bits of rsi

MikeHolman:
@Cellule Ah ok. Yeah, was a bit confused by the names.

From intel manual:

>If the register is a destination operand, the resulting value in the two high-order bytes of the register is implementation dependent

Looks like on new processors it fills 0, but don't know if that behavior is present on all hardware we support.

Cellule:
Do you suggest we do a `Mov r64, 0` before the conversion ?

MikeHolman:
Yeah, I'm not convinced that it is necessary.

PR #2069, opened 2016-11-20
swb: annotate nested field type, union

jianchun:
plugin:
- Track down and annotate field type if an annotated field is of
  struct/class/union (RecordType) or array of such a type. To opt out, use
  FieldNoBarrier.
- Add logs to reason about annotation errors. To see why the plugin
  requires a class to be annotated, run
  `build.sh -n -d --wb-check THE_CPP_FILE --wb-args=-verbose`
  and read the traces at the bottom. It shows if the requirement is from
  a direct write barriered allocation, or a base class or a field type of
  such allocated class.

Fixed all newly discovered missing annotations. Notes:
- DoublyLinkedList/DoubleLinkedListElement now both takes TAllocator
  template parameter, and they must use the same TAllocator type.
- Fixed union with WriteBarrierPtr members with union constructor (c++11
  feature, thanks @leirocks for suggestion).
- FunctionBody.h was forced write barrier but some types should not force.
  Solved by moving "force" to only bottom region. To be refactored later
  after merged with master.
- Fixed bugs discovered from test: DynamicObject inline and aux slots were
  not annotated. Fixed.
- There are global static SimpleTypeHandler instances created from static
  BuildInPropertyIds that are not GC objects at all. They are used before
  WriteBarrierManager initialization. Fixed by using NoBarrierSet.

Fixed qsort_s overload to handle type that contains WriteBarrierPtr (use
a policy type similarly to CopyArray). Also emits compile error if qsort_s
is used on WriteBarrierPtr array.

Enhanced "runtests.py" to be able to pass ch flags. e.g. to disable a
default ch switch: `runtests.py -d --flag=-WriteBarrierTest-`.

### DIFF
--- a/lib/Common/Memory/RecyclerPointers.h
+++ b/lib/Common/Memory/RecyclerPointers.h
@@ -216,6 +216,11 @@ void ClearArray(WriteBarrierPtr<T>& dst, size_t count)
 {
     ClearArray(static_cast<T*>(dst), count);
 }
+template <class T, size_t N>
+void ClearArray(T (&dst)[N])
+{
+    ClearArray(dst, N);
+}
 
 
 template <typename T>
@@ -446,6 +451,22 @@ inline T* const& PointerValue(T* const& ptr) { return ptr; }
 template <class T>
 inline T* const& PointerValue(const WriteBarrierPtr<T>& ptr) { return ptr; }
 
+// Unsafe NoWriteBarrierSet. Use only when necessary and you are sure skipping
+// write barrier is ok.
+//
+template <class T>
+inline void NoWriteBarrierSet(T*& dst, T* ptr) { dst = ptr; }
+template <class T>
+inline void NoWriteBarrierSet(WriteBarrierPtr<T>& dst, T* ptr)
+{
+    dst.NoWriteBarrierSet(ptr);
+}
+template <class T>
+inline void NoWriteBarrierSet(WriteBarrierPtr<T>& dst, const WriteBarrierPtr<T>& ptr)
+{
+    dst.NoWriteBarrierSet(ptr);
+}
+
 }  // namespace Memory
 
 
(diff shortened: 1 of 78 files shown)


jianchun:

jianchun:

obastemur:
Quickly looked through non swb related parts. A concern and rest LGTM

jianchun:

lib/Common/Memory/RecyclerPointers.h
```diff
@@ -468,10 +489,41 @@ const T& max(const T& a, const NoWriteBarrierField<T>& b) { return a > b ? a : b
...
+// QuickSort Array content
+//
+template <class Policy>
+struct _QuickSortImpl
+{
+    template<class T, class Comparer>
+    static void qsort_s(T* arr, size_t count, const Comparer& comparer, void* context)
+    {
+        // by default use system qsort_s
+        ::qsort_s(arr, count, sizeof(T), comparer, context);
```
jianchun: I recall this qsort_s is implemented in PAL and calls to qsort_r.

obastemur: Yep. That final `qsort_r` call is the problem :)

jianchun: @obastemur Sorry, misunderstood your comment. Do you recall details of that qsort_r issue? We have brought back a custom QuickSort implementation, but at the moment not fine tuned, so would like to use system version if available.

jianchun:
@dotnet-bot  test Linux tests please

lib/Common/Memory/RecyclerPointers.h
obastemur: > but at the moment not fine tuned, so would like to use system version if available.

@jianchun The biggest issue with system qsort_r is that the problems are hard to track once we hit. I feel your reasoning and makes sense.

How about if we create a Github issue OR add todo there -> `Make our quicksort great and use it instead of system qsort_r because it is unreliable cross platform. It really is..`  ?

lib/Common/Memory/RecyclerPointers.h
```diff
@@ -468,10 +489,41 @@ const T& max(const T& a, const NoWriteBarrierField<T>& b) { return a > b ? a : b
...
 
-template<typename T, typename Comparer>
-void qsort_s(Memory::WriteBarrierPtr<T>* _Base, size_t _NumOfElements, size_t _SizeOfElements, Comparer comparer, void* _Context)
+// QuickSort Array content
+//
+template <class Policy>
+struct _QuickSortImpl
+{
+    template<class T, class Comparer>
+    static void qsort_s(T* arr, size_t count, const Comparer& comparer, void* context)
```
leirocks: >qsort_s [](start = 16, length = 7)

should we change a name to distinguish with the CRT one?

leirocks:
:shipit:

lib/Common/Memory/RecyclerPointers.h
jianchun: I'll change this to `QuickSort` next. Thanks!

lib/Common/Memory/RecyclerPointers.h
jianchun: @obastemur Opened new issue to track this. Thanks!

jianchun:
Thanks @obastemur @leirocks for review.

PR #2070, opened 2016-11-20
Jsrt: Use ArrayBuffer instead of BYTE for *Serialize

obastemur:


bin/ch/ch.cpp
```diff
@@ -80,76 +80,50 @@ void __stdcall PrintUsage()
 
 // On success the param byteCodeBuffer will be allocated in the function.
```
jianchun: Please update the outdated comment.

jianchun:
@liminzhu I read this PR code and it solves the API inconsistency issue by use "JsValueRef" (ArrayBuffer) consistently in both byte code buffer generation and consumption. Looks like a nice approach to me.

@obastemur There might be a Var lifetime issue. I didn't see clearly how the JsParse/RunSerialized input `bufferVal` being referenced in the runtime. This is similar to the `scriptVal` needing to be stored (in Utf8SourceInfo) to keep alive. If `bufferVal` not stored in runtime, its finalizer could be called early and underlying byte buffer be destroyed early.

liminzhu:
@obastemur What's the benefit of replacing byte buffer with var? 

A very common scenario of using serialization is that someone implements a service that creates a new context and serves the same script for every incoming tenant. The host needs to serialize the common script, store the buffer, and share it with every context. By changing byte buffer to var, the host would need extra getArrayBufferStorage call (one-time) and createArrayBuffer calls (per-context) for the case I mentioned.

obastemur:
> @obastemur I'm still not getting why not just an unload callback.

@liminzhu Actually I was also pro `unloadCallback` but after discussing with @curtisman I'm convinced with this one. This one is compatible with others and so simple to use.

Coming back to your scenario; You can create multiple ExternalArrayBuffer from a single source and performance wise, this is not much a killer IIRC.

agarwal-sandeep:
>         (const byte*)((Js::JavascriptString*)(scriptVal))->GetSz();

Use Js::JavascriptString::FromVar instead, same below

---
Refers to: lib/Jsrt/Jsrt.cpp:4447 in d33b7df. [](commit_id = d33b7df98c11b9880772ea10c9a43842f9b1f24c, deletion_comment = False)

liminzhu:
@obastemur I do like the consistency as it is similar with using arraybuffer for the script input.

How expensive is `JsCreateExternalArrayBuffer`? If that basically just wraps a byte buffer into an arraybuffer (and is not trying to do anything crazy like copying the buffer), then I guess it may be cheap enough, though it's still unfortunate because serialization is a perf feature and this action has to be done for every context in the scenario I mentioned.

Also trying to understand the lifetime of the arraybuffer created by `JsSerialize`. Does the call require an active script context and is the result buffer affiliated to the context? I'd imagine so, then the host has to either keep that context alive, or copy the arraybuffer storage into a different buffer (adding a chunk to peak memory) in order for other context to use it. In short, it does add quite a bit of hassle if someone wants to use serialized bytecode in multiple contexts.

obastemur:
> How expensive is JsCreateExternalArrayBuffer?

IMO, the answer is not expensive + `insignificant`. Especially comparing to `creating a whole new sub runtime / context / run stuff there...`

> Also trying to understand the lifetime of the arraybuffer created by JsSerialize...

This is one case vs others. IMO, once again this is not as expensive as it sounds'. What we are talking on this `edge case` is that an extra`alloc` and copy/move.

lib/Jsrt/Jsrt.cpp
```diff
@@ -4460,19 +4462,40 @@ CHAKRA_API JsSerialize(
...
+    }
+
+    if (bufferSize == 0)
+    {
+        return JsErrorScriptCompile;
+    }
+
+    if ((errorCode = JsCreateArrayBuffer(bufferSize, bufferVal)) == JsNoError)
+    {
+        byte* buffer = ((Js::ArrayBuffer*)(*bufferVal))->GetBuffer();
```
obastemur: Code is updated.

bin/ch/ch.cpp
```diff
@@ -80,76 +80,50 @@ void __stdcall PrintUsage()
...
         "// Licensed under the MIT license. See LICENSE.txt file in the project root for full license information.\r\n"
         "//-------------------------------------------------------------------------------------------------------\r\n"
         "#if 0\r\n";
+
+
+    HRESULT hr = GetSerializedBuffer(contentsRaw, &bufferVal);
+
+    if (FAILED(hr)) return hr;
+
+    IfFalseGo(ChakraRTInterface::JsGetArrayBufferStorage(bufferVal, &bcBuffer,
```
jianchun: I see...
You could use `IfJsErrorFailLog`, used a lot in this file.

---
In reply to: [89010064](https://github.com/Microsoft/ChakraCore/pull/2070#discussion_r89010064) [](ancestors = 89010064)

bin/ch/ch.cpp
obastemur: > Please update the outdated comment.

done

lib/Runtime/Base/ScriptContext.h
```diff
@@ -52,6 +52,8 @@ class SRCINFO
     Js::ModuleID moduleID;
     ULONG grfsi;
 
+    Js::Var serializedScriptBuffer; // in order to keep a reference to serialized Script Buffer
```
obastemur: > Also the byteCodeBuffer does not apply to all scenarios, only needed for JsParse/RunSerialized. Would JsrtSourceHolder be a better place? 

SRCINFO is the place exactly for this reason.

jianchun: Looks SRCINFO has more impact than JsrtSourceHolder? SRCINFO also used for other inputs (Edge). JsrtSourceHolder only used for JsParseRunSerialized.

obastemur: @jianchun  That can be ifdef'ed for ChakraFull. I just don't want to update all the way down to JsrtSourceHolder. Eventually one thing is part of another.

jianchun: Why ifdef or "all the way down to" -- I see it's just a matter of doing it at https://github.com/obastemur/ChakraCore/blob/a2ab359471b48c7c5dade5eae4c0a2934f6bfce9/lib/Jsrt/Jsrt.cpp#L3261 vs. https://github.com/obastemur/ChakraCore/blob/a2ab359471b48c7c5dade5eae4c0a2934f6bfce9/lib/Jsrt/Jsrt.cpp#L3233 passing one more parameter to JsrtSourceHolder constructor.

bin/ch/ch.cpp
```diff
@@ -79,77 +79,50 @@ void __stdcall PrintUsage()
...
+        &bcBufferSize) != JsNoError);
+
+    bcFileHandle = CreateFile(bcFullPath, GENERIC_WRITE, FILE_SHARE_DELETE,
```
ThomsonTan: Log GetLastError()? Or it would be hard to figure out why CreateFile failed just from the log.

bin/ch/ChakraRtInterface.h
```diff
@@ -309,7 +309,7 @@ class ChakraRTInterface
     static JsErrorCode WINAPI JsGetExternalData(JsValueRef object, void **data) { return HOOK_JS_API(GetExternalData(object, data)); }
     static JsErrorCode WINAPI JsCreateArray(unsigned int length, JsValueRef *result) { return HOOK_JS_API(CreateArray(length, result)); }
     static JsErrorCode WINAPI JsCreateArrayBuffer(unsigned int byteLength, JsValueRef *result) { return m_jsApiHooks.pfJsrtCreateArrayBuffer(byteLength, result); }
```
ThomsonTan: >m_jsApiHooks [](start = 104, length = 12)

m_jsApiHooks in JsCreateArrayBuffer is not updated the same as JsGetArrayBufferStorage?

bin/ch/ChakraRtInterface.h
```diff
@@ -366,8 +366,8 @@ class ChakraRTInterface
...
-    static JsErrorCode WINAPI JsSerialize(JsValueRef script, ChakraBytePtr buffer, unsigned int *bufferSize, JsParseScriptAttributes parseAttributes) { return HOOK_JS_API(Serialize(script, buffer, bufferSize, parseAttributes)); }
-    static JsErrorCode WINAPI JsRunSerialized(ChakraBytePtr buffer, JsSerializedLoadScriptCallback scriptLoadCallback, JsSourceContext sourceContext, JsValueRef sourceUrl, JsValueRef * result) { return HOOK_JS_API(RunSerialized(buffer, scriptLoadCallback, sourceContext, sourceUrl, result)); }
+    static JsErrorCode WINAPI JsSerialize(JsValueRef script, JsValueRef *buffer, JsParseScriptAttributes parseAttributes) { return HOOK_JS_API(Serialize(script, buffer, parseAttributes)); }
+    static JsErrorCode WINAPI JsRunSerialized(JsValueRef buffer, JsSerializedLoadScriptCallback scriptLoadCallback, JsSourceContext sourceContext, JsValueRef sourceUrl, JsValueRef * result) { return HOOK_JS_API(RunSerialized(buffer, scriptLoadCallback, sourceContext, sourceUrl, result)); }
```
ThomsonTan: >JsValueRef buffer [](start = 46, length = 17)

Gives name buffer to variable in both type "JsValueRef" and "JsValueRef *" looks a little confuse. Renaming "JsValueRef *buffer" to "JsValueRef *pBuffer"?

bin/ch/ch.cpp
```diff
@@ -79,77 +79,50 @@ void __stdcall PrintUsage()
...
     // For validating the header file against the library file
     auto outputStr =
         "//-------------------------------------------------------------------------------------------------------\r\n"
         "// Copyright (C) Microsoft. All rights reserved.\r\n"
         "// Licensed under the MIT license. See LICENSE.txt file in the project root for full license information.\r\n"
         "//-------------------------------------------------------------------------------------------------------\r\n"
         "#if 0\r\n";
+
+
+    HRESULT hr = GetSerializedBuffer(contentsRaw, &bufferVal);
```
obastemur: See below `if` .. similar to what we had before.

bin/ch/ch.cpp
obastemur: Sounds like a nice improvement in general. Could you please create a Github issue?

bin/ch/ChakraRtInterface.h
obastemur: Both were not in use for xplat static lib. This PR benefits `JsGetArrayBufferStorage` so it is updated.

bin/ch/ChakraRtInterface.h
obastemur: Could be nice. Updated only the type there. See the original naming. Besides; This is just a proxy variable name. Jsrt reference for the function is available from `ChakraCore.h`

lib/Jsrt/Jsrt.cpp
jianchun: Here it is still using cast instead of FromVar suggested by Sandeep.

lib/Runtime/Base/ScriptContext.h
jianchun: "serializedScriptBuffer" added in "ScriptContext.h" should be reverted?

jianchun:
Reviewed again and left 3 nit comments, otherwise LGTM. Thanks!

lib/Jsrt/ChakraCore.h
```diff
@@ -497,7 +492,7 @@ CHAKRA_API
 ///     Requires an active script context.
 ///     </para>
 /// </remarks>
-/// <param name="buffer">The serialized script.</param>
+/// <param name="buffer">The serialized script as an ArrayBuffer (preferably ExternalArrayBuffer)</param>
```
dilijev: nit: we're pretty consistent about ending `<param>` descriptions with a period.

lib/Jsrt/Jsrt.cpp
```diff
@@ -3248,15 +3248,15 @@ JsErrorCode RunSerializedScriptCore(
...
-            /* grfsi               */ 0
+            /* sourceContextInfo      */ sourceContextInfo,
+            /* dlnHost                */ 0,
+            /* ulColumnHost           */ 0,
+            /* lnMinHost              */ 0,
+            /* ichMinHost             */ 0,
+            /* ichLimHost             */ 0, // xplat-todo: need to compute this?
+            /* ulCharOffset           */ 0,
+            /* mod                    */ kmodGlobal,
+            /* grfsi                  */ 0
```
dilijev: nit: What prompted this change?

lib/Jsrt/Jsrt.cpp
```diff
@@ -4460,19 +4462,40 @@ CHAKRA_API JsSerialize(
...
+    {
+        return errorCode;
+    }
+
+    if (bufferSize == 0)
+    {
+        return JsErrorScriptCompile;
+    }
+
+    if ((errorCode = JsCreateArrayBuffer(bufferSize, bufferVal)) == JsNoError)
```
dilijev: I think we generally regard assignment within an if-statement as a code smell. Can you make the assignment in the line above?

lib/Jsrt/Jsrt.cpp
```diff
@@ -4509,9 +4540,17 @@ CHAKRA_API JsRunSerialized(
...
+        return JsErrorInvalidArgument;
+    }
+
+    byte* buffer = Js::ArrayBuffer::FromVar(bufferVal)->GetBuffer();
+
     return RunSerializedScriptCore(
         scriptLoadCallback, DummyScriptUnloadCallback,
-        sourceContext, // use the same user provided sourceContext as scriptLoadSourceContext
-        buffer, sourceContext, url, false, result);
+        sourceContext,// use the same user provided sourceContext as scriptLoadSourceContext
```
dilijev: nit: restore space before `//`

jianchun:
Should this go to 1.4 by this time?

obastemur:
@jianchun you mean redirect this to 1.4 and merge?

digitalinfinity:
>     // We currently don't free SRCINFO object so we don't want to add extra variables here.

Is this comment still accurate? If SRCINFO doesn't ever get freed, is there a possibility of a leak here with the new serializedScriptBuffer field?

---
Refers to: lib/Runtime/Base/ScriptContext.h:41 in ce4f197. [](commit_id = ce4f197e24a664b065d0cbb025c2e12b6d85df8d, deletion_comment = False)

obastemur:
> Is this comment still accurate? If SRCINFO doesn't ever get freed, is there a possibility of a leak here with the new serializedScriptBuffer field?

@digitalinfinity It doesn't use SRCINFO now. Most probably that was one of the reasons @jianchun was asking me to change from SRCINFO to JsrtSourceHolder.

lib/Jsrt/Jsrt.cpp
obastemur: @dilijev I'm not sure if I understood the problem here. Generally this anti pattern reduces the readability of code but here IMO it doesn't.

dilijev: It's fine either way. It's worth noting that in other places, such as above, the assignment to error code is done before the condition that checks it. Also separating the two doesn't hurt readability either.

PR #2081, opened 2016-11-22
Very Trivial Block Reordering

Krovatkin:
A very trivial and unopinionated block reordering in a sense 
* it only moves a block at a time rather than identify a longest sequence and move it all at once. The end result will be the same meaning the longest possible sequence will be moved. This approach makes the opt's logic much simpler at a slight overhead of repeating some checks. 
* it is very conservative and unjit-like; it doesn't speculate about the blocks with conditional branches to  increase fall-through block sequence and reduce the number of taken branches. 
* it has only the most basic awareness of loops to make sure it won't separate a landing pad and header
* it is disabled in generators

### DIFF
--- a/lib/Backend/FlowGraph.h
+++ b/lib/Backend/FlowGraph.h
@@ -166,7 +166,7 @@ public:
     static void  SafeRemoveInstr(IR::Instr *instr);
     void         SortLoopLists();
     FlowEdge *   FindEdge(BasicBlock *predBlock, BasicBlock *succBlock);
-
+    void         MoveBlocksBefore(BasicBlock *blockStart, BasicBlock *blockEnd, BasicBlock *insertBlock);
 #if DBG_DUMP
     void         Dump();
     void         Dump(bool verbose, const char16 *form);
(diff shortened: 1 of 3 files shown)


Krovatkin:
@LouisLaf @MikeHolman could you please take a look when you have a mnt?

MikeHolman:
Can you please give some reasoning for this change? Can you give an example where this change improves codegen?

Krovatkin:
@MikeHolman 

We've noticed that ChakraCore sometimes generates the following assembly sequence. We saw it in Octane, Massive (Asm.js), and JetStream (if I remember correctly)

ORIGINAL (2 jumps on true & 1 jump on false) : 

```asm
mov ebx, 1
comisd xmm0, xmm1
jnbe true2

false2
xor ebx, ebx
jmp false1

true2:
mov eax, 7
jmp done2

false1:
mov eax, 5

done2:
mov ecx,4 
```

The block reordering is able to rectify this to 

OPTIMIZED (1 jump on both)

```asm
mov ebx, 1
comisd xmm0, xmm1
jnbe true2

false2
xor ebx, ebx

false1:
mov eax, 5 
jmp done2

true2:
mov eax, 7

done2:
mov ecx,4 
```

Block reordering should also be able to minimize the number of jumps in longer nested if sequences, though I don't have an example off top of my head.

Krovatkin:
@MikeHolman 
If you guys decide that this opt useful I'll rebase this PR against the latest `master`. Otherwise, we could close it.

MikeHolman:
I think this seems like a pretty reasonable change. Can you also share benchmark numbers?

Krovatkin:
@MikeHolman I'm having trouble matching numbers I got back to their respective benchmarks. I'll get back to you on this in a bit 😖

MikeHolman:
Closing for now. If you can rediscover the perf wins feel free to reopen.

Issue #1811, opened 2016-10-24
xplat: Make shared library usage requirements more reasonable


obastemur:
At the moment, ChakraCore shared library on xplat requires an explicit call to DllMain. 
This is not practical for 3rd party apps using ChakraCore.

A potential solution may benefit `static library` xplat mechanism. Please keep in mind, we need to consider command line args/flags handling for SO separately.

kphillisjr:
I believe the best solution would be to use the constructor and destructor attributes. These can be prioritized, and offer a beginning solution to this issue.

```C
// Beginning of file
void SharedLibraryEntry() __attribute__((constructor));
void SharedLibraryExit() __attribute__((constructor));

// Somewhere later.
void SharedLibraryEntry()
{
	// TODO: Library initialization. This is called before Main.
}

void SharedLibraryExit()
{
	//TODO: Library Cleanup/Finalization. This is called after Main, or on unload.
}

/* This will result in following run order...
SharedLibraryEntry() 
main()
SharedLibraryExit()
*/
```

Also as a side note you can prioritize the constructor/destructor. The major part of that is that for constructors the lower number gets ran first, and on destructors the higher number is ran first. An example is as follows...

```C
void SharedLibraryEntryA() __attribute__((constructor(1)));
void SharedLibraryExitA() __attribute__((destructor(1)));
void SharedLibraryEntryB() __attribute__((constructor(2)));
void SharedLibraryExitB() __attribute__((destructor(2)));
void SharedLibraryEntryC() __attribute__((constructor(3)));
void SharedLibraryExitC() __attribute__((destructor(3)));
```

With the above declarations you will get the Following run order:
```
// Run Order:
SharedLibraryEntryA() // 1
SharedLibraryEntryB() // 2
SharedLibraryEntryC() // 3
Main()
SharedLibraryExitC() // 3
SharedLibraryExitB() // 2
SharedLibraryExitA()// 1
```

kphillisjr:
I submitted a pull request that provides a highly usable fix for this bug under #2124. The only dependency is that pull request #2111 be applied first. The reason for this is that I have already patched the ChakraCore samples to take advantage of pull request #2111, and these samples require pull request #2124 to run properly. The changes for the ChakraCore Samples are located at [GitHub: kphillisjr/Chakra-Samples at samples_cmake_v3](https://github.com/kphillisjr/Chakra-Samples/tree/samples_cmake_v3).

obastemur:
[...]

Issue #1616, opened 2016-09-21
Feature Request: Enable ChakraCore embedding using CMake.


kphillisjr:
I noticed that the current system for integrating ChakraCore using CMake is not exactly one of the most friendly. All this really boils down to is simplifying the auto-detection for CMake to where all of the platform dependent configuration is handled easily, and build options are easily configured by ChakraCore. In general this can be broken down into several tasks.

- [ ] Improve CMake build configuration on linux/OS X 
    - [ ] add basic Platform Detection options to cmake. [1]
    - [ ] Use find_package command on non-windows systems[2]
        - [ ] locate libunwind using find_package.
        - [ ] locate libuuid using find_package.
        - [ ] locate ICU Common library and Data files ( icu-uc ) using find_package
        - [ ] locate ICU Internationalization library  (icu-i18n) using find_package
	- [ ] add CPack support [3] [4]
- [ ] Add more configuration options to CMAKE. 
    - [ ] rename ChakraCore  specific options/defines.
    - [ ] add option to Disable ChakraCore install options.
    - [ ] add option to Disable ChakraCore programs.
    - [ ] add option to Disable ChakraCore  JIT features
- [ ] Implement ChakraCore embedding using CMake in a fashion similar to GLFW. [5]
    - [ ] add configuration option for embedding ChakraCore through CMake.
    - [ ] add a custom target to  copy ChakraCore.h, ChakraCommon.h, ChakraCommonWindows.h and ChakraDebug.h to a different folder when ChakraCore embedding option is true. [6]
    - [ ] Verify embedded ChakraCore using CMake works on Linux.
    - [ ] Verify embedded ChakraCore using CMake works on OS X.
    - [ ] Verify embedded ChakraCore using CMake works on Windows. [7]

[1] [CMake Wiki: How To Write Platform Checks](https://cmake.org/Wiki/CMake:How_To_Write_Platform_Checks)
[2] [CMake Wiki: How To Find Libraries](https://cmake.org/Wiki/CMake:How_To_Find_Libraries)
[3] CPack is highly useful for when users want to generate installable packages using for chakra.
[3] [CMake Wiki: Packaging With CPack](https://cmake.org/Wiki/CMake:Packaging_With_CPack)
[4] [GLFW: Building Applications](http://www.glfw.org/docs/latest/build_guide.html) 
[...]

PR #2084, opened 2016-11-23
Cmake Fixes

kphillisjr:
This includes many of the changes/fixes outlined in bug report #1616. This also includes a solution to bug report  #1811.

**NOTE 1:** many of the changes in this pull request need testing on windows.
**NOTE 2:** This should include fixes to help with running Static Library on windows, Just remember to call the new Jsrt calls before running any other jsrt function.


obastemur:
@kphillisjr couple of pointers;

- We can and should fix shared lib initialization without introducing new API to Jsrt
- Unless there is a serious reason, we wouldn't like to update preprocessors we have
- CMake defs with `_SH` ending are there because of CMake Cache cleanup. We would like to keep that.
- We have more flags on ICU support, including `no-icu` and `static`. We wouldn't like to break them.
- Shared library should be capable of initialization of the config flags as given from #1811 
- We tend to have a single commit / PR per issue

Thanks

kphillisjr:
@obastemur, I'll rework the patches and push a new set of commits. Some of the commits are no longer needed anymore.

kphillisjr:
I reworked the automatic configuration to avoid directly modifying CMakeListst.txt. The new code is found in pull #2111.

PR #2087, opened 2016-11-24
swb: fix AllocateArray annotations

jianchun:
RecyclerChecker plugin was not able to detect (write barriered) types
allocated through AllocateArray because AllocateArray was an inlined
function. The plugin works by checking "new" operators.

Fixed by enhancing the plugin to check AllocateArray. (Originally
attempted to revert AllocateArray back to macro, but run into prefast
failures.)

Fixed a plugin bug introduced in last commit that results in over
annotating. I was checking wrong bits and ended up requiring any Recycler
allocation including leaf annotation to be annotated. Fixed the bits.

Annotated newly discovered types from AllocateArray.


jianchun:

jianchun:

lib/Runtime/Base/FunctionBody.h
```diff
@@ -1179,19 +1179,19 @@ namespace Js
     struct LoopHeader
```
leirocks: >LoopHeader [](start = 11, length = 10)

do we need to move this out of FunctionBody.h?

jianchun: Double checked and this seems ok for now. "FORCE_USE_WRITE_BARRIER" starts after this class definition at line 1298. (I avoided splitting this file to prevent merge hell.)

---
In reply to: [89894881](https://github.com/Microsoft/ChakraCore/pull/2087#discussion_r89894881) [](ancestors = 89894881)

leirocks:
:shipit:

PR #2088, opened 2016-11-24
xplat: don't use private heap

ianwjhalliday:
Do not use private heap on xplat.

Currently with our static linking in node-chakracore our overridden new/delete operators are picked up by node itself.  Use of the private heap leads to instability on shutdown due to the chance that HeapAllocator::Instance is disposed before other allocations.  Using alloc and free instead circumvents this problem.


ianwjhalliday:
@digitalinfinity

digitalinfinity:
LGTM

Issue #2083, opened 2016-11-23
Setting an element outside the bounds of a frozen array does not throw an exception


EdMaurer:
Open the developer console for logs.

Call Object.freeze on an array and set an element outside of the array's bounds.
Expected result: An exception is thrown
Actual result: No exception is thrown

This does not repro when setting elements inside the bounds of the array.

        'use strict';

        var arr = ["one", "two", "three"];
        Object.freeze(arr);

        console.log("Initial arr.length: " + arr.length);

        try {
            console.log("Setting arr[0]...");
            arr[0] = "";
            console.log("No exception thrown when setting arr[0]");
        }
        catch(e) {
            console.log("Exception thrown when setting arr[0]");
        }
        console.log("arr[0] after setting - Expected: one, Actual: " + arr[0]);
        console.log("arr.length: " + arr.length);
        console.log();

        try {
            console.log("Setting arr[5]...");
            arr[5] = "";
            console.log("No exception thrown when setting arr[5]");
        }
        catch(e) {
            console.log("Exception thrown when setting arr[5]");
        }
        console.log("arr[5] after setting - Expected: undefined, Actual: " + arr[5]);
        console.log("arr.length: " + arr.length);

dilijev:
I believe this is fixed by #2089. @ThomsonTan can you confirm and close this issue?

ThomsonTan:
Yes, fix https://github.com/Microsoft/ChakraCore/pull/2089 had been fixed.

PR #2089, opened 2016-11-24
Throw TypeError when accessing array with out-of-bound index in strict mode

ThomsonTan:
This is for issue https://github.com/Microsoft/ChakraCore/issues/2083

### DIFF
--- a/lib/Runtime/Types/ES5ArrayTypeHandler.cpp
+++ b/lib/Runtime/Types/ES5ArrayTypeHandler.cpp
@@ -371,7 +371,7 @@ namespace Js
         // Reject if we need to grow non-writable length
         if (!CanSetItemAt(arr, index))
         {
-            return false;
+            return CantExtend(flags, scriptContext);
         }
 
         IndexPropertyDescriptor* descriptor;


digitalinfinity:
@dotnet-bot test this please

ThomsonTan:
@jianchun could you please help take a look?

lib/Runtime/Types/ES5ArrayTypeHandler.cpp
```diff
@@ -371,7 +371,7 @@ namespace Js
         // Reject if we need to grow non-writable length
         if (!CanSetItemAt(arr, index))
         {
```
jianchun: There are 2 more similar fragments in this file ("if (!CanSetItemAt...) { return false; }"), please double check if they need fixes too.

ThomsonTan: @jianchun I checked some others cases like Object.defineProperty which doesn't enter the other occurrences in the code. How about check in this with a comment? Probably extend it later.

ThomsonTan:
Talked with @jianchun offline and it is Ok to take the fix now.

PR #2090, opened 2016-11-24
Prevent redeferring a function body if GC happens while creating a function object of that function body

rajatd:
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/2090)


digitalinfinity:
LGTM- I'd move the new class out of the function though- I can't imagine this is the only place to guard against redeferral?

PR #2091, opened 2016-11-24
Another fix for redeferral with function objects with inline caches

rajatd:
If the function object needs to use the inline caches from the function body and the body was redeferred, point them to the undeferred function body's caches right after undeferring it.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/2091)


digitalinfinity:
LGTM

PR #2093, opened 2016-11-25
Trace: Implement built-in trace for xplat

obastemur:
Implements a standalone solution to trace function calls in a *nix application
See additional embedder notes inside the source code comments


obastemur:
moved into #2296

PR #2100, opened 2016-11-28
WASM - fast array implementation

### LINKED ISSUE #1888: WASM - fast array out of bounds traps
Use the fast array implementation to detect out of bounds access and trap.

Cellule:
This removes the bounds check in jited code on x64 Windows for heap access.

I targeted this change against release/1.4 to get the review going. 
We should discuss if we want to push this change to the release branch or if we should keep it in master 

Fixes #1888 

Since we have more control on how the buffer is created in WebAssembly then asm.js, I decided to create the class WebAssemblyArrayBuffer which would 
- allocate 8G instead of 4G of virtual memory.
- always create a buffer even if the initial size is 0.
- always use virtual memory when transfering the buffer.

@MikeHolman Are there other places other than TransferInternal that can detach the buffer ?

@liminzhu @LouisLaf @EdMaurer Do we want to integrate this change to release/1.4 since it has a massive perf benefit or is it too late ?

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/2100)


Cellule:
Closing for now.

PR #2102, opened 2016-11-28
Memory leak fixes

kunalspathak:
- Today the type transition map was a `BaseDictionary` of `DynamicType *` to `DynamicType *`. This would have kept both old and new type alive. Changed it to `WeaklyReferencedKeyDictionary`
  where key is weak reference of old `DynamicType` and value is new `DynamicType *`. This will guarantee that when oldType is collected, we no longer use its corresponding newType but instead
  create a newer type.

- Added a map on `scriptContext` to share the `JsrtExternalType` for `JsrtExternalObject` based on `finalizeCallback`. So `JsrtExternalObject` created with same callback will get same type.
  This faciliate to use `JsrtExternalObject` in type sharing for `__proto__` optimization. The map on `scriptContext` is registered on scriptContext weakReferenceDictionaryList so the entries
  inside it gets cleanup up during PostCollectionCallback.

- Add Jsrttest for setPrototype

- `PromoteType` called from `SetPrototype` was treating the instance as object literal and was creating a new `DynamicType` with `jsFunctionCallback` field of `JsrtExternalType` set to nullptr. Fixed it.

### DIFF
--- a/bin/NativeTests/JsRTApiTest.cpp
+++ b/bin/NativeTests/JsRTApiTest.cpp
@@ -1192,6 +1192,82 @@ namespace JsRTApiTest
         JsRTApiTest::RunWithAttributes(JsRTApiTest::ObjectMethodTest);
     }
 
+    void SetPrototypeTest(JsRuntimeAttributes attributes, JsRuntimeHandle runtime)
+    {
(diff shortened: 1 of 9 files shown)


kunalspathak:
Already reviewed by @digitalinfinity and @pleath .

PR #2103, opened 2016-11-28
Follow up from MSFT 9232298, MSFT 9232299, MSFT 9232230: change go.microsoft.com links to use HTTPS.

dilijev:


dilijev:
Note: CI is now working properly for release/1.4

PR #2105, opened 2016-11-28
enable write barrier test on windows

leirocks:
make built with WriteBarrierPtr<> globaly enabled -- will turn this off before merging to master
by default still use write barrier for function body and write watch for the rest
use -ForceSoftwareWriteBarrier to use write barrier for all none leaf allocation
use -KeepRecyclerTrackData to keep the recycler track data longer until reuse

### DIFF
--- a/lib/Common/Memory/Recycler.cpp
+++ b/lib/Common/Memory/Recycler.cpp
@@ -266,14 +266,20 @@ Recycler::Recycler(AllocationPolicyManager * policyManager, IdleDecommitPageAllo
     verifyEnabled =  GetRecyclerFlagsTable().IsEnabled(Js::RecyclerVerifyFlag);
     if (verifyEnabled)
     {
-        ForRecyclerPageAllocator(EnableVerify());
+        ForEachPageAllocator([](IdleDecommitPageAllocator* pageAlloc)
+        {
+            pageAlloc->EnableVerify();
+        });
     }
 #endif
 
 #ifdef RECYCLER_NO_PAGE_REUSE
     if (GetRecyclerFlagsTable().IsEnabled(Js::RecyclerNoPageReuseFlag))
     {
-        ForRecyclerPageAllocator(DisablePageReuse());
+        ForEachPageAllocator([](IdleDecommitPageAllocator* pageAlloc)
+        {
+            pageAlloc->DisablePageReuse();
+        });
     }
 #endif
 
@@ -433,7 +439,10 @@ void
 Recycler::ResetThreadId()
 {
     // Transfer all the page allocator to the current thread id
-    ForRecyclerPageAllocator(ClearConcurrentThreadId());
+    ForEachPageAllocator([](IdleDecommitPageAllocator* pageAlloc)
+    {
+        pageAlloc->ClearConcurrentThreadId();
+    });
 #if ENABLE_CONCURRENT_GC
     if (this->IsConcurrentEnabled())
     {
@@ -558,7 +567,10 @@ Recycler::~Recycler()
 
 #if DBG
     // Disable idle decommit asserts
-    ForRecyclerPageAllocator(ShutdownIdleDecommit());
+    ForEachPageAllocator([](IdleDecommitPageAllocator* pageAlloc)
+    {
+        pageAlloc->ShutdownIdleDecommit();
+    });
 #endif
     Assert(this->collectionState == CollectionStateExit || this->collectionState == CollectionStateNotCollecting);
 #if ENABLE_CONCURRENT_GC
@@ -765,7 +777,7 @@ Recycler::Initialize(const bool forceInThread, JsUtil::ThreadService *threadServ
             , captureAllocCallStack
             , captureFreeCallStack
 #endif
-            );
+        );
     }
     else
     {
@@ -775,7 +787,7 @@ Recycler::Initialize(const bool forceInThread, JsUtil::ThreadService *threadServ
             , captureAllocCallStack
             , captureFreeCallStack
 #endif
-            );
+        );
     }
 #else
     autoHeap.Initialize(this
@@ -784,7 +796,7 @@ Recycler::Initialize(const bool forceInThread, JsUtil::ThreadService *threadServ
         , captureAllocCallStack
         , captureFreeCallStack
 #endif
-        );
+    );
 #endif
 
     markContext.Init(Recycler::PrimaryMarkStackReservedPageCount);
@@ -823,7 +835,7 @@ Recycler::Initialize(const bool forceInThread, JsUtil::ThreadService *threadServ
 #endif
 #endif
 
-#if ENABLE_WRITE_WATCH
+#ifdef RECYCLER_WRITE_WATCH
     bool needWriteWatch = false;
 #endif
 
@@ -844,8 +856,8 @@ Recycler::Initialize(const bool forceInThread, JsUtil::ThreadService *threadServ
         this->disableConcurrent = true;
     }
     else if (CUSTOM_PHASE_OFF1(GetRecyclerFlagsTable(), Js::ConcurrentMarkPhase) &&
-             CUSTOM_PHASE_OFF1(GetRecyclerFlagsTable(), Js::ParallelMarkPhase) &&
-             CUSTOM_PHASE_OFF1(GetRecyclerFlagsTable(), Js::ConcurrentSweepPhase))
+        CUSTOM_PHASE_OFF1(GetRecyclerFlagsTable(), Js::ParallelMarkPhase) &&
+        CUSTOM_PHASE_OFF1(GetRecyclerFlagsTable(), Js::ConcurrentSweepPhase))
     {
         // All concurrent collection phases disabled
         this->disableConcurrent = true;
@@ -857,11 +869,8 @@ Recycler::Initialize(const bool forceInThread, JsUtil::ThreadService *threadServ
 
         if (deferThreadStartup || EnableConcurrent(threadService, false))
         {
-#if ENABLE_WRITE_WATCH
-            if (!CONFIG_FLAG(ForceSoftwareWriteBarrier))
-            {
-                needWriteWatch = true;
-            }
+#ifdef RECYCLER_WRITE_WATCH
+            needWriteWatch = true;
 #endif
         }
     }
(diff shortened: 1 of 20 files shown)


leirocks:

leirocks:

lib/Common/Memory/Recycler.h
```diff
@@ -152,25 +150,6 @@ struct InfoBitsWrapper{};
...
-#define RecyclerNewStructPlus               RecyclerNewWithBarrierStructPlus
-#define RecyclerNewArray                    RecyclerNewWithBarrierArray
-#define RecyclerNewArrayZ                   RecyclerNewWithBarrierArrayZ
-#define RecyclerNewFinalized                RecyclerNewWithBarrierFinalized
-#define RecyclerNewFinalizedPlus            RecyclerNewWithBarrierFinalizedPlus
-#define RecyclerNewTracked                  RecyclerNewWithBarrierTracked
-#define RecyclerNewEnumClass                RecyclerNewWithBarrierEnumClass
-#define RecyclerNewWithInfoBits             RecyclerNewWithBarrierWithInfoBits
-#define RecyclerNewFinalizedClientTracked   RecyclerNewWithBarrierFinalizedClientTracked
-#endif
```
jianchun: This breaks the plugin? The plugin relies on allocFunc name to tell if a type is allocated in writebarrier memory (allocFunc name contains "WithBarrier", or "InfoBitsWrapper" contains "WithBarrierBit").

If you really want this change, maybe we can change the plugin: any "Recycler" new without "Leaf" in allocFunc name, or any "InvoBitsWrapper" variant recycler new, the type must be annotated. Is this precise?

lib/Common/Memory/SmallFinalizableHeapBucket.h
```diff
@@ -174,7 +173,7 @@ class HeapBucketGroup
         typedef typename SmallHeapBlockType<objectAttributes, TBlockAttributes>::BucketType BucketType;
         static BucketType& GetBucket(HeapBucketGroup<TBlockAttributes> * heapBucketGroup)
         {
-            CompileAssert(objectAttributes & FinalizeBit);
+            Assert(objectAttributes & FinalizeBit);
```
jianchun: Why changing "CompileAssert" to "Assert"?

leirocks: because the compiler is not smart enough, that while allocating with LeafBit, we already pick the Leaf buckets via the specialization, and should not fall through here. see Recycler.inl line 97 for detail.

---
In reply to: [89901040](https://github.com/Microsoft/ChakraCore/pull/2105#discussion_r89901040) [](ancestors = 89901040)

lib/Common/Memory/Recycler.h
leirocks: I think we should make the change, currently we need the assumption that all recycler allocation are made through RecyclerNew* macros, to get the plugin work. what if there's recycler->Alloc() call? (I don't think there is one for now though), I remember there's RecyclerFastAllocator might be close to the pattern "recycler->Alloc", let me check, may be we need to do something special there

---
In reply to: [89900815](https://github.com/Microsoft/ChakraCore/pull/2105#discussion_r89900815) [](ancestors = 89900815)

lib/Common/Memory/SmallFinalizableHeapBucket.h
jianchun: I see, thanks!

---
In reply to: [89909533](https://github.com/Microsoft/ChakraCore/pull/2105#discussion_r89909533) [](ancestors = 89909533,89901040)

leirocks:

jianchun:
:shipit:

lib/Common/Memory/Recycler.inl
```diff
@@ -91,7 +91,19 @@ Recycler::AllocWithAttributesInlined(DECLSPEC_GUARD_OVERFLOW size_t size)
     }
 #endif
 
-    char* memBlock = RealAlloc<(ObjectInfoBits)(attributes & InternalObjectInfoBitMask), nothrow>(&autoHeap, allocSize);
+    char* memBlock = nullptr;
+#if GLOBAL_ENABLE_WRITE_BARRIER
+    if (CONFIG_FLAG(ForceSoftwareWriteBarrier)
+        && ((attributes & LeafBit) != LeafBit
+            || (attributes & FinalizeBit) == FinalizeBit)) // there's no Finalize Leaf bucket
```
curtisman: Why do we need to special case this?   FinalizeLeaf is allocated in the FinalizeBucket,  Shouldn't FinalizeLeaf also go to FinalizeWithWriteBarrier bucket with ForceSoftwareWriteBarrier?

lib/Common/Memory/RecyclerWriteBarrierManager.cpp
```diff
@@ -66,6 +66,16 @@ X64WriteBarrierCardTableManager::OnThreadInit()
...
 
+#ifdef X64_WB_DIAG
+    this->_stackbase = (char*)stackBase;
+    this->_stacklimit = (char*)stackEnd;
+#endif
+
+    // on Windows server 2012 stack limit can expand with process running, and causes
+    // accessing uncommitted card table page.
+    // TODO: use VirtualQuery twice to get the max possible stack limit
+    stackEnd -= AutoSystemInfo::PageSize * AutoSystemInfo::PageSize;
```
curtisman: Huh?  Where does this value come from?

lib/Common/Memory/SmallFinalizableHeapBucket.h
curtisman: But the compiler has to know. because to instantiate this function it needs to know the value of objectAttributes already?   If there is an compiler error, you are calling the function with the incorrect template parameter.

lib/Runtime/Library/ArrayBuffer.h
```diff
@@ -150,7 +150,7 @@ namespace Js
         Field(JsUtil::List<RecyclerWeakReference<ArrayBufferParent>*>*) otherParents;
 
 
-        Field(BYTE  *) buffer;             // Points to a heap allocated RGBA buffer, can be null
+        FieldNoBarrier(BYTE*) buffer;             // Points to a heap allocated RGBA buffer, can be null
```
curtisman: In your comment: can this be any raw buffer instead of RGBA buffer?

lib/Common/Memory/Recycler.inl
leirocks: yes, that's why special case this right? :) 
Leaf (without Finalize)->leaf bucket, 
Finalize (whatever leaf or not) -> FinalizeBucket(with barrier here) 

 a little bit confusing though, let me make a more detailed comment in my next change


---
In reply to: [90006103](https://github.com/Microsoft/ChakraCore/pull/2105#discussion_r90006103) [](ancestors = 90006103)

lib/Common/Memory/RecyclerWriteBarrierManager.cpp
leirocks: the value is one page of card table, that represents 16MB address space, should be enough with stack growth. otherwise we can use virtual query to get the possible growth limit, which can be huge. 

---
In reply to: [90006505](https://github.com/Microsoft/ChakraCore/pull/2105#discussion_r90006505) [](ancestors = 90006505)

lib/Common/Memory/SmallFinalizableHeapBucket.h
leirocks: we are not actually calling, with following:

#if GLOBAL_ENABLE_WRITE_BARRIER
    if (CONFIG_FLAG(ForceSoftwareWriteBarrier)
        && ((attributes & LeafBit) != LeafBit
            || (attributes & FinalizeBit) == FinalizeBit)) // there's no Finalize Leaf bucket
    {
        memBlock = RealAlloc<(ObjectInfoBits)((attributes | WithBarrierBit) & InternalObjectInfoBitMask), nothrow>(&autoHeap, allocSize);
    }
    else
#endif
    {
        memBlock = RealAlloc<(ObjectInfoBits)(attributes & InternalObjectInfoBitMask), nothrow>(&autoHeap, allocSize);
    }


the LeafBit without FinalizeBit has already going through the path with LeafBit specialization and won't fall through here. when compiler doing optimization, it should remove such symbols. even if it's not, the runtime assertion should catch it.
just because the compiler didn't analyze the 'if' statement while expanding the template. 

more clear way to explain this is, compiler thought there's (WithBarrierBit | LeafBit) combination (deducting from (attributes | WithBarrierBit)), which won't hit the LeafBit specialization and fall through the default one which we Assertion on FinalizeBit is set.
however we actually don't have (WithBarrierBit | LeafBit) combination, from the if statement, if the original attribute is LeafBit only, I don't adding WithBarrierBit even with ForceSoftwareWriteBarrier on


---
In reply to: [90006829](https://github.com/Microsoft/ChakraCore/pull/2105#discussion_r90006829) [](ancestors = 90006829)

lib/Common/Memory/Recycler.cpp
```diff
@@ -896,6 +902,130 @@ Recycler::Initialize(const bool forceInThread, JsUtil::ThreadService *threadServ
...
+{
+    // TODO: SWB this is for Finalizable leaf allocation, which we didn't implement leaf bucket for it
+    // remove this after the finalizable leaf bucket is implemented
+#if GLOBAL_ENABLE_WRITE_BARRIER
+    if (CONFIG_FLAG(ForceSoftwareWriteBarrier))
+    {
+        return &this->recyclerWithBarrierPageAllocator;
+    }
+#endif
+    else
```
ThomsonTan: >else [](start = 4, length = 4)

Could it be possible that "else{}" is invalid because GLOBAL_ENABLE_WRITE_BARRIER is not defined?

lib/Common/Memory/Recycler.h
ThomsonTan: Which plugin depends on this?

---
In reply to: [89916552](https://github.com/Microsoft/ChakraCore/pull/2105#discussion_r89916552) [](ancestors = 89916552,89911621,89900815)

lib/Common/Memory/RecyclerWriteBarrierManager.cpp
```diff
@@ -66,6 +66,16 @@ X64WriteBarrierCardTableManager::OnThreadInit()
     ::GetCurrentThreadStackLimits(&stackEnd, &stackBase);
 #endif
 
+#ifdef X64_WB_DIAG
+    this->_stackbase = (char*)stackBase;
+    this->_stacklimit = (char*)stackEnd;
+#endif
+
+    // on Windows server 2012 stack limit can expand with process running, and causes
```
ThomsonTan: >Windows server 2012 [](start = 10, length = 19)

Is growing stack limit only a feature for Windows server 2012, ant not for all others?

leirocks: I only hit this on CI build which is running windows server 2012. it might happen in win10 as well but I didn't get a repro on my machine for running 1000 times of the test

---
In reply to: [90075652](https://github.com/Microsoft/ChakraCore/pull/2105#discussion_r90075652) [](ancestors = 90075652)

lib/Common/Memory/Recycler.h
leirocks: our clang plugin to check recycler pointer usage

---
In reply to: [90073284](https://github.com/Microsoft/ChakraCore/pull/2105#discussion_r90073284) [](ancestors = 90073284,89916552,89911621,89900815)

Issue #2101, opened 2016-11-28
WASM - exported function properties


Cellule:
Exported Functions also have the following data properties: the length property is set to the exported function's signature's arity
the name is set to index as a Number value

PR #2106, opened 2016-11-28
reassigning name and length props for exported functions

Krovatkin:


Krovatkin:
@Cellule @MikeHolman Took the first crack at Issue #2101 
Unfortunately, it looks like `SetPropertyWithAttributes` updates the type of a function and an entry point is set on the new type. Yet, the next time the same WASM function is exported the old type is re-used and the old type's entry point triggers bytecode generation again. 
My current solution to this is a cludge, since I'm super familiar with property APIs 😞

Krovatkin:
@dotnet-bot test Ubuntu tests please

dilijev:
@dotnet-bot test this please
(Testing to see what the current state is.)

Cellule:
I started testing/investigating this, but I hit a crash #2451 which prevented me from looking deeper until it was fixed

dilijev:
Reopened to reset CI

Krovatkin:
@dilijev @Cellule I actually might need some guidance to finish up this one. The issue is that we reusing the same function body for different function bodies (when multiple instances of the same module are instantiated, for instance). The function body contains an entry point info. Adding either length or name properties via a call to `SetPropertyWithAttributes` (seems to be the only way to modify system properties) results in replacing the existing type/type handler and entry point info (entryPoint in DynamicType and ProxyEntryPointInfo in ScriptFunctionType) with brand new objects.  When a Web Assembly method is called, thunks (a deferred thunk, AsmJsExternalEntry, DynamicInterpreterThunk) update the entry point info in both, function body and function objects). This behaviour creates inconsistency between entry point information in the function body object and entry point info in the function object which typically manifests as a crash or an assertion depending on an order web assembly modules are instantiated and their methods are called. 

I'm trying to figure out what the easiest way to deal with this issue.

dilijev:
I think this should be retargeted to `master` because it's too large a change at this time in the stabilization branch `release/1.4`. Marked milestone as `vNext` to capture this.

Cellule:
@Krovatkin @dilijev I think we should simply close it for now since it is not ready. 
@Krovatkin I will, once rs2 settles, re investigate this and will use your changes as a start if that's alright with you

Krovatkin:
@dilijev Yes, I was told that I would need to rebase it against `master`. I was wondering if we could keep this PR open for a little longer so we could use it as a starting point for a discussion on the issue I described in https://github.com/Microsoft/ChakraCore/pull/2106#issuecomment-277395586 ?

dilijev:
I don't see any reason not to leave it open, but it should be retargeted to `master` since we keep picking this up in our queries for PRs targeted to `release/1.4` which we're trying to drive to completion. :)

Krovatkin:
@Cellule @dilijev Alright, let's close this one, then and, in one-two weeks (after rs2 settles), I'll open a new PR w/ this change rebased against `master` Hopefully, I'll come up with a less ugly to way to deal with the issue by that time, so you guys won't have to invest too much time in this.

dilijev:
Yep, in your new issue you can reference the discussion in this issue, so we won't lose any context. Thanks @Krovatkin

PR #2108, opened 2016-11-29
Redeferral checks should also be done on LoopEntryPointInfo's

rajatd:
While deciding whether we can redefer a function body or not, if none of the function entry point infos suggest that the function could be jitted while we redefer the body, we should also do the same checks for loop entry point infos registered with the function body

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/2108)

### DIFF
--- a/lib/Runtime/Base/FunctionBody.cpp
+++ b/lib/Runtime/Base/FunctionBody.cpp
@@ -798,8 +798,6 @@ namespace Js
 
     bool FunctionBody::DoRedeferFunction(uint inactiveThreshold) const
     {
-        bool isJitCandidate = false;
-
         if (!(this->GetFunctionInfo()->GetFunctionProxy() == this &&
               this->CanBeDeferred() &&
               this->GetByteCode() &&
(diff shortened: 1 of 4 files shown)


rajatd:
@dotnet-bot test this please

lib/Runtime/Base/FunctionBody.h
```diff
@@ -591,6 +591,9 @@ namespace Js
 #endif
     public:
         uint frameHeight;
+#if ENABLE_NATIVE_CODEGEN
+        bool nativeEntryPointProcessed;
```
pleath: Do you know how we got away with having this only on FunctionEntryPointInfo? Does it never get used on loop entry points?

rajatd: It was really only used, among other things, to decide whether to expire a function entry point info. LoopEntryPointInfo doesn't have an Expire method.

lib/Runtime/Base/FunctionBody.cpp
```diff
@@ -828,6 +828,22 @@ namespace Js
...
+        if (!isJitCandidate)
+        {
+            // Now check loop body entry points
+            this->MapLoopHeaders([&](uint loopNumber, LoopHeader* header)
+            {
+                header->MapEntryPoints([&](int index, LoopEntryPointInfo* entryPointInfo)
+                {
+                    if ((entryPointInfo->IsCodeGenPending() && isJitModeFunction) || entryPointInfo->IsCodeGenQueued() || entryPointInfo->IsCodeGenRecorded() || (entryPointInfo->IsCodeGenDone() && !entryPointInfo->nativeEntryPointProcessed))
+                    {
+                        isJitCandidate = true;
```
ThomsonTan: >isJitCandidate = true; [](start = 24, length = 22)

break or return from the lambda, and no need to check the remaining loops if any?

ThomsonTan: >isJitCandidate = true; [](start = 24, length = 22)

Does this imply if any loop in a function is JIT candidate, we cannot redefer the whole function?

lib/Runtime/Base/FunctionBody.cpp
```diff
@@ -828,6 +828,22 @@ namespace Js
...
+                    }
+                });
+            });
+        }
+
         if (isJitCandidate)
```
ThomsonTan: >if (isJitCandidate) [](start = 8, length = 19)

Simplify this as "return !isJitCandidate"?

pleath:
LGTM. Thanks.

rajatd:
@dotnet-bot test this please

rajatd:
@dotnet-bot test OSX osx_osx_test please

lib/Runtime/Base/FunctionBody.h
```diff
@@ -591,6 +591,9 @@ namespace Js
...
+#endif
```
ThomsonTan: >ENABLE_NATIVE_CODEGEN [](start = 4, length = 21)

Why need nativeEntryPointProcessed if ENABLE_NATIVE_CODEGEN is not defined?

rajatd: That's what I had thought too. But, it has been used in places in the code without checking for ENABLE_NATIVE_CODEGEN. I didn't want to change all those places.

PR #2113, opened 2016-11-29
SAL annotation fixes

meg-gupta:


lib/Backend/InterpreterThunkEmitter.cpp
```diff
@@ -393,7 +393,7 @@ void InterpreterThunkEmitter::FillBuffer(
     _In_ bool asmJsThunk,
     _In_ intptr_t finalAddr,
     _In_ size_t bufferSize,
-    _Out_writes_bytes_all_(bufferSize) BYTE* buffer,
+    _Out_writes_bytes_all_(BlockSize) BYTE* buffer,
```
ThomsonTan: >buffer [](start = 44, length = 6)

Could the function use buffer out side of bufferSize?

meg-gupta: Doesn't look like. It doesn't uses BlockSize.

meg-gupta:
@curtisman , @ThomsonTan Can you please review ?

lib/Backend/InterpreterThunkEmitter.cpp
ThomsonTan: Why annotating buffer with `BlockSize` if it doesn't use `BlockSize`?

ThomsonTan:
:shipit:

PR #2114, opened 2016-11-29
Prevent redeferral of FunctionBody's that are being processed when GC happens. 

pleath:
Fix a class of bugs involving redeferral of functions that are being accessed by some runtime method that allocates memory, triggering GC and possibly redeferral. When a GC that will result in redeferral scans the stack, mark any FunctionBody's we find as having active references. Do this by for now adding a method to FinalizableObject and overriding it in FunctionBody. We'll replace this with a new object info bit if and when bits become available.


pleath:
@curtisman , @rajatd 
Thanks.

rajatd:
LGTM. You can also get rid of the AutoDisableRedeferral class and its objects, or defer it until we are certain of stability of this change, upto you.

PR #2115, opened 2016-11-29
Fix crash due to label breaking precise captures

ianwjhalliday:
Precise closure capture breaks when a label exists with the same name as a variable or function.

Label parsing used to rely on ParseExpr(), checking after the fact to see if the expression was a single identifier and if it is followed by a tkColon. Since it parsed, however, a pidref was pushed for the identifier setting up a new reference to that identifier which could then trick the front-end into thinking there was a closure capture where there actually wasn't.  This leads to invalid StSlots and LdSlots.

Fix it by looking for a label before parsing as an expression statement. Labels are by spec exactly a tkID followed by a tkColon.  Chakra unfortunately allows parenthesized label names, so we also check for balanced open and close parentheses here to maintain that behavior.  We will remove this legacy behavior in a future release.

If we don't find a label pattern in the syntax then we rewind and parse as an expression statement.

### DIFF
--- a/lib/Parser/Parse.cpp
+++ b/lib/Parser/Parse.cpp
@@ -9389,7 +9389,7 @@ LFunctionStatement:
         default:
             {
 LDefaultTokenFor:
-               RestorePoint exprStart;
+                RestorePoint exprStart;
                 tokens beforeToken = tok;
                 m_pscan->Capture(&exprStart);
                 if (IsPossiblePatternStart())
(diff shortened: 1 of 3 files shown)


ianwjhalliday:
@pleath @Microsoft/chakra-es please review

lib/Parser/Parse.cpp
```diff
@@ -10110,9 +10110,79 @@ ParseNodePtr Parser::ParseStatement()
...
+                        if (PnodeLabelNoAST(&tokInner, pLabelIdList))
+                        {
+                            Error(ERRbadLabel);
+                        }
+                        LabelId* pLabelId = CreateLabelId(&tokInner);
+                        pLabelId->next = pLabelIdList;
+                        pLabelIdList = pLabelId;
+                    }
+                    m_pscan->Scan();
+                    goto LRestart;
```
akroshg: where this goto goes, do you need to call SeekTo before we go to LRestart?

lib/Parser/Parse.cpp
```diff
@@ -10110,9 +10110,79 @@ ParseNodePtr Parser::ParseStatement()
...
+        if (m_token.tk == tkLParen || m_token.tk == tkID)
+        {
+            RestorePoint idStart;
+            m_pscan->Capture(&idStart);
+
+            // Support legacy behavior of allowing parentheses around label identifiers.
+            // Require balanced parentheses for correcting parsing.  Note unbalanced cases
+            // take care of themselves correctly by resulting in rewind and parsing as
+            // an expression statement.
+            // REVIEW[ianhall]: Can this legacy functionality be removed? Chrome does not support this parsing behavior.
```
akroshg: We should remove this legacy behavior. (I didn't even know this was existed)

akroshg:
LGTM.

---
In reply to: [263724757](https://github.com/Microsoft/ChakraCore/pull/2115#issuecomment-263724757) [](ancestors = 263724757)

lib/Parser/Parse.cpp
tcare: Agreed. I don't think it would be that risky.

test/Closures/rlexe.xml
```diff
@@ -158,4 +158,10 @@
       <tags>BugFix</tags>
     </default>
   </test>
+  <test>
+    <default>
+      <files>bug_OS_9008744.js</files>
+      <tags>BugFix,exclude_dynapogo</tags>
```
tcare: Do we want to exercise both the defer and non-defer paths here?

lib/Parser/Parse.cpp
ianwjhalliday: No, this is the positive case where we've found a label.  The above code includes the new label in the current label list (which is tracked differently between buildAST and !buildAST), and then it goes back to restart statement parsing to look at the next (labeled) statement.

lib/Parser/Parse.cpp
tcare: Sounds good. Would you mind creating an issue?

test/Closures/rlexe.xml
ianwjhalliday: We could I suppose.  The parsing logic that was causing the bug is now independent of buildAST flag though.  What do you think?

tcare: Fair enough. I wish we had a forcedeferparse/nodeferparse/forceundodefer nightly run :)

ianwjhalliday: Maybe we should do that?  I think we're not seeing a whole lot of deferred parse bugs these days anymore though.

In my ideal world our parser has no !buildAST unique code paths.  Only code paths for buildAST that do the actual AST construction.  All other logic is independent.  That would significantly reduce deferparse bugs.

tcare: I agree. And you're right we haven't had any recently, but I'd say that's mostly because we haven't been doing much language work. Enabling these runs is non-trivial I'm afraid because of a ton of tests that need to be examined and excluded.

lib/Parser/Parse.cpp
ianwjhalliday: Yep, #2119

tcare:
LGTM

PR #2116, opened 2016-11-29
WASM - Debugger

Cellule:
Do not crash when debugging with WebAssembly functions present.
Correctly reset the entry points for the Wasm function when attaching/detaching the debugger.

Tested in Edge with AngryBots

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/2116)


lib/Runtime/Base/CrossSite.cpp
```diff
@@ -313,6 +313,21 @@ namespace Js
 
         TTD_XSITE_LOG(callable->GetScriptContext(), "DefaultOrProfileThunk", callable);
 
+#ifdef ENABLE_WASM
+        if (AsmJsScriptFunction::IsWasmScriptFunction(function))
```
MikeHolman: can we avoid coming here at all?

Cellule: I am not entirely sure we come here at all, I was trying to play it safe.

PR #2117, opened 2016-11-30
Fix LdLen operand tags

Penguinwizzard:
In the change to defer handling of length field loads until later in the optimizer, the need arose to move from a LdFld on a PropertySym SymOpnd to a StackSym RegOpnd (ldfld a->length to ldlen a). Unfortunately, in this transition a few flags were lost on the operand; the most notable is the isJITOptimized flag, which was causing an assert to be hit a good number of times.

### DIFF
--- a/lib/Backend/GlobOpt.cpp
+++ b/lib/Backend/GlobOpt.cpp
@@ -4909,30 +4909,37 @@ GlobOpt::OptInstr(IR::Instr *&instr, bool* isInstrRemoved)
     // Change LdFld on arrays, strings, and 'arguments' to LdLen when we're accessing the .length field
     if ((instr->GetSrc1() && instr->GetSrc1()->IsSymOpnd() && instr->m_opcode == Js::OpCode::ProfiledLdFld) || instr->m_opcode == Js::OpCode::LdFld || instr->m_opcode == Js::OpCode::ScopedLdFld)
     {
-        IR::Opnd * opnd = instr->GetSrc1();
-        Sym *sym = opnd->AsSymOpnd()->m_sym;
+        IR::SymOpnd * opnd = instr->GetSrc1()->AsSymOpnd();
+        Sym *sym = opnd->m_sym;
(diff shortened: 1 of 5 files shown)


Penguinwizzard:
@rajatd @LouisLaf please review

lib/Backend/Sym.cpp
```diff
@@ -1090,9 +1091,14 @@ Sym::Dump(IRDumpFlags flags, const ValueType valueType)
                             uint index = stackSym->GetByteCodeRegSlot() - stackSym->GetByteCodeFunc()->GetJITFunctionBody()->GetConstCount();
                             Js::PropertyId propertyId = functionBody->GetPropertyIdOnRegSlotsContainer()->propertyIdsForRegSlots[index];
                             Output::Print(_u("(%s)"), stackSym->GetByteCodeFunc()->GetInProcThreadContext()->GetPropertyRecord(propertyId)->GetBuffer());
+                            printed = true;
                         }
                     }
                 }
+                if (!printed)
+                {
+                    Output::Print(_u("{%d}"), stackSym->GetByteCodeRegSlot());
```
ThomsonTan: >{%d} [](start = 38, length = 4)

Is it necessary to introduce the new format `{%d}` which doesn't look like a general usage?

Penguinwizzard: Good point; I've removed the added debug code for the time being. It might be useful to add something like this in the future though (under an additional flag, perhaps), in order to aid in diagnosis of bytecodereg lifetime issues.

LouisLaf:
>     }

BTW, it'd be good to move this whole if() to a separate function in an attempt to make OptInstr() more readable.

---
Refers to: lib/Backend/GlobOpt.cpp:4949 in 9809b1d. [](commit_id = 9809b1de0c1b1115d9b8c18c6efe79298854ba42, deletion_comment = False)

LouisLaf:
:shipit:

PR #2121, opened 2016-11-30
Removed U+180E MONGOLIAN VOWEL SEPARATOR from Whitespace classification.

### LINKED ISSUE #2120: [RegExp] Remove U+180E MONGOLIAN VOWEL SEPARATOR from Whitespace classification.
This character was recategorized from Zs to Cf in Unicode 6.3, and remains categorized as such in Unicode 9.0 (current target version). The current version of ECMAScript requires that only characters classed as Zs in the target version of Unicode be recognized as Whitespace. This change is now reflected in Test262 so making this change will improve our Test262 score rather than regress it.

See: https://github.com/tc39/ecma262/pull/300

See: https://github.com/tc39/test262/commit/3a5a09eba2adb6a8c24cf6e603aefe9bfa8d75b9

See: https://github.com/mathiasbynens/regexpu-core/commit/9b10d2a597d4e56b236cae0f9aa2f21a9c6d1122

Discovered while working on #1181

dilijev:
This character was recategorized from Zs to Cf in Unicode 6.3, and remains categorized as such in Unicode 9.0 (current target version). The current version of ECMAScript requires that only characters classed as Zs in the target version of Unicode be recognized as Whitespace. This change is now reflected in Test262 so making this change will improve our Test262 score rather than regress it.

Additionally, update comments about location of UnicodeData.txt and about the definition of Whitespace characters.

See: https://github.com/tc39/ecma262/pull/300

See: https://github.com/tc39/test262/commit/3a5a09eba2adb6a8c24cf6e603aefe9bfa8d75b9

See: https://github.com/mathiasbynens/regexpu-core/commit/9b10d2a597d4e56b236cae0f9aa2f21a9c6d1122

Fixes #2120


lib/Parser/CaseInsensitive.cpp
```diff
@@ -206,7 +206,7 @@ END {
 */
 
         // For case-folding entries, version 8.0.0 of CaseFolding.txt located at [1] was used.
```
dilijev: Need to update this version number as well as the link. I will be updating this transforms table in a subsequent PR, so I'll revert this number to 8.0.0 for now.

PR #2111, opened 2016-11-29
Cmake Automatic Configuration

kphillisjr:
This is a rework for the ChakraCore automatic configuration. This is two patches that will allow many developers to quickly start developing using ChakraCore. I will be pushing a rework of the patches to the ChakraCore examples later, however the general process for this is as follows:

An example of how to use this is as follows...
```CMake
# include Auto-Configure macro.
include("${CMAKE_SOURCE_DIR}/ChakraCore/ChakraAutoConfigure.cmake")
# Configure ChakraCore:
CHAKRACORE_AUTOCONFIGURE()
add_subdirectory(ChakraCore)
add_custom_target(CopyChakraHeaders ALL
	COMMAND ${CMAKE_COMMAND} -E make_directory "${CMAKE_BINARY_DIR}/include"
	COMMAND ${CMAKE_COMMAND} -E copy_if_different
		"${CMAKE_SOURCE_DIR}/ChakraCore/lib/Jsrt/ChakraCore.h"
		"${CMAKE_BINARY_DIR}/include"
	COMMAND ${CMAKE_COMMAND} -E copy_if_different
		"${CMAKE_SOURCE_DIR}/ChakraCore/lib/Jsrt/ChakraCommon.h"
		"${CMAKE_BINARY_DIR}/include"
	COMMAND ${CMAKE_COMMAND} -E copy_if_different
		"${CMAKE_SOURCE_DIR}/ChakraCore/lib/Jsrt/ChakraCommonWindows.h"
		"${CMAKE_BINARY_DIR}/include"
	COMMAND ${CMAKE_COMMAND} -E copy_if_different
		"${CMAKE_SOURCE_DIR}/ChakraCore/lib/Jsrt/ChakraDebug.h"
		"${CMAKE_BINARY_DIR}/include"
)
# enable C++11
add_executable(ChakraSample sample.cpp)
set_target_properties(ChakraSample PROPERTIES COMPILE_FLAGS "-std=c++11")
# link to ChakraCore.
target_link_libraries(ChakraSample ChakraCore)
# Make sure Chakra Core Headers are in a location that is easy to use.
add_dependencies(ChakraSample CopyChakraHeaders)
# Actually include the headers.
target_include_directories(ChakraSample PUBLIC "${CMAKE_BINARY_DIR}/include")
```

PR #2123, opened 2016-11-30
Fix issue 1811

kphillisjr:
This implements the changes required to fix issue #1811. There is only two changes and this ensures that ch runs correctly after on Linux/OSX.

Also, this pull request is branched off of another pull request ( see: https://github.com/Microsoft/ChakraCore/pull/2111 )

### DIFF
--- a/lib/Runtime/Language/DynamicProfileStorage.cpp
+++ b/lib/Runtime/Language/DynamicProfileStorage.cpp
@@ -296,7 +296,7 @@ char16 const * DynamicProfileStorage::GetMessageType()
 
 bool DynamicProfileStorage::Initialize()
 {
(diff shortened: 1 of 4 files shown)


lib/Runtime/Language/DynamicProfileStorage.cpp
```diff
@@ -296,7 +296,7 @@ char16 const * DynamicProfileStorage::GetMessageType()
 
 bool DynamicProfileStorage::Initialize()
 {
-    AssertMsg(!initialized, "Initialize called multiple times");
```
kphillisjr: This change was supposed to be an definition check against _WIN32. For xplat this function is safe to call repeatedly.

kphillisjr:
Resubmitted as pull request  #2124

PR #2124, opened 2016-11-30
Fix issue 1811 -- Version 2

kphillisjr:
This implements the changes required to fix issue #1811. There is only two changes and this ensures that ch runs correctly after on Linux/OSX.

Also, this pull request is branched off of another pull request ( see: #2111 )

**NOTE:** The only change between this version and the other is that the DynamicProfileStorage change is only applied to xplat.

### DIFF
--- a/ChakraAutoConfigure.cmake
+++ b/ChakraAutoConfigure.cmake
@@ -1,0 +1,68 @@
+
+
+# Simple Macro that Autoconfigures ChakraCore
+MACRO(CHAKRACORE_AUTOCONFIGURE)
+  #-------------------------------
+  # BEGIN CPU Autodetection
+  #-------------------------------
+
+  # Clear up variables that get Modified by this Macro.
+  unset(CC_TARGETS_X86_SH CACHE)
+  unset(CC_TARGETS_AMD64_SH CACHE)
+  unset(CC_TARGETS_X86 CACHE)
+  unset(CC_TARGETS_AMD64 CACHE)
+
+  # Configure CPU Architecture.
+  # Currently only x86 and AMD64 are all that are autocnfigured.
+  if( ${CMAKE_SYSTEM_PROCESSOR} MATCHES "x86_64|amd64|AMD64")
+    set(CC_TARGETS_AMD64 1)
+    message(STATUS "ChakraCore: Building 64-bit x86 code (AMD64)")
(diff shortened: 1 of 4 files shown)


obastemur:
@kphillisjr could you please point me the part that fixes #1811 ? In other words, why we are doing these changes?

kphillisjr:
The commit two commits specific to issue #1811 are https://github.com/Microsoft/ChakraCore/pull/2124/commits/75182f3e50280ac154d9397ba381bd4272ecaee9 and https://github.com/Microsoft/ChakraCore/pull/2124/commits/b45d0b5b9886eaaa4976b47f2b6935b076545a00. However the rest of the commits are for the earlier pull request (#2111) which is in regards to issue #1616.

obastemur:
@kphillisjr I do really appreciate the overall effort.

> Simplify static linking of ChakraCore using CMake.
> Automatically Copy ChakraCore Headers to a better Location that will help prevent excessive header exposure outside the ChakraCore environment.

From these two, we could use a solution like this; After build is done, copy header files to BuildLinux/<BUILD_TYPE>/include folder. 

Feel free to send a single commit PR only doing this. You may even create a Github issue on this particular issue and send a PR separately fixing that issue.

>    Improve automatic Detection of libunwind on xplat.
>    Improve automatic Detection of libuuid on xplat

We already removed our dependency to libunwind. We may approach libuuid in a similar manner. So, we better skip this.

> Add option to disable building of optional ChakraCore elements - This is currently GCStress, and ch on xplat.

IMO; We may need to skip this (too) for now as it kinda complicates the number of build arguments.

> Implement support of building ChakraCore using CMake and MSVC.

We have so many differences between Windows and Xplat builds. IMO `msbuild` works really good with Visual Studio sln files. If we could have a batch file similar to build.sh on Windows, that could be nice to have. We could use `msbuild` with that batch file. I was planning to contribute this as a weekend fun but if you are into doing so, go ahead!

something like;
```
build.bat /--test-build/--debug/--release/ --arch=x86/amd64/ARM 
```

Actually we could combine batch file and sh file and have a single file instead of build.bat and build.sh separately. There are couple of ways to trigger either Windows or Bash specific command set under the same file. This one is not that critical, once we have the build.bat is contributed, I can also combine them later.

Thanks again.

obastemur:
https://github.com/Microsoft/ChakraCore/issues/2575#issuecomment-283128713

PR #2129, opened 2016-12-01
xplat: add ChakraCoreHeaders target.

kphillisjr:
Currently only the ChakraCore shared library uses this target, however this
target is also extremely useful for Static Library builds. To use this
target with any other target library/executable, simply add the following
line to cmake:

add_dependencies(myTarget ChakraCoreHeaders)

/cc @obastemur - Here is your request for adding a target for ChakraCoreHeaders. Also, this is future proofed a little for if/when using cmake to target MSVC/windows is added.

### DIFF
--- a/bin/ChakraCore/CMakeLists.txt
+++ b/bin/ChakraCore/CMakeLists.txt
@@ -75,3 +75,6 @@ if(NOT CC_XCODE_PROJECT)
     ${CHAKRACORE_BINARY_DIR}/
     )
 endif()
+
+add_dependencies(ChakraCore ChakraCoreHeaders)
+
--- a/lib/Jsrt/CMakeLists.txt
+++ b/lib/Jsrt/CMakeLists.txt
@@ -52,3 +52,25 @@ target_include_directories (
     ../Runtime/Debug
     ../Parser
     )
+
+
+add_custom_target(ChakraCoreHeaders ALL
+  COMMAND ${CMAKE_COMMAND} -E make_directory "${CMAKE_BINARY_DIR}/include"
+  COMMAND ${CMAKE_COMMAND} -E copy_if_different
+    "${CMAKE_CURRENT_SOURCE_DIR}/ChakraCore.h"
+    "${CMAKE_BINARY_DIR}/include"
+  COMMAND ${CMAKE_COMMAND} -E copy_if_different
+    "${CMAKE_CURRENT_SOURCE_DIR}/ChakraCommon.h"
+    "${CMAKE_BINARY_DIR}/include"
+  COMMAND ${CMAKE_COMMAND} -E copy_if_different
+    "${CMAKE_CURRENT_SOURCE_DIR}/ChakraDebug.h"
+    "${CMAKE_BINARY_DIR}/include"
+  )
+
+if(MSVC OR WIN32)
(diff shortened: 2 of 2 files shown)


lib/Jsrt/CMakeLists.txt
```diff
@@ -52,3 +52,25 @@ target_include_directories (
     ../Runtime/Debug
     ../Parser
     )
+
+
+add_custom_target(ChakraCoreHeaders ALL
```
obastemur: We need to have this under `if(NOT CC_XCODE_PROJECT)` otherwise xcode build would break.

lib/Jsrt/CMakeLists.txt
```diff
@@ -52,3 +52,25 @@ target_include_directories (
...
+    "${CMAKE_BINARY_DIR}/include"
+  COMMAND ${CMAKE_COMMAND} -E copy_if_different
+    "${CMAKE_CURRENT_SOURCE_DIR}/ChakraCommon.h"
+    "${CMAKE_BINARY_DIR}/include"
+  COMMAND ${CMAKE_COMMAND} -E copy_if_different
+    "${CMAKE_CURRENT_SOURCE_DIR}/ChakraDebug.h"
+    "${CMAKE_BINARY_DIR}/include"
+  )
+
+if(MSVC OR WIN32)
```
obastemur: This is not necessary better not add.

bin/ChakraCore/CMakeLists.txt
```diff
@@ -75,3 +75,6 @@ if(NOT CC_XCODE_PROJECT)
     ${CHAKRACORE_BINARY_DIR}/
     )
 endif()
+
+add_dependencies(ChakraCore ChakraCoreHeaders)
```
obastemur: Do we need to add the same to `/bin/ch/CMakeLists.txt` for STATIC builds?

kphillisjr: I am in the process of adding major revisions to the xplat build system. These revisions have three main features...

1. Greatly reduces compile time for swap between Shared and Static Libraries... This means that a minimal number of files get changed each time.
2. A new ChakraCoreStatic target. This is simply the ChakraCore target with the static preprocessor define added.
3. The number of static libraries is reduced by converting the existing library to an object.

lib/Jsrt/CMakeLists.txt
kphillisjr: I got ChakraCore to compile using MSVC 2015 and cmake, so this change is useful.

obastemur: @kphillisjr I'm not saying we can't do that. The problem is that preprocessor flags we set and source files we do include on CMake xplat is not necessarily matching what we supposed to do for Windows. For the best ChakraCore experience on Windows, we better use msbuild with sln file.

bin/ChakraCore/CMakeLists.txt
obastemur: We better add it here now and if we another PR will change that later, that is something else.

obastemur:
@kphillisjr ping?

kphillisjr:
Closing this pull request... It is superseded by pull request #2678.

PR #2130, opened 2016-12-01
xplat: reduce source changes between Static and Shared Libraries.

kphillisjr:
This greatly reduces the complexity of switching between static and shared libraries by converting most of the static libraries to Shared libraries. Now it is possible to test a static build of chakracore by simply linking
aginst libChakraCoreStatic. Also, this greatly reduces compile time when changing between the two since the changes between static and shared libraries has been isolated to within the bin folder.

As a quick note, This patch greatly reduces the runtime to run the Following commands...
``` sh
./build.sh -n
./build.sh -n --static
./build.sh -n --debug
./build.sh -n --static --debug
```

Also as a quick note, This has not been tested against OSX/Windows, but I do not believe this will break any builds.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -167,10 +167,6 @@ if (CMAKE_CXX_COMPILER_ID STREQUAL AppleClang
     add_definitions("-fdiagnostics-color=always")
 endif()
 
-if(STATIC_LIBRARY)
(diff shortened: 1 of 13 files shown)


kphillisjr:
Resubmitting with fixes to compile tests, and MSVC... It'll probably take at least 2 more revisions to fix all of the breaks.

PR #2131, opened 2016-12-01
xplat: reduce source changes between Static and Shared Libraries.

kphillisjr:
This greatly reduces the complexity of switching between static and shared libraries by converting most of the static libraries to Shared libraries. Now it is possible to test a static build of chakracore by simply linking
aginst libChakraCoreStatic. Also, this greatly reduces compile time when changing between the two since the changes between static and shared libraries has been isolated to within the bin folder.

As a quick note, This patch greatly reduces the runtime to run the Following commands...
``` sh
./build.sh -n
./build.sh -n --static
./build.sh -n --debug
./build.sh -n --static --debug
```

Also as a quick note, This has not been tested against OSX/Windows, but I do not believe this will break any builds.

v2 changes: attempt to fix compile/runtime errors with MSVC/OSX, and errors on native tests.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -167,10 +167,6 @@ if (CMAKE_CXX_COMPILER_ID STREQUAL AppleClang
     add_definitions("-fdiagnostics-color=always")
 endif()
 
-if(STATIC_LIBRARY)
-    add_definitions(-DCHAKRA_STATIC_LIBRARY=1)
-endif()
-
 if(CLR_CMAKE_PLATFORM_XPLAT)
     add_definitions(-DFEATURE_PAL)
     add_definitions(-DPLATFORM_UNIX=1)
--- a/bin/CMakeLists.txt
+++ b/bin/CMakeLists.txt
@@ -1,5 +1,5 @@
 add_subdirectory (GCStress)
+add_subdirectory (ChakraCore)
 add_subdirectory (ch)
-if(NOT STATIC_LIBRARY)
-    add_subdirectory (ChakraCore)
-endif()
+
+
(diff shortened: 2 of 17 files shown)


kphillisjr:
I fixed the issue with most of the Compile/Test errors, so all that is left is to fix the OSX build. Could someone with OSX please help me to fix these errors. The problem seems limited to linker errors.

Once OSX has a fix, I will squash all of these commits and submit this version 3.

/CC @obastemur

kphillisjr:
I think I just fixed the OSX Build by simplifying the linker flags. That means all that is required is probably some code cleanups, suggestions, and then finally squashing all of the commits on this pull request into a single pull request and commit.

kphillisjr:
@obastemur - With this Pull request, The required minimum version for cmake must change to version 3.5.0 or newer. The currently failed OSX test uses a version earlier than this, and in my tests cmake 3.4.3 or older all produce this error due to bugs in how macros, and target_sources. are handled.

@mmitche - I think that dci-macpro-10 needs updating. It appears that CMake version 3.2 includes bugs that lead to the failure of the "osx_osx_test" build parameter.

obastemur:
@kphillisjr The CI OSX issues are due to linker problems. We are unable to manage cross dependencies as we would like to. They are being optimized away. We could however manage that by some extra fake calls on final interface. However, this workaround would give us headaches later. Good intention but we would probably keep it as is.

kphillisjr:
@obastemur, I believe that the CI OSX Linker issue is caused by OSX bugs within CMake. Is there any chance to retry the test with an updated version of CMake only? I know that CMake 3.6 pre-built binaries target OS X 10.7 or newer.

kphillisjr:
Made changes, and submitting as pull request #2138.

PR #2137, opened 2016-12-02
grow_memory should return -1 on OOM

Krovatkin:


Krovatkin:
@Cellule @MikeHolman pls take a look when you have a mnt. thanks!

lib/Runtime/Language/InterpreterStackFrame.cpp
```diff
@@ -7811,10 +7811,29 @@ const byte * InterpreterStackFrame::OP_ProfiledLoopBodyStart(const byte * ip)
     int InterpreterStackFrame::OP_GrowMemory(int32 delta)
     {
 #ifdef ENABLE_WASM
-        int32 oldPageCount = m_wasmMemory->GrowInternal((uint32)delta);
+
+        int32 oldPageCount = -1;
+        JavascriptExceptionObject* caughtExceptionObject = nullptr;
+        try
+        {
+            oldPageCount = m_wasmMemory->GrowInternal((uint32)delta);
```
Cellule: The try/catch should be done in GrowInternal around `TransferInternal` right now this only covers the grow_memory operator in the interpreter, it won't work in JIT and it won't work for the JS api on the WebAssembly.Memory

Cellule:
Could you write a test that covers the exception path. It'll make it easier to verify the feature works.
You should be able to have just the WebAssembly.Memory object and grow it by the max number of pages that fits in a uint32 so 0xFFFF pages.
If this doesn't cause an out of memory, you can probably do it a few times on different WebAssembly.Memory until it throws/returns -1

Krovatkin:
i'll add a test

test/wasm/oom.js
```diff
@@ -0,0 +1,35 @@
...
+        throw `Expected ${expected}, received ${actual}`;
+    }
+}
+
+function createModulesUntilOOM (module, newSize) {
+    const n = 128; //~ 128GB which, hopefully, should be enough to counter a page file/swap
+    let arr = [];
+    for (let i = 0; i < n; i++)
+    {
+        let mod = new WebAssembly.Module(readbuffer('oom.wasm'));
```
Cellule: You don't actually need a webassembly module to test this.
Just do
```js
const mem = new WebAssembly.Memory({initial});
// loop until returns -1
  mem.grow(delta);
```

Krovatkin: good point!

Krovatkin:
@Cellule I'm actually running into a lot of issues with enabling my OOM tests 😢 

The major one is that due to the virtue of virtual memory and to the seemingly unconstrained growth of a page file on windows 64bit, OOM isn't triggered until the very last byte on a hard drive is used up. If the hard drive on a test machine has lots of free space the test hits a timeout. (https://ci.dot.net/job/Microsoft_ChakraCore/job/master/job/x64_test_prtest/397/consoleFull#170124036524fd4879-baf5-4ffe-a15a-c2a47c84523c). 

I was wondering if off the top of your head, you could advise a way to specify a maximum heap size/cap (like -Xmx256M in Java)? I couldn't find such an option in `E:\chakras\known_ops\ChakraCore-private\lib\Common\ConfigFlagsList.h` I'm going to look into GC/Recycler in a bit to see if there's maybe an env variable or hardcoded limit after i figure out all the smaller issues.

Cellule:
@MikeHolman @Krovatkin The variable `ArrayBuffer::MaxArrayBufferLength` doesn't seem to be used when growing array buffers.
I am wondering if we should explicitly check for this and throw out of memory in this case

MikeHolman:
@digitalinfinity Do you know a good way to add UT to test OOM?

@Cellule What do you mean? Looks like EntryTransfer does check for this.

Cellule:
@MikeHolman I meant in TransferInternal
However, I saw that when we'll recreate an array it check and throws, but not OutOfMemory
```cpp
ArrayBuffer::ArrayBuffer(...)
        if (length > MaxArrayBufferLength)
        {
            JavascriptError::ThrowTypeError(GetScriptContext(), JSERR_FunctionArgument_Invalid);
        }
```

Krovatkin:
@Cellule

> However, I saw that when we'll recreate an array it check and throws, but not OutOfMemory

Maybe, this is expected for typed arrays?

I think we do a subsuming check in  `WebAssemblyMemory::GrowInternal` (https://github.com/Krovatkin/ChakraCore/blob/a4e67657364883fdc6fc82d9a28f95110ad4f1d1/lib/Runtime/Library/WebAssemblyMemory.cpp#L120) 

```cpp
    const uint64 deltaBytes = (uint64)deltaPages * WebAssembly::PageSize;
    if (deltaBytes > ArrayBuffer::MaxArrayBufferLength)
    {
        return -1;
    }
```
Another interesting idiosyncrasy is that `TransferInternal` does NOT always throw an OOM error. For example, no OOM is produced even if `TransferInternal` fails to allocate a byte array on x32. Instead, it creates a new zero-length arraybuffer with a null byte array.
In contrast, in certain cases, on x64, `TransferInternal` can throw OOM BEFORE it gets a chance to run `ReportExternalMemoryFree` which in turn results in an assertion in `recycler`. 

I didn't try to refactor `TransferInternal` since since this is probably not the right PR for that

digitalinfinity:
Historically, the most reliable way I've found is adding new fault injection switches, and then running tests to simulate OOM. Otherwise, OOM is hard to do reliably (different system configurations/CI machines start to time out etc).

Krovatkin:
@digitalinfinity so i need to add stn similar to, say `ForceMemoryLeak` ?
Out of curiosity, were min/max heap size switches considered in the past? I have noticed that .NET/CoreCLR doesn't seem to have these switches either. Since both projects are much younger than Java, I am curious as to why neither one has the flags for tuning heap size?

Krovatkin:
I have adapted my tests to use injectionfault mechanism for triggering OOM errors and I have also switched to using javascript API for instantiating WASM memory objects.

Could you guys please take a look when you have a mnt?  @digitalinfinity @MikeHolman @Cellule  fyi: @arunetm

Krovatkin:
@MikeHolman fyi: @Cellule curious as to why `WebAssembly.Memory.prototype.grow` throws a `JSERR_ArgumentOutOfRange` specifically?

MikeHolman:
@Krovatkin There was no reason in particular. It just needs to throw a JS Type Error. In some of the ArrayBuffer code we throw OutOfRange for these type of cases, so I did same here.

test/wasm/oom.js
```diff
@@ -0,0 +1,40 @@
...
+
+    return 1;
+}
+
+assertEquals(2, WScript.Arguments.length);
+
+const INITIAL_SIZE = parseInt(WScript.Arguments[0]);
+const GROW_SIZE = parseInt(WScript.Arguments[1]);
+
+assertEquals(0, wasmAlloc(INITIAL_SIZE, GROW_SIZE));
```
Cellule: I don't understand the test. You are checking that the function returns 0, but it returns 0 when an exception happens.
We are not supposed to throw when running `.grow`.
Plus you are not checking that we correctly return `-1` when we get an out of memory exception (in c++)

Krovatkin: @Cellule 
Spec seems to suggest that `Memory.prototype.grow` is supposed to throw in case of OOM

> Let ret be the result of performing a grow_memory operation given delta d.
> If ret is -1, a RangeError is thrown.

https://github.com/WebAssembly/design/blob/master/JS.md#webassemblymemoryprototypegrow

I could add a test that calls `grow_memory` directly. However, it looks like testing `Memory.prototype.grow`  is sufficient since the only two other possible exceptions `Memory.prototype.grow` could throw are `WASMERR_NeedMemoryObject` and `JSERR_DetachedTypedArray` are unlikely to be thrown in this particular scenario. I could add an extra test to check the error's message to make sure the correct error (`JSERR_ArgumentOutOfRange` ) is thrown if you think this is necessary. 

Could you please elaborate on this one? 

> Plus you are not checking that we correctly return -1 when we get an out of memory exception (in c++)

Krovatkin: @Cellule I think you are right. I need to make sure that the error thrown is indeed `JSERR_ArgumentOutOfRange ` rather than OOM or something else totally unexpected.

Cellule: I think I got confused between the js and the web assembly implementations.
I guess we should test both then since they don't behave the same, wasm grow_memory operator should always return -1 and not throw, the Js WebAssembly.Memory.grow should throw a RangeError if it would return -1

Krovatkin:
@Cellule Refined the original tests to make sure that an exception is indeed `JSERR_ArgumentOutOfRange`. Also added 3 more tests that directly call `grow_memory`

MikeHolman:
@leirocks Can you take a look at fault injection change? I think we shouldn't have specific UTs with fault injection flags. We have infra which runs and causes fault at each legal location.

lib/Runtime/Library/WebAssemblyMemory.cpp
```diff
@@ -139,9 +139,34 @@ WebAssemblyMemory::GrowInternal(uint32 deltaPages)
...
-    m_buffer = newBuffer;
+    ArrayBuffer * newBuffer = nullptr;
+    JavascriptExceptionObject* caughtExceptionObject = nullptr;
+    try
+    {
+        newBuffer = m_buffer->TransferInternal(newBytes);
+    }
+    catch (const JavascriptException& err)
+    {
+        caughtExceptionObject = err.GetAndClear();
```
Cellule: I think here we could Assert that this is an OOM exception, but in every cases we should just return -1 and drop the exception

lib/Runtime/Library/WebAssemblyMemory.cpp
```diff
@@ -139,9 +139,34 @@ WebAssemblyMemory::GrowInternal(uint32 deltaPages)
...
+        Assert(caughtExceptionObject);
+        //Propagate if not OOM
+        if (caughtExceptionObject != ThreadContext::GetContextForCurrentThread()->GetPendingOOMErrorObject())
+        {
+            caughtExceptionObject = caughtExceptionObject->CloneIfStaticExceptionObject(GetScriptContext());
+            JavascriptExceptionOperators::DoThrow(caughtExceptionObject, GetScriptContext());
+        }
+        GetRecycler()->ReportExternalMemoryAllocation(oldBytes); //transferInternal overestimates the amount of freed memory
```
Cellule: I don't understand the case where this can happen.
If this really happens then it's a bug in TransferInternal, not in GrowMemory

lib/Runtime/Library/WebAssemblyMemory.cpp
```diff
@@ -139,9 +139,34 @@ WebAssemblyMemory::GrowInternal(uint32 deltaPages)
...
+        return -1;
+    }
 
+    Assert(newBuffer);
+    if (!newBuffer->GetByteLength() && m_buffer->GetByteLength() < newBytes)
+    {
+        GetRecycler()->ReportExternalMemoryAllocation(oldBytes); //transferInternal overestimates the amount of freed memory
```
Cellule: I can't see this case when looking at TransferInternal, it looks like it throws on every failure cases or at least it should

MikeHolman:
@Krovatkin I talked with @leirocks offline, and seems that there are dependencies that need to be updated in the internal Chakra repo. Can you remove the fault injection changes and assign issue to me for adding this? I can take care of porting that over.

lib/Runtime/Library/WebAssemblyMemory.cpp
Krovatkin: The createArrayBuffer path doesn't throw on x32.

lib/Runtime/Library/WebAssemblyMemory.cpp
Krovatkin: Looks like this happens every time an existing buffer is "resized". 
`transferInternal` calls `recycler->ReportExternalMemoryAllocation(newBufferLength - this->bufferLength)` and then `ReportExternalMemoryFailure(newBufferLength)` in case of OOM.

If OOM exception is thrown, the old buffer is left unmodified and its memory (`this->bufferLength`) also gets freed at shutdown. 

https://github.com/Microsoft/ChakraCore/pull/2137#issuecomment-267388954 provides a testcase for exposing this issue.

Krovatkin:
@MikeHolman @Cellule fyi: @arunetm 

Removed injection faults, fixed the cases where RangeErrors or OOMs are expected but not thrown, refined memory tracking in `transferInternal` to prevent assertions (overestimation of freed memory).  Also, confirmed that `ArrayBuffer.prototype.slice` produces an exception if OOM.

```javascript
const uint8 = new Uint8Array(0x3FFF0000);
const uint8_2 = uint8.slice(1, 0x3FFF0000); 
```

Cellule:
@Krovatkin I am so sorry, with the holidays I totally forgot about this PR.
I've reviewed and it looks good. I will take take of merging this. Thanks!

Krovatkin:
@Cellule that's totally fine! i'm very sorry that it took me so long (and 33 comments) to realize and refine the scope of this PR.  Do you want me to close this one and open another against master?

Cellule:
Yeah, I think it would be safer at this point to merge this in master, since it's not really fixing a bug, but making it more spec compliant.
Thanks

Krovatkin:
Needs to be rebased against master. Opened https://github.com/Microsoft/ChakraCore/pull/2393

PR #2144, opened 2016-12-05
swb: AddressOf, FieldWithBarrier, build Full

jianchun:
swb: WriteBarrierPtr: swap "&" and AddressOf
---------------------------------------------------
Overriding "&" operator keeps running into new trouble. This change
removes "&" operator override and switches "AddressOf" to do that.

swb: add FieldWithBarrier macro
-----------------------------------
And remove FORCE_USE_WRITE_BARRIER based macro redefinitions. We've run
into confusing effects of macro redefinition several times.

Included one local inline slot annotation point fix.

swb: RecyclerJITTypeHolder rename
---------------------------------------
Rename JITTypeHolderObject to RecyclerJITTypeHolder (previous CR feedback)

swb: misc fixes to build chakra full
--------------------------------------

### DIFF
--- a/lib/Runtime/Base/FunctionBody.h
+++ b/lib/Runtime/Base/FunctionBody.h
@@ -1259,7 +1259,7 @@ namespace Js
         template <class DebugSite, class Fn>
         HRESULT MapEntryPoints(DebugSite site, Fn fn) const // external debugging version
         {
-            return Map(site, this->entryPoints, [&](int index, LoopEntryPointInfo * entryPoint)
+            return Map(site, PointerValue(this->entryPoints), [&](int index, LoopEntryPointInfo * entryPoint)
             {
                 if (entryPoint != nullptr)
                 {
--- a/lib/Runtime/Base/FunctionBody.h
+++ b/lib/Runtime/Base/FunctionBody.h
@@ -1285,26 +1285,16 @@ namespace Js
         }
 
     };
-}
 
-// ----------------------------------------------------------------------------
-// SWB-TODO: Below force uses write barrier. Move top code to new header file.
-//
+    // ----------------------------------------------------------------------------
+    // SWB-TODO: Below explicitly uses write barrier. Move top code to new header file.
+    //
 
-// In general, write-barriers are used only on non-Windows platforms
-// However, classes in this file have been allocated in software write-barrier
-// memory for historic reasons, because there is some small perf wins here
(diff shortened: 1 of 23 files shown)


jianchun:

lib/Runtime/Base/FunctionBody.h
```diff
@@ -1765,60 +1755,60 @@ namespace Js
...
-        Field(bool) m_isWasmFunction : 1;
-        Field(bool) m_isGlobalFunc : 1;
-        Field(bool) m_doBackendArgumentsOptimization : 1;
-        Field(bool) m_doScopeObjectCreation : 1;
-        Field(bool) m_usesArgumentsObject : 1;
-        Field(bool) m_isEval : 1;              // Source code is in 'eval'
-        Field(bool) m_isDynamicFunction : 1;   // Source code is in 'Function'
-        Field(bool) m_hasImplicitArgIns : 1;
-        Field(bool) m_dontInline : 1;            // Used by the JIT's inliner
+        FieldWithBarrier(bool) m_hasBeenParsed : 1;       // Has function body been parsed- true for actual function bodies, false for deferparse
```
leirocks: ah, FieldWithBarrier will also use no barrier policy right? but still, I think it should be explicit FieldNoBarrier for non poiner field

---
In reply to: [90974531](https://github.com/Microsoft/ChakraCore/pull/2144#discussion_r90974531) [](ancestors = 90974531)

leirocks:
:shipit:

lib/Runtime/Base/FunctionBody.h
jianchun: Yes, whatever we put here are effectively the same. There are pros and cons either way. Let's think about this.

leirocks: what's the cons for the primitive types? since even some pointers we should mark it as no barrier.

---
In reply to: [90976930](https://github.com/Microsoft/ChakraCore/pull/2144#discussion_r90976930) [](ancestors = 90976930)

leirocks: also I think we should make FieldWithBarrier to be force write barrier, regardless of the type, to be able to use write barrier on non-pointer field like intptr_t, there's no such use case for now though.

---
In reply to: [90977538](https://github.com/Microsoft/ChakraCore/pull/2144#discussion_r90977538) [](ancestors = 90977538,90976930)

PR #2146, opened 2016-12-06
WASM - f64.mul/div interpreter precision on x86

### LINKED ISSUE #2145: WASM - f64.mul/div interpreter precision on x86
The compiler we use in ChakraFull doesn't generate the same thing for mul/div with doubles as it does for ch.exe
The instruction set generated is 
```asm
fld aLeft
fdiv/fmul aRight
```

The instructions we should use are
```asm
movsd xmm0, aLeft
divsd/mulsd xmm0, aRight
```

After fixing this, need to re-enable tests
test\WasmSpec\testsuite\float_misc.wast:313-317,404-409

Cellule:
Wasm: Manually implement div/mul for double to have the same precision in interpreter and jitted code.

Fixes #2145

Revert "Disable tests in float_misc.wast until fix has been found"

This reverts commit 1e77788b7d40cf77f9a80baffc535e498ba2c081.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/2146)


Cellule:
@MikeHolman @LouisLaf Can you guys reviews this please ?

lib/Runtime/Math/AsmJsMath.inl
```diff
@@ -108,12 +108,36 @@ namespace Js
     {
         return aLeft / aRight;
     }
+#if _M_IX86
+    template<>
+    inline double AsmJsMath::Div( double aLeft, double aRight )
+    {
+        __asm {
+            movsd xmm0, aLeft;
```
ThomsonTan: >xmm0 [](start = 18, length = 4)

Is xmm0 free to use at anytime?

MikeHolman:
IMO better is to use SSE2 library like we have for JavascriptMath.

Cellule:
@MikeHolman I'm not sure I understand what exactly makes JavascriptSSE2MathOperators different from JavascriptMathOperators other than the `#define SSE2MATH`
Are you saying there is a way to tell the compiler to use SSE2 math on a file by file basis ?

LouisLaf:
Yes, the JavascriptSSE2MathOperators file is compiled with -arch:SSE2 which allows the compiler to use SSE2 instructions (note however that it is not guaranteed).  You can also use the new __vectorcall calling convention to pass arguments in SSE2 registers, which will make it way more likely for the compiler to use SSE2 instrs for the computations.  Note that __vectorcall is theoratically not legal at the call-site which is probably not compiled with -arch:SSE2....  We don't support however non-SSE2 machines anymore, so it's ok to break that path.

---
In reply to: [265549927](https://github.com/Microsoft/ChakraCore/pull/2146#issuecomment-265549927) [](ancestors = 265549927)

Cellule:
I think I found the real source of the problem
AsmJsMath.inl resides in Chakra.Runtime.Math which sets the /arch:sse2 flag.
However, since the implementations are in the .inl file that is included in JavascriptMathOperator which resides in Chakra.Runtime.Language, it is compiled without the /arch:sse2 flag.
I'll try to move the includes around see if I can make asmjsmath to be compiled with /arch:sse2

Cellule:
So the only way to make sure that code is compiled with /arch:sse2 it had to be moved to the cpp file.
I moved the implementation of div & mul there and did my best to keep it clean I also took the opportunity to clean up some includes and an outdated div implementation in WasmMath

Cellule:
ping

LouisLaf:
:shipit:

PR #2148, opened 2016-12-06
xplat: implement internal qsort_r

obastemur:
Tested with an array of 100M ordered/random numbers.
Comparing to STL qsort_r ~10% to ~30% ~~~40%~~ slower.

### DIFF
--- a/pal/src/cruntime/misc.cpp
+++ b/pal/src/cruntime/misc.cpp
@@ -251,14 +251,6 @@ PAL_qsort(void *base, size_t nmemb, size_t size,
     PERF_EXIT(qsort);
 }
 
-//
-// qsort_s is a variant of qsort that takes in a context parameter
-// It's similar to qsort_r that GCC provides, but with the difference
-// that qsort_s takes in the context as the first param to the compare
-// function, while qsort_r takes it as the last.
-// So we have to do this wrapper dance to deal with the translation
-// Lambda doesn't work here as qsort_r needs a function pointer
-//
 typedef int (__cdecl * qsort_s_comparer)(void*, const void *, const void *);
 struct CompareWrapperContext
 {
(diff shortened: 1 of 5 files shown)


obastemur:
Also enabled a previously disabled test case due to implementation inconsistencies.

pal/src/cruntime/misc.cpp
```diff
@@ -251,57 +251,241 @@ PAL_qsort(void *base, size_t nmemb, size_t size,
...
+    cc_qsort_r(base, pos++, size, compar, thunk);
+    cc_qsort_r(base + (pos * size), nmemb - pos, size, compar, thunk);
 }
-#endif
 
 PALIMPORT
 void __cdecl
 PAL_qsort_s(void *base, size_t nmemb, size_t size,
-    int (__cdecl *compar )(void*, const void *, const void *), void* context)
+    int (__cdecl *compar )(void*, const void *, const void *), void* context) noexcept
```
jianchun: >noexcept [](start = 78, length = 8)

See several "noexcept" like this. Is it safe? "compar" callback may throw?

obastemur: Sure it is safe. We don't catch anything here and we don't want to catch `here`

jianchun: I'm not familiar with this. Reading http://en.cppreference.com/w/cpp/language/noexcept_spec, "If a search for a matching exception handler leaves a function marked noexcept or noexcept(true), std::terminate is called immediately. ". If ```compar``` throws from within, it will terminate instead of searching for handlers in callers. Won't it?

obastemur: > std::terminate is called immediately

The `immediately` part is IMHO open to misinterpretation. Otherwise, that would be similarly expensive. Because, there has to be an additional code to track that, right? However AFAIK compiler doesn't put anything there (when noexcept is set) and exception slips to a caller. There, we have the exception handler in place.

Agree this is a trick and saves ~4%

obastemur: BTW; thanks for the point. I will add some additional testing for throw/catch to make sure it doesn't break later.

pal/src/cruntime/misc.cpp
```diff
@@ -251,57 +251,241 @@ PAL_qsort(void *base, size_t nmemb, size_t size,
...
 }
-#else
-// the qsort_r comparer is backwards from the way it's defined on linux
-static int qsort_compare_wrapper(void* context, const void* e1, const void* e2)
+
+union swap8
 {
-    CompareWrapperContext* wrapper = (CompareWrapperContext*) context;
-    return wrapper->real_comparer(wrapper->real_context, e1, e2);
+    char chr[8]; // optimized for Js::Var
```
ThomsonTan: >optimized for Js::Var [](start = 20, length = 21)

Is it only optimized for Js::Var on 64-bit platform?

test/Array/array_qsortr.js
```diff
@@ -0,0 +1,73 @@
+//-------------------------------------------------------------------------------------------------------
+// Copyright (C) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE.txt file in the project root for full license information.
+//-------------------------------------------------------------------------------------------------------
+
+// Test our custom qsort_r implementation on xplat.
```
ThomsonTan: >Test our custom qsort_r implementation on xplat [](start = 3, length = 47)

Would this implementation benefit win32?

pal/src/cruntime/misc.cpp
obastemur: swap4 and swap2 should take care of 32-bit?

obastemur:
Also added `BUILD_TYPE` property to `WScript.Platform`. Testing 500.000 array several times sometimes timeouts with CI Debug builds.

pal/src/cruntime/misc.cpp
obastemur: Removed `fexceptions` from PAL instead (on a separate PR). Also added some extra tests for throwing from compare.

test/Array/array_qsortr.js
obastemur: Good question. If there is a reason for enabling these tests on Windows, I can do that. (we particularly test custom`qsortr` here)

test/Array/array_qsortr.js
```diff
@@ -0,0 +1,113 @@
...
+            return x > y ? +1 : (x < y ? -1 : 0) ;
+        },
+        arraySize /* test 500K items */,
+        "NoThrow");
+
+    testArray(
+        function(x, y) {
+            try {
+                if (x % 2) throw new Error("Crash!!!! ?");
+            } catch (e) { /* no. this shouldn't crash! */ }
```
jianchun: nit: In the "catch(e) { ... }" blocks, you can try verify the exception caught is indeed the exception you throw.

jianchun:
LGTM

obastemur:

PR #2149, opened 2016-12-06
Undeferral fix

pleath:
With the refactor of FunctionInfo/FunctionBody and the undeferral path, we could leave the FunctionInfo pointing to an incomplete FunctionBody if we hit an exception before compilation was finished. In such a case, make sure we restore the FunctionInfo to its previous state.


pleath:
@rajatd This is the fix for the compilecount==0, bytecodeblock==null cases.

rajatd:
:shipit:

PR #2150, opened 2016-12-06
Fix opeq assert firing in x86

meg-gupta:
Debug asserts that fire when operand sizes differ for mov and opeq
instructions were added recently. Fix a missing change that fires this
assert.


meg-gupta:

rajatd:
:shipit:

PR #2151, opened 2016-12-07
Recheck array type before checking element is MissingItem after prototype lookup on the array

ThomsonTan:
When slicing an array (Array.prototype.slice),  we make a shadow copy of original array, then tries to populate the MissingItems in the copied array with elements from the prototype chain of the source array. Lookup in the prototype chain could change its array type so we need recheck the array type after the lookup returns before retrieving remaining items as MissItems, or we get the wrong element to check against.

### DIFF
--- a/lib/Runtime/Library/JavascriptArray.cpp
+++ b/lib/Runtime/Library/JavascriptArray.cpp
@@ -475,7 +475,7 @@ namespace Js
         return SparseArraySegment<T>::IsMissingItem(&headSeg->elements[index]);
     }
 
-    bool JavascriptArray::IsMissingItemForUnkownArrayTypeAt(uint32 index)
+    bool JavascriptArray::IsMissingItem(uint32 index)
     {
         bool isIntArray = false, isFloatArray = false;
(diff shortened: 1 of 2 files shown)


ThomsonTan:
@LouisLaf please help review.

lib/Runtime/Library/JavascriptArray.cpp
```diff
@@ -467,6 +467,33 @@ namespace Js
...
 
+    template<typename T>
+    bool JavascriptArray::IsMissingItemAt(uint32 index) const
+    {
+        SparseArraySegment<T>* headSeg = (SparseArraySegment<T>*)this->head;
+
+        return SparseArraySegment<T>::IsMissingItem(&headSeg->elements[index]);
+    }
+
+    bool JavascriptArray::IsMissingItemForUnkownArrayTypeAt(uint32 index)
```
LouisLaf: This is somewhat inconsistent naming with the rest of the array APIs.  Just call that one IsMissingItemAt() and leave the template one as is, or rename to something like DirectIsMissingItemAt().

LouisLaf:
:shipit:

PR #2152, opened 2016-12-07
CheckIfTypeIsEquivalent for fixed field

kunalspathak:
With my CheckIfTypeIsEquivalent optimization, we were reusing the cached equivalent types even though guard value i.e. type got collected. This is fine as long as fields against which we are guarding are not fixed. If they are fixed, then we should not be referring to the cached types. Added a new helper for fixed fields/methods which will return false if the guard was invalidated or it was collected.

Refer: [9514714](https://microsoft.visualstudio.com//defaultcollection/OS/_workItems?_a=edit&id=9514714&triage=true)


kunalspathak:
@pleath , @rajatd - Please help review.

pleath:
LGTM. Thanks.

PR #2157, opened 2016-12-07
Use Sections for OOP JIT

MikeHolman:
This change adds new SectionAllocWrapper classes, which manages Sections for JIT code pages.

When you call Alloc with these APIs, sections are mapped into the runtime process with PAGE_EXECUTE permissions. For SectionAllocWrapper, we maintain a map of outstanding sections modeled after the GC's HeapBlockMap for fast lookup.
If you want to map a section into the JIT process, you can call AllocLocal, which will map (a portion) of your section in as PAGE_READWRITE. It is the callers responsibility to call FreeLocal on the local section mapping.

Sections don't support decommit, which is used most importantly in PreReservedSectionAllocWrapper. Instead, to remove pages from the working set we ZeroMemory and set PAGE_NOACCESS protection on the runtime view.

InterpreterThunkEmitter was also modified to use sections. On the server we map everything in and then throw away the section handle. When the runtime is done, it can directly call to UnmapViewOfSection.
Now for OOP JIT we use the thunkBlocks lists for tracking outstanding interpreter thunks rather than maintaining an EmitBufferManager on the JIT process.


MikeHolman:
@satheeshravi @LouisLaf @leirocks can you take a look?

kphillisjr:
You need to specify new allocator types in ```lib/Common/Memory/PageAllocatorDefines.h``` seeing as how this file contains the Constants for ```lib/Common/Memory/PageAllocator.h```

MikeHolman:
@kphillisjr I was thinking reusing `PageAllocatorType_CustomHeap` should be fine?

kphillisjr:
I believe the reason for the Allocator types is to help with finding/diagnosing memory problems.  This is why I believe this type of change should include a custom entry because others might use it elsewhere... That said, I believe that ```PageAllocatorType_SectionAlloc``` is reasonable enough.

lib/Common/Core/DelayLoadLibrary.h
```diff
@@ -45,20 +72,59 @@ class NtdllLibrary : protected DelayLoadLibrary
 
     typedef VOID (NTAPI *PFnRtlGrowFunctionTable)(_Inout_ PVOID DynamicTable, _In_ ULONG NewEntryCount);
     PFnRtlGrowFunctionTable growFunctionTable;
+#endif
+
+    typedef NTSTATUS(NTAPI *PFnNtCreateSection)(
```
leirocks: is the section APIs blocked by ACG on client process?

MikeHolman: Yes, I believe so.

lib/Common/Memory/SectionAllocWrapper.cpp
```diff
@@ -0,0 +1,949 @@
...
+        status = NtdllLibrary::Instance->MapViewOfSection(sectionHandle, this->process, &address, NULL, NULL, NULL, &viewSize, NtdllLibrary::ViewUnmap, NULL, PAGE_EXECUTE);
+        if (status != 0 || address == nullptr)
+        {
+            NtdllLibrary::Instance->Close(sectionHandle);
+            return nullptr;
+        }
+        SectionInfo * section = HeapNewNoThrowStruct(SectionInfo);
+        if (!section)
+        {
+            HeapDelete(section);
```
leirocks: >HeapDelete(section); [](start = 12, length = 20)

when OOM section is null, should not delete

lib/Common/Memory/SectionAllocWrapper.cpp
```diff
@@ -0,0 +1,949 @@
...
+        NtdllLibrary::Instance->InitializeObjectAttributes(&attr, NULL, NtdllLibrary::OBJ_KERNEL_HANDLE, NULL, NULL);
+        LARGE_INTEGER size = { 0 };
+#if TARGET_32
+        size.LowPart = dwSize;
+#elif TARGET_64
+        size.QuadPart = dwSize;
+#endif
+        HANDLE sectionHandle = nullptr;
+        const ULONG allocAttributes = ((allocationType & MEM_COMMIT) == MEM_COMMIT) ? SEC_COMMIT : SEC_RESERVE;
+        int status = NtdllLibrary::Instance->CreateSection(&sectionHandle, SECTION_MAP_READ | SECTION_MAP_WRITE | SECTION_QUERY | SECTION_MAP_EXECUTE, &attr, &size, PAGE_EXECUTE_READWRITE, allocAttributes, NULL);
```
leirocks: >PAGE_EXECUTE_READWRITE [](start = 165, length = 22)

will this make it executable on JIT server, or is this necessary to be executable here since we specify PAGE_EXECUTE while MapViewOfSection?

lib/Common/Memory/SectionAllocWrapper.cpp
```diff
@@ -0,0 +1,949 @@
...
+        if (status != 0)
+        {
+            return nullptr;
+        }
+
+        SIZE_T viewSize = 0;
+        status = NtdllLibrary::Instance->MapViewOfSection(sectionHandle, this->process, &address, NULL, NULL, NULL, &viewSize, NtdllLibrary::ViewUnmap, NULL, PAGE_EXECUTE);
+        if (status != 0 || address == nullptr)
```
leirocks: >0 [](start = 22, length = 1)

nit: use STATUS_SUCESS macro

lib/Common/Memory/SectionAllocWrapper.cpp
MikeHolman: Yeah, this is necessary in order to map a view that has PAGE_EXECUTE permissions. It doesn't actually map into any process' address space at this point.

leirocks:
:shipit:

lib/Common/Memory/SectionAllocWrapper.cpp
MikeHolman: we don't have header with that macro defined.

PR #2158, opened 2016-12-07
Redeferral and deferred stubs

pleath:
Do not copy the deferred stub tree to the new ParseableFunctionInfo on redeferral, as nested functions may not be deferred in that case. Recompute them instead if necessary.


pleath:
@rajatd Could you look at this very small fix? Should fix some web issues. (Thanks.)

rajatd:
:shipit:

PR #2165, opened 2016-12-07
WASM - function import signature comparison

Cellule:
When comparing the function signature on imports, the counter represents the function index, not the signature index

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/2165)


lib/Runtime/Library/WebAssemblyInstance.cpp
```diff
@@ -322,7 +322,7 @@ void WebAssemblyInstance::LoadImports(
             {
                 Assert(env->GetWasmFunction(counter) == nullptr);
                 AsmJsScriptFunction* func = AsmJsScriptFunction::FromVar(prop);
-                if (!wasmModule->GetSignature(counter)->IsEquivalent(func->GetSignature()))
+                if (!wasmModule->GetFunctionSignature(counter)->IsEquivalent(func->GetSignature()))
```
Cellule: I agree, it was useful before we had wasm function thunks for imported functions

PR #2166, opened 2016-12-08
WASM - Fix bug in -wasmi64 implementation

Cellule:
The case `someInt64Fn({high: 5, low: -2})` would incorrectly think the input is `-2` as the current implementation was doing signed extended conversion from int32 to int64 when merging low and high

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/2166)


Cellule:

PR #2168, opened 2016-12-08
relax the rpc failfast

MikeHolman:
This failfast was a bit aggressive, and firing in scenarios where we don't necessarily want to get dumps (e.g. a user kills the server process as part of their shutdown). This excludes from firing a failfast for the known scenarios. Instead we will throw a codegen aborted error if this happened during codegen, or OOM in other cases.

### DIFF
--- a/lib/Backend/NativeCodeGenerator.cpp
+++ b/lib/Backend/NativeCodeGenerator.cpp
@@ -904,7 +904,7 @@ NativeCodeGenerator::CodeGen(PageAllocator * pageAllocator, CodeGenWorkItem* wor
             // script context may close after codegen call starts, consider this as aborted codegen
             hr = E_ABORT;
         }
-        JITManager::HandleServerCallResult(hr, true);
+        JITManager::HandleServerCallResult(hr, RemoteCallType::CodeGen);
     }
     else
     {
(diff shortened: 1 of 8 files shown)


lib/Backend/NativeCodeGenerator.cpp
```diff
@@ -3639,6 +3640,24 @@ JITManager::HandleServerCallResult(HRESULT hr)
     case VBSERR_OutOfStack:
         throw Js::StackOverflowException();
     default:
-        RpcFailure_fatal_error(hr);
+        if (!GetJITManager()->IsServerAlive() && (
+            hr == HRESULT_FROM_WIN32(RPC_S_CALL_FAILED) ||
+            hr == HRESULT_FROM_WIN32(RPC_S_CALL_FAILED_DNE)
+            ))
+        {
+            if (isCodeGenCall)
```
sermetiskin: is there no way to gracefully fallback to interpreted?

lib/JITClient/JITManager.h
```diff
@@ -208,7 +220,7 @@ class JITManager
 
     static JITManager * GetJITManager()
         { return &s_jitManager; }
-    static void HandleServerCallResult(HRESULT hr);
+    static void HandleServerCallResult(HRESULT hr, RemoteCallType callType) { Assert(UNREACHED); }
```
ThomsonTan: >RemoteCallType callType [](start = 51, length = 23)

Declare a default value for callType for typical or unknown scenario?

PR #2169, opened 2016-12-08
WASM - call_indirect no table error

Cellule:
call_indirect operator without a table is a compile error 
Add WasmLazyTrapCallback as a valid function entry point

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/2169)


Cellule:

PR #2172, opened 2016-12-08
2 redeferral-related fixes in Recycler

pleath:
1. Do not try to take special mark actions in BackgroundScanStack, only in ScanStack, as memory may be unstable in the background case. 2. In marking, if MarkInternal returns true (indicating no further work is necessary), and if we're required to take special mark actions, detect finalizable object and take the action.

### DIFF
--- a/lib/Common/Memory/HeapBlockMap.inl
+++ b/lib/Common/Memory/HeapBlockMap.inl
@@ -68,6 +68,10 @@ HeapBlockMap32::Mark(void * candidate, MarkContext * markContext)
 
     if (MarkInternal<interlocked>(chunk, candidate))
     {
+        if (doSpecialMark)
+        {
+            this->OnSpecialMark(chunk, candidate);
+        }
         return;
     }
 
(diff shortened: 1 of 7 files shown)


pleath:
@dotnet-bot test Ubuntu ubuntu_linux_debug when you've got a chance

pleath:
@curtisman I know you're watching
@digitalinfinity In case Curtis isn't watching

lib/Common/Memory/HeapBlockMap.inl
```diff
@@ -66,16 +66,61 @@ HeapBlockMap32::Mark(void * candidate, MarkContext * markContext)
...
-    }
-
     uint id2 = GetLevel2Id(candidate);
     HeapBlock::HeapBlockType blockType = chunk->blockInfo[id2].blockType;
 
     Assert(blockType == HeapBlock::HeapBlockType::FreeBlockType || chunk->map[id2]->GetHeapBlockType() == blockType);
 
+    if (MarkInternal<interlocked>(chunk, candidate))
+    {
+        if (doSpecialMark)
```
curtisman: May be we can put this code in a separate function?

curtisman:
Other then the refactor comment.  LGTM.

PR #2173, opened 2016-12-08
WASM - Added checks for exports

Cellule:
We were not behaving as spec for some aspects in the exports table.
We need to check for duplicate names and verify that the global exists.

Disallow export/import of mutable globals

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/2173)

### DIFF
--- a/lib/WasmReader/WasmBinaryReader.cpp
+++ b/lib/WasmReader/WasmBinaryReader.cpp
@@ -778,6 +778,10 @@ void WasmBinaryReader::ReadExportTable()
             {
                 ThrowDecodingError(_u("Unknown global %u for export %s"), index, exportName);
             }
+            if (m_module->GetGlobal(index)->IsMutable())
+            {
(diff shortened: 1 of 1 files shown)


Cellule:

lib/WasmReader/WasmBinaryReader.cpp
```diff
@@ -963,6 +1001,10 @@ WasmBinaryReader::ReadImportEntries()
         {
             WasmTypes::WasmType type = ReadWasmType(len);
             bool isMutable = ReadConst<UINT8>() == 1;
+            if (isMutable)
```
MikeHolman: Different issue, but noticed it now. We should also check for other values and throw decoding error (i.e. if it is anything other than 0 or 1).

Cellule: Thanks! Created #2193 to track this

PR #2174, opened 2016-12-08
WASM - elem/data segment validation

Cellule:
Check if a table is present when reading an element section
Validate the init_expr for element and data segments

Disallow internal global in init_expr
The issue https://github.com/WebAssembly/spec/issues/367 made me realize we were doing the same thing the spec interpreter was which was not according to the design documents

Check the type of init_expr for globals

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/2174)


Cellule:
Hold off a moment to review this, I want to include another fix related to this

Cellule:
Should be okay to review now

lib/Runtime/Library/WebAssemblyModule.cpp
```diff
@@ -553,29 +553,37 @@ WebAssemblyModule::AddTableImport(const char16* modName, uint32 modNameLen, cons
...
-        Wasm::WasmGlobal* global = m_globals->Item(initExpr.var.num);
+        Wasm::WasmGlobal* global = GetGlobal(initExpr.var.num);
+        Assert(global->GetType() == Wasm::WasmTypes::I32);
+        offset = env->GetGlobalValue(global).i32;
+    }
+    return offset;
+}
+
+void
+WebAssemblyModule::ValidateInitExporForOffset(const Wasm::WasmNode& initExpr) const
```
MikeHolman: nit: spelling (ValidateInitExprForOffset)

lib/Runtime/Library/WebAssemblyModule.cpp
```diff
@@ -553,29 +553,37 @@ WebAssemblyModule::AddTableImport(const char16* modName, uint32 modNameLen, cons
 uint
 WebAssemblyModule::GetOffsetFromInit(const Wasm::WasmNode& initExpr, const WebAssemblyEnvironment* env) const
 {
-    if (initExpr.op != Wasm::wbI32Const && initExpr.op != Wasm::wbGetGlobal)
-    {
-        throw Wasm::WasmCompilationException(_u("Invalid init_expr for element offset"));
-    }
+    // Should have been checked at compile time, but let's just make sure one more time
+    ValidateInitExporForOffset(initExpr);
```
MikeHolman: then maybe we should put under #if DBG and assert?

lib/WasmReader/WasmBinaryReader.cpp
```diff
@@ -804,12 +808,12 @@ WasmBinaryReader::ReadElementSection()
     for (uint32 i = 0; i < count; ++i)
     {
         uint32 index = LEB128(length); // Table id
-        if (index != 0)
+        if (index != 0 || !(m_module->HasTable() || m_module->HasTableImport()))
```
MikeHolman: the spec seemed unclear to me about this. it looked like it might be legal to have no table or table import and have element section, as long as it does nothing (e.g. any segments have length 0)

MikeHolman: true, but the design says that every module has one default memory/table (not up to one). there is no mention what the state is if there is no explicit table section. so if this is how the spec interpreter behaves then i'm fine deferring to it. just one of those corners that will needed to be worked out in the formal spec

Cellule: It feels weird to allow to set data or elements in a memory/table that hasn't been defined.
In our current implementation, we would throw a linking error when treating the data/element segments.
This is only moving the error to compile time effectively changing from WebAssembly.RuntimeError to WebAssembly.CompileError which is more in line with the spec interpreter.
In the end, it really doesn't matter, but early error should be favorable if we have to read the data anyway

Cellule:
ping

lib/Runtime/Library/WebAssemblyModule.cpp
```diff
@@ -553,29 +553,46 @@ WebAssemblyModule::AddTableImport(const char16* modName, uint32 modNameLen, cons
 uint
 WebAssemblyModule::GetOffsetFromInit(const Wasm::WasmNode& initExpr, const WebAssemblyEnvironment* env) const
 {
-    if (initExpr.op != Wasm::wbI32Const && initExpr.op != Wasm::wbGetGlobal)
+    try
+    {
+        ValidateInitExportForOffset(initExpr);
```
MikeHolman: A bit strange, but seems ok to me.

PR #2175, opened 2016-12-08
Fix for...in enumeration in prototype chain for shadowed property

ThomsonTan:
Current we ignore to visit a prototype in for...in enumerator if the prototype has hasNoEnumerableProperties as true to accelerate prototype lookup, like for many built-in prototype objects, but ignore visit the non-enumerable properties could miss process shadowed property in the base prototype.


ThomsonTan:
@curtisman @rajatd could you please review?

lib/Runtime/Library/ForInObjectEnumerator.cpp
```diff
@@ -182,7 +203,30 @@ namespace Js
                 if (TestAndSetEnumerated(propertyId) //checks if the property is already enumerated or not
                     && (attributes & PropertyEnumerable))
                 {
-                    return currentIndex;
+                    bool propertyShadowed = false;
+
+                    if (this->enumeratingPrototype)
```
rajatd: add some more comments around this code

lib/Runtime/Library/ForInObjectEnumerator.cpp
```diff
@@ -182,7 +207,32 @@ namespace Js
                 if (TestAndSetEnumerated(propertyId) //checks if the property is already enumerated or not
                     && (attributes & PropertyEnumerable))
                 {
-                    return currentIndex;
+                    bool propertyShadowed = false;
```
curtisman: Umm... Is it worth it to of paying the price of looking up the property on all the intervening prototype chain object on every enumerable property, instead of just enumerating all the property in the chain without skipping to the first object with enumerable properties in the property chain so that the bit vector check is enough?

ThomsonTan: Usually almost all the properties of built-in prototype are non-enumerable properties, and skip them for look-up is a good acceleration, like we could ignore visiting ~20 methods on Array.prototype which save us time. Array.prototype and Object.prototype share no  properties so bitmap doesn't help this much.  Also lookup in descendant prototype is guarded by bitmap test, so there would be minimum extra look-up introduced.
I will run benchmark to see whether there is some significant perf impact.

rajatd:
:shipit:

lib/Runtime/Library/ForInObjectEnumerator.cpp
curtisman: But this will slow down the case where you have a plain prototype chain significantly since every enumerable prototype property will have to ask the whole chain if it is shadowed again.   That defeat the purpose of having the bit vector (where now it only have benefit of property that has been shadowed before, and not property that hasn't shadowed.

ThomsonTan: That's right. For long prototype chain and most of the prototype objects are non built-in objects, this will make the lookup much slower. For this case, I think we could move firstPrototype up after a prototype has been enumerated instead of keeping it fixed for a given prototype chain. This will fix the perf that non-built objects inherits from built-in objects in prototype chain, but it will still be an issue if built-in objects are inherited from non built-in objects. Wondering whether the latter case is typical? If it is, we could add some heuristics like traverse the whole chain after some number of downward look-ups? I will do the former optimization. May do the latter if necessary. What do you think, @curtisman ?

curtisman: I don't have data to back this up, but I think it is common on the web (with mix in, and monkey patching) that the prototype chain may have a fix of built in prototype version non-built in prototypes.

I am wondering whether this optimization is worth it at all to skip the any prototype if the prototype chain has any enumerable properties at all.  Can we measure how much that is going to help us?  Does it impact benchmark if we remove it?

curtisman: Just to illustrate the problem:

`function createPrototypeChainCtor(depth, protoCtor)
{
    var ctor = function() {};
    if (protoCtor)
        ctor.prototype = new protoCtor();
    for (var i = 0; i < 100; i++)
    {
       ctor.prototype["prop_" + depth + "_" + i] = i;
    }
    if (depth == 0) { return ctor; }
    return createPrototypeChainCtor( depth - 1, ctor);
}

var ctor = createPrototypeChainCtor(5);

var obj = new ctor();
var d = new Date()

for (var i = 0; i < 10000; i++)
{
    for (var s in obj)
    {
    }
}


console.log(new Date() - d);
`

This takes 3.5x more time after your change

ThomsonTan: Yes, downward look-up is slow in this case, but the first optimization I proposed above could eliminate all the extra look-ups. I will verify the perf impact of plain look-up.

ThomsonTan: I've made 2 changes:
1. Return the first prototype object and don't go up even if current prototype object is marked as no enumerable property.
2. Remove downward lookup loop.
For the perf benchmark for...in is optimized for, there is a stable score drop (~2%) when comparing with chakra before this change.
Seems skipping the prototype without any enumerable properties is necessary option. For an simple case, for...in on `Array` (this might not be a good pattern, but others are similar) will benefit from this optimization (no need to visit all the properties in Array.prototype and Object.prototype)

curtisman: Oh... I think you misunderstood.  I think should keep the optimization where the whole prototype chain has no enumerable properties. 

I am suggesting is that we don't skip any prototype if there are any enumerable property on the prototype chain, which I think is what you are suggesting, right?

ThomsonTan: Ok. With skipping prototype chain lookup when all prototype objects have no enumerable objects, the perf impact is flat now. But this makes the lookup very easy to be affected by any change in the prototype chain, like a minor change to any prototype object (JS framework is likely to do this?) will fallback almost all prototype to full lookup but this looks more conservative. 
Which option do you prefer, this option or the optimization I suggested on the existing ineffectiveness?

curtisman: Sorry, I guess I don't understand what your solution is
> Return the first prototype object and don't go up even if current prototype object is marked as no enumerable property.
Isn't that the same at not skip any prototype at all if there are any enumerable object in the prototype change?

ThomsonTan: My suggestion is that we still skip to the first prototype with enumerable properties, but don't skip non-visited prototype for downward lookup (the assumption is even some JS framework may add method to built-in prototype, like `Object.prototype`, the number is very small so the downward lookup number is also small, not like the test case you've made before). The revision is excluding prototype from downward lookup when it is visited (this led to 3.5 times slow in your previous example), it eliminates the unnecessary downward lookup if the most of the prototype have enumerable properties (they will be visited in sequential).

In some extreme cases like a lot of properties needs downward lookup (like if any code only adds tons of methods to `Object.prototype`, we could add a heuristic that fallback to full lookup, as you suggested, after some number of downward search.

This is trying avoid full lookup if possible, but I think the plain heuristic for full lookup (do full search if there is any enumerable properties in the chain) should be good in practical.

PR #2176, opened 2016-12-09
VSO 9848500: Change Assert to FailFast for when DisposeObjects is called from NoScriptScope. The caller should ensure this doesn't happen.

dilijev:


lib/Runtime/Base/ThreadContext.cpp
```diff
@@ -2111,15 +2111,19 @@ ThreadContext::ExecuteRecyclerCollectionFunction(Recycler * recycler, Collection
...
     {
         // Don't dispose objects when implicit calls are disabled, since disposing may cause implicit calls. Objects will remain
         // in the dispose queue and will be disposed later when implicit calls are not disabled.
         return;
     }
 
-    // we shouldn't dispose in noscriptscope as it might lead to script execution.
-    Assert(!this->IsNoScriptScope());
+    // We shouldn't DisposeObjects in NoScriptScope as this might lead to script execution.
+    // Callers of DisposeObjects should ensure !IsNoScriptScope() before calling DisposeObjects.
```
ThomsonTan: >Callers of DisposeObjects should ensure !IsNoScriptScope() before calling DisposeObjects. [](start = 7, length = 89)

How could script be executed when disposing?

PR #2177, opened 2016-12-09
VSO 8207235: Reflect.apply and Reflect.construct require `argumentsList` params (and the value must be an array or array-like object).

dilijev:


dilijev:
Thanks @Cellule for the help. Mind giving a quick review?

/cc @jdalton

Cellule:
I am just wondering if `object` and `arrays` are the only structures to be considered `ArrayLike`.

Otherwise, the error message bothers me a little because `Reflect.apply` doesn't expect the arguments object. It expects an `ArrayLike` object.
If we don't want to create a new error message for this case, I don't particularly mind.

dilijev:
See:
https://tc39.github.io/ecma262/#sec-reflect.apply
https://tc39.github.io/ecma262/#sec-createlistfromarraylike
https://tc39.github.io/ecma262/#sec-ecmascript-language-types

According to the spec, if I'm reading it correctly, there is no spec type for Array or TypedArray, and they are classed as special kinds of Object.

CreateListFromArrayLike requires that the input type is an Object -- any array-like object will do. There's no problem with passing an Object with no keys that look like array indexes, the resulting array will be empty. Passing an array-like Object or Array or TypedArray works.

I'm going to test native arrays and refactor the condition to a new method called  `JavascriptOperators::IsArrayLike`

I'll also create a new error to improve the message.

dilijev:
Okay, good call. It looks like the previous condition didn't cover TypedArray.

Improved check, added tests, and added a new error code.

Cellule:
FYI, I just merged a change to isolate Wasm related error messages, so you will have to rebase on release/1.4 and change the value of your new error message

dilijev:
@Cellule - just noticed and fixed, thanks!

Cellule:
At https://tc39.github.io/ecma262/#sec-createlistfromarraylike
The main check is `If Type(obj) is not Object, throw a TypeError exception.`
The current implemented check is insufficient because it misses a lot of other types that returns `"object"` to `typeof`

I suggest doing `Typeof(argArray, scriptContext) == scriptContext->GetLibrary()->GetObjectTypeDisplayString()` instead which will check if the typeof argArray is object 

I tested with types like Promise, Proxy, Error and more which all return `"object"` to `typeof`, but were not allowed in `Reflect.apply`

dilijev:
@Cellule how's this look?

Cellule:
Humm, `typeof null === "object"`, but from the spec, null doesn't have the object type... this is weird

ljharb:
@Cellule `Type(null)` is Null in the spec, but indeed `typeof null === 'object'` in JS. The `Type` abstract operation isn't the same as `typeof`.

Cellule:
Yeah that's what I see... I don't know if we have something in our code that matches the spec Types.
https://tc39.github.io/ecma262/#sec-typeof-operator-runtime-semantics-evaluation clearly shows the different spec type and why I was wrong to suggest to use typeof operator since Functions are accepted (Object type with typeof returning "Function") and null is not accepted (Null type with typeof returning "object")

dilijev:
Yeah apparently I forgot to rebuild before re-running the tests locally.

Cellule:
Okay, I think I got it... Try `JavascriptOperators::IsObjectType` that should give us what we need

dilijev:
@Cellule Looking good so far.

lib/Parser/rterrors.h
```diff
@@ -363,6 +363,7 @@ RT_ERROR_MSG(JSERR_InvalidOperationOnTypedArray, 5664, "", "The operation is not
 RT_ERROR_MSG(JSERR_CannotSuspendBuffer, 5665, "", "Current agent cannot be suspended", kjstRangeError, 0)
 RT_ERROR_MSG(JSERR_CantDeleteNonConfigProp, 5666, "Cannot delete non-configurable property '%s'", "Cannot delete non-configurable property", kjstTypeError, 0)
 RT_ERROR_MSG(JSERR_CantRedefineProp, 5667, "Cannot redefine property '%s'", "Cannot redefine property", kjstTypeError, 0)
+RT_ERROR_MSG(JSERR_FunctionArgument_NeedArrayLike, 5668, "%s: argument is not an array-like object", "Array-like object expected", kjstTypeError, 0)
```
ThomsonTan: >Array-like object expected [](start = 102, length = 26)

Should this be array or array-like object expected?

dilijev: Arrays are array-like objects. The spec uses the term ArrayLike in this context.

dilijev: I realize the semantics of the message might be clearer if it said "array or array-like object". I can make this change.

PR #2178, opened 2016-12-09
tweak rpc failfast

MikeHolman:
tweak rpc failfast to allow additional hresult (RPC_X_SS_IN_NULL_CONTEXT). This is because we may have null context handle if context handle for ScriptContext wasn't able to be initialized (due to jit process going away).

Also, add retry logic on the server liveness check. Server may be in process of closing but still alive. Worth waiting a bit to find out rather than having failfast right away.

### DIFF
--- a/lib/Backend/NativeCodeGenerator.cpp
+++ b/lib/Backend/NativeCodeGenerator.cpp
@@ -3646,17 +3646,10 @@ JITManager::HandleServerCallResult(HRESULT hr, RemoteCallType callType)
     // if jit process is terminated, we can handle certain abnormal hresults
 
     // we should not have RPC failure if JIT process is still around
-    DWORD sleepInterval = 0;
-    uint retries = 0;
-    while (WaitForSingleObject(GetJITManager()->GetServerHandle(), sleepInterval) != WAIT_OBJECT_0)
+    // since this is going to be a failfast, lets wait a bit in case server is in process of terminating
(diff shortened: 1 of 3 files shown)


MikeHolman:
@sermetiskin, @leirocks, @LouisLaf can you take a look?

lib/Backend/NativeCodeGenerator.cpp
```diff
@@ -3629,43 +3629,57 @@ bool NativeCodeGenerator::TryAggressiveInlining(Js::FunctionBody *const topFunct
...
-        }
+        break;
+    }
+    // if jit process is terminated, we can handle certain abnormal hresults
 
-        // we only expect to see these hresults in case server has been closed. failfast otherwise
-        if (hr != HRESULT_FROM_WIN32(RPC_S_CALL_FAILED) && hr != HRESULT_FROM_WIN32(RPC_S_CALL_FAILED_DNE))
+    // we should not have RPC failure if JIT process is still around
+    DWORD sleepInterval = 0;
+    uint retries = 0;
```
sermetiskin: Instead of this retry logic, why not use 250ms (which may or mat not be sufficient depending on the speed of the system, but will certainly decrease the odds) timeout value? WaitForSingleObject will return as soon as the process is signaled (dead). It would only wait until 250ms expires when process stays alive throughout the period.

MikeHolman:
@dotnet-bot test Ubuntu ubuntu_linux_debug

LouisLaf:
:shipit:

PR #2181, opened 2016-12-09
WASM - defer parsing js hook

Cellule:
Add a js test hook to force deferred parsing or not of WebAssembly Modules
This helps a lot to test spec invalid tests

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/2181)


Cellule:
ping

MikeHolman:
I think spec invalid tests should use validate. I'd prefer to change that instead.

Cellule:
I thought about this, the problem is we won't check if we are throwing the right type of error with those tests and we lose the change the check the error message (only while debugging as we don't print those).
I did find some errors regarding the type of errors using this approach

Cellule:
Closed in favor of #2199

PR #2182, opened 2016-12-09
VSO 9947214: Fix OACR warning in GetSharedPropertyGuards (using outparam count as index into allocated array).

dilijev:


dilijev:
@ianwjhalliday have a look?

PR #2184, opened 2016-12-10
Add a write barrier to FunctionInfo::functionBodyImpl. 

pleath:
This has not been needed in the past, but redeferral now experiences symptoms of missing write barrier, i.e., memory pointed to by a ParseableFunctionInfo created by redeferral is freed/overwritten.


pleath:
@leirocks Can you look at this one?

PR #2185, opened 2016-12-10
validate module addresses sent to JIT process

MikeHolman:


lib/JITServer/JITServer.cpp
```diff
@@ -166,24 +166,119 @@ ServerInitializeThreadContext(
...
+    {
+        goto FailureCleanup;
+    }
+
+    return hr;
+
+FailureCleanup:
+    if (moduleArray != nullptr && moduleArray != stackArray)
+    {
+        HeapDeleteArray(arraySize, moduleArray);
```
LouisLaf: Don't you want to free this in the success path as well?

lib/JITServer/JITServer.cpp
```diff
@@ -166,24 +166,119 @@ ServerInitializeThreadContext(
...
+        goto FailureCleanup;
+    }
+
+    // if our stack array does not have enough size, allocate an array large enough to handle all the modules
+    if (!EnumProcessModules(processHandle, stackArray, arraySize, &neededBytes))
+    {
+        Assert(neededBytes % sizeof(HMODULE) == 0);
+
+        // add 32 to give us reasonable buffer in case more modules are loading while we are querying
+        arraySize = neededBytes / sizeof(HMODULE) + 32;
```
leirocks: > + 32 [](start = 49, length = 5)

why not just use a while loop to query until succeed or non-size related error happened, instead of using this speculated size?

lib/JITServer/JITServer.cpp
```diff
@@ -166,24 +166,119 @@ ServerInitializeThreadContext(
...
+                hr = E_ACCESSDENIED;
+                goto FailureCleanup;
+            }
+        }
+        else
+        {
+            continue;
+        }
+
+        if(GetModuleFileNameExW(processHandle, moduleArray[i], modName, MAX_PATH) == 0)
```
leirocks: >moduleArray[i] [](start = 47, length = 14)

can we directly use the passed in module address to query the module name cross process? if yes you don't need to enumerate all the modules, right?

MikeHolman: That's a good question, something I wasn't entirely sure on as well. Asked our security people, and they think it would be best not to allow user to control the HMODULE.

leirocks: interesting, the HMODULE is just a number, even after enumerating we are still using the same number to get the module name. with the enumerating we have extra verification that make sure the number is in the list, not sure how useful this is though.

just thought about this a bit, I think the concern is on the processHandle part. if the attacker fake both processHandle and the moduleAddress (to be another running content process), the new verification will succeed.

looks we don't even need the client to provide the module address, we can just query it in the server side with the processHandle. we need to make sure the processHandle's fidelity. what we can do is:
  - use processHandle to get a processId
  - use RpcServerInqCallAttributes to get the caller(client) processId
  - verify the two processIds are the same.

MikeHolman: Good idea, I'll add validation of processHandle as well. I guess we could avoid user passing module address, but probably better to avoid doing name lookup and string compare for every loaded module.

MikeHolman:
Apparently psapi methods are sufficiently unreliable that we should do something else.

lib/JITServer/JITServer.cpp
ThomsonTan: >arraySize = neededBytes / sizeof(HMODULE) + 32; [](start = 8, length = 47)

Check GetLastError() before assuming arraySize is too small?

PR #2188, opened 2016-12-12
xplat: link stdlibc++ statically by default

obastemur:
We already had `--embed-icu` for ICU static linking. Besides, we had also removed libunwind dependency.

This PR forward we should have a smoother cross platform binary distribution. Statically embedding the dependencies, increases the final `CH` binary size drastically. However, embedders of ChakraCore are free to decide whether they want to link the dependencies statically or not as before.

Also minor cleanup on PAL's CMakeLists

### DIFF
--- a/bin/ChakraCore/CMakeLists.txt
+++ b/bin/ChakraCore/CMakeLists.txt
@@ -30,6 +30,7 @@ if(CMAKE_SYSTEM_NAME STREQUAL Linux)
   set(LINKER_END_GROUP
     -Wl,--no-whole-archive
     -Wl,--end-group
+    -static-libstdc++
     )
 elseif(CMAKE_SYSTEM_NAME STREQUAL Darwin)
   set(LINKER_START_GROUP -Wl,-force_load,)
@@ -44,7 +45,6 @@ set(lib_target "${lib_target}"
   Chakra.Jsrt
   ${LINKER_END_GROUP}
   pthread
-  stdc++
   dl
   ${ICULIB}
   )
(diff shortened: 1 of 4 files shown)


bin/ChakraCore/CMakeLists.txt
```diff
@@ -30,6 +30,7 @@ if(CMAKE_SYSTEM_NAME STREQUAL Linux)
   set(LINKER_END_GROUP
     -Wl,--no-whole-archive
     -Wl,--end-group
+    -static-libstdc++
```
obastemur: Apple doesn't support system libraries are statically linked and we target OSX 10.8 SDK.. So it is expected to work 10.8+

jianchun: Makes sense.

jianchun: Does this line means "libChakraCore.so" always statically linked to stdc++? Did you observe significant size increase?

When you say "However, embedders of ChakraCore are free to decide whether they want to link the dependencies statically or not as before", is that option only for static libs, since no customization exists for "libChakraCore.so"?

obastemur: Yes. Aprox 1Mb change in size and yes, only for static lib.

Maybe we could have `--embed-static` or something like that for `build.sh` but I didn't want to separate embedded libc++ testing from not embedded one. This is the package we will release and better we use it and test it all the time ?

jianchun: I agree.

obastemur:

PR #2189, opened 2016-12-12
Fix for missed stack references to FunctionBody in redeferral

pleath:
(Thanks to Curtis for help on this one.) Make the redeferral callback earlier in GC so that new FunctionBody references can't appear on the stack between the time that GC scans the stack and the time we gather our set of legal redeferral candidates.


pleath:
@curtisman

digitalinfinity:
:shipit:

PR #2190, opened 2016-12-12
Fix type assert in DbCheckPostLower hitting in swb branch

meg-gupta:


meg-gupta:
@jianchun Can you please review ?

PR #2195, opened 2016-12-13
need marshal in JavascriptProxy::GetValueFromDescriptor

leirocks:
my change 6aa7967 revealed this bug


test/es6/proxybug.js
```diff
@@ -2,6 +2,9 @@
 // Copyright (C) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE.txt file in the project root for full license information.
 //-------------------------------------------------------------------------------------------------------
+
+var passed;
```
leirocks: no actually for now, other test is actually hitting assertion. can move the assertion tests to the top. however adding the passed can make it more flexible if we have multiple failure locations in the future

PR #2196, opened 2016-12-13
Change to address CVE-2016-7202

suwc:
Heap overflow in Array.prototype.reverse
In Array.prototype.reverse, array length is cached and used in ReverseHelper().
ReverseHelper() could invoke FillFromPrototypes(), which can cause a side-effect on the array,
including changing its length. Therefore, the use of cached array length to calculate segment left index could result in overflow. Fix by clamping array length at zero.


suwc:
@pleath or @boingoing could you approve pls?

boingoing:
LGTM

PR #2197, opened 2016-12-13
Move etwtest.js test back to ChakraFull

ianwjhalliday:
This test doesn't work in ChakraCore because it requires
NativeUnitTests.dll which only builds in ChakraFull.

Move the test to unittest\Miscellaneous from core\test\Miscellaneous.

Fixes VSO 9760284


ianwjhalliday:
@tcare @digitalinfinity

tcare:
LGTM

PR #2198, opened 2016-12-13
Keep property records live on IteratorObjectEnumerator

rajatd:
We need to keep property records for properties being enumerated live during enumeration. For dynamic objects, type handlers hold references to property records and for non-existent properties, we have code in ForInObjectEnumerator::MoveAndGetNext to keep the new records live. 

Taking care of this case for Proxies which use IteratorObjectEnumerator

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/2198)

### DIFF
--- a/lib/Runtime/Library/IteratorObjectEnumerator.cpp
+++ b/lib/Runtime/Library/IteratorObjectEnumerator.cpp
@@ -45,6 +45,11 @@ namespace Js
             {
                 JavascriptString* propertyName = JavascriptConversion::ToString(currentIndex, scriptContext);
                 GetScriptContext()->GetOrAddPropertyRecord(propertyName->GetString(), propertyName->GetLength(), &propertyRecord);
+
+                // Need to keep property records alive during enumeration to prevent collection
+                // and eventual reuse during the same enumeration. For DynamicObjects, property
+                // records are kept alive by type handlers.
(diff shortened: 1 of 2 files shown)


rajatd:
@digitalinfinity could you please take a look?

ThomsonTan:
>         Assert(FALSE);

Reset `this->propertyRecords` when reset the enumerator? Also will `this->propertyRecords` be freed when destructing IteratorObjectEnumerator?

---
Refers to: lib/Runtime/Library/IteratorObjectEnumerator.cpp:63 in ba3244d. [](commit_id = ba3244d1249d361545ccc1b9a06732e46ecda2f1, deletion_comment = False)

tcare:
Looks good, if they get cleaned up

tcare:
I looked through this, and we generally follow the convention of throwing away the enumerator at the end unless it can be reset (e.g. CustomEnumerator.) We don't expect to reset IteratorObjectEnumerator, so it is fine to leave the references attached to the enumerator reference. OK to merge.

rajatd:
@tcare @digitalinfinity - thanks for the reviews.
@ThomsonTan The Assert in the Reset method implies that we should never call that method. Also, when the enumerator is thrown away (as @tcare mentioned), the propertyRecord list will become available for collection

PR #2199, opened 2016-12-13
WScript.Flag

Cellule:
Add WScript.Flag to change config flag from javascript

The main reason for this is to allows to change the behavior of WebAssembly features from javascript to support the spec tests which needs different config within a single file

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/2199)

### DIFF
--- a/bin/ch/WScriptJsrt.cpp
+++ b/bin/ch/WScriptJsrt.cpp
@@ -1035,6 +1035,8 @@ JsValueRef __stdcall WScriptJsrt::FlagCallback(JsValueRef callee, bool isConstru
     JsErrorCode errorCode = JsNoError;
 
     IfJsrtErrorSetGo(ChakraRTInterface::JsGetUndefinedValue(&returnValue));
+
+#if ENABLE_DEBUG_CONFIG_OPTIONS
     if (argumentCount > 1)
     {
         AutoString cmd;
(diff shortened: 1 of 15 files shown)


Cellule:
@MikeHolman @LouisLaf @dilijev Can one of you guys review this please and also matching jshost review ?

Cellule:
In order to make this feature more relevant I changed how we parse phases to avoid discrepancies between the on and off phases.
Since `OnFlag` and `OffFlag` are separate flags, you can end up with turning on and off the same feature at the same time. 
This means, it now falls on how the flags is looked at to determine if it is actually on or off.

The change I made will remove the opposite flag when another is used.
This means, if you do `-off:FullJit -on:FullJit:1.4` now, it will correctly turn off full jit for everything except function 1.4 where as before it would turn off FullJit for everything because implementation looks at the off flag and not the on flag.

It also means, that if you run
```js
WScript.Flag("-on:FullJit"); // FullJit is on
...
WScript.Flag("-off:FullJit"); // FullJit on flag is removed and off flag is added
...
WScript.Flag("-on:FullJit"); // FullJit off flag is removed and on flag is added
```
it will behave as you would expect.

Cellule:
ping

LouisLaf:
:shipit:

dilijev:
LGTM

PR #2200, opened 2016-12-14
xplat: shared library no longer requires an explicit call to dllmain

obastemur:


obastemur:
Fixes #1811

bin/ChakraCore/ChakraCoreShared.cpp
```diff
@@ -0,0 +1,33 @@
...
+    JsRuntimeAttributes attr;
+    JsCreateRuntime(attr, nullptr, runtime);
+}
+#endif
+
+__attribute__ ((visibility ("default")))
+EXTERN_C BOOL WINAPI DllMain(HINSTANCE hmod, DWORD dwReason, PVOID pvReserved)
+{
+    // xplat doesn't need DllMain is being called.
+    // Instead, we handle attach / detach stuff internally
```
jianchun: Could you please add more comment here to describe where/how we handle those attach/detach  internally?

obastemur: Okay. Added the location of the file that is responsible from the internal mechanism. There are additional comments there.

obastemur:

bin/ChakraCore/ChakraCoreDllFunc.cpp
```diff
@@ -19,26 +19,8 @@
...
-}
-#endif
-
 extern HANDLE g_hInstance;
-#ifdef _WIN32
 static ATOM  lockedDll = 0;
-#endif
-
-#ifdef _MSC_VER
-#define EXPORT_FUNC
```
Cellule: The removal of this break nojit builds. At the end of this file there are 2 functions under `#if !ENABLE_NATIVE_CODEGEN` that are still using `EXPORT_FUNC`. I am not sure what the exact fix is, could you take a look please

PR #2134, opened 2016-12-01
cross-site fix for unshift api

akroshg:

PR #2201, opened 2016-12-14
Array.prototype.unshift does not marshal parameters correctly

tcare:
Fixes OS 9357224.

Array.prototype.unshift prepends array values to 'this' array by calling JavascriptArray::FillFromArgs. FillFromArgs makes the assumption that values are already marshalled to the same context as 'this'. It is possible to have a situation in Array.prototype.unshift where the source array is in another context than the parameters passed in, triggering the assert. Other users of FillFromArgs (e.g. new Array()) don't seem to be able to hit this situation. Fix is to marshal values as we iterate in FillFromArgs.

### DIFF
--- a/lib/Runtime/Library/JavascriptArray.cpp
+++ b/lib/Runtime/Library/JavascriptArray.cpp
@@ -7563,7 +7563,7 @@ Case0:
         {
            return res;
         }
-        if (JavascriptArray::Is(args[0]))
+        if (JavascriptArray::Is(args[0]) && !JavascriptArray::FromVar(args[0])->IsCrossSiteObject())
         {
 #if ENABLE_COPYONACCESS_ARRAY
             JavascriptLibrary::CheckAndConvertCopyOnAccessNativeIntArray<Var>(args[0]);
--- a/test/Array/shift_unshift.baseline
+++ b/test/Array/shift_unshift.baseline
@@ -48,6 +48,7 @@ e instanceOf TypeError = true
 a.length = 1
 ary.length = 18
 arr.length = 6
+Crosssite new length: 2
 Overridden unshift
 Overridden unshift
 Overridden unshift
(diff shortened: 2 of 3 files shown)


tcare:
@pleath @aneeshdk please review

pleath:
LGTM. Thanks.

tcare:
It looks like Akrosh fixed this in #2134 but did not merge it. He takes a slightly different approach, by avoiding the if (JavascriptArray()) path if we have a cross site object. I am fine with either fix, do you have any preference?

dilijev:
I like this PR because it includes a test (and Akrosh is OOF) -- but of course, you could adopt his change if you decided that was the better approach.

dilijev:
@tcare Should this be retargeted to release/1.4?

tcare:
Yes, you're right. I will await Paul's offline signoff and push there.

tcare:
After speaking with Curtis, Akrosh's fix in #2134 is correct. I will check that in.

dilijev:
Maybe update this PR to defer to that fix but still add the tests?

tcare:
Yep that's what 348cd33 is :)

dilijev:
@tcare oh gotcha, didn't notice the update

lib/Runtime/Library/JavascriptArray.cpp
```diff
@@ -7563,7 +7563,7 @@ namespace Js
         {
            return res;
         }
-        if (JavascriptArray::Is(args[0]))
+        if (JavascriptArray::Is(args[0]) && !JavascriptArray::FromVar(args[0])->IsCrossSiteObject())
```
dilijev: Looks like this is still present in this change?

tcare: That's Akrosh's fix. Mine was in FillFromArgs.

dilijev: Ok I guess it will just merge down. Anyway you'll want to retarget this branch to release/1.4 and rebase.

tcare: Ah I missed this targeted master. Will update

dilijev:

PR #2202, opened 2016-12-14
Fix handling of inlined ArgIn_Rest while mapping formals during inlining

meg-gupta:
While mapping actuals during inlining, do not handle ArgIn_Rest when the
instruction was inlined originally in the inlinee. Existing mapping code was
resulting in incorrect codegen when mapping ArgIn_Rest instruction which was
originally inlined into the inlinee. Because the mapping code queried
formalCount of the inlinee instead of the originally inlined function of
which ArgIn_Rest was part of, and created an incorrect size NewScArray.


meg-gupta:

rajatd:
Looks good to me. Add a unit test and make sure to run exprgen for a long time

lib/Backend/Inline.cpp
```diff
@@ -4667,6 +4667,10 @@ Inline::MapFormals(Func *inlinee,
             {
                 break;
             }
+            if (instr->m_func != inlinee)
```
ThomsonTan: >instr->m_func != inlinee [](start = 16, length = 24)

Does this mean there could be some instruction before inlinee->m_headInstr but still belongs to inlinee?

meg-gupta: The check only verifies if the ArgIn_rest belongs to the inline or the inner inlinee.

PR #2203, opened 2016-12-14
Object type spec bug fix

rajatd:
If the flow based type doesn't match the type on the objTypeSpecFldInfo and we need a monomorphic type check, indicate a type mismatch for that operand and avoid fast path.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/2203)


rajatd:
@pleath please take a look. Let me know if you think that existing callsites of `opnd->SetTypeMismatch` should be augmented to cover this case instead.

rajatd:

lib/Backend/BackwardPass.cpp
```diff
@@ -4475,6 +4475,10 @@ BackwardPass::TrackObjTypeSpecProperties(IR::PropertySymOpnd *opnd, BasicBlock *
                         // fixed field not loaded from prototype.) Note the IsTypeAvailable test above: only do this at
                         // the initial type check that protects this path.
                         opnd->SetMonoGuardType(bucket->GetMonoGuardType());
+                        if (opnd->GetMonoGuardType() != (opnd->IsMono() ? opnd->GetType() : opnd->GetFirstEquivalentType()))
+                        {
+                            opnd->SetTypeMismatch(true);
```
ThomsonTan: >opnd->SetTypeMismatch(true); [](start = 28, length = 28)

Is it possible to add a test case for this?

pleath:
What does Lower generate for this type check if "type mismatch" is set?

rajatd:
@pleath , lowerer generates fast path with inline cache lookup and a helper call if the inline cache misses.

ThomsonTan:
Why the lower can still generates fast path even if type mismatch?

---
In reply to: [267412547](https://github.com/Microsoft/ChakraCore/pull/2203#issuecomment-267412547) [](ancestors = 267412547)

rajatd:
@ThomsonTan, hmm, it does sound unnecessary. @pleath what do you think?

ThomsonTan:
:shipit:

rajatd:
the underlying issue was fixed by #3341

PR #2205, opened 2016-12-14
Trivial redeferral fix

pleath:
In a compiled function nested in a redeferred function, we won't make use of the literal regex index. Don't let byte code gen increment the literal regex count to avoid breaking thread-safety requirements of the compact counters.


pleath:
@rajatd This is a tiny one.

rajatd:
:shipit:

PR #1990, opened 2016-11-11
Fix VS2013 build break

MikeHolman:
Fixes #2026

Issue #1918, opened 2016-11-04
JSRT: uint8_t vs char for Utf8 interface


obastemur:
Some JSRT API accept `char` for Utf8 named functions while others `uint8_t`.

We had a small discussion on this. Shared my opinion below;

 - `char` is more developer friendly and all C / IO comes with `char`.
     If a developer has nothing to do with utf8 encoding / flags etc.
     there is no reason for her/him to use `unsigned`. At least I can’t think of any.
 
 - No matter what we choose for external type, internally we will continue using `unsigned`.

 - Having `char` argument may make ASCII based string interface obsolete.

/cc @agarwal-sandeep  @curtisman @jianchun @liminzhu

liminzhu:
What I care most is that we use one consistent type across. @agarwal-sandeep @curtisman @jianchun on which one is better?

kphillisjr:
For user facing interfaces, I agree that the use of ```char``` is the best argument choice for 8-bit characters.

jianchun:
> Having char argument may make ASCII based string interface obsolete.

The reason for one more variant was to allow node-chakracore to support v8's OneByte string. Thinking it over, if we don't care about perf, node-chakracore can implement that variant in chakrashim. chakrashim can just use the Utf16 variant and convert back and forth. @kunalspathak @agarwal-sandeep

liminzhu:
Had an offline chat with @jianchun, we can probably leave the ASCII variant and only ship utf8/16 versions of Js[Create/Copy]String in JSRT. NodeChakraCore can support the ASCII variant through the shim. Should it turn out to be a problem, we can always add that variant support later.

It also sounds like `char` is a more friendly interface for utf8, so I advocate we use that for `JsCreateStringUtf8`, `JsCreateStringUtf16` and `JsCreatePropertyIdUtf8`. 

@obastemur @agarwal-sandeep @digitalinfinity @curtisman ?

obastemur:
@liminzhu Sounds good to me.

> Thinking it over, if we don't care about perf,.....

@jianchun I kinda agree. IMO, the perf is more critical for CreateProperty from ASCII source instead of create `a string` from ASCII source. Most of the time we / others approach external or even internal (JS land) strings as Utf8.

obastemur:
We've been discussing this internally. We decided to use `char` type and drop `Utf8` from names.

dilijev:
[...]

PR #2207, opened 2016-12-15
Fix Legacy & NoJit

Cellule:
Includes #1990 

This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/2207)

### DIFF
--- a/bin/ch/ChakraRtInterface.h
+++ b/bin/ch/ChakraRtInterface.h
@@ -73,8 +73,8 @@ struct JsAPIHooks
     typedef JsErrorCode(WINAPI *JsrtParse)(JsValueRef script, JsSourceContext sourceContext, JsValueRef sourceUrl, JsParseScriptAttributes parseAttributes, JsValueRef *result);
     typedef JsErrorCode(WINAPI *JsrtSerialize)(JsValueRef script, JsValueRef *buffer, JsParseScriptAttributes parseAttributes);
     typedef JsErrorCode(WINAPI *JsrtRunSerialized)(JsValueRef buffer, JsSerializedLoadScriptCallback scriptLoadCallback, JsSourceContext sourceContext, JsValueRef sourceUrl, JsValueRef * result);
-    typedef JsErrorCode(WINAPI *JsrtCopyStringUtf8)(JsValueRef value, uint8_t* buffer, size_t bufferSize, size_t* written);
-    typedef JsErrorCode(WINAPI *JsrtCreateStringUtf8)(const uint8_t *content, size_t length, JsValueRef *value);
+    typedef JsErrorCode(WINAPI *JsrtCopyStringUtf8)(JsValueRef value, unsigned char* buffer, size_t bufferSize, size_t* written);
+    typedef JsErrorCode(WINAPI *JsrtCreateStringUtf8)(const unsigned char *content, size_t length, JsValueRef *value);
     typedef JsErrorCode(WINAPI *JsrtCreateExternalArrayBuffer)(void *data, unsigned int byteLength, JsFinalizeCallback finalizeCallback, void *callbackState, JsValueRef *result);
     typedef JsErrorCode(WINAPI *JsrtCreatePropertyIdUtf8)(const char *name, size_t length, JsPropertyIdRef *propertyId);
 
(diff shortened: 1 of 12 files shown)


Cellule:
@dotnet-bot test nojit tests please

Cellule:
@dotnet-bot test nojit tests please

Cellule:
The failure seems to all be slow tests.
I guess we run slow tests on legacy build.
I will merge this for now and investigate the slow tests failures

dilijev:
Yes, it seems to be the case that slow tests are failing right now in release/1.4; https://github.com/Microsoft/ChakraCore/wiki/Build-Status-%28release-1.4%29#builds-with-slow-tests

And yes, I believe we run the slow tests in legacy config because the only automatic legacy builds are daily.

lib/Jsrt/ChakraCore.h
```diff
@@ -212,7 +212,7 @@ CHAKRA_API
 /// </returns>
 CHAKRA_API
     JsCreateStringUtf8(
-        _In_ const uint8_t *content,
+        _In_ const unsigned char *content,
```
obastemur: Please do not update this file with `unsigned char`. See #1918

bin/ch/ChakraRtInterface.h
```diff
@@ -368,8 +368,8 @@ class ChakraRTInterface
...
     static JsErrorCode WINAPI JsSerialize(JsValueRef script, ChakraBytePtr buffer, unsigned int *bufferSize, JsParseScriptAttributes parseAttributes) { return HOOK_JS_API(Serialize(script, buffer, bufferSize, parseAttributes)); }
     static JsErrorCode WINAPI JsRunSerialized(ChakraBytePtr buffer, JsSerializedLoadScriptCallback scriptLoadCallback, JsSourceContext sourceContext, JsValueRef sourceUrl, JsValueRef * result) { return HOOK_JS_API(RunSerialized(buffer, scriptLoadCallback, sourceContext, sourceUrl, result)); }
-    static JsErrorCode WINAPI JsCopyStringUtf8(JsValueRef value, uint8_t* buffer, size_t bufferSize, size_t* written) { return HOOK_JS_API(CopyStringUtf8(value, buffer, bufferSize, written)); }
-    static JsErrorCode WINAPI JsCreateStringUtf8(const uint8_t *content, size_t length, JsValueRef *value) { return HOOK_JS_API(CreateStringUtf8(content, length, value)); }
+    static JsErrorCode WINAPI JsCopyStringUtf8(JsValueRef value, unsigned char* buffer, size_t bufferSize, size_t* written) { return HOOK_JS_API(CopyStringUtf8(value, buffer, bufferSize, written)); }
```
obastemur: similarly, all these `unsigned char` will be updated to `char`

Cellule:
@obastemur Would it be possible to submit like this so this PR doesn't change any functionality, but simply make compilation work.
We will be able to change `unsigned char` to `char` afterward in a more relevant PR that addresses the issue

Cellule:
Okay so I will be submitting this PR as it is.
We should address #1918 in a separate PR

obastemur:
> Okay so I will be submitting this PR as it is.
We should address #1918 in a separate PR

Can someone share a `TECHNICAL` reason for merging this AS IS ?

@Cellule @dilijev ?

Cellule:
The technical reason is to avoid technical problems that could occur with such a change. 
The issue you raised can have other impacts or compatibility issues with type conversions, plus the point of this pr is to fix a build break, not implement a change in the api.

As far as api wise, nothing has changed except we don't use a typedef that can be missing in some compilers.
The issue #1918 can still be addressed by someone who has better context than me in the matter.

obastemur:
> we don't use a typedef that can be missing in some compilers.

I can't think of a compiler (in last XX years) that is missing `typedef`. Could you give me one that we support but that doesn't have `typedef` support?

FYI; we already use `typedef` on user facing header files. (see ChakraCore.h)

 > The issue you raised can have other impacts or compatibility issues with type conversions

Sorry, I don't understand this either. How it is different than defining `unsigned short` directly ? My suggestion was `use typedef ... when char_t is not available`.. What exactly this supposed to break?

> The issue #1918 can still be addressed by someone who has better context than me in the matter.

I thought that's why we had the reviews here.

Cellule:
Sorry, what I meant is not to support compiler that don't have the ability to use typedef, I meant these typedef `uint8_t` and `uint16_t` in particular in those files.

I'm starting to think there was some confusion here. 
I thought your initial review comment was asking me to change from `uint8_t` to `char` which is a change in the typing instead of changing `uint8_t` to `unsigned char`.
If not, I'm sorry, but I didn't understand what you were asking.

dilijev:
This PR was intended to fix legacy (Dev12) compiler errors only. No typed or semantics were changed (the only types changed were substituted for equivalent types that fixed the legacy compiler errors).

Since that was precisely the scope of this PR, and precisely what this PR achieved, I don't see why anything else needs to be addressed here. Other issues should be discussed in the appropriate issues, and fixed in a separate PR.

obastemur:
> No actual types or code semantics were changed (the only source-code types changed were substituted for equivalent types that fixed the legacy compiler errors, as apparently uint8_t was not supported in the legacy compiler).

@dilijev this is a very interesting approach to builtin C++11 types and the history of Unicode. Why all these new types are appeared with C++11 could be a good starting point.

IMO the change here is beyond a style update and more importantly this sort of PRs shouldn't be rushed on merging. Also they should be reviewed carefully.

We have been discussing a type change in ChakraCore.h for a while just in order to update from one type to another virtually. No rush. Better we do this discussion on Github so we can receive the valuable opinions of our community.

If we are into fixing a legacy build break, we can perfectly target <C++11 compilers as given below;

```
#if (defined(_MSC_VER) && _MSC_VER < 1900) || (__cplusplus <= 199711L) // !C++11
typedef unsigned short char16_t;
typedef unsinged int uint16_t;
#endif
``` 
I will update this file back with the planned Jsrt changes.

Thanks for the discussion.

dilijev:
@obastemur You're right that a change which could affect API compatibility shouldn't be rushed. However there is already good issue in which to have the discussion continue rather than in an already-closed PR.

IMO this change was fine to fix a build issue (which was urgent given the impending release), with no API incompatibilities being introduced as of this change (except in APIs which are already incompatible with a previous release). It may be the case that the better approach here was to add the conditionally-compiled typedefs you mentioned, but on the other hand, that would mean that any VS 2013 project wanting to use these APIs would have to either use the compatible existing type or create the same typedefs on their side as well. I discussed this point with @Cellule offline and he decided this was the better approach to fix the build problem and unblock the test coverage for those builds.

Given that a bigger (compatibility) change to these APIs had already been decided as of #1918 and your PR implements it, I think we can just move on to worrying about the final shape of the APIs, rather than the nature of this change, and of course there are better issues and PRs to discuss this.

PR #2211, opened 2016-12-15
xplat: fix position independent compiler direction

obastemur:
+ minor fixes to JavascriptOperatorA.S

### DIFF
--- a/bin/ChakraCore/CMakeLists.txt
+++ b/bin/ChakraCore/CMakeLists.txt
@@ -23,6 +23,7 @@ target_include_directories (
 #
 if(CC_TARGET_OS_ANDROID OR CC_TARGET_OS_LINUX)
   set(LINKER_START_GROUP
+    -fPIC
     -Wl,--start-group
     -Wl,--whole-archive
     )
--- a/bin/ch/CMakeLists.txt
+++ b/bin/ch/CMakeLists.txt
@@ -34,7 +34,7 @@ target_include_directories (ch
 
 if(STATIC_LIBRARY)
   if(CC_TARGET_OS_ANDROID OR CC_TARGET_OS_LINUX)
-    set(LINKER_START_GROUP -Wl,--start-group)
+    set(LINKER_START_GROUP -pie -Wl,--start-group)
     set(LINKER_END_GROUP -Wl,--end-group -static-libstdc++)
   elseif(CC_TARGET_OS_OSX)
     set(LINKER_START_GROUP -Wl,-force_load,)
--- a/lib/Runtime/Language/JavascriptStackWalker.cpp
+++ b/lib/Runtime/Language/JavascriptStackWalker.cpp
@@ -272,7 +272,7 @@ namespace Js
         {
             return inlinedFrameWalker.GetArgv(/* includeThis = */ false);
         }
-        else 
+        else
 #endif
             if (this->GetCurrentFunction()->GetFunctionInfo()->IsCoroutine())
         {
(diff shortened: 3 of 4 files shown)


obastemur:
Fixes #2009

jianchun:
@obastemur Apologize for my CR delay. I had to refresh my memory of this "fake frame" thing, esp. seeing you adding debug info directives to `amd64_CallWithFakeFrame` and adjusting its stack layout.

It appears to me adding debug info directives here is unnecessary. They are (almost) never needed. The goal of `amd64_CallWithFakeFrame` is to simulate our JIT frame stack layout. When an exception occurs in JIT code (e.g. a try block) and the system needs to rewind native stack, the .eh_frame data for the `container` JIT function (the function containing the try block) applies. Since `amd64_CallWithFakeFrame` creates exactly the same stack layout, the `container` JIT function's .eh_frame data helps rewind stack to parent caller (OP_TryCatch/OP_TryFinally) correctly.

You are correct that in `OP_TryFinally` the 2nd call to `amd64_CallWithFakeFrame` fails because the first call corrupted the stack. The corruption is actually from

```
        // The stack walker uses this marker to skip this frame.
        // Note that we use rip-relative addressing here since this
        // library is shared and therefore the code has to be position independent
        lea rax, [rip + C_FUNC(amd64_ReturnFromCallWithFakeFrame)]
        mov [rsp+8h], rax
```

Note that it puts address of `amd64_ReturnFromCallWithFakeFrame` to *`[rsp + 8h]`* -- a stack location in parent frame (below return address). In the case of `OP_TryFinally`, that location happens to store parameter `finallyAddr`, causing the 2nd call skip real finally block and directly jumps to `amd64_ReturnFromCallWithFakeFrame`.

Could you please try a different fix instead? Just delete the lines above. It should not write to parent frame. I searched and did not find any proof of the statements in the comment. That is probably outdated. You can even delete the `extern "C" void *amd64_ReturnFromCallWithFakeFrame(void);` declaration from JavascriptStackWalker.cpp.

The same code may need to be deleted from "JavascriptOperatorsA.asm" too.

obastemur:
@jianchun After seeing the usages below, I was more into fixing the problem on the place it happened.

see https://github.com/Microsoft/ChakraCore/blob/master/lib/Backend/LowerMDShared.cpp#L472-L500 

https://github.com/Microsoft/ChakraCore/blob/master/lib/Backend/amd64/LowererMDArch.cpp#L3110-L3133

in case the usage above is also `outdated`, agree cleanup is the best idea.

Opinions ?

obastemur:
BTW; @jianchun Thanks for the in details explanation above, for sure I really admire the quality comments like that. 👍

jianchun:
> @jianchun After seeing the usages below, I was more into fixing the problem on the place it happened.
> see https://github.com/Microsoft/ChakraCore/blob/master/lib/Backend/LowerMDShared.cpp#L472-L500
> https://github.com/Microsoft/ChakraCore/blob/master/lib/Backend/amd64/LowererMDArch.cpp#L3110-L3133
> in case the usage above is also outdated, agree cleanup is the best idea.
> Opinions ?

I also searched and read those code. I thought they were legitimate code, independent and unrelated to the problematic code I posted above. They push `amd64_ReturnFromCallWithFakeFrame` to stack and `RET`, that's how the `try` block code returns normally without exception. They don't use/need the problematic code above to store RET address. Any reason you concerned about them?

Thanks!

obastemur:
@jianchun Thanks for the review and awesome comments

PR #2216, opened 2016-12-15
Don't remove dead syms which are referenced in callSequence in infinite loop blocks

ThomsonTan:
When function with infinite loop is inlined as parameter evaluation for another function, callSequence is propagated to infinite loop blocks in forward pass, but usually the operands in callSequence are not live because there is no way to exit the loop (so it is uncontional infinite loop). This triggers assertion when filling bailout from callSequence (because no value for given sym). 

This change fixes it by not removing dead sym in symToValueMap if it is referenced in callSequence.


ThomsonTan:
@LouisLaf @rajatd could you please help review?

rajatd:
:shipit:

PR #2217, opened 2016-12-16
Remove '-oopjit-' flag from ChakraCore POGO training. Fixes #2213

### LINKED ISSUE #2213: pogo_training.ps1 using `-oopjit-` flag breaks POGO training for ChakraCore release builds
In ChakraFull we POGO Test builds (rather than Release builds) because we need performance-optimized Test builds for testing purposes. The pogo_training.ps1 script is used to do the POGO training in ChakraFull as well. In the case of ChakraFull, extra parameters like `-oopjit-` work just fine.

We had not yet noticed the ChakraCore problem because we have not yet made a release of ChakraCore since these flags were added (they were not present in release/1.3, but the flag (currently -oopjit-) was added after OOP JIT was added to the code base to ensure we train without using OOP JIT.

OOP JIT is explicitly disabled by default in ChakraCore. Therefore, we can remove this additional flag for ChakraCore POGO, and move that flag into the appropriate automation specific to ChakraFull.

See discussion here: https://github.com/Microsoft/ChakraCore/pull/1793#issuecomment-267065346

/cc @matthargett

dilijev:
Fixes #2213

### DIFF
--- a/Build/scripts/pgo/pogo_training.ps1
+++ b/Build/scripts/pgo/pogo_training.ps1
@@ -21,12 +21,14 @@ param (
     [Parameter(Mandatory=$True)]
     [string]$arch,
 
-    [ValidateSet("default", "codecoverage", "pogo")]
-    [string]$subtype = "default",
-
     # force callers to specify this in case of future use
     [Parameter(Mandatory=$True)]
     [string]$flavor,
+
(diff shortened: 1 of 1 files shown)


dilijev:
/cc @Cellule @boingoing 

@matthargett Could you give this a try and see if it fixes your issue?

PR #2224, opened 2016-12-16
android: ChakraCore compiles to Android ARMv7

obastemur:
disclaimer: Android is not officially supported.


obastemur:
Things left to do;

- fix linker issues and convert couple of MSVC ASM
- fix couple of todo items
- test on ADB

lib/Common/Core/SysInfo.cpp
```diff
@@ -254,7 +254,7 @@ AutoSystemInfo::SSE2Available() const
     return false; // TODO: xplat support
 #endif
 #else
-    #error Unsupported platform.
+    return false;
```
digitalinfinity: I'd leave the error here and add an explicit ifdef for ARMv7/Android here so that anyone adding new platform support is aware that this function is something they need to think about

lib/Common/CommonPal.h
```diff
@@ -637,9 +648,9 @@ namespace PlatformAgnostic
 #endif
     }
 
-    __forceinline unsigned char _BitTest(const LONG *_BitBase, int _BitPos)
+    __forceinline unsigned char _BitTest(LONG *_BitBase, int _BitPos)
```
digitalinfinity: Why was the const removed here? seems like allowing for const can result in better codegen by the compiler- if it's android specific, I'd prefer doing a const_cast when _bittest is called scoped to Android

lib/Common/CommonPal.h
```diff
@@ -621,7 +632,7 @@ namespace PlatformAgnostic
 {
     __forceinline unsigned char _BitTestAndSet(LONG *_BitBase, int _BitPos)
     {
-#if defined(__clang__)
+#if defined(__clang__) && !defined(_ARM_)
```
digitalinfinity: My ARM fu is weak, so a basic question here- is _ARM_ the same for Windows ARM vs Android ARM v7? Should it be? I know that Windows supports Thumb2, which I don't know if Android supports so the Windows ARM code might end up being slightly different from the Android ARM- I'm not sure what the macro model here should look like. @pleath @satheeshravi thoughts?

lib/Runtime/Language/Arguments.h
```diff
@@ -13,9 +13,8 @@
     va_start(_vl, callInfo);                                        \
     Js::Var* va = (Js::Var*)_vl
 #else
-#if defined(_M_X64) || defined(_M_IX86)
 // We use a custom calling convention to invoke JavascriptMethod based on
-// System V AMD64 ABI. At entry of JavascriptMethod the stack layout is:
+// System ABI. At entry of JavascriptMethod the stack layout is:
```
digitalinfinity: Interesting- are the calling conventions in the ABIs same for x86/AMD64/ARM on Linux?

lib/Runtime/Language/Arguments.h
```diff
@@ -75,6 +70,10 @@ inline int _count_args(const T1&, const T2&, const T3&, const T4&, Js::CallInfo
 #define CALL_ENTRYPOINT(entryPoint, function, callInfo, ...) \
     entryPoint(function, callInfo, nullptr, nullptr, nullptr, nullptr, \
                function, callInfo, ##__VA_ARGS__)
+#elif defined(_ARM_)
+// xplat-todo: fix me
```
digitalinfinity: Can you add some more description here? What needs to be fixed?

lib/Runtime/Library/JavascriptError.cpp
```diff
@@ -371,7 +371,9 @@ namespace Js
 
         if (FACILITY_CONTROL == HRESULT_FACILITY(hr) || FACILITY_JSCRIPT == HRESULT_FACILITY(hr))
         {
+#ifndef __ANDROID__
```
digitalinfinity: What is this change for? Does this function work xplat? It looks like its mostly getting windows specific errors so perhaps the entire if condition shouldn't disabled for Android, no?

pal/src/configure.cmake
```diff
@@ -1,3 +1,5 @@
+if (NOT CC_TARGET_OS_ANDROID)
```
digitalinfinity: On this note- have we cleaned up this file to make sure that it's checking for stuff that ChakraCore actually needs? When I imported the PAL, I didn't bother to clean this up but if our PAL is diverging, can you make a note to clean up this file at some point?

lib/Common/Core/SysInfo.cpp
obastemur: @digitalinfinity if it's not Intel, IMO `SSE2Available()` should return false. What other possibilities here?

lib/Common/CommonPal.h
obastemur: Clang intrinsic definition has no const.

lib/Runtime/Language/Arguments.h
obastemur: `AddressOfReturnAddress` is intrinsic

lib/Runtime/Library/JavascriptError.cpp
obastemur: `argList != nullptr` is wrong. I'm not sure how this works or if it works at all. Clearly it doesn't work with Clang 3.8 abi-arm-android. 

> `entire if condition shouldn't disabled for Android`

Sounds Good. Thanks

pal/src/configure.cmake
obastemur: PAL uses this file to feature detect on target. This is not an option when cross-compiling the code. 

> can you make a note to clean up this file at some point?

Good idea, will do. Thanks

obastemur:
> Can you also write up something on how to build CC on Android if anyone else wanted to try? Maybe just update our BUILDING file or add it to the wiki?

@digitalinfinity we need https://github.com/Microsoft/ChakraCore/pull/2224#issuecomment-267624637 steps are done for anyone to experience ChakraCore literally.

lib/Runtime/Language/Arguments.h
obastemur: updated this to `fix me ARM`

lib/Runtime/Library/JavascriptError.cpp
obastemur: Actually, shouldn't we consider argList will never be null and keep it as is here?

pal/src/configure.cmake
obastemur: added extra notes

obastemur:
Steps to compile: 

- Download Android NDK and extract
- Build/android_toolchain.sh <path to android NDK root folder>
- ./build.sh --target=android --arch=arm --no-icu --static --debug --no-jit

After some updates to this PR, currently there are only few linker errors left.

@digitalinfinity do you have any additional comments?

obastemur:
Android ARMv7 linker issues are also fixed (along with the implementation of ChakraCore ASM files)

obastemur:
Also targets RaspberryPI (see #2262 for additional requirements)

digitalinfinity:
No more comments from my side- thanks @obastemur

digitalinfinity:
LGTM

obastemur:
@digitalinfinity @ThomsonTan Thanks for the Reviews!!

PR #2225, opened 2016-12-16
Don't attempt to create a deferred stub for a non-deferred function. 

pleath:
Works around a parser anomaly that appeared with re-deferral. I'll keep investigating the cause of the anomaly.


pleath:
@tcare @aneeshdk Can either of you look at this one? (Thanks.)

tcare:
Looks good

PR #2227, opened 2016-12-16
swb: a few more unit test verification fixes

jianchun:
A few fixes from running unit test verification.

Previously missed disabling "memcpy" on "WriteBarrierPtr*".

List uses "bool isLeaf" to indicate leaf/non-leaf. Use the same info to
determine write barrier in CopyArray/qsort_s inside List.

### DIFF
--- a/lib/Backend/JnHelperMethod.cpp
+++ b/lib/Backend/JnHelperMethod.cpp
@@ -215,7 +215,7 @@ DECLSPEC_GUARDIGNORE  _NOINLINE intptr_t GetNonTableMethodAddress(ThreadContextI
         return SHIFT_CRT_ADDR(context, (int(*)(void *, void *, size_t))memcmp);
 
     case HelperMemCpy:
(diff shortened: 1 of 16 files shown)


jianchun:

jianchun:

lib/Backend/JnHelperMethod.cpp
```diff
@@ -215,7 +215,7 @@ DECLSPEC_GUARDIGNORE  _NOINLINE intptr_t GetNonTableMethodAddress(ThreadContextI
         return SHIFT_CRT_ADDR(context, (int(*)(void *, void *, size_t))memcmp);
 
     case HelperMemCpy:
-        return SHIFT_CRT_ADDR(context, (int(*)(void *, void *, size_t))memcpy);
+        return SHIFT_CRT_ADDR(context, (void*(*)(void *, void const*, size_t))memcpy);
```
jianchun: Because I now added missing memcpy overload. The original cast signature does not match and cannot figure out which overload to use. With this change it can.

leirocks:
:shipit:

PR #2230, opened 2016-12-17
Change to address CVE-2016-7287,CVE-2016-7286,CVE-2016-7288,CVE-2016-7296

boingoing:
Fix for SpreadArgs overrun
The array's length gets changed when we do the get-item this will overrun already allocated buffer. Fixed that by creating a local initialized that length.

Use-after-free in TypedArray.sort
%TypedArray%.prototype.sort has a helper method which performs the compare between two elements of the array. This helper takes an optional compare function callback, which is user code, and executes it (if present) to compute the sort order in a user-defined way. We call ToNumber on the return value from the compare function callback. The issue here is that this compare function can return an object which executes user code in the ToNumber conversion (via valueOf). This valueOf function is user code which can detach the buffer in the TypedArray. We check to see if the buffer has been detached after calling the compare function but we don't check again after calling ToNumber. If the valueOf call detaches the buffer, our sort function will re-order elements in the buffer's memory after it was detached and potentially free'd.
Fix is to detect the buffer detach after calling ToNumber and throw out of the helper to make sure we don't shuffle elements in the detached memory.

Uninitialized Memory in SIMD.toLocaleString
JavascriptSIMDObject::ToLocaleString explicitly handles the number of arguments it copies into a temporary array before passing this array to a toLocaleString helper. We have a check for 1, 2, or 3 arguments but calling the function with more than 3 arguments causes it to skip copying any of the incoming arguments into the temporary array. This leads to the toLocaleString helper using uninitialized values in the temporary array since we access this array using the original argument count. There's also another bug here which is that the toLocaleString helper can throw an exception which leaks the temporary array memory since it was allocated on the heap and is not free'd when this helper throws.
Fix both issues by allocating the temporary array on the stack with a fixed size of 3, clamping the number of incoming arguments to 3, and removing the explicit checks for the number of incoming arguments. Now we insert into the temporary array the two optional arguments if our original set of args includes those optional arguments.

Uninitialized Memory in SIMD.Load
All of the EntryLoad variants in the SIMD Javascript library directly access elements in the arguments array regardless of how many arguments are actually in the array. If user code calls this API with no arguments we will end up loading values which are unknown and potentially uninitiailized. Simple fix is to check the count of arguments passed-in to the function and use undefined for missing arguments.

Type Confusion in Internationalization Initialization
IntlEngineInterfaceExtensionObject::deletePrototypePropertyHelper loads a property from an object and uses it as if it is a DynamicObject even if the property isn't an object. We can run into two different problems here. First, we can end up using an uninitialized stack var if the object we're loading the property from doesn't have this property. Second, we can treat a non-object as an object leading to reading from arbitrary memory addresses.

### DIFF
--- a/lib/Runtime/Library/IntlEngineInterfaceExtensionObject.cpp
+++ b/lib/Runtime/Library/IntlEngineInterfaceExtensionObject.cpp
@@ -295,35 +295,46 @@ namespace Js
 
     void IntlEngineInterfaceExtensionObject::deletePrototypePropertyHelper(ScriptContext* scriptContext, DynamicObject* intlObject, Js::PropertyId objectPropertyId, Js::PropertyId getterFunctionId)
     {
-        DynamicObject *prototypeVal = nullptr;
+        DynamicObject *prototypeObject = nullptr;
         DynamicObject *functionObj = nullptr;
-        Var propertyValue;
-        Var getter;
-        Var setter;
+        Var propertyValue = nullptr;
+        Var prototypeValue = nullptr;
+        Var resolvedOptionsValue = nullptr;
+        Var getter = nullptr;
+        Var setter = nullptr;
 
-        if (!Js::JavascriptOperators::GetProperty(intlObject, objectPropertyId, &propertyValue, scriptContext))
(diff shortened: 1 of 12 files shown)


boingoing:
@pleath Please take a look if you don't mind

pleath:
:shipit:

PR #2232, opened 2016-12-17
Replace FunctionBody reference in ScopeSlots with FunctionInfo

pleath:
The ScopeSlots structure used to hold closure-captured variables has a reference to the scope in the form of a FunctionBody*. But that FunctionBody may become a stale reference after redeferral. Replace with an immutable FunctionInfo*.


rajatd:
:shipit:

PR #2233, opened 2016-12-17
Fix build errors with legacy compiler with nojit

Cellule:
I accidentally built with nojit and legacy at the same time and it found some interesting oversight

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/2233)

### DIFF
--- a/lib/Runtime/Language/InterpreterStackFrame.cpp
+++ b/lib/Runtime/Language/InterpreterStackFrame.cpp
@@ -1163,8 +1163,6 @@ namespace Js
             !(this->executeFunction->GetHasTry() && (PHASE_OFF((Js::JITLoopBodyInTryCatchPhase), this->executeFunction) || this->executeFunction->GetHasFinally())) &&
             (this->executeFunction->ForceJITLoopBody() || this->executeFunction->IsJitLoopBodyPhaseEnabled()) &&
             !this->executeFunction->IsInDebugMode();
-#else
-        const bool doJITLoopBody = false;
 #endif
 
         // Pick a version of the LoopBodyStart OpCode handlers that is hardcoded to do loop body JIT and
--- a/lib/Runtime/Library/WebAssemblyMemory.h
+++ b/lib/Runtime/Library/WebAssemblyMemory.h
@@ -9,6 +9,9 @@ namespace Js
 {
     class WebAssemblyMemory : public DynamicObject
     {
+    protected:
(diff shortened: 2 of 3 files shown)


Cellule:
I don't think we got any coverage for this scenario and I don't know if we care about it, but I figured I might as well fix them since I stumbled upon it

Cellule:
@MikeHolman Can you take a look at the missing constructors in WebAssembly new objects?

lib/Runtime/Language/InterpreterStackFrame.cpp
```diff
@@ -1163,8 +1163,6 @@ namespace Js
             !(this->executeFunction->GetHasTry() && (PHASE_OFF((Js::JITLoopBodyInTryCatchPhase), this->executeFunction) || this->executeFunction->GetHasFinally())) &&
             (this->executeFunction->ForceJITLoopBody() || this->executeFunction->IsJitLoopBodyPhaseEnabled()) &&
             !this->executeFunction->IsInDebugMode();
-#else
-        const bool doJITLoopBody = false;
```
dilijev: What happened here?

Cellule: This variable is unused in nojit

lib/Runtime/Library/WebAssemblyMemory.h
```diff
@@ -9,6 +9,9 @@ namespace Js
 {
     class WebAssemblyMemory : public DynamicObject
     {
+    protected:
+        DEFINE_VTABLE_CTOR( WebAssemblyMemory, DynamicObject );
+        DEFINE_MARSHAL_OBJECT_TO_SCRIPT_CONTEXT( WebAssemblyMemory );
```
MikeHolman: Is this related to legacy no jit? I talked with @boingoing and I think there is more to do here.

Cellule: When I tried to build nojit with vs2013 it complained that WebAssemblyMemory/Table didn't have a constructor.
I added this by looking at other javascript objects, but I don't know if it needs something else

MikeHolman: ok. I think this is fine to add, but we may still have more to do for marshalling to really work. We can talk to Curtis in January for that though.

dilijev:
Okay, sounds like this is good to merge?

Cellule:
Yes I just didn't have time before the weekend, thanks for reminding me

PR #1300, opened 2016-07-20
[Updated] #110 Convert rlexe.xml tags to use new build architecture and type terms 

RexSong:
Using this PR to replace old one (#991), due to there are so many changes after that change.

Fixes #110

PR #2239, opened 2016-12-18
(do not merge) [PR #1300 @RexSong] #110 Convert rlexe.xml tags to use new build architecture and type terms

### LINKED ISSUE #110: rlexe.xml tags: add support for new build architecture and type terms

dilijev:
Includes PR #1300 by @RexSong

Fixes #110


dilijev:
@RexSong Feel free to update in #1300. This PR was just meant to run CI when runtests.py has the execute bit set. I may also use this PR to add any additional changes that I would need to make after your PR is updated.

PR #2243, opened 2016-12-19
swb: fixes from stressing Basics and Strings tests

jianchun:
Lower.cpp: needs WB from JIT set string item slots.

Added bunch of missing annotations that may contain GC pointers.

CompoundString: block buffer may contain chars or pointers. Ammended to
handle pointers correctly.

### DIFF
--- a/lib/Backend/Lower.cpp
+++ b/lib/Backend/Lower.cpp
@@ -22435,7 +22435,8 @@ Lowerer::LowerSetConcatStrMultiItem(IR::Instr * instr)
     InsertAdd(false, dstLength, dstLength, srcLength, instr);
 
     dstOpnd->SetOffset(dstOpnd->GetOffset() * sizeof(Js::JavascriptString *) + Js::ConcatStringMulti::GetOffsetOfSlots());
-    this->m_lowererMD.ChangeToAssign(instr);
+
+    LowererMD::ChangeToWriteBarrierAssign(instr);
 }
 
 IR::RegOpnd *
--- a/lib/Runtime/Language/DynamicProfileInfo.cpp
+++ b/lib/Runtime/Language/DynamicProfileInfo.cpp
@@ -1921,7 +1921,7 @@ namespace Js
 
             if (callSiteInfoCount != 0)
             {
-                callSiteInfo = RecyclerNewArrayLeaf(recycler, CallSiteInfo, callSiteInfoCount);
+                callSiteInfo = RecyclerNewArray(recycler, CallSiteInfo, callSiteInfoCount);
                 if (!reader->ReadArray(callSiteInfo, callSiteInfoCount))
                 {
                     goto Error;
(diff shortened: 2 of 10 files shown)


jianchun:

lib/Runtime/Language/DynamicProfileInfo.cpp
```diff
@@ -1921,7 +1921,7 @@ namespace Js
 
             if (callSiteInfoCount != 0)
             {
-                callSiteInfo = RecyclerNewArrayLeaf(recycler, CallSiteInfo, callSiteInfoCount);
+                callSiteInfo = RecyclerNewArray(recycler, CallSiteInfo, callSiteInfoCount);
```
jianchun: CallSiteInfo contains a GC pointer in a union: polymorphicCallSiteInfo

lib/Runtime/Language/DynamicProfileInfo.h
```diff
@@ -84,26 +84,26 @@ namespace Js
...
     struct CallSiteInfo
     {
-        uint16 isArgConstant : 13;
-        uint16 isConstructorCall : 1;
-        uint16 dontInline : 1;
-        uint16 isPolymorphic : 1;
-        ValueType returnType;
-        InlineCacheIndex ldFldInlineCacheId;
-        union
+        Field(uint16) isArgConstant : 13;
```
digitalinfinity: The macros work with bit-lengths? Nice :)

digitalinfinity:
:shipit:

lib/Runtime/Language/DynamicProfileInfo.h
jianchun: Yes, it works because it expands to uint16 unchanged.

lib/Backend/Lower.cpp
```diff
@@ -22435,7 +22435,8 @@ Lowerer::LowerSetConcatStrMultiItem(IR::Instr * instr)
     InsertAdd(false, dstLength, dstLength, srcLength, instr);
 
     dstOpnd->SetOffset(dstOpnd->GetOffset() * sizeof(Js::JavascriptString *) + Js::ConcatStringMulti::GetOffsetOfSlots());
-    this->m_lowererMD.ChangeToAssign(instr);
+
+    LowererMD::ChangeToWriteBarrierAssign(instr);
```
jianchun: Yes, Lei's verification is awesome. Only constraint is that is it depends on TTT repro, and for some we can't catch a TTT.

lib/Runtime/Language/DynamicProfileInfo.cpp
leirocks: then it was a bug in master branch! we are just lucky and didn't hit a crash?

jianchun: After more look and discussion offline, I'll revert this change and add a comment. the Leaf here is safe because we explicitly save the only pointer elsewhere.

jianchun:
Thanks @digitalinfinity @leirocks

PR #2245, opened 2016-12-20
Proxy ownkeys trap needs marshal

leirocks:


kunalspathak:
Can you also double check other places in proxy code where marshalling might be needed?

leirocks:
yeah, I already spent some time to read the code to see if there's still missing marshalling, but can't find one. seems recent exprgen change is pretty good to find this kind of issue, let me checkin and see.

PR #2246, opened 2016-12-20
WASM - Custom sections

### LINKED ISSUE #2241: WASM - Custom Sections
Support https://github.com/WebAssembly/design/blob/master/JS.md#webassemblymodulecustomsections

Cellule:
Support WebAssembly.Module.customSections
Fixes #2241

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/2246)


Cellule:
ping

lib/Runtime/Library/WebAssemblyModule.cpp
```diff
@@ -772,6 +827,32 @@ uint WebAssemblyModule::GetGlobalsByteSize() const
...
+{
+    return m_customSections ? (uint32)m_customSections->Count() : 0;
+}
+
+Wasm::CustomSection
+WebAssemblyModule::GetCustomSection(uint32 index) const
+{
+    if (index >= GetCustomSectionCount())
+    {
+        throw Wasm::WasmCompilationException(_u("Custom section index out of bounds %u"), index);
```
MikeHolman: we shouldn't call with out of bounds length, and even if we did this exception kind would be invalid (it isn't a JS exception so it would go uncaught)

Cellule: It's true that it shouldn't happen, but lets say this api is used differently and we pass an out of bounds index, it should throw an exception we control. From what I can see we have the same kind of pattern in `WebAssemblyModule::EntryImports`.
Either we 
- don't use this api in `EntryImports` and `EntryCustomSections`
- Try/catch for `WasmCompilationException`
- remove the check
What do you think ?

PR #2249, opened 2016-12-20
AutoRestoreFunctionInfo does not handle throw during initialization

tcare:
Fixes OS 10106951.

Paul's recent change in #2149 misses a case where we throw during creation of the new function body. We then assert because we did not restore the original function body.

Fixed by detecting the case where the function has not been parsed and we did not finish creating the new function body, using the ParsableFunctionInfo to restore the original state.

### DIFF
--- a/lib/Runtime/Base/FunctionBody.cpp
+++ b/lib/Runtime/Base/FunctionBody.cpp
@@ -2138,24 +2138,24 @@ namespace Js
         public:
             AutoRestoreFunctionInfo(ParseableFunctionInfo *pfi) : pfi(pfi), funcBody(nullptr) {}
             ~AutoRestoreFunctionInfo() {
-                if (this->funcBody && this->funcBody->GetFunctionInfo()->GetFunctionProxy() == this->funcBody)
+                if (this->pfi != nullptr && this->pfi->GetFunctionInfo()->GetFunctionProxy() != this->pfi)
                 {
-                    FunctionInfo *functionInfo = funcBody->GetFunctionInfo();
+                    FunctionInfo *functionInfo = this->pfi->functionInfo;
                     functionInfo->SetAttributes(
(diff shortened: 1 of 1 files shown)


tcare:
Updated based on Paul's feedback. We can always get the FunctionBody from the ParsableFunctionInfo.

lib/Runtime/Base/FunctionBody.cpp
```diff
@@ -2138,24 +2138,24 @@ namespace Js
         public:
             AutoRestoreFunctionInfo(ParseableFunctionInfo *pfi) : pfi(pfi), funcBody(nullptr) {}
             ~AutoRestoreFunctionInfo() {
-                if (this->funcBody && this->funcBody->GetFunctionInfo()->GetFunctionProxy() == this->funcBody)
+                if (this->pfi != nullptr)
                 {
-                    FunctionInfo *functionInfo = funcBody->GetFunctionInfo();
+                    FunctionInfo *functionInfo = this->pfi->functionInfo;
```
pleath: Before we (re-)set these attributes, can we confirm that this->funcBody is non-null and that functionInfo->GetFunctionProxy()==this->funcBody? (We can still get functionInfo from this->pfi in all cases.)

tcare: In the repro, this->funcBody *will* be null because we haven't set it yet.

tcare: Knowing the above, would the following suffice:

if (this->pfi != nullptr && this->pfi->GetFunctionInfo()->GetFunctionProxy() == this->pfi)

tcare: Sorry that should have been !=. That works, though.

pleath:
:shipit:

PR #2257, opened 2016-12-20
WASM - misc fixes

Cellule:
Fix an error type when trying to convert int64 from Js value. https://github.com/WebAssembly/design/blob/master/JS.md#towebassemblyvalue

I don't know when it was changed, but there is no longer a concept of default namespace for imports https://github.com/WebAssembly/design/blob/master/JS.md#webassemblyinstance-constructor. The steps shows that we should always perform 2 GetProperty to get teh import

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/2257)


Cellule:
Rebased on latest release/1.4 and rerun tests since it was pretty old

PR #2264, opened 2016-12-21
silence failures for connecting to jit process

MikeHolman:
We are seeing failures here due to external causes (leading to double counting of issues). Silence this for now.


lib/Runtime/Language/SourceTextModuleRecord.cpp
```diff
@@ -866,7 +866,7 @@ namespace Js
             localSlotCount = currentSlotCount;
 
 #if ENABLE_NATIVE_CODEGEN
-            if (JITManager::GetJITManager()->IsOOPJITEnabled())
+            if (JITManager::GetJITManager()->IsOOPJITEnabled() && JITManager::GetJITManager()->IsConnected())
```
leirocks: IsConntected implicitly mean oop jit enabled?

PR #2265, opened 2016-12-21
Modified Array.prototype.toLocaleString to use the length of the internal slot for TypedArray.

atulkatti:
Modified Array.prototype.toLocaleString to use the length of the internal slot for TypedArray. Added tests for different types of typed arrays. Also, added a test to ensure that the normal array behavior has not changed after this change.

### DIFF
--- a/lib/Runtime/Library/JavascriptArray.cpp
+++ b/lib/Runtime/Library/JavascriptArray.cpp
@@ -7858,7 +7858,18 @@ Case0:
     template <typename T>
     JavascriptString* JavascriptArray::ToLocaleString(T* arr, ScriptContext* scriptContext)
     {
-        uint32 length = ItemTrace<T>::GetLength(arr, scriptContext);
+        uint32 length = 0;
+        if (TypedArrayBase::Is(arr))
+        {
+            // For a TypedArray use the actual length of the array.
+            length = TypedArrayBase::FromVar(arr)->GetLength();
+        }
+        else
+        {
+            //For anything else, use the "length" property if present.
+            length = ItemTrace<T>::GetLength(arr, scriptContext);
+        }
+
         if (length == 0 || scriptContext->CheckObject(arr))
         {
             return scriptContext->GetLibrary()->GetEmptyString();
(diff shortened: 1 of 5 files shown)


atulkatti:
@Microsoft/chakra-developers

test/UnitTestFramework/UnitTestFramework.js
```diff
@@ -1,357 +1,362 @@
-//-------------------------------------------------------------------------------------------------------
```
dilijev: Looks like this entire file was overwritten. Can you check your line-endings settings?

tcare:
Would you mind squashing your commits together when you update?

lib/Runtime/Library/JavascriptArray.cpp
```diff
@@ -7858,7 +7858,18 @@ namespace Js
     template <typename T>
     JavascriptString* JavascriptArray::ToLocaleString(T* arr, ScriptContext* scriptContext)
     {
-        uint32 length = ItemTrace<T>::GetLength(arr, scriptContext);
+        uint32 length = 0;
+        if (Js::TypedArrayBase::Is(arr))
+        {
+            // For a TypedArray use the actual length of the array.
+            length = (static_cast<Js::ArrayObject*>(arr))->GetLength();
```
tcare: Why do we need to cast this?

tcare:
Can we put a reference to the issue this fixed?

dilijev:
echo @tcare 
See: https://help.github.com/articles/closing-issues-via-commit-messages/

lib/Runtime/Library/JavascriptArray.cpp
atulkatti: The actual parameter arr has the type Js::RecyclableObject. We have to explicitly call the ArrayObject::GetLength to get the actual length of the array.

test/UnitTestFramework/UnitTestFramework.js
atulkatti: Thanks. Fixed.

lib/Runtime/Library/JavascriptArray.cpp
tcare: In this case, we can use TypedArrayBase::FromVar(arr). We use this convention to add a type assert.

PR #2266, opened 2016-12-22
ChakraCore NuGet Package Implementation

pre10der89:
Adding implementation for several nuget packages that allow the ChakraCore DLLs to be referenced in .NET projects... This main tasks in git issue #85...

A nuspec has been provided for the following flavors:

ARM, x86, x64 (combined)
ARM (only)
x86 (only)
x64 (only)
Symbols [PDBs] (only)

v140 (CPP only)


digitalinfinity:
@boingoing can you fix the copyright check script to handle this PR. @pre10der89, once we fix the copyright check script, you can rebase your PR on top of that change, and also update the ps1 file to change tabs to spaces.

Build/NuGet/.pack-version
```diff
@@ -0,0 +1 @@
+2.0.2-pre
```
dilijev: We might want to update this version number.

dilijev:
Hi @pre10der89, thanks so much for this!

Build/NuGet/package.ps1
```diff
@@ -0,0 +1,45 @@
+$root = (split-path -parent $MyInvocation.MyCommand.Definition) + '\..'
```
dilijev: Could you add the Microsoft copyright header to this file?

```ps1
```

dilijev:
@boingoing @digitalinfinity these are the only files that weren't excluded form the check (we already have an exclusion for .nuspec files):

```
18:15:03 Check Copyright > Checking Build/NuGet/.pack-version
18:15:03 Check Copyright > Checking Build/NuGet/_._
18:15:03 Check Copyright > Checking Build/NuGet/package.ps1
```

`.pack-version` is failing because it's non-empty and doesn't have the header. We should add an exclusion for that here: https://github.com/Microsoft/ChakraCore/blob/master/jenkins/check_copyright.sh#L33

`_._` is fine because the script explicitly passes empty files.

`package.ps1` will need the copyright header added: https://github.com/Microsoft/ChakraCore/pull/2266#discussion_r93702269

Build/NuGet/.pack-version
pre10der89: @dilijev I simply chose that value for internal testing... Is there a version you prefer?

pre10der89:
@dilijev I don't think we can add the copyright to the pack-version file since the packaging script (as is) reads the entire content of the file and uses that for the package versioning...  

The _._ file shouldn't have the copyright either because its meant to be blank...

I was able to add the copyright to the package.ps1

The use of the .pack-version and the package.ps1 is ultimately up to your build team... I  utilize the .pack-version so that I only had to change the version in one location instead of in every NuSpec... The package.ps1 does build all the different flavors of packages, but doesn't go much beyond that...

pre10der89:
@dilijev I have also signed the CLA twice... I see the CLA required tag is still there, however... Perhaps just a backlog...

dilijev:
@pre10der89 We'll look into that, thanks! /cc @tcare @digitalinfinity

jenkins/check_copyright.sh
```diff
@@ -31,6 +31,7 @@ git diff --name-only `git merge-base origin/master HEAD` HEAD |
     grep -v -E '\.filters$' |
     grep -v -E '\.targets$' |
     grep -v -E '\.nuspec$' |
+	grep -v -E '\.pack-version$' |
```
dilijev: Looks like you introduced a tab here ;P

Build/NuGet/.pack-version
dilijev: @digitalinfinity what do you think? IMO it's not necessary to change this for this PR. We will update it as needed.

pre10der89:
@dilijev Andrew Forster <anforste@bluejeansnet.com> is my primary email associated with my GitHub account

dilijev:
@pre10der89 Apparently it was a glitch with the GitHub API Hook. It's resolved now. :)

dilijev:
@pre10der89 Would you guys like us to release a package using these definitions for release/1.4? If so, could you retarget this pull request to release/1.4?

image

Then apply your changes onto release/1.4, and force-push that ref to the branch you used to open this PR.

pre10der89:
@dilijev I have changed the base branch to release/1.4 and force pushed... This created a merge conflict on one of the test files... I then rebased my local branch against master and pushed again... This seems to have cleared the conflicts and it appears the CI is running... 

Let me know if I've done this right :)

dilijev:
@pre10der89 Looks great! Thanks for your patience and indulging us on this adventure :)

PR #2267, opened 2016-12-22
Fix inconsistent type spec for Math.abs(INT32_MIN) in loops

ThomsonTan:
When there is Math.abs(INT32_MIN) in loop, we do int type spec on it in LoopPrePass, and const fold it after that. But const fold changes its dst type to float because Math.abs(INT32_MIN) is not valid int32 value. This inconsistency causes error if dst of Math.abs(INT32_MIN) flows back loop header then lossy conversion is needed on the float var. The lossy conversion violates the original assumption in loop header that the var is lossless.


ThomsonTan:

lib/Backend/GlobOpt.cpp
```diff
@@ -9611,8 +9621,10 @@ GlobOpt::OptConstFoldUnary(
             }
             else
             {
-                isInt = false;
-                fValue = -(FloatConstType)INT32_MIN;
+                // disable const fold for Math.abs(INT32_MIN) because it causes dst to be float type which
+                // could be different with previous type spec result in LoopPrePass. The latter does int
+                // type spec for Math.abs(INT32_MIN).
+                return false;
```
LouisLaf: Why not simply " throw Js::RejitException(RejitReason::AggressiveIntTypeSpecDisabled);" here?  We will bailout anyway...

ThomsonTan: RejitReason::AggressiveIntTypeSpecDisabled might be too aggressive just because we cannot do typespec on constant INT32_MIN?

---
In reply to: [94697935](https://github.com/Microsoft/ChakraCore/pull/2267#discussion_r94697935) [](ancestors = 94697935)

LouisLaf: But your change will cause us to type specialize the abs, which will bailout at runtime until we rejit and disable aggressive int type spec.  So it had the same negative cost, plus the bailouts/rejit.
The only other solution would be to not type specialize that case, but then you need to add the lowering to handle the unspecialized abs.  This is a corner case and it doesn't seem worth the cost.

ThomsonTan: You are right, the only advantage of continuing compilation is that if Math.abs(INT32_MIN) doesn't run in the jitted code, no bailout/rejit will be needed. I will switch to raise RejitException as simple workaround.

---
In reply to: [94718982](https://github.com/Microsoft/ChakraCore/pull/2267#discussion_r94718982) [](ancestors = 94718982)

ThomsonTan:
@LouisLaf any more comments?

lib/Backend/GlobOpt.cpp
```diff
@@ -8909,21 +8910,30 @@ GlobOpt::TypeSpecialization(
             {
                 return instr;
             }
+
+            // need to run typespec if cannot const fold InlineMathAbs, or dst will be dead
+            if (instr->m_opcode != Js::OpCode::InlineMathAbs)
+            {
+                runTypeSpec = false;
```
LouisLaf: You shouldn't need this code anymore right?

ThomsonTan: This is still necessary because user can specify `-off:constfold` to ignore constant folding so no JIT exception in that way. Still need to handle that.

ThomsonTan:
@LouisLaf any more comments?

lib/Backend/GlobOpt.cpp
LouisLaf: No, but we should bailout at runtime in this case.  There should be a bailout for this situation for when the source isn't a constant, but a variable equal to MinInt.

ThomsonTan: Yes, bailout handles run-time behavior, but the error here is compile time assertion. LoopPre pass does type specialize on dest of Math.abs(MIN_INT) to int32, and convert this dest to float when folding. Assertion is triggered when merging current block to successor loop header block which restores states on back edge, because it cannot convert float to lossless int32.

LouisLaf: That does not seem to be the problem at all.  Inlining of built-ins seems completely broken with -off:constfold.  Math.floor(1) is also broken.

All of this code:
        if (!this->IsLoopPrePass() &&
            !instr->IsBranchInstr() &&
            src1Val->GetValueInfo()->TryGetIntConstantValue(&intConstantValue) &&
            !(
                // Nothing to fold for element stores. Go into type specialization to see if they can at least be specialized.
                instr->m_opcode == Js::OpCode::StElemI_A ||
                instr->m_opcode == Js::OpCode::StElemI_A_Strict ||
                instr->m_opcode == Js::OpCode::StElemC ||
                instr->m_opcode == Js::OpCode::MultiBr ||
                instr->m_opcode == Js::OpCode::InlineArrayPop
            ))

Should really be in OptConstFoldUnary().  When OptConstFold returns false, we should go to the TypeSpecUnary() code.

---
In reply to: [99457671](https://github.com/Microsoft/ChakraCore/pull/2267#discussion_r99457671) [](ancestors = 99457671)

LouisLaf: Best thing to do right now would be to remove the useless code here, and then do the proper fix for -off:constfold in master.

---
In reply to: [99963378](https://github.com/Microsoft/ChakraCore/pull/2267#discussion_r99963378) [](ancestors = 99963378,99457671)

ThomsonTan:
It makes sense. I will remove the extra code and create a new issue for
-off:constfold

ThomsonTan:
@LouisLaf any more comments?

LouisLaf:
No, but you still have the unnecessary code in TypeSpec()...


---
In reply to: [278737239](https://github.com/Microsoft/ChakraCore/pull/2267#issuecomment-278737239) [](ancestors = 278737239)

ThomsonTan:
@LouisLaf, missed to push it to GitHub. Please help take a look again.

LouisLaf:
:shipit:

PR #2272, opened 2016-12-22
asmjs: anonymous function crash

obastemur:


obastemur:
Also added a test case

obastemur:

MikeHolman:
:shipit:

lib/Runtime/Language/AsmJsModule.cpp
```diff
@@ -1080,6 +1080,11 @@ namespace Js
     AsmJsFunc* AsmJsModuleCompiler::CreateNewFunctionEntry( ParseNode* pnodeFnc )
     {
         PropertyName name = ParserWrapper::FunctionName( pnodeFnc );
+        if ( !name )
```
ThomsonTan: >name [](start = 14, length = 4)

Do we need use a default name like 'anonymous function' for it?

Cellule: No, this should be invalid asm.js. Returning null here will invalidate the asm.js module and then be parsed as normal javascript

obastemur:
@Cellule @MikeHolman @ThomsonTan Thanks for the reviews!

PR #2273, opened 2016-12-22
Simple memory corruption fix

pleath:
ByteCodeGenerator::StartBindFunction uses a union in a way that allows a ParseableFunctionInfo to be accessed as though it were a FunctionBody. This led to memory corruption when redeferral changed the meaning of a flag that was meant to protect the accesses. Fixed by removing the union and the flag and using IsFunctionBody/GetFunctionBody to guard against illegal access.


pleath:
@dotnet-bot please test Windows x86_debug

pleath:
@rajatd, @digitalinfinity can either of you look at this one? (Thanks.)

PR #2274, opened 2016-12-22
Initialize ScriptFunction::computedNameVar in ctor

tcare:


lib/Runtime/Library/ScriptFunction.cpp
```diff
@@ -28,13 +28,13 @@ namespace Js
     ScriptFunction::ScriptFunction(DynamicType * type) :
         ScriptFunctionBase(type), environment((FrameDisplay*)&NullFrameDisplay),
         cachedScopeObj(nullptr), hasInlineCaches(false), hasSuperReference(false), homeObj(nullptr),
-        isActiveScript(false)
+        isActiveScript(false), computedNameVar(nullptr)
```
akroshg: Nit: could you move this initialization just after homeObj?

akroshg:
:shipit:

PR #2279, opened 2016-12-22
Fix sal warnings

meg-gupta:
Add annotation in CodeGenNumberAllocator.cpp to suppress sal warning.
Restructure code in Encoder.cpp, the two function calls were throwing off
the analyzer.

### DIFF
--- a/lib/Backend/CodeGenNumberAllocator.cpp
+++ b/lib/Backend/CodeGenNumberAllocator.cpp
@@ -324,6 +324,9 @@ Js::JavascriptNumber* XProcNumberPageSegmentImpl::AllocateNumber(Func* func, dou
                 Recycler::FillPadNoCheck(pLocalNumber, sizeof(Js::JavascriptNumber), sizeCat, false);
                 pLocalNumber = new (pLocalNumber) Js::JavascriptNumber(localNumber);
             }
+#else
+            Assert(sizeCat == sizeof(Js::JavascriptNumber));
+            __analysis_assume(sizeCat == sizeof(Js::JavascriptNumber));
 #endif
             // change vtable to the remote one
             *(void**)pLocalNumber = (void*)func->GetScriptContextInfo()->GetVTableAddress(VTableValue::VtableJavascriptNumber);
--- a/lib/Backend/Encoder.cpp
+++ b/lib/Backend/Encoder.cpp
@@ -295,8 +295,9 @@ Encoder::Encode()
     {
         if (m_func->IsOOPJIT())
         {
-            Js::ThrowMapEntry * throwMap = NativeCodeDataNewArrayNoFixup(m_func->GetNativeCodeDataAllocator(), Js::ThrowMapEntry, m_pragmaInstrToRecordMap->Count());
-            for (int32 i = 0; i < m_pragmaInstrToRecordMap->Count(); i++)
+            int allocSize = m_pragmaInstrToRecordMap->Count();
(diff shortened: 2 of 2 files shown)


meg-gupta:
@ThomsonTan, @leirocks can you please review ?

lib/Backend/Encoder.cpp
```diff
@@ -295,8 +295,9 @@ Encoder::Encode()
     {
         if (m_func->IsOOPJIT())
         {
-            Js::ThrowMapEntry * throwMap = NativeCodeDataNewArrayNoFixup(m_func->GetNativeCodeDataAllocator(), Js::ThrowMapEntry, m_pragmaInstrToRecordMap->Count());
-            for (int32 i = 0; i < m_pragmaInstrToRecordMap->Count(); i++)
```
ThomsonTan: >i < m_pragmaInstrToRecordMap->Count() [](start = 30, length = 37)

What's the problem in `i < m_progmaInstrToRecordMap`, type mismatch?

lib/Backend/Encoder.cpp
```diff
@@ -295,8 +295,9 @@ Encoder::Encode()
...
+            int allocSize = m_pragmaInstrToRecordMap->Count();
```
ThomsonTan: >int [](start = 12, length = 3)

Use `int32` to declare `allocSize` to be consistent?

lib/Backend/CodeGenNumberAllocator.cpp
```diff
@@ -324,6 +324,9 @@ Js::JavascriptNumber* XProcNumberPageSegmentImpl::AllocateNumber(Func* func, dou
                 Recycler::FillPadNoCheck(pLocalNumber, sizeof(Js::JavascriptNumber), sizeCat, false);
                 pLocalNumber = new (pLocalNumber) Js::JavascriptNumber(localNumber);
             }
+#else
+            Assert(sizeCat == sizeof(Js::JavascriptNumber));
+            __analysis_assume(sizeCat == sizeof(Js::JavascriptNumber));
```
ThomsonTan: >__analysis_assume(sizeCat == sizeof(Js::JavascriptNumber)); [](start = 12, length = 59)

Why only need this if `RECYCLER_MEMORY_VERIFY` is not defined? I don't see it is set when `RECYCLER_MEMORY_VERIFY` is defined.

lib/Backend/Encoder.cpp
meg-gupta: The calls to m_pragmaInstrToRecordMap->Count() for  AllocatorArray and loop counter is not recognized as equivalent by the analyzer.

lib/Backend/CodeGenNumberAllocator.cpp
leirocks: RECYCLER_MEMORY_VERIFY means it's checked build, which looks we don't run pre-fast with

meg-gupta: There is no warning when `RECYCLER_MEMORY_VERIFY` is defined,  `pLocalNumber` is allocated as : 
`pLocalNumber = (Js::JavascriptNumber*)alloca(sizeCat);`
So at `WriteProcessMemory(hProcess, (void*)number, pLocalNumber, sizeCat, NULL)` there will be no warning from the analyzer, saying `pLocalNumber` should be >= `sizeCat`.

Also looks like when `RECYCLER_MEMORY_VERIFY` is defined `sizeCat` can be larger than `sizeof(Js::JavascriptNumber)`.

lib/Backend/Encoder.cpp
meg-gupta: Done

ThomsonTan:
:shipit:

leirocks:
:shipit:

PR #2283, opened 2016-12-23
swb: avoid write barrier for global static data

jianchun:
Added WriteBarrierPtr(nullptr) constructor, no write barrier for nullptr.
Changed such initializers to use nullptr instead of 0 or NULL.

For all global static data that may contain WriteBarrierPtr fields, add
a constructor variant that takes a dummy parameter to indicate that
constructor is to initialize from static data. The corresponding
WriteBarrierPtr field uses a special variant to skip write barrier. This
avoids write barrier on static data address (which is in binary image
address, not recycler memory, and we have no cardTable for them on x64).

### DIFF
--- a/lib/Common/DataStructures/DoublyLinkedListElement.inl
+++ b/lib/Common/DataStructures/DoublyLinkedListElement.inl
@@ -7,7 +7,7 @@
 namespace JsUtil
 {
     template<class T, class TAllocator>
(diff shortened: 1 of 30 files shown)


jianchun:

jianchun:
@dotnet-bot test Ubuntu daily_ubuntu_linux_test_static please

PR #2285, opened 2016-12-23
fix virtual alloc upper limit

### LINKED ISSUE #2268: Assertion failure: "_cardTable" when memory address space is limited to 2GB
Save the following code as a python file, create an empty "empty.js" file, update the path of the chakracore binary and run it:

```
import os
import subprocess

def f1():
    import resource  # module only available on POSIX

    GB = 2**30  # Limit address space to 2GB
    resource.setrlimit(resource.RLIMIT_AS, (2*GB, 2*GB))

# Update the following line to your "ch" location
chakracoreShell = os.path.join(os.path.expanduser('~'), 'chakracore', 'BuildLinux', 'Debug', 'ch')
subprocess.check_output([chakracoreShell, 'empty.js'], preexec_fn=f1)
```
You will get the following assertion failure:
ASSERTION 79532: (/home/skywinlinvm/chakracore/lib/Common/Memory/RecyclerWriteBarrierManager.cpp, line 69) _cardTable
 Failure: (_cardTable)

The assertion failure still exists if the resource.RLIMIT_AS value is increased to 33*GB.

The assertion failure changes to "ERROR: ChakraRTInterface::JsCreateContext(runtime, &context) failed. JsErrorCode=0x20001 (JsErrorOutOfMemory)" if the resource.RLIMIT_AS value is increased to 34*GB.

The assertion failure goes away if the resource.RLIMIT_AS value is increased to 35*GB. Tested on Python 2.7.12 and Ubuntu Linux 16.04.1 LTS.

obastemur:
Default `lpMaximumApplicationAddress` can be beyond what is allowed for the process.
Get AS/RSS limit before sending an allocation request.

Also, crash early in case of allocation fails.

Fixes #2268

### DIFF
--- a/lib/Common/Memory/RecyclerWriteBarrierManager.cpp
+++ b/lib/Common/Memory/RecyclerWriteBarrierManager.cpp
@@ -226,10 +226,28 @@ X64WriteBarrierCardTableManager::Initialize()
         // On Win8, reserving 32 GB is fine since reservations don't incur a cost. On Win7, the cost
         // of a reservation can be approximated as 2KB per MB of reserved size. In our case, we take
         // an overhead of 96KB for our card table.
-        const unsigned __int64 maxUmProcessAddressSpace = (__int64) AutoSystemInfo::Data.lpMaximumApplicationAddress;
+
+        // xplat: GetRLimit AS / RSS for ``the maximum size of the process's virtual memory``
+        size_t memoryLimit;
+        if (!PlatformAgnostic::SystemInfo::GetMaxVirtualMemory(&memoryLimit))
+        {
+            memoryLimit = (size_t) AutoSystemInfo::Data.lpMaximumApplicationAddress; // try upper limit
+        }
+        else
+        {
+            // Safest option : Max RSS can be beyond what we can allocate, aim the smaller one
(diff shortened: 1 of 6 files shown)


obastemur:
Fixes #2268

lib/Common/Memory/RecyclerWriteBarrierManager.cpp
```diff
@@ -226,10 +226,29 @@ X64WriteBarrierCardTableManager::Initialize()
...
+        // xplat: GetRLimit AS / RSS for ``the maximum size of the process's virtual memory``
+        size_t memoryLimit;
+        if (!PlatformAgnostic::SystemInfo::GetMaxVirtualMemory(&memoryLimit))
+        {
+            memoryLimit = (size_t) AutoSystemInfo::Data.lpMaximumApplicationAddress; // try upper limit
+        }
+        else
+        {
+            // Safest option : Max RSS can be beyond what we can allocate, aim the smaller one
+            memoryLimit = memoryLimit > (size_t) AutoSystemInfo::Data.lpMaximumApplicationAddress ?
```
jianchun: >memoryLimit > (size_t) AutoSystemInfo::Data.lpMaximumApplicationAddress [](start = 26, length = 71)

might try use "min" template function?

jianchun:
:shipit:

obastemur:

PR #2288, opened 2016-12-27
Use WCHAR instead of wchar_t

obastemur:
ChakraCore uses char16_t on xplat and wchar_t on Windows for 2 bytes wide strings.
The reason we alter wchar_t to two bytes char16_t is that Unix wchar_t is 4 bytes long.

This PR removes the code that alters wchar_t and replaces all the instances of wchar_t with WCHAR.

Only ChakraCommonWindows.h is not updated.

### DIFF
--- a/bin/External/catch.hpp
+++ b/bin/External/catch.hpp
@@ -1571,8 +1571,8 @@ std::string toString( std::string const& value );
 std::string toString( std::wstring const& value );
 std::string toString( const char* const value );
 std::string toString( char* const value );
-std::string toString( const wchar_t* const value );
-std::string toString( wchar_t* const value );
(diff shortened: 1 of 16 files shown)


Penguinwizzard:
lgtm; the only small and fairly pedantic comment I have is that theoretically it'd be more predictable using uint16_t to define WCHAR in the two places where it's defined, as that aligns more narrowly with what windows' wchar_t is specified as. In both of the situations where that'd be an issue, though, there'd likely be innumerable other issues from the difference (not to mention windows itself might do things differently there), so doing a char16_t if possible and an unsigned short otherwise should be more than good enough.

obastemur:
@Penguinwizzard Thanks for the review!

PR #2292, opened 2016-12-28
android: fix va_* and cleanup pal + first output

obastemur:
andro-chakra

### DIFF
--- a/pal/inc/pal.h
+++ b/pal/inc/pal.h
@@ -65,6 +65,28 @@ namespace std {
 #endif
 #endif // __APPLE__ ?
 
+#ifdef __ANDROID__
+#define S_IREAD   0000400
+#define S_IWRITE  0000200
+#define S_IEXEC   0000100
+
+#ifndef CC_AND_TAG
+#define CC_AND_TAG "chakracore-log"
+#endif
+#include <android/log.h>
+#include <stdarg.h>
+#define PRINT_LOG(...) \
+    __android_log_print(ANDROID_LOG_INFO, CC_AND_TAG, __VA_ARGS__)
+#define PRINT_ERROR(...) \
+    __android_log_print(ANDROID_LOG_ERROR, CC_AND_TAG, __VA_ARGS__)
+#else
+typedef __builtin_va_list va_list;
+#define PRINT_LOG(...) \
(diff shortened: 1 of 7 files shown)


obastemur:

digitalinfinity:
Very cool! @EdMaurer FYI

pal/inc/pal.h
```diff
@@ -187,36 +209,12 @@ extern "C" {
 
 /******************* Compiler-specific glue *******************************/
 
-#ifndef _MSC_VER
 #define FEATURE_PAL_SXS 1
```
digitalinfinity: Why is this change needed?

pal/inc/pal.h
```diff
@@ -187,36 +209,12 @@ extern "C" {
...
-#endif // !_MSC_VER
-
-#if defined(_MSC_VER) || defined(__llvm__)
 #define DECLSPEC_ALIGN(x)   __declspec(align(x))
```
digitalinfinity: I'm confused- __declspec(align) is a VC++ concept, no? Wasn't the ifdef that was there correct?

pal/inc/pal.h
```diff
@@ -187,36 +209,12 @@ extern "C" {
...
-#else
-#define DECLSPEC_ALIGN(x)
-#endif
-
 #define DECLSPEC_NORETURN   PAL_NORETURN
-
-#ifndef _MSC_VER
 #define __assume(x) (void)0
```
digitalinfinity: I'm trying to understand the changes to this file- is the idea basically that this file should never be built with VC++? Otherwise, for instance, this change breaks prefast, no?

pal/inc/pal.h
obastemur: Cleanup.

pal/inc/pal.h
obastemur: We use either Clang or VC++.

pal/inc/pal.h
obastemur: > the idea basically that this file should never be built with VC++?

Yes

obastemur:
@digitalinfinity Thanks for the review!!

PR #2293, opened 2016-12-28
some write barrier fixes

leirocks:
- false poisitive on EntryPointInfo->jsMethod after native address reuse in recycler
- false poisitive in arena allocator cacheBlockEnd reused in leaf block in recycler (sharing same page allocator)
- overload ChangeToAssign in JIT code, check all posible assign for write barrier
- missing barriers in various location


leirocks:

leirocks:

leirocks:

leirocks:

leirocks:

leirocks:

leirocks:

leirocks:

leirocks:

leirocks:

leirocks:

leirocks:

leirocks:

leirocks:
@dotnet-bot test OSX daily_osx_osx_test_static please

leirocks:

leirocks:

lib/Runtime/Base/FunctionBody.cpp
```diff
@@ -9611,6 +9611,15 @@ namespace Js
 #endif
                 if (validationCookie == currentCookie)
                 {
+                    if (this->jsMethod == reinterpret_cast<Js::JavascriptMethod>(this->GetNativeAddress()))
```
jianchun: >if (this->jsMethod == reinterpret_cast<Js::JavascriptMethod>(this->GetNativeAddress())) [](start = 20, length = 87)

When "jsMethod != ...", what would "jsMethod" contain, could it still create other false positives? Is it ok to always do this without checking "jsMethod != ..."?

test/runtests.py
```diff
@@ -94,7 +94,7 @@
 binary = args.binary
 if binary == None:
     if sys.platform == 'win32':
-        binary = 'Build/VcBuild/bin/{}_{}/ch.exe'.format(arch, flavor)
+        binary = 'Build\\VcBuild\\bin\\{}_{}\\ch.exe'.format(arch, flavor)
```
jianchun: >Build\\VcBuild\\bin\\{}_{}\\ch.exe [](start = 18, length = 34)

Just curious, is this change necessary? I've been running it on Windows and "/" worked.

leirocks: if you copy the command line with /, ttt won't work...

lib/Runtime/Base/FunctionBody.h
```diff
@@ -2428,6 +2431,9 @@ namespace Js
         FieldWithBarrier(bool) m_hasBailoutInstrInJittedCode : 1; // Indicates whether function has bailout instructions. Valid only if hasDoneCodeGen is true
         FieldWithBarrier(bool) m_pendingLoopHeaderRelease : 1; // Indicates whether loop headers need to be released
         FieldWithBarrier(bool) hasExecutionDynamicProfileInfo : 1;
+#ifdef _M_X64_OR_ARM64
+        FieldWithBarrier(bool) m_tag : 1;                     // Used to tag the low bit to prevent possible GC false references
+#endif
```
jianchun: Why moving "m_tag"... is it because x64 difference with x86? If so would this make x86 worse?

Also this should be a build break on x86, as now "m_tag" only declared on x64 but initialized on both. (I'm reading in commit order, this could be fixed in later commit.)

bin/ch/ch.cpp
```diff
@@ -615,6 +615,7 @@ HRESULT ExecuteTest(const char* fileName)
 HRESULT ExecuteTestWithMemoryCheck(char* fileName)
 {
     HRESULT hr = E_FAIL;
+#ifdef _WIN32 // looks on linux it always leak ThreadContextTLSEntry since there's no DllMain
```
jianchun: Recall Oguz did some change in master, not sure if resolves this.

lib/Runtime/Base/FunctionBody.h
leirocks: yeah, in later commit, I carefully added tag for every 4 bytes. and this won't be build break either, for x86 m_tag is defined in another .h file :)

bin/ch/ch.cpp
leirocks: seems not, hit assertions in linux test

jianchun:
> 

Do you remove this because it overlaps with the other constructor? If so wondering if keeping this constructor makes more sense?

---
Refers to: lib/Runtime/Base/FunctionBody.h:3702 in 50b4d1a. [](commit_id = 50b4d1a56acc645506ef1182955f1d7b96b34d8e, deletion_comment = True)

leirocks:
> 

yes, it's a duplicate if global write barrier off. and I kept the plain one because many other places we are using this pattern and do the conversion inside function body

---
In reply to: [271458604](https://github.com/Microsoft/ChakraCore/pull/2293#issuecomment-271458604) [](ancestors = 271458604)

---
Refers to: lib/Runtime/Base/FunctionBody.h:3702 in 50b4d1a. [](commit_id = 50b4d1a56acc645506ef1182955f1d7b96b34d8e, deletion_comment = True)

lib/Runtime/SerializableFunctionFields.h
```diff
@@ -68,6 +68,7 @@ PRIVATE_FIELDS
     DECLARE_SERIALIZABLE_FIELD(ProfileId, profiledReturnTypeCount, UInt16);
     DECLARE_SERIALIZABLE_FIELD(ProfileId, profiledSlotCount, UInt16);
     DECLARE_SERIALIZABLE_ACCESSOR_FIELD(uint, LoopCount, RegSlot);
+    DECLARE_SERIALIZABLE_FIELD(bool, m_tag31, Bool); // Used to tag the low bit to prevent possible GC false references
```
jianchun: >DECLARE_SERIALIZABLE_FIELD(bool, m_tag31, Bool); [](start = 4, length = 48)

Are the "tag" fields being serialized? Is that the reason your change affected bc.h? If so can we avoid serializing the tag fields?

leirocks: yes, it will be serialized. it's one byte more space in the bytecode. if we choose to not serializing it but add tag here, we might need to change the serializing logic. I didn't spent time on investigate how much work to do though

---
In reply to: [95427632](https://github.com/Microsoft/ChakraCore/pull/2293#discussion_r95427632) [](ancestors = 95427632)

lib/Runtime/Library/JavascriptLibrary.cpp
```diff
@@ -6153,7 +6153,16 @@ namespace Js
...
+#if GLOBAL_ENABLE_WRITE_BARRIER
+        if (CONFIG_FLAG(ForceSoftwareWriteBarrier))
+        {
+            return RecyclerNewFinalized(this->GetRecycler(), JavascriptGenerator, generatorType, args, scriptFunction);
+        }
+        else
+#endif
+        {
+            return RecyclerNew(this->GetRecycler(), JavascriptGenerator, generatorType, args, scriptFunction);
+        }
```
jianchun: Consider moving this into a function JavascriptGenerator::New (and make constructor private) so that the logic at the same place.

leirocks: good idea, let me make the change in next commit. 

---
In reply to: [95430032](https://github.com/Microsoft/ChakraCore/pull/2293#discussion_r95430032) [](ancestors = 95430032)

lib/Runtime/SerializableFunctionFields.h
jianchun: Should be very easy to do. Add a new macro DECLARE_TAG_FIELD in this file which by default does nothing. In FunctionBody.h define it to declare a tag field. Serializer does not see it so no impact.

---
In reply to: [95429096](https://github.com/Microsoft/ChakraCore/pull/2293#discussion_r95429096) [](ancestors = 95429096,95427632)

jianchun:
:shipit:

PR #2296, opened 2016-12-29
Fix PAL double initialize

obastemur:


bin/ch/Helpers.cpp
```diff
@@ -725,3 +733,97 @@ void CALLBACK Helpers::TTFlushAndCloseStreamCallback(JsTTDStreamHandle handle, b
     fclose((FILE*)handle);
 }
 
+#define SET_BINARY_PATH_ERROR_MESSAGE(path, msg) \
+    str_len = (int) strlen(msg);                 \
+    memcpy(path, msg, (size_t)str_len);          \
+    path[str_len] = char(0)
+
+void GetBinaryLocation(char *path, const unsigned size)
```
ThomsonTan: >size [](start = 50, length = 4)

Specify type for size explicitly?

obastemur:
Adding additional [ Android / ARMv7 / iOS ] related commits straight to this PR. Fixes are related.

A full review on this PR is highly appreciated.

obastemur:
moved custom `trace` support to this PR and added Android LogCat output support.

Build/compile_clang.sh
```diff
@@ -0,0 +1,122 @@
+#!/bin/bash
+#-------------------------------------------------------------------------------------------------------
+# Copyright (C) Microsoft. All rights reserved.
+# Licensed under the MIT license. See LICENSE.txt file in the project root for full license information.
+#-------------------------------------------------------------------------------------------------------
+
+LLVM_VERSION="3.9.1"
+
+CC_URL="git://sourceware.org/git/binutils-gdb.git\nhttp://llvm.org/releases/${LLVM_VERSION}/\n"
```
obastemur: /cc @digitalinfinity

dilijev:

jianchun:
"Fix PAL double initialize" -- Could you document in comment or commit message in more details, how previously PAL double initializes and now does not? From the surface it looks we have 3+ places calling PAL_InitializeChakraCore().

lib/Runtime/PlatformAgnostic/Platform/Common/Trace.cpp
```diff
@@ -0,0 +1,320 @@
...
+static _Thread_local size_t LAST_TICK = -1;
+static bool IN_TRACE = 0;
+static unsigned long CALL_IN_COUNTER = 0;
+static char SELF_PATH[SELF_PATH_SIZE];
+static char FILE_BUFFER[FILE_BUFFER_SIZE];
+static int BUFFER_POS = 0;
+
+__attribute__((no_instrument_function))
+static void
+GetBinaryLocation()
```
jianchun: multiple copies of this function...

lib/Runtime/Language/Arguments.h
```diff
@@ -28,6 +28,9 @@ inline Js::Var* _get_va(void* addrOfReturnAddress, int n)
 {
     // All args are right after ReturnAddress by custom calling convention
     Js::Var* pArgs = reinterpret_cast<Js::Var*>(addrOfReturnAddress) + 1;
+#ifdef _ARM_
+    n += 2; // ip + fp
```
jianchun: Could you please document more details for this "+ 2"? Android ARM stores "ip + fp" in between return address and args?

lib/Runtime/PlatformAgnostic/Platform/Common/Trace.cpp
obastemur: Intentional. See comments, this file shouldn't take dependencies

jianchun: Yes, I was aware of that. Questionable though if it is worth duplicating code for that goal.

Also, is this only useful for diagnostics builds and can be excluded form "release"?

obastemur: It is already excluded for any configuration with no `--trace`. -> `#ifdef ENABLE_CC_XPLAT_TRACE`

obastemur:
>  Could you document in comment or commit message in more details, how previously PAL double initializes and now does not? From the surface it looks we have 3+ places calling PAL_InitializeChakraCore().

Previously we were calling 2 sep. methods and those were ending up calling the same final initializer multiple times. ATM, we have multiple entry points (Shared / CH / Static) and we trigger base initializer only once.

lib/Runtime/Language/Arguments.h
obastemur: Yes. Not just Android ARM, so `ifdef _ARM_`. Still work in progress here though.

bin/ch/Helpers.cpp
obastemur: We had discussed on this when it was implemented for the first time. It should be reachable from the `blame` part.

obastemur:
@ThomsonTan  @jianchun Thanks for the review

PR #2298, opened 2016-12-29
Remove downward lookup when searching shadowed property in prototype chain in for...in

ThomsonTan:
The previous fix of shadowing non-enumerable property for for...in (https://github.com/Microsoft/ChakraCore/pull/2175) added downward lookup in prototype chain which could degrade performance in edge cases. This fix followed suggestion from Curtis and only ignores lookup in prototype chain if there is no enumerable properties in the whole chain, otherwise search from the first prototype.

The perf data for which we optimized for in enumerator looks flat.


ThomsonTan:
@curtisman @rajatd please help review.

ThomsonTan:
@curtisman could you please help take a look?

PR #2300, opened 2016-12-30
Marshal value when traversing property keys via proxy

ThomsonTan:
We missed to marshal property key string when traversing (Object.keys()) an object via a proxy on it. Marshal is necessary because the object underlying the proxy could be from a difference script context.


ThomsonTan:
@rajatd please help review. A test will be added to Full.

rajatd:
:shipit:

PR #2301, opened 2016-12-30
Don't marshal symbol when it is just created from the same script context

ThomsonTan:
Marshal symbol in JavascriptObject::CreateKeysHelper is unnecessary because it is just created from the same script context.


ThomsonTan:
Pointed out by @rajatd. @rajatd @kunalspathak please help review.

rajatd:
Looks good to me.

kunalspathak:
LGTM

PR #2303, opened 2017-01-03
TypeSharing Bug fix

kunalspathak:
See: https://microsoft.visualstudio.com/DefaultCollection/OS/_workItems?id=10058868&triage=true&_a=edit

When we share type for ChangePrototype scenario, there is an instance where we could assign type with slotCapacity different than the layout of underline object.
For example, initially we cache the map of oldType to newCachedType (both didn’t have properties) in prototype object `P`. Later, newCachedType evolves when properties were added on it.
Also, oldType’s slotCapacity got shrunk when we update the ctor cache of an object.
In future, when we try to set same object `P` as prototype to another object, we notice that currentType aka oldType’s slotcapacity was shrunk. So we try to shrink the newCachedType’s slotCapacity
as well, but it doesn’t shrink to the extent oldType was shrunk. The reason being because newCachedType has successors added that were not present on oldType. Thus we end up assigning a type to an
object with different slotCapacity than object’s layout. Any further operations on this object can give undefined behavior.

Fix:
We try to shrink slotCapacity of newCachedType and after shrinking if it still doesn’t match oldType’s slotCapacity, we create newType and set it on object. We also update the cache with this newType.

### DIFF
--- a/lib/Runtime/Types/PathTypeHandler.cpp
+++ b/lib/Runtime/Types/PathTypeHandler.cpp
@@ -1552,6 +1552,16 @@ namespace Js
                     Assert(cachedDynamicTypeHandler->GetInlineSlotCapacity() >= roundedInlineSlotCapacity);
                     Assert(cachedDynamicTypeHandler->GetInlineSlotCapacity() >= GetPropertyCount());
                     cachedDynamicTypeHandler->ShrinkSlotAndInlineSlotCapacity();
+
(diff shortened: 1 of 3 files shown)


rajatd:
LGTM. Make sure to run appropriate perf tests.

kunalspathak:
Thanks @rajatd 
>  Make sure to run appropriate perf tests.

Ran perf test as well as used tracing to compare how many times we create new type. I don't see any regression.

PR #2304, opened 2017-01-03
Utf8Codex DecodeTail read past end of buffer

boingoing:
Decoding a Unicode byte-sequence can cause us to read extra bytes past the end of the source buffer if the sequence is truncated by the end of buffer immediately after the starting character.
    
    The read characters are unused as the unicode sequence will include a NULL character and, thus, be considered badly-formed resulting in the decoded character becoming the unknown character (0xFFFD).
    
    Bug is in DecodeInto, where we always add 4 to the source pointer and use that as a pointer to the end of the buffer. Because we always add 4, we always assume there are 4 more characters in the buffer, even when there are less.
    
    Fix is to calculate the real end of the source buffer and pass that down to DecodeTail which will allow us to avoid reading past the end of the buffer.
    
    Fixes https://microsoft.visualstudio.com/OS/_workitems?id=8989055&_a=edit

### DIFF
--- a/lib/Common/Codex/Utf8Codex.cpp
+++ b/lib/Common/Codex/Utf8Codex.cpp
@@ -397,7 +397,9 @@ LFastPath:
 LSlowPath:
         while (cch-- > 0)
         {
-            *buffer++ = Decode(ptr, ptr + 4, localOptions); // WARNING: Assume cch correct, suppress end-of-buffer checking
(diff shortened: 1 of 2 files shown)


boingoing:
@dotnet-bot test Ubuntu_linux_test_static please

boingoing:
@digitalinfinity, @LouisLaf This is the truncated Unicode escape-sequence buffer overflow. Would you mind taking a quick look?

PR #2289, opened 2016-12-28
Resolves #1391: Implement TypedArray updates to index handling.

atulkatti:
Resolves #1391: Implement TypedArray updates to index handling.
1. Also, made changes to ArrayBuffer and DataView as per the spec updates.
2. Added some more tests for DataView.

PR #2306, opened 2017-01-04
Resolves Bug#8971506. Defer parse case, assertions due to ByteCodeEmitter aggressively emitting nodes unprocessed yet by ByteCodeGenerator.

atulkatti:
In defer parse cases, sometimes byte code emitter seems to aggressively process nested scopes that the byte code generator hasn't visited. This results in assertion failures as certain flags are not set as expected. This surfaces more in cases where parameter scope has a class declaration and either the class declaration has an eval or the enclosing function is in an eval/global eval.

### DIFF
--- a/lib/Runtime/ByteCode/ByteCodeEmitter.cpp
+++ b/lib/Runtime/ByteCode/ByteCodeEmitter.cpp
@@ -3654,15 +3654,20 @@ void ByteCodeGenerator::EmitScopeList(ParseNode *pnode, ParseNode *breakOnBodySc
                     PushFuncInfo(_u("StartEmitFunction"), funcInfo);
                 }
 
-                if (paramScope && !paramScope->GetCanMergeWithBodyScope())
+                // In DeferParse scenario, ByteCodeGenerator would not have yet visited the nested
+                // scopes, so the emitter should also not process these nodes either.
+                if (pnode->sxFnc.pnodeBody)
                 {
-                    // Before emitting the body scoped functions let us switch the special scope slot to use the body ones
-                    pnode->sxFnc.funcInfo->UseInnerSpecialScopeSlots();
-                    this->EmitScopeList(pnode->sxFnc.pnodeBodyScope->sxBlock.pnodeScopes);
-                }
-                else
-                {
-                    this->EmitScopeList(pnode->sxFnc.pnodeScopes);
+                    if (paramScope && !paramScope->GetCanMergeWithBodyScope())
+                    {
+                        // Before emitting the body scoped functions let us switch the special scope slot to use the body ones
+                        pnode->sxFnc.funcInfo->UseInnerSpecialScopeSlots();
+                        this->EmitScopeList(pnode->sxFnc.pnodeBodyScope->sxBlock.pnodeScopes);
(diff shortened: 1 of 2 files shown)


lib/Runtime/ByteCode/ByteCodeEmitter.cpp
```diff
@@ -3654,20 +3654,25 @@ void ByteCodeGenerator::EmitScopeList(ParseNode *pnode, ParseNode *breakOnBodySc
...
+                    else
+                    {
+                        this->EmitScopeList(pnode->sxFnc.pnodeScopes);
+                    }
                 }
 
                 this->EmitOneFunction(pnode);
                 this->EndEmitFunction(pnode);
-
+
```
obastemur: extra space here.

atulkatti:
@Microsoft/chakra-developers

lib/Runtime/Library/TypedArray.cpp
```diff
@@ -509,12 +510,15 @@ namespace Js
             if (args.Info.Count > 3 && !JavascriptOperators::IsUndefinedObject(args[3]))
             {
                 mappedLength = ArrayBuffer::ToIndex(args[3], JSERR_InvalidTypedArrayLength, scriptContext, ArrayBuffer::MaxArrayBufferLength / elementSize, false);
+                newByteLength = mappedLength * elementSize;
 
-                if ((uint32)mappedLength > (byteLength - offset)/ elementSize)
+                if (offset + newByteLength > byteLength)
```
kunalspathak: Aren't these and many other changes in this PR part of #2289?

atulkatti: No. Please only look at the last commit with message "Resolves Bug#8971506...". The other commits are a result of a merge I did after pushing my branch. But those commits are already merged to release/1.4 branch. Sorry about the confusion.

aneeshdk:
LGTM

tcare:
Looks good

pleath:
The symptom you describe puzzles me. Can you may out an example of how this happens?

test/es6/default-splitscope.js
```diff
@@ -1175,6 +1175,32 @@ var tests = [
...
+        // Bug 8971506: This used to throw an ASSERT. The test is considered to be passing if no ASSERT is thrown.
+        (eval(`
+        function f6(jirfmx = class c3 {}) { };
+        `));
+
+        // Bug 8971506: This used to throw an ASSERT. The test is considered to be passing if no ASSERT is thrown.
+        (eval(`
+        function f7(a3 = class c4 extends false {
+        }) { };
+        `));
```
dilijev: Is the extra set of parens around the `eval(...)` necessary to set up the scopes correctly to repro the issue?

atulkatti: No, the extra set of parenthesis aren't necessary.

ThomsonTan:
Could the similar problem happen in other places?

atulkatti:
A more appropriate fix for this bug is to remove the isEnclosedInParamScope flag. Once that happens this fix will not be required.

PR #2307, opened 2017-01-04
Recover perf regression related to stack args

rajatd:
A perf regression had crept in with a refactor I did a while ago. 

We dont have block-level merging of arguments object syms in the globopt. So, as much as possible, we try to keep the set of arguments object syms same across blocks. To this end, we ignore tracking the arguments object alias introduced by ByteCodeArgoutCapture. To do this, the globopt expects the argout instructions to be immediately preceding the InlineBuiltInStart instr for .apply.

The refactor I did had broken this assumption. Reverting to that.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/2307)

### DIFF
--- a/lib/Backend/Inline.cpp
+++ b/lib/Backend/Inline.cpp
@@ -2442,6 +2442,14 @@ IR::Instr * Inline::InlineApplyWithArgumentsObject(IR::Instr * callInstr, IR::In
     IR::Instr* builtInStartInstr;
     InsertInlineeBuiltInStartEndTags(callInstr, 3, &builtInStartInstr); //3 args (implicit this + explicit this + arguments = 3)
 
+    // Move argouts close to call. Globopt expects this for arguments object tracking.
+    IR::Instr* argInsertInstr = builtInStartInstr;
+    builtInStartInstr->IterateArgInstrs([&](IR::Instr* argInstr) {
+        argInstr->Move(argInsertInstr);
+        argInsertInstr = argInstr;
+        return false;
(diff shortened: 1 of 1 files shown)


rajatd:
@satheeshravi could you review, please?

lib/Backend/Inline.cpp
```diff
@@ -2442,6 +2442,14 @@ IR::Instr * Inline::InlineApplyWithArgumentsObject(IR::Instr * callInstr, IR::In
     IR::Instr* builtInStartInstr;
     InsertInlineeBuiltInStartEndTags(callInstr, 3, &builtInStartInstr); //3 args (implicit this + explicit this + arguments = 3)
 
+    // Move argouts close to call. Globopt expects this for arguments object tracking.
```
ThomsonTan: >Move argouts close to call. Globopt expects this for arguments object tracking. [](start = 7, length = 79)

Is this causing the regression without OOP JIT?

rajatd: No, this has nothing to do with OOP jit. This got inadvertently introduced over 2 months ago.

satheeshravi:
:shipit:

satheeshravi:
Where was the regression noticed? And how much?

rajatd:
Regression was observed in jetstream\box2d,js; ~27%. However, I dont understand why we IB didn't show a regression in Octane box2d.

rajatd:

PR #2309, opened 2017-01-05
Fix TTD issues found during Node master merge

kfarnung:
* The RuntimeInfoTracker wasn't tracking the ArrayIteratorPrototype as part of the well known root objects.
* The comparison of doubles between two snapshots wasn't accounting for NaN values.

### DIFF
--- a/lib/Runtime/Debug/TTRuntmeInfoTracker.cpp
+++ b/lib/Runtime/Debug/TTRuntmeInfoTracker.cpp
@@ -697,6 +697,8 @@ namespace TTD
         this->EnqueueRootPathObject(_u("_stackTraceAccessor"), ctx->GetLibrary()->GetStackTraceAccessorFunction());
         this->EnqueueRootPathObject(_u("_throwTypeErrorRestrictedPropertyAccessor"), ctx->GetLibrary()->GetThrowTypeErrorRestrictedPropertyAccessorFunction());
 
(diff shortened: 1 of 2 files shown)


agarwal-sandeep:
Why only arrayIteratorPrototype? what about mapIteratorPrototype, setIteratorPrototype and many more?

kfarnung:
I suspect there are many others that will be an issue.  I'm working to unblock the current set of test cases and will investigate further with Mark to determine a systematic way to keep this list up-to-date.

kfarnung:
@mrkmarron Is there a better way that we can classify which objects need to be added to the map and which don't?

mrkmarron:
What this code is really doing is keeping track of objects that the runtime maintains via well known persistent pointers. Ideally we would like a "JsWellKnownObject" macro + list of fields that would notify us of what these fields/objects are (and give us stable names/ids for them). 

Unfortunately the declaration/creation is not in a single place and there is a large number of them (e.g. pretty much all of the RecyclableObject subtypes member fields in JavascriptLibraryBase etc.). Doing, this seemed a bit disruptive, error prone, and potentially easy to forget when adding a new field. So, the current approach is a stopgap until we have more test coverage (which we seem to be getting close) and were ready to commit to the code impact.

mrkmarron:
LGTM

PR #2310, opened 2017-01-05
Fix VSO 10259529: Add assertion to satisfy analysis of descriptors[descCount]

dilijev:
Fix VSO 10259529: Add assertion to satisfy analysis of descriptors[descCount] in JavascriptObject::DefinePropertiesHelperForGenericObjects.

Fix some OACR warnings about non-const locals in JavascriptObject.cpp since I was already fixing an OACR warning in the area.


ianwjhalliday:
LGTM

PR #2315, opened 2017-01-05
Remove InternalErrorException and FailFast instead

tcare:
There are a series of bugs we are getting where the dump has happened after we have thrown a InternalErrorException and returned E_FAIL. We don't even really need an InternalErrorException anyway, so we can FailFast in this case and get a more useful dump.


tcare:
@curtisman

lib/Common/Exceptions/Throw.h
```diff
@@ -137,12 +133,30 @@ namespace Js {
...
+    } \
+    else if (hr == VBSERR_OutOfStack) \
+    { \
+        JavascriptError::ThrowStackOverflowError(scriptContext); \
+    } \
+    else if (hr == E_ABORT) \
+    { \
+        throw Js::ScriptAbortException(); \
+    } \
+    else if (FAILED(hr)) \
```
curtisman: This need to be the inverse translate of what is caught and translated by END_TRANSLATE_KNOWN_EXCEPTION_TO_HRESULT.

PR #2322, opened 2017-01-05
Fix for an issue with redeferral and stack func init in jitted code

rajatd:
For initializing a nested function object on the stack, the jitted code was using hardcoded field values and offsets from the function's functionProxy. This is not legal with redeferral as the proxy could have been a full function body and could be redeferred from the time its enclosing function was jitted to when it was executed. Redeferral will throw away the existing function body and reallocate a parseable function info which will then be set as function proxy on the function info. This reallocation means that the hardcoded values and offsets in the jitted code are no longer valid.

Fix is to not use compile time values but to jit indirections. We already do that for oop jit, just made in-proc jit also do the same.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/2322)

### DIFF
--- a/lib/Backend/Lower.cpp
+++ b/lib/Backend/Lower.cpp
@@ -6579,73 +6579,38 @@ Lowerer::GenerateScriptFunctionInit(IR::RegOpnd * regOpnd, IR::Opnd * vtableAddr
     Js::FunctionInfoPtrPtr nestedInfo, IR::Opnd * envOpnd, IR::Instr * insertBeforeInstr, bool isZeroed)
 {
     Func * func = this->m_func;
-    IR::Opnd * functionProxyOpnd;
-    IR::Opnd * functionInfoOpnd = nullptr;
-    IR::Opnd * typeOpnd = nullptr;
-    bool doCheckTypeOpnd = true;
-    if (m_func->IsOOPJIT() || !CONFIG_FLAG(OOPJITMissingOpts) || (*nestedInfo)->IsDeferred())
-    {
-        functionInfoOpnd = IR::RegOpnd::New(TyMachPtr, func);
-        InsertMove(functionInfoOpnd, IR::MemRefOpnd::New(nestedInfo, TyMachPtr, func), insertBeforeInstr);
-        functionProxyOpnd = IR::RegOpnd::New(TyMachPtr, func);
-        InsertMove(functionProxyOpnd, IR::IndirOpnd::New(functionInfoOpnd->AsRegOpnd(), Js::FunctionInfo::GetOffsetOfFunctionProxy(), TyMachPtr, func), insertBeforeInstr);
-        typeOpnd = IR::RegOpnd::New(TyMachPtr, func);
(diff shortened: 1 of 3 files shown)


lib/Backend/Lower.cpp
```diff
@@ -6579,73 +6579,40 @@ Lowerer::GenerateScriptFunctionInit(IR::RegOpnd * regOpnd, IR::Opnd * vtableAddr
...
-        {
-            functionInfoOpnd = IR::IndirOpnd::New(functionProxyOpnd->AsRegOpnd(), Js::FunctionProxy::GetOffsetOfFunctionInfo(), TyMachReg, func);
-        }
-        else
-        {
-            functionInfoOpnd = IR::MemRefOpnd::New((BYTE*)functionProxyOpnd->AsAddrOpnd()->m_address + Js::FunctionProxy::GetOffsetOfFunctionInfo(), TyMachReg, func);
-        }
-    }
+    
+    functionInfoOpnd = IR::IndirOpnd::New(functionProxyOpnd->AsRegOpnd(), Js::FunctionProxy::GetOffsetOfFunctionInfo(), TyMachReg, func);
```
rajatd: @pleath - i was in two minds whether to directly use the nestedInfo for initializing functionInfo. I'm sure I have asked you this before, but do we expect functionInfo to be different from functionInfo->functionBodyImpl->functionInfo in some cases?

lib/Backend/Lower.cpp
```diff
@@ -6579,73 +6579,40 @@ Lowerer::GenerateScriptFunctionInit(IR::RegOpnd * regOpnd, IR::Opnd * vtableAddr
...
     GenerateMemInitNull(regOpnd, Js::ScriptFunction::GetOffsetOfObjectArray(), insertBeforeInstr, isZeroed);
     GenerateMemInit(regOpnd, Js::ScriptFunction::GetOffsetOfConstructorCache(),
-        LoadLibraryValueOpnd(insertBeforeInstr, LibraryValue::ValueConstructorCacheDefaultInstance),
+    LoadLibraryValueOpnd(insertBeforeInstr, LibraryValue::ValueConstructorCacheDefaultInstance),
         insertBeforeInstr, isZeroed);
-    if (!functionInfoOpnd)
-    {
-        if (functionProxyOpnd->IsRegOpnd())
-        {
-            functionInfoOpnd = IR::IndirOpnd::New(functionProxyOpnd->AsRegOpnd(), Js::FunctionProxy::GetOffsetOfFunctionInfo(), TyMachReg, func);
```
ThomsonTan: >functionInfoOpnd [](start = 12, length = 16)

I am also wondering why we need re-check functionInfo from functionProxy, should it be the same?

rajatd:
@pleath please review

lib/Backend/Lower.cpp
pleath: No. Never. Once we allocate a FunctionInfo for a given user function, it never changes. So, yes, once we've loaded it from the nested array we can just reuse it throughout the code sequence.

dilijev:

dilijev:
@dotnet-bot test this please

pleath:
Change checked in under another PR.

PR #2323, opened 2017-01-05
2 very simple redeferral fixes

pleath:
1. Now that the ScopeSlots structure uses a FunctionInfo*, make sure that the FunctionInfo points to a full FunctionBody before attempting to box a StackScriptFunction. 2. In addition to changing the type's entry point when we redefer, make sure we change a ScriptFunctionType's entry point info's entry point. This is necessary in the case of a cross-site call to a redeferred function.


dilijev:

Issue #2324, opened 2017-01-05
Collect profile information of array's length 


kunalspathak:
We should collect profile information of array's length created by user so we can use it to allocate array of appropriate size while JITing.

agarwal-sandeep:
This is a good idea. If this show up on websites we will revisit. Don't see a benefit in tracking this as an issue.

PR #2325, opened 2017-01-06
Allocate more elements for `new Array()`

kunalspathak:
With #1363, I changed the allocation buckets to allocate space for 2 elements if user calls `new Array()` or size passed to Array ctor is 0. However there can be cases where user creates array using `new Array()` but then set elements to indices beyond size 2. With my change, we will go through slow path to set these elements.
    
 As part of the fix, revert the `new Array()` case to allocate same amount of space before my #1363 changes. This doesn't change the `new Array(0) or (foo = 0; new Array(foo);` case and will continue to allocate space for just 2 elements.

I have opened #2324 to gather profile data of array's length which will help us specify the right length for such scenarios during array allocation in JIT.

### DIFF
--- a/lib/Backend/Lower.cpp
+++ b/lib/Backend/Lower.cpp
@@ -3901,7 +3901,7 @@ Lowerer::GenerateProfiledNewScArrayFastPath(IR::Instr *instr, Js::ArrayCallSiteI
         }
         GenerateArrayInfoIsNativeIntArrayTest(instr, arrayInfo, arrayInfoAddr, helperLabel);
         Assert(Js::JavascriptNativeIntArray::GetOffsetOfArrayFlags() + sizeof(uint16) == Js::JavascriptNativeIntArray::GetOffsetOfArrayCallSiteIndex());
-        headOpnd = GenerateArrayAlloc<Js::JavascriptNativeIntArray>(instr, &size, arrayInfo, &isZeroed);
+        headOpnd = GenerateArrayLiteralsAlloc<Js::JavascriptNativeIntArray>(instr, &size, arrayInfo, &isZeroed);
         const IR::AutoReuseOpnd autoReuseHeadOpnd(headOpnd, func);
 
         GenerateMemInit(dstOpnd, Js::JavascriptNativeIntArray::GetOffsetOfWeakFuncRef(), IR::AddrOpnd::New(weakFuncRef, IR::AddrOpndKindDynamicFunctionBodyWeakRef, m_func), instr, isZeroed);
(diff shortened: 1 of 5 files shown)


kunalspathak:
@satheeshravi , @rajatd - Please help review the changes.

lib/Backend/Lower.cpp
```diff
@@ -3993,7 +3993,22 @@ IR::JnHelperMethod GetArrayAllocMemHelper<Js::JavascriptNativeFloatArray>()
 
 template <typename ArrayType>
 IR::RegOpnd *
-Lowerer::GenerateArrayAlloc(IR::Instr *instr, uint32 * psize, Js::ArrayCallSiteInfo * arrayInfo, bool * pIsHeadSegmentZeroed, bool isArrayObjCtor /* = false */)
+Lowerer::GenerateArrayLiteralsAlloc(IR::Instr *instr, uint32 * psize, Js::ArrayCallSiteInfo * arrayInfo, bool * pIsHeadSegmentZeroed)
```
satheeshravi: >GenerateArrayLiteralsAlloc [](start = 9, length = 26)

ArrayLiterals means something else and i guess it doesn't suit this scenario.
Moreover, why do you need 2 wrappers ? All you need is one extra argument in the existing and last two arguments can be default params.

lib/Runtime/Library/JavascriptArray.cpp
```diff
@@ -1109,15 +1109,8 @@ namespace Js
 
         if (callInfo.Count < 2)
         {
-            if (pNew == nullptr)
```
satheeshravi: >if (pNew == nullptr) [](start = 12, length = 20)

what happens to the != nullptr cases ? You recreate new array everytime ? When will pNew is not nullptr ?

lib/Backend/Lower.cpp
kunalspathak: This method is used to generate memory allocation for array literals. I kept it this way for easy readability. I don't like having 2 defaults because then caller has to make sure that it is calling the method with right parameters.

---
In reply to: [95441552](https://github.com/Microsoft/ChakraCore/pull/2325#discussion_r95441552) [](ancestors = 95441552)

lib/Runtime/Library/JavascriptArray.cpp
kunalspathak: pNew can never be nullpre. If you read the code above, we declare it and start checking if it is nullptr. I think there was legitimate case where pNew was non-nullptr, and during refactoring, we removed that code. The code that have removed is dead code that can't happen atleast in current state of code.

---
In reply to: [95442434](https://github.com/Microsoft/ChakraCore/pull/2325#discussion_r95442434) [](ancestors = 95442434)

satheeshravi:
:shipit:

PR #2337, opened 2017-01-09
bug-fix: 9575461 - IsConcatSpreadable is called twice

obastemur:


lib/Runtime/Library/JavascriptArray.cpp
```diff
@@ -3019,8 +3023,8 @@ namespace Js
         for (uint idxArg = start; idxArg < args.Info.Count; idxArg++)
         {
             Var aItem = args[idxArg];
-            BOOL spreadable = false;
-            if (scriptContext->GetConfig()->IsES6IsConcatSpreadableEnabled())
+            BOOL spreadable = spreadableCheckedAndTrue;
```
akroshg: nit: change BOOL to bool as the spreadableCheckedAndTrue is bool?

akroshg:
looks good to me.

lib/Runtime/Library/JavascriptArray.cpp
obastemur: > change BOOL to bool as the spreadableCheckedAndTrue is bool?

Right! Thanks!

obastemur:
@akroshg Thanks for the review

PR #2341, opened 2017-01-09
Fix SAL annotations

meg-gupta:


meg-gupta:
@satheeshravi, @MikeHolman, @Penguinwizzard Can one of you review please ?

meg-gupta:
Ping

satheeshravi:
>         uint i = 0;

Should we even need this counter ? It checking if the SList is working fine.
I think you can consider removing the counter itself.

---
Refers to: lib/Backend/JITTimePolymorphicInlineCacheInfo.cpp:35 in 9347382. [](commit_id = 934738200ac24413e913649ed8db4d54c6bf9631, deletion_comment = False)

satheeshravi:
:shipit:

meg-gupta:
`i` is used to iterate the array inlineeInfoIDL, and the iterator is needed for the SList. Looks like the loop calls a function to populate the array elements from the list elements. Both kind of iterators are needed and SAL cant figure out they have the same upper bound.

@satheeshravi thanks for the review!

satheeshravi:
Oh yeah. I missed to see the index usage. Nvm.

---
In reply to: [271673423](https://github.com/Microsoft/ChakraCore/pull/2341#issuecomment-271673423) [](ancestors = 271673423)

Penguinwizzard:
:shipit:

PR #2343, opened 2017-01-09
Added assert to check deleting last property doesn't change the prototype chain

agarwal-sandeep:

### DIFF
--- a/lib/Runtime/Types/DynamicObject.cpp
+++ b/lib/Runtime/Types/DynamicObject.cpp
@@ -305,6 +305,8 @@ namespace Js
         Assert(!predecessorType->GetIsLocked() || predecessorType->GetTypeHandler()->GetIsLocked());
         Assert(!predecessorType->GetIsShared() || predecessorType->GetTypeHandler()->GetIsShared());
 
+        Assert(this->GetType()->GetPrototype() == predecessorType->GetPrototype());
+
         PathTypeHandlerBase* currentPathTypeHandler = (PathTypeHandlerBase*)this->GetTypeHandler();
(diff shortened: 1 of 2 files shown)


agarwal-sandeep:

pleath:
Out of curiosity, how do we wind up with a path type predecessor that has a different prototype from its successor?

agarwal-sandeep:
@pleath setting \_\_proto\_\_ can change the type so the predecessor prototype is different
```javascript
var obj1 = {};
obj1.o1 = 1;
var obj2 = {};
obj2.o2 = 1; // Set a property (last)
obj2.__proto__ = obj1; // Modify the prototype
delete obj2.o2; // Trigger delete last property
```

agarwal-sandeep:
@kunalspathak this case was found by Exprgen so it already exercise this code path and the DeleteStatementWeight is 0.7 which seems good enough.

pleath:
@kunalspathak is this how the optimization of __proto__-setting works? I thought you built an entirely new path in which all types had the new prototype. I didn't think we simply advanced along the same type path.

kunalspathak:
@pleath - Yeah, in `ChangePrototype` optimization I retain the oldtype on the entire chain and just set the new prototype on final evolved type from existing object (in this example. type obtained after adding property `o2`). We won't get benefit of delete last property optimization in this scenario, but is there any thing else that can functionally go wrong with this if we don't set the new prototype on the chain?

kunalspathak:
>  @kunalspathak this case was found by Exprgen
Right, but i think we found this after a while. So worth adding a snippet at least.

kunalspathak:
Spoke offline with @pleath  and we are on same page now. @agarwal-sandeep , we don't need this change. Can you just have an assert instead for `this->GetPredecessorType()->GetPrototype() == object->GetType()->GetPrototype()` ?

PR #2345, opened 2017-01-10
Fix leak in deferred function dictionary. 

pleath:
Now that the deferred function dictionary is leaf memory, we need to remove ParseableFunctionInfo's from it as they are undeferred.


pleath:
@kunalspathak Can you look at this one?

pleath:
RemoveFunctionBody only needs to happen when something is redeferred, and we're doing that currently.

PR #2349, opened 2017-01-10
OS9791203:Failfast in Tweetium and other UWA background task

suwc:
Multi-thread race condition causes inconsistency in chakra's DynamicSourceHolder vs. WWAHost's internal mapping count. Fix by removing call to dynamic source holder from backend JIT code.


suwc:
@MikeHolman @leirocks @Microsoft/chakra-developers could you take a look pls?

PR #2350, opened 2017-01-10
Revert the earlier TypedArray change as it may lead to integer overflow.

atulkatti:


atulkatti:
@Microsoft/chakra-developers

lib/Runtime/Library/TypedArray.cpp
```diff
@@ -510,15 +509,12 @@ namespace Js
             if (args.Info.Count > 3 && !JavascriptOperators::IsUndefinedObject(args[3]))
             {
                 mappedLength = ArrayBuffer::ToIndex(args[3], JSERR_InvalidTypedArrayLength, scriptContext, ArrayBuffer::MaxArrayBufferLength / elementSize, false);
-                newByteLength = mappedLength * elementSize;
 
-                if (offset + newByteLength > byteLength)
+                if ((uint32)mappedLength > (byteLength - offset)/ elementSize)
```
akroshg: In the above ToIndex call you could pass (byteLength - offset)/ elementSize instead of ArrayBuffer::MaxArrayBufferLength / elementSize. That should be same, right?

atulkatti: Yes. I will make that change.

tcare:
Looks good, please get Akrosh's signoff as well.

akroshg:
:shipit:

ljharb:
What was the earlier change? Can you please link to the PR number?

tcare:
@ljharb #2289

PR #2355, opened 2017-01-11
swb: trivial cleanup

jianchun:
Annotate "Arguments" fields with FieldNoBarrier, so that the plugin checker
is happy.

Revert a JavascriptGenerator change Lei accidentally removed (credit to
Lei).

One StackTrace test config seems incorrect. It contained 2 "compile-flags"
and the 2nd overrided 1st, making the test exactly the same as a previous
one.


jianchun:
@dotnet-bot test Linux tests please

Issue #2316, opened 2017-01-05
[Slow Tests, Legacy] UnifiedRegex failures in {x64,x86}_{debug,test} (branches: release/1.4, master)


dilijev:
https://github.com/Microsoft/ChakraCore/wiki/Build-Status#builds-with-slow-tests

https://ci.dot.net/job/Microsoft_ChakraCore/job/master/job/daily_slow_x86_debug/changes
Shows @boingoing's recent commits as the most likely source of the break.

Note: The Slow Tests issue is showing up in Legacy Tests.

dilijev:
Fixed by #2396 in `release/1.4` and `master`.

PR #2356, opened 2017-01-11
Update CaseInsensitive table from hybrid of (Unicode 6.3 and later, up to 8.0) to full Unicode 8.0 support. Fixes #517

### LINKED ISSUE #517: [RegExp] case-insensitive matching misses characters
When we generate the character casing equivalence sets for RegExp using [UnicodeData.txt](http://ftp.unicode.org/Public/UNIDATA/UnicodeData.txt), we start from lowercase letters and then add the characters that they map to. However, this misses non-letter characters. For example, '\u0345', which is in the "Mn" (Mark, Nonspacing) category, should match '\u0399' when the case-insensitive flag is passed (even without the "unicode" flag), but we fail to do so.

When this is fixed, the following test should pass:

``` JS
var didMatch = /\u0345/i.test('\u0399');
WScript.Echo(didMatch);
```

dilijev:
Update CaseInsensitive table from hybrid of (Unicode 6.3 and later, up to 8.0) to full Unicode 8.0 support.

Fixes #517

Note: The current standard wants Unicode 9.0 but it might be too risky to update that far in a stabilization branch. Opened #2367 to track this work item.

The table was generated in the past but then was (mostly) manually edited to include various optimizations and to fix bugs over the years. To make sure we got a complete update, I wrote a tool to generate the table.

## CaseInsensitive mapping generator tool
PR: https://github.com/dilijev/ChakraCore/pull/3
Source: https://github.com/dilijev/ChakraCore/tree/CaseInsensitive/tools/Unicode/CaseInsensitive

From this tool I was able to see and apply the differences from the current implementation to the correct implementation. In order to keep the change as small as possible, I used the diff as a reference for what needed changing and left out non-essential diffs. Additionally, the tool generates a suite of tests to track regressions against the update and ensure that the implementation does what is expected. I took some key tests from that suite and created the test file contained in this PR.

# Overview of Changes

I have staged the changes to hopefully make this easier to review. Here's an overview.

NOTE: The individual commits list or summarize the relevant lines of UnicodeData.txt where applicable.

First, I normalized the existing table to a reasonable format (same as the output of the tool) to make the later commits more clear. This involves fixing the casing and sorting deltas on each line in ascending order.

3d0f37f34aba206a796a14dad2f711bd8c53661a

Next, I fixed a few bugs with the current table that were preventing some cases from being matched correctly.

abb5d91947dc421fdaa7caed49a332a46d508f43
4894d244f6844bc8ec695bdecdea05cfb31c09e4
25049de90fba35afac8ef30a6e36cfb2f43e66e8

Added new codepoints:

f197902d952a89365240fef5660683ba2c95d3e6 - GREEK LETTER YOT
af2d08327b6355cf57a693b7af63dbcd6540f519 - Cyrillic
cba5439f9c1fe11c1ac2b995b1641c3c5e490981 - Cherokee
6c25a51d6a983ed3bd95c167d030161873d4161b - Latin extensions

Other tests and fixes:

cb736ab3e355c5e2d649e6dbf150b4046ab063fe - Add test cases from #517 to ensure those issues are fixed.
fbfb953f955171cc7a0f635624617cf80475102b - 0x0345 and case-insensitive equivalent characters with or without /u flag.
dc3e750344fcf5c10346aba3857b1db5f9961ac2 - Case-insensitive matching for Cherokee only with /u. [1]
d96eed51546aa6ebce6da71e8b689cc8796d503d - All other Unicode 8.0 cases of case-insensitive matching only with /u. [1] Added generated tests.

[1] These were with a focus on compat with v8 as determined by running the full regression test suite I generated against node-6.9.4-LTS and node-7.4.0 (latest), and double-checking a handful of tests against the latest stable Chrome (v 55).

# Test Coverage

* **Regex test run successful!** `Summary: E:\d\RegexTestCollateral had 151147 tests; 0 failures`
* Internal and slow tests pass.
* Note that PRs are merged with the target branch before running Jenkins checks so attempting to run slow tests on this PR would result in failures as per #2316 -- but running them locally on this branch, the tests pass.

# Reviewers

@tcare @bterlson @boingoing @Cellule - Thank you for your assistance with this change and for your code reviews.


kunalspathak:
What was the scenario that you are trying to fix? Is there need to add some test cases?

dilijev:
@kunalspathak - This is a normalization change that I just wanted to put up for review. Coming soon (before merging this) will be additional commits which fix #517 and add tests for the same.

dilijev:
/cc @bterlson @mathiasbynens - Care to take a look?

Also, any chance you could point me to where in the spec it mentions which case-insensitive matches should not be done without /u flag? For this PR I focused on compat with v8 for that decision.

mathiasbynens:
> Also, any chance you could point me to where in the spec it mentions which case-insensitive matches should not be done without `/u` flag?

See step 2 of https://tc39.github.io/ecma262/#sec-runtime-semantics-canonicalize-ch (“If Unicode is true, then … else …”).

I’ll take a closer look at this patch as soon as possible!

tcare:
For me it seems like reviewing the tool is just as important as the change in this PR, so I would encourage the other reviewers to do the same. Perhaps we need a separate PR for that?

dilijev:
@tcare I can submit a PR for that but since I don't plan on merging that in, at least not to `release/1.4`, maybe a PR against my fork would be better -- just to keep things organized. Also I'd say it's not high-enough quality to merge.

PR: https://github.com/dilijev/ChakraCore/pull/3
Source: https://github.com/dilijev/ChakraCore/tree/CaseInsensitive/tools/Unicode/CaseInsensitive

dilijev:
@mathiasbynens Thanks! Basically it seems the difference is without /u mapping under toUpperCase, and with /u, relying on CaseFolding.txt. In a lot of cases, the CaseFolding.txt file provides exactly the same mapping as UnicodeData.txt, but I guess there might be some differences in toUpperCase?

In the case of our table, we use the `MappingSource::UnicodeData` line to mark when we do the case mapping irrespective of /u flag (the simple toUpperCase case mappings), and `MappingSource::CaseFolding` to handle the cases where the CaseFolding mapping provided another alternative that should be considered equivalent.

Basically under `/iu` we use `CaseFolding` entries, then fall back on `UnicodeData` entries if an entry was not found for the code point in question marked as `MappingSource::CaseFolding`. Under `/i` we only use the `MappingSource::UnicodeData` entries.

I'm still surprised about the Cherokee letters, because the Cherokee letters are now split into uppercase and lowercase ranges since Unicode 8.0. It seems like the lowercase/uppercase pairs should still match considering there are a lot of multilingual plane code points for which there is a match without /u.

dilijev:
I actually generated much larger test files which test all of the equivalences captured in this table. I'll add those tests to this PR.

dilijev:
I might look into reducing the size of these test files to save space. Minification!

@curtisman Thinks that the tool which generates the table could be probably be checked in to ChakraCore in master (under the tools directory) after it is cleaned up a bit, although he hasn't had time to review it yet.

dilijev:
Didn't save many lines by minifying the tests, but saved about 100K of repo size.

tcare:
Looks good, I also had a quick look at the tool and a discussion with Doug offline. We have v8 compat achieved with the tests on this change so I am happy.

dilijev:
Thanks @tcare and @bterlson for the offline discussion and for your help as I worked on this fix!

dilijev:
@mathiasbynens I'm planning to merge this today since I'm going on vacation for 2 weeks and don't want to let this sit here without additional internal coverage. However, I would still greatly appreciate your feedback!

boingoing:
LGTM

dilijev:
Thanks @Cellule and @boingoing for your help reviewing!

mathiasbynens:
Belated LGTM, although I skipped the generated files in favor of the code that generates them, i.e. https://github.com/dilijev/ChakraCore/pull/3. 👍 Glad to see this merged!

Issue #2352, opened 2017-01-10
Intl should cache and re-use various formatters


kunalspathak:
[InitializeDateTimeFormat](https://github.com/Microsoft/ChakraCore/blob/master/lib/Runtime/Library/InJavascript/Intl.js#L1113) method takes an empty object dateTimeFormat and sets various properties on it
based on `options` and `localeList`. Inside `platform.createDateTime()`, it caches an instance of `dateTimeFormatter` as `hiddenObject` property on `dateTimeFormat` object.
If InitializeDateTimeFormat() is called from ctor of Intl.DateTimeFormat, then various methods like `Intl.DateTimeFormat.prototype.format()` called on the instance will reuse the `dateTimeFormatter` and other other cached properties on `dateTimeFormat`. 

However InitializeDateTimeFormat() is also called from non Intl methods like `Date.prototype.toLocale*String()`, in which case a new `dateTimeFormatter` is created every time and is thrown away. Below code can be written better, but is a good example to demonstrate the creation of `dateTimeFormatter` every time `toLocaleString` is called.

```js
  var utcDate = new Date(Date.UTC(96, 11, 1, 0, 0, 0));
  for(var i = 0;i < 10000; i++){
    print(typeof utcDate);
    print(utcDate.toLocaleString()); // dateTimeFormatter is created every time.
  }
```

Ideal solution would be to cache dateTimeFormat based on `localList` and `options` parameters, so we don't have to
1. Initialize `dateTimeFormat` object
2. Create dateTimeformatters

Above comment holds true for [InitializeNumberFormat()](https://github.com/Microsoft/ChakraCore/blob/master/lib/Runtime/Library/InJavascript/Intl.js#L661) and [InitializeCollator()](https://github.com/Microsoft/ChakraCore/blob/master/lib/Runtime/Library/InJavascript/Intl.js#L432) methods and APIs that consume these methods.

dilijev:
@jackhorton is this done?

jackhorton:
DateTimeFormat gets cached as of #4390 but Collator still isn't with no timeline on when that will happen.

PR #2358, opened 2017-01-11
Cache resolved locales in Intl.js

kunalspathak:
In order to resolve a locale, there is a cost to create date time formatter. We should simply cache the resolved locale for a given input locale. 

See : https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/10156769/

I verified https://www.flixbus.com/ and it no longer is unresponsive.

Here is the example where there is performance bottleneck and is slow.

```js
function sorter(a, b) {
    // here, every time we try to resolve locale
    return a.localeCompare(b, 'en');
  }

  // imagine this array contains lot of elements
  var arr = ['a', 'b', 'c', 'd', 'e'];
  arr.sort(sorter);  
```

There is more room for improvement as mentioned in #2352.


kunalspathak:
@curtisman - Can you please review?

kunalspathak:
@EdMaurer - FYI.

lib/Runtime/Library/InJavascript/Intl.js
```diff
@@ -81,13 +87,26 @@
     };
 
     var resolveLocaleBestFit = function (locale, defaultLocale) {
-        var resolvedLocale = platform.resolveLocaleBestFit(locale);
+        var resolvedLocale = platform.localeBestFitCache.get(locale);
+        if (resolvedLocale === undefined) {
+            resolvedLocale = platform.resolveLocaleBestFit(locale);
```
agarwal-sandeep: >resolvedLocale [](start = 12, length = 14)

This can be undefined. Before adding to map should we check that typeof is string?

kunalspathak: Good point. I think we should also cache the fact that we got `undefined` for a specific `locale` parameter so we don't try to resolve it again. I will update the code.

lib/Runtime/Library/InJavascript/Intl.js
```diff
@@ -7,11 +7,18 @@
...
+    }
+    if (platform.localeBestFitCache === undefined) {
+        platform.localeBestFitCache = new Map();
+    }
 
     // Built-Ins
     var setPrototype = platform.builtInSetPrototype;
     var getArrayLength = platform.builtInGetArrayLength;
     var callInstanceFunc = platform.builtInCallInstanceFunction;
+    const notFound = "Not found";
```
dilijev: nit: maybe call this const `NOT_FOUND`?

lib/Runtime/Library/InJavascript/Intl.js
```diff
@@ -1166,7 +1201,7 @@
                 // Deal with the locale
                 localeList = CanonicalizeLocaleList(localeList);
                 var resolvedLocaleInfo = resolveLocales(localeList, matcher, ["nu", "ca"], strippedDefaultLocale);
-
+
```
dilijev: nit: whitespace

kunalspathak:
@dotnet-bot please test Ubuntu ubuntu_linux_test_static

PR #2363, opened 2017-01-11
Change policy on root removal

digitalinfinity:
Currently, if RootRelease is called on a pinned object, we indicate that we have a candidate for an exhaustive GC. However, we don't do this for other sources of root removal- this change expands the RootRelease behavior to all cases of explicit root unpinning, since any such unpinning indicates a candidate for an exhaustive GC.


digitalinfinity:
@curtisman can you review please?

PR #2364, opened 2017-01-12
swb: a few fixes (module namespace, Debugger)

jianchun:
### swb: fix module namespace annotation

OP_StModuleSlot was changing slot without write barrier. Annotate all
related code.

(Credit to Lei for finding it is module namespace related.)

### Revert "fix a pal build issue on my machine"

This reverts commit c49e4af5d4242f07ed74c98b0afd9fcd1eb604a2.

It should now work after Oguz removed wchar_t redef from PAL.

### pin Debugger::m_context

Found under stress unit testing. Debugger::m_context was not pinned
and could be GC collected when not in use.

### DIFF
--- a/lib/Backend/IRBuilder.cpp
+++ b/lib/Backend/IRBuilder.cpp
@@ -3823,7 +3823,7 @@ IRBuilder::BuildElementSlotI2(Js::OpCode newOpcode, uint32 offset, Js::RegSlot r
         case Js::OpCode::LdModuleSlot:
         case Js::OpCode::StModuleSlot:
         {
-            Js::Var* moduleExportVarArrayAddr = Js::JavascriptOperators::OP_GetModuleExportSlotArrayAddress(slotId1, slotId2, m_func->GetScriptContextInfo());
+            Field(Js::Var)* moduleExportVarArrayAddr = Js::JavascriptOperators::OP_GetModuleExportSlotArrayAddress(slotId1, slotId2, m_func->GetScriptContextInfo());
             IR::AddrOpnd* addrOpnd = IR::AddrOpnd::New(moduleExportVarArrayAddr, IR::AddrOpndKindConstantAddress, m_func, true);
             regOpnd = IR::RegOpnd::New(TyVar, m_func);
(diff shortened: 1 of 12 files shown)


jianchun:

lib/Backend/IRBuilder.cpp
```diff
@@ -3823,7 +3823,7 @@ IRBuilder::BuildElementSlotI2(Js::OpCode newOpcode, uint32 offset, Js::RegSlot r
         case Js::OpCode::LdModuleSlot:
         case Js::OpCode::StModuleSlot:
         {
-            Js::Var* moduleExportVarArrayAddr = Js::JavascriptOperators::OP_GetModuleExportSlotArrayAddress(slotId1, slotId2, m_func->GetScriptContextInfo());
+            Field(Js::Var)* moduleExportVarArrayAddr = Js::JavascriptOperators::OP_GetModuleExportSlotArrayAddress(slotId1, slotId2, m_func->GetScriptContextInfo());
```
leirocks: why do you need this change?

leirocks: ah, I see, you changed the return type. however I think we should avoid using Field() in backend code. may be create a version of OP_GetModuleExportSlotArrayAddress_Backend, or force cast to Js::Var*

agarwal-sandeep:
Debugger change to pin/unpin looks good to me

pal/src/cruntime/finite.cpp
```diff
@@ -106,8 +98,8 @@ _isnan(
 
 See MSDN doc
 --*/
-double
-__cdecl
+double 
+__cdecl
```
obastemur: nit: some empty spaces here and below

lib/Backend/IRBuilder.cpp
jianchun: Here it is a pointer use only, I hate cast if unnecessary...

pal/src/cruntime/finite.cpp
jianchun: Yes, I simply reverted to previous version. It contains spaces but would be identical and easier to merge back to master.

jianchun:
@leirocks @agarwal-sandeep @obastemur Thanks for CR!

PR #2371, opened 2017-01-12
Math::FitsInDWord bugfix

Cellule:
Fix Math::FitsInDWord(int64) to account for negative value. Also, values like 0x00F0000000 were invalid because (0x00F0000000 & 0xFFFFFFFF) == 0xF0000000 (int64)0xF0000000 == 0xFFFFFFFFF0000000

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/2371)


Cellule:
@LouisLaf @MikeHolman please review.

Cellule:
@Microsoft/chakra-developers could someone take a look at this please ?

suwc:
@Cellule shall we add some tests?

Cellule:
@suwc The repro I had was pretty big, I'll see if I can reduce it to make it a test

lib/Common/Common/MathUtil.h
```diff
@@ -44,17 +44,8 @@ class Math
...
-    {
-#if TARGET_32
-        CompileAssert(sizeof(size_t) == 4);
-        return FitsInDWord((int32) value);
-#elif TARGET_64
-        CompileAssert(sizeof(size_t) == 8);
-        return FitsInDWord((int64)value);
-#endif
-    }
+    static bool     FitsInDWord(size_t value) { return ((size_t)(signed int)(value & 0xFFFFFFFF) == value); }
```
obastemur: Curious; Why don't you compare `value` to `INT_MIN` and `INT_MAX` range instead of forcing to overflow (signed int)?

Cellule: This was the original implementation, I believe it requires fewer instructions this way than doing 2 signed comparisons.

lib/Common/Common/MathUtil.h
```diff
@@ -44,17 +44,8 @@ class Math
...
+    static bool     FitsInDWord(int64 value) { return FitsInDWord((size_t)value); }
```
LouisLaf: >static bool     FitsInDWord(int64 value) { return FitsInDWord((size_t)value); } [](start = 4, length = 79)

This doesn't make much sense on 32bits...

Cellule: humm right...
Should I reimplement it like the size_t version aka
`return ((int64)(signed int)(value & 0xFFFFFFFF) == value);`?

LouisLaf: Yes, that's what I'd do.

LouisLaf:
:shipit:

PR #2374, opened 2017-01-13
Proxy : fixing a case related to GetOwnProperty.

akroshg:
According to OridinarySet(9.1.9.1) if the proxy returns the existing reciever we need to apply few checks and return from there.
Let existingDescriptor be ? Receiver.[[GetOwnProperty]](P).
If existingDescriptor is not undefined, then
    If IsAccessorDescriptor(existingDescriptor) is true, return false.
    If existingDescriptor.[[Writable]] is false, return false.

Fixed the codebase now, which fixes one assert as well.

### DIFF
--- a/lib/Runtime/Library/JavascriptProxy.cpp
+++ b/lib/Runtime/Library/JavascriptProxy.cpp
@@ -645,6 +645,21 @@ namespace Js
         }
         else
         {
+            // ES2017 Spec'ed (9.1.9.1): 
+            // If existingDescriptor is not undefined, then
+            //    If IsAccessorDescriptor(existingDescriptor) is true, return false.
+            //    If existingDescriptor.[[Writable]] is false, return false.
+
+            if (proxyPropertyDescriptor.IsAccessorDescriptor())
+            {
+                return FALSE;
+            }
+
+            if (proxyPropertyDescriptor.WritableSpecified() && !proxyPropertyDescriptor.IsWritable())
+            {
+                return FALSE;
+            }
+
             proxyPropertyDescriptor.SetValue(value);
             proxyPropertyDescriptor.SetOriginal(nullptr);
             return Js::JavascriptOperators::DefineOwnPropertyDescriptor(this, propertyId, proxyPropertyDescriptor, true, scriptContext);
--- a/test/es6/proxybugs.js
+++ b/test/es6/proxybugs.js
@@ -37,6 +37,70 @@ var tests = [
                 assert.areEqual(target.x, 20, "target object should work as expected even after proxy object is added to map");
             });
         }
+    },
+    {
+        name: "Assertion validation : returning descriptor during getOwnPropertyDescriptor should not pollute the descriptor",
+        body() {
+            var target = {};
+            var handler = {};
+            var getOwnPropertyDescriptorCalled = false;
+            handler['defineProperty'] = function () {
+                assert.fail("This function will not be called as 'getOwnPropertyDescriptor' will add accessor");
+            };
+            
+            handler['getOwnPropertyDescriptor'] = function (t, property) {
+                getOwnPropertyDescriptorCalled = true;
+                Object.defineProperty(t, 'abc', { set: function () { } });
+                return Reflect.getOwnPropertyDescriptor(t, property);
+            };
+            
+            var proxy = new Proxy(target, handler);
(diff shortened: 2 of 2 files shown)


akroshg:
@kunalspathak please have a look.

test/es6/proxybugs.js
```diff
@@ -37,6 +37,24 @@ var tests = [
...
+        name: "Assertion validation : returning descriptor during getOwnPropertyDescriptor should not pollute the descriptor",
+        body() {
+            var target = {};
+            var handler = {};
+            handler['defineProperty'] = function () {
+                assert.fail("This function will not be called as 'getOwnPropertyDescriptor' will return a descriptor");
+            };
+            
+            handler['getOwnPropertyDescriptor'] = function (t, property) {
+                Object.defineProperty(t, 'abc', { set: function () { } });
```
kunalspathak: >defineProperty [](start = 23, length = 14)

can you add print statement here to make sure we are calling trap?

lib/Runtime/Library/JavascriptProxy.cpp
```diff
@@ -645,6 +645,21 @@ namespace Js
...
+            }
+
+            if (proxyPropertyDescriptor.WritableSpecified() && !proxyPropertyDescriptor.IsWritable())
+            {
+                return FALSE;
+            }
+
             proxyPropertyDescriptor.SetValue(value);
             proxyPropertyDescriptor.SetOriginal(nullptr);
             return Js::JavascriptOperators::DefineOwnPropertyDescriptor(this, propertyId, proxyPropertyDescriptor, true, scriptContext);
```
kunalspathak: >DefineOwnPropertyDescriptor [](start = 44, length = 27)

Is there existing test case that will make sure we call defineOwnPropertyDescriptor handler at this point?

lib/Runtime/Library/JavascriptProxy.cpp
```diff
@@ -645,6 +645,21 @@ namespace Js
...
+            if (proxyPropertyDescriptor.IsAccessorDescriptor())
+            {
+                return FALSE;
+            }
+
+            if (proxyPropertyDescriptor.WritableSpecified() && !proxyPropertyDescriptor.IsWritable())
+            {
+                return FALSE;
+            }
+
```
kunalspathak: Is there a way to add test case for return value FALSE?

test/es6/proxybugs.js
akroshg: No we are not calling trap. while we are setting the value we identified there is no trap but getOwnPropertyDescriptor adds that value so we return from there (that is the fix in here).

---
In reply to: [95927783](https://github.com/Microsoft/ChakraCore/pull/2374#discussion_r95927783) [](ancestors = 95927783)

lib/Runtime/Library/JavascriptProxy.cpp
akroshg: That is the test case added - the handler['defineProperty'] was called before my fix. After my fix it is not called. Is that what you are looking for?

---
In reply to: [95928059](https://github.com/Microsoft/ChakraCore/pull/2374#discussion_r95928059) [](ancestors = 95928059)

test/es6/proxybugs.js
kunalspathak: Right, what I am asking is to have print statement before `Object.defineProperty()` statement so in future we still know this test case calls into this handler which defineProperty on target. Makes sense?

---
In reply to: [96045829](https://github.com/Microsoft/ChakraCore/pull/2374#discussion_r96045829) [](ancestors = 96045829,95927783)

lib/Runtime/Library/JavascriptProxy.cpp
akroshg: There should be one I believe - but it will be hard to find. I can add a test here that would be simpler.

---
In reply to: [95927839](https://github.com/Microsoft/ChakraCore/pull/2374#discussion_r95927839) [](ancestors = 95927839)

test/es6/proxybugs.js
akroshg: Ok - will do that.

---
In reply to: [96047714](https://github.com/Microsoft/ChakraCore/pull/2374#discussion_r96047714) [](ancestors = 96047714,96045829,95927783)

PR #2377, opened 2017-01-13
add a null check to JsTTDNotifyContextDestroy

jianchun:
JsTTDNotifyContextDestroy may be called during shutdown / destroying
threadContext. ThreadContext::GetContextForCurrentThread() may return
nullptr.


agarwal-sandeep:
LGTM

jianchun:
Thanks @agarwal-sandeep

PR #2380, opened 2017-01-14
Fix debugging return value inspection and continue after exception scenarios

agarwal-sandeep:
During redeferal change a check was removed, this broke return value
inspection and continue after exception scenario for builtins and
external functions. Basically we for debugging we wrap non script functions
in a try catch for continue after exception.

If CSP policy says eval is disabled, we should call the right entrypoints during debugging


agarwal-sandeep:
@pleath @akroshg please take a look

lib/Runtime/Base/ScriptContext.h
```diff
@@ -806,6 +806,7 @@ namespace Js
         // this is to indicate the actual close is called once only.
         bool isScriptContextActuallyClosed;
         bool isFinalized;
+        bool isEvalRestrict;
```
akroshg: >isEvalRestrict [](start = 13, length = 14)

nit: isEvalRestricted

akroshg:
LGTM

---
In reply to: [274124377](https://github.com/Microsoft/ChakraCore/pull/2380#issuecomment-274124377) [](ancestors = 274124377)

akroshg:
:shipit:

PR #2382, opened 2017-01-16
QuickSort: improve perf. and move our custom qsort_s

### LINKED ISSUE #2074: swb: tune JsUtil::QuickSort implementation
We brought back custom `JsUtil::QuickSort` in swb branch to solve the problem that system qsort routine does not update write barrier when array element contains WriteBarrierPtr fields.

We need to fine tune `JsUtil::QuickSort` performance. Current implementation does not work well with already sorted or many equal items array.

Moreover, @obastemur said xplat `qsort_r` is not reliable on some platforms (#2069). We might prefer custom reliable implementation on xplat too.

obastemur:
Recently we had implemented our custom qsort_s under PAL. SWB uses QuickSort from `lib/Common/DataStructures` and the implementation there is a bit slow on particular cases. 

This PR, combines both under QuickSort and enables the disabled test case back.

Fixes #2074


obastemur:

obastemur:

obastemur:
@mmitche  the message below started to appear often on the CI

```
05:30:15 [1m/mnt/j/workspace/Microsoft_ChakraCore/master/ubuntu_linux_debug_static_prtest/lib/Common/CommonBasic.h:13:10: [0m[0;1;31mfatal error: [0m[1m'typeinfo' file not found[0m
05:30:15 #include <typeinfo>
05:30:15 [0;1;32m         ^
05:30:15 [0m1 error generated.
```

Can we filter - reset tests for this case too ?

lib/Common/Memory/RecyclerPointers.h
```diff
@@ -489,21 +489,26 @@ template <class Policy>
 struct _QuickSortImpl
 {
     template<class T, class Comparer>
-    static void Sort(T* arr, size_t count, const Comparer& comparer, void* context)
+    static void Sort(T* arr, size_t count, const Comparer& comparer, void* context, size_t elementSize = sizeof(T))
     {
+#ifndef _WIN32
```
leirocks: use GLOBAL_ENABLE_WRITE_BARRIER instead, and combine with ForceSoftwareWriteBarrier, so we can test this on windows?

lib/Common/DataStructures/QuickSort.h
```diff
@@ -6,45 +6,277 @@
...
+    };
+#endif
 
-        inline static void swap(T& x, T& y)
+    template <class Policy, class T, class Comparer>
+    class QuickSort
+    {
+    public:
+        #define CCQ_SWAP QuickSortSwap<Policy, T>::swap
+        static void Sort(T* base, size_t nmemb, size_t size, Comparer comparer, void* context)
```
jianchun: >static void Sort(T* base, size_t nmemb, size_t size, [](start = 8, length = 52)

Could you please clean up this further? Current implementation is ported from old non-typed one ("void* base") and only works if "T" is "char/byte" or "size == 1" because of code like "base + i * size". This signature has redundant info and is confusing. Either "T* base, size_t count" or "void* base, size_t count, size_t size". Element "size" should be sizeof(T).

lib/Common/Memory/RecyclerPointers.h
```diff
@@ -489,21 +489,26 @@ template <class Policy>
...
+        JsUtil::QuickSort<Policy, char, Comparer>::Sort((char*)arr, count, elementSize, comparer, context);
```
jianchun: >JsUtil::QuickSort<Policy, char, Comparer>::Sort((char*)arr, count, elementSize, comparer, context); [](start = 8, length = 99)

Related to other comment. This has type and should be able to call like "JsUtil::QuickSort<Policy, T, Comparer>::Sort(arr, count, comparer, context)", below WB instantiation could call like "JsUtil::QuickSort<_write_barrier_policy, T, Comparer>::Sort(arr, count, comparer, context)", without casting and tweaking elementSize.

lib/Common/DataStructures/QuickSort.h
obastemur: sizeof(T) wouldn't apply to `write_barrier_policy` and all the others uses sizeof(T)

lib/Common/Memory/RecyclerPointers.h
obastemur: On XPLAT, this implementation replaces qsort_s. Are we sure all the current and future usages of this will comply with what you ask? In other words, T could be the type that `=` operator of it wouldn't perform what we want. 

This implementation, once the type is no_write_barrier_policy, it copies the memory.

lib/Common/Memory/RecyclerPointers.h
```diff
@@ -515,13 +520,23 @@ void qsort_s(T* arr, size_t count, const Comparer& comparer, void* context)
...
+
+#ifndef _WIN32
+template<class T, class PolicyType = T, class Allocator = Recycler, class Comparer>
+void qsort_s(T* arr, size_t count, size_t size, const Comparer& comparer, void* context)
+{
+    typedef typename _ArrayItemWriteBarrierPolicy<PolicyType>::Policy ItemPolicy;
+    typedef typename AllocatorWriteBarrierPolicy<Allocator, ItemPolicy>::Policy Policy;
+    _QuickSortImpl<Policy>::Sort(arr, count, comparer, context, size);
+}
+#else
```
jianchun: Why add this? It is already typed and does not need "size". The existing version above already meets requirements, and with a different signature, it requires the user to double examine this call (make sure to use correct PolicyType -- default T might be incorrect). Please don't make this "xplat vs. Windows" difference.

lib/Common/DataStructures/QuickSort.h
jianchun: sizeof(T) can apply to write_barrier_policy, that was my ask, by doing that this API is more clean.

lib/Common/Memory/RecyclerPointers.h
obastemur: Once again, we use custom qsort on xplat. Instead of implementing qsort on 2 different places, combined all in here.

lib/Common/Memory/RecyclerPointers.h
jianchun: Yes, I see that... I was asking to cleanup the JsUtil::QuickSort API so both calls are clean and consistent without casting/tweaking elementSize. That doesn't conflict with current behavior.

lib/Common/Memory/RecyclerPointers.h
jianchun: I see... Instead of template, could you simply use standard qsort_s signature here? That makes its purpose clearer.

lib/Common/Memory/RecyclerPointers.h
obastemur: > I was asking to cleanup the JsUtil::QuickSort API so both calls are clean and consistent without casting/tweaking elementSize.

Casting is inevitable for non_write_barrier ones.

lib/Common/Memory/RecyclerPointers.h
jianchun: Also, your "#else/#endif" commented out a qsort_s overload whose purpose was to emit build time failure on a misuse.

lib/Common/DataStructures/QuickSort.h
obastemur: `write_barrier_policy` swap is a basic `=` operator. Also, the algorithm doesn't cast / iterate over swb enabled array. That's why  `sizeof(T) wouldn't apply to write_barrier_policy`. However I could iterate `T` type instead of `char` for other non_write_barrier_policy types. In this case, code would have to cast back and forward on swap.

jianchun:
:shipit:

obastemur:

PR #2383, opened 2017-01-16
Type confusion in JavascriptArray indexOf, lastIndexOf, slice and splice with -es6toprimitive.

atulkatti:
In cases where a well known symbol like toPrimitive is defined on any of the arguments of array methods, it is possible that the array type is changed during the execution of the method. If this happens the method needs to take that into account for selecting fast/slow paths of processing for that method.


atulkatti:
@Microsoft/chakra-developers

PR #2387, opened 2017-01-18
fix Utf8 encoder

### LINKED ISSUE #2386: Utf8 encoding fails to encode
Utf8 encoding fails for the test case below. This introduced to master recently.

```
function toHexCP(c, cp) {
    var hex = "0123456789abcdef";
    return String.fromCharCode(hex.charCodeAt((c >> (cp * 4)) & 0xf));
}

function toHex(str) {
    var result = "";
    for(var i = 0; i < str.length; i++) {
        var c = str.charCodeAt(i);
        for (var cp = 3; cp >= 0; cp--) {
            result += toHexCP(c, cp);
        }
    }
    return "0x" + result;
}

var hex_str = String.fromCharCode("0x0080");
var pattern = eval("/" + hex_str + "/");
if (toHex(hex_str) != toHex(pattern.source)) throw new Error("String encoding has failed?")

console.log("PASS");
```

obastemur:
Added test case for currently failing scenarios.

Fixes #2386

### DIFF
--- a/lib/Common/Codex/Utf8Codex.cpp
+++ b/lib/Common/Codex/Utf8Codex.cpp
@@ -377,30 +377,93 @@ LFourByte:
             return ptr;
     }
 
+    inline static size_t GetUtf8Length(LPCUTF8 ptr, size_t cch)
+    {
+        size_t n = 0;
+        for (size_t i = 0; i < cch; i++)
+        {
+            size_t len = 0;
+            char chr = ptr[n];
+
+            if ((chr & 0x80) == 0)
+            {
+                len = 1;
+            }
+            else if ((chr & 0xE0) == 0xC0)
+            {
+                len = 2;
+            }
+            else if ((chr & 0xF0) == 0xE0)
+            {
+                len = 3;
+            }
+            else if ((chr & 0xF8) == 0xF0)
+            {
+                len = 4;
+            }
+            else
+            {
+                len = 4; // len >= 3
+            }
+
+            n += len;
+        }
+        return n;
+    }
+
     void DecodeInto(__out_ecount_full(cch) char16 *buffer, LPCUTF8 ptr, size_t cch, DecodeOptions options)
     {
-        DecodeOptions localOptions = options;
-
-        if (!ShouldFastPath(ptr, buffer)) goto LSlowPath;
-
-LFastPath:
-        while (cch >= 4)
+        size_t n = 0;
+        for (size_t i = 0; i < cch; i++)
         {
-            uint32 bytes = *(uint32 *)ptr;
-            if ((bytes & 0x80808080) != 0) goto LSlowPath;
-            ((uint32 *)buffer)[0] = (bytes & 0x7F) | ((bytes << 8) & 0x7F0000);
-            ((uint32 *)buffer)[1] = ((bytes >> 16) & 0x7F) | ((bytes >> 8) & 0x7F0000);
-            ptr += 4;
-            buffer += 4;
-            cch -= 4;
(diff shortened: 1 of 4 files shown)


lib/Common/Codex/Utf8Codex.cpp
```diff
@@ -395,10 +394,11 @@ namespace utf8
             cch -= 4;
         }
 LSlowPath:
+        // WARNING: Assume cch correct, suppress end-of-buffer checking
+        // we might have failed to measure cch, +1 for odd length. sizeof(char16) == 2
+        LPCUTF8 end = ptr + cch + (cch % 2 ? 2 : 1);
         while (cch-- > 0)
         {
-            LPCUTF8 end = ptr + cch + 1; // WARNING: Assume cch correct, suppress end-of-buffer checking
```
jianchun: cc @boingoing 

I wonder if both before and after change are incorrect. We don't know "end" at all based on "cch" (number of characters). The original code before Taylor's change makes some sense: "*buffer++ = Decode(ptr, ptr + 4, localOptions)", because a character can map up to 3 utf8 code units. Now either in-loop "ptr + cch + 1" (assume input cch == 1, end == ptr + 1, we can only decode 1 byte instead of up to 3) or out-of-loop "ptr + cch + 2/1" (assume input cch == 10, 20-byte utf8 stream, end == ptr + 12/11 instead of ptr + 20) could cut the input stream to much shorter and produce wrong result. How can this work?

test/utf8/rlexe.xml
```diff
@@ -20,4 +20,10 @@
       <baseline />
     </default>
   </test>
+  <test>
+    <default>
+      <files>bugGH2386.js</files>
+      <baseline />
```
jianchun: I recall to check "pass" output we should omit "baseline" tag. "<baseline/>" means ignore output.

lib/Common/Codex/Utf8Codex.cpp
obastemur: @jianchun this algorithm works differently. If you look at closer, it is almost the same with original one and this one saves reading 1 extra byte

obastemur: > I wonder if both before and after change are incorrect. 

I was looking into this with a random string generator. Looks like this algorithm is broken right from the beginning. Improving the test case and impl.

obastemur:
Updated the algorithm for decoder. /cc @boingoing @jianchun

obastemur:
Added some more tests.

lib/Common/Codex/Utf8Codex.cpp
jianchun: The original one was "*buffer++ = Decode(ptr, ptr + 4, localOptions)" and not broken... except that it ran into issues in some boundary case that Taylor had to fix.

obastemur: > The original one was "*buffer++ = Decode(ptr, ptr + 4, localOptions)" and not broken... except that it ran into issues in some boundary case that Taylor had to fix.

It is/was broken.

Lets keep the boundary cases, and fastPath problems aside.. `EncodedBytes` may perfectly return <1 or >4 and when it does...

jianchun: It depends on how this function was supposed to be used. The comment said assuming passed in "cch" is known and correct. This function might intend to be used in scenarios that we've already decoded the stream before and know for sure the character counts. @boingoing may know more context.

obastemur:

PR #2388, opened 2017-01-18
fix legacy builds

meg-gupta:
Fix build errors raised in legacy builds. Majority of the changes are related to "error C2899:  typename cannot be used outside a template declaration" due to declaration with type FieldWithBarrier outside template. This is an error with older compilers.

### DIFF
--- a/lib/Common/Memory/WriteBarrierMacros.h
+++ b/lib/Common/Memory/WriteBarrierMacros.h
@@ -16,12 +16,12 @@
 //
 
 #define FieldWithBarrier(type, ...) \
-    typename WriteBarrierFieldTypeTraits<type, ##__VA_ARGS__>::Type
+    WriteBarrierFieldTypeTraits<type, ##__VA_ARGS__>::Type
 
 #if GLOBAL_ENABLE_WRITE_BARRIER
 
 #define Field(type, ...) \
-    FieldWithBarrier(type, ##__VA_ARGS__)
+    typename FieldWithBarrier(type, ##__VA_ARGS__)
 #define FieldNoBarrier(type) \
     typename WriteBarrierFieldTypeTraits<type, _no_write_barrier_policy, _no_write_barrier_policy>::Type
 
--- a/lib/Runtime/Base/AuxPtrs.h
+++ b/lib/Runtime/Base/AuxPtrs.h
@@ -20,7 +20,7 @@ namespace Js
     {
         static const uint8 MaxCount;
         FieldWithBarrier(uint8) count;                 // always saving maxCount
-        FieldWithBarrier(FieldsEnum) type[_MaxCount];  // save instantiated pointer enum
+        typename FieldWithBarrier(FieldsEnum) type[_MaxCount];  // save instantiated pointer enum
         FieldWithBarrier(void*) ptr[_MaxCount];        // save instantiated pointer address
         AuxPtrsFix();
         AuxPtrsFix(AuxPtrsFix<FieldsEnum, 16>* ptr16); // called when promoting from AuxPtrs16 to AuxPtrs32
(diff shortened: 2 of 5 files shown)


meg-gupta:
@dotnet-bot test legacy

meg-gupta:

meg-gupta:

meg-gupta:

meg-gupta:
The test failures in legacy builds are outstanding issues in slow tests and unreleated to this change.

meg-gupta:
@leirocks , @jianchun , @aneeshdk Can you review this please ?

lib/Common/Memory/WriteBarrierMacros.h
```diff
@@ -17,6 +17,8 @@
 
 #define FieldWithBarrier(type, ...) \
     typename WriteBarrierFieldTypeTraits<type, ##__VA_ARGS__>::Type
+#define FieldWithBarrierT(type, ...) \
```
leirocks: only need to remove the typename for legacy compiler right? if so can you use _MSC_VER to differentiate it so you don't need to change FunctionBody.h?

meg-gupta: Do you mean replace  FieldWithBarrier with  FieldWithBarrierT for legacy ? For declarations within a template we still need  FieldWithBarrier .

leirocks: I meant this
```
#if !defined(__clang__) && _MSC_VER < xxxx
#define FieldWithBarrier(type, ...) \
      WriteBarrierFieldTypeTraits<type, ##__VA_ARGS__>::Type
#else
#define FieldWithBarrier(type, ...) \
      typename WriteBarrierFieldTypeTraits<type, ##__VA_ARGS__>::Type
#endif
```

meg-gupta: That will not work because there are some declarations with FieldWithBarrier inside a template and some not within a template. For declarations within a template we require 'typename' whereas for those not inside a template 'typename' should not be used with legacy compiler.

meg-gupta: So we will end up needing both 
`
#define FieldWithBarrier(type, ...) \
      WriteBarrierFieldTypeTraits<type, ##__VA_ARGS__>::Type
#define FieldWithBarrierT(type, ...) \
      typename WriteBarrierFieldTypeTraits<type, ##__VA_ARGS__>::Type`

for legacy builds.

leirocks: can you give an example that 'declarations within a template'? I think FieldWithBarrier is now only used in FunctionBody related structures, it's not broadly used.

meg-gupta: Declarations within template classes with dependent name need `typename`. 
```  template<class T>
    class InlineCachePointerArray
    {
    public:
        FieldWithBarrier(Field(T*)*) inlineCaches;
   ...
   }
```
Whereas the uses in `FunctionBody` structures should not have `typename`.

leirocks: I think we should explicitly adding a typename here for vs2013 then, like 
```
#if _MSC_VER<XXX
#define TYPENAME typename
#endif
        TYPENAME FieldWithBarrier(Field(T*)*) inlineCaches;
```

and @jianchun found we already macro _TYPENAME to handle similar case.

meg-gupta:

lib/Common/Memory/WriteBarrierMacros.h
meg-gupta: @leirocks Thanks, that reduces the changes required. I have changed the macro to not include `typename`. I have added `typename` explicitly for declarations with template dependent name. This is required by both legacy and non legacy.
Submitted a new commit consisting of the changes.

leirocks:
:shipit:

meg-gupta:
@dotnet-bot test  OSX osx_osx_debug_static  please

PR #2389, opened 2017-01-18
fix for swb to keep LeafBit in Finalizable Leaf allocation case

leirocks:
previously I removed LeafBit Finalizable Leaf allocation to avoid an compilation assertion. There's no functional issue - we just scan more objects and creating more false positives while VerifyMark()
fix to keep LeafBit and remove the unecessary workaround


jianchun:
:shipit:

PR #2390, opened 2017-01-18
Prototype on cached type chain

kunalspathak:
While investigating #2343, it was realized that when we set prototype, the
type chain that we cache doesn't have same prototype. All the predecessors
in the cached type chain before existing type of object still has oldPrototype
and only the existing type has new prototype.

Fix: Start with a type that has `newPrototype` object set because of which entire
chain will have `newPrototype` object.

### DIFF
--- a/lib/Runtime/Types/PathTypeHandler.cpp
+++ b/lib/Runtime/Types/PathTypeHandler.cpp
@@ -1229,6 +1229,9 @@ namespace Js
             {
                 // The new type isn't shared yet.  We will make it shared when the second instance attains it.
                 nextType = instance->DuplicateType();
+                // nextType's prototype and predecessorType's prototype can only be different here
+                // only for SetPrototype scenario where predecessorType is the cachedType with newPrototype
+                nextType->SetPrototype(predecessorType->GetPrototype());
                 nextType->typeHandler = nextPath;
                 markTypeAsShared ? nextType->SetIsLockedAndShared() : nextType->SetIsLocked();
             }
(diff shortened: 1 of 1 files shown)


kunalspathak:
@pleath  - Can you please review?

kunalspathak:
@dotnet-bot please test Ubuntu ubuntu_linux_release

agarwal-sandeep:
>             }

Can we add couple of test case covering if (isObjectLiteral) and else paths separately?

---
Refers to: lib/Runtime/Types/PathTypeHandler.cpp:1227 in 9210c36. [](commit_id = 9210c36a6c0945f794a17f260ac461c85ae0c07e, deletion_comment = False)

kunalspathak:
>             }

We already have existing test cases that covering it.

---
In reply to: [273591176](https://github.com/Microsoft/ChakraCore/pull/2390#issuecomment-273591176) [](ancestors = 273591176)

---
Refers to: lib/Runtime/Types/PathTypeHandler.cpp:1227 in 9210c36. [](commit_id = 9210c36a6c0945f794a17f260ac461c85ae0c07e, deletion_comment = False)

agarwal-sandeep:
>             }

How come none of them caught this issue or issue in PR#2343 before?

---
In reply to: [273592287](https://github.com/Microsoft/ChakraCore/pull/2390#issuecomment-273592287) [](ancestors = 273592287,273591176)

---
Refers to: lib/Runtime/Types/PathTypeHandler.cpp:1227 in 9210c36. [](commit_id = 9210c36a6c0945f794a17f260ac461c85ae0c07e, deletion_comment = False)

kunalspathak:
>             }

`isObjectLiteral` condition is not related to `SetPrototype`. It is only called when we promote type when we see `{a: 1, b: 2}`. I was mistakenly passing this template parameter as `true` and when i added test cases in #2102, this was caught and fixed. Additionally, the test case you added in #2343 should be sufficient to test the change.

---
In reply to: [273594150](https://github.com/Microsoft/ChakraCore/pull/2390#issuecomment-273594150) [](ancestors = 273594150,273592287,273591176)

---
Refers to: lib/Runtime/Types/PathTypeHandler.cpp:1227 in 9210c36. [](commit_id = 9210c36a6c0945f794a17f260ac461c85ae0c07e, deletion_comment = False)

pleath:
LGTM. Thanks.

PR #2391, opened 2017-01-18
WASM: memory import validation message

Cellule:
Change error type for invalid memory import and give more meaningful message.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/2391)

### DIFF
--- a/lib/Runtime/Library/WebAssemblyInstance.cpp
+++ b/lib/Runtime/Library/WebAssemblyInstance.cpp
@@ -328,9 +328,14 @@ void WebAssemblyInstance::LoadImports(
                     JavascriptError::ThrowWebAssemblyLinkError(ctx, WASMERR_NeedMemoryObject);
                 }
                 WebAssemblyMemory * mem = WebAssemblyMemory::FromVar(prop);
-                if (!wasmModule->IsValidMemoryImport(mem))
+
+                if (mem->GetInitialLength() < wasmModule->GetMemoryInitSize())
                 {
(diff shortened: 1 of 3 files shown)


lib/Runtime/Library/WebAssemblyInstance.cpp
```diff
@@ -328,9 +328,14 @@ void WebAssemblyInstance::LoadImports(
                     JavascriptError::ThrowWebAssemblyLinkError(ctx, WASMERR_NeedMemoryObject);
                 }
                 WebAssemblyMemory * mem = WebAssemblyMemory::FromVar(prop);
-                if (!wasmModule->IsValidMemoryImport(mem))
+
+                if (mem->GetInitialLength() < wasmModule->GetMemoryInitSize())
                 {
-                    JavascriptError::ThrowWebAssemblyLinkError(ctx, WASMERR_NeedMemoryObject);
+                    throw Wasm::WasmCompilationException(_u("Imported memory initial size (%u) is smaller than declared (%u)"), mem->GetInitialLength(), wasmModule->GetMemoryInitSize());
```
MikeHolman: we should avoid throwing WasmCompilationException here. Did you mean this is supposed to be WebAssemblyCompileError?

Cellule: `CreateInstance` has a try/catch for Compilation exception and converts them into LinkError, This allows more control over the error message.
I don't think we can avoid the try/catch as a lot of our module methods can throw.
The alternative would be to create another entry in rterrors.h, but since this particular error can really only happen here, I felt this made more sense

Cellule:
ping

PR #2393, opened 2017-01-19
WASM: returning -1 instead of throwing OOM

Krovatkin:
Rebased from **release/1.4** against **master**


Krovatkin:
@Cellule Rebased against master. Would you like me to add the tests mentioned in https://github.com/Microsoft/ChakraCore/pull/2137#issuecomment-267388954 ? 

fyi: @arunetm

Cellule:
@Krovatkin if you have simple tests that doesn't require fault injection sure

Krovatkin:
@Cellule I was thinking if we could maybe add `exclude_amd64` tag and, at least, run the tests on x32?

Cellule:
That could work, since there's no arch specific code in there, it would at least give us a little coverage

Krovatkin:
@Cellule There is a *little bit* of arch specific code, but as you pointed out the tests could "give us a little coverage." 
I'm a bit worried that `ArrayBuffer` constructor and `ArrayBuffer.transfer` produce two different errors when OOM. However, es7 doesn't explicitly state the exact type of the exception `ArrayBuffer.transfer` is required to throw.

Cellule:
Looks good, I am running some additional tests then will merge if everything is green.

PR #2395, opened 2017-01-20
Ensure scope slot is allocated for arguments symbol when captured by lambda

aneeshdk:
When a lambda captures arguments symbol, the symbol gets the scope slots allocated while it is emitted in the lambda function. But there can be a case where the function gets reparsed and the lambda does not get emitted again. So if the function has arguments then make sure scope slot is allocated each time.


aneeshdk:
@pleath @tcare @akroshg Can you please take a look?

tcare:
Looks good

PR #2396, opened 2017-01-20
Some string APIs operating on UTF-8 buffers do not property compute the length of the UTF-8 buffer in bytes

boingoing:
Many of our string APIs which take LPCUTF8 buffers do not know how many
bytes are in those buffers. Typically, these APIs need to walk over the
characters in the decoded string so they take in parameters for the
number of decoded characters. But the number of decoded characters is
not necessarily the same as the number of bytes. This is the case when
there are Unicode sequences encoded into the UTF-8 string.

As a consequence of this missing info, APIs such as utf8::CharsAreEqual
made the assumption that the UTF-8 buffer always contains at least 4
bytes - the maximum number of bytes in a Unicode sequence - without
knowing if that was actually true or not. It turned out that this was
not always true.

This change computes the correct byte length for the APIs which
previously made assumptions about the number of bytes in the buffer.

Fixes:
https://microsoft.visualstudio.com/OS/_workitems?id=10440534&_a=edit


lib/Common/Codex/Utf8Codex.cpp
```diff
@@ -377,7 +377,7 @@ namespace utf8
             return ptr;
     }
 
-    void DecodeInto(__out_ecount_full(cch) char16 *buffer, LPCUTF8 ptr, size_t cch, DecodeOptions options)
+    void DecodeInto(__out_ecount_full(cch) char16 *buffer, LPCUTF8 ptr, LPCUTF8 end, size_t cch, DecodeOptions options)
```
jianchun: Are these two functions now duplicates of "DecodeUnitsInto" and "DecodeUnitsIntoAndNullTerminate" below?

obastemur:
> As a consequence of this missing info, APIs such as utf8::CharsAreEqual
made the assumption that the UTF-8 buffer always contains at least 4
bytes - the maximum number of bytes in a Unicode sequence - without
knowing if that was actually true or not. It turned out that this was
not always true.

@boingoing `4` number there is not necessarily mean that `Decode` is going to use next 4 bytes or buffer should contain at least 4 bytes.

If are going to use the Codex methods with *Units and remove the other `cch` based functions. I'm all for it, otherwise `end` is good to have and maybe fixing a particular issue but the problem is still there.

lib/Common/Codex/Utf8Codex.cpp
boingoing: These two sets are closer but not duplicates. DecodeUnitsInto advances the source pointer and the fast path for reading characters is a little different.

jianchun: Maybe not worth maintaining 2 sets of functions. One set can be easily implemented by calling the other.

boingoing:
@obastemur Sorry, I meant to say that a previous fix had removed the assumption that the buffer always had 4 bytes in it but that the fix calculated end of the buffer incorrectly because it operated on cch (characters not bytes).

I didn't convert all our code to use the *Units versions because they advance the UTF-8 source pointer and have slightly different behavior. Was a little worried about differences in perf as well. I'll make an experiment to remove DecodeInto and DecodeIntoAndNullTerminate and see if there are some problems.

obastemur:
>  I'll make an experiment to remove DecodeInto and DecodeIntoAndNullTerminate and see if there are some problems.

Cool! Once you update this PR, if you prefer, you may take the test cases from my PR #2387 and make them part of this PR instead.

boingoing:
I updated this PR with the changes to remove DecodeInto and variants. There are still some API which operate on char counts in Utf8Codex but it was tricky enough to merge DecodeInto and DecodeUnitsInto that I didn't want to get crazy and go further right now.

> if you prefer, you may take the test cases from my PR #2387 and make them part of this PR instead
Added your unit tests to this PR, thank you.

test/utf8/surrogatepair.js
```diff
@@ -7,4 +7,9 @@
 // For this test case to work, please save this file with UTF-8 encoding
 var y = "function () { 'চ𐌲𐌿𐍄𐌹𐍃𐌺নির্বাচিত নিবন্ধ।' ;WScript.Echo('hello'); }"
 var x = function () { 'চ𐌲𐌿𐍄𐌹𐍃𐌺নির্বাচিত নিবন্ধ।' ;WScript.Echo('hello'); }
-WScript.Echo(x.toString() === y ? "PASS" : "FAIL");
\ No newline at end of file
+
+// 2 bytes 
+var y2 = "function () { '�� kugu' ;WScript.Echo('hello'); }"
```
obastemur: this and above looks like corrupted. If you have copy-pasted the contents, could you please move the file as is from my PR instead?

lib/Parser/HashFunc.cpp
```diff
@@ -16,31 +16,31 @@
  *  of the hash function so things don't go out of sync.
  */
 
-ULONG CaseSensitiveComputeHashCch(LPCOLESTR prgch, int32 cch)
+ULONG CaseSensitiveComputeHashCch(LPCOLESTR prgch, LPCOLESTR end, int32 cch)
```
obastemur: why keeping both `end` and `cch` ?

lib/Parser/Hash.h
```diff
@@ -404,34 +407,39 @@ class HashTbl
...
-    static bool CharsAreEqual(__in_z LPCOLESTR psz1, LPCUTF8 psz2, int32 cch2)
+    static bool CharsAreEqual(__in_z LPCOLESTR psz1, LPCUTF8 psz2, LPCUTF8 psz2end, int32 cch2)
     {
-        return utf8::CharsAreEqual(psz1, psz2, cch2, utf8::doAllowThreeByteSurrogates);
+        return utf8::CharsAreEqual(psz1, psz2, psz2end, cch2, utf8::doAllowThreeByteSurrogates);
     }
-    static bool CharsAreEqual(__in_z LPCOLESTR psz1, __in_ecount(cch2) char const * psz2, int32 cch2)
+    static bool CharsAreEqual(__in_z LPCOLESTR psz1, __in_ecount(cch2) char const * psz2, char const * psz2end, int32 cch2)
     {
+        Unused(psz2end);
```
obastemur: `psz2end != psz2` instead of sending cch? what if cch2 or psz2end-psz2 > psz1 length ?

lib/Parser/Hash.h
```diff
@@ -404,34 +407,39 @@ class HashTbl
...
+    static void CopyString(__in_ecount(cch + 1) LPOLESTR psz1, LPCUTF8 psz2, LPCUTF8 psz2end, int32 cch)
     {
-        utf8::DecodeIntoAndNullTerminate(psz1, psz2, cch);
+        Unused(cch);
+        utf8::DecodeUnitsIntoAndNullTerminate(psz1, psz2, psz2end);
     }
-    static void CopyString(__in_ecount(cch + 1) LPOLESTR psz1, __in_ecount(cch) char const * psz2, int32 cch)
+    static void CopyString(__in_ecount(cch + 1) LPOLESTR psz1, __in_ecount(cch) char const * psz2, char const * psz2end, int32 cch)
     {
+        Unused(psz2end);
```
obastemur: again, can we use psz2end instead of cch?

test/utf8/rlexe.xml
```diff
@@ -20,4 +20,17 @@
       <baseline />
```
obastemur: we need to remove `<baseline />` in order to make result from this test case effective

boingoing:
@obastemur Thanks for feedback, updated with a new commit. Can you take a look?

lib/Parser/Hash.h
```diff
@@ -351,14 +351,17 @@ class HashTbl
     }
 
     template <typename CharType>
+    IdentPtr PidHashNameLen(CharType const * psz, CharType const * end, uint32 cch);
+    template <typename CharType>
     IdentPtr PidHashNameLen(CharType const * psz, uint32 cch);
```
jianchun: A bit confused seeing these 2 overrides, one "end & cch" and one "cch" only...

jianchun:
:shipit:

dilijev:
Addresses #2316

PR #2399, opened 2017-01-20
fix: calling a non-static member function of a null object

obastemur:
previous impl. works mostly (as long as compiler plays nice with it..)

making it right.


jianchun:
:shipit:

obastemur:

PR #2400, opened 2017-01-21
new way to do the writebarrier verification

leirocks:
verify the barrier bits while marking, instead of doing it in VerifyMark, to catch issue earier and more accurate
put it under VerifyBarrierBit flag, so don't need to run with RecyclerVerifyMark
set the write barrier bit and verification bit before updating the pointer first, to avoid some race issue
fix the logic on recording the write barrier bit has ever been set and then cleared

avoid assertion a false positive on ScopeInfo->scope, this is arena address resuing and causing false positive. here's just avoiding the verification code to assertion on it, added todo to really fix the false positive

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -40,6 +40,7 @@ Build/.vs/
 Build/ipch/
 Build/swum-cache.txt
 Build/VCBuild.NoJIT/
+Build/VCBuild.SWB/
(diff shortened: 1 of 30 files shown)


jianchun:
:shipit:

PR #2405, opened 2017-01-24
xplat: fixes/improvements to PAL init and tracing

obastemur:


pal/src/memory/heap.cpp
```diff
@@ -150,6 +150,13 @@ HeapCreate(
 }
 
 
+__attribute__((no_instrument_function, noinline))
+static bool PAL_Initialize_Check_Once()
+{
+    int error = PAL_InitializeChakraCore();
+    return error == ERROR_SUCCESS;
+}
```
jianchun: Why is this code repeated in different files?

jianchun:
:shipit:

pal/src/memory/heap.cpp
obastemur: I didn't want this function is available externally or dependent to another binary or exec. order.

obastemur:

PR #2407, opened 2017-01-24
Bug fixes for Section API

MikeHolman:
We need process handle for ServerThreadContext destructor. In some cases this would be unavailable. Have changed to give each ServerThreadContext its own handle to keep lifetime management simpler.

We weren't correctly handling some cases where AllocLocal fails.

### DIFF
--- a/lib/JITServer/JITServer.cpp
+++ b/lib/JITServer/JITServer.cpp
@@ -123,17 +123,6 @@ ServerShutdown(
     return ShutdownCommon();
 }
 
-HRESULT
-ServerCleanupProcess(
-    /* [in] */ handle_t binding,
-    /* [in] */ intptr_t processHandle)
-{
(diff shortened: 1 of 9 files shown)


lib/JITServer/JITServer.cpp
```diff
@@ -174,6 +163,7 @@ ServerInitializeThreadContext(
     }
     catch (Js::OutOfMemoryException)
     {
+        CloseHandle((HANDLE)threadContextData->processHandle);
```
Cellule: What about other types of exception ? Are we guarantied only OOM can happen here ?

MikeHolman: Only thing that can throw here is the allocation, which should only throw OOM. Any other exception would be a bug (and will trigger a failfast in the content process).

PR #2409, opened 2017-01-25
Let redeferral work around block scope limitation of deferred parsing. 

pleath:
Detect cases where only block-scoping prevents upfront deferral. Also detect cases where it is safe to treat a lexically-scoped function as a var-scoped one. For functions that meet both criteria, use the var binding for it in place of the lexical binding. If at byte code gen time there are no lexical/with/catch scopes in the current chain, mark the function as a redeferral candidate.


pleath:
@aneeshdk, @boingoing Can you take a look? (Not actually a terribly complicated change.)

boingoing:
Looks good to me. 👍

PR #2410, opened 2017-01-25
Do not proactively set the stack nested func flag on recompiled functions. 

pleath:
If the function has been compiled previously, the stack nested func flag already has the correct value, and resetting it may ignore an escape that has since been discovered.


boingoing:
LGTM 👍

PR #2413, opened 2017-01-25
swb: some annotation fixes

jianchun:
Enhance diagnostics util class StackBackTraceRing a bit.

Update IDL_PADx to pass write barrier plugin checker.

GetParseableFunctionInfoRef() returns raw address of functionBodyImpl,
which is used and modified by GenerateByteCode() bypassing write barrier.
Removed this dangerous API and replaced with SetParseableFunctionInfo().

Enhanced m_constTable annotation and removed some casts.

Corrected a test file name case.

Tagged StackTraceLimitOOS test which times out when verifyMark on xplat,
because by default Linux uses 8MB stack (Windows 1MB), causing significant
more stack frames to mark and verify. Can use the tag to exclude them to
reduce noise.

Update the plugin to work with recent FieldWithBarrier macro change (
removing "typename" to support legacy compiler).

### DIFF
--- a/lib/Runtime/Library/SparseArraySegment.inl
+++ b/lib/Runtime/Library/SparseArraySegment.inl
@@ -241,7 +241,7 @@ namespace Js
             }
             for (i = (start + step-1)/step; i < (size/step); i++)
             {
-                ((Var*)(this->elements))[i] = fill;
+                ((Var*)(this->elements))[i] = fill; // swb: no write barrier, set to non-GC pointer
             }
             if ((i *= step) < size)
             {
@@ -257,7 +257,7 @@ namespace Js
 
             for (uint i = start; i < size * step; i++)
             {
-                ((Var*)(this->elements))[i] = fill;
+                ((Var*)(this->elements))[i] = fill; // swb: no write barrier, set to non-GC pointer
             }
         }
     }
--- a/lib/Common/Core/StackBackTrace.h
+++ b/lib/Common/Core/StackBackTrace.h
@@ -120,6 +120,22 @@ private:
 };
 
 //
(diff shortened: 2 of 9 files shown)


obastemur:
:shipit:

obastemur:
> Tagged StackTraceLimitOOS test which times out when verifyMark on xplat,
because by default Linux uses 8MB stack (Windows 1MB), causing significant
more stack frames to mark and verify. Can use the tag to exclude them to
reduce noise.

We might set/limit the stack on xplat for testing purposes?

lib/Runtime/Library/SparseArraySegment.inl
```diff
@@ -241,7 +241,7 @@ namespace Js
             }
             for (i = (start + step-1)/step; i < (size/step); i++)
             {
-                ((Var*)(this->elements))[i] = fill;
+                ((Var*)(this->elements))[i] = fill; // swb: no write barrier, set to non-GC pointer
```
leirocks: do we need to add a global NoWriteBarrierSet to do this? the casting looks not straightforward.

jianchun:
> We might set/limit the stack on xplat for testing purposes?

That should work too. Did not bother as mostly verifyMark with unit tests is private custom run. We can consider that in future if it becomes a pain.

lib/Runtime/Library/SparseArraySegment.inl
jianchun: We can... I remember I had that but deleted, worried it might be abused. We can think more about the best solution for cases like this. (However the cast here is mainly because "elements" are of "T[]" type, so it needs to cast to "Var[]" before assigning to "Var fill" value.)

PR #2414, opened 2017-01-26
Remove JsShutdownJITServer API

tcare:
The API is currently unused and has issues when the server never finished connecting.

### DIFF
--- a/lib/JITServer/JITServer.cpp
+++ b/lib/JITServer/JITServer.cpp
@@ -99,23 +99,6 @@ ShutdownCommon()
     return status;
 }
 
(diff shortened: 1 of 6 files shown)


lib/JITServer/JITServer.cpp
```diff
@@ -99,23 +99,6 @@ ShutdownCommon()
     return status;
 }
 
-__declspec(dllexport)
```
leirocks: remove it from ChakraCore.def as well?

tcare: Oops, missing a bunch of files. Fixing.

PR #2415, opened 2017-01-26
Fix bad binding in the presence of destructured arrow function parameters. 

pleath:
Do the same adjustment of block ID's in the case of parentheses in a destructured parameter expression that we do on entering a parenthetical expression.


pleath:
@aneeshdk Can you look at this one?

aneeshdk:
:shipit:

PR #2416, opened 2017-01-26
Ensure PropertyRecordList is set on new FunctionBody's. 

pleath:
An undeferral path could allow this to remain null.


akroshg:
LGTM

PR #2418, opened 2017-01-26
Modified parser to support basic type annotations

fcastellacci:


lib/Parser/Scan.cpp
```diff
@@ -1588,6 +1641,98 @@ tokens Scanner<EncodingPolicy>::ScanStringConstant(OLECHAR delim, EncodedCharPtr
...
+                    //+5 so we skip the space (change that)
+					p += 5;//FCASTE: don't like this code, revisit 
+                    switch ((ch = this->ReadFirst(p, last))) 
+                    {
+                    case '{':
+                        *pp = p;
+                        return tkTypeAnnBegin;
+                    }
+				}
+				Error(ERRsyntax);
```
MikeHolman: shouldn't throw an error (instead maybe just update current character and break?)

lib/Parser/ptree.h
```diff
@@ -600,9 +600,21 @@ struct PnFinally : PnStmt
     ParseNodePtr pnodeBody;
 };
 
+enum JsType : byte
```
MikeHolman: nit: maybe make an enum class? otherwise we pollute global namespace, then we can do JsType::Object, etc. instead of t_

lib/Parser/ptree.h
```diff
@@ -600,9 +600,21 @@ struct PnFinally : PnStmt
...
+	t_float,
+	t_bool,
+	t_array,
+	t_function
+};
+
 struct ParseNode
 {
     OpCode nop;
+    JsType typeHint; //Not sure if putting it here is the best idea
```
MikeHolman: i think it's ok to put in here. one nit is that we should move this def to be near bools, otherwise struct will use extra memory due to padding. usually not that important to worry about, but functions will have very many parse nodes, so a few bytes saved can add up

lib/Parser/ptree.h
```diff
@@ -600,9 +600,21 @@ struct PnFinally : PnStmt
     ParseNodePtr pnodeBody;
 };
 
+enum JsType : byte 
+{
+	t_object,
```
MikeHolman: I think we should add a Unknown/None type, to tell if we didn't get an annotation for this node

MikeHolman:
> void Parser::InitNode(OpCode nop,ParseNodePtr pnode) {

we don't zero memory for ParseNode allocations, so here we should initialize type field. otherwise when there is no annotation we will have some random value

---
Refers to: lib/Parser/Parse.cpp:670 in 05196f0. [](commit_id = 05196f095f49c703c70b4c07222988af678068a8, deletion_comment = False)

lib/Parser/Scan.cpp
fcastellacci: I added this to make it easier to find errors. I would rather change this after we do some testing, in order to know when a type annotation is poorly formatted or invalid.

MikeHolman: Fair enough. In that case, we should add a runtime flag and put the type annotations behind it. Otherwise we will break any code that has @ inside of a block comment (looks like one of our UTs is hitting this issue).

lib/Parser/Scan.cpp
```diff
@@ -162,6 +162,8 @@ Scanner<EncodingPolicy>::Scanner(Parser* parser, HashTbl *phtbl, Token *ptoken,
 
     m_fYieldIsKeyword = false;
     m_fAwaitIsKeyword = false;
+
+    m_typeAnnotationsOn = Js::Configuration::Global.flags.TypeAnnotations;
```
MikeHolman: use CONFIG_FLAG macro here

lib/Common/ConfigFlagsList.h
```diff
@@ -1362,6 +1362,7 @@ FLAGNR(Boolean, TraceAsyncDebugCalls  , "Trace calls to async debugging API (def
 #ifdef TRACK_DISPATCH
 FLAGNR(Boolean, TrackDispatch         , "Save stack traces of where JavascriptDispatch/HostVariant are created", false)
 #endif
+FLAGNR(Boolean, TypeAnnotations, "Enables the usage of type annotations in comments (default: false)", false)
```
MikeHolman: add DEFAULT_CONFIG_TypeAnnotations with value false, otherwise release build will break

PR #2420, opened 2017-01-26
Make jstoc.py compatible with both python 2 and python 3

greysAcademicCode:


dilijev:
Looks like this causes problems on OSX.

I'm assuming that python is linked to python3 on your system? Perhaps a better approach here would be to  make this program compatible with both python 2 and python 3 to avoid any issues with the version of python used?

greysAcademicCode:
Yeah on my distro (Arch Linux) 
```
$ python --version
Python 3.6.0
```
I just pushed an attempt to make the script run in both 2 and 3.

bin/ch/jstoc.py
```diff
@@ -4,15 +4,15 @@
 # Licensed under the MIT license. See LICENSE.txt file in the project root for full license information.
 #-------------------------------------------------------------------------------------------------------
 
-from __future__ import with_statement # py 2.5
```
dilijev: https://docs.python.org/2/library/__future__.html

`from __future__ import ...` should be a no-op in Python 3 and in order to use the print function you'll need to have at least the following:

```
from __future__ import print_function
```
I'm assuming the existing import is necessary to work in python 2.

So probably you'd want to do:

```py
from __future__ import print_function, with_statement
```

I think the operative change here which makes things compatible in both python 2 and python 3 is changing the print keyword to the print function.

dilijev:
Looks like the CI failure is not related to your change.

greysAcademicCode:
Shall I put
```python
from __future__ import print_function, with_statement
```
or keep it like this since the builds were "successful?"

dilijev:
@greysAcademicCode jstoc.py is a utility script that is not exercised in the CI, so the CI here is not important. Please make the change, and if you're able to, please verify that it runs correctly in both python 2 and python 3.

greysAcademicCode:
This builds on my end when I force it to both python3 and 2.

dilijev:

dilijev:

dilijev:
Windows 7 break is #2442

dilijev:

obastemur:
Normally we don't expect any of the py files in this repo to run with Python 3 and our future changes may break it.

However the overall effort and amount of the changes on this file... LGTM.

dilijev:
@obastemur Yes, going forward, I don't think we should worry if the scripts end up being fundamentally incompatible with Python 3, if there's a reason for it, but as long as there's no real reason, we should go with best-effort to keep them compatible with both 2 and 3.

If they do become incompatible, we should document the requirement for specifically python 2 and try to make the shebang lines so that they invoke python 2 explicitly, instead of assuming that `python` maps to a python 2 exe.

obastemur:
> If they do become incompatible, we should document the requirement for specifically python 2 and try to make the shebang lines so that they invoke python 2 explicitly, instead of assuming that python maps to a python 2 exe.

Please go ahead and add the notice wherever you feel it is suitable. Our python script files require Python 2. I can't recall any reasonable posix/unix distro with python3 is default.

Forcing with `shebang` is not a good idea as the aliases vary among the distros.. Python provides runtime version so if we really want to check that, we can do that programmatically.

I wouldn't invest any time on this though.

If a particular dev doesn't have Python 2 is installed. We expect her/him to do so.

Thanks.

dilijev:
Notices about Python 2 requirement added to https://github.com/Microsoft/ChakraCore/wiki/Building-ChakraCore#linux and https://github.com/Microsoft/ChakraCore/wiki/Building-ChakraCore#os-x

dilijev:
@greysAcademicCode ping

PR #2426, opened 2017-01-28
OOPJIT: Avoid attempting to initialize script context if JIT server process has terminated.

atulkatti:


atulkatti:
@Microsoft/chakra-developers

PR #2427, opened 2017-01-28
WASM: section id read past buffer

Cellule:
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/2427)


lib/WasmReader/WasmBinaryReader.cpp
```diff
@@ -217,38 +217,30 @@ WasmBinaryReader::ReadSectionHeader()
...
     UINT32 sectionSize = LEB128(len);
     header.end = m_pc + sectionSize;
     CheckBytesLeft(sectionSize);
 
-    const char *sectionName = nullptr;
-    UINT32 nameLength = 0;
-
-    if (sectionId > 0)
-    {
-        SectionCode sectCode = (SectionCode)(sectionId - 1);
```
Cellule: Casting from unsigned to enum type (aka signed) caused the next check to pass and then read from invalid memory.

PR #2429, opened 2017-01-29
Updating log emit code to allow for programatic emit from JS.

mrkmarron:
- Simplified emit host code in ch
- Update TTD code to support multiple different emit directories.
- Add JS call to programatically do emit.
- Update unit tests.

### DIFF
--- a/bin/ch/Helpers.cpp
+++ b/bin/ch/Helpers.cpp
@@ -16,244 +16,92 @@
 
 //TODO: x-plat definitions
 #ifdef _WIN32
-typedef char16 TTDHostCharType;
-typedef struct _wfinddata_t TTDHostFileInfo;
-typedef intptr_t TTDHostFindHandle;
-typedef struct _stat TTDHostStatType;
+#define MAX_URI_LENGTH 512
+#define TTD_HOST_PATH_SEP "\\"
 
-#define TTDHostPathSeparator _u("\\")
-#define TTDHostPathSeparatorChar _u('\\')
-#define TTDHostFindInvalid -1
-
-size_t TTDHostStringLength(const TTDHostCharType* str)
-{
-    return wcslen(str);
-}
-
-void TTDHostInitEmpty(TTDHostCharType* dst)
-{
-    dst[0] = _u('\0');
-}
-
-void TTDHostInitFromUriBytes(TTDHostCharType* dst, const byte* uriBytes, size_t uriBytesLength)
-{
-    memcpy_s(dst, MAX_PATH * sizeof(TTDHostCharType), uriBytes, uriBytesLength);
-    dst[uriBytesLength / sizeof(TTDHostCharType)] = _u('\0');
-
-    AssertMsg(wcslen(dst) == (uriBytesLength / sizeof(TTDHostCharType)), "We have an null in the uri or our math is wrong somewhere.");
-}
-
-void TTDHostAppend(TTDHostCharType* dst, size_t dstLength, const TTDHostCharType* src)
-{
-    size_t srcLength = TTDHostStringLength(src);
-    size_t dpos = TTDHostStringLength(dst);
-    Helpers::TTReportLastIOErrorAsNeeded(dpos < dstLength, "The end of the string already exceeds the buffer");
-
-    size_t srcByteLength = srcLength * sizeof(TTDHostCharType);
-    size_t dstRemainingByteLength = (dstLength - dpos - 1) * sizeof(TTDHostCharType);
-    Helpers::TTReportLastIOErrorAsNeeded(srcByteLength <= dstRemainingByteLength, "The source string must be able to fit within the destination buffer");
-
-    memcpy_s(dst + dpos, dstRemainingByteLength, src, srcByteLength);
-    dst[dpos + srcLength] = _u('\0');
-}
-
-void TTDHostAppendWChar(TTDHostCharType* dst, size_t dstLength, const wchar* src)
-{
-    size_t srcLength = wcslen(src);
-    size_t dpos = TTDHostStringLength(dst);
-    Helpers::TTReportLastIOErrorAsNeeded(dpos < dstLength, "The end of the string already exceeds the buffer");
-
-    size_t dstRemainingLength = dstLength - dpos - 1;
-    Helpers::TTReportLastIOErrorAsNeeded(srcLength <= dstRemainingLength, "The source string must be able to fit within the destination buffer");
-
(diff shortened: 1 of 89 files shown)


mrkmarron:
@agarwal-sandeep @kfarnung please review.

bin/ch/ch.cpp
```diff
@@ -314,6 +314,16 @@ HRESULT RunScript(const char* fileName, LPCSTR fileContents, JsValueRef bufferVa
...
+                JsValueRef ttString = nullptr;
+                JsValueRef global = nullptr;
+
+                ChakraRTInterface::JsCreatePropertyId("ttdLogURI", strlen("ttdLogURI"), &ttProperty);
+                ChakraRTInterface::JsCreateString(ttUri, ttUriLength, &ttString);
+                ChakraRTInterface::JsGetGlobalObject(&global);
+
+                ChakraRTInterface::JsSetProperty(global, ttProperty, ttString, false);
+
                 ChakraRTInterface::JsTTDStart();
```
agarwal-sandeep: No error checking?

bin/ch/ch.cpp
```diff
@@ -783,13 +792,13 @@ int _cdecl wmain(int argc, __in_ecount(argc) LPWSTR argv[])
         {
             doTTRecord = true;
             wchar* ruri = argv[i] + wcslen(_u("-TTRecord="));
-            Helpers::GetTTDDirectory(ruri, &ttUriByteLength, ttUri);
+            Helpers::GetTTDDirectory(ruri, &ttUriLength, ttUri);
         }
-        else if(wcsstr(argv[i], _u("-TTDebug=")) == argv[i])
```
agarwal-sandeep: >wcsstr [](start = 16, length = 6)

Looking at this again, should all of these be case insensitive compare?

lib/Runtime/Library/JavascriptLibrary.cpp
```diff
@@ -1243,6 +1243,7 @@ namespace Js
         if(scriptContext->GetThreadContext()->IsRuntimeInTTDMode())
         {
             AddFunctionToLibraryObjectWithPropertyName(globalObject, _u("telemetryLog"), &GlobalObject::EntryInfo::TelemetryLog, 3);
+            AddFunctionToLibraryObjectWithPropertyName(globalObject, _u("emitTTDLog"), &GlobalObject::EntryInfo::EmitTTDLog, 2);
         }
```
agarwal-sandeep: Personally I don't like the idea of polluting globalObject with non-spec'd properties without user being aware of it. I had this concern initially as well and now we are adding another property.
If we can't do without having all this utility methods then we should have a JsRT API like JsTTDEnableObjectOnGlobal(const char *ttdName, size_t length); when this is called we will add a object (named ttdName) on global object with all private methods like telemetryLog, emitTTDLog etc.

lib/Runtime/Library/GlobalObject.cpp
```diff
@@ -1627,6 +1627,33 @@ namespace Js
...
         return function->GetScriptContext()->GetLibrary()->GetUndefined();
     }
+
+    //Write a copy of the current TTD log to a specified location
+    Var GlobalObject::EntryEmitTTDLog(RecyclableObject* function, CallInfo callInfo, ...)
+    {
+        PROBE_STACK(function->GetScriptContext(), Js::Constants::MinStackDefault);
+        ARGUMENTS(args, callInfo);
+
+        TTDAssert(args.Info.Count == 2 && Js::JavascriptString::Is(args[1]), "Bad arguments!!!");
```
agarwal-sandeep: >TTDAssert(args.Info.Count == 2 && Js::JavascriptString::Is(args[1]), "Bad arguments!!!"); [](start = 8, length = 89)

TTDAssert today is fatal error but if we change it to assert later all this type checking will be gone. Should we explicit have type checking as we do in other built-ins?

lib/Runtime/Library/JavascriptLibrary.cpp
jianchun: Or consider adding those to global "Debug" object to reduce global namespace pollution?

bin/ch/Helpers.cpp
```diff
@@ -16,244 +16,92 @@
...
 }
 
-void TTDHostBuildCurrentExeDirectory(TTDHostCharType* path, size_t pathBufferLength)
+int TTDHostMKDir(const char* path, size_t pathLength)
 {
-    wchar exePath[MAX_PATH];
-    GetModuleFileName(NULL, exePath, MAX_PATH);
+    char16 cpath[MAX_PATH];
+    LPCUTF8 pathbase = (LPCUTF8)path;
+    utf8::DecodeUnitsIntoAndNullTerminate(cpath, pathbase, pathbase + pathLength);
```
kfarnung: We should check to make sure we don't overrun the buffer.

bin/ch/Helpers.cpp
```diff
@@ -16,244 +16,92 @@
...
-    TTDHostAppendWChar(path, pathBufferLength, exePath);
+    return _wmkdir(cpath);
 }
 
-JsTTDStreamHandle TTDHostOpen(const TTDHostCharType* path, bool isWrite)
+JsTTDStreamHandle TTDHostOpen(size_t pathLength, const char* path, bool isWrite)
 {
+    char16 wpath[MAX_PATH];
+    LPCUTF8 pathbase = (LPCUTF8)path;
+    utf8::DecodeUnitsIntoAndNullTerminate(wpath, pathbase, pathbase + pathLength);
```
kfarnung: Here too, it seems like these UTF8 methods should take in an output buffer length to be able to ensure that the buffer isn't overrun.

bin/ch/Helpers.cpp
```diff
@@ -16,244 +16,92 @@
...
-
-    TTDHostAppend(path, pathBufferLength, exePath);
+    memcpy_s(path, MAX_URI_LENGTH, exePath, *pathLength);
 }
 
-JsTTDStreamHandle TTDHostOpen(const TTDHostCharType* path, bool isWrite)
+int TTDHostMKDir(const char* path, size_t pathLength)
 {
-    return (JsTTDStreamHandle)fopen(TTDHostCharConvert(path), isWrite ? "w+b" : "r+b");
+    return mkdir(path, 0777);
```
obastemur: Do we want to give an exec. rights? otherwise, folder access should be `0666` instead of `0777`.  Actually reader and writer are the same users, even `0600` should be fine.

mrkmarron:
Closing, as included in PR #2474.

Issue #2250, opened 2016-12-20
Module object Object.prototype.toString value casing seems incorrect


guybedford:
With the following inside of a `<script type="module">` tag:

```javascript
import * as m from './m.js';
Object.prototype.toString.call(m) === '[object Module]';
```

The assertion here is returning false as the result is `[object module]` instead of what should I believe be `[object Module]` since the `Symbol.toStringTag` value is `Module` and not `module`.

Fixing this would help to ensure consistency across browsers for checking if a given object is a module namespace object, which may be needed for interop scenarios in future.

dilijev:
@boingoing re: modules

dilijev:
Candidate line where the issue is:

```
lib/Runtime/Library/JavascriptLibrary.cpp:1122:        moduleTypeDisplayString = CreateStringFromCppLiteral(_u("module"));
```

dilijev:
For this script:

```js
import * as m from './m.js';
console.log(Object.prototype.toString.call(m));
console.log(m[Symbol.toStringTag]);
console.log(Object.prototype.toString.call(m) === '[object Module]');
```

Output:

```
[object module]
module
false
```

It appears that our current implementation for @@toStringTag for modules is also lowercase (I think it uses the same string, so that's not surprising -- I just wanted to check on this since it sounded like you were saying our @@toStringTag for modules had different casing).

guybedford:
Yes it seems both `@@toStringTag` and the display string should be using capital letters, as in the spec at https://tc39.github.io/ecma262/#sec-@@tostringtag.

Thanks for looking into this and so quickly. It would really help to avoid a painful edge case in detection.

ljharb:
(fwiw, by convention every `Symbol.toStringTag` value should be PascalCased, a la `[object Object]`) Great catch!

guybedford:
Just checking if there's any movement here?

boingoing:
Sorry for being late to the party on this one. Doug is right, this is just a simple oversight. Quick fix is over here https://github.com/Microsoft/ChakraCore/pull/2430

PR #2430, opened 2017-01-30
Module namespace objects should have @@toStringTag own property set to "Module"

boingoing:
Currently this value is set to "module" which is wrong according to the spec:
https://tc39.github.io/ecma262/#sec-module-namespace-objects

This fixes https://github.com/Microsoft/ChakraCore/issues/2250


boingoing:
@dilijev, can you take a look at this one-character change ?

PR #2432, opened 2017-01-30
Merge new PR checks from release/1.4-ci

### LINKED ISSUE #2229: Add one flavor each of Legacy and DisableJIT builds to PR checks to catch regressions
It is often the case that regressions in daily configurations go unnoticed in PRs. I believe we can pick a representative sample of builds from other configurations which will catch issues without significantly bloating the number of checks we run per PR.

The extra checks we add will not be exhaustive because they won't cover all possible configurations, but the intent is to cover the ones that break most often. This means that we are more likely to catch regressions before they are checked-in, but we will still need to keep an eye on the daily builds in case of more unusual regressions.

# Slow Tests

**x64_debug** should catch the most likely Slow Test regressions.

# DisableJIT Tests

**x64_debug** should catch the most likely breaks in build and test.

# Legacy Tests

~~**x64_release**~~ [1] **x64_debug** should catch the most frequent issues introduced in Legacy builds: compile errors ~~and PreFAST (static analysis) errors~~ [1] that are specific to the Dev12 compiler ~~, and which only run under Release builds~~ [1]. Test regressions are almost always because of Javascript semantics, so any test failures seen in Legacy builds are usually also seen in Slow Tests.

If the above is not sufficient, we could also add **x64_debug**. ~~However, **x64_release** build is a *must* because of the potential to introduce PreFAST issues.~~ [1]

**Edit:**
[1] Review of netci.groovy reveals that we are not running code analysis on dev12 because dev14 has superior analysis and dev12 gives false positives (see e95e58139dba8a81bce9ee02a989aed67f4c6b94). Because of this, I think it is more valuable to use **x64_debug** instead of **x64_release**.

----

/cc @Cellule 

@obastemur While we're discussing this, it might be a good idea to discuss adding additional checks to cover the most commonly-broken scenarios for the x-plat builds we support.

Depending on the depth of discussion needed and the complexity of the changes there, we might make those changes at a later time / under a different issue. Because of the stabilization nature of the `release/1.4` branch I propose all new x-plat checks be added in master, but the new Windows builds should be added in `release/1.4` to aid with stabilization at this time.

dilijev:
Fixes #2229

### DIFF
--- a/netci.groovy
+++ b/netci.groovy
@@ -91,10 +91,7 @@ def CreateBuildTask = { isPR, buildArch, buildType, machine, configTag, buildExt
     if (nonDefaultTaskSetup == null) {
         if (isPR) {
             def osTag = machineTypeToOSTagMap.get(machine)
-            // Set up checks which apply to PRs targeting any branch
-            Utilities.addGithubPRTrigger(newJob, "${osTag} ${config}")
-            // To enable PR checks only for specific target branches, use the following instead:
-            // Utilities.addGithubPRTriggerForBranch(newJob, branch, checkName)
+            Utilities.addGithubPRTriggerForBranch(newJob, branch, "${osTag} ${config}")
         } else {
             Utilities.addGithubPushTrigger(newJob)
         }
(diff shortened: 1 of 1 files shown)


dilijev:
@dotnet-bot
test slow tests please
test linux tests please
test dev12 tests please

PR #2433, opened 2017-01-30
Enable delete last property optimization if obj is in for-in

agarwal-sandeep:
DeleteLastProperty optimization was relying on the map which holds type
participating in for-in enumeration, but the map is cleared during
PreSweepCallback causing wrong for in enumeration. Fix this by not relying
on the map.

### DIFF
--- a/lib/Backend/Lower.cpp
+++ b/lib/Backend/Lower.cpp
@@ -9519,17 +9519,12 @@ Lowerer::GenerateFastBrBReturn(IR::Instr * instr)
         Js::OpCode::BrNeq_A, labelHelper, instr);
 
     // MOV cachedDataOpnd, forInEnumeratorOpnd->enumerator.cachedData
-    // CMP cachedDataOpnd, nullptr
-    // JEQ $helper
     // MOV enumeratedCountOpnd, forInEnumeratorOpnd->enumerator.enumeratedCount
     // CMP enumeratedCountOpnd, cachedDataOpnd->cachedCount
     // JLT $loopBody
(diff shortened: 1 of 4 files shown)


agarwal-sandeep:
@jianchun @curtisman please take a look

lib/Backend/Lower.cpp
```diff
@@ -9519,12 +9519,17 @@ Lowerer::GenerateFastBrBReturn(IR::Instr * instr)
...
+    // CMP cachedDataOpnd, nullptr
+    // JEQ $helper
     // MOV enumeratedCountOpnd, forInEnumeratorOpnd->enumerator.enumeratedCount
     // CMP enumeratedCountOpnd, cachedDataOpnd->cachedCount
     // JLT $loopBody
     IR::RegOpnd * cachedDataOpnd = IR::RegOpnd::New(TyMachPtr, m_func);
     InsertMove(cachedDataOpnd,
         GetForInEnumeratorFieldOpnd(forInEnumeratorOpnd, Js::ForInObjectEnumerator::GetOffsetOfEnumeratorCachedData(), TyMachPtr), instr);
+
+    InsertCompareBranch(cachedDataOpnd, IR::AddrOpnd::NewNull(m_func), Js::OpCode::BrEq_A, labelHelper, instr);
```
curtisman: Can you just set "canUseJitFastPath" to false do you don't need to have addition check here?
There is also the cachedDataType you can use to invalidate the cache maybe?

agarwal-sandeep:
@jianchun @curtisman I removed the JIT check, please take a look.

PR #2434, opened 2017-01-31
Modify cfg checks to maintain the funcptr in register, use __guard_disptach_icall_fptr

meg-gupta:


meg-gupta:
Already reviewed. Will merge directly.

PR #2436, opened 2017-01-31
Update totalOpHelperLengthByEnd on lifetime accurately for unstructured control flow.

meg-gupta:
We can encounter loops such as the following:

L1: // loop L1 header
...
jmp L1
L2: // loop L2 header
...
jmp L1
...
jmp L2

The outer loop (L1) lexically ends before the inner loop (L2) here. The extended lifetimes in
such outer loops are extended until the end of inner loop. Also, the
loopEnd of the outer loop is extended till the end of the inner loop.
However fields such as totalOpHelperLengthByEnd on the outer loop are not
updated.
In this change, we loop over the parents of curLoop which originally ended
lexically before the curLoop and update totalOpHelperLengthByEnd on its
extended lifetimes and other fields.

### DIFF
--- a/lib/Backend/SccLiveness.cpp
+++ b/lib/Backend/SccLiveness.cpp
@@ -53,24 +53,30 @@ SCCLiveness::Build()
             AssertMsg(instr->IsBranchInstr(), "Loop tail should be a branchInstr");
             AssertMsg(instr->AsBranchInstr()->IsLoopTail(this->func), "Loop tail not marked correctly");
 
-            FOREACH_SLIST_ENTRY(Lifetime *, lifetime, this->curLoop->regAlloc.extendedLifetime)
+            Loop *loop = this->curLoop;
+            while (loop && loop->regAlloc.loopEnd == this->curLoop->regAlloc.loopEnd)
             {
-                if (this->curLoop->regAlloc.hasNonOpHelperCall)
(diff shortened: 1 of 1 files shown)


meg-gupta:
@LouisLaf Can you please take a look ?

lib/Backend/SccLiveness.cpp
```diff
@@ -53,24 +53,30 @@ SCCLiveness::Build()
             AssertMsg(instr->IsBranchInstr(), "Loop tail should be a branchInstr");
             AssertMsg(instr->AsBranchInstr()->IsLoopTail(this->func), "Loop tail not marked correctly");
 
-            FOREACH_SLIST_ENTRY(Lifetime *, lifetime, this->curLoop->regAlloc.extendedLifetime)
+            Loop *loop = this->curLoop;
+            while (loop && loop->regAlloc.loopEnd == this->curLoop->regAlloc.loopEnd)
```
LouisLaf: >== [](start = 50, length = 2)

Shouldn't this be <= or is it guaranteed to have been extended by now?  Would still be safer...

LouisLaf:
:shipit:

lib/Backend/SccLiveness.cpp
meg-gupta: It is guaranteed to be extended now. I have added an Assert for parent's loopEnd to make sure of this.

PR #2439, opened 2017-02-01
fix the order on setting write barrier bits

leirocks:
order:
1. update the verification bits for verification in background thread marking
2. update the target address
3. update card table

make sure the JIT code use the same order


jianchun:
:shipit:

Issue #2328, opened 2017-01-06
Float32Array memory leak problem


goleoh:
If run below code, Edge's total memory increases continuously.
But the memory size of snap shot doesn't increase.

Our web-app game is crashed and refreshed because of this memory problem.
We suspect that there is the same problem at Uint8Array, Float16Array, etc..

```javascript
var bufList = [];
var num = 256;
var bufSize = 96;
var buffer = new ArrayBuffer( bufSize * num);
var testMemFunc = function() {
    for(var j=0;j<num; j++) {
        var floatBuf = new Float32Array( buffer, j*bufSize + 12,  3 );
        floatBuf[0] = 7;
        floatBuf[1] = 7;
        floatBuf[2] = 7;
        bufList.push(floatBuf);
    }
    bufList.length = 0;
};
setInterval(testMemFunc, 1);
```

kunalspathak:
@goleoh - It is expected that memory grows because you are creating objects in loop. 
Did you check this test case on other browsers like Chrome and Firefox?

Penguinwizzard:
I think the issue is that the code intends to make it such that all of the views on the buffer are automatically freed at the completion of testMemFunc, and that this apparently isn't happening. The edge instance I'm writing this from is slowly growing in memory usage after running that snippet, up to 2.9GB from ~300MB right now.

goleoh:
@kunalspathak - memory is freed at the end of testMemFunc by "bufList.length = 0". The total memory usage doesn't grow up at Chrome. If we use Float32Array with no buffer, it works well. But if we use buffer the memory usage grows up continuously.

kunalspathak:
Thanks @goleoh for checking. I overlooked your test case. I too verified and the problem persist even on IE11. The problem is that every time typed array is created with `buffer` , we [add the typed array to the list of parents](https://github.com/Microsoft/ChakraCore/blob/master/lib/Runtime/Library/ArrayBuffer.cpp#L119) of that `buffer` but never remove its entry from parents list of `buffer`. This is the reason the list keeps growing.  Since parents stored in the list are weak references to the typed array, we should periodically (every time we GC) iterate through the list and delete all the dead references, thus compacting the list.

dilijev:
Closed by #2465

PR #2444, opened 2017-02-01
Float32Array memory leak problem #2328

leirocks:
the List keeping the weakreference that from ArrayBuffer to TypedArray kept expanding, there's no clean up after the TypedArray is collected.
use WeakReferenceDictionay instead of the list, and register it to ScriptContext so it can be cleaned up on post collection stage.


leirocks:
targeted wrong branch

PR #2447, opened 2017-02-01
WebAssembly On by default

Cellule:
Make WebAssembly on by default and remove it from the Experimental Flags

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/2447)


Cellule:
@agarwal-sandeep for visibility since I am changing the debugger filter list (I have a matching change in chakrafull)

Cellule:
ping

chicoxyzzy:
will wasm be available starting from Edge 15 or Edge 16?

chicoxyzzy:
Seems like this will land in Edge 16 in best case scenario?

MikeHolman:
I don't know Edge version numbers, but we are targeting the Windows release after creators update for turning on by default.

PR #2451, opened 2017-02-02
WASM: deferred parsing entrypoint fixes

Cellule:
Fixes entrypoints when 2 functions not sharing the same type are being deferred parsed

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/2451)


Cellule:
ping

lib/Runtime/Library/WasmLibrary.cpp
```diff
@@ -106,49 +106,72 @@ Js::JavascriptMethod Js::WasmLibrary::WasmDeferredParseEntryPoint(Js::AsmJsScrip
...
-        JavascriptError *pError = library->CreateWebAssemblyCompileError();
-        JavascriptError::SetErrorMessage(pError, WASMERR_WasmCompileError, msg, scriptContext);
-
-        func->GetDynamicType()->SetEntryPoint(WasmLazyTrapCallback);
-        entypointInfo->jsMethod = WasmLazyTrapCallback;
-        info->SetLazyError(pError);
+        // This can happen if another function had its type changed and then was parsed
+        // They still share the function body, so just change the entry point
+        Assert(body->GetByteCodeCount() > 0);
+        Js::JavascriptMethod externalEntryPoint = info->GetLazyError() ? WasmLazyTrapCallback : Js::AsmJsExternalEntryPoint;
```
MikeHolman: internal and external calls may have different calling convention (on x86 for instance), so I think in general external we should avoid setting both internal and external entrypoint to same function. (i think in this case it works because we always throw). imo we should keep external as AsmJsExternalEntryPoint.
i think we may also have an issue here in case of with CrossSite, but not totally sure. and if we do, seems really hard to deal with. maybe for release/1.4 we just leave that since i suspect a holistic will be too invasive

MikeHolman:
:shipit:

PR #2454, opened 2017-02-02
make it still working with disabling concurrent GC

leirocks:


jianchun:
:shipit:

leirocks:
replaced with #2591

PR #2456, opened 2017-02-02
add missing copy on access conversion for GetItemSetter

MikeHolman:


suwc:
:shipit:

suwc:
LGTM

akroshg:
>         virtual DescriptorFlags GetItemSetter(uint32 index, Var* setterValue, ScriptContext* requestContext) override

what about this case?


---
Refers to: lib/Runtime/Library/JavascriptArray.h:1113 in a9cb55a. [](commit_id = a9cb55ac121b4df344f539e7e0f776be798bc771, deletion_comment = False)

test/Array/CopyOnAccessArray_bugs.js
```diff
@@ -129,6 +129,13 @@ var tests = [
         }
     },
     {
+        name: "Reflect.set",
+        body: function ()
+        {
+            assert.isTrue(Reflect.set([1950, 1960, 1970, 1980, 1990], "0", 1), "Should be able to set property on array");
```
akroshg: add the test case for float and mix floats and ints.

MikeHolman:
@akroshg I think that case shouldn't happen, because you won't have a CopyOnAccessIntArray if you have NativeFloatArray vtable.

PR #2457, opened 2017-02-02
Trivial fix for a mistyping that can cause infinite loops.

pleath:


akroshg:
LGTM, is there a easy way to add unittest to exercise this code path?

pleath:
Probably so, yes. I’d like to push this quickly to unblock anyone who’s hitting this, but I’ll put in a UT later this afternoon.

-- Paul

PR #2458, opened 2017-02-02
VSO 8294189: Disable two incorrect, high-hitting asserts.

dilijev:
In the function

```
RegexHelper::StringReplace(JavascriptString* match, JavascriptString* input, JavascriptFunction* replacefn)
```

These two asserts would fail when `replacefn` is cross-site:

```
Assert(match->GetScriptContext() == scriptContext);
Assert(input->GetScriptContext() == scriptContext);
```

These asserts were overzealous. There is no actual problem here because the `replacefn`'s `ScriptContext` is not used for any outputs from this function and therefore is not observable.

The `scriptContext` in which `EntryReplace` is called should match `match->GetScriptContext()` and `input->GetScriptContext()`.

For now, disable these asserts to stop high-hitting noise from these assertions in the current release.

In master, the clean-up fix would be to pass in the `scriptContext` from the calling function and change these asserts to check against that `scriptContext`. It is okay for the `scriptContext` of the `replacefn` to be different, as the parameters and results will be marshalled correctly by other code.

### DIFF
--- a/lib/Runtime/Library/RegexHelper.cpp
+++ b/lib/Runtime/Library/RegexHelper.cpp
@@ -1392,8 +1392,9 @@ namespace Js
     {
         CharCount indexMatched = JavascriptString::strstr(input, match, true);
         ScriptContext* scriptContext = replacefn->GetScriptContext();
-        Assert(match->GetScriptContext() == scriptContext);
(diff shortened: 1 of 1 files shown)


dilijev:
This change is not worth targeting to release/1.4. Will simply fix in master.

PR #2459, opened 2017-02-02
VSO 8294189: Fix two incorrect, high-hitting asserts in RegexHelper::StringReplace

dilijev:
In the function

```
RegexHelper::StringReplace(JavascriptString* match, JavascriptString* input, JavascriptFunction* replacefn)
```

These two asserts would fail when `replacefn` is cross-site (where `scriptContext` is the context of the `replacefn`):

```
Assert(match->GetScriptContext() == scriptContext);
Assert(input->GetScriptContext() == scriptContext);
```

These asserts were overzealous. There is no actual problem here because the `replacefn`'s `ScriptContext` is not used for any outputs from this function and therefore is not observable.

The `scriptContext` in which `EntryReplace` is called should match `match->GetScriptContext()` and `input->GetScriptContext()`.

The fix for these asserts is to pass in the `scriptContext` from the calling function and check against that `scriptContext`. It is okay for the `scriptContext` of `replacefn` to be different, as the parameters and results will be marshalled correctly in other code.

Replaces #2458

### DIFF
--- a/lib/Runtime/Library/RegexHelper.cpp
+++ b/lib/Runtime/Library/RegexHelper.cpp
@@ -1389,10 +1389,9 @@ namespace Js
         return concatenated.ToString();
     }
 
-    Var RegexHelper::StringReplace(JavascriptString* match, JavascriptString* input, JavascriptFunction* replacefn)
+    Var RegexHelper::StringReplace(ScriptContext* scriptContext, JavascriptString* match, JavascriptString* input, JavascriptFunction* replacefn)
     {
         CharCount indexMatched = JavascriptString::strstr(input, match, true);
-        ScriptContext* scriptContext = replacefn->GetScriptContext();
         Assert(match->GetScriptContext() == scriptContext);
         Assert(input->GetScriptContext() == scriptContext);
 
@@ -1412,6 +1411,7 @@ namespace Js
                                     postfixStr, postfixLength);
             return bufferString.ToString();
         }
+
         return input;
     }
 
--- a/test/Strings/replace-xsite.js
+++ b/test/Strings/replace-xsite.js
@@ -1,0 +1,28 @@
+//-------------------------------------------------------------------------------------------------------
+// Copyright (C) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE.txt file in the project root for full license information.
+//-------------------------------------------------------------------------------------------------------
+
+WScript.LoadScriptFile("..\\UnitTestFramework\\UnitTestFramework.js");
+
+var tests = [
+    {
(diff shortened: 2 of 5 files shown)


dilijev:

dilijev:
@curtisman or @tcare could you take a look?

test/Strings/replace-xsite.js
```diff
@@ -0,0 +1,21 @@
...
+
+WScript.LoadScriptFile("..\\UnitTestFramework\\UnitTestFramework.js");
+
+var tests = [
+    {
+        name: "String.Replace ( Pattern as String , ReplaceFunction )",
+        body: function () {
+            assert.doesNotThrow(function() {
+                var str = "AAoooAAooo";
+                var replacer = WScript.LoadScriptFile('replace-xsite-replacer.js', 'samethread')
```
akroshg: you can call WScript.LoadScript('source code'), instead of LoadScriptFile - that way you can avoid creating one more file.

dilijev: @akroshg Thanks, that's a handy tip! I'll update.

tcare: Yes that will still be cross context. +1

lib/Runtime/Library/RegexHelper.cpp
```diff
@@ -1389,10 +1389,9 @@ namespace Js
         return concatenated.ToString();
     }
 
-    Var RegexHelper::StringReplace(JavascriptString* match, JavascriptString* input, JavascriptFunction* replacefn)
+    Var RegexHelper::StringReplace(ScriptContext* scriptContext, JavascriptString* match, JavascriptString* input, JavascriptFunction* replacefn)
     {
         CharCount indexMatched = JavascriptString::strstr(input, match, true);
-        ScriptContext* scriptContext = replacefn->GetScriptContext();
         Assert(match->GetScriptContext() == scriptContext);
         Assert(input->GetScriptContext() == scriptContext);
```
ThomsonTan: Do we still need these 2 asserts if `replacefn` could be in cross context?

dilijev: That's exactly the thing -- we don't need these asserts. The scriptContext of replaceFn doesn't matter here. Everything is done in the current scriptContext.

ThomsonTan: I meant is it required that both `math` and `input` should be in currentScriptContext?

---
In reply to: [103047018](https://github.com/Microsoft/ChakraCore/pull/2459#discussion_r103047018) [](ancestors = 103047018)

dilijev:
@akroshg @tcare I updated the test. Look good?

tcare:
LGTM

PR #2460, opened 2017-02-03
xplat: make sure linker doesn't pass gluing sub libs

obastemur:
- remove unused definitions and fix mixing empty object file (context)
- make sure sub libs (static) are combined (use an empty object file)


bin/ch/WScriptJsrt.cpp
```diff
@@ -789,8 +789,10 @@ bool WScriptJsrt::Initialize()
     JsValueRef buildValue;
 #ifdef _DEBUG
 #define BUILD_TYPE_STRING_CH "Debug" // (O0)
+#elif !defined(_DEBUG) && defined(ENABLE_DEBUG_CONFIG_OPTIONS)
```
jianchun: Isn't "!defined(_DEBUG)" redundant?

obastemur: Yes, indeed it is. Thanks!

jianchun:
:shipit:

obastemur:

PR #2461, opened 2017-02-03
Symbol : Equal leads to recursive call

akroshg:
JavascriptSymbol::Equal leads to recursive call when the right variable
is not symbol. Fixed that according to the spec.

### DIFF
--- a/lib/Runtime/Library/JavascriptSymbol.cpp
+++ b/lib/Runtime/Library/JavascriptSymbol.cpp
@@ -242,7 +242,14 @@ namespace Js
 
     BOOL JavascriptSymbol::Equals(JavascriptSymbol* left, Var right, BOOL* value, ScriptContext * requestContext)
     {
-        switch (JavascriptOperators::GetTypeId(right))
+        TypeId typeId = JavascriptOperators::GetTypeId(right);
+        if (typeId != TypeIds_Symbol && typeId != TypeIds_SymbolObject)
+        {
+            right = JavascriptConversion::ToPrimitive(right, JavascriptHint::None, requestContext);
+            typeId = JavascriptOperators::GetTypeId(right);
+        }
+
+        switch (typeId)
         {
         case TypeIds_Symbol:
             *value = left->GetValue() == JavascriptSymbol::FromVar(right)->GetValue();
@@ -251,7 +258,7 @@ namespace Js
             *value = left->GetValue() == JavascriptSymbolObject::FromVar(right)->GetValue();
             break;
         default:
-            *value = JavascriptOperators::Equal_Full(right, left, requestContext);
+            *value = FALSE;
             break;
         }
 
--- a/test/es6/ES6Symbol.js
+++ b/test/es6/ES6Symbol.js
@@ -972,6 +972,28 @@ var tests = [
             assert.throws(function () { o[Symbol.iterator](); }, TypeError, "Calling non-existent built-in symbol property with description fails", "Object doesn't support property or method 'Symbol(Symbol.iterator)'");
         }
     },
+    {
+        name: "Symbol equality logic",
+        body() {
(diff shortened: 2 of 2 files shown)


akroshg:
@boingoing please have a look

lib/Runtime/Library/JavascriptSymbol.cpp
```diff
@@ -251,8 +257,7 @@ namespace Js
             *value = left->GetValue() == JavascriptSymbolObject::FromVar(right)->GetValue();
             break;
         default:
-            *value = JavascriptOperators::Equal_Full(right, left, requestContext);
-            break;
+            return FALSE;
```
boingoing: Doesn't this function return the equality via the value parameter? We're not setting *value = false here. I think returning FALSE has the effect of marking the call to RecyclableObject::Equals as a failure. Not sure if that will have an impact but I think we should set value to false and return TRUE in this case. The default branch would be empty, then.

lib/Runtime/Library/JavascriptSymbol.cpp
```diff
@@ -242,6 +242,12 @@ namespace Js
 
     BOOL JavascriptSymbol::Equals(JavascriptSymbol* left, Var right, BOOL* value, ScriptContext * requestContext)
     {
+        TypeId typeId = JavascriptOperators::GetTypeId(right);
+        if (typeId != TypeIds_Symbol && typeId != TypeIds_SymbolObject)
+        {
+            right = JavascriptConversion::ToPrimitive(right, JavascriptHint::None, requestContext);
```
boingoing: Can we update typeId with the new type of right inside this if? Then we can switch on typeId which would save us a call to GetTypeId on the Symbol or SymbolObject case.
```c++
if (typeId != TypeIds_Symbol && typeId != TypeIds_SymbolObject) 
{  
  right = JavascriptConversion::ToPrimitive(right, JavascriptHint::None, requestContext); 
  typeId = JavascriptOperators::GetTypeId(right); 
}
```

test/es6/ES6Symbol.js
```diff
@@ -972,6 +972,24 @@ var tests = [
...
+            
+            var valueOfCalled = false;
+            var a = Symbol('f');
+            var b = {
+                valueOf : function() {
+                    valueOfCalled = true;
+                    return a;
+                }
+            };
+            assert.isTrue(a == b);
```
boingoing: Probably valuable to have a test for the symbol object case. Like:
```js
assert.isTrue(Object(a) == b); 
```

lib/Runtime/Library/JavascriptSymbol.cpp
akroshg: ah - good point. Fixed that.

---
In reply to: [99263244](https://github.com/Microsoft/ChakraCore/pull/2461#discussion_r99263244) [](ancestors = 99263244)

test/es6/ES6Symbol.js
akroshg: I didn't understand? Why 
```Object(a) == a
```? 
```typeof Object(a)``` is actually ```object```. 
For Chrome as well ```Object(a) == a``` is ```false```

---
In reply to: [99263716](https://github.com/Microsoft/ChakraCore/pull/2461#discussion_r99263716) [](ancestors = 99263716)

boingoing: The code here says that a symbol object is equal to a symbol if the symbol wrapped by the symbol object is the same symbol. Does that make sense? Sorry, symbol is in there a bunch.

I think Symbol.prototype[@@toPrimitive] indicates this. If we do toPrimitive of Object(a) and a is a symbol, the result should be a. Yes?

akroshg:
@boingoing have a look at again.

boingoing:
LGTM

PR #2465, opened 2017-02-03
Float32Array memory leak problem #2328

leirocks:
the List keeping the weakreference that from ArrayBuffer to TypedArray kept expanding, there's no clean up after the TypedArray is collected.
cleanup the list while adding parent once a while. changing the list to be a hashset, because the removal on the list causes linear scan and very slow

### DIFF
--- a/lib/Runtime/Library/ArrayBuffer.cpp
+++ b/lib/Runtime/Library/ArrayBuffer.cpp
@@ -95,7 +95,7 @@ namespace Js
 
         if (this->otherParents != nullptr)
         {
-            this->otherParents->Map([&](int index, RecyclerWeakReference<ArrayBufferParent>* item)
+            this->otherParents->Map([&](RecyclerWeakReference<ArrayBufferParent>* item)
             {
                 this->ClearParentsLength(item->Get());
             });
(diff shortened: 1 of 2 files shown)


lib/Runtime/Library/ArrayBuffer.cpp
```diff
@@ -128,23 +139,17 @@ namespace Js
         }
         else
         {
-            int foundIndex = -1;
-            bool parentFound = this->otherParents != nullptr && this->otherParents->MapUntil([&](int index, RecyclerWeakReference<ArrayBufferParent>* item)
+            int foundCount = 0;
+            this->otherParents->MapAndRemoveIf([&](RecyclerWeakReference<ArrayBufferParent>* item)->bool
             {
-                if (item->Get() == parent)
+                if(item->Get() == nullptr)
```
kunalspathak: >nullptr [](start = 34, length = 7)

Don't we have to match the `parent` here instead of `nullptr`?

lib/Runtime/Library/ArrayBuffer.cpp
```diff
@@ -128,23 +139,17 @@ namespace Js
...
                 return false;
-
             });
-
-            if (parentFound)
-            {
-                this->otherParents->RemoveAt(foundIndex);
-            }
-            else
+            if (foundCount != 1)
```
kunalspathak: >foundCount [](start = 16, length = 10)

Confused with this condition. Above IIRC `MapAndRemoveIf` will continue iterating as long as `item-Get() == nullptr` in which case `foundCount can be > 1` ?

lib/Runtime/Library/ArrayBuffer.cpp
leirocks: nullptr means that parent has gone, so we should clean it up the weakRef object and the list

lib/Runtime/Library/ArrayBuffer.cpp
leirocks: that can happen with the old way with a List, if that particular parent is collected, and reused.
Since I changed to use a hashset, this should not happen any more, let me fix this.

kunalspathak: Ok. If you are using `foundCount` just for assert, could you add its use under `#ifdef debug`?

---
In reply to: [99727464](https://github.com/Microsoft/ChakraCore/pull/2465#discussion_r99727464) [](ancestors = 99727464)

leirocks: sure. the compiler should optimize this away though..

---
In reply to: [99728161](https://github.com/Microsoft/ChakraCore/pull/2465#discussion_r99728161) [](ancestors = 99728161,99727464)

PR #2468, opened 2017-02-03
enhance TraceRing, enable xplat STACK_BACK_TRACE

jianchun:
Make diagnostic helper TraceRing (renamed from StackBackTraceRing)
thread safe and strongly typed so it is easier to navigate to any frame
and easier to read frame data in debugger (windbg/gdb).

Use backtrace()/backtrace_symbols() to support STACK_BACK_TRACE on xplat.

### DIFF
--- a/lib/Common/Core/StackBackTrace.cpp
+++ b/lib/Common/Core/StackBackTrace.cpp
@@ -34,9 +34,10 @@ ULONG StackBackTrace::Capture(ULONG framesToSkip)
 size_t
 StackBackTrace::Print()
 {
-    DbgHelpSymbolManager::EnsureInitialized();
-
     size_t retValue = 0;
+
+#ifdef DBGHELP_SYMBOL_MANAGER
+    DbgHelpSymbolManager::EnsureInitialized();
 
     for(ULONG i = 0; i < this->framesCount; i++)
     {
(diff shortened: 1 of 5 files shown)


lib/Common/Core/StackBackTrace.cpp
```diff
@@ -34,19 +34,40 @@ ULONG StackBackTrace::Capture(ULONG framesToSkip)
...
+
     for(ULONG i = 0; i < this->framesCount; i++)
     {
         PVOID address = this->stackBackTrace[i];
         retValue += Output::Print(_u(" "));
         retValue += DbgHelpSymbolManager::PrintSymbol(address);
         retValue += Output::Print(_u("\n"));
     }
+#else
+    char** f = backtrace_symbols(this->stackBackTrace, this->framesCount);
```
obastemur: What sort of output is expected from this call? function addresses ?

obastemur:
`backtrace_symbols` as is may not be much useful on xplat but we can improve it over time. LGTM

lib/Common/Core/StackBackTrace.cpp
jianchun: This is for printing call stack back traces from code for diagnostics. Basic function name decoding should suffice here.

Thanks for review!

obastemur: This shouldn't be decoding function names on xplat with current config. Actually, addresses are also useless there (we compile with PIC...) anyway, we take care of this later.

PR #2469, opened 2017-02-04
Array type change to ES5Array due to side-effects results in incorrect behavior in various Array methods.

atulkatti:
Addresses 2 types of issues:
1. If side-effects while reading array method arguments change the array type to ES5Array we should process the array as an ES5Array would be.
2. If side-effects during executing callback function for methods such as map, filter etc. change the array type to ES5Array we should process the array as an ES5Array would be.


atulkatti:
This has already been code reviewed in code flow.
@Microsoft/chakra-developers

PR #2470, opened 2017-02-04
Fix inconsistency in F12 rundown/reparse after redeferral. 

pleath:
F12 rundown/reparse counts on having function bodies stay around so that top-level functions can be discovered and reparsed. With the changes in redeferral, and making the function body dictionary into a LeafValueDictionary, function bodies can be discarded, so the functions that appear to be top-level at reparse time are missing information about enclosing scopes. Fix this by detecting top-level functions at byte code gen time and reparsing them even if they've been kicked out of the dictionary.

### DIFF
--- a/lib/Runtime/Base/Utf8SourceInfo.h
+++ b/lib/Runtime/Base/Utf8SourceInfo.h
@@ -130,6 +130,14 @@ namespace Js
         void SetFunctionBody(FunctionBody * functionBody);
         void RemoveFunctionBody(FunctionBody* functionBodyBeingRemoved);
 
+        void AddTopLevelFunctionInfo(Js::FunctionInfo * functionInfo, Recycler * recycler);
+        void ClearTopLevelFunctionInfoList();
+        JsUtil::List<Js::FunctionInfo *, Recycler> * EnsureTopLevelFunctionInfoList(Recycler * recycler);
+        JsUtil::List<Js::FunctionInfo *, Recycler> * GetTopLevelFunctionInfoList() const
+        {
+            return this->topLevelFunctionInfoList;
+        }
+
         // The following functions could get called even if EnsureInitialized hadn't gotten called
         // (Namely in the OOM scenario), so we simply guard against that condition rather than
         // asserting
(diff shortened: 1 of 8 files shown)


pleath:
@dotnet-bot please test Windows x64_debug Windows ci_slow_x64_debug

pleath:
@dotnet-bot please test Windows ci_slow_x64_debug

lib/Runtime/Base/Utf8SourceInfo.h
```diff
@@ -376,6 +383,7 @@ namespace Js
 
         FunctionBodyDictionary* functionBodyDictionary;
         DeferredFunctionsDictionary* m_deferredFunctionsDictionary;
+        JsUtil::List<Js::FunctionInfo *, Recycler> *topLevelFunctionInfoList;
```
agarwal-sandeep: >JsUtil::List<Js::FunctionInfo *, Recycler> *topLevelFunctionInfoList; [](start = 8, length = 69)

Top level functions are never collected right?

Rest looks good to me. Please run all tests including crawler before check-in

pleath: The top-level FunctionInfo is not collected, no. It may be redeferred.

lib/Runtime/Debug/DebugContext.cpp
```diff
@@ -221,21 +220,20 @@ namespace Js
...
                     continue;
                 }
 
                 if (shouldReparseFunctions)
                 {
-
                     BEGIN_JS_RUNTIME_CALL_EX_AND_TRANSLATE_EXCEPTION_AND_ERROROBJECT_TO_HRESULT_NESTED(cachedScriptContext, false)
                     {
-                        pFuncBody->Parse();
+                        functionInfo->GetParseableFunctionInfo()->Parse();
```
akroshg: is it possible that top level functionInfo is deferred ? Which means there is no ParseableFunctionInfo?

pleath: A deferred function does have ParseableFunctionInfo; that's the deferred FunctionProxy representation. The ParseableFunctionInfo is replaced by a FunctionBody when it's fully compiled.

akroshg: ah - thanks. LGTM. 

---
In reply to: [99685768](https://github.com/Microsoft/ChakraCore/pull/2470#discussion_r99685768) [](ancestors = 99685768)

PR #2476, opened 2017-02-06
Pass the type information from the parser to the instruction

fcastellacci:
I had to comment out an Assert to make everything work. Need to find a proper work-around

### DIFF
--- a/lib/Backend/IRBuilder.cpp
+++ b/lib/Backend/IRBuilder.cpp
@@ -4522,7 +4522,7 @@ IRBuilder::BuildElementCP(Js::OpCode newOpcode, uint32 offset, Js::RegSlot insta
                 instr->AsProfiledInstr()->u.FldInfo().valueType = ValueType::Float.SetCanBeTaggedValue(true);
                 break;
             case Js::TypeHint::Bool:
-                instr->AsProfiledInstr()->u.FldInfo().valueType = ValueType::Boolean.SetCanBeTaggedValue(true);
+                instr->AsProfiledInstr()->u.FldInfo().valueType = ValueType::Boolean;
                 break;
             case Js::TypeHint::Object:
                 instr->AsProfiledInstr()->u.FldInfo().valueType = ValueType::UninitializedObject;
@@ -4538,7 +4538,6 @@ IRBuilder::BuildElementCP(Js::OpCode newOpcode, uint32 offset, Js::RegSlot insta
         InsertBailOnNoProfile(instr);
     }
 }
-
 
 ///----------------------------------------------------------------------------
 ///
--- a/lib/Backend/IRBuilder.cpp
+++ b/lib/Backend/IRBuilder.cpp
@@ -4374,6 +4374,11 @@ IRBuilder::BuildElementRootCP(Js::OpCode newOpcode, uint32 offset)
     BuildElementCP(newOpcode, offset, Js::FunctionBody::RootObjectRegSlot, layout->Value, layout->inlineCacheIndex);
 }
 
+int __cdecl compareTypeInfos(const void * key, const void * elem)
+{
+    return (*(int*)key - ((Js::FunctionBody::TypeInformation*)elem)->bytecodeOffset);
+}
+
 void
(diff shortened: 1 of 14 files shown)


lib/Backend/IRBuilder.cpp
```diff
@@ -4501,6 +4505,32 @@ IRBuilder::BuildElementCP(Js::OpCode newOpcode, uint32 offset, Js::RegSlot insta
...
+            switch (((Js::FunctionBody::TypeInformation*)typeAnnotation)->type)
+            {
+            case Js::TypeHint::Int:
+                instr->AsProfiledInstr()->u.FldInfo().valueType = ValueType::Int.SetCanBeTaggedValue(true);
+                break;
+            case Js::TypeHint::Float:
+                instr->AsProfiledInstr()->u.FldInfo().valueType = ValueType::Float.SetCanBeTaggedValue(true);
+                break;
+            case Js::TypeHint::Bool:
+                instr->AsProfiledInstr()->u.FldInfo().valueType = ValueType::Boolean.SetCanBeTaggedValue(true);
```
MikeHolman: I'm not sure this will work. I think we store booleans as Objects

lib/Backend/IRBuilder.cpp
```diff
@@ -4501,6 +4505,32 @@ IRBuilder::BuildElementCP(Js::OpCode newOpcode, uint32 offset, Js::RegSlot insta
...
+            m_func->GetJITFunctionBody()->GetTypeAnnotationsArray()->count,
+            sizeof(Js::FunctionBody::TypeInformation), compareTypeInfos)) != nullptr)
+        {
+            switch (((Js::FunctionBody::TypeInformation*)typeAnnotation)->type)
+            {
+            case Js::TypeHint::Int:
+                instr->AsProfiledInstr()->u.FldInfo().valueType = ValueType::Int.SetCanBeTaggedValue(true);
+                break;
+            case Js::TypeHint::Float:
+                instr->AsProfiledInstr()->u.FldInfo().valueType = ValueType::Float.SetCanBeTaggedValue(true);
```
MikeHolman: i think for x86, CanBeTagged should be false

lib/JITIDL/JITTypes.h
```diff
@@ -466,6 +466,19 @@ typedef struct ConstTableContentIDL
...
+typedef struct TypeInformationIDL {
+    unsigned int bytecodeOffset;
+    unsigned int regSlot;
+    unsigned char type;
+} TypeInformationIDL;
+
+typedef struct TypeAnnotationsArrayIDL
+{
+    unsigned int count;
+    //Should I use some padding?
```
MikeHolman: yeah, but don't worry about it right now. that is just for perf for marshalling

lib/Parser/Parse.cpp
```diff
@@ -6307,7 +6314,33 @@ void Parser::ParseFncFormals(ParseNodePtr pnodeFnc, ParseNodePtr pnodeParentFnc,
                 }
 
                 m_pscan->Scan();
+                //FCASTE: Process type annotation on param here
+                //FCASTE: after parsing a term, check if there is a type annotation and parse it
+                if (CONFIG_FLAG(TypeAnnotations) && m_token.tk == tkTypeAnnBegin)
```
MikeHolman: can you move this out to its own method to avoid copy/paste?

MikeHolman:
:shipit:

lib/Backend/IRBuilder.cpp
fcastellacci: Yeah, you're right. Looks like booleans can't be set as tagged values.
I'll change that

lib/Backend/IRBuilder.cpp
fcastellacci: I tested it with annotated js in x86 and it didn't break. 
Should I change it anyway?

lib/Parser/Parse.cpp
fcastellacci: Done

PR #2477, opened 2017-02-06
Fix lowering of BrOnNotEmpty

curtisman:
When we finish iterating the ForInCache in JIT'ed code and if the cache is completed, we can exit the loop.
For BrOnNotEmpty, the target of the branch is the loop body instead of loop end.  We need to use the label after the instruction


curtisman:
@pleath  pls review.

pleath:
LGTM 👍

PR #2478, opened 2017-02-06
Fix lowering of BrOnNotEmpty

curtisman:
When we finish iterating the ForInCache in JIT'ed code and if the cache is completed, we can exit the loop.
We always generate BrOnEmpty in the byte code and it doesn't get inverted to BrOnNotEmpty usually.
But for BrOnNotEmpty, the target of the branch is the loop body instead of loop end.  We need to use the label after the instruction when the branch inversion happens.


curtisman:
@pleath  pls review.

curtisman:
Going to check into release/1.4

PR #2485, opened 2017-02-07
Remove default arguments from member definitions

joemmett:
Default arguments are not permitted on out-of-line definitions of members
of class templates. They can only be used in the declaration in the class.

This is a pre-emptive fix for upcoming MSVC compiler changes that will
diagnose this code. The Clang build only allows these currently
because they are built with -fms-extensions -Wno-microsoft.

### DIFF
--- a/lib/Common/Memory/PageAllocator.cpp
+++ b/lib/Common/Memory/PageAllocator.cpp
@@ -2493,7 +2493,7 @@ HeapPageAllocator<T>::ReleaseDecommittedSegment(__in SegmentBase<T>* segment)
 // decommit the page but don't release it
 template<typename T>
 void
-HeapPageAllocator<T>::DecommitPages(__in char* address, size_t pageCount = 1)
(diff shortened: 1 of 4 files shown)


joemmett:
@dotnet-bot test ubuntu_linux_debug please

joemmett:

dilijev:
I think it's helpful to devs to have that information inline. Would you mind commenting out the `= value` part, as we did with some functions like:

lib/Backend/GlobOpt.cpp:10332:
```cpp
bool
GlobOpt::TypeSpecializeIntBinary(IR::Instr **pInstr, Value *src1Val, Value *src2Val,
Value **pDstVal, int32 min, int32 max, bool skipDst /* = false */)
```

Corresponding definition in the header:

lib/Backend/GlobOpt.h:1436:    
```cpp
bool
TypeSpecializeIntBinary(IR::Instr **pInstr, Value *src1Val, Value *src2Val,
Value **pDstVal, int32 min, int32 max, bool skipDst = false);
```

PR #2486, opened 2017-02-08
Discard extra bits for left shift on arm

ThomsonTan:
We use `<<` in Lowerer::PeepShl to convert SHR+SHL to AND. `<<` on ARM is implemented via LSL which generates all 0 when shift count is larger than 32. This is inconsistent to implement `<<` in JavaScript which only shift the lower 5-bit. 

See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Bitwise_Operators#Left_shift.


ThomsonTan:
@LouisLaf please help review.

lib/Backend/PreLowerPeeps.cpp
```diff
@@ -98,6 +98,14 @@ IR::Instr *Lowerer::PeepShl(IR::Instr *instrShl)
     instrDef->Remove();
 
     IntConstType oldValue = src2->AsIntConstOpnd()->GetValue();
+
+    // Left shift operator (<<) on arm32 is implemented by LSL which doesn't discard bits beyond lowerest 5-bit.
```
obastemur: @ThomsonTan AFAIK shift range is specified on a 5 bits field in the instruction. 5 bits give you the range of 0-31 (32 different values) Isn't it enough to shift all the bits on a 32 bit value? Could you please elaborate a bit more?

LouisLaf: Tom's point is that the (1 << oldValue) behaves differently on x86/x64 vs ARM.  x86/x64 only use the 5 bits of the shift count as ARM effectively uses all the bits.  JavaScript says to only use the bottom 5 bits, so the x86/x64 behavior is correct.  

For code clarity however, I would remove the #ifdef.

---
In reply to: [100117486](https://github.com/Microsoft/ChakraCore/pull/2486#discussion_r100117486) [](ancestors = 100117486)

ThomsonTan: Louis is right. Both LSL on ARM and SHL on x86 accept shift count operand in register which could be 32-bit value at least. This code is handling the different behavior when the register has value larger than 32 (shift more than 32 bits).

Will remove the #ifdef.

---
In reply to: [100138094](https://github.com/Microsoft/ChakraCore/pull/2486#discussion_r100138094) [](ancestors = 100138094,100117486)

obastemur: @LouisLaf @ThomsonTan Thanks for the answers. `x86 masks the shift count to 5bits` is the key difference then. I had confused with the comment above.

ThomsonTan:
@LouisLaf any more comments?

LouisLaf:
:shipit:

PR #2489, opened 2017-02-08
Catch cases of escaping nested functions at parse time. 

pleath:
Let the parser detect cases where a symbol is used on the LHS of a function expression assignment and then escapes in a nested function. Use this information to avoid generating stack-script-function byte code that will necessitate boxing at execution time.

### DIFF
--- a/lib/Parser/Parse.cpp
+++ b/lib/Parser/Parse.cpp
@@ -1695,6 +1695,9 @@ void Parser::BindPidRefsInScope(IdentPtr pid, Symbol *sym, int blockId, uint max
         sym->SetIsModuleExportStorage(true);
     }
 
+    bool hasFuncAssignment = sym->GetHasFuncAssignment();
+    bool doesEscape = false;
+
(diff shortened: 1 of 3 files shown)


pleath:
@ianwjhalliday or @boingoing can you look at this one? (Thanks.)

lib/Parser/Parse.cpp
```diff
@@ -8842,7 +8863,7 @@ ParseNodePtr Parser::ParseVariableDeclaration(
                         pnodeInit->sxFnc.hint = pNameHint;
                         pnodeInit->sxFnc.hintLength = nameHintLength;
                         pnodeInit->sxFnc.hintOffset = nameHintOffset;
-
+                        pnodeThis->sxVar.pid->GetTopRef()->isFuncAssignment = true;
```
boingoing: Do we have this same problem in destructured assignments?

pleath: Potentially, yes. The current logic in the parser (including this change) is trying to detect common patterns to prevent obvious cases of boxing at runtime. We may at some point shift the whole legality check from byte code gen to the parser, but that's for a later release.

Issue #2472, opened 2017-02-04
Support -v / --version in `ch` binary


mathiasbynens:
There is currently no way to figure out the version of a given `ch` binary built in release mode:

```
$ ch -v
Error in opening file: -v: No such file or directory

$ ch --version
Error in opening file: --version: No such file or directory
```

Please consider making `-v` / `--version` (and possibly `-h` / `--help`) output something useful.

dilijev:
Duplicate of https://github.com/Microsoft/ChakraCore/issues/109

mathiasbynens:
Whoops, apologies! My GitHub issue-searching skills are lacking…

dilijev:
@mathiasbynens - Note that `-?` displays help.

mathiasbynens:
@dilijev Not in release mode builds, it seems:

```
$ ch -?
Error in opening file: -?: No such file or directory
```

`-h` / `--help` is a more conventional, less surprising argument name. Please consider changing it + supporting it in release mode builds :)

Anyway, I meant to say that `-?` (for test builds) doesn’t output any version information either.

dilijev:
@mathiasbynens That's correct, we don't support any flags in release builds. That decision is because of configuration flags but run-mode flags like printing help or version info should probably be supported in both. Thanks for reminding that any change in this area should enable those flags for release builds as well.

dilijev:
(Also `-?` is probably because of the convention of `/?` in windows cmd commands. We support the leading slash for parameters on Windows only as a legacy feature, so in that sense it still makes sense. There's no conflict with `-h`, `--help`, `-v`, or `--version` so we can add those.

dilijev:
@tcare Believes that this is potentially misleading as the version being reported is really just the version of ch, not the version of ChakraCore that ch loads. At a minimum, we should report `ch version 1.4.1` at this time and do the right thing. Will discuss with others on the team for opinions about how to proceed before merging this.

Issue #109, opened 2016-01-14
ch.exe needs --version


bterlson:
At a minimum it could simply pull the current commit's metadata and report commit hash and commit date.

tcare:
I believe we already store this sort of data in the DLL - it shouldn't be too hard to get it out. @curtisman and @digitalinfinity know the details of the encoding.

dilijev:
@curtisman recently made a change in this area. https://github.com/Microsoft/ChakraCore/commit/0813dc1d3847478e420842dbf6ee9015ad1ce4c6

See: https://github.com/Microsoft/ChakraCore/blob/master/lib/common/CommonDefines.h

ianwjhalliday:
This should be better defined.  After a quick discussion with @curtisman the version should at least be the major and minor numbers and build flavor and whether it is prerelease or not.

I believe all this information can be read from the resource file in chakracore.dll so ch.exe need not directly reference macros to get this information.

The `FileDescription` and `FileVersion` resource fields include all the above mentioned pieces of version/build information.  `Timestamp` is also available.

As for getting the commit information that would require embedding that information during build time.  More work to get that going.

curtisman:
There is actually an mechanism to embed commit hash in the version description in the resources already; by defining CHAKRA_VERSION_BUILD_COMMIT in the environment.  Our internal build system defines it for us, but in order to get it automatically, the msbuild project needs to be update to get it from the git repo.  

Note that the commit hash is still not 100% accurate, as it may be built with uncommitted changes.

digitalinfinity:
Brian, what exactly is the use case here? Is it for internal build system binaries or ones built locally?

-Hitesh.

On Jan 15, 2016, at 10:16, Curtis Man <notifications@github.com<mailto:notifications@github.com>> wrote:


There is actually an mechanism to embed commit hash in the version description in the resources already; by defining CHAKRA_VERSION_BUILD_COMMIT in the environment. Our internal build system defines it for us, but in order to get it automatically, the msbuild project needs to be update to get it from the git repo.

Note that the commit hash is still not 100% accurate, as it may be built with uncommitted changes.

-

bterlson:
Sorry @digitalinfinity missed your comment. The use case is knowing what version a .exe comes from, whether built locally at some point or pulled from a build server.  A specific use case is for https://github.com/bterlson/eshost - I want to report a version hash next to the host name in the output.

bterlson:
[...]

PR #2490, opened 2017-02-08
Add -v and --version flags to ch to print ChakraCore version.

dilijev:
Partially addresses #109 and #2472.

This is intended as a first pass at the issue to provide some information at this point.

Later, we plan to add more information such as whether a binary was produced as a developer build or an official build, what specific commit the binary was built from, which branch, which machine, etc. The specific set of information that we plan to provide is still being discussed in #109.

```
ch /v
ch /version
ch -v
ch --v
ch -version
ch --version
```

(Supporting `-` and `--` prefixes for short and long flag names is consistent with how we process all behaviors. The logic I used here is a replica of that in `CmdLineArgsParser::Parse`.)

The above all display the version information for `ch` and `ChakraCore` as `<major>.<minor>.<patch>.<QFE>` (note: by convention, ChakraCore's `QFE` is always `0`).

For example on Windows for ChakraCore 1.4.1 with `ch.exe` and `ChakraCore.dll` from the same build:

```
$ ch -v
ch.exe version 1.4.1.0
chakracore.dll version 1.4.1.0
```

If with the same `ch.exe`, if `ChakraCore.dll` was from, e.g., a 2.0.0 build:

```
$ ch -v
ch.exe version 1.4.1.0
chakracore.dll version 2.0.0.0
```

On Linux:

```
~/dev/ChakraCore$ ./BuildLinux/Release/ch -v
ch version 1.4.1.0
```

This flag is available in all flavors (including Release).

---

Updating the help message. Also enable `-h` and `-help` as per request in #2472.

Enables `-h` and `-help` for non-release builds, and `-?`, `-h`, and `-help` for release builds.

Messages will look as follows.

debug/test builds:

```
Usage: ch.exe [-v|-version] [-h|-help] [-?] [flaglist] <source file>
        -v|-version             Displays version info
        -h|-help                Displays this help message
        -?                      Displays this help message with complete [flaglist] info
```

release builds:

```
Usage: ch.exe [-v|-version] [-h|-help|-?] <source file>
Note: [flaglist] is not supported in Release builds; try a Debug or Test build to enable these flags.
        -v|-version             Displays version info
        -h|-help|-?             Displays this help message
```

The distinction is that `-?` is already implemented in debug/test builds to display full information about the `[flaglist]` options. This is a minimal change here that will result in guiding the user more accurately in all build flavors, the effect of which is that `-?` in debug/test builds is still the only way to get information about the `[flaglist]` options.

### DIFF
--- a/bin/ch/ChakraRtInterface.cpp
+++ b/bin/ch/ChakraRtInterface.cpp
@@ -23,6 +23,11 @@ ChakraRTInterface::ArgInfo* ChakraRTInterface::m_argInfo = nullptr;
 TestHooks ChakraRTInterface::m_testHooks = { 0 };
 JsAPIHooks ChakraRTInterface::m_jsApiHooks = { 0 };
 
+LPCSTR GetChakraDllName()
+{
+    return chakraDllName;
+}
+
 // Wrapper functions to abstract out loading ChakraCore
 // and resolving its symbols
 // Currently, these functions resolve to the PAL on Linux
--- a/bin/ch/ch.cpp
+++ b/bin/ch/ch.cpp
@@ -65,17 +65,28 @@ int HostExceptionFilter(int exceptionCode, _EXCEPTION_POINTERS *ep)
 
 void __stdcall PrintUsageFormat()
 {
-    wprintf(_u("\nUsage: %s [flaglist] <source file>\n"), hostName);
+    wprintf(_u("\nUsage: %s [-v|-version] [-h|-help] [-?] [flaglist] <source file>\n"), hostName);
+    wprintf(_u("\t-v|-version\t\tDisplays version info\n"));
+    wprintf(_u("\t-h|-help\t\tDisplays this help message\n"));
+    wprintf(_u("\t-?\t\t\tDisplays this help message with complete [flaglist] info\n"));
 }
+
+#if !defined(ENABLE_DEBUG_CONFIG_OPTIONS)
+void __stdcall PrintReleaseUsage()
+{
+    wprintf(_u("\nUsage: %s [-v|-version] [-h|-help|-?] <source file> %s"), hostName,
+        _u("\nNote: [flaglist] is not supported in Release builds; try a Debug or Test build to enable these flags.\n"));
+    wprintf(_u("\t-v|-version\t\tDisplays version info\n"));
+    wprintf(_u("\t-h|-help|-?\t\tDisplays this help message\n"));
+}
+#endif
 
 void __stdcall PrintUsage()
 {
 #if !defined(ENABLE_DEBUG_CONFIG_OPTIONS)
-    wprintf(_u("\nUsage: %s <source file> %s"), hostName,
-            _u("\n[flaglist] is not supported for Release mode\n"));
+    PrintReleaseUsage();
 #else
     PrintUsageFormat();
-    wprintf(_u("Try '%s -?' for help\n"), hostName);
 #endif
 }
 
@@ -802,6 +813,18 @@ int _cdecl wmain(int argc, __in_ecount(argc) LPWSTR argv[])
             (arglen == 7 && wcsncmp(arg, _u("version"), arglen) == 0))
         {
             PrintVersion();
+            PAL_Shutdown();
+            return EXIT_SUCCESS;
+        }
+        else if (
+#if !defined(ENABLE_DEBUG_CONFIG_OPTIONS) // release builds can display some kind of help message
+            (arglen == 1 && wcsncmp(arg, _u("?"),    arglen) == 0) ||
+#endif
+            (arglen == 1 && wcsncmp(arg, _u("h"),    arglen) == 0) ||
+            (arglen == 4 && wcsncmp(arg, _u("help"), arglen) == 0)
+            )
+        {
+            PrintUsage();
             PAL_Shutdown();
             return EXIT_SUCCESS;
         }
--- a/bin/ch/ch.cpp
+++ b/bin/ch/ch.cpp
@@ -78,6 +78,11 @@ void __stdcall PrintUsage()
 #endif
 }
 
+void __stdcall PrintVersion()
+{
+    wprintf(_u("%d.%d.%d\n"), CHAKRA_CORE_MAJOR_VERSION, CHAKRA_CORE_MINOR_VERSION, CHAKRA_CORE_PATCH_VERSION);
+}
+
 // On success the param byteCodeBuffer will be allocated in the function.
 HRESULT GetSerializedBuffer(LPCSTR fileContents, JsValueRef *byteCodeBuffer)
 {
@@ -770,7 +775,36 @@ int _cdecl wmain(int argc, __in_ecount(argc) LPWSTR argv[])
     int cpos = 1;
     for(int i = 1; i < argc; ++i)
     {
-        if(wcsstr(argv[i], _u("-TTRecord=")) == argv[i])
+        wchar *arg = argv[i];
+        size_t arglen = wcsnlen(arg, 2); // just look for prefixes for now
+
+        // support - or / prefix for flags
+        if (arglen >= 1 && (arg[0] == _u('-')
+#ifdef _WIN32
+            || arg[0] == _u('/') // '/' prefix for legacy (Windows-only because it starts a path on Unix)
+#endif
+            ))
+        {
+            // support -- prefix for flags
+            if (arglen >= 2 && arg[0] == _u('-') && arg[1] == _u('-'))
+            {
+                arg += 2; // advance past -- prefix
+            }
+            else
+            {
+                arg += 1; // advance past - or / prefix
+            }
+        }
+
+        arglen = wcsnlen(arg, 8); // ensure that we get the entirety of "version" so we don't match e.g. "versions"
+        if (arglen == 1 && wcsncmp(arg, _u("v"), arglen) == 0 ||
+            arglen == 7 && wcsncmp(arg, _u("version"), arglen) == 0)
+        {
+            PrintVersion();
+            PAL_Shutdown();
+            return EXIT_SUCCESS;
+        }
+        else if(wcsstr(argv[i], _u("-TTRecord=")) == argv[i])
         {
             doTTRecord = true;
             wchar* ruri = argv[i] + wcslen(_u("-TTRecord="));
--- a/bin/ch/ch.cpp
+++ b/bin/ch/ch.cpp
@@ -6,6 +6,7 @@
 #include "Core/AtomLockGuids.h"
 #include <CommonPal.h>
 #ifdef _WIN32
+#include <winver.h>
 #include <process.h>
 #endif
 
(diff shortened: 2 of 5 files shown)


dilijev:
@bterlson @liminzhu We could probably merge this before the 1.4.1 release, so at a minimum 1.4.1 and 2.0.0 binaries can be distinguished at this point.
/cc @mathiasbynens @obastemur @ThomsonTan @ianwjhalliday 
What do you think?

dilijev:
What is an xplat alternative for `wcsnlen`?

ThomsonTan:
Linux has wcslen. The build problem is probably about const conversion.

---
In reply to: [278212465](https://github.com/Microsoft/ChakraCore/pull/2490#issuecomment-278212465) [](ancestors = 278212465)

bin/ch/ch.cpp
```diff
@@ -770,7 +775,36 @@ int _cdecl wmain(int argc, __in_ecount(argc) LPWSTR argv[])
     int cpos = 1;
     for(int i = 1; i < argc; ++i)
     {
-        if(wcsstr(argv[i], _u("-TTRecord=")) == argv[i])
+        wchar *arg = argv[i];
```
ThomsonTan: const?

mathiasbynens:
> We could probably merge this before the 1.4.1 release, so at a minimum 1.4.1 and 2.0.0 binaries can be distinguished at this point.

🎉

obastemur:
should we include this into `-?` list?

liminzhu:
This sounds pretty useful. Is there a way for any embedder (not just ch) to programmatically know ChakraCore version #?

jianchun:
> This sounds pretty useful. Is there a way for any embedder (not just ch) to programmatically know ChakraCore version #?

We intend to ship "ChakraCoreVersion.h" file, and embedder knows it by including that header.

dilijev:
Yes, additionally, the file info of the compiled binaries already contains information about the version (that's how these macros were being consumed before):

Here's the info for some binaries I built locally:

```
ch.exe:
        Verified:       Unsigned
        Link date:      6:27 PM 2/7/2017
        Publisher:      Microsoft Corporation
        Description:    Microsoft « Chakra Core Host (Private)
        Product:        Microsoft « Chakra Core Host
        Prod version:   2.0.0.0
        File version:   2.0.0.0
        MachineType:    64-bit

ChakraCore.dll:
        Verified:       Unsigned
        Link date:      5:57 PM 2/7/2017
        Publisher:      Microsoft Corporation
        Description:    Microsoft « Chakra Core (Private)
        Product:        Microsoft « Chakra Core
        Prod version:   2.0.0.0
        File version:   2.0.0.0
        MachineType:    64-bit
```

And from the preview package I recently published to myget:

```
microsoft.chakracore.vc140.1.4.1-preview-00010-42060\lib\native\v140\x64\debug\ch.exe:
        Verified:       Signed
        Signing date:   7:17 PM 1/30/2017
        Publisher:      Microsoft Corporation
        Description:    Microsoft « Chakra Core Host
        Product:        Microsoft « Chakra Core Host
        Prod version:   1.4.1.0
        File version:   1.4.1.0 (users/doilij/nufeed:3e2e2f353d71d8e3e8522fa04b9789cf1b5c35ac.00010.42060.170130-1858)
        MachineType:    64-bit

microsoft.chakracore.vc140.1.4.1-preview-00010-42060\lib\native\v140\x64\debug\ChakraCore.dll:
        Verified:       Signed
        Signing date:   7:15 PM 1/30/2017
        Publisher:      Microsoft Corporation
        Description:    Microsoft « Chakra Core
        Product:        Microsoft « Chakra Core
        Prod version:   1.4.1.0
        File version:   1.4.1.0 (users/doilij/nufeed:3e2e2f353d71d8e3e8522fa04b9789cf1b5c35ac.00010.42060.170130-1858)
        MachineType:    64-bit
```

dilijev:
It looks like the problem is that the signature of main for xplat specifies:

```cpp
#ifndef _WIN32
static char16** argv = nullptr;
int main(int argc, char** c_argv)
```

Instead of the Windows version which uses:

```cpp
int _cdecl wmain(int argc, __in_ecount(argc) LPWSTR argv[])
```

Note `LPWSTR` is typedef'd to `wchar_t`.

Additionally, I confused things initially by using `wchar` instead of `wchar_t` -- apparently not a problem on Windows but it is for xplat.

dilijev:
Apparently we have PAL_wcslen but not PAL_wcsnlen

dilijev:
I'm going to look into using HostConfigFlags instead to avoid introducing the PAL dependency in `ch`. If that works, I'll update the PR away from the current approach.

obastemur:
> Additionally, I confused things initially by using wchar instead of wchar_t -- apparently not a problem on Windows but it is for xplat.

We shouldn't use `wchar_t` I had cleaned up that one. Its size can be 4 bytes on one platform and 2 on another..

obastemur:
LGTM. Thanks @dilijev for taking care of this!

dilijev:
@jianchun re: shipping ChakraCoreVersion.h -- the NuGet packages do not include this file since it's so minor. Would it be worth including it? For, say, .NET projects which might want to report the ChakraCore version without hardcoding is there any kind of solution that would work that doesn't involve exposing a new API that reports the version info? Should we consider implementing such an API for the benefit of hosting? Is there some other approach that could work?
/cc @liminzhu

dilijev:
> I'm going to look into using `HostConfigFlags` instead to avoid introducing the PAL dependency in ch. If that works, I'll update the PR away from the current approach.

From my investigation so far, this seems like a bit of a rabbit hole. There seems to be good reason (at least, it was thought about before) not to initialize the `HostConfigFlags` in release builds. Working around it to enable these specific parameters seems silly and difficult, so if we went this way, it would involve enabling all `HostConfigFlags`.

I agree that taking a dependency on PAL for this may be a bit extreme. One workaround is to copy a definition of `PAL_wcslen` to support taking length of `char16` strings to avoid the dependency problem, but I'd like to avoid doing something like that if it's not totally necessary. Any suggestions?

Think it's safe to merge this as-is and follow-up later if the dependency becomes a problem?

dilijev:
Updating the help message. Also enables -h and -help for non-release builds, and `-?`, `-h`, and `-help` for release builds, as per request in https://github.com/Microsoft/ChakraCore/issues/2472

dilijev:
@tcare Believes that this is potentially misleading as the version being reported is really just the version of `ch`, not the version of `ChakraCore` that `ch` loads. At a minimum, we should report at this time:

```
ch version 1.4.1
```

We could read the `ChakraCore.dll` binary information (see above), but the `ChakraCore` binaries produced in xplat builds don't seem to contain this info at this time, so we would be unable to do the same for xplat right now. It may be reasonable to make this a separate work item for later.

Assuming that in the common case `ch` will only be used with the `ChakraCore` library that was built at the same time, the version of `ch` can be used as a proxy for the version of `ChakraCore` it will load, and this could mostly resolve the common scenarios that @bterlson @mathiasbynens and others are interested in.

Will discuss with others on the team for opinions about how to proceed before merging this.

dilijev:
https://stackoverflow.com/questions/940707/how-do-i-programatically-get-the-version-of-a-dll-or-exe-file/940743#940743 shows how to do this.

Used this to prototype a solution. Output of `ch -v` etc. is now:

```
ch.exe version 1.4.1.0
chakracore.dll version 1.4.1.0
```

Or, because it reads the DLL info, when the DLL built from a different version, that is reflected in the output as well:

```
ch.exe version 1.4.1.0
chakracore.dll version 2.0.0.0
```

/cc @tcare

dilijev:
Windows x86_release error:

```
warning C6388: 'verHandle' might not be '0':  this does not adhere to the specification for the function 'GetFileVersionInfoA'. 
```

```cpp
DWORD
APIENTRY
GetFileVersionInfoSizeA(
        _In_        LPCSTR lptstrFilename, /* Filename of version stamped file */
        _Out_opt_ LPDWORD lpdwHandle       /* Information for use by GetFileVersionInfo */
        );
```

```cpp
BOOL
APIENTRY
GetFileVersionInfoA(
        _In_                LPCSTR lptstrFilename, /* Filename of version stamped file */
        _Reserved_          DWORD dwHandle,          /* Information from GetFileVersionSize */
        _In_                DWORD dwLen,             /* Length of buffer for info */
        _Out_writes_bytes_(dwLen) LPVOID lpData            /* Buffer to place the data structure */
        );
```

What? Seems like a warning that should be suppressed because clearly the value of `&verHandle` output from `GetFileVersionInfoSizeA` is supposed to be passed into `GetFileVersionInfoA`...

tcare:
dwHandle is reserved and is ignored. Just pass in nullptr. https://msdn.microsoft.com/en-us/library/windows/desktop/ms647003(v=vs.85).aspx

dilijev:
@tcare Thanks for confirming, I was thinking that might be the right thing to do but I couldn't find info about it.

bin/ch/ch.cpp
```diff
@@ -90,9 +91,65 @@ void __stdcall PrintUsage()
...
+
+    char modulename[_MAX_PATH];
+    GetModuleFileNameA(NULL, modulename, _MAX_PATH);
+    _splitpath_s(modulename, drive, _MAX_DRIVE, dir, _MAX_DIR, nullptr, 0, nullptr, 0);
+    _makepath_s(filename, drive, dir, chakraDllName, nullptr);
+
+    DWORD verHandle = NULL;
+    UINT size = 0;
+    LPBYTE lpBuffer = NULL;
+    DWORD verSize = GetFileVersionInfoSizeA(filename, &verHandle);
```
tcare: verHandle is optional for this and we don't use it later, so you can also pass nullptr here.

bin/ch/ChakraRtInterface.cpp
```diff
@@ -23,6 +23,11 @@ ChakraRTInterface::ArgInfo* ChakraRTInterface::m_argInfo = nullptr;
 TestHooks ChakraRTInterface::m_testHooks = { 0 };
 JsAPIHooks ChakraRTInterface::m_jsApiHooks = { 0 };
 
+LPCSTR GetChakraDllName()
+{
+    return chakraDllName;
```
tcare: Where is this defined?

bin/ch/ch.vcxproj
```diff
@@ -37,6 +37,7 @@
         ole32.lib;
         kernel32.lib;
         Rpcrt4.lib;
+        $(ChakraCommonLinkDependencies);
```
tcare: Do we really need to add this link dependency? Is there a more lightweight way to get the Dll name?

dilijev: Yes, I'll just link the specific lib.

bin/ch/ChakraRtInterface.cpp
dilijev: Top of the file. Different string for each platform.

bin/ch/ch.cpp
```diff
@@ -64,17 +66,88 @@ int HostExceptionFilter(int exceptionCode, _EXCEPTION_POINTERS *ep)
...
+    if (verSize != NULL)
+    {
+        LPSTR verData = new char[verSize];
+
+        if (GetFileVersionInfoA(filename, NULL, verSize, verData) &&
+            VerQueryValue(verData, _u("\\"), (VOID FAR * FAR *)&lpBuffer, &size) &&
+            (size != 0))
+        {
+            VS_FIXEDFILEINFO *verInfo = (VS_FIXEDFILEINFO *)lpBuffer;
+            if (verInfo->dwSignature == 0xfeef04bd)
```
tcare: The magic number can be VS_FFI_SIGNATURE

tcare:
LGTM, please run by @curtisman.

bin/ch/ch.cpp
dilijev: Ah thanks, that was bugging me too.

PR #2492, opened 2017-02-08
change debug attach asserts about OOM to failfasts

MikeHolman:


lib/Runtime/Debug/DebugContext.cpp
```diff
@@ -142,11 +142,7 @@ namespace Js
         });
         END_TRANSLATE_OOM_TO_HRESULT(hr);
 
-        if (hr != S_OK)
-        {
-            Assert(FALSE);
-            return hr;
-        }
+        DEBUGGER_ATTACHDETACH_FATAL_ERROR_IF_FAILED(hr);
```
akroshg: why failfast when it is OOM? Why not replace with Assert(hr==S_OK || hr==E_OUTOFMEMORY)?

agarwal-sandeep: Yes we should return OOM and let the caller which have fail fast handle it

MikeHolman: Apparently Paul is removing this function anyway, so I'll wait for that to merge.

MikeHolman:

akroshg:
:shipit:

MikeHolman:

MikeHolman:

PR #2493, opened 2017-02-08
change how we do our seh failfast

MikeHolman:
We were seeing on jscript9 our SEH failfast was misbehaving, which led to bad misattribution of Watson crashes. It is fixed there, but same may apply to ChakraCore


lib/Runtime/Library/JavascriptFunction.cpp
```diff
@@ -691,10 +697,10 @@ namespace Js
         __finally
         {
             // 0xE06D7363 is C++ exception code
-            if (exceptionCode != 0 && !IsDebuggerPresent() && exceptionCode != 0xE06D7363 && exceptionAction != EXCEPTION_CONTINUE_EXECUTION)
+            if (AbnormalTermination() && exceptionCode != 0 && !IsDebuggerPresent() && exceptionCode != 0xE06D7363)
```
agarwal-sandeep: >AbnormalTermination [](start = 16, length = 19)

Good one but does this have performance impact?
[https://msdn.microsoft.com/en-us/library/windows/desktop/ms679265(v=vs.85).aspx](https://msdn.microsoft.com/en-us/library/windows/desktop/ms679265(v=vs.85).aspx)

MikeHolman: I looked at implementation and seems not that expensive. I'll swap the condition around though so we only come here if exceptionCode != 0

ThomsonTan:
>                 exceptionAction = CallRootEventFilter(exceptionCode, GetExceptionInformation()))

Remove this variable if we don't need keep the action?

---
Refers to: lib/Runtime/Library/JavascriptFunction.cpp:692 in b30069b. [](commit_id = b30069bca8f470096f34675b4c7fa9ef55d27bdd, deletion_comment = False)

lib/Runtime/Library/JavascriptFunction.cpp
LouisLaf: It is actually very cheap.  It is just a local variable managed by the compiler.

---
In reply to: [100167801](https://github.com/Microsoft/ChakraCore/pull/2493#discussion_r100167801) [](ancestors = 100167801)

LouisLaf:
:shipit:

lib/Runtime/Library/JavascriptFunction.cpp
```diff
@@ -683,18 +688,18 @@ namespace Js
...
         __finally
         {
             // 0xE06D7363 is C++ exception code
-            if (exceptionCode != 0 && !IsDebuggerPresent() && exceptionCode != 0xE06D7363 && exceptionAction != EXCEPTION_CONTINUE_EXECUTION)
+            if (exceptionCode != 0 && exceptionCode != 0xE06D7363 && AbnormalTermination() && !IsDebuggerPresent() )
             {
-                // ensure that hosts are not doing SEH across Chakra frames, as that can lead to bad state (e.g. destructors not being called)
-                UnexpectedExceptionHandling_fatal_error(&exceptionInfo);
+                scriptContext->GetThreadContext()->SetAbnormalExceptionCode(exceptionCode);
+                scriptContext->GetThreadContext()->SetAbnormalExceptionRecord(&exceptionInfo);
```
ThomsonTan: >SetAbnormalExceptionRecord [](start = 51, length = 26)

Is it intended to ignore running fatal error after this and propagate exception up?

MikeHolman: Yes. Case we had in jscript9 was the top level filter handles the exception and then terminates process. So all finally blocks are run (and we were doing failfast at that point, which messed up attribution).

The idea behind this is so that in case someone handles the exception, that they can't call back into Chakra later.

ThomsonTan: Ok, this makes sense. But will it make troubleshooting harder? Since we'll lose information of the last exception context (not just exception info).

---
In reply to: [100175696](https://github.com/Microsoft/ChakraCore/pull/2493#discussion_r100175696) [](ancestors = 100175696)

MikeHolman: You're right, it definitely can make troubleshooting harder, but hopefully if someone is doing this when they call back they will be calling back with their same bad try/except. We don't have a concrete case of this happening anymore, so I guess we'll have to cross this bridge if we get there.

PR #2496, opened 2017-02-08
Adding func addr range cache on script context for IsNativeAddr (WER case)

satheeshravi:
- This is to enable WER scenario to perform IsNativeAddress check with OOPjit case.
- IsNativeAddress in DAC wasn't handling OOPjit case correctly today. 

Fix:
- Since we will not be able to make RPC call during WER, we cache the func ranges on the script context and check this cache during IsNativeAddress.

Pending:
- IE Perf lab for memory regression
- Unit tests

### DIFF
--- a/lib/Runtime/Base/ScriptContext.cpp
+++ b/lib/Runtime/Base/ScriptContext.cpp
@@ -49,6 +49,8 @@ namespace Js
         HeapDelete(scriptContext);
     }
 
+    CriticalSection JITPageAddrToFuncRangeCache::cs;
+
     ScriptContext::ScriptContext(ThreadContext* threadContext) :
         ScriptContextBase(),
         interpreterArena(nullptr),
@@ -194,6 +196,7 @@ namespace Js
         , bailOutOffsetBytes(0)
 #endif
         , debugContext(nullptr)
+        , jitFuncRangeCache(nullptr)
     {
        // This may allocate memory and cause exception, but it is ok, as we all we have done so far
        // are field init and those dtor will be called if exception occurs
@@ -478,6 +481,11 @@ namespace Js
         {
             DeleteNativeCodeGenerator(this->nativeCodeGen);
             nativeCodeGen = NULL;
+        }
+        if (jitFuncRangeCache != nullptr)
+        {
+            HeapDelete(jitFuncRangeCache);
+            jitFuncRangeCache = nullptr;
         }
 #endif
 
@@ -1242,6 +1250,7 @@ namespace Js
 #if ENABLE_NATIVE_CODEGEN
         // Create the native code gen before the profiler
         this->nativeCodeGen = NewNativeCodeGenerator(this);
+        this->jitFuncRangeCache = HeapNew(JITPageAddrToFuncRangeCache);
 #endif
 
 #ifdef PROFILE_EXEC
@@ -4347,6 +4356,12 @@ void ScriptContext::RegisterConstructorCache(Js::PropertyId propertyId, Js::Cons
 }
 #endif
 
+JITPageAddrToFuncRangeCache * 
+ScriptContext::GetJitFuncRangeCache()
+{
+    return jitFuncRangeCache;
+}
+
 void ScriptContext::RegisterPrototypeChainEnsuredToHaveOnlyWritableDataPropertiesScriptContext()
 {
     Assert(!IsClosed());
@@ -4419,7 +4434,7 @@ void ScriptContext::RegisterPrototypeChainEnsuredToHaveOnlyWritableDataPropertie
 #if ENABLE_NATIVE_CODEGEN
     BOOL ScriptContext::IsNativeAddress(void * codeAddr)
     {
-        return this->GetThreadContext()->IsNativeAddress(codeAddr);
+        return this->GetThreadContext()->IsNativeAddress(codeAddr, this);
     }
 #endif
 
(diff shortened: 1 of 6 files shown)


satheeshravi:
@agarwal-sandeep @LouisLaf : Please take a look.

lib/Runtime/Base/ThreadContext.cpp
```diff
@@ -4099,7 +4099,15 @@ BOOL ThreadContext::IsNativeAddress(void * pCodeAddr)
         boolean result;
         HRESULT hr = JITManager::GetJITManager()->IsNativeAddr(this->m_remoteThreadContextInfo, (intptr_t)pCodeAddr, &result);
         JITManager::HandleServerCallResult(hr, RemoteCallType::HeapQuery);
```
agarwal-sandeep: This should be under DBG for verification only?

lib/Runtime/Base/ThreadContext.cpp
```diff
@@ -4099,7 +4099,15 @@ BOOL ThreadContext::IsNativeAddress(void * pCodeAddr)
...
-        return result;
+
+        bool isNativeAddr = false;
+        for (Js::ScriptContext *scriptContext = scriptContextList; scriptContext && !isNativeAddr; scriptContext = scriptContext->next)
+        {
+            isNativeAddr = scriptContext->GetJitFuncRangeCache()->IsNativeAddr(pCodeAddr);
+        }
+
+        Assert(result == (isNativeAddr? 1:0));
```
agarwal-sandeep: Should we pass scriptContext to IsNativeAddress and first check the current scriptContext before walking the list? I think most of calls should be from current scriptContext.

lib/Runtime/Base/ThreadContext.h
```diff
@@ -481,9 +481,9 @@ class ThreadContext sealed :
     typedef JsUtil::BaseDictionary<Js::PropertyId, TypeHashSet *, Recycler, PowerOf2SizePolicy> PropertyIdToTypeHashSetDictionary;
     typedef JsUtil::WeaklyReferencedKeyDictionary<const Js::PropertyRecord, PropertyGuardEntry*, Js::PropertyRecordPointerComparer> PropertyGuardDictionary;
 
+    intptr_t m_prereservedRegionAddr;
```
agarwal-sandeep: >intptr_t m_prereservedRegionAddr; [](start = 4, length = 33)

Why public? have a getter instead?

lib/Runtime/Base/ScriptContext.cpp
```diff
@@ -310,6 +313,8 @@ namespace Js
 
     void ScriptContext::InitializeAllocations()
     {
+        this->jitFuncRangeCache = HeapNew(JITPageAddrToFuncRangeCache);
```
agarwal-sandeep: >this->jitFuncRangeCache = HeapNew(JITPageAddrToFuncRangeCache); [](start = 8, length = 63)

Should be in ScriptContext::InitializePreGlobal where we create nativeCodeGen and under ENABLE_NATIVE_CODEGEN

lib/Runtime/Base/ScriptContext.cpp
```diff
@@ -400,6 +405,12 @@ namespace Js
         // TODO: Can we move this on Close()?
         ClearHostScriptContext();
 
+        if (jitFuncRangeCache != nullptr)
+        {
+            HeapDelete(jitFuncRangeCache);
+            jitFuncRangeCache = nullptr;
+        }
```
agarwal-sandeep: Should this be just before we call DeleteNativeCodeGenerator?

lib/Runtime/Base/ScriptContext.cpp
```diff
@@ -5915,4 +5932,118 @@ void ScriptContext::RegisterPrototypeChainEnsuredToHaveOnlyWritableDataPropertie
...
+
+            int byteCount = 0;
+            Assert(!largeJitFuncToSizeMap->TryGetValue(address, &byteCount));
+            largeJitFuncToSizeMap->Add(address, bytes);
+        }
+    }
+
+    void* JITPageAddrToFuncRangeCache::GetPageAddr(void * address)
+    {
+        return (void*)((uintptr_t)address & ~0xFFF);
```
agarwal-sandeep: >0xFFF [](start = 45, length = 5)

Should we use AutoSystemInfo::PageSize instead of hardcoded value?

lib/Runtime/Base/ScriptContext.cpp
```diff
@@ -5915,4 +5932,118 @@ void ScriptContext::RegisterPrototypeChainEnsuredToHaveOnlyWritableDataPropertie
...
+        {
+            if (jitPageAddrToFuncRangeMap == nullptr)
+            {
+                jitPageAddrToFuncRangeMap = HeapNew(JITPageAddrToFuncRangeMap, &HeapAllocator::Instance);
+            }
+
+            void * pageAddr = GetPageAddr(address);
+            RangeMap * rangeMap = nullptr;
+            jitPageAddrToFuncRangeMap->TryGetValue(pageAddr, &rangeMap);
+            if (rangeMap == nullptr)
```
agarwal-sandeep: >if (rangeMap == nullptr) [](start = 12, length = 24)

Check TryGetValue return instead and assert on out rangeMap?

lib/Runtime/Language/AsmJsCodeGenerator.cpp
```diff
@@ -37,6 +37,10 @@ namespace Js
 #if ENABLE_DEBUG_CONFIG_OPTIONS
             funcEntrypointInfo->SetIsTJMode(true);
 #endif
+            if (!PreReservedVirtualAllocWrapper::IsInRange((void*)mScriptContext->GetThreadContext()->m_prereservedRegionAddr, (void*)address))
+            {
+                mScriptContext->GetJitFuncRangeCache()->AddFuncRange((void*)address, (int)entrypointInfo->GetCodeSize());
```
agarwal-sandeep: >(int) [](start = 85, length = 5)

Any issue with typecast to int? GetCodeSize return ptrdiff_t which on x64 is 8 bytes whereas int is just 4?

lib/Runtime/Base/ScriptContext.cpp
```diff
@@ -5915,4 +5932,118 @@ void ScriptContext::RegisterPrototypeChainEnsuredToHaveOnlyWritableDataPropertie
...
+        }
+
+        if (largeJitFuncToSizeMap != nullptr)
+        {
+            HeapDelete(largeJitFuncToSizeMap);
+            largeJitFuncToSizeMap = nullptr;
+        }
+    }
+
+    void JITPageAddrToFuncRangeCache::AddFuncRange(void * address, int bytes)
```
agarwal-sandeep: >int bytes [](start = 67, length = 9)

JITOutputIDL.codeSize is unsigned int

lib/Runtime/Language/AsmJsCodeGenerator.cpp
satheeshravi: Checked the upper bound for uint 32bit

---
In reply to: [100214944](https://github.com/Microsoft/ChakraCore/pull/2496#discussion_r100214944) [](ancestors = 100214944)

agarwal-sandeep:
:shipit:

Issue #2428, opened 2017-01-28
Exponential function reports overflow in MS Edge JavaScript


JdeH:
I am using plotly.js to plot an exponential plot with MS Edge.
It runs fine on Firefox, Safari, Chrome and Internet Explorer, but on Edge it reports the argument of the exp function being out of range, reporting:

> SCRIPT5026: The number of fractional digits is out of range

The code is compiled from [Transcrypt](http://www.transcrypt.org).

The Python / Transcrypt source is:

```
    xValues = list (range (10))
    yValues = [math.exp (x**2) for x in xValues] 
    kind = 'logarithmic'
    Plotly.plot (
        kind,
        [
            {
                x: xValues,
                y: yValues
            }
        ],
        {
            title: kind,
            xaxis: {title: 'x'},
            yaxis: {type: 'log', tickformat: '2e', title: 'exp (x**2)'}
        }
    )
```

The generated JavaScript, running directly on Edge is:

```
			var xValues = list (range (10));
			var yValues = function () {
				var __accu0__ = [];
				var __iterable0__ = xValues;
				for (var __index0__ = 0; __index0__ < __iterable0__.length; __index0__++) {
					var x = __iterable0__ [__index0__];
					__accu0__.append (math.exp (Math.pow (x, 2)));
				}
				return __accu0__;
			} ();
			var kind = 'logarithmic';
			Plotly.plot (kind, list ([dict ({x: xValues, y: yValues})]), dict ({title: kind, xaxis: dict ({title: 'x'}), yaxis: dict ({type: 'log', tickformat: '2e', title: 'exp (x**2)'})}));
		}
```

You can run / debug this code live at: http://www.transcrypt.org/live/transcrypt/demos/plotly_demo/plotly_demo.min.html

N.B. It's just JavaScript, the fact that it's generated from Transcrypt shouldn't make any difference.
Seems the error emerges somewhere from the plotly library.
However it runs fine on all other browsers, so there seems to be an anomaly in Edge.

Kind regards
Jacques de Hooge

P.S. Probably the error  is caused somehow by the Chakra code found here, spot marked with // !!!

https://github.com/Microsoft/ChakraCore/blob/e295155dac83b3479020f47cb5450cb8ab2f5176/lib/Runtime/Library/JavascriptNumber.cpp

containing the following:

```
    Var JavascriptNumber::EntryToExponential(RecyclableObject* function, CallInfo callInfo, ...)
    {
[...]

PR #2497, opened 2017-02-09
Don't throw RangeError for toExponential(undefined)

ThomsonTan:
RangeError was raised for `Number.prototype.toExponential(undefined)` unexpectedly after refactoring range check (fraction should be between 1 to 20) out of else statement which converts non-integer and not undefined fractionDigits to integer. The spec also doesn't imply range error for it (see below link).

http://www.ecma-international.org/ecma-262/6.0/#sec-number.prototype.toexponential

### DIFF
--- a/lib/Runtime/Library/JavascriptNumber.cpp
+++ b/lib/Runtime/Library/JavascriptNumber.cpp
@@ -585,6 +585,7 @@ namespace Js
         {
             //use the first arg as the fraction digits, ignore the rest.
             Var aFractionDigits = args[1];
+            bool noRangeCheck = false;
 
             // shortcut for tagged int's
             if(TaggedInt::Is(aFractionDigits))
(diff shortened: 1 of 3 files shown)


lib/Runtime/Library/JavascriptNumber.cpp
```diff
@@ -585,19 +585,23 @@ namespace Js
...
             //use the first arg as the fraction digits, ignore the rest.
             Var aFractionDigits = args[1];
+            bool noRangeCheck = false;
 
             // shortcut for tagged int's
             if(TaggedInt::Is(aFractionDigits))
             {
                 fractionDigits = TaggedInt::ToInt32(aFractionDigits);
             }
             else if(JavascriptOperators::GetTypeId(aFractionDigits) == TypeIds_Undefined)
```
akroshg: there is no need for this check, the ToInteger will convert it properly, right?

lib/Runtime/Library/JavascriptNumber.cpp
```diff
@@ -585,19 +585,23 @@ namespace Js
...
-                ; // fraction digits = -1
+            {
```
akroshg: could you change this code to?
```js
            else if(JavascriptOperators::GetTypeId(aFractionDigits) != TypeIds_Undefined)
            {
                fractionDigits = (int)JavascriptConversion::ToInteger(aFractionDigits, scriptContext);
                if (fractionDigits < 0 || fractionDigits >20)
                {
                    JavascriptError::ThrowRangeError(scriptContext, JSERR_FractionOutOfRange);
                }
            }
        }


```

ThomsonTan: That will make the code behave a little different (like before the refactor). `fractionDigits` as `TaggedInt` will not be checked.

---
In reply to: [100205438](https://github.com/Microsoft/ChakraCore/pull/2497#discussion_r100205438) [](ancestors = 100205438)

akroshg: There was no point checking for taggedInt, the ToInteger is doing the same thing. Anyway - this is good to go.


---
In reply to: [100207392](https://github.com/Microsoft/ChakraCore/pull/2497#discussion_r100207392) [](ancestors = 100207392,100205438)

akroshg:
:shipit:

akroshg:
fixes #2428

PR #2498, opened 2017-02-09
debug fill interpreter stack frame

MikeHolman:
For debugging purposes, it is helpful to have interpreter stack frame to be initialized


satheeshravi:
:shipit:

PR #2501, opened 2017-02-09
Fix kraken regressions

LouisLaf:
My previous checking caused a regressions in a couple of the Kraken crypto benchmarks. Loading int32 from native arrays is ok even if int-type-spec is disabled, because we will convert these to float64, as long as float-type-spec is enabled.


LouisLaf:
@pleath, @meg-Gupta, @satheeshravi , please review.

pleath:
LGTM.

PR #2502, opened 2017-02-09
fix incorrect debugger mode reparse after function redefer

jianchun:
Debug mode reparse should still set "reuseNestedFunc", otherwise we'll
discard existing nested func info and create new FunctionInfo/FunctionBody
for nested functions, conflicting with existing ones.

Existing debug mode reparse also not expecting function redefer introduced
ParseableFunctionInfo in nested FunctionBody chain. It did not reuse
existing child FunctionBody and not clean up them for reparse properly
when a parent function was redeferred. Patch the code to fix this.

Credits to Paul/Akrosh.


pleath:
Aha. Mystery(-ies) solved, then. LGTM.

akroshg:
:shipit:

Issue #2482, opened 2017-02-07
"length" property assertion in JSON.cpp can be triggered with proxy


anba:
Version: cb7817bda7e4b265e7a7364641123e20204facfd
Platform: Ubuntu x64

Test case:
```js
var i = 0;
JSON.stringify(new Proxy([], {
    get(t, pk, r){
        if (pk === "length") {
            return ++i;
        }
        return Reflect.get(t, pk, r);
    }
}));
```

Triggers this assertion:
```
ASSERTION 5573: (/home/andre/git/ChakraCore/lib/Runtime/Library/JSON.cpp, line 780) Js::JavascriptConversion::ToLength(Js::JavascriptOperators::OP_GetLength(value, scriptContext), scriptContext) == length
 Failure: (Js::JavascriptConversion::ToLength(Js::JavascriptOperators::OP_GetLength(value, scriptContext), scriptContext) == length)
```

akroshg:
Thanks @anba, the assert is indeed wrong. PR is on the way.

bterlson:
I think this is fixed now?

akroshg:
@bterlson yup.

PR #2503, opened 2017-02-09
Fix for assertion fired in JSON.stringify.

akroshg:
We called the GetLength twice assuming that the length will not change. However with proxy you can give different length both time and that will causes an assert. No fre impact so just removed the assert. While testing I found that we can pass more that 2^32 and that will cause the overflow in the length. However the CompoundString cannot handle more than MaxCharLength (INT_MAX-1) and it will eventually OOM. Instead of waiting till that point I am raising the RangeError.


akroshg:
Fixes #2482

akroshg:
@suwc @kunalspathak please have a look.

kunalspathak:
:shipit:

dilijev:

suwc:
:shipit:

PR #2508, opened 2017-02-10
Sync with master

fcastellacci:


MikeHolman:
We shouldn't sync from type-annotations to master, we should sync from master to type-annotations.

PR #2515, opened 2017-02-10
Eliminate noisy assert involving thread-safe counters. 

pleath:
When we redefer, reset the flag indicating that a background thread may access the counters, as we can't be redeferring if we're jitting, and jitting from a redeferred state is impossible.


pleath:
@leirocks This is the one we were just discussing. (Thanks.)

Issue #2483, opened 2017-02-07
Crash/Assertion error when revoking proxy during proxy operation


anba:
Version: cb7817bda7e4b265e7a7364641123e20204facfd
Platform: Ubuntu x64

Test case:
```js
var {proxy, revoke} = Proxy.revocable(() => {}, {
  get apply() {
    revoke();
  }
});

proxy();
```

Results in nullptr-crash resp. assertion error in debug mode:
```
ASSERTION 5592: (/home/andre/git/ChakraCore/lib/Jsrt/../Runtime/Types/RecyclableObject.inl, line 13) RecyclableObject::Is aValue is null
 Failure: (aValue != nullptr)
```

Similar issues are present for the other proxy traps.

akroshg:
@anba, this is a good find. I'll work on this one.

PR #2516, opened 2017-02-11
Proxy fixes when it is revoked while in trap

akroshg:
When we are proxy trap, it is possible that we invoke the user script and revoke current proxy.
In that case we made the target and handler to be nullptr. After revoking
once we return back we probably be working on null object.
According to spec we need to take the local variables when we begin in the
trap, that way the target and handler will be not be gone.
Added test cases for all traps.

### DIFF
--- a/lib/Runtime/Library/JavascriptProxy.cpp
+++ b/lib/Runtime/Library/JavascriptProxy.cpp
@@ -194,6 +194,17 @@ namespace Js
     {
         PROBE_STACK(GetScriptContext(), Js::Constants::MinStackDefault);
 
+        //1. Let handler be the value of the[[ProxyHandler]] internal slot of O.
+        Js::RecyclableObject *handlerObj = this->handler;
+
+        //2. If handler is null, then throw a TypeError exception.
+        if (handlerObj == nullptr)
+        {
+            JavascriptError::ThrowTypeError(requestContext, JSERR_ErrorOnRevokedProxy, _u("preventExtensions"));
+        }
+        //3. Let target be the value of the[[ProxyTarget]] internal slot of O.
+        Js::RecyclableObject *targetObj = this->target;
+
         Assert((static_cast<DynamicType*>(GetType()))->GetTypeHandler()->GetPropertyCount() == 0 ||
             (static_cast<DynamicType*>(GetType()))->GetTypeHandler()->GetPropertyId(GetScriptContext(), 0) == InternalPropertyIds::WeakMapKeyMap);
         JavascriptFunction* gOPDMethod = GetMethodHelper(PropertyIds::getOwnPropertyDescriptor, requestContext);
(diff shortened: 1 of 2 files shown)


akroshg:
@kunalspathak @leirocks please have a look.

akroshg:
Fixes #2483

test/es6/proxybugs.js
```diff
@@ -101,7 +101,208 @@ var tests = [
...
+            Object.preventExtensions(obj.proxy);
+            assert.isTrue(trapCalled);
+        }
+    },
+    {
+        name: "Assertion validation : revoking the proxy in getOwnPropertyDescriptor trap",
+        body() {
+            var trapCalled = false;
+            var handler = {
+                getOwnPropertyDescriptor : function(a, b, c) {
```
kunalspathak: >getOwnPropertyDescriptor [](start = 16, length = 24)

may be return descriptor object using Reflect.get() ?

test/es6/proxybugs.js
```diff
@@ -101,7 +101,208 @@ var tests = [
...
+    },
+    {
+        name: "Assertion validation : revoking the proxy in deleteProperty trap",
+        body() {
+            var trapCalled = false;
+            var handler = {
+                deleteProperty : function (a, b, c) {
+                    trapCalled = true;
+                    obj.revoke();
+                    return {};
```
kunalspathak: >{} [](start = 27, length = 2)

nit: return boolean?

lib/Runtime/Library/JavascriptProxy.cpp
```diff
@@ -194,6 +194,17 @@ namespace Js
     {
         PROBE_STACK(GetScriptContext(), Js::Constants::MinStackDefault);
 
+        //1. Let handler be the value of the[[ProxyHandler]] internal slot of O.
+        Js::RecyclableObject *handler = this->handler;
```
kunalspathak: >handler [](start = 30, length = 7)

can you rename `handler` to something else like `handleObj` and vice-versa for `target` object to avoid this bug in future when someone make changes to proxy code? It will be easy to spot if we using `handler` of instance or cached one.

lib/Runtime/Library/JavascriptProxy.cpp
```diff
@@ -744,6 +765,8 @@ namespace Js
             return FALSE;
         }
```
kunalspathak: This is inconsistent compared to other methods. Can we move this before `handler == nullptr` check?

lib/Runtime/Library/JavascriptProxy.cpp
```diff
@@ -1177,6 +1208,12 @@ namespace Js
 
     BOOL JavascriptProxy::GetDefaultPropertyDescriptor(PropertyDescriptor& descriptor)
     {
+        if (target == nullptr)
+        {
+            // REVIEW : Should we throw error here?
```
kunalspathak: >REVIEW [](start = 15, length = 6)

If i read the spec correctly, this should throw TypeError.

kunalspathak:
LGTM the changes you have made. I assumed you have visited all the places where this needed fix.

test/es6/proxybugs.js
akroshg: All needed is revoke call in the trap, no need to complicate what it returns.

lib/Runtime/Library/JavascriptProxy.cpp
akroshg: Converted to throw type error.

lib/Runtime/Library/JavascriptProxy.cpp
akroshg: Fixed it.

PR #2518, opened 2017-02-13
Add missing StSlot for non-temp registers for JITLoopBody

ThomsonTan:
In JITLoopBody, we gnerate StSlot for the last instruction we imported from given bytecode instruction. This works fine when one bytecode instruction is mapped to one IR instruction. But in some cases, one bytecode instruction (`NewConcatStrMulti`) could be converted to multiple IR instructions (`Conv_primStr/NewConcatStrMulti/SetConcatStrMultiItem`), then we only check the very last IR instruction to generate StSlot and miss it for previous instructions.

This fix keeps the last processed IR instruction and processes all the IR instructions after that in backward order.

### DIFF
--- a/lib/Backend/IRBuilder.cpp
+++ b/lib/Backend/IRBuilder.cpp
@@ -670,6 +670,7 @@ IRBuilder::Build()
     JsUtil::BaseDictionary<IR::Instr*, int, JitArenaAllocator> ignoreExBranchInstrToOffsetMap(m_tempAlloc);
 
     Js::LayoutSize layoutSize;
+    IR::Instr* lastProcessedInstrForJITLoopBody = nullptr;
     for (Js::OpCode newOpcode = m_jnReader.ReadOp(layoutSize); (uint)m_jnReader.GetCurrentOffset() <= lastOffset; newOpcode = m_jnReader.ReadOp(layoutSize))
     {
         Assert(newOpcode != Js::OpCode::EndOfBlock);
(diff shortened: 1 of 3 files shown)


ThomsonTan:

lib/Backend/IRBuilder.cpp
```diff
@@ -750,28 +751,41 @@ IRBuilder::Build()
             }
         }
 #endif
-        if (IsLoopBodyInTry() && m_lastInstr->GetDst() && m_lastInstr->GetDst()->IsRegOpnd() && m_lastInstr->GetDst()->GetStackSym()->HasByteCodeRegSlot())
+
+        if (IsLoopBodyInTry() && lastProcessedInstrForJITLoopBody != m_lastInstr)
```
LouisLaf: > lastProcessedInstrForJITLoopBody != m_lastInstr [](start = 32, length = 48)

Can this ever be false?
 lastProcessedInstrForJITLoopBody != m_lastInstr

ThomsonTan: Yes, I hit a case that one loop iteration doesn't generate any new IR instr, then added this check to make it general.

---
In reply to: [100933833](https://github.com/Microsoft/ChakraCore/pull/2518#discussion_r100933833) [](ancestors = 100933833)

lib/Backend/IRBuilder.cpp
```diff
@@ -750,28 +751,41 @@ IRBuilder::Build()
...
         {
-            StackSym * dstSym = m_lastInstr->GetDst()->GetStackSym();
-            Js::RegSlot dstRegSlot = dstSym->GetByteCodeRegSlot();
-            if (!this->RegIsTemp(dstRegSlot) && !this->RegIsConstant(dstRegSlot))
+            FOREACH_INSTR_BACKWARD_EDITING_IN_RANGE(
+                instr,
+                instrPrev,
+                m_lastInstr,
+                lastProcessedInstrForJITLoopBody ? lastProcessedInstrForJITLoopBody->m_next : m_lastInstr)
```
LouisLaf: m_lastInstr is technically wrong here as the first instr might have generated multiple.  You can initialize lastProcessedInstrForJitLoopBody to m_func->m_headInstr.

lib/Backend/IRBuilder.cpp
```diff
@@ -750,28 +751,41 @@ IRBuilder::Build()
...
         }
 #endif
-        if (IsLoopBodyInTry() && m_lastInstr->GetDst() && m_lastInstr->GetDst()->IsRegOpnd() && m_lastInstr->GetDst()->GetStackSym()->HasByteCodeRegSlot())
+
+        if (IsLoopBodyInTry() && lastProcessedInstrForJITLoopBody != m_lastInstr)
         {
-            StackSym * dstSym = m_lastInstr->GetDst()->GetStackSym();
-            Js::RegSlot dstRegSlot = dstSym->GetByteCodeRegSlot();
-            if (!this->RegIsTemp(dstRegSlot) && !this->RegIsConstant(dstRegSlot))
+            FOREACH_INSTR_BACKWARD_EDITING_IN_RANGE(
```
LouisLaf: >FOREACH_INSTR_BACKWARD_EDITING_IN_RANGE [](start = 12, length = 39)

Why backwards instead of forward?

ThomsonTan: There is `m_stSlots` to prevent reset of the same `symId`, traverse in backward so only the later store of given `symId` takes effect. Or we could get stale result if the earlier store takes effect if traverse in forward order.

---
In reply to: [100936203](https://github.com/Microsoft/ChakraCore/pull/2518#discussion_r100936203) [](ancestors = 100936203)

LouisLaf: Please add a comment to make this more obvious.

---
In reply to: [100937161](https://github.com/Microsoft/ChakraCore/pull/2518#discussion_r100937161) [](ancestors = 100937161,100936203)

lib/Backend/IRBuilder.cpp
```diff
@@ -750,28 +751,43 @@ IRBuilder::Build()
...
 #endif
-        if (IsLoopBodyInTry() && m_lastInstr->GetDst() && m_lastInstr->GetDst()->IsRegOpnd() && m_lastInstr->GetDst()->GetStackSym()->HasByteCodeRegSlot())
+
+        if (IsLoopBodyInTry() && lastProcessedInstrForJITLoopBody != m_lastInstr)
         {
-            StackSym * dstSym = m_lastInstr->GetDst()->GetStackSym();
-            Js::RegSlot dstRegSlot = dstSym->GetByteCodeRegSlot();
-            if (!this->RegIsTemp(dstRegSlot) && !this->RegIsConstant(dstRegSlot))
+            // traverse in backward so we get new/later value of given symId for storing instead of the earlier/stale
+            // symId value. m_stStores is used to prevent multiple stores to the same symId.
```
rajatd: >m_stStores [](start = 28, length = 10)

it should be m_stSlots

LouisLaf:
:shipit:

PR #2519, opened 2017-02-13
Rethrow of exception in TryFinally can lead to running out of stack space on a Javascript StackOverflow Exception, redesign TryFinally to avoid rethrow

meg-gupta:


meg-gupta:
@LouisLaf , @curtisman Can you please review ?

lib/Common/CommonPal.h
```diff
@@ -611,19 +611,20 @@ HRESULT ULongMult(ULONG ulMultiplicand, ULONG ulMultiplier, ULONG* pulResult);
...
-    finallyFunc(hasException);
+    public:
+        FinallyObject(const FinallyFunc& finallyFunc) : finallyFunc(finallyFunc), abnormalTermination(true) {}
+        ~FinallyObject() { finallyFunc(abnormalTermination); }
+
+        void SetHasNoAbnormalTermination() { abnormalTermination = false; }
+    private:
+        const FinallyFunc& finallyFunc;
+        bool abnormalTermination;
+    } finallyObject(finallyFunc);
```
ThomsonTan: >finallyObject(finallyFunc); [](start = 6, length = 27)

Will it run correctly if allocate finallyObject on the stack triggered stackoverflow?

lib/Common/CommonPal.h
```diff
@@ -611,19 +611,20 @@ HRESULT ULongMult(ULONG ulMultiplicand, ULONG ulMultiplier, ULONG* pulResult);
...
-
-    finallyFunc(hasException);
+    public:
+        FinallyObject(const FinallyFunc& finallyFunc) : finallyFunc(finallyFunc), abnormalTermination(true) {}
+        ~FinallyObject() { finallyFunc(abnormalTermination); }
+
+        void SetHasNoAbnormalTermination() { abnormalTermination = false; }
+    private:
+        const FinallyFunc& finallyFunc;
+        bool abnormalTermination;
```
ThomsonTan: >abnormalTermination [](start = 13, length = 19)

We probably could specify the default value of `abnormalTermination` as `true` here instead of put it in the initialization list (supported by c++11).

lib/Common/CommonPal.h
```diff
@@ -611,19 +611,20 @@ HRESULT ULongMult(ULONG ulMultiplicand, ULONG ulMultiplier, ULONG* pulResult);
 template <class TryFunc, class FinallyFunc>
 void TryFinally(const TryFunc& tryFunc, const FinallyFunc& finallyFunc)
 {
-    bool hasException = true;
-    try
+    class FinallyObject
```
ThomsonTan: >class FinallyObject [](start = 4, length = 19)

Still wondering why rethrow in catch can cause stackoverflow and how could destructor avoid that?

meg-gupta: Quoting @LouisLaf's explanation from an internal discussion -  Rethrow causes a new instead of the CRT/Windows code to raise and exception, and the code to then process it.  Both save copies of the context record, which is over 1k each.  If the exception we are trying to process is a stack overflow, we’re in trouble quickly if we have a few of the TryFinally.

LouisLaf: Catch is called before the stack is unwound.  Rethrow costs a lot of stack space.  WIth the Dtors, you don't consume extra stack space when you have many of these TryFinally nested.

---
In reply to: [100933897](https://github.com/Microsoft/ChakraCore/pull/2519#discussion_r100933897) [](ancestors = 100933897)

lib/Common/CommonPal.h
LouisLaf: No, it won't if the dtor was called because of an exception.
Meghana, please add a big warning comment that the finallyFunc isn't allowed to trigger any exceptions.  This also means it can't probe the stack or allocate memory...

---
In reply to: [100933437](https://github.com/Microsoft/ChakraCore/pull/2519#discussion_r100933437) [](ancestors = 100933437)

LouisLaf:
:shipit:

PR #2520, opened 2017-02-14
Fix byte code gen for function-in-block in eval.

pleath:
An optimization to permit redeferral of function-in-block was missing a legality check in eval code.


pleath:
@Cellule This is the fix for the issue we were looking at together.

Cellule:
@pleath oh, I didn't think to look if that source was coming from an eval!
So basically we were not preventing eval code from being redeffered if I understand correctly.
LGTM

pleath:
Thanks. The presence of eval was not at all obvious, because the eval helper was not on the call stack. The enclosing function had been instantiated in eval and invoked later.

PR #2521, opened 2017-02-14
swb: wasm write barrier fixes

jianchun:
This change fixes wasm related write barrier issues (enabled/introduced
recently).

GrowingArray: Change SWB annotations and use CopyArray.

ArenaAllocator: Tag cacheBlockEnd to avoid GC false positive. Refactor
related code to make TaggedPointer (renamed from GC_MARKED_OBJECT)
available.

HeapBlock::PrintVerifyMarkFailure: Fix previously missed addresses and
ignore target internal address false positives.

Wasm: Add annotations, and specifically remove bad casts that bypass SWB
(Get/SetVarElement).

runtests.py: Decode "byte" output so that when printed to console "\n"
works.

### DIFF
--- a/lib/Common/Memory/RecyclerPointers.h
+++ b/lib/Common/Memory/RecyclerPointers.h
@@ -4,6 +4,8 @@
 //-------------------------------------------------------------------------------------------------------
 #pragma once
 
+#include "WriteBarrierMacros.h"
+
 namespace Memory
 {
 class Recycler;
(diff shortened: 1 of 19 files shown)


lib/Common/Memory/RecyclerPointers.h
```diff
@@ -596,4 +598,35 @@ void *  __cdecl memset(_Out_writes_bytes_all_(_Size) WriteBarrierPtr<T> * _Dst,
     CompileAssert(false);
 }
 
-#include <Memory/WriteBarrierMacros.h>
\ No newline at end of file
+
+// This class abstract a pointer value with its last 2 bits set to avoid conservative GC tracking.
+template <class T>
+class GC_MARKED_OBJECT
```
leirocks: change the name to be taggedpointer? the mark has a strict definition in recycler which makes it confusing to use this name

jianchun: Agree, changed

jianchun:

Issue #2517, opened 2017-02-11
Exit code -11 causing xplat tests to fail intermittently


dilijev:
Example:

https://ci.dot.net/job/Microsoft_ChakraCore/job/master/job/osx_osx_debug_static_prtest/549/consoleFull#-1719471233c7453dd6-bf4f-4ae9-9fca-50f9ec7ab19e

Snippet:

```
[1261/1973 3.06] Failed -> WasmSpec/spec.js                           
14:45:31 /Users/dotnet-bot/j/workspace/Microsoft_ChakraCore/master/osx_osx_debug_static_prtest/out/Debug/ch -WERExceptionSupport -ExtendedErrorStackForTestHost -BaselineMode -maxInterpretCount:1 -maxSimpleJitRunCount:1 -bgjit- -dynamicprofilecache:profile.dpl.2492 -wasm -args testsuite-bin/f64.json -endargs /Users/dotnet-bot/j/workspace/Microsoft_ChakraCore/master/osx_osx_debug_static_prtest/test/WasmSpec/spec.js
14:45:31 
14:45:31 Output:
14:45:31 ----------------------------
14:45:31 2395/2395 tests passed.
14:45:31 
14:45:31 ----------------------------
14:45:31 exit code: -11
```

(Test output is correct, it's only the exit code that's a problem.)

Tor reference -11 = 0xFFFFFFF5, but I haven't gone looking for the error code yet.

jianchun:
Not sure if relevant, we found wasm code has software write barrier bugs. I'm working on fixing those issues.

jianchun:
> I'm not sure I understand the issue. ScriptContext close shouldn't release any memory, and I don't understand why this would cause AV. It will decommit entry points, but not free them. I can believe that ours aren't getting cleaned up at that point. However, they still should be freed once there are no more recycler references to the FunctionBody.

@MikeHolman There is a release order dependency. FunctionEntryPointInfo cleanup needs to deregister debug data (only on xplat). Debug data exists in native code generator managed memory. ScriptContext close cleans up known FunctionBodies/EntryPointInfos then destroys native code generator. In your case the order doesn't guarantee -- When cleaning up your FunctionEntryPointInfo from finalizer, script context already closed and native code generator destroyed, but the FunctionEntryPointInfo cleanup still needs to read the (native code generator) released memory.

MikeHolman:
[...]

PR #2524, opened 2017-02-14
fix issues creating Utf8SourceInfo for wasm

MikeHolman:
Utf8SourceInfo::New makes a copy of srcinfo and it wasn't the correct size for wasm. also, we should be allocating our copy of the buffer as leaf to prevent false references

### DIFF
--- a/lib/Runtime/Library/WebAssemblyModule.cpp
+++ b/lib/Runtime/Library/WebAssemblyModule.cpp
@@ -219,11 +219,12 @@ WebAssemblyModule::CreateModule(
     {
         Js::AutoDynamicCodeReference dynamicFunctionReference(scriptContext);
         SRCINFO const * srcInfo = scriptContext->cache->noContextGlobalSourceInfo;
-        Js::Utf8SourceInfo* utf8SourceInfo = Utf8SourceInfo::New(scriptContext, (LPCUTF8)buffer, lengthBytes / sizeof(char16), lengthBytes, srcInfo, false);
 
         // copy buffer so external changes to it don't cause issues when defer parsing
-        byte* newBuffer = RecyclerNewArray(scriptContext->GetRecycler(), byte, lengthBytes);
+        byte* newBuffer = RecyclerNewArrayLeaf(scriptContext->GetRecycler(), byte, lengthBytes);
         js_memcpy_s(newBuffer, lengthBytes, buffer, lengthBytes);
+
+        Js::Utf8SourceInfo* utf8SourceInfo = Utf8SourceInfo::NewWithNoCopy(scriptContext, (LPCUTF8)newBuffer, lengthBytes / sizeof(char16), lengthBytes, srcInfo, false);
 
         Wasm::WasmModuleGenerator bytecodeGen(scriptContext, utf8SourceInfo, newBuffer, lengthBytes);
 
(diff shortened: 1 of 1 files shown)


lib/Runtime/Library/WebAssemblyModule.cpp
```diff
@@ -219,12 +219,13 @@ WebAssemblyModule::CreateModule(
...
         Js::AutoDynamicCodeReference dynamicFunctionReference(scriptContext);
         SRCINFO const * srcInfo = scriptContext->cache->noContextGlobalSourceInfo;
-        Js::Utf8SourceInfo* utf8SourceInfo = Utf8SourceInfo::New(scriptContext, (LPCUTF8)buffer, lengthBytes / sizeof(char16), lengthBytes, srcInfo, false);
 
         // copy buffer so external changes to it don't cause issues when defer parsing
-        byte* newBuffer = RecyclerNewArray(scriptContext->GetRecycler(), byte, lengthBytes);
+        byte* newBuffer = RecyclerNewArrayLeaf(scriptContext->GetRecycler(), byte, lengthBytes);
         js_memcpy_s(newBuffer, lengthBytes, buffer, lengthBytes);
 
+        Js::Utf8SourceInfo* utf8SourceInfo = Utf8SourceInfo::NewWithNoCopy(scriptContext, (LPCUTF8)newBuffer, lengthBytes / sizeof(char16), lengthBytes, srcInfo, false);
```
Cellule: I checked a little closer at the `Utf8SourceInfo` constructors and I am wondering now if we should use `Utf8SourceInfo::New` and then get its copy of the buffer instead of making the copy ourselves.

lib/Runtime/Library/WebAssemblyModule.cpp
```diff
@@ -281,9 +282,14 @@ WebAssemblyModule::ValidateModule(
     {
         Js::AutoDynamicCodeReference dynamicFunctionReference(scriptContext);
         SRCINFO const * srcInfo = scriptContext->cache->noContextGlobalSourceInfo;
-        Js::Utf8SourceInfo* utf8SourceInfo = Utf8SourceInfo::New(scriptContext, (LPCUTF8)buffer, lengthBytes / sizeof(char16), lengthBytes, srcInfo, false);
 
-        Wasm::WasmModuleGenerator bytecodeGen(scriptContext, utf8SourceInfo, (byte*)buffer, lengthBytes);
+        // review: unsure if we need copy here, but seems safer to do it
```
Cellule: Since the created module doesn't escape this scope, it should be garbage collected asap and there should be no further references to the buffer after this call. Therefore, we don't need to get a copy of the buffer, simply validate it then trash all references to it.

Cellule:
Does this need to be in release/1.4 ? Is there a bug associated ?

jianchun:
@MikeHolman Piggyback for #2517. I was investigating that wasm crash at shutdown on Linux, and now suspect the cause is broken link in wasm created Utf8SourceInfo. You do not have a call of "ScriptContext::SaveSourceNoCopy()" anywhere, thus your utf8SourceInfo isn't in scriptContext->sourceList, and scriptContext close will not see your utf8SourceInfo or the FunctionBodies inside and won't clean up them. ScriptContext close deletes nativeCodeGenerator and releases the memory. Later on wasm generated FunctionBody/EntryPoint finalizer needs to access that memory to deregister debug data. This results in AV. Does this make sense?

lib/Runtime/Library/WebAssemblyModule.cpp
MikeHolman: But I don't know about if the lifetime of the UTF8SourceInfo is different. It has a finalizer and the actual ArrayBuffer buffer isn't recycler allocated, so a reference to that wouldn't keep it alive.

MikeHolman:
I'm not sure I understand the issue. ScriptContext close shouldn't release any memory, and I don't understand why this would cause AV. It will decommit entry points, but not free them. I can believe that ours aren't getting cleaned up at that point. However, they still should be freed once there are no more recycler references to the FunctionBody.

Cellule:
LGTM

PR #2525, opened 2017-02-14
port two fixes back to github

leirocks:
1. Flush correct code buffer in case of OOP JIT.
after section change, there are 3 buffers, source buffer, local buffer, remote buffer. the local buffer and remote buffer are pointing to same address in kernel, but have different mapping in client and server process. before the section
change don't have 3 buffers, we were using WriteProcessMemory, so local buffer above is actually using address from client process

2. Fix a hang
For some reason ChakraJIT.acf is not compiled with midl in razzle build(it built correctly in VSO), the context creating server call is serialized, which waits for all on-going codegen call to complete, hence blocks the UI thread and resulting a hang


dilijev:
@leirocks Thanks!

PR #2528, opened 2017-02-14
Fixes Bug#10330888 and duplicates. Fix Block ID accounting. Do not eat left parentheses when parsing destructured var declarations.

atulkatti:


atulkatti:
@Microsoft/chakra-developers 
This has already been reviewed in codeflow.

PR #2529, opened 2017-02-15
OS10895818: F12 edge crash when debugging google.com

suwc:
For global function inside direct eval, FuncAllowsDirectSuper() looks up JavaScript caller through stackwalker to determine if the eval is called from a class constructor. Need to check if functionBody exists to cover the case where eval is called by F12 from inside a built-in function.

### DIFF
--- a/lib/Runtime/ByteCode/ByteCodeGenerator.cpp
+++ b/lib/Runtime/ByteCode/ByteCodeGenerator.cpp
@@ -2585,7 +2585,7 @@ bool FuncAllowsDirectSuper(FuncInfo *funcInfo, ByteCodeGenerator *byteCodeGenera
     if (funcInfo->IsGlobalFunction() && ((byteCodeGenerator->GetFlags() & fscrEval) != 0))
     {
         Js::JavascriptFunction *caller = nullptr;
-        if (Js::JavascriptStackWalker::GetCaller(&caller, byteCodeGenerator->GetScriptContext()))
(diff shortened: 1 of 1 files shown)


suwc:
@curtisman @pleath @Microsoft/chakra-es review please

lib/Runtime/ByteCode/ByteCodeGenerator.cpp
```diff
@@ -2585,7 +2585,7 @@ bool FuncAllowsDirectSuper(FuncInfo *funcInfo, ByteCodeGenerator *byteCodeGenera
     if (funcInfo->IsGlobalFunction() && ((byteCodeGenerator->GetFlags() & fscrEval) != 0))
     {
         Js::JavascriptFunction *caller = nullptr;
-        if (Js::JavascriptStackWalker::GetCaller(&caller, byteCodeGenerator->GetScriptContext()))
+        if (Js::JavascriptStackWalker::GetCaller(&caller, byteCodeGenerator->GetScriptContext()) && Js::ScriptFunction::Is(caller))
         {
```
agarwal-sandeep: Can we add a test case in test\DebuggerCommon?

suwc: I haven't been able to repro the issue with ch.exe/jshost.exe

pleath:
LGTM. Thanks.

lib/Runtime/ByteCode/ByteCodeGenerator.cpp
agarwal-sandeep: I can help you with that

PR #2534, opened 2017-02-15
Repair incorrect fix for double increment of literal regex count. 

pleath:
On redeferral of an enclosing function, where nested functions are still in a compiled state, we do not want to re-increment the literal regex count for the nested functions. But we do still want to count the regexes and allocate new regex storage for the function we're recompiling. A new simple fix handles both cases.


pleath:
@dilijev This is the one you were interested in.

dilijev:
@pleath Fix looks reasonable to me. Thanks!

PR #2535, opened 2017-02-15
increase wait time before rpc failfast happens

MikeHolman:


lib/Backend/NativeCodeGenerator.cpp
```diff
@@ -3673,7 +3673,7 @@ JITManager::HandleServerCallResult(HRESULT hr, RemoteCallType callType)
 
     // we should not have RPC failure if JIT process is still around
     // since this is going to be a failfast, lets wait a bit in case server is in process of terminating
-    if (WaitForSingleObject(GetJITManager()->GetServerHandle(), 250) != WAIT_OBJECT_0)
+    if (WaitForSingleObject(GetJITManager()->GetServerHandle(), 1000) != WAIT_OBJECT_0)
```
agarwal-sandeep: How frequently are we seeing this fail fast?

In any case since we will fail fast should this be a bit more higher (say 2-5 sec). Also can you have it as a #define instead of a hardcoded number

MikeHolman: We are seeing it pretty rarely, but still hitting sometimes. I guess we can make it even higher, though. I'll bump up to 3 seconds and put behind #define.

PR #2538, opened 2017-02-16
Fix operand types after hoisting IR instrs to landingpad

ThomsonTan:
We don't fix value type of many IR instructions when hositing it into landingpad (like for `CheckFixFld`, we did it for `FromVar`). Then incorrect value type info in landingpad causes bad codegen, like missing some type checks required by the type info in landingpad (some type assertions in the loop are false in the landingpad because they are from bailout in the loop) , and the missing checked causes access violation finally.

I got this problem from crawler. The fix and simple repro was mostly made by Louis. Thanks @LouisLaf.

### DIFF
--- a/lib/Backend/GlobOpt.cpp
+++ b/lib/Backend/GlobOpt.cpp
@@ -19060,9 +19060,15 @@ GlobOpt::OptDstIsInvariant(IR::RegOpnd *dst)
 void
 GlobOpt::OptHoistToLandingPadUpdateValueType(
     BasicBlock* landingPad,
+    IR::Instr* instr,
     IR::Opnd* opnd,
     Value* opndVal)
 {
+    if (instr->m_opcode == Js::OpCode::FromVar)
+    {
+        return;
+    }
+
     Sym* opndSym = opnd->GetSym();;
 
     if (opndSym)
@@ -19106,7 +19112,7 @@ GlobOpt::OptHoistInvariant(
     if (src1)
     {
         // We are hoisting this instruction possibly past other uses, which might invalidate the last use info. Clear it.
-        OptHoistToLandingPadUpdateValueType(landingPad, src1, src1Val);
+        OptHoistToLandingPadUpdateValueType(landingPad, instr, src1, src1Val);
 
         if (src1->IsRegOpnd())
         {
@@ -19116,7 +19122,7 @@ GlobOpt::OptHoistInvariant(
         IR::Opnd* src2 = instr->GetSrc2();
         if (src2)
         {
-            OptHoistToLandingPadUpdateValueType(landingPad, src2, nullptr);
+            OptHoistToLandingPadUpdateValueType(landingPad, instr, src2, nullptr);
 
             if (src2->IsRegOpnd())
             {
--- a/lib/Backend/GlobOpt.cpp
+++ b/lib/Backend/GlobOpt.cpp
@@ -19058,6 +19058,38 @@ GlobOpt::OptDstIsInvariant(IR::RegOpnd *dst)
 }
 
 void
+GlobOpt::OptHoistToLandingPadUpdateValueType(
+    BasicBlock* landingPad,
+    IR::Opnd* opnd,
+    Value* opndVal)
+{
+    Sym* opndSym = opnd->GetSym();;
+
+    if (opndSym)
+    {
+        if (opndVal == nullptr)
+        {
+            opndVal = FindValue(opndSym);
+        }
+
+        Value* opndValueInLandingPad = FindValue(landingPad->globOptData.symToValueMap, opndSym);
+        Assert(opndVal->GetValueNumber() == opndValueInLandingPad->GetValueNumber());
+
+        opnd->SetValueType(opndValueInLandingPad->GetValueInfo()->Type());
+
+        if (opndSym->IsPropertySym())
+        {
+            // Also fix valueInfo on objPtr
+            StackSym* opndObjPtrSym = opndSym->AsPropertySym()->m_stackSym;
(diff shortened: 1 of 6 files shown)


ThomsonTan:
@LouisLaf @rajatd please help review

lib/Backend/GlobOpt.cpp
```diff
@@ -19069,6 +19107,30 @@ GlobOpt::OptHoistInvariant(
     IR::BailOutKind bailoutKind)
 {
     BasicBlock *landingPad = loop->landingPad;
+
+    IR::Opnd* src1 = instr->GetSrc1();
```
rajatd: Why do you need to move this block of code at the beginning of the method?

lib/Backend/GlobOpt.cpp
```diff
@@ -19058,6 +19058,44 @@ GlobOpt::OptDstIsInvariant(IR::RegOpnd *dst)
 }
 
 void
+GlobOpt::OptHoistToLandingPadUpdateValueType(
+    BasicBlock* landingPad,
+    IR::Instr* instr,
+    IR::Opnd* opnd,
+    Value* opndVal)
+{
+    if (instr->m_opcode == Js::OpCode::FromVar)
```
rajatd: Is the reason for not updating the value types in case of a FromVar that there is FromVar specific code in OptHoistInvariant that needs to know if the value types were different? If so, would you still need this FromVar check if you call this method after adding the bailout to FromVar?

lib/Backend/GlobOpt.cpp
```diff
@@ -19069,6 +19107,30 @@ GlobOpt::OptHoistInvariant(
     IR::BailOutKind bailoutKind)
 {
     BasicBlock *landingPad = loop->landingPad;
+
+    IR::Opnd* src1 = instr->GetSrc1();
+    if (src1)
+    {
+        // We are hoisting this instruction possibly past other uses, which might invalidate the last use info. Clear it.
+        OptHoistToLandingPadUpdateValueType(landingPad, instr, src1, src1Val);
```
rajatd: Actually, how does updating the value type on the operand in the hoisted instruction help? The hoisted instr won't be processed by the forward pass, correct? Is it that the deadstore pass takes some unwanted action if the value type is left unchanged?

lib/Backend/GlobOpt.cpp
ThomsonTan: Moving this to the beginning of method is for cleanup from Louis.

---
In reply to: [101555390](https://github.com/Microsoft/ChakraCore/pull/2538#discussion_r101555390) [](ancestors = 101555390)

lib/Backend/GlobOpt.cpp
ThomsonTan: The incorrect value type (if no update) for hoisted instruction could lead incorrect codegen, like missing some type checks which is required for the landingpad. 

---
In reply to: [101557970](https://github.com/Microsoft/ChakraCore/pull/2538#discussion_r101557970) [](ancestors = 101557970)

lib/Backend/GlobOpt.cpp
ThomsonTan: Updating type for `FromVar` causes some problem in lowerer because it has some type restrictions, such as the source cannot be a definite type. This could be violated if we update it to landingpad. Better leave it to the code in `OptHoistInvariant` to handle it.

---
In reply to: [101556159](https://github.com/Microsoft/ChakraCore/pull/2538#discussion_r101556159) [](ancestors = 101556159)

LouisLaf: I don't understand this...  We copied the FromVar code here.  What is the difference?

---
In reply to: [101593051](https://github.com/Microsoft/ChakraCore/pull/2538#discussion_r101593051) [](ancestors = 101593051,101556159)

lib/Backend/GlobOpt.cpp
LouisLaf: It's the lowering that looked at the valueType and incorrectly omitted tag checks....

---
In reply to: [101592247](https://github.com/Microsoft/ChakraCore/pull/2538#discussion_r101592247) [](ancestors = 101592247,101557970)

lib/Backend/GlobOpt.cpp
ThomsonTan: I don't see the source of the copy, where is it?

---
In reply to: [101616597](https://github.com/Microsoft/ChakraCore/pull/2538#discussion_r101616597) [](ancestors = 101616597,101593051,101556159)

lib/Backend/GlobOpt.cpp
rajatd: Ok. Thanks @louislaf

ThomsonTan:
@LouisLaf, the other code (see below) in `OptHoistInvariant` does the below check for `FromVar`. Seems leave the extra handling code here is fine, or we need dup these checks in the new function `OptHoistToLandingPadUpdateValueType` which seems longer than the type hoisting code we save (2 lines).

```
        case Js::OpCode::FromVar:
        {
            StackSym* src1StackSym = IR::RegOpnd::TryGetStackSym(instr->GetSrc1());

            if (instr->HasBailOutInfo())
            {
                IR::BailOutKind instrBailoutKind = instr->GetBailOutKind();
                Assert(instrBailoutKind == IR::BailOutIntOnly ||
                    instrBailoutKind == IR::BailOutExpectingInteger ||
                    instrBailoutKind == IR::BailOutOnNotPrimitive ||
                    instrBailoutKind == IR::BailOutNumberOnly ||
                    instrBailoutKind == IR::BailOutPrimitiveButString ||
                    instrBailoutKind == IR::BailOutSimd128F4Only ||
                    instrBailoutKind == IR::BailOutSimd128I4Only);
            }
            else if (src1StackSym && bailoutKind != IR::BailOutInvalid)
            {
```

Issue #2536, opened 2017-02-15
Array destructuring throws an error


dustinsoftware:
This throws an error: Destructuring expressions can only have identifier references
```
class Foo {
	flip() {
		[this.indexA, this.indexB] = [this.indexB, this.indexA];
	}
}
```

This works:
```
class Foo {
	flip() {
		let { indexA, indexB } = this;
		[indexA, indexB] = [this.indexB, this.indexA];
		this.indexA = indexA;
		this.indexB = indexB;
	}
}
```

Microsoft Edge 38.14393.0.0
Confirmed working in Firefox and Chrome

ianwjhalliday:
Note this repros in any context, not just class methods.  Minimal repro at global scope:
```javascript
[this.indexA, this.indexB] = [this.indexB, this.indexA];
```

akroshg:
below code works
```js
    var obj = {x:1, y:2};
    [obj.x, obj.y] = [obj.y, obj.x];
```
but this does not.
```js
    [this.x, this.y]  = [this.y, this.x];
```

akroshg:
ah, found it. 
@dustinsoftware thanks for reporting this. Fix is on the way.

akroshg:
This is merged.

PR #2539, opened 2017-02-16
this is referenced in destructuring.

akroshg:
tkTHIS was not handled when we allow the reference for destructuring expression. Fixed that.


akroshg:
@aneeshdk please have a look.

akroshg:
fixes #2536

aneeshdk:
:shipit:

PR #2543, opened 2017-02-16
Fix thread-unsafe modification of literal regex counter. 

pleath:
A fully compiled function nested within a function that we're recompiling should not have its literal regex counter and storage reset. We won't regenerate byte code, so we won't use the new counter value.


pleath:
@leirocks Can you look at this (very small) one?

PR #2544, opened 2017-02-16
Don't attempt to get redeferred FunctionBody from JavascriptExceptionObject

pleath:


akroshg:
How a function body gets redeferred when it throws exception just now. will it not be part of the active functions set?

pleath:
The frame it was executing in has been unwound, apparently, at the time redeferral occurs.

akroshg:
alright - LGTM.

---
In reply to: [280493925](https://github.com/Microsoft/ChakraCore/pull/2544#issuecomment-280493925) [](ancestors = 280493925)

Issue #2481, opened 2017-02-07
Crash when yielding variable in TDZ in CatchParameter default expression


anba:
Version: cb7817bda7e4b265e7a7364641123e20204facfd
Platform: Ubuntu x64

Test case:
```js
var it = function*() {
  try {
    throw [void 0];
  } catch ([c = (yield c)]) {
  }
}();

it.next();
```

Stack trace from gdb:
```
0x00007ffff3117438 in Memory::RecyclerRootPtr<ThreadContext::RecyclableData>::operator-> (
    this=0x11b8) at /home/andre/git/ChakraCore/lib/Common/Memory/RecyclerRootPtr.h:18
18	    T * operator->() const { Assert(ptr != nullptr); return ptr; }
(gdb) bt
#0  0x00007ffff3117438 in Memory::RecyclerRootPtr<ThreadContext::RecyclableData>::operator-> (
    this=0x11b8) at /home/andre/git/ChakraCore/lib/Common/Memory/RecyclerRootPtr.h:18
#1  0x00007ffff3d3c01a in ThreadContext::SaveTempUncaughtException (this=0x0, 
    exceptionObject=0x7fffefc9c000)
    at /home/andre/git/ChakraCore/lib/Runtime/./Base/ThreadContext.h:1359
#2  0x00007ffff3d3668a in Js::JavascriptExceptionOperators::DoThrow (
    exceptionObject=0x7fffefc9c000, scriptContext=0x7fffefd80000)
    at /home/andre/git/ChakraCore/lib/Runtime/Language/JavascriptExceptionOperators.cpp:873
#3  0x00007ffff3d6b1cd in Js::JavascriptOperators::OP_ResumeYield (yieldData=0x7fffefd51680, 
    iterator=0x0) at /home/andre/git/ChakraCore/lib/Runtime/Language/JavascriptOperators.cpp:9785
#4  0x00007ffff3cf0cfc in Js::InterpreterStackFrame::OP_ResumeYield (this=0x7fffefd0c000, 
    yieldDataVar=0x7fffefd51680, yieldStarIterator=4294967295)
...
```

Penguinwizzard:
This test case leads to the same crash on windows chakracore. The crash is immediately due to the threadContext reference on the scriptContext is null, but is needed for the exception handling.

akroshg:
@anba thanks for reporting this. I understood the issue. The fix is coming soon.

akroshg:
The fix is merged.

PR #2547, opened 2017-02-16
Yield at catch param binding pattern.

akroshg:
The yielding at catch param place gone wrong when there is destructuring. It had happened because the we were not doing correct book-keeping for catch block for the pattern. Fixed that by recording just before assignment.


akroshg:
@aneeshdk please have a look.

akroshg:
Fixes #2481

aneeshdk:
LGTM

PR #2550, opened 2017-02-17
Fix byte code gen with block-scoped function called from with and captured elsewhere.

pleath:
Make sure that if some references to a block-scoped function are optimized to refer to the var-scoped binding and others are not, and both bindings are marked for closure capture, then both scopes are processed accordingly.


pleath:
@boingoing Very simple byte code gen change.

PR #2555, opened 2017-02-17
Remove legacy parenthesized label parsing; Remove redundant PnLabel parse tree node and handling

ianwjhalliday:
This PR includes two changes:
1. Remove legacy parenthesized label parsing, follow up to #2115 
2. Simplify label parsing by removing the label parse tree node type and corresponding bifurcation on the buildAST mode

In the latter change, use the NoAST label tracking logic for both NoAST and buildAST modes.  The label parse tree nodes were unused after parsing, so no need to create and keep them.

This is a post v1.4 change.


lib/Parser/Parse.h
```diff
@@ -472,29 +472,23 @@ class Parser
     bool NextTokenIsPropertyNameStart() const { return m_token.tk == tkID || m_token.tk == tkStrCon || m_token.tk == tkIntCon || m_token.tk == tkFltCon || m_token.tk == tkLBrack || m_token.IsReservedWord(); }
 
     template<bool buildAST>
-    void PushStmt(StmtNest *pStmt, ParseNodePtr pnode, OpCode op, ParseNodePtr pnodeLab, LabelId* pLabelIdList)
+    void PushStmt(StmtNest *pStmt, ParseNodePtr pnode, OpCode op, LabelId* pLabelIdList)
     {
-        AssertMem(pStmt);
```
suwc: >AssertMem(pStmt); [](start = 8, length = 17)

Remind me why this has to go too?

ianwjhalliday: Asserting a pointer is not null before using it is redundant, since it will AV anyway.  You could say perhaps we gain benefit by asserting earlier, but I have never seen any benefit to such asserts in practice.

I actually have another change pending that removes all the legacy assert macro usage.

tcare:
Looks good :)

lib/Parser/Parse.h
dilijev: @ianwjhalliday Looking forward to your first external PR to remove legacy assert macros :)

test/Basics/label1.js
```diff
@@ -4,7 +4,7 @@
 //-------------------------------------------------------------------------------------------------------
 
 function f() {
-    // Label in parenthesis is bad syntax, but we allow it. Verify consistency in deferred parsing.
+    // Label in parenthesis is bad syntax. Verify consistency in deferred parsing.
```
dilijev: If this syntax is now disallowed, shouldn't we have a parse error, or would this be something that now throws at runtime, so it's okay to have it in a defer-parsed function?

test/Basics/rlexe.xml
```diff
@@ -73,14 +73,17 @@
...
+      <baseline>label1.baseline</baseline>
       <compile-flags>-Intl-</compile-flags>
+      <tags>exclude_dynapogo</tags>
     </default>
   </test>
   <test>
     <default>
       <files>label1.js</files>
+      <baseline>label1.baseline</baseline>
       <compile-flags>-Force:Deferparse -Intl-</compile-flags>
```
dilijev: Just curious, why is  `-Intl-` necessary for this test?

tcare: This is from the old days of Intl. I don't think it's needed now.

dilijev:
@ianwjhalliday if you have a minute, could you make the bot happy?

PR #2564, opened 2017-02-18
disable a write barrier verifier false positive

jianchun:
In some cases ScriptFunctionType::entryPoint is outdated but uncleared,
pointing to memory reused by recycler. This may cause noise in write barrier
verifier. Filter it for now.

### DIFF
--- a/lib/Common/Memory/HeapBlock.cpp
+++ b/lib/Common/Memory/HeapBlock.cpp
@@ -930,6 +930,20 @@ void HeapBlock::PrintVerifyMarkFailure(Recycler* recycler, char* objectAddress, 
                     return;
                 }
 
+                // Js::Type::entryPoint may contain outdated data uncleared, and reused by recycler
+                // Most often occurs with script function Type
+                if (offset ==
+#if TARGET_64
+                    0x18
+#else
+                    0x10
+#endif
+                    && strstr(typeName, "Js::ScriptFunctionType"))
+                {
+                    dumpFalsePositive();
+                    return;
+                }
+
                 //TODO: (leish)(swb) analyze pdb to check if the field is a pointer field or not
                 Output::Print(_u("Missing Barrier\nOn type %S+0x%x\n"), typeName, offset);
             }


obastemur:
```
18:02:18 ERROR: Connection was broken: java.io.IOException: Unexpected termination of the channel
18:02:18 	at hudson.remoting.SynchronousCommandTransport$ReaderThread.run(SynchronousCommandTransport.java:73)
18:02:18 Caused by: java.io.EOFException
18:02:18 	at java.io.ObjectInputStream$PeekInputStream.readFully(ObjectInputStream.java:2624)
18:02:18 	at java.io.ObjectInputStream$BlockDataInputStream.readShort(ObjectInputStream.java:3099)
18:02:18 	at java.io.ObjectInputStream.readStreamHeader(ObjectInputStream.java:853)
18:02:18 	at java.io.ObjectInputStream.<init>(ObjectInputStream.java:349)
18:02:18 	at hudson.remoting.ObjectInputStreamEx.<init>(ObjectInputStreamEx.java:48)
18:02:18 	at hudson.remoting.AbstractSynchronousByteArrayCommandTransport.read(AbstractSynchronousByteArrayCommandTransport.java:34)
18:02:18 	at hudson.remoting.SynchronousCommandTransport$ReaderThread.run(SynchronousCommandTransport.java:59)
18:02:18 
```
https://ci.dot.net/job/Microsoft_ChakraCore/job/master/job/ubuntu_linux_test_static_prtest/679/ 

@mmitche is this known?

lib/Common/Memory/HeapBlock.cpp
```diff
@@ -930,6 +930,20 @@ void HeapBlock::PrintVerifyMarkFailure(Recycler* recycler, char* objectAddress,
                     return;
                 }
 
+                // Js::Type::entryPoint may contain outdated data uncleared, and reused by recycler
+                // Most often occurs with script function Type
+                if (offset ==
+#if TARGET_64
+                    0x18
```
digitalinfinity: Can we not use the offsetof macro here instead of hardcoding the offset? type->entrypoint is generally safe to take a dependency on since its offset hasn't moved since Chakra started but it'd still be better to not hardcode this here. Also, I'd take a dependency on specific platforms rather than just x64 vs everything else since new compilers could do strange things 😄

jianchun: Thanks for review! Did not use "offsetof" to avoid dependency noise (same reason of a bunch of this pattern in existing code). This is in "memory" lib which so far only includes basic core and memory headers. Will need to introduce runtime headers to use "offsetof" -- will consider. This is for debug and probably temporary as ideally we'd fix such false positives instead of filtering them.

jianchun:

PR #2569, opened 2017-02-21
Disabling stackoverflow checks (interrupts) in loops by default in ch.exe

Krovatkin:


dilijev:
ch.exe is not part of the ChakraCore.dll product, so as long as this causes no problems in ch.exe, I approve this -- since we use this to run the tests, and all tests pass, it seems fine.

Curious though, since ch.exe is not the product itself and its running time does not affect the intrinsic performance of ChakraCore, why bother?

curtisman:
I also wonder the reason behind this change.  Can we just expose a switch to disable it, if it isn't there already?

Krovatkin:
@curtisman @dilijev 

Indeed, there's a switch (`off:interruptprobe`) to turn loop probes off. We felt that ch.exe was probably the most **convenient** and **accessible** way to compare the **latest/greatest** ChakraCore against other JS engines and it seemed a bit *suboptimal* (especially, for smaller loops doing little work per iteration) to keep the interrupts on by *default* given that there didn't seem to be **any** logic that actually uses them. We decided to bring it up to see what you guys think.

curtisman:
@Krovatkin 
Yeah,  I am not totally against it per se.  But here are the reasons that I am hesitant, and want to keep it the way it is:
1. Node-chakracore current uses this mode
2. This provide testing coverage with interrupt probe on, since our internal Edge testing have it off.

@LouisLaf @digitalinfinity Any thoughts?

Krovatkin:
@curtisman 

> This provide testing coverage with interrupt probe on, since our internal Edge testing have it off.


> Node-chakracore current uses this mode

Maybe, we could consider adding an "interrupt" mode running a subset of unit tests or would that be too much overhead on CI? 

Regardless, thank you very much for pointing all these use cases out!

LouisLaf:
I'd rather see it off by default in ch.exe, as it most closely match Edge and other normal usage.  We can add a test which forces on.

Krovatkin:
@dilijev @curtisman 

Did you guys get a chance to reach a consensus on this item?
Should I close it or should I add another mode for running w/ loop interrupts on?

curtisman:
Sorry.  I thought I hit reply and didn't.
Yes.   We can change the default and bring up more testing with it  on separately.

dilijev:
I'll merge this soon.

Krovatkin:
@dilijev @curtisman Would you like me to add a new variant(mode) that enables interrupts and can be hooked into CI? 

Optionally, we could also add it as a default variant. 

```
  :: If the user didn't specify explicit variants then do the defaults
  if "%_Variants%"=="" set _Variants=interpreted,dynapogo
```

dilijev:
Those variants are pretty heavyweight (as they run our entire test suite for each variant). That might be the right approach. I think another PR is the best place for that additional work. We can merge this as-is and discuss strategy in a GitHub issue before the next set of changes for this?

LouisLaf:
Rerunning all the tests with that flag is way to big of a hammer.  Adding the flag to a couple unittests would be good, and I have asked @satheeshravi to add the flag to the pool ExprGen uses at random.  That should be enough.

dilijev:
Addresses https://github.com/Microsoft/ChakraCore/issues/2572

PR #2570, opened 2017-02-21
Refactor add_msbuild_path.cmd and locate_msbuild.ps1

dilijev:
Refactor logic of legacy script to use locate_msbuild.ps1

### DIFF
--- a/Build/scripts/add_msbuild_path.cmd
+++ b/Build/scripts/add_msbuild_path.cmd
@@ -5,61 +5,42 @@
 
 :: add_msbuild_path.cmd
 ::
-:: Locate msbuild.exe and add it to the PATH
+:: Locate msbuild.exe and add it to the PATH.
+:: This script wraps the logic in locate_msbuild.ps1.
+::
+:: Note: we use the ugly pattern below because it seems to be the best (or only) way
+:: to take the output of a command and assign it to an environment variable.
+::     for /f "delims=" %%a in ('command') do @set VARIABLE=%%a
 
+@echo off
 set USE_MSBUILD_12=%1
 
 if "%USE_MSBUILD_12%" == "True" (
-    echo Skipping Dev14 and trying Dev12...
-    goto :LABEL_USE_MSBUILD_12
+    echo Trying Dev12 directly...
+    for /f "delims=" %%a in ('powershell -Command ". %~dp0\locate_msbuild.ps1; Split-Path -Parent (Locate-MSBuild-Version -version 12.0)"') do @set MSBUILD_PATH=%%a
+) else (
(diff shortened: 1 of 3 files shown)


Build/scripts/add_msbuild_path.cmd
```diff
@@ -5,61 +5,42 @@
...
-if not exist %MSBUILD_PATH%\msbuild.exe (
-    set MSBUILD_PATH="%ProgramFiles(x86)%\msbuild\%MSBUILD_VERSION%\Bin\amd64"
-)
-
-if exist %MSBUILD_PATH%\msbuild.exe (
+    echo Trying Dev12 directly...
+    for /f "delims=" %%a in ('powershell -Command ". %~dp0\locate_msbuild.ps1; Split-Path -Parent (Locate-MSBuild-Version -version 12.0)"') do @set MSBUILD_PATH=%%a
+) else (
+    for /f "delims=" %%a in ('powershell -Command "[bool](Get-Command msbuild.exe -ErrorAction SilentlyContinue)"') do @set MSBUILD_FOUND=%%a
+    if "%MSBUILD_FOUND%" == "True" (
```
Cellule: This won't work, you need to add `setlocal EnableDelayedExpansion` at the top and `endlocal` in cleanup
then you have to change to `if "!MSBUILD_FOUND!" == "True" (`

dilijev: Really? This seemed to be working for me.

Cellule: The variable `%MSBUILD_FOUND%` is evaluated upon entering the else block, therefore the modification in the for-loop comes too late.

dilijev: Okay I'll confirm and then make the change.

dilijev:
@Cellule I added delayed expansion. Since we need to export a new PATH from this script, and `setlocal` (even with the param `EnableDelayedExpansion`) creates a scope for the script, I needed to move the `setlocal ... endlocal` as close to the usage as possible. Because of the exact scope where I did this I was worried it would not work at all, but it does fix the issue (if msbuild.exe is already on the path, the script reports that it could not locate msbuild.exe).

On the other hand, I'm now wondering whether it would be better to just add a function to the powershell script to compute and output the new PATH (whether modified or not) and assign that to path. That way we can keep all the logic in powershell and only wrap the logic for the purposes of exporting the updated path. That was essentially the goal of the refactoring I did here in the first place.

Cellule:
One thing we could do differently in the script is, in the dev12 check put a goto like `goto:checkMSBuild` then we don't need the else and the we shouldn't need the delayed expansion.

tcare:
Let's just merge this for now :)

dilijev:
Merged #2842 which included #2835. I'll rebase and update this PR to just be the refactoring in `add_msbuild_path.cmd`

dilijev:
Will replace with a new PR when ready.

PR #2571, opened 2017-02-21
Fix failure to initialize stack nested func parent pointer. 

pleath:
When a fully compiled function is nested inside a redeferred function, we can quit byte code generation early for the nested function. But don't do so until its stack nested function parent pointer has been initialized (if necessary).


pleath:
@curtisman This is the stack nested func parent init issue that was mystifying us.

PR #2580, opened 2017-02-22
Fix parsing a lambda consisting of a single statement wrapped in parens

boingoing:
If we parse a lambda which only has a single statement which is wrapped
in parens but not curly braces, we calculate the extent of that lambda
parse node incorrectly.

var a = () => (123)

In above, the parse node for the lambda has it's limit character set to
the limit character of the return node - which is the numeric constant
123. The return node throws away the parens and, thus, the character
extents for the lambda as a whole are incorrectly pointing to the last
character of 123.

This is not a regression in behavior. In existing code, if you try to
toString a lambda like this it will get cutoff:

print('"' + a + '"'); // "() => (123"

However, after improving the utf8 parsing such that we correctly
calculate the extents of the bytes we need to parse, toString of the
lambda will hit an assert due to the mismatch in expected length and
actual characters parsed.

Fix here is to keep track of the last right paren we parsed when we're
parsing the lambda body expression and use that instead of the character
extent of the return parse node.

### DIFF
--- a/lib/Parser/Parse.h
+++ b/lib/Parser/Parse.h
@@ -824,7 +824,7 @@ private:
         _Inout_opt_ IdentToken* pToken = NULL,
         bool fUnaryOrParen = false,
         _Inout_opt_ bool* pfLikelyPattern = nullptr,
-        _Out_opt_ charcount_t *plastRParen = nullptr);
+        _Inout_opt_ charcount_t *plastRParen = nullptr);
 
     template<bool buildAST> ParseNodePtr ParseTerm(
         BOOL fAllowCall = TRUE,
@@ -836,7 +836,7 @@ private:
         _Out_opt_ BOOL* pfCanAssign = nullptr,
         _Inout_opt_ BOOL* pfLikelyPattern = nullptr,
         _Out_opt_ bool* pfIsDotOrIndex = nullptr,
-        _Out_opt_ charcount_t *plastRParen = nullptr);
+        _Inout_opt_ charcount_t *plastRParen = nullptr);
 
     template<bool buildAST> ParseNodePtr ParsePostfixOperators(
         ParseNodePtr pnode,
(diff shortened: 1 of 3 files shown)


boingoing:
@pleath @akroshg 

Please take a look if you get a minute. Pretty small fix.

lib/Parser/Parse.h
```diff
@@ -195,6 +195,7 @@ class Parser
     bool                m_hasParallelJob;
     bool                m_doingFastScan;
     int                 m_nextBlockId;
+    charcount_t         m_lastRParen;
```
pleath: Rather than making this a member of the parser instance, can it be local to ParseExpr and passed by reference to ParseTerm?

test/es6/lambda1.js
```diff
@@ -454,6 +454,23 @@ var tests = [
             assert.throws(function () { eval('var a = {}; a.x \n => d;'); }, SyntaxError, "Verify that badly formed arrow functions return correct error even if a newline is before the => token", "Syntax error");
             assert.throws(function () { eval('var a = {}; a\n.x => d;'); }, SyntaxError, "Verify that badly formed arrow functions return correct error even if a newline is before the => token", "Syntax error");
         }
+    },
+    {
+        name: "Lambda consisting of single expression wrapped in parens should have correct source string",
```
pleath: Can you add nested lambda cases?

boingoing: Yes, good call.

lib/Parser/Parse.h
boingoing: Needs to flow out of ParseExpr but can be passed as optional to both. I didn't do this initially because I thought it might be hard to pass the optional param down all paths but actually seems to be pretty simple.

akroshg:
looks good to me.

Issue #2541, opened 2017-02-16
[JSRT] Promise API enhancements


jaredwy:
I apologies if this is the wrong place to ask this question.  I couldn't find a more appropriate place.

I am looking to create a function from C++ that will return a promise to the callee in JS. Then be able to resolve/reject it when a async task has been completed natively. Something similar to `v8::Promise::Resolver`

I can't see anything in the docs, nor the source. Is there a way to do such a think in ChakraCore at the moment?

liminzhu:
I don't believe we have it in JSRT now. These's an active discussion about this on [Gitter](https://gitter.im/Microsoft/ChakraCore?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge). 

@galvesribeiro @ReubenBond @boingoing @dilijev maybe we should continue in this issue?

jaredwy:
Great thanks. I can ask my follow up question in gitter.

galvesribeiro:
For the record, that is the API we are reffering to https://v8docs.nodesource.com/node-7.4/d3/d8f/classv8_1_1_promise.html

liminzhu:
Let me post a short summary for the conversation with Orleans folks (@galvesribeiro @ReubenBond) and @jaredwy on Gitter from [here](https://gitter.im/Microsoft/ChakraCore?at=589e70b5567b0fc8138d1ba0) to [here](https://gitter.im/Microsoft/ChakraCore?at=58a63573de50490822da2d55) and let's discuss further matters here to preserve context. And folks, feel free to suggest edits if I miss anything. 

There're general two asks: 
1. Ability create a promise from the native side. Something to the effect of `JsCreatePromise(JsValueRef resolve, JsValueRef resolve, JsValueRef *promise)` which mirrors `new Promise(resolveFunc, rejectFunc)` in JS. This is already do-able in JSRT by running a small script which returns a Promise, but the code gets really messy if handlers are native function. For what it's worth, V8 has similar [APIs](https://v8docs.nodesource.com/node-7.4/d3/d8f/classv8_1_1_promise.html). 
2. Similar to 1), ability to resolve/reject from native.  
3. Task/Promise interop between C# and JavaScript. Essentially, support for this pattern (copying [stuff](https://gitter.im/Microsoft/ChakraCore?at=58a3b1f6567b0fc8138d8222) from @ReubenBond), 

_We provide some asynchronous methods to JS code_
```
public interface IDatabase
{
[...]

PR #2581, opened 2017-02-23
Added JsCreatePromise to JsRT

kfarnung:


liminzhu:
+@curtisman , this is the context https://github.com/Microsoft/ChakraCore/issues/2541

lib/Jsrt/ChakraCommon.h
```diff
@@ -2378,6 +2378,30 @@ typedef unsigned short uint16_t;
...
 
+    /// <summary>
+    ///     Creates a new JavaScript Promise object.
+    /// </summary>
+    /// <remarks>
+    ///     <para>
+    ///     Requires an active script context.
+    ///     </para>
+    /// </remarks>
+    /// <param name="promise">The new Promise object.</param>
```
boingoing: Just a nitpick, this param tag is all one line while the others are split up.

lib/Jsrt/Jsrt.cpp
```diff
@@ -2866,6 +2867,32 @@ CHAKRA_API JsSetPromiseContinuationCallback(_In_ JsPromiseContinuationCallback p
...
+    Js::JavascriptPromise *jsPromise = scriptContext->GetLibrary()->CreatePromise();
+    Js::JavascriptPromise::InitializePromise(jsPromise, &jsResolve, &jsReject, scriptContext);
+
+    *promise = (JsValueRef)jsPromise;
+    *resolve = (JsValueRef)jsResolve;
+    *reject = (JsValueRef)jsReject;
+
+    return JsNoError;
+  });
+}
```
dilijev: nit: indentation level in the rest of this file is 4 spaces.

kfarnung: Good catch, VS doesn't seem to auto-detect indentation.

lib/Jsrt/Jsrt.cpp
```diff
@@ -2866,6 +2867,32 @@ CHAKRA_API JsSetPromiseContinuationCallback(_In_ JsPromiseContinuationCallback p
     /*allowInObjectBeforeCollectCallback*/true);
 }
 
+CHAKRA_API JsCreatePromise(_Out_ JsValueRef *promise, _Out_ JsValueRef *resolve, _Out_ JsValueRef *reject)
+{
+  return ContextAPIWrapper<true>([&](Js::ScriptContext *scriptContext, TTDRecorder& _actionEntryPopper) -> JsErrorCode {
```
boingoing: Nitpick, indentation here.

lib/Jsrt/ChakraCommon.h
dilijev: We're inconsistent about this elsewhere in this file as well. I wonder if it has to do with line length and VS doing something automatically with formatting?

liminzhu:
I'm a little confused about what `resolveFunction` and `rejectFunction` output are and how I would use them.

Let's say JsCreatePromise is equivalent to `new Promise((res, rej) => {...})`, would this make more sense? 
```
JsCreatePromise(_In_ JsValueRef executor, _Out_ JsValueRef *promise)
``` 
`executor` could be a native function created by `JsCreateFunction`.

boingoing:
@liminzhu 

> I'm a little confused about what resolveFunction and rejectFunction output are and how I would use them.

These are essentially the same functions passed to the executor. Another method to fetch them might be like this:

```JavaScript
var resolve, reject;
function executor(res, rej) { resolve = res; reject = rej; };
var promise = new Promise(executor);
```
Goal would be to allow our native code to resolve the promise whenever it wants to. One use might be an API which reads a file from the network and returns a promise resolved with the contents of that file. Native code can construct the promise using JsCreatePromise and hold-on to resolveFunction for that promise. Assume the native-side performs some processing to fetch the result data and when that is done it can call the resolveFunction with that result data which will handle resolving the promise in the engine. I guess the key thing here is that the native code can resolve the promise in an async way without having to create an executor function which would manually suck out the resolve/reject functions.

liminzhu:
Makes much sense now. Thanks @boingoing !

lib/Jsrt/Jsrt.cpp
```diff
@@ -4508,4 +4509,30 @@ CHAKRA_API JsRunSerialized(
...
+CHAKRA_API JsCreatePromise(_Out_ JsValueRef *promise, _Out_ JsValueRef *resolve, _Out_ JsValueRef *reject)
+{
+    return ContextAPIWrapper<true>([&](Js::ScriptContext *scriptContext, TTDRecorder& _actionEntryPopper) -> JsErrorCode {
+        PERFORM_JSRT_TTD_RECORD_ACTION_NOT_IMPLEMENTED(scriptContext);
+
+        PARAM_NOT_NULL(promise);
+        PARAM_NOT_NULL(resolve);
+        PARAM_NOT_NULL(reject);
+
+        *promise = nullptr;
```
obastemur: is this needed? (same applies to two below)

galvesribeiro: It is not mandatory to have either resolve() or reject() methods FYI.

galvesribeiro:
LGTM. I just would like to mention that it is not required to have both reject() and response() methods. And if possible, it would be good to accept multiple methods accepted because people can catch multiple types of errors on diff methods. For example:

`mypromise.then(...).then(...).catch(...).catch(...)` <- that is totally viable.

kfarnung:
@galvesribeiro I think you might have it backwards, the `resolve` and `reject` parameters are out parameters that will contain pointers to the functions that the native code will call when it has either resolved or rejected the promise.  These are equivalent to the parameters that come into the `executor` method when creating a promise in JavaScript.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise

For continuations you would just use the JsRT functions to get the property (`then`, `catch`, etc.) and then call it normally with `JsCallFunction`.

In C#:
```
JavaScriptPropertyId thenId = JavaScriptPropertyId.FromString("then");
JavaScriptValue thenFunction = promise.GetProperty(thenId);

JavaScriptValue resolveFunc = JavaScriptValue.CreateFunction(ResolveCallback);
JavaScriptValue rejectFunc = JavaScriptValue.CreateFunction(RejectCallback);
JavaScriptValue thenPromise = thenFunction.CallFunction(promise, resolveFunc, rejectFunc);
```

Since it's just a JavaScript function, you can choose to leave out the second parameter when calling it.

lib/Jsrt/Jsrt.cpp
kfarnung: @obastemur I was just following the pattern of other methods.  The idea was to ensure that regardless of how the flow changed we would always have set some value.  I realize that in this flow we will always set a value, so I'm fine removing it if others agree.

galvesribeiro:
Oh! Right, I understood it completely wrong, sorry. You are right! Good work!

lib/Jsrt/Jsrt.cpp
dilijev: I don't see a reason to remove this. In this case the compiler will be able to figure out that this operation is not needed -- if the flow ever changes, we are protected.

lib/Jsrt/Jsrt.cpp
```diff
@@ -4523,4 +4524,30 @@ CHAKRA_API JsRunSerialized(
         sourceContext, // use the same user provided sourceContext as scriptLoadSourceContext
         buffer, bufferVal, sourceContext, url, false, result);
 }
+
+CHAKRA_API JsCreatePromise(_Out_ JsValueRef *promise, _Out_ JsValueRef *resolve, _Out_ JsValueRef *reject)
+{
+    return ContextAPIWrapper<true>([&](Js::ScriptContext *scriptContext, TTDRecorder& _actionEntryPopper) -> JsErrorCode {
+        PERFORM_JSRT_TTD_RECORD_ACTION_NOT_IMPLEMENTED(scriptContext);
```
jianchun: Do we need to implement this for TTD? @mrkmarron

kfarnung:
Closing this one and retargeting to release/2.0 in #2594.

lib/Jsrt/Jsrt.cpp
kfarnung: I spoke with @mrkmarron and this was his recommendation.  If the user is running TTD and runs into this we'll fail fast and deal with it at that time.

PR #2582, opened 2017-02-23
Fix incorrect deletion of unreachable InlineeEnd

ThomsonTan:
When there is `InlineeEnd` IR instruction in successor block following the block with `BailOnNoProfile`, the succesor block will be considered as dead if there is no other incoming block (or other incoming blocks are also dead). GlobOpt removes such dead block in the begining of `GlobOpt::OptBlock`  so no fall through instructions after `BailOnNoProfile`. But LinearScan only generates `InlineeFrameRecord` when handling `InlineeEnd`, and previous deletion of `InlineeEnd` could lead to incorrect generation of `InlineeFrameRecord`.

### DIFF
--- a/lib/Backend/FlowGraph.cpp
+++ b/lib/Backend/FlowGraph.cpp
@@ -2637,6 +2637,54 @@ FlowGraph::RemoveInstr(IR::Instr *instr, GlobOpt * globOpt)
     }
 }
 
+// We cannot just remove Js::OpCode::InlineeEnd from cold path, need to make a copy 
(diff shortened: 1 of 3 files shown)


ThomsonTan:

ThomsonTan:
Ping:)

rajatd:
Add a test case. Otherwise LGTM

ThomsonTan:
@dotnet-bot test OSX please

ThomsonTan:
@rajatd, thanks for review. Seems some simple scripts snippet from the repro website cannot repro.

PR #2583, opened 2017-02-23
build-bug: fix C linkage

### LINKED ISSUE #2574: Duplicate symbols from constants when linking
This is fairly easy to demonstrate. I believe that declaring these three symbols as constants in ChakraCommon.h is a mistake as you can then never include ChakraCommon.h in multiple compilation units. Then again I don't have the greatest familiarity with the minutiae of C standards.

a.c
```c
#include <ChakraCore.h>

void do_hello_world(void);

int
main(int argc, char* argv[])
{
    do_hello_world();
}
```

b.c
```c
#include <stdio.h>
#include <ChakraCore.h>

void
do_hello_world(void)
{
    printf("Hello, World!\n");
}
```

Link failure
```bash
gcc -o bug a.c b.c
duplicate symbol _JS_SOURCE_CONTEXT_NONE in:
    /var/folders/4p/51v2mtgj0rzg_66d5vr3mygh0000gn/T/a-a26b4b.o
    /var/folders/4p/51v2mtgj0rzg_66d5vr3mygh0000gn/T/b-8eb111.o
duplicate symbol _JS_INVALID_RUNTIME_HANDLE in:
    /var/folders/4p/51v2mtgj0rzg_66d5vr3mygh0000gn/T/a-a26b4b.o
    /var/folders/4p/51v2mtgj0rzg_66d5vr3mygh0000gn/T/b-8eb111.o
duplicate symbol _JS_INVALID_REFERENCE in:
    /var/folders/4p/51v2mtgj0rzg_66d5vr3mygh0000gn/T/a-a26b4b.o
    /var/folders/4p/51v2mtgj0rzg_66d5vr3mygh0000gn/T/b-8eb111.o
ld: 3 duplicate symbols for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

obastemur:
Global variable values are set on common header file breaks the link process. This happens only when two or more C source files includes ChakraCore.h

Also updated one of the native test cases to test this particular scenario and prevent future mistakes.

Fixes #2574

### DIFF
--- a/test/native-tests/test-char/Platform.js
+++ b/test/native-tests/test-char/Platform.js
@@ -24,7 +24,7 @@ LDIR=$(LIBRARY_PATH)/libChakraCoreStatic.a \n\
 \n\
 ifeq (darwin, ${PLATFORM})\n\
 \tICU4C_LIBRARY_PATH ?= /usr/local/opt/icu4c\n\
-\tCFLAGS=-lstdc++ -std=c++11 -I$(IDIR)\n\
+\tCFLAGS=-lstdc++ -I$(IDIR)\n\
 \tFORCE_STARTS=-Wl,-force_load,\n\
 \tFORCE_ENDS=\n\
 \tLIBS=-framework CoreFoundation -framework Security -lm -ldl -Wno-c++11-compat-deprecated-writable-strings \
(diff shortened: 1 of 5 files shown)


test/native-tests/test-char/Platform.js
```diff
@@ -33,15 +33,15 @@ ifeq (darwin, ${PLATFORM})\n\
     $(ICU4C_LIBRARY_PATH)/lib/libicuuc.a \
     $(ICU4C_LIBRARY_PATH)/lib/libicui18n.a\n\
 else\n\
-\tCFLAGS=-lstdc++ -std=c++0x -I$(IDIR)\n\
+\tCFLAGS=-lstdc++ -I$(IDIR)\n\
```
jianchun: Does removing "-std=c++11" and "-std=c++0x" reduce coverage?

obastemur: This is just another one of the tests. We have other tests those we compile with `c++` flags. Since we test `C` files for this particular test, no `c++` flag is allowed / needed.

obastemur:
@jianchun @kfarnung Thanks for the review

PR #2588, opened 2017-02-23
No need to mark use of scope object if there are no bailouts in the function

rajatd:
Marking a use of scope object sym for stack args optimization was creating bytecodeUpwardExposedUsed bit vector for the current block without checking if the function had a bailout. This led to an assert that expects no byteCodeUpwardExposedUsed bv to be allocated in the absence of any bailout in the function.

Fix is to add the mentioned check.


satheeshravi:
LGTM

satheeshravi:
:shipit:

PR #2589, opened 2017-02-23
Removes the function body from the source info in case ByteBlock allocation hits OOM.

atulkatti:


atulkatti:
@Microsoft/chakra-developers 
This has already been reviewed in codeflow.

lib/Runtime/Base/FunctionBody.cpp
```diff
@@ -6257,6 +6257,13 @@ namespace Js
         this->SetConstTable(nullptr);
         this->byteCodeBlock = nullptr;
 
+        // Also, remove the function body from the source info to prevent any further processing 
+        // of the function such as attempts to set breakpoints.
+        if (GetIsFuncRegistered())
```
akroshg: do we retry again somewhere when BytecodeGen failed in the first time?

atulkatti: I don't think we do. In this particular case because it's OOM we will throw and come out of the EmitProgram call.

PR #2591, opened 2017-02-23
make it still working with disabling concurrent GC

leirocks:


leirocks:
@dotnet-bot  test this please

dilijev:
Should we add a CI definition to test with concurrent GC disabled?

leirocks:
I don't think that is still needed. This is just make sure it's still easy to be disabled, and in the future if someone wants to disable concurrent GC to do some testing, it will be less work to do.

dilijev:
This replaces https://github.com/Microsoft/ChakraCore/pull/2454 right? If so, can you close that PR?

PR #2592, opened 2017-02-24
Fix leak in JSRT

leirocks:
The leak is because of a circular reference and more than one node on the circle are pinned.

1. Pinned object fakeGlobalFuncForUndefer (as well as profileInfoList in test/debug build) reference to javascriptLibrary -- directly or indirectly, these are rely on ScriptContext::Close() to unpin.
2. javascriptLibrary has a reference to JsrtContext
3. JsrtContext is pinned while setting to current thread, and unpinned when getting out of current thread
4. if user code didn't explicited pin JsrtContext (in following POC), at this stage it should be disposed in next GC, and hence call ScriptContext::Close()
5. the disposal in # 4 didn't because JsrtContext is reachable through fakeGlobalFuncForUndefer->javascriptLibrary->JsrtContext(# 2), so the whole graph is leaked
6. when there's external call to JsDisposeRuntime, it will directly dispose JsrtContext, and then ScriptContext::Close, unpin fakeGlobalFuncForUndefer then everything is collectable

the POC:
```c++
    JsRuntimeHandle runtime;
    unsigned currentSourceContext = 0;
    JsCreateRuntime(JsRuntimeAttributeNone, nullptr, &runtime);
    auto runJob = [&](wstring script, int i)
    {
        {
            JsValueRef result;
            JsContextRef context;
            JsCreateContext(runtime, &context);
            JsSetCurrentContext(context);
            JsRunScript(script.c_str(), currentSourceContext++, L"", &result);
            JsSetCurrentContext(JS_INVALID_REFERENCE);
            context = nullptr;
            result = nullptr;
        }

        if (i % 5 == 0) {
            JsCollectGarbage(runtime); // JsrtContext in above scope should be collectible at this point,
                                       // but the Finalize/Dispose of JsrtContext didn't happen
        }
    };

    for (int i = 0; i < 100; i++)
    {
        runJob(L"(()=>{return \'Hello world!\';})()", i);
    }

    printf("JsDisposeRuntime\n");
    JsDisposeRuntime(runtime); // all JsrtContext will be collected at this point
    printf("After JsDisposeRuntime\n");
```

The fix is, do not pin fakeGlobalFuncForUndefer and profileInfoList. However, there are a lot of code(mostly debugger related code) rely on the leak to do the cleanup. Most of the work is to make sure the cleanup working correctly (without either UAF or leak).

### DIFF
--- a/lib/Runtime/Base/ThreadContext.cpp
+++ b/lib/Runtime/Base/ThreadContext.cpp
@@ -591,7 +591,8 @@ ThreadContext::~ThreadContext()
 void
 ThreadContext::SetJSRTRuntime(void* runtime)
 {
-    Assert(jsrtRuntime == nullptr); jsrtRuntime = runtime;
+    Assert(jsrtRuntime == nullptr);
+    jsrtRuntime = runtime;
 #ifdef ENABLE_BASIC_TELEMETRY
     Telemetry::EnsureInitializeForJSRT();
 #endif
@@ -3030,7 +3031,10 @@ ThreadContext::ClearScriptContextCaches()
 {
     for (Js::ScriptContext *scriptContext = scriptContextList; scriptContext != nullptr; scriptContext = scriptContext->next)
     {
-        scriptContext->ClearScriptContextCaches();
+        if (!scriptContext->IsClosed())
+        {
+            scriptContext->ClearScriptContextCaches();
+        }
     }
 }
 
--- a/lib/Runtime/Base/ThreadContext.cpp
+++ b/lib/Runtime/Base/ThreadContext.cpp
@@ -3031,10 +3031,7 @@ ThreadContext::ClearScriptContextCaches()
 {
     for (Js::ScriptContext *scriptContext = scriptContextList; scriptContext != nullptr; scriptContext = scriptContext->next)
     {
-        if (!scriptContext->IsClosed())
-        {
-            scriptContext->ClearScriptContextCaches();
-        }
+        scriptContext->ClearScriptContextCaches();
     }
 }
 
--- a/lib/Runtime/Library/JavascriptLibrary.cpp
+++ b/lib/Runtime/Library/JavascriptLibrary.cpp
@@ -1054,6 +1054,7 @@ namespace Js
             scriptContext->ResetWeakReferenceDictionaryList();
             scriptContext->SetIsFinalized();
             scriptContext->GetThreadContext()->UnregisterScriptContext(scriptContext);
+            scriptContext->MarkForClose();
         }
     }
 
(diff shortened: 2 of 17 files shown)


lib/Runtime/Base/ThreadContext.cpp
```diff
@@ -3030,7 +3031,10 @@ ThreadContext::ClearScriptContextCaches()
 {
     for (Js::ScriptContext *scriptContext = scriptContextList; scriptContext != nullptr; scriptContext = scriptContext->next)
     {
-        scriptContext->ClearScriptContextCaches();
+        if (!scriptContext->IsClosed())
```
curtisman: Why do we have to check now?

lib/Runtime/Library/JavascriptLibrary.cpp
```diff
@@ -1054,29 +1054,29 @@ namespace Js
             scriptContext->ResetWeakReferenceDictionaryList();
             scriptContext->SetIsFinalized();
             scriptContext->GetThreadContext()->UnregisterScriptContext(scriptContext);
+            scriptContext->MarkForClose();
```
curtisman: Is this here just for ScriptContext::SaveStartupProfileAndRelease?  Can you just move that function to JAvascriptLibrary and call it directly?
Then you don't need to add the ISClose check in MarkForClose too, right?

lib/Runtime/Library/JavascriptLibrary.cpp
```diff
@@ -1054,29 +1054,29 @@ namespace Js
...
         }
     }
 
     void JavascriptLibraryBase::Dispose(bool isShutdown)
     {
         if (scriptContext)
         {
-            if (isShutdown)
-            {
-                // during shut down the global object might not be closed yet.
```
curtisman: Why is this not needed anymore?

lib/Runtime/Library/JavascriptLibrary.h
```diff
@@ -28,6 +28,61 @@ namespace Projection
...
+        static const int kMaxCachedPropStrings = 16;
+        Field(DynamicObject*) object;
+        Field(DynamicType*) type;
+        Field(PropertyString*) propertyStrings[kMaxCachedPropStrings];
+        Field(int) validPropStrings;
+    };
+
+    // this is allocated in GC directly to avoid force pinning the object, it is linked from JavascriptLibrary such that it has
+    // the same lifetime as JavascriptLibrary, and it can be collected without ScriptContext Close.
+    // Allocate it as Finalizable such that it will be still available during JavascriptLibrary Dispose time
```
curtisman: This comment is no longer true.

lib/Runtime/Library/JavascriptLibrary.cpp
leirocks: these work was duplicate in ScriptContext::Close(true) which called from ~ScriptContext(), which will be triggered by the HeapDelete right after here

lib/Runtime/Library/JavascriptLibrary.cpp
leirocks: not only for that, after this point, the recycler pointers hold by or can-be-reached-through scriptContext may be collected/invalid. most of those pointers was in Cache struct (which I moved to directly on JavascriptLibrary to simplify the cleanup path). I am using isClosed on scriptContext as barrier, and do those cleanup. if we do this at JavascriptLibrary::Dispose() time, it's too late because those recycler pointer might be collected already

lib/Runtime/Base/ThreadContext.cpp
leirocks: looks this is not necessary, I added it while tweaking the change. let me remove it

lib/Runtime/Library/JavascriptLibrary.h
leirocks: will remove it

curtisman:
:shipit:

PR #2596, opened 2017-02-24
ResetObject is setting undefined for internal properties

akroshg:
Once the GetInternalProperty fails the stackTraceValue (and others) are set to GetMissingPropertyResult. Later below in that function we  set the undefined object for such property which actually leads to crashes in the Weakmap functionality as we never expect it to be undefined.
Fixed this and put an assert just incase in the Weakmap we get undefined again.


akroshg:
This is reviewed and signed off by @kunalspathak

PR #1569, opened 2016-09-12
int32x4 const folding for asm.js/wasm? + tests

Krovatkin:
@LouisLaf @dilijev @MikeHolman 

Could you please take a look?

This is an initial version of const folding for int32x4 in asm.js (possibly WASM). We could add support for more simd types (int8x16, int16x8, uint32x4, uint8x16, uint16x8, etc) in future installments.

FYI:
@commonlisp @arunetm

PR #2597, opened 2017-02-24
Trivial Int32/Int64 constant folding for WASM/asm.js

Krovatkin:
This PR implements simple constant folding for int32/int64 arithmetic (e.g. `Add_I4`, `Sub_I4`, etc) and conditional (e.g. `BrGe_I4`, `BrEq_I4`, etc) expressions for WASM and asm.js.

* Louis suggested that we introduced a new class for representing int64 constants rather than re-use the same class for both, int32 and int64 constants. Two classes seem to make logic a little more verbose. Even more so since we couldn't use inheriting constructors due to an older compiler on win7.  We also have another version using one class just in case.
* We didn't implement caching of int64 constants as it involves tracking liveness information and we weren't quite sure how it would fit in the existing infrastructure.
* We didn't extend `IntBounds` and `IntRanges` to support int64 which could enable more opts.
* All Int64 constants are un-taggable, since the process of introducing new types to represent int64 in JavaScript is still ongoing. 
* We could extend this to support more operators (e.g. unary ops) if there's interest.
* We didn't explore the idea of smaller int64 constants "standing in" for int32 since WASM is strictly typed and asm.js doesn't have int64, so we weren't even sure if there was any value in looking into that idea.
* We ran smaller tests (e.g. fasta, skinning, etc) included in emscripten benchmark to make we aren't degrading performance. There seems to be no obvious performance degradation and no apparent perf improvement since these tests seem to expose very few opportunities for constant folding. 

Below are a few basic scenarios we tested this PR on. 

Wasm tests
```lisp
(module
  (func (export "test") (result i32) (local i64 i64 i64)
    (set_local 0 (i64.const 3))
    (set_local 1 (i64.const 2))
    (set_local 2 (i64.mul (get_local 0) (get_local 1)))
    (set_local 2 (i64.mul (get_local 0) (i64.const 10)))
    (i32.wrap/i64 (get_local 2))
   )

  (func (export "while") (result i32) (local i64 i64 i64 i64 i64)
    (set_local 0 (i64.const 5))
    (set_local 1 (i64.const 1))
    (set_local 2 (i64.const 3))
    (set_local 3 (i64.const 2))
    (set_local 4 (i64.const 7))
    (block
      (loop
        (br_if 1 (i64.eqz (get_local 0)))
        (set_local 3 (i64.mul (get_local 2) (get_local 4)))
        (set_local 1 (i64.mul (get_local 0) (get_local 1)))
        (set_local 1 (i64.mul (get_local 0) (get_local 3)))
        (set_local 0 (i64.sub (get_local 0) (i64.const 1)))
        (br 0)
      )
    )
    (i32.wrap/i64 (get_local 1))
  )
)
```

Asm.js tests
```javascript
    function f3() {
        var x=0;
        var e = 0;
        var f = 2;
        var g = 0;
        e = 2|0;
        f = 3|0;

        while((x|0) < (3|0)) {
            g = (e * f) | 0;
            x=x+g|0;
        }

        return x|0;
    }

    function f1(){
        var e = 0;
        var f = 2;
        var g = 0;
        e = 2|0;
        f = 3|0;
        g = (e * f) | 0;
        g = (g + e) | 0;
        return g|0;
```


dilijev:
Hi @Krovatkin - Is this related to https://github.com/Microsoft/ChakraCore/pull/1569?

Krovatkin:
@dilijev #1569 is not related. #1569 was a prototype targeting one particular SIMD type, int32x4. We'll probably revamp it in a little bit when start working on WASM.SIMD. 

This one, #2597 , is on Mike's roadmap to improve WASM perf, **Int64 const folding**.

Krovatkin:
@LouisLaf @dilijev @MikeHolman could you guys please take a look when you have a moment?

dilijev:
> We also have another version using one class just in case.

Is that change dependent on this one being merged or is it an alternative to this PR?

Krovatkin:
@dilijev 

> Is that change dependent on this one being merged or is it an alternative to this PR?

It's an alternative in case you guys would be interested to see how one class implementation looks like.

dilijev:
Before merging, maybe we could take a look at that, too? I'm not an expert in the area, but I'm curious enough. Could you push your alternate branch and provide a link to the diff against master, or make an alternate PR for that?

Krovatkin:
@dilijev 

I'll open another PR shortly w/ one-class implementation shortly. One-class implementation looks a bit cleaner, IMO. However, Louis' main concern was an increase in memory footprint i.e. extra 4 bytes per each non int64 constant

dilijev:
> Below are a few basic scenarios we tested this PR on.

Could you convert those into unit tests?

dilijev:
> However, Louis' main concern was an increase in memory footprint i.e. extra 4 bytes per each non int64 constant

Maybe there's a compromise implementation that can simplify the coding while still saving the space? I'm imagining maybe something with macros but I wouldn't necessarily encourage that.

Krovatkin:
@dilijev 

> Could you convert those into unit tests?

I could definitely do that. I am wondering if we should add **only** non-loopy ones, though, since nearly every 3rd/4th test in WasmSpec triggers constant folding ? 🔨

dilijev:
@Krovatkin - sure, whichever you think is best. If you have a test that covers a scenario not already tested, then please add it, otherwise the existing tests are fine.

Krovatkin:
@dotnet-bot
test OSX osx_osx_test_static please

Krovatkin:
@dotnet-bot
test Windows x86_debug please

Krovatkin:
@MikeHolman removed unnecessary tests and rebased against the latest `master`. Please let me know I missed stn!

Krovatkin:
@dotnet-bot
test Windows x86_debug please

PR #2599, opened 2017-02-25
Fix F12 issue with block scope optimization. 

pleath:
We were not doing block scope optimization related to redeferral in F12 debug mode, but this causes inconsistency when a function object gets byte code accessing an enclosing block scope, and the function was instantiated in non-debug mode and thus has a closure environment not containing the block scope.


rajatd:
LGTM

PR #2600, opened 2017-02-25
clear scopeInfo->scope to avoid GC false positive

jianchun:
The given field holds a pointer to arena memory temporarily.
Clear it when done to avoid GC false positive (memory released
and reused by GC later).

This issue blocks -VerifyBarrierBit (without -KeepRecyclerTrackData
thus the filter doesn't work) test on Linux.


pleath:
LGTM. Thanks.

jianchun:
Thanks @leirocks @digitalinfinity @pleath

PR #2601, opened 2017-02-25
Add checks on SHR's src1 during UnsignedCmpPeep in FGPeeps

meg-gupta:


meg-gupta:

lib/Backend/FlowGraph.cpp
```diff
@@ -3014,7 +3014,15 @@ bool FlowGraph::IsUnsignedOpnd(IR::Opnd *src, IR::Opnd **pShrSrc1)
         return false;
     }
 
-    *pShrSrc1 = shrUInstr->GetSrc1();
+    IR::Opnd *shrSrc = shrUInstr->GetSrc1();
+
+    Assert (shrSrc->IsRegOpnd());
+    if (!shrSrc->AsRegOpnd()->m_sym->m_isIntConst)
```
ThomsonTan: >if (!shrSrc->AsRegOpnd()->m_sym->m_isIntConst) [](start = 4, length = 46)

Could we made a test case for this?

meg-gupta: Added testcase.

MikeHolman:
Why are we doing an |0? Semantics of js are that |0 converts to signed. The point of the  >>> 0 in source is to convert to unsigned.

lib/Backend/FlowGraph.cpp
```diff
@@ -3014,7 +3014,11 @@ bool FlowGraph::IsUnsignedOpnd(IR::Opnd *src, IR::Opnd **pShrSrc1)
         return false;
     }
 
-    *pShrSrc1 = shrUInstr->GetSrc1();
+    IR::Opnd *shrSrc = shrUInstr->GetSrc1();
+
+    Assert (shrSrc->IsRegOpnd());
```
ThomsonTan: >Assert (shrSrc->IsRegOpnd()); [](start = 4, length = 29)

Why assert here, got some violation?

lib/Backend/FlowGraph.cpp
```diff
@@ -3034,6 +3038,7 @@ bool FlowGraph::UnsignedCmpPeep(IR::Instr *cmpInstr)
     //
     //  t1 = ShrU_A x, 0
     //  t2 = 10;
+    //  t3 = Or_A x, 0
     //  BrUnGt x, t2, L
```
ThomsonTan: >x [](start = 15, length = 1)

Sounds like this should be `BrUnGt t3, t2, L`.

ThomsonTan:
I guess `Or_A` is not exactly the same as `|` in JavaScript, and it should be fine to calculate temporary result because its result is not affected by signed/unsigned. Or we probably could use `MOV_TRUNK`?

---
In reply to: [283729990](https://github.com/Microsoft/ChakraCore/pull/2601#issuecomment-283729990) [](ancestors = 283729990)

meg-gupta:
@MikeHolman We do this optimization only when one of the compare branch's sources is a non negative int constant. | 0 was done to convert to 32-bits. Since we are converting signed GT/GE/LT/LE to unsigned compare branches, we produce correct results. For signed EQ/NEQ compare branches, we produce correct results because one of the operands is always non negative int constant.

lib/Backend/FlowGraph.cpp
meg-gupta: Yes, will fix this.

meg-gupta:

MikeHolman:
I looked into this more, and I don't really like this peep... It seems like a hack to tell globopt that `a >>> 0` will make dst into a uint (which we don't support well). But, this fixes the bug, so LGTM.

PR #2603, opened 2017-02-25
TTD -- misc fixes and enhancements.

mrkmarron:
- Add support for additional promise object.
- Disable code cache for TTD.
- Fix handling of names on snap/restore external function.
- Add missing well known functions to core enumeration.
- Revise script context reuse heuristics in re-inflate. Check for problematic objects and enable more re-use.

### DIFF
--- a/lib/Runtime/Debug/TTSnapObjects.cpp
+++ b/lib/Runtime/Debug/TTSnapObjects.cpp
@@ -107,25 +107,78 @@ namespace TTD
         Js::DynamicObject* ReuseObjectCheckAndReset(const SnapObject* snpObject, InflateMap* inflator)
         {
             Js::RecyclableObject* robj = inflator->FindReusableObjectIfExists(snpObject->ObjectPtrId);
-            if(robj == nullptr)
+            if(robj == nullptr || Js::DynamicObject::FromVar(robj)->GetTypeId() != snpObject->SnapType->JsTypeId || Js::DynamicObject::FromVar(robj)->IsCrossSiteObject() != snpObject->IsCrossSite)
             {
                 return nullptr;
             }
             TTDAssert(Js::DynamicType::Is(robj->GetTypeId()), "You should only do this for dynamic objects!!!");
 
             Js::DynamicObject* dynObj = Js::DynamicObject::FromVar(robj);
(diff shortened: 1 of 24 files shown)


lib/Runtime/Debug/TTSnapObjects.cpp
```diff
@@ -153,23 +147,73 @@ namespace TTD
...
+                    {
                         propertyReset.Clear();
-                        return dynObj->GetLibrary()->CreateObject();
+                        return true;
                     }
+                }
+            }
+
+            propertyReset.Clear();
+            ////
```
kfarnung: Unneeded slashes?

lib/Runtime/Debug/TTSnapObjects.cpp
```diff
@@ -153,23 +147,73 @@ namespace TTD
...
+        }
+
+        Js::DynamicObject* ObjectPropertyReset_WellKnown(const SnapObject* snpObject, Js::DynamicObject* dynObj, InflateMap* inflator)
+        {
+            TTDAssert(snpObject->OptWellKnownToken != TTD_INVALID_WELLKNOWN_TOKEN, "Should only call this on well known objects.");
+
+            JsUtil::BaseHashSet<Js::PropertyId, HeapAllocator>& propertyReset = inflator->GetPropertyResetSet();
+            propertyReset.Clear();
+
+            ////
```
kfarnung: nit: Here too

lib/Runtime/Library/GlobalObject.cpp
```diff
@@ -617,7 +617,15 @@ namespace Js
                 Throw::FatalInternalError();
             }
 #endif
+
+#if ENABLE_TTD
+            if(!scriptContext->IsTTDRecordOrReplayModeEnabled())
+            {
+                scriptContext->AddToEvalMap(key, isIndirect, pfuncScript);
```
digitalinfinity: What is the reason for this change?

digitalinfinity:
:shipit:

lib/Runtime/Library/GlobalObject.cpp
mrkmarron: We disabled the cache during record (since trying to replay the behavior was problematic) so no need to reset cache values here.

PR #2610, opened 2017-02-28
Trivial Int32/Int64 constant folding for WASM/asm.js (1-class IntConstantValueInfo version)

Krovatkin:
Doug suggested that I should open another PR w/ one-class implementation of int64 constants. 
This is almost identical to https://github.com/Microsoft/ChakraCore/pull/2597 except for the implementation of int64 constants.

Please disregard this one until it passes all the CI tests


Krovatkin:
@dotnet-bot
pretty-pretty please 😄

lib/Backend/GlobOpt.cpp
```diff
@@ -8958,19 +8979,36 @@ GlobOpt::TypeSpecialization(
...
         if (!this->IsLoopPrePass())
         {
-            // OptConstFoldBinary doesn't do type spec, so only deal with things we are sure are int (IntConstant and IntRange)
-            // and not just likely ints  TypeSpecializeBinary will deal with type specializing them and fold them again
-            IntConstantBounds src1IntConstantBounds, src2IntConstantBounds;
-            if (src1Val && src1Val->GetValueInfo()->TryGetIntConstantBounds(&src1IntConstantBounds))
+            if (GetIsAsmJSFunc())
             {
-                if (src2Val && src2Val->GetValueInfo()->TryGetIntConstantBounds(&src2IntConstantBounds))
+                if (OPT_ON(WasmFold))
```
MikeHolman: can you use CONFIG_FLAG macro? if you are having build issues, you might need to rename DEFAULT_CONFIG_WASM_FOLD to DEFAULT_CONFIG_WasmFold

lib/Common/ConfigFlagsList.h
```diff
@@ -387,6 +387,7 @@ PHASE(All)
 #else
     #define DEFAULT_CONFIG_WasmFastArray    (false)
 #endif
+#define DEFAULT_CONFIG_WASM_FOLD          (false)
```
MikeHolman: any reason not to enable it?

Krovatkin: @MikeHolman 

None, AFAIK. 

* W/ the flag always on, there are quite a few existing unit tests (22+ files) that trigger constant folding which provides me w/ some level of confidence in the functional correctness of this opt.
* Also, I didn't observe any apparent perf regressions on emscripten microbenchmarks I ran. Didn't see any breath-taking improvements either since emscripten benches seem to expose very few opportunities.

I simply added the flag to err on the side of caution.

lib/Backend/GlobOpt.cpp
Krovatkin: @MikeHolman 

If you don't mind I'd rather remove this flag 🛩 I added it proactively in case new opts should be guarded by an "aggressive" flag.

MikeHolman: I think flag is always good to have, e.g. in future it might help with bug investigation. But we should default enable it

Krovatkin: okay! i'll fix this tmr!

lib/Backend/GlobOpt.h
```diff
@@ -318,18 +323,14 @@ template<> ValueNumber JsUtil::ValueToKey<ValueNumber, Value *>::ToKey(Value *co
...
 {
-private:
-    const int32 intValue;
 
-protected:
-    IntConstantValueInfo(const int32 intValue)
-        : ValueInfo(GetInt(IsTaggable(intValue)), ValueStructureKind::IntConstant),
-        intValue(intValue)
+public:
+    static IntConstantValueInfo *New(JitArenaAllocator *const allocator, const int32 intValue)
```
MikeHolman: I think Louis is right and we should use your template version of this. Otherwise we risk introducing memory regression for js

test/WasmSpec/rlexe.xml
```diff
@@ -892,4 +892,33 @@
...
+      <baseline>baselines/chakra_i64.baseline</baseline>
+      <compile-flags>-wasm -wasmfold -args testsuite-bin/chakra_i64.json -endargs</compile-flags>
+      <tags>exclude_xplat</tags>
+    </default>
+  </test>
+  <test>
+    <default>
+      <files>spec.js</files>
+      <baseline>baselines/int_literals.baseline</baseline>
+      <compile-flags>-wasm -wasmfold -args testsuite-bin/int_literals.json -endargs</compile-flags>
```
MikeHolman: with it on by default, we don't need explicit test variation with the flag

MikeHolman:
Overall looks good, just couple nits. And let's go with the 2 class version.

lib/Backend/GlobOpt.h
Krovatkin: The template version initially looked a little bit less complicated & verbose, because I was using inheriting c-tors. Unfortunately, some CI jobs use an older cl compiler (on win7) which gets confused by this feature.

test/WasmSpec/rlexe.xml
Krovatkin: ah, sorry about that! will fix shortly!

Krovatkin:
@MikeHolman Should I close this one since we decided to go w/ the 2-class version?

MikeHolman:
Yeah, I guess so

Krovatkin:
closing this in favour of PR #2597 fyi @MikeHolman  @arunetm

PR #2612, opened 2017-02-28
swb: fix over FieldWithBarrier

jianchun:
Made some batch replacement errors before in "FunctionBody.h". Only direct
FunctionBody fields need "FieldWithBarrier" (FunctionBody always uses
software write barrier). Change others to "Field".

### DIFF
--- a/lib/Runtime/Base/FunctionBody.h
+++ b/lib/Runtime/Base/FunctionBody.h
@@ -1304,8 +1304,8 @@ namespace Js
 
     class FunctionProxy;
 
-    typedef FieldWithBarrier(FunctionInfo*)* FunctionInfoArray;
-    typedef FieldWithBarrier(FunctionInfo*)* FunctionInfoPtrPtr;
+    typedef Field(FunctionInfo*)* FunctionInfoArray;
+    typedef Field(FunctionInfo*)* FunctionInfoPtrPtr;
 
     //
     // FunctionProxy represents a user defined function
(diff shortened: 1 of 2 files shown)


lib/Runtime/Base/FunctionBody.h
```diff
@@ -1779,12 +1779,12 @@ namespace Js
...
     private:
-        FieldWithBarrier(const byte*) m_functionBytes;
-        FieldWithBarrier(ByteCodeCache*) m_cache;
-        FieldWithBarrier(const char16 *) m_displayName;  // Optional name
-        FieldWithBarrier(uint) m_displayNameLength;
-        FieldWithBarrier(uint) m_displayShortNameOffset;
-        FieldWithBarrier(NativeModule *) m_nativeModule;
+        Field(const byte*) m_functionBytes;
+        Field(ByteCodeCache*) m_cache;
+        Field(const char16 *) m_displayName;  // Optional name
```
leirocks: this sometimes is recycler allocated, sometimes not, should use FieldWithBarrier.
check m_displayNameIsRecyclerAllocated and FunctionProxy::SetDisplayName() for detail

leirocks: or maybe not. looks SetDisplayName() is taking care of the barrier bit. did you run stress test?

jianchun: This is not "FunctionBody". It is "DeferDeserializeFunctionInfo" which is not allocated in write barrier memory.

leirocks: ah, OK

PR #2613, opened 2017-02-28
fix assertion on AuxPtr

leirocks:
While fixing a dead lock bug and adding assertion for that the work item queue should not be locked at the time of getting the auxPtrs. The assertion is wrong when JIT is collecting data in foreground thread causes stack probing and hence call dispose then re-enter script, which can call GetAuxPtrsWithLock. Fixing with checking if JIT is in foreground.
Also change to not call dispose in codeGen/gathering data, to prevent unpredicted reentrant issue.

Fix the real possible points of dead lock on etw rundown and remove IsLockedByAnyThread (which is hacky and not accurate, should not allow the etw rundown thread to updated the counters on function body, if any we should fix it)

### DIFF
--- a/lib/Runtime/Base/EtwTrace.cpp
+++ b/lib/Runtime/Base/EtwTrace.cpp
@@ -95,6 +95,9 @@ void EtwTrace::PerformRundown(bool start)
 
     while(threadContext != nullptr)
     {
+        // Take the auxPtrs updating lock, in case auxPtrs is expanding causes GC
+        // which locks Etw Rundown lock, hence dead lock.
+        AutoCriticalSection autoCs(FunctionProxy::GetLock());
         // Take etw rundown lock on this thread context
         AutoCriticalSection autoEtwRundownCs(threadContext->GetEtwRundownCriticalSection());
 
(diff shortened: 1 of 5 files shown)


lib/Runtime/Base/FunctionBody.cpp
```diff
@@ -130,7 +133,7 @@ namespace Js
         }
 
         // when setting ptr to null we never need to promote
-        AutoCriticalSection aucoCS(&GlobalLock);
+        AutoCriticalSection aucoCS(GetLock());
```
dilijev: Was this name supposed to be `autoCS`?

lib/Runtime/Base/EtwTrace.cpp
```diff
@@ -95,6 +95,9 @@ void EtwTrace::PerformRundown(bool start)
 
     while(threadContext != nullptr)
     {
+        // Take the auxPtrs updating lock, in case auxPtrs is expanding causes GC
+        // which locks Etw Rundown lock, hence dead lock.
+        AutoCriticalSection autoCs(FunctionProxy::GetLock());
```
dilijev: nit: consistency with casing of `autoCS` used elsewhere.

dilijev:
Oh whoops, late to the party. Feel free to ignore my nits.

lib/Runtime/Base/FunctionBody.cpp
leirocks: let me take it in my next PR. thanks!

PR #2615, opened 2017-03-01
Remove problematic (performance) assert.

mrkmarron:
Remove an assert that seems to cause performance irregularities and is not adding much value.


mrkmarron:
@dotnet-bot test OSX osx_osx_test_static please

PR #2616, opened 2017-03-01
WASM: Empty buffer error type + source refactor

Cellule:
By spec, we're supposed to throw a WebAssemblyCompile exception when the source buffer is empty. We were throwing a TypeError.

Add class WebAssemblySource to centralize how we create sources for WebAssembly

### DIFF
--- a/lib/Runtime/Language/WebAssemblySource.cpp
+++ b/lib/Runtime/Language/WebAssemblySource.cpp
@@ -1,0 +1,80 @@
+//-------------------------------------------------------------------------------------------------------
+// Copyright (C) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE.txt file in the project root for full license information.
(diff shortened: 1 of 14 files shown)


lib/Runtime/Language/WebAssemblySource.cpp
```diff
@@ -0,0 +1,85 @@
...
+    {
+        // ArrayBuffer and TypedArray return nullptr when the buffer is empty
+        // assign a dummy buffer to let the flow continue because this doesn't always result in an exception
+        buffer = emptyBuffer;
+        Assert(bufferLength == 0);
+        bufferLength = 0;
+    }
+
+    // copy buffer so external changes to it don't cause issues when defer parsing
+    byte* newBuffer = RecyclerNewArrayLeaf(scriptContext->GetRecycler(), byte, bufferLength);
```
MikeHolman: i think we shouldn't be creating array of length 0

PR #2619, opened 2017-03-01
rework JIT process memory errors

MikeHolman:
In case of memory exhaustion type scenarios, we don't want to crash the JIT process,  instead crash the runtime process, since the JIT may serve many runtimes.

Also, this resolves issue where we get memory violation due to runtime crashing. In such a case, now we will have runtime crash after the JIT returns

### DIFF
--- a/lib/Backend/NativeCodeGenerator.cpp
+++ b/lib/Backend/NativeCodeGenerator.cpp
@@ -3164,18 +3164,15 @@ bool NativeCodeGenerator::TryReleaseNonHiPriWorkItem(CodeGenWorkItem* workItem)
 void
 NativeCodeGenerator::FreeNativeCodeGenAllocation(void* address)
 {
-    if(this->backgroundAllocators)
+    if (JITManager::GetJITManager()->IsOOPJITEnabled())
     {
         ThreadContext * context = this->scriptContext->GetThreadContext();
-        if (JITManager::GetJITManager()->IsOOPJITEnabled())
-        {
-            // OOP JIT TODO: need error handling?
-            JITManager::GetJITManager()->FreeAllocation(context->GetRemoteThreadContextAddr(), (intptr_t)address);
-        }
-        else
-        {
-            this->backgroundAllocators->emitBufferManager.FreeAllocation(address);
-        }
+        HRESULT hr = JITManager::GetJITManager()->FreeAllocation(context->GetRemoteThreadContextAddr(), (intptr_t)address);
+        JITManager::HandleServerCallResult(hr, RemoteCallType::MemFree);
+    }
+    else if(this->backgroundAllocators)
+    {
+        this->backgroundAllocators->emitBufferManager.FreeAllocation(address);
     }
 }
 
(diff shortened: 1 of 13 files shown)


lib/Common/Memory/PageAllocator.cpp
```diff
@@ -977,29 +977,18 @@ PageAllocatorBase<TVirtualAlloc, TSegment, TPageSegment>::FillAllocPages(__in vo
...
     {
         // new pages are filled with zeros, old pages are filled with DbgMemFill
-        Assert(readBuffer[i] == 0 || readBuffer[i] == DbgMemFill);
-    }
-    if (!isLocalProc)
-    {
-        HeapDeleteArray(bufferSize, readBuffer);
+        Assert(localAddr[i] == 0 || localAddr[i] == DbgMemFill);
     }
+    this->GetVirtualAllocator()->FreeLocal(localAddr);
```
satheeshravi: >this->GetVirtualAllocator()->FreeLocal(localAddr); [](start = 4, length = 50)

is this just a return true?

lib/Common/Memory/CustomHeap.cpp
```diff
@@ -439,12 +439,12 @@ Allocation* Heap<TAlloc, TPreReservedAlloc>::AllocLargeObject(size_t bytes, usho
     size_t resultBytes = VirtualQueryEx(this->processHandle, address, &memBasicInfo, sizeof(memBasicInfo));
     if (resultBytes == 0)
     {
-        if (this->processHandle != GetCurrentProcess())
```
satheeshravi: >if (this->processHandle != GetCurrentProcess()) [](start = 8, length = 47)

any idea, why this check was there in the first place?

lib/Backend/NativeCodeGenerator.cpp
```diff
@@ -3202,23 +3199,9 @@ NativeCodeGenerator::QueueFreeNativeCodeGenAllocation(void* address)
     }
 
     // The foreground allocators may have been used
-    ThreadContext * context = this->scriptContext->GetThreadContext();
-    if(this->foregroundAllocators)
+    if(this->foregroundAllocators && this->foregroundAllocators->emitBufferManager.FreeAllocation(address))
     {
-        if (JITManager::GetJITManager()->IsOOPJITEnabled())
-        {
-            // TODO: OOP JIT, should we always just queue this in background?
```
satheeshravi: >OOP JIT, should we always just queue this in background? [](start = 21, length = 56)

can you add this as a comment ?

MikeHolman:
@dotnet-bot test this

lib/Common/Memory/PageAllocator.cpp
MikeHolman: for in proc, yes, for oop it unmaps the local view

---
In reply to: [104337005](https://github.com/Microsoft/ChakraCore/pull/2619#discussion_r104337005) [](ancestors = 104337005)

lib/Common/Memory/CustomHeap.cpp
MikeHolman: we didn't want to assert for oop jit, because it might be that runtime died, so here we threw

---
In reply to: [104337250](https://github.com/Microsoft/ChakraCore/pull/2619#discussion_r104337250) [](ancestors = 104337250)

PR #2622, opened 2017-03-01
fix simd buildbreak with linux

MikeHolman:
Recently merged change causes buildbreak on linux


MikeHolman:
Actually, I'm going to undo the change, as I also noticed an issue with this for OOP JIT

PR #2624, opened 2017-03-02
Fix mapping rest parameters when inlining a function with an inner inlinee

meg-gupta:
While mapping formals during inlining, ArgIn_Rest was handled incorrectly when it belonged to
another inlined function originally in the inlinee. For such ArgIn_Rest instructions we should generate code depending on the formalCount of ArgIn_Rest's function instead of the formalCount of the current inlinee.

### DIFF
--- a/lib/Backend/Inline.cpp
+++ b/lib/Backend/Inline.cpp
@@ -4675,16 +4675,22 @@ Inline::MapFormals(Func *inlinee,
             {
                 break;
             }
+
+            int excess;
+            Js::ArgSlot restFuncFormalCount = 0;
             if (instr->m_func != inlinee)
             {
-                // this can happen only when we are inlining a function which has inlined an apply call with the arguments object
-                formalCount = instr->m_func->GetJITFunctionBody()->GetInParamsCount();
+                restFuncFormalCount = instr->m_func->GetJITFunctionBody()->GetInParamsCount();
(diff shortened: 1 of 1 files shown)


meg-gupta:
@rajatd and @ThomsonTan Can you please review ?

https://github.com/Microsoft/ChakraCore/pull/2202 was the PR that resulted into this problem.

meg-gupta:
Looking at the test failure.

ThomsonTan:
Also is it possible to add a test case to expose this?

lib/Backend/Inline.cpp
```diff
@@ -4677,8 +4677,7 @@ Inline::MapFormals(Func *inlinee,
             }
             if (instr->m_func != inlinee)
             {
-                // this can happen only when we are inlining a function which has inlined an apply call with the arguments object
-                formalCount = instr->m_func->GetJITFunctionBody()->GetInParamsCount();
```
satheeshravi: >formalCount [](start = 16, length = 11)

So the earlier fix about formalCount being wrong during NewScArray creation - is also taken care? - can you also pls  verify that?

meg-gupta: Yes, the updated commit handles correct codegen for ArgIn_Rest.

rajatd:
:shipit:

PR #2626, opened 2017-03-03
Set the implicit call flag on partial length change

Penguinwizzard:
In the presence of non-configurable, numeric properties, setting the
length of an ES5Array can only decrease the length to the maximum of
the highest numerically indexed length property less than the length
and the new length. We were overeagerly propagating the newer length
with the copy prop machinery, and not checking that we actually were
able to assign this length value correctly. This changes the helpers
for setting the length property on ES5Arrays to set the implicitcall
flag when the new length isn't fully applied.

Relevant test case included.

### DIFF
--- a/lib/Runtime/Library/ES5Array.cpp
+++ b/lib/Runtime/Library/ES5Array.cpp
@@ -221,7 +221,11 @@ namespace Js
             }
 
             uint32 newLen = ToLengthValue(value, scriptContext);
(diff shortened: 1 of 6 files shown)


rajatd:
This is much better than what we were discussing. 👍

Penguinwizzard:
@dotnet-bot test osx and Ubuntu please

Penguinwizzard:
@dotnet-bot test osx please

Penguinwizzard:
@dotnet-bot test Ubuntu please

PR #2628, opened 2017-03-03
Add a StatementDebugBreak flag for debug builds

Penguinwizzard:
This flag allows you to insert a debugger break on a given statment.

(Also enables GenerateDebugBreak on arm, in non-macro form).

Note: was originally rolled into the es5 array length cache fix, split
for atomicity of pr purpose.

### DIFF
--- a/lib/Backend/Lower.cpp
+++ b/lib/Backend/Lower.cpp
@@ -2900,9 +2900,10 @@ Lowerer::LowerRange(IR::Instr *instrStart, IR::Instr *instrEnd, bool defaultDoFa
             // If we have a JITStatementBreakpoint, then we should break on this statement
             {
                 uint32 statementIndex = instr->AsPragmaInstr()->m_statementIndex;
-                if (Js::Configuration::Global.flags.StatementDebugBreak.Contains(instr->m_func->GetFunctionNumber(), statementIndex))
+                if (Js::Configuration::Global.flags.StatementDebugBreak.Contains(instr->m_func->GetSourceContextId(), instr->m_func->GetLocalFunctionId(), statementIndex))
                 {
                     IR::Instr* tempinstr = instr;
+                    Assert(tempinstr != nullptr);
                     // go past any labels, and then add a debug breakpoint
                     while (tempinstr->m_next != nullptr && tempinstr->m_next->m_opcode == Js::OpCode::Label)
                     {
--- a/lib/Backend/Lower.cpp
+++ b/lib/Backend/Lower.cpp
@@ -2896,7 +2896,23 @@ Lowerer::LowerRange(IR::Instr *instrStart, IR::Instr *instrEnd, bool defaultDoFa
         case Js::OpCode::StatementBoundary:
             // This instruction is merely to help convey source info through the IR
             // and eventually generate the nativeOffset maps.
-            break;
+#if DBG_DUMP && DBG
+            // If we have a JITStatementBreakpoint, then we should break on this statement
+            {
+                uint32 statementIndex = instr->AsPragmaInstr()->m_statementIndex;
+                if (Js::Configuration::Global.flags.StatementDebugBreak.Contains(instr->m_func->GetFunctionNumber(), statementIndex))
+                {
+                    IR::Instr* tempinstr = instr;
+                    // go past any labels, and then add a debug breakpoint
+                    while (tempinstr->m_next != nullptr && tempinstr->m_next->m_opcode == Js::OpCode::Label)
+                    {
+                        tempinstr = tempinstr->m_next;
+                    }
+                    this->m_lowererMD.GenerateDebugBreak(tempinstr);
+                }
+            }
+#endif
+        break;
 
         case Js::OpCode::BailOnNotPolymorphicInlinee:
             instrPrev = LowerBailOnNotPolymorphicInlinee(instr);
--- a/lib/Backend/arm/LowerMD.cpp
+++ b/lib/Backend/arm/LowerMD.cpp
@@ -9303,7 +9303,7 @@ LowererMD::LowerTypeof(IR::Instr* typeOfInstr)
 //
 void LowererMD::GenerateDebugBreak( IR::Instr * insertInstr )
 {
-    IR::Instr *int3 = IR::Instr::New(Js::OpCode::DEBUGBREAK, insertInstr->m_func);\
-    insertInstr->InsertBefore(int3);\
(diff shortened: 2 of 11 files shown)


lib/Backend/arm/LowerMD.cpp
```diff
@@ -9295,4 +9295,15 @@ LowererMD::LowerTypeof(IR::Instr* typeOfInstr)
...
\ No newline at end of file
+}
+
+#if DBG
+//
+// Helps in debugging of fast paths.
+//
+void LowererMD::GenerateDebugBreak( IR::Instr * insertInstr )
+{
+    IR::Instr *int3 = IR::Instr::New(Js::OpCode::DEBUGBREAK, insertInstr->m_func);\
```
ThomsonTan: Better call it debugBreakInstr? Also remove the tailing `\`?

lib/Backend/Lower.cpp
```diff
@@ -2896,7 +2896,23 @@ Lowerer::LowerRange(IR::Instr *instrStart, IR::Instr *instrEnd, bool defaultDoFa
         case Js::OpCode::StatementBoundary:
             // This instruction is merely to help convey source info through the IR
             // and eventually generate the nativeOffset maps.
-            break;
+#if DBG_DUMP && DBG
+            // If we have a JITStatementBreakpoint, then we should break on this statement
```
ThomsonTan: This functionality is very useful especially to debug some large JIT'd code. I am wondering could we design it as generating `__debugbreak` for some special source statement, like `debugger;`? It is more easy for revising the location of bp. But the a drawback is it makes break unmodified JS code hard (like for some live website).

lib/Backend/Lower.cpp
```diff
@@ -2896,7 +2896,23 @@ Lowerer::LowerRange(IR::Instr *instrStart, IR::Instr *instrEnd, bool defaultDoFa
...
+            {
+                uint32 statementIndex = instr->AsPragmaInstr()->m_statementIndex;
+                if (Js::Configuration::Global.flags.StatementDebugBreak.Contains(instr->m_func->GetFunctionNumber(), statementIndex))
```
ThomsonTan: >GetFunctionNumber() [](start = 96, length = 19)

Use source context Id and local function Id could be better than function number.

lib/Backend/Lower.cpp
```diff
@@ -2896,7 +2896,23 @@ Lowerer::LowerRange(IR::Instr *instrStart, IR::Instr *instrEnd, bool defaultDoFa
...
+                {
+                    IR::Instr* tempinstr = instr;
+                    // go past any labels, and then add a debug breakpoint
+                    while (tempinstr->m_next != nullptr && tempinstr->m_next->m_opcode == Js::OpCode::Label)
```
ThomsonTan: >tempinstr [](start = 27, length = 9)

Could it be the end statement so tempinstr can be `nullptr`?

lib/Backend/arm/LowerMD.cpp
Penguinwizzard: The naming as "GenerateDebugBreak" is to have it match up with the name of the x86/amd64 version at LowerMDShared.cpp:7897. The reason that this function needed to be added here was that while the code added to Lower.cpp built on x86 and amd64, it didn't build on ARM due to this symbol not being defined in the platform's LowerMD. I've removed the trailing '\\'; it was due to copying the break generation from an arm-specific preprocessor macro.

lib/Backend/Lower.cpp
Penguinwizzard: It's not necessarily an either/or situation; this functionality should be useful on its own, at least, for the reasons you mentioned earlier (work with unmodified JS code), and additionally that it should allow you to keep much more of the pipeline in a steady state (since a debugbreak instruction might have different semantics or effects on earlier stages in the pipeline, at the very least by changing parse structures).

lib/Backend/Lower.cpp
Penguinwizzard: I've added the necessary machinery to have triples in a command line argument (in this case, sourceContextID, localFunctionID, statementIndex), and changed this check to use those values.

lib/Backend/Lower.cpp
Penguinwizzard: I don't believe so; tempinstr is initialized to instr, which, if null, would cause issues much earlier in the surrounding function, and any values which might be assigned to tempinstr are first checked for nullity. I've added an assert in the new version though.

rajatd:
Could you provide a snippet/add a test case to show the usage of this flag?

Penguinwizzard:
I can't add a test case (due to this flag causing a crash without a debugger), but I can provide an example of a use case:

Say I want to break when I get to line 3281 of octane\box2d.js. A quick look at any of the dump flags will tell you that this is sourceContextId 1, localFunctionId 332, statement 8. When you run ch with "-statementdebugbreak:1,332,8 -bgjit- -oopjit-" in VS, you'll hit a breakpoint in the output jitted code, conveniently at the top of the code generated by statement 8. This isn't always perfectly accurate (as stuff can get moved around a bit), but due to being very late in the process it has a low likelihood of impacting reproducibility of results. (If you want to get a little fancier, you can also do "-statementdebugbreak:1,332,8;1,332,9 -bgjit- -oopjit-", which will put an 'int 3' breakpoint instruction both before and after your statement - nice for outlining parts of code that you want to look at.)

lib/Backend/arm/LowerMD.h
```diff
@@ -92,6 +92,9 @@ class LowererMD
             IR::Instr *     LowerLdEnv(IR::Instr *instr);
             IR::Instr *     LowerLdSuper(IR::Instr * instr, IR::JnHelperMethod helperOpCode);
             IR::Instr *     GenerateSmIntPairTest(IR::Instr * instrInsert, IR::Opnd * opndSrc1, IR::Opnd * opndSrc2, IR::LabelInstr * labelFail);
+#if DBG
+    static  void            GenerateDebugBreak( IR::Instr * insertInstr );
```
ThomsonTan: >GenerateDebugBreak( IR::Instr * insertInstr ); [](start = 28, length = 46)

Remove spaces around parenthesis to be consistent?

ThomsonTan:
:shipit:

lib/Backend/arm/LowerMD.h
Penguinwizzard: Good catch, I've fixed this (and rebased to catch up to master) with the new commit.

PR #2629, opened 2017-03-03
fold pattern of a+(b<<pow2) into lea

MikeHolman:
This pattern shows up a fair bit in wasm and asm.js code. It's about a 3% win in zlib, 2% in jetstream, and 1% in wasm unity.

I added peep to our PreLowerPeeps phase for catching this.

It also required adding support in the encoder and couple other places to allow IndirOpnds to not have a baseOpnd post-lower.


lib/Backend/PreLowerPeeps.cpp
```diff
@@ -28,6 +35,139 @@ IR::Instr *Lowerer::PreLowerPeepInstr(IR::Instr *instr, IR::Instr **pInstrPrev)
...
+IR::Instr *
+Lowerer::PeepShiftAdd(IR::Instr *instrAdd)
+{
+    // Peep:
+    //    t1 = SHL X, 0|1|2|3    / t1 = MUL X, 1|2|4|8
+    //    t2 = ADD t1, Y
+    //
+    // Into:
+    //    t2 = LEA [X * scale + Y]
+
```
Krovatkin: @MikeHolman I'm really sorry to bother you, but I was wondering if you could point me to where we double check that there are no subsequent uses of `t1` **after** `t2 = ... ` ?

MikeHolman: No problem! The isDead flag should still be valid here, which is set at point of last use. So I check against that. Ideally what I want is a single use flag, but isDead combined with walking the basic block is sufficient, if somewhat inefficient.

Krovatkin: Ah, that's subtle! Thank you!

> Ideally what I want is a single use flag, but isDead combined with walking the basic block is sufficient, if somewhat inefficient.

I wish we had DU & UD chains.

lib/Backend/IR.cpp
```diff
@@ -2859,7 +2860,7 @@ Instr::FindRegUse(StackSym *sym)
         else if (src1->IsIndirOpnd())
         {
             IR::IndirOpnd *indirOpnd = src1->AsIndirOpnd();
-            if (indirOpnd->GetBaseOpnd()->m_sym == sym)
+            if (indirOpnd->GetBaseOpnd() && indirOpnd->GetBaseOpnd()->m_sym == sym)
```
ThomsonTan: >indirOpnd->GetBaseOpnd() [](start = 16, length = 24)

In which scenario there could be null base opnd for indir opnd?

lib/Backend/Opnd.cpp
```diff
@@ -2288,6 +2285,35 @@ IndirOpnd::New(RegOpnd *baseOpnd, RegOpnd *indexOpnd, byte scale, IRType type, F
 ///----------------------------------------------------------------------------
 
 IndirOpnd *
+IndirOpnd::New(RegOpnd *indexOpnd, int32 offset, byte scale, IRType type, Func *func)
+{
+    IndirOpnd * indirOpnd;
+
+    indirOpnd = JitAnew(func->m_alloc, IR::IndirOpnd);
```
ThomsonTan: >IR:: [](start = 39, length = 4)

nit: IR:: prefix is unnecessary. Add it might cause some confuse because `InDirOpnd` is the current class.

lib/Backend/Opnd.cpp
```diff
@@ -3180,11 +3207,14 @@ Opnd::Dump(IRDumpFlags flags, Func *func)
         IndirOpnd *indirOpnd = this->AsIndirOpnd();
 
         Output::Print(_u("["));
-        indirOpnd->GetBaseOpnd()->Dump(flags, func);
+        if (indirOpnd->GetBaseOpnd())
+        {
+            indirOpnd->GetBaseOpnd()->Dump(flags, func);
+            Output::Print(_u("+"));
```
ThomsonTan: >Output::Print(_u("+")); [](start = 12, length = 23)

Better output `null + index` to be aware that it is an indir without baseopnd?

mmitche:
@dotnet-bot test this please

lib/Backend/IR.cpp
MikeHolman: Could be null in case addition is with constant. There we will need index * scale + displacement.

lib/Backend/PreLowerPeeps.cpp
```diff
@@ -28,6 +35,139 @@ IR::Instr *Lowerer::PreLowerPeepInstr(IR::Instr *instr, IR::Instr **pInstrPrev)
...
+    if (!instrAdd->GetDst()->IsRegOpnd())
+    {
+        return instrAdd;
+    }
+
+    IR::Opnd * src2 = instrAdd->GetSrc2();
+    IR::Opnd * src1 = instrAdd->GetSrc1();
+    IR::Instr * resultInstr = TryShiftAdd(instrAdd, src1, src2);
+    if (resultInstr->m_opcode == Js::OpCode::Add_I4)
```
Cellule: Here you assume that the opcode was `Add_I4` before calling `TryShiftAdd`. Should put an assert at the top of the function.

lib/Backend/PreLowerPeeps.cpp
```diff
@@ -28,6 +35,139 @@ IR::Instr *Lowerer::PreLowerPeepInstr(IR::Instr *instr, IR::Instr **pInstrPrev)
...
+        // if any branch between def-use, don't do peeps on it because branch target might depend on the def
+        if (instrIter->IsBranchInstr())
+        {
+            return instrAdd;
+        }
+        if (instrIter->HasBailOutInfo())
+        {
+            return instrAdd;
+        }
+        if (instrIter->FindRegDef(instrDef->GetSrc1()->AsRegOpnd()->m_sym))
```
Cellule: > instrDef->GetSrc1()->AsRegOpnd()->m_sym

Should make this a variable outside the loop

lib/Backend/PreLowerPeeps.cpp
```diff
@@ -28,6 +35,139 @@ IR::Instr *Lowerer::PreLowerPeepInstr(IR::Instr *instr, IR::Instr **pInstrPrev)
...
+        {
+            return instrAdd;
+        }
+        if (instrIter->FindRegUse(opndFold->AsRegOpnd()->m_sym))
```
Cellule: > opndFold->AsRegOpnd()->m_sym

Should make this a variable outside the loop

lib/Backend/PreLowerPeeps.cpp
```diff
@@ -28,6 +35,139 @@ IR::Instr *Lowerer::PreLowerPeepInstr(IR::Instr *instr, IR::Instr **pInstrPrev)
...
+        {
+            return instrAdd;
+        }
+    } NEXT_INSTR_IN_RANGE;
+
+    IR::IndirOpnd * leaOpnd = nullptr;
+    if (opndAdd->IsInt32() && opndAdd->IsIntConstOpnd())
```
Cellule: nit: I would swap the 2 blocks as checking for IsRegOpnd is simpler than making sure it's an int32 const

lib/Backend/i386/EncoderMD.cpp
```diff
@@ -337,26 +337,36 @@ EncoderMD::EmitModRM(IR::Instr * instr, IR::Opnd *opnd, BYTE reg1)
...
+        AssertMsg(!indexOpnd || indexOpnd->GetReg() != RegESP, "ESP cannot be the index of an indir.");
+        if (baseOpnd == nullptr)
         {
+            Assert(indexOpnd != nullptr);
             regIndex = this->GetRegEncode(indexOpnd);
-            *(m_pc++) = (this->GetMod(indirOpnd, &dispSize) | reg1 | 0x4);
-            *(m_pc++) = (((indirOpnd->GetScale() & 3) << 6) | ((regIndex & 7) << 3) | (regBase & 7));
+            dispSize = 4;
+            *(m_pc++) = (0x00 | reg1 | 0x4);
+            *(m_pc++) = (((indirOpnd->GetScale() & 3) << 6) | ((regIndex & 7) << 3) | 0x5);
```
ThomsonTan: >0x5 [](start = 86, length = 3)

What does 0x5 mean here as it is difference in else statement below?

lib/Backend/PreLowerPeeps.cpp
```diff
@@ -13,6 +13,13 @@ IR::Instr *Lowerer::PreLowerPeepInstr(IR::Instr *instr, IR::Instr **pInstrPrev)
 
     switch (instr->m_opcode)
     {
+#if defined(_M_IX86) || defined(_M_X64)
+        // this sort of addressing mode magic only applies to x86 platforms
```
ThomsonTan: >// this sort of addressing mode magic only applies to x86 platforms [](start = 8, length = 67)

ARM has some similar instruction form as `ADD r4, r4, r4, LSL #4`, could we also peep that?

lib/Backend/PreLowerPeeps.cpp
MikeHolman: I assumed the compiler would realize this is loop invariant, but I can move outside loop.

lib/Backend/PreLowerPeeps.cpp
MikeHolman: I did this on purpose because I thought this is much more likely to fail.

lib/Backend/i386/EncoderMD.cpp
MikeHolman: The 5 here means a disp32 with no base if the MOD is 0x00. (From Intel developer manual section 2.1.5 table 2-3)

lib/Backend/Opnd.cpp
```diff
@@ -2244,9 +2244,6 @@ IndirOpnd::New(RegOpnd *baseOpnd, RegOpnd *indexOpnd, IRType type, Func *func)
 {
     IndirOpnd * indirOpnd;
 
-    AssertMsg(baseOpnd, "An IndirOpnd needs a valid baseOpnd.");
```
MikeHolman: My new one covers index * scale + offset, but not base + index * scale. Technically it's only required to remove the second assert though.

lib/Backend/Opnd.cpp
MikeHolman: I thought about the placement of + being weird in some scenarios, but felt people can figure it out. However, as Tom suggests I can put null here, which seems reasonable, since otherwise we can't distinguish between index and base in case scale is 0. So I'll go with that, solution which should also solve issue of the +

lib/Backend/IR.cpp
dilijev: Is GetBaseOpnd() const? If not, could you save the result to a variable to avoid a potential deoptimization from the compiler?

lib/Backend/Opnd.cpp
dilijev: Same comment as above about result of GetBaseOpnd -- save to variable before the if

lib/Backend/PreLowerPeeps.cpp
dilijev: Any data on how often the respective cases check and fail versus pass?

MikeHolman: I don't have data, but almost everything is a RegOpnd (probably >90%). Very little are IntConstOpnd (likely 1% or less).

Cellule: I think that's an argument in favor to swap the 2 blocks. Usually we want to have the most common case first (if we can easily check for it)

MikeHolman: Oh I misunderstood, I thought you were asking about the escape condition where it's not RegOpnd or IntConstOpnd. Sure I can swap this around.

lib/Backend/PreLowerPeeps.cpp
MikeHolman: Ah I didn't know about that. I'll look into adding it.

MikeHolman: Ah I didn't know about that. I'll look into adding it.

MikeHolman: It seems like our ARM encoder would not support it without a lot of work, and since this is mostly for asm.js/wasm where we don't have arm support i'll open an issue and leave this for now.

lib/Backend/PreLowerPeeps.cpp
ThomsonTan: Ok. Makes sense.

Cellule:
:lgtm:

---

Reviewed 13 of 13 files at r1, 3 of 3 files at r2.
Review status: all files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/microsoft/chakracore/2629#-:-KekGao-ZGMdDBpNIaQ-:bnfp4nl)*

PR #2631, opened 2017-03-03
Fixing and refactoring the load of Func object operand in lower code

satheeshravi:
A hot fix for an ARM bug related to StackArgs opt was provided in the source depot branch sometime back to unblock the RI. 
This change is a more elaborate fix which has other refactoring around the fix.

### DIFF
--- a/lib/Backend/Lower.cpp
+++ b/lib/Backend/Lower.cpp
@@ -308,30 +308,10 @@ Lowerer::LowerRange(IR::Instr *instrStart, IR::Instr *instrEnd, bool defaultDoFa
                     // s3 = formals are let decls
                     this->m_lowererMD.LoadHelperArgument(instr, IR::IntConstOpnd::New(currFunc->GetHasNonSimpleParams() ? TRUE : FALSE, TyUint8, currFunc));
 
-                    if (currFunc->IsInlinee())
-                    {
-                        // s2 = current function.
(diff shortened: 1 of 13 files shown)


lib/Backend/Lower.cpp
```diff
@@ -12791,6 +12771,63 @@ void Lowerer::LowerBailOnCreatedMissingValue(IR::Instr *const instr, const bool
...
+        paramOpnd = func->GetInlineeFunctionObjectSlotOpnd();
+    }
+    else
+    {
+#if defined(_M_ARM)
+        StackSym * paramSym = this->m_lowererMD.GetImplicitParamSlotSym(0);
+#else
+        StackSym *paramSym = StackSym::New(TyMachReg, this->m_func);
+        this->m_func->SetArgOffset(paramSym, 2 * MachPtr);
+        this->m_func->SetHasImplicitParamLoad();
```
rajatd: Can we have GetImplicitParamSlotSym on the Lowerer object and pass 0 for arm and 2 for x86/64?

rajatd:
:shipit:

lib/Backend/Lower.cpp
satheeshravi: Discussed offline.

satheeshravi:
@dotnet-bot please test OSX static_osx_osx_debug

PR #2632, opened 2017-03-03
Fix Hoisting ValueType of String concat

ThomsonTan:
When hoisting `SetConcatStrMultiItemBE` to loop landingpad, the type of src operands could be loosed to be non-string (this was ensured by some bailout in the loop, but the type guarantee will disappear in landing pad if the bailout was not also hoisted). In this case we need add `Conv_PrimStr` back to make sure `String` is passed to `SetConcatStrMultiItemBE` as src operands.

### DIFF
--- a/lib/Backend/GlobOpt.cpp
+++ b/lib/Backend/GlobOpt.cpp
@@ -14811,7 +14811,7 @@ GlobOpt::ToTypeSpecUse(IR::Instr *instr, IR::Opnd *opnd, BasicBlock *block, Valu
                     ))
                 {
                     Assert(!this->IsLoopPrePass());
-                    this->OptHoistInvariant(newInstr, block, block->loop, val, val, false);
+                    this->OptHoistInvariant(newInstr, block, block->loop, val, val, nullptr, false);
                 }
             }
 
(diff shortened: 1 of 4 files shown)


ThomsonTan:

lib/Backend/GlobOpt.cpp
```diff
@@ -19198,30 +19198,54 @@ GlobOpt::OptDstIsInvariant(IR::RegOpnd *dst)
...
+        BasicBlock* landingPad = loop->landingPad;
         Value* opndValueInLandingPad = FindValue(landingPad->globOptData.symToValueMap, opndSym);
         Assert(opndVal->GetValueNumber() == opndValueInLandingPad->GetValueNumber());
 
-        opnd->SetValueType(opndValueInLandingPad->GetValueInfo()->Type());
+        ValueType opndValueTypeInLandingPad = opndValueInLandingPad->GetValueInfo()->Type();
+
+        if (srcOpnd->GetValueType() != opndValueTypeInLandingPad)
+        {
+            if (instr->m_opcode == Js::OpCode::SetConcatStrMultiItemBE)
```
rajatd: We are not checking if the Conv_PrimStr instr is actually required, are we? I mean, if the bailout ensuring the string type was also hoisted, we wouldn't need conv_primstr.

ThomsonTan: If there is bailout to ensure string type is also hoisted, the value type of src1 for `SetConcatStrMultiItemBE` should be `string`, then a new `Conv_PrimStr` will not be prepended here.

---
In reply to: [104359344](https://github.com/Microsoft/ChakraCore/pull/2632#discussion_r104359344) [](ancestors = 104359344)

rajatd: Makes sense.

PR #2639, opened 2017-03-06
Don't assert again if we are already in assert.  Just FailFast.

curtisman:


curtisman:
@tcare Please review.

tcare:
Looks good. Was this hitting anywhere?

PR #2641, opened 2017-03-06
fix an issue on scriptContext close on OOM

leirocks:
while initializing scriptContext, OOM can happen between initializeing debugContext and debugContext->diagProbesContainer.  should check this while closing scriptContext

### DIFF
--- a/lib/Runtime/Base/ScriptContext.cpp
+++ b/lib/Runtime/Base/ScriptContext.cpp
@@ -658,8 +658,11 @@ namespace Js
         this->EnsureClearDebugDocument();
         if (this->debugContext != nullptr)
         {
-            this->debugContext->GetProbeContainer()->UninstallInlineBreakpointProbe(NULL);
-            this->debugContext->GetProbeContainer()->UninstallDebuggerScriptOptionCallback();
+            if(this->debugContext->GetProbeContainer())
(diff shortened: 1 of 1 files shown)


lib/Runtime/Base/ScriptContext.cpp
```diff
@@ -658,8 +658,11 @@ namespace Js
         this->EnsureClearDebugDocument();
         if (this->debugContext != nullptr)
         {
-            this->debugContext->GetProbeContainer()->UninstallInlineBreakpointProbe(NULL);
-            this->debugContext->GetProbeContainer()->UninstallDebuggerScriptOptionCallback();
+            if(this->debugContext->GetProbeContainer())
```
akroshg: what about other places in the same file? say InitializePostGlobal?

leirocks: that's a thowing OOM, you won't get there if OOM happens.  here it's in the cleaning path.

akroshg: Makes sense ... LGTM

---
In reply to: [104546815](https://github.com/Microsoft/ChakraCore/pull/2641#discussion_r104546815) [](ancestors = 104546815)

PR #2642, opened 2017-03-06
Set promise continuation callback so that promise task can be enqueued correctly.

atulkatti:


atulkatti:
@Microsoft/chakra-developers 
This has already been reviewed internally.

PR #2644, opened 2017-03-06
perf test script fixes

jianchun:
Add "WScript.Echo" definitions to all SunSpider js files to support
non-Chakra hosts.

`perftest.pl`:
- Fix dir name cases.
- Change to "-?" in test_switch ("/?" not supported on xplat).
- Add "-testfile:<file>" switch to save perf test result to a file as well,
  by default to "perftest???.txt". This is useful if we want to use the
  test numbers as baseline for subsequent perf runs.
- Add "-skipCheckSwitch" to support non-Chakra hosts. It is not meaningful
  to test for Chakra switches, and in addition may run into trouble (host
  may start command loop instead of exit).


test/benchmarks/SunSpider/3d-cube.js
```diff
@@ -20,9 +20,16 @@
...
  OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
  (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
- OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE. 
+ OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 */
 
+if(typeof(WScript) === "undefined")
+{
+    var WScript = {
+        Echo: print
```
obastemur: can we use `console.log` here instead? It works both for node and ch

obastemur: or -> `(typeof(console) === 'undefined' ? print : console.log)`

jianchun: @obastemur That looks better. Can we do that in a future change? This one is copied from other benchmark files. To make that change we better change them all.

obastemur:
:shipit:

jianchun:
Thanks @obastemur !

Issue #2645, opened 2017-03-06
ES6 module : unable to find an object from imports


akroshg:
If we have import and export for a module and if that module is exported as namespace object then we unable to find that export from the namespace object.
consider this
**test.js**
```js
WScript.LoadScriptFile("test1.js", "module");
```

**test1.js (R)**
```js
import * as module1 from "module1.js";

function test() {
    return module1.simpleExport();
}

console.log(test());
```

**module1.js (D1)**
```js
import * as module2 from "module2.js";

export function simpleExport() { return 'exported'; }
```

**module2.js (D2)**
```js
export function simpleExport() { return 'module2.exported'; }
```

ch -es6module test.js produces
TypeError: Object doesn't support property or method 'simpleExport'
   at test (:4:5)
   at Anonymous function (:7:1)
   at module (:1:1)

akroshg:
Fix pushed.

PR #2646, opened 2017-03-07
Module export field issue with namespace.

akroshg:
The export function is not put to the module namespace object as the logic of namespace's initialize has an issue. We should be putting the exports to the namespace if we don't find it in the importlist. However the logic was saying otherwise. Fixed that.


akroshg:
Fixes #2645 . @boingoing please have a look.

akroshg:
@dotnet-bot test OSX static_osx_osx_test

akroshg:
thanks @boingoing for helping me in fix.

PR #2647, opened 2017-03-07
Fix miss OptInstr for InlineeEnd after upward it to follow BailOnNoProfile.

ThomsonTan:
In the previous fix #2582 which upwards `InlineeEnd` to follow `BailOnNoProfile`, it missed to call `OptInstr` on the upwarded `InlineeEnd` because this is in forward pass. Also added checks to ensure each block is visited one time at most in upward traversal.

### DIFF
--- a/lib/Backend/FlowGraph.cpp
+++ b/lib/Backend/FlowGraph.cpp
@@ -1722,7 +1722,7 @@ FlowGraph::RemoveUnreachableBlocks()
 
 // If block has no predecessor, remove it.
 bool
-FlowGraph::RemoveUnreachableBlock(BasicBlock *block, GlobOpt * globOpt)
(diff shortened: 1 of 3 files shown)


ThomsonTan:
@dotnet-bot test arm_release please.

ThomsonTan:
@dotnet-bot test Windows arm_release please.

rajatd:
👍 
But I would like for us to come up with a fix that doesn't involve messing with InlineeEnds

ThomsonTan:
@rajatd thanks. Yes, create a new opcode will make it more separated than `InlineeEnd`, but anyway, the upward search process and other handling here are still necessary for the new opcode so create that  based on the current change will make it easier.

PR #2648, opened 2017-03-08
Change error msg for Array.sort (Fix #2411)

### LINKED ISSUE #2411: `[1, 2].sort({})` has unhelpful error message
`[1, 2].sort(true)` and `[1, 2].sort({})` throw a TypeError with the `message` "Array.prototype.sort: argument is not a JavaScript object".

This is unhelpful for `true`, and incorrect for `{}`. A better error message would perhaps be, `Array.prototype.sort: comparefn argument must be a function or undefined".

(discovered while looking into https://github.com/tc39/ecma262/pull/785)

xiaoyinl:
Unfortunately, there is no "argument must be a function or undefined" message defined in `rterrors.h`. The closest one is `JSERR_FunctionArgument_NeedFunction`: argument is not a Function object.

See #2411 for the issue.

### DIFF
--- a/lib/Runtime/Library/JavascriptArray.cpp
+++ b/lib/Runtime/Library/JavascriptArray.cpp
@@ -6594,7 +6594,7 @@ Case0:
                 bool useDefaultComparer = typeId == TypeIds_Undefined;
                 if (!useDefaultComparer)
                 {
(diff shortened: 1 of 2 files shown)


xiaoyinl:
@ljharb @dilijev @MikeHolman Could you review this patch?

dilijev:
Doing a test against internal test baselines and will merge this soon, thanks.

PR #2650, opened 2017-03-08
Disable float typespec of StElem, when it is off

meg-gupta:


meg-gupta:
Ping

lib/Backend/GlobOpt.cpp
```diff
@@ -13477,8 +13477,11 @@ GlobOpt::TypeSpecializeStElem(IR::Instr ** pInstr, Value *src1Val, Value **pDstV
     case ObjectType::Float64VirtualArray:
     case ObjectType::Float64MixedArray:
     Float64Array:
-        toType = TyFloat64;
-        break;
+    if (this->DoFloatTypeSpec())
```
LouisLaf: We also need to disable all of the int cases if DoAggressiveIntTypeSpec or DoFloatTypeSpec is disabled, like I did in ValueNumberLdElemDst()

lib/Backend/GlobOpt.cpp
```diff
@@ -13450,7 +13450,10 @@ GlobOpt::TypeSpecializeStElem(IR::Instr ** pInstr, Value *src1Val, Value **pDstV
...
     case ObjectType::Int32MixedArray:
     Int32Array:
-        toType = TyInt32;
+        if (this->DoAggressiveIntTypeSpec() || this->DoFloatTypeSpec())
+        {
+            toType = TyInt32;
+        }
         break;
 
     case ObjectType::Uint32Array:
```
meg-gupta: @LouisLaf When the sym is already int/float specialized, I have not added the check to see if float/int typespec was disabled. This should be okay ?

LouisLaf:
:shipit:

PR #2651, opened 2017-03-08
add CSE attributes for conv_prim to differentiate on signedness

MikeHolman:
This was long-standing bug on asm.js and wasm where CSE would get incorrectly kick in when converting the same sym with signed and unsigned conversion.


Cellule:
Other than the signedness , I wonder if different types should not be CSE'd altogether for ConvPrim.

Cellule:
LGTM after offline discussion

PR #2652, opened 2017-03-08
Remove extra optimization on InlineeEnd which is not necessary

ThomsonTan:
We do `OptInstr` on upwarded `InlineeEnd`, but the block is already been removed. So it is actually no need to do the extra optimizations, just need `RecordInlineeFrame`.


rajatd:
LGTM

PR #2654, opened 2017-03-09
WASM: Use wabt in chakra

Cellule:
We've been lacking a good way to convert wast to wasm. In the past we've used wabt to convert wast files and checkin the compiled version.
Since Wabt looks like something to stay and is reliable, we decided to integrate it directly in chakra to help do testing.
Right now the only api exposed is `WebAssembly.wabt.convertWast2Wasm(<text>)` but in the near future I would like to expose a method to convert spec like wast files with asserts.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/2654)


lib/Runtime/Library/WabtInterface.cpp
```diff
@@ -0,0 +1,206 @@
...
+#include "ast-parser.h"
+#include "binary-writer.h"
+#include "binary-writer-spec.h"
+#include "resolve-names.h"
+#include "validator.h"
+#include "Codex/Utf8Helper.h"
+using namespace wabt;
+
+ArenaAllocator* tempArenaPtr;
+struct utf8Allocator
```
MikeHolman: i don't really like this, and it seems incorrect e.g. in case of multiple threads running. can we update the utf8 api to be able to use non-static allocator?

Cellule: I totally agree, I didn't like the solution either, but I wanted to make things work and I forgot to clean this up afterward. Thanks for pointing it out

Cellule:
@obastemur Could you check if I set things up correctly to build wabt on xplat ?
Also, I was a little confused about build configuration. I wasn't able to figure out how to make wabt part of the test build, but not the release build.

Cellule:
@obastemur I decided to merge this as it is to be able to move forward with other wasm changes.
If you have any suggestions regarding the xplat build you can let me know and I will address them in a later change.

PR #2655, opened 2017-03-09
xplat: add --libs-only build.sh option

obastemur:
This is to keep `build.sh` from compiling `ch/gcstress`. So, the embedder doesn't have to feed chakracore with third party libs for static library builds.


obastemur:

kunalspathak:
LGTM. nit: Can you add a line in `README.md` about this option?

PR #2659, opened 2017-03-09
Fix type mismatch for mono type check

ThomsonTan:
For `PropertySymOpnd` which needs mono type check, we should check the `monoGuardType` is the same `Type` as the operand, or the slot index (cached type) in the operand is invalid for property access, like `LdSlot` and `StSlot`, even if mono type check passed.


ThomsonTan:
I tried bench mark perf which looks flat.

pleath:
Is this identical to a fix that @rajatd tried for essentially the same issue? My sense is that the dead store pass is too late to be marking opnd type mismatch, because bailout points are already in place. It's not good enough just to do this here, because downstream field accesses rely on the type check to either succeed or cause bailout.

rajatd:
Yes, identical to #2203. :)

ThomsonTan:
Close it since it is a duplicate.

PR #2660, opened 2017-03-10
remove global operator new/delete override

jianchun:
Global operator new/delete override replaces the default used by hosts
unintendedly (not sure if there is reliable way to avoid that). This change
removes the global overrides and replaces all ChakraCore library new/delete
calls with HeapNew/HeapDelete.

HeapDelete: Added an optional flag "AllocatorDeleteFlags::UnknownSize"
for cases of unknown size to avoid size check. Default still checks size.

JITType.h: Add "const" to avoid an unintended copy constructor call, which
causes software write barrier access failure in JIT thread (not expecting
write barrier access on stack value).

### DIFF
--- a/lib/Common/Memory/ArenaAllocator.h
+++ b/lib/Common/Memory/ArenaAllocator.h
@@ -802,16 +802,21 @@ class RefCounted
     volatile LONG refCount;
 
 protected:
-    virtual ~RefCounted()
-    {
-    }
-
-public:
     RefCounted()
         : refCount(1)
     {
     }
 
+    virtual ~RefCounted()
+    {
+    }
+
+    void operator delete(void* p, size_t size)
+    {
+        HeapFree(p, size);
+    }
+
+public:
     uint32 AddRef(void)
     {
         return (uint32)InterlockedIncrement(&refCount);
@@ -823,7 +828,7 @@ public:
 
         if (0 == refs)
         {
-            delete this;
+            delete this;  // invokes overrided operator delete
         }
 
         return refs;
@@ -882,7 +887,6 @@ class ReferencedArenaAdapter : public RefCounted
     bool deleteFlag;
 
 public:
-
     ~ReferencedArenaAdapter()
     {
         if (this->arena)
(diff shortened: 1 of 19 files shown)


jianchun:
cc: @curtisman @digitalinfinity @obastemur @leirocks

lib/Common/Memory/ArenaAllocator.h
```diff
@@ -801,17 +801,16 @@ class RefCounted
...
-    {
-    }
-
 public:
     RefCounted()
         : refCount(1)
     {
     }
 
+    virtual ~RefCounted()
```
curtisman: I think the intent here is that RefCounted should never be instantiated by itself.
May be it should put the constructor as protected instead.

lib/Parser/Hash.h
```diff
@@ -394,7 +394,6 @@ class HashTbl
         m_perr = perr;
         memset(&m_rpid, 0, sizeof(m_rpid));
     }
-    ~HashTbl(void) {}
```
curtisman: I think this is to enforce that new/delete are not called on HashTbl, but uses Create/Release instead.
We probably can keep the destructor private that if we override new/delete of the HashTbl class and put that private as well.

lib/Parser/Scan.h
```diff
@@ -712,7 +712,6 @@ class Scanner : public IScanner, public EncodingPolicy
     size_t m_iecpLimTokPrevious;
 
     Scanner(Parser* parser, HashTbl *phtbl, Token *ptoken, ErrHandler *perr, Js::ScriptContext *scriptContext);
-    ~Scanner(void);
```
curtisman: Same comment as HashTbl.

jianchun:
@curtisman Added fix, could you please review again? Thanks!

lib/Common/Memory/HeapAllocatorOperators.cpp
```diff
@@ -8,6 +8,15 @@
...
 
+// Update:
+//
+// Overriding global operator new/delete causes problems in host.
+// Do not use global operator new/delete.
+//
+// To support memory tracking, use explicit HeapNew/HeapDelete.
+//
+
+#if 0
```
curtisman: Why keep these?  Delete?

PR #2661, opened 2017-03-10
Add option -debugbreakonphasebegin to break on a backend phase for given function

ThomsonTan:
This simple change breaks into debugger for a phase and function passed from command line. This is very helpful when I was looking at some bad transformation in a given phase for some known function. Usually we could look at the dump output to see which phase produces the incorrect IR, then trace the transformation and understand how it is generated. This breakpoint would make it easier to do so, especially for looking at such issue from website.


rajatd:
:shipit:

PR #2663, opened 2017-03-10
JSRT staticlib shutdown partial fix

jianchun:
A lot of shutdown/finalizer code expects current context. The current
staticlib shutdown atexit runs into problems. Since atexit is at process
shutdown, releasing memory isn't strictly required. This change reduces
current atexit to only shutdown threads. We'll investigate more in future
how to shutdown cleanly.

### DIFF
--- a/lib/Jsrt/JsrtHelper.cpp
+++ b/lib/Jsrt/JsrtHelper.cpp
@@ -109,12 +109,7 @@ void JsrtCallbackState::ObjectBeforeCallectCallbackWrapper(JsObjectBeforeCollect
     // we do not track the main thread. When it exits do the cleanup below
 #ifdef CHAKRA_STATIC_LIBRARY
     atexit([]() {
-        ThreadContext *threadContext = ThreadContext::GetContextForCurrentThread();
-        if (threadContext)
-        {
-            if (threadContext->IsInScript()) return;
-            ThreadBoundThreadContextManager::DestroyContextAndEntryForCurrentThread();
-        }
+        ThreadBoundThreadContextManager::DestroyContextAndEntryForCurrentThread();
 
         JsrtRuntime::Uninitialize();
 
--- a/lib/Jsrt/JsrtRuntime.cpp
+++ b/lib/Jsrt/JsrtRuntime.cpp
@@ -53,16 +53,20 @@ void JsrtRuntime::Uninitialize()
     while (currentThreadContext)
     {
         Assert(!currentThreadContext->IsScriptActive());
-#ifdef CHAKRA_STATIC_LIBRARY
-        if (currentThreadContext->IsInScript()) break;
-#endif
         JsrtRuntime* currentRuntime = static_cast<JsrtRuntime*>(currentThreadContext->GetJSRTRuntime());
         tmpThreadContext = currentThreadContext;
         currentThreadContext = currentThreadContext->Next();
 
+#ifdef CHAKRA_STATIC_LIBRARY
+        // xplat-todo: Cleanup staticlib shutdown. This only shuts down threads.
+        // Other closing contexts / finalizers having trouble with current
+        // runtime/context.
+        RentalThreadContextManager::DestroyThreadContext(tmpThreadContext);
+#else
         currentRuntime->CloseContexts();
         RentalThreadContextManager::DestroyThreadContext(tmpThreadContext);
         HeapDelete(currentRuntime);
+#endif
     }
 }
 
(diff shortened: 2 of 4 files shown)


lib/Jsrt/JsrtHelper.cpp
```diff
@@ -89,40 +91,32 @@ void JsrtCallbackState::ObjectBeforeCallectCallbackWrapper(JsObjectBeforeCollect
     }
 
 #ifndef _WIN32
-    static pthread_key_t s_threadLocalDummy;
+    static THREAD_LOCAL struct _OnThreadExit
```
obastemur: This fails. Fix->separate the definition/declaration.

lib/Jsrt/JsrtRuntime.cpp
```diff
@@ -53,16 +53,20 @@ void JsrtRuntime::Uninitialize()
...
         Assert(!currentThreadContext->IsScriptActive());
-#ifdef CHAKRA_STATIC_LIBRARY
-        if (currentThreadContext->IsInScript()) break;
-#endif
         JsrtRuntime* currentRuntime = static_cast<JsrtRuntime*>(currentThreadContext->GetJSRTRuntime());
         tmpThreadContext = currentThreadContext;
         currentThreadContext = currentThreadContext->Next();
 
+#ifdef CHAKRA_STATIC_LIBRARY
+        // TODO: Cleanup staticlib shutdown. This only shuts down threads.
```
obastemur: nit: can we update `TODO` to `xplat-todo` ?

lib/Runtime/Base/ThreadBoundThreadContextManager.cpp
```diff
@@ -223,10 +223,19 @@ JsUtil::JobProcessor * ThreadBoundThreadContextManager::GetSharedJobProcessor()
 
 void RentalThreadContextManager::DestroyThreadContext(ThreadContext* threadContext)
 {
-    ShutdownThreadContext(threadContext);
+    bool deleteThreadContext = true;
+
+#ifdef CHAKRA_STATIC_LIBRARY
+    // TODO: Cleanup staticlib shutdown. Deleting contexts / finalizers having
```
obastemur: nit: todo -> xplat-todo

lib/Jsrt/JsrtHelper.cpp
```diff
@@ -89,40 +91,32 @@ void JsrtCallbackState::ObjectBeforeCallectCallbackWrapper(JsObjectBeforeCollect
...
 #ifdef CHAKRA_STATIC_LIBRARY
-    atexit([]() {
-        ThreadContext *threadContext = ThreadContext::GetContextForCurrentThread();
-        if (threadContext)
-        {
-            if (threadContext->IsInScript()) return;
-            ThreadBoundThreadContextManager::DestroyContextAndEntryForCurrentThread();
-        }
+        atexit([]() {
+            ThreadBoundThreadContextManager::DestroyAllContextsAndEntries();
```
obastemur: shall we check whether the current thread has an active context and that context isInScript?

obastemur:
couple of nits and questions. 

LGTM

lib/Jsrt/JsrtHelper.cpp
```diff
@@ -89,40 +91,32 @@ void JsrtCallbackState::ObjectBeforeCallectCallbackWrapper(JsObjectBeforeCollect
...
+    {
+        ~_OnThreadExit()
+        {
+            ThreadBoundThreadContextManager::DestroyContextAndEntryForCurrentThread();
+        }
+    } _s_onThreadExit;
 #endif
-    static THREAD_LOCAL bool s_threadWasEntered = false;
 
-    _NOINLINE void DISPOSE_CHAKRA_CORE_THREAD(void *_)
```
obastemur: BTW; prev. pthread approach was guaranteed to run at the `last phase` of thread exit. Current approach doesn't have a priority in place. That makes me wonder that; If altering the timing for `ThreadBoundThreadContextManager::DestroyContextAndEntryForCurrentThread()` call has improved the test results positively, we might have a `race` issue on cleanup that we might have to target.

lib/Jsrt/JsrtHelper.cpp
jianchun: It looks the added THREAD_LOCAL destructor is called before this which handled it already.

lib/Jsrt/JsrtHelper.cpp
jianchun: I don't know as it is hard to repro/investigate. (Some say pthread_key_create destructors might be skipped if pthread_exit is not called, e.g. by calling exit -- not sure if that plays a role).

jianchun:
@obastemur I reverted THREAD_LOCAL changes back to pthread_create_key, removed/tweaked a IsInScript() check and related code (we can do that because now I only shutdown threads), and now it looks works as well.

obastemur:
Great! :shipit:

jianchun:
Thanks @obastemur

PR #2664, opened 2017-03-10
Added support to VSO and Jenkins test scripts to know that we run by default on Server 2012 R2.

dilijev:


dilijev:
@dotnet-bot test ci please

PR #2666, opened 2017-03-10
Defer functions enclosed in all contexts. 

pleath:
Support deferral of functions enclosed in scopes other than function body and function expression scope -- in particular, ES6-style lexical scopes and parameter scopes. This requires changing ScopeInfo so that it is associated with a single scope rather than a function body. Instead of carrying per-function information for body/var scope, parameter scope, and function expression scope, ScopeInfo will describe one scope, with a value identifying the scope type, a pointer to the FunctionInfo that contains it, and a pointer to the enclosing ScopeInfo. A FunctionProxy will point to the ScopeInfo that is the nearest enclosing scope. At parse time, we will reconstitute the closure environment by walking the list of enclosing ScopeInfo's. The code that allowed redeferral to work around the context limitation of deferring parsing is deleted, and the OptimizeBlockScope feature is off by default, as it doesn't play well with this new logic and isn't required to make (re-)deferral effective. (We will want to revisit it, though, so I've left it there.)

### DIFF
--- a/lib/Runtime/ByteCode/ByteCodeGenerator.cpp
+++ b/lib/Runtime/ByteCode/ByteCodeGenerator.cpp
@@ -968,84 +968,101 @@ Js::RegSlot ByteCodeGenerator::EnregisterStringTemplateCallsiteConstant(ParseNod
 //
 // Restore all outer func scope info when reparsing a deferred func.
 //
-void ByteCodeGenerator::RestoreScopeInfo(Js::ParseableFunctionInfo* functionBody)
+void ByteCodeGenerator::RestoreScopeInfo(Js::ScopeInfo *scopeInfo, FuncInfo * func)
 {
-    if (functionBody && functionBody->GetScopeInfo())
+    if (scopeInfo)
     {
         PROBE_STACK(scriptContext, Js::Constants::MinStackByteCodeVisitor);
 
-        Js::ScopeInfo* scopeInfo = functionBody->GetScopeInfo();
-        RestoreScopeInfo(scopeInfo->GetParent()); // Recursively restore outer func scope info
+        Js::ParseableFunctionInfo * pfi = scopeInfo->GetFunctionInfo()->GetParseableFunctionInfo();
+        bool newFunc = (func == nullptr || func->byteCodeFunction != pfi);
 
-        Js::ScopeInfo* paramScopeInfo = scopeInfo->GetParamScopeInfo();
-        Scope* paramScope = nullptr;
-        if (paramScopeInfo != nullptr)
+        if (newFunc)
         {
-            paramScope = paramScopeInfo->GetScope();
-            Assert(paramScope);
-            if (!paramScopeInfo->GetCanMergeWithBodyScope())
+            func = Anew(alloc, FuncInfo, pfi->GetDisplayName(), alloc, nullptr, nullptr, nullptr, pfi);
+            newFunc = true;
+        }
+
+        // Recursively restore enclosing scope info so outermost scopes/funcs are pushed first.
+        this->RestoreScopeInfo(scopeInfo->GetParentScopeInfo(), func);
+        this->RestoreOneScope(scopeInfo, func);
+
+        if (newFunc)
+        {
+            PushFuncInfo(_u("RestoreScopeInfo"), func);
+            if (!pfi->DoStackNestedFunc())
             {
-                paramScope->SetCannotMergeWithBodyScope();
+                func->hasEscapedUseNestedFunc = true;
             }
-            // We need the funcInfo before continuing the restoration of the param scope, so wait for the funcInfo to be created.
         }
-
-        Scope* bodyScope = scopeInfo->GetScope();
-
-        Assert(bodyScope);
-        bodyScope->SetHasOwnLocalInClosure(scopeInfo->GetHasOwnLocalInClosure());
-
-        FuncInfo* func = Anew(alloc, FuncInfo, functionBody->GetDisplayName(), alloc, paramScope, bodyScope, nullptr, functionBody);
-
-        if (bodyScope->GetScopeType() == ScopeType_GlobalEvalBlock)
-        {
-            func->bodyScope = this->currentScope;
-        }
-        PushFuncInfo(_u("RestoreScopeInfo"), func);
-
-        if (!functionBody->DoStackNestedFunc())
-        {
-            func->hasEscapedUseNestedFunc = true;
-        }
-
-        Js::ScopeInfo* funcExprScopeInfo = scopeInfo->GetFuncExprScopeInfo();
-        if (funcExprScopeInfo)
-        {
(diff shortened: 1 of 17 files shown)


pleath:
@aneeshdk This is the feature work we discussed last week. Can you take a look when you get a chance? (Thanks.)

lib/Runtime/ByteCode/ByteCodeGenerator.cpp
```diff
@@ -968,84 +968,101 @@ Js::RegSlot ByteCodeGenerator::EnregisterStringTemplateCallsiteConstant(ParseNod
...
-        Assert(bodyScope);
-        bodyScope->SetHasOwnLocalInClosure(scopeInfo->GetHasOwnLocalInClosure());
-
-        FuncInfo* func = Anew(alloc, FuncInfo, functionBody->GetDisplayName(), alloc, paramScope, bodyScope, nullptr, functionBody);
-
-        if (bodyScope->GetScopeType() == ScopeType_GlobalEvalBlock)
-        {
-            func->bodyScope = this->currentScope;
+            func = Anew(alloc, FuncInfo, pfi->GetDisplayName(), alloc, nullptr, nullptr, nullptr, pfi);
+            newFunc = true;
```
aneeshdk: Don't need it here?

lib/Runtime/ByteCode/ByteCodeGenerator.cpp
```diff
@@ -2023,7 +2041,7 @@ void ByteCodeGenerator::CheckDeferParseHasMaybeEscapedNestedFunc()
     else
     {
         // We have to wait until it is parsed before we populate the stack nested func parent.
-        FuncInfo * parentFunc = top->GetBodyScope()->GetEnclosingFunc();
+        FuncInfo * parentFunc = top->GetParamScope() ? top->GetParamScope()->GetEnclosingFunc() : top->GetBodyScope()->GetEnclosingFunc();
```
aneeshdk: >top->GetParamScope()->GetEnclosingFunc() : top->GetBodyScope()->GetEnclosingFunc() [](start = 55, length = 82)

Aren't these two the same?

lib/Runtime/ByteCode/ByteCodeGenerator.cpp
```diff
@@ -3629,15 +3653,21 @@ void PostVisitBlock(ParseNode *pnode, ByteCodeGenerator *byteCodeGenerator)
...
-        Scope *scope = pnode->sxBlock.scope;
         bool scopeIsEmpty = scope->IsEmpty();
         scope->SetIsObject();
         scope->SetCapturesAll(true);
         scope->SetMustInstantiate(!scopeIsEmpty);
     }
 
+    if (scope->GetHasOwnLocalInClosure())
+    {
+        byteCodeGenerator->ProcessScopeWithCapturedSym(scope);
```
aneeshdk: >byteCodeGenerator->ProcessScopeWithCapturedSym(scope); [](start = 8, length = 54)

Why do we have to do this here? Doesn't this happen when we find the captured symbol itself? Or is it to cover the case where the scope is recreated from the ScopeInfo so we have to mark it again? If that is the case should we do this while Restoring the scope?

lib/Runtime/ByteCode/ScopeInfo.cpp
```diff
@@ -74,26 +68,29 @@ namespace Js
...
         scopeInfo->isCached = (scope->GetFunc()->GetBodyScope() == scope) && scope->GetFunc()->GetHasCachedScope();
-        scopeInfo->isGlobalEval = scope->GetScopeType() == ScopeType_GlobalEvalBlock;
         scopeInfo->canMergeWithBodyScope = scope->GetCanMergeWithBodyScope();
         scopeInfo->hasLocalInClosure = scope->GetHasOwnLocalInClosure();
+        
 
-        TRACE_BYTECODE(_u("\nSave ScopeInfo: %s parent: %s #symbols: %d %s\n"),
-            scope->GetFunc()->name, parent ? parent->GetDisplayName() : _u(""), count,
+        TRACE_BYTECODE(_u("\nSave ScopeInfo: %s #symbols: %d %s\n"),
+            scope->GetFunc()->name, /*parent ? parent->GetDisplayName() : _u(""),*/ count,
```
aneeshdk: >/*parent ? parent->GetDisplayName() : _u(""),*/ [](start = 36, length = 47)

Don't need this

lib/Runtime/ByteCode/ScopeInfo.cpp
```diff
@@ -74,26 +68,29 @@ namespace Js
...
             scopeInfo->isObject ? _u("isObject") : _u(""));
 
-        MapSymbolData mapSymbolData = { byteCodeGenerator, scope->GetFunc(), 0 };
+        MapSymbolData mapSymbolData = { /*byteCodeGenerator,*/ scope->GetFunc(), 0 };
```
aneeshdk: >/*byteCodeGenerator,*/ [](start = 40, length = 22)

Don't need this

lib/Runtime/ByteCode/ScopeInfo.cpp
```diff
@@ -107,75 +104,46 @@ namespace Js
...
+        if (scope)
+        {
+            scopeInfo->SetParentScopeInfo(ScopeInfo::SaveScopeInfo(scope, scriptContext));
+        }
+
+        return scopeInfo;
     }
 
-    void ScopeInfo::SaveScopeInfoForDeferParse(ByteCodeGenerator* byteCodeGenerator, FuncInfo* parentFunc, FuncInfo* funcInfo)
+    void ScopeInfo::SaveEnclosingScopeInfo(ByteCodeGenerator* byteCodeGenerator, /*FuncInfo* parentFunc,*/ FuncInfo* funcInfo)
```
aneeshdk: >/*FuncInfo* parentFunc,*/ [](start = 81, length = 25)

Remove this

aneeshdk:
:shipit:

lib/Runtime/ByteCode/ByteCodeGenerator.cpp
pleath: We've been doing this when we find the capturing reference to the symbol. Now that reference may be deferred, so byte code gen won't see it. It would be great to go through and remove the calls at the points where they're no longer needed, but I'm not doing that cleanup as part of this change.

lib/Runtime/ByteCode/ByteCodeGenerator.cpp
pleath: Thanks for spotting these.

pleath: Sorry -- can you explain what you mean?

lib/Runtime/ByteCode/ScopeInfo.cpp
pleath: Thanks for spotting these.

Issue #2672, opened 2017-03-10
Atomics.store throws "TypeError: The operation is not supported on this typed array type"


brion:
Testing cross-compiled code using emscripten's pthreads/SharedArrayBuffer mode in Edge on Windows Insider build 15048, I found that code that runs in Firefox and Safari tech previews fails here due to a TypeError being thrown in Atomics.store():

Simplified code:

```
var s = new SharedArrayBuffer(65536);
var u = new Uint32Array(s);
Atomics.store(u, 0, 1234);
```

Live example: https://jsfiddle.net/yejks8ja/

This works in Firefox Developer Edition and Safari Technical Preview, but fails in Edge (requires Experimental JS Features flag on) with "TypeError: The operation is not supported on this typed array type"

akroshg:
This should work. I'll take a look.

akroshg:
Hmm, typedarray checks were failing even though they are typedarray, that is because the virtual table is set differently for count 2**16.

akroshg:
@brion thanks for reporting that.

PR #2674, opened 2017-03-10
Atomics check fails for virtual typedarray.

akroshg:
Atomics operation failed on the virtual typedarray (when SharedArrayBuffer is more than 2^16). However In our case we don't need to check for virtual table check as the final operation happes on the typedarray itself. Fixed that by checking just the typeid.

### DIFF
--- a/lib/Runtime/Library/AtomicsObject.cpp
+++ b/lib/Runtime/Library/AtomicsObject.cpp
@@ -24,17 +24,18 @@ namespace Js
             JavascriptError::ThrowTypeError(scriptContext, JSERR_NeedTypedArrayObject);
         }
 
+        TypeId typeId = JavascriptOperators::GetTypeId(typedArray);
         if (onlyInt32)
         {
-            if (!Int32Array::Is(typedArray))
+            if (typeId != TypeIds_Int32Array)
             {
                 JavascriptError::ThrowTypeError(scriptContext, JSERR_InvalidOperationOnTypedArray);
             }
         }
         else
         {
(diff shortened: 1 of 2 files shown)


akroshg:
@boingoing please have a look at that

akroshg:
Fixes #2672

lib/Runtime/Library/AtomicsObject.cpp
```diff
@@ -24,17 +24,18 @@ namespace Js
...
+            if (typeId != TypeIds_Int32Array)
             {
                 JavascriptError::ThrowTypeError(scriptContext, JSERR_InvalidOperationOnTypedArray);
             }
         }
         else
         {
-            if (!(Int8Array::Is(typedArray) || Uint8Array::Is(typedArray) || Int16Array::Is(typedArray) ||
-                Uint16Array::Is(typedArray) || Int32Array::Is(typedArray) || Uint32Array::Is(typedArray)))
+            if (!(typeId == TypeIds_Int8Array || typeId == TypeIds_Uint8Array || typeId == TypeIds_Int16Array ||
```
boingoing: Should this code handle Uint8ClampedArray or is that one not supported by the AtomicsObject?

akroshg: This is per specs : http://tc39.github.io/ecmascript_sharedmem/shmem.html#Atomics.ValidateSharedIntegerTypedArray
It should not support unit8clampedarray.

PR #2677, opened 2017-03-13
WASM: memory validation

Cellule:
Add more validation to the memory descriptor and opcodes to be compliant with the spec.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/2677)


MikeHolman:
:shipit:

PR #2684, opened 2017-03-15
Avoid calling CompileScriptException::ProcessError in the catch handler, where the stack is not unwounded.

curtisman:
If we are hitting a stack overflow, we should unwind the stack before running more code (like ProcessError)
Also clean up ErrHandler, which really does nothing but throw ParseException.  Change it to directly throw ParseException.

Clean up some debug only ifdef and disable background parsing code in release build, which is not enabled by default.

Fixes VSO: 11101148


curtisman:
@pleath  code review pls.

PR #2687, opened 2017-03-15
WASM: Validate unreachable code

Cellule:


lib/Runtime/ByteCode/IWasmByteCodeWriter.h
```diff
@@ -0,0 +1,45 @@
...
+namespace Js
+{
+    struct IWasmByteCodeWriter
+    {
+    public:
+        virtual void Create() PURE;
+        virtual void Begin(FunctionBody* functionWrite, ArenaAllocator* alloc) PURE;
+        virtual void End() PURE;
+        virtual void Reset() PURE;
+        virtual ByteCodeLabel DefineLabel() PURE;
```
MikeHolman: nit: we do = 0 everywhere else, so would prefer that. i had to look up the macro definition

lib/WasmReader/WasmByteCodeGenerator.cpp
```diff
@@ -676,42 +650,49 @@ WasmBytecodeGenerator::EmitSetLocal(bool tee)
...
         break;
     default:
         throw WasmCompilationException(_u("Unknown type %u"), dst.type);
     }
 }
 
+WasmConstLitNode
+WasmBytecodeGenerator::GetZeroCnst()
+{
+    WasmConstLitNode cnst;
```
MikeHolman: pretty sure you don't need memset and can just do `= {0};`, though maybe the compiler can figure that out too and avoid the call

lib/WasmReader/WasmByteCodeGenerator.cpp
```diff
@@ -869,6 +844,7 @@ WasmBytecodeGenerator::EmitCall()
         case WasmTypes::F64:
             argOp = isImportCall ? Js::OpCodeAsmJs::ArgOut_Db : Js::OpCodeAsmJs::I_ArgOut_Db;
             break;
+        case WasmTypes::Any:
```
MikeHolman: i feel like this should have a comment

lib/WasmReader/WasmByteCodeGenerator.cpp
```diff
@@ -1182,62 +1123,59 @@ WasmBytecodeGenerator::EmitReturnExpr(EmitInfo* explicitRetInfo)
...
 
-    m_writer.AsmReg2(op, trueInfo.location, falseInfo.location);
+    m_writer->AsmReg2(op, trueInfo.location, falseInfo.location);
     ReleaseLocation(&falseInfo);
 
-    m_writer.MarkAsmJsLabel(doneLabel);
-
+    m_writer->MarkAsmJsLabel(doneLabel);
+    
+    if (trueInfo.type == WasmTypes::Any && falseInfo.type != WasmTypes::Any)
```
MikeHolman: i don't understand this. e.g. `a = cond ? return : 5`, don't we need to give 5 as the result? or is that not the condition you are checking here?

lib/WasmReader/WasmByteCodeGenerator.cpp
Cellule: I thought some compilers didn't like `= {0}`, but I'll give it a try.

lib/WasmReader/WasmByteCodeGenerator.cpp
Cellule: Actually, in a select we have to evaluate both the true and the false result since it's a stack machine.
By the time we reach the select op we've already became unreachable therefore, the select is not executed.
The code really is
```js
var true = return;
var false = 5;
var selected = cond ? true : false;
```

This is remnant of when I was trying to emit valid bytecode in unreachable state, but with EmptyWasmBYteCodeWriter this became useless.

It make me think however that we probably should convert the select op to
```js
var selected;
if (cond) {
  selected = "true expr"; 
} else {
  selected = "false expr";
}
```
if neither expr has side effects

MikeHolman:
Ok I'll probably need to talk to you in person about this. For case of true unreachable, I feel we still need to evaluate false, but maybe I'm missing something.

_____________________________

lib/WasmReader/WasmByteCodeGenerator.cpp
Cellule: I tried removing that code and then we failed a test.
I remember why I had to add this.
in the case
```js
var res = cond ? unreachable : (i64); // res must be type i64, not Any
var res2 = (i32) + res; // invalid since res is i64 type.
```

I decided to rewrite EmitSelect and I hope the new version is clearer

PR #2691, opened 2017-03-16
Remove unused and probably broken syntax coloring support in scanner

curtisman:
Also remove instantiation of the Scanner<NullTerminatedUtf8EncodingPolicy>

### DIFF
--- a/lib/Parser/Scan.cpp
+++ b/lib/Parser/Scan.cpp
@@ -35,60 +35,6 @@ int CountNewlines(LPCOLESTR psz, int cch)
     return cln;
 }
 
-template< typename CharT >
-struct AorW
-{
-};
-
-// Specialization for UTF8Char
-template<>
-struct AorW< UTF8Char >
-{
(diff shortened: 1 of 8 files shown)


curtisman:
@EdMaurer @pleath pls review

lib/Parser/Scan.cpp
```diff
@@ -1552,7 +1460,7 @@ tokens Scanner<EncodingPolicy>::ScanStringConstant(OLECHAR delim, EncodedCharPtr
 LBreak:
     bool createPid = true;
 
-    if (m_fSyntaxColor || (m_DeferredParseFlags & ScanFlagSuppressStrPid) != 0)
+    if ((m_DeferredParseFlags & ScanFlagSuppressStrPid) != 0)
```
pleath: I think ScanFlagSuppressStrPid can be deleted. I thought background parsing was using it, but I'm thinking of the ScanNoKeywords path.l

curtisman: Cool.  I will remove it in the next check in.

PR #2692, opened 2017-03-16
add JIT fastpath for hasOwnProperty calls using for-in property name

MikeHolman:
Added hasOwnProperty as CallDirect candidate, and also made a pretty simple lowering fastpath, which detects if we are loading property name from a for-in loop. if we are, then we can take advantage of state of our enumerator to know if property is an own property or from prototype.

This improves perf of React by about 10%.


rajatd:
:shipit:

lib/Backend/Lower.cpp
```diff
@@ -17931,6 +17934,70 @@ Lowerer::GenerateFastInlineArrayPop(IR::Instr * instr)
     GenerateHelperToArrayPopFastPath(instr, doneLabel, bailOutLabelHelper);
 }
 
+void
+Lowerer::GenerateFastInlineHasOwnProperty(IR::Instr * instr)
+{
+    Assert(instr->m_opcode == Js::OpCode::CallDirect);
```
Penguinwizzard: Could you add a comment at the top of this function outlining the inline code? (Similar to how e.g. GenerateFastCondBranch has it)

curtisman:
canUseJITFastpath is only enabled if the object doesn't have any enumerable property from the start, and no index property.  So the win is mostly from optimizating the call the built in.

I wonder whether we will get the same win if do the partial inlining of the PropertyString cache check instead checking the for in enumerator. (or do both?)

lib/Backend/Lower.cpp
```diff
@@ -17931,6 +17934,70 @@ Lowerer::GenerateFastInlineArrayPop(IR::Instr * instr)
...
+    }
+
+    IR::LabelInstr *doneLabel = InsertLabel(false, instr->m_next);
+    IR::LabelInstr *labelHelper = InsertLabel(true, instr);
+    IR::Instr * insertInstr = labelHelper;
+    IR::RegOpnd * thisObj = argsOpnd[0]->AsRegOpnd();
+    IR::Opnd * forInEnumeratorOpnd = argsOpnd[1]->AsRegOpnd()->m_sym->m_instrDef->GetSrc1();
+
+    // go to helper if we can't use JIT fastpath
+    IR::Opnd * canUseJitFastPathOpnd = GetForInEnumeratorFieldOpnd(forInEnumeratorOpnd, Js::ForInObjectEnumerator::GetOffsetOfCanUseJitFastPath(), TyInt8);
```
curtisman: Please add a comment about what checking this flag will rule out (e.g. the for in object doesn't have any enumerable prototype property, it doesn't have any index property, the type has to be "lockable" (e.g. currently no DictionaryTypeHandler).  Also add a comment in the ForInObjectEnumerator for this flag so that if it ever expand the cases that we can use jit fast path, we will have to look at this fast path as well.

curtisman:
Please add test cases for all the variants of for..in related to this fastpath.  Probably want to add ExprGen testing too.

lib/Backend/Lower.cpp
MikeHolman: Sure. I ended up going more conservative than originally coded to be safe and win seemed about same so I stuck with this variant. Wonder though if maybe it would be better to go more aggressive and check only that we have initial type, and that type is lockable.

I still see a bit of time left (3% under HasOwnProperty), but I think remaining cases are mostly when using different object than you are enumerating against.

MikeHolman:
String cache lookup was showing as very costly. The win here is mostly by changing that dictionary lookup into checking against a boolean.

lib/Backend/Lower.cpp
curtisman: I wonder whether you will catch those case as well if you also emit property string cache check?

MikeHolman: I can look into that for my next checkin. Probably a few places where that fastpath will be useful.

MikeHolman: Looks like in this case it won't buy much: the remaining hot call site isn't calling with a property string (it is building a string via some odd concatenation)

curtisman:
:shipit:

lib/Runtime/Library/ForInObjectEnumerator.h
```diff
@@ -21,7 +21,7 @@ namespace Js
         } *shadowData;
 
         // States
-        bool canUseJitFastPath;
+        bool canUseJitFastPath; // used by GenerateFastBrBReturn and GenerateFastInlineHasOwnProperty
```
curtisman: Would be good to document what this being true means here now that there are other use of this boolean.

curtisman:
@MikeHolman Please make sure ExprGen has coverage.

PR #2698, opened 2017-03-17
Fix bad code with call to closure-captured local function from inside 'with'. 

pleath:
The case where the with statement is in a function nested within the function that instantiates the called function is already handled. Now that functions nested within with can be deferred, we need to handle the case where the with is in the enclosing function itself.


pleath:
@aneeshdk Can you look at this one? (Thanks.)

aneeshdk:
LGTM

PR #2699, opened 2017-03-17
fix warning in build after merging 17-03 servicing fixes

MikeHolman:
changes around isConcatSpreadable in release/2.0 required me to change the fix a bit for this branch. My local build had no issues, but cloud build gave error for implicit cast of BOOL to bool.


Cellule:
LGTM

PR #2700, opened 2017-03-17
Added support to post_build.ps1 and runcitests.cmd to have extra test params.

boingoing:
The build scripts pass a new parameter to post_build.ps1 (-testparams)
which post_build.ps1 on release/1.4 does not know about. This causes post_build.ps1 to
throw an error.

Fixed by porting the change from master. Cherry-picked 803abd5a30.


boingoing:
@dilijev can you take a look ?

dilijev:
Yes, this should get pumped to every branch we currently work with. Thanks for catching this. LGTM :shipit:

PR #2703, opened 2017-03-17
Reduce memory consumption of Utf8SourceInfo objects

kunalspathak:
Today we hold on to `scriptSource` objects inside `Utf8SourceInfo` even if we make a copy of the script source inside source holder. This is unnecessary and increase the working set. The fix is to not hold reference anymore.

However if the `scriptSource` is external array buffer and is managing its own life time, we don't need to make copy of it and instead we just hold the reference to `scriptSource` for scenarios like `function.name.toString()` to work (which is how it is done today).

### DIFF
--- a/lib/Runtime/Base/ScriptContext.cpp
+++ b/lib/Runtime/Base/ScriptContext.cpp
@@ -1814,7 +1814,7 @@ namespace Js
             // Free unused bytes
             Assert(cbNeeded + 1 <= cbUtf8Buffer);
             *ppSourceInfo = Utf8SourceInfo::New(this, utf8Script, (int)length,
-                cbNeeded, pSrcInfo, isLibraryCode, scriptSource);
+                cbNeeded, pSrcInfo, isLibraryCode);
         }
         else
         {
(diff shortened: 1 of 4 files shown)


kunalspathak:
@digitalinfinity , @obastemur , @curtisman  - Please help review.

obastemur:
:shipit:

lib/Runtime/Base/ScriptContext.cpp
```diff
@@ -1814,7 +1814,7 @@ namespace Js
             // Free unused bytes
             Assert(cbNeeded + 1 <= cbUtf8Buffer);
             *ppSourceInfo = Utf8SourceInfo::New(this, utf8Script, (int)length,
-                cbNeeded, pSrcInfo, isLibraryCode, scriptSource);
+                cbNeeded, pSrcInfo, isLibraryCode, nullptr);
```
digitalinfinity: consider commenting on the parameter name (nullptr /* scriptSourceVar */) - alternatively, isn't that parameter default? If so, maybe just remove it altogether?

kunalspathak: Thanks @digitalinfinity , i didn't realize that it has default parameter. No need to pass `nullptr` here. I will update the PR.

digitalinfinity: One additional comment I had was, it would be good to leave comments on why we don't need to pass in the parameter here so that people looking at the code later aren't confused

lib/Runtime/Base/ScriptContext.cpp
```diff
@@ -1814,7 +1814,7 @@ namespace Js
             // Free unused bytes
             Assert(cbNeeded + 1 <= cbUtf8Buffer);
             *ppSourceInfo = Utf8SourceInfo::New(this, utf8Script, (int)length,
-                cbNeeded, pSrcInfo, isLibraryCode, scriptSource);
```
curtisman: Can we remove the scriptSource argument from the definition too?  It serves no purpose if we always copies

PR #2705, opened 2017-03-17
Build Fix

aneeshdk:
Commenting the failing test for now. I am working on a fix for it. Will
uncomment it once the fix is in.


MikeHolman:
If fix will take more than just today, can you open an issue so it doesn't get lost?

aneeshdk:
Created the issue https://github.com/Microsoft/ChakraCore/issues/2707

Issue #2702, opened 2017-03-17
ES6 module: multiple imports of same module


suwc:
```javascript
// script.js
WScript.LoadScriptFile('mod0.js', "module");
```
```javascript
// mod0.js
import 'mod1a.js';
import 'mod1b.js';
console.log('mod0');
```
```javascript
// mod1a.js
//import 'mod1b.js'  // uncomment to repro
console.log("mod1a");
```
```javascript
// mod1b.js
console.log("mod1b");
```
`>ch -es6module script.js`
Expected output
```
mod1a
mod1b
mod0
```
Actual output (with `import 'mod1b.js'` uncommented)
``````

PR #2709, opened 2017-03-18
Issue2681/2702: Multiple imports of same module

suwc:
While resolving a module, a counter (`numUnInitializedChildModule`)
is maintained to reflect the number of its child modules that have
not been instantiated. Once a child module is parsed and all its children
resolved, its parent modules will be notified through `OnChildModuleReady()`,
where the counter is decremented. The counter has to be zero for a module
to be instantiated.

This causes a problem when there are multiple imports referring to
the same child module, and the child module is parsed and resolved
before some of its other parent modules. Because the child module is only parsed
once, parent modules that are parsed and linked to the child module after
the notification will never see their counter going back to zero. As a result,
these parent modules and all modules importing them will never be instantiated.

Proposed fix is to add a flag (`readyAsChildModule`) and flip it from false to true
upon notification of a module's parents. Future parent modules will see the
flag and deem that child module resolved. Also proposed a name change of the
counter from `numUnInitializedChildModule` to `numPendingChildModule` to
reflect the logic change.

### DIFF
--- a/test/es6/moduletest2_mod0.js
+++ b/test/es6/moduletest2_mod0.js
@@ -1,0 +1,9 @@
+//-------------------------------------------------------------------------------------------------------
+// Copyright (C) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE.txt file in the project root for full license information.
+//-------------------------------------------------------------------------------------------------------
+import './moduletest2_mod1a.js';
+import './moduletest2_mod1b.js';
+
+loaded++;
+if (loaded == 5) console.log("Pass");
--- a/lib/Runtime/ByteCode/ByteCodeGenerator.cpp
+++ b/lib/Runtime/ByteCode/ByteCodeGenerator.cpp
@@ -3139,7 +3139,7 @@ void ByteCodeGenerator::ProcessCapturedSym(Symbol *sym)
     FuncInfo *funcHome = sym->GetScope()->GetFunc();
     FuncInfo *funcChild = funcHome->GetCurrentChildFunction();
 
-    Assert(sym->NeedsSlotAlloc(funcHome) || sym->GetIsGlobal());
+    Assert(sym->NeedsSlotAlloc(funcHome) || sym->GetIsGlobal() || sym->GetIsModuleImport());
 
     // If this is not a local property, or not all its references can be tracked, or
     // it's not scoped to the function, or we're in debug mode, disable the delayed capture optimization.
--- a/lib/Runtime/Language/SourceTextModuleRecord.cpp
+++ b/lib/Runtime/Language/SourceTextModuleRecord.cpp
@@ -36,10 +36,11 @@ namespace Js
         resolvedExportMap(nullptr),
         wasParsed(false),
         wasDeclarationInitialized(false),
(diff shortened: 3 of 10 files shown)


suwc:
Addresses https://github.com/Microsoft/ChakraCore/issues/2702

suwc:
@boingoing @akroshg @Yongqu @Microsoft/chakra-es review pls?

boingoing:
Is this the case where a child module has been parsed already because it was already included as a child of a previous root module? Or because the same module record can be returned from the host via different module specifiers?

suwc:
There was only one root module, but import clauses in different source files share the same specifier. Take the updated unit test: 
moduletest1.js > passmodule.js > exportmodule.js > passmodule.js (2nd encounter, resolved, parsed but not instantiated)
The counter for exportmodule is incremented to 1 for passmodule but will never go to 0.

---
In reply to: [287507214](https://github.com/Microsoft/ChakraCore/pull/2709#issuecomment-287507214) [](ancestors = 287507214)

boingoing:
Sounds reasonable. This can only happen in the cyclical case, then?

akroshg:
@suwc what about the cases where another module throws error? do we still parse rest of the imports?

suwc:
No, this happens in non-cyclic cases too, e.g. in the github issue we are trying to address:
script.js > mod0 > mod1b
                             > mod1a > mod1b (2nd encounter)


---
In reply to: [287508635](https://github.com/Microsoft/ChakraCore/pull/2709#issuecomment-287508635) [](ancestors = 287508635)

suwc:
In this case module resolution happens before byteCodeGen, therefore only early error will be seen.


---
In reply to: [287509096](https://github.com/Microsoft/ChakraCore/pull/2709#issuecomment-287509096) [](ancestors = 287509096)

suwc:
@boingoing @akroshg @Yongqu could you take a look at the new iteration please?

test/es6/moduletest2_mod0.js
```diff
@@ -0,0 +1,9 @@
+//-------------------------------------------------------------------------------------------------------
+// Copyright (C) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE.txt file in the project root for full license information.
+//-------------------------------------------------------------------------------------------------------
+import './moduletest2_mod1a.js';
+import './moduletest2_mod1b.js';
+
+loaded++;
+if (loaded == 5) console.log("Pass");
```
boingoing: Should this log "Fail" if loaded != 5? Don't remember if empty output is treated as passing or not.

suwc: Empty output is a failed case. For example, when module instantiation doesn't happen, this test fails (confirmed).

---
In reply to: [107269987](https://github.com/Microsoft/ChakraCore/pull/2709#discussion_r107269987) [](ancestors = 107269987)

boingoing: Ok, thanks for checking. 👍

suwc:
@dotnet-bot 
test Ubuntu static_ubuntu_linux_test

PR #2710, opened 2017-03-18
Fix bad byte code when block-scoped function is called from with. 

pleath:
Yesterday's fix was incomplete, because it failed to address the case in which the nested function called from a with statement is declared in a lexical scope. The concept we're conservatively using, of a scope containing a with statement as well as closure-captured data, has to be tracked on individual scopes rather than functions.

### DIFF
--- a/lib/Runtime/ByteCode/Scope.h
+++ b/lib/Runtime/ByteCode/Scope.h
@@ -41,6 +41,7 @@ private:
     BYTE canMergeWithBodyScope : 1;
     BYTE hasLocalInClosure : 1;
     BYTE isBlockInLoop : 1;
+    BYTE containsWith : 1;
 public:
 #if DBG
     BYTE isRestored : 1;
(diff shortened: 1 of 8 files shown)


pleath:
@aneeshdk This is a variant on the one you looked at last week.

lib/Runtime/ByteCode/Scope.h
```diff
@@ -265,6 +267,9 @@ class Scope
     void SetIsBlockInLoop(bool is = true) { isBlockInLoop = is; }
     bool IsBlockInLoop() const { return isBlockInLoop; }
 
+    void SetContainsWith(bool does = true) { containsWith = does; }
```
aneeshdk: Can we assert that this is not a param scope?

aneeshdk:
LGTM

lib/Runtime/ByteCode/Scope.h
pleath: No, we can't. The attribute means that there's a with scope nested somewhere within the current scope.

PR #2712, opened 2017-03-18
Fix #2670. Set stdout in ch.exe to UTF-8 mode so console.log gets Unicode output support.

### LINKED ISSUE #2670: Text output functions (WScript.Echo / console.log / print) corrupt non-ASCII characters on output.
Note: all of the text output functions named above are implemented as the same native function, `WScriptJsrt::EchoCallback`, with multiple JS name bindings.

['NO-BREAK SPACE' (U+00A0)](http://www.fileformat.info/info/unicode/char/00a0/index.htm)
becomes
[á 'LATIN SMALL LETTER A WITH GRAVE' (U+00E0)](http://www.fileformat.info/info/unicode/char/00e0/index.htm)

[¡ 'INVERTED EXCLAMATION MARK' (U+00A1)](http://www.fileformat.info/info/unicode/char/00A1/index.htm)
becomes
[í 'LATIN SMALL LETTER I WITH GRAVE' (U+00E1)](http://www.fileformat.info/info/unicode/char/00e0/index.htm)

[– 'EN DASH' (U+2013)](http://www.fileformat.info/info/unicode/char/2013/index.htm)
becomes
[? 'QUESTION MARK' (U+003F)](http://www.fileformat.info/info/unicode/char/003f/index.htm)

Same transformation for [EM-DASH (U+2014)](http://www.fileformat.info/info/unicode/char/2014/index.htm)

Here's a test:

```js
console.log("\u00a0\u00a1\u2013\u2014");
// U+00A0 ( ) no-break space
// U+00A1 (¡) inverted exclamation mark
// U+2014 (—) em dash
// U+2013 (–) en dash
```

Output from ch: `áí??`
Output from node: ` ¡–—` (correct output, so apparently not a problem with conhost on Windows)

From looking at a hex dump of the output, the first two cases are apparently incorrectly rendered as a single byte (since the code point is 0 except the bottom 8 bits, we might be mistakenly interpreting that as a code point which is valid to output as a single byte):

```
> ch test.js | xxd
00000000: a0a1 3f3f 0d0a                           ..??..
```

Note a0a1 in the first two bytes (the ISO/IEC 8859-1 / Latin-1 / Windows-1252, one-byte encoding of those codepoints, but not a correct UTF-8 encoding), and 3f (`?`) in the second two bytes. Final 2 bytes are CRLF.

```
> node test.js | xxd
00000000: c2a0 c2a1 e280 93e2 8094 0a              ...........
```

Note c2a0 and c2a1, the correct UTF-8 encodings of U+00A0 and U+00A1, followed by e28093 and e28094, the correct UTF-8 encodings of U+2013 and U+2014 respectively. (LF line terminator at the end.)

In the latter two cases, I think this is internally converting to `?` as a replacement character for an unknown character.

Discovered when looking into https://github.com/Microsoft/ChakraCore/issues/2656

dilijev:
Only use _setmode on Windows; it seems stdout is UTF-8 mode by default on Linux.

This change would break printf because it does not understand UTF-8. To that end, this change also incorporates the following changes:

* Replace printf with wprintf and update formatter strings.
* Replace wprintf with Output::Print except in ch where it would be a new dependency.
* Convert L"" strings to _u("") strings in more places where they were apparently overlooked before.

Fixes #2670


dilijev:
I made the change to print out the DLL version under the --version flag with wprintf because printf apparently breaks under this change. We don't have that many instances of printf throughout the code but it would be a good idea to figure out a workaround or convert all of those printf instances to be wprintf instead.

bin/ch/ch.cpp
```diff
@@ -3,6 +3,7 @@
 // Licensed under the MIT license. See LICENSE.txt file in the project root for full license information.
 //-------------------------------------------------------------------------------------------------------
 #include "stdafx.h"
+#include <fcntl.h>
```
dilijev: Move this under _WIN32

dilijev:
@jianchun The test here is to ensure that it also works correctly on xplat. Apparently no change was needed to enable UTF-8 output.

Conhost seems to support UTF-8 and convert characters to UCS-2/UTF-16 codepoints, but does not combine surrogates. Conemu combines surrogates and I was able to confirm that the text printed on the console and the UTF-8 saved in the file are correct.

Differences in the bytes in the files are strictly in line endings \r\n versus \n between ch and node respectively. Linux version of ch also outputs \n but the test runner will normalize line endings so that's not a problem (applies to all existing tests anyway).

demo-conhost

demo-conemu

bin/ch/ch.cpp
```diff
@@ -831,7 +831,17 @@ int _cdecl wmain(int argc, __in_ecount(argc) LPWSTR argv[])
...
 
 #ifdef _WIN32
-    _setmode(_fileno(stdout), _O_U8TEXT); // enable Unicode output as UTF-8
+    // Set the output mode of stdout so we can display non-ASCII characters on the console and redirect to file as UTF-8
+    {
+        int result = _setmode(_fileno(stdout), _O_U8TEXT); // set stdout to UTF-8 mode
+        if (result == -1)
+        {
+            // Failed to set mode. Undefined behavior may result, so exit now.
+            wprintf(_u("Failed to set output stream mode. Exiting...\n"));
```
dilijev: There's a little bit of irony here that failing to set the mode of stdout might cause undefined behavior when printing to stdout, and then we print an error message to stdout. Here I'm assuming that the mode will remain as the default and ASCII output should work just fine, but if the message comes out corrupted, exiting early was the right decision in any case.

bin/ch/ch.cpp
```diff
@@ -129,8 +128,8 @@ void __stdcall PrintChakraCoreVersion()
                 // Doesn't matter if you are on 32 bit or 64 bit,
                 // DWORD is always 32 bits, so first two revision numbers
                 // come from dwFileVersionMS, last two come from dwFileVersionLS
-                printf("%s version %d.%d.%d.%d\n",
-                    chakraDllName,
+                wprintf(_u("%s version %d.%d.%d.%d\n"),
+                    GetChakraDllNameW(),
```
jianchun: >GetChakraDllNameW(), [](start = 20, length = 20)

Wonder if you can use format specifier "%S" with narrow chakraDllName. Save creating a "W" version.

jianchun:
LGTM 👍

bin/ch/ch.cpp
dilijev: I thought of that but I read somewhere that there's a bit of disagreement about what %s and %S should mean for wprintf. Apparently on Linux, %S is wide string even in wprintf and %s is ASCII string even in wprintf, but on Windows the meanings are flipped for wprintf. It might be the case that this is not true or the PAL version of wprintf takes care of this so that they behave the same on both platforms. I'll check.

dilijev: Looks like %S does what we'd expect in PAL (behaves the same on Linux and Windows). On Linux, it does seem like the implementation wprintf flips the meanings of those formatters.

In any case for this bit of output it doesn't matter because it's Windows-only.

jianchun: Ok, thanks for checking!

dilijev: For completeness, posting my standalone test:

```
$ cat test.cpp
#include <cstdio>
#include <wchar.h>
int main() {
    wprintf(L"%S", L"hello\n");
    return 0;
}

$ make test
g++     test.cpp   -o test

$ ./test
hello

$ sed -i test.cpp -e "s/%S/%s/"

$ make test
g++     test.cpp   -o test

$ ./test       # note: UTF-16 LE second byte is 0x00 NUL when wchar* interpreted as char*
h

$ vim test.cpp

$ cat test.cpp
#include <cstdio>
#include <wchar.h>
int main() {
    wprintf(L"%S", "hello\n"); // removed L prefix from arg 2, now ASCII string
    return 0;
}

$ make test
g++     test.cpp   -o test

$ ./test        # output is garbage because trying to interpret char* as wchar*
??%S
```

dilijev:
Added "Waiting" tag because I'm concerned about breakage of printf under this change.

dilijev:
Combining #2742 into this PR.

dilijev:

dilijev:
Merging. /cc @Penguinwizzard

PR #2715, opened 2017-03-20
WASM: wabt for spec tests

Cellule:
Use wabt for spec tests instead of precompiled binaries.
I had to reimplement wabt's binary-writer-spec because we didn't want to create .json files, but instead create javascript objects directly.
I hope/think this won't require too much maintenance.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/2715)


Cellule:
The error in legacy build is a bug in the implementation of `strtof` in that version.
I have to look into workaround about this one.

Cellule:
I managed to fix the main issue regarding float parsing in wabt for legacy build, but there is still 1 more issue with precision when using `strtod`. I won't try to fix that issue since it's a bug in MSVC++ 12.0 which has been reported before and decided not to service.
Since we are even considering dropping support for VS2013 soon, I won't pursue the issue further and decided to exclude the remaining failing test.

PR #2722, opened 2017-03-21
Date missing cross-site marshalling

jianchun:
A lot of Date methods (the family of to???String) miss cross-site
marshalling. This change fixes that by removing
`DateImplementation::m_scriptContext` and pass in `requestContext`
everywhere.


jianchun:
Thanks @kunalspathak

PR #2723, opened 2017-03-21
Set release/1.4 version to 1.4.2

boingoing:


Build/NuGet/.pack-version
```diff
@@ -1 +1 @@
-1.4.1
+1.4.2
```
dilijev: I can't remember but we might have said we just don't care about the value of this file since we explicitly set the NuGet version in the automation that builds the package and don't use this file. This file is for 3rd party. That said, it's good to keep them in sync.

boingoing:

PR #2724, opened 2017-03-21
Fix inconsistent parser state when deferral is undone. 

pleath:
On stoppage of deferred parsing (i.e., when we discover mid-script that we want to stop deferring, and at the end we go back and undefer), complete binding of vars in global scope and restore scope and block info state before we start undeferral.


aneeshdk:
LGTM

PR #2725, opened 2017-03-21
use PropertyString for some builtin string values

MikeHolman:
We see in some cases built-in string values get used as property names in objects. There is big upside from turning these into PropertyStrings in case they are used as PropertyName.

This improves perf of React by about 5%.


lib/Runtime/Library/JavascriptLibrary.h
```diff
@@ -651,7 +650,6 @@ namespace Js
         JavascriptString* GetCommaSpaceDisplayString() { return commaSpaceDisplayString; }
         JavascriptString* GetTrueDisplayString() { return trueDisplayString; }
         JavascriptString* GetFalseDisplayString() { return falseDisplayString; }
-        JavascriptString* GetLengthDisplayString() { return lengthDisplayString; }
```
boingoing: No one uses this ?

MikeHolman: Seems not.

PR #2726, opened 2017-03-21
Migrate Intl tests to ChakraCore.

dilijev:
Closes #2726


test/utf8/OS_2977448.js
```diff
@@ -1,8 +1,14 @@
...
-
-function TrimStackTracePath(line) {
-    return line && line.replace(/\(.+\\test.StackTrace./ig, "(");
-}
+//-------------------------------------------------------------------------------------------------------
+// Copyright (C) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE.txt file in the project root for full license information.
+//-------------------------------------------------------------------------------------------------------
+
+try {
```
boingoing: Is this a rename + edit or a new file? Diff tool seems confused here.

test/utf8/rlexe.xml
```diff
@@ -15,6 +15,13 @@
   </test>
   <test>
     <default>
+      <files>OS_2977448.js</files>
+      <compile-flags>-Intl-</compile-flags>
+      <tags>fail_mutators,exclude_winglob,BugFix</tags>
```
boingoing: I'm not sure but I think fail_mutators may be a dead tag.

test/utf8/rlexe.xml
```diff
@@ -15,6 +15,13 @@
   </test>
   <test>
     <default>
+      <files>OS_2977448.js</files>
+      <compile-flags>-Intl-</compile-flags>
```
boingoing: Is turning off Intl required? I don't know if it would make a difference in this test since the test is not using Intl or any of the built-ins which plumb into platform APIs.

test/Bugs/randombug.baseline
```diff
@@ -63,10 +63,10 @@ Test case 30
 Passed
 Test case 31
 Error
-	at func2 (randombug.js:287:9)
```
boingoing: Consider removing the line numbers and columns from the baseline completely if this test isn't explicitly testing them.

test/utf8/rlexe.xml
dilijev: Yeah, actually this was originally under Intl tests and I moved it since it has nothing to do with Intl features (I added the -Intl- flag to double-check that it wasn't dependent on Intl). I can remove the tag.

test/utf8/rlexe.xml
dilijev: It's used in enough places that I didn't want to make that call at this time.

test/Bugs/randombug.baseline
dilijev: Fair enough. I didn't really intend to change this test, it was just part of a refactoring.

test/utf8/OS_2977448.js
dilijev: Looks like it got confused with the TrimStackTracePath file because it's also a small file and mostly similar thanks to the copyright header. This is a test migrated from full to core. It was in the Intl directory but it turned out not to be an Intl test but rather a Unicode test so I moved it to the utf8 directory (which also isn't a great fit since it's not about the encoding so much as the Unicode code point -- but it's a better fit anyway).

---
In reply to: [107275264](https://github.com/Microsoft/ChakraCore/pull/2726#discussion_r107275264) [](ancestors = 107275264)

test/Bugs/randombug.baseline
dilijev: I think we could change the implementation of TrimStackTracePath or add an alternate function which additionally trims line numbers. But I'd say that's out of scope of this change. I don't want to consider the intent behind testing (or not) the line numbers of error stack traces.

---
In reply to: [107286850](https://github.com/Microsoft/ChakraCore/pull/2726#discussion_r107286850) [](ancestors = 107286850)

jianchun: Just to chime in about `TrimStackTracePath`: a few days ago I saw in Full that when unit tests run with `-BVT` (by default), we strip away paths from stack traces inside runtime. Maybe we can consolidate these in future. There are probably more flags for related/similar purposes: `-BaselineMode` (and `ExtendedErrorStackForTestHost`? not sure).

dilijev: Interesting. Three of those test folders also had their own version TrimStackTracePath which I refactored to the one file now found under UnitTestFramework in core.

dilijev: I'll mark this as Won't Fix and follow up on this in the future. As I work on Intl these tests will likely evolve and I'll probably revisit some issues like this.

---
In reply to: [107526651](https://github.com/Microsoft/ChakraCore/pull/2726#discussion_r107526651) [](ancestors = 107526651)

test/utf8/OS_2977448.js
dilijev: To be specific, the actual refactoring here deleted multiple copies of this script across full and core and moved one definitive copy to UnitTestFramework. Here's what the command line has to say:

```
>git show --name-status | grep --color StackTracePath
R083    test/Error/TrimStackTracePath.js        test/UnitTestFramework/TrimStackTracePath.js
D       test/inlining/TrimStackTracePath.js
R064    test/StackTrace/TrimStackTracePath.js   test/utf8/OS_2977448.js
```

---
In reply to: [107290021](https://github.com/Microsoft/ChakraCore/pull/2726#discussion_r107290021) [](ancestors = 107290021,107275264)

PR #2727, opened 2017-03-21
fix proxy enumerator

leirocks:
should not use an iterator to implement the enumerator, use a customized enumerator instead
incorperate @curtisman change for all enumerator MoveAndGetNext to return JavascriptString


lib/Runtime/Library/JavascriptProxy.cpp
```diff
@@ -956,55 +956,75 @@ namespace Js
...
+            ProxyOwnkeysEnumerator(ScriptContext* scriptContext, JavascriptProxy* proxy, JavascriptArray* trapResult)
+                :JavascriptEnumerator(scriptContext), scriptContext(scriptContext), proxy(proxy), trapResult(trapResult)
             {
-                return FALSE;
+                visited = RecyclerNew(scriptContext->GetRecycler(), VisitedNamesHashSet, scriptContext->GetRecycler());
             }
-            while ((propertyName = trapEnumerator.MoveAndGetNext(propertyId)) != NULL)
+            JavascriptStaticEnumerator* GetInnerEnumerator() { return &innerEnumerator; }
+            virtual uint32 GetCurrentItemIndex()  override { return index; }
+            virtual void Reset() override
```
curtisman: You don't have to override this.  This is only neededfor JSONParser to quickly determine if the value of an array index.    You are not returning an array index here.

lib/Runtime/Library/JavascriptProxy.cpp
```diff
@@ -956,55 +956,75 @@ namespace Js
...
-                PropertyId  propId = JavascriptOperators::GetPropertyId(propertyName, requestContext);
-                Var prop = JavascriptOperators::GetProperty(RecyclableObject::FromVar(trapResult), propId, requestContext);
-                // if (typeof key === "string") {
-                if (JavascriptString::Is(prop))
+                index = 0;
+                innerEnumerator.Reset();
+                visited->Reset();
+            }
+
+            virtual JavascriptString * MoveAndGetNext(PropertyId& propertyId, PropertyAttributes* attributes = nullptr) override
```
curtisman: >propertyId [](start = 66, length = 10)

You not return in the propertyId of the array, but the propertyId of the property Name.

lib/Runtime/Library/JavascriptProxy.cpp
```diff
@@ -956,55 +956,75 @@ namespace Js
...
+            {
+                Var var = nullptr;
```
curtisman: Separate the return from MoveAndGetNext and JavascriptOperators::GetProperty, so you don't need to demote the return MoveAndGetNext form JavascriptString to Var.

lib/Runtime/Library/JavascriptProxy.cpp
```diff
@@ -956,55 +956,75 @@ namespace Js
...
+                // 13.7.5.15 EnumerateObjectProperties(O) (https://tc39.github.io/ecma262/#sec-enumerate-object-properties)
+                // for (let key of Reflect.ownKeys(obj)) {
+                while ((var = innerEnumerator.MoveAndGetNext(propertyId, attributes)) != nullptr)
```
curtisman: It may be safer to walk the value of the array index directly instead of using iterator and JavascirptOperators::GetProperty to avoid unintentional activation of user reentrencies (getter/setter, etc.)
(There isn't any in the spec, right?)

lib/Runtime/Library/JavascriptProxy.cpp
leirocks: however Reset is a pure virtual function, remove the pure virtual in the base?

leirocks: ah you meant GetCurrentItemIndex?

lib/Runtime/Library/JavascriptProxy.cpp
```diff
@@ -956,55 +956,71 @@ namespace Js
...
+            }
+            virtual void Reset() override
+            {
+                index = 0;
+                visited->Reset();
             }
-            while ((propertyName = trapEnumerator.MoveAndGetNext(propertyId)) != NULL)
+
+            virtual JavascriptString * MoveAndGetNext(PropertyId& propertyId, PropertyAttributes* attributes = nullptr) override
             {
```
curtisman: propertyId needs to be initialized to Constants::NoProperty.

Yongqu: Do we have spec change here? I remember the spec requires an iterator and the behavior is observable. OTOH, isn't the enumerator trap removed, only iterator trap left?

leirocks: the trapResult array is a new created internal array, not the original user specified array. the elements in the trapResult array is added from user specified array after some filtering.
going through with index is essentially the same as going through with enumerator.

see an issue (#2755) I opened while going through the code path

PR #2729, opened 2017-03-22
Enabling eval in param scope

aneeshdk:
Enabling eval usage in param scope by default. Removed the duplicate
symbols from body scope. Also removed the work around used for special
symbols like this, super etc.


aneeshdk:

lib/Parser/Parse.cpp
```diff
@@ -938,7 +924,8 @@ Symbol* Parser::AddDeclForPid(ParseNodePtr pnode, IdentPtr pid, SymbolType symbo
             AnalysisAssert(pnodeFnc);
             if (pnodeFnc->sxFnc.pnodeName &&
                 pnodeFnc->sxFnc.pnodeName->nop == knopVarDecl &&
-                pnodeFnc->sxFnc.pnodeName->sxVar.pid == pid)
+                pnodeFnc->sxFnc.pnodeName->sxVar.pid == pid &&
+                (pnodeFnc->sxFnc.IsBodyAndParamScopeMerged() || scope->GetScopeType() == ScopeType_Parameter))
             {
                 // Named function expression has its name hidden by a local declaration.
```
pleath: As we discussed, it should be possible to remove the func-expr-name-is-hidden attribute, either now or in a separate change.

aneeshdk: I will do that in a separate change

PR #2730, opened 2017-03-23
Numerous fixes to get ASAN working for CI

Penguinwizzard:
This is approximately a final version; I still need to run perf tests, but the code is in a state where a number of bugs have been fixed, the ASAN builds pass all tests (excluding ones which I've tagged exclude_sanitize_address and filed GitHub issues for), and it's ready for thorough review.

### DIFF
--- a/bin/ch/ch.cpp
+++ b/bin/ch/ch.cpp
@@ -825,6 +825,7 @@ int main(int argc, char** c_argv)
 // xplat-todo: PAL free CH ?
     PAL_InitializeChakraCore();
 #endif
+    int origargc = argc; // store for clean-up later
     argv = new char16*[argc];
     for (int i = 0; i < argc; i++)
     {
@@ -839,22 +840,29 @@ int _cdecl wmain(int argc, __in_ecount(argc) LPWSTR argv[])
 #ifdef _WIN32
     bool runJITServer = HandleJITServerFlag(argc, argv);
 #endif
+    int retval = -1;
+    HRESULT exitCode = E_FAIL;
+    int cpos = 1;
+    HINSTANCE chakraLibrary = nullptr;
+    bool success = false;
+    ChakraRTInterface::ArgInfo argInfo;
 
     if (argc < 2)
     {
         PrintUsage();
         PAL_Shutdown();
-        return EXIT_FAILURE;
+        retval = EXIT_FAILURE;
+        goto return_cleanup;
     }
 
 #ifdef _WIN32
     if (runJITServer)
     {
-        return RunJITServer(argc, argv);
+        retval = RunJITServer(argc, argv);
+        goto return_cleanup;
     }
 #endif
 
-    int cpos = 1;
     for(int i = 1; i < argc; ++i)
     {
         const wchar *arg = argv[i];
@@ -884,7 +892,8 @@ int _cdecl wmain(int argc, __in_ecount(argc) LPWSTR argv[])
         {
             PrintVersion();
             PAL_Shutdown();
-            return EXIT_SUCCESS;
+            retval = EXIT_SUCCESS;
+            goto return_cleanup;
         }
         else if (
 #if !defined(ENABLE_DEBUG_CONFIG_OPTIONS) // release builds can display some kind of help message
@@ -896,7 +905,8 @@ int _cdecl wmain(int argc, __in_ecount(argc) LPWSTR argv[])
         {
             PrintUsage();
             PAL_Shutdown();
-            return EXIT_SUCCESS;
+            retval = EXIT_SUCCESS;
+            goto return_cleanup;
         }
         else if(wcsstr(argv[i], _u("-TTRecord=")) == argv[i])
         {
@@ -927,7 +937,9 @@ int _cdecl wmain(int argc, __in_ecount(argc) LPWSTR argv[])
         }
         else
         {
+            wchar *temp = argv[cpos];
             argv[cpos] = argv[i];
+            argv[i] = temp;
             cpos++;
         }
     }
@@ -951,9 +963,8 @@ int _cdecl wmain(int argc, __in_ecount(argc) LPWSTR argv[])
 
     HostConfigFlags::HandleArgsFlag(argc, argv);
 
-    ChakraRTInterface::ArgInfo argInfo = { argc, argv, PrintUsage, nullptr };
-    HINSTANCE chakraLibrary = nullptr;
-    bool success = ChakraRTInterface::LoadChakraDll(&argInfo, &chakraLibrary);
+    argInfo = { argc, argv, PrintUsage, nullptr };
+    success = ChakraRTInterface::LoadChakraDll(&argInfo, &chakraLibrary);
 
 #if defined(CHAKRA_STATIC_LIBRARY) && !defined(NDEBUG)
     // handle command line flags
@@ -965,7 +976,6 @@ int _cdecl wmain(int argc, __in_ecount(argc) LPWSTR argv[])
         WideStringToNarrowDynamic(argv[1], &argInfo.filename);
     }
 
-    HRESULT exitCode = E_FAIL;
     if (success)
     {
 #ifdef _WIN32
@@ -1006,5 +1016,20 @@ int _cdecl wmain(int argc, __in_ecount(argc) LPWSTR argv[])
 #endif
 
     PAL_Shutdown();
-    return (int)exitCode;
+    retval = (int)exitCode;
+
+return_cleanup:
+#ifndef _WIN32
+    if(argv != nullptr)
+    {
+        for(int i=0;i<origargc;i++)
+        {
+            free(argv[i]);
+            argv[i] = nullptr;
+        }
+    }
+    delete[] argv;
+    argv = nullptr;
+#endif
+    return retval;
 }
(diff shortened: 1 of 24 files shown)


lib/Common/DataStructures/FixedBitVector.h
```diff
@@ -317,7 +318,7 @@ void BVFixed::SetRange(Container* value, BVIndex start, BVIndex len)
     {
         // Calculation above might overflow the original container.
         // normalize the overflow value. LE only
-        temp = (*((char*)bits)) + (*(((char*)bits) + 1));
+        //temp = (*((char*)bits)) + (*(((char*)bits) + 1));
```
Penguinwizzard: Review: This variable looks like it's never used. Are we setting a flag bit for the calling function?

lib/Backend/NativeCodeData.h
```diff
@@ -42,6 +42,7 @@ class NativeCodeData
 
     struct DataChunkNoFixup
     {
+        unsigned int len; // only used for memory management purposes.
```
Penguinwizzard: Review: This is one of the things that I'm looking to perf and functionality test.

lib/Common/DataStructures/FixedBitVector.h
```diff
@@ -242,18 +242,19 @@ void BVFixed::SetRange(Container* value, BVIndex start, BVIndex len)
     BVIndex oStart = BVUnit::Offset(start);
     BVIndex oEnd = BVUnit::Offset(end);
 
-    BVUnit::BVUnitTContainer temp;
-    BVUnit::BVUnitTContainer* bits;
+    typedef typename std::conditional<sizeof(Container) == 1,uint8_t,BVUnit::BVUnitTContainer>::type operandtype;
```
Penguinwizzard: Note: I'm aware that this change could have a performance penalty if it goes the wrong way; I'm looking at it.

Penguinwizzard:
@obastemur  You mentioned earlier that you were interested in this, so tagging you for the notification.

lib/Common/DataStructures/FixedBitVector.h
obastemur: Appreciate if we don't introduce more from std on core.

lib/Common/Memory/MarkContext.inl
```diff
@@ -46,6 +46,11 @@ bool MarkContext::AddTrackedObject(FinalizableObject * obj)
 #endif
 
 template <bool parallel, bool interior, bool doSpecialMark>
+#if defined(__has_feature)
+#if __has_feature(address_sanitizer)
```
obastemur: You may introduce a NO_SANITIZE definition into CommonPal instead? Same for other declarations below

test/runtests.py
```diff
@@ -125,11 +126,16 @@
 if sys.platform != 'win32':
     not_tags.add('exclude_xplat')
     not_tags.add('Intl')
+    not_tags.add('require_simd')
+    not_tags.add('require_asmjs')
```
obastemur: ?

Penguinwizzard: This is all related to the move from excluding anything with the argument -simdjs to actually tagging the tests appropriately. As this changeset doesn't include the changes in #2716, it doesn't have working asmjs/wasm on xplat. As I expect that #2716 will get into master before this does, I'll rebase on top of that and remove the require_asmjs and require_wasm flags as appropriate from this changeset before checking in.

lib/Common/Memory/MarkContext.inl
Penguinwizzard: I'll add a NO_SANITIZE_ADDRESS definition into CommonPal.

lib/Common/DataStructures/FixedBitVector.h
Penguinwizzard: Fair point; I'll just mark this whole function as NO_SANITIZE_ADDRESS instead, and that should make the changeset significantly simpler too.

lib/Common/DataStructures/FixedBitVector.h
Penguinwizzard: Changing back to the previous version, since I'm just marking this function as NO_SANITIZE_ADDRESS anyway.

test/runtests.py
Penguinwizzard: Note: in the new version of the PR, this is changed so the require_asmjs flag is no longer there, since those tests now work on xplat.

lib/Common/Memory/MarkContext.inl
Penguinwizzard: Change is introduced and now in the PR.

bin/ch/ch.cpp
```diff
@@ -671,6 +671,10 @@ HRESULT ExecuteTest(const char* fileName)
         }
     }
 Error:
+    if( fileContents != nullptr)
+    {
+        free((char*)fileContents);
```
obastemur: IIRC `fileContents` is used as the source of an ExternalArrayBuffer and we shouldn't free it this way

Penguinwizzard: It is, but the ExternalArrayBuffer is cleaned up before this. I'll rescope this fix properly.

Penguinwizzard: Actually, virtually no instance of ExternalArrayBuffer in ch actually cleans up its memory properly... I'll roll that set of fixes in here.

jianchun: > Actually, virtually no instance of ExternalArrayBuffer in ch actually cleans up its memory properly

What do you mean exactly? Current runtime stores a reference to the ExternalArrayBuffer when it's used as source, and when GC finally collects the ExternalArrayBuffer instance it'll clean up.

Penguinwizzard: Each ExternalArrayBuffer can take a finalizeCallback argument and a callbackState argument. On finalize, the finalizeCallback, if it's non-null, is called on the callbackState. There are indeed a few cases where a finalizeCallback is specified in ch (as a lambda wrapper around free()). However, in the few cases where it is specified, the provided callbackState is nullptr. This means that it does not actually do anything. In the case of calling free, free(nullptr) is specified as doing nothing. This changeset already includes fixes to the cases in WScriptJsrt where ExternalArrayBuffer was just being called without finalizeCallbacks (=nullptr) and was leaking the source body of every script run. There's two locations in ch.cpp where the finalizeCallback was specified, but the callbackState is nullptr, and one in ch.cpp and one in Debugger.cpp where both are nullptr. Manual memory management of the style at the top of this comment chain is the reason that the non-leaking cases weren't leaking. There's a few situations in which this is actually complicated a bit; For instance, there's at least one path that creates multiple ExternalArrayBuffers for the same backing buffer.

jianchun: I see, you are right, those are bugs.

Penguinwizzard: I added a commit to the changeset that attempts to address this, and an ASAN run after adding it came back clean. There's likely some work that could be done on the JSRT api related to this though; there doesn't seem to be a good way to, for example, handle a case where you have some data that you want to hold around needed to create the source for a serialized function with a deferred source initializer; even if we were to add the finalize callback for it, it doesn't handle the case where it's never needed, where we'd still need to free the data that would be needed in the case that the deferred initializer were to run.

Penguinwizzard:
@dotnet-bot test OSX _no_jit_shared_osx_osx_test please

bin/ch/ch.cpp
jianchun: > a deferred source initializer; even if we were to add the finalize callback for it, it doesn't handle the case where it's never needed, where we'd still need to free the data that would be needed in the case that the deferred initializer were to run

In that case I'd assume that data's lifetime is attached to the serialized bytecode buffer's lifetime. When the serialized buffer is being released, the deferred source initializer data can be released. I can try manage the data in JsCreateExternalArrayBuffer's callBackState.

Penguinwizzard: Yeah, that's what I'm doing in the attempt that I added, and I shoehorned in some other data, but it only really works because I know externally to the actual code that we're completely done with the source buffer and all data needed to create it. Let's say we have a zip file containing both cached bytecode and source code as separate sections, and we want to create a function to pass around with JsParseSerializedScript. While we can use this function to have a loadCallback, and in that return an ExternalArrayBuffer with a finalizeCallback, we don't have a good way to know "we no longer need the ability to construct the source buffer for this, so you can free the data needed to construct it", which would let us free the memory being taken up by the compressed version of the source code. Here we only get away with it because we're using JsRunSerializedScript, and I know afterwards that we're not going to do any more calls so I can free the buffer that would be used to construct the ExternalArrayBuffer if it were to be needed, but in more complex cases I think we'd need another callback or something to properly handle the scoping.

jianchun: I meant the ExternalArrayBuffer for the bytecode buffer created to call JsParseSerializedScript, not ExternalArrayBuffer for the script source created in the loadCallback. When the bytecode ExternalArrayBuffer is being finalized, you "no longer need the ability to construct the source".

Penguinwizzard: The bytecode buffers are being passed in as just byte*s, not as ExternalArrayBuffers, so we can't register the callback that way, as far as I can tell.

jianchun: That's true for the old API. The new API takes ArrayBuffer: https://github.com/Microsoft/ChakraCore/blob/master/lib/Jsrt/Jsrt.cpp#L4491

Penguinwizzard: Ah, I didn't see that due to the wiki being incorrect. I've updated the wiki with the proper types now for JsParseSerialized and JsRunSerialized.

PR #2740, opened 2017-03-24
xplat: back short interval GetSystemTime calls via cpu-ticks

obastemur:


lib/Runtime/PlatformAgnostic/Platform/Common/HiResTimer.cpp
```diff
@@ -56,18 +52,34 @@ namespace DateTime
...
     }
 
+    double HiResTimer::GetSystemTime()
+    {
+        ULONGLONG current = GetTickCount64();
+        ULONGLONG diff = current - data.cacheTick;
+
+        if (diff > 5) // max 5ms to respond system time changes
+        {
+            data.cacheSysTime = GetSystemTimeREAL();
```
obastemur: self-note: add `Assert(GetSystemTimeREAL() >= data.cacheSysTime + 5)` and add a slow test case

jianchun:
:shipit:

obastemur:
@curtisman @jianchun Thanks for the review

PR #2741, opened 2017-03-24
Revert "Fix delete InlineeEnd incorrectly"

rajatd:
This reverts changes from #2582, #2647, and #2652. #2582 fixed a bug around InlineeEnd and BailOnNoProfile but introduced more bugs which the later two commits attempted to fix. Currently there are a lot of bugs coming from these set of commits. So, reverting them for now - I have a different fix in mind for which I will create a PR soon.


rajatd:
@LouisLaf please review

PR #2742, opened 2017-03-25
Replace printf and wprintf with Output::Print where possible and update formatter strings

dilijev:
* Replace printf with wprintf and update formatter strings.
* Replace wprintf with Output::Print except in ch where it would be a new dependency.
* Convert L"" strings to _u("") strings in more places where they were apparently overlooked before.

---

Under #2712 printf will be broken on Windows because it does not understand utf8. See discussion in that PR.

This PR includes #2712's changes which are relevant for testing this change. However, I opened this as a separate PR for the sake of keeping discussion and CI separate. Please review the new commit which is specific to this PR.

The most concerning area will be usages of printf in the PAL. I left alone any code which is not part of the core engine and any code which will not run under ch.exe.


dilijev:
The command I used to find instances of printf calls is the following (the files excluded are files I determined to not be a part of the regular build 

```sh
git grep -n "\bprintf(" *.cpp *.h | grep -v -E "((^tools)|(^bin\/rl)|(native\-tests)|GCStress|CaseInsensitive|JavascriptExceptionOperators|printfcpp|pal\.h)" | grep --color "\bprintf("
```

@jianchun Is it alright to use L"" instead of _u("") in the PAL for the purposes of wprintf since we don't need those strings to be u"" (UTF-8 strings) on xplat? This will avoid the problems with _u not being defined in the PAL.

dilijev:
@jianchun Is it alright to use L"" instead of _u("") in the PAL for the purposes of wprintf since we don't need those strings to be u"" (UTF-8 strings) on xplat? This will avoid the problems with _u not being defined in the PAL.

digitalinfinity:
Doesn't the PAL define it's own wprintf, which actually operates on char16_t? If that one is used, using L"" would be incorrect. It's been a while since I looked at that code so I could be wrong though...

dilijev:
@digitalinfinity The PAL does indeed define it's own wprintf but it wasn't clear whether the system's version or the PAL's version would be used in this case. However, throughout ChakraCore we are always using the PAL's version of wprintf (which apparently uses L"" on Windows and u"" on Linux, based on the behavior of existing code which does use wprintf), I could try to debug into it I suppose. If it is the PAL's version, using u"" within the PAL should work xplat.

obastemur:
Can we define/use something like 'OUTPUTW/OUTPUT' and target per platform? i.e. You could use wprintf for both on Windows... 

This was on my personal todo list for Android logcat / delayed function call tracing etc.. There we change everything... Can we make it once for all?

dilijev:
We've already got a routine `Output::Print` which might be a good place to do this. It is used in various places already and the instances of printf are anomalies. We could switch everything over to that and then target platforms as needed.

jianchun:
How about skipping `pal`? `pal` is only used on xplat and as I understand xplat did not have the console output encoding issue. Could let it continue to use `printf`?

I was going to mention `Output::Print` too. Replacing everywhere might incur some header/lib dependency issue though.

dilijev:
@jianchun Thanks for confirming that PAL is only used on non-Windows. I thought that might be the case, but I wasn't sure. I'll revert the PAL changes (which were the most concerning ones anyway).

dilijev:
Reverted changes to PAL. Changed wprintf to Output::Print in ChakraCore (not in ch.exe or other test code where it would be a new dependency). Also fixed a few instances of L"" being used instead of _u("").

dilijev:
Will combine this change into #2712

PR #2743, opened 2017-03-25
build: Support ChakraCore with VC++ Build Tools

digitalinfinity:
DelayLoadLibrary was including cor.h for metadata reading However, this is used only for projection support in Chakra.dll

Disabled this in ChakraCore so that node-chakracore can build with VC++ build tools.

### DIFF
--- a/lib/Runtime/Base/DelayLoadLibrary.cpp
+++ b/lib/Runtime/Base/DelayLoadLibrary.cpp
@@ -484,30 +484,4 @@ namespace Js
 #endif
         return FALSE;
     }
-
-    // Implement this function inlined so that WinRT.lib can be used without the runtime.
-    HRESULT DelayLoadWinType::RoGetMetaDataFile(
-        _In_ const HSTRING name,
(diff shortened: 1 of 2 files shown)


digitalinfinity:
@curtisman please review (cc @joaocgreis)

This also fixes Nodejs/Node-ChakraCore#200

digitalinfinity:
@curtisman I assume you meant RoGetMetaDataFile? It looks like RoGetParameterizedTypeInstanceIID is still used by projection. I deleted the former (running a full build now to verify)

curtisman:
@digitalinfinity  Yes.  Probably want to move RoGetParameterizedTypeInstanceID out of core too?

digitalinfinity:
@curtisman module lifetime management is done on the ThreadContext, so unfortunately, it's not trivial to move this out of core- my vote is to #ifdef it out for now, and we should have a separate issue tracking management of module lifetimes independent of the thread context.

curtisman:
Ok.
:shipit:

PR #2744, opened 2017-03-27
WASM: update wabt and spec tests

Cellule:
Had to completely remove wabt from legacy build because it just doesn't build anymore.
Account for the 2 new nan assert types.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/2744)


Cellule:
@MikeHolman @curtisman Can you review this, mostly the changes I made for the build (ie: project file/property sheets). 
Aside from that the only things I changed are in `test/WasmSpec/spec.js` `WabtInterface.cpp`.
YOu can review by commits and skip the commits updating wabt and the spec tests

Build/Common.Build.props
```diff
@@ -86,6 +86,8 @@
         %(PreprocessorDefinitions);
         BYTECODE_TESTING=1
       </PreprocessorDefinitions>
+      <PreprocessorDefinitions Condition="'$(BuildWabt)'=='true'">%(PreprocessorDefinitions);EXCLUDE_WABT=0</PreprocessorDefinitions>
```
curtisman: nit:  May want to use a positive condition like ENABLE_WABT, instead of a negative one like EXCLUE_WABT to be slightly clearer?

Cellule: Right now we already use `ENABLE_WABT` in the code. `EXCLUDE_WABT` is only use to not set `ENABLE_WABT` along with some other conditions in CommonDefines.h

Cellule:
Right now I don't know how to fix the xplat break.
It looks like wabt's header files included in WabtInterface are conflicting with other includes in the Runtime.Library project.
I am not sure how to get a project that can use both wabt and chakra. 
I might have to create an api file in wabt with the minimal set of things exposed.

MikeHolman:
:shipit:

PR #2756, opened 2017-03-29
pass prop string as optional arg for hasOwnProperty; remove dictionary lookup

MikeHolman:
HasOwnProperty helper was a bit inefficient where it often gets a PropertyString as an argument, gets the propertyId from that and passes it to another helper which does a dictionary lookup to see if string is cached.

Technically the dictionary lookup might find PropertyString if argument was passed in as some other type of JavascriptString, but mostly we will be there in the negative case, so I removed the dictionary lookup. It's also unclear if that dictionary lookup is even faster than the actual property lookup.

Improves react-redux by 4%.

### DIFF
--- a/lib/Runtime/Base/ScriptContext.cpp
+++ b/lib/Runtime/Base/ScriptContext.cpp
@@ -851,7 +851,7 @@ namespace Js
         return propertyRecord->GetPropertyId();
     }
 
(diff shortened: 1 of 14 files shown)


tcare:
LGTM, but please get additional signoff.

PR #2760, opened 2017-03-29
Memory allocation fixes and improvements

obastemur:
#### xplat: fix memory allocations. [align-allocate memory address to 64KB]
 - Improves xplat perf and reduces memory usage significantly.

#### xplat: do not reserve executable memory
 - Reduces xplat memory usage slightly

#### improve allocation loop on main thread
 - Improves perf slightly on both Win/Xplat

#### assert address_ptr % 64KB [ so we don't break in the future ]

### DIFF
--- a/lib/Common/Memory/PageAllocator.cpp
+++ b/lib/Common/Memory/PageAllocator.cpp
@@ -124,6 +124,8 @@ SegmentBase<T>::Initialize(DWORD allocFlags, bool excludeGuardPages)
         return false;
     }
 
+    Assert( ((ULONG_PTR)this->address % (64 * 1024)) == 0 );
+
     originalAddress = this->address;
     bool committed = (allocFlags & MEM_COMMIT) != 0;
     if (addGuardPages)
--- a/lib/Common/Memory/PageAllocator.cpp
+++ b/lib/Common/Memory/PageAllocator.cpp
@@ -284,7 +284,7 @@ PageSegmentBase<T>::Prime()
 
 template<typename T>
 bool
-PageSegmentBase<T>::IsAllocationPageAligned(__in char* address, size_t pageCount)
+PageSegmentBase<T>::IsAllocationPageAligned(__in char* address, size_t pageCount, uint *nextIndex)
 {
     // Require that allocations are aligned at a boundary
     // corresponding to the page count
@@ -297,6 +297,11 @@ PageSegmentBase<T>::IsAllocationPageAligned(__in char* address, size_t pageCount
     if ((reinterpret_cast<uintptr_t>(address)& mask) == 0)
     {
         return true;
+    }
+
+    if (nextIndex != nullptr)
+    {
+        *nextIndex = (uint) ((reinterpret_cast<uintptr_t>(address) % (mask + 1)) / AutoSystemInfo::PageSize);
     }
 
     return false;
@@ -332,9 +337,14 @@ PageSegmentBase<T>::AllocPages(uint pageCount)
 
             if (pageCount > 1 && !notPageAligned)
             {
-                if (!IsAllocationPageAligned(allocAddress, pageCount))
+                uint nextIndex = 0;
+                if (!IsAllocationPageAligned(allocAddress, pageCount, &nextIndex))
                 {
-                    index = this->freePages.GetNextBit(index + 1);
+                    if (index + nextIndex >= this->GetAllocator()->GetMaxAllocPageCount())
+                    {
+                        return nullptr;
+                    }
+                    index = this->freePages.GetNextBit(index + nextIndex);
                     continue;
                 }
             }
@@ -400,9 +410,14 @@ PageSegmentBase<TVirtualAlloc>::AllocDecommitPages(uint pageCount, T freePages, 
 
             if (!notPageAligned)
             {
-                if (!IsAllocationPageAligned(pages, pageCount))
+                uint nextIndex = 0;
+                if (!IsAllocationPageAligned(pages, pageCount, &nextIndex))
                 {
-                    index = freeAndDecommitPages.GetNextBit(index + 1);
+                    if (index + nextIndex >= this->GetAllocator()->GetMaxAllocPageCount())
+                    {
+                        return nullptr;
+                    }
+                    index = freeAndDecommitPages.GetNextBit(index + nextIndex);
                     continue;
                 }
             }
--- a/pal/src/include/pal/virtual.h
+++ b/pal/src/include/pal/virtual.h
@@ -1,6 +1,6 @@
 //
 // Copyright (c) Microsoft. All rights reserved.
-// Licensed under the MIT license. See LICENSE file in the project root for full license information. 
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
 //
 
 /*++
@@ -27,7 +27,7 @@ extern "C"
 #endif // __cplusplus
 
 typedef struct _CMI {
-    
+
     struct _CMI * pNext;        /* Link to the next entry. */
     struct _CMI * pLast;        /* Link to the previous entry. */
 
@@ -53,7 +53,7 @@ enum VIRTUAL_CONSTANTS
     /* Allocation type. */
     VIRTUAL_COMMIT_ALL_BITS     = 0xFF,
     VIRTUAL_RESERVE_ALL_BITS    = 0x0,
-    
+
     /* Protection Type. */
     VIRTUAL_READONLY,
     VIRTUAL_READWRITE,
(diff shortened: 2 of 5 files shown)


pal/src/include/pal/virtual.h
```diff
@@ -105,108 +105,6 @@ BOOL VIRTUALOwnedRegion( IN UINT_PTR address );
...
 #ifdef __cplusplus
 }
-
-/*++
-Class:
-    ExecutableMemoryAllocator
-
-    This class implements a virtual memory allocator for JIT'ed code.
-    The purpose of this allocator is to opportunistically reserve a chunk of virtual memory
-    that is located near the coreclr library (within 2GB range) that can be later used by
```
curtisman: Although we don't do this now because it is a hassle to do it in Windows, may be can make use of this in our JIT'ed code as well?

pal/src/map/virtual.cpp
```diff
@@ -1682,6 +1671,81 @@ VirtualAlloc(
...
+                addr64 = address + (KB64 - diff);
+                VirtualFree(address, 0, MEM_RELEASE);
+
+                address = (char*) VirtualAlloc_(addr64, dwSize, MEM_RESERVE, flProtect);
+                if (!address)
+                {
+                    allocationFailedCount++;
+                    if (allocationFailedCount > 2) return nullptr;
+                    // address is still not available. OutOfMemory?
+                    return VirtualAlloc(nullptr, dwSize, flAllocationType, flProtect);
```
curtisman: Can we avoid recursion?  I am sure you can write this in a loop.  This avoid having the TLS as well.

pal/src/include/pal/virtual.h
obastemur: On XPLAT, both has the same backing proxy hence no difference apart from making our bootstrap process heavy and complicated.

pal/src/map/virtual.cpp
obastemur: If we are hitting this, most likely we are out of memory. Looping there has no practical difference. I could indeed implement a loop there but didn't want to make it look like (if we wait here enough, we may get free memory). When the function re-enters if we are out of juice, it won't end up here again. If we are not,unlikely to come back to here.

obastemur: Wrapped last trial with a critical section to make sure the reason here is OOM.

lib/Common/Memory/PageAllocator.cpp
```diff
@@ -124,6 +124,8 @@ SegmentBase<T>::Initialize(DWORD allocFlags, bool excludeGuardPages)
         return false;
     }
 
+    Assert( ((ULONG_PTR)this->address % (64 * 1024)) == 0 );
```
digitalinfinity: Instead of 64 * 1024, can you use AutoSystemInfo::Data.dwAllocationGranularity here? Also, in SysInfo.cpp, after the call to GetSystemInfo in AutoSystemInfo's constructor, it's worth adding a similar assert (that dwAllocationGranularity is 64k)

lib/Common/Memory/PageAllocator.cpp
```diff
@@ -284,7 +286,7 @@ PageSegmentBase<T>::Prime()
 
 template<typename T>
 bool
-PageSegmentBase<T>::IsAllocationPageAligned(__in char* address, size_t pageCount)
+PageSegmentBase<T>::IsAllocationPageAligned(__in char* address, size_t pageCount, uint *nextIndex)
```
digitalinfinity: Can you add a comment about what the intent of nextIndex here is?

lib/Common/Memory/PageAllocator.cpp
```diff
@@ -299,6 +301,11 @@ PageSegmentBase<T>::IsAllocationPageAligned(__in char* address, size_t pageCount
         return true;
     }
 
+    if (nextIndex != nullptr)
+    {
+        *nextIndex = (uint) ((reinterpret_cast<uintptr_t>(address) % (mask + 1)) / AutoSystemInfo::PageSize);
```
digitalinfinity: Does nextIndex need to be computed this way or would it also work to just set nextIndex to pageCount? (in which case we don't need this out parameter in the first place)

digitalinfinity: Actually, probably not- nextIndex is intended to be the index of the next *aligned* page, yes? That makes sense

pal/src/map/virtual.cpp
```diff
@@ -878,15 +870,14 @@ static LPVOID VIRTUALReserveMemory(
...
 
-    InternalEnterCriticalSection(pthrCurrent, &virtual_critsec);
-
-    // If this is a request for special executable (JIT'ed) memory then, first of all,
-    // try to get memory from the executable memory allocator to satisfy the request.
-    if (((flAllocationType & MEM_RESERVE_EXECUTABLE) != 0) && (lpAddress == NULL))
+    if ((flAllocationType & MEM_RESERVE_EXECUTABLE) != 0)
     {
-        pRetVal = g_executableMemoryAllocator.AllocateMemory(MemSize);
+        fprintf(stderr, "MEM_RESERVE_EXECUTABLE is not supported!");
```
digitalinfinity: Actually, we do have a preReservedSegment in the JIT (for CFG cases)- it's turned off on xplat because CFG support isn't there but I think clang's CFI (https://clang.llvm.org/docs/ControlFlowIntegrity.html) provides the same support so we should look to turning it back on soon. In which case, if the preReservedSegment mimicked this mechanism, I think it could be nice. @curtisman thoughts?

pal/src/map/virtual.cpp
```diff
@@ -1682,6 +1673,73 @@ VirtualAlloc(
...
+            {   // looks like ``pushed new address + dwSize`` is not available
+                // try on a bigger surface
+                address = (char*) VirtualAlloc_(nullptr, dwSize + KB64, MEM_RESERVE, flProtect);
+                if (!address) return nullptr;
+
+                diff = ((ULONG_PTR)address % KB64);
+                addr64 = address + (KB64 - diff);
+
+                CPalThread *pthrCurrent = InternalGetCurrentThread();
+                InternalEnterCriticalSection(pthrCurrent, &virtual_realloc);
```
digitalinfinity: More a question than a comment- what is the intention behind this critical section? For example, why is this not entered when you first call VirtualAlloc_(add64)

digitalinfinity:
Nice job! One more question- when you say "Improves xplat perf and reduces memory usage significantly", how much is the improvement in each metric?

pal/src/map/virtual.cpp
```diff
@@ -1682,6 +1673,73 @@ VirtualAlloc(
...
+        SIZE_T diff = ((ULONG_PTR)address % KB64);
+        if ( diff != 0 )
+        {
+            VirtualFree(address, 0, MEM_RELEASE);
+            char *addr64 = address + (KB64 - diff);
+
+            // try reserving from the same address space
+            address = (char*) VirtualAlloc_(addr64, dwSize, MEM_RESERVE, flProtect);
+
+            if (!address)
```
leirocks: >if (!address) [](start = 12, length = 13)

just a question: how about doing the alignment at mmap level?

obastemur: how we would end-up doing with mmap and did here are similar. Didn't want to introduce release into alloc for unsuccessful attempts.  Just a reminder here, once we reserve the first big chunk for CardTable and make sure that one is aligned, the rest is aligned as well.

obastemur:
@digitalinfinity saw deltablue at 1GB consumption prior to this update. Now it's ~220 fixed. Perf improvement is somewhere between 1% to 30% depending to type of the app / bench. Octane numbers should be better/reasonable now on xplat.

lib/Common/Memory/PageAllocator.cpp
obastemur: Yes. nextIndex is the first index that can be counted as aligned.

digitalinfinity: Makes sense- please leave a comment to make that nuance clear

pal/src/map/virtual.cpp
obastemur: @curtisman also asked for keeping CLR related stuff that I removed with one of the commits on this PR. ATM, it just makes it more complicated and error prone. IMHO once we want to enable it, we may better approach VirtualAlloc/Free stuff similar to other PlatformAgnostic items. i.e. On some of the bench runs, I saw whole run took ~10 secs while the total time we spent on VirtualFree on xplat was ~1.52 secs.. I want to make some serious changes here but not pre-release indeed.

pal/src/map/virtual.cpp
obastemur: This keeps the memory address for the consumer before it is being allocated by another consumer.

lib/Common/Memory/PageAllocator.cpp
obastemur: Isn't `AutoSystemInfo::Data.dwAllocationGranularity` equal to pageSize ? == 4096?

obastemur:
@curtisman @digitalinfinity @leirocks Thanks for the review

Issue #2667, opened 2017-03-10
Fix and Re-enable flaky test test/es6/proto_disable.js


dilijev:
Test output includes the following line but is otherwise correct:

`WARNING: DynamicProfileStorage: Dynamic Profile Data corrupted: 'e:\a\_work\2\s\test\unittestframework\unittestframework.js'`

Disabled in https://github.com/Microsoft/ChakraCore/pull/2668

dilijev:
Fixed in #2763 
Re-enabled in #2776

dilijev:
Apparently not fixed. Disabling the test again in #3087 and reactivating this issue.

Penguinwizzard:
I suspect that this is also the cause for the issue in proto_initializer, so I've disabled that test as well.

PR #2763, opened 2017-03-30
register scriptContext to the list after globalObject and javascriptLibrary are initialized(2.0)

leirocks:
while ETW kicks in, it goes through the list to do source rundown, which can hit a state that the scriptContext is in the list but javascriptLibrary/utf8sourceList is not initialized yet.
change to register the scriptContext after it's fully initialized
ensure scriptContext is regerstered before unregistering
Clear ScriptContext caches while ScriptContext closing to make the caches to be collected earlier


tcare:
Looks good, but I would like @curtisman to also verify since he has context on the previous change.

dilijev:
(@leirocks says this should fix #2667)

PR #2764, opened 2017-03-30
xplat: Try CXX/CC in case Clang wasn't on the PATH

obastemur:


build.sh
```diff
@@ -377,11 +384,19 @@ else
...
-            echo "You could use clang++ from a custom location."
-            PRINT_USAGE
-            exit 1
+            # try env CXX and CC
+            if [[ ! -f $CXX ]]; then
+                ERROR_CLANG
+            fi
+
+            _CXX=$CXX
+            _CC=$CC
```
digitalinfinity: is it worth validating that both cxx and cc are set in this case?

obastemur: We weren't validating _CC so far. At least I could validate the availability of the one from ENV. Let me update. Thanks!

obastemur:

PR #2765, opened 2017-03-30
remove assertion on existence of expirable object on the list

leirocks:
this can cause hang in chk build

### DIFF
--- a/lib/Runtime/Base/ThreadContext.cpp
+++ b/lib/Runtime/Base/ThreadContext.cpp
@@ -3007,8 +3007,6 @@ ThreadContext::UnregisterExpirableObject(ExpirableObject* object)
 {
     Assert(this->expirableObjectList);
     Assert(object->registrationHandle != nullptr);
-    Assert(this->expirableObjectList->HasElement(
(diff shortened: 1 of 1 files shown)


leirocks:

leirocks:

digitalinfinity:
I wonder if there is a cheaper check we can do to check for the existence of an element in a list. One idea is to add a cookie field to the DListNode class in debug builds- let this cookie simply the be the list that the node is a part of- on removal, we null it out- then we can have a "fastHasElement" version which simply checks if the list is the same as the cookie. Thoughts?

leirocks:
how about clear prev/next pointer after removing node from the DList, like I did for scriptContextList(https://github.com/Microsoft/ChakraCore/pull/2763/files#diff-6858ff2e7f3b97763dd5be71dfc9ae56R421)

```c++
    static void RemoveNode(NodeBase * node)
    {
        node->Prev()->Next() = node->Next();
        node->Next()->Prev() = node->Prev();

        // adding following two lines:
        node->Prev() = nullptr;
        node->Next() = nullptr;
    }
```

PR #2768, opened 2017-03-31
xplat: consider sys.time adjustments on hiResTimer

obastemur:
System time may have small adjustments since the last cache point we have. We will observe the adjustment as if the time wasn't passed. This may create a difference between tickCount and real time.


obastemur:

PR #2777, opened 2017-04-04
xplat: Improve GetCurrent*Thread performance

obastemur:
This PR targets node-chakracore server load testing scenarios.

Perf hungry JS application triggers GetCurrent*Thread more than 10K
times per second. Use front thread_local caching if possible.

### DIFF
--- a/pal/src/thread/pal_thread.cpp
+++ b/pal/src/thread/pal_thread.cpp
@@ -2911,3 +2911,47 @@ bool IsAddressOnStack(ULONG_PTR address)
 
     return false;
 }
+
+#ifndef __IOS__
+// why _Thread_local ? Because it is faster(.)
+// why not replace PAL to use _Thread_local instead of front caching? _Thread_local is not cross platform
+
+// Why ULONG_PTR? keeping type for localThread `simple` may affect implementation hence the perf. positively.
+THREAD_LOCAL ULONG_PTR localThread = 0;
+CPalThread *CorUnix::GetCurrentPalThread(bool force)
+{
+    ULONG_PTR pThread = localThread;
+    if (pThread == 0)
+    {
+        pThread = (ULONG_PTR) reinterpret_cast<CPalThread*>(pthread_getspecific(thObjKey));
+#ifdef FEATURE_PAL_SXS
+        if (pThread == 0 && force)
+        {
+            pThread = (ULONG_PTR) CreateCurrentThreadData();
+        }
+#endif
+        localThread = pThread;
+    }
+
+    return (CPalThread*)pThread;
+}
+#else // !__IOS__
+CPalThread *CorUnix::GetCurrentPalThread(bool force)
+{
+    CPalThread *pThread = reinterpret_cast<CPalThread*>(pthread_getspecific(thObjKey));
+
+#ifdef FEATURE_PAL_SXS
+    if (pThread == nullptr && force)
+    {
+        pThread = CreateCurrentThreadData();
+    }
+#endif
+
+    return pThread;
+}
+#endif
+
+CPalThread *CorUnix::InternalGetCurrentThread()
+{
+    return GetCurrentPalThread(true);
+}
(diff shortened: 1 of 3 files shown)


pal/src/thread/pal_thread.cpp
```diff
@@ -2911,3 +2911,58 @@ bool IsAddressOnStack(ULONG_PTR address)
 
     return false;
 }
+
+#ifndef __IOS__
+// why _Thread_local ? Because it is faster(.)
+// why not replace PAL to use _Thread_local instead of front caching? _Thread_local is not cross platform
+
+// Why size_t? keeping type for localThread `simple` may affect implementation hence the perf. possitively.
+THREAD_LOCAL size_t localThread = 0;
```
digitalinfinity: since thObjKey is a pointer, would uintptr_t be a better type here?

pal/src/thread/pal_thread.cpp
```diff
@@ -2911,3 +2911,58 @@ bool IsAddressOnStack(ULONG_PTR address)
 
     return false;
 }
+
+#ifndef __IOS__
+// why _Thread_local ? Because it is faster(.)
+// why not replace PAL to use _Thread_local instead of front caching? _Thread_local is not cross platform
+
+// Why size_t? keeping type for localThread `simple` may affect implementation hence the perf. possitively.
```
digitalinfinity: Nit: Positively spelling

pal/src/thread/pal_thread.cpp
```diff
@@ -2911,3 +2911,58 @@ bool IsAddressOnStack(ULONG_PTR address)
...
+        pThread = (size_t) reinterpret_cast<CPalThread*>(pthread_getspecific(thObjKey));
+#ifdef FEATURE_PAL_SXS
+        if (pThread == 0 && force)
+        {
+            pThread = (size_t) CreateCurrentThreadData();
+        }
+#endif
+        localThread = pThread;
+    }
+
```
digitalinfinity: Consider adding a Debug mode assert that if localThread != 0, localThread and pthread_getspecific(thObjKey) return the same value

pal/src/thread/pal_thread.cpp
```diff
@@ -2911,3 +2911,58 @@ bool IsAddressOnStack(ULONG_PTR address)
...
+CPalThread *CorUnix::GetCurrentPalThread(bool force)
+{
+    size_t pThread = localThread;
+    if (pThread == 0)
+    {
+        pThread = (size_t) reinterpret_cast<CPalThread*>(pthread_getspecific(thObjKey));
+#ifdef FEATURE_PAL_SXS
+        if (pThread == 0 && force)
+        {
+            pThread = (size_t) CreateCurrentThreadData();
```
digitalinfinity: Should we just change CreateCurrentThreadData to set localThread instead of calling pthread_getspecific for the non-iOS cases? Then you can get rid of the additional if-check and avoid storing it in two different places

pal/src/thread/pal_thread.cpp
```diff
@@ -2911,3 +2911,58 @@ bool IsAddressOnStack(ULONG_PTR address)
...
+        localThread = pThread;
+    }
+
+    return (CPalThread*)pThread;
+}
+
+CPalThread *CorUnix::InternalGetCurrentThread()
+{
+    size_t pThread = localThread;
+    if (pThread == 0)
```
digitalinfinity: Since GetCurrentPalThread does this check anyway, maybe we should just change the caller of InternalGetCurrentThread to call GetCurrentPalThread directly?

pal/src/thread/pal_thread.cpp
obastemur: But it could be 0 when !force and/or !FEATURE_PAL_SXS

pal/src/thread/pal_thread.cpp
obastemur: That would change the behavior of GetCurrentPalThread.

pal/src/thread/pal_thread.cpp
obastemur: sure, why not. Updated

pal/src/thread/pal_thread.cpp
obastemur: Thanks

pal/src/thread/pal_thread.cpp
obastemur: Makes sense

obastemur:
@digitalinfinity Thanks for the review!

pal/src/thread/pal_thread.cpp
digitalinfinity: Curious- why is that a problem?

obastemur: How we raise a callback on thread exit? That's why.

Issue #2125, opened 2016-11-30
Wasm: Inconsistent representation of signalling NaNs in x86 test


meg-gupta:
The C++ compiler in x86 generates fstp instruction for float return values. This instruction changes the bit pattern of signaling float NaN (For example: from 0x7f80f1e2 to 0x7fc0f1e2). This is problematic in Wasm, as it relies on NaN bit patterns. In x86 debug all signaling NaNs will be canonicalized and we see uniform behavior. In x86 test however, due to optimizations, fstp instruction can get optimized away, and we see different bit patterns for signaling NaNs.

Issue #2049, opened 2016-11-18
WASM: float32 NAN value representation in JIT


meg-gupta:
We use double to store all FloatConstOpnd's in JIT.
While building a NAN float32 constant in IRBuilder we call:
`IR::FloatConstOpnd::New(constVal, TyFloat32, m_func), m_func);`

The function is declared as : 
`FloatConstOpnd *
FloatConstOpnd::New(FloatConstType value, IRType type, Func *func)`

where `FloatConstType` is double

The conversion instruction from float to double will turn a float 0x7fa00000 to 0x7fffc0000000 and then finally the value built will be 0x7fffc000.

This violates some tests in the spec.

arunetm:
This problem change the bits for the nan's when some are stored and read back. not a problem for js as the read back value is still a nan. wasm nan's has a more stricter layout than js nan's making this a problem for wasm. 
Would it make sense to canonicalize wasm nan's? or should we fix the above issue with FloatConstOpnd?
@Cellule @MikeHolman

obastemur:
I had spent some time on this. Actually we may even have a workaround PR fixing an issue on xplat. The bigger problem here is that nan is not a single signed/non-signed value. However internally we convert any `nan` value to a single representation and the rest of the solution behaves based on that constant.

ThomsonTan:
Probably we need some special check before converting `FloatConstantOpnd` to float32 NaN?

Issue #2037, opened 2016-11-17
WASM - handle float NAN literals


meg-gupta:
Add support to handle literals such as nan:0x0f1e2. The following tests for copysign and min/max are failing:

in `testsuite/float_misc.wast`
-noNative only:
(assert_return (invoke "f32.copysign" (f32.const nan:0x0f1e2) (f32.const nan)) (f32.const nan:0x0f1e2))
(assert_return (invoke "f32.copysign" (f32.const nan:0x0f1e2) (f32.const -nan)) (f32.const -nan:0x0f1e2))
(assert_return (invoke "f32.copysign" (f32.const -nan:0x0f1e2) (f32.const nan)) (f32.const nan:0x0f1e2))
(assert_return (invoke "f32.copysign" (f32.const -nan:0x0f1e2) (f32.const -nan)) (f32.const -nan:0x0f1e2)

Both -noNative and forceNative:
(assert_return (invoke "f32.min" (f32.const 0.0) (f32.const nan:0x200000)) (f32.const nan:0x600000))
(assert_return (invoke "f32.min" (f32.const nan:0x200000) (f32.const 0.0)) (f32.const nan:0x600000))
(assert_return (invoke "f32.max" (f32.const nan:0x200000) (f32.const 0.0)) (f32.const nan:0x600000))
(assert_return (invoke "f64.min" (f64.const 0.0) (f64.const nan:0x4000000000000)) (f64.const nan:0xc000000000000))
(assert_return (invoke "f64.min" (f64.const nan:0x4000000000000) (f64.const 0.0)) (f64.const nan:0xc000000000000))
(assert_return (invoke "f64.max" (f64.const 0.0) (f64.const nan:0x4000000000000)) (f64.const nan:0xc000000000000))
(assert_return (invoke "f64.max" (f64.const nan:0x4000000000000) (f64.const 0.0)) (f64.const nan:0xc000000000000)

Cellule:
We have to hold off on this, the problem here can and probably is tooling. Because Wabt generates a compare of 2 floats and `nan != nan`.
The spec interpreter generates a i32 reinterpret than an i32 compare which works.
We might still have problems, but we should wait until we've updated tooling

PR #2779, opened 2017-04-04
Preserve Nan Bits

Krovatkin:
@Cellule @MikeHolman  fyi: @arunetm @meg-gupta @LouisLaf

This is a tentative solution for https://github.com/Microsoft/ChakraCore/issues/2125 

Our understanding of the problem is that both ss->sd (float to double)  and m32->fld (32->80->32) are considered to be conversions and if sNaN are not unmasked by an application, an SNaN is converted to a QNaN

ss->sd conversions occur due to float to double conversions investigated by @meg-gupta in https://github.com/Microsoft/ChakraCore/issues/2125 

m32->fld occur as a result of a calling convention for returning float values on x86-32

A few examples:

0x7fa00000 (SNaN) is converted 0x7fe00000 (QNaN), hardware sets the most significant bit to convert an SNaN to a QNaN

However, 
QNaNs (e.g. 0x7fc60000, 0x7fd00000, 0x7ff00000) survive the roundtrip.  

WebAssembly spec states:

> Except when otherwise specified, when an arithmetic operator returns NaN, there is nondeterminism in determining the specific bits of the NaN. However, wasm does still provide the guarantee that NaN values returned from an operation will not have 1 bits in their fraction field that aren't set in any NaN values in the input operands, except for the most significant bit of the fraction field (which most operators set to 1).

The spec seems to imply that there's no guarantee that the most significant bit will be unchanged, yet they do have a few tests that test this behaviour. 

This change makes sure that the available tests in WasmSpec pass, however, we aren't sure if we need to go down the rabbit hole and `__vectorcall` every function returning floats that Wasm **might** call? 

What do you guys think?


Cellule:
This also addresses #2037 #2049 right ?

Krovatkin:
@Cellule 

The difference between **this** and the other NaN issues is that this one also occurs on **x86-64**. Namely, the absence of a separate float constant operand result in **redundant** conversions when we store float literals into `m_value` of type `double` in `FloatConstOpnd`

Now, the other issues seem to exclusively occur on x86-32 due to
> m32->fld occur as a result of a calling convention for returning float values on x86-32

The part that confuzzles me is that the spec specifically points out 
> "... for the most significant bit of the fraction field (which most operators set to 1)." 


there are no guarantees if it will be set or not. 

Yet, all our NaN mishaps occur because the most significant bit is flipped **on** by hardware for all signalling NaNs and the spec tests are designed to test **all** the bits.

Krovatkin:
@Cellule 

I realized that I didn't  quite answer your original question. Unfortunately, this PR doesn't solve the other NaN issues 😿

Cellule:
I did find a few places where we currently use FloatConstOpnd and some of them probably needs support for Float32ConstOpnd

---

Reviewed 18 of 18 files at r1.
Review status: all files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[lib/Backend/GlobOpt.cpp, line 4569 at r1](https://reviewable.io:443/reviews/microsoft/chakracore/2779#-Kh8ieRR5OefXGAhMKxw:-Kh8ieRR5OefXGAhMKxx:b-cgv2nh) ([raw file](https://github.com/microsoft/chakracore/blob/38a19e5fdda64815c069e0ab54ad3203504b6394/lib/Backend/GlobOpt.cpp#L4569)):*
> ```cpp
>         constant.InitFloatConstValue(srcDef->AsFloatConstOpnd()->m_value);
>     }
>     else if (srcDef->IsFloat32ConstOpnd())
> ```

I'm not sure I understand why this assert is here.
We don't have stElem ldElem instr in asm.js/wasm and thus we shouldn't reach this point.
It would make sense to put a check that we're in asm.js/wasm in OptSrc (also handle that new constOpnd there)

---

*[lib/Backend/IRBuilderAsmJs.cpp, line 781 at r1](https://reviewable.io:443/reviews/microsoft/chakracore/2779#-Kh8j7yGmcsNhpN1SzHv:-Kh8j7yGmcsNhpN1SzHw:bmwms27) ([raw file](https://github.com/microsoft/chakracore/blob/38a19e5fdda64815c069e0ab54ad3203504b6394/lib/Backend/IRBuilderAsmJs.cpp#L781)):*
> ```cpp
>             break;
>         case WAsmJs::FLOAT32:
>             CreateLoadConstInstrForType<float, IR::Float32ConstOpnd>(
> ```

There is also a case in BuildFloat1Const1 that we should use Float32ConstOpnd

---

*[lib/Backend/LowerMDShared.cpp, line 7393 at r1](https://reviewable.io:443/reviews/microsoft/chakracore/2779#-Kh9Q9oHH649bBWEcrqZ:-Kh9Q9oHH649bBWEcrq_:bla4fc3) ([raw file](https://github.com/microsoft/chakracore/blob/38a19e5fdda64815c069e0ab54ad3203504b6394/lib/Backend/LowerMDShared.cpp#L7393)):*
> ```cpp
> LowererMD::LoadFloatValue(IR::Opnd * opndDst, float value, IR::Instr * instrInsert)
> {
>     if (value == 0.0 && !Js::JavascriptNumber::IsNegZero(value))
> ```

I think we should only templatize the other version and keep the same implementation.
Is it really more performant to decode an int32 ? keep in mind that we need to encode the int32 for security purposes.

---

*[lib/Backend/LowerMDShared.cpp, line 7399 at r1](https://reviewable.io:443/reviews/microsoft/chakracore/2779#-Kh9Qbkej_QvJ2Kt-ziP:-Kh9Qbkej_QvJ2Kt-ziQ:b-uc0l6e) ([raw file](https://github.com/microsoft/chakracore/blob/38a19e5fdda64815c069e0ab54ad3203504b6394/lib/Backend/LowerMDShared.cpp#L7399)):*
> ```cpp
>     }
>     int intReprOfFloat = *(reinterpret_cast<int*>(&value));
>     IR::Opnd* intConstOpnd = IR::IntConstOpnd::New(intReprOfFloat, TyInt32, instrInsert->m_func, true);
> ```

If we're to keep this implementation the `true` at the end must be removed because this is a user controlled constant

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/microsoft/chakracore/2779#-:-Kh9TOBP7Y4LK4uJMlhk:bbhvxaa)*

Krovatkin:
> I did find a few places where we currently use FloatConstOpnd and some of them probably needs support for Float32ConstOpnd

I'll double check to see which ones I missed. 

> There is also a case in BuildFloat1Const1 that we should use Float32ConstOpnd

This one? https://github.com/Microsoft/ChakraCore/pull/2779/files#diff-68e402c15675d9f46c381239ffdedcfaL2275

> I'm not sure I understand why this assert is here.
> We don't have stElem ldElem instr in asm.js/wasm and thus we shouldn't reach this point.
> It would make sense to put a check that we're in asm.js/wasm in OptSrc (also handle that new > constOpnd there)

I added the assert for completeness w/ other constant types. The assert should have be saying `Assert(UNREACHABLE)`. I'll remove it. 

> I think we should only templatize the other version and keep the same implementation.
> Is it really more performant to decode an int32 ? keep in mind that we need to encode the int32 for > security purposes.

I'll happily revert to using the templatized version of the original implementation.

Cellule:
> This one? https://github.com/Microsoft/ChakraCore/pull/2779/files#diff-68e402c15675d9f46c381239ffdedcfaL2275

Oops sorry about that, the diff was so small I missed it

Krovatkin:
@Cellule 

> I did find a few places where we currently use FloatConstOpnd and some of them probably needs support for Float32ConstOpnd

Alright, I did a full sweep over all uses of `AsFloatConstOpnd`. I skipped a bunch of uses where constant info is used for bailout book-keeping since we don't have bailouts enabled in asm.js/wasm **yet** and I intentionally decided to not implement value prop for float32 since it would require a brand new class & some plumbing to store float constants and we seem to do very little float constant folding anyways. We could probably enable a special case where float constants are actually integers (i.e. no fractional part), although this behaviour is probably more useful for regular javascript than WASM, since, in WASM, these would probably start out as integers anyways.

Cellule:
:lgtm:

---

Reviewed 2 of 5 files at r2, 6 of 6 files at r3.
Review status: all files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/microsoft/chakracore/2779#-:-KhPGrthyCYNZ61qWaj8:bnfp4nl)*

Krovatkin:
@dotnet-bot test please

Cellule:
@dotnet-bot test please

Cellule:
@Krovatkin I just want to let you know that I started to make fixes on top of your changes.
I haven't got 100% tests passing yet, but I will probably get back on this early next week.

I will close this PR for the time being and I will open a new one including your changes and then some.
Thanks a lot for getting this moving!

Krovatkin:
@Cellule I got us compliant w/ the **spec** versions of the tests on x64.  https://github.com/Krovatkin/ChakraCore/tree/nan2 

Hopefully, you'll find these useful too.

 Also, I did some reading on NaNs on the WebAssembly design repo and I learned that we are supposed to always get the most significant fraction bit right. It goes both ways, some operations like **min/max** require us to  **set** the bit by forcing the conversion from the signalling to quiet NaNs, whereas we should avoid redundant conversions for the other operations like loads and  reinterpret casts.

FYI: @arunetm

Krovatkin:
@Cellule 

Namely, these two commits: 

https://github.com/Krovatkin/ChakraCore/commit/21790fd83ea22b39469d200a1861c721b63258f2
https://github.com/Krovatkin/ChakraCore/commit/400b08b34f74d6ebdeef3bb92768aab02c22c4e7

Cellule:
@Krovatkin Thanks! I wasn't sure if you were still working on this or not.
I'll check with the changes that I got to see how to proceed from there

PR #2781, opened 2017-04-05
Disable randombug.js on CCRobot until it is fixed

obastemur:


jianchun:
I don't understand your 2nd commit -- why do you ignore output of xplatinterval test? On success it is supposed to print "pass" and before change is fine.

obastemur:
It fails on Windows slow runs

obastemur:
> As a note, <baseline/> is mostly only useful for one scenario -- your test wants to verify an ASSERT or AV which crashes and returns non-zero exit code.

Just realized.. on CH we don't set exit code to none 0 when we throw from JS land. Anyone knows why?

obastemur:

PR #2782, opened 2017-04-05
Fix expanding rest args in MapFormals while inlining apply call

meg-gupta:
While expanding rest parameters of an apply call in a nested inlinee,
we have to create StElems for args by reading from argOuts and argOutsExtra
arrays built for the current inlinee.
argOuts consists of args from 0 - formalCount. argOutsExtra consists of
args from formalCount - actualCount.
While creating StElems we were not considering the case when actuals are
less than formals in the inlinee. Fixing this.

### DIFF
--- a/lib/Backend/Inline.cpp
+++ b/lib/Backend/Inline.cpp
@@ -4680,7 +4680,7 @@ Inline::MapFormals(Func *inlinee,
             }
 
             int excess;
-            Js::ArgSlot restFuncFormalCount = 0;
+            uint restFuncFormalCount = 0;
             if (instr->m_func != inlinee)
             {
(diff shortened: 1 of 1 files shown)


lib/Backend/Inline.cpp
```diff
@@ -4713,7 +4713,7 @@ Inline::MapFormals(Func *inlinee,
 
             if (instr->m_func != inlinee)
             {
-                for (uint i = restFuncFormalCount; i < formalCount; ++i)
+                for (uint i = restFuncFormalCount; i < min(actualCount, formalCount); ++i)
```
akroshg: why not do min once outside of the loop instead within the loop?

PR #2783, opened 2017-04-05
let parser detect constant strs used in LdElem and make them PropertyString

MikeHolman:
It's a relatively common pattern to declare some constant string in an outer scope and then use that as a property key inside inner functions.

Perf win here is about 4% in react, with smaller (~1%) wins in backbone and ember.


MikeHolman:
@dotnet-bot test Windows arm_release

MikeHolman:
@dotnet-bot test Ubuntu _no_jit_shared_ubuntu_linux_test

PR #2784, opened 2017-04-05
add profiling for object LdElem; support mixed in Br/CmEq fast paths

MikeHolman:
We weren't profiling LdElem out of an object, which impacts fastpath generation among other things. A lot of these cases have mixed type, so I also added some support for fastpaths of mixed types on Br/CmEq. There were some duplicated code paths in lowering, so I combined them a bit as well.

This is meant to optimize cases like `if(obj1[a] === obj2[a]) ...`

Preliminary perf run showed about a 5% win in flightjs, with a few smaller wins scattered about.

### DIFF
--- a/lib/Backend/LowerMDShared.cpp
+++ b/lib/Backend/LowerMDShared.cpp
@@ -2458,8 +2458,8 @@ LowererMD::GenerateFastBrOrCmString(IR::Instr* instr)
         !srcReg2 ||
         srcReg1->IsTaggedInt() ||
         srcReg2->IsTaggedInt() ||
-        !srcReg1->GetValueType().IsLikelyString() ||
-        !srcReg2->GetValueType().IsLikelyString())
+        !srcReg1->GetValueType().HasHadStringTag() ||
+        !srcReg2->GetValueType().HasHadStringTag())
     {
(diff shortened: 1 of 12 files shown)


lib/Backend/LowerMDShared.cpp
```diff
@@ -2458,8 +2458,8 @@ LowererMD::GenerateFastBrOrCmString(IR::Instr* instr)
         !srcReg2 ||
         srcReg1->IsTaggedInt() ||
         srcReg2->IsTaggedInt() ||
-        !srcReg1->GetValueType().IsLikelyString() ||
-        !srcReg2->GetValueType().IsLikelyString())
+        !srcReg1->GetValueType().HasHadStringTag() ||
+        !srcReg2->GetValueType().HasHadStringTag())
```
Penguinwizzard: Here you use HasHadStringTag, but on arm you use HasBeenString, which is slightly stricter; is this intended?

MikeHolman: Nope, good catch. I did ARM change after my earlier testing and forgot the quirk of this flag. I'll change that to use HasHadStringTag as well.

PR #2790, opened 2017-04-06
Improve JsCreateString et al. perf.

obastemur:
Rather heavy `NarrowStringToWide` follows the slow path no matter if the source string is entirely non unicode.

- pre-allocate memory `sourceCount * sizeof(WCHAR)`
- cast/copy chars from uint8 buffer to WCHAR until hit the first potential unicode
- if there was no unicode, return as is, otherwise continue from the remaining part of the string and use the old ways.

An `empty!` http request ends up at NarrowStringToWide (ASCII) >40.000 times each second.

### DIFF
--- a/lib/Common/Codex/Utf8Helper.h
+++ b/lib/Common/Codex/Utf8Helper.h
@@ -14,7 +14,7 @@ namespace utf8
     /// The returned string is null terminated.
     ///
     template <class Allocator>
-    HRESULT WideStringToNarrow(_In_ LPCWSTR sourceString, size_t sourceCount, _Out_ LPSTR* destStringPtr, _Out_ size_t* destCount)
+    HRESULT WideStringToNarrow(_In_ LPCWSTR sourceString, size_t sourceCount, _Out_ LPSTR* destStringPtr, _Out_ size_t* destCount, size_t* allocateCount = nullptr)
     {
         size_t cchSourceString = sourceCount;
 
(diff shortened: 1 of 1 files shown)


lib/Common/Codex/Utf8Helper.h
```diff
@@ -70,14 +69,44 @@ namespace utf8
...
+            WCHAR *remDestString = destString + sourceStart;
+
+            charcount_t cchDestString = utf8::ByteIndexIntoCharacterIndex(remSourceString, cbSourceString - sourceStart);
+            cchDestString += sourceStart;
+            Assert (cchDestString <= sourceCount);
+
+            // Some node tests depend on the utf8 decoder not swallowing invalid unicode characters
+            // instead of replacing them with the "replacement" chracter. Pass a flag to our
+            // decoder to require such behavior
+            utf8::DecodeUnitsIntoAndNullTerminateNoAdvance(remDestString, remSourceString, (LPCUTF8) sourceString + cbSourceString, DecodeOptions::doAllowInvalidWCHARs);
```
boingoing: I think you could take the return value from `DecodeUnitsIntoAndNullTerminateNoAdvance()` and assert that it equals `cchDestString - sourceStart`.

obastemur:
CI errors due to Java/Jenkins.

obastemur:
@dotnet-bot test OSX please

obastemur:

lib/Common/Codex/Utf8Helper.h
obastemur: We send the end buffer ptr to `DecodeUnitsIntoAndNullTerminateNoAdvance` and there is an assert for `\0` check

obastemur:
CI is not triggering rebasing..

obastemur:
@boingoing Thanks for the review. Yes, we allocate bigger (worst case 2X) `temp` buffer instead of looping over for the second time. Besides, the loop there has small perf advantage over full decode too.

PR #2739, opened 2017-03-24
clear expirableObjectDisposeList at the end of disposal all objects

leirocks:
currently removing one by one from this linked list is slow, might cause hang
remove an uncessary Assert (expirableObjectDisposeList is not null as long as recycler is not null)

PR #2792, opened 2017-04-06
clear expirableObjectDisposeList on DisposeObjects

leirocks:
to avoid linear searching the list to do the removal which can cause busy hang
this is a more completed fix for #2739


curtisman:
:shipit:

PR #2804, opened 2017-04-06
WASM - function execution tracing

Cellule:
Previously we had the option to trace asmjs function with -trace:asmjsfunctionentry.
However this was insufficient when dealing with jit bugs as the tracing didn't work with jitted function.
This PR adds intermediary thunks to wasm function calls to trace their argument and return values.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/2804)


MikeHolman:
:shipit:

PR #2807, opened 2017-04-06
Fix assert in cases of calls to captured function expressions inside 'with'. 

pleath:
Make sure that the function expression scope is not marked for merging with the body scope if the function expression scope will be reified as an ActivationObject. This prevents firing of an assert on useless creation of empty scope slots.


pleath:
@aneeshdk Very simple fix.

aneeshdk:
LGTM

PR #2810, opened 2017-04-07
check loopbody codegen in advance to avoid locking done function body counters

leirocks:
when ETW rundown kicks in the functionBody may have not have bytecode generated yet before registering to utf8SourceInfo
accessing MapLoopHeaders in background thread can causes the functionBody counters locked for updating
and cause assertion when bytecode generation is done and updating bytecodeCount counter on functionBody
so check if the functionBody has done loopbody codegen in advance to not call into Map function in case
loopbody codegen is not done yet


curtisman:
:shipit:

PR #2812, opened 2017-04-07
Invalidate cached scopes in the call stack on Function.caller. 

pleath:
Prevent a cached scope's variables from being cleared out if the function that escapes via Function.caller can access them.


pleath:
@boingoing

boingoing:
LGTM 👍 

Is the code cleanup in parse.cpp around class constructor intentional? Also, is the test added here related to the Function.caller issue?

pleath:
No, the presence of the first 2 commits is due to non-pumping of the Git pump.

PR #2814, opened 2017-04-07
Fix a few Wasm-related leaks.

Penguinwizzard:
Also tighten stuff up so that it's a little harder to make these
errors again later.

### DIFF
--- a/bin/ch/WScriptJsrt.cpp
+++ b/bin/ch/WScriptJsrt.cpp
@@ -907,7 +907,7 @@ JsValueRef __stdcall WScriptJsrt::LoadTextFileCallback(JsValueRef callee, bool i
     }
     else
     {
-        const char *fileContent;
+        const char* fileContent;
         AutoString fileName;
 
         IfJsrtErrorSetGo(fileName.Initialize(arguments[1]));
@@ -927,6 +927,7 @@ JsValueRef __stdcall WScriptJsrt::LoadTextFileCallback(JsValueRef callee, bool i
                 JsValueRef stringObject;
                 IfJsrtErrorSetGo(ChakraRTInterface::JsCreateString(
                     fileContent, lengthBytes, &stringObject));
+                free((void*)fileContent);
                 return stringObject;
             }
         }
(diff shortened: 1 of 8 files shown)


bin/ch/WScriptJsrt.cpp
```diff
@@ -981,6 +982,8 @@ JsValueRef __stdcall WScriptJsrt::LoadBinaryFileCallback(JsValueRef callee,
                         returnValue = arrayBuffer;
                     }
                 }
+ErrorStillFree:
+                HeapFree(GetProcessHeap(), 0, (void*)fileContent);
```
Cellule: Why are we using `free` on one and `HeapFree` on the other ?

Cellule: Can we just make `fileContent` an AutoString ?

lib/Runtime/Library/WebAssemblyModule.cpp
```diff
@@ -234,15 +234,24 @@ WebAssemblyModule::CreateModule(
     }
     catch (Wasm::WasmCompilationException& ex)
     {
-        Wasm::WasmCompilationException newEx = ex;
+        // The reason that we use GetTempErrorMessageRef here, instead of just doing simply
```
Cellule: I think this comment could have just stayed in the pull request hahaha

lib/WasmReader/WasmByteCodeGenerator.h
```diff
@@ -49,18 +49,83 @@ namespace Wasm
...
+        WasmCompilationException(WasmCompilationException&& other)
+        {
+            errorMsg = other.errorMsg;
+            other.errorMsg = nullptr;
+        }
 
-        char16* ReleaseErrorMessage()
+        ~WasmCompilationException()
+        {
+            if(errorMsg != nullptr)
```
Cellule: SysFreeString already does nullptr check

lib/WasmReader/WasmByteCodeGenerator.h
```diff
@@ -49,18 +49,83 @@ namespace Wasm
...
             errorMsg = nullptr;
             return msg;
-        };
+        }
+
+        WasmCompilationException& operator=(WasmCompilationException&& other)
+        {
+            if (this != &other)
+            {
+                if(errorMsg != nullptr)
```
Cellule: Unnecessary check

bin/ch/WScriptJsrt.cpp
Penguinwizzard: > Why are we using free on one and HeapFree on the other ?

Because one is malloc'd and one is HeapAlloc'd. I didn't want to make more changes than immediately necessary.

> Can we just make fileContent an AutoString ?

Not without changing how LoadScriptFromFile works; there's no good way to have an AutoString get externally initialized and set all of the properties correctly (e.g. length).

lib/WasmReader/WasmByteCodeGenerator.h
Penguinwizzard: Removed, thanks.

lib/WasmReader/WasmByteCodeGenerator.h
Penguinwizzard: Removed, thanks.

bin/ch/WScriptJsrt.cpp
Cellule: I see, I guess this is good enough for now

PR #2816, opened 2017-04-07
Install exception filters only once for FaultInjection

leirocks:
this was broken by a change that removing the static local var usage. causes exception filter installing and symInit are called on each allocation, which should causes the faultinjection run very slow

### DIFF
--- a/lib/Common/Core/FaultInjection.cpp
+++ b/lib/Common/Core/FaultInjection.cpp
@@ -850,8 +850,15 @@ namespace Js
     }
 
     static bool faultInjectionDebug = false;
-    bool FaultInjection::InstallExceptionFilters()
+    static bool triedToInstallExceptionFilter = false;
+    void FaultInjection::InstallExceptionFilters()
     {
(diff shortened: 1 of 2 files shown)


MikeHolman:
Nice find, thanks!

lib/Common/Core/FaultInjection.cpp
```diff
@@ -850,8 +850,15 @@ namespace Js
...
-    bool FaultInjection::InstallExceptionFilters()
+    static bool triedToInstallExceptionFilter = false;
+    void FaultInjection::InstallExceptionFilters()
     {
+        if (triedToInstallExceptionFilter)
+        {
+            return;
+        }
+
+        triedToInstallExceptionFilter = true;
```
curtisman: Need to protect against multithread access?

leirocks:
@dotnet-bot test Ubuntu _no_jit_shared_ubuntu_linux_test please

leirocks:
@dotnet-bot test OSX static_osx_osx_debug  please

PR #2821, opened 2017-04-08
Fix cases of bad binding in arrow function parameter lists. 

pleath:
Fix some cases where block enumeration differed between buildAST and !buildAST compiles. If the initial parse of an arrow function parameter was deferred, and the reparse on discovery of the arrow token was (necessarily) not deferred, then the inconsistent state of the pid refs left by the initial parse led to incorrect bindings.

### DIFF
--- a/lib/Parser/Parse.cpp
+++ b/lib/Parser/Parse.cpp
@@ -6509,11 +6509,6 @@ ParseNodePtr Parser::GenerateEmptyConstructor(bool extends)
         (*m_pnestedCount)++;
     }
 
-    if (!buildAST)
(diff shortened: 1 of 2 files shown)


pleath:
@dotnet-bot please(!) test OSX static_osx_osx_debug

pleath:
@aneeshdk (Thanks.)

pleath:
@aneeshdk Will you be able to look at this? Thanks.

akroshg:
:shipit:

aneeshdk:
LGTM

Issue #2681, opened 2017-03-15
ES6 modules not working in some cases


herrernst:
I've found a case where trying to load ES6 modules doesn't work reliable: Most of the time, nothing happens, although the source files are requested from server. Sometimes it does work though, maybe it's some race condition. Here's a test case to reproduce:
http://rawgit.com/herrernst/msedge-es6module-bug/master/index.html
I've tested with the "Microsoft Edge on Win 10 Preview (15.15014)" VM from https://developer.microsoft.com/en-us/microsoft-edge/tools/vms/

akroshg:
adding few folks @boingoing @suwc

suwc:
This is at least partially due to https://github.com/Microsoft/ChakraCore/issues/2702, but there might be other issues.

PR #2822, opened 2017-04-10
ES6module cyclic imports fix and test enhancements

### LINKED ISSUE #2800: ES6module: cyclic imports
Cyclic module imports not function properly:
```javascript
// script.js
WScript.LoadScriptFile('./mod0.js', "module");
```
```javascript
// mod0.js
import './mod1.js';
console.log('mod0');
```
```javascript
// mod1.js
import './mod2.js';
console.log('mod1');
```
```javascript
// mod2.js
import './mod0.js';
console.log('mod2');
```
Expected:
```
>ch -es6module script.js
mod2
mod1
mod0
>
```
Actual:
```
>ch -es6module script.js
>
```

suwc:
ES6module cyclic imports fix and test enhancements

	1) ES6module: cyclic imports (Issue#2800)

(Reference: Issue2681/2702)

For ES6 module support, module import and resolution depends on a counter ("numPendingChildModule") to increment when a child module is identified and decrement when the child module "notifies" its parent module.

This breaks down when there imports forms a cycle, e.g. `mod0 > mod1 > mod2 > mod0` (">" represents left-side module imports right-side module). In this case, the counter stays at one indefinitely and none of these modules involved is instantiated.

Proposed fix is to break the cycle by checking the "wasParsed" flag of the child module, changed from false to true right after the parser has done processing the module code successfully. Once the "wasParsed" flag is set to "true", the child module will either resolve all its dependencies and be ready for instantiation, or find itself/dependencies in error state. Both cases will cause its parent module to be notified and decrement the counter, except when loading fails in the host.

The "readyAsChildModule" flag that indicates parent notification is renamed to "parentsNotified" and retained for assertion purpose.

	2) Unit test framework improvement for module

Cyclic imports were nominally covered in unit tests but did not catch Issue#2800. This is because module-related unit tests use WScript.LoadModule() that failed silently when modules fail to instantiate. We need an update on unit test framework to collect and tally results from tests that are launched asynchronously.

Proposed changes to the unit test framework augments testRunner APIs by 1) adding tracking of asynchronously launched code in runTest(); 2) adding testRunner.LoadModule() as a promise-friendly wrapper for Wscript.LoadModule(). Module tests therefore go through the extra step of checking asynchronously launched code, which have to complete with no error in order for the tests to pass.

	3) ModuleNamespace initialization triggers singleton instance assertion (Issue #2840)

Instances of ModuleNamespace share the same type object moduleNamespaceType in JavascriptLibrary. Need to mark the same as shared to avoid singleton instance logic in type handler classes.
	
	4) Tracing for module

This changeset also includes tracing and assertions for module.

Fixes #2800, #2840 
Ref #2681, #2702


suwc:
Address https://github.com/Microsoft/ChakraCore/issues/2800 and more.
@boingoing @akroshg @tcare @Microsoft/chakra-es could you review please?

test/UnitTestFramework/UnitTestFramework.js
```diff
@@ -12,12 +12,14 @@
 //     }
 //   How to define and run tests:
 //     var tests = [ { name: "test name 1", body: function () {} }, ...];
-//     testRunner.run(tests);
```
dilijev: >testRunner.run(tests); [](start = 7, length = 22)

.run(tests) still works, right? Probably don't need to remove this.

test/UnitTestFramework/UnitTestFramework.js
```diff
@@ -12,12 +12,14 @@
 //     }
 //   How to define and run tests:
 //     var tests = [ { name: "test name 1", body: function () {} }, ...];
-//     testRunner.run(tests);
+//     testRunner.runTests(tests);       // regular tests
+//     testRunner.runAsyncTests(tests);  // tests loaded with WScript.LoadModule
```
dilijev: >testRunner.runAsyncTests(tests);  // tests loaded with WScript.LoadModule [](start = 7, length = 73)

Can this in theory be used with async patterns besides module loading? If it's module-only, should this be named differently?

test/es6/module-syntax.js
```diff
@@ -8,7 +8,7 @@
 WScript.LoadScriptFile("..\\UnitTestFramework\\UnitTestFramework.js");
 
 function testModuleScript(source, message, shouldFail) {
```
dilijev: >shouldFail [](start = 43, length = 10)

relying on default value = undefined? could explicitly have `shouldFail = false` instead to be more explicit about this.

test/UnitTestFramework/UnitTestFramework.js
```diff
@@ -167,8 +191,120 @@ var testRunner = function testRunner() {
...
+
+        asyncTestEnd : function asyncTestEnd(testIndex, testCount) {
+            _asyncMode = false;
+            asyncTest.resolve[testIndex][testCount]();
+        },
+
+        async : function async(source, shouldFail) {
+            var testIndex = asyncTest.testIndex;
+            if (!shouldFail) {
+                var testCount = asyncTest.count[testIndex]++;
```
dilijev: >var testCount = asyncTest.count[testIndex]++; [](start = 16, length = 45)

Trying to get count before or after increment? Should testIndex ever be the same? I don't quite understand the intent of this pattern

dilijev:
> var4=undefined

Is it expected that the output of this baseline file could be different even if all tests pass? If not, could this test be an `-args summary -endargs` test?
Considering the last line of the test file is `testRunner.runAsyncTests(tests, { verbose: WScript.Arguments[0] != "summary" });` is there a test config which runs this test in summary mode? Was it planned to run in summary mode?

---
Refers to: test/es6/module-namespace.baseline:18 in 4c08174. [](commit_id = 4c08174edb82f5dbeb96431ebdf0509e78d38de7, deletion_comment = False)

dilijev:
:shipit:

test/UnitTestFramework/UnitTestFramework.js
```diff
@@ -167,8 +191,120 @@ var testRunner = function testRunner() {
...
+                var promise = new Promise((resolve, reject) => {
+                    asyncTest.resolve[testIndex].push(resolve);
+                });
+                asyncTest.promise[testIndex].push(promise);
+                return 'testRunner.asyncTestBegin(' + testIndex + ", " + testCount + ');' + source + ";\ntestRunner.asyncTestEnd(" + testIndex + ", " + testCount + ");";
```
boingoing: String template would also be pretty nice here. 👍 

But actually, would it make more sense to have a helper here instead of requiring consumers of this API call `WScript.LoadModule(testRunner.async(...))` and munging up the module source code? Could be `testRunner.runModule(source, shouldFail)` which just loads/runs the module inline, returns a promise which does that, or returns a lambda which would do the same? Not sure I understand why we need to edit the module code itself here since that seems more complicated for new tests.

test/UnitTestFramework/UnitTestFramework.js
```diff
@@ -167,8 +191,120 @@ var testRunner = function testRunner() {
...
+                var message = ex.stack || ex.message || ex;
+                logTestNameIf(!_verbose);
+                helpers.writeln("Test threw exception: ", message);
+                isSuccess = false;
+            }
+            if (isSuccess) {
+                if (_verbose) {
+                    helpers.writeln("LOADED");
+                }
+                ++passedTestCount;
```
suwc: It is in fact loadedSuccessCount, but reusing the same variable. Maybe we should add a new variable to make it more readable.

---
In reply to: [110760516](https://github.com/Microsoft/ChakraCore/pull/2822#discussion_r110760516) [](ancestors = 110760516)

suwc:
> var4=undefined

I think a lot of the outputs in this tests do need to be checked against the baseline. I will see if they can be converted into asserts and run in summary mode.
Note: summary mode with runAsyncTests() works the same way as run()

---
In reply to: [293070315](https://github.com/Microsoft/ChakraCore/pull/2822#issuecomment-293070315) [](ancestors = 293070315)

---
Refers to: test/es6/module-namespace.baseline:18 in 4c08174. [](commit_id = 4c08174edb82f5dbeb96431ebdf0509e78d38de7, deletion_comment = False)

lib/Runtime/Language/SourceTextModuleRecord.h
```diff
@@ -47,15 +47,18 @@ namespace Js
 
         void SetSpecifier(Var specifier) { this->normalizedSpecifier = specifier; }
         Var GetSpecifier() const { return normalizedSpecifier; }
+        const char16 *GetSpecifierSz() const { return JavascriptString::FromVar(this->normalizedSpecifier)->GetSz(); }
```
suwc: Currently it is used in debug-only traces, but it can be used in test-build traces too.

---
In reply to: [110757106](https://github.com/Microsoft/ChakraCore/pull/2822#discussion_r110757106) [](ancestors = 110757106)

test/UnitTestFramework/UnitTestFramework.js
suwc: An unique testIndex is assigned to each item in the 'testsToRun' array. Each testIndex is associated with a 'name' and 'body', etc.
Inside each item, there can be multiple calls that spawn asynchronous execution, each will be assigned a promise and tracked by a separate counter, "testCount".
Every piece of code loaded asynchronously must know its testIndex and testCount to resolve the correct promise when the execution completes.

---
In reply to: [110755539](https://github.com/Microsoft/ChakraCore/pull/2822#discussion_r110755539) [](ancestors = 110755539)

test/UnitTestFramework/UnitTestFramework.js
suwc: Yes, this can be used with any code that is executed asynchronously and not module-only.

---
In reply to: [110754143](https://github.com/Microsoft/ChakraCore/pull/2822#discussion_r110754143) [](ancestors = 110754143)

test/UnitTestFramework/UnitTestFramework.js
suwc: Here is why we add stuff to the source, remind me if I am missing something:
WScript.LoadModule() loads and parses module code synchronously but executes it asynchronously. Ideally we will need it to return a promise and resolve/reject it later depending on the parsing and the execution. Essentially making it a full-fledged async API. This unit test framework change tries to bridge the gap by handling the promise part for APIs like this that are 1) asynchronous in nature, and 2) not tracking its execution through promise. At this point, only the loaded code that knows exactly when it ends can resolve the promise.


---
In reply to: [110764912](https://github.com/Microsoft/ChakraCore/pull/2822#discussion_r110764912) [](ancestors = 110764912)

boingoing: Yes, I suppose that's fair. When it's complete, we can replace `WScript.LoadModule()` with dynamic `import` statements and chain promises off of them which could be used to settle the test case.

test/UnitTestFramework/UnitTestFramework.js
```diff
@@ -167,8 +191,120 @@ var testRunner = function testRunner() {
...
+                    var passed = testRunner.asyncTestHasNoErr(i);
+                    if (_verbose) {
+                        helpers.writeln("*** Result of test #", i + 1, " (", i, "): ", testsToRun[i].name);
+                        if (!passed) helpers.writeln(asyncTest.error[i]);
+                        helpers.writeln(passed ? "PASSED" : "FAILED");
+                    }
+                    !passed || passedTestCount++;
+                }
+                if (_verbose || executedTestCount != passedTestCount) {
+                    helpers.writeln("Summary of tests: total executed: ", executedTestCount,
```
dilijev: (Side note: any reason not to use a string template when it seems appropriate? I try to use them whenever it makes sense. Are there any performance concerns?)

suwc: There's no particular reason other than consistency with the older code.

---
In reply to: [110987334](https://github.com/Microsoft/ChakraCore/pull/2822#discussion_r110987334) [](ancestors = 110987334)

akroshg:
:shipit:

test/UnitTestFramework/UnitTestFramework.js
suwc: In the new rev: merging runAsyncTests and runAsyncTest back into runTests and runTest.


---
In reply to: [110792175](https://github.com/Microsoft/ChakraCore/pull/2822#discussion_r110792175) [](ancestors = 110792175,110760516)

suwc:
> var4=undefined

Next iteration: converting this test to run in summary mode.

---
In reply to: [293116299](https://github.com/Microsoft/ChakraCore/pull/2822#issuecomment-293116299) [](ancestors = 293116299,293070315)

---
Refers to: test/es6/module-namespace.baseline:18 in 4c08174. [](commit_id = 4c08174edb82f5dbeb96431ebdf0509e78d38de7, deletion_comment = False)

suwc:
@pleath let me know if you agree with the change in `JavascriptLibrary.cpp` to address #2840. Thanks!

PR #2824, opened 2017-04-11
resolve ophelper assert; remove MD paths for FastBrOrCmString

MikeHolman:
I added check in GenerateFastStringCheck to have a direct branch to success label if opnds are equal. however helper blocks are still needed e.g. in case string test fails, and they may still have branch to failure label.

This may lead to failure label being accessible only from helper blocks. Failure label may correspond to a large sequence of blocks. We don't have any convenient way to flip all of that over to helper, and this hit in an artificial test case (not real scenario), so to resolve the issue I set `m_noHelperAssert = true` for that label.

There were very similar code paths in ARM and x86/x64 for the fastpath in question, but I didn't see any need to have the fastpath be MD, so I moved it out to Lowerer.


rajatd:
:clock1:

MikeHolman:

PR #2825, opened 2017-04-11
avoid updating auxPtrs while shutting down

leirocks:
while shutting down there's no need to update the auxPtrs and counters.
and a couple code cleanup fix.


curtisman:
:shipit:

PR #2826, opened 2017-04-12
Merge from unreleased/rs2 to release/1.4

leirocks:
This includes all internal fixes in windows RS2


lib/Common/Memory/CustomHeap.cpp
```diff
@@ -621,6 +619,13 @@ bool Heap<TAlloc, TPreReservedAlloc>::AllocInPage(Page* page, size_t bytes, usho
         return false;
     }
 
+    //Section of the Page should already be freed.
```
curtisman: Why this is duplicated?

leirocks: it's a bad merge on 9/2016...

---
In reply to: [111493100](https://github.com/Microsoft/ChakraCore/pull/2826#discussion_r111493100) [](ancestors = 111493100)

PR #2829, opened 2017-04-12
Jsrt: External Data support for all the Javascript Objects

obastemur:
Previously, Js Set/Get/Has ExternalData API were supporting API for External
Object only. This `ChakraCore only` change, alters their functionality to serve
for both JavascriptObject and ExternalObject.

In order to accomplish this, JsSetExternalDataWithCallback is added.

#### Why
ChakraCore dependent app. might specialize a JS object with additional properties
without changing its base/parent. Confirming the object with previously defined
specializations is expensive. ChakraCore's use scenario may require external data
support for native land created objects while not all the objects created in native
land require the external data support. Creating all the objects from an ExternalObject 
base type by default and/or wrapping them is `memory` expensive. Apart from native
land objects, JS land objects might also need / use external data.

Besides the extra use of memory, and complications in implementations, application
will end up checking if a proxy object wraps an object with external data. Not knowing
the JS land object is wrapped by a proxy, we end up checking each and every object
whether they were wrapped this way. This naturally brings a `performance` drop.

#### Remarks / Discussion
This PR updates the behavior of existing Jsrt API for ChakraCore target only.
Since, the API is intact, behavior change may also take a place in Chakra as well ?


obastemur:
@dotnet-bot test Windows ci_disablejit_x64_debug please

liminzhu:
Apology for a late response.

Trying to understand the premise - today you can attach a native pointer to a JS object via creating externalObject and setting that as a property on the JS object. With the change, you no longer have to deal with property lookup and all that, and both attaching the native pointer and accessing the pointer natively become cheaper. Is that correct? This seems useful in cases where a host is storing some metadata and doesn't care about exposing things to JS land. Since this is mostly about perf, do you think it would be helpful to do some prototyping in node-chakracore and confirm wins there? 

One feedback is that I'm not a big fan of introducing different semantics for APIs in Chakra and ChakraCore. And since this is a breaking change (e.g. for some reason I was using JsHasExternalData to tell me whether something is an external obj or not), I'm not sure about just changing the behavior in Chakra as well. Maybe having a new set of APIs is the better option.

obastemur:
@liminzhu The problem with existing model is given on the description of this PR and having an API that creates a new ExternalObject doesn't solve that.

Also the `wins` part explained. All these and previous change proposals are based on perf measurements.

Please be more specific if description on PR doesn't address your questions.

About API; IMHO, I believe the API does what its name says for the first time. `JsHasExternalData`, now, answers whether and object has external data or not regardless from its type. We have other facilities to identify the JsValue type. Same argument for Get and Set.

liminzhu:
> Also the wins part explained. All these and previous change proposals are based on perf measurements.

Understand where the wins would come from. I was thinking along the lines of whether we can quantify speed/memory wins in AcmeAir or other workflows.

> About API; IMHO, I believe the API does what its name says for the first time. JsHasExternalData, now, answers whether and object has external data or not regardless from its type. We have other facilities to identify the JsValue type. Same argument for Get and Set.

Totally agree with you that `JsHasExternalData` does what it says on the tin. The point I'm trying to make is that the current semantics in Chakra separate the idea of JS and external objects. If we make this change as-is in ChakraCore only, then we have different semantics for the same APIs in ChakraCore and Chakra. If we port the change to Chakra, it'd be a change that breaks Windows APIs' backward compatibility, which we shouldn't. Hence the concern.

obastemur:
> If we port the change to Chakra, it'd be a change that breaks Windows APIs' backward compatibility, which we shouldn't. Hence the concern.

See your point. My argument; the behavior of API changes with the language itself anyways. Aren't we supposed to keep ABI intact for Windows API. I'm not sure if there is a way to keep behavior intact over time. Maybe I'm wrong?

For performance; I believe there are more to do there. Not rushing on landing this though.

obastemur:
Adding `JavascriptExternalFunction` i.e. `externalData` considering it separately inside the new API helps reducing ~50% of the expensive calls.

@curtisman @digitalinfinity @jianchun @liminzhu see original proposal (and gains) [above](https://github.com/Microsoft/ChakraCore/pull/2829#issue-221255749) and extension below;

Previously, Js Set/Get/Has ExternalData API were providing external data API for External
Object only. This (and upcoming couple of more) ChakraCore only changes, alter
their functionality to serve for `JavascriptObject`, `ExternalObject` and `JavascriptExternalFunction`.

`JavascriptExternalFunction` has enough space (prior to alignment) to introduce `externalData`. So, for both `ExternalObject` and `JavascriptExternalFunction` we don't need to use Get/Set property internally. Introducing `JavascriptExternalFunction` this way helps reducing number of Get/Set calls ~50% (http load). 

Thoughts?

obastemur:
#3639 brings full external data support. closing this one.

Issue #2823, opened 2017-04-11
Nullptr assertion error/crash in GetOwnPropertyDescriptor with revoked proxy


anba:
Test case:
```js
var {proxy, revoke} = Proxy.revocable({a: 0}, new Proxy({}, {
    get(t, pk) {
        if (pk === "getOwnPropertyDescriptor")
            revoke();
    }
}));
Object.keys(proxy);
```

```gdb
0x00007ffff3b80ba5 in Js::JavascriptOperators::GetOwnPropertyDescriptor (obj=0x0, propertyId=739, scriptContext=0x55555589f288, propertyDescriptor=0x7fffffff89c8)
    at /home/andre/git/ChakraCore/lib/Runtime/Language/JavascriptOperators.cpp:1227
1227	        Assert(obj);
(gdb) bt
#0  0x00007ffff3b80ba5 in Js::JavascriptOperators::GetOwnPropertyDescriptor (obj=0x0, propertyId=739, scriptContext=0x55555589f288, propertyDescriptor=0x7fffffff89c8)
    at /home/andre/git/ChakraCore/lib/Runtime/Language/JavascriptOperators.cpp:1227
#1  0x00007ffff3ea0de6 in Js::JavascriptProxy::GetOwnPropertyDescriptor(Js::RecyclableObject*, int, Js::ScriptContext*, Js::PropertyDescriptor*)::$_16::operator()() const (this=0x7fffffff8800)
    at /home/andre/git/ChakraCore/lib/Runtime/Library/JavascriptProxy.cpp:1656
[...]

PR #2830, opened 2017-04-12
Proxy got revoked when in the trap.

akroshg:
During the GetOwnPropertyDescriptor trap we were looking for the getOwnPropertyHandler, which actually causes the revoke of the proxy, though we were using the proxy->target instead of using the targetObj we have cached locally.
Fixed that.


akroshg:
Fixes #2823

Cellule:
LGTM

PR #2831, opened 2017-04-13
ch: add console.error

obastemur:
I manually add something like this to separate outputs to the file and console. If it was already there, that could save some time.

Simply, having stderr output capabilities is helpful.


obastemur:
Errors due to DebuggerCommon baselines. Have another PR updating those files, will keep this PR until the other one is merged in order not to deal with rebasing the baselines.

Cellule:
@obastemur You might want to take a look at #2447
I had to do something similar regarding baseline and determined the right solution was to exclude console from the DbgController.js filterList. 
I didn't merge it because Wasm is still not ready to be turned on by Default, but I probably should have extracted the part for the debugger baselines
I think you might need to update the full repository as well I think we have some debugger baselines there

Cellule:
Otherwise this PR looks good to me

akroshg:
I have recently added to dbgcontroller.js, you can have a look that way no baseline adjustment needed.

obastemur:
@Cellule Thanks for the review! 

 @akroshg @Cellule Thanks for the tips! 

> I had to do something similar regarding baseline and determined the right solution was to exclude console from the DbgController.js filterList.

Yeah, but sending this PR to release/2.0 even if I remove `Console` from the filter list, I would have to endup updating the baselines. :/

MikeHolman:
any reason this needs to be in release/2.0 rather than master?

dilijev:
Spoke with @obastemur offline, since this PR needs more work and is not a high priority, adding the `Waiting` label so we know not to worry about this PR being merged in any particular time frame.

obastemur:
No point keeping this around.

PR #2832, opened 2017-04-13
Fixes to support node inspector

kfarnung:
* Added support for "forceSetValueProp" which forces the raw value to
  be added to the diagnostics object.
* Handled the possibility that although a context may exist, it may
  not be fully initialized.

### DIFF
--- a/bin/ch/Debugger.cpp
+++ b/bin/ch/Debugger.cpp
@@ -174,7 +174,7 @@ JsValueRef Debugger::Evaluate(JsValueRef callee, bool isConstructCall, JsValueRe
     if (argumentCount > 2)
     {
         IfJsErrorFailLogAndRet(ChakraRTInterface::JsNumberToInt(arguments[1], &stackFrameIndex));
-        ChakraRTInterface::JsDiagEvaluate(arguments[2], stackFrameIndex, JsParseScriptAttributeNone, &result);
+        ChakraRTInterface::JsDiagEvaluate(arguments[2], stackFrameIndex, JsParseScriptAttributeNone, /* forceSetValueProp */ false, &result);
     }
 
     return result;
--- a/lib/Jsrt/JsrtDebugUtils.cpp
+++ b/lib/Jsrt/JsrtDebugUtils.cpp
@@ -122,7 +122,7 @@ void JsrtDebugUtils::AddVarPropertyToObject(Js::DynamicObject * object, const ch
     }
 }
 
-void JsrtDebugUtils::AddPropertyType(Js::DynamicObject * object, Js::IDiagObjectModelDisplay* objectDisplayRef, Js::ScriptContext * scriptContext)
+void JsrtDebugUtils::AddPropertyType(Js::DynamicObject * object, Js::IDiagObjectModelDisplay* objectDisplayRef, Js::ScriptContext * scriptContext, bool forceSetValueProp)
 {
     Assert(objectDisplayRef != nullptr);
     Assert(scriptContext != nullptr);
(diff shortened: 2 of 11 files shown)


kfarnung:
@dotnet-bot test Ubuntu Ubuntu_linux_debug_static please

bin/ch/Debugger.cpp
```diff
@@ -174,7 +174,7 @@ JsValueRef Debugger::Evaluate(JsValueRef callee, bool isConstructCall, JsValueRe
     if (argumentCount > 2)
     {
         IfJsErrorFailLogAndRet(ChakraRTInterface::JsNumberToInt(arguments[1], &stackFrameIndex));
-        ChakraRTInterface::JsDiagEvaluate(arguments[2], stackFrameIndex, JsParseScriptAttributeNone, &result);
+        ChakraRTInterface::JsDiagEvaluate(arguments[2], stackFrameIndex, JsParseScriptAttributeNone, false, &result);
```
digitalinfinity: Nit: for boolean params, consider adding the param name next it it in a comment to improve readability
Eg. false /* forceSetValueProp */

lib/Jsrt/JsrtDebugUtils.cpp
```diff
@@ -389,6 +394,16 @@ void JsrtDebugUtils::AddPropertyToObject(Js::DynamicObject * object, JsrtDebugPr
...
 }
 
+bool JsrtDebugUtils::HasProperty(Js::DynamicObject * object, JsrtDebugPropertyId propertyId, Js::ScriptContext * scriptContext)
+{
+    const char16* propertyName = GetDebugPropertyName(propertyId);
+
+    const Js::PropertyRecord* propertyRecord;
+    scriptContext->GetOrAddPropertyRecord(propertyName, static_cast<int>(wcslen(propertyName)), &propertyRecord);
+
+    return !!Js::JavascriptOperators::HasProperty(object, propertyRecord->GetPropertyId());
```
digitalinfinity: Why do you need the !! here? Isn't it sufficient to just return the result of HasProperty?

digitalinfinity: Also, you probably don't need GetOrAddPropertyRecord here- if FindPropertyRecord returns null, then you could return false here. There's no need to do another allocation here.

---
In reply to: [111523435](https://github.com/Microsoft/ChakraCore/pull/2832#discussion_r111523435) [](ancestors = 111523435)

digitalinfinity:
cc @liminzhu - a JSRT API has changed here, FYI

lib/Runtime/Base/ScriptContext.cpp
```diff
@@ -1271,6 +1272,13 @@ namespace Js
 
         this->GetDebugContext()->GetProbeContainer()->Initialize(this);
 
+        isDebugContextInitialized = true;
+
+#if defined(_M_ARM32_OR_ARM64)
+        // We need to ensure that the above write to the isDebugContextInitialized is visible to the debugger thread.
+        MemoryBarrier();
```
obastemur: Do we really need this here?

lib/Jsrt/JsrtDebugUtils.cpp
kfarnung: I was getting warnings about conversion from `BOOL` (`typedef int`) to `bool`.  I didn't see any other usages of `BOOL` in this file, so I wanted to keep the signature as is.  Is there a better way to "BOOLIFY" this?

lib/Runtime/Base/ScriptContext.cpp
kfarnung: This was the outcome of a discussion with @digitalinfinity, the concern was that setting a Boolean value was not guaranteed to be committed to memory on ARM.

liminzhu:
`forceSetValueProp` adds a value property on the returned json object? Why not always add value prop?

kfarnung:
@liminzhu I didn't want to change the existing behavior where complex types are never returned by value.  It changes the JSON payload and may break callers.  Since we are going to be removing the legacy node debug code anyway, it probably isn't a big deal.  This is basically equivalent to the `returnByValue` parameter from the [inspector API](https://chromedevtools.github.io/debugger-protocol-viewer/v8/Runtime/#method-evaluate) surface.

liminzhu:
@kfarnung makes sense. Is there a better name for the param? `forceSetValueProp` is really confusing when I first see it.

kfarnung:
@liminzhu I agree with that.  I had a hard time coming up with a proper name for it.  I went back and forth between `returnByValue` and `forceSetValueProp`, but wasn't sure which was clearer.  Do you have any suggestions?

kfarnung:
@liminzhu Let's plan to land this as-is and then we can discuss name changes afterwards.  Once the CI is clean I'll push.

kfarnung:

kfarnung:

kfarnung:

kfarnung:

PR #2835, opened 2017-04-13
Add Dev15 MSBuild to MSBuild search path

tcare:
Search Dev15 Enterprise path for MSBuild.

It should search other variants and perhaps use amd64 if appropriate, but I don't have the time to fix that right now. :)


dilijev:
Shall we combine efforts? :P https://github.com/Microsoft/ChakraCore/pull/2570

PR #2838, opened 2017-04-14
Improve StrictEqual

obastemur:
#### 1 -     Improve String::StrictEqual

Neither `PAL_wmemcmp` / `Windows wmemcmp` checks for `ptr1 == ptr2` hence we end up with the worst case scenario / perf.

#### 2 -     StrictEqual overall improvement

CC calls `JavascriptOperators::StrictEqual`with final calculated values. So, when ptr1 == ptr2 and typeOf (ptr*) != TypeIds_Number, we may not need to go further for strict equal check. The importance with the `TypeIds_Number` check is due to `NaN !== NaN`

#### Why

mt. http req. triggers a strict check for ptr1 === ptr2 aprox. 200.000 times each second.

### DIFF
--- a/lib/Runtime/Language/JavascriptOperators.cpp
+++ b/lib/Runtime/Language/JavascriptOperators.cpp
@@ -879,8 +879,13 @@ namespace Js
     BOOL JavascriptOperators::StrictEqual(Var aLeft, Var aRight, ScriptContext* requestContext)
     {
         double dblLeft, dblRight;
-        TypeId leftType = JavascriptOperators::GetTypeId(aLeft);
-        TypeId rightType = JavascriptOperators::GetTypeId(aRight);
+        TypeId rightType, leftType;
+        leftType = JavascriptOperators::GetTypeId(aLeft);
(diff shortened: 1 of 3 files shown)


lib/Runtime/Language/JavascriptOperators.cpp
```diff
@@ -882,6 +882,9 @@ namespace Js
         TypeId leftType = JavascriptOperators::GetTypeId(aLeft);
         TypeId rightType = JavascriptOperators::GetTypeId(aRight);
 
+        // Because NaN !== NaN, we may not return TRUE when typeId is Number
+        if (aLeft == aRight && leftType != TypeIds_Number) return TRUE;
```
MikeHolman: you can move this even higher up, above def of rightType

obastemur: makes sense. will do.

obastemur:
@bterlson @MikeHolman Thanks for the review!

PR #2640, opened 2017-03-06
Integer indexed exotic object behavior

suwc:
Implement integer indexed exotic object behaviors (9.4.5) in typed arrays:

1. Replace HasProperty(), GetProperty(), HasItem(), GetPropertyReference(), GetItem(), GetItemReference() with -Query version with tri-state return value
   TRUE  => Property_Found (1)
   FALSE => Property_NotFound (0)
         => Property_NotFound_NoProto (2) Property not found but stop going up prototype chain

2. Add new DescriptorFlags (None_NoProto) to stop going up prototype chain for
   GetSetter(), GetItemSetter()

Fixes #374

PR #2845, opened 2017-04-17
Array.prototype.IndexOf is now a builtin function made with JS instead of C++

veler:
- Remove the internal function that corresponds to Array.prototype.IndexOf
- Replace it by a builtin function in JS that match with the spec.


lib/Runtime/Library/BuiltIn/IndexOf.js
```diff
@@ -0,0 +1,48 @@
...
+
+(function indexOf(searchElement, fromIndex) {
+
+    var k;
+
+    if (this == null) {
+        throw new TypeError('"this" is null or not defined');
+    }
+
+    var o = this;
```
MikeHolman: I think if this is a number it might not work. Maybe want to check if typeof this isn't an array then create Object, otherwise use this.

lib/Runtime/Library/BuiltIn/IndexOf.js
```diff
@@ -0,0 +1,48 @@
+//-------------------------------------------------------------------------------------------------------
+// Copyright (C) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE.txt file in the project root for full license information.
+//-------------------------------------------------------------------------------------------------------
+
+"use strict";
+
+(function indexOf(searchElement, fromIndex) {
```
MikeHolman: This is only for array, right? Maybe worth renaming to arrayIndexOf

lib/Runtime/Library/BuiltIn/IndexOf.js
```diff
@@ -0,0 +1,48 @@
...
+    } else {
+        k = len - Math.abs(n);
+    }
+
+    if (k < 0) {
+        k = 0;
+    }
+
+    while (k < len) {
+        if (o[k] === searchElement && (o[k] !== undefined || searchElement !== undefined)) {
```
MikeHolman: If you are searching for undefined I think you still have to skip holes in the array.

lib/Runtime/Library/JavascriptLibrary.cpp
```diff
@@ -1680,6 +1726,9 @@ namespace Js
 
         ScriptContext* scriptContext = arrayPrototype->GetScriptContext();
         JavascriptLibrary* library = arrayPrototype->GetLibrary();
+
+        library->EnsureIndexOfByteCode();
```
MikeHolman: This is ok now, but we may want to make more generic when we start adding more builtins.

veler: It will be one of the part of my mission if we want to use this builtin system. For the moment Michael and Louis told me to don't "really" care about it.

---
In reply to: [111827172](https://github.com/Microsoft/ChakraCore/pull/2845#discussion_r111827172) [](ancestors = 111827172)

lib/Runtime/Library/BuiltIn/IndexOf.js
veler: Yes it does not work, it throws a TypeError. The spec does not precise if it should work for a number (for example or any object different from array or string). Under Google Chrome, the behavior is the same : it throws a TypeError. I'm not sure that we really should do this check.

---
In reply to: [111824987](https://github.com/Microsoft/ChakraCore/pull/2845#discussion_r111824987) [](ancestors = 111824987)

lib/Runtime/Library/BuiltIn/IndexOf.js
dilijev: Do we not have tests for arrays with holes already?

lib/Common/ConfigFlagsList.h
```diff
@@ -501,6 +501,7 @@ PHASE(All)
 
 #define DEFAULT_CONFIG_Intl                    (true)
 #define DEFAULT_CONFIG_IntlBuiltIns            (true)
+#define DEFAULT_CONFIG_BuiltIn         (true)
```
dilijev: nit: spacing

lib/Runtime/Base/FunctionBody.cpp
```diff
@@ -4354,7 +4354,7 @@ namespace Js
             AssertMsg(false, "Non-matching start of document");
         }
 
-        GetLineCharOffsetFromStartChar(cchStartOffset, &line, &col, false /*canAllocateLineCache*/);
+        GetLineCharOffsetFromStartChar(cchStartOffset, &line, &col, false /*canAllocateLineCache*/, true /*canDumpLibraryCode*/);
```
dilijev: you should probably pass `canDumpLibraryCode` directly instead of hardcoding the true value here. In release builds, Asserts are compiled out and won't hit. We want to ensure we still get the intended (and reasonable) behavior even if the Assert didn't cause execution to stop.

lib/Runtime/Library/Chakra.Runtime.Library.vcxproj
```diff
@@ -170,6 +170,8 @@
     <ClInclude Include="..\RuntimeCommon.h" />
     <ClInclude Include="..\SerializableFunctionFields.h" />
     <ClInclude Include="AtomicsObject.h" />
+    <ClInclude Include="BuiltIn\IndexOf.js.bc.32b.h" />
+    <ClInclude Include="BuiltIn\IndexOf.js.bc.64b.h" />
```
dilijev: What about the NoJIT bytecodes?

lib/Runtime/Library/Chakra.Runtime.Library.vcxproj.filters
```diff
@@ -252,6 +252,8 @@
     <ClInclude Include="WebAssemblyTable.h" />
     <ClInclude Include="WebAssemblyEnvironment.h" />
     <ClInclude Include="WabtInterface.h" />
+    <ClInclude Include="BuiltIn\IndexOf.js.bc.32b.h" />
+    <ClInclude Include="BuiltIn\IndexOf.js.bc.64b.h" />
```
dilijev: Again, what about the NoJIT bytecodes?

lib/Runtime/Library/EngineInterfaceObject.cpp
```diff
@@ -192,8 +192,11 @@ namespace Js
     {
         typeHandler->Convert(commonNativeInterfaces, mode, 38);
 
-        ScriptContext* scriptContext = commonNativeInterfaces->GetScriptContext();
-        JavascriptLibrary* library = scriptContext->GetLibrary();
+        JavascriptLibrary* library = commonNativeInterfaces->GetScriptContext()->GetLibrary();
+
+        library->EnsureIndexOfByteCode();
+
+        library->DefaultCreateFunction(library->indexOfByteCode->GetNestedFunctionForExecution(0), 1, commonNativeInterfaces, Js::PropertyIds::builtInJavascriptArrayEntryIndexOf);
```
dilijev: > library->indexOfByteCode->GetNestedFunctionForExecution(0)

Is it going to be a common pattern to hardcode the offset like this? After a certain number of functions I think this would be difficult to maintain.

lib/Runtime/Library/JavascriptLibrary.h
```diff
@@ -413,6 +413,9 @@ namespace Js
 
         Field(ConstructorCache*) builtInConstructorCache;
 
+
```
dilijev: nit: unnecessary newline

test/DebuggerCommon/ES6_proto_invalidation.js.dbg.baseline
```diff
@@ -461,10 +461,9 @@
             },
             "indexOf": {
               "#__proto__": "function <large string>",
-              "length": "number 1",
+              "prototype": "Object {...}",
```
dilijev: This change in behavior is a bit concerning. Can you double check the value of the prototype before and after your change?

test/es6/es6toLength.js
```diff
@@ -37,15 +37,15 @@ var tests = [
             assert.areEqual(-1,a.indexOf(3, 4294967290), "indexOf on an array is bound by array max length");
             assert.areEqual(-1,a.indexOf(4, 4294967290), "indexOf on an array is bound by array max length");
 
-            assert.areEqual(4294967294,Array.prototype.indexOf.call(o32, 2, 4294967290), "confirm we can find value 2 in the index range given");
-            assert.areEqual(4294967295,Array.prototype.indexOf.call(o32, 3, 4294967290), "objects don't have a bounded length so we should find the index given any length");
+            assert.areEqual(-1,Array.prototype.indexOf.call(o32, 2, 4294967290), "confirm we can find value 2 in the index range given");
+            assert.areEqual(-1,Array.prototype.indexOf.call(o32, 3, 4294967290), "objects don't have a bounded length so we should find the index given any length");
```
dilijev: The changes to this test look like a change in behavior. Can you explain what happened here?

test/Intl/DateTimeFormatOptions.js
```diff
@@ -51,7 +51,7 @@ var tests = [
                 }
                 catch (e) {
                     if (!(e instanceof RangeError)) {
-                        fail("Incorrect exception was thrown.");
+                        assert.fail("Incorrect exception was thrown.");
```
dilijev: @veler I guess this is the issue you were talking about? Good catch. I guess we didn't hit this path before so nothing complained that the `fail` function was missing. What behavior changed that exposed this path for you?

lib/Runtime/Library/BuiltIn/IndexOf.js
```diff
@@ -0,0 +1,48 @@
...
+
+    var k;
+
+    if (this == null) {
+        throw new TypeError('"this" is null or not defined');
+    }
+
+    var o = this;
+
+    var len = o.length >>> 0;
```
dilijev: Isn't `>>>0` the same as `|0` that you use below? If so, stick to a consistent notation :P If it's not the same, what's the difference?
Is this actually useful information (i.e. do we use this type information) outside of asm.js?

lib/Runtime/Library/BuiltIn/IndexOf.js
```diff
@@ -0,0 +1,48 @@
...
+        return -1;
+    }
+
+    var n = fromIndex | 0;
+
+    if (n >= len) {
+        return -1;
+    }
+
+    if (n >= 0) {
```
dilijev: probably change to `else if` since we're checking the same var

lib/Runtime/Library/BuiltIn/IndexOf.js
```diff
@@ -0,0 +1,48 @@
...
+
+    var o = this;
+
+    var len = o.length >>> 0;
+
+    if (o.length <= 0) {
+        return -1;
+    }
+
+    var n = fromIndex | 0;
```
dilijev: nit: prefer `let`

lib/Runtime/Library/BuiltIn/IndexOf.js
```diff
@@ -0,0 +1,48 @@
...
+
+    while (k < len) {
+        if (o[k] === searchElement && (o[k] !== undefined || searchElement !== undefined)) {
+            return k;
+        }
+        k++;
+    }
+
+    return -1;
+});
```
dilijev: nit: newline at end of file

lib/Runtime/Library/BuiltIn/IndexOf.js
dilijev: nit: prefer `let`

lib/Runtime/Library/BuiltIn/IndexOf.js
```diff
@@ -0,0 +1,48 @@
+//-------------------------------------------------------------------------------------------------------
+// Copyright (C) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE.txt file in the project root for full license information.
+//-------------------------------------------------------------------------------------------------------
+
+"use strict";
+
+(function indexOf(searchElement, fromIndex) {
+
+    var k;
```
dilijev: nit: to make things more clear, declare `k` right before you use it (and prefer `let` over `var`)

lib/Runtime/Library/BuiltIn/IndexOf.js
veler: We don't. I just fixed the bug and added a unit test. Please see "test/Array/array_indexOf.js".

---
In reply to: [111840475](https://github.com/Microsoft/ChakraCore/pull/2845#discussion_r111840475) [](ancestors = 111840475)

test/es6/es6toLength.js
veler: Those unit test were not complaining with the specs. Here we are trying to get an item from an id > int32.Max. That is why we use o.length >>> 0; It is to shift o.length bits to the right, discarding bits shifted off, and shifting in zeroes from the left.

test/Intl/DateTimeFormatOptions.js
veler: I don't remember exactly. It was a bug linked to my changes and it was failing on this test, but not only. I fixed the problem and everything was working fine again. There was no direct link with this file but I considered that it was good to fix this line by the way.

lib/Runtime/Library/BuiltIn/IndexOf.js
veler: That is why We use o.length >>> 0; It is to shift o.length bits to the right, discarding bits shifted off, and shifting in zeroes from the left. It is asked by the spec.

veler: Also, fromIndex | 0 check whether fromIndex if defined and set before assigning the value. If it is not, the default value will be 0 instead of undefined for example.

lib/Runtime/Library/BuiltIn/IndexOf.js
MikeHolman: By work I meant be spec compliant. Spec stays to convert this with ToObject, which creates new Number object in case argument is a number, so I think we should not throw. It also mentions that "indexOf function is intentionally generic; it does not require that its this value be an Array object".

http://www.ecma-international.org/ecma-262/7.0/index.html#sec-array.prototype.indexof

http://www.ecma-international.org/ecma-262/7.0/index.html#sec-toobject

However, our current implementation as well as Firefox and Chrome all throw type errors, so it's possible I'm misunderstanding something. Maybe @bterlson can help clarify what the correct behavior should be?

lib/Common/ConfigFlagsList.h
```diff
@@ -960,6 +961,7 @@ FLAGNR(Boolean, CollectGarbage        , "Enable CollectGarbage API", DEFAULT_CON
 FLAGR (Boolean, Intl                  , "Intl object support", DEFAULT_CONFIG_Intl)
 FLAGNR(Boolean, IntlBuiltIns          , "Intl built-in function support", DEFAULT_CONFIG_IntlBuiltIns)
 
+FLAGR (Boolean, BuiltIn               , "Built-in function support", DEFAULT_CONFIG_BuiltIn)
```
MikeHolman: Looks like the config flag is unused? If so, please remove it.

lib/Runtime/Library/BuiltIn/IndexOf.js
veler: After chatting with @dilijev , we cannot use | 0 it doesn't work with an unsigned integer.

lib/Runtime/Library/BuiltIn/IndexOf.js
dilijev: well, maybe this is clearer separately like this, because before you were bounds checking, and here you're checking whether n is positive or negative

---
In reply to: [111842688](https://github.com/Microsoft/ChakraCore/pull/2845#discussion_r111842688) [](ancestors = 111842688)

lib/Runtime/Library/BuiltIn/IndexOf.js
MikeHolman: `>>>0` is not the same as `|0`. `>>>0` does a coercion `ToUint32`, whereas `|0` does `ToInt32`. For `indexOf`, we need to perform `ToUint32` on the length property.

lib/Runtime/Library/BuiltIn/IndexOf.js
dilijev: >| 0 [](start = 22, length = 3)

This should be >>>0 as well to force UInt32 because it could be as large as length and still be valid.

lib/Runtime/Library/BuiltIn/IndexOf.js
dilijev: Ah, I see now. The implementation actually defines the behavior of this operator should do that conversion. I was focused on how that's sort of a true thing for asm.js but it shouldn't be observable in the result which would be a JS Number. And since Arrays have a max size of UInt32, that coercion (and not ToInt32) is needed.

Learned something today :P

---
In reply to: [111851902](https://github.com/Microsoft/ChakraCore/pull/2845#discussion_r111851902) [](ancestors = 111851902)

lib/Runtime/Library/BuiltIn/IndexOf.js
```diff
@@ -0,0 +1,48 @@
...
+
+    var o = this;
+
+    var len = o.length >>> 0;
+
+    if (o.length <= 0) {
+        return -1;
+    }
+
+    var n = fromIndex | 0;
```
dilijev: > | 0 [](start = 21, length = 4)

Also use >>>0 here because fromIndex can be as long as the array and still needs to be handled correctly (also as a UInt32).

lib/Runtime/Library/BuiltIn/IndexOf.js
```diff
@@ -0,0 +1,48 @@
...
+
+    if (n >= len) {
+        return -1;
+    }
+
+    if (n >= 0) {
```
dilijev: >n >= 0 [](start = 8, length = 6)

Note -0 and +0 are both === 0.
The spec notes that when n is -0 k should be set to +0. If you do (fromIndex >>> 0) above, -0 will be coerced to +0, so you don't need to do anything here.

test/es6/es6toLength.js
MikeHolman: Actually, I think we were looking at an old version of the spec. The unit test is correct.

dilijev: The old version was correct or the change is correct?

lib/Runtime/Library/BuiltIn/IndexOf.js
dilijev: Just a note, from testing it seems that no annotation is fine here, `|0` would be wrong, and `>>>0` is correct but not strictly necessary, except for potentially providing type information.

dilijev: > Edit: actually, that's not true anymore. It used to be that we need to call ToUint32, now we need to do ToLength, so our coercing ToUint32 isn't valid.

It seems the difference is that any negative value for lenValue must become 0, but since an array cannot produce a lenValue < 0 it doesn't matter. Additionally, the higher upper range for length is not necessary to consider because Array length is capped by the spec at 2^32-1 anyway.

IOW, <<<0 will do the right thing even though the semantics in the spec have changed.

lib/Runtime/Library/BuiltIn/IndexOf.js
```diff
@@ -38,7 +38,7 @@
     }
 
     while (k < len) {
-        if (o[k] === searchElement && (o[k] !== undefined || searchElement !== undefined)) {
+        if (o[k] === searchElement && (o[k] !== undefined || (searchElement === undefined && o.hasOwnProperty(k)))) {
```
MikeHolman: I think you need to say `k in o`. Likely now we are going to incorrectly skip case where `o.prototype[k]` is `undefined`.

dilijev: ^ oh that's tricky. I was going to say that for arrays, holes in the set of own properties are just undefined and the prototype is not checked. But then I remembered that that weirdness is only true if the indexes are outside the length of the array or are non-integer numbers (like 3.14). I think. See https://github.com/Microsoft/ChakraCore/pull/2640 and the linked issue it resolves.

dilijev: Oh then again that weirdness is only TypedArrays. Nevermind.

MikeHolman:
You might also want to try running Array.prototype.indexOf tests in [Test262](https://github.com/tc39/test262/tree/master/test/built-ins/Array/prototype/indexOf) (I haven't done this before, but I think there is a test harness that will work with ChakraCore).

lib/Runtime/Library/BuiltIn/IndexOf.js
MikeHolman: You are allowed to pass non-arrays as the `this` object, so I think we must still consider the case where length > 2^32.

lib/Runtime/Library/BuiltIn/IndexOf.js
veler: It returns `-1` actually.
We removed the ToObject because it was regressing the performances.

lib/Runtime/Library/EngineInterfaceObject.cpp
veler: As it is a prototype right now, it's OK to hardcode.

lib/Runtime/Base/FunctionBody.cpp
veler: As it is a prototype right now, it's OK to hardcode. Also `canDumpLibraryCode` is an optional argument, by default equals to `false`.

test/es6/es6toLength.js
veler: For me the change is correct.

lib/Runtime/Library/BuiltIn/IndexOf.js
veler: Sooo, I changed `k in o` for performances reason. If you remember, with this expression we had a regression of 20% on Splay. As @dilijev said, that weirdness is only true if the indexes are outside the length of the array or are non-integer numbers. So the behavior is similar to `k in o` and is fast, except in some improbable weird case like searching for undefined in an almost empty array where the length is defined manually.

lib/Common/ConfigFlagsList.h
veler: It is used by `RegenAllByteCode.cmd` to generate the bytecode for IndexOf.js

dilijev:
Yep, @MikeHolman makes a good point. You can use @bterlson's [test262-harness](https://github.com/bterlson/test262-harness) with a `ch.exe` binary to run the tests.

lib/Runtime/Library/BuiltIn/IndexOf.js
dilijev: It's worth writing a test case like @MikeHolman said. I think `k in o` might be the only correct way to get `undefined` at indexes in the prototype which would otherwise look like holes.

lib/Runtime/Library/BuiltIn/IndexOf.js
dilijev: Like by doing `Array.prototype.bind(nonArrayObject)`? I see.

lib/Runtime/Base/FunctionBody.cpp
```diff
@@ -3018,9 +3018,9 @@ namespace Js
         return this->GetLineCharOffsetFromStartChar(startCharOfStatement, _line, _charOffset, canAllocateLineCache);
     }
 
-    bool FunctionBody::GetLineCharOffsetFromStartChar(int startCharOfStatement, ULONG* _line, LONG* _charOffset, bool canAllocateLineCache /*= true*/)
+    bool FunctionBody::GetLineCharOffsetFromStartChar(int startCharOfStatement, ULONG* _line, LONG* _charOffset, bool canAllocateLineCache /*= true*/, bool canDumpLibraryCode /*= false*/)
     {
-        Assert(!this->GetUtf8SourceInfo()->GetIsLibraryCode());
+        Assert(canDumpLibraryCode || !this->GetUtf8SourceInfo()->GetIsLibraryCode());
```
dilijev: Thought exercise: in release builds, Asserts are compiled out and won't hit. We want to ensure we still get the intended (and reasonable) behavior even if the Assert didn't cause execution to stop.

test/DebuggerCommon/ES6_proto_invalidation.js.dbg.baseline
```diff
@@ -740,10 +739,9 @@
             },
             "indexOf": {
               "#__proto__": "function <large string>",
-              "length": "number 1",
-              "name": "string indexOf",
-              "caller": "object null",
-              "arguments": "object null"
+              "prototype": "Object {...}",
+              "name": "string arrayIndexOf",
+              "length": "number 1"
```
dilijev: I think my comment got buried, but this change looks like a change in observable semantics. We want to make sure that that js implementation is not observably different from the native implementation we had before. Some of these behaviors are by spec, like the name being "arrayIndexOf" instead of "indexOf".

In order to manipulate this you might actually need to follow the patterns toward the bottom of Intl.js for adding functions to the various Intl objects.

PR #2846, opened 2017-04-17
Update BP & debugger code load handling, misc bugfixes.

mrkmarron:
- Refactored debugger interaction code into seperate file.
- Only notify debugger of first load of a source file.
- Ensure all sources have unique id (shared with debugger) even when re-created.
- Breakpoints are loaded incrementally. 
- Updated reverse continue logic.

PR #2847, opened 2017-04-17
Optimize log representation & improve weak ref/external lifetime handling.

mrkmarron:
- Improve log representation memory usage.
- Fix weak set/map handling to remove memory leak in recording.
- Improve handling of long-lived external reference tracking in log (addref and makeweak).
- Includes changes from PR #2846 (will remove after that merges).


mrkmarron:
Closing and retargeting to release/2.0 in PR #2853.

PR #2849, opened 2017-04-19
fix couple bugs with ShiftAdd peep

MikeHolman:
Don't peep case where both sources to add use the same temp -- because the shl/mul instr will still be needed.

SCCLiveness::FoldIndir didn't handle case where base is null.

### DIFF
--- a/lib/Backend/PreLowerPeeps.cpp
+++ b/lib/Backend/PreLowerPeeps.cpp
@@ -82,8 +82,8 @@ Lowerer::TryShiftAdd(IR::Instr *instrAdd, IR::Opnd * opndFold, IR::Opnd * opndAd
         switch (constVal)
         {
         case 1:
-            scale = 0;
-            break;
+scale = 0;
+break;
         case 2:
             scale = 1;
             break;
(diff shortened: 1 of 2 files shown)


lib/Backend/PreLowerPeeps.cpp
```diff
@@ -82,8 +82,8 @@ Lowerer::TryShiftAdd(IR::Instr *instrAdd, IR::Opnd * opndFold, IR::Opnd * opndAd
         switch (constVal)
         {
         case 1:
-            scale = 0;
-            break;
+scale = 0;
```
Cellule: nit: could you revert this ?

lib/Backend/SccLiveness.cpp
```diff
@@ -781,30 +781,36 @@ SCCLiveness::FoldIndir(IR::Instr *instr, IR::Opnd *opnd)
         {
             return false;
         }
-        indir->SetOffset(offset);
+        int32 newOffset = 0;
+        if(!Int32Math::Add(offset,indir->GetOffset(), &newOffset))
```
Cellule: I don't understand this part.
The check above already does add the indir offset and the scaled index.
This added check returns false if an additional add doesn't overflow... And if it does overflow, use it ?

MikeHolman: i agree, this makes no sense

---
In reply to: [116345511](https://github.com/Microsoft/ChakraCore/pull/2849#discussion_r116345511) [](ancestors = 116345511)

PR #2850, opened 2017-04-19
add coercion to uint for IntConstOpnd during const folding in wasm

MikeHolman:
This coercion shouldn't have impact on release build, but causes assert to trigger on debug builds.


MikeHolman:
No problem. I feel the assert may be a bit too conservative since we will truncate back to 32 bit before using the value anyway, but I guess it could catch issues where someone was using the wrong type.
Anyway, thanks for the review!

PR #2852, opened 2017-04-19
fix: Visual Studio build on master

obastemur:
ChakraCore master doesn't compile with VisualStudio IDE due to missing dependency link to wabt. If you have it running somehow, it's because it might have compiled previously or you compile the whole solution all the time.


obastemur:
@MikeHolman @dilijev Thanks for the review

PR #2854, opened 2017-04-20
xplat: cover additional edge cases for interval caching

obastemur:
- test update: Opt out the case for system time backward jumps.
- improve implementation: CPU tick value is circular. Make sure we don't miss updating `now`

### DIFF
--- a/lib/Runtime/PlatformAgnostic/Platform/Common/HiResTimer.cpp
+++ b/lib/Runtime/PlatformAgnostic/Platform/Common/HiResTimer.cpp
@@ -64,28 +64,33 @@ namespace DateTime
                                             utc_tm.tm_mday - 1, milliseconds);
     }
 
+// Important! When you update 5ms below to any other number, also update test/Date/xplatInterval.js 0->5 range
 #define INTERVAL_FOR_TICK_BACKUP 5
     double HiResTimer::GetSystemTime()
     {
         ULONGLONG current = GetTickCount64();
         ULONGLONG diff = current - data.cacheTick;
 
-        if (diff >= INTERVAL_FOR_TICK_BACKUP) // max *ms to respond system time changes
+        if (diff <= data.previousDifference || diff >= INTERVAL_FOR_TICK_BACKUP) // max *ms to respond system time changes
         {
             double currentTime = GetSystemTimeREAL();
 
             // in case the system time wasn't updated backwards, and cache is still beyond...
             if (currentTime >= data.cacheSysTime && currentTime < data.cacheSysTime + INTERVAL_FOR_TICK_BACKUP)
             {
+                data.previousDifference = (ULONGLONG) -1; // Make sure next request won't use cache
                 return data.cacheSysTime + INTERVAL_FOR_TICK_BACKUP; // wait for real time
             }
 
+            data.previousDifference = 0;
             data.cacheSysTime = currentTime;
             data.cacheTick = current;
 
             return currentTime;
         }
 
+        // tick count is circular. So, make sure tick wasn't cycled since the last request
+        data.previousDifference = diff;
         return data.cacheSysTime + (double)diff;
     }
 #undef INTERVAL_FOR_TICK_BACKUP
--- a/test/Date/xplatInterval.js
+++ b/test/Date/xplatInterval.js
@@ -8,13 +8,20 @@ if (WScript.Platform && WScript.Platform.OS != "win32") {
     function rand() {
         return parseInt(Math.random() * 1e2) + 50;
     }
+
     for (var j = 0; j < 1e2; j++) {
         var pre_time = Date.now(), now;
         for(var i = 0; i < 1e6; i++) {
             now = Date.now();
             var diff = now - pre_time
 
-            if (diff < 0) throw new Error ("Timer interval has failed. diff < 0");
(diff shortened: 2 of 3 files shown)


lib/Runtime/PlatformAgnostic/Platform/Common/HiResTimer.cpp
```diff
@@ -64,28 +64,33 @@ namespace DateTime
...
 
-        if (diff >= INTERVAL_FOR_TICK_BACKUP) // max *ms to respond system time changes
+        if (diff <= data.previousDifference || diff >= INTERVAL_FOR_TICK_BACKUP) // max *ms to respond system time changes
         {
             double currentTime = GetSystemTimeREAL();
 
             // in case the system time wasn't updated backwards, and cache is still beyond...
             if (currentTime >= data.cacheSysTime && currentTime < data.cacheSysTime + INTERVAL_FOR_TICK_BACKUP)
             {
+                data.previousDifference = /*ULONGLONG*/-1; // Make sure next request won't use cache
```
jianchun: >/*ULONGLONG*/-1 [](start = 42, length = 15)

`previousDifference` type is actually `int`, why not ULONGLONG?

obastemur: I'm also surprised to see `int` there. Updated.

lib/Runtime/PlatformAgnostic/Platform/Common/HiResTimer.cpp
```diff
@@ -64,28 +64,33 @@ namespace DateTime
...
 
-        if (diff >= INTERVAL_FOR_TICK_BACKUP) // max *ms to respond system time changes
+        if (diff <= data.previousDifference || diff >= INTERVAL_FOR_TICK_BACKUP) // max *ms to respond system time changes
         {
             double currentTime = GetSystemTimeREAL();
 
             // in case the system time wasn't updated backwards, and cache is still beyond...
             if (currentTime >= data.cacheSysTime && currentTime < data.cacheSysTime + INTERVAL_FOR_TICK_BACKUP)
             {
+                data.previousDifference = (ULONGLONG) -1; // Make sure next request won't use cache
```
jianchun: >// Make sure next request won't use cache [](start = 58, length = 41)

Just to understand this, what's the case that this time it didn't use cache but next time it uses cache? Is it because of tick cycle?

test/Date/xplatInterval.js
```diff
@@ -8,13 +8,20 @@ if (WScript.Platform && WScript.Platform.OS != "win32") {
...
+
     for (var j = 0; j < 1e2; j++) {
         var pre_time = Date.now(), now;
         for(var i = 0; i < 1e6; i++) {
             now = Date.now();
             var diff = now - pre_time
 
-            if (diff < 0) throw new Error ("Timer interval has failed. diff < 0");
+            // INTERVAL_FOR_TICK_BACKUP = 5
+            // So, anything beyond 5ms is not subject to our testing here.
```
jianchun: What is the exact reason for current test failure? System time jumped back 5+ ms?

What if system jump back 5- ms?

lib/Runtime/PlatformAgnostic/Platform/Common/HiResTimer.cpp
obastemur: we don't know when `next time` is. This method can be called much later right? so, previous state was -> `cache is expired` so, whatever the upcoming `tickCount` is, we don't care and update the real time.

test/Date/xplatInterval.js
obastemur: The reason is parallel python jobs under tzset env. We do not control real system time. The one fails here the real system time backwards. We do not cache or control that.

obastemur: in case system time jumps back <= 5 ms, we most likely recover that internally.

jianchun: > in case system time jumps back <= 5 ms, we most likely recover that internally. (due to our cache is 5ms)

How does it recover? Suppose last time it returned `cacheSysTime + 2` using cache. Now 3 ticks later, system time jumps to `cacheSysTime - 1`. It'll return from line 89 and fail the test (now - prev == -3).

(Just try to understand this test stability.)

lib/Runtime/PlatformAgnostic/Platform/Common/HiResTimer.cpp
jianchun: Sorry, I was trying to understand the purpose of adding `previousDifference`. What error scenario exactly does it try to fix?

obastemur: We previously didn't cover circularity. We cover now + use same veritable for forced recache

test/Date/xplatInterval.js
obastemur: This scenario makes sense but isn't system timer granularity is somewhere around 10ms? I'm not saying we can't see that test is failing but we most likely don't and I prefer to see that test is failing on different environments instead of discarding the comparison whenever the system timer backwards.

jianchun:
:shipit:

obastemur:

PR #2855, opened 2017-04-20
Add missing cases to unsigned cmp peep

meg-gupta:


lib/Backend/FlowGraph.cpp
```diff
@@ -3072,6 +3072,10 @@ bool FlowGraph::UnsignedCmpPeep(IR::Instr *cmpInstr)
     case Js::OpCode::BrNeq_A:
     case Js::OpCode::BrSrEq_A:
     case Js::OpCode::BrSrNeq_A:
+    case Js::OpCode::CmEq_A:
+    case Js::OpCode::CmNeq_A:
+    case Js::OpCode::CmSrEq_A:
```
MikeHolman: makes sense. maybe worth adding the CmLe_A, CmGt_A etc. as well?

meg-gupta: There are handled in the far end of the switch https://github.com/Microsoft/ChakraCore/pull/2855/files#diff-ff2373c80d7d67870bf34ea643081a7eL3088

MikeHolman: Ah I missed that. LGTM

PR #2865, opened 2017-04-22
Fix failure to reparse children of redeferred functions. 

pleath:
On debug attach, we reparse the tree of compiled functions for debug mode, but sub-trees rooted at redeferred functions will not be reparsed, and this can leave functions still compiled for non-debug mode. Fix this by detecting redeferred functions, compiling them for debug mode, and continuing to reparse the sub-tree. Functions that have never been fully parsed will still be skipped, as they are guaranteed to have no fully parsed children.


rajatd:
:shipit:

PR #2868, opened 2017-04-24
some pageheap fixes

leirocks:
- fix that pageheap allocation didn't count in heuristic, which causes very high memory usage
and often make the program not usable

- while page reusing is disabled, locked the object page as soon as it's swept to increase the
chance of catching memory corruption

- some performance improvement with IsAll instead of looping byte by byte while doing verification

- move allocation close to guard page when guard page is at end of allocation, to increase chance
of catching buffer overrun

- move pageheap metadata to seperate structure, so we don't allocate the memory for it while not
in pageheap mode

### DIFF
--- a/lib/Common/Memory/ArenaAllocator.cpp
+++ b/lib/Common/Memory/ArenaAllocator.cpp
@@ -1134,10 +1134,10 @@ void InlineCacheAllocator::CheckIsAllZero(bool lockdown)
             // will be debug pattern filled (specifically, at least their weak reference slots).
             // All other caches must be zeroed out (again, at least their weak reference slots).
 #ifdef ARENA_MEMORY_VERIFY
-            Assert(IsAll((char*)weakRefBytes, sizeof(cache->weakRefs), '\0') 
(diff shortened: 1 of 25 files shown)


lib/Common/Memory/ArenaAllocator.cpp
```diff
@@ -1477,20 +1477,20 @@ void InlineCacheAllocator::CheckIsAllZero()
...
-        Assert(IsAll(blockp->GetBytes(), blockp->currentByte, '\0'));
+        Assert((byte*)IsAll(blockp->GetBytes(), blockp->currentByte, 0));
         blockp = blockp->nextBigBlock;
     }
 
     ArenaMemoryBlock * memoryBlock = this->mallocBlocks;
     while (memoryBlock != nullptr)
     {
-        Assert(IsAll(memoryBlock->GetBytes(), memoryBlock->nbytes, '\0'));
+        Assert((byte*)IsAll(memoryBlock->GetBytes(), memoryBlock->nbytes, 0'));
```
curtisman: huh?

PR #2871, opened 2017-04-25
Fix fgPeepTmp annotation and re-enable an optimization that uses it

Penguinwizzard:
Marking the result load for fgPeepTmp as an fgPeepTmp load allows us to correctly handle this case.


dilijev:
@Penguinwizzard are you planning to merge this?

Penguinwizzard:
I'll make a new PR when it's time for that.

PR #2876, opened 2017-04-26
xplat: Support Valgrind

### LINKED ISSUE #2873: Applications using libchakracore crash in Valgrind with OOM
Regardless of the application, the stack trace at termination always looks like this, when running an application that is linked with ChakraCore using valgrind:

```
Out of Memory
==10928== 
==10928== Process terminating with default action of signal 6 (SIGABRT)
==10928==    at 0x61943F8: raise (in /lib64/libc-2.23.so)
==10928==    by 0x6195FF9: abort (in /lib64/libc-2.23.so)
==10928==    by 0x4EC9DB: Memory::X64WriteBarrierCardTableManager::Initialize()
==10928==    by 0x42100C: _GLOBAL__sub_I_RecyclerWriteBarrierManager.cpp 
==10928==    by 0xBB9D1C: __libc_csu_init
==10928==    by 0x617F75E: (below main) (in /lib64/libc-2.23.so)

```

This is happening on a Linux x64 system with 32G of ram.

Some of the exact addresses may vary, depending on which branch of ChakraCore I link to, but the result is always the same.  Even a very tiny application that does nothing more than read user input and execute the javascript in a REPL will crash on startup when running it with valgrind.  The applications do not appear to use very much memory otherwise, and operate as intended outside of Valgrind, but I do require a method of using a tool such as this for programs that I write.

I've been troubleshooting things on my end as much as I know how, but it appears to me that the problem lies in Chakacore's memory usage at startup.

obastemur:
Valgrind doesn't support (by default) > 32GB address space

Fixes #2873

### DIFF
--- a/lib/Common/Memory/RecyclerWriteBarrierManager.cpp
+++ b/lib/Common/Memory/RecyclerWriteBarrierManager.cpp
@@ -268,13 +268,30 @@ X64WriteBarrierCardTableManager::Initialize()
         }
         const unsigned __int64 maxUmProcessAddressSpace = (__int64) memoryLimit;
 
-        _cardTableNumEntries = Math::Align<size_t>(maxUmProcessAddressSpace / AutoSystemInfo::PageSize, AutoSystemInfo::PageSize) /* s_writeBarrierPageSize */;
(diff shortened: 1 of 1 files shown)


obastemur:
Will send another PR. Wrong branch.

PR #2860, opened 2017-04-21
xplat: add package creator

obastemur:

PR #2880, opened 2017-04-27
xplat: Valgrind Support

### LINKED ISSUE #2873: Applications using libchakracore crash in Valgrind with OOM

obastemur:
adds `--valgrind` parameter to build arguments. `--valgrind` build disables concurrent GC support.

Valgrind supports max 32GB address space (3.12 forward 64GB) yet WriteBarrier card table reserves all the process address space and expects the control over. We might limit the address space for mmap etc. to 32GB yet we don't have control over the stack allocations. Inconsistency in between breaks the card table.

The intention with this new parameter is to give ChakraCore users a way to use Valgrind on their applications. Fixes #2873 . Since this is an initial support, we may have issues (with Valgrind) that we may need to address going forward.


lib/Common/CommonDefines.h
```diff
@@ -140,7 +140,11 @@
 #define ENABLE_JS_ETW                               // ETW support
 #else
 #define SYSINFO_IMAGE_BASE_AVAILABLE 0
+#ifndef ENABLE_VALGRIND
 #define ENABLE_CONCURRENT_GC 1
+#else
+#define ENABLE_CONCURRENT_GC 0
```
jianchun: nit: Don't need this (no define to 0 elsewhere for this symbol)

jianchun:
:shipit:

lib/Common/CommonDefines.h
obastemur: most of the places check whether its value is 1 instead of if it is defined. Made this way for consistency

jianchun: Check 5 lines above. We don't have a "define to 0" on _WIN32. Hence it is inconsistency.

obastemur: Not sure if it was the places you were pointing but it defines to 1 there. see also line 152 -> `#if ENABLE_CONCURRENT_GC`

leirocks: 32 GB card table is only for x64, x86 it's only 1 MB card table. so you don't need to disable concurrent gc for x86

leirocks: also, if Valgrind supports max 32GB address space, we can change it to  32GB/4096 bytes to be reserved for card table. so you can change AutoSystemInfo to specify max possible address range and still keep concurrent GC enabled

jianchun: @obastemur Yes, I was saying "no matching define to 0" elsewhere.
@leirocks Good point

obastemur: X86 argument makes sense and will apply. However as given from the description 32GB approach doesn't work because we don't have control over stack ptr addresses. Will start an offline discussion on the second one. We might have a bug with write barrier.

digitalinfinity: For the stack addresses, we do two approaches to mitigate this- we try to map in the write barrier cards corresponding to the script threads stack up front, and we also try to be disciplined about not setting barriers on writes to stack addresses (they're unnecessary). Still, @obastemur I think you're right that the 32GB approach might not work if under valgrind doesn't guarantee that the memory range is contiguous and starts from 0 (if it is contiguous and we know the start range, we can make the smaller card table idea work)

obastemur: > I think you're right that the 32GB approach might not work if under valgrind doesn't guarantee that the memory range is contiguous and starts from 0

@digitalinfinity thanks for the details there. IIRC, process stack is part of user mode space and accommodated under .text section hence it won't start from 0. I was (initially) interested implementing a smaller card table idea. However the inconsistency among the stack and process address space require changes to card table indexing approach. Besides, we will have to limit stack space too.

This PR should just work. I don't think user will need Concurrent GC for Valgrind. Only pain here is to maintain `--valgrind` flag on a host project.

obastemur:
@jianchun Thanks for the review! @leirocks @digitalinfinity Thanks for the discussion.

x86 condition is applied. Closing this PR and will cherry pick the commits into #2860 in order not to rebase / conflict fix multiple times.

PR #2883, opened 2017-04-28
Use PolymorphicInlineCache for PropertyString

MikeHolman:
PropertyStrings were earlier using their own special kind of inline caches. I have changed for them to now use our PolymorphicInlineCache, which can improve cases where we have polymorphic types., and also take advantage of existing InlineCache optimizations that weren't present in the specialized PropertyString cache.

This gives about 8% improvement in react and 4% in angular.


lib/Runtime/Language/InlineCache.h
```diff
@@ -540,6 +539,65 @@ namespace Js
 #endif
     };
 
+
+    class FunctionBodyPolymorphicInlineCache sealed : public PolymorphicInlineCache
+    {
+    private:
+        Field(FunctionBody *) functionBody;
```
curtisman: I don't think you need to annotate class that are allocated as leaf.

lib/Runtime/Library/PropertyString.cpp
```diff
@@ -9,71 +9,81 @@ namespace Js
...
     {
     }
 
-    PropertyString* PropertyString::New(StaticType* type, const Js::PropertyRecord* propertyRecord, ArenaAllocator *arena)
+    PropertyString* PropertyString::New(StaticType* type, const Js::PropertyRecord* propertyRecord, Recycler *recycler)
     {
-        PropertyString * propertyString = (PropertyString *)Anew(arena, ArenaAllocPropertyString, type, propertyRecord);
-        propertyString->propCache = AllocatorNewStructZ(InlineCacheAllocator, type->GetScriptContext()->GetInlineCacheAllocator(), PropertyCache);
+        PropertyString * propertyString = RecyclerNewZ(recycler, PropertyString, type, propertyRecord);
+        propertyString->ldElemInlineCache = ScriptContextPolymorphicInlineCache::New(MinPropertyStringInlineCacheSize, type->GetScriptContext());
```
curtisman: Probably can inline these into property string to avoid the extra allocation?

lib/Runtime/Language/InlineCache.cpp
```diff
@@ -697,7 +674,7 @@ namespace Js
 #endif
             if (!PHASE_OFF1(Js::CloneCacheInCollisionPhase))
             {
-                if (!inlineCaches[inlineCacheIndex].IsEmpty() && !inlineCaches[inlineCacheIndex].NeedsToBeRegisteredForStoreFieldInvalidation())
+                if (!inlineCaches[inlineCacheIndex].IsEmpty() && !inlineCaches[inlineCacheIndex].NeedsToBeRegisteredForStoreFieldInvalidation() && GetSize() != 1)
```
rajatd: >1 [](start = 160, length = 1)

is this '1' representing MinPropertyStringInlineCacheSize? If yes, could you use MinPropertyStringInlineCacheSize instead?

lib/Runtime/Language/InlineCache.h
```diff
@@ -374,40 +373,33 @@ namespace Js
...
         static uint16 GetNextSize(uint16 currentSize)
         {
             if (currentSize == MaxPolymorphicInlineCacheSize)
             {
                 return 0;
             }
+            else if (currentSize == MinPropertyStringInlineCacheSize)
+            {
+                CompileAssert(MinPropertyStringInlineCacheSize < MinPolymorphicInlineCacheSize);
+                return MinPolymorphicInlineCacheSize;
```
rajatd: >MinPolymorphicInlineCacheSize [](start = 23, length = 29)

So, we're not gonna grow the cache?

lib/Runtime/Base/ScriptContext.cpp
```diff
@@ -1665,10 +1670,28 @@ namespace Js
...
+#ifdef ENABLE_DEBUG_CONFIG_OPTIONS
+                    if (PHASE_TRACE1(PropertyStringCachePhase))
+                    {
+                        Output::Print(_u("PropertyString '%s' : Invalidating LdElem cache for type %p\n"), string->GetString(), type);
+                    }
+#endif
+                    cache->GetInlineCaches()[cache->GetInlineCacheIndexForType(type)].Clear();
+                }
+                cache = string->GetStElemInlineCache();
+                if (cache->PretendTryGetProperty(type, &info))
```
rajatd: >PretendTryGetProperty [](start = 27, length = 21)

Shouldn't this be a call to PretendTry-SET-Property ?

lib/Runtime/Library/PropertyString.cpp
MikeHolman: right now PIC don't ever grow, we just replace with a new one when we want to grow, so all the paths that deal with growing PIC would need to be updated.

lib/Runtime/Language/InlineCache.cpp
MikeHolman: It's really about the specific size value 1 rather than the abstract variable. Point is if there is only 1 slot, we can't clone to empty slot.

lib/Runtime/Language/InlineCache.h
MikeHolman: this case will grow cache from 1 to 4.

lib/Runtime/Language/InlineCache.cpp
rajatd: Ok, makes sense.

---
In reply to: [115339004](https://github.com/Microsoft/ChakraCore/pull/2883#discussion_r115339004) [](ancestors = 115339004)

lib/Runtime/Language/InlineCache.h
rajatd: Oh right, I misread MinPolymorphicInlineCacheSize and MinPropertyStringInlineCacheSize as the same

---
In reply to: [115339150](https://github.com/Microsoft/ChakraCore/pull/2883#discussion_r115339150) [](ancestors = 115339150)

MikeHolman:

MikeHolman:

MikeHolman:
Can you guys take another look? I have more changes that I want to check in on top of this and don't have much time.

rajatd:
Looks good to me.

PR #2884, opened 2017-04-28
Minor perf improvements

obastemur:
#### type-checking: reduce unnecessary calls
Looking at the implementation, type checking methods are a pretty
inexpensive calls. However CC calls them so many times (mostly
unnecessary) hence they produce a significant heat signature on final
perf result. Reducing the unnecessary calls.

#### ISF-Process: condition-less
In case `!ehBailoutData` it is unlikely the other 2 comparisons will
result true. However, other two may result true but  actually we may
not have `ehBailoutData`. So, condition less.

#### pal: dismiss wcsncmp worst case

### DIFF
--- a/lib/Runtime/Language/JavascriptOperators.cpp
+++ b/lib/Runtime/Language/JavascriptOperators.cpp
@@ -1692,9 +1692,9 @@ CommonNumber:
             RootObjectBase* rootObject = static_cast<RootObjectBase*>(object);
             foundProperty = rootObject->GetRootProperty(instance, propertyId, value, info, requestContext);
         }
+
         while (!foundProperty && JavascriptOperators::GetTypeId(object) != TypeIds_Null)
         {
-
             if (unscopables && IsPropertyUnscopable(object, propertyId))
             {
                 break;
@@ -5521,7 +5521,15 @@ CommonNumber:
 
     Var JavascriptOperators::OP_InitCachedScope(Var varFunc, const Js::PropertyIdArray *propIds, Field(DynamicType*)* literalType, bool formalsAreLetDecls, ScriptContext *scriptContext)
     {
-        ScriptFunction *func = JavascriptGeneratorFunction::Is(varFunc) || JavascriptAsyncFunction::Is(varFunc) ?
+        bool isGAFunction = JavascriptFunction::Is(varFunc);
+        Assert(isGAFunction);
+        if (isGAFunction)
+        {
+            JavascriptFunction *function = JavascriptFunction::FromVar(varFunc);
+            isGAFunction = JavascriptGeneratorFunction::Test(function) || JavascriptAsyncFunction::Test(function);
+        }
+
+        ScriptFunction *func = isGAFunction ?
             JavascriptGeneratorFunction::FromVar(varFunc)->GetGeneratorVirtualScriptFunction() :
             ScriptFunction::FromVar(varFunc);
 
--- a/lib/Runtime/Language/JavascriptStackWalker.cpp
+++ b/lib/Runtime/Language/JavascriptStackWalker.cpp
@@ -444,7 +444,7 @@ namespace Js
                 inlineeOffset = inlinedFrameWalker.GetCurrentInlineeOffset();
             }
         }
-        else if (ScriptFunction::Is(parentFunction) && HasInlinedFramesOnStack())
+        else if (ScriptFunction::Test(parentFunction) && HasInlinedFramesOnStack())
         {
             // Inlined frames are not being walked right now. However, if there
             // are inlined frames on the stack the InlineeCallInfo of the first inlined frame
--- a/lib/Runtime/Library/ScriptFunction.h
+++ b/lib/Runtime/Library/ScriptFunction.h
@@ -45,6 +45,7 @@ namespace Js
     public:
         ScriptFunction(FunctionProxy * proxy, ScriptFunctionType* deferredPrototypeType);
         static bool Is(Var func);
+        inline static BOOL Test(JavascriptFunction *func) { return func->GetFunctionInfo()->HasBody(); }
         static ScriptFunction * FromVar(Var func);
         static ScriptFunction * OP_NewScFunc(FrameDisplay *environment, FunctionInfoPtrPtr infoRef);
 
--- a/lib/Runtime/Math/JavascriptMath.cpp
+++ b/lib/Runtime/Math/JavascriptMath.cpp
@@ -169,45 +169,38 @@ namespace Js
             Assert(aRight != nullptr);
             Assert(scriptContext != nullptr);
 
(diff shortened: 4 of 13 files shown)


lib/Runtime/Library/ScriptFunction.h
```diff
@@ -45,6 +45,7 @@ namespace Js
     public:
         ScriptFunction(FunctionProxy * proxy, ScriptFunctionType* deferredPrototypeType);
         static bool Is(Var func);
+        inline static BOOL Test(JavascriptFunction *func) { return func->GetFunctionInfo()->HasBody(); }
```
MikeHolman: nit: inline is implicit

MikeHolman: nit: i would slightly prefer this to be an overload of `Is` rather than a new function

lib/Runtime/Language/JavascriptOperators.cpp
```diff
@@ -5441,7 +5441,15 @@ namespace Js
 
     Var JavascriptOperators::OP_InitCachedScope(Var varFunc, const Js::PropertyIdArray *propIds, Field(DynamicType*)* literalType, bool formalsAreLetDecls, ScriptContext *scriptContext)
     {
-        ScriptFunction *func = JavascriptGeneratorFunction::Is(varFunc) || JavascriptAsyncFunction::Is(varFunc) ?
+        bool isGAFunction = JavascriptFunction::Is(varFunc);
+        Assert(isGAFunction);
```
MikeHolman: Can you change the function signature to take a `JavascriptFunction *`? It looks like the 2 call sites are doing the right thing, but this is asking for trouble in the future

MikeHolman:
Sorry I didn't see this review until just now. I think at this point release/1.6 is only taking bug fixes, so we should probably retarget to master.

lib/Runtime/Library/ScriptFunction.h
obastemur: Well, agree but that messes with the rest of the usage.

obastemur: updated.

lib/Runtime/Language/JavascriptOperators.cpp
obastemur: I could but not sure the rest of the code would comply with that.

obastemur:
@MikeHolman Thanks for the review. Anything else?

lib/Runtime/Library/ScriptFunction.h
```diff
@@ -45,6 +45,7 @@ namespace Js
     public:
         ScriptFunction(FunctionProxy * proxy, ScriptFunctionType* deferredPrototypeType);
         static bool Is(Var func);
+        static BOOL Test(JavascriptFunction *func) { return func->GetFunctionInfo()->HasBody(); }
```
digitalinfinity: Curious- why does this Test not do a vtable check too?

lib/Runtime/Math/JavascriptMath.cpp
```diff
@@ -169,45 +169,38 @@ namespace Js
...
                     __int64 sum = TaggedInt::ToInt64(aLeft) + TaggedInt::ToInt64(aRight);
                     return JavascriptNumber::ToVar(sum, scriptContext);
                 }
+                else if (typeLeft == TypeIds_String)
+                {
+                    return JavascriptString::Concat(JavascriptString::FromVar(aLeft), JavascriptString::FromVar(aRight));
+                }
             }
-            else if (TaggedInt::Is(aRight))
+            else if(typeLeft == TypeIds_Number && typeRight == TypeIds_Integer)
```
digitalinfinity: Super nit: in some cases, you have a space in between the if and (, and in some cases you don't- can you pick one and stick to it 😄?

obastemur: Great attention! Will do :)

lib/Runtime/Library/ScriptFunction.h
obastemur: Vtable works for sure. That's the problem. Intention with Test is not the same what "Is" provides

lib/Runtime/Language/JavascriptOperators.cpp
MikeHolman: Rest of the code has to comply or we have a type confusion bug. I checked the call sites, and they do comply.

My concern is that someone can add another call to this function in the future where varFunc is not guaranteed to be a JavascriptFunction. This was a problem with the code before, but if you are changing these casts, it's better to eliminate that possibility.

MikeHolman:
I really think you should change the signature for OP_InitCachedScope. Otherwise looks good.

lib/Runtime/Language/JavascriptStackWalker.cpp
```diff
@@ -444,7 +444,7 @@ namespace Js
                 inlineeOffset = inlinedFrameWalker.GetCurrentInlineeOffset();
             }
         }
-        else if (ScriptFunction::Is(parentFunction) && HasInlinedFramesOnStack())
+        else if (ScriptFunction::Test(parentFunction) && HasInlinedFramesOnStack())
```
kunalspathak: >ScriptFunction::Test( [](start = 17, length = 21)

there are couple of other places in `JavascriptObject.cpp`, `FixedFieldInfo.cpp` and `Bailout.cpp` where we do `ScriptFunction::Is`. Can you change those as well?

obastemur: @kunalspathak that's a good point. I just rebased the code (3 months old :) ) to master. I will rescan and see what else I might do it here.

kunalspathak: Let me correct, you need to only change to `ScriptFunction::Test` at call sites that is guarded by `JavascriptFunction::Is()`

lib/Runtime/Math/JavascriptMath.cpp
obastemur: Well, I found 40+ similar usages on the same file, replaced them all ;)

obastemur:
@MikeHolman @kunalspathak ping

MikeHolman:
Can you have GetFunctionExpression return a JavascriptFunction* to avoid the FromVar?

I don't mean to harp on this, but I think it's really important for us to avoid unchecked FromVar calls as much as possible, because it is easy to inadvertently break these assumptions in the future.

obastemur:
> Can you have GetFunctionExpression return a JavascriptFunction* to avoid the FromVar?

After the recent changes, previous definition was surely redundant. Good to go?

obastemur:
@MikeHolman @digitalinfinity @kunalspathak Thanks for the review!

PR #2885, opened 2017-04-28
fix issue where int32 isn't being sign extended when coerced to int64

MikeHolman:


Krovatkin:
LGTM. I didn't realize that `int32` wasn't reconciled to `int64` :-(

lib/Backend/GlobOptIntBounds.cpp
```diff
@@ -174,7 +174,14 @@ bool ValueInfo::TryGetInt64ConstantValue(int64 *const intValueRef, const bool is
         int32 int32ValueRef;
         if (TryGetIntConstantValue(&int32ValueRef, false))
         {
-            *intValueRef = (isUnsigned) ? (uint)int32ValueRef : int32ValueRef;
+            if (isUnsigned)
```
Cellule: That's a very subtle bug hahaha
I guess implicit conversion warning would have caught this.
Would `*intValueRef = (isUnsigned) ? (int64)(uint)int32ValueRef : (int64)int32ValueRef;` have also been a valid fix ?

Krovatkin: As long as there's `int64` in the "else" arm, it seems to work. 

`t2` is indeed sign-extended in the example below:

```cpp
int t = -1;
int64_t t2 = (isUnsigned) ? (uint32_t)t : (int64_t)t;
```
I don't which one is more ChakraCore-idiomatic :-)

PR #2887, opened 2017-04-28
Updating CI script

### LINKED ISSUE #2877: Jenkins Style Checks (EOL Check, Tab Check, Copyright Check) running under task for incorrect branch
See checks that ran for https://github.com/Microsoft/ChakraCore/pull/2826, in particular: https://ci.dot.net/job/Microsoft_ChakraCore/job/release_1.3/job/ubuntu_check_eol_prtest/1835/console

All of that tasks that ran should have had `release_1.4` in their URL, but the Style Check tasks were `release_1.3`.

* Verify that the style check tasks are correctly generated for `release_1.4`
* Track down why the wrong tasks were triggered

Not sure if this has caused any problems because these tasks are all the same and the diff (for files affected by the change) is taken by comparing against `master` (in the future this behavior will be changed to diff against the target branch (https://github.com/Microsoft/ChakraCore/issues/962)), but I would be suspicious about whether whatever the problem is could cause problems in the future or affect more complex non-style checks.

MSLaguana:
Prior to this change, we were building some configurations for all branches when a PR came in.
This is just backporting the netci.groovy script from `release/1.4` to be more consistent.

Fixes #2877

### DIFF
--- a/netci.groovy
+++ b/netci.groovy
@@ -4,7 +4,7 @@
 //-------------------------------------------------------------------------------------------------------
 
 // Import the utility functionality.
-import jobs.generation.Utilities;
+import jobs.generation.Utilities
 
 // Grab the github project name passed in
 def project = GithubProject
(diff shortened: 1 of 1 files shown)


MSLaguana:
@dotnet-bot test ci please

dilijev:
See https://github.com/Microsoft/ChakraCore/issues/2877,

dilijev:
@MSLaguana look at that propagation :P

Issue #2862, opened 2017-04-21
Issues found when building with MSVC + /permissive-


xiangfan-ms:
The C++ compiler team is currently doing daily RWC (Read World Code) testing which includes Chakra.

We are also working with internal partners and the community to fix source issues exposed by "/permissive-".
The option isn’t finalized yet (we plan to add some large impact features into it in the incoming VS update), so we don’t suggest to adopt it yet.
In the meantime, we want to address these source issues as early as possible so that the community can adopt it when it is finalized.

We discover some source issues in Chakra when we build it with "/permissive-".

Here is the issue:

src\lib\runtime\Library/JavascriptLibrary.h(770): error C4596: 'CreatePromiseAllResolveElementFunction_TTD': illegal qualified name in member declaration

        Js::RecyclableObject* JavascriptLibrary::CreatePromiseAllResolveElementFunction_TTD(Js::JavascriptPromiseCapability* capabilities, uint32 index, Js::JavascriptPromiseAllResolveElementFunctionRemainingElementsWrapper* wrapper, Js::RecyclableObject* values, bool alreadyCalled);

The member function declaration above is inside the class ‘JavascriptLibrary’ and the prefix 'JavascriptLibrary::' before the member function name isn’t allowed. The fix is to remove it.

mrkmarron:
I have updated my open TTD related PR (PR #2853) with a fix for this issue. That is targeting release/2.0 and should  flow to master shortly afterwords.

xiangfan-ms:
Hi, there is a similar issue in src/vm/sampleprofiler.h

static DWORD WINAPI SampleProfiler::ThreadProc(void *args);
->
static DWORD WINAPI ThreadProc(void *args);

xiangfan-ms:
We find another issue in bin\ch\ch.cpp.

ch.cpp(1085): error C2362: initialization of 'lock' is skipped by 'goto return_cleanup'

It is something like this which is ill-formed:

	void f()
	{
		goto return_cleanup;
		int lock = 0;
	return_cleanup:;
	}

The offending initialization (which is skipped by goto):

	// The following code is present to make sure we don't load
	// jscript9.dll etc with ch. Since that isn't a concern on non-Windows
	// builds, it's safe to conditionally compile it out.
	#ifdef _WIN32
	ATOM lock = ::AddAtom(szChakraCoreLock);
	AssertMsg(lock, "failed to lock chakracore.dll");
	#endif // _WIN32

[...]

PR #2889, opened 2017-04-29
Fix an issue pointed out by VS's /permissive-

Penguinwizzard:
See issue #2862


Penguinwizzard:
@xiangfan-ms please take a look as this is related to your issue.

PR #2818, opened 2017-04-08
Fix #2770: Implement Intl.getCanonicalLocales

dilijev:
https://tc39.github.io/ecma402/#sec-intl.getcanonicallocales
https://tc39.github.io/ecma402/#sec-canonicalizelocalelist

See:
https://github.com/tc39/ecma402/pull/78
https://github.com/tc39/ecma402/issues/46

Fixes #2770.

PR #2898, opened 2017-05-02
[Prototype] Designed a way to write built ins functions in JavaScript.

veler:
New prototype that gives the possibility to write a built in function, like Array.prototype.indexOf, in JavaScript.

The current prototype cannot be used in a lot of use case for the moment. If it becomes the good direction to take, it will be improved to be more customizable.


lib/Runtime/Library/BuiltIn/BuiltIn.js
```diff
@@ -0,0 +1,60 @@
...
+"use strict";
+
+(function (intrinsic) {
+    var platform = intrinsic.BuiltIn;
+
+    let FunctionsEnum = {
+        ArrayIndexOf: { className: "Array", methodName: "indexOf", argumentsCount: 1 },
+    };
+
+    platform.registerFunction(FunctionsEnum.ArrayIndexOf, function (searchElement, fromIndex) {
```
MikeHolman: I think this needs to be `function indexOf`, looking at some baseline changes this seems like it is user visible

dilijev: I think that was supposed to be handled by the `ArrayIndexOf: { className: "Array", methodName: "indexOf", argumentsCount: 1 }` above, but it looks like it didn't work?

test/DebuggerCommon/ES6_proto_invalidation.js.dbg.baseline
```diff
@@ -459,13 +465,6 @@
               "caller": "object null",
               "arguments": "object null"
             },
-            "indexOf": {
-              "#__proto__": "function <large string>",
-              "length": "number 1",
-              "name": "string indexOf",
-              "caller": "object null",
-              "arguments": "object null"
```
MikeHolman: we need to understand why we no longer have `caller` or `arguments` and we have new `prototype` property.

lib/Backend/IRBuilder.cpp
```diff
@@ -1797,6 +1797,15 @@ IRBuilder::BuildReg1(Js::OpCode newOpcode, uint32 offset, Js::RegSlot R0)
         }
         dstIsCatchObject = true;
         break;
+
+    case Js::OpCode::LdChakraLib:
+        {
+            const auto addrOpnd = IR::AddrOpnd::New(m_func->GetScriptContextInfo()->GetChakraLibAddr(), IR::AddrOpndKindDynamicVar, m_func, true);
+            addrOpnd->SetValueType(ValueType::Boolean);
```
Cellule: I think this should be `ValueType::PrimitiveOrObject`

lib/Common/ConfigFlagsList.h
```diff
@@ -501,6 +501,9 @@ PHASE(All)
 
 #define DEFAULT_CONFIG_Intl                    (true)
 #define DEFAULT_CONFIG_IntlBuiltIns            (true)
+#define DEFAULT_CONFIG_BuiltIn                 (true)
```
Cellule: This should be false if `#ifndef ENABLE_BUILTIN_OBJECT`

Cellule:
Since you created a `builtin` flag, you should not remove the old `indexOf` code.
That way if we run ch with `ch.exe test.js -builtin-` we should be using the old c++ indexOf version

lib/Common/CommonDefines.h
```diff
@@ -102,6 +102,7 @@
 // xplat-todo: revisit these features
 #ifdef _WIN32
 #define ENABLE_INTL_OBJECT                          // Intl support
+#define ENABLE_BUILTIN_OBJECT                       // Built In functions support
```
dilijev: >ENABLE_BUILTIN_OBJECT [](start = 8, length = 21)

Is there going to be a Builtin object? Certainly not one exposed to the user though. Probably best to rename this to something like ENABLE_JS_BUILTINS

lib/Parser/rterrors.h
```diff
@@ -366,6 +366,9 @@ RT_ERROR_MSG(JSERR_CantRedefineProp, 5667, "Cannot redefine property '%s'", "Can
 RT_ERROR_MSG(JSERR_FunctionArgument_NeedArrayLike, 5668, "%s: argument is not an array or array-like object", "Array or array-like object expected", kjstTypeError, 0)
 RT_ERROR_MSG(JSERR_FatalMemoryExhaustion, 5669, "", "Encountered a non-recoverable OOM", kjstError, 0)
 
+// Built In functions Errors
+RT_ERROR_MSG(JSERR_BuiltInNotAvailable, 5670, "", "Built In functions are not available.", kjstTypeError, 0)
+
```
dilijev: This should never be seen by the user, right?

lib/Common/ConfigFlagsList.h
```diff
@@ -501,6 +501,9 @@ PHASE(All)
...
+
+#define DEFAULT_CONFIG_LdChakraLib (false)
```
dilijev: > (false) [](start = 34, length = 8)

nit: Formatting

lib/Runtime/ByteCode/ByteCodeEmitter.cpp
```diff
@@ -5433,6 +5434,11 @@ void ByteCodeGenerator::EmitPropLoad(Js::RegSlot lhsLocation, Symbol *sym, Ident
             case Js::PropertyIds::undefined:
                 opcode = Js::OpCode::LdUndef;
                 break;
+            case Js::PropertyIds::__chakraLibrary:
+                if (CONFIG_FLAG(LdChakraLib)) {
+                    opcode = Js::OpCode::LdChakraLib;
```
dilijev: >opcode = Js::OpCode::LdChakraLib; [](start = 20, length = 33)

Maybe I didn't see the logic, but I think after the first time you run this bytecode, a flag could be set so that any future generated JIT does not need to run this instruction to load the lib. The helper should also ensure it is only loaded once.

lib/Runtime/Library/Chakra.Runtime.Library.vcxproj.filters
```diff
@@ -118,6 +118,7 @@
     <ClCompile Include="$(MSBuildThisFileDirectory)WebAssemblyTable.cpp" />
     <ClCompile Include="$(MSBuildThisFileDirectory)WebAssemblyEnvironment.cpp" />
     <ClCompile Include="$(MSBuildThisFileDirectory)WabtInterface.cpp" />
+    <ClCompile Include="BuiltInEngineInterfaceExtensionObject.cpp" />
```
dilijev: >BuiltInEngineInterfaceExtensionObject.cpp [](start = 24, length = 41)

Please add $(MSBuildThisFileDirectory)

test/DebuggerCommon/ES6_proto_invalidation.js.dbg.baseline
veler: Status : 'caller' and 'arguments' were not set anymore. It's fixed. The 'name' property was not set correctly. It's fixed. The 'prototype' is under investigation

---
In reply to: [114233872](https://github.com/Microsoft/ChakraCore/pull/2898#discussion_r114233872) [](ancestors = 114233872)

lib/Runtime/Library/Chakra.Runtime.Library.vcxproj
```diff
@@ -164,12 +164,18 @@
     <ClCompile Include="$(MSBuildThisFileDirectory)WebAssemblyTable.cpp" />
     <ClCompile Include="$(MSBuildThisFileDirectory)WebAssemblyEnvironment.cpp" />
     <ClCompile Include="$(MSBuildThisFileDirectory)WabtInterface.cpp" />
+    <ClCompile Include="BuiltInEngineInterfaceExtensionObject.cpp" />
```
dilijev: >"BuiltInEngineInterfaceExtensionObject.cpp" [](start = 23, length = 43)

Please add $(MSBuildThisFileDirectory)

lib/Runtime/Library/BuiltIn/BuiltIn.js
veler: I confirm, it is supposed to be handled by the ArrayIndexOf object. The name was not set correctly. it is fixed now.

---
In reply to: [114233688](https://github.com/Microsoft/ChakraCore/pull/2898#discussion_r114233688) [](ancestors = 114233688)

dilijev: >FunctionsEnum.ArrayIndexOf [](start = 30, length = 26)

Maybe add a link to the spec section for ergonomics in referencing the implementation?

* [https://tc39.github.io/ecma262/#sec-array.prototype.indexof](https://tc39.github.io/ecma262/#sec-array.prototype.indexof)

In your first commit, the message said array+string.prototype.indexOf -- but the implementations are different, right? (It appears you're doing just Array here.)

[https://tc39.github.io/ecma262/#sec-string.prototype.indexof](https://tc39.github.io/ecma262/#sec-string.prototype.indexof)

lib/Runtime/Library/BuiltIn/BuiltIn.js
```diff
@@ -0,0 +1,60 @@
+//-------------------------------------------------------------------------------------------------------
+// Copyright (C) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE.txt file in the project root for full license information.
+//-------------------------------------------------------------------------------------------------------
+
+"use strict";
+
+(function (intrinsic) {
+    var platform = intrinsic.BuiltIn;
```
dilijev: >platform [](start = 8, length = 8)

What is the distinction between platform and the global __chakraLibrary objects?

Please add a test to double check that the __chakraLibrary code is not exposed to the user-facing runtime.

lib/Runtime/Library/BuiltIn/BuiltIn.js
```diff
@@ -0,0 +1,60 @@
...
+
+    let FunctionsEnum = {
+        ArrayIndexOf: { className: "Array", methodName: "indexOf", argumentsCount: 1 },
+    };
+
+    platform.registerFunction(FunctionsEnum.ArrayIndexOf, function (searchElement, fromIndex) {
+        let o = __chakraLibrary.Object(this);
```
dilijev: >Object [](start = 32, length = 6)

Is this the Object constructor? The spec defines the Object constructor and the ToObject internal operation differently: 

* [https://tc39.github.io/ecma262/#sec-object-value](https://tc39.github.io/ecma262/#sec-object-value) 
* [https://tc39.github.io/ecma262/#sec-toobject](https://tc39.github.io/ecma262/#sec-toobject)

lib/Runtime/Library/BuiltIn/BuiltIn.js
```diff
@@ -0,0 +1,60 @@
...
+
+        let k;
+
+        if (n >= 0) {
+            if (len == -0) {
+                k = +0;
+            } else {
+                k = n;
+            }
+        } else if (n < 0) {
```
dilijev: >if (n < 0) [](start = 15, length = 10)

nit: drop this if (n < 0) since it is redundant.

lib/Runtime/Library/BuiltIn/BuiltIn.js
```diff
@@ -0,0 +1,60 @@
...
+            k = len + n;
+
+            if (k < 0) {
+                k = 0;
+            }
+        }
+
+        while (k < len) {
+            if (k in o) {
```
dilijev: >k in o [](start = 16, length = 6)

Is this definitely equivalent to `Let kPresent be ? HasProperty(O, ! ToString(k)).` from the spec (step 8.a.)?

lib/Runtime/Library/BuiltIn/BuiltIn.js
```diff
@@ -0,0 +1,60 @@
...
+                let elementK = o[k];
```
dilijev: >o[k] [](start = 31, length = 4)

Similarly, is this definitely equivalent to 8.b.i.: `Let elementK be ? Get(O, ! ToString(k)).`?

lib/Runtime/Library/BuiltIn/BuiltIn.js
```diff
@@ -0,0 +1,60 @@
...
+
+                if (elementK === searchElement) {
+                    return k;
+                }
+            }
+
+            k++;
```
dilijev: >k++ [](start = 12, length = 3)

nit: ++k requires less analysis to optimize to an increment with no side effect.

test/Array/array_indexOf.baseline
```diff
@@ -664,6 +664,7 @@
 -1
 -1
 -1
+5
```
dilijev: This baseline is totally unreadable -- as long as you're in the area, if you have time, maybe you can rewrite this with the UnitTestFramework to use asserts and then delete the baseline file entirely.

test/BuiltIns/ChakraLib.js
```diff
@@ -0,0 +1,22 @@
...
+// Licensed under the MIT license. See LICENSE.txt file in the project root for full license information.
+//-------------------------------------------------------------------------------------------------------
+
+WScript.LoadScriptFile("..\\UnitTestFramework\\UnitTestFramework.js");
+
+var tests = [
+    {
+        name: "Access to __chakralib",
+        body: function () {
+            assert.areEqual(__chakralib, undefined, "__chakralib must be undefined outside of the BuiltIns methods.");
```
dilijev: >__chakralib, undefined [](start = 28, length = 22)

Oh I see you've already added the test I asked for. Carry on :)

test/BuiltIns/rlexe.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="utf-8"?>
+<regress-exe>
+  <test>
+    <default>
+      <files>ChakraLib.js</files>
+      <compile-flags>-args summary -endargs</compile-flags>
+      <tags>BuiltIn LdChakraLib</tags>
```
dilijev: ><tags>BuiltIn LdChakraLib</tags> [](start = 6, length = 32)

Tags need to be a comma-separated list of tokens with no spaces (not even around the comma, as that causes anything after the first tag to be ignored by rl.exe).

Tags are used to select which tests to run or exclude from a given run of unit tests. If you intend to add either of these tags to be excluded from certain test runs, you should do that, for example a test run of a build where the binary was built with the BuiltIns explicitly turned off, you would want to run your test suite with `-nottags BuiltIn`. We did this with Intl because the Intl object is not necessarily enabled (for example it is not currently enabled on xplat). All the Intl tests are tagged as Intl.

It makes sense to have a config which disables tests tagged BuiltIn, but I'm not sure if it makes sense to disable tests when the LdChakraLib bytecode is not available. Will this ever be the case?

test/Debugger/JsDiagGetFunctionPosition.js.dbg.baseline
```diff
@@ -31,8 +31,8 @@
   },
   {
     "functionPosition": {
-      "scriptId": 5,
-      "parentScriptId": 4,
+      "scriptId": 6,
+      "parentScriptId": 5,
```
dilijev: Are there any tools that rely on these indexes directly? Just food for thought that some updates might be needed.

test/DebuggerCommon/ES6_proto_invalidation.js.dbg.baseline
```diff
@@ -363,6 +363,12 @@
               "__lookupGetter__": "function <large string>",
               "__lookupSetter__": "function <large string>"
             },
+            "indexOf": {
+              "#__proto__": "function <large string>",
+              "prototype": "Object {...}",
+              "name": "string ",
+              "length": "number 1"
+            },
```
dilijev: Is this change expected?

test/Intl/DateTimeFormatOptions.js
```diff
@@ -51,7 +51,7 @@ var tests = [
                 }
                 catch (e) {
                     if (!(e instanceof RangeError)) {
-                        fail("Incorrect exception was thrown.");
+                        assert.fail("Incorrect exception was thrown.");
```
dilijev: >assert.fail("Incorrect exception was thrown."); [](start = 24, length = 47)

I'll add this change to my PR here: [https://github.com/Microsoft/ChakraCore/pull/2818](https://github.com/Microsoft/ChakraCore/pull/2818)

lib/Runtime/Library/BuiltIn/BuiltIn.js
dilijev: [](http://msn.com)

test/DebuggerCommon/ES6_proto_invalidation.js.dbg.baseline
veler: The fact that the position of IndexOf in this list has changed is normal, as we don't load/declare IndexOf at the same place anymore. But it miss a few info compare to before. I actually fixed that earlier.

---
In reply to: [115389957](https://github.com/Microsoft/ChakraCore/pull/2898#discussion_r115389957) [](ancestors = 115389957)

lib/Runtime/Library/BuiltIn/BuiltIn.js
veler: Unit test added.
Platform is a private variable sent in argument to this function that expose (for the moment) ONE private function called "registerFunction" which basically link a given method to a prototype.
Inside the given method, as the scope is now the scope of the given prototype (here, Array.prototype), we don't have access to the platform variable. That is why we use __chakraLibrary which is a global variable available only when the -LdChakraLib is set. This variable gives access to a set of private internal functions like toLength or toNumber that the consumers must not be able to exploit.

---
In reply to: [115388059](https://github.com/Microsoft/ChakraCore/pull/2898#discussion_r115388059) [](ancestors = 115388059)

lib/Runtime/Library/BuiltIn/BuiltIn.js
veler: Yes it is, we checked the behavior of this syntax in C++ and it is exactly the same.

---
In reply to: [115389001](https://github.com/Microsoft/ChakraCore/pull/2898#discussion_r115389001) [](ancestors = 115389001)

test/BuiltIns/rlexe.xml
Cellule: Technically LdChakraLib should always be turned off except when serializing the bytecode for the builtins if I understood correctly

lib/Runtime/Library/BuiltIn/BuiltIn.js
veler: Yes it is. HasProperty(O, ! ToString(k)) === IsIn. But HasOwnProperty != IsIn.

---
In reply to: [115388973](https://github.com/Microsoft/ChakraCore/pull/2898#discussion_r115388973) [](ancestors = 115388973)

test/BuiltIns/rlexe.xml
veler: That's right. I've made a new unit test to make it clearer and fixed the arguments.

---
In reply to: [115563779](https://github.com/Microsoft/ChakraCore/pull/2898#discussion_r115563779) [](ancestors = 115563779)

lib/Runtime/ByteCode/ByteCodeEmitter.cpp
veler: That's what is happening actually. When we generate the bytecode for the JS built ins function, the __chakraLibrary is loaded and gives access to ToLength, ToNumber...etc. Then, when we run ChakraCore without the flag -LdChakraLib, we don't load __chakraLibrary  anymore. But the JS Built ins methods still have access to it, privately.

---
In reply to: [115379529](https://github.com/Microsoft/ChakraCore/pull/2898#discussion_r115379529) [](ancestors = 115379529)

lib/Parser/rterrors.h
veler: Good idea !

---
In reply to: [115377532](https://github.com/Microsoft/ChakraCore/pull/2898#discussion_r115377532) [](ancestors = 115377532)

lib/Runtime/Library/BuiltIn/BuiltIn.js
veler: Yes it's the object constructor. Actually from what I understand here, [https://tc39.github.io/ecma262/#sec-toobject](https://tc39.github.io/ecma262/#sec-toobject) , we throw an error in case of Null or Undefined value, then we simply instantiate a new Object. I was think that our object constructor was already doing that, but it's not checking for null or undefined value. I fixed the problem.

---
In reply to: [115388233](https://github.com/Microsoft/ChakraCore/pull/2898#discussion_r115388233) [](ancestors = 115388233)

test/Array/array_indexOf.baseline
veler: I written a post it to keep it in mind. I'll do it late June if I have the time.

---
In reply to: [115389276](https://github.com/Microsoft/ChakraCore/pull/2898#discussion_r115389276) [](ancestors = 115389276)

lib/Runtime/Library/BuiltIn/BuiltIn.js
veler: After discussing with Doug, there is no reason to change this line.

---
In reply to: [115389078](https://github.com/Microsoft/ChakraCore/pull/2898#discussion_r115389078) [](ancestors = 115389078)

test/Debugger/JsDiagGetFunctionPosition.js.dbg.baseline
MikeHolman: i think this should be fine.

---
In reply to: [115389856](https://github.com/Microsoft/ChakraCore/pull/2898#discussion_r115389856) [](ancestors = 115389856)

lib/Backend/Lower.cpp
```diff
@@ -18550,6 +18551,78 @@ Lowerer::GenerateFastInlineRegExpExec(IR::Instr * instr)
     RelocateCallDirectToHelperPath(tmpInstr, labelHelper);
 }
 
+void Lowerer::GenerateFastInlineIsIn(IR::Instr * instr)
```
MikeHolman: can you add a comment at the top of this function for what code this generates?

lib/Backend/Lower.cpp
```diff
@@ -18550,6 +18551,78 @@ Lowerer::GenerateFastInlineRegExpExec(IR::Instr * instr)
...
+    IR::IndirOpnd* indirOpnd = IR::IndirOpnd::New(src2RegOpnd, Js::JavascriptArray::GetOffsetOfHead(), TyMachPtr, this->m_func);
+    IR::RegOpnd* headSegmentOpnd = IR::RegOpnd::New(TyMachPtr, this->m_func);
+    autoReuseHeadSegmentOpnd.Initialize(headSegmentOpnd, m_func);
+    InsertMove(headSegmentOpnd, indirOpnd, instr);
+
+    IR::Opnd* headSegmentLengthOpnd = IR::IndirOpnd::New(headSegmentOpnd, Js::SparseArraySegmentBase::GetOffsetOfLength(), TyUint32, m_func);
+    autoReuseHeadSegmentLengthOpnd.Initialize(headSegmentLengthOpnd, m_func);
+
+    InsertCompare(src1Untagged, headSegmentLengthOpnd, instr);
+    InsertBranch(Js::OpCode::BrGe_A /* >= */, true /* isUnsigned */, helperLabel, instr);
```
MikeHolman: nit: you can combine these 2 into InsertCompareBranch

lib/Backend/Lower.cpp
```diff
@@ -18550,6 +18551,78 @@ Lowerer::GenerateFastInlineRegExpExec(IR::Instr * instr)
     RelocateCallDirectToHelperPath(tmpInstr, labelHelper);
 }
 
+void Lowerer::GenerateFastInlineIsIn(IR::Instr * instr)
+{
+    // operator "foo in bar"
+    IR::Opnd* src1 = instr->GetSrc1(); // foo
+    IR::Opnd* src2 = instr->GetSrc2(); // bar
+
+    IR::LabelInstr* helperLabel = IR::LabelInstr::New(Js::OpCode::Label, m_func, true);
```
MikeHolman: can you only emit this fastpath if valuetype for object isLikelyArray and HasNoMissingValues before emitting this fast path?

lib/Runtime/Base/FunctionBody.cpp
```diff
@@ -3018,9 +3018,9 @@ namespace Js
         return this->GetLineCharOffsetFromStartChar(startCharOfStatement, _line, _charOffset, canAllocateLineCache);
     }
 
-    bool FunctionBody::GetLineCharOffsetFromStartChar(int startCharOfStatement, ULONG* _line, LONG* _charOffset, bool canAllocateLineCache /*= true*/)
+    bool FunctionBody::GetLineCharOffsetFromStartChar(int startCharOfStatement, ULONG* _line, LONG* _charOffset, bool canAllocateLineCache /*= true*/, bool canDumpLibraryCode /*= false*/)
     {
-        Assert(!this->GetUtf8SourceInfo()->GetIsLibraryCode());
+        Assert(canDumpLibraryCode || !this->GetUtf8SourceInfo()->GetIsLibraryCode());
```
MikeHolman: since we found this doesn't work for library code, let's remove the argument

lib/Runtime/Library/JavascriptArray.cpp
```diff
@@ -3827,20 +3827,21 @@ namespace Js
     // Array.prototype.indexOf as defined in ES6.0 (final) Section 22.1.3.11
     Var JavascriptArray::EntryIndexOf(RecyclableObject* function, CallInfo callInfo, ...)
     {
+        //return nullptr;
```
MikeHolman: nit: remove this comment

lib/Runtime/Base/ThreadConfigFlagsList.h
```diff
@@ -7,6 +7,7 @@
 FLAG(IsCollectGarbageEnabled, CollectGarbage)
 FLAG(IsErrorStackTraceEnabled, errorStackTrace)
 FLAG(IsES6UnicodeVerboseEnabled, ES6UnicodeVerbose)
+FLAG_RELEASE(IsJsBuiltInEnabled, JsBuiltIn)
```
MikeHolman: i think this shouldn't be release flag

lib/Common/ConfigFlagsList.h
```diff
@@ -960,6 +968,9 @@ FLAGNR(Boolean, CollectGarbage        , "Enable CollectGarbage API", DEFAULT_CON
 FLAGR (Boolean, Intl                  , "Intl object support", DEFAULT_CONFIG_Intl)
 FLAGNR(Boolean, IntlBuiltIns          , "Intl built-in function support", DEFAULT_CONFIG_IntlBuiltIns)
 
+FLAGR (Boolean, JsBuiltIn             , "JS Built-in function support", DEFAULT_CONFIG_JsBuiltIn)
```
MikeHolman: make this FLAGNR

lib/Common/CommonDefines.h
```diff
@@ -102,6 +102,7 @@
 // xplat-todo: revisit these features
 #ifdef _WIN32
 #define ENABLE_INTL_OBJECT                          // Intl support
+#define ENABLE_JS_BUILTINS                       // Built In functions support
```
MikeHolman: i think we should make it so this works on linux too

lib/Runtime/Library/JavascriptLibrary.cpp
```diff
@@ -1713,8 +1749,10 @@ namespace Js
 
         builtinFuncs[BuiltinFunction::JavascriptArray_Unshift]            = library->AddFunctionToLibraryObject(arrayPrototype, PropertyIds::unshift,         &JavascriptArray::EntryInfo::Unshift,           1);
 
-
-        builtinFuncs[BuiltinFunction::JavascriptArray_IndexOf]        = library->AddFunctionToLibraryObject(arrayPrototype, PropertyIds::indexOf,         &JavascriptArray::EntryInfo::IndexOf,           1);
+        if (!scriptContext->IsJsBuiltInEnabled())
```
MikeHolman: you can remove the scriptContext function for this and check `CONFIG_FLAG(JsBuiltin)` instead

lib/Runtime/Library/JsBuiltIn/JsBuiltIn.js
```diff
@@ -0,0 +1,66 @@
...
+        let n = __chakraLibrary.toInteger(fromIndex);
+
+        if (n >= len) {
+            return -1;
+        }
+
+        let k;
+
+        if (n >= 0) {
+            if (len == -0) {
```
MikeHolman: this should be `if(n === -0)` but also, I think we should change check to be `if(n === 0) k = 0;` This is valid because -0 == +0 == 0. However, when you use -0 or +0, our engine might think you are meaning to use floating point numbers which can hurt performance.
Also, we can change the statement around to be like `if (n===0) { k = 0 } else if (n > 0) { k = n } else { ... }`, which should be more optimal

lib/Runtime/Library/JavascriptArray.cpp
```diff
@@ -3828,19 +3828,19 @@ namespace Js
     Var JavascriptArray::EntryIndexOf(RecyclableObject* function, CallInfo callInfo, ...)
     {
         PROBE_STACK(function->GetScriptContext(), Js::Constants::MinStackDefault);
-
+
```
dilijev: nit: trim trailing whitespace

dilijev:
>     // Array.prototype.indexOf as defined in ES6.0 (final) Section 22.1.3.11

Should probably use ES 2017 final spec instead: [https://tc39.github.io/ecma262/2017/](https://tc39.github.io/ecma262/2017/)

---
Refers to: lib/Runtime/Library/JavascriptArray.cpp:3827 in 95e59de. [](commit_id = 95e59deba35f59053cbed49cbdd3a3d9dcdf71ba, deletion_comment = False)

dilijev:
>     // Array.prototype.indexOf as defined in ES6.0 (final) Section 22.1.3.11

Please use the anchor tag for the spec location instead, as it's less likely to change between versions. In 2017, 22.1.3.11 is [https://tc39.github.io/ecma262/2017/#sec-array.prototype.includes](https://tc39.github.io/ecma262/2017/#sec-array.prototype.includes)

Array.prototype.indexOf is now 22.1.3.12 and will always be [https://tc39.github.io/ecma262/2017/#sec-array.prototype.indexof](https://tc39.github.io/ecma262/2017/#sec-array.prototype.indexof)

In the source, reference this as #sec-array.prototype.indexof 

---
Refers to: lib/Runtime/Library/JavascriptArray.cpp:3827 in 95e59de. [](commit_id = 95e59deba35f59053cbed49cbdd3a3d9dcdf71ba, deletion_comment = False)

lib/Runtime/Library/JsBuiltIn/JsBuiltIn.js
```diff
@@ -0,0 +1,65 @@
...
+
+        let n = __chakraLibrary.toInteger(fromIndex);
+
+        if (n >= len) {
+            return -1;
+        }
+
+        let k;
+
+        if (n === 0) {
```
dilijev: >if (n === 0) { [](start = 8, length = 14)

Probably add a comment to this per Michael's feedback explaining why it works even though it doesn't match the spec text.

test/JsBuiltIns/rlexe.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="utf-8"?>
+<regress-exe>
+  <test>
+    <default>
+      <files>ChakraLib.js</files>
+      <compile-flags>-args summary -endargs -LdChakraLib</compile-flags>
```
dilijev: > -LdChakraLib [](start = 43, length = 13)

Is this totally separate from the builtins? Does the JsBuiltIn code work without -LdChakraLib? Should these flags be connected?

test/JsBuiltIns/ChakraLib_disabled.js
```diff
@@ -0,0 +1,24 @@
...
+        name: "Access to __chakraLibrary must not work",
+        body: function () {
+            assert.isUndefined(__chakraLibrary, "__chakraLibrary must be undefined outside of the BuiltIns methods.");
+        },
+        name: "Local __chakraLibrary variable must not have an impact on indexOf",
+        body: function () {
+            var __chakraLibrary = -1;
+            var array = [1, 2, 3, 4];
+            assert.areEqual(__chakraLibrary, -1, "__chakraLibrary must be -1 outside of the BuiltIns methods.");
+            assert.areEqual(array.indexOf(1), 0, "The __chakraLibrary in the indexOf method must give access to built ins methods instead of -1.");
```
dilijev: >1 [](start = 42, length = 1)

Just to be extra clear, maybe look for indexOf(4) and the result should be 3.

dilijev: >__chakraLibrary [](start = 54, length = 15)

I see, so the -LdChakraLib logic always applies to the JsBuiltIn code and the flag just allows you to run this set of tests like this.

test/JsBuiltIns/ChakraLib.js
```diff
@@ -0,0 +1,17 @@
...
+// Licensed under the MIT license. See LICENSE.txt file in the project root for full license information.
+//-------------------------------------------------------------------------------------------------------
+
+WScript.LoadScriptFile("..\\UnitTestFramework\\UnitTestFramework.js");
+
+var tests = [
+    {
+        name: "Access to __chakraLibrary work",
+        body: function () {
+            assert.isNotUndefined(__chakraLibrary, "__chakraLibrary must NOT be undefined as we're using the -LdChakraLib flag.");
```
dilijev: >__chakraLibrary [](start = 34, length = 15)

What if you try to overwrite this like you did in the other test? You should ensure that inside of the JsBuiltIn code it is still the correct object.

test/JsBuiltIns/ChakraLib_disabled.js
veler: Exactly :)

---
In reply to: [115902542](https://github.com/Microsoft/ChakraCore/pull/2898#discussion_r115902542) [](ancestors = 115902542)

test/JsBuiltIns/rlexe.xml
veler: The LdChakraLib is used to enable/disable the __chakraLibrary global variable used to access to the internal C++ methods from Chakra like toLength and toNumber. It must be used only when we generate the bytecode of the JsBuiltIn.js file.
The JsBuiltIn flag is used to enable/disable the JsBuiltInEngineExtension in C++. When this flag is disabled, the previous version of the built ins functions (the C++ version of IndexOf) take the relay.
Those two flags are independent but will mainly be used together. A use case of disabling JsBuiltIn would be to do some debugging stuff or compare performances without having to retrieve an exe or dll from chakrafs.

---
In reply to: [115902305](https://github.com/Microsoft/ChakraCore/pull/2898#discussion_r115902305) [](ancestors = 115902305)

test/Array/array_indexOf.baseline
dilijev: Fair enough.

veler:
>     // Array.prototype.indexOf as defined in ES6.0 (final) Section 22.1.3.11

Please consider that this implementation of IndexOf might be deprecated (not following the last specs) and that this C++ code will be removed once this experimentation is done.

---
In reply to: [300677156](https://github.com/Microsoft/ChakraCore/pull/2898#issuecomment-300677156) [](ancestors = 300677156)

---
Refers to: lib/Runtime/Library/JavascriptArray.cpp:3827 in c8a3408. [](commit_id = c8a340828c83c7c9195f3d5bdd1998f348eb70c2, deletion_comment = False)

test/JsBuiltIns/ChakraLib.js
veler: With the LdChakraLib flag, overriding must not work. Test added. :)

---
In reply to: [115902571](https://github.com/Microsoft/ChakraCore/pull/2898#discussion_r115902571) [](ancestors = 115902571)

lib/Runtime/Library/JavascriptArray.cpp
```diff
@@ -3824,7 +3824,7 @@ namespace Js
         }
     }
 
-    // Array.prototype.indexOf as defined in ES6.0 (final) Section 22.1.3.11
+    // Array.prototype.indexOf as defined in ES6.0 (final) Section 22.1.3.11 https://tc39.github.io/ecma262/2017/#sec-array.prototype.includes
```
dilijev: >ES6.0 (final) [](start = 45, length = 13)

nit: ES6 was ES 2015 ([https://www.ecma-international.org/ecma-262/6.0/](https://www.ecma-international.org/ecma-262/6.0/)) -- if you intend this reference to refer specifically to ES 2017 spec, also update this text to say ES 2017

lib/Runtime/Library/JsBuiltIn/JsBuiltIn.js
```diff
@@ -0,0 +1,69 @@
...
+        if (n >= len) {
+            return -1;
+        }
+
+        let k;
+
+        /* We refactored the code but it still respect the spec.
+           When using -0 or +0, the engine might think we are meaning
+           to use floating point numbers which can hurt performance.
+           So we refactored to use integers instead. */
```
dilijev: >So we refactored to use integers instead. [](start = 11, length = 41)

nit: also add something about why `=== 0` works. This basically explains why we differ from the spec, but not why it is okay to do so.
To be extra transparent, include in the comment the spec lines where we are approximately different.

test/JsBuiltIns/ChakraLib.js
dilijev: New test LGTM, thanks.

---
In reply to: [116263296](https://github.com/Microsoft/ChakraCore/pull/2898#discussion_r116263296) [](ancestors = 116263296,115902571)

lib/Runtime/ByteCode/ByteCodeEmitter.cpp
dilijev: Forgive any rambling as I'm not a JIT expert...
What if those JS Built-ins are inlined into the JIT? You'd need to ensure that the library has been loaded and is visible only to the built-in and that it doesn't leak. But you also want to make sure that it is loaded the minimum number of times (ideally 1). If it is still necessary to run a bytecode to get a reference to that library object, that's alright, just ensure that it is cached.

lib/Runtime/Library/JsBuiltIn/JsBuiltIn.js
```diff
@@ -0,0 +1,69 @@
...
+
+(function (intrinsic) {
+    var platform = intrinsic.JsBuiltIn;
+
+    let FunctionsEnum = {
+        ArrayIndexOf: { className: "Array", methodName: "indexOf", argumentsCount: 1 },
+    };
+
+    platform.registerFunction(FunctionsEnum.ArrayIndexOf, function (searchElement, fromIndex) {
+        // https://www.ecma-international.org/ecma-262/7.0/index.html#sec-array.prototype.indexof
```
dilijev: >// https://www.ecma-international.org/ecma-262/7.0/index.html#sec-array.prototype.indexof [](start = 8, length = 89)

This is essentially an outdated spec at this point. ES 2017 has been all-but-ratified at this point. Double check with @bterlson but I'd recommend implementing against the latest version of the spec ES 2017.

dilijev: change to "ECMAScript 2017 #sec-array.prototype.indexof" LGTM.

---
In reply to: [116853125](https://github.com/Microsoft/ChakraCore/pull/2898#discussion_r116853125) [](ancestors = 116853125)

veler:

veler:
@dotnet-bot test all please

veler:
@dotnet-bot test please

veler:
@dotnet-bot test Windows all please

veler:

veler:
@dotnet-bot test all please

dilijev:
@veler did you mean:
@dotnet-bot test this please

dilijev:
But it does looks like you have a test failure: 
`23:16:01 disable_jit\rl.results.log:DebuggerCommon (ES6_proto_invalidation.js ( -debuglaunch -dbgbaseline:ES6_proto_invalidation.js.dbg.baseline) exe) -- failed : exec time=0`

veler:
I know I'm investigating it :)

---
In reply to: [303529497](https://github.com/Microsoft/ChakraCore/pull/2898#issuecomment-303529497) [](ancestors = 303529497)

dilijev:
new change LGTM

dilijev:
🎉

PR #2902, opened 2017-05-02
Fixing the Bytecode testing flag in code coverage builds

aneeshdk:
The bytecode testing flag was not enabled for code coverage builds


aneeshdk:
@MikeHolman Can you please take a look?

PR #2906, opened 2017-05-03
Some small improvements to wasm exception messages.

Penguinwizzard:
Changes the message when a function fails to link to name the function
and print out its signature, and fixes a typo in a format specifier.

### DIFF
--- a/lib/Parser/rterrors.h
+++ b/lib/Parser/rterrors.h
@@ -391,7 +391,7 @@ RT_ERROR_MSG(JSERR_OutOfBoundString, 7020, "", "String length is out of bound", 
 RT_ERROR_MSG(JSERR_InvalidIterableObject, 7021, "%s : Invalid iterable object", "Invalid iterable object", kjstTypeError, 0)
 RT_ERROR_MSG(JSERR_InvalidIteratorObject, 7022, "%s : Invalid iterator object", "Invalid iterator object", kjstTypeError, 0)
 RT_ERROR_MSG(WASMERR_NeedResponse, 7023, "%s is not a Reponse", "Response expected", kjstTypeError, 0)
-RT_ERROR_MSG(WASMERR_LinkSignatureMismatch, 7024, "Cannot link import %s in link table called with invalid signature", "Function called with invalid signature", kjstWebAssemblyRuntimeError, 0)
+RT_ERROR_MSG(WASMERR_LinkSignatureMismatch, 7024, "Cannot link import %s in link table due to a signature mismatch", "Function called with invalid signature", kjstWebAssemblyRuntimeError, 0)
 
 // Wabt Errors
 RT_ERROR_MSG(WABTERR_WabtError, 7200, "%s", "Wabt Error.", kjstTypeError, 0)
--- a/lib/Parser/rterrors.h
+++ b/lib/Parser/rterrors.h
@@ -391,6 +391,7 @@ RT_ERROR_MSG(JSERR_OutOfBoundString, 7020, "", "String length is out of bound", 
 RT_ERROR_MSG(JSERR_InvalidIterableObject, 7021, "%s : Invalid iterable object", "Invalid iterable object", kjstTypeError, 0)
 RT_ERROR_MSG(JSERR_InvalidIteratorObject, 7022, "%s : Invalid iterator object", "Invalid iterator object", kjstTypeError, 0)
 RT_ERROR_MSG(WASMERR_NeedResponse, 7023, "%s is not a Reponse", "Response expected", kjstTypeError, 0)
+RT_ERROR_MSG(WASMERR_LinkSignatureMismatch, 7024, "Cannot link import %s in link table called with invalid signature", "Function called with invalid signature", kjstWebAssemblyRuntimeError, 0)
 
 // Wabt Errors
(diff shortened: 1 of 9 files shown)


lib/Parser/rterrors.h
```diff
@@ -390,6 +390,7 @@ RT_ERROR_MSG(WASMERR_WasmLinkError, 7019, "%s", "Linking failed.", kjstWebAssemb
 RT_ERROR_MSG(JSERR_OutOfBoundString, 7020, "", "String length is out of bound", kjstRangeError, 0)
 RT_ERROR_MSG(JSERR_InvalidIterableObject, 7021, "%s : Invalid iterable object", "Invalid iterable object", kjstTypeError, 0)
 RT_ERROR_MSG(JSERR_InvalidIteratorObject, 7022, "%s : Invalid iterator object", "Invalid iterator object", kjstTypeError, 0)
```
Penguinwizzard: Note: I'm probably going to send in another PR shortly to move these JSERR messages to the right block.

lib/Runtime/Library/WebAssemblyInstance.cpp
```diff
@@ -300,7 +300,14 @@ void WebAssemblyInstance::LoadImports(
                 AsmJsScriptFunction* func = AsmJsScriptFunction::FromVar(prop);
                 if (!wasmModule->GetWasmFunctionInfo(counter)->GetSignature()->IsEquivalent(func->GetSignature()))
                 {
-                    JavascriptError::ThrowWebAssemblyLinkError(ctx, WASMERR_SignatureMismatch);
+                    char16 temp[1024] = { 0 };
```
agarwal-sandeep: Add a test case exercising this code?

lib/Parser/rterrors.h
Cellule: Thanks!

lib/Runtime/Library/WebAssemblyInstance.cpp
Cellule: Agreed, also, the signatures can be very long, are we sure the buffers are big enough ?

lib/WasmReader/WasmSignature.cpp
```diff
@@ -181,53 +181,31 @@ WasmSignature::FromIDL(WasmSignatureIDL* sig)
...
-                break;
-            case WasmTypes::WasmType::F32:
-                Output::Print(_u("f32"));
-                break;
-            case WasmTypes::WasmType::F64:
-                Output::Print(_u("f64"));
-                break;
+            numwritten += swprintf_s(out+numwritten, maxlen-numwritten, _u(", "));
         }
+        numwritten += swprintf_s(out+numwritten, maxlen-numwritten, _u("%ls"), WasmTypes::GetTypeName(this->GetParam(i)));
```
agarwal-sandeep: >numwritten += swprintf_s(out+numwritten, maxlen-numwritten, _u("%ls"), WasmTypes::GetTypeName(this->GetParam(i))); [](start = 8, length = 114)

Add another test where the combined string length of parameters is more than 511

lib/WasmReader/WasmSignature.cpp
```diff
@@ -181,53 +181,31 @@ WasmSignature::FromIDL(WasmSignatureIDL* sig)
...
-    }
+    numwritten += swprintf_s(out+numwritten, maxlen-numwritten, _u(")->%ls"), WasmTypes::GetTypeName(this->GetResultType()));
+    return numwritten;
+}
+
+void
+WasmSignature::Dump()
+{
+#ifdef ENABLE_DEBUG_CONFIG_OPTIONS
+    char16 buf[128] = { 0 };
```
Cellule: Might want to make this `512` like in `WebAssemblyInstance::LoadImports`

lib/Runtime/Library/WebAssemblyInstance.cpp
Penguinwizzard: > Add a test case exercising this code?

We do test this code, which is why the PR is currently failing CI - I need to update the baselines.

> The signatures can be very long, are we sure the buffers are big enough ?

I can change to heap allocation and growing the buffer when needed, but that complicates the lifetime tracking slightly.

Cellule: Actually, right now we fail the CI because of a prefast warning.
I don't think we have any tests that actually prints this. At most I think we catch the exception. Then again, I don't know how interesting it is to have a test just printing an error message.

As for size, we just have to make sure swprintf_s correctly crops the message and we don't have any errors if it becomes too long.

agarwal-sandeep: If we already have a test case which catch this exception we should modify it print the message as it will cover this additional code as well. It is always a good idea to cover maximum code through unittests.

Penguinwizzard: I see a failure to match the baseline on test\wasm\api.js due to it printing this message, but I can add some more cases as well.

lib/WasmReader/WasmSignature.cpp
Penguinwizzard: I've mirrored the value for now, but I'll probably change the interface again to make it able to handle arbitrarily long signatures.

PR #2907, opened 2017-05-03
AsmJs Templatized Jit fixes

Cellule:
- Use the common function to determine if it's time to change from TemplatizedJit to FullJit.
- Pass the local slot and args addresses instead of calculating it. It was error prone and had some weird compiler bugs when changing locals in that functions.
- Respect the `-debugbreak:#` flag for templatized jit

This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/2907)


Cellule:
ping

MikeHolman:
:shipit:

PR #2908, opened 2017-05-04
Exposing the diagnostic APIs through static methods

aneeshdk:
Currently the methods debugEval, getStackTrace and editSource are part of the diagnostics object. This changelist provides static API to access these methods from the host.


aneeshdk:
@akroshg @curtisman Can you please take a look?

agarwal-sandeep:
LGTM

PR #2909, opened 2017-05-04
DebugBreak flag as phase

Cellule:
Add flags `-on:debugbreak:1.0` to match more closely how we usually reference functions in flags.

This is something we talked about years ago and I ended up working around that code so I figured I'd fix it while I'm at it.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/2909)


Cellule:
@Microsoft/chakra-developers Can someone review this please ? It should be pretty trivial change

lib/Runtime/Language/i386/AsmJsJitTemplate.cpp
```diff
@@ -1507,6 +1507,16 @@ namespace Js
             int32 stackSize = asmInfo->GetTotalSizeinBytes();
             stackSize = ::Math::Align<int32>(stackSize, 8);
 
+#if DBG_DUMP
+            if (Js::Configuration::Global.flags.DebugBreak.Contains(funcBody->GetFunctionNumber()) ||
+                PHASE_ON(DebugBreakPhase, funcBody))
+            {
+                // int 3
+                *buffer++ = 0xCC;
```
Penguinwizzard: Should we add an Int::EncodeInstruction for this?

Cellule: I thought about it, but I don't think it's worth it since we will not use int instruction much.
Also, 0xCC is a special case for `int 3` on x86
image

LouisLaf: It is debug only code.  No need to encode, plus it isn't user controllable.

---
In reply to: [114687392](https://github.com/Microsoft/ChakraCore/pull/2909#discussion_r114687392) [](ancestors = 114687392)

LouisLaf:
:shipit:

PR #2913, opened 2017-05-04
Dynamic Module Import

suwc:
This PR adds support for `import()` dynamic module import sematic.

Per https://github.com/tc39/proposal-dynamic-import:

"A call to `import(specifier)` returns a promise for the module namespace object of the requested module, which is created after fetching, instantiating, and evaluating all of the module's dependencies, as well as the module itself.

"Here specifier will be interpreted the same way as in an import declaration (i.e., the same strings will work in both places). However, while specifier is a string it is not necessarily a string literal; thus code like import(`./language-packs/${navigator.language}.js`) will work—something impossible to accomplish with the usual import declarations.

"`import()` is proposed to work in both scripts and modules. This gives script code an easy asynchronous entry point into the module world, allowing it to start running module code."

This PR includes following changes:
	- Update parser and bytecode generator to support `import()` sematic in module and in script
	- Add new bytecode `ImportCall`
	- Add runtime function for `import()` that:
		○ Uses caller from stack to look up module record or source context that are associated with the module or script from which `import()` is called
		○ Requests host to load target module source file (gets module record in return)
		○ Creates promise unless the module record has one
		○ Resolves/rejects promise if appropriates
		○ Returns promise
	- Add new host callback (`FetchImportedModuleFromScript`) for fetching imported module from script (accepts source context)
	- Add `promiseCapability` field to module record class
	- Update `SourceTextModuleRecord`'s methods to accept callback from host and to handle dynamically imported module and its promise capability
	- Update exception checks and assertions to cover new usage scenario of importing and evaluating module code with active script
Add unit tests for dynamic import functionality

### DIFF
--- a/bin/ch/WScriptJsrt.cpp
+++ b/bin/ch/WScriptJsrt.cpp
@@ -304,14 +304,22 @@ JsErrorCode WScriptJsrt::InitializeModuleInfo(JsValueRef specifier, JsModuleReco
 {
     JsErrorCode errorCode = JsNoError;
     errorCode = ChakraRTInterface::JsSetModuleHostInfo(moduleRecord, JsModuleHostInfo_FetchImportedModuleCallback, (void*)WScriptJsrt::FetchImportedModule);
+
     if (errorCode == JsNoError)
     {
-        errorCode = ChakraRTInterface::JsSetModuleHostInfo(moduleRecord, JsModuleHostInfo_NotifyModuleReadyCallback, (void*)WScriptJsrt::NotifyModuleReadyCallback);
+        errorCode = ChakraRTInterface::JsSetModuleHostInfo(moduleRecord, JsModuleHostInfo_FetchImportedModuleFromScriptCallback, (void*)WScriptJsrt::FetchImportedModuleFromScript);
+
+        if (errorCode == JsNoError)
+        {
+            errorCode = ChakraRTInterface::JsSetModuleHostInfo(moduleRecord, JsModuleHostInfo_NotifyModuleReadyCallback, (void*)WScriptJsrt::NotifyModuleReadyCallback);
+
+            if (errorCode == JsNoError)
+            {
+                errorCode = ChakraRTInterface::JsSetModuleHostInfo(moduleRecord, JsModuleHostInfo_HostDefined, specifier);
+            }
+        }
     }
-    if (errorCode == JsNoError)
-    {
-        errorCode = ChakraRTInterface::JsSetModuleHostInfo(moduleRecord, JsModuleHostInfo_HostDefined, specifier);
-    }
+
     IfJsrtErrorFailLogAndRetErrorCode(errorCode);
     return JsNoError;
 }
@@ -351,9 +359,10 @@ JsErrorCode WScriptJsrt::LoadModuleFromString(LPCSTR fileName, LPCSTR fileConten
     JsValueRef errorObject = JS_INVALID_REFERENCE;
 
     // ParseModuleSource is sync, while additional fetch & evaluation are async.
+    unsigned int fileContentLength = (fileContent == nullptr) ? 0 : (unsigned int)strlen(fileContent);
     errorCode = ChakraRTInterface::JsParseModuleSource(requestModule, dwSourceCookie, (LPBYTE)fileContent,
-        (unsigned int)strlen(fileContent), JsParseModuleSourceFlags_DataIsUTF8, &errorObject);
-    if ((errorCode != JsNoError) && errorObject != JS_INVALID_REFERENCE)
+        fileContentLength, JsParseModuleSourceFlags_DataIsUTF8, &errorObject);
+    if ((errorCode != JsNoError) && errorObject != JS_INVALID_REFERENCE && fileContent != nullptr)
     {
         ChakraRTInterface::JsSetException(errorObject);
         return errorCode;
@@ -875,8 +884,22 @@ bool WScriptJsrt::Initialize()
     IfJsrtErrorFail(CreatePropertyIdFromString("console", &consoleName), false);
     IfJsrtErrorFail(ChakraRTInterface::JsSetProperty(global, consoleName, console, true), false);
 
+    IfJsrtErrorFail(InitializeModuleCallbacks(), false);
+
 Error:
     return hr == S_OK;
+}
+
+JsErrorCode WScriptJsrt::InitializeModuleCallbacks()
+{
+    JsModuleRecord moduleRecord = JS_INVALID_REFERENCE;
+    JsErrorCode errorCode = ChakraRTInterface::JsInitializeModuleRecord(nullptr, nullptr, &moduleRecord);
+    if (errorCode == JsNoError)
+    {
+        errorCode = InitializeModuleInfo(nullptr, moduleRecord);
+    }
+
+    return errorCode;
 }
 
 bool WScriptJsrt::Uninitialize()
@@ -1217,7 +1240,12 @@ HRESULT WScriptJsrt::ModuleMessage::Call(LPCSTR fileName)
             hr = Helpers::LoadScriptFromFile(specifierStr.GetString(), fileContent);
             if (FAILED(hr))
             {
-                fprintf(stderr, "Couldn't load file.\n");
+                if (!HostConfigFlags::flags.AsyncModuleLoadIsEnabled)
+                {
+                    fprintf(stderr, "Couldn't load file.\n");
+                }
+
+                LoadScript(nullptr, specifierStr.GetString(), nullptr, "module", true, WScriptJsrt::FinalizeFree);
             }
             else
             {
@@ -1228,12 +1256,8 @@ HRESULT WScriptJsrt::ModuleMessage::Call(LPCSTR fileName)
     return errorCode;
 }
 
-// Callback from chakracore to fetch dependent module. In the test harness,
-// we are not doing any translation, just treat the specifier as fileName.
-// While this call will come back directly from ParseModuleSource, the additional
-// task are treated as Promise that will be executed later.
-JsErrorCode WScriptJsrt::FetchImportedModule(_In_ JsModuleRecord referencingModule,
-    _In_ JsValueRef specifier, _Outptr_result_maybenull_ JsModuleRecord* dependentModuleRecord)
+JsErrorCode WScriptJsrt::FetchImportedModuleHelper(JsModuleRecord referencingModule,
+    JsValueRef specifier, __out JsModuleRecord* dependentModuleRecord)
 {
     JsModuleRecord moduleRecord = JS_INVALID_REFERENCE;
     AutoString specifierStr;
@@ -1265,6 +1289,27 @@ JsErrorCode WScriptJsrt::FetchImportedModule(_In_ JsModuleRecord referencingModu
         *dependentModuleRecord = moduleRecord;
     }
     return errorCode;
+}
+
+// Callback from chakracore to fetch dependent module. In the test harness,
+// we are not doing any translation, just treat the specifier as fileName.
+// While this call will come back directly from ParseModuleSource, the additional
+// task are treated as Promise that will be executed later.
+JsErrorCode WScriptJsrt::FetchImportedModule(_In_ JsModuleRecord referencingModule,
+    _In_ JsValueRef specifier, _Outptr_result_maybenull_ JsModuleRecord* dependentModuleRecord)
+{
+    return FetchImportedModuleHelper(referencingModule, specifier, dependentModuleRecord);
+}
+
+// Callback from chakracore to fetch module dynamically during runtime. In the test harness,
+// we are not doing any translation, just treat the specifier as fileName.
+// While this call will come back directly from runtime script or module code, the additional
+// task can be scheduled asynchronously that executed later.
+JsErrorCode WScriptJsrt::FetchImportedModuleFromScript(_In_ JsSourceContext dwReferencingSourceContext,
+    _In_ JsValueRef specifier, _Outptr_result_maybenull_ JsModuleRecord* dependentModuleRecord)
+{
+    // ch.exe assumes all imported source files are located at .
+    return FetchImportedModuleHelper(nullptr, specifier, dependentModuleRecord);
 }
 
 // Callback from chakraCore when the module resolution is finished, either successfuly or unsuccessfully.
--- a/lib/Jsrt/ChakraCore.h
+++ b/lib/Jsrt/ChakraCore.h
@@ -38,7 +38,8 @@ typedef enum JsModuleHostInfoKind
     JsModuleHostInfo_Exception = 0x01,
     JsModuleHostInfo_HostDefined = 0x02,
     JsModuleHostInfo_NotifyModuleReadyCallback = 0x3,
-    JsModuleHostInfo_FetchImportedModuleCallback = 0x4
+    JsModuleHostInfo_FetchImportedModuleCallback = 0x4,
+    JsModuleHostInfo_FetchImportedModuleFromScriptCallback = 0x5
 } JsModuleHostInfoKind;
 
 /// <summary>
@@ -65,6 +66,21 @@ typedef JsErrorCode(CHAKRA_CALLBACK * FetchImportedModuleCallBack)(_In_ JsModule
 /// holds the exception. Otherwise the referencingModule is ready and the host should schedule execution afterwards.
 /// </remarks>
 /// <param name="referencingModule">The referencing module that have finished running ModuleDeclarationInstantiation step.</param>
+/// <param name="exceptionVar">If nullptr, the module is successfully initialized and host should queue the execution job
+///                           otherwise it's the exception object.</param>
+/// <returns>
+///     true if the operation succeeded, false otherwise.
+/// </returns>
+typedef JsErrorCode(CHAKRA_CALLBACK * FetchImportedModuleFromScriptCallBack)(_In_ JsSourceContext dwReferencingSourceContext, _In_ JsValueRef specifier, _Outptr_result_maybenull_ JsModuleRecord* dependentModuleRecord);
+
+/// <summary>
+///     User implemented callback to get notification when the module is ready.
+/// </summary>
+/// <remarks>
+/// Notify the host after ModuleDeclarationInstantiation step (15.2.1.1.6.4) is finished. If there was error in the process, exceptionVar
+/// holds the exception. Otherwise the referencingModule is ready and the host should schedule execution afterwards.
+/// </remarks>
+/// <param name="dwReferencingSourceContext">The referencing script that calls import()</param>
 /// <param name="exceptionVar">If nullptr, the module is successfully initialized and host should queue the execution job
 ///                           otherwise it's the exception object.</param>
 /// <returns>
--- a/lib/Jsrt/Core/JsrtContextCore.cpp
+++ b/lib/Jsrt/Core/JsrtContextCore.cpp
@@ -102,6 +102,25 @@ void JsrtContextCore::OnScriptLoad(Js::JavascriptFunction * scriptFunction, Js::
 
 HRESULT ChakraCoreHostScriptContext::FetchImportedModule(Js::ModuleRecordBase* referencingModule, LPCOLESTR specifier, Js::ModuleRecordBase** dependentModuleRecord)
 {
+    return FetchImportedModuleHelper(
+        [=](Js::JavascriptString *specifierVar, JsModuleRecord *dependentRecord) -> JsErrorCode
+        {
+            return fetchImportedModuleCallback(referencingModule, specifierVar, dependentRecord);
+        }, specifier, dependentModuleRecord);
+}
+
+HRESULT ChakraCoreHostScriptContext::FetchImportedModuleFromScript(JsSourceContext dwReferencingSourceContext, LPCOLESTR specifier, Js::ModuleRecordBase** dependentModuleRecord)
+{
+    return FetchImportedModuleHelper(
+        [=](Js::JavascriptString *specifierVar, JsModuleRecord *dependentRecord) -> JsErrorCode
+    {
+        return fetchImportedModuleFromScriptCallback(dwReferencingSourceContext, specifierVar, dependentRecord);
+    }, specifier, dependentModuleRecord);
+}
+
+template<typename Fn>
+HRESULT ChakraCoreHostScriptContext::FetchImportedModuleHelper(Fn fetch, LPCOLESTR specifier, Js::ModuleRecordBase** dependentModuleRecord)
+{
     if (fetchImportedModuleCallback == nullptr)
     {
         return E_INVALIDARG;
@@ -110,7 +129,7 @@ HRESULT ChakraCoreHostScriptContext::FetchImportedModule(Js::ModuleRecordBase* r
     JsModuleRecord dependentRecord = JS_INVALID_REFERENCE;
     {
         AUTO_NO_EXCEPTION_REGION;
-        JsErrorCode errorCode = fetchImportedModuleCallback(referencingModule, specifierVar, &dependentRecord);
+        JsErrorCode errorCode = fetch(specifierVar, &dependentRecord);
         if (errorCode == JsNoError)
         {
             *dependentModuleRecord = static_cast<Js::ModuleRecordBase*>(dependentRecord);
--- a/lib/Jsrt/Core/JsrtContextCore.h
+++ b/lib/Jsrt/Core/JsrtContextCore.h
@@ -168,6 +168,7 @@ public:
     }
 
     HRESULT FetchImportedModule(Js::ModuleRecordBase* referencingModule, LPCOLESTR specifier, Js::ModuleRecordBase** dependentModuleRecord) override;
+    HRESULT FetchImportedModuleFromScript(JsSourceContext dwReferencingSourceContext, LPCOLESTR specifier, Js::ModuleRecordBase** dependentModuleRecord) override;
 
     HRESULT NotifyHostAboutModuleReady(Js::ModuleRecordBase* referencingModule, Js::Var exceptionVar) override;
 
@@ -176,6 +177,9 @@ public:
 
     void SetFetchImportedModuleCallback(FetchImportedModuleCallBack fetchCallback) { this->fetchImportedModuleCallback = fetchCallback ; }
     FetchImportedModuleCallBack GetFetchImportedModuleCallback() const { return this->fetchImportedModuleCallback; }
+
+    void SetFetchImportedModuleFromScriptCallback(FetchImportedModuleFromScriptCallBack fetchCallback) { this->fetchImportedModuleFromScriptCallback = fetchCallback; }
+    FetchImportedModuleFromScriptCallBack GetFetchImportedModuleFromScriptCallback() const { return this->fetchImportedModuleFromScriptCallback; }
 
 #if DBG_DUMP || defined(PROFILE_EXEC) || defined(PROFILE_MEM)
     void EnsureParentInfo(Js::ScriptContext* scriptContext = NULL) override
@@ -186,6 +190,9 @@ public:
 #endif
 
 private:
+    template<typename Fn>
+    HRESULT FetchImportedModuleHelper(Fn fetch, LPCOLESTR specifier, Js::ModuleRecordBase** dependentModuleRecord);
     FetchImportedModuleCallBack fetchImportedModuleCallback;
+    FetchImportedModuleFromScriptCallBack fetchImportedModuleFromScriptCallback;
     NotifyModuleReadyCallback notifyModuleReadyCallback;
 };
--- a/lib/Parser/Parse.cpp
+++ b/lib/Parser/Parse.cpp
@@ -2435,11 +2435,36 @@ bool Parser::IsImportOrExportStatementValidHere()
         && this->m_tryCatchOrFinallyDepth == 0;
 }
 
+template<bool buildAST> ParseNodePtr Parser::ParseImportCall()
+{
+    m_pscan->Scan();
+    ParseNodePtr specifier = ParseExpr<buildAST>(koplCma, nullptr, /* fAllowIn */FALSE, /* fAllowEllipsis */FALSE);
+    if (m_token.tk != tkRParen)
+    {
+        Error(ERRnoRparen);
+    }
+
+    m_pscan->Scan();
+    return CreateCallNode(knopCall, CreateNodeWithScanner<knopImport>(), specifier);
+}
+
 template<bool buildAST>
-ParseNodePtr Parser::ParseImportDeclaration()
+ParseNodePtr Parser::ParseImport()
 {
     Assert(m_scriptContext->GetConfig()->IsES6ModuleEnabled());
     Assert(m_token.tk == tkIMPORT);
+
+    RestorePoint parsedImport;
+    m_pscan->Capture(&parsedImport);
+    m_pscan->Scan();
+
+    // import()
+    if (m_token.tk == tkLParen)
+    {
+        return ParseImportCall<buildAST>();
+    }
+
+    m_pscan->SeekTo(parsedImport);
 
     if (!IsImportOrExportStatementValidHere())
     {
@@ -3220,6 +3245,23 @@ LFunction :
         if (m_scriptContext->GetConfig()->IsES6ClassAndExtendsEnabled())
         {
             pnode = ParseSuper<buildAST>(pnode, !!fAllowCall);
+        }
+        else
+        {
+            goto LUnknown;
+        }
+        break;
+
+    case tkIMPORT:
+        if (m_scriptContext->GetConfig()->IsES6ModuleEnabled())
+        {
+            m_pscan->Scan();
+            if (m_token.tk == tkLParen)
+            {
+                return ParseImportCall<buildAST>();
+            }
+
+            Error(ERRnoLparen);
         }
         else
         {
@@ -10112,12 +10154,7 @@ LGetJumpStatement:
         goto LNeedTerminator;
 
     case tkIMPORT:
-        if (!(m_grfscr & fscrIsModuleCode))
-        {
-            goto LDefaultToken;
-        }
-
-        pnode = ParseImportDeclaration<buildAST>();
+        pnode = ParseImport<buildAST>();
 
         goto LNeedTerminator;
 
--- a/lib/Runtime/ByteCode/ByteCodeEmitter.cpp
+++ b/lib/Runtime/ByteCode/ByteCodeEmitter.cpp
@@ -10803,6 +10803,15 @@ void Emit(ParseNode *pnode, ByteCodeGenerator *byteCodeGenerator, FuncInfo *func
         {
             byteCodeGenerator->EmitSuperCall(funcInfo, pnode, fReturnValue);
         }
+        else if (pnode->sxCall.pnodeTarget->nop == knopImport)
+        {
+            ParseNodePtr args = pnode->sxCall.pnodeArgs;
+            Assert(CountArguments(args) == 2); // import() takes one argument
+            Emit(args, byteCodeGenerator, funcInfo, false);
+            funcInfo->ReleaseLoc(args);
+            funcInfo->AcquireLoc(pnode);
+            byteCodeGenerator->Writer()->Reg2(Js::OpCode::ImportCall, pnode->location, args->location);
+        }
         else
         {
             if (pnode->sxCall.isApplyCall && funcInfo->GetApplyEnclosesArgs())
--- a/lib/Runtime/Language/JavascriptOperators.cpp
+++ b/lib/Runtime/Language/JavascriptOperators.cpp
@@ -9639,6 +9639,112 @@ CommonNumber:
         return superCtor;
     }
 
+    Var JavascriptOperators::OP_ImportCall(__in JavascriptFunction *function, __in Var specifier, __in ScriptContext* scriptContext)
+    {
+        ModuleRecordBase *moduleRecordBase = nullptr;
+        SourceTextModuleRecord *moduleRecord = nullptr;
+
+        FunctionBody* parentFuncBody = function->GetFunctionBody();
+        JavascriptString *specifierString = nullptr;
+
+        try
+        {
+            specifierString = JavascriptConversion::ToString(specifier, scriptContext);
+        }
+        catch (const JavascriptException &err)
+        {
+            Var errorObject = err.GetAndClear()->GetThrownObject(scriptContext);
+            AssertMsg(errorObject != nullptr, "OP_ImportCall: null error object thrown by ToString(specifier)");
+            if (errorObject != nullptr)
+            {
+                return SourceTextModuleRecord::ResolveOrRejectDynamicImportPromise(false, errorObject, scriptContext);
+            }
+
+            Throw::InternalError();
+        }
+
+        DWORD_PTR dwReferencingSourceContext = parentFuncBody->GetHostSourceContext();
+        if (!parentFuncBody->IsES6ModuleCode() && dwReferencingSourceContext == Js::Constants::NoHostSourceContext)
+        {
+            // import() called from eval
+            if (parentFuncBody->GetUtf8SourceInfo()->GetCallerUtf8SourceInfo() == nullptr)
+            {
+                JavascriptError *error = scriptContext->GetLibrary()->CreateError();
+                JavascriptError::SetErrorMessageProperties(error, E_FAIL, _u("Unable to locate active script or module that calls import()"), scriptContext);
+                return SourceTextModuleRecord::ResolveOrRejectDynamicImportPromise(false, error, scriptContext);
+            }
+
+            dwReferencingSourceContext = parentFuncBody->GetUtf8SourceInfo()->GetCallerUtf8SourceInfo()->GetSourceContextInfo()->dwHostSourceContext;
+
+            if (dwReferencingSourceContext == Js::Constants::NoHostSourceContext)
+            {
+                // Walk the call stack if caller function is neither module code nor having host source context
+
+                JavascriptFunction* caller = nullptr;
+                Js::JavascriptStackWalker walker(scriptContext);
+                walker.GetCaller(&caller);
+
+                do
+                {
+                    if (walker.GetCaller(&caller) && caller != nullptr && caller->IsScriptFunction())
+                    {
+                        parentFuncBody = caller->GetFunctionBody();
+                        dwReferencingSourceContext = parentFuncBody->GetHostSourceContext();
+                    }
+                    else
+                    {
+                        JavascriptError *error = scriptContext->GetLibrary()->CreateError();
+                        JavascriptError::SetErrorMessageProperties(error, E_FAIL, _u("Unable to locate active script or module that calls import()"), scriptContext);
+                        return SourceTextModuleRecord::ResolveOrRejectDynamicImportPromise(false, error, scriptContext);
+                    }
+
+                } while (!parentFuncBody->IsES6ModuleCode() && dwReferencingSourceContext == Js::Constants::NoHostSourceContext);
+            }
+        }
+
+        LPCOLESTR moduleName = specifierString->GetSz();
+        HRESULT hr = 0;
+
+        if (parentFuncBody->IsES6ModuleCode())
+        {
+            SourceTextModuleRecord *referenceModuleRecord = parentFuncBody->GetScriptContext()->GetLibrary()->GetModuleRecord(parentFuncBody->GetModuleID());
+            BEGIN_LEAVE_SCRIPT(scriptContext);
+            BEGIN_TRANSLATE_TO_HRESULT(static_cast<ExceptionType>(ExceptionType_OutOfMemory | ExceptionType_StackOverflow));
+            hr = scriptContext->GetHostScriptContext()->FetchImportedModule(referenceModuleRecord, moduleName, &moduleRecordBase);
+            END_TRANSLATE_EXCEPTION_TO_HRESULT(hr);
+            END_LEAVE_SCRIPT(scriptContext);
+        }
+        else
+        {
+            Assert(dwReferencingSourceContext != Js::Constants::NoHostSourceContext);
+            BEGIN_LEAVE_SCRIPT(scriptContext);
+            BEGIN_TRANSLATE_TO_HRESULT(static_cast<ExceptionType>(ExceptionType_OutOfMemory | ExceptionType_StackOverflow));
+            hr = scriptContext->GetHostScriptContext()->FetchImportedModuleFromScript(dwReferencingSourceContext, moduleName, &moduleRecordBase);
+            END_TRANSLATE_EXCEPTION_TO_HRESULT(hr);
+            END_LEAVE_SCRIPT(scriptContext);
+        }
+
+        if (FAILED(hr))
+        {
+            Js::JavascriptError *error = scriptContext->GetLibrary()->CreateURIError();
+            JavascriptError::SetErrorMessageProperties(error, hr, moduleName, scriptContext);
+            return SourceTextModuleRecord::ResolveOrRejectDynamicImportPromise(false, error, scriptContext);
+        }
+
+        moduleRecord = SourceTextModuleRecord::FromHost(moduleRecordBase);
+
+        if (moduleRecord->GetErrorObject() != nullptr)
+        {
+            return SourceTextModuleRecord::ResolveOrRejectDynamicImportPromise(false, moduleRecord->GetErrorObject(), scriptContext, moduleRecord);
+        }
+        else if (moduleRecord->WasEvaluated())
+        {
+            return SourceTextModuleRecord::ResolveOrRejectDynamicImportPromise(true, moduleRecord->GetNamespace(), scriptContext, moduleRecord);
+        }
+
+        return moduleRecord->PostProcessDynamicModuleImport();
+    }
+
     Var JavascriptOperators::ScopedLdHomeObjFuncObjHelper(Var scriptFunction, Js::PropertyId propertyId, ScriptContext * scriptContext)
     {
         ScriptFunction *instance = ScriptFunction::FromVar(scriptFunction);
--- a/lib/Runtime/Language/SourceTextModuleRecord.cpp
+++ b/lib/Runtime/Language/SourceTextModuleRecord.cpp
@@ -7,6 +7,7 @@
 #include "Types/SimpleDictionaryPropertyDescriptor.h"
 #include "Types/SimpleDictionaryTypeHandler.h"
 #include "ModuleNamespace.h"
+#include "Library/JavascriptPromise.h"
 
 namespace Js
 {
@@ -36,15 +37,14 @@ namespace Js
         resolvedExportMap(nullptr),
         wasParsed(false),
         wasDeclarationInitialized(false),
-#if DBG
         parentsNotified(false),
-#endif
         isRootModule(false),
         hadNotifyHostReady(false),
         localExportSlots(nullptr),
         numPendingChildrenModule(0),
         moduleId(InvalidModuleIndex),
         localSlotCount(InvalidSlotCount),
+        promise(nullptr),
         localExportCount(0)
     {
         namespaceRecord.module = this;
@@ -74,15 +74,11 @@ namespace Js
         localExportRecordList = nullptr;
         indirectExportRecordList = nullptr;
         starExportRecordList = nullptr;
-        childrenModuleSet = nullptr;
         parentModuleList = nullptr;
         if (!isShutdown)
         {
-            if (parser != nullptr)
-            {
-                AllocatorDelete(ArenaAllocator, scriptContext->GeneralAllocator(), parser);
-                parser = nullptr;
-            }
+            AdeleteUnlessNull(scriptContext->GeneralAllocator(), parser);
+            AdeleteUnlessNull(scriptContext->GeneralAllocator(), childrenModuleSet);
         }
     }
 
@@ -91,6 +87,7 @@ namespace Js
         OUTPUT_TRACE_DEBUGONLY(Js::ModulePhase, _u("ParseSource(%s)\n"), this->GetSpecifierSz());
         Assert(!wasParsed);
         Assert(parser == nullptr);
+        Assert(exceptionVar != nullptr);
         HRESULT hr = NOERROR;
         ScriptContext* scriptContext = GetScriptContext();
         CompileScriptException se;
@@ -98,6 +95,9 @@ namespace Js
         *exceptionVar = nullptr;
         if (!scriptContext->GetConfig()->IsES6ModuleEnabled())
         {
+            JavascriptError *pError = scriptContext->GetLibrary()->CreateError();
+            JavascriptError::SetErrorMessageProperties(pError, hr, _u("ES6Module not supported"), scriptContext);
+            *exceptionVar = pError;
             return E_NOTIMPL;
         }
         // Host indicates that the current module failed to load.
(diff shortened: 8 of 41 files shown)


suwc:
@Microsoft/chakra-es @Microsoft/chakra-developers, please help code review. Thanks!

lib/Jsrt/ChakraCore.h
```diff
@@ -70,6 +71,21 @@ typedef JsErrorCode(CHAKRA_CALLBACK * FetchImportedModuleCallBack)(_In_ JsModule
 /// <returns>
 ///     true if the operation succeeded, false otherwise.
 /// </returns>
+typedef JsErrorCode(CHAKRA_CALLBACK * FetchImportedModuleFromScriptCallBack)(_In_ DWORD_PTR dwReferencingSourceContext, _In_ JsValueRef specifier, _Outptr_result_maybenull_ JsModuleRecord* dependentModuleRecord);
```
agarwal-sandeep: >DWORD_PTR [](start = 82, length = 9)

Use JsSourceContext instead of DWORD_PTR

test/es6/dynamic-module-functionality.js
```diff
@@ -0,0 +1,359 @@
...
+
+var tests = [
+    {
+        name: "Runtime evaluation of module specifier",
+        body: function () {
+            let functionBody =
+                `testDynamicImport(
+                    ()=>{
+                        var getName = ()=>{ return 'ModuleSimpleExport'; };
+                        return import( getName() + '.js');
```
agarwal-sandeep: >import [](start = 31, length = 6)

What if a non-string is passed to import? like import() or import({}) etc. Can we add test cases covering error scenarios?

lib/Runtime/Language/JavascriptOperators.cpp
```diff
@@ -9639,6 +9639,82 @@ namespace Js
...
+                ASYNC_HOST_OPERATION_START(scriptContext->GetThreadContext());
+                hr = scriptContext->GetHostScriptContext()->FetchImportedModuleFromScript(dwReferencingSourceContext, moduleName, &moduleRecordBase);
+                ASYNC_HOST_OPERATION_END(scriptContext->GetThreadContext());
+            }
+
+            if (FAILED(hr))
+            {
+                Js::JavascriptError * error = scriptContext->GetLibrary()->CreateURIError();
+                Var value = JavascriptString::NewCopySz(moduleName, scriptContext);
+                JavascriptOperators::OP_SetProperty(error, PropertyIds::message, value, scriptContext);
```
agarwal-sandeep: >OP_SetProperty [](start = 37, length = 14)

Should you use JavascriptError::SetErrorMessageProperties instead?

lib/Runtime/Language/JavascriptOperators.cpp
```diff
@@ -9639,6 +9639,82 @@ namespace Js
...
+                {
+                    moduleRecord->InstantiateAndNotifyHostAboutDynamicModuleReady();
+                }
+
+                return promiseCapability->GetPromise();
+            }
+        }
+        else
+        {
+            Assert(false);
```
agarwal-sandeep: >Assert [](start = 12, length = 6)

nit: Using AssertMsg with a message would be more useful

lib/Runtime/Language/SourceTextModuleRecord.cpp
```diff
@@ -983,4 +1035,41 @@ namespace Js
...
+        {
+            promiseCapability = moduleRecord->GetPromiseCapability();
+        }
+
+        if (promiseCapability == nullptr)
+        {
+            promiseCapability = JavascriptPromise::NewPromiseCapability(scriptContext->GetLibrary()->GetPromiseConstructor(), scriptContext);
+        }
+
+        Var resolveOrRejectVar = toResolve ? promiseCapability->GetResolve() : promiseCapability->GetReject();
```
boingoing: So `toResolve` indicates the promise should be resolved instead of rejected? Nitpicky, but I think I would prefer this to be named `isResolve`, `isReject`, `success`, or something along those lines.

lib/Runtime/Language/JavascriptOperators.cpp
```diff
@@ -9639,6 +9639,86 @@ namespace Js
...
+            {
+                return SourceTextModuleRecord::ResolveOrRejectDynamicImportPromise(false, moduleRecord->GetErrorObject(), scriptContext, moduleRecord);
+            }
+            else
+            {
+                JavascriptPromiseCapability *promiseCapability = moduleRecord->GetPromiseCapability();
+                if (promiseCapability == nullptr)
+                {
+                    BEGIN_TRANSLATE_TO_HRESULT(static_cast<ExceptionType>(ExceptionType_OutOfMemory | ExceptionType_StackOverflow));
+                    promiseCapability = JavascriptPromise::NewPromiseCapability(scriptContext->GetLibrary()->GetPromiseConstructor(), scriptContext);
```
boingoing: Does the spec call-out that we should be storing the `PromiseCapabilityRecord` here or is this more of an implementation detail? I would suspect it should be simpler to store a `JavascriptPromise*` here and call `JavascriptPromise::Resolve()` or `JavascriptPromise::Reject()` on it when it's time to resolve/reject it.

Looks like we're storing `promiseCapability` here to avoid casting the promise into a `JavascriptPromise*` because `JavascriptPromiseCapability::promise` is a `Var`? I think that cast should be safe to do since we pass `%Promise%` to `JavascriptPromise::NewPromiseCapability` which is guaranteed to return something that's a `JavascriptPromise*` because the return object is constructed via `JavascriptLibrary::CreatePromise` and we aren't doing a super call. Actually might be even simpler than that since we don't even need to call an executor function for this promise, right? So we could just call `library->CreatePromise()` and pass the result to `JavascriptPromise::InitializePromise`.

Might make sense to add a helper to `JavascriptPromise` itself

```c++
    JavascriptPromise* JavascriptPromise::CreateEnginePromise(ScriptContext* scriptContext)
    {
        // Unused
        JavascriptPromiseResolveOrRejectFunction* resolve;
        JavascriptPromiseResolveOrRejectFunction* reject;

        JavascriptPromise* promise = library->CreatePromise();
        InitializePromise(promise, &resolve, &reject, scriptContext);

        return promise;
    }
```

I think it would be simpler to resolve the promise at least if we stored the promise itself instead of the capability record. We would also avoid constructing a bunch of objects and making a bunch of calls by constructing the promise as the above helper.

lib/Runtime/Language/i386/StackFrame.cpp
```diff
@@ -51,9 +51,13 @@ X86StackFrame::InitializeByReturnAddress(void * returnAddress, ScriptContext* sc
 bool
 X86StackFrame::Next()
 {
-    this->addressOfCodeAddr = this->GetAddressOfReturnAddress();
-    this->codeAddr = this->GetReturnAddress();
-    this->frame = (void **)this->frame[0];
+    if (frame != nullptr)
+    {
```
tcare: Which situation needs this and why only x86?

lib/Runtime/Language/SourceTextModuleRecord.cpp
tcare: +1, I also found this hard to read.

---
In reply to: [114906734](https://github.com/Microsoft/ChakraCore/pull/2913#discussion_r114906734) [](ancestors = 114906734)

lib/Parser/Parse.cpp
```diff
@@ -2436,19 +2436,26 @@ bool Parser::IsImportOrExportStatementValidHere()
...
     Assert(m_scriptContext->GetConfig()->IsES6ModuleEnabled());
     Assert(m_token.tk == tkIMPORT);
 
+    m_pscan->Scan();
+
+    // import()
+    if (m_token.tk == tkLParen)
+    {
+        ParseNodePtr specifier = ParseTerm<buildAST>();
+        return CreateCallNode(knopCall, CreateNodeWithScanner<knopImport>(), specifier);
```
tcare: >CreateCallNode [](start = 15, length = 14)

Since we parse this like a call, I would like to make sure we have sufficient coverage for parsing cases where import() could be mistaken for a function call, e.g. with multiple parameters, spread, etc.

boingoing: This is a good point

lib/Runtime/ByteCode/ByteCodeEmitter.cpp
```diff
@@ -10803,6 +10803,14 @@ void Emit(ParseNode *pnode, ByteCodeGenerator *byteCodeGenerator, FuncInfo *func
         {
             byteCodeGenerator->EmitSuperCall(funcInfo, pnode, fReturnValue);
         }
+        else if (pnode->sxCall.pnodeTarget->nop == knopImport)
+        {
+            ParseNodePtr args = pnode->sxCall.pnodeArgs;
+            Emit(args, byteCodeGenerator, funcInfo, false);
```
tcare: >args [](start = 17, length = 4)

Since we're piggybacking off a call node, perhaps we want to assert some restrictions e.g. single parameter

lib/Common/Exceptions/ExceptionCheck.cpp
```diff
@@ -58,7 +58,8 @@ void ExceptionCheck::SetHandledExceptionType(ExceptionType e)
 #if DBG
     if(!(e == ExceptionType_None ||
          e == ExceptionType_DisableCheck ||
-         !JsUtil::ExternalApi::IsScriptActiveOnCurrentThreadContext() ||
+         (e & ExceptionType_OutOfMemory) == ExceptionType_OutOfMemory ||  // InitializeModuleRecord handles OOM during dynamic import
```
curtisman: >InitializeModuleRecord [](start = 77, length = 22)

Why add this?  The point here is that if you are in script, you should be throwing the JS exception object instead of OutOfMemory.

lib/Runtime/Library/JavascriptLibrary.cpp
```diff
@@ -7134,12 +7134,12 @@ namespace Js
...
     {
-        Assert((moduleRecordList->Count() >= 0) && (moduleId < (uint)moduleRecordList->Count()));
-        if (moduleId >= (uint)moduleRecordList->Count())
+        Assert((moduleRecordList->Count() > 0) && (moduleId <= (uint)moduleRecordList->Count()));
+        if (moduleId > (uint)moduleRecordList->Count())
         {
             Js::Throw::FatalInternalError();
         }
-        return moduleRecordList->Item(moduleId);
+        return moduleRecordList->Item(moduleId - 1); // 0 == kmodGlobal
```
curtisman: Why use kmodGlobal?  That is the value to distingish between the different global scopes (the real global scope of form global scope).  It has nothing to do with Javascript modules)
Please define a different constant to mean: this is not a module?  May be use the value -1 instead os you don't have to do this weird shifting?

lib/Runtime/Language/SourceTextModuleRecord.cpp
```diff
@@ -609,6 +645,15 @@ namespace Js
         }
     }
 
+    void SourceTextModuleRecord::EnsureChildModuleSet(ScriptContext *scriptContext)
+    {
+        if (nullptr == this->childrenModuleSet)
+        {
+            ArenaAllocator* allocator = scriptContext->GeneralAllocator();
+            this->childrenModuleSet = (ChildModuleRecordSet*)AllocatorNew(ArenaAllocator, allocator, ChildModuleRecordSet, allocator);
```
curtisman: >childrenModuleSet [](start = 18, length = 17)

Use Anew?

curtisman:
>         childrenModuleSet = nullptr;

childrenModuleSet is leaked?

---
Refers to: lib/Runtime/Language/SourceTextModuleRecord.cpp:77 in ea66233. [](commit_id = ea66233b2394329953901fc69aea3625b631d40a, deletion_comment = False)

lib/Runtime/Language/JavascriptOperators.cpp
```diff
@@ -9639,6 +9639,86 @@ namespace Js
         return superCtor;
     }
 
+    Var JavascriptOperators::OP_ImportCall(Var specifier, ScriptContext* scriptContext)
+    {
+        ModuleRecordBase *moduleRecordBase = nullptr;
+        SourceTextModuleRecord *moduleRecord = nullptr;
+
+        JavascriptFunction* pfuncCaller;
+        if (JavascriptStackWalker::GetCaller(&pfuncCaller, scriptContext) && pfuncCaller && pfuncCaller->IsScriptFunction())
```
curtisman: Why do we need to use the stack walker here?    Didn't the interpreter just called this OP_ImportCall and know what the function is already?  Can't the interpreter pass it in?

lib/Runtime/Language/JavascriptOperators.cpp
```diff
@@ -9639,6 +9639,86 @@ namespace Js
...
+        {
+            FunctionBody* parentFuncBody = pfuncCaller->GetFunctionBody();
+            ModuleID id = parentFuncBody->GetModuleID();
```
curtisman: The ModuleID here has no relationship with Javascript modules.
This is in reference to different global scope (either global, or in a  HTML form).
The type should rename ModuleRootID so it is not being confused.

curtisman:
>                 srcInfo->moduleID = moduleId;

moduleID in SRCINFO is unrelated to javascript modules.  It is probably wron't to store this here.

---
Refers to: lib/Runtime/Language/SourceTextModuleRecord.cpp:122 in ea66233. [](commit_id = ea66233b2394329953901fc69aea3625b631d40a, deletion_comment = False)

lib/Runtime/Base/LeaveScriptObject.cpp
```diff
@@ -63,7 +63,7 @@ namespace Js
             // not run and we might be in an inconsistent state
 
             // Put any code that may raise an exception in OnScriptStart
-            scriptContext->GetThreadContext()->EnterScriptStart(entryExitRecord, doCleanup);
+            scriptContext->GetThreadContext()->EnterScriptStart(entryExitRecord, doCleanup, &this->isScriptActive);
```
curtisman: are we trying to allow nested EnterScript without LeaveScript?  Why?

suwc: This is to allow e.g. ImportCall() -> FetchImportedModule() -> InitializeModuleRecord().
When FetchImportedModule() from host calls into chakra, it goes through EnterScript-LeaveScript.
Non-nested EnterScript-LeaveScript resets threadContext->isScriptActive to false.


---
In reply to: [114931889](https://github.com/Microsoft/ChakraCore/pull/2913#discussion_r114931889) [](ancestors = 114931889)

lib/Runtime/Language/SourceTextModuleRecord.cpp
```diff
@@ -61,7 +61,7 @@ namespace Js
         // in practice the modulerecord lifetime should be the same as the scriptcontext so it could be retrieved for the same
         // site. Host might hold a reference to the module as well after initializing the module.
         // In our implementation, we'll use the moduleId in bytecode to identify the module.
-        childModuleRecord->moduleId = scriptContext->GetLibrary()->EnsureModuleRecordList()->Add(childModuleRecord);
+        childModuleRecord->moduleId = scriptContext->GetLibrary()->EnsureModuleRecordList()->Add(childModuleRecord) + 1; // 0 == kmodGlobal
```
suwc: Found a better way to check module code so this shift is no longer needed.

---
In reply to: [114905856](https://github.com/Microsoft/ChakraCore/pull/2913#discussion_r114905856) [](ancestors = 114905856)

lib/Runtime/Language/SourceTextModuleRecord.cpp
```diff
@@ -230,15 +259,15 @@ namespace Js
             OUTPUT_TRACE_DEBUGONLY(Js::ModulePhase, _u("\t>NotifyParentsAsNeeded\n"));
             NotifyParentsAsNeeded();
 
-            if (!WasDeclarationInitialized() && isRootModule)
+            if (this->promiseCapability != nullptr || (!WasDeclarationInitialized() && isRootModule))
             {
                 // TODO: move this as a promise call? if parser is called from a different thread
```
suwc: This PR doesn't enable the case where parser is called from a different thread. Therefore I leave this TODO here for future enhancement.

---
In reply to: [114906209](https://github.com/Microsoft/ChakraCore/pull/2913#discussion_r114906209) [](ancestors = 114906209)

lib/Runtime/Language/i386/StackFrame.cpp
suwc: Only with nested enter_script, which I have removed.

---
In reply to: [114913205](https://github.com/Microsoft/ChakraCore/pull/2913#discussion_r114913205) [](ancestors = 114913205)

lib/Parser/Parse.cpp
suwc: Added dynamic-module-import-specifier.js
More to come.

---
In reply to: [114915006](https://github.com/Microsoft/ChakraCore/pull/2913#discussion_r114915006) [](ancestors = 114915006)

lib/Common/Exceptions/ExceptionCheck.cpp
```diff
@@ -58,7 +58,8 @@ void ExceptionCheck::SetHandledExceptionType(ExceptionType e)
...
+        !JsUtil::ExternalApi::IsScriptActiveOnCurrentThreadContext() ||
```
suwc: Fixed in 2nd iteration

---
In reply to: [114900876](https://github.com/Microsoft/ChakraCore/pull/2913#discussion_r114900876) [](ancestors = 114900876)

lib/Jsrt/ChakraCore.h
suwc: Updated in 2nd iteration.

---
In reply to: [114836637](https://github.com/Microsoft/ChakraCore/pull/2913#discussion_r114836637) [](ancestors = 114836637)

suwc:
>                 srcInfo->moduleID = moduleId;

Switched to using fscrIsModuleCode to detect ES6 module in ParseableFuncitonInfo::IsES6ModuleCode().
Removed shifting of moduleId by 1 and going back to starting from 0.
Will follow up SRCINFO->moduleID usage separately.

---
In reply to: [299373242](https://github.com/Microsoft/ChakraCore/pull/2913#issuecomment-299373242) [](ancestors = 299373242)

---
Refers to: lib/Runtime/Language/SourceTextModuleRecord.cpp:122 in ea66233. [](commit_id = ea66233b2394329953901fc69aea3625b631d40a, deletion_comment = False)

lib/Runtime/Language/JavascriptOperators.cpp
suwc: Updated interpreter and lowerer code to pass in function and removed stack walker call.

---
In reply to: [114930380](https://github.com/Microsoft/ChakraCore/pull/2913#discussion_r114930380) [](ancestors = 114930380)

lib/Common/Exceptions/ExceptionCheck.cpp
suwc: Removed along with nested EnterScript.

---
In reply to: [114929334](https://github.com/Microsoft/ChakraCore/pull/2913#discussion_r114929334) [](ancestors = 114929334)

suwc:
>         childrenModuleSet = nullptr;

Added Adelete in 2nd iteration.

---
In reply to: [299368705](https://github.com/Microsoft/ChakraCore/pull/2913#issuecomment-299368705) [](ancestors = 299368705)

---
Refers to: lib/Runtime/Language/SourceTextModuleRecord.cpp:77 in ea66233. [](commit_id = ea66233b2394329953901fc69aea3625b631d40a, deletion_comment = False)

lib/Runtime/Language/SourceTextModuleRecord.cpp
suwc: Updated in 2nd iteration.

---
In reply to: [114930171](https://github.com/Microsoft/ChakraCore/pull/2913#discussion_r114930171) [](ancestors = 114930171)

lib/Runtime/ByteCode/ByteCodeEmitter.cpp
suwc: Added assertion in 2nd iteration.

---
In reply to: [114915361](https://github.com/Microsoft/ChakraCore/pull/2913#discussion_r114915361) [](ancestors = 114915361)

lib/Runtime/Language/JavascriptOperators.cpp
suwc: Good idea. Updated in 2nd iteration.

---
In reply to: [114912538](https://github.com/Microsoft/ChakraCore/pull/2913#discussion_r114912538) [](ancestors = 114912538)

lib/Runtime/Language/SourceTextModuleRecord.cpp
suwc: Changed to 'isResolve'

---
In reply to: [114906734](https://github.com/Microsoft/ChakraCore/pull/2913#discussion_r114906734) [](ancestors = 114906734)

lib/Runtime/Language/JavascriptOperators.cpp
suwc: Good point. This line is removed along with stack walker call though.

---
In reply to: [114845379](https://github.com/Microsoft/ChakraCore/pull/2913#discussion_r114845379) [](ancestors = 114845379)

lib/Runtime/Language/JavascriptOperators.cpp
suwc: Updated in 2nd iteration.

---
In reply to: [114844985](https://github.com/Microsoft/ChakraCore/pull/2913#discussion_r114844985) [](ancestors = 114844985)

lib/Parser/Parse.cpp
```diff
@@ -2436,19 +2436,26 @@ bool Parser::IsImportOrExportStatementValidHere()
 }
 
 template<bool buildAST>
-ParseNodePtr Parser::ParseImportDeclaration()
+ParseNodePtr Parser::ParseImport()
 {
     Assert(m_scriptContext->GetConfig()->IsES6ModuleEnabled());
     Assert(m_token.tk == tkIMPORT);
 
+    m_pscan->Scan();
```
suwc: Yes, updated in 2nd iteration.

---
In reply to: [114901862](https://github.com/Microsoft/ChakraCore/pull/2913#discussion_r114901862) [](ancestors = 114901862)

test/es6/dynamic-module-functionality.js
suwc: Added new test dynamic-module-import-specifier.js.
More tests to come.

---
In reply to: [114840008](https://github.com/Microsoft/ChakraCore/pull/2913#discussion_r114840008) [](ancestors = 114840008)

suwc:
@dotnet-bot 
test Windows arm_release
test Windows x64_release
test Windows x86_release

lib/Parser/Parse.cpp
```diff
@@ -2435,12 +2435,37 @@ bool Parser::IsImportOrExportStatementValidHere()
         && this->m_tryCatchOrFinallyDepth == 0;
 }
 
+template<bool buildAST> ParseNodePtr Parser::ParseImportCall()
+{
+    m_pscan->Scan();
+    ParseNodePtr specifier = ParseExpr<buildAST>(koplCma, nullptr, /* fAllowIn */FALSE, /* fAllowEllipsis */FALSE);
+    if (m_token.tk != tkRParen)
+    {
+        Error(ERRsyntax);
```
akroshg: nit: proper error would be useful.

suwc: Changing to 'ERRnoRparen'.

---
In reply to: [115324907](https://github.com/Microsoft/ChakraCore/pull/2913#discussion_r115324907) [](ancestors = 115324907)

bin/ch/WScriptJsrt.cpp
```diff
@@ -1267,6 +1295,45 @@ JsErrorCode WScriptJsrt::FetchImportedModule(_In_ JsModuleRecord referencingModu
     return errorCode;
 }
 
+// Callback from chakracore to fetch module dynamically during runtime. In the test harness,
+// we are not doing any translation, just treat the specifier as fileName.
+// While this call will come back directly from runtime script or module code, the additional
+// task can be scheduled asynchronously that executed later.
+JsErrorCode WScriptJsrt::FetchImportedModuleFromScript(_In_ JsSourceContext dwReferencingSourceContext,
```
akroshg: >FetchImportedModuleFromScript [](start = 25, length = 29)

wondering if this function can be refactored with above function. It seems most of the code is shared.

lib/Jsrt/Core/JsrtContextCore.cpp
```diff
@@ -120,6 +120,36 @@ HRESULT ChakraCoreHostScriptContext::FetchImportedModule(Js::ModuleRecordBase* r
...
+HRESULT ChakraCoreHostScriptContext::FetchImportedModuleFromScript(JsSourceContext dwReferencingSourceContext, LPCOLESTR specifier, Js::ModuleRecordBase** dependentModuleRecord)
+{
+    if (fetchImportedModuleFromScriptCallback == nullptr)
+    {
+        return E_INVALIDARG;
+    }
+
+    HRESULT hr = NOERROR;
+    Js::JavascriptString* specifierVar = nullptr;
+    specifierVar = Js::JavascriptString::NewCopySz(specifier, GetScriptContext());
```
akroshg: nit: combine these two lines?

lib/Jsrt/Core/JsrtContextCore.cpp
```diff
@@ -120,6 +120,36 @@ HRESULT ChakraCoreHostScriptContext::FetchImportedModule(Js::ModuleRecordBase* r
...
+    
+    if (FAILED(hr))
```
akroshg: is this redundant check? where the hr is changed in above code?

lib/Jsrt/Core/JsrtContextCore.cpp
```diff
@@ -120,6 +120,36 @@ HRESULT ChakraCoreHostScriptContext::FetchImportedModule(Js::ModuleRecordBase* r
...
+    {
+        return hr;
+    }
+
+    JsModuleRecord dependentRecord = JS_INVALID_REFERENCE;
```
akroshg: there is not much difference between this function and above function (FetchImportedModule). Did you consider creating a lambda and just stub relevant part in that?

lib/Jsrt/Core/JsrtContextCore.h
```diff
@@ -177,6 +178,9 @@ class ChakraCoreHostScriptContext sealed : public HostScriptContext
     void SetFetchImportedModuleCallback(FetchImportedModuleCallBack fetchCallback) { this->fetchImportedModuleCallback = fetchCallback ; }
     FetchImportedModuleCallBack GetFetchImportedModuleCallback() const { return this->fetchImportedModuleCallback; }
 
+    void SetFetchImportedModuleFromScriptCallback(FetchImportedModuleFromScriptCallBack fetchCallback) { this->fetchImportedModuleFromScriptCallback = fetchCallback; }
```
akroshg: Is this function allowed to be called multiple times? Is it good to assert that this->fetchImportedModuleFromScriptCallback is null when it is called?

akroshg:
>         case JsModuleHostInfo_NotifyModuleReadyCallback:

any update required in this function?

---
Refers to: lib/Jsrt/Core/JsrtCore.cpp:209 in db8d7fd. [](commit_id = db8d7fdc5b27af9f9814887a781d6e8c924e07b7, deletion_comment = False)

akroshg:
>         if ((this->GetGrfscr() & fscrIsModuleCode) == fscrIsModuleCode || this->GetModuleID() == kmodGlobal)

nit: can this be changed with IsES6ModuleCode()?

---
Refers to: lib/Runtime/Base/FunctionBody.cpp:3166 in db8d7fd. [](commit_id = db8d7fdc5b27af9f9814887a781d6e8c924e07b7, deletion_comment = False)

lib/Parser/Parse.cpp
```diff
@@ -2435,12 +2435,37 @@ bool Parser::IsImportOrExportStatementValidHere()
         && this->m_tryCatchOrFinallyDepth == 0;
 }
 
+template<bool buildAST> ParseNodePtr Parser::ParseImportCall()
+{
+    m_pscan->Scan();
+    ParseNodePtr specifier = ParseExpr<buildAST>(koplCma, nullptr, /* fAllowIn */FALSE, /* fAllowEllipsis */FALSE);
```
akroshg: >ParseExpr [](start = 29, length = 9)

Clarification : will this function stop after first comma?

lib/Runtime/Language/JavascriptOperators.cpp
```diff
@@ -9639,6 +9639,91 @@ namespace Js
...
+        {
+            specifierString = JavascriptConversion::ToString(specifier, scriptContext);
+        }
+        catch (const JavascriptException &err)
+        {
+            Var errorObject = err.GetAndClear()->GetThrownObject(scriptContext);
+            if (errorObject != nullptr && Js::JavascriptError::Is(errorObject))
+            {
+                return SourceTextModuleRecord::ResolveOrRejectDynamicImportPromise(false, errorObject, scriptContext);
+            }
```
akroshg: don't we need to rethrow if above checks fail?

lib/Runtime/Language/JavascriptOperators.cpp
```diff
@@ -9639,6 +9639,91 @@ namespace Js
...
+
+            BEGIN_LEAVE_SCRIPT(scriptContext);
+            BEGIN_TRANSLATE_TO_HRESULT(static_cast<ExceptionType>(ExceptionType_OutOfMemory | ExceptionType_StackOverflow));
+            hr = scriptContext->GetHostScriptContext()->FetchImportedModule(referenceModuleRecord, moduleName, &moduleRecordBase);
+            END_TRANSLATE_EXCEPTION_TO_HRESULT(hr);
+            END_LEAVE_SCRIPT(scriptContext);
+        }
+        else
+        {
+            DWORD_PTR dwReferencingSourceContext = parentFuncBody->GetFunctionInfo()->GetFunctionProxy()->GetSourceContextInfo()->dwHostSourceContext;
```
akroshg: what if the parent function is an eval? the dwHostSourceContext will be -1 in that case?

test/es6/dynamic-module-import-syntax.js
```diff
@@ -0,0 +1,83 @@
+//-------------------------------------------------------------------------------------------------------
+// Copyright (C) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE.txt file in the project root for full license information.
+//-------------------------------------------------------------------------------------------------------
+
+// ES6 Generators Syntax tests -- verifies function* and yield syntax spec conformance
```
boingoing: Nitpick, this comment doesn't look right.

lib/Runtime/Library/JavascriptPromise.cpp
```diff
@@ -1863,4 +1863,16 @@ namespace Js
 
         return args[0];
     }
+
+    //static
+    JavascriptPromise* JavascriptPromise::CreateEnginePromise(ScriptContext *scriptContext)
```
boingoing: Cool 👍

lib/Jsrt/Core/JsrtContextCore.h
suwc: These functions are being called multiple times.
For example:

REQUIRE(JsSetModuleHostInfo(requestModule, JsModuleHostInfo_FetchImportedModuleCallback, Success_FIMC) == JsNoError);


---
In reply to: [115328835](https://github.com/Microsoft/ChakraCore/pull/2913#discussion_r115328835) [](ancestors = 115328835)

lib/Parser/Parse.cpp
suwc: Yes, it stops at ',' and doesn't accept '...'

---
In reply to: [115330698](https://github.com/Microsoft/ChakraCore/pull/2913#discussion_r115330698) [](ancestors = 115330698)

lib/Runtime/Language/SourceTextModuleRecord.cpp
```diff
@@ -220,6 +224,38 @@ namespace Js
...
+    {
+        HRESULT hr = NOERROR;
+        if (!WasDeclarationInitialized())
+        {
+            ScriptContext* scriptContext = GetScriptContext();
+            ModuleDeclarationInstantiation();
+
+            if (this->errorObject != nullptr)
+            {
+                SourceTextModuleRecord::ResolveOrRejectDynamicImportPromise(false, this->GetErrorObject(), this->GetScriptContext(), this);
```
akroshg: >this->GetScriptContext() [](start = 107, length = 24)

nit: just scriptContext, as we have got that in the local variable?

lib/Runtime/Language/SourceTextModuleRecord.cpp
```diff
@@ -742,7 +801,15 @@ namespace Js
         Assert(this->errorObject == nullptr);
         if (this->errorObject != nullptr)
         {
-            JavascriptExceptionOperators::Throw(errorObject, scriptContext);
+            if (this->promise != nullptr)
```
akroshg: can this codepath hit? there is an assert above which says Assert(this->errorObject == nullptr); ?

lib/Runtime/Language/SourceTextModuleRecord.cpp
```diff
@@ -765,7 +832,17 @@ namespace Js
         CleanupBeforeExecution();
 
         Arguments outArgs(CallInfo(CallFlags_Value, 0), nullptr);
-        return rootFunction->CallRootFunction(outArgs, scriptContext, true);
+        
+        Var ret = nullptr;
+        AUTO_NESTED_HANDLED_EXCEPTION_TYPE((ExceptionType)(ExceptionType_OutOfMemory | ExceptionType_JavascriptException));
+        ENTER_SCRIPT_IF(scriptContext, true, false, false, !scriptContext->GetThreadContext()->IsScriptActive(),
+        {
+            ret = rootFunction->CallRootFunction(outArgs, scriptContext, true);
```
akroshg: If this throws exception will the promise reject properly?

test/es6/dynamic-module-import-syntax.js
```diff
@@ -0,0 +1,83 @@
...
+    }
+}
+
+var tests = [
+    {
+        name: "Valid cases for import()",
+        body: function () {
+                let functionBody =
+                    `
+                    assert.doesNotThrow(function () { eval("import(undefined)"); }, "undefined");
```
akroshg: aren't these tests repeated from import-specifier.js file?

test/es6/dynamic-module-import-specifier.js
```diff
@@ -0,0 +1,142 @@
...
+                    message = specifier;
+                }
+
+                let functionBody =
+                    `testDynamicImport(
+                        ()=>{
+                            return import(${specifier});
+                        },
+                        (v)=>{
+                            assert.isTrue(false, 'Expected: promise rejected; actual: promise resolved: ' + '${message}');
```
akroshg: nit: change this to assert.fail() .

test/es6/dynamic-module-functionality.js
```diff
@@ -0,0 +1,359 @@
...
+                    (err)=>{ assert.isTrue(false, err.message); }, _asyncEnter, _asyncExit);
+                `;
+
+            testScript(functionBody, "Implicitly re-exporting an import binding (import { foo } from ''; export { foo };)", false, true);
+            testModuleScript(functionBody, "Implicitly re-exporting an import binding (import { foo } from ''; export { foo };)", false, true);
+        }
+    },
+];
+
+testRunner.runTests(tests, { verbose: WScript.Arguments[0] != "summary" });
```
akroshg: Is there a test case which covers the nested dynamic imports ?

lib/Parser/Parse.cpp
```diff
@@ -2435,12 +2435,37 @@ bool Parser::IsImportOrExportStatementValidHere()
...
 template<bool buildAST>
-ParseNodePtr Parser::ParseImportDeclaration()
+ParseNodePtr Parser::ParseImport()
 {
     Assert(m_scriptContext->GetConfig()->IsES6ModuleEnabled());
     Assert(m_token.tk == tkIMPORT);
 
+    RestorePoint parsedImport;
+    m_pscan->Capture(&parsedImport);
+    m_pscan->Scan();
```
tcare: Do we need to do a rewind? We are just going to scan again a few lines down anyway.

test/es6/dynamic-module-import-syntax.js
suwc: They are not exactly the same. Import-specifier.js tests ToString operation applied on specifier. This one tests parser.

---
In reply to: [115364519](https://github.com/Microsoft/ChakraCore/pull/2913#discussion_r115364519) [](ancestors = 115364519)

lib/Parser/Parse.cpp
suwc: This is just to make sure col# reported by error msg is not off due to the premature scan.

---
In reply to: [115371147](https://github.com/Microsoft/ChakraCore/pull/2913#discussion_r115371147) [](ancestors = 115371147)

test/es6/dynamic-module-functionality.js
suwc: Nest import syntax is test in dynamic-module-import-specifier.js


---
In reply to: [115367185](https://github.com/Microsoft/ChakraCore/pull/2913#discussion_r115367185) [](ancestors = 115367185)

test/es6/dynamic-module-import-syntax.js
akroshg: I don't see this file is added to the rlexe.xml, am I missing something?

---
In reply to: [115372781](https://github.com/Microsoft/ChakraCore/pull/2913#discussion_r115372781) [](ancestors = 115372781,115364519)

bin/ch/WScriptJsrt.cpp
suwc: updated in latest iteration

---
In reply to: [115327127](https://github.com/Microsoft/ChakraCore/pull/2913#discussion_r115327127) [](ancestors = 115327127)

lib/Jsrt/Core/JsrtContextCore.cpp
suwc: NA in latest iteration (file refactored)

---
In reply to: [115328093](https://github.com/Microsoft/ChakraCore/pull/2913#discussion_r115328093) [](ancestors = 115328093)

lib/Jsrt/Core/JsrtContextCore.cpp
suwc: NA in latest iteration (function refactored).

---
In reply to: [115328224](https://github.com/Microsoft/ChakraCore/pull/2913#discussion_r115328224) [](ancestors = 115328224)

lib/Jsrt/Core/JsrtContextCore.cpp
suwc: Refactored in latest iteration.

---
In reply to: [115328606](https://github.com/Microsoft/ChakraCore/pull/2913#discussion_r115328606) [](ancestors = 115328606)

lib/Runtime/Language/JavascriptOperators.cpp
suwc: Changed to assert + internal error.

---
In reply to: [115331615](https://github.com/Microsoft/ChakraCore/pull/2913#discussion_r115331615) [](ancestors = 115331615)

lib/Runtime/Language/JavascriptOperators.cpp
suwc: Case handled in latest iteration. Added test.

---
In reply to: [115331981](https://github.com/Microsoft/ChakraCore/pull/2913#discussion_r115331981) [](ancestors = 115331981)

lib/Runtime/Language/SourceTextModuleRecord.cpp
suwc: Moved assert to below in the latest iteration.

---
In reply to: [115362106](https://github.com/Microsoft/ChakraCore/pull/2913#discussion_r115362106) [](ancestors = 115362106)

lib/Runtime/Language/SourceTextModuleRecord.cpp
suwc: Added promise reject in latest iteration.

---
In reply to: [115362554](https://github.com/Microsoft/ChakraCore/pull/2913#discussion_r115362554) [](ancestors = 115362554)

suwc:
>         case JsModuleHostInfo_NotifyModuleReadyCallback:

Done.

---
In reply to: [299962864](https://github.com/Microsoft/ChakraCore/pull/2913#issuecomment-299962864) [](ancestors = 299962864)

---
Refers to: lib/Jsrt/Core/JsrtCore.cpp:209 in db8d7fd. [](commit_id = db8d7fdc5b27af9f9814887a781d6e8c924e07b7, deletion_comment = False)

suwc:
>         if ((this->GetGrfscr() & fscrIsModuleCode) == fscrIsModuleCode || this->GetModuleID() == kmodGlobal)

Done.

---
In reply to: [299963481](https://github.com/Microsoft/ChakraCore/pull/2913#issuecomment-299963481) [](ancestors = 299963481)

---
Refers to: lib/Runtime/Base/FunctionBody.cpp:3166 in db8d7fd. [](commit_id = db8d7fdc5b27af9f9814887a781d6e8c924e07b7, deletion_comment = False)

lib/Jsrt/Core/JsrtContextCore.cpp
```diff
@@ -110,7 +129,7 @@ HRESULT ChakraCoreHostScriptContext::FetchImportedModule(Js::ModuleRecordBase* r
     JsModuleRecord dependentRecord = JS_INVALID_REFERENCE;
     {
         AUTO_NO_EXCEPTION_REGION;
-        JsErrorCode errorCode = fetchImportedModuleCallback(referencingModule, specifierVar, &dependentRecord);
+        JsErrorCode errorCode = fetch(specifierVar, &dependentRecord);//fetchImportedModuleCallback(referencingModule, specifierVar, &dependentRecord);
```
akroshg: nitpick : comments can be removed

lib/Runtime/Language/JavascriptOperators.cpp
```diff
@@ -9639,6 +9639,115 @@ namespace Js
...
+        catch (const JavascriptException &err)
+        {
+            Var errorObject = err.GetAndClear()->GetThrownObject(scriptContext);
+            AssertMsg(errorObject != nullptr && Js::JavascriptError::Is(errorObject), "OP_ImportCall: Unhandled exception thrown by ToString(specifier)");
+            if (errorObject != nullptr && Js::JavascriptError::Is(errorObject))
+            {
+                return SourceTextModuleRecord::ResolveOrRejectDynamicImportPromise(false, errorObject, scriptContext);
+            }
+
+            Throw::InternalError();
```
akroshg: what happens if 
import( { toString : function() { 
throw 1;
}});

lib/Runtime/Language/JavascriptOperators.cpp
```diff
@@ -9639,6 +9639,115 @@ namespace Js
...
+        }
+
+        // Walk the call stack if caller function is neither module code nor having host source context
+        DWORD_PTR dwReferencingSourceContext = dwReferencingSourceContext = parentFuncBody->GetFunctionInfo()->GetFunctionProxy()->GetSourceContextInfo()->dwHostSourceContext;
+        if (!parentFuncBody->IsES6ModuleCode() && dwReferencingSourceContext == Js::Constants::NoHostSourceContext)
```
akroshg: if dwReferencingSourceContext == Js::Constants::NoHostSourceContext
could you not use callerUtf8SourceInfo (defined in the utf8sourceinfo)? Confirm with @agarwal.sandeep@microsoft.com that this is populated in the non-debug scenario as well.

lib/Runtime/Language/JavascriptOperators.cpp
```diff
@@ -9639,6 +9639,115 @@ namespace Js
...
+            if (moduleRecord->ParentsNotified())
+            {
+                hr = moduleRecord->InstantiateAndNotifyHostAboutDynamicModuleReady();
+                if (FAILED(hr))
+                {
+                    Js::JavascriptError * error = scriptContext->GetLibrary()->CreateURIError();
+                    JavascriptError::SetErrorMessageProperties(error, hr, moduleName, scriptContext);
+                    return SourceTextModuleRecord::ResolveOrRejectDynamicImportPromise(false, error, scriptContext);
+                }
+            }
```
akroshg: would be nice if this code moved to SoureTextModuleRecord. I don't think javascriptoperator.cpp needs to know any of this.

lib/Runtime/Language/SourceTextModuleRecord.cpp
```diff
@@ -175,6 +175,12 @@ namespace Js
             }
 
             OUTPUT_TRACE_DEBUGONLY(Js::ModulePhase, _u("\t>NotifyParentAsNeeded\n"));
+            if (this->promise != nullptr)
+            {
+                SourceTextModuleRecord::ResolveOrRejectDynamicImportPromise(false, this->GetErrorObject(), this->GetScriptContext(), this);
+                this->SetPromise(nullptr);
```
akroshg: >SetPromise [](start = 22, length = 10)

would it be wise to reset the promise to null first and do the ResolveOrReject.. ?

lib/Runtime/Language/SourceTextModuleRecord.cpp
```diff
@@ -175,6 +175,12 @@ namespace Js
             }
 
             OUTPUT_TRACE_DEBUGONLY(Js::ModulePhase, _u("\t>NotifyParentAsNeeded\n"));
+            if (this->promise != nullptr)
+            {
+                SourceTextModuleRecord::ResolveOrRejectDynamicImportPromise(false, this->GetErrorObject(), this->GetScriptContext(), this);
```
akroshg: >this->GetScriptContext() [](start = 107, length = 24)

nitpick : use the local scriptContext (also you can use the errorObject as well)

lib/Runtime/Language/SourceTextModuleRecord.cpp
```diff
@@ -220,6 +224,38 @@ namespace Js
...
+    {
+        HRESULT hr = NOERROR;
+        if (!WasDeclarationInitialized())
+        {
+            ScriptContext* scriptContext = GetScriptContext();
+            ModuleDeclarationInstantiation();
+
+            if (this->errorObject != nullptr)
+            {
+                SourceTextModuleRecord::ResolveOrRejectDynamicImportPromise(false, this->GetErrorObject(), scriptContext, this);
```
akroshg: >scriptContext [](start = 107, length = 13)

nitpick : could you not use the this->scriptContext? I am not sure why we need to have local variable for a scriptcontext.

lib/Runtime/Language/SourceTextModuleRecord.cpp
```diff
@@ -765,7 +834,41 @@ namespace Js
...
+                ret = rootFunction->CallRootFunction(outArgs, scriptContext, true);
+            });
+        }
+        catch (const Js::JavascriptException& err)
+        {
+            Var errorObject = err.GetAndClear()->GetThrownObject(scriptContext);
+            AssertMsg(errorObject != nullptr && Js::JavascriptError::Is(errorObject), "ModuleEvaluation: Unhandled exception thrown by calling root function");
+            if (errorObject != nullptr && Js::JavascriptError::Is(errorObject))
+            {
+                ResolveOrRejectDynamicImportPromise(false, errorObject, scriptContext);
```
akroshg: setPromise to null?

lib/Runtime/Language/SourceTextModuleRecord.cpp
```diff
@@ -765,7 +834,41 @@ namespace Js
...
+            {
+                Throw::InternalError();
+            }
+        }
+        catch (Js::OutOfMemoryException)
+        {
+            Js::JavascriptError * error = scriptContext->GetLibrary()->CreateError();
+            JavascriptError::SetErrorMessageProperties(error, E_OUTOFMEMORY, this->GetSpecifierSz(), scriptContext);
+            ResolveOrRejectDynamicImportPromise(false, error, scriptContext);
+            return scriptContext->GetLibrary()->GetUndefined();
```
akroshg: we are returning undefined in the OOM case? why not re-thorw?

lib/Runtime/Language/JavascriptOperators.cpp
agarwal-sandeep: callerUtf8SourceInfo is populated in non-debug scenario as well. You will still need to walk the chain to get to first non evaled parent.

lib/Runtime/Language/JavascriptOperators.cpp
suwc: I think the reasonable thing to expect is to reject the promise with 1.

Here spec says:
IfAbruptRejectPromise(specifierString, promiseCapability).

which is questionable because specifierString will be nullptr.
I will seek clarification.


---
In reply to: [115379478](https://github.com/Microsoft/ChakraCore/pull/2913#discussion_r115379478) [](ancestors = 115379478)

lib/Runtime/Language/SourceTextModuleRecord.cpp
suwc: In fact ResolveOrRejectDynamicImportPromise() takes care of this already. Removing this line and in similar cases

---
In reply to: [115382160](https://github.com/Microsoft/ChakraCore/pull/2913#discussion_r115382160) [](ancestors = 115382160)

lib/Runtime/Language/SourceTextModuleRecord.cpp
suwc: Adding 'this' as last param and SetPromise(nullptr) will be taken care of.

---
In reply to: [115383242](https://github.com/Microsoft/ChakraCore/pull/2913#discussion_r115383242) [](ancestors = 115383242)

lib/Runtime/Language/SourceTextModuleRecord.cpp
suwc: Only in the dynamic import case, we reject the promise with error object and return undefined.
In the static case, we should throw the error object. Making the change.

---
In reply to: [115383826](https://github.com/Microsoft/ChakraCore/pull/2913#discussion_r115383826) [](ancestors = 115383826)

lib/Runtime/Language/JavascriptOperators.cpp
suwc: Yes it works for a single eval case, but fails for nested eval.

---
In reply to: [115380766](https://github.com/Microsoft/ChakraCore/pull/2913#discussion_r115380766) [](ancestors = 115380766)

lib/Runtime/Language/JavascriptOperators.cpp
suwc: Good point. Moved and renamed the function.

---
In reply to: [115381049](https://github.com/Microsoft/ChakraCore/pull/2913#discussion_r115381049) [](ancestors = 115381049)

test/es6/dynamic-module-import-syntax.js
suwc: This file was redundant and was removed.

---
In reply to: [115339634](https://github.com/Microsoft/ChakraCore/pull/2913#discussion_r115339634) [](ancestors = 115339634)

test/es6/dynamic-module-import-specifier.js
suwc: Done.

---
In reply to: [115364860](https://github.com/Microsoft/ChakraCore/pull/2913#discussion_r115364860) [](ancestors = 115364860)

suwc:
@dotnet-bot 
run Windows x64_release

lib/Runtime/Language/JavascriptOperators.cpp
```diff
@@ -9639,6 +9639,104 @@ namespace Js
...
+                JavascriptFunction* caller = nullptr;
+                Js::JavascriptStackWalker walker(scriptContext);
+                walker.GetCaller(&caller);
+
+                do
+                {
+                    if (walker.GetCaller(&caller) && caller != nullptr && caller->IsScriptFunction())
+                    {
+                        parentFuncBody = caller->GetFunctionBody();
+                        dwReferencingSourceContext = dwReferencingSourceContext = parentFuncBody->GetFunctionInfo()->GetFunctionProxy()->GetSourceContextInfo()->dwHostSourceContext;
```
akroshg: nit: function body has function GetHostSourceContext(), which will give you the dwHostSOurceContext.

suwc:
@dotnet-bot 
test Windows x64_release

lib/Runtime/Language/JavascriptOperators.cpp
```diff
@@ -9639,6 +9639,104 @@ namespace Js
...
+            {
+                // Walk the call stack if caller function is neither module code nor having host source context
+
+                JavascriptFunction* caller = nullptr;
+                Js::JavascriptStackWalker walker(scriptContext);
+                walker.GetCaller(&caller);
+
+                do
+                {
+                    if (walker.GetCaller(&caller) && caller != nullptr && caller->IsScriptFunction())
```
akroshg: interesting to see what happens when the import is called in the setTimeout. Will you add a test case just to see that we don't get the fatal error.

lib/Runtime/Language/SourceTextModuleRecord.cpp
```diff
@@ -765,7 +844,52 @@ namespace Js
...
+                    ResolveOrRejectDynamicImportPromise(false, errorObject, scriptContext, this);
+                    return scriptContext->GetLibrary()->GetUndefined();
+                }
+                else
+                {
+                    Throw::InternalError();
+                }
+            }
+
+            throw(err);
```
akroshg: throw outside of catch handler. (look at other example in the code base, where they use different function for re-throw).

lib/Runtime/Language/SourceTextModuleRecord.cpp
```diff
@@ -765,7 +844,52 @@ namespace Js
...
+            {
+                ret = rootFunction->CallRootFunction(outArgs, scriptContext, true);
+            });
+        }
+        catch (const Js::JavascriptException &err)
+        {
+            if (this->promise != nullptr)
+            {
+                Var errorObject = err.GetAndClear()->GetThrownObject(scriptContext);
+                AssertMsg(errorObject != nullptr, "ModuleEvaluation: null error object thrown from root function");
```
akroshg: >AssertMsg [](start = 16, length = 9)

nit : convert to AssertOrFailFast so that you don't have to have if/else blocks

lib/Runtime/Language/SourceTextModuleRecord.cpp
```diff
@@ -739,19 +813,24 @@ namespace Js
...
+                return scriptContext->GetLibrary()->GetUndefined();
+            }
+            else
+            {
+                JavascriptExceptionOperators::Throw(errorObject, this->scriptContext);
+            }
         }
+
+        Assert(this->errorObject == nullptr);
         Assert(!WasEvaluated());
```
akroshg: nit: SetWasEvaluated has this assert covered.

lib/Runtime/Language/JavascriptOperators.cpp
akroshg: I couldn't find a scenario which involves the nested imports also dynamic->static->dynamic. sort of?

---
In reply to: [115561112](https://github.com/Microsoft/ChakraCore/pull/2913#discussion_r115561112) [](ancestors = 115561112)

akroshg:
Is import() configurable through config flags or it is enabled by default on?

lib/Runtime/Language/JavascriptOperators.cpp
suwc: I am still working on these test cases. They will come in later.

---
In reply to: [115577016](https://github.com/Microsoft/ChakraCore/pull/2913#discussion_r115577016) [](ancestors = 115577016,115561112)

suwc:
We are enabling it by default.

---
In reply to: [300271647](https://github.com/Microsoft/ChakraCore/pull/2913#issuecomment-300271647) [](ancestors = 300271647)

suwc:
Thanks for the code review.

PR #2917, opened 2017-05-05
LEAVE script macros for Init and next function of external iterators.

akroshg:


akroshg:
@curtisman please have a look.

akroshg:
pushed.

PR #2929, opened 2017-05-06
Test release/2.0-ci (fix #2928) [do not merge]

### LINKED ISSUE #2928: Jenkins: Style Checks depend on Ubuntu14.04 for no good reason, update to 16.04
It seems not many Jenkins CI users are using 14.04 anymore so we're paying an extra cost (style check throughput as well as CI cost) to use that OS as we wait for VMs to be provisioned.

There's no reason to have the style checks depend on 14.04. It was just the latest Ubuntu LTS available when we set up the checks.

Switch the style checks to use 16.04 like the Linux CI checks we use for testing ChakraCore on Linux. This way we are using the same pool of VMs across all of our checks, and will share that pool with more other users of the CI (which means there will more likely already be a VM available and VMs are not as likely to be decommissioned/recommissioned due to changes in our load).

Using OS name "Ubuntu" defaults to 14.04: https://github.com/dotnet/dotnet-ci/blob/master/jobs/generation/Utilities.groovy#L119

See our usage of Ubuntu OS's:

* https://github.com/Microsoft/ChakraCore/blob/master/netci.groovy#L23
* https://github.com/Microsoft/ChakraCore/blob/master/netci.groovy#L218 (14.04 used for style checks)
* https://github.com/Microsoft/ChakraCore/blob/master/netci.groovy#L299 (16.04 for Linux xplat checks)

/cc @MSLaguana @boingoing @mmitche

dilijev:
Testing change to fix #2928

### DIFF
--- a/dummy.cpp
+++ b/dummy.cpp
@@ -1,0 +1,10 @@
+//-------------------------------------------------------------------------------------------------------
+// Copyright (C) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE.txt file in the project root for full license information.
+//-------------------------------------------------------------------------------------------------------
+
+#include <stdio.h>
+int main() {
+    printf("Hello\n");
+}
+


dilijev:
@MSLaguana 

```
Copyright Check — Triggered. (54/59 on ubuntu1604-20170216)
EOL Check — Triggered. (51/59 on ubuntu1604-20170216)
```

Looks like it works :)

PR #2935, opened 2017-05-08
Minor refactor of code accessing the eval map. 

pleath:
Let all callers get the benefit of debug checks, etc.

### DIFF
--- a/lib/Runtime/Base/ScriptContext.cpp
+++ b/lib/Runtime/Base/ScriptContext.cpp
@@ -2310,7 +2310,30 @@ namespace Js
         return success;
     }
 
-    void ScriptContext::AddToEvalMap(FastEvalMapString const& key, BOOL isIndirect, ScriptFunction *pFuncScript)
+    void ScriptContext::AddToEvalMap(FastEvalMapString const& key, BOOL isIndirect, ScriptFunction *pfuncScript)
+    {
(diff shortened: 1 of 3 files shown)


pleath:
@mrkmarron is this correct for the call from NSSnapValues::InflateTopLevelEvalFunctionBodyInfo? (Thanks.)

mrkmarron:
Yes, this should be correct as the guard is `IsTTDRecordOrReplayModeEnabled` which is true if the runtime is started in record or replay mode (even if recording/replaying is temporarily disabled). Clearly if we are in inflating a snapshot we should already be in replay mode.

mrkmarron:
Otherwise LGTM from a TTD standpoint.

lib/Runtime/Base/ScriptContext.cpp
```diff
@@ -2293,7 +2293,30 @@ namespace Js
         return success;
     }
 
-    void ScriptContext::AddToEvalMap(FastEvalMapString const& key, BOOL isIndirect, ScriptFunction *pFuncScript)
+    void ScriptContext::AddToEvalMap(FastEvalMapString const& key, BOOL isIndirect, ScriptFunction *pfuncScript)
```
agarwal-sandeep: LGTM

PR #2936, opened 2017-05-08
Exposing a new JSRT method to get additional information about exceptions

MSLaguana:
This information includes the source file and line/column position of where the exception
came from, as well as the actual source content of the line.

The motivation for this method is for node to print it out on uncaught exceptions.


lib/Jsrt/ChakraCore.h
```diff
@@ -165,6 +165,31 @@ JsGetModuleHostInfo(
...
+///     <para>
+///     Requires an active script context.
+///     </para>
+/// </remarks>
+/// <param name="metadata">The exception metadata for the runtime of the current context.</param>
+/// <returns>
+///     The code <c>JsNoError</c> if the operation succeeded, a failure code otherwise.
+/// </returns>
+CHAKRA_API
+JsGetAndClearExceptionWithMetadata(
```
agarwal-sandeep: >JsGetAndClearExceptionWithMetadata [](start = 0, length = 34)

@liminzhu Should this be in ChakraCommon.h instead of core only?

MSLaguana: At the moment this is still experimental and not yet officially supported, which is why I put it in `ChakraCore.h`

lib/Jsrt/Jsrt.cpp
```diff
@@ -2511,6 +2512,92 @@ CHAKRA_API JsHasException(_Out_ bool *hasException)
     return JsNoError;
 }
 
+CHAKRA_API JsGetAndClearExceptionWithMetadata(_Out_ JsValueRef *metadata)
```
obastemur: Apart from the return part, this implementation looks like an exact match to `JsGetAndClearException` implementation below. In order not to duplicate, could it be better to define a common function and make these two public methods using that?

MSLaguana: They do share a common prelude, but they do slightly different things for TTD, and I need to reference a few things from the prelude in the latter part of this function like the `scriptContext` and the `recordedException`. We could split the prelude out into a common function but it would be very specific.

obastemur:
Is it possible to add a test under `test/native-tests` using this new interface? or using it from CH and a test touching it somehow? ATM CI doesn't test this new interface and if we break it later, we won't know.

MSLaguana:
I've extended one of the tests in `JsRTApiTest.cpp` to cover the new method.

jianchun:
@MSLaguana A thought of API design. Instead of this, how about keep using the existing `JsGetAndClearException`, but add a new API like `JsGetExceptionMetaData(exception)` in which to populate the info you needed? That'll avoid deprecating API and duplicating code.

MSLaguana:
How do you see that flow working? The additional information that we need is accessed through the `JavascriptExceptionObject` which is cleared and forgotten in `JsGetAndClearException`. It isn't a javascript value so I don't believe that I can attach it to the `exception` object, which is a javascript value (and may be any javascript value since it is whatever was thrown).

jianchun:
> How do you see that flow working? The additional information that we need is accessed through the JavascriptExceptionObject which is cleared and forgotten in JsGetAndClearException. It isn't a javascript value so I don't believe that I can attach it to the exception object, which is a javascript value (and may be any javascript value since it is whatever was thrown).

Indeed that is a problem. I can only think of an ugly workaround that may have problems (use an `InternalProperty` to store the `exceptionRecord` in the `thrownObject`. Give up, or do ToObject on `thrownObject` if it is a tagged number.)

Your approach seems fine.

liminzhu:
What's the difference in behavior between `JsGetAndClearExceptionWithMetadata` and `JsGetAndClearException`? Is the former exactly same as latter except for extra metadata info in the returned object?

lib/Jsrt/ChakraCore.h
```diff
@@ -165,6 +165,31 @@ JsGetModuleHostInfo(
 
 #ifdef CHAKRACOREBUILD_
 /// <summary>
+///     Returns metadata relating to the exception that caused the runtime of the current context
+///     to be in the exception state and resets the exception state for that runtime. The metadata
+///     includes a reference to the exception itself.
+/// </summary>
+/// <remarks>
```
liminzhu: Can you lay out the returned `metadata` object in `<remarks>`? I've always hated that I've no clue what's in `exception` of [JsGetAndClearException](https://github.com/Microsoft/ChakraCore/wiki/JsGetAndClearException) without looking into source code.

lib/Runtime/Library/JavascriptExceptionMetadata.cpp
```diff
@@ -0,0 +1,144 @@
...
+        charcount_t startCharOffset;
+        charcount_t endCharOffset;
+
+        if (line >= cache->GetLineCount())
+        {
+            return false;
+        }
+        startCharOffset = cache->GetCharacterOffsetForLine(line, &startByteOffset);
+
+        if (line + 1 >= cache->GetLineCount())
```
akroshg: can you use unit32math helper ?

lib/Runtime/Library/JavascriptExceptionMetadata.cpp
```diff
@@ -0,0 +1,144 @@
...
+            endCharOffset = functionBody->LengthInChars();
+        }
+        else
+        {
+            endCharOffset = cache->GetCharacterOffsetForLine(line + 1, &endByteOffset);
+
+            // The offsets above point to the start of the following line,
+            // while we need to find the end of the current line.
+            // To do so, just step back over the preceeding newline character(s)
+            if (functionSource[endByteOffset - 1] == _u('\n'))
```
akroshg: could this be 0 ? assertorfailfast in beginning?

lib/Runtime/Library/JavascriptExceptionMetadata.cpp
```diff
@@ -0,0 +1,144 @@
...
+
+        Js::JavascriptOperators::OP_SetProperty(metadata, Js::PropertyIds::length,
+            Js::JavascriptNumber::New(0, scriptContext), scriptContext);
+
+        Js::JavascriptOperators::OP_SetProperty(metadata, Js::PropertyIds::url,
+            Js::JavascriptString::NewCopySz(functionBody->GetSourceContextInfo()->url, scriptContext), scriptContext);
+
+        Js::Utf8SourceInfo* sourceInfo = functionBody->GetUtf8SourceInfo();
+        sourceInfo->EnsureLineOffsetCache();
+        JsUtil::LineOffsetCache<Recycler> *cache = sourceInfo->GetLineOffsetCache();
```
akroshg: Did you consider to move and encapsulate this to functionbody?

lib/Runtime/Library/JavascriptExceptionMetadata.cpp
```diff
@@ -0,0 +1,144 @@
...
+        LPCUTF8 functionSource = sourceInfo->GetSource(_u("Jsrt::JsExperimentalGetAndClearExceptionWithMetadata"));
+
+        charcount_t startByteOffset;
+        charcount_t endByteOffset;
+        charcount_t startCharOffset;
+        charcount_t endCharOffset;
+
+        if (line >= cache->GetLineCount())
+        {
+            return false;
```
akroshg: if we return false here, all the setproperty we have done above is wasted, isn't it? make sense to move this to top of the function.

lib/Jsrt/Jsrt.cpp
akroshg: did you consider using ContextAPIWrapper, it does bunch of checks already which you are performing in this function?

---
In reply to: [115546477](https://github.com/Microsoft/ChakraCore/pull/2936#discussion_r115546477) [](ancestors = 115546477)

lib/Jsrt/Jsrt.cpp
```diff
@@ -2511,6 +2512,92 @@ CHAKRA_API JsHasException(_Out_ bool *hasException)
...
+
+    return ContextAPIWrapper<false>([&](Js::ScriptContext* scriptContext, TTDRecorder& _actionEntryPopper) -> JsErrorCode {
+        Js::Var exceptionMetadata = Js::JavascriptExceptionMetadata::CreateMetadataVar(scriptContext);
+        Js::JavascriptOperators::OP_SetProperty(exceptionMetadata, Js::PropertyIds::exception, exception, scriptContext);
+
+        Js::FunctionBody *functionBody = recordedException->GetFunctionBody();
+        if (functionBody == nullptr)
+        {
+            // This is probably a parse error. We can get the error location metadata from the thrown object.
+            Js::JavascriptExceptionMetadata::PopulateMetadataFromCompileException(exceptionMetadata, exception, scriptContext);
```
akroshg: do you have the test case for this syntax error part in the jsRTApiTest?

bin/NativeTests/JsRTApiTest.cpp
```diff
@@ -751,6 +754,39 @@ namespace JsRTApiTest
         REQUIRE(JsSetException(exception) == JsNoError);
         REQUIRE(JsHasException(&value) == JsNoError);
         CHECK(value == true);
+
+        REQUIRE(JsGetAndClearExceptionWithMetadata(&exceptionMetadata) == JsNoError);
```
akroshg: would be nice to verify this API in the case where there is no exception thrown?

lib/Runtime/Library/JavascriptExceptionMetadata.cpp
MSLaguana: This must not be 0 since it is the offset for at least the second line. However, the next check could potentially be 0, so I've added a check there.

MSLaguana:
@liminzhu `JsGetAndClearException` returns the thrown object from an exception while also clearing the current exception, and exposes no other information. `JsGetAndClearExceptionWithMetadata` has the same side effect of clearing the current exception, but returns a javascript object which has the thrown object as a property (the `exception` property) as well as other information: line number, column number, length, line of source code, and script name.

bin/NativeTests/JsRTApiTest.cpp
MSLaguana: Done, which required some fixes in the JSRT functions.

lib/Jsrt/Jsrt.cpp
MSLaguana: I do now.

lib/Runtime/Library/JavascriptExceptionMetadata.cpp
MSLaguana: Done.

lib/Jsrt/ChakraCore.h
MSLaguana: Done.

liminzhu:
@MSLaguana gotcha. Is there any reason to use the old API when the new one's added? From an API design perspective, how do you feel about using and old API instead and piling new properties on the returned err object?

MSLaguana:
If you only care about the thrown object, then it's slightly cheaper to invoke the original method, but once this is accepted then I'll be changing node-chakracore to use only the new method since it always wants the additional information.

I would have preferred to re-use the previous method, but that's not possible. The object returned in the original method is the thrown javascript object, which can be any javascript object including a number, so it may not be something with properties (e.g. it may be a number).

liminzhu:
@MSLaguana makes sense. Thanks for the explanation.

lib/Runtime/Library/JavascriptExceptionMetadata.cpp
```diff
@@ -0,0 +1,163 @@
...
+                }
+                else if (functionSource[endByteOffset - 1] == _u('\r'))
+                {
+                    endCharOffset--;
+                    endByteOffset--;
+                }
+                else
+                {
+                    // This should not be possible
+                    return false;
```
akroshg: assert here, just to catch those issues?

lib/Runtime/Library/JavascriptExceptionMetadata.cpp
```diff
@@ -0,0 +1,163 @@
...
+
+        Js::JavascriptOperators::OP_SetProperty(metadata, Js::PropertyIds::length,
+            Js::JavascriptNumber::New(0, scriptContext), scriptContext);
+
+        Js::JavascriptOperators::OP_SetProperty(metadata, Js::PropertyIds::url,
+            Js::JavascriptString::NewCopySz(functionBody->GetSourceContextInfo()->url, scriptContext), scriptContext);
+
+        LPCUTF8 functionSource = sourceInfo->GetSource(_u("Jsrt::JsExperimentalGetAndClearExceptionWithMetadata"));
+
+        charcount_t startByteOffset;
```
akroshg: nit: initialize locals.

lib/Runtime/Library/JavascriptExceptionMetadata.cpp
```diff
@@ -0,0 +1,163 @@
...
+        charcount_t endByteOffset;
+        charcount_t startCharOffset;
+        charcount_t endCharOffset;
+
+        
+        startCharOffset = cache->GetCharacterOffsetForLine(line, &startByteOffset);
```
akroshg: did you consider moving this whole thing to function body and encapsulate there?

akroshg:
:shipit:

liminzhu:
Added API reference in wiki. Please update them if needed in the future.

https://github.com/Microsoft/ChakraCore/wiki/JsGetAndClearExceptionWithMetadata

PR #2939, opened 2017-05-08
some changes for SharedArrayBuffer

leirocks:
- Due to spec change, SharedArrayBuffer is not transferrable, remove related code
- Hardening the allocation path for OOM case
- Fix a possible race issue in waiter list


lib/Runtime/Language/JavascriptOperators.cpp
```diff
@@ -5351,13 +5351,12 @@ namespace Js
...
         {
-        case TypeIds_SharedArrayBuffer:
-            return Js::SharedArrayBuffer::NewFromSharedState(state, library);
-        case TypeIds_ArrayBuffer:
             return Js::ArrayBuffer::NewFromDetachedState(state, library);
-        default:
+        }
+        else
+        {
             AssertMsg(false, "We should explicitly have a case statement for each object which has detached state.");
```
akroshg: convert to fatal error?

lib/Runtime/Library/SharedArrayBuffer.cpp
```diff
@@ -23,6 +23,15 @@ namespace Js
     }
 #endif
 
+    uint SharedContents::AddRef()
+    {
+        return InterlockedIncrement(&refCount);
+    }
+    uint SharedContents::Release()
+    {
+        return InterlockedDecrement(&refCount);
```
akroshg: assert of failfast of refCount should be 0 to begin with?

lib/Runtime/Library/SharedArrayBuffer.cpp
```diff
@@ -452,12 +452,9 @@ namespace Js
             {
                 HeapDeleteArray(sharedContents->bufferLength, sharedContents->buffer);
             }
-
-            if (!isShutdown)
-            {
-                Recycler* recycler = GetType()->GetLibrary()->GetRecycler();
-                recycler->ReportExternalMemoryFree(sharedContents->bufferLength);
-            }
+
```
obastemur: why this fails on Linux?

lib/Runtime/Language/JavascriptOperators.cpp
akroshg: nit: you can do AssertOrFailFastMsg(state->GetTypeId() != TypeIds_ArrayBuffer, "blah blah"); This will be lot cleaner. 

---
In reply to: [115386419](https://github.com/Microsoft/ChakraCore/pull/2939#discussion_r115386419) [](ancestors = 115386419)

lib/Runtime/Library/SharedArrayBuffer.cpp
```diff
@@ -23,6 +23,20 @@ namespace Js
...
+    long SharedContents::AddRef()
+    {
+        return InterlockedIncrement(&refCount);
+    }
+    long SharedContents::Release()
+    {
+        long ret = InterlockedDecrement(&refCount);
+        if (ret < 0)
+        {
+            Js::Throw::FatalInternalError();
```
akroshg: nit: you can do AssertOrFailFastMsg (ret >= 0).

akroshg:
:shipit:

lib/Runtime/Library/SharedArrayBuffer.cpp
leirocks: didn't look into the detail but I guess somehow in Linux the finalGC is not called in linux.

obastemur: @jianchun maybe I remember wrong but.. do you think this has anything to do with exit time cleanup we were dealing?

obastemur: > didn't look into the detail but I guess somehow in Linux the finalGC is not called in linux.

If  that's the case, shall we make it non _WIN32 only and add TODO investigate?

lib/Common/Memory/Allocator.h
```diff
@@ -227,6 +227,7 @@ struct AllocatorInfo
 template <typename TAllocator>
 struct ForceNonLeafAllocator
 {
+    static const bool FakeZeroLengthArray = true;
```
curtisman: Why?

lib/Common/Memory/Recycler.h
```diff
@@ -2318,9 +2318,21 @@ class TypeAllocatorFunc<Recycler, T *> : public _RecyclerAllocatorFunc<_Recycler
 #endif
 
 // Dummy class to choose the allocation function
-class RecyclerLeafAllocator;
-class RecyclerNonLeafAllocator;
-class RecyclerWriteBarrierAllocator;
+class RecyclerLeafAllocator
+{
+public:
+    static const bool FakeZeroLengthArray = true;
```
curtisman: Why?

lib/Common/Memory/Allocator.h
```diff
@@ -382,16 +384,19 @@ void DestructArray(size_t count, T* obj)
 template <typename TAllocator, typename T>
 void DeleteArray(typename AllocatorInfo<TAllocator, T>::AllocatorType * allocator, size_t count, T * obj)
 {
-    if (count == 0)
+    if (count == 0 && TAllocator::FakeZeroLengthArray)
```
curtisman: Should this be AllocatorInfo<TAllocator, T>::AllocatorType::FakeZeroLengthArray instead?

lib/Common/Memory/Allocator.h
leirocks: see https://github.com/Microsoft/ChakraCore/pull/2939/commits/5224e0583a1da50c529e1bbfbf67d81c04354a39

while Heap allocation zero length array, we allocated the track data, but we didn't delete the track data in DeleteArray, causes memory leak in chk build.

with changing DeleteArray to use the FakeZeroLengthArray, every allocator need to provide such definition.

lib/Common/Memory/Allocator.h
leirocks: hmm, could be, then I can remove the FakeZeroLengthArray on RecyclerLeafAllocator I guess.

leirocks: AllocatorInfo<TAllocator, T>::AllocatorType is just same as TAllocator...

we care because for Heap allocated zero length array, we still allocate memory for the tracking data. and for Heap allocator the tracking data is in front of the user address, need to call into the free function to clear the tracking data and reset the memory leaking status

PR #2941, opened 2017-05-09
Fixing the evaulation order of default parameters in generator functions

aneeshdk:
Currently the default parameter expressions are evaluated as part of the body. So when generator function is called (which creates the generator object) default parameter expressions are not executed. This changelist is to fix this behavior. A yield opcode is inserted at the beginning of the function body and after creating generator object a next call is made.


aneeshdk:
@dotnet-bot test OSX _no_jit_shared_osx_osx_test please

aneeshdk:

aneeshdk:

lib/Runtime/ByteCode/ByteCodeEmitter.cpp
```diff
@@ -3359,6 +3360,16 @@ void ByteCodeGenerator::EmitOneFunction(ParseNode *pnode)
             }
         }
 
+        // If the function has non simple parameter list, the params needs to be evaluated when the generator object is created
+        // (that is when the function is called). This yield opcode is to mark the  begining of the function body.
```
pleath: Looks good. Could you just add a todo comment here to confirm that there's no perf hit from the additional next call? (Thanks.)

PR #2947, opened 2017-05-09
xplat: minor changes

obastemur:
- do not check QCMoreFreq.. on non mobile platforms
   Removes redundant check on unrelated platforms.

- use force marshal instead of double condition
   If scriptContext doesn't match, no point comparing it again.


agarwal-sandeep:
LGTM

obastemur:
@agarwal-sandeep Thanks for the review!

PR #2948, opened 2017-05-09
Add JSRT weak reference APIs

jasongin:
- Add a new typedef: `JsWeakRef`
 - Add two new APIs: `JsCreateWeakReference()`, `JsGetWeakReferenceValue()`
 - Add a native test for the new APIs

The weak reference APIs are required for N-API in Node-ChakraCore. See https://github.com/nodejs/node-chakracore/pull/238


lib/Jsrt/Jsrt.cpp
```diff
@@ -2890,6 +2890,40 @@ CHAKRA_API JsSetPromiseContinuationCallback(_In_ JsPromiseContinuationCallback p
...
+}
+
+CHAKRA_API JsGetWeakReferenceValue(
+    _In_ JsWeakRef weakRef,
+    _Out_ JsValueRef* value)
+{
+    VALIDATE_JSREF(weakRef);
+
+    return GlobalAPIWrapper_NoRecord([&]() -> JsErrorCode {
+        Memory::RecyclerWeakReference<char>* recyclerWeakReference = reinterpret_cast<Memory::RecyclerWeakReference<char>*>(weakRef);
```
kfarnung: Can we add a break on this line to make it easier to read?

lib/Jsrt/Jsrt.cpp
```diff
@@ -2890,6 +2890,40 @@ CHAKRA_API JsSetPromiseContinuationCallback(_In_ JsPromiseContinuationCallback p
     /*allowInObjectBeforeCollectCallback*/true);
 }
 
+CHAKRA_API JsCreateWeakReference(
+    _In_ JsValueRef value,
+    _Out_ JsWeakRef* weakRef)
```
agarwal-sandeep: >weakRef [](start = 21, length = 7)

PARAM_NOT_NULL check for out parameters in both APIs

lib/Jsrt/Jsrt.cpp
```diff
@@ -2890,6 +2890,40 @@ CHAKRA_API JsSetPromiseContinuationCallback(_In_ JsPromiseContinuationCallback p
...
+CHAKRA_API JsGetWeakReferenceValue(
+    _In_ JsWeakRef weakRef,
+    _Out_ JsValueRef* value)
+{
+    VALIDATE_JSREF(weakRef);
+
+    return GlobalAPIWrapper_NoRecord([&]() -> JsErrorCode {
+        Memory::RecyclerWeakReference<char>* recyclerWeakReference = reinterpret_cast<Memory::RecyclerWeakReference<char>*>(weakRef);
+        *value = reinterpret_cast<JsValueRef>(recyclerWeakReference->Get());
+        return JsNoError;
```
kfarnung: Does it make sense to return a specific error code when the reference was already released?  Having it return an invalid reference seems ok, I just wasn't sure whether this was consistent with other APIs.

/cc @liminzhu

lib/Jsrt/Jsrt.cpp
```diff
@@ -2890,6 +2890,40 @@ CHAKRA_API JsSetPromiseContinuationCallback(_In_ JsPromiseContinuationCallback p
     /*allowInObjectBeforeCollectCallback*/true);
 }
 
+CHAKRA_API JsCreateWeakReference(
```
agarwal-sandeep: >JsCreateWeakReference [](start = 11, length = 21)

I think this API should not be callable during collection callbacks as this allocates memory.

test/native-tests/test-weakref/sample.cpp
```diff
@@ -0,0 +1,71 @@
...
+    // Now set the current execution context.
+    JsSetCurrentContext(context);
+
+    JsValueRef valueRef;
+    FAIL_CHECK(JsCreateString("sample", strlen("sample"), &valueRef));
+
+    JsWeakRef weakRef;
+    FAIL_CHECK(JsCreateWeakReference(valueRef, &weakRef));
+
+    FAIL_CHECK(JsGetWeakReferenceValue(weakRef, &valueRef));
```
MSLaguana: Would it make sense to fetch the value to a new reference, and check that `valueRef == weakRefValue`?

lib/Jsrt/Jsrt.cpp
```diff
@@ -2890,6 +2890,40 @@ CHAKRA_API JsSetPromiseContinuationCallback(_In_ JsPromiseContinuationCallback p
...
+    return GlobalAPIWrapper_NoRecord([&]() -> JsErrorCode {
+        ThreadContext* threadContext = ThreadContext::GetContextForCurrentThread();
+        if (threadContext == nullptr)
+        {
+            return JsErrorNoCurrentContext;
+        }
+
+        Recycler* recycler = threadContext->GetRecycler();
+        Memory::RecyclerWeakReference<char>* recyclerWeakReference =
+            recycler->CreateWeakReferenceHandle<char>(reinterpret_cast<char*>(value));
```
agarwal-sandeep: >CreateWeakReferenceHandle [](start = 22, length = 25)

Should it be ok to use FindOrCreateWeakReferenceHandle instead of CreateWeakReferenceHandle?

lib/Jsrt/Jsrt.cpp
agarwal-sandeep: >weakRef [](start = 21, length = 7)

Initialize weakRef to nullptr

lib/Jsrt/Jsrt.cpp
```diff
@@ -2890,6 +2890,40 @@ CHAKRA_API JsSetPromiseContinuationCallback(_In_ JsPromiseContinuationCallback p
...
+        *weakRef = reinterpret_cast<JsWeakRef>(recyclerWeakReference);
+        return JsNoError;
+    });
+}
+
+CHAKRA_API JsGetWeakReferenceValue(
+    _In_ JsWeakRef weakRef,
+    _Out_ JsValueRef* value)
```
agarwal-sandeep: >value [](start = 22, length = 5)

Initialize value to JS_INVALID_REFERENCE

lib/Jsrt/Jsrt.cpp
```diff
@@ -2890,6 +2890,40 @@ CHAKRA_API JsSetPromiseContinuationCallback(_In_ JsPromiseContinuationCallback p
...
+{
+    VALIDATE_JSREF(weakRef);
+
+    return GlobalAPIWrapper_NoRecord([&]() -> JsErrorCode {
```
MSLaguana: Is GC deterministic? What would happen in a TTD scenario when the object backing a weak reference is collected at different points?

MSLaguana:
Looks like the OSX static test failed your freshly added test; after invoking the GC the reference wasn't invalidated.

MSLaguana:

lib/Jsrt/Jsrt.cpp
jasongin: I'm not sure. I think @boingoing actually wrote this code originally. Taylor can you comment on this?

lib/Jsrt/Jsrt.cpp
jasongin: I would not consider it a _failure_ when the weak-referenced value is not available, so I don't think it should return an error code then.

kfarnung: That's true and consistent with other libraries, there just aren't many APIs in JSRT where the out param can be JS_INVALID_REFERENCE after a successful call.  I think the documentation is pretty clear, so that should cover it.

lib/Jsrt/Jsrt.cpp
jasongin: I'll add a check for that.

lib/Jsrt/Jsrt.cpp
jasongin: I don't know enough about how TTD works to answer that. I'd appreciate input from someone else about how weak references should interact with TTD.

jasongin:
> Looks like the OSX static test failed your freshly added test; after invoking the GC the reference wasn't invalidated.

I see ubuntu_linux_debug_static failed the same way. I ran the test (and it passed) on Ubuntu with a release build. Maybe the GC behavior is different with a debug build? I'll try to repro.

lib/Jsrt/Jsrt.cpp
MSLaguana: @mrkmarron can hopefully offer an opinion when he's got the time

jasongin:
I confirmed the test passes with a Release build and fails with a Debug build. I assume because the GC has different configuration in a debug build. I don't know yet what to do about that.

jasongin:
Is the test even in the right place? I just noticed there are other (more complete?) JSRT tests at [`bin/NativeTests/JsRTApiTest.cpp`](https://github.com/Microsoft/ChakraCore/blob/release/2.0/bin/NativeTests/JsRTApiTest.cpp). Should I move my test over to there?

kunalspathak:
@jasongin - That should be the right place to add jsrt native test.

jasongin:
When I put the test into `bin/NativeTests/JsRTApiTest.cpp`, it passes on Windows with _both_ debug and release builds. But, those tests are Windows-only, is that OK?

Why is there native test code in 2 places, separated for Windows and non-Windows platforms?

kunalspathak:
@akroshg  - do you know why?

akroshg:
Why not add test in core\bin\NativeTests ?

jasongin:
@akroshg `bin/NativeTests` is what I just mentioned above. But those tests are Windows-only, aren't they? At least I don't see any way to build them on non-windows platforms. There is only a `.vcxproj` file.

I want to ensure the new APIs work on all supported platforms. Originally in this PR I had put tests under then `test/native-tests` directory, which seems a more obvious place to find tests than a `bin` directory.

akroshg:
They were written in different branches. the windows only depends on the catch framework and during that time the cross-plat was not picked up. However there were compilation and other sort of problems happened when we directly use the windows to crossplat which needed to be looked at. (It just we haven't  got time to consolidate them yet). 

---
In reply to: [300545940](https://github.com/Microsoft/ChakraCore/pull/2948#issuecomment-300545940) [](ancestors = 300545940)

akroshg:
@  yes they are for windows only (although no effort made to consolidate them). 

NativeTest were put along with ch under the bin directory. When I did that I put there so that all application (test or product) will be found under bin folder. But it seems like it can be confusing.

---
In reply to: [300549055](https://github.com/Microsoft/ChakraCore/pull/2948#issuecomment-300549055) [](ancestors = 300549055)

jasongin:
I pushed a couple commits to this PR, to address feedback above and move the test to `bin/NativeTests`.

Remaining issues are the [question about `FindOrCreateWeakReferenceHandle`](https://github.com/Microsoft/ChakraCore/pull/2948#pullrequestreview-37153722) and [concern about interaction with TTD](https://github.com/Microsoft/ChakraCore/pull/2948#discussion_r115602232)

agarwal-sandeep:
:shipit:

liminzhu:
Do APIs get shipped to Chakra when in `ChakraCommon.h`? If that's the case might want to consider just putting the APIs in ChakraCore first in `ChakraCore.h`. Also curious how is this used in Node? Got kind of lost on https://github.com/nodejs/node-chakracore/pull/238.

jasongin:
> Also curious how is this used in Node?

This is necessary to implement the [reference functions in the new Node.js addon API](https://nodejs.org/api/n-api.html#n_api_references_to_objects_with_a_lifespan_longer_than_that_of_the_native_method) in Node-ChakraCore. Those reference APIs are used by native addons that hold a reference to JavaScript values that have associated native data, such as a typed array with external data, or a [native class instance that is "wrapped" in a JavaScript object](https://nodejs.org/api/n-api.html#n_api_object_wrap). The _weak_ references are useful because often the native code only needs to access the value as long as some JS code is still holding it.

In Node-V8, those APIs are implemented using a `v8::Persistent`, which supports weak-reference functionality via [`v8::PersistentBase::SetWeak()`](https://v8docs.nodesource.com/node-7.4/d4/dca/classv8_1_1_persistent_base.html#a9a1e1d92935d6fac29091cff63592854).

liminzhu:
@jasongin exactly what I look for, thanks! Do you imagine an additional finalizer param for `JsCreateWeakReference` called upon invalidating the weak reference useful, similar to what [`v8::PersistentBase::SetWeak()`](https://v8docs.nodesource.com/node-7.4/d4/dca/classv8_1_1_persistent_base.html#a9a1e1d92935d6fac29091cff63592854) does?

jasongin:
> Do you imagine an additional finalizer param for JsCreateWeakReference called upon invalidating the weak reference useful, similar to what v8::PersistentBase::SetWeak() does?

That would be redundant because [`JsSetObjectBeforeCollectCallback`](https://github.com/Microsoft/ChakraCore/wiki/JsSetObjectBeforeCollectCallback) already provides that functionality.

jasongin:
> Do APIs get shipped to Chakra when in `ChakraCommon.h`? If that's the case might want to consider just putting the APIs in ChakraCore first in `ChakraCore.h`

What does everyone think? While these APIs should work fine in Chakra, I don't know any reason they will be really needed there. (There's no plan to implement N-API for Node-Chakra AFAIK.)

liminzhu:
@jasongin my argument - NAPI is still experimental and in case we need to change the new JSRT APIs for w/e reason it's a lot easier when they're just in ChakraCore.

I want them in Chakra eventually ofc. In general I don't like API differences between Chakra and ChakraCore unless there's legitimate reason not to ship something to Chakra, but there's no need to rush for the reason you mentioned. My mental model is to 'stage' experimental APIs in ChakraCore and once they mature we push them in Chakra.

jasongin:
I pushed another commit:
 - Moved new APIs from `ChakraCommon.h` to `ChakraCore.h`, and made them conditional on `CHAKRACOREBUILD_`
 - Using `FindOrCreateWeakReferenceHandle()` instead of `CreateWeakReferenceHandle()`

jasongin:
I think the TTD issue should not be blocking for now. These changes can only impact TTD if the new APIs are called, and in the short term that's only likely to be done via N-API in Node-ChakraCore, and N-API is still officially experimental.

lib/Jsrt/JsrtCommonExports.inc
```diff
@@ -103,6 +103,8 @@
     JsGetRuntime
     JsIdle
     JsSetPromiseContinuationCallback
+    JsCreateWeakReference
+    JsGetWeakReferenceValue
```
MSLaguana: Since you moved the headers to `ChakraCore.h`, you should also move these lines below the `#ifndef` below as well, I believe.

jasongin: Fixed!

liminzhu:
Added API reference in wiki. Please update them if needed in the future.

https://github.com/Microsoft/ChakraCore/wiki/JsWeakRef
https://github.com/Microsoft/ChakraCore/wiki/JsCreateWeakReference
https://github.com/Microsoft/ChakraCore/wiki/JsGetWeakReferenceValue

PR #2894, opened 2017-05-01
Implement IsOpndNegZero in the JIT

meg-gupta:

PR #2950, opened 2017-05-10
IsOpndNegZero bugfix + direct branching

Cellule:
There was a bug on x64 for float32 IsNegZero
the conversion to uint32 was not done correctly (MOVQ with uint32 and float32 didn't work).

I took the opportunity to reuse `LowerReinterpretPrimitive` for the conversion and `FindOrCreateInt64Pair` for the x86 double case.
This allowed me to move the code back to LowerMDShared.cpp instead of arch specific implementation between x64 and x86.

Added some test cases to cover the problematic case.
Related #2894 

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/2950)


Cellule:
@meg-gupta @MikeHolman Can you please review this ?
Also, be careful with the arm changes, I think I got it right, but it's hard to test

lib/Backend/arm64/LowerMD.h
```diff
@@ -233,7 +233,6 @@ class LowererMD
               template <bool verify = false>
               static void         Legalize(IR::Instr *const instr, bool fPostRegAlloc = false) { __debugbreak(); }
 
-              IR::Opnd*           IsOpndNegZero(IR::Opnd* opnd, IR::Instr* instr) { __debugbreak(); return 0; }
```
Cellule: I don't think we need to implement my new apis here since they're only used internally

lib/Backend/LowerMDShared.cpp
```diff
@@ -8497,17 +8497,22 @@ LowererMD::LowerReinterpretPrimitive(IR::Instr* instr)
 #elif _M_IX86
```
meg-gupta: nit: I think it will be helpful to spot this api LowerReinterpretPrimitive if you rename to the lines of other LowerMDShared converts like EmitInt64toFloat, EmitFloatToInt etc?

Cellule: Yeah, this was initially for wasm `reinterpret` operator.
It converts an int64 to double or double to int64 or int32 to float or float to int32, so I'm not sure exactly how I would name it.

Cellule:
@meg-gupta  I have reworked `LowerReinterpretPrimitive` to have, I think, a better api, more error checking and useful alias.
Let me know if that makes sense to you

meg-gupta:
@Cellule Looks good. Thanks!

PR #2951, opened 2017-05-10
Add PropertyString pointer to some LiteralStrings

MikeHolman:
There are some types of strings which we convert into LiteralString (e.g. ConcatStrings). However, they generally have a number of fields so there is extra space that is unused after conversion. This change utilizes that space with a PropertyString pointer (which we populate if you use the string as an index for a StElem).

This is about a 2% improvement in Speedometer.


lib/Runtime/Language/JavascriptConversion.cpp
```diff
@@ -305,6 +305,20 @@ namespace Js
                 PropertyString * propertyString = (PropertyString *)propName;
                 *propertyRecord = propertyString->GetPropertyRecord();
             }
+            else if (VirtualTableInfo<Js::LiteralStringWithPropertyStringPtr>::HasVirtualTable(propName))
```
rajatd: >VirtualTableInfo<Js::LiteralStringWithPropertyStringPtr>::HasVirtualTable(propName [](start = 21, length = 82)

Would it make sense to add equivalent code at other places where we check for PropertyStrings? For example, ForInObjectEnumerator::MoveAndGetNext, and TMapKey_ConvertKey and TMapKey_ConvertKey_TTD in SimpleDictionaryTypeHandler.cpp.

lib/Runtime/Library/ConcatString.inl
```diff
@@ -82,6 +91,8 @@ namespace Js
         // Allow slots to be garbage collected if no more refs.
         ClearArray(m_slots, N);
 
+        LiteralStringWithPropertyStringPtr::ConvertString(this);
```
rajatd: >LiteralStringWithPropertyStringPtr::ConvertString [](start = 8, length = 49)

Can we move calls to this function to ConcatStringBase::GetSzImpl?

lib/Runtime/Language/JavascriptConversion.cpp
MikeHolman: Possibly, but I thought by having it here we won't really ever end up with these type of strings inside any type handlers.

lib/Runtime/Library/ConcatString.inl
MikeHolman: We could, but I didn't like the idea of doing work with fields of the old class after changing type.

rajatd:
:shipit:

PR #2956, opened 2017-05-10
WASM: int64 cse and const prop

Cellule:
- ~~Do int64 const prop~~
- Do int64 cse
- Added some tests

This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/2956)


Cellule:
@Penguinwizzard For you to see regarding your upcoming value map change

Cellule:

lib/Backend/GlobOpt.cpp
```diff
@@ -6590,6 +6594,22 @@ GlobOpt::CopyProp(IR::Opnd *opnd, IR::Instr *instr, Value *val, IR::IndirOpnd *p
...
         return opnd;
     }
+    else if (valueInfo->TryGetIntConstantValue(&int64ConstantValue, false))
+    {
+        if (PHASE_OFF(Js::ConstPropPhase, this->func))
+        {
+            return opnd;
+        }
+
+        if (this->func->GetJITFunctionBody()->IsWasmFunction() && opnd->IsInt64())
```
MikeHolman: this should only happen in case of wasm, right? maybe we should assert?

Cellule:
I started writing an extensive test for unsigned int64 and I ended up with tons of bugs.
Most of them related to const prop. I will have to spend some more time to debug all of this.

LouisLaf:
:shipit:

Cellule:
@MikeHolman @LouisLaf Could you double check my change ?
I made a bunch of fixes for int64 const prop, however in the end I decided to turn it off because there are still too many bugs and it is not ready.
This PR contains some bug fixes, tests and also CSE for int64 so it is worth checking in as it is.

PR #2957, opened 2017-05-11
Fix bugs from PropertyString cache changes

MikeHolman:
Fixes following:
We were not disabling StoreFieldCache for SimpleDictionaryTypeHandle if PropertyDescriptor is not initialized.
We were attempting to cache enumeration of InternalPropertyId in case it was the last property.
We were attempting to cache property reads when object of a LdElem was a tagged number.


MikeHolman:
These are small fixes, so rather than making separate PRs I decided to just throw them all in here.

MikeHolman:

rajatd:
:shipit:

PR #2958, opened 2017-05-11
Fix some issues in CmdLineArgsParser

dilijev:
Add `=` as an equivalent character to `:` in ch arguments lists for things like `-dump:bytecode` (`-dump=bytecode`).

### DIFF
--- a/lib/Common/Core/CmdParser.cpp
+++ b/lib/Common/Core/CmdParser.cpp
@@ -68,11 +68,12 @@ CmdLineArgsParser::ParseString(__inout_ecount(ceBuffer) LPWSTR buffer, size_t ce
             case ' ':
             case ',':
             case 0:
-                fDone = 1;
+                fDone = true;
                 break;
             case '-':
+            case '=':
             case ':':
-                if(fTreatColonAsSeparator)
+                if (fTreatColonAsSeparator)
                 {
                     fDone = true;
                     break;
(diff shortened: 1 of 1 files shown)


dilijev:
FYI @navidR

dilijev:

lib/Common/Core/CmdParser.cpp
```diff
@@ -68,11 +68,12 @@ CmdLineArgsParser::ParseString(__inout_ecount(ceBuffer) LPWSTR buffer, size_t ce
             case ' ':
             case ',':
             case 0:
-                fDone = 1;
+                fDone = true;
                 break;
-            case '-':
+            case '-': // REVIEW is this correct? What syntax results in this path being taken?
```
dilijev: Can someone take a look at this?

jianchun: Could this be the case of parsing the ending `-` in disabling flag like `-OOPJIT-`?

dilijev: Yeah I was thinking it was something like that. Since this isn't actually broken I'm going to assume that this is okay.

lib/Common/Core/CmdParser.cpp
```diff
@@ -68,11 +68,12 @@ CmdLineArgsParser::ParseString(__inout_ecount(ceBuffer) LPWSTR buffer, size_t ce
...
+            case '=':
```
dilijev: Is there any potential negative to adding this as a separator? I don't believe we have a special meaning for `=` in any of our arguments syntax.

dilijev:

PR #2967, opened 2017-05-12
xplat: build compiler toolchain for LTO

obastemur:
Linux: check if `ld` supports LTO, otherwise build
clang toolchain and use that instead

Once this merged, `node-chakracore` builds may set `--lto-thin` argument by default for `Release` builds.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -375,10 +375,26 @@ if [[ ${#_VERBOSE} > 0 ]]; then
 fi
 
 # if LTO build is enabled and cc-toolchain/clang was compiled, use it instead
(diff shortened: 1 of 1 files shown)


digitalinfinity:
@obastemur thanks for adding this- LGTM (although CI needs to pass 😄)

@MSLaguana @kunalspathak FYI- can one of you bring this change over to node-chakracore and update node-chakracore to use the lto flag?

kunalspathak:
Sure, will do.

obastemur:
@digitalinfinity Thanks for the review!

PR #2968, opened 2017-05-12
Reverting MARSHAL_OBJECT change

kfarnung:
The change introduced in https://github.com/Microsoft/ChakraCore/commit/cd3407b40a79a4f63bb08a5f0a26e582d3407e6e
which passes `true` to MarshalVar broke a lot of tests in
node-chakracore. Reverting the change to unblock node-chakracore.

### DIFF
--- a/lib/Jsrt/JsrtInternal.h
+++ b/lib/Jsrt/JsrtInternal.h
@@ -34,7 +34,7 @@ typedef struct {} TTDRecorder;
             {  \
                 return JsErrorWrongRuntime;  \
             }  \
-            p = Js::CrossSite::MarshalVar(scriptContext, __obj, true); \
+            p = Js::CrossSite::MarshalVar(scriptContext, __obj); \
         }
 
 #define VALIDATE_INCOMING_RUNTIME_HANDLE(p) \


agarwal-sandeep:
My bad, this is correct. fRequestWrapper is used inside DoRequestWrapper which now always return true whereas previously it was false as fRequestWrapper was false.

LGTM

obastemur:
:shipit:

digitalinfinity:
I think the intent behind @obastemur's initial change is still valid- we do have a double compare still - it's just that `fRequestWrapper` was not the right proxy to use for comparison here. We could do something like introduce a new function (or templatize MarshalVar?)? Also, @obastemur in your original change, was there a perf win?
```
    Var CrossSite::MarshalVarDifferentScriptContexts(ScriptContext* scriptContext, Var value)
    {
        // value might be null from disable implicit call
        if (value == nullptr || Js::TaggedNumber::Is(value))
        {
            return value;
        }
        Js::RecyclableObject* object =  RecyclableObject::FromVar(value);
        return MarshalVarInner(scriptContext, object, false /* fRequestWrapper */);
    }
```

obastemur:
@digitalinfinity thanks for the details. 

MarshalVar shows significant heat signature (on main thread) due to its use in string conversation interface. While preparing the changes for that part in focus, saw an unnecessary double check and sent it separately. I don't think the comparison there deserves a whole new template interface. However, I'm working on those warm places and may come up with another change on the same interface if there is a perf gain. 

@kfarnung thanks for catching this early.

PR #2971, opened 2017-05-15
Reorganize some code in GlobOpt

Penguinwizzard:
This shouldn't have changed any actual runtime semantics yet, but it
does change pointer use somewhat - initial tests show no perf swing.

The aim in this code is to isolate some components related to the value map, so that the data structure can be changed.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/2971)


Cellule:
Seriously... way too much `const` all over the place. To begin with our code base doesn't use const much except in trivial cases which makes this look messy.
There is Way too much copy/paste for the sake of using `const` and I disagree with this. Either change the original api to const if possible otherwise drop it.

---

Reviewed 18 of 31 files at r1.
Review status: 18 of 31 files reviewed at latest revision, 6 unresolved discussions, some commit checks failed.

---

*[lib/Backend/GlobOpt.h, line 780 at r1](https://reviewable.io:443/reviews/microsoft/chakracore/2971#-KkF45GOc63DyeSpi8X_:-KkF45GOc63DyeSpi8Xa:b-k1ykc6) ([raw file](https://github.com/microsoft/chakracore/blob/5dc1cd3b4dcb68dc6f13e14c4ac3f7d1072c7d66/lib/Backend/GlobOpt.h#L780)):*
> ```objc
>     bool                    DoTypedArrayTypeSpec() const { return GlobOpt::DoTypedArrayTypeSpec(this->func); }
>     bool                    DoNativeArrayTypeSpec() const { return GlobOpt::DoNativeArrayTypeSpec(this->func); }
>     bool                    DoLdLenIntSpec(IR::Instr const * const instr, const ValueType baseValueType) const;
> ```

I think this shouldn't be const. We are doing an optimization here, it is possible and valid to modify the instr.

---

*[lib/Backend/IR.cpp, line 3240 at r1](https://reviewable.io:443/reviews/microsoft/chakracore/2971#-KkF0rI3dz1R9DJx_p2q:-KkF0rI3dz1R9DJx_p2r:b-s8o5bg) ([raw file](https://github.com/microsoft/chakracore/blob/5dc1cd3b4dcb68dc6f13e14c4ac3f7d1072c7d66/lib/Backend/IR.cpp#L3240)):*
> ```cpp
> }
> 
> bool IR::Instr::TransfersSrcValue() const
> ```

nit: We are already in the namespace IR. Remove the extra `IR::`

---

*[lib/Backend/IR.cpp, line 3491 at r1](https://reviewable.io:443/reviews/microsoft/chakracore/2971#-KkF4cYI7rNQnWqPXRBm:-KkF4cYI7rNQnWqPXRBn:b-6g5l1u) ([raw file](https://github.com/microsoft/chakracore/blob/5dc1cd3b4dcb68dc6f13e14c4ac3f7d1072c7d66/lib/Backend/IR.cpp#L3491)):*
> ```cpp
> }
> 
> bool Instr::CallsSetter(IR::PropertySymOpnd const* methodOpnd) const
> ```

Good opportunity to remove the unused parameter here and in CallsGetter

---

*[lib/Backend/IR.h, line 181 at r1](https://reviewable.io:443/reviews/microsoft/chakracore/2971#-KkF4rT8QSRECBNEDOma:-KkF4rT8QSRECBNEDOmb:bdmyh7f) ([raw file](https://github.com/microsoft/chakracore/blob/5dc1cd3b4dcb68dc6f13e14c4ac3f7d1072c7d66/lib/Backend/IR.h#L181)):*
> ```objc
>     bool            IsProfiledInstr() const;
>     ProfiledInstr * AsProfiledInstr();
>     ProfiledInstr const * AsProfiledInstr() const;
> ```

It really feels weird that `AsProfiledInstr` is the only one that has a const version...

---

*[lib/Backend/Opnd.h, line 179 at r1](https://reviewable.io:443/reviews/microsoft/chakracore/2971#-KkF60NV6IcJ4ak0M8y2:-KkF60NWRUUMBLUEmGY2:b-yehimh) ([raw file](https://github.com/microsoft/chakracore/blob/5dc1cd3b4dcb68dc6f13e14c4ac3f7d1072c7d66/lib/Backend/Opnd.h#L179)):*
> ```objc
>     bool                IsRegOpnd() const;
>     RegOpnd *           AsRegOpnd();
>     RegOpnd const *     AsRegOpnd() const;
> ```

`const RegOpnd*` is equivalent to `RegOpnd const *` I would honestly favor the former as it matches more what we have in the code base.

---

*[lib/Backend/Sym.cpp, line 931 at r1](https://reviewable.io:443/reviews/microsoft/chakracore/2971#-KkF82fSgsASzWFllmdA:-KkF82fSgsASzWFllmdB:b-s5ddpv) ([raw file](https://github.com/microsoft/chakracore/blob/5dc1cd3b4dcb68dc6f13e14c4ac3f7d1072c7d66/lib/Backend/Sym.cpp#L931)):*
> ```cpp
> 
> StackSym const *
> StackSym::GetTypeEquivSym_NoCreate(IRType type) const
> ```

This is not needed (also I highly disagree with copy/pasting functions with more than 2 lines of code for sake of const)
Just use `StackSym* StackSym::GetTypeEquivSym_NoCreate(IRType type) const`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/microsoft/chakracore/2971#-:-KkF8UgPFSfMSXxo-DsF:bbk8rmt)*

Penguinwizzard:
Since Reviewable is still erroring out every time I try to reply there:

> Seriously... way too much const all over the place. To begin with our code base doesn't use const much except in trivial cases which makes this look messy.
There is Way too much copy/paste for the sake of using const and I disagree with this. Either change the original api to const if possible otherwise drop it.

I've removed any cases where I copied more than two lines of code for a const version of a method. Should I just drop all of the const stuff entirely? Some of it allows us to get a number of other trivial cases - there's a fairly common pattern of `x->As_()->Is_()` being used in checks, so having `const _*As_() const`  methods makes const actually usable in the situations that use those, since most of our `Is_()` functions are const.

> I think this shouldn't be const. We are doing an optimization here, it is possible and valid to modify the instr.

You're correct, I've removed the const.

> nit: We are already in the namespace IR. Remove the extra `IR::`

Woops, thanks, I've removed the extra `IR::`

> Good opportunity to remove the unused parameter here and in CallsGetter

I've removed the unused argument in the new version.

> It really feels weird that `AsProfiledInstr` is the only one that has a const version...

I've removed the const version of AsProfiledInstr; it was added because it allowed for CallsGetter/CallsSetter/CallsAccessor/TransfersSrcValue to all be const, because of the use of the As_()->Is_() pattern.

> `const RegOpnd*` is equivalent to `RegOpnd const *` I would honestly favor the former as it matches more what we have in the code base.

I've changed this file to use the former everywhere, but the rest of the codebase is honestly somewhat mixed...not to mention that there's a bunch of places where const is used with actually no effect (e.g. "X* const" as parameters to functions), which leads me to believe that a more thorough pass over our use of const is really needed. Additionally, to make things in this file more uniform, I've added the const versions of the other `As_Opnd` functions.

> This is not needed (also I highly disagree with copy/pasting functions with more than 2 lines of code for sake of const)
Just use `StackSym* StackSym::GetTypeEquivSym_NoCreate(IRType type) const`

I've made this a little cleaner by having the non-const version call the const version and explicitly discard the const on return (this is what Effective C++ recommends; it's not terribly pretty, but it avoids duplication and is type safe). I feel that making the function const and return a non-const equiv sym can lead to some unsafe assumptions from the type signatures.

Cellule:
Reviewed 8 of 31 files at r1, 9 of 13 files at r2, 4 of 4 files at r3.
Review status: all files reviewed at latest revision, 4 unresolved discussions.

---

*[lib/Backend/GlobOpt.cpp, line 661 at r3](https://reviewable.io:443/reviews/microsoft/chakracore/2971#-KkHfO-T8ZJtIpmLqI3b:-KkHfO-T8ZJtIpmLqI3c:b-8egfha) ([raw file](https://github.com/microsoft/chakracore/blob/239a4b8156f7cad7ee7f63ea649cf00a58ce5783/lib/Backend/GlobOpt.cpp#L661)):*
> ```cpp
>         }
>     }
>     block->globOptData.hasCSECandidates = this->currentBlock->globOptData.hasCSECandidates;
> ```

Should remove this line

---

*[lib/Backend/GlobOpt.h, line 696 at r3](https://reviewable.io:443/reviews/microsoft/chakracore/2971#-KkHhHCmmTelDainTD7g:-KkHhHCn3kIzMM9mS31P:bkl23nt) ([raw file](https://github.com/microsoft/chakracore/blob/239a4b8156f7cad7ee7f63ea649cf00a58ce5783/lib/Backend/GlobOpt.h#L696)):*
> ```objc
>     JsArrayKills            CheckJsArrayKills(IR::Instr *const instr);
> 
>     GlobOptBlockData const * GetCurrentBlockData() const;
> ```

Should make a non-const version and use it instead of `this->currentBlock->globoptData`

---

*[lib/Backend/GlobOptFields.cpp, line 3001 at r3](https://reviewable.io:443/reviews/microsoft/chakracore/2971#-KkHbuf28C5A1QZ2NhXz:-KkHbuf28C5A1QZ2NhY-:b-skdnsm) ([raw file](https://github.com/microsoft/chakracore/blob/239a4b8156f7cad7ee7f63ea649cf00a58ce5783/lib/Backend/GlobOptFields.cpp#L3001)):*
> ```cpp
> 
> void
> GlobOpt::KillAllObjectTypes(GlobOptBlockData *blockData, BVSparse<JitArenaAllocator>* liveFields)
> ```

Should not pass blockData and just Assert/check that liveFields is not null since all callers right now don't pass null

---

*[lib/Backend/ValueInfo.h, line 324 at r3](https://reviewable.io:443/reviews/microsoft/chakracore/2971#-KkH_2LUdmMIeyGZiFiX:-KkH_2LUdmMIeyGZiFiY:bt4l38t) ([raw file](https://github.com/microsoft/chakracore/blob/239a4b8156f7cad7ee7f63ea649cf00a58ce5783/lib/Backend/ValueInfo.h#L324)):*
> ```objc
> };
> 
> template<> inline ValueNumber JsUtil::ValueToKey<ValueNumber, Value *>::ToKey(Value *const &value);
> ```

You should remove this declaration

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/microsoft/chakracore/2971)*

Penguinwizzard:
> Should remove this line

Removed line.

> Should make a non-const version and use it instead of `this->currentBlock->globoptData`

Added non-const version, did a few find+replace with decreasingly conservative patterns to use it in GlobOpt.cpp.

> Should not pass blockData and just Assert/check that liveFields is not null since all callers right now don't pass null

Stopped passing blockData, handles nullptr args better now.

> You should remove this declaration

Removed.

Cellule:
:lgtm:

---

Reviewed 5 of 5 files at r4.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/microsoft/chakracore/2971#-:-KkIcYFUtH8erw7sjeNl:bnfp4nl)*

PR #2972, opened 2017-05-15
Update .gitignore for VS 2017

dilijev:
It is possible for .vs/ to be created in the repo root by using the "Folder View" feature.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,6 +27,7 @@
 *.vcxproj.user
 *.vspscc
 *.wrn
+.vs/
 
 # Visual Studio Extensions
 *.vadbg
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -36,7 +37,6 @@
 build_*.err
 build_*.log
 build_*.wrn
-Build/.vs/
```
Penguinwizzard: Why remove this? (is it just something that never happens anyway?)

kfarnung: I think it's redundant since `.vs/` doesn't require that the folder be in the root, if it appears anywhere in the folder structure it will be ignored.

dilijev: Exactly. Lines in .gitignore apply to any subdirectory unless you prefix it with a "root" directory `/`.

Penguinwizzard: Ah, right, that makes sense.

PR #2985, opened 2017-05-17
Prevent clearing of closure environment from live eval function object. 

pleath:
We clear the environment from cached eval functions to avoid pinning captured variables. But if the same function is fetched from the cache while another instance of the same eval is executing, then when the newer invocation returns we clear the environment out from under the instance that is still running. Fix this by making sure we use a different function object to execute a direct eval if there is already a running instance, indicated by a non-null environment on the function.


akroshg:
:shipit:

pleath:
@dotnet-bot please test OSX static_osx_osx_test

PR #2988, opened 2017-05-18
xplat: extern non-pal interface for GetTickCount

obastemur:
Provide necessarry interface, so we can enable IdleCollect on node-chakracore

See https://github.com/nodejs/node-chakracore/pull/248

### DIFF
--- a/pal/inc/pal_external.h
+++ b/pal/inc/pal_external.h
@@ -1,0 +1,22 @@
+//-------------------------------------------------------------------------------------------------------
+// Copyright (C) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE.txt file in the project root for full license information.
+//-------------------------------------------------------------------------------------------------------
+
+// The interface on this file can be consumed from none chakracore dependents
+// Keep this file independent from the rest of PAL
+
+#ifndef PAL_INC_PAL_EXTERNAL_H
+#define PAL_INC_PAL_EXTERNAL_H
+#ifndef _WIN32
+
+namespace CCPAL
+{
+    unsigned long GetMilliseconds();
+}
+
+#define GetTickCount()   CCPAL::GetMilliseconds()
+#define GetTickCount64() CCPAL::GetMilliseconds()
+
+#endif // !_WIN32
+#endif // PAL_INC_PAL_EXTERNAL_H
--- a/lib/Common/Memory/Recycler.cpp
+++ b/lib/Common/Memory/Recycler.cpp
@@ -2364,8 +2364,9 @@ void Recycler::ResetPartialHeuristicCounters()
 void
 Recycler::ScheduleNextCollection()
 {
-    this->tickCountNextCollection = ::GetTickCount() + RecyclerHeuristic::TickCountCollection;
-    this->tickCountNextFinishCollection = ::GetTickCount() + RecyclerHeuristic::TickCountFinishCollection;
+    DWORD tick = ::GetTickCount();
+    this->tickCountNextCollection = tick + RecyclerHeuristic::TickCountCollection;
+    this->tickCountNextFinishCollection = tick + RecyclerHeuristic::TickCountFinishCollection;
 }
 
 #if ENABLE_CONCURRENT_GC
@@ -7987,7 +7988,7 @@ Recycler::VerifyMarkStack()
     }
 }
 
-bool 
+bool
 Recycler::VerifyMark(void * target)
 {
     return VerifyMark(nullptr, target);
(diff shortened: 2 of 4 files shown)


obastemur:
Q: Why forcing the same backend for GetTickCount instead of implementing something similar into ChakraShim?

A: Our version of GetTickCount is based on `rdtsc` on x86_64/amd64 arch. We may slightly end-up comparing apples to oranges.

pal/inc/pal_external.h
```diff
@@ -0,0 +1,22 @@
+//-------------------------------------------------------------------------------------------------------
+// Copyright (C) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE.txt file in the project root for full license information.
+//-------------------------------------------------------------------------------------------------------
+
+// The interface on this file can be consumed from none chakracore dependents
```
kfarnung: nit: "non-chakracore dependents"

digitalinfinity:
Thanks for taking a stab at this, @obastemur! I understand why you implemented this, but the real issue is the documentation of JsIdle. JsIdle is documented as returning time in system ticks, but it should really be documented as returning time in milliseconds (which is what the GetTickCount documentation says). Given that, from the host's side, they should just treat the time passed back as milliseconds, and they can call whatever system API to get the current number of milliseconds since epoch.

Given that, I think we should do the following:
1. No changes needed in ChakraCore- don't need this PR
2. In Node-ChakraCore, remove the call to GetTickCount and instead implement something similar

And more generally, the API exposed by ChakraCore should be limited to JSRT- we shouldn't expose PAL APIs from ChakraCore, it's not our API to expose.

cc @curtisman

obastemur:
>  .....but it should really be documented as returning time in milliseconds (which is what the GetTickCount documentation says). Given that, from the host's side, they should just treat the time passed back as milliseconds, and they can call whatever system API to get the current number of milliseconds since epoch.

@digitalinfinity I thought I had answered the `why?` part above.

Well, I read that differently. First of all, `GetTickCounts` (given on MSDN) doesn't return system time and has nothing to do with epoch. It (also on ChakraCore xplat) returns time in ms. and the measurements is based on ticks. However, the method return value is `milliseconds` that corresponds to tick counts in MS.

and -> https://github.com/Microsoft/ChakraCore/pull/2988#issuecomment-302434720

digitalinfinity:
Sorry @obastemur- I still don't understand https://github.com/Microsoft/ChakraCore/pull/2988#issuecomment-302434720

And yes, sorry, epoch was a bad choice of words- I meant that to indicate it returns the number of milliseconds since a fixed point in time (in this case, system startup). As you point out, in both cases (win32 and PAL), the time is returned in milliseconds. Given these, we don't need to document anything about ticks.

The more important principle I want to state is, the only dependency an embedder can take on ChakraCore is an API exposed through JSRT. Given this, I do not think it's a good idea to have Node-ChakraCore take a dependency on GetTickCount implemented in the PAL.

Given the above, can you help me understand what I'm missing? Thanks!

obastemur:
> The more important principle I want to state is, the only dependency an embedder can take on ChakraCore is an API exposed through JSRT. Given this, I do not think it's a good idea to have Node-ChakraCore take a dependency on GetTickCount implemented in the PAL.

@digitalinfinity On one hand, I agree with you. It feels like a cross dependency. On the other hand, the external interface I would like to introduce is something separate from ChakraCore. However, the necessity here has nothing to do with my wishes.

We generate GetTickCount based on rdtsc and our initial estimate on cpu cycles per microseconds. So, lets say;

- T1 ChakraCore time is at 100ms, since then 9.000.000 cpu cycles passed. Our initial measurements was 1500 cycles per 1 microseconds. Imagine external host is measuring it as 1497 or 1505 ?

- External interface can't! use system time because GetTickCount doesn't reflect that. So, it has to use cpu cycles similarly. 

- We can't rely on putting the same implementation there and expect exact same CPU cycles / microsecond ratio. Any difference will result very badly over the time.

- We can't rely on a slower interface backed by the system time because once again, we have to ask embedder to do the same and use the same approach exactly! Besides, it will be slow.

So, my intention is to provide an easy to embed set of header/source files for other developers to use for their cross platform apps and make sure they will have the best experience.

curtisman:
I think the basic thing is that JsIdle should return a time interval value in millisecond.   The host doesn't need to know whether it is a tick or source from something else. How we implement it shouldn't be concern of the host.  As long as there is an universal agreeable time unit, the host can do whatever they want to do to get/calculate this time info and that make this API unnecessary.

obastemur:
How we/host is going to handle time jumps? IMHO, interacting with a separate entity on milliseconds level and not sharing the source of timer doesn't work.

We might give the host an interface to just ping us (whenever host is idle).

obastemur:
This works on Windows because all the applications use the same source for milliseconds measurements. That measurement is not based on real time.

curtisman:
This is a time interval, the source of the timer doesn't matter, as long as it can time things in milliseconds.

dilijev:
This was closed when release/2.0 was deleted. Feel free to reopen.

PR #2991, opened 2017-05-18
Fix perf regression in typed array tests. 

pleath:
If we bail out on array access because we needed a helper call to, for instance, convert the value stored to a typed array, then when we interpret the array access the profile info needs to be updated to indicate that the helper call was needed. Otherwise, on re-jit, we'll generate the same code again, and we'll keep bailing out. Try to be proactive in the case of non-number value stored to a typed array, but also make use of the fact that we had BailOutOnArrayAccessHelperCall, since helper call cases can't be perfectly predicted in the interpreter.

### DIFF
--- a/lib/Runtime/Language/ProfilingHelpers.cpp
+++ b/lib/Runtime/Language/ProfilingHelpers.cpp
@@ -11,7 +11,8 @@ namespace Js
         const Var base,
         const Var varIndex,
         FunctionBody *const functionBody,
-        const ProfileId profileId)
+        const ProfileId profileId,
+        bool didArrayAccessHelperCall)
     {
         Assert(base);
         Assert(varIndex);
@@ -43,6 +44,11 @@ namespace Js
         TypeId arrayTypeId;
         JavascriptArray *const array =
             JavascriptArray::GetArrayForArrayOrObjectWithArray(base, &isObjectWithArray, &arrayTypeId);
+
+        if (didArrayAccessHelperCall)
+        {
+            ldElemInfo.neededHelperCall = true;
+        }
 
         do // while(false)
         {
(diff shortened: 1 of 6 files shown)


lib/Runtime/Language/ProfilingHelpers.cpp
```diff
@@ -271,13 +278,23 @@ namespace Js
             {
                 length = headSegmentLength;
                 bool isVirtual = (VirtualTableInfoBase::GetVirtualTable(base) == ValueType::GetVirtualTypedArrayVtable(arrayTypeId));
-                stElemInfo.arrayType = ValueType::FromTypeId(arrayTypeId, isVirtual).ToLikely();
+                stElemInfo.arrayType = ValueType::FromTypeId(arrayTypeId, isVirtual).ToLikely();        
+                if (!TaggedNumber::Is(value) && !JavascriptNumber::Is_NoTaggedIntCheck(value))
```
rajatd: >Is_NoTaggedIntCheck [](start = 67, length = 19)

We dont need the JavascriptNumber check on amd64, do we?

pleath:
No, I suppose not. Is there an API that does the right machine specific thing under the covers? I'd rather not write #ifdef'd code at this level.

lib/Runtime/Language/ProfilingHelpers.cpp
MikeHolman: I thought TaggedNumber::Is check took care of both int and doubles?

rajatd: it only checks for ints on x86

MikeHolman: oh, duh! of course

rajatd:
@pleath , I couldn't find any existing API that would do what we want here.

pleath:
Thanks for looking. In that case, I'd prefer to stick with something generic for this change. Maybe we could think about adding an API and changing our usage everywhere.

-- Paul

rajatd:
If we do add an API, we may not have to change our usage everywhere as in most of the places we are actually only looking for a tagged value, as opposed to this case where we want to look for both ints and doubles.

pleath:
True. Any other comments on this change?

rajatd:
Sorry, no, no other comments. looked good

pleath:
Thanks.

PR #2992, opened 2017-05-18
WASM: memory 

Cellule:
- Alwasy create a WebAssemblyArrayBuffer for WebAssemblyMemory
- Prevent detaching the WebAssemblyArrayBuffer except in grow_memory scenario
- Fix bug in Interpreter with negative indexes in load/store wasm
- Do not use ArrayBufferRegister slot for wasm, use WebAssemblyMemory directly instead.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/2992)

### DIFF
--- a/lib/Runtime/Library/ArrayBuffer.cpp
+++ b/lib/Runtime/Library/ArrayBuffer.cpp
@@ -950,32 +950,42 @@ namespace Js
 #endif
 
 
-
-    WebAssemblyArrayBuffer::WebAssemblyArrayBuffer(uint32 length, DynamicType * type) :
-#ifndef ENABLE_FAST_ARRAYBUFFER
-        // Treat as a normal JavascriptArrayBuffer
-        JavascriptArrayBuffer(length, type) {}
-#else
-        JavascriptArrayBuffer(length, type, WasmVirtualAllocator)
+    template<typename Allocator>
+    Js::WebAssemblyArrayBuffer::WebAssemblyArrayBuffer(uint32 length, DynamicType * type, Allocator allocator):
+        JavascriptArrayBuffer(length, type, allocator)
     {
+#ifndef ENABLE_FAST_ARRAYBUFFER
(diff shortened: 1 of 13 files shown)


lib/Runtime/Library/ArrayBuffer.cpp
```diff
@@ -998,38 +1017,70 @@ namespace Js
...
 #if ENABLE_FAST_ARRAYBUFFER
-        return true;
+        return CONFIG_FLAG(WasmFastArray);
 #else
         return false;
 #endif
     }
 
-    ArrayBuffer * WebAssemblyArrayBuffer::TransferInternal(uint32 newBufferLength)
+    WebAssemblyArrayBuffer* WebAssemblyArrayBuffer::GrowMemory(uint32 newBufferLength)
```
MikeHolman: While we're at it, it's probably worth it to make this non-throwing. In case of failure return nullptr, and if we get nullptr then return -1.

lib/Runtime/Library/ArrayBuffer.cpp
```diff
@@ -998,38 +1017,70 @@ namespace Js
...
     {
-#if ENABLE_FAST_ARRAYBUFFER
-        ReportDifferentialAllocation(newBufferLength);
-        Assert(this->buffer);
+        if (newBufferLength <= this->bufferLength)
+        {
+            JavascriptError::ThrowTypeError(GetScriptContext(), WASMERR_BufferGrowOnly);
```
MikeHolman: Assert(UNREACHED)?

lib/Runtime/Library/ArrayBuffer.cpp
Cellule: I wanted to do this initially, but there are multiple places in there that can throw, so either I catch the exception in GrowMemory or in WebAssemblyMemory::GrowInternal

MikeHolman:
:shipit:

Issue #2733, opened 2017-03-23
ASAN: TypedArray transfer can leak memory


Penguinwizzard:
Running ASAN on the test/typedarray/transfer.js test reports a memory leak on application exit. The leaked memory is sized exactly the same as the TypedArray in the test. The affected test has been marked as exclude_sanitize_address for now in PR #2730 to allow ASAN CI to be enabled.

Penguinwizzard:
This issue still repros on 759c8ce0c10c13e070a033ddf7cd18a4b61395cb, which includes the linked change; the ASAN report on the failed test is as follows:

```
[1872/1900 0.68] Failed -> typedarray/transfer.js
/home/demorri/ChakraCore_VSO/out/Debug/ch -WERExceptionSupport -ExtendedErrorStackForTestHost -BaselineMode -maxInterpretCount:1 -maxSimpleJitRunCount:1 -bgjit- -dynamicprofilecache:profile.dpl.3373 -ArrayBufferTransfer /home/demorri/ChakraCore_VSO/test/typedarray/transfer.js

Output:
----------------------------
0
4194304
43
42
44
0
1048576
undefined
42
0
80
42
0
0
0
2097152
0
0

=================================================================
==17302==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 2097152 byte(s) in 1 object(s) allocated from:
    #0 0x7f9d79f367f8 in __interceptor_malloc (/home/demorri/ChakraCore_VSO/out/Debug/ch+0xc3e7f8)
    #1 0x7f9d79ff3567 in CorUnix::InternalMalloc(unsigned long) /home/demorri/ChakraCore_VSO/pal/src/cruntime/malloc.cpp:114:20
    #2 0x7f9d79ff3534 in PAL_malloc /home/demorri/ChakraCore_VSO/pal/src/cruntime/malloc.cpp:98:12
    #3 0x7f9d7b5b5d8f in Js::ArrayBuffer::ArrayBuffer<void* (*)(unsigned long)>(unsigned int, Js::DynamicType*, void* (*)(unsigned long)) /home/demorri/ChakraCore_VSO/lib/Runtime/Library/ArrayBuffer.cpp:599:33
    #4 0x7f9d7b5b079f in Js::JavascriptArrayBuffer::JavascriptArrayBuffer(unsigned int, Js::DynamicType*) /home/demorri/ChakraCore_VSO/lib/Runtime/Library/ArrayBuffer.cpp:666:9
[...]

PR #2993, opened 2017-05-18
reenable disabled typedarray test

leirocks:
resolving #2733

the memory leak should have been fixed with following change:
https://github.com/Microsoft/ChakraCore/pull/2939/files#diff-5b386c0529e526ce786f663dc7b2933fR387


akroshg:
:shipit:

PR #2996, opened 2017-05-19
Fixing the issue while copying properties to console scope

aneeshdk:
While copying the properties from empty scope to the console scope, the attributes of the properties were copied wrongly which was making the var declared properties not enumerables.


aneeshdk:
@akroshg @agarwal-sandeep Can you take a look?

agarwal-sandeep:
:shipit:

akroshg:
:shipit:

Issue #2606, opened 2017-02-27
Expression-body arrow function parameter destructuring "Use before declaration" error when parameter is only used in an inner function


erikdesjardins:
#2401 still happens for expression-body (but not block-body) arrow functions.

Reproduced with Chakra 1.4.0 and 1.4.1 (from the github releases).

```js
// works: block
(({ x }) => {
	(() => x)();
})({});
console.log('A done');

// works: block with return (behaviourally equivalent)
(({ x }) => {
	return (() => x)();
})({});
console.log('B done');

// fails: expression
(({ x }) =>
	(() => x)() // "Use before declaration"
)({});
console.log('C done');
```

jods4:
I stumbled upon this today with Edge 14.4393. 
It impacts Aurelia framework, see aurelia/bootstrapper#59.
I came up with the following minimal repro:
```js
function g(cb) { return g() }
let f = ({ x }) => g(() => x);
f({ x: 42 }); // should return 42 but throws "use before declaration".
```
As noted by OP the bug is quite specific. Not having nested lambdas, or using a block instead of an expression after the arrow makes it go away.

Code above works in Firefox.

RReverser:
Same issue here on latest Edge and ChakraCore 1.4.3, although I thought it's specific to Promises:

```javascript
let f = ({ data }) => Promise.resolve().then(() => {
  console.log(data);
});

f({
  data: [1,2,3]
}).catch(err => console.log(err.stack));
```

```
ReferenceError: Use before declaration
   at Anonymous function (d:\chakra-bug.js:2:3)
```

BenMcGarry:
Was #2998 the fix for this issue?

aneeshdk:
@BenMcGarry https://github.com/Microsoft/ChakraCore/pull/2998 should fix this issue. Does it still repro?

PR #2998, opened 2017-05-19
Fixing the closure capturing issue in expression arrow function body

aneeshdk:
While creating the parse node for the body of an arrow function we should be using the return statement to create the statement list. Right now we were using the param block scope.

### DIFF
--- a/test/es6/destructuring_params.js
+++ b/test/es6/destructuring_params.js
@@ -222,6 +222,10 @@ var tests = [
             assert.areEqual(k, 62,  "Identifiers from patterns are captured and initialized correctly");
         }
         f3({x:10}, [20], 30);
+
+        function f4(a) { return a() }
+        var f5 = ({ x }) => f4(() => x);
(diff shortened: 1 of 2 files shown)


aneeshdk:
@akroshg @pleath @tcare Can you please take a look?

aneeshdk:
https://github.com/Microsoft/ChakraCore/issues/2606

test/es6/destructuring_params.js
```diff
@@ -222,6 +222,10 @@ var tests = [
             assert.areEqual(k, 62,  "Identifiers from patterns are captured and initialized correctly");
         }
         f3({x:10}, [20], 30);
+
+        function f4(a) { return a() }
```
akroshg: add a test which has the eval in the lambda, which captures that var.

akroshg:
:shipit:

test/es6/destructuring_params.js
aneeshdk: Done

PR #3000, opened 2017-05-19
WASM: Use ArgSlot in signature

### LINKED ISSUE #2833: WASM - signature param fields type
Change WasmSignature::m_paramSize & m_paramsCount to ArgSlot and put necessary checks to make sure we don't overflow.

Right now, we use uint32 for the number of arguments in wasm, but at some places we cast this w/o checks to ArgSlot (because that's what the rest of the engine uses).

Cellule:
Closes #2833

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/3000)


bterlson:
Grats on 3k issues! :-D 🎆 🍾

PR #3002, opened 2017-05-20
Mark some tests as slow

Penguinwizzard:
Some of tests tests take a long time; we should run them as part of the slow run, not as part of the normal runs.

### DIFF
--- a/test/Object/rlexe.xml
+++ b/test/Object/rlexe.xml
@@ -171,7 +171,7 @@
   <test>
     <default>
       <files>Slow.js</files>
-      <tags>exclude_chk</tags>
+      <tags>exclude_chk,Slow</tags>
     </default>
   </test>
   <test>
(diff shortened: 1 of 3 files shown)


test/Object/rlexe.xml
```diff
@@ -171,7 +171,7 @@
   <test>
     <default>
       <files>Slow.js</files>
-      <tags>exclude_chk</tags>
+      <tags>exclude_chk,slow</tags>
```
dilijev: Not sure whether tags are case-sensitive or not, but we've always used uppercase `Slow` for slow tests.

Penguinwizzard: I've fixed the capitalization, and re-targeted to release/1.6.

Penguinwizzard:
@dotnet-bot test Ubuntu please

Penguinwizzard:
Rebasing and resubmitting since some of the CI got lost in the short downtime last week.

PR #3007, opened 2017-05-22
xplat perf improvements

obastemur:
#### osx: cache expensive threadId retrieval
`pthread_threadid_np(pthread_self(), &tid);` call is very expensive. Cache expensive threadId retrieval instead. Same commit also removes unused `THREADGetThreadProcessId`

#### xplat: Improve PAL wmemcmp perf.
Since we know the char count, comparing with larger blocks reduces the total cycles to spend.

### DIFF
--- a/pal/src/cruntime/wchar.cpp
+++ b/pal/src/cruntime/wchar.cpp
@@ -1230,7 +1230,7 @@ PAL_wmemcmp(
         const char16_t *string2,
         size_t count)
 {
-    size_t i;
+    size_t i, wi = 0;
     int diff = 0;
 
     PERF_ENTRY(wmemcmp);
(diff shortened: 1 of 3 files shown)


pal/src/cruntime/wchar.cpp
```diff
@@ -1241,14 +1241,21 @@ PAL_wmemcmp(
 
     if (string1 == string2) return diff;
 
-    for (i = 0; i < count; i++)
+    constexpr size_t blockSize = sizeof(size_t) / sizeof(char16_t);
+    const     size_t *num1     = (const size_t*)(string1);
+    const     size_t *num2     = (const size_t*)(string2);
+
+    while( (count >= blockSize * wi) && !(num1[wi] - num2[wi]) ) ++wi;
```
jianchun: >count >= blockSize * wi [](start = 12, length = 23)

`count >= blockSize * wi` can result in out-of-bound read at the end (when they equal).

obastemur: Yep, a silly edit issue on my end. That's why the test fails on CI. Thanks for pointing it out though.

obastemur:
retargeted to release/1.6

pal/src/cruntime/wchar.cpp
```diff
@@ -1241,14 +1241,21 @@ PAL_wmemcmp(
 
     if (string1 == string2) return diff;
 
-    for (i = 0; i < count; i++)
+    constexpr size_t blockSize = sizeof(size_t) / sizeof(char16_t);
+    const     size_t *num1     = (const size_t*)(string1);
+    const     size_t *num2     = (const size_t*)(string2);
+
+    while( (count > blockSize * wi) && !(num1[wi] - num2[wi]) ) ++wi;
```
jianchun: `!(num1[wi] - num2[wi])` ==> `num1[wi] == num2[wi]` would be easier to read?

obastemur:

PR #3020, opened 2017-05-22
Use the right script context while retrieving the debug document context

aneeshdk:
While retrieving the debug document context we should be using the script context of the function.


aneeshdk:

akroshg:
:shipit:

PR #3022, opened 2017-05-23
xplat: Improve spinLock and criticalSection implementations

obastemur:
Measured on amd64, new CCSpinLock works twice faster comparing to older
implementation and uses a lighter interface. Besides, previous
interface was interfering with blocking threads.


obastemur:
No CI ? /cc @dilijev

obastemur:
@digitalinfinity @jianchun ping

pal/inc/CCSpinLock.hpp
```diff
@@ -0,0 +1,34 @@
...
+
+#ifndef CC_PAL_INC_CCSPINLOCK_H
+#define CC_PAL_INC_CCSPINLOCK_H
+
+class CCSpinLock
+{
+    unsigned int  enterCount;
+    bool          shouldTrackThreadId;
+    char          lock;
+    size_t        threadId;
```
jianchun: >size_t [](start = 4, length = 6)

Why you use "size_t" instead of "DWORD" for threadId? GetCurrentThreadId() returns DWORD.

pal/src/thread/pal_thread.cpp
```diff
@@ -191,6 +191,7 @@ static void InternalEndCurrentThreadWrapper(void *arg)
 --*/
 BOOL TLSInitialize()
 {
+    free_threads_spinlock.Reset();
```
jianchun: >free_threads_spinlock.Reset(); [](start = 4, length = 30)

Why you need to call "Reset()" for this one? Doesn't the constructor work? Or this could be called multiple times? If the spinlock is already Entered by some thread, wouldn't this call corrupts the existing data?

pal/inc/CCSpinLock.hpp
```diff
@@ -0,0 +1,34 @@
...
+public:
+    void Reset(bool trackThread = false)
```
jianchun: >void Reset(bool trackThread = false) [](start = 4, length = 36)

Is this "Reset" API really needed? Seems it adds risk. Assume you need this API... do you need to allow changing "shouldTrackThreadId"? If not, seems there be an opportunity to make that a const field or a template argument. Those are 2 styles of SpinLock.

pal/src/sync/CCSpinLock.cpp
```diff
@@ -0,0 +1,114 @@
...
+        enterCount = 1;
+    }
+}
+
+bool CCSpinLock::TryEnter()
+{
+    int spin = 0;
+    bool locked = true;
+
+    while(__sync_lock_test_and_set(&lock, 1))
```
jianchun: >while(__sync_lock_test_and_set(&lock, 1)) [](start = 4, length = 41)

Seems you penalize "TryEnter" by not checking if the thread already "Enter"ed first... why?

pal/src/sync/CCSpinLock.cpp
```diff
@@ -0,0 +1,114 @@
...
+            enterCount++;
+            return;
+        }
+    }
+
+    while(__sync_lock_test_and_set(&lock, 1))
+    {
+        while(lock)
+        {
+            if (++spin > 11)
```
jianchun: >11 [](start = 25, length = 2)

Why magic number 11... could you add support for user determined spinCount?

pal/src/sync/CCSpinLock.cpp
```diff
@@ -0,0 +1,114 @@
...
+    }
+    __sync_lock_release(&lock);
+}
+
+bool CCSpinLock::IsLocked() const
+{
+    return threadId == GetCurrentThreadId();
+}
+
+bool CCSpinLock::IsLockedByAnyThread()
```
jianchun: Is this API needed / used anywhere? If not, consider delete it for less APIs.

lib/Common/Memory/Recycler.cpp
```diff
@@ -7586,7 +7585,6 @@ Recycler::InitializeProfileAllocTracker()
         trackerDictionary = NoCheckHeapNew(TypeInfotoTrackerItemMap, &NoCheckHeapAllocator::Instance, 163);
 
 #pragma prefast(suppress:6031, "InitializeCriticalSectionAndSpinCount always succeed since Vista. No need to check return value");
-        InitializeCriticalSectionAndSpinCount(&trackerCriticalSection, 1000);
```
jianchun: >1000 [](start = 71, length = 4)

Not sure if this "1000" has any impact. If you support spinCount, you could use it here and not changing any behavior.

lib/Common/Memory/ArenaAllocator.h
```diff
@@ -944,21 +944,19 @@ class ReferencedArenaAdapter : public RefCounted
...
-    {
-        InitializeCriticalSection(&adapterLock);
-    }
+    { }
 
     bool AddStrongReference()
     {
-        EnterCriticalSection(&adapterLock);
+        bool retval = false;
+        adapterLock.Enter();
```
jianchun: >adapterLock.Enter(); [](start = 8, length = 20)

Could use "AutoCriticalSection" to simplify some code like this.

pal/inc/CCSpinLock.hpp
obastemur: no particular reason. They both do the same on xplat yet agree DWORD should be the type there.

pal/src/thread/pal_thread.cpp
obastemur: This will be called once and can't be called during/after another lock in place due to `PAL` is not yet initialized. The underlying issue here is related to whole lib/pal initialization on start. I did some improvements there in the past, there will be more to come.

obastemur: Adding a comment.

pal/inc/CCSpinLock.hpp
obastemur: Moving `shouldTrackThreadId` to template argument is a good idea.

pal/src/sync/CCSpinLock.cpp
obastemur: Otherwise, ABA could happen. We need to make sure that we are able/unable to enter before checking the rest of the conditions.

pal/src/sync/CCSpinLock.cpp
obastemur: small enough to let cache is updated. At the moment, this is best for the performance under high load. Adding a todo note for better tuning / parameterizing

pal/src/sync/CCSpinLock.cpp
obastemur: Yep, I've been rebasing this implementation since 2.0 days. Safe to remove.

lib/Common/Memory/ArenaAllocator.h
obastemur: not sure how that applies to here?

pal/src/sync/CCSpinLock.cpp
jianchun: Why doesn't the same statement apply to "Enter"? If you can do that in "Enter", you should be able to do it in "TryEnter", no? Basically "Enter" does not spin if the thread already has the lock, but "TryEnter" still spins, why is the difference?

lib/Common/Memory/ArenaAllocator.h
jianchun: Your change is fine; I was just saying we could use "AutoCriticalSection a(&adapterLock)" for some code like this and the code would be simpler.

pal/src/sync/CCSpinLock.cpp
obastemur: Ohhh, see your point. The reason; `TryEnter` is more aggressive. Since we won't spin or sched_yield until we acquire the lock, we try a.s.a.p instead (very lightweight) and check other conditions after that.

obastemur:
@jianchun Thanks for the review! Anything else?

PR #3026, opened 2017-05-24
fix bug with inlining isArray where src isn't RegOpnd

MikeHolman:
In such a case, we need to load into a RegOpnd, but we were doing this in the wrong location.


rajatd:
:shipit:

PR #3028, opened 2017-05-24
fix assert in ObjTypeSpecFldInfo::GetFixedFieldIfAvailableAsFixedFunction

MikeHolman:
Assert isn't valid, as there isn't always a guarantee about equivalence. Instead, we should do a check in the API.

### DIFF
--- a/lib/Backend/Inline.cpp
+++ b/lib/Backend/Inline.cpp
@@ -2575,6 +2575,11 @@ IR::Instr * Inline::InlineApplyWithoutArrayArgument(IR::Instr *callInstr, const 
         return callInstr;
     }
 
+    if (!applyTargetInfo)
+    {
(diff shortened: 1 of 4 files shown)


rajatd:
Do you have a test case for this?

MikeHolman:
Sure, I've added one.

rajatd:
1. We dont have polymorphic inlining for apply targets.
2. We collect fixed field info at workitem creation for polymorphic inlining only if the method properties are at equivalent slots in their owner object types.

By (2), the assert makes sense. We should, however, not call TryOptimizeCallInstrWithFixedMethod from InlineApplyWithoutArrayArgument if applyTargetInfo is null.

MikeHolman:
Makes sense. I'll make that change.

PR #3044, opened 2017-05-24
Enable VTune JIT Profiling in Linux

litian2025:
Enable VTune JIT Profiling in Linux. Added VTuneChakraProfile::Register(); along with EtwTrace::Register(), also added undef and def for getenv in jitprofiling.cpp on Linux to resolve the conflict introduced by pal.h definition for getenv.

### DIFF
--- a/lib/Jsrt/JsrtHelper.cpp
+++ b/lib/Jsrt/JsrtHelper.cpp
@@ -12,6 +12,8 @@
 #include "jsrtHelper.h"
 #include "Base/ThreadContextTlsEntry.h"
 
+#include "Base/VTuneChakraProfile.h"
+
 #ifdef DYNAMIC_PROFILE_STORAGE
 #include "Language/DynamicProfileStorage.h"
 #endif
@@ -143,6 +145,9 @@ void JsrtCallbackState::ObjectBeforeCallectCallbackWrapper(JsObjectBeforeCollect
 
     #ifdef ENABLE_JS_ETW
         EtwTrace::Register();
+    #endif
+    #ifdef VTUNE_PROFILING
+        VTuneChakraProfile::Register();
     #endif
         ValueType::Initialize();
         ThreadContext::GlobalInitialize();
--- a/pal/src/cruntime/misc.cpp
+++ b/pal/src/cruntime/misc.cpp
@@ -378,6 +378,10 @@ char *MiscGetenv(const char *name)
     }
 done:
     InternalLeaveCriticalSection(pthrCurrent, &gcsEnvironment);
+    if (pRet == NULL)
+    {
+        return getenv(name);
+    }
     return pRet;
 }
 


lib/Runtime/Base/jitprofiling.cpp
```diff
@@ -6,6 +6,12 @@
 
 #ifdef VTUNE_PROFILING
 
+// For Linux getenv is been defined as PAL_getenv in pal/inc/pal.h
+// We need undef here so that VTune can use it to get environment variable 
+#ifdef PLATFORM_UNIX
+#undef getenv
```
obastemur: How exactly `getenv` is used for posix/unix and why PAL_getenv fails?

litian2025: getenv is used at: https://github.com/Microsoft/ChakraCore/blob/master/lib/Runtime/Base/jitprofiling.cpp#L199_L201
VTune needs to retrieve environment variable via getenv in unix so that VTune JIT profiling can work, on Windows the environement variable is retrieved via GetEnvironmentVariableA, thus no problem on Windows. 
PAL_getenv fails because this file is used by both Windows and Linux, at the top of the file it needs include "RuntimeBasePch.h" so that it can be built correctly on Windows and Unix: https://github.com/Microsoft/ChakraCore/blob/master/lib/Runtime/Base/jitprofiling.cpp#L5, In Unix, "RuntimeBasePch.h" pulls in the include file of pal.h (pal/inc/pal.h), in that file pal.h: getenv is defined in PAL_getenv at https://github.com/Microsoft/ChakraCore/blob/master/pal/inc/pal.h#L6051, however PAL_getenv is empty when this jitprofiling.cpp is been built on Linux, thus causing VTune JIT Profiling failed. 
Another way to do it is to build pal cruntime earlier than building this component, since the implementation of PAL_getenv is at https://github.com/Microsoft/ChakraCore/blob/master/pal/src/cruntime/misc.cpp#L171_#L196. However, if we take this approach, not only we need make changes in the build procedure, we also need to make change in this jitprofiling.cpp file, change the original getenv to PAL_getenv.  For the purpose of easy maintenance in the future, I don't think it's a good idea to deviate too much from the original VTune file, especially making changes in the middle of the file. 
So compare to the 2nd approach, current approach is cleaner, less code change and less distruptive.

obastemur: Well, instead of disabling `PAL_getenv` here, we may fix it.

We could replace;
https://github.com/Microsoft/ChakraCore/blob/master/pal/src/cruntime/misc.cpp#L381

to

```
if (pRet == NULL)
    {
        return getenv(name);
    }
    return pRet;
```

litian2025: I like this idea since it fixed the bug in misc.cpp. I will make the changes, run tests and update the PR.

litian2025:
@obastemur Please let me know if there's anything else needs to be done, tested it builds fine on Ubuntu system, the build failure https://ci.dot.net/job/Microsoft_ChakraCore/job/master/job/shared_ubuntu_linux_debug_prtest/566/consoleFull#-116972114519119f5c-76e5-4da8-b294-eb427d612413 is not related to the change.

obastemur:

lib/Runtime/Base/jitprofiling.cpp
```diff
@@ -196,9 +196,9 @@ static int loadiJIT_Funcs()
         }
     }
 #else  /* ITT_PLATFORM==ITT_PLATFORM_WIN */
-    dllName = getenv(NEW_DLL_ENVIRONMENT_VAR);
+    dllName = PAL_getenv(NEW_DLL_ENVIRONMENT_VAR);
```
obastemur: Any reason to update this? (and the one below)

obastemur:
@litian2025 Thanks for the contribution. Changes should work without updating `getenv` calls to `PAL_getenv` since it will eventually call PAL_getenv on xplat.

Once you update that, I will merge this in.

obastemur:
@litian2025 Could you please squash your changes into a single commit?

litian2025:
@obastemur Sure, I will squash it into 1 commit and update

obastemur:
LGTM

litian2025:
@obastemur Could you please merge it? Please let me know if there's anything I need to do, I am not sure that I need to update the branch to pull in the latest changes for the merge or not needed. Thank you.

PR #3045, opened 2017-05-24
Remove obsolete ES6 flag tests

### LINKED ISSUE #58: Remove ES6 flag baselines
We probably don't need es6\es6_stable.js and friends anymore - can we remove them?

akroshg  commented on Nov 30 2015
Agreed.
Should we also remove all stable flags from the .xml files?

tcare  commented on Nov 30 2015
We could, but if we turn a feature off by default (e.g. due to spec change) then we would have to add them back again for that test. That has been fairly rare, so I'm ok with removing the flags.

ianwjhalliday  commented on Nov 30 2015 
There is still value in the idea of validating the flags so that when a new one is added we have to pay attention to its default to prevent accidentally introducing features too early.  However, as we saw previously with the async feature snafu earlier, the current test does not make it obvious enough for us to catch such a mistake.

I recommend that if we spend time here changing things we should take a stab at improving the test itself after reducing it down to all that is needed.

Specific improvements I'd like to see are:
- don't use `-verbose` to print out the flag information; use something like `-DumpEsFlags` instead
  - currently this reliance on `-verbose` pollutes other uses of `-verbose`, like when looking at verbose bailout trace and dump
- get rid of `-es6` and `-es6all`; instead introduce `-esall` to turn on off-by-default language features
  - perhaps also a `-esexperimental` to turn on all the experimental features
- beautify the flag dump information into a table showing the feature flag and its default setting (`off`, `experimental`, `on`)

I don't think we care about testing the semantics for specifying the switches multiple times (e.g. the `-es6 -es6-` test), nor do we care about seeing the current state (as opposed to the default) of each switch.

tcare:
Fixes #58


tcare:
@dilijev

tcare:

akroshg:
LGTM
curious what is the motivation behind this change?

tcare:
These were originally added to test the introduction of disabling ES6 flags selectively. It's long been stable and they are a minor pain to update when adding a flag so I don't think we need them anymore

dilijev:
Basically "low-value, high maintenance cost" -- something we're trying to fix across our tests.

PR #3048, opened 2017-05-25
fix issue with rejit exception and no profile info

MikeHolman:
In some cases it's possible to have a rejit exception which wants to disable some optimization even if you don't have profile info. For this to work it required refactoring a bit in how we check if an optimization is disabled, and while I was here I decided to do a bit of cleanup around the RejitReason enum as well.

### DIFF
--- a/lib/Backend/Func.cpp
+++ b/lib/Backend/Func.cpp
@@ -299,42 +299,35 @@ Func::Codegen(JitArenaAllocator *alloc, JITTimeWorkItem * workItem,
         catch (Js::RejitException ex)
         {
             // The work item needs to be rejitted, likely due to some optimization that was too aggressive
-            if (ex.Reason() == RejitReason::AggressiveIntTypeSpecDisabled)
+            switch (ex.Reason())
             {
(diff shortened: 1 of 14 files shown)


lib/Backend/Func.cpp
```diff
@@ -299,42 +299,35 @@ Func::Codegen(JitArenaAllocator *alloc, JITTimeWorkItem * workItem,
...
-                workItem->GetJITFunctionBody()->GetProfileInfo()->DisableArrayCheckHoist(func.IsLoopBody());
+                break;
+            case RejitReason::ArrayCheckHoistDisabled:
+            case RejitReason::ArrayAccessHelperCallEliminationDisabled:
                 outputData->disableArrayCheckHoist = TRUE;
-            }
-            else
-            {
-                Assert(ex.Reason() == RejitReason::TrackIntOverflowDisabled);
-                workItem->GetJITFunctionBody()->GetProfileInfo()->DisableTrackCompoundedIntOverflow();
```
Penguinwizzard: It looks like you've removed the last use of a few of these functions; should those functions themselves be removed as part of this PR?

MikeHolman:
@dotnet-bot test this

PR #3056, opened 2017-05-25
implement $262 for Atomics test262 tests

leirocks:
also exposes SharedArrayBuffer APIs for JSRT
and couple fixes for Atomics dealing with the index input argument
all Atomics test262 tests passed


test/$262/$262test.js
```diff
@@ -0,0 +1,147 @@
...
+  } else {
+    print('test262/error Test262Error: ' + err);
+  }
+}
+
+function $DONE(err) {
+  if (err) {
+    $ERROR(err);
+  }
+  print('test262/done');
```
akroshg: just print "pass", so that you don't have to define the baseline file.

lib/Runtime/Library/AtomicsObject.cpp
```diff
@@ -53,23 +53,27 @@ namespace Js
 
     uint32 AtomicsObject::ValidateAtomicAccess(Var typedArray, Var requestIndex, ScriptContext *scriptContext)
     {
+        Assert(TypedArrayBase::Is(typedArray));
+
         int32 accessIndex = -1;
         if (TaggedInt::Is(requestIndex))
```
akroshg: we should make use of ArrayBuffer::ToIndex. That function should do the exactly the same.

lib/Runtime/Library/AtomicsObject.cpp
```diff
@@ -217,7 +221,10 @@ namespace Js
             DWORD_PTR agent = (DWORD_PTR)scriptContext;
             Assert(sharedArrayBuffer->GetSharedContents()->IsValidAgent(agent));
             awoken = waiterList->AddAndSuspendWaiter(agent, timeout);
-            waiterList->RemoveWaiter(agent);
+            if (!awoken) 
+            {
+                waiterList->RemoveWaiter(agent);
```
akroshg: Did you have test to hit scenario?

bin/ch/WScriptJsrt.cpp
```diff
@@ -511,6 +554,114 @@ JsValueRef WScriptJsrt::LoadScript(JsValueRef callee, LPCSTR fileName,
     return value;
 }
 
+ThreadData::ThreadData()
```
curtisman: Separate .h/.cpp file for ThreadData?

leirocks: sure. for naming this is intended to generally support 'crossthread', not only for $262.agent. I'm going with RuntimeThreadData

bin/ch/WScriptJsrt.cpp
```diff
@@ -34,6 +34,25 @@
...
+{
+    ThreadLocalData()
+        :threadData(nullptr)
+    {
+    }
+    ~ThreadLocalData()
+    {
+        if (threadData)
+        {
+            delete threadData;
```
obastemur: could be nice to set threadData back to nullptr after deleting.

bin/ch/WScriptJsrt.cpp
```diff
@@ -473,6 +493,29 @@ JsValueRef WScriptJsrt::LoadScript(JsValueRef callee, LPCSTR fileName,
...
     }
+    else if (strcmp(scriptInjectType, "crossthread") == 0)
+    {
+        if (threadLocalData.threadData == nullptr) 
+        {
+            threadLocalData.threadData = new ThreadData();
+        }
+
+        ThreadData* child = new ThreadData();
+        child->initialSource = fileContent;
```
obastemur: Curious; if life time of `fileContent` could be a problem?

leirocks: initialSource is a std::string, the content is copied

bin/ch/$262.js
```diff
@@ -0,0 +1,24 @@
+R"(
```
obastemur: Do you need this to be included as `c char array` ?

bin/ch/WScriptJsrt.cpp
leirocks: sure, will do.

btw, just notice semaphore is not implemented in our PAL, is it hard to wrap linux semaphore to be like the Win32 one?

bin/ch/$262.js
leirocks: yes, this is raw string in c++1x. not sure if the VS2012 support this. if not might need change this back to a messy byte array or load script file in running time

bin/ch/WScriptJsrt.cpp
obastemur: Please add an `xplat-todo: semaphore could be nice here` note to the places. I'm updating xplat locking backend hence I can visit those comments later

bin/ch/$262.js
obastemur: see https://github.com/Microsoft/ChakraCore/blob/master/build.sh#L536 

I had wrote a small python tool to convert string to `c char array` on pre build step. We can use the same approach here.

leirocks: actually, I just checked we are supporting VS2013 and VS2015, both support c++11 raw string.
so no pre-build is necessary.
see https://msdn.microsoft.com/en-us/library/69ze775t(v=vs.120).aspx

leirocks:
@liminzhu @bterlson can you review the new JSRT API for SharedArrayBuffer?

bin/ch/$262.js
obastemur: >  I just checked we are supporting VS2013 and VS2015, both support c++11 raw string.
so no pre-build is necessary.

How this will work for xplat?

lib/Jsrt/ChakraCommon.h
```diff
@@ -350,6 +350,15 @@ typedef unsigned short uint16_t;
     typedef void *JsRef;
 
     /// <summary>
+    ///     A reference to an object owned by the SharedArrayBuffer.
```
leirocks: moved. I'm creating PR base on 1.6 branch. will be updated there

lib/Jsrt/ChakraCommon.h
```diff
@@ -2063,6 +2072,74 @@ typedef unsigned short uint16_t;
...
+    ///     Requires an active script context.
+    /// </remarks>
+    /// <param name="sharedContents">
+    ///     The storage object of a SharedArrayBuffer which can be shared between multiple thread.
+    /// </param>
+    /// <param name="result">The new SharedArrayBuffer object.</param>
+    /// <returns>
+    ///     The code <c>JsNoError</c> if the operation succeeded, a failure code otherwise.
+    /// </returns>
+    CHAKRA_API
```
leirocks: >>Just to confirm, one can only create sab from sharedContents returned from a JsGetSharedArrayBufferContent call somewhere, correct?

currently yes, in order to run the test262 tests. we can add API to create SAB if user specify a length and chakra core manage the buffer. if user want to create SAB with arbitrary buffer that will be complicated, the free function to free that buffer will be needed

>>Is the output JsValueRef *result shareable between threads

only JsSharedArrayBufferContent can be shared between threads, and require JsSharedContentAddRef/JsSharedContentRelease call to manage the lifetime. the JsValueRef output of JsCreateSharedArrayBuffer is same as other javascript object which should only be used in current thread

>> A JsCreateExternalSharedArrayBuffer can also be tricky since I don't know if native access to the buffer can break atomics

I don't think that will break atomics as long as the host guarantee that it does not mutate the buffer, this is mostly about memory management, we need a way to notify the host that chakracore has done using the buffer (aka. refcount get to 0). we can discuss about this if you have a use case. for this PR it's not necessary for running test262

I'll change JsCreateSharedArrayBuffer to JsCreateSharedArrayBufferWithSharedContent or you can suggest a better name?

leirocks:
replaced with #3137

PR #3058, opened 2017-05-26
Add trampoline for JIT calls

MikeHolman:
This adds additional mitigation for JIT function calls. With this trampoline, we can range check and safely call to any 16 byte aligned address inside the block. If range check fails (e.g. function we are calling has not been JITed), we do a CFG check. Benchmarks are mostly flat, but there is about a 2.5% loss in Gameboy.

### DIFF
--- a/lib/Backend/JITThunkEmitter.cpp
+++ b/lib/Backend/JITThunkEmitter.cpp
@@ -1,0 +1,285 @@
+//-------------------------------------------------------------------------------------------------------
+// Copyright (C) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE.txt file in the project root for full license information.
+//-------------------------------------------------------------------------------------------------------
+
+#include "Backend.h"
+
+#if defined(ENABLE_NATIVE_CODEGEN) && defined(_CONTROL_FLOW_GUARD) && (_M_IX86 || _M_X64)
+
+template class JITThunkEmitter<VirtualAllocWrapper>;
+
+#if ENABLE_OOP_NATIVE_CODEGEN
+template class JITThunkEmitter<SectionAllocWrapper>;
+#endif
+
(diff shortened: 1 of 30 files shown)


MikeHolman:
@dotnet-bot test Ubuntu shared_ubuntu_linux_release

lib/Backend/JITThunkEmitter.cpp
```diff
@@ -0,0 +1,285 @@
...
+    return pageBV->IsAllSet();
+}
+
+template <> inline
+void
+JITThunkEmitter<VirtualAllocWrapper>::ProtectPage(void * address)
+{
+    DWORD oldProtect;
+    VirtualProtectEx(this->processHandle, address, ThunkSize, PAGE_EXECUTE, &oldProtect);
+    AssertOrFailFast(oldProtect == PAGE_EXECUTE_READWRITE);
```
leirocks: >oldProtect == PAGE_EXECUTE_READWRITE [](start = 21, length = 36)

check return value as well, or initialize oldProtect, in case VirtualProtectEx failed

lib/Common/DataStructures/FixedBitVector.h
```diff
@@ -681,6 +681,19 @@ class BVStatic
         return true;
     }
 
+    bool IsAllSet() const
+    {
+        for (BVIndex i = 0; i < wordCount; i++)
+        {
+            if (!this->data[i].IsFull())
```
leirocks: >IsFull() [](start = 31, length = 8)

is this correct? what if there's unused bits at the end? IsAllClear is OK because I think we zeroed while initializing.

I think we need to check with rangeSize for last word, in case rangeSize%wordSize !=0

leirocks:
:shipit:

MikeHolman:
@dotnet-bot test this

PR #3062, opened 2017-05-30
xplat: cache threadId

obastemur:
GetCurrentThreadId is one of the hot functions and on linux it makes a syscall.


digitalinfinity:
There's no scenario where the thread id can change out from underneath us, yes? This is a good change though- LGTM

obastemur:
@digitalinfinity Yes. Thanks for the review!

PR #3068, opened 2017-05-31
Handle thumb bit for CFG (thanks MicHol) and 8-byte align arm64 interpreter thunks.

Penguinwizzard:
Also fix a couple of other arm-related issues.


lib/Backend/NativeCodeGenerator.cpp
```diff
@@ -1117,7 +1117,7 @@ NativeCodeGenerator::CodeGen(PageAllocator * pageAllocator, CodeGenWorkItem* wor
 
     if (!CONFIG_FLAG(OOPCFGRegistration))
     {
-        scriptContext->GetThreadContext()->SetValidCallTargetForCFG((PVOID)jitWriteData.codeAddress);
```
MikeHolman: I think this is only a partial fix. There is a different place to do this if config flag is set. And presumably the unregister call also needs it.

PR #3070, opened 2017-05-31
xplat: fix llvm ld compatibility problem

### LINKED ISSUE #2891: Can't compile with ld.lld
Hi,
LLVM linker doesn't support "--no-keep-memory" option, so the build fails.

obastemur:
Fixes #2891


obastemur:
@digitalinfinity @kfarnung Thanks for the review

PR #3072, opened 2017-05-31
Update return symbol assignment in the presence of early return in try finally

meg-gupta:
Currently we skip generating a Ld_A after LdElem for return symbol s0 to save a temporary register.
In the presence of try finally with an early return, we have to run the finally code after thfunction return.
This can lead to problems later on in globopt, this change fixes this.

### DIFF
--- a/lib/Backend/GlobOpt.cpp
+++ b/lib/Backend/GlobOpt.cpp
@@ -17467,6 +17467,7 @@ GlobOpt::PreOptPeep(IR::Instr *instr)
             IR::Opnd *oldDst = instr->UnlinkDst();
             instr->SetDst(newDst);
             IR::Instr *ld = IR::Instr::New(Js::OpCode::Ld_A, oldDst, newDst, this->func);
+            oldDst->SetIsJITOptimizedReg(true);
             instr->InsertAfter(ld);
         }
     }
--- a/lib/Backend/GlobOpt.cpp
+++ b/lib/Backend/GlobOpt.cpp
@@ -17460,7 +17460,7 @@ GlobOpt::PreOptPeep(IR::Instr *instr)
         // With try finallys, these blocks are not moved out due to the early return -> finally and finally -> early return edges
         //
         // This happens only with return s0 symbol, because we avoid creating new temporary destination register for the s0 case
-        if (dst && dst->GetSym() && dst->GetSym()->m_id == 0 && !dst->AsRegOpnd()->IsArrayRegOpnd() && instr->m_opcode != Js::OpCode::Ld_A)
+        if (dst && dst->GetSym() && dst->GetSym()->m_id == 0 && instr->m_opcode != Js::OpCode::Ld_A)
         {
             IR::RegOpnd *newDst = dst->AsRegOpnd()->Copy(this->func)->AsRegOpnd();
             newDst->m_sym = StackSym::New(dst->GetType(), this->func);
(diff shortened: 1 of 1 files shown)


lib/Backend/GlobOpt.cpp
```diff
@@ -17435,7 +17435,18 @@ GlobOpt::PreOptPeep(IR::Instr *instr)
         }
         RemoveCodeAfterNoFallthroughInstr(instr);
     }
-
+    if (this->func->HasFinally())
+    {
+        IR::Opnd * dst = instr->GetDst();
+        if (dst && dst->GetSym() && dst->GetSym()->m_id == 0 && instr->m_opcode == Js::OpCode::LdElemI_A)
```
LouisLaf: >Js::OpCode::LdElemI_A [](start = 83, length = 21)

Why only LdElem?  Isn't this a problem for all opcodes?

lib/Backend/GlobOpt.cpp
```diff
@@ -17435,7 +17435,18 @@ GlobOpt::PreOptPeep(IR::Instr *instr)
         }
         RemoveCodeAfterNoFallthroughInstr(instr);
     }
-
+    if (this->func->HasFinally())
+    {
```
LouisLaf: Add a good comment here.  This is a weird hack unfortunately, and we need to document it well.

meg-gupta: Added comment.

lib/Backend/GlobOpt.cpp
meg-gupta: Right, it can potentially happen with any opcode with a preop bailout. Corrected this.

lib/Backend/GlobOpt.cpp
```diff
@@ -17441,7 +17441,36 @@ GlobOpt::PreOptPeep(IR::Instr *instr)
...
+        // If this instruction was in a loop, gets added to liveOnBackEdge syms
+        //
+        // This causes asserts in register allocator, when it extends the lifetime of the type specialized symbol to the entire loop
+        //
+        // This is a problem only with try-finallys
+        // For non-EH functions, the early return block gets moved out of the loop due to break block removal
+        // With try finallys, these blocks are not moved out due to the early return -> finally and finally -> early return edges
+        //
+        // This happens only with return s0 symbol, because we avoid creating new temporary destination register for the s0 case
+        if (dst && dst->GetSym() && dst->GetSym()->m_id == 0 && !dst->AsRegOpnd()->IsArrayRegOpnd() && instr->m_opcode != Js::OpCode::Ld_A)
```
LouisLaf: >IsArrayRegOpnd [](start = 83, length = 14)

It seems like the bug you had with IsArrayRegOpnd was because Copy() doesn't handle these operands...  Handling that would be a better fix.

LouisLaf: It actually does seem to handle these...  So why is this needed?


---
In reply to: [120736073](https://github.com/Microsoft/ChakraCore/pull/3072#discussion_r120736073) [](ancestors = 120736073)

meg-gupta: Corrected it, that remained as a residue from the time I did RegOpnd::New instead of Copy and replace m_sym.

meg-gupta:
Closed this because this fix is insufficient. If we bailout on the instruction whose dst we changed from s0, in the interpreter path we will not have the value for s0's bytecode reg, because of the dst replacement. The correct fix is here - https://github.com/Microsoft/ChakraCore/pull/3460

PR #3075, opened 2017-05-31
WASM: Prevent folding cmp instr

Cellule:
If we run out of registers during regalloc we'll try to fold the instruction and sometimes swap the operands.
In the case of `cmp` we swap the operands if we're able to reverse the logic of the next branch.

However, it is possible that another instr will want to use the flags set by `cmp` other than the branch.

The exact case I am seeing is this on x86
```asm
; Globopt
s1.i64 = Conv_Prim -2.u32
... ; some other operations
s2.i64 = Ld_I4 0xC5C6358EC332B00A.i64
s3.i32 = CmUnGe_I4 s1.i64, s2.i64
...

; after regalloc
s4.i32 = MOV 0
s4<-10>.i32 = MOV s4.i32 //spill
s5.i32 = MOV -2
... ; some other operations
s6.i32 = MOV 0xC5C6358E.i32
s7.i32 = MOV 0xC332B00A.i32

s10.i32 = MOV 0 ; comparison result
                ; Compare high bits of the i64
                CMP s6.i32, s4<-10> ; BUG: opnd swapped to avoid read spilled reg
                JNE $L1
                CMP s5.i32, s7.i32 ; compare low bits of the i64
$L1:
s10.i32 = SETAE s10.i32 ; Use the flags set by cmp to the comparison result
```

The _clean_ solution would be to track which instr uses/set the register flags to know if we're allowed to do such an optimization. However, this seems to be a lot of overhead so I opted for the quick fix.

I was also thinking of using the unused `dst` of the `cmp` instr to track which instr uses the flags it sets.
What do you guys think ?

I wasn't able to write a small enough test, I couldn't find a way to use all the registers and force trying to fold.

This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/3075)

### DIFF
--- a/lib/Backend/IR.h
+++ b/lib/Backend/IR.h
@@ -751,7 +751,7 @@ public:
     static BranchInstr * New(Js::OpCode opcode, Opnd* destOpnd, LabelInstr * branchTarget, Opnd *srcOpnd, Func *func);
     static BranchInstr * New(Js::OpCode opcode, LabelInstr * branchTarget, Opnd *src1Opnd, Opnd *src2Opnd, Func *func);
 
-    BranchInstr(bool hasBailOutInfo = false) : Instr(hasBailOutInfo), m_branchTarget(nullptr), m_isAirlock(false), m_isSwitchBr(false), m_isOrphanedLeave(false)
+    BranchInstr(bool hasBailOutInfo = false) : Instr(hasBailOutInfo), m_branchTarget(nullptr), m_isAirlock(false), m_isSwitchBr(false), m_isOrphanedLeave(false), m_areCmpRegisterFlagsUsedLater(false)
     {
 #if DBG
         m_isMultiBranch = false;
--- a/lib/Backend/IR.h
+++ b/lib/Backend/IR.h
@@ -739,6 +739,7 @@ public:
     bool                 m_isAirlock : 1;
     bool                 m_isSwitchBr : 1;
     bool                 m_isOrphanedLeave : 1; // A Leave in a loop body in a try, most likely generated because of a return statement.
+    bool                 m_areCmpRegisterFlagsUsedLater : 1; // Indicate that this branch is not the only instr using the register flags set by cmp
 #if DBG
     bool                 m_isMultiBranch;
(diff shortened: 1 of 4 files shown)


Cellule:

Cellule:
ping, this issue is hitting a lot

lib/Backend/IR.h
```diff
@@ -739,6 +739,7 @@ class BranchInstr : public Instr
     bool                 m_isAirlock : 1;
     bool                 m_isSwitchBr : 1;
     bool                 m_isOrphanedLeave : 1; // A Leave in a loop body in a try, most likely generated because of a return statement.
+    bool                 m_areCmpRegisterFlagsUsedLater : 1; // Indicate that this branch is not the only instr using the register flags set by cmp
```
LouisLaf: >m_areCmpRegisterFlagsUsedLater [](start = 25, length = 30)

IR::Instrs are not zero allocated.  You need to initialize the flag in the ctor.

lib/Backend/amd64/EncoderMD.cpp
```diff
@@ -1891,7 +1891,7 @@ bool EncoderMD::TryFold(IR::Instr *instr, IR::RegOpnd *regOpnd)
         {
             IR::Instr *instrNext = instr->GetNextRealInstrOrLabel();
 
-            if (instrNext->IsBranchInstr() && instrNext->AsBranchInstr()->IsConditional())
+            if (instrNext->IsBranchInstr() && instrNext->AsBranchInstr()->IsConditional() && instrNext->AsBranchInstr()->m_areCmpRegisterFlagsUsedLater)
```
LouisLaf: >instrNext->AsBranchInstr()->m_areCmpRegisterFlagsUsedLater [](start = 93, length = 58)

You forgot the ! here...

Cellule: Fixed, thanks!

lib/Backend/IR.h
Cellule: Fixed, thanks

LouisLaf:
:shipit:

Issue #3049, opened 2017-05-25
Incorrect cast and assert when calling eval in a with statement


natashenka:
When eval references this, OP_GetThisScoped casts each link in the scope chain to a DynamicObject to look for the correct this object. However, if eval is called inside a with statement, the scope chain could contain a WithObject, which is not a Dynamic object. This causes an assert to fire, and could cause incorrect functionality.

To reproduce the issue, use the following script:

var arrowInsideWithEval;

with ({}) {
  arrowInsideWithEval = (s) => eval(s);
  arrowInsideWithEval("this");
}

natashenka:
The assert is:

ASSERTION 21118: (/home/user/test_everywhere/ChakraCore/lib/Runtime/Types/DynamicObject.cpp, line 110) Ensure instance is actually a DynamicObject
 Failure: (obj->DbgIsDynamicObject())
Illegal instruction (core dumped)

I'm running:

ChakraCore/out/Debug/ch with.js to get this result

PR #3078, opened 2017-05-31
Avoid casting with scope object to Dynamic Object

aneeshdk:
We were wrongly casting with scope object to a Dynamic Object while trying to access this.

### DIFF
--- a/lib/Runtime/Language/JavascriptOperators.cpp
+++ b/lib/Runtime/Language/JavascriptOperators.cpp
@@ -1913,7 +1913,7 @@ CommonNumber:
         for (int i = 0; i < length; i += 1)
         {
             Var value;
-            DynamicObject *obj = DynamicObject::FromVar(pScope->GetItem(i));
+            RecyclableObject *obj = RecyclableObject::FromVar(pScope->GetItem(i));
             if (JavascriptOperators::GetProperty(obj, Js::PropertyIds::_lexicalThisSlotSymbol, &value, scriptContext))
             {
                 return value;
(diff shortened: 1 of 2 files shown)


aneeshdk:
@curtisman @pleath @akroshg Can you please take a look?

aneeshdk:
https://github.com/Microsoft/ChakraCore/issues/3049

lib/Runtime/Language/JavascriptOperators.cpp
```diff
@@ -1913,7 +1913,7 @@ namespace Js
         for (int i = 0; i < length; i += 1)
         {
             Var value;
-            DynamicObject *obj = DynamicObject::FromVar(pScope->GetItem(i));
+            RecyclableObject *obj = RecyclableObject::FromVar(pScope->GetItem(i));
```
agarwal-sandeep: Can there be a scenario here where framedisplay contains inline slot array instead of RecyclableObject?

pleath:
ActivationObject is a DynamicObject. What kind of object were we casting?

lib/Runtime/Language/JavascriptOperators.cpp
pleath: Presumably this is only called from within eval, and so all the contents of the FrameDisplay are objects.

aneeshdk:
It was a WithScopeObject which is not a Dynamic Object

pleath:
Okay. Thanks.

lib/Runtime/Language/JavascriptOperators.cpp
aneeshdk: No this is specific to eval case

PR #3081, opened 2017-05-31
UTF8 decode index signedness

Cellule:
Do not convert index from unsigned to signed when null termination in DecodeUnitsIntoAndNullTerminate

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/3081)


dilijev:
LGTM

PR #2757, opened 2017-03-29
Check return value for GetModuleFileName in ConfigParser::ParseConfig

xiaoyinl:
If `GetModuleFileName` failed and we ignore the return value, then `_wsplitpath_s` will read uninitialized stack buffer `modulename`.

Issue #2297, opened 2016-12-29
PAL: Cleanup GetModuleFileName** and implement corresponding interface


obastemur:
This is a follow-up issue for #2296 

```
+// xplat-todo: Implement a corresponding solution for GetModuleFileNameW
+// and cleanup PAL. [ https://github.com/Microsoft/ChakraCore/pull/2288 should be merged first ]
+// GetModuleFileName* PAL is not reliable and forces us to explicitly double initialize PAL
+// with argc / argv....
```

~~Waiting for #2288~~

Move `GetBinaryLocation` (https://github.com/Microsoft/ChakraCore/blob/7541e228ef42a97895a88f46d5427d07bf2675ba/bin/ch/Helpers.cpp#L577) into PlatformAgnostic and use that instead of `GetModuleFileName*`

edit-2, Some more details:

- Create `SystemInfo.cpp` under `lib/Runtime/PlatformAgnostic/Platform/Common/`

```
#include "Common.h"
#include "ChakraPlatform.h"

namespace PlatformAgnostic
{
    void SystemInfo::GetBinaryLocation(........) {   .......  }
    ....
```

- After moving GetBinaryLocation under the new SystemInfo.cpp, we should add it into `lib/Runtime/PlatformAgnostic/CMakeLists.txt` (i.e. at line 4)

- Similarly add that source file into `lib/Runtime/PlatformAgnostic/Chakra.Runtime.PlatformAgnostic.vcxproj` (i.e. line 41)

- Finally add SystemInfo's GetBinaryLocation static method definition into `lib/Common/PlatformAgnostic/SystemInfo.h`

dilijev:
@obastemur Any update on this?

obastemur:
@dilijev this is a nice to have hence not rushing on. Also added `accepting PRs` label. If anyone is interested implementing this, highly appreciated.

obastemur:
#3112 takes care of moving GetBinaryLocation into PlatformAgnostic. 

Remaining task: remove GetModuleFileName** interface and replace the usages to GetBinaryLocation.

PR #3082, opened 2017-05-31
Check return value for GetModuleFileName

xiaoyinl:
This is a recreated pull request of PR #2757, as I messed up the commits there.

This patch checks the return values of calls to GetModuleFileName in ConfigParser.cpp. I am not so sure about how to test this. I saw most test cases in the `test` folder are js files, but this change is not related with JavaScript functionality.

### DIFF
--- a/lib/Common/Core/ConfigParser.cpp
+++ b/lib/Common/Core/ConfigParser.cpp
@@ -318,7 +318,10 @@ void ConfigParser::ParseConfig(HANDLE hmod, CmdLineArgsParser &parser)
     char16 modulename[_MAX_PATH];
     char16 filename[_MAX_PATH];
 
-    GetModuleFileName((HMODULE)hmod, modulename, _MAX_PATH);
+    if (!GetModuleFileName((HMODULE)hmod, modulename, _MAX_PATH))
+    {
+        return;
+    }
     char16 drive[_MAX_DRIVE];
     char16 dir[_MAX_DIR];
 
--- a/lib/Common/Core/ConfigParser.cpp
+++ b/lib/Common/Core/ConfigParser.cpp
@@ -521,7 +521,7 @@ HRESULT ConfigParser::SetOutputFile(const WCHAR* outputFile, const WCHAR* openMo
     char16 moduleName[_MAX_PATH];
     if (!GetModuleFileName(0, moduleName, _MAX_PATH))
     {
-        AssertMsg(FALSE, "Get executable file name failed");
+        AssertMsg(FALSE, "Get filename of the current executable failed");
     }
     _wsplitpath_s(moduleName, nullptr, 0, nullptr, 0, fileName, _MAX_PATH, nullptr, 0);
     if (_wcsicmp(fileName, _u("WWAHost")) == 0 ||
--- a/lib/Common/Core/ConfigParser.cpp
+++ b/lib/Common/Core/ConfigParser.cpp
@@ -368,7 +368,10 @@ void ConfigParser::ProcessConfiguration(HANDLE hmod)
     bool hasOutput = false;
     char16 modulename[_MAX_PATH];
 
-    GetModuleFileName((HMODULE)hmod, modulename, _MAX_PATH);
+    if (!GetModuleFileName((HMODULE)hmod, modulename, _MAX_PATH))
+    {
+        return;
+    }
 
     // Win32 specific console creation code
     // xplat-todo: Consider having this mechanism available on other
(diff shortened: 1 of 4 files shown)


dilijev:
@suwc I think we can take this change without additional test cases, as this sort of thing is fairly difficult to test. If we decide we need to add tests later, we can do that.

lib/Common/Core/FaultInjection.cpp
```diff
@@ -1228,16 +1228,20 @@ namespace Js
         GetModuleHandleEx(GET_MODULE_HANDLE_EX_FLAG_FROM_ADDRESS, reinterpret_cast<LPCTSTR>(ip), &mod);
         offset = ip - (uintptr_t)mod;
         auto& faultModule = modulePath;
-        GetModuleFileName(mod, faultModule, MAX_PATH);
-        fwprintf(stderr, _u("***FI: Exception: %08x, module: %s, offset: 0x%p\n"),
-            ep->ExceptionRecord->ExceptionCode, faultModule, (void*)offset);
+        if (GetModuleFileName(mod, faultModule, MAX_PATH))
+        {
+            fwprintf(stderr, _u("***FI: Exception: %08x, module: %s, offset: 0x%p\n"),
+                ep->ExceptionRecord->ExceptionCode, faultModule, (void*)offset);
```
dilijev: Is it okay to continue execution if `GetModuleFileName` call fails above?

lib/Common/Core/FaultInjection.cpp
```diff
@@ -1228,16 +1228,20 @@ namespace Js
...
         uintptr_t savedOffset = 0;
         auto& mainModule = modulePath;
-        GetModuleFileName(NULL, mainModule, MAX_PATH);
-        // multiple session of Fault Injection run shares the single crash offset recording file
-        _snwprintf_s(filename, _TRUNCATE, _u("%s.FICrashes.txt"), mainModule);
+        if (GetModuleFileName(NULL, mainModule, MAX_PATH))
+        {
+            // multiple session of Fault Injection run shares the single crash offset recording file
+            _snwprintf_s(filename, _TRUNCATE, _u("%s.FICrashes.txt"), mainModule);
+        }
```
dilijev: Is it okay to continue execution if `GetModuleFileName` call fails above?

lib/Common/Core/ConfigParser.cpp
```diff
@@ -513,7 +519,10 @@ HRESULT ConfigParser::SetOutputFile(const WCHAR* outputFile, const WCHAR* openMo
 
     char16 fileName[_MAX_PATH];
     char16 moduleName[_MAX_PATH];
-    GetModuleFileName(0, moduleName, _MAX_PATH);
+    if (!GetModuleFileName(0, moduleName, _MAX_PATH))
+    {
+        AssertMsg(FALSE, "Get filename of the current executable failed");
```
dilijev: Is it okay to continue execution if `GetModuleFileName` fails in release builds (where Asserts don't hit)?

suwc:
@dilijev I agree with you.

---
In reply to: [305615082](https://github.com/Microsoft/ChakraCore/pull/3082#issuecomment-305615082) [](ancestors = 305615082)

obastemur:
My 2 cents; `GetModuleFileName` (cross platform) has some issues. `GetBinaryLocation` should replace that. See #2297 .

lib/Common/Core/FaultInjection.cpp
xiaoyinl: The buffer faultModule that GetModuleFileName fills is not used after the fwprintf, so I guess it's fine to only skip fwprintf and continue, when it fails.

dilijev:
@obastemur May be true, but I think that's out of scope of this change (unless you disagree?). Let's continue to track that discussion separately.

obastemur:
>  May be true, but I think that's out of scope of this change (unless you disagree?)

Well, what's the scope of this change?

This update is redundant and makes the future update unnecessarily complicated. There is an open Github issue with a direction and reason.

dilijev:
The scope of this change is what the title says.

We can hold off if you think the alternative is going to make progress soon, however #2297 hasn't seen any discussion in nearly a month, and in general, we'd like to merge PRs from external folks without unnecessary delays.

This PR strictly improves the state things by checking the return values. I don't especially see why this PR makes a future change more complicated.

obastemur:
@xiaoyinl Thanks for the effort. I talked @dilijev offline, we reached to a common understanding.

How `GetModuleFileName` used in ChakraCore, we don't need the return value. The replacement function is designed this in mind and already takes care of other cases.

If you want to help us on making `GetModuleFileName` usage right, I would like to have a chat with you on gitter about #2297 . Feel free to ping me on gitter so we can start talking on this (you can find me at @obastemur)

I added some details on fixing #2297, hope they are sufficient. See https://github.com/Microsoft/ChakraCore/issues/2297#issue-198068489 for implementation details.

xiaoyinl:
Thank you @dilijev, @suwc and @obastemur!

@obastemur: Yes, I'd like to work on #2297. I will leave you a message on gitter. Thank you!

obastemur:
Closing this one. #3112 is done, we will be moving to GetBinaryLocation instead. @xiaoyinl Thanks again! 👍

PR #3084, opened 2017-06-01
WASM: Limits and code style

Cellule:
Add limits when reading wasm.
Fix code style in WasmReader project

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/3084)


test/wasm/superlongsignaturemismatch.js
```diff
@@ -3,2061 +3,53 @@
 // Licensed under the MIT license. See LICENSE.txt file in the project root for full license information.
 //-------------------------------------------------------------------------------------------------------
 
-
-var module1src =`
+const module1src = `
 (module
```
Penguinwizzard: Sorry about the mess, and thanks for cleaning it up!

Cellule:
> Could you add test cases for the new limit checks? It'd help to identify cases where we might leak due to the differing execution flow.

I was trying to write such tests, but they take so long to run ( and write ).
I guess it's worth writing the tests and just marking it as slow test though.

Cellule:
So I added tests for the limits, however they take forever to run and use a ton of memory, so I have marked them as slow tests.
I decided to start using yargs-parser for command-line argument parsing, I was tired of implementing my own sloppy version every time.

Penguinwizzard:
LGTM

Issue #3085, opened 2017-06-01
Environment variable OptimizedBuild=false should prevent build optimizations


ddv239:
It is valuable for the user to have ability to set a specific environment variable to prevent all build optimizations. Chakra Core should allow this.

PR #3086, opened 2017-06-01
Allow disabling build optimizations by setting the OptimizedBuild property to anything other than "true"

ddv239:
#3085

### DIFF
--- a/Build/Common.Build.Default.props
+++ b/Build/Common.Build.Default.props
@@ -29,8 +29,9 @@
 
   <!-- Common build settings -->
   <PropertyGroup>
-    <OptimizedBuild />
-    <OptimizedBuild Condition="'$(Configuration)'=='Release' or '$(Configuration)'=='Test'">true</OptimizedBuild>
+    <DebugBuild />
+    <DebugBuild Condition="'$(Configuration)'!='Release' and '$(Configuration)'!='Test'">true</DebugBuild>
+    <OptimizedBuild Condition="'$(OptimizedBuild)'=='' and '$(DebugBuild)'!='true'">true</OptimizedBuild>
     <WindowsSDKDesktopARMSupport Condition="'$(Platform)'=='ARM'">true</WindowsSDKDesktopARMSupport>
   </PropertyGroup>
   <PropertyGroup Label="Configuration">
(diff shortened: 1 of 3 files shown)


digitalinfinity:
Thanks @ddv239 - however, in general, we prefer not to use windows build specific variables in ChakraCore builds- can you retarget this PR to ChakraFull instead? You should be able to accomplish the same result there and leave ChakraCore unaffected.

ddv239:
I would not mind retargeting it, but I don't see how it is possible: as far as i understand, certain binaries/libraries are built by chakracore and there is no way arount it.

dilijev:
(Taking discussion about ChakraFull specifics offline.)

ddv239:
Implemented the suggestion and changed the environment variable used for preventing optimizations into "OptimizedBuild".

dilijev:
@digitalinfinity FWIW proj files will treat variables the same whether they are specified in environment or msbuild command line so this change will also work when passing `"/p:OptimizedBuild=true"` or similar.

Thanks @ddv239!

ddv239:
Sorry, I just realized that making OptimizedBuild false will essentially convert it into a checked build. The goal is to have an unoptimized free build, so some more changes are needed.

ddv239:
Updated with more appropriate changes.

Build/Common.Build.Default.props
```diff
@@ -29,8 +29,9 @@
 
   <!-- Common build settings -->
   <PropertyGroup>
-    <OptimizedBuild />
-    <OptimizedBuild Condition="'$(Configuration)'=='Release' or '$(Configuration)'=='Test'">true</OptimizedBuild>
+    <FreeBuild />
+    <FreeBuild Condition="'$(Configuration)'=='Release' or '$(Configuration)'=='Test'">true</FreeBuild>
```
curtisman: Can we flip this to 'DebugBuild' instead?

ddv239: Why not. I'll make "DebugBuild" have the value of negation of the current value of FreeBuild.

curtisman: Or just change all reference of '$(FreeBuild)'!='true' to '$(Configuration)'=='Debug' and vice versa.

ddv239: I think having a bool property is cleaner: if we ever introduce another variant of a "debug build", this will work better.

ddv239:
Addressing the comment with the update

dilijev:

PR #3090, opened 2017-06-01
Avoid adding empty blocks while inserting an edge from finally block to early exit

meg-gupta:
When we add an edge from end of finally block to an early exit, we need not create additional blocks
if the Leave/LeaveNull at the end of finally was in a separate block preceeded by a Label.
Creating empty blocks like this can lead to issues with DelateLeaveChains later on, moreover we want to keep the IR clean.

Also piggybacking 2 minor fixes:
1. Don't zero out region info on dead blocks, affects liveness info in the presence of regions
2. Don't call ResetOut on non exception path in ProcessTryHandlerBailout

### DIFF
--- a/lib/Backend/FlowGraph.cpp
+++ b/lib/Backend/FlowGraph.cpp
@@ -636,29 +636,40 @@ void FlowGraph::InsertEdgeFromFinallyToEarlyExit(BasicBlock *finallyEndBlock, IR
     IR::LabelInstr *leaveLabel = IR::LabelInstr::New(Js::OpCode::Label, this->func);
     lastInstr->InsertBefore(leaveLabel);
 
-    this->AddBlock(leaveLabel, lastInstr, finallyEndBlock->GetNext(), finallyEndBlock /*prevBlock*/);
+    this->AddBlock(leaveLabel, lastInstr, nextBB, finallyEndBlock /*prevBlock*/);
     leaveLabel->SetRegion(lastLabel->GetRegion());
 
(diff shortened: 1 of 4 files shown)


meg-gupta:
Ping..Ping..

meg-gupta:
Easy to review this PR commit by commit

lib/Backend/FlowGraph.cpp
```diff
@@ -636,29 +636,40 @@ void FlowGraph::InsertEdgeFromFinallyToEarlyExit(BasicBlock *finallyEndBlock, IR
...
+        this->AddBlock(brLabel, brToExit, finallyEndBlock->GetNext(), finallyEndBlock /*prevBlock*/);
+        brLabel->SetRegion(lastLabel->GetRegion());
 
-    brLabel->SetRegion(lastLabel->GetRegion());
+        this->AddEdge(finallyEndBlock, brLabel->GetBasicBlock());
+    }
+    else
+    {
+        // If the Leave/LeaveNull at the end of finally was preceeded by a Label, we reuse the block inserting BrOnException to early exit in it
+        IR::BranchInstr *brToExit = IR::BranchInstr::New(Js::OpCode::BrOnException, exitLabel, this->func);
```
meg-gupta: Only else part is new here.

LouisLaf:
:shipit:

meg-gupta:

meg-gupta:
@dotnet-bot Please test Windows 7 ci_dev12_x64_debug

PR #3092, opened 2017-06-01
fix handle leak for recycler concurrent thread

leirocks:


mjsabby:
@leirocks when can this be merged into 1.4/1.5?

PR #3098, opened 2017-06-02
Allow allocations during concurrent sweep.

atulkatti:
1. We enable allocations during concurrent sweep using an interlocked SLIST for allocations. This list is used only during the concurrent sweep.
     2. As part of this change the heap blocks will become available for the allocators as soon as either SweepPartialReusePages or SweepPendingObjects make them available.
     3. The heap blocks we allocated from during concurrent sweep are kept track of separately and are processed during the next sweep so they will land in the heapBlockList or the fullBlockList as appropriate.


atulkatti:
This has already been reviewed.

atulkatti:
@Microsoft/chakra-developers

PR #3099, opened 2017-06-02
Fixing the Is check for console scope activation object

aneeshdk:
While checking whether an activation object is console scope or not we were not checking it for the cross context object.


aneeshdk:
@akroshg @agarwal-sandeep Can you please take a look?

agarwal-sandeep:
Do you have a test case for full repro?

aneeshdk:

PR #3107, opened 2017-06-05
xplat: fix script location

obastemur:
`dirname $0` may return relative path, while some part of the script require it to be absolute.


obastemur:
@kfarnung that would not be practical for this one.

Thanks for reviewing!

PR #3108, opened 2017-06-05
Adding a missing null check

rajatd:


lib/Backend/GlobOpt.cpp
```diff
@@ -13470,7 +13470,7 @@ GlobOpt::OptArraySrc(IR::Instr * *const instrRef)
             // Unless we're in asm.js (where it is guaranteed that virtual typed array accesses cannot read/write beyond 4GB),
             // check the range of the index to make sure we won't access beyond the reserved memory beforing eliminating bounds
             // checks in jitted code.
-            if (!GetIsAsmJSFunc())
+            if (!GetIsAsmJSFunc() && baseOwnerIndir)
```
rajatd: This is only in the case of those special virtual typed arrays. push and pop will update the length of the array and .length only loads the length. So, I think its safe

Cellule: I'm wondering if you put additional checks for whenever these are not the only case

LouisLaf:
:shipit:

Issue #638, opened 2016-03-24
toPrimitiveJitBug654983.js test sometimes fails


ianwjhalliday:
The test `test\es6\toPrimitiveJitBug654983.js` has disabled since late October 2015 (marked with the `fail` tag) because it was occassionally failing with spurious bailout traces in its output.

_Note the `Symbol.toPrimitive` feature is not fully implemented and may change when it is fully implemented rendering this test obsolete or irrelevant.  **This issue should only be investigated once work begins on the feature again.**_

_Ported from VSO bug [5244610](https://microsoft.visualstudio.com/web/wi.aspx?pcguid=cb55739e-4afe-46a3-970f-1b49d8ee7564&id=5244610)_

The test appears to have been added to verify that bailouts no longer occur in the repro test code, but seems they are.

Further investigation is needed.  The test is disabled for now.

E.g. rl output when the test fails:

``` text
3> ch.exe -DumpOnCrash -WERExceptionSupport -ExtendedErrorStackForTestHost -forceNative -off:simpleJit -bgJitDelay:0 -dynamicprofileinput:profile.dpl.UnnamedTest1438 -es6toprimitive -trace:bailout D:\164\src\test\es6\toPrimitiveJitBug654983.js >testout1177 2>&1
3>ERROR: name of output file: D:\164\src\test\es6\testout1177; size: 202; creation: Thu Oct 29 10:03:19 2015, last access: Thu Oct 29 10:03:19 2015, now: Thu Oct 29 10:03:21 2015
3>ERROR: bad output file follows ============
3>BailOut: function: foobar ( (#1.2), #3) offset: #0025 Opcode: Add_A Kind: BailOutOnImplicitCalls
3>BailOut: function: foobar ( (#1.2), #3) offset: #0025 Opcode: Add_A Kind: BailOutOnImplicitCalls
3>PASS
3>ERROR: end of bad output file ============
```

tcare:
Fixed by #3110

PR #3110, opened 2017-06-06
toPrimitiveJitBug654983.js test sometimes fails #638

tcare:
Removing test as it is incorrect after a spec update. No repro of the bailout trace issue.


tcare:
Spoke with Suwei offline, the expected TypeError is covered in toPrimitive.js under "Object toPrimitive must be Function or null else Throws typeError"

`
var o = { toString : function () {return "o"}, valueOf : function() { return 0;}};

            o[Symbol.toPrimitive]  = {}; // can only be a  function else type error
            assert.throws(function() {var a = o+1;}, TypeError, "o[Symbol.toPrimitive] must be a function", "The value of the property 'Symbol.toPrimitive' is not a Function object");

            o[Symbol.toPrimitive] = null;
            assert.doesNotThrow(function() {var a = o+1;}, "If o[Symbol.toPrimitive] is null it is ignored");
`

suwc:
:shipit:

Issue #3076, opened 2017-05-31
Assertion Failure in Parsing


natashenka:
The attached file causes an assert to fire in the Debug build.

[parseassert.txt](https://github.com/Microsoft/ChakraCore/files/1042546/parseassert.txt)

ChakraCore/out/Debug/ch parseassert.txt 
ASSERTION 8981: (/home/user/test_everywhere/ChakraCore/lib/Parser/Parse.cpp, line 722) !this->m_deferringAST
 Failure: (!this->m_deferringAST)
Illegal instruction (core dumped)

dilijev:
@natashenka It looks like this repro is incomplete -- there are missing module files and the fail function is not present in the repro. Could you provide the full set of repro files?

natashenka:
The repro should work exactly as is, with the missing modules and the undefined method. Are you trying this on the Linux build, it's possible it's a Linux-only problem.

MikeHolman:
I'm having no problem with repro. Looks like bug with defer parsing module import calls.

@suwc Can you take a look?

dilijev:
I was able to repro with a recent debug build on Linux. Trying to figure out why it didn't repro on Windows.

suwc:
Fix merged. Thanks!

PR #3111, opened 2017-06-06
Issue3076: Assertion Failure in Parsing

suwc:
Assertion is triggered in parse.cpp on this->m_deferringAST
when CreateCallNode() is called in ParseImportCall().
Fix by adding check for buildAST.

### DIFF
--- a/test/es6/bug_issue_3076.js
+++ b/test/es6/bug_issue_3076.js
@@ -1,0 +1,10 @@
+//-------------------------------------------------------------------------------------------------------
+// Copyright (C) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE.txt file in the project root for full license information.
+//-------------------------------------------------------------------------------------------------------
+
+function test() {
+    import('passmodule.js');
+}
+
+test();
--- a/test/es6/rlexe.xml
+++ b/test/es6/rlexe.xml
@@ -1421,6 +1421,13 @@
 </test>
 <test>
   <default>
+    <files>bug_issue_3076.js</files>
+    <compile-flags>-force:deferparse</compile-flags>
+    <tags>BugFix</tags>
+  </default>
+</test>
+<test>
+  <default>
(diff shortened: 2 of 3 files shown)


suwc:
Addressing https://github.com/Microsoft/ChakraCore/issues/3076

suwc:
@Microsoft/chakra-es could you review pls?

test/es6/rlexe.xml
```diff
@@ -1421,6 +1421,12 @@
 </test>
 <test>
   <default>
+    <files>bug_issue_3076.js</files>
+    <tags>BugFix</tags>
```
akroshg: don't you have to pass the forcedeferparse to repro this?

test/es6/bug_issue_3076.js
```diff
@@ -0,0 +1,328 @@
+//-------------------------------------------------------------------------------------------------------
+// Copyright (C) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE.txt file in the project root for full license information.
+//-------------------------------------------------------------------------------------------------------
+
+function f0 (a, b) {
```
akroshg: not sure why this many expression needed. Does this repro in the forcedeferparse?

test/es6/rlexe.xml
suwc: No, but it is better to use forcedeferparse. Updated.

---
In reply to: [120246928](https://github.com/Microsoft/ChakraCore/pull/3111#discussion_r120246928) [](ancestors = 120246928)

test/es6/bug_issue_3076.js
suwc: Yes, it is better to just use forcedeferparse

---
In reply to: [120246993](https://github.com/Microsoft/ChakraCore/pull/3111#discussion_r120246993) [](ancestors = 120246993)

suwc:
Thanks for the code review.

PR #3117, opened 2017-06-07
Fix region propagation of finally end blocks

meg-gupta:
We add edge from infinite loops in finally region to end of finally block, so that unreachable block elimination does not remove finally end block. We have to retain finally end block, so that we can add edge from finally end block to early exit. 

When the only edge to finally block is an edge from a loop within some child region, we can end up in propagating incorrect region info from the predecessor. We have to use the region on the finally label for such an end of finally block. 

So create a new map called leaveNullLabelToFinallyLabelMap and pick up region info from here.

Also piggybacking other small changes:
- dumping
- fix asserts for Br instructions : 
Leaves in non exception finally are converted to Br instructions, for such branches the target region is the parent of the current region.
Also, a BrOnException from finally to early exit can be converted to BrOnNoException and Br, we cannot assert for such a Br that the target region is same as Br region

### DIFF
--- a/lib/Backend/FlowGraph.cpp
+++ b/lib/Backend/FlowGraph.cpp
@@ -173,6 +173,7 @@ FlowGraph::Build(void)
         this->finallyLabelStack = JitAnew(this->alloc, SList<IR::LabelInstr*>, this->alloc);
         this->leaveNullLabelStack = JitAnew(this->alloc, SList<IR::LabelInstr*>, this->alloc);
         this->regToFinallyEndMap = JitAnew(this->alloc, RegionToFinallyEndMapType, this->alloc, 0);
+        this->leaveNullLabelToFinallyLabelMap = JitAnew(this->alloc, LeaveNullLabelToFinallyLabelMapType, this->alloc, 0);
     }
 
     IR::Instr * currLastInstr = nullptr;
@@ -301,7 +302,8 @@ FlowGraph::Build(void)
 
             this->finallyLabelStack->Push(finallyLabel);
             Assert(!this->leaveNullLabelStack->Empty());
-            this->leaveNullLabelStack->Pop();
+            IR::LabelInstr * leaveNullLabel = this->leaveNullLabelStack->Pop();
+            this->leaveNullLabelToFinallyLabelMap->Item(leaveNullLabel, leaveTarget /*new finally label*/);
 
             Assert(leaveTarget->labelRefs.HasOne());
             IR::BranchInstr * brOnException = IR::BranchInstr::New(Js::OpCode::BrOnException, finallyLabel, instr->m_func);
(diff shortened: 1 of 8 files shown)


meg-gupta:
It is easy to review this commit by commit. This PR is stacked on top of https://github.com/Microsoft/ChakraCore/pull/3090. 

The only commits to review are  on top of "Fix region asserts for Leaves converted to Br in non exception finally"

meg-gupta:
@dotnet-bot Please test Ubuntu _no_jit_shared_ubuntu_linux_test

lib/Backend/IR.h
```diff
@@ -748,10 +750,11 @@ class BranchInstr : public Instr
     static BranchInstr * New(Js::OpCode opcode, Opnd* destOpnd, LabelInstr * branchTarget, Opnd *srcOpnd, Func *func);
     static BranchInstr * New(Js::OpCode opcode, LabelInstr * branchTarget, Opnd *src1Opnd, Opnd *src2Opnd, Func *func);
 
-    BranchInstr(bool hasBailOutInfo = false) : Instr(hasBailOutInfo), m_branchTarget(nullptr), m_isAirlock(false), m_isSwitchBr(false), m_isOrphanedLeave(false), m_areCmpRegisterFlagsUsedLater(false)
+    BranchInstr(bool hasBailOutInfo = false) : Instr(hasBailOutInfo), m_branchTarget(nullptr), m_isAirlock(false), m_isSwitchBr(false), m_isOrphanedLeave(false), m_areCmpRegisterFlagsUsedLater(false), m_brFinallyToEarlyExit(nullptr)
```
rajatd: >nullptr [](start = 224, length = 7)

initialize to false

lib/Backend/FlowGraph.cpp
```diff
@@ -1721,21 +1767,26 @@ FlowGraph::Destroy(void)
                         Assert(!this->func->DoGlobOpt());
                         break;
                     case Js::OpCode::BrOnNoException:
-                        Assert(region->GetType() == RegionTypeTry || region->GetType() == RegionTypeCatch || region->GetType() == RegionTypeFinally);
+                        Assert(region->GetType() == RegionTypeTry || region->GetType() == RegionTypeCatch || region->GetType() == RegionTypeFinally ||
+                            (predRegion->GetType() == RegionTypeFinally && predBlock->GetLastInstr()->AsBranchInstr()->m_brFinallyToEarlyExit));
```
rajatd: >predRegion->GetType() == RegionTypeFinally && predBlock->GetLastInstr()->AsBranchInstr()->m_brFinallyToEarlyExit)) [](start = 29, length = 114)

This condition will be true when the early exit code was moved out as a break block?

lib/Backend/FlowGraph.cpp
```diff
@@ -636,6 +673,13 @@ void FlowGraph::InsertEdgeFromFinallyToEarlyExit(BasicBlock *finallyEndBlock, IR
     IR::LabelInstr *leaveLabel = IR::LabelInstr::New(Js::OpCode::Label, this->func);
     lastInstr->InsertBefore(leaveLabel);
 
+    if (this->leaveNullLabelToFinallyLabelMap->ContainsKey(finallyEndBlock->GetFirstInstr()->AsLabelInstr()))
+    {
+        // Add new LeaveNull label to the map, we don't delete old LeaveNull label from the map
+        Assert(this->leaveNullLabelToFinallyLabelMap->ContainsKey(finallyEndBlock->GetFirstInstr()->AsLabelInstr()));
```
rajatd: >this->leaveNullLabelToFinallyLabelMap->ContainsKey(finallyEndBlock->GetFirstInstr()->AsLabelInstr()) [](start = 15, length = 100)

Looks like the same condition you checked in the above `if`

meg-gupta: Fixed.

lib/Backend/IR.h
meg-gupta: Fixed.

lib/Backend/FlowGraph.cpp
meg-gupta: Added comment and test case for this. This happens when the early exit code is moved inside the finally block. And the early exit has 2 edges one from a root region and one from the preceeding BrOnNoException block.

LouisLaf:
Now that I look back at this:
"We add edge from infinite loops in finally region to end of finally block, so that unreachable block elimination does not remove finally end block. We have to retain finally end block, so that we can add edge from finally end block to early exit."

Is there a way to have your code work without the edge?  If the finally block has an infinite loop, the code after the early exit isn't reachable...

meg-gupta:
@LouisLaf I don't suspect globopt to run into problems without this edge. But one issue that comes up with not adding these edges is in building loops/processing basic blocks in flowgraphs. Mostly due to the fact that we don't cascade edge removal in unreachable block elimination.

rajatd:
Looks good to me as it is now, barring @LouisLaf 's question.

meg-gupta:
@LouisLaf First look without the edge, seems a like a few FlowGraph methods need changes. Ok if I visit this issue in a different PR ? Can I merge this one ?

meg-gupta:
Spoke to @LouisLaf offline. Will be merging this.

PR #3128, opened 2017-06-08
Fix some WebAssembly spec bugs

Cellule:
- Fix string tags for WebAssembly Types
- Detach Wasm array buffer on `.grow(0)`
- Set wasm export prototype to `null`
- Prevent extensions on the exports object
- Check that WebAssembly.Memory maximum > initial
- Check that WebAssembly.Table maximum > initial
- Throw exception on `new wasmFunction()`
- WebAssembly.Module.customSections sectionName argument is optional

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/3128)


test/WasmSpec/testsuite/js-api/jsapi.js
```diff
@@ -492,6 +492,14 @@ test(() => {
...
     assert_equals(buf, mem.buffer);
+
+    mem = new Memory({initial:0, maximum:1});
+    buf = mem.buffer;
+    assert_equals(buf.byteLength, 0);
+    assert_equals(mem.grow(0), 0);
+    assert_equals(buf !== mem.buffer, true);
+    assert_equals(buf.byteLength, 0);
+    assert_equals(mem.buffer.byteLength, 0);
 }, "'WebAssembly.Memory.prototype.grow' method");
```
Penguinwizzard: Would you mind duplicating this test block for an arraybuffer with length 1? There's a difference in what happens in the !ENABLE_FAST_ARRAYBUFFER case between 0-length and nonzero-length arraybuffers, and checking that growing by 0 works there too shouldn't be too much additional code.

lib/Runtime/Library/JavascriptLibrary.cpp
```diff
@@ -2901,21 +2902,21 @@ namespace Js
...
+            library->AddMember(constructor, PropertyIds::name, library->CreateStringFromCppLiteral(_u("Module")), PropertyConfigurable);
         }
         constructor->SetHasNoEnumerableProperties(true);
     }
 
     void JavascriptLibrary::InitializeWebAssemblyObject(DynamicObject* webAssemblyObject, DeferredTypeHandlerBase * typeHandler, DeferredInitializeMode mode)
     {
         JavascriptLibrary* library = webAssemblyObject->GetLibrary();
-        int slots = 8;
+        int slots = 9;
```
Penguinwizzard: Does the increment on this need to be conditional on IsES6FunctionNameEnabled, similarly to how it's conditionally incremented when ENABLE_WABT is defined?

lib/WasmReader/WasmLimits.h
```diff
@@ -16,7 +16,7 @@ namespace Wasm {
         static const uint32 MaxGlobals = 1000000;
         static const uint32 MaxDataSegments = 100000;
         static const uint32 MaxElementSegments = 10000000;
-        static const uint32 MaxTableSize = 10000000;
+        static const uint32 MaxTableSize = DEFAULT_CONFIG_WasmMaxTableSize;
```
Penguinwizzard: This is now unused, but I see why you kept it. Going forward, is this maybe something we want to change in how ConfigFlags list can work, such that we could privately scope these things somewhere and have the config flag list include them somehow?

test/WasmSpec/testsuite/js-api/jsapi.js
Cellule: This is already a test case, see above

lib/Runtime/Library/JavascriptLibrary.cpp
Cellule: I am unclear whether we need to have the exact number of inline slots.
From what I understand it's a best effort and it's better to have more then less.

lib/WasmReader/WasmLimits.h
Cellule: I don't think we can really change how config flags work because it has to be top-level.
I decided to keep it there for visibility even if it is now unused

Issue #2735, opened 2017-03-23
ASAN: Module tests commonly fail to free memory.


Penguinwizzard:
There are a number of ES6 module tests that fail to free memory appropriately. This appears to be due to WScriptJSRT not freeing the source code of modules it loads. The affected tests have been marked as exclude_sanitize_address for now in PR #2730 to allow ASAN CI to be enabled. These tests are:

* moduletest1.js
* moduletest2.js
* module-syntax.js
* module-syntax1.js
* module-functionality.js
* module-syntax.js (force:deferparse)
* module-syntax1.js (force:deferparse)
* module-namespace.js
* test_bug_2645.js
* bug_issue_3076.js

Note: This may be an API issue; an (admittedly cursory) look around the APIs related to JsParseModuleSource didn't reveal an easy way to register a finalize callback.

akroshg:
Fixes are merged.

Penguinwizzard:
@akroshg Do you have a commit ref for when these issues were fixed?

akroshg:
isn't this fixed by #3129 ?

Penguinwizzard:
No, that changeset just disables the test. Re-opening.

Issue #2378, opened 2017-01-13
AsmJs/WAsm calling convention issues on xplat


Penguinwizzard:
When attempting to pass a floating-point argument to an asmjs/wasm function on xplat, the caller and callee don't agree on the calling convention. This causes correctness issues, as we end up often using the wrong value. This occurs in JITted code.

**Repro:**
Remove these lines from test/runtest.py ([link](https://github.com/Microsoft/ChakraCore/blob/master/test/runtests.py#L129-L130)):
```python
not_compile_flags = set(['-simdjs']) \  
    if sys.platform != 'win32' else None
```
Run `runtests.py -d AsmJSFloat`.

Note: There's a separate bug filed related to removing the lines hiding this issue from runtest.py at https://github.com/Microsoft/ChakraCore/issues/2412.

**Repro file:**
Here's a separate file by @Cellule that reproduces the issue with -maic:0
```js
function AsmModuleDouble() {
        "use asm";

        function add(x) {
                    x = +x;
                    return +sub(1.5);
                }
        function sub(x) {
                    x = +x;
                    return +(x);
                }
        return {
                    add: add
                };
}

var asmModuleDouble = AsmModuleDouble();     // produces AOT-compiled version
print(asmModuleDouble.add(2, 1))
```

jianchun:
@Penguinwizzard Indeed, didn't get to complete Simd when I worked on xplat JIT. That's why Simd tests were disabled on xplat by that line.

Cellule:
From what I was able to gather this happens because `LowererMDArch::LowerCallArgs` calls in `LowererMDArch::GetArgSlotOpnd` which tries to use xmm registers for float argument when we are supposed to put them on the stack on xplat

Cellule:
@jianchun The problem is almost all asm.js test have that flag even though simd is not used (I am not entirely sure the reason why, but I think there's a good one).

Penguinwizzard:
@jianchun If Simd doesn't work on xplat, should we disable building Simd/not recognize the flag on xplat?

jianchun:
@Penguinwizzard @Cellule At the time I felt Simd was close to completion, but I was not exactly sure of the calling conventions. Left Simd code in build so it would not diverge further (Simd code kept introducing build failures at the time).

MikeHolman:
Is wrong calling convention still an issue?

Cellule:
[...]

PR #3129, opened 2017-06-08
Fix a few issues blocking ASAN runs

Penguinwizzard:
Disable one test in ASAN due to a known issue with modules in ch (bug #2735), disable one test on xplat due to wasm calling convention issues (bug #2378), keep a loop from accessing data beyond an array bound in the pal code, and fix the casing on one test file.


obastemur:
:shipit:

PR #3135, opened 2017-06-09
xplat: Improve ICU fallback conditions

obastemur:
node-chakracore was silently falling back to `no-icu`. Couple of changes to be more proactive when a custom ICU path is given. Similarly, do not fallback to `no-icu` when custom ICU path is given.


obastemur:
@kfarnung @kunalspathak Thanks for the review

PR #3136, opened 2017-06-09
lock down counters for inlinee function body

leirocks:
we assume a function body counters won't change after JIT started so we did lock down the counters while the function begins execution or creating codeGen workitem on it. however, the inlinee function body can be jitted without creating work item or begin execution. this change is to lockdown the counters while creating JIT time data.

a better place to lockdown for function body might be right after bytecode generated, however in Asmjs and WAsm design, the counters(like varcount) can be changed after bytecode generated. that change will be pretty messy.


curtisman:
:shipit:

PR #3138, opened 2017-06-09
Remove RFG code from Chakra

meg-gupta:


meg-gupta:
@kunalspathak I don't think so, none of the tests broke : )

PR #3144, opened 2017-06-12
Disable Optimizing try finally when we bailout on early exit

meg-gupta:


lib/Backend/JITTimeProfileInfo.h
```diff
@@ -63,6 +63,7 @@ class JITTimeProfileInfo
     bool IsLoopImplicitCallInfoDisabled() const;
     bool IsPowIntIntTypeSpecDisabled() const;
     bool IsTagCheckDisabled() const;
+    bool IsOptimizeTryFinally() const;
```
rajatd: >IsOptimizeTryFinally [](start = 9, length = 20)

Can you call this function IsOptimizeTryFinallyDisabled?

meg-gupta: Thanks! Done.

rajatd:
:shipit:

lib/Backend/BailOut.cpp
```diff
@@ -2200,6 +2200,12 @@ void BailOutRecord::ScheduleFunctionCodeGen(Js::ScriptFunction * function, Js::S
                     rejitReason = RejitReason::PowIntIntTypeSpecDisabled;
                 }
             }
+            case IR::BailOutOnEarlyExit:
+            {
+                profileInfo->DisableOptimizeTryFinally();
```
LouisLaf: You should follow the same pattern as above:
if (already disabled)
   reThunk to the latest JIT'd code
else
   disable it

lib/Backend/BailOut.cpp
```diff
@@ -2200,6 +2200,12 @@ void BailOutRecord::ScheduleFunctionCodeGen(Js::ScriptFunction * function, Js::S
...
+                rejitReason = RejitReason::OptimizeTryFinallyDisabled;
+                break;
+            }
```
LouisLaf: Add the same code to the JIT loop body version of this function.

rajatd: we don't optimize try finally's in jitted loop bodies

lib/Backend/BailOut.cpp
```diff
@@ -2199,6 +2199,20 @@ void BailOutRecord::ScheduleFunctionCodeGen(Js::ScriptFunction * function, Js::S
                     profileInfo->DisablePowIntIntTypeSpec();
                     rejitReason = RejitReason::PowIntIntTypeSpecDisabled;
                 }
+                break;
```
LouisLaf: >break [](start = 16, length = 5)

Yikes!  Good catch....

LouisLaf:
:shipit:

PR #3146, opened 2017-06-13
Asm.js jitLoopBody slot signed type

Cellule:
Make sure we create loop body slot opnd as signed int.

[VSO bug](https://microsoft.visualstudio.com/OS/_workitems?id=9883547)

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/3146)

### DIFF
--- a/lib/Backend/IRType.cpp
+++ b/lib/Backend/IRType.cpp
@@ -57,6 +57,24 @@ bool IRType_IsSimd128(IRType type)
     return type >= TySimd128F4 && type <= TySimd128D2;
 }
 
+IRType IRType_EnsureSigned(IRType type)
+{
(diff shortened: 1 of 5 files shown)


lib/Backend/IRType.cpp
```diff
@@ -57,6 +57,24 @@ bool IRType_IsSimd128(IRType type)
     return type >= TySimd128F4 && type <= TySimd128D2;
 }
 
+IRType IRType_EnsureSigned(IRType type)
+{
+    if (IRType_IsUnsignedInt(type))
+    {
+        return (IRType)(type - (TyUint8 - TyInt8));
+    }
```
MikeHolman: maybe worth a compile assert since you are making some assumptions about these enum values.

PR #3152, opened 2017-06-13
wait for agent thread before shutdown

leirocks:
resloving #3149 and re-enable the tests

### DIFF
--- a/bin/ch/WScriptJsrt.cpp
+++ b/bin/ch/WScriptJsrt.cpp
@@ -1130,17 +1130,24 @@ JsValueRef __stdcall WScriptJsrt::BroadcastCallback(JsValueRef callee, bool isCo
 
             LONG count = (LONG)threadData->children.size();
             threadData->hSemaphore = CreateSemaphore(NULL, 0, count, NULL);
-            //Clang does not support "for each" yet
-            for (auto i = threadData->children.begin(); i != threadData->children.end(); i++)
+            if (threadData->hSemaphore)
             {
-                auto child = *i;
-                SetEvent(child->hevntReceivedBroadcast);
+                //Clang does not support "for each" yet
+                for (auto i = threadData->children.begin(); i != threadData->children.end(); i++)
+                {
+                    auto child = *i;
+                    SetEvent(child->hevntReceivedBroadcast);
+                }
+
+                WaitForSingleObject(threadData->hSemaphore, INFINITE);
+                CloseHandle(threadData->hSemaphore);
+                threadData->hSemaphore = INVALID_HANDLE_VALUE;
             }
-
-            WaitForSingleObject(threadData->hSemaphore, INFINITE);
-            CloseHandle(threadData->hSemaphore);
-            threadData->hSemaphore = INVALID_HANDLE_VALUE;
-
+            else
+            {
+                fwprintf(stderr, _u("Couldn't create semaphore.\n"));
+                fflush(stderr);
+            }
 
             ChakraRTInterface::JsReleaseSharedArrayBufferContentHandle(threadData->sharedContent);
         }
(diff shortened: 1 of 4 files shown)


leirocks:
@dotnet-bot test all please

leirocks:
@dotnet-bot test this please

bin/ch/WScriptJsrt.cpp
```diff
@@ -3,6 +3,7 @@
 // Licensed under the MIT license. See LICENSE.txt file in the project root for full license information.
 //-------------------------------------------------------------------------------------------------------
 #include "stdafx.h"
+#include <vector>
```
dilijev: Do we already depend on `<vector>` anywhere else? If not, do we already have an implementation of a vector-like object in ChakraCore that you could depend on?

bin/ch/WScriptJsrt.cpp
```diff
@@ -960,25 +961,23 @@ bool WScriptJsrt::Uninitialize()
...
 
-        WaitForSingleObject(GetRuntimeThreadLocalData().threadData->hSemaphore, INFINITE);
-        CloseHandle(GetRuntimeThreadLocalData().threadData->hSemaphore);
-        GetRuntimeThreadLocalData().threadData->hSemaphore = INVALID_HANDLE_VALUE;
+        DWORD waitRet = WaitForMultipleObjects(count, &childrenHandles[0], TRUE, INFINITE);
+        Assert(waitRet == WAIT_OBJECT_0);
+
+        for (auto i = GetRuntimeThreadLocalData().threadData->children.begin(); i != GetRuntimeThreadLocalData().threadData->children.end(); i++)
+        {
+            delete *i;
```
dilijev: We use `new`/`delete` in `ch`? Is this the right approach?

bin/ch/WScriptJsrt.cpp
leirocks: that's in ch, we are already using std::list, std::string etc

bin/ch/WScriptJsrt.cpp
leirocks: yes, I think we are using it in ch in many places already

leirocks:
@dilijev this change is not changing $262 itself. it's waiting the agent thread to complete before the process terminated

bin/ch/WScriptJsrt.cpp
```diff
@@ -960,25 +961,23 @@ bool WScriptJsrt::Uninitialize()
     if (GetRuntimeThreadLocalData().threadData && !GetRuntimeThreadLocalData().threadData->children.empty())
     {
         LONG count = (LONG)GetRuntimeThreadLocalData().threadData->children.size();
```
akroshg: >GetRuntimeThreadLocalData().threadData [](start = 27, length = 38)

nit: should we have local variable for GetRuntimeThreadLocalData().threadData ?

bin/ch/WScriptJsrt.cpp
akroshg: >delete *i [](start = 12, length = 9)

should we remove from the threadDAta->children set once this loop is done?

leirocks: we can but we don't need to though, since the thread is shutting down the whole set will be deleted while the TLS data is destructed

bin/ch/WScriptJsrt.cpp
leirocks: sure

bin/ch/WScriptJsrt.cpp
dilijev: I think @Penguinwizzard and ASAN would appreciate the cleanup even if it's technically unnecessary :)

akroshg:
:shipit:

PR #3153, opened 2017-06-13
fix invalid assert during property enumeration

MikeHolman:
In case cache was disabled, PropertyValueInfo might not have the expected PropertyString. Change assert to ignore this case.

OS: 12327564

### DIFF
--- a/lib/Runtime/Types/DynamicObjectPropertyEnumerator.cpp
+++ b/lib/Runtime/Types/DynamicObjectPropertyEnumerator.cpp
@@ -253,7 +253,7 @@ namespace Js
 
         if (info.GetPropertyString() != nullptr && info.GetPropertyString()->ShouldUseCache())
         {
-            Assert(propertyString == info.GetPropertyString());
+            Assert(info.IsNoCache() || propertyString == info.GetPropertyString());
             CacheOperators::CachePropertyRead(startingObject, this->object, false, propertyId, false, &info, scriptContext);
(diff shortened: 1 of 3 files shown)


lib/Runtime/Types/DynamicObjectPropertyEnumerator.cpp
```diff
@@ -253,7 +253,7 @@ namespace Js
 
         if (info.GetPropertyString() != nullptr && info.GetPropertyString()->ShouldUseCache())
         {
-            Assert(propertyString == info.GetPropertyString());
+            Assert(info.IsNoCache() || propertyString == info.GetPropertyString());
```
rajatd: >IsNoCache [](start = 24, length = 9)

if the cache was disabled, is it ok to cache the property string on it?

MikeHolman: I thought it was fine, and would be a nop, but probably better off to add this check as a condition of the if anyway.

rajatd: Ok, just that the new condition is duplicated :)

---
In reply to: [126219071](https://github.com/Microsoft/ChakraCore/pull/3153#discussion_r126219071) [](ancestors = 126219071)

MikeHolman: oops, fixed!

rajatd:
:shipit:

PR #3155, opened 2017-06-13
OS12095746: Missed ModuleReady callback

suwc:
This change addresses three issue:

1) Early error in root module
Chakra fails to call host's ModuleReady callback when it encounters
an early error during parsing of a root module (no parent)
Fix: add 'ModuleReady' notification to ParseSource()

2) Error in dependent module used by a 2nd root module
Error in a dependent module is bubbled up the hierarchy and notified
through the root module (or the dynamically-imported module). The 2nd
encounter of the dependent module referenced by a 2nd root module fails
to trigger a ModuleReady notification to the host
Fix: add checks for dependent modules' errorObject and report failure through HRESULT

3) import("mod.js").then().catch()
This sematic should be supported.
Fix: add parsing for post-fix operators for "import()" code path in parser.

### DIFF
--- a/lib/Parser/Parse.cpp
+++ b/lib/Parser/Parse.cpp
@@ -2465,7 +2465,10 @@ ParseNodePtr Parser::ParseImport()
     // import()
     if (m_token.tk == tkLParen)
     {
-        return ParseImportCall<buildAST>();
+        ParseNodePtr pnode = ParseImportCall<buildAST>();
+        BOOL fCanAssign;
+        IdentToken token;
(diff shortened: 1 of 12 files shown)


suwc:
@Microsoft/chakra-es @Microsoft/chakra-developers could you review pls?

lib/Parser/Parse.cpp
```diff
@@ -3262,10 +3265,12 @@ LFunction :
             m_pscan->Scan();
             if (m_token.tk == tkLParen)
             {
-                return ParseImportCall<buildAST>();
```
boingoing: I think this would read a little cleaner using `ChkCurTok(tkLParen, ERRnoLparen);`

PR #3156, opened 2017-06-14
OS: 12292687: TVS fix for missing lower bound check in enum

tcare:


tcare:
@Microsoft/chakra-developers please review

lib/Runtime/Base/ThreadContext.cpp
```diff
@@ -2252,7 +2252,7 @@ void ThreadContext::SetWellKnownHostTypeId(WellKnownHostType wellKnownType, Js::
 {
     AssertMsg(wellKnownType <= WellKnownHostType_Last, "ThreadContext::SetWellKnownHostTypeId called on unknown type");
```
MSLaguana: Shouldn't you also add the `>= 0` check to the assert here as well?

tcare: This is mainly just to please the analysis, but I could add one here

PR #3158, opened 2017-06-14
fix couple of checks to look at trampoline addrs if applicable

MikeHolman:
This was incorrectly used in a debug only validation path, which could lead to assert incorrectly firing.
However, I noticed this when investigating a different bug and haven't seen any case where this is actually an issue.
Also, removed few reinterpret casts that are no longer necessary.

### DIFF
--- a/lib/Runtime/Base/FunctionBody.cpp
+++ b/lib/Runtime/Base/FunctionBody.cpp
@@ -3451,9 +3451,7 @@ namespace Js
     bool FunctionBody::IsSimpleJitOriginalEntryPoint() const
     {
         const FunctionEntryPointInfo *const simpleJitEntryPointInfo = GetSimpleJitEntryPointInfo();
-        return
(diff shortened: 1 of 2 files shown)


lib/Runtime/Base/FunctionBody.cpp
```diff
@@ -9708,7 +9706,7 @@ namespace Js
                 currentCookie = (void*)currentNativeCodegen;
 #endif
 
-                if (this->jsMethod == reinterpret_cast<Js::JavascriptMethod>(this->GetNativeAddress()))
+                if (this->jsMethod == this->GetNativeEntrypoint())
```
Penguinwizzard: The functional part of the change is the change of the function called here so that we consider the thunk if we have one, with the rest of the changes being entirely clean-up, correct? If so, then LGTM.

MikeHolman:
@dotnet-bot test this

MikeHolman:

PR #3159, opened 2017-06-14
Support reading config file on xplat

digitalinfinity:
Add simple xplat support for reading jscript.config

### DIFF
--- a/lib/Common/Core/ConfigParser.cpp
+++ b/lib/Common/Core/ConfigParser.cpp
@@ -313,7 +313,6 @@ void ConfigParser::ParseConfig(HANDLE hmod, CmdLineArgsParser &parser)
     Assert(!_hasReadConfig);
     _hasReadConfig = true;
 
-    char16 configBuffer[MaxTokenSize];
     int err = 0;
     char16 modulename[_MAX_PATH];
     char16 filename[_MAX_PATH];
@@ -326,6 +325,7 @@ void ConfigParser::ParseConfig(HANDLE hmod, CmdLineArgsParser &parser)
     _wmakepath_s(filename, drive, dir, _configFileName, _u(".config"));
 
     FILE* configFile;
+#ifdef _WIN32
     if (_wfopen_s(&configFile, filename, _u("r, ccs=UNICODE")) != 0 || configFile == nullptr)
     {
         WCHAR configFileFullName[MAX_PATH];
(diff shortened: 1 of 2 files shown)


kunalspathak:
Nice, I had this change sitting in one of my [fork](https://github.com/kunalspathak/node-chakracore/commit/07adb237ed8fd8303be02d9e31f8cf02b513cc76) since October last year. HHm.

lib/Common/Core/ConfigParser.cpp
```diff
@@ -337,19 +337,70 @@ void ConfigParser::ParseConfig(HANDLE hmod, CmdLineArgsParser &parser)
...
         }
+
         if (_wfopen_s(&configFile, filename, _u("r, ccs=UNICODE")) != 0 || configFile == nullptr)
         {
             return;
         }
     }
+#else
+    // Two-pathed for a couple reasons
+    // 1. PAL doesn't like the ccs option passed in.
```
obastemur: IIRC; PAL doesn't care anything but the first char. Thus, Windows code above works as is.

lib/Common/Core/ConfigParser.cpp
```diff
@@ -337,19 +337,70 @@ void ConfigParser::ParseConfig(HANDLE hmod, CmdLineArgsParser &parser)
...
+    // 2. _wfullpath is not implemented in the PAL.
+    // Instead, on xplat, we'll check the HOME directory to see if there is
```
obastemur: Why Home folder? Why `_wfullpath` is needed here? We may define `wfullpath` as no op here.

lib/Common/Core/ConfigParser.cpp
```diff
@@ -337,19 +337,70 @@ void ConfigParser::ParseConfig(HANDLE hmod, CmdLineArgsParser &parser)
...
+            if (curChar == FINISHED)
+            {
+                break;
+            }
+
+            index = 0;
+        }
+        else
+        {
+            configBuffer[index++] = (char16) curChar;
```
obastemur: Correct me if I'm wrong but if the buffer is `unicode` this fails.

lib/Common/Core/ConfigParser.cpp
digitalinfinity: When I tried this out, wfopen failed if I passed in ccs=UNICODE- I didn't dig in too deeply since this is test code and I need something up quickly to unblock myself on xplat 😄

lib/Common/Core/ConfigParser.cpp
digitalinfinity: no particular reason other than it's the closest equivalent on xplat to the desktop on Windows and allows you to keep a jscript.config outside the repo in case you clean the repo

lib/Common/Core/ConfigParser.cpp
digitalinfinity: Yup, correct- in this particular case, Unicode is not particularly interesting- again, test code and I can't think of any config switch that is unicode

lib/Common/Core/ConfigParser.cpp
obastemur: FYI; Just tested and it works this way on xplat. In case if it's not a big hassle, we might use the same code for both?

lib/Common/Core/ConfigParser.cpp
obastemur: Gotcha. Sure, home folder is more convenient on xplat.

lib/Common/Core/ConfigParser.cpp
obastemur: I see. so probably we don't even need to put a comment there? i.e. user path is unicode? but I guess there is no path variable within config file?

obastemur:
Couple of minor make-up things, LGTM otherwise.

lib/Common/Core/ConfigParser.cpp
digitalinfinity: I'm missing something- are you saying that this line of code works on Linux:
`_wfopen_s(&configFile, filename, _u("r, ccs=UNICODE")`

How does it work? I see the following use of mode in the PAL implementation of PAL_fopen:
```
PAL_fopen(const char * fileName, const char * mode)
{
    PAL_FILE *f = NULL;
    LPSTR supported = NULL;
    LPSTR UnixFileName = NULL;
    struct stat stat_data;
    BOOL bTextMode = TRUE;

    PERF_ENTRY(fopen);
    ENTRY("fopen ( fileName=%p (%s) mode=%p (%s))\n", fileName, fileName, mode , mode );

    _ASSERTE(fileName != NULL);
    _ASSERTE(mode != NULL);

    if ( *mode == 'r' || *mode == 'w' || *mode == 'a' )
    {
        supported = MapFileOpenModes( (char*)mode,&bTextMode);
```

This defers to MapFileOpenModes, and when I look at that function, I don't see any handling of `ccs=Unicode`. 

This matches what I see when I run the code too (that calling _wfopen_s with `ccs=Unicode` fails but calling it without passes. Can you help me understand what I missed?

obastemur: Well, when I copy and paste the `wfopen` calls to non-Windows part, it works as is.

`MapFileOpenModes` doesn't handle `ccs=Unicode`, it simply ignores.

Looking at the 2 versions of your implementation, isn't the same thing you are trying to achieve ? What I'm trying to say is that, well, it's not needed to code twice.

```
    if (_wfopen_s(&configFile, filename, _u("r, ccs=UNICODE")) != 0 || configFile == nullptr)
VS
    if (_wfopen_s(&configFile, filename, _u("r")) != 0 || configFile == nullptr)
```

digitalinfinity: heh, you forced me to debug this- I understand what's going on now. 
- _wfopen_s's implementation [here](https://github.com/Microsoft/ChakraCore/blob/master/pal/src/cruntime/file.cpp#L358) limits the mode to 10 chars (!) - so calling _wfopen with "r, ccs=UNICODE" fails since the mode string is longer than 10 chars
- If I fix that issue, then I hit the issue that I was expecting because of [this issue in MapFileOpenModes](https://github.com/Microsoft/ChakraCore/blob/master/pal/src/cruntime/file.cpp#L123) - D is not supported (lowercase-d would be ignored I think). That's probably why ccs=Unicode might have worked?

I'm leaning on just keeping the two different calls as is and not touching the PAL. Thoughts?

obastemur: Well, second should be fine then ? For the first one, opened #3176 . In case you would like to keep two separate implementations there, please comment on this issue. So, whoever takes care of that part, should also update the usage here.

digitalinfinity: Sorry, wasn't clear- if you pass a mode string with "D", PAL prints out an error and returns NULL causing wfopen_s to fail. I'm going to leave the code as is for now, but I'll comment on the issue to point out the second issue - I agree that whoever fixes #3176 should fix both issues

lib/Common/Core/ConfigParser.cpp
digitalinfinity: Added a comment here to make it clear

digitalinfinity:
@obastemur @kfarnung @kunalspathak thanks for the reviews!

PR #3162, opened 2017-06-14
On source rundown we don't have to validate the entry point for nested functions

aneeshdk:
On debugger attach while reparsing the function we have to save the entry point as it is for the nested function. We don't have to validate the entry point.


aneeshdk:

pleath:
LGTM. Thanks.

PR #3163, opened 2017-06-14
dump writebarrier cardtable symbol

leirocks:
fix swb build on windows
ignore Build/VcBuild so we can create symbolic link for VcBuild

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -101,3 +101,6 @@ deps/
 android-toolchain-arm/
 cc-toolchain/
 /TraceOutput.txt
+Build/VcBuild
+Build/VcBuild.NoJIT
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -101,3 +101,6 @@ deps/
 android-toolchain-arm/
 cc-toolchain/
 /TraceOutput.txt
+Build/VcBuild
```
jianchun: These already in .gitignore? https://github.com/Microsoft/ChakraCore/blob/master/.gitignore#L42

leirocks: that was directories, they are files in my change. I'm creating symbolic link to my SSD

PR #3167, opened 2017-06-15
Fix xplat test warning: fix casing of test in rlexe.xml

dilijev:


dilijev:
@Penguinwizzard I considered that, but re-casing files is a pain and in this case we can avoid it with no ill effects.

PR #3168, opened 2017-06-15
Fix PAL VS project and add newer files

Cellule:
This should stop VS from asking to upgrade pal.vcxproj the first time you open the solution (or after git clean)

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/3168)


Cellule:
Should I target master for this change ?

dilijev:
LGTM. Does the Windows build even depend on PAL? Whether to target here or master is a matter of: (1) how annoying is it to go unfixed; (2) how risky is this? I'd say not that risky -- if changing the project file doesn't break the build, there should be no other problems.

Cellule:
It is not included in the build at all. This is only to give visibility in the solution when making changes to function that are reimplemented in pal.
I think I will target 1.6 then and just call it a day

PR #3169, opened 2017-06-15
use caller profile data for inline functions for array access

sigatrev:


lib/Backend/GlobOpt.cpp
```diff
@@ -3514,7 +3514,7 @@ GlobOpt::OptSrc(IR::Opnd *opnd, IR::Instr * *pInstr, Value **indirIndexValRef, I
     if (val)
     {
         ValueType valueType(val->GetValueInfo()->Type());
-        if (valueType.IsLikelyNativeArray() && !valueType.IsObject() && instr->IsProfiledInstr())
+        if (valueType.IsLikelyNativeArray() && !valueType.IsObject() && instr->IsProfiledInstr() && !instr->m_func->IsInlined())
```
LouisLaf: >veArray() && !valueType.IsObject() && instr->IsProfiledInstr() && !instr->m_func->IsInlined()) [](start = 34, length = 94)

Add a comment explaining this block of code, and the situation you are fixing here.

PR #3170, opened 2017-06-16
JS_REENTRANT macro was missing in the sort function.

akroshg:
FillFromPrototype was missing the JS_REENTRANT, fixed that.


akroshg:
@suwc please have a look.

suwc:
:shipit:

PR #3171, opened 2017-06-16
Fix MOVSXD encoding for exception filter

Cellule:
We were incorrectly reading the opcode for MOVSXD causing the AV in virtual memory to escape instead of throwing a WebAssembly runtime exception.
Since the AV occurs in the 8GB array, it is not a security concern.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/3171)


Cellule:
@dotnet-bot test 
Windows ci_disablejit_x64_debug

PR #3173, opened 2017-06-16
Fix bug inlining apply with no args

sigatrev:


test/inlining/rlexe.xml
```diff
@@ -274,4 +274,10 @@
       <compile-flags>-loopinterpretcount:1 -bgjit- -force:inline</compile-flags>
     </default>
   </test>
+  <test>
+    <default>
+      <files>bug9936017.js</files>
+      <compile-flags>-off:simplejit -oopjit-</compile-flags>
```
rajatd: so, the default compile flags that the interpreted (misnomer) flavor of our unittests get run with are -maxinterpretcount:1 and -maxsimplejitruncount:1. While -off:simplejit and -oopjit- are good to use for focussed private testing, as much as possible, we should try to checkin tests which can test the intended behaviour with -maxinterpretcount:1 and -maxsimplejitruncount:1. For example, in your case, having a third call to test1 and running with -mic:1 -msjrc:1 would also repro the issue.

sigatrev: Adding a third call to test1 removes the need for -off:simplejit, but I can't seem to get the test to fail without -oopjit-. It takes much longer to run and eventually prints "Passed" regardless of the number of calls to test1. Is there something I am missing?

---
In reply to: [122547975](https://github.com/Microsoft/ChakraCore/pull/3173#discussion_r122547975) [](ancestors = 122547975)

sigatrev:
creating a new PR starting from release/1.6

PR #3190, opened 2017-06-20
xplat: fix ICU lib tilde path problem

obastemur:
Bash doesn't auto expand tilde on the path hence directory exists condition fails


obastemur:
@kfarnung @kunalspathak Thanks for the review

PR #3193, opened 2017-06-20
Adding the ability to dump perf.map files on non-windows systems

MSLaguana:
To enable tools such as `perf` to provide diagnostic information on
generated code, we add the ability to dump out a mapping from memory
addresses to function names. Sending a `SIGUSR2` signal to a ChakraCore
process on Linux or OSX will cause it to write to `/tmp/perf-<pid>.map`
the next time the process performs a stack probe.

### DIFF
--- a/lib/Jsrt/JsrtHelper.cpp
+++ b/lib/Jsrt/JsrtHelper.cpp
@@ -149,6 +149,10 @@ void JsrtCallbackState::ObjectBeforeCallectCallbackWrapper(JsObjectBeforeCollect
     #ifdef VTUNE_PROFILING
         VTuneChakraProfile::Register();
     #endif
+#if PERFMAP_TRACE_ENABLED
+        PlatformAgnostic::PerfTrace::Register();
+#endif
+
         ValueType::Initialize();
         ThreadContext::GlobalInitialize();
 
--- a/lib/Runtime/Base/ThreadContext.cpp
+++ b/lib/Runtime/Base/ThreadContext.cpp
@@ -1735,6 +1735,13 @@ ThreadContext::ProbeStack(size_t size, Js::ScriptContext *scriptContext, PVOID r
 {
     this->ProbeStackNoDispose(size, scriptContext, returnAddress);
 
+#if PERFMAP_TRACE_ENABLED
+    if (PlatformAgnostic::PerfTrace::mapsRequested)
+    {
+        PlatformAgnostic::PerfTrace::WritePerfMap();
+    }
+#endif
+    
     // BACKGROUND-GC TODO: If we're stuck purely in JITted code, we should have the
     // background GC thread modify the threads stack limit to trigger the runtime stack probe
     if (this->callDispose && this->recycler->NeedDispose())
--- a/lib/Runtime/PlatformAgnostic/PerfTrace.h
+++ b/lib/Runtime/PlatformAgnostic/PerfTrace.h
@@ -1,0 +1,33 @@
+//-------------------------------------------------------------------------------------------------------
+// Copyright (C) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE.txt file in the project root for full license information.
+//-------------------------------------------------------------------------------------------------------
+#pragma once
+
+#include <signal.h>
+
+//
+// perf or perf_events is a tool for tracing in linux.
+// In order to make sense of perf traces in the presence of a JIT,
+// some metadata must be provided describing what memory address ranges
+// correspond to what compiled function.
+//
+
+
+namespace PlatformAgnostic
+{
+
+//
+// Encapsulates all perf event logging and registration related to symbol decoding.
+//
+class PerfTrace
+{
+public:
+    static void Register();
+
+    static void WritePerfMap();
+
+    static volatile sig_atomic_t mapsRequested;
+};
+
+};
--- a/lib/Runtime/PlatformAgnostic/Platform/CMakeLists.txt
+++ b/lib/Runtime/PlatformAgnostic/Platform/CMakeLists.txt
@@ -7,6 +7,7 @@ set(PL_SOURCE_FILES
   Linux/UnicodeText.ICU.cpp
   Linux/NumbersUtility.cpp
   Linux/Thread.cpp
+  Linux/PerfTrace.cpp
   Common/Trace.cpp
   Common/SystemInfo.Common.cpp
   )
--- a/lib/Runtime/PlatformAgnostic/Platform/Linux/PerfTrace.cpp
+++ b/lib/Runtime/PlatformAgnostic/Platform/Linux/PerfTrace.cpp
@@ -1,0 +1,152 @@
+//-------------------------------------------------------------------------------------------------------
+// Copyright (C) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE.txt file in the project root for full license information.
+//-------------------------------------------------------------------------------------------------------
+#include "Common.h"
+#include "Runtime.h"
+#include "ChakraPlatform.h"
+
+#include <signal.h>
+#include <errno.h>
+#include <unistd.h>
+
+using namespace Js;
+
+//
+//  Handle the SIGUSR2 signal by setting a flag
+//  indicating that we should call WritePerfMap later on
+//
+void handle_signal(int signum)
+{
+    PlatformAgnostic::PerfTrace::mapsRequested = 1;
+}
+
+namespace PlatformAgnostic
+{
+
+volatile sig_atomic_t PerfTrace::mapsRequested = 0;
+  
+//
+// Registers a signal handler for SIGUSR2 
+//
+void PerfTrace::Register()
+{
+#if PERFMAP_TRACE_ENABLED
+    struct sigaction newAction = {0};
+    newAction.sa_flags = SA_RESTART;
+    newAction.sa_handler = handle_signal;
+    sigemptyset(&newAction.sa_mask);
+
+    struct sigaction previousAction = {0};
+
+    if (-1 == sigaction(PERFMAP_SIGNAL, &newAction, &previousAction))
+    {
+        AssertMsg(errno, "PerfTrace::Register: sigaction() call failed\n");
+    }
+#endif
+}
+
+void  PerfTrace::WritePerfMap()
+{
+#if ENABLE_NATIVE_CODEGEN
+    // Lock threadContext list during etw rundown
+    AutoCriticalSection autoThreadContextCs(ThreadContext::GetCriticalSection());
+
+    ThreadContext * threadContext = ThreadContext::GetThreadContextList();
+
+    FILE * perfMapFile;
+
+    {
+        const size_t PERFMAP_FILENAME_MAX_LENGTH = 30;
+        char perfMapFilename[PERFMAP_FILENAME_MAX_LENGTH];
+        pid_t processId = getpid();
+        snprintf(perfMapFilename, PERFMAP_FILENAME_MAX_LENGTH, "/tmp/perf-%d.map", processId);
+
+        perfMapFile = fopen(perfMapFilename, "w");
+        if (perfMapFile == NULL) {
+            return;
+        }
+    }
+
+    while(threadContext != nullptr)
+    {
+        // Take etw rundown lock on this thread context
+        AutoCriticalSection autoEtwRundownCs(threadContext->GetFunctionBodyLock());
+
+        ScriptContext* scriptContext = threadContext->GetScriptContextList();
+        while(scriptContext != NULL)
+        {
+            if(scriptContext->IsClosed())
+            {
+                scriptContext = scriptContext->next;
+                continue;
+            }
+
+            scriptContext->MapFunction([=] (FunctionBody* body)
+            {
+#if DYNAMIC_INTERPRETER_THUNK
+                if(body->HasInterpreterThunkGenerated())
+                {
+                    const char16* functionName = body->GetExternalDisplayName();
+                    fwprintf(perfMapFile, _u("%llX %llX %s(Interpreted)\n"),
+                        body->GetDynamicInterpreterEntryPoint(),
+                        body->GetDynamicInterpreterThunkSize(),
+                        functionName);
+                }
+#endif
+
+#if ENABLE_NATIVE_CODEGEN
+                body->MapEntryPoints([&](int index, FunctionEntryPointInfo * entryPoint)
+                {
+                    if(entryPoint->IsCodeGenDone())
+                    {
+                        const ExecutionMode jitMode = entryPoint->GetJitMode();
+                        if (jitMode == ExecutionMode::SimpleJit)
+                        {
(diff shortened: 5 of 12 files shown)


lib/Runtime/PlatformAgnostic/Platform/Linux/PerfTrace.cpp
```diff
@@ -0,0 +1,152 @@
...
+
+    {
+      const size_t PERFMAP_FILENAME_MAX_LENGTH = 30;
+      char perfMapFilename[PERFMAP_FILENAME_MAX_LENGTH];
+      pid_t processId = getpid();
+      snprintf(perfMapFilename, PERFMAP_FILENAME_MAX_LENGTH, "/tmp/perf-%d.map", processId);
+      
+      perfMapFile = fopen(perfMapFilename, "w");
+      if (perfMapFile == NULL) {
+	return;
```
kunalspathak: >return [](start = 1, length = 6)

nit: indentation.

lib/Runtime/PlatformAgnostic/Platform/Linux/PerfTrace.cpp
```diff
@@ -0,0 +1,152 @@
...
+            scriptContext->MapFunction([=] (FunctionBody* body)
+            {
+#if DYNAMIC_INTERPRETER_THUNK
+                if(body->HasInterpreterThunkGenerated())
+                {
+		    const char16* functionName = body->GetExternalDisplayName();
+		    fwprintf(perfMapFile, _u("%llX %llX %s\n"),
+			    body->GetDynamicInterpreterEntryPoint(),
+			    body->GetDynamicInterpreterThunkSize(),
+			    functionName);
```
kunalspathak: indentation.

lib/Runtime/PlatformAgnostic/Platform/Linux/PerfTrace.cpp
```diff
@@ -0,0 +1,152 @@
...
+
+#if ENABLE_NATIVE_CODEGEN
+                body->MapEntryPoints([&](int index, FunctionEntryPointInfo * entryPoint)
+                {
+                    if(entryPoint->IsCodeGenDone())
+                    {
+		      const ExecutionMode jitMode = entryPoint->GetJitMode();
+		      if (jitMode == ExecutionMode::SimpleJit)
+		      {
+			fwprintf(perfMapFile, _u("%llX %llX %s(Simple)\n"),
```
kunalspathak: >Simple [](start = 42, length = 6)

May be call this `(SimpleJIT)` and the one below as `(FullJIT)`?

lib/Runtime/PlatformAgnostic/Platform/Linux/PerfTrace.cpp
```diff
@@ -0,0 +1,152 @@
...
+                continue;
+            }
+
+            scriptContext->MapFunction([=] (FunctionBody* body)
+            {
+#if DYNAMIC_INTERPRETER_THUNK
+                if(body->HasInterpreterThunkGenerated())
+                {
+		    const char16* functionName = body->GetExternalDisplayName();
+		    fwprintf(perfMapFile, _u("%llX %llX %s\n"),
```
kunalspathak: >\n [](start = 44, length = 2)

can we add `(Interpreter)` here, just to be clear?

lib/Common/PlatformAgnostic/DateTime.h
```diff
@@ -7,6 +7,7 @@
 #define RUNTIME_PLATFORM_AGNOSTIC_COMMON_DATETIME
 
 #include "PlatformAgnostic/DateTimeInternal.h"
+#include "CommonPal.h"
```
kunalspathak: why is this change? can you remove it and see if it builds?

MSLaguana: Looks like it's fine to be removed. I must have made a mistake earlier.

lib/Runtime/Base/ThreadContext.cpp
```diff
@@ -1735,6 +1735,11 @@ ThreadContext::ProbeStack(size_t size, Js::ScriptContext *scriptContext, PVOID r
 {
     this->ProbeStackNoDispose(size, scriptContext, returnAddress);
 
+    if (PlatformAgnostic::PerfTrace::mapsRequested)
+    {
+        PlatformAgnostic::PerfTrace::WritePerfMap();
```
digitalinfinity: On Windows since the implementation is empty, can you look at the code generated for ProbeStack and confirm that this if-check is optimized away?

lib/Jsrt/JsrtHelper.cpp
```diff
@@ -149,6 +149,8 @@ void JsrtCallbackState::ObjectBeforeCallectCallbackWrapper(JsObjectBeforeCollect
     #ifdef VTUNE_PROFILING
         VTuneChakraProfile::Register();
     #endif
+        PlatformAgnostic::PerfTrace::Register();
```
digitalinfinity: I would just move the #ifdef that's in Register to here instead

lib/Runtime/PlatformAgnostic/Platform/Linux/PerfTrace.cpp
```diff
@@ -0,0 +1,150 @@
...
+  
+//
+// Registers a signal handler for SIGUSR2 
+//
+void PerfTrace::Register()
+{
+    struct sigaction newAction = {0};
+    newAction.sa_flags = SA_RESTART;
+    newAction.sa_handler = handle_signal;
+    sigemptyset(&newAction.sa_mask);
```
obastemur: Running this on xplat no matter what may not be a good idea for a `host` embedding ChakraCore. We better enable this stuff with a compile time switch?

lib/Runtime/PlatformAgnostic/PerfTrace.h
```diff
@@ -0,0 +1,31 @@
...
+// Encapsulates all perf event logging and registration related to symbol decoding.
+//
+class PerfTrace
+{
+public:
+    static void Register();
+
+    static void WritePerfMap();
+
+    static volatile bool mapsRequested;
```
obastemur: Why volatile?

lib/Runtime/PlatformAgnostic/Platform/Linux/PerfTrace.cpp
```diff
@@ -0,0 +1,150 @@
...
+
+    ThreadContext * threadContext = ThreadContext::GetThreadContextList();
+
+    FILE * perfMapFile;
+
+    {
+        const size_t PERFMAP_FILENAME_MAX_LENGTH = 30;
+        char perfMapFilename[PERFMAP_FILENAME_MAX_LENGTH];
+        pid_t processId = getpid();
+        snprintf(perfMapFilename, PERFMAP_FILENAME_MAX_LENGTH, "/tmp/perf-%d.map", processId);
```
obastemur: Since we write to `/tmp` can we use more distinctive naming for the file name? i.e. `cc-perf-%d.map` ?

lib/Runtime/PlatformAgnostic/Platform/Linux/PerfTrace.cpp
```diff
@@ -0,0 +1,150 @@
...
+    {
+        AssertMsg(errno, "PerfTrace::Register: sigaction() call failed\n");
+    }
+}
+
+void  PerfTrace::WritePerfMap()
+{
+#if ENABLE_NATIVE_CODEGEN
+    // Lock threadContext list during etw rundown
+    AutoCriticalSection autoThreadContextCs(ThreadContext::GetCriticalSection());
```
obastemur: Shall we halt the runtime prior to all these or it is unnecessary ?

lib/Runtime/PlatformAgnostic/PerfTrace.h
MSLaguana: This is volatile so that the main program loop won't optimize out reads on the variable which will be written to asynchronously in a signal handler.

lib/Runtime/PlatformAgnostic/Platform/Linux/PerfTrace.cpp
MSLaguana: We do; the `PERFMAP_TRACE_ENABLED` variable is used where `PerfTrace::Register` is invoked, as well as around where `PerfTrace::mapsRequested` is checked before invoking `PerfTrace::WritePerfMap`.

lib/Runtime/PlatformAgnostic/Platform/Linux/PerfTrace.cpp
MSLaguana: This code is based on the ETW tracing code so I believe this is ok.

lib/Runtime/PlatformAgnostic/Platform/Linux/PerfTrace.cpp
MSLaguana: I've gone with `/tmp/perf-<pid>.map` because that is the format that `perf` expects. If we gave it a different name, you would need to rename it before it worked with other tools.

lib/Runtime/PlatformAgnostic/Platform/Linux/PerfTrace.cpp
MSLaguana: I should note, this dumping process is "best-effort": It's possible (and I've seen it happen) that you may have a perf trace that refers to memory addresses that are not covered in the final `perf-<pid>.map` because at the time we dump the map some sections of code have been collected.

lib/Runtime/PlatformAgnostic/Platform/Linux/PerfTrace.cpp
obastemur: Cool, but how about registering signal handler by default? This is where I'm concerned the most

lib/Runtime/PlatformAgnostic/PerfTrace.h
obastemur: volatile only applies to compile time. If we write this variable from multiple threads, should not be atomic instead?

lib/Runtime/PlatformAgnostic/Platform/CMakeLists.txt
```diff
@@ -7,6 +7,7 @@ set(PL_SOURCE_FILES
   Linux/UnicodeText.ICU.cpp
   Linux/NumbersUtility.cpp
   Linux/Thread.cpp
+  Linux/PerfTrace.cpp
```
obastemur: Perf is Linux specific. We better move this into Linux only instead?

lib/Runtime/PlatformAgnostic/PerfTrace.h
MSLaguana: It's just a bool, shouldn't that be guaranteed to be atomic? Are you suggesting that I should use `std::atomic<bool>` instead?

lib/Runtime/PlatformAgnostic/Platform/CMakeLists.txt
MSLaguana: The `perf` tool is, but the idea of a mapping from memory addresses to functions is not, and I don't believe any of the mechanisms I have used are not compatible with OSX.

lib/Runtime/PlatformAgnostic/Platform/Linux/PerfTrace.cpp
MSLaguana: If `PERFMAP_TRACE_ENABLED` is false, then we will not call `Register` and so no signal handler will be registered.

lib/Runtime/PlatformAgnostic/PerfTrace.h
obastemur: There is no such guarantee if system has multiple cores

lib/Runtime/PlatformAgnostic/Platform/CMakeLists.txt
obastemur: well, not sure what do you mean by OSX compatible. I just don't know how this output will be useful there.

lib/Runtime/PlatformAgnostic/Platform/Linux/PerfTrace.cpp
obastemur: Well from what I see above, that's defined as long as it's not WIN32 system.

MSLaguana: Yes; by default I thought we would want to include this (so that people can get useful performance analysis) but if an embedder does not want the signal handler, they can recompile with it disabled?

lib/Runtime/PlatformAgnostic/Platform/CMakeLists.txt
MSLaguana: If you use a tool like dtrace on OSX, is there a pre-existing way to determine that the function at `0x7deadbeefcafe` is actually the javascript function `foo`? The `perf-<pid>.map` file provides that information because I didn't see anything that exposed it previously.

lib/Runtime/PlatformAgnostic/Platform/Linux/PerfTrace.cpp
obastemur: IMHO, we shouldn't register signal handler by default. We may surprise host application or even worse cause runtime instabilities.

lib/Runtime/PlatformAgnostic/Platform/CMakeLists.txt
obastemur: Not sure how dtrace could fit in the scenario but on OSX `Instruments` application does the job pretty well. Thus; for `instruments` not sure how this output would work. Actually it is already easy to identify the JS calls there. Only missing (default) functionality is naming the JS functions. We could left a todo for osx here though

MSLaguana:
I've changed it so that now by default any build will have all the perf map code disabled, but it can be re-enabled with appropriate build flags, e.g. `./build.sh --extra-defines=PERFMAP_TRACE_ENABLED=1`

Penguinwizzard:
If you're adding a build flag like this, you should probably also add some form of CI to ensure that it doesn't get broken... maybe add the flag to one of the Jenkins jobs?

MSLaguana:
@dotnet-bot test ci please

MSLaguana:
CI test run at https://github.com/Microsoft/ChakraCore/pull/3215

MSLaguana:
Manually triggered a daily run at https://ci.dot.net/job/Microsoft_ChakraCore/job/master/job/GenPRTest/job/shared_daily_ubuntu_linux_release/1/ since the ci-testing functionality doesn't include daily jobs.

obastemur:
Just to clarify couple of things. (there might be a missed review above)

@MSLaguana 

- `volatile` doesn't guarantee atomic ops. See https://github.com/Microsoft/ChakraCore/pull/3193#pullrequestreview-45401524 

- This PR introduces a Linux specific feature. We better don't make it part of OSX builds. Actually, IMHO, not sure why we have an interface for Windows too ? https://github.com/Microsoft/ChakraCore/pull/3193#pullrequestreview-45493842

PR #3195, opened 2017-06-20
For external types reject deferred initializer calls if disableimplicit call is set

agarwal-sandeep:
If JIT has disabled implicit call, we should not call the DeferredTypeHandler
initializer as it can potentially go out of engine and change stuff.


agarwal-sandeep:
@curtisman please take a look

lib/Runtime/Types/DeferredTypeHandler.h
```diff
@@ -580,8 +621,9 @@ namespace Js
     {
         if (!EnsureObjectReady(instance, DeferredInitializeMode_Default))
         {
-            return FALSE;
+            return TRUE;
```
curtisman: You need to check all the caller to make sure it can handle this property.
For this particular one, I only find one call site in NativeCodeGenerator::GatherCodeGenData.
It should never be under disabled implicit call. But  If you return TRUE here while it does, it will start using uninitialized value in attributes.
This case probably should be a fail fast.

curtisman: Please review other API that you changed the return value as well to make sure we are doing the right thing.

curtisman:
:shipit:

agarwal-sandeep:
Thanks @curtisman

PR #3198, opened 2017-06-20
Fix bookkeeping of unregisteredInlineCacheCount

kunalspathak:
We were removing inlineCache from invalidation list but were not
updating the count at `ThreadContext` level. In between we were calling
compacting the invalidation list and find that there are more null
entries in the invalidation list than we thought (as per `threadContext->unregisteredInlineCacheCount`).

Fix by calling compact only after we update the variable at threadContext.


kunalspathak:
Ref: https://microsoft.visualstudio.com//defaultcollection/OS/_workItems?_a=edit&id=9779316&triage=true

digitalinfinity:
LGTM but someone who understands this code better (@rajatd perhaps?) should take a look too

rajatd:
Looks good.

kunalspathak:
Thanks @digitalinfinity  and @rajatd for the review.

PR #3199, opened 2017-06-21
jsrt: don't check runtime on each API call for non-browser

obastemur:
JSRT doesn't know how host is using the API. Verifying runtime state on each call redundant and time taking.

Simple JsHasException call returns all the useful stuff about runtime when it is needed.

### DIFF
--- a/lib/Jsrt/Jsrt.cpp
+++ b/lib/Jsrt/Jsrt.cpp
@@ -43,26 +43,28 @@ public:
 typedef utf8::NarrowWideConverter<CodexHeapAllocatorInterface, LPCSTR, LPWSTR> NarrowToWideChakraHeap;
 typedef utf8::NarrowWideConverter<CodexHeapAllocatorInterface, LPCWSTR, LPSTR> WideToNarrowChakraHeap;
 
-JsErrorCode CheckContext(JsrtContext *currentContext, bool verifyRuntimeState, bool allowInObjectBeforeCollectCallback)
+JsErrorCode CheckContext(JsrtContext *currentContext, bool verifyRuntimeState,
+    bool allowInObjectBeforeCollectCallback)
 {
     if (currentContext == nullptr)
     {
         return JsErrorNoCurrentContext;
     }
 
-    Js::ScriptContext *scriptContext = currentContext->GetScriptContext();
-    Assert(scriptContext != nullptr);
-    Recycler *recycler = scriptContext->GetRecycler();
-    ThreadContext *threadContext = scriptContext->GetThreadContext();
-
     // We don't need parameter check if it's checked in previous wrapper.
     if (verifyRuntimeState)
     {
+        Js::ScriptContext *scriptContext = currentContext->GetScriptContext();
+        Assert(scriptContext != nullptr);
+        Recycler *recycler = scriptContext->GetRecycler();
+        ThreadContext *threadContext = scriptContext->GetThreadContext();
+
         if (recycler && recycler->IsHeapEnumInProgress())
         {
             return JsErrorHeapEnumInProgress;
         }
-        else if (!allowInObjectBeforeCollectCallback && recycler && recycler->IsInObjectBeforeCollectCallback())
+        else if (!allowInObjectBeforeCollectCallback &&
+            recycler && recycler->IsInObjectBeforeCollectCallback())
         {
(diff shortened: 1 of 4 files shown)


obastemur:
@curtisman @digitalinfinity ping.

obastemur:
It could be nice if this one lands in 1.7. Anything left here to do?

lib/Jsrt/Jsrt.cpp
```diff
@@ -43,26 +43,28 @@ struct CodexHeapAllocatorInterface
...
+        Recycler *recycler = scriptContext->GetRecycler();
+        ThreadContext *threadContext = scriptContext->GetThreadContext();
+
         if (recycler && recycler->IsHeapEnumInProgress())
         {
             return JsErrorHeapEnumInProgress;
         }
-        else if (!allowInObjectBeforeCollectCallback && recycler && recycler->IsInObjectBeforeCollectCallback())
+        else if (!allowInObjectBeforeCollectCallback &&
+            recycler && recycler->IsInObjectBeforeCollectCallback())
```
kunalspathak: >IsInObjectBeforeCollectCallback [](start = 34, length = 31)

we don't get this information from `JsHasException`. How can host determine that?

obastemur: Not sure I understood. Could you elaborate?

kunalspathak: You said that host can use `JsHasException` to check if `chakracore` is in bad state and take decision based on that. We get most of the bad state information from `JsHasException` except if it is in `JsErrorInObjectBeforeCollectCallback` or not. Does that make sense?

obastemur: Once that happens, what application is expected to do?

obastemur: If the answer is not much! that's what unsafe mode is.

Or; if there is a particular JSRT interface that we should not unsafe?

kunalspathak: > If the answer is not much! that's what unsafe mode is. 

That is true for any other exceptions that `JsHasException` would return after your change. My concern is we won't have a way to know if Chakra is in a state `JsErrorInObjectBeforeCollectCallback` which was possible before your change.

kunalspathak: @liminzhu  - in case he has any opinion.

lib/Jsrt/JsrtInternal.h
```diff
@@ -14,6 +14,17 @@ typedef TTD::TTDJsRTActionResultAutoRecorder TTDRecorder;
...
 
+#ifndef NTBUILD // non-browser
+#define JSRT_UNSAFE_MODE
+#endif
+
+// JSRT Unsafe mode leaves runtime health-state responsibility to the host
+#ifdef JSRT_UNSAFE_MODE
+#define JSRT_MAYBE_TRUE false
+#else
+#define JSRT_MAYBE_TRUE true
```
dilijev: I think this name is confusing to say the least. Since it will be used [only] as the `verifyRuntimeState` parameter, I think should call it `JSRT_VERIFY_RUNTIME_STATE` or similar. This suggests the meaning of the value, rather than what the value might be. I can't imagine a "maybe-true" value controlled by boolean compile-time flag for `UNSAFE` being used in more a general sense.

obastemur: I liked the `JSRT_VERIFY_RUNTIME_STATE` name. Thanks

obastemur:
@dilijev This was discussed offline. Once the PR is final, there will be an addition to docs.

dilijev:
What about questions re: keeping asserts in debug builds, why is this good enough for non-browser, etc?

obastemur:
@kunalspathak please review.

kunalspathak:
:shipit:

obastemur:
@dilijev It's not about being good enough. This PR may require some changes to host app. That's why it is limited to ChakraCore.

obastemur:
@kfarnung @dilijev Thanks for the review

dilijev:
Can a future PR include Debug asserts in case of violation of the contract? This may hurt performance in Debug builds but would help a host developer ensure correctness.

obastemur:
No need for that. If runtime is broken, we already have asserts in place

PR #3200, opened 2017-06-21
runtests: lldb debugger support

obastemur:
An helpful addition to runtests.py to get crash logs and call stack from lldb. This is especially useful when crash only occurs under python child process.


obastemur:
@Penguinwizzard I'm not sure why we should maintain two debugger outputs ? This just prints call stack that normally very hard to trigger / print.

Thanks for the review!

dilijev:
Looks fine but I don't know much about LLDB. What will the output look like?

obastemur:
@Penguinwizzard @dilijev Thanks for the review.

@dilijev The output is similar to lldb `bt` output.

PR #3202, opened 2017-06-22
nativetests: fix thread racing in JsDisableRuntimeExecution tests

jianchun:
The tests were calling REQUIRE in multiple threads. This can corrupt CATCH
internal data. Fixed by postponing the REQUIRE call to main thread.


akroshg:
LGTM

jianchun:
Thanks @akroshg @obastemur for review

PR #3205, opened 2017-06-22
DRT: Test cleanup.

atulkatti:


Cellule:
is there an issue tracking these ?

MikeHolman:
Why are you excluding these? What configuration is failing?

atulkatti:
These are currently failing in the nightly tests. I am temporarily disabling them only from DRTs configuration to unblock onboarding our DRTs. There should be bugs opened by nightly runs for these failures. Once the bugs are addressed the tests should be enabled in DRTs once they become table. Please refer to the nightly test runs to see which configurations are failing.

PR #3210, opened 2017-06-22
fix an invalid DynamicObject* cast

jianchun:
Scope objects are not always "DynamicObject"s (e.g. may be of type
"WithScopeObject"). Use RecyclerObject instead.


rajatd:
similar issue in `PatchGetPropertyScoped` as well

jianchun:
> similar issue in PatchGetPropertyScoped as well

You are right, fixed in update

jianchun:
Thanks @Penguinwizzard @rajatd

Issue #3035, opened 2017-05-24
Memory leak in Array.splice()


ChristianStroyer:
In some cases an object removed from an array with `Array.splice` is still allocated in memory, even though it is not seen in a heap dump. This results in IE(11.0.42) and Edge(38.14393) running out of memory in Windows in our web app. Chrome(58) does not have this issue.

Array.splice is a common way to remove entries in arrays. If it leaks I believe it is a huge issue for Edge.

I've created a small example. It creates an array with a large objects, and then removes the object with `splice(0,1)`. Running this for a few seconds takes both IE and Edge to a memory consumption of more than 1GB, which is never freed. An inspection shows only empty arrays.

Example at https://jsfiddle.net/aefxf0v9/
```
// create 80kb Object
function createLargeObject(){
	var temparray = [];
	for(var i=0;i<10000;i++){
		temparray[i]=i;
	}
  return { obj:temparray };
}

function leak() { 
	globalarray = [];
	for(var i=0;i<20000;i++) {
		globalarray[i]=[];
		var innerarray = globalarray[i];
		var largeObj = createLargeObject();
		innerarray[0] = largeObj ;
		//innerarray[0] = null;   // Adding this will fix the problem
		innerarray.splice( 0, 1 );  
	}
}
```

Setting the array entry to null before removing it with splice fixes the problem. I have made a shim for our project, but would very much like to see a solution.

ChristianStroyer:
Exploring a bit further it seems the problem gravitates around removing the final entry in an array. The following examples supports it:
```
// Leak
innerarray[0] = largeObj ;
innerarray.splice( 0, 1 );  
```
```
// No issue
innerarray[0] = 0 ;
innerarray[1] = largeObj ;
innerarray.splice( 1, 1 );  
```
```
// No issue
innerarray[0] = largeObj ;
innerarray[1] = 0 ;
innerarray.splice( 0, 1 ); 
```
```
// Leak
innerarray[0] = 0 ;
innerarray[1] = largeObj ;
innerarray.splice( 0, 2 );  
```

michielcrivits:
We are experiencing a memory leak in our web application as of IE v11.0.42. We have a strong feeling it's related to this bug. Any idea yet when this would be solved in IE 11?

Are there certain ways to avoid/lower the impact of the issue? IE settings, alternatives, ...

suwc:
[...]

PR #3211, opened 2017-06-23
Issue3035: Memory leak in Array.splice()

suwc:
Memory leak occurs when the entire inline segment is removed
from the array's linked list, because the inline segment is
not collected like other segments. Fix by clearing the elements
of the inline segment when it is replaced.

### DIFF
--- a/lib/Runtime/Library/JavascriptArray.cpp
+++ b/lib/Runtime/Library/JavascriptArray.cpp
@@ -529,6 +529,38 @@ namespace Js
                 VirtualTableInfo<JavascriptNativeFloatArray>::HasVirtualTable(aValue));
     }
 
+    bool JavascriptArray::IsInlineSegment(SparseArraySegmentBase *seg, JavascriptArray *pArr)
+    {
+        if (seg == nullptr)
+        {
+            return false;
+        }
+
+        SparseArraySegmentBase* inlineHeadSegment = nullptr;
+        if (JavascriptNativeArray::Is(pArr))
+        {
+            if (JavascriptNativeFloatArray::Is(pArr))
+            {
+                inlineHeadSegment = DetermineInlineHeadSegmentPointer<JavascriptNativeFloatArray, 0, true>((JavascriptNativeFloatArray*)pArr);
+            }
+            else
+            {
+                AssertOrFailFast(JavascriptNativeIntArray::Is(pArr));
+                inlineHeadSegment = DetermineInlineHeadSegmentPointer<JavascriptNativeIntArray, 0, true>((JavascriptNativeIntArray*)pArr);
+            }
+
+            Assert(inlineHeadSegment);
+            return (seg == inlineHeadSegment);
+        }
+
+        // This will result in false positives. It is used because DetermineInlineHeadSegmentPointer
+        // does not handle Arrays that change type e.g. from JavascriptNativeIntArray to JavascriptArray
+        // This conversion in particular is problematic because JavascriptNativeIntArray is larger than JavascriptArray
+        // so the returned head segment ptr never equals pArr->head. So we will default to using this and deal with
+        // false positives. It is better than always doing a hard copy.
+        return pArr->head != nullptr && HasInlineHeadSegment(pArr->head->length);
+    }
+
     DynamicObjectFlags JavascriptArray::GetFlags() const
     {
         return GetArrayFlags();
@@ -5213,28 +5245,7 @@ Case0:
     template <typename T>
     void JavascriptArray::CopyHeadIfInlinedHeadSegment(JavascriptArray *array, Recycler *recycler)
     {
-        SparseArraySegmentBase* inlineHeadSegment = nullptr;
-        bool hasInlineSegment = false;
-
-        if (JavascriptNativeArray::Is(array))
-        {
-            if (JavascriptNativeFloatArray::Is(array))
-            {
-                inlineHeadSegment = DetermineInlineHeadSegmentPointer<JavascriptNativeFloatArray, 0, true>((JavascriptNativeFloatArray*)array);
-            }
-            else if (JavascriptNativeIntArray::Is(array))
-            {
-                inlineHeadSegment = DetermineInlineHeadSegmentPointer<JavascriptNativeIntArray, 0, true>((JavascriptNativeIntArray*)array);
-            }
-            Assert(inlineHeadSegment);
-            hasInlineSegment = (array->head == (SparseArraySegment<T>*)inlineHeadSegment);
-        }
-        else
-        {
-            hasInlineSegment = HasInlineHeadSegment(array->head->length);
-        }
-
-        if (hasInlineSegment)
+        if (JavascriptArray::IsInlineSegment(array->head, array))
         {
             AnalysisAssert(array->head);
             SparseArraySegment<T>* newHeadSeg = array->ReallocNonLeafSegment((SparseArraySegment<T>*)PointerValue(array->head), array->head->next);
@@ -5378,8 +5389,8 @@ Case0:
             }
 
             // During the loop below we are going to reverse the segments list. The head segment will become the last segment.
-            // We have to verify that the current head segment is not the inilined segement, otherwise due to shuffling below, the inlined segment will no longer
-            // be the head and that can create issue down the line. Create new segment if it is an inilined segment.
+            // We have to verify that the current head segment is not the inlined segement, otherwise due to shuffling below, the inlined segment will no longer
+            // be the head and that can create issue down the line. Create new segment if it is an inlined segment.
             if (pArr->head && pArr->head->next)
             {
                 if (isIntArray)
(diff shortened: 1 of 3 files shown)


suwc:
Addresses https://github.com/Microsoft/ChakraCore/issues/3035

lib/Runtime/Library/JavascriptArray.cpp
```diff
@@ -529,6 +529,31 @@ namespace Js
...
+    bool JavascriptArray::IsInlineSegment(SparseArraySegmentBase *seg, JavascriptArray *pArr)
+    {
+        SparseArraySegmentBase* inlineHeadSegment = nullptr;
+        if (JavascriptNativeArray::Is(pArr))
+        {
+            if (JavascriptNativeFloatArray::Is(pArr))
+            {
+                inlineHeadSegment = DetermineInlineHeadSegmentPointer<JavascriptNativeFloatArray, 0, true>((JavascriptNativeFloatArray*)pArr);
+            }
+            else if (JavascriptNativeIntArray::Is(pArr))
```
agarwal-sandeep: >if (JavascriptNativeIntArray::Is(pArr) [](start = 17, length = 38)

This can be converted to an assert instead as JavascriptNativeArray::Is already verified that it is either a NativeFloat or NativeInt array

lib/Runtime/Library/JavascriptArray.cpp
```diff
@@ -2033,6 +2064,7 @@ namespace Js
         SparseArraySegmentBase *seg, *nextSeg, *prevSeg = nullptr;
         for (seg = fArray->head; seg; seg = nextSeg)
         {
+            bool isInlineSegment = JavascriptArray::IsInlineSegment(seg, fArray);
```
agarwal-sandeep: >isInlineSegment [](start = 17, length = 15)

Have isInlineSegment as false and move isInlineSegment = JavascriptArray::IsInlineSegment(seg, fArray); in if part where we are allocating newSeg that way when we call ClearElements we need to check only isInlineSegment

lib/Runtime/Library/JavascriptArray.cpp
```diff
@@ -5615,6 +5633,11 @@ namespace Js
                 head->length = offset + next->length;
                 head->CheckLengthvsSize();
                 pArr->head = head;
+
+                if (oldHead != head && isInlineSegment)
```
agarwal-sandeep: >if (oldHead != head && isInlineSegment) [](start = 16, length = 39)

I think if we reverse the && to have isInlineSegment first compiler can do a test for bool and then a pointer comparison which will be faster

lib/Runtime/Library/JavascriptArray.cpp
```diff
@@ -6500,7 +6523,13 @@ namespace Js
             uint32 newLength = head->length + countUndefined;
             if (newLength > head->size)
             {
+                SparseArraySegmentBase *oldHead = head;
+                bool isInlineSegment = JavascriptArray::IsInlineSegment(head, this);
                 head = SparseArraySegment<Var>::From(head)->GrowByMin(recycler, newLength - head->size);
```
agarwal-sandeep: >GrowByMin [](start = 60, length = 9)

Seems like this will always allocate a new segment so the check oldHead != head is not needed

lib/Runtime/Library/JavascriptArray.cpp
```diff
@@ -6908,33 +6943,7 @@ namespace Js
...
-        {
-            // This will result in false positives. It is used because DetermineInlineHeadSegmentPointer
-            // does not handle Arrays that change type e.g. from JavascriptNativeIntArray to JavascriptArray
-            // This conversion in particular is problematic because JavascriptNativeIntArray is larger than JavascriptArray
-            // so the returned head segment ptr never equals pArr->head. So we will default to using this and deal with
-            // false positives. It is better than always doing a hard copy.
-            hasInlineSegment = HasInlineHeadSegment(pArr->head->length);
-        }
-
+        bool hasInlineSegment = IsInlineSegment(startSeg, pArr);
```
agarwal-sandeep: >startSeg [](start = 48, length = 8)

Seems like we can have scenario where startSeg is nullptr (check below), after your change for those scenario we will have AV in FromVar during JavascriptNativeArray::Is

lib/Runtime/Library/JavascriptArray.cpp
```diff
@@ -7137,6 +7149,11 @@ namespace Js
                     {
                         segInsert = segInsert->GrowByMinMax(recycler, insertLen - spaceLeft, segInsert->next->left - segInsert->left - segInsert->size);
                     }
+
+                    if (oldSegInsert != segInsert && isInlineSegment)
```
agarwal-sandeep: >oldSegInsert != segInsert [](start = 24, length = 25)

Redundant GrowBy* will always allocate a new segment

lib/Runtime/Library/JavascriptArray.inl
```diff
@@ -129,6 +131,12 @@ namespace Js
         {
             segmentMap->SwapSegment(seg->left, seg, newSeg);
         }
+
+        if (isInlineSegment)
+        {
+            this->ClearElements(seg, 0);
+        }
```
agarwal-sandeep: Instead of clearing the elements can't we free it? next/previous is already adjusted so no one should be using seg. Will save recycler some work at same cost of clearing elements.

leirocks: I think the inline segment is a plus alloc, it's part of the array structure, you can't free only a part of memory of one allocation

lib/Runtime/Library/JavascriptArray.cpp
akroshg: This should be NativeIntArray. FailFast would be good.

---
In reply to: [123650724](https://github.com/Microsoft/ChakraCore/pull/3211#discussion_r123650724) [](ancestors = 123650724)

lib/Runtime/Library/JavascriptArray.cpp
```diff
@@ -529,6 +529,31 @@ namespace Js
                 VirtualTableInfo<JavascriptNativeFloatArray>::HasVirtualTable(aValue));
     }
 
+    bool JavascriptArray::IsInlineSegment(SparseArraySegmentBase *seg, JavascriptArray *pArr)
```
akroshg: >seg [](start = 66, length = 3)

return false when seg is null? could be a valid case.

lib/Runtime/Library/JavascriptArray.inl
akroshg: correct.

---
In reply to: [123792263](https://github.com/Microsoft/ChakraCore/pull/3211#discussion_r123792263) [](ancestors = 123792263)

lib/Runtime/Library/JavascriptArray.cpp
```diff
@@ -7579,6 +7605,10 @@ namespace Js
             pArr->head = SparseArraySegment<T>::From(pArr->head)->GrowByMinMax(recycler, unshiftElements, ((nextToHeadSeg->left + unshiftElements) - pArr->head->left - pArr->head->size));
         }
 
+        if (oldHead != pArr->head && isInlineSegment)
```
agarwal-sandeep: >oldHead != pArr->head [](start = 12, length = 21)

Redundant?

agarwal-sandeep:
>                         current = current->GrowByMin(recycler, growby);

Only in this case current is changed, we can prevent oldCurrent != current below by initializing isInlineSegment to false and moving IsInlineSegment here

---
Refers to: lib/Runtime/Library/JavascriptArray.inl:919 in 05b1641. [](commit_id = 05b16413a31da0fda7aeb6798a4fce97fe3a97aa, deletion_comment = False)

lib/Runtime/Library/JavascriptArray.cpp
suwc: Point taken and updated in other occurrences. Decided JavascriptNativeFloatArray::ConvertToVarArray() does not need to call clearElements() because its inlined head segment only contains native data. Same thing for JavascriptNativeIntArray::ConvertToVarArray().

---
In reply to: [123652599](https://github.com/Microsoft/ChakraCore/pull/3211#discussion_r123652599) [](ancestors = 123652599)

lib/Runtime/Library/JavascriptArray.cpp
suwc: Added nullptr checking in JavascriptArray::IsInlineSegment().

---
In reply to: [123654814](https://github.com/Microsoft/ChakraCore/pull/3211#discussion_r123654814) [](ancestors = 123654814)

suwc:
Any other questions or comments?

lib/Runtime/Library/JavascriptArray.inl
```diff
@@ -1357,6 +1397,8 @@ SECOND_PASS:
                     const bool currentWasFull = current->length == current->size;
 
                     Assert(itemIndex == current->left + current->size);
+                    SparseArraySegmentBase* oldSegment = current;
+                    bool isInlineSegment = JavascriptArray::IsInlineSegment(oldSegment, this);
                     current = SparseArraySegment<T>::CopySegment(recycler, (SparseArraySegment<T>*)current, next->left, next, next->left, next->length);
```
agarwal-sandeep: >SparseArraySegment<T>::CopySegment [](start = 30, length = 34)

CopySegment doesn't always allocate a new segment, can we assert before clearing that current != oldSegment

agarwal-sandeep:
Can we add the test case you have as a snippet in Exprgen? #Resolved

agarwal-sandeep:
Rest LGTM

suwc:
Sure

---
In reply to: [311135664](https://github.com/Microsoft/ChakraCore/pull/3211#issuecomment-311135664) [](ancestors = 311135664)

PR #3213, opened 2017-06-23
Duplicate let or const definition in the body of a split scope funciton should throw redeclaration error

aneeshdk:
For non-split scope functions we add an additional reference to the body for each param symbol. When a declaration occurs in the body we will find that sym already existing and throw redeclaration. But in case of split scope functions we can't add the additional reference, so we have to check explicitly for duplicates.


aneeshdk:
@pleath @akroshg @bterlson Can you please take a look?

aneeshdk:

test/es6/default.js
```diff
@@ -55,6 +55,23 @@ var tests = [
...
+      assert.throws(function () { eval("function foo({a, b = () => a}) { let b; };"); }, SyntaxError, "Duplicate let decalration in the body of a function with destructured param should throw redeclaration error", "Let/Const redeclaration");
+      assert.throws(function () { eval("function foo([a], b = () => a) { const b = 1; };"); }, SyntaxError, "Duplicate const decalration in the body of a function with destructured param should throw redeclaration error", "Let/Const redeclaration");
+      assert.throws(function () { eval("function foo([arguments, b = () => arguments]) { let arguments; };"); }, SyntaxError, "Duplicate let definition of arguments in a function with destructured params should throw redeclaration error", "Let/Const redeclaration");
+      assert.throws(function () { eval("function foo(arguments, {b = () => arguments}) { const arguments = 1; };"); }, SyntaxError, "Duplicate const definition of arguments in a function with destructured params should throw redeclaration error", "Let/Const redeclaration");
```
akroshg: I believe this case is covered already 
function foo({a}) { let a = 1; }
isn't it?

lib/Parser/Parse.cpp
```diff
@@ -915,10 +916,21 @@ Symbol* Parser::AddDeclForPid(ParseNodePtr pnode, IdentPtr pid, SymbolType symbo
...
             }
         }
+        else if (!pnodeFnc->sxFnc.IsBodyAndParamScopeMerged()
+            && scope->GetScopeType() == ScopeType_FunctionBody
+            && (pnode->nop == knopLetDecl || pnode->nop == knopConstDecl))
+        {
+            // In case of split scope function when we add a new let or const declaration to the body
+            // we have to check whether the param scope already has the same symbol defined.
+            PidRefStack *paramScopeRef = pid->GetPidRefForScopeId(pnodeFnc->sxFnc.pnodeScopes->sxBlock.blockId);
+            if (paramScopeRef && paramScopeRef->GetSym() && !paramScopeRef->GetSym()->GetIsArguments())
```
akroshg: >paramScopeRef->GetSym() [](start = 33, length = 23)

if the sym is null then why GetPidRefForScopeId will return paramScopeRef at all?

akroshg: >paramScopeRef->GetSym()->GetIsArguments() [](start = 61, length = 41)

When I ask for current pid reference then why will it give the arguments.

akroshg:
>             PidRefStack *pidRefOld = pid->GetPidRefForScopeId(1);

I looked at the usage GetPidRefForScopeId and I found that at caller side we are checking both pidRefOld && pidRefOld->GetSym(). why? 
Why not change the function to return valid sym if it found one?

---
Refers to: lib/Parser/Parse.cpp:913 in 0387e9c. [](commit_id = 0387e9c86f8cf6bdb117b887f009e516ac98a161, deletion_comment = False)

lib/Parser/Parse.cpp
aneeshdk: on the ref stack if the node is just a reference to the pid, not a declaration, then sym will be null but the refstack will have the scopeid and function id. We later bind the right symbol to the ref stack. If it is a declaration then sym won't be null.

akroshg:
:shipit:

Issue #2919, opened 2017-05-05
Implement Intl using ICU (enables Intl on xplat) and provide a compilation option


dilijev:
Windows Globalization does not work on xplat targets, so we will re-implement Intl using ICU to allow Intl to work on xplat. Additionally, a compilation option will make this implementation available as an option on Windows as well.

PR #3214, opened 2017-06-23
Partial implementation of Intl with ICU for Windows and xplat

dilijev:
Provides a build option for Windows and xplat builds to use the ICU implementation of Intl. This work will enable using Intl on xplat and Node-ChakraCore.

See #2919

* Implemented operations needed for `CanonicalizeLocaleList` (`Intl.getCanonicalLocales`)
* Optionally build using ICU for Intl operations (see below for build instructions)
  * With this feature compiled out (default), builds and passes tests on ChakraFull, ChakraCore (Windows), ChakraCore (xplat), except for Intl tests on xplat which don't apply since there is no Intl object.
  * With this feature compiled in, builds and passes tests (except Intl tests) on ChakraCore (Windows, xplat).
    * Feature is still a work-in-progress and more work needs to be done to pass Intl tests.
    * Note: this feature does not yet work in ChakraFull.
  * With hybrid mode (WinGlob fallback for things not yet implemented in ICU), passes all Intl tests.
  * Build with ICU requires that user has ICU libraries and headers on disk (optionally, allows user to specify paths to them).
    * Currently, only supports statically linking the ICU.
    * Use `"/p:IcuLibDir=..."` and `"/p:IcuIncludeDir=..."` to specify paths.
    * See **Chakra.Build.Paths.props** for specifics.

I invite reviewers to build on Windows or Linux with the INTL_ICU feature enabled (see below) and try out the following sample code:

```js
print(42);
print(Intl.getCanonicalLocales('en-us')); // en-US
print(Intl.getCanonicalLocales(['de-de'])); // de-DE
print(Intl.getCanonicalLocales(['en-us', 'de-de', 'ja-jp', 'zh-hans-cn'])); // en-US,de-DE,ja-JP,zh-Hans-CN
```

(Deviating too far from this example will likely cause errors or incorrect behavior due to the unimplemented parts.)

## Build on Windows

See https://github.com/Microsoft/ChakraCore/wiki/Building-ChakraCore#windows (with option `"/p:IntlICU=true"`)

```bat
set IntlICU=true
REM (optional) set IcuLibDir=... (default location ~= .\deps\static\icu-small\%Configuration%)
set IcuLibDir=...
REM (optional) set IcuIncludeDir=... (default location = .\deps\icu-small\source)
set IcuIncludeDir=...

msbuild /m /p:Platform=... /p:Configuration=... Build\Chakra.Core.sln
```

OR inline:

```bat
msbuild /m /p:Platform=... /p:Configuration=... Build\Chakra.Core.sln "/p:IntlICU=true"
msbuild /m /p:Platform=... /p:Configuration=... Build\Chakra.Core.sln "/p:IntlICU=true" "/p:IcuLibDir=..." "/p:IcuIncludeDir"
```

## Build on Linux

See https://github.com/Microsoft/ChakraCore/wiki/Building-ChakraCore#linux (with option `--with-intl`).

```sh
./build.sh -n --debug --static --with-intl
```

## Portions previously reviewed:

* Intl.js: Language-Tag regexes implementing RFC 5646 #section-2.1 (numerous commits)
* Implement Intl using ICU (first pass). (f3449ae0924990c2dd2bc41280982707257ce114)
  * Original message follows:

```
    Implement Intl using ICU (first pass).

    * Implemented operations needed for CanonicalizeLocaleList.
    * Optionally build using ICU for Intl operations (build with "/p:IntlICU=true".)
    * Build with ICU requires that user has ICU .libs and headers on disk and specifies a path to them.
      * Use "/p:IcuLibDir=..." and "/p:IcuIncludeDir=..." to specify. See Chakra.Build.Paths.props for specifics.
```


dilijev:
Whoops -- forgot to test the clang build without `--intl-icu` -- looking into the failures now.

CMakeLists.txt
```diff
@@ -387,6 +392,13 @@ else()
     set(BuildJIT 1)
 endif()
 
+if(INTL_ICU_SH)
+    unset(INTL_ICU_SH CACHE)    # don't cache
+    unset(IntlICU CACHE)        # also clear it just in case
```
obastemur: This is not needed.

dilijev: I'm not that familiar with CMake. I welcome feedback for how to do this more effectively.

build.sh
```diff
@@ -225,6 +225,10 @@ while [[ $# -gt 0 ]]; do
         NO_JIT="-DNO_JIT_SH=1"
         ;;
 
+    --intl-icu)
```
obastemur: on the top of this file we list all the available args. This should also documented there. ++ it could be nice to have this param as `--with-intl` .. so, if user defines `--icu=...` and then `--with-intl`.. it might look nicer?

dilijev: Yes, that makes sense. Since on Linux the presence of Intl will imply an ICU implementation, `--intl-icu` is redundant.

lib/Runtime/Base/WindowsGlobalizationAdapter.cpp
```diff
@@ -694,7 +699,153 @@ if (this->object) \
...
+    bool IcuIntlAdapter::ResolveLocaleBestFit(_In_ ScriptContext *scriptContext, _In_z_ const char16 *locale, _Out_ char16 *resolved)
+    {
+        // Note: the "best fit" matcher is implementation-defined, so it is okay to return the same result as ResolveLocaleLookup.
+        // TODO (doilij): implement a better "best fit" matcher
+        return ResolveLocaleLookup(scriptContext, locale, resolved);
+    }
+
+    int IcuIntlAdapter::GetUserDefaultLocaleName(_Out_ LPWSTR lpLocaleName, _In_ int cchLocaleName)
+    {
+        // FIXME (doilij): implement GetUserDefaultLocaleName
```
obastemur: can you please tag it as XPLAT-TODO ? or we already have the interface available for all the platforms?

pal/src/misc/strutil.cpp
```diff
@@ -95,3 +95,63 @@ CPalString::FreeBuffer()
     _ASSERTE(NULL != m_pwsz);
     InternalFree(const_cast<WCHAR*>(m_pwsz));
 }
+
+/* TODO (doilij) cleanup
```
obastemur: cleanup before merge?

CMakeLists.txt
obastemur: just remove this line

build.sh
obastemur: well, I would keep `--with-intl` anyways. Not everyone needs that and it consumes more memory + bigger binary size.

CMakeLists.txt
dilijev: Done, thanks.

build.sh
dilijev: Yes I'm changing the flag to `--with-intl` (definitely not making it the default at this time, maybe never will, or in any case it will be optional) and I'm keeping the macro named `INTL_ICU` for consistency with Windows and so I don't have to churn that at this time.

lib/Runtime/Base/WindowsGlobalizationAdapter.cpp
```diff
@@ -694,7 +699,153 @@ if (this->object) \
...
+        // REVIEW (doilij): should we / do we need to zero these stack-allocated arrays?
+        char icuLocaleId[ULOC_FULLNAME_CAPACITY] = { 0 };
+        char icuLangTag[ULOC_FULLNAME_CAPACITY] = { 0 };
+        int32_t parsedLength = 0;
+        int32_t forLangTagResultLength = 0;
+        int32_t toLangTagResultLength = 0;
+
+        // Convert input language tag to a locale ID for use in uloc_toLanguageTag API.
+        forLangTagResultLength = uloc_forLanguageTag(reinterpret_cast<const char *>(inputLangTagUtf8), icuLocaleId, ULOC_FULLNAME_CAPACITY, &parsedLength, &error);
+        success = forLangTagResultLength && parsedLength && U_SUCCESS(error) && ((size_t)parsedLength == inputLangTagUtf8SizeActual);
```
kfarnung: nit:  Shorten these long lines

lib/Runtime/Library/IntlEngineInterfaceExtensionObject.cpp
```diff
@@ -610,34 +693,51 @@ namespace Js
...
+            IcuIntlAdapter::GetUserDefaultLocaleName(defaultLocale, _countof(defaultLocale)) == 0
+#endif
+            || defaultLocale[0] == '\0')
         {
+#if defined(INTL_WINGLOB)
             return scriptContext->GetLibrary()->GetUndefined();
+#else
+            Output::Print(_u("EntryIntl_GetDefaultLocale > IcuIntlAdapter::GetUserDefaultLocaleName returned 0: returning null to fallback to JS\n"));
+
+            return scriptContext->GetLibrary()->GetNull();
```
MSLaguana: Why `GetNull` rather than `GetUndefined` like in the `INTL_WINGLOB` case?

MSLaguana: It looks like there are a lot of cases in this file where for the non-`WINGLOB` case you return null on errors, while the Winglob code returns Undefined instead. Is that deliberate?

Build/Chakra.Build.Paths.props
```diff
@@ -8,6 +8,12 @@
     <BuildConfigPropsPath>$(MSBuildThisFileDirectory)</BuildConfigPropsPath>
 
     <ChakraBuildPathImported>true</ChakraBuildPathImported>
-
+  </PropertyGroup>
+  <PropertyGroup Label="ICU">
```
kunalspathak: >"> [](start = 27, length = 2)

should this be under IntlICU check?

lib/Runtime/Library/Chakra.Runtime.Library.vcxproj
```diff
@@ -34,6 +34,7 @@
         $(MSBuildThisFileDirectory)..\..\Runtime\ByteCode;
         $(MSBuildThisFileDirectory)..\..\JITClient;
         $(ChakraJITIDLIntDir);
+        $(IcuIncludeDir)\common; <!-- // TODO cleanup: Is this needed here in Runtime.Library? -->
```
kunalspathak: > $(IcuIncludeDir)\common;  [](start = 7, length = 26)

can you convert it to something like `$(IcuIncludeDirCommon)` instead?

lib/Common/CommonDefines.h
```diff
@@ -98,11 +98,20 @@
...
 #define ENABLE_INTL_OBJECT                          // Intl support
+//#define INTL_ICU 1 // FIXME (doilij) remove this (need to #define this in the source code for the IDE to know about it, but the actual def will come from the project file)
+#ifdef INTL_ICU
+//#define INTL_ICU_ALLOW_HYBRID 1 // FIXME (doilij) this macro is only needed as long as INTL_ICU is in-development.
 #endif
+#if !defined(INTL_ICU) || (defined(INTL_ICU) && defined(INTL_ICU_ALLOW_HYBRID))
+// FIXME (doilij) INTL_WINGLOB only defined this way (for convenience) as long as INTL_ICU_ALLOW_HYBRID is defined, otherwise it should just be !defined(INTL_ICU)
+#define INTL_WINGLOB 1
+#endif
+// #endif // FIXME (doilij) remove or uncomment
```
kunalspathak: nit: I believe this will be changing soon, so skipped reviewing this portion.

kunalspathak:
>     <ClInclude Include="Scan.h" />

You might want to skip include `windows.globalization.h` for Intl_ICU

---
Refers to: lib/Parser/Chakra.Parser.vcxproj:119 in bf1b75b. [](commit_id = bf1b75bc6ff22eb145664b7ca8dd2db8dcfd8061, deletion_comment = False)

lib/Runtime/Base/WindowsGlobalizationAdapter.h
```diff
@@ -142,6 +156,24 @@ namespace Js
         static HRESULT VerifyResult(HSTRING * result, HRESULT errCode);
 #endif
     };
-}
-#endif
+#endif // INTL_WINGLOB
+
+#ifdef INTL_ICU
```
kunalspathak: >INTL_ICU [](start = 7, length = 8)

Eventually, is there a plan to have `IntlAdapter` class from which we inherit `WindowsGlobalizationAdapter` and `ICUAdapter` ?

lib/Runtime/Library/InJavascript/Intl.js
```diff
@@ -1035,30 +1076,33 @@
 
             // Certain formats have similar patterns on both ecma and windows; will use helper methods for them
             function correctWeekdayEraMonthPattern(patternString, userValue, searchParam) {
-                // parts[1] is either dayofweek.solo, dayofweek, era or month; parts[3] is either abbreviated or full
-                var parts = platform.builtInRegexMatch(patternString, RegExp("{(" + searchParam + "(\\.solo)?)\\.([a-z]*)(\\([0-9]\\))?}"));
+                // parts[1] is either dayofweek.solo, dayofweek, era or month; parts[2] is either abbreviated or full
```
kunalspathak: >parts[2]  [](start = 79, length = 9)

has this changed with your implementation?

lib/Runtime/Base/WindowsGlobalizationAdapter.cpp
```diff
@@ -694,7 +699,153 @@ if (this->object) \
...
+        return success;
+    }
+
+    HRESULT IcuIntlAdapter::NormalizeLanguageTag(_In_z_ const char16 *languageTag, _In_ const charcount_t cch,
+        _Out_ char16 *normalized, _Out_ size_t *normalizedLength)
+    {
+        // Allocate memory for the UTF8 output buffer. Need 3 bytes for each (code point + null) to satisfy SAL.
+        const size_t inputLangTagUtf8SizeAllocated = AllocSizeMath::Mul(AllocSizeMath::Add(cch, 1), 3);
+        // REVIEW (doilij): not perf critical so I used HeapNewArrayZ to zero-out the allocated array
+        utf8char_t *inputLangTagUtf8 = HeapNewArrayZ(utf8char_t, inputLangTagUtf8SizeAllocated);
```
kunalspathak: >HeapNewArrayZ [](start = 39, length = 13)

use `ArenaAllocator` instead? See `GlobalObject::ValidateSyntax()`.

lib/Runtime/Library/IntlEngineInterfaceExtensionObject.cpp
dilijev: Yes. Undefined is when the answer should actually return undefined to the JS logic, null is a sentinel to tell that the JS logic should fall back on another method of determining a useful value (basically, API is not implemented yet in native, but JS can use some reasonable default or an alternate JS-only algorithm to approximate the answer).

Build/Chakra.Build.Paths.props
dilijev: Possibly, in theory these options control where to look for ICU on disk, and could be used even if Intl is disabled. I'll look into the various options.

lib/Runtime/Library/Chakra.Runtime.Library.vcxproj
dilijev: I was not sure whether other directories under this path would be needed. I'll come back to this and refactor when the work is complete.

lib/Common/CommonDefines.h
dilijev: Yep, fixed.

dilijev:
@kunalspathak 

> You might want to skip include windows.globalization.h for Intl_ICU

That's true. For development, I support a hybrid of Intl and WinGlob so for now this is still "sometimes-needed". I'll refactor this later.

lib/Runtime/Library/InJavascript/Intl.js
dilijev: Yes, I optimized out a capture group as I was cleaning this area up.

lib/Runtime/Base/WindowsGlobalizationAdapter.cpp
dilijev: Good point. I'll make this change.

lib/Common/CommonDefines.h
dilijev: Updated.

---
In reply to: [124411993](https://github.com/Microsoft/ChakraCore/pull/3214#discussion_r124411993) [](ancestors = 124411993)

lib/Runtime/Base/WindowsGlobalizationAdapter.h
dilijev: That might be a good idea. But, since this is conditional compilation anyway and calling convention is going to be different (as informed by interfaces to the libraries), it might not be worth the extra overhead that polymorphism would add.

I'll consider it.

---
In reply to: [124413902](https://github.com/Microsoft/ChakraCore/pull/3214#discussion_r124413902) [](ancestors = 124413902)

dilijev:

PR #3218, opened 2017-06-23
AsmJs: Fix bad cast in Lower

Cellule:
Fix potentially bad cast in LowerArgInAsmJs. 
Since the size of the arguments is also an ArgSlot, the max value we can have as the number of arguments is 0x4000. This change is just to future proof this code. 
Add additional check to make sure we are lowering ArgIn_A.
Related [OSG 12366674](https://microsoft.visualstudio.com/os/_workitems?id=12366674)

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/3218)

### DIFF
--- a/lib/Backend/Lower.cpp
+++ b/lib/Backend/Lower.cpp
@@ -10916,30 +10916,24 @@ Lowerer::LoadGeneratorObject(IR::Instr * instrInsert)
 }
 
 IR::Instr *
-Lowerer::LowerArgInAsmJs(IR::Instr * instrArgIn)
+Lowerer::LowerArgInAsmJs(IR::Instr * instr)
 {
     Assert(m_func->GetJITFunctionBody()->IsAsmJsMode());
 
-    int argCount = (int)(int16)m_func->argInsCount;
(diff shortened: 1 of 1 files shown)


lib/Backend/Lower.cpp
```diff
@@ -10920,11 +10920,12 @@ Lowerer::LowerArgInAsmJs(IR::Instr * instrArgIn)
 {
     Assert(m_func->GetJITFunctionBody()->IsAsmJsMode());
 
-    int argCount = (int)(int16)m_func->argInsCount;
+    Js::ArgSlot argCount = m_func->argInsCount;
     IR::Instr * instr = instrArgIn;
 
-    for (int argNum = argCount - 1; argNum >= 0; --argNum)
+    for (Js::ArgSlot argNum = 0; argNum < argCount; ++argNum)
```
MikeHolman: I don't think we need this loop at all. We can let the main lowering loop go one by one. This seems superfluous.

Cellule: Good call!

PR #3220, opened 2017-06-24
Fix Dll handle leak in ch.cpp

xiaoyinl:
In `RunJITServer`, if `UuidFromStringW` or `initRpcServer` fail, it should free the `chakraLibrary` handle.

### DIFF
--- a/bin/ch/ch.cpp
+++ b/bin/ch/ch.cpp
@@ -850,6 +850,7 @@ int _cdecl RunJITServer(int argc, __in_ecount(argc) LPWSTR argv[])
     ChakraRTInterface::ArgInfo argInfo = { argc, argv, PrintUsage, nullptr };
     HINSTANCE chakraLibrary = nullptr;
     bool success = ChakraRTInterface::LoadChakraDll(&argInfo, &chakraLibrary);
+    int status = 0;
 
     if (!success)
     {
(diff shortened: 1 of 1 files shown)


bin/ch/ch.cpp
```diff
@@ -861,6 +861,7 @@ int _cdecl RunJITServer(int argc, __in_ecount(argc) LPWSTR argv[])
     DWORD status = UuidFromStringW((RPC_WSTR)connectionUuidString, &connectionUuid);
     if (status != RPC_S_OK)
     {
+        ChakraRTInterface::UnloadChakraDll(chakraLibrary);
```
Penguinwizzard: I'd rather not duplicate this call like this, especially since we're duplicating it without duplicating the check for its non-null-ness (as seen on L877). Could you instead make this function single-return and have a `goto cleanup` here instead? That way we'd avoid having different or duplicated logic about when to clean-up chakraLibrary.

xiaoyinl:
@Penguinwizzard Thanks for reviewing! Also in general when should I target `release/1.6` and when target `master`?

Penguinwizzard:
Unfortunately, it's hard to give a general rule for that, since our branch targeting is an area of active discussion and continual improvement, and changes due to a variety of factors (e.g. release branches to coincide with windows releases, release branches to coincide with node-chakracore releases, serviced releases, so on). For the time being, bugfixes should be targeted at release/1.6, at least as long as they're not too large, as that's the release that we're currently working on stabilizing. Our automatic pumps should pick up the changes from there and then bring them into downstream branches (such as master). Any feature work, as well as larger bugfixes that may themselves introduce issues, should be targeted at master at the moment. Since this change (and your other PR) are bugfixes for ch-specific code, they're fine to get into release/1.6, and leak fixes are definitely welcome there as well, since that means that we can have our automated testing have more clean runs.

xiaoyinl:
@Penguinwizzard I see. Thank you for your explanation!

Issue #3178, opened 2017-06-19
ch.exe -v doesn't work correctly if path contains non-ascii characters


xiaoyinl:
`PrintChakraCoreVersion()` in `bin/ch/ch.cpp` uses ANSI version of Windows APIs, like `GetModuleFileNameA`, `GetFileVersionInfoA`, etc., so if the current path contains non-ascii characters and the current code page doesn't support these characters, `GetFileVersionInfoA` fails even though the file is there.

![unicodetest](https://user-images.githubusercontent.com/16713498/27271829-fd8e24e6-5494-11e7-8036-115e268dcb04.png)

Maybe related to #2712.

dilijev:
Yeah I noticed this while adding support for printing version number and in working on #2712. This should be fixed, but might take some work to factor out the ASCII versions of APIs.

PR #3112, opened 2017-06-06
Move GetBinaryLocation to SystemInfo.Common.cpp

xiaoyinl:
This is the first step towards fixing #2297. This patch copies the `GetBinaryLocation` function from `bin/ch/Helpers.cpp`
to `lib/Runtime/PlatformAgnostic/Platform/Common/SystemInfo.Common.cpp` (new file) under namespace `PlaatformAgnostic`.

PR #3221, opened 2017-06-24
Replace GetModuleFileName with GetBinaryLocation (fix #2297)

### LINKED ISSUE #2297: PAL: Cleanup GetModuleFileName** and implement corresponding interface

xiaoyinl:
This will fix the remaining task in Issue #2297: replace `GetModuleFileName` with `PlatformAgnostic::SystemInfo::GetBinaryLocation`.

This task is easy if the caller of `GetModuleFileName` can easily work with a UTF-8 string, like `TTDHostBuildCurrentExeDirectory`. But if the caller needs to pass the returned string to a Unicode version of Windows API (see #3178 for example), then it seems the UTF-8 string returned by `GetBinaryLocation` is not so suitable here. What to do in this case?

### DIFF
--- a/bin/ch/Helpers.cpp
+++ b/bin/ch/Helpers.cpp
@@ -7,30 +7,41 @@
 #include <sys/stat.h>
 
 //TODO: x-plat definitions
-#ifdef _WIN32
 #define MAX_URI_LENGTH 512
+#ifdef _WIN32
+#define TTD_MAX_FILE_LENGTH MAX_PATH
 #define TTD_HOST_PATH_SEP "\\"
+#else
+#define TTD_MAX_FILE_LENGTH MAX_URI_LENGTH
+#define TTD_HOST_PATH_SEP "/"
+#endif
 
 void TTDHostBuildCurrentExeDirectory(char* path, size_t* pathLength, size_t bufferLength)
 {
-    wchar exePath[MAX_PATH];
-    GetModuleFileName(NULL, exePath, MAX_PATH);
+    char exePath[TTD_MAX_FILE_LENGTH];
+    PlatformAgnostic::SystemInfo::GetBinaryLocation(exePath, TTD_MAX_FILE_LENGTH);
 
-    size_t i = wcslen(exePath) - 1;
-    while(exePath[i] != _u('\\'))
+    size_t i = strlen(exePath) - 1;
+    while (exePath[i] != TTD_HOST_PATH_SEP[0] && i != 0)
     {
         --i;
     }
-
-    if(i * 3 > bufferLength)
+    if (i == 0)
     {
-        wprintf(_u("Don't overflow path buffer during conversion"));
+        fwprintf(stderr, _u("Can't get current exe directory"));
         exit(1);
     }
-    *pathLength = utf8::EncodeInto((LPUTF8)path, exePath, (charcount_t)(i + 1));
+    if (i + 2 > bufferLength)
+    {
+        fwprintf(stderr, _u("Don't overflow path buffer during copy"));
+        exit(1);
+    }
+    memcpy_s(path, bufferLength, exePath, i + 1);
+    *pathLength = i + 1;
     path[*pathLength] = '\0';
 }
 
+#ifdef _WIN32
 int TTDHostMKDir(const char* path, size_t pathLength)
 {
     char16 cpath[MAX_PATH];
@@ -67,42 +78,6 @@ JsTTDStreamHandle TTDHostOpen(size_t pathLength, const char* path, bool isWrite)
 #define TTDHostRead(buff, size, handle) fread_s(buff, size, 1, size, (FILE*)handle);
 #define TTDHostWrite(buff, size, handle) fwrite(buff, 1, size, (FILE*)handle)
 #else
-
-#ifdef __APPLE__
-#include <mach-o/dyld.h>
-#else
-#include <unistd.h>
-#endif
-#define MAX_URI_LENGTH 512
-#define TTD_HOST_PATH_SEP "/"
-
-void TTDHostBuildCurrentExeDirectory(char* path, size_t* pathLength, size_t bufferLength)
-{
-    char exePath[MAX_URI_LENGTH];
-    //TODO: xplattodo move this logic to PAL
-    #ifdef __APPLE__
-    uint32_t tmpPathSize = sizeof(exePath);
-    _NSGetExecutablePath(exePath, &tmpPathSize);
-    size_t i = strlen(exePath) - 1;
-    #else
-    size_t i = readlink("/proc/self/exe", exePath, MAX_URI_LENGTH) - 1;
-    #endif
-
-    while(exePath[i] != '/')
-    {
-        --i;
-    }
-    *pathLength = i + 1;
-
-    if(*pathLength > bufferLength)
-    {
-        wprintf(_u("Don't overflow path buffer during copy."));
-        exit(1);
-    }
-
-    memcpy_s(path, bufferLength, exePath, *pathLength);
-}
-
 int TTDHostMKDir(const char* path, size_t pathLength)
 {
     return mkdir(path, 0700);
--- a/bin/ch/ch.cpp
+++ b/bin/ch/ch.cpp
@@ -116,9 +116,8 @@ void __stdcall PrintChakraCoreVersion()
     char16 dir[_MAX_DIR];
 
     LPCWSTR chakraDllName = GetChakraDllNameW();
-
     char16 modulename[_MAX_PATH];
-    if (!GetModuleFileNameW(NULL, modulename, _MAX_PATH))
+    if (!PlatformAgnostic::SystemInfo::GetBinaryLocation(modulename, _MAX_PATH))
     {
         return;
     }
--- a/lib/Common/Core/ConfigParser.cpp
+++ b/lib/Common/Core/ConfigParser.cpp
@@ -594,7 +594,7 @@ HRESULT ConfigParser::SetOutputFile(const WCHAR* outputFile, const WCHAR* openMo
 
     char16 fileName[_MAX_PATH];
     char16 moduleName[_MAX_PATH];
-    GetModuleFileName(0, moduleName, _MAX_PATH);
+    PlatformAgnostic::SystemInfo::GetBinaryLocation(moduleName, _MAX_PATH);
     _wsplitpath_s(moduleName, nullptr, 0, nullptr, 0, fileName, _MAX_PATH, nullptr, 0);
     if (_wcsicmp(fileName, _u("WWAHost")) == 0 ||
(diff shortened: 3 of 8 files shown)


xiaoyinl:
How about we implement a UTF-16 version of `GetBinaryLocation`, i.e.  `PlatformAgnostic::SystemInfo::GetBinaryLocationW` for WIN32 only? What do you think @obastemur?

bin/ch/Helpers.cpp
```diff
@@ -7,30 +7,41 @@
...
+    while (exePath[i] != TTD_HOST_PATH_SEP[0] && i != 0)
     {
         --i;
     }
-
-    if(i * 3 > bufferLength)
+    if (i == 0)
     {
-        printf("Don't overflow path buffer during conversion");
+        printf("%s", exePath);  // If exePath doesn't contain path separator, it's an error message
```
obastemur: Could be nice to show an error message here (along with the exePath)

obastemur:
@xiaoyinl Thanks for this! overall looks good. Could you please take a look at the review comment and squash them all ?

bin/ch/Helpers.cpp
xiaoyinl: But based on what we did in #3112, `GetBinaryLocation` already returns an error message in `exePath`: `"GetBinaryLocation: GetModuleFileName has failed."`, etc. 

Or do you want me to add another error message along with that message?

obastemur: Could be nice to see why we print the path there. i.e. `this xxx path is empty or not found etc..`

xiaoyinl: I mean this line doesn't print a path. `exePath` itself is an error message.
The `GetBinaryLocation` function sets `exePath` to either an actual path or an error message. (such as [L52 of SystemInfo.h](https://github.com/xiaoyinl/ChakraCore/blob/eed45f6ae78e004502614afc34edefd5367a325a/lib/Common/PlatformAgnostic/SystemInfo.h#L52))
Here the `while` loop finds the index of the path separator ("\\" or "/") in `exePath`. If `i == 0`, the path separator doesn't exist, so `exePath` is not a path, but an error message.

obastemur: You comment seems sufficient there.

obastemur:
BTW; can we also replace the other usages under bin/ch ? and lib/Common/Core/ConfigParser.cpp?

xiaoyinl:
@obastemur Not so easily. For example, `ConfigParser::SetOutputFile` in `lib/Common/Core/ConfigParser.cpp` needs `moduleName` to be an UTF-16 string because it later passes it to `GetEnvironmentVariableW`, but our GetBinaryLocation returns UTF-8.

bin/ch/Helpers.cpp
xiaoyinl: So does this line look good? Or do you mean to change the printf message to the comment?

obastemur:
@xiaoyinl could it be done by overloading `PlatformAgnostic::SystemInfo::GetBinaryLocation` ?

xiaoyinl:
@obastemur I will try it.

xiaoyinl:
I have implemented `bool SystemInfo::GetBinaryLocation(char16 *path, const unsigned size)`. It needs to return `bool` because some callers check the return value of `GetModuleFileName`. On Linux, this function calls the original `GetBinaryLocation` to get the `path` and then uses `utf8::DecodeUnitsIntoAndNullTerminate` to convert the string from UTF-8 to UTF-16. For this reason, the original `GetBinaryLocation` has to return `bool` so that the overloaded one can propagate the result to its caller.

Does it look good to you, @obastemur?

obastemur:
@xiaoyinl great effort! Let me check the handle part.

obastemur:
@xiaoyinl have you tried with a Unicode path? 

Curious if we could add a test file (name containing Unicode chars)

xiaoyinl:
@obastemur #3178 is an issue when the name contains Unicode chars. I didn't find other issues with Unicode path. I'm not sure how to add test cases for that.

dilijev:
@obastemur @xiaoyinl any update here?

obastemur:
Before fixing conflicts, wait for #4088

dilijev:
Note: https://github.com/Microsoft/ChakraCore/pull/4088 was closed in favor of https://github.com/Microsoft/ChakraCore/pull/4141 which was merged. This PR should be unblocked now.

obastemur:
@xiaoyinl could you please rebase?

xiaoyinl:
@obastemur  Sorry for the delay. I have rebased the change. Is there anything else I need to do in this PR?

obastemur:
@xiaoyinl Thanks for the effort! LGTM. I will merge.

bin/ch/Helpers.cpp
```diff
@@ -7,30 +7,41 @@
 #include <sys/stat.h>
 
 //TODO: x-plat definitions
-#ifdef _WIN32
 #define MAX_URI_LENGTH 512
+#ifdef _WIN32
+#define TTD_MAX_FILE_LENGTH MAX_PATH
 #define TTD_HOST_PATH_SEP "\\"
+#else
+#define TTD_MAX_FILE_LENGTH MAX_URI_LENGTH
```
obastemur: `MAX_URI_LENGTH` definition looks redundant.

bin/ch/Helpers.cpp
```diff
@@ -7,30 +7,41 @@
...
+    while (exePath[i] != TTD_HOST_PATH_SEP[0] && i != 0)
     {
         --i;
     }
-
-    if(i * 3 > bufferLength)
+    if (i == 0)
     {
-        wprintf(_u("Don't overflow path buffer during conversion"));
+        wprintf(_u("Can't get current exe directory"));
```
obastemur: `fprintf(stderr, ...` instead?

bin/ch/Helpers.cpp
```diff
@@ -7,30 +7,41 @@
...
         exit(1);
     }
-    *pathLength = utf8::EncodeInto((LPUTF8)path, exePath, (charcount_t)(i + 1));
+    if (i + 2 > bufferLength)
+    {
+        printf("Don't overflow path buffer during copy");
```
obastemur: `fprintf(stderr, ...` instead?

obastemur:
@xiaoyinl I was about to merge but couple of tiny details bugged me. I do appreciate if you may look at them?

bin/ch/Helpers.cpp
dilijev: Also printf is not allowed anymore in ChakraCore because it will cause an assertion on Windows because printf can't handle our upgrade to UTF-8 output.

dilijev: We prefer using `Output::Print`. At a minimum use `wprintf` and `fwprintf`

obastemur: I don't get the part.. why we use wprintf once the string is not unicode? i.e. the string above? All the `_u(x)` etc. stuff is an unnecessary hassle.

dilijev: It is probably okay because the string is ASCII, but as a general policy we don't use `printf`, and instead use `Output::Print` (which takes a unicode string).

dilijev: (I think in `ch` we don't have access to the `Output` class so we use `wprintf` instead.)

xiaoyinl: So use `fwprintf(stderr, ...);`?

dilijev: Yeah, I think that would be best. @Penguinwizzard since you're making more changes in the area of stdlib output, what do you think?

Penguinwizzard: Yeah, use fwprintf(stderr, ...) for error for now; the _u is a hassle, but I'm not entirely sure if mixing printf and wprintf works nicely in all cases (especially since setting up output encoding properly is already a bit nuanced).

xiaoyinl: I have changed both to `fwprintf(stderr, ...);`

xiaoyinl: @Penguinwizzard Do you want me to change other occurrences of `fprintf` and `wprintf` in this file to `fwprintf`?

dilijev: @xiaoyinl up to you, but I'd say since it's out of scope of this change and it hasn't broken yet, if you want to make changes like that you could do it in a follow-up PR.

bin/ch/Helpers.cpp
xiaoyinl: `MAX_URI_LENGTH` is needed in functions `Helpers::CreateTTDDirectoryAsNeeded` (L393) and `Helpers::TTCreateStreamCallback` (L457).

xiaoyinl: Or do you mean change this line to `#define TTD_MAX_FILE_LENGTH 512`?

bin/ch/Helpers.cpp
xiaoyinl: I see.

Penguinwizzard: @xiaoyinl Thanks for offering, but I actually have a change in the works (but backlogged because of more important work) that does a lot of those changes.

bin/ch/Helpers.cpp
obastemur: Yes.

PR #3222, opened 2017-06-25
NumberToString: Improve performance and reduce cache memory

obastemur:
- NumberToString ~100% faster on xplat and ~50% faster on Windows
- Reduce cache size to 512 JsString from 1024
- introduce new definition `CC_LOW_MEMORY_TARGET` to prevent caching on
low memory targets
- Add Todo note for making cache more effective on long running apps.

### DIFF
--- a/lib/Runtime/Base/ScriptContext.cpp
+++ b/lib/Runtime/Base/ScriptContext.cpp
@@ -1733,25 +1733,37 @@ namespace Js
 
         JavascriptString *string;
 
+// TODO: (obastemur) Could this be dynamic instead of compile time?
+#ifndef CC_LOW_MEMORY_TARGET // we don't need this on a target with low memory
         if (!this->integerStringMap.TryGetValue(value, &string))
         {
             // Add the string to hash table cache
             // Don't add if table is getting too full.  We'll be holding on to
             // too many strings, and table lookup will become too slow.
-            if (this->integerStringMap.Count() > 1024)
+            // TODO: Long term running app, this cache doesn't provide much value?
+            //       i.e. what is the importance of first 512 number to string calls?
+            //       a solution; count the number of times we couldn't use cache
+            //       after cache is full. If it's bigger than X ?? the discard the
+            //       previous cache?
+            if (this->integerStringMap.Count() > 512)
             {
+#endif
                 // Use recycler memory
                 string = TaggedInt::ToString(value, this);
+
+#ifndef CC_LOW_MEMORY_TARGET
             }
             else
             {
-                char16 stringBuffer[20];
+                char16 stringBuffer[22];
 
-                TaggedInt::ToBuffer(value, stringBuffer, _countof(stringBuffer));
-                string = JavascriptString::NewCopySzFromArena(stringBuffer, this, this->GeneralAllocator());
+                int pos = TaggedInt::ToBuffer(value, stringBuffer, _countof(stringBuffer));
+                string = JavascriptString::NewCopySzFromArena(stringBuffer + pos,
+                    this, this->GeneralAllocator(), (_countof(stringBuffer) - 1) - pos);
                 this->integerStringMap.AddNew(value, string);
             }
         }
+#endif
 
         return string;
     }
--- a/lib/Runtime/Language/TaggedInt.cpp
+++ b/lib/Runtime/Language/TaggedInt.cpp
@@ -428,21 +428,66 @@ LblDone:
         return JavascriptNumber::ToVar(uValue >> (nShift & 0x1F), scriptContext);
     }
 
-    void TaggedInt::ToBuffer(Var aValue, __out_ecount_z(bufSize) char16 * buffer, uint bufSize)
+    int TaggedInt::ToBuffer(Var aValue, __out_ecount_z(bufSize) char16 * buffer, uint bufSize)
     {
         return ToBuffer(ToInt32(aValue), buffer, bufSize);
     }
 
-    void TaggedInt::ToBuffer(int value, __out_ecount_z(bufSize) char16 * buffer, uint bufSize)
+    // fills the buffer from the end and returns the start index
+    static int UnsignedToString(unsigned long value, char16 *buffer, int bufferSize)
     {
-        Assert(bufSize > 10);
-        _itow_s(value, buffer, bufSize, 10);
+        static_assert(sizeof(unsigned long) <= 8, "This method may not support the target architecture");
+        AssertMsg(bufferSize >= 22, "Error: bufferSize is too small. value may not be represented properly");
+
+        buffer[bufferSize - 1] = char16(0);
+        int pos = bufferSize - 2;
+        while(value > 9)
+        {
+            const int val100 = value % 100;
+            value /= 100;
+
+            if (val100 < 10)
+            {
+                buffer[pos--] = _u('0') + static_cast<char>(val100);
+                buffer[pos--] = _u('0');
+                continue;
+            }
+
+            buffer[pos--] = _u('0') + static_cast<char>(val100 % 10);
+            buffer[pos--] = _u('0') + static_cast<char>(val100 / 10);
+        }
+
+        if (value && value < 10)
+        {
+            buffer[pos--] = _u('0') + static_cast<char>(value);
+        }
+        else if (pos == bufferSize - 2) // if it was 0
+        {
+            buffer[pos--] = _u('0');
+        }
+
+        return pos + 1;
     }
 
-    void TaggedInt::ToBuffer(uint value, __out_ecount_z(bufSize) char16 * buffer, uint bufSize)
+    static int SignedToString(long value, char16 *buffer, int bufferSize)
     {
-        Assert(bufSize > 10);
-        _ultow_s(value, buffer, bufSize, 10);
+        bool neg = value < 0;
+        unsigned long val = (unsigned long) (neg ? -1 * value : value);
+        int pos = UnsignedToString(val, buffer, bufferSize);
+        if (neg) buffer[--pos] = _u('-');
+        return pos;
+    }
+
+    int TaggedInt::ToBuffer(int value, __out_ecount_z(bufSize) char16 * buffer, uint bufSize)
+    {
+        Assert(bufSize >= 22);
+        return SignedToString(value, buffer, bufSize);
+    }
+
+    int TaggedInt::ToBuffer(uint value, __out_ecount_z(bufSize) char16 * buffer, uint bufSize)
+    {
+        Assert(bufSize >= 22);
+        return UnsignedToString(value, buffer, bufSize);
     }
 
     JavascriptString* TaggedInt::ToString(Var aValue,ScriptContext* scriptContext)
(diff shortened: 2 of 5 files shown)


obastemur:
+ small change to `NewCopySzFromArena` so we don't ''strlen'' the generated number string

lib/Runtime/Language/TaggedInt.cpp
```diff
@@ -428,21 +428,65 @@ namespace Js
...
+            buffer[pos--] = _u('0');
+        }
+
+        return pos + 1;
+    }
+
+    static int SignedToString(long value, char16 *buffer, int bufferSize)
+    {
+        bool neg = value < 0;
+        unsigned long val = (unsigned long) (neg ? -1 * (long)value : value);
```
MSLaguana: Won't this be wrong on `LONG_MIN` since `LONG_MIN * -1 = LONG_MIN`?

lib/Runtime/Language/TaggedInt.cpp
```diff
@@ -428,21 +428,65 @@ namespace Js
...
+        int pos = UnsignedToString(val, buffer, bufferSize);
+        if (neg) buffer[--pos] = _u('-');
```
MSLaguana: This will underflow the buffer if you pass in a large negative number with insufficient buffer size

obastemur: good catch! done.

lib/Runtime/Language/TaggedInt.cpp
obastemur: Could you explain why it is wrong?

MSLaguana: Because signed integral values have one more negative value than positive value I thought it would cause issues, but I hadn't accounted for twos complement meaning that `(signed long) LONG_MIN` has the same representation as `(unsigned long) (-LONG_MIN)`, so after playing around with a compiler it looks like it works. Not sure if it's a case of undefined behavior that just happens to work though.

Penguinwizzard: Overflow and such are undefined behavior for signed ints. As -LONG_MIN is outside the range of a long, it's UB, but in a practical sense it _should_ work as expected on all architectures that we support at the moment. The biggest danger is an aggressive compiler assuming that the resultant `val` has a 0 high-bit, and omitting some range check, but since we don't have any such checks here, it should be non-issue.

MSLaguana: Would it be less undefined-behavior to do something like `unsigned long val = (neg ? 1+ ~(unsigned long)value : value)`?

lib/Runtime/Base/ScriptContext.cpp
```diff
@@ -1733,25 +1733,36 @@ namespace Js
...
             }
             else
             {
                 char16 stringBuffer[20];
 
-                TaggedInt::ToBuffer(value, stringBuffer, _countof(stringBuffer));
-                string = JavascriptString::NewCopySzFromArena(stringBuffer, this, this->GeneralAllocator());
+                int pos = TaggedInt::ToBuffer(value, stringBuffer, _countof(stringBuffer));
+                string = JavascriptString::NewCopySzFromArena(stringBuffer + pos,
+                    this, this->GeneralAllocator(), 19 - pos);
```
jianchun: >19 [](start = 52, length = 2)

Avoid hard code `19` here.

lib/Runtime/Language/TaggedInt.cpp
```diff
@@ -453,16 +497,16 @@ namespace Js
...
+        return JavascriptString::NewCopyBuffer(szBuffer + pos, 19 - pos, scriptContext);
     }
     JavascriptString* TaggedInt::ToString(uint value, ScriptContext* scriptContext)
     {
         char16 szBuffer[20];
-        ToBuffer(value, szBuffer, _countof(szBuffer));
+        int pos = ToBuffer(value, szBuffer, _countof(szBuffer));
 
-        return JavascriptString::NewCopySz(szBuffer, scriptContext);
+        return JavascriptString::NewCopyBuffer(szBuffer + pos, 19 - pos, scriptContext);
```
jianchun: >19 [](start = 63, length = 2)

Similarly avoid hard coded `19` (a few more places).

lib/Runtime/Language/TaggedInt.cpp
```diff
@@ -428,21 +428,65 @@ namespace Js
...
 
-    void TaggedInt::ToBuffer(Var aValue, __out_ecount_z(bufSize) char16 * buffer, uint bufSize)
+    int TaggedInt::ToBuffer(Var aValue, __out_ecount_z(bufSize) char16 * buffer, uint bufSize)
     {
         return ToBuffer(ToInt32(aValue), buffer, bufSize);
     }
 
-    void TaggedInt::ToBuffer(int value, __out_ecount_z(bufSize) char16 * buffer, uint bufSize)
+    // fills the buffer from the end and returns the start index
+    static int UnsignedToString(unsigned long value, char16 *buffer, int bufferSize)
```
jianchun: >bufferSize [](start = 73, length = 10)

Code / logic to validate `bufferSize` is big enough? (Update: what if "pos < 0" below?)

lib/Runtime/Language/TaggedInt.cpp
```diff
@@ -428,21 +428,65 @@ namespace Js
...
+                buffer[pos--] = _u('0') + static_cast<char>(val100);
+                buffer[pos--] = _u('0');
+                continue;
+            }
+
+            buffer[pos--] = _u('0') + static_cast<char>(val100 % 10);
+            buffer[pos--] = _u('0') + static_cast<char>(val100 / 10);
+        }
+
+        if (pos == 0) return 0; // well, the buffer wasn't enough. just return whatever we had
```
jianchun: >if (pos == 0) return 0 [](start = 8, length = 22)

Why not write at pos 0?

lib/Runtime/Language/TaggedInt.cpp
```diff
@@ -428,21 +428,65 @@ namespace Js
...
+        return pos + 1;
+    }
+
+    static int SignedToString(long value, char16 *buffer, int bufferSize)
+    {
+        bool neg = value < 0;
+        unsigned long val = (unsigned long) (neg ? -1 * (long)value : value);
+        int pos = UnsignedToString(val, buffer, bufferSize);
+        if (neg && pos) buffer[--pos] = _u('-');
+        return pos;
```
MSLaguana: It looks like the caller has no way of knowing whether the number was successfully converted to a string or not. If the return value is `> 0` then it succeeded, but if it is `= 0` then it could be that it fits exactly, that the number fit but the `-` sign did not, or that the number has been truncated and is missing digits. Shouldn't this information be exposed to the caller?

lib/Runtime/Language/TaggedInt.cpp
jianchun: >(long)value [](start = 56, length = 11)

`(long)value` -- value is already of type `long`, why cast?

lib/Runtime/Language/TaggedInt.cpp
obastemur: see static_assert for bufferSize.

lib/Runtime/Language/TaggedInt.cpp
obastemur: see static_assert for bufferSize. There is a minimum buffer limit.

lib/Runtime/Language/TaggedInt.cpp
obastemur: done

obastemur:
@MSLaguana @Penguinwizzard @jianchun Thanks for the reviews. Anything else?

MSLaguana:
You use the number 20 in a bunch of places, what's the reasoning there? Is it just a "sufficiently large number"? `ULONG_MAX` seems to be 10+1 digits for a 32bit long with optional sign character, or if it was a 64bit long then it would be 20+1 digits with an optional sign character (neither of those counts including the null terminator). I'd prefer if you converted the magic number 20 into a const/macro to ensure it is used consistently, and some insight into how it was determined.

obastemur:
> You use the number 20 in a bunch of places, what's the reasoning there? Is it just a "sufficiently large number"? 

@MSLaguana  That magical definition was there previously and it is perfectly sufficient for JS. I just made it minimum after reviews on what if the bufferSize is 3... However, in order to make it future proof, Increased to 22 instead. There is a static assert for sizeof(long) hence no need to do other calculations for the sizeof the buffer?

obastemur:
Previous update forces `22` charcount for buffer length. Thus, removed the checks for `pos`.

@MSLaguana @Penguinwizzard @jianchun anything else?

lib/Runtime/Language/TaggedInt.cpp
```diff
@@ -452,17 +502,18 @@ namespace Js
...
 
-        return JavascriptString::NewCopySz(szBuffer, scriptContext);
+        return JavascriptString::NewCopyBuffer(szBuffer + pos, (_countof(szBuffer) - 1) - pos, scriptContext);
     }
+
     JavascriptString* TaggedInt::ToString(uint value, ScriptContext* scriptContext)
     {
-        char16 szBuffer[20];
-        ToBuffer(value, szBuffer, _countof(szBuffer));
+        char16 szBuffer[22];
```
MSLaguana: Please change these magic numbers into a macro or a const to make sure that if it changes in future everything is updated correctly.

obastemur: I prefer the force-check under new toString implementation is hard coded instead of define once for all. Since that one is hard coded and doesn't allow anything below for known architectures, remaining changes wouldn't go beyond the make-up? 

p.s.: I don't have a strong opinion on make-up part but would like to keep buffer checking hard coded. There is enough explanation there.

lib/Runtime/Base/ScriptContext.cpp
```diff
@@ -1733,25 +1733,36 @@ namespace Js
 
         JavascriptString *string;
 
+#ifndef CC_LOW_MEMORY_TARGET // we don't need this on a target with low memory
```
curtisman: Can this be a dynamic flag instead of a build flag?

curtisman: Or just a dynamic flag for a the side of this cache, which can go down to 0?

lib/Runtime/Language/TaggedInt.cpp
```diff
@@ -428,21 +428,71 @@ namespace Js
...
 
-    void TaggedInt::ToBuffer(int value, __out_ecount_z(bufSize) char16 * buffer, uint bufSize)
+    // fills the buffer from the end and returns the start index
+    static int UnsignedToString(unsigned long value, char16 *buffer, int bufferSize)
     {
-        Assert(bufSize > 10);
-        _itow_s(value, buffer, bufSize, 10);
+#ifdef DEBUG
+        static_assert(sizeof(unsigned long) <= 8, "This method may not support the target architecture");
+        if (bufferSize < 22)
```
curtisman: Use assert instead?

lib/Runtime/Base/ScriptContext.cpp
obastemur: I got your point on making it dynamic and scenarios we could test. This is just a side addition to this PR. There are places we can use a runtime flag and others we might have to stick with compile time definitions. 

Since this is just a reminder for future and no-one is actually using this flag, I've added an additional TODO note instead? So, I will update this part while marking others.

lib/Runtime/Language/TaggedInt.cpp
obastemur: @curtisman  Done. Anything else?

obastemur:
It could be nice if this one lands in 1.7. Anything left here to do?

jianchun:
:shipit:

obastemur:
@MSLaguana @jianchun @curtisman Thanks for the review!

PR #3223, opened 2017-06-26
Fix memory leak in Debugger::SetBaseline

xiaoyinl:
If `ChakraRTInterface::JsCreateString` fails, `IfJsrtErrorFailLogAndRetFalse` causes the function `Debugger::SetBaseline()` to return false, without deleting `script` and closing `file`.

### DIFF
--- a/bin/ch/Debugger.cpp
+++ b/bin/ch/Debugger.cpp
@@ -327,7 +327,7 @@ bool Debugger::SetBaseline()
                 script[numChars] = '\0';
 
                 JsValueRef wideScriptRef;
-                IfJsrtErrorFailLogAndRetFalse(ChakraRTInterface::JsCreateString(
+                IfJsErrorFailLogAndHR(ChakraRTInterface::JsCreateString(
                   script, strlen(script), &wideScriptRef));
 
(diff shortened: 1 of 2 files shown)


obastemur:
@Penguinwizzard please review

agarwal-sandeep:
:shipit:

bin/ch/Debugger.cpp
```diff
@@ -327,7 +327,7 @@ bool Debugger::SetBaseline()
                 script[numChars] = '\0';
 
                 JsValueRef wideScriptRef;
-                IfJsrtErrorFailLogAndRetFalse(ChakraRTInterface::JsCreateString(
+                IfJsErrorFailLog(ChakraRTInterface::JsCreateString(
```
Penguinwizzard: This change in the macro changes the return value of the function in this case; since hr will still be S_OK, the function will now return true (where previously it was returning false). It might be more appropriate here to either make a macro that allows assignment to a variable in the error case (so you could set hr = E_FAIL) before jumping to Error.

xiaoyinl:
Done. Thank you, @Penguinwizzard!

PR #3227, opened 2017-06-26
prevent re-entering GC in freeing ThunkBlocks

leirocks:
While system memory is very low, it's trying to free some memory from recycler, which trigger's functionBody Cleanup, then triggers Releasing Interpreter ThunkBlocks, while releasing ThunkBlocks it tries to add to a free list, which is a dynamic allocated freelist, hence trigger another memory allocation and re-entering recycler memory recover routine
Fix is to use no-recovery alloc for this free list in arena


leirocks:
@dotnet-bot test Ubuntu static_ubuntu_linux_release please

PR #3232, opened 2017-06-27
Fix file handle leak in Helpers.cpp

xiaoyinl:
We need to make sure `fclose(file)` is always called after the file has been successfully opened, so move it after `Error` label.

Another change is to move error message printing outside `#ifdef _WIN32`, so that it is printed on Linux.

### DIFF
--- a/bin/ch/Helpers.cpp
+++ b/bin/ch/Helpers.cpp
@@ -353,10 +353,10 @@ HRESULT Helpers::LoadBinaryFile(LPCSTR filename, LPCSTR& contents, UINT& lengthB
     {
         if (printFileOpenError)
         {
+            fprintf(stderr, "Error in opening file '%s' ", filename);
 #ifdef _WIN32
             DWORD lastError = GetLastError();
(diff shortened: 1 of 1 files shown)


bin/ch/Helpers.cpp
```diff
@@ -370,7 +370,7 @@ HRESULT Helpers::LoadBinaryFile(LPCSTR filename, LPCSTR& contents, UINT& lengthB
             }
 #endif
             fprintf(stderr, "\n");
-            IfFailGo(E_FAIL);
+            return E_FAIL;
```
kfarnung: Looks like the `return E_FAIL` can be moved out a level and eliminate the `else` condition?

xiaoyinl: Yes, you're right. I didn't notice this. I will fix it soon.

xiaoyinl: Done. Thanks!

kfarnung:
Once all the checks pass I'll merge this PR.

PR #3235, opened 2017-06-27
Fix memory leak in AutoSystemInfo::InitPhysicalProcessorCount

xiaoyinl:
The heap buffer `pBufferStart` is not deleted if `GetLogicalProcessorInformation` fails.


digitalinfinity:
Thanks @xiaoyinl - how did you find this bug, was it a tool?
@kfarnung - we can go ahead and merge this commit- are you going to do that or do you want me to?

xiaoyinl:
@digitalinfinity Thanks for your review! No, I didn't use a tool. I just found it while reading the source code in VS.

PR #3238, opened 2017-06-28
Change thrown non-Error objects to rethrow as uncaught exceptions

tcare:
Fixes OS: 11743399.

The macro error handling for thrown objects makes an assumption that if we throw non-Error objects, it has to be null. In this case, we revert to OOM error handling since we expect that a null error object indicates a fatal error.

However, it's possible someone will throw a non-Error object, in which case we can rethrow to let the host handle it.

Tested with Crawler and things seem to be clean. We might get a spike of bugs that previously showed as OOM lower on the stack. These should be more actionable with this change.

### DIFF
--- a/lib/Common/Exceptions/Throw.h
+++ b/lib/Common/Exceptions/Throw.h
@@ -208,6 +208,10 @@ namespace Js {
         {       \
             hr = GetRuntimeErrorFunc(Js::RecyclableObject::FromVar(errorObject), nullptr);   \
         }   \
+        else if (errorObject != nullptr) \
+        {  \
+            hr = JSERR_UncaughtException; \
+        }  \
         else \
         {  \
             AssertMsg(errorObject == nullptr, "errorObject should be NULL");  \


lib/Common/Exceptions/Throw.h
```diff
@@ -208,6 +208,10 @@ namespace Js {
         {       \
             hr = GetRuntimeErrorFunc(Js::RecyclableObject::FromVar(errorObject), nullptr);   \
         }   \
+        else if (errorObject != nullptr) \
```
jianchun: Just curious of the bug's effect... where do I see an OOM if my script "throw 123"? Tried Edge, F12, ch -- none shows OOM?

tcare:
```
"use strict";

const m = new MessageEvent('type', { get ports() { throw { name: 'Error' }; } });
```

Gives us:

> SCRIPT14: SCRIPT14: Not enough storage is available to complete this operation.

PR #3239, opened 2017-06-28
move recycler verify mark false reference checks to runtime

jianchun:
The checks were in Common.Memory and had no access to runtime headers. We had
to hard code runtime data field offsets. This was inconvenient and unstable.

This change moves the checks to runtime so that we can use exact field offsets
with "offsetof".

Added one more GC false reference filter (native array elements).

Added one unused field initialization to avoid false reference.


jianchun:
Thanks @digitalinfinity @leirocks

PR #3240, opened 2017-06-28
Fix memory leak in WScriptJsrt::LoadTextFileCallback

xiaoyinl:
`fileContent` should be freed when `ChakraRTInterface::JsCreateString` fails.


kfarnung:
Hey @xiaoyinl7, thanks for the fix.  Was there a particular scenario where you were hitting this leak?

xiaoyinl:
Thank you! I didn't find a particular scenario that hit the leak. I just read the source code and noticed the resource isn't freed in all paths.

kfarnung:
@xiaoyinl Thanks for the fix!  I've just merged it.

xiaoyinl:
Thanks for reviewing and merging!

PR #3243, opened 2017-06-28
WASM - Debug trace memory writes

Cellule:
Add config flag to trace memory writes in wasm
Very useful to find bugs in jitted code because we can determine where things went wrong when used with `-trace:WasmInOut`

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/3243)

### DIFF
--- a/lib/Runtime/Library/WebAssemblyMemory.cpp
+++ b/lib/Runtime/Library/WebAssemblyMemory.cpp
@@ -170,6 +170,53 @@ WebAssemblyMemory::GrowHelper(WebAssemblyMemory * mem, uint32 deltaPages)
     return mem->GrowInternal(deltaPages);
 }
 
+void WebAssemblyMemory::TraceMemWrite(WebAssemblyMemory* mem, uint32 index, uint32 offset, Js::ArrayBufferView::ViewType viewType, uint32 bytecodeOffset, ScriptContext* context)
+{
+#if DBG
+    // Must call after the write
+    Assert(mem);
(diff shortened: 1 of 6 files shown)


lib/Runtime/Library/WebAssemblyMemory.cpp
```diff
@@ -170,6 +170,53 @@ WebAssemblyMemory::GrowHelper(WebAssemblyMemory * mem, uint32 deltaPages)
...
+    case ArrayBufferView::ViewType::TYPE_FLOAT32: Output::Print(_u("%.4f\n"), *(float*)(buffer + bigIndex)); break;
+    case ArrayBufferView::ViewType::TYPE_FLOAT64: Output::Print(_u("%.8f\n"), *(double*)(buffer + bigIndex)); break;
+    case ArrayBufferView::ViewType::TYPE_INT64: Output::Print(_u("%lld\n"), *(int64*)(buffer + bigIndex)); break;
+    default:
+        CompileAssert(ArrayBufferView::ViewType::TYPE_COUNT == 15);
+        Assert(UNREACHED);
+    }
+    return;
+#else
+    Assert(UNREACHED);
```
Penguinwizzard: Rather than do it this way, why not wrap the whole function in #if DBG and cause a link error if a non-DBG-only code block attempts to use it?

Cellule: Initially I wanted to do this, then I changed my mind for some reason. But I don't see any reason to not wrap it with `#if DBG` so yeah I think it's a good idea

PR #3244, opened 2017-06-28
WASM - x86 int64 gt_s & lt_s

Cellule:
On x86 for int64 compare, we need to compare the low bits using unsigned comparisons.

This removes the need for the hack to prevent cmp src swapping.

Also, lower int64 branching directly instead of relying on cmp then branch on cmp result as it was incorrectly branching on a signed comparison of the lower bits.

Fixes [OS#12197588](https://microsoft.visualstudio.com/os/_workitems?id=12197588)

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/3244)


Cellule:
Hold off on the review, I found a related bug with branches for int64 on x86. I will include the fix in this PR

Cellule:
Ping, can someone review this, it is hitting a lot in WasmExprgen

Cellule:
Thanks @Penguinwizzard

PR #3258, opened 2017-06-28
Add missing project dependency to ChakraCore.vcxproj

digitalinfinity:
wabt was not listed as a dependency in ChakraCore.vcxproj. In VS2017 with lightweight solution loading enabled, this was consistently reproing as a link failure. Adding an explicit project dependency to wabt appears to fix the problem.

### DIFF
--- a/bin/ChakraCore/ChakraCore.vcxproj
+++ b/bin/ChakraCore/ChakraCore.vcxproj
@@ -162,6 +162,9 @@
     <ProjectReference Include="..\..\lib\WasmReader\Chakra.WasmReader.vcxproj">
       <Project>{53D52B0B-86D9-4D31-AD09-0D6B3C063ADD}</Project>
     </ProjectReference>
+    <ProjectReference Condition="'$(BuildWabt)'=='true'" Include="..\..\lib\wabt\wabt.vcxproj">
+      <Project>{F48B3491-81DF-4F49-B35F-3308CBE6A379}</Project>
+    </ProjectReference>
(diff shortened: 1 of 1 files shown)


digitalinfinity:
Looks like CI is failing because wabt isn't built in Jenkins? Probably need to modify the project reference to be conditional to `$(BuildWabt)` being set? Testing this change locally, will update PR if it works

bin/ChakraCore/ChakraCore.vcxproj
```diff
@@ -166,6 +166,11 @@
       <Project>{706083f7-6aa4-4558-a153-6352ef9110f6}</Project>
     </ProjectReference>
   </ItemGroup>
+  <ItemGroup Condition="'$(BuildWabt)'=='true'">
+    <ProjectReference Include="..\..\lib\wabt\wabt.vcxproj">
```
Cellule: You can also put the condition on the ProjectReference to avoid the need for an extra ItemGroup

digitalinfinity:
@dotnet-bot test Windows x64_release Windows x86_debug please

digitalinfinity:
@dotnet-bot test Windows ci_slow_x64_debug please

digitalinfinity:

bin/ChakraCore/ChakraCore.vcxproj
digitalinfinity: Ok- I thought about it and went this route in case wabt had more than one project but I can change it to be on the ProjectReference instead

PR #3259, opened 2017-06-29
Update the expected results for Percent formatter based on latest Windows::Globalization behavior.

atulkatti:
I have verified this with a C# app as well and it looks like a change in the Windows::Globalization API.

### DIFF
--- a/test/Intl/NumberFormat.js
+++ b/test/Intl/NumberFormat.js
@@ -4,7 +4,6 @@
 //-------------------------------------------------------------------------------------------------------
 
 // WinTH: Changed to use non-breaking space in some output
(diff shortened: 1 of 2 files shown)


atulkatti:
@Microsoft/chakra-developers

kunalspathak:
What do you mean by latest windows.globalization behavior? What happens when chakracore is ran on machines having older `windows.globalization.dll`?

atulkatti:
You are right Kunal. Actually, this will fail while running unit tests on dev boxes with say RS2. I just found Issue#3014 where Doug has suggested we do regex matching as the spec is open about the output format.

Issue #3262, opened 2017-06-29
ChakraCore hello scenario generates segmentation fault on Linux Skylake server


litian2025:
We encountered an issue that on a hello world scenario with ChakraCore, Linux Skylake server generates a segmentation fault. This issue only happens on Skylake server, not on Haswell server, not on Skylake client system either. It only happens on Linux, not on Windows. 

Following is hello world test case and call stack info when the segmentation fault happens:

/home/git/ttt/ChakraCore/out/Debug$ cat hello.js
console.log("hello");
/home/git/ttt/ChakraCore/out/Debug$ ./ch hello.js
Segmentation fault (core dumped)
/home/git/ttt/ChakraCore/out/Debug$ lldb ch hello.js
(lldb) target create "ch"
Current executable set to 'ch' (x86_64).
(lldb) settings set -- target.run-args  "hello.js"
(lldb) br set -n NewThunkBlock
Breakpoint 1: no locations (pending).
WARNING:  Unable to resolve breakpoint to any actual locations.
(lldb) r
Process 238376 launched: '/home/git/ttt/ChakraCore/out/Debug/ch' (x86_64)
1 location added to breakpoint 1
Process 238376 stopped
* thread #1: tid = 238376, 0x00007ffff4b14248 libChakraCore.so`InterpreterThunkEmitter::NewThunkBlock(this=0x00005555558ba3a8) + 24 at InterpreterThunkEmitter.cpp:328, name = 'ch', stop reason = breakpoint 1.1
    frame #0: 0x00007ffff4b14248 libChakraCore.so`InterpreterThunkEmitter::NewThunkBlock(this=0x00005555558ba3a8) + 24 at InterpreterThunkEmitter.cpp:328
   325      }
   326  #endif
   327
-> 328      Assert(this->thunkCount == 0);
   329      BYTE* buffer;
   330
   331      EmitBufferAllocation<VirtualAllocWrapper, PreReservedVirtualAllocWrapper> * allocation = emitBufferManager.AllocateBuffer(BlockSize, &buffer);
(lldb) continue
Process 238376 resuming
Process 238376 stopped
* thread #1: tid = 238376, 0x00007ffff7126d80 libgcc_s.so.1`__register_frame, name = 'ch', stop reason = signal SIGSEGV
    frame #0: 0x00007ffff7126d80 libgcc_s.so.1`__register_frame
libgcc_s.so.1`__register_frame:
[...]

PR #3263, opened 2017-06-29
Fix Segmentation fault on Linux Skylake server

litian2025:
The root cause of the issue is because on Windows, if memory is protected as  PAGE_EXECUTE, it is implies that it is also PAGE_EXECUTE_READ. This works fine on Haswell server and Skylake client on Linux. However, On Linux Skylake server, it is more restrict and explict, PROT_EXE is Executable only, not readable.  The fix is to ensure PAL emulates Windows API. This fix the issue #3262.


litian2025:
@curtisman PTAL

litian2025:
@curtisman @digitalinfinity Thank you for approving the change, could you please merge it? Please let me know what else that I need to do from our end, I don't think I have the permission right to do the merge myself. Thanks.

curtisman:
@obastemur Can you help merging?

pal/src/map/common.cpp
```diff
@@ -49,7 +49,7 @@ INT W32toUnixAccessControl( IN DWORD flProtect )
         MemAccessControl = PROT_EXEC | PROT_READ | PROT_WRITE;
         break;
     case PAGE_EXECUTE :
-        MemAccessControl = PROT_EXEC;
+        MemAccessControl = PROT_EXEC | PROT_READ; // WinAPI PAGE_EXECUTE also implies readable
```
obastemur: @curtisman @digitalinfinity this looks to me a workaround rather than a fix. See `PAGE_EXECUTE_READ` below. If our code needs to read the memory that has `PAGE_EXECUTE` rights, shouldn't we better set `PAGE_EXECUTE_READ` instead ? I see on Windows, `PAGE_EXECUTE` gives a limited READ capabilities? But adding `PROT_READ` doesn't look to me as a similar behavior.

curtisman: The goal here is to mimic windows behavior.  I believe @litian2025  has experiement that `PAGE_EXECUTE` gives you read permission.    What do you mean "limited" READ?   I though reading is reading is reading.   How do you distinguish different "kind" of reading?

obastemur: When `PAGE_EXECUTE` is set, for example not sure `ReadProcessMemory` could read that area? however, if we set `PAGE_EXECUTE_READ` that would be another story. There is no such thing on Linux and we mimic these tags to `PROT_*` And so, if we set `PROT_READ` we make that area truly readable. Do we really want that?

curtisman: Try this:

    #include "windows.h"
    #include "stdio.h"
    void main()
    {

        void * address = VirtualAlloc(NULL, 64 * 1024, MEM_COMMIT | MEM_RESERVE, PAGE_EXECUTE);
        int i;
        SIZE_T readCount;
        if (!ReadProcessMemory(GetCurrentProcess(), address, &i, sizeof(i), &readCount))
        {
            printf("Unable to read\n");
        }
        else
        {
            printf("Read %d, count %d\n", i, readCount);
        }
    }

I got
`Read 0, count 4`

I am not sure you can distinguish different reads.  If you can read, you can read, and it seems like on windows PAGE_EXECUTE grants you read permission?

I don't mind us switching to PAGE_EXECUTE_READ to be more explicit as well.  But the PAL should emulate what windows API does.

obastemur: > I am not sure you can distinguish different reads. If you can read, you can read, and it seems like on windows PAGE_EXECUTE grants you read permission?

Isn't this to be an issue when reading from another process ?

> I don't mind us switching to PAGE_EXECUTE_READ to be more explicit as well.

+1 to this. It could be nice to pin down the area we actually need to read and don't affect others?

curtisman: I don't know about other process, but the rules should be the same, no?

litian2025: This boils down to that if we want PAL implementation carries the same meaning as Windows or not. Since  on Windows : https://msdn.microsoft.com/en-us/library/windows/desktop/aa366786(v=vs.85).aspx, PAGE_EXECUTE on Windows implies readable, only writes will generate AV. There is no such thing on Windows that PAGE_EXECUTE means execute only, not readable. This PR is to ensure that on PAL implementation emulates the Windows API.

obastemur: > Since on Windows : https://msdn.microsoft.com/en-us/library/windows/desktop/aa366786(v=vs.85).aspx, PAGE_EXECUTE on Windows implies readable, only writes will generate AV. 

Well, the problem is that I don't see any promise regarding to `READ` access on that document. Isn't `PAGE_EXECUTE_READ` supposed to be used for that?

Even more, memory dump tools show different marks for PAGE_EXECUTE_READ and PAGE_EXECUTE on a particular region. IMHO, PAL matches to Windows API. The problem here is that we read on xplat while we don't on Windows. This PR showed us that even if we would read that area, we could do it without an issue.

I don't have the environment to test it; we have very few `PAGE_EXECUTE` usage in the code, can we test which one is causing this problem and replace only that/those to `PAGE_EXECUTE_READ` instead?

obastemur: P.S: I don't want to hold this PR. My argument here is; `PROT_READ` gives much more access than what we wish. I'm not sure this is really what we want there.

litian2025: Update the PR to replace PAGE_EXECUTE with PAGE_EXECUTE_READ when appropriate, while still keeping PAL matches Windows, this change makes the code easier to read and causes less confusion since we use PAGE_EXECUTE_READ explicitly instead of using the implicit meaning of PAGE_EXECUTE when we actually read from it. 

@curtisman @obastemur PTAL

obastemur:
LGTM. @litian2025 Thanks for the changes.

litian2025:
@obastemur Thank you, I will squash the commits to single commit only and update.

obastemur:
@litian2025 no need to squash these (they exactly do what they supposed to). I will merge. Thanks again.

litian2025:
Great, thank you! @obastemur

PR #3264, opened 2017-06-29
Extend return register lifetime

Cellule:
Keep the `Ret` instruction to extend the lifetime of the return register(s) then remove it in FinalLower. 

Fixing a bug in wasm int64 returns on x86  where edx is reused to reload a spilled value for eax.

Fixes [OSG 12195427](https://microsoft.visualstudio.com/os/_workitems?id=12195427)

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/3264)

### DIFF
--- a/lib/Backend/Lower.cpp
+++ b/lib/Backend/Lower.cpp
@@ -23589,6 +23589,7 @@ Lowerer::ValidOpcodeAfterLower(IR::Instr* instr, Func * func)
     }
     switch (opcode)
     {
+    case Js::OpCode::Ret:
     case Js::OpCode::Label:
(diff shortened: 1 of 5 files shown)


LouisLaf:
> {

Should also fix ARM's LowerRet() for consistency.

---
Refers to: lib/Backend/arm/LowerMD.cpp:2426 in bb4e9fa. [](commit_id = bb4e9fa4f39e62bb34bfa0b9fea7f0b65e7232a1, deletion_comment = False)

LouisLaf:
:shipit:

PR #3266, opened 2017-06-29
Proxy should allow duplicate ownKeys when returning getOwnPropertyDescriptors

tcare:
Fixes test262 test\built-ins\Object\getOwnPropertyDescriptors\duplicate-keys.js and OS: 8729344

Simple spec update to allow duplicate keys to be returned in Object.getOwnPropertyDescriptors by always assigning the return array value. We do not need to keep track of the count because we always remove all instances from the property map in spec code.

### DIFF
--- a/lib/Runtime/Library/JavascriptProxy.h
+++ b/lib/Runtime/Library/JavascriptProxy.h
@@ -185,11 +185,13 @@ namespace Js
                     {
                         targetToTrapResultMap.Add(propertyId, true);
                     }
+                }
 
-                    if (fn(propertyRecord))
-                    {
-                        trapResult->DirectSetItemAt(trapResultIndex++, element);
-                    }
+                // We explicitly allow duplicates in the results. A map is sufficient since the spec steps that remove entries
+                // remove ALL of them at the same time.
+                if (fn(propertyRecord))
+                {
+                    trapResult->DirectSetItemAt(trapResultIndex++, element);
                 }
             }
         }


lib/Common/Exceptions/Throw.h
```diff
@@ -208,6 +208,10 @@ namespace Js {
         {       \
             hr = GetRuntimeErrorFunc(Js::RecyclableObject::FromVar(errorObject), nullptr);   \
         }   \
+        else if (errorObject != nullptr) \
+        {  \
+            hr = JSERR_UncaughtException; \
+        }  \
```
kunalspathak: is this change related to proxy?

lib/Runtime/Library/JavascriptProxy.h
```diff
@@ -185,11 +185,13 @@ namespace Js
                     {
                         targetToTrapResultMap.Add(propertyId, true);
                     }
+                }
 
-                    if (fn(propertyRecord))
-                    {
-                        trapResult->DirectSetItemAt(trapResultIndex++, element);
-                    }
+                // We explcitly allow duplicates in the results. A map is sufficient since the spec steps that remove entries
```
kunalspathak: >explcitly [](start = 22, length = 9)

can we have a unit test?

tcare: Test262 is run in nightlies and should sufficiently cover this case. I could add that one though..

lib/Common/Exceptions/Throw.h
tcare: Nope, I need to push again.

kunalspathak:
I think it will be good to have it so bugs are caught early in CI run itself.

Thanks,
Kunal

akroshg:
:shipit:

kunalspathak:
:shipit:

PR #3267, opened 2017-06-29
Asm.js check return type from last stmt

Cellule:
Fix how we check if the last asm.js stmt is a return and throw if the type doesn't match

Fixes [OS#9592312](https://microsoft.visualstudio.com/os/_workitems?id=9592312)

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/3267)


Cellule:

lib/Runtime/Language/AsmJsByteCodeGenerator.cpp
```diff
@@ -426,18 +426,22 @@ namespace Js
...
+            stmt = ParserWrapper::GetBinaryLeft(varStmts);
             EmitTopLevelStatement( stmt );
             varStmts = ParserWrapper::GetBinaryRight(varStmts);
         }
         Assert(!varStmts->CapturesSyms());
 
         // if last statement isn't return, type must be void
-        if (varStmts->nop != knopReturn)
+        if (!stmt || stmt->nop != knopReturn)
         {
```
meg-gupta: Shouldn't we still check the right node of the binary statement ? (varStmts->nop)

meg-gupta: Ah, looks like rightNode will be knopEndCode, and we have to check left node for the list. LGTM.

Cellule:
Rebased, squashed and added OS bug number to commit message for tracking

PR #3269, opened 2017-06-30
modify InterpreterThunkEmitter::IsInHeap to handle JIT process crashing

MikeHolman:
This is debug only method used in asserts to validate whether an entrypoint is valid. In case JIT process crashed, let's just return true to keep asserts from firing.


lib/Backend/NativeCodeGenerator.cpp
```diff
@@ -3716,7 +3716,6 @@ JITManager::HandleServerCallResult(HRESULT hr, RemoteCallType callType)
         // inform job manager that JIT work item has been cancelled
         throw Js::OperationAbortedException();
     case RemoteCallType::HeapQuery:
```
leirocks: make HeapQuery debug only?

MikeHolman: sure

PR #3278, opened 2017-07-01
When clearing inline cache remove it from invalidation list

agarwal-sandeep:
During non-shutdown scenarios clearing the inline cache zeroes the
invalidationListSlotPtr but the node in various inlinecachelist on
threadcontext still point to inlinecache. In release build we only check
for values which are null and compact them but in debug build we have
validation which access the type stored thus reading invalid memory.
Invalidation should also help in more compaction.

### DIFF
--- a/lib/Runtime/Base/ScriptContext.cpp
+++ b/lib/Runtime/Base/ScriptContext.cpp
@@ -1677,7 +1677,7 @@ namespace Js
                         Output::Print(_u("PropertyString '%s' : Invalidating LdElem cache for type %p\n"), string->GetString(), type);
                     }
 #endif
-                    cache->GetInlineCaches()[cache->GetInlineCacheIndexForType(type)].Clear();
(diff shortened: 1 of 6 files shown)


agarwal-sandeep:
@pleath please take a look

rajatd:
:shipit:

agarwal-sandeep:
Thanks @rajatd

pleath:
Is it possible that these caches will be populated again and need to be invalidated later?

agarwal-sandeep:
In clear the invalidationListSlotPtr is zeroed so the link between the node in the list and this inlinecache is broken. When the cache is populated again we check if invalidationListSlotPtr is nullptr and re-register them so they get a new entry in list and invalidationListSlotPtr is updated to address of new entry.

---
In reply to: [313236723](https://github.com/Microsoft/ChakraCore/pull/3278#issuecomment-313236723) [](ancestors = 313236723)

pleath:
Sounds perfect. Thanks.

Issue #3014, opened 2017-05-22
test/Intl/NumberFormat*.js: Make expected output more general to support different implementations


dilijev:
Various versions of WinGlob across supported Windows OS's as well as supporting ICU (#2919) could fracture these tests further with different implementation-specific expected outputs. For example, formatting a number as a `percent` could have either a regular space (`\x20`), NON-BREAKING SPACE (`\u00a0`) or no space at all.

The goal of Intl tests is to ensure the output is reasonable, not that the output matches specifically a spec-defined template (because no such template exists, the spec allows implementation-defined behavior). Therefore it makes sense to capture various reasonable possibilities (as output by the various implementations we support) via a RegExp match or something similar.

For example we could change expected output like `12,300${NON_BREAKING_SPACE}%` to a test against the RegExp `/12,300[\x20\u00a0]?%/`.

/cc @tcare @jianchun

PR #3279, opened 2017-07-03
Use a Regex to match Percent formatter results to take into account behavior differences on different platforms.

atulkatti:


atulkatti:
@Microsoft/chakra-developers

dilijev:
Looks like this would close https://github.com/Microsoft/ChakraCore/issues/3014

While you're doing this, note that the only different between the platform specific Win10 etc is this difference in output. With this change, those tests should be combined. Also, anywhere a NON_BREAKING_SPACE is used, this pattern should be used as well.

test/Intl/NumberFormat.js
```diff
@@ -4,7 +4,6 @@
 //-------------------------------------------------------------------------------------------------------
 
 // WinTH: Changed to use non-breaking space in some output
```
dilijev: I think you can remove this line at this point as well. It has varied a bit between certain releases, the regex covers it now.

PR #3284, opened 2017-07-03
Fix #3257, Fix #3247, Fix #3306: Module import relative path

### LINKED ISSUE #3247: Modules evaluated twice when referenced circularly
```js
// bug.js
WScript.LoadModule(`import "bug-dep.js"`);

// bug-dep.js
import './bug-dep.js';
print('eval');
```

Actual: eval printed twice
Expected: eval printed once

### LINKED ISSUE #3257: Module imports in ch.exe/jshost.exe should be relative to the current module
Currently, all imports are relative to CWD. It should instead be relative to the module doing the import.

### LINKED ISSUE #3306: Module: premature instantiation of dynamic import modules
Hit this bug while working on #3257. To check in fix as part of #3284.
```javascript
// bug.js
WScript.LoadScriptFile("./mod0.js", "module");
WScript.LoadScriptFile("./script0.js");
```
```javascript
// mod0.js
import './mod1.js';
import './mod2.js';
console.log("mod0");
```
```javascript
// mod1.js
import './mod2.js';
console.log("mod1");
```
```javascript
// mod2.js
import './mod0.js';
console.log("mod2");
```
```javascript
// script0.js
console.log("script0");
import('./mod1.js');
import('./mod2.js');
```
Expected:
```javascript
script0
mod2
mod1
mod0
```
Actual
```javascript
script0
ASSERTION 179256: (lib\Runtime\Language\SourceTextModuleRecord.cpp, line 817) moduleRecord->WasParsed()
 Failure: (moduleRecord->WasParsed())
```

suwc:
Issue3257/3247/3306: Module import relative path
1) Change module imports (including dynamic imports) to be relative to the
source file that calls import (#3257)
2) Change to using full path as key for module record map (#3247)
3) A module can be evaluated after a 2nd request for module evaluation
is queued, therefore 2nd evaluation attempt should just return instead
of an error.
4) When dynamic import happens in the middle of module instantiation of
staticly imported modules, using the presence of promise to trigger
instantiation can be premature. Fix by going back to instantiate from
root module only. (#3306)
5) Remove canonicalization of file and path names by converting to lower
case because this is not UTF8 nor xplat friendly. Update baselines.


suwc:
@Microsoft/chakra-es @Microsoft/chakra-developers please review. Thanks!

bterlson:
This is WONDERFUL.

curtisman:
>         JsValueRef returnRef = moduleRecord->ModuleEvaluation();

This now returns nullptr when it was evaluated already.  We still want JsNoError?
If so, we don't use JsErrorModuleEvaluated any more, should that be removed?


---
Refers to: lib/Jsrt/Core/JsrtCore.cpp:125 in a94be95. [](commit_id = a94be95f4ab1dabc03c7f6ddaaa85cd505d62ae8, deletion_comment = False)

bin/ch/WScriptJsrt.cpp
```diff
@@ -37,6 +38,8 @@
 
 MessageQueue* WScriptJsrt::messageQueue = nullptr;
 std::map<std::string, JsModuleRecord>  WScriptJsrt::moduleRecordMap;
+std::map<JsModuleRecord, std::string>  WScriptJsrt::moduleDirMap;
```
curtisman: it is fine for limited current use for now, but this will not work when running with multiple threads.
Also, JsModuleRecord is Js::SourceTexttMdouleRecord. with is GC allocated.
That means that if the JsModuleRecord is collected, you will have dangling records here.

bin/ch/WScriptJsrt.cpp
```diff
@@ -325,11 +334,46 @@ JsErrorCode WScriptJsrt::InitializeModuleInfo(JsValueRef specifier, JsModuleReco
...
+        len = strlen(fullPathNarrow);
+        for (size_t i = 0; i < len; i++)
+        {
+            fullPathNarrow[i] = (char)tolower(fullPathNarrow[i]);
+        }
+    }
+    else
+    {
+        // No fileName provided (WScript.LoadScript()), use dummy "script.js"
+        strcpy_s(fullPathNarrow, 10, "script.js");
```
curtisman: Shouldn't this be expanded to full path too?
It is probably better to default script.js outside of GetFullPath.

bin/ch/WScriptJsrt.cpp
```diff
@@ -325,11 +334,46 @@ JsErrorCode WScriptJsrt::InitializeModuleInfo(JsValueRef specifier, JsModuleReco
...
+{
+    size_t len = 0;
+    if (fileName)
+    {
+        if (_fullpath(fullPathNarrow, fileName, _MAX_PATH) == nullptr)
+        {
+            return false;
+        }
+        // canonicalize that path name to lower case for the profile storage
+        // REVIEW: This doesn't work for UTF8...
```
curtisman: Doesn't work for cross path either.  Path name may be case sensitive.

bin/ch/WScriptJsrt.cpp
```diff
@@ -325,11 +334,46 @@ JsErrorCode WScriptJsrt::InitializeModuleInfo(JsValueRef specifier, JsModuleReco
...
 
-JsErrorCode WScriptJsrt::LoadModuleFromString(LPCSTR fileName, LPCSTR fileContent)
+bool WScriptJsrt::GetFullPath(LPCSTR fileName, __out char* const fullPathNarrow)
+{
+    size_t len = 0;
+    if (fileName)
+    {
+        if (_fullpath(fullPathNarrow, fileName, _MAX_PATH) == nullptr)
+        {
+            return false;
```
curtisman: Should we just error in this case?

curtisman: Looks like we are in WScriptJsrt::LoadScript, but not in FetchImportedModuleHelper

---
In reply to: [125370693](https://github.com/Microsoft/ChakraCore/pull/3284#discussion_r125370693) [](ancestors = 125370693)

bin/ch/WScriptJsrt.cpp
```diff
@@ -1490,7 +1527,22 @@ HRESULT WScriptJsrt::ModuleMessage::Call(LPCSTR fileName)
...
+            bool useReferenceModulePath = false;
+            if (this->moduleRecord)
+            {
+                auto moduleDirEntry = moduleDirMap.find(this->moduleRecord);
+                if (moduleDirEntry != moduleDirMap.end())
+                {
+                    _getcwd(oldpath, _MAX_PATH);
+                    useReferenceModulePath = true;
+                    std::string dir = moduleDirEntry->second;
+                    _chdir(dir.c_str());
```
curtisman: Changing current directory in an multi-thread application is usually frown upon. , because is the property is process wide.

suwc: Chdir happens in FetchImportedModule*() or ModuleMessage::Call(). I thought only the main thread will go to these two places. No?

---
In reply to: [125371146](https://github.com/Microsoft/ChakraCore/pull/3284#discussion_r125371146) [](ancestors = 125371146)

suwc:
>         JsValueRef returnRef = moduleRecord->ModuleEvaluation();

It probably can be removed.

---
In reply to: [312751510](https://github.com/Microsoft/ChakraCore/pull/3284#issuecomment-312751510) [](ancestors = 312751510)

---
Refers to: lib/Jsrt/Core/JsrtCore.cpp:125 in a94be95. [](commit_id = a94be95f4ab1dabc03c7f6ddaaa85cd505d62ae8, deletion_comment = False)

bin/ch/WScriptJsrt.cpp
suwc: It's probably cleaner to return error in FetchImportedModuleHelper when it sees false here.

---
In reply to: [125370885](https://github.com/Microsoft/ChakraCore/pull/3284#discussion_r125370885) [](ancestors = 125370885,125370693)

bin/ch/WScriptJsrt.cpp
suwc: True. I will make this WIN32 only.

---
In reply to: [125370583](https://github.com/Microsoft/ChakraCore/pull/3284#discussion_r125370583) [](ancestors = 125370583)

bin/ch/WScriptJsrt.cpp
curtisman: You can have multiple thread with different JsRuntime/ThreadContext.   ch support testing with multiple thread and cross thread scenarios.

---
In reply to: [125371765](https://github.com/Microsoft/ChakraCore/pull/3284#discussion_r125371765) [](ancestors = 125371765,125371146)

suwc: Makes sense. I will get rid of these chdirs.

---
In reply to: [125382458](https://github.com/Microsoft/ChakraCore/pull/3284#discussion_r125382458) [](ancestors = 125382458,125371765,125371146)

bin/ch/WScriptJsrt.cpp
```diff
@@ -325,11 +334,46 @@ JsErrorCode WScriptJsrt::InitializeModuleInfo(JsValueRef specifier, JsModuleReco
...
+    else
+    {
+        // No fileName provided (WScript.LoadScript()), use dummy "script.js"
+        strcpy_s(fullPathNarrow, 10, "script.js");
+    }
+
+    return true;
+}
+
+char* WScriptJsrt::GetDir(LPCSTR fullPathNarrow, __out char* const fullDirNarrow)
```
akroshg: >char* [](start = 0, length = 5)

would it be better to return std::string instead of char*, as you are converting them at the caller side?

bin/ch/WScriptJsrt.cpp
suwc: Adding critical sections to lock global state updates in these hosts.
Confirmed with @bterlson that the lifetime of JsModuleRecord is tied to the Realm.
Yes, will have to figure out a way to clear moduleDirMap once the module record gets collected.

---
In reply to: [125369988](https://github.com/Microsoft/ChakraCore/pull/3284#discussion_r125369988) [](ancestors = 125369988)

bin/ch/WScriptJsrt.cpp
suwc: Removed this practice and updated all baselines.

---
In reply to: [125371926](https://github.com/Microsoft/ChakraCore/pull/3284#discussion_r125371926) [](ancestors = 125371926,125370583)

bin/ch/WScriptJsrt.cpp
suwc: Filed a separate test issues to drive the necessary changes in ch.exe's global states. Merging this PR to unblock test262 updates.

Issues:
[https://github.com/Microsoft/ChakraCore/issues/3319](https://github.com/Microsoft/ChakraCore/issues/3319)
[https://github.com/Microsoft/ChakraCore/issues/3318](https://github.com/Microsoft/ChakraCore/issues/3318)


---
In reply to: [126265384](https://github.com/Microsoft/ChakraCore/pull/3284#discussion_r126265384) [](ancestors = 126265384,125369988)

bin/ch/WScriptJsrt.cpp
```diff
@@ -325,12 +333,25 @@ JsErrorCode WScriptJsrt::InitializeModuleInfo(JsValueRef specifier, JsModuleReco
     return JsNoError;
 }
 
-JsErrorCode WScriptJsrt::LoadModuleFromString(LPCSTR fileName, LPCSTR fileContent)
+char* WScriptJsrt::GetDir(LPCSTR fullPathNarrow, __out char* const fullDirNarrow)
+{
+    char dir[_MAX_DIR];
+#pragma prefast(push)
+#pragma prefast(disable:6386, "_MAX_PATH(260) allocated for fullDirNarrow")
```
curtisman: You shouldn't need this if you SAL annotate fullDirNarrow

curtisman:
:shipit:

PR #3289, opened 2017-07-05
Allow runtests.cmd to be executed from any directory

agarwal-sandeep:
No need to be in tests directory explicitly.

### DIFF
--- a/test/runtests.cmd
+++ b/test/runtests.cmd
@@ -27,6 +27,12 @@
 @echo off
 setlocal
 
+rem <repo root>\test\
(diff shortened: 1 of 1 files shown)


test/runtests.cmd
```diff
@@ -27,6 +27,12 @@
 @echo off
 setlocal
 
+rem <repo root>\test\
+set script_dir=%~d0%~p0
```
jianchun: can be simpler "%~dp0%"

agarwal-sandeep:
Thanks @jianchun

PR #3290, opened 2017-07-06
Fix lambda function source extents for a couple of cases

boingoing:
When we convert a lambda function to a string, we decode characters from the UTF-8 source based on the extents of the function we calculated during parse-time. In some cases, the extents are wrong and we will crash during decode because the count of characters we decoded will not match the count of characters we expected to decode. This affects lambda functions rather than regular functions because curly braces are optional for lambdas which consist of a single expression and not all expressions handle trailing right-parens correctly. This change fixes three such cases.

Lambda consisting of a single question mark expression in which the false-branch is wrapped in parens:
`() => x ? 1 : (0)`

Lambda consisting of a single new expression in which the constructor is wrapped in parens:
`() => new (Boolean)`

Async lambda consisting of a single await expression where the argument to await is wrapped in parens:
`async() => await (5)`

### DIFF
--- a/lib/Parser/Parse.cpp
+++ b/lib/Parser/Parse.cpp
@@ -3141,7 +3141,7 @@ ParseNodePtr Parser::ParseTerm(BOOL fAllowCall,
         }
         else
         {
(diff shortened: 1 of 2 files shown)


boingoing:
Thanks, @dilijev

PR #3291, opened 2017-07-06
Fix #2809: Throw TypeError immediately when getter/setter is not callable.

### LINKED ISSUE #2809: test262/annexB: When getter/setter not callable, should not call toString on key because TypeError is thrown first
Failures (4 tests):
        test262/test/annexB/built-ins/Object/prototype/__defineGetter__/getter-non-callable.js: Expected SameValue("5", "0") to be true
        test262/test/annexB/built-ins/Object/prototype/__defineGetter__/this-non-obj.js: Expected SameValue("2", "0") to be true
        test262/test/annexB/built-ins/Object/prototype/__defineSetter__/setter-non-callable.js: Expected SameValue("5", "0") to be true
        test262/test/annexB/built-ins/Object/prototype/__defineSetter__/this-non-obj.js: Expected SameValue("2", "0") to be true

    Resources:
        https://tc39.github.io/ecma262/#sec-additional-properties-of-the-object.prototype-object
        https://tc39.github.io/ecma262/#sec-object.prototype.__defineGetter__ (step 2.)
        https://tc39.github.io/ecma262/#sec-object.prototype.__defineSetter__ (step 2.)

    Analysis:
        Should throw error as soon as
            2. If IsCallable(setter) is false throw TypeError exception
        We wait until after
            4. Let key be ? ToPropertyKey(P).

atulkatti:
Fixes #2809

### DIFF
--- a/lib/Runtime/Library/JavascriptObject.cpp
+++ b/lib/Runtime/Library/JavascriptObject.cpp
@@ -1302,12 +1302,11 @@ namespace Js
 #if ENABLE_COPYONACCESS_ARRAY
         JavascriptLibrary::CheckAndConvertCopyOnAccessNativeIntArray<Var>(args[0]);
 #endif
-        Var thisArg = JavascriptOperators::OP_GetThisNoFastPath(args[0], 0, scriptContext);
-        RecyclableObject* obj = RecyclableObject::FromVar(thisArg);
-
-        Var propertyKey = args.Info.Count > 1 ? args[1] : obj->GetLibrary()->GetUndefined();
-        const PropertyRecord* propertyRecord;
-        JavascriptConversion::ToPropertyKey(propertyKey, scriptContext, &propertyRecord);
+        RecyclableObject* obj = nullptr;
+        if (!JavascriptConversion::ToObject(args[0], scriptContext, &obj))
+        {
+            JavascriptError::ThrowTypeError(scriptContext, JSERR_This_NullOrUndefined, _u("Object.prototype.__defineGetter__"));
+        }
 
         Var getterFunc = args.Info.Count > 2 ? args[2] : obj->GetLibrary()->GetUndefined();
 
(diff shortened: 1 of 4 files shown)


atulkatti:
@Microsoft/chakra-developers

lib/Runtime/Library/JavascriptObject.cpp
```diff
@@ -1305,17 +1305,17 @@ namespace Js
...
 
-        Var propertyKey = args.Info.Count > 1 ? args[1] : obj->GetLibrary()->GetUndefined();
-        const PropertyRecord* propertyRecord;
-        JavascriptConversion::ToPropertyKey(propertyKey, scriptContext, &propertyRecord);
-
         Var getterFunc = args.Info.Count > 2 ? args[2] : obj->GetLibrary()->GetUndefined();
 
         if (!JavascriptConversion::IsCallable(getterFunc))
         {
             JavascriptError::ThrowTypeError(scriptContext, JSERR_FunctionArgument_NeedFunction, _u("Object.prototype.__defineGetter__"));
```
atulkatti: >ThrowTypeError [](start = 29, length = 14)

[https://github.com/Microsoft/ChakraCore/issues/2809](https://github.com/Microsoft/ChakraCore/issues/2809) 
The TypeError should be thrown earlier as per spec.

pleath: Is this the issue we were discussing yesterday?

atulkatti: Not exactly. It's the same GitHub issue I was looking at but this fix is fixing another set of test failures. I need to follow up with Brian on the issue we were talking about.

pleath: I see. What's the issue we're fixing here? Does something observable happen in JavascriptConversion::ToPropertyKey?

atulkatti: Yes, we end up calling ToString(). However, the according to the spec we should fail early if the function isn't callable. Because of this some test262 tests are failing.

---
In reply to: [126040046](https://github.com/Microsoft/ChakraCore/pull/3291#discussion_r126040046) [](ancestors = 126040046)

PR #3292, opened 2017-07-06
fix issue where call will be in helper block if JIT thunk allocation fails

MikeHolman:
We don't need to insert the helper label at all unless we are emitting range check.

Also, added a couple of config flags to help debugging.

OS#12460279

### DIFF
--- a/lib/Backend/LowerMDShared.cpp
+++ b/lib/Backend/LowerMDShared.cpp
@@ -5783,9 +5783,8 @@ LowererMD::GenerateNumberAllocation(IR::RegOpnd * opndDst, IR::Instr * instrInse
 void
 LowererMD::GenerateCFGCheck(IR::Opnd * entryPointOpnd, IR::Instr * insertBeforeInstr)
 {
-    bool useJITTrampoline = m_func->GetThreadContextInfo()->IsCFGEnabled();
-    IR::LabelInstr * callLabelInstr = IR::LabelInstr::New(Js::OpCode::Label, m_func);
-    IR::LabelInstr * cfgLabelInstr = IR::LabelInstr::New(Js::OpCode::Label, m_func, useJITTrampoline);
+    bool useJITTrampoline = CONFIG_FLAG(UseJITTrampoline);
+    IR::LabelInstr * callLabelInstr = nullptr;
     uintptr_t jitThunkStartAddress = NULL;
     if (useJITTrampoline)
     {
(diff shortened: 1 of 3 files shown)


lib/Backend/LowerMDShared.cpp
```diff
@@ -5856,12 +5859,15 @@ LowererMD::GenerateCFGCheck(IR::Opnd * entryPointOpnd, IR::Instr * insertBeforeI
...
-#if DBG
-        //Always generate CFG check in DBG build to make sure that the address is still valid
-        movInstrEntryPointToRegister->InsertBefore(callLabelInstr);
-#else
-        insertBeforeInstr->InsertBefore(callLabelInstr);
-#endif
+        if (CONFIG_FLAG(ForceJITCFGCheck))
+        {
+            // Always generate CFG check to make sure that the address is still valid
+            movInstrEntryPointToRegister->InsertBefore(callLabelInstr);
```
MikeHolman: this is a deviation from what we had before on debug build, but seems more valuable for codegen on debug/release to be close to each other by default.

Cellule: Is `ForceJITCFGCheck` a flag we use in our tools (exprgen, jsfunfuzz, crawler, etc...) ?
If not, I think we should add it to the pool of flag to test so we have coverage on this.

lib/Backend/LowerMDShared.cpp
```diff
@@ -5810,6 +5809,9 @@ LowererMD::GenerateCFGCheck(IR::Opnd * entryPointOpnd, IR::Instr * insertBeforeI
             IR::AddrOpnd * endAddressOfSegmentConstOpnd = IR::AddrOpnd::New(endAddressOfSegment, IR::AddrOpndKindDynamicMisc, m_func);
```
Cellule: Unrelated to this change, I noticed we add `InProcJITThunkEmitter::TotalThunkSize` to `jitThunkStartAddress`. It does look like it is "impossible" for this to overflow assuming the allocation was done properly. But since it is an assumption, we probably should `Assert(jitThunkStartAddress + InProcJITThunkEmitter::TotalThunkSize > jitThunkStartAddress)` or even do a runtime overflow check although I doubt it's necessary.

lib/Backend/LowerMDShared.cpp
MikeHolman: Yup, I plan on adding this, at least for exprgen.

MikeHolman:
@Cellule I thought I was following the new convention. I added message about OS bug in the message (which will be in the merge commit). Does it need to be in the title? Feels like it hurts readability having it in the PR title. And as it is for internal tracking only, it seems better to have that somewhere less prominent.

Cellule:
> @Cellule I thought I was following the new convention. I added message about OS bug in the message (which will be in the merge commit). Does it need to be in the title? Feels like it hurts readability having it in the PR title. And as it is for internal tracking only, it seems better to have that somewhere less prominent.
--

I am not a 100% sure about that. I don't know if we need the marker to be on the initial commit or if on the merge commit is good enough.
Either way, the syntax should be `OS#12460279` (using # and no spaces)

lib/Backend/LowerMDShared.cpp
Cellule: :+1:

lib/Backend/LowerMDShared.cpp
Cellule: :+1:

MikeHolman:
Ok, I changed formatting to OS#BugNumber.

PR #3298, opened 2017-07-07
Fix early exception throws in the presence of function calls within finally code

meg-gupta:
When there is an exception in tryfinally, we set the pendinfFinallyException object on the threadContext and bailout to the interpreter.
In the interpreter we read the pendingFinallyException from the threadContext and throw.
But if the finally code calls a function which has try finally, and bails out in the finally code due to a different reason,
we will end up reading the pendingFinallyExceptionObject on the threadContext and throw early.
With this change, we read the pendingFinallyExceptionObject from the threadContext before running finally code.


meg-gupta:
ah ok! Thanks for the review.

PR #3301, opened 2017-07-07
Clear the auto restore data if we have a valid function already

aneeshdk:
We don't have to track the restore data if the function body is valid.


aneeshdk:

PR #3303, opened 2017-07-07
OS#12416927: Fix line ending of test\Intl\IntlTaintingPreInitTests.baseline to be CRLF

dilijev:
The test baseline was apparently checked-in with LF instead of the intended CRLF, which was fine for most things, which will normalize line endings of files not marked as binary. Nightlies apparently do not do this normalization.

```
$ xxd test/Intl/IntlTaintingPreInitTests.baseline
00000000: 5061 7373 6564 2070 7265 2d69 6e69 7420  Passed pre-init
00000010: 7461 696e 7469 6e67 210d 0a              tainting!..
```

(Note the `0d 0a` at the end, before was only `0a`.)


dilijev:
@tcare @atulkatti @digitalinfinity @curtisman Didn't we used to have a gitattributes to mark baseline files as binary and not do this conversion, either when committing or checking-out?

tcare:
Looks like it was removed a few months ago accidentally: https://github.com/Microsoft/ChakraCore/commit/02a02c3a59f4f281d5817ca86358e4d293ec13c4

Feel free to add it back.

dilijev:
Makes sense. We should probably take a closer look at those types of merges to make sure we don't lose any changes or content.

tcare:
LGTM

dilijev:
Restoring .gitattributes caused a bunch of tests to start failing -- probably baselines which were checked in since .gitattributes changed. See: https://ci2.dot.net/job/Microsoft_ChakraCore/job/release_1.6/job/x64_debug_prtest/341/

I'm removing that commit and checking in the minimal fix here for now. Will open another issue.

Issue #3256, opened 2017-06-28
Module import/export should not be allowed in the statement of an if/else/while/etc.


bterlson:
```js
// some module.js
if (false) export default null;
```
Actual: no error
Expected: early syntax error

This also applies to (at least) the else block, the body of loop constructs, and labels.

Issue #3255, opened 2017-06-28
Modules should not allow `return` at top level


bterlson:
```js
// some module.js
return;
```
Actual: bails out of the module (kind of handy actually)
Expected: Early syntax error

Issue #3252, opened 2017-06-28
Module namespace objects should not have Symbol.iterator property


bterlson:
```js
// bug-dep.js
import * as ns from './bug-dep.js';
print(Object.prototype.hasOwnProperty.call(ns, Symbol.iterator));
```

Actual: true
Expected: false ([spec changed](https://github.com/tc39/ecma262/commit/814463f4819f30fe2bc2db56a00b63e3f76167af))

Issue #3246, opened 2017-06-28
Module namespace entries are not sorted properly


bterlson:
When constructing a module namespace object, the keys [should be ordered](https://tc39.github.io/ecma262/#sec-modulenamespacecreate) as if passed to Array.prototype.sort with the default comparator.

```js
// root.js
WScript.LoadModule(`import "bug-dep.js"`);

// bug-dep.js
var x;
export { x as π }; // u03c0
export { x as az };
export { x as __ };
export { x as za };
export { x as Z };
export { x as \u03bc };
export { x as z };
export { x as zz };
export { x as a };
export { x as A };
export { x as aa }
export { x as λ }; // u03bb
export { x as _ };
export { x as $$ };
export { x as $ };
export default null;

import * as ns from './bug-dep.js';

var stringKeys = Object.getOwnPropertyNames(ns);
print(JSON.stringify(stringKeys));
```
Actual: ["default","$","$$","\_","?","aa","A","a","zz","z","?","Z","za","\_\_","az","?"] (reverse order?)
Expected: ["$","$$","A","Z","\_","__","a","aa","az","default","z","za","zz","?","?","?"] (declaration order)

(Note: ?'s above are due to console output, in expected output the ordering is \u03bb, \u03bc, \u03c0)

PR #3307, opened 2017-07-09
Issues#3256,#3255,#3252 Module bugfixes

suwc:
- Module import/export in if/else/while/for or function block should be syntax error (#3256)
 - Module should not allow 'return at top-level (#3255)
 - Module namespace should not have Symbol.iterator property (#3252)

### DIFF
--- a/lib/Parser/Parse.cpp
+++ b/lib/Parser/Parse.cpp
@@ -131,6 +131,8 @@ Parser::Parser(Js::ScriptContext* scriptContext, BOOL strictMode, PageAllocator 
     m_deferEllipsisError = false;
     m_hasDeferredShorthandInitError = false;
     m_parsingSuperRestrictionState = ParsingSuperRestrictionState_SuperDisallowed;
+
+    m_disallowImportExportStmt = false;
 }
 
 Parser::~Parser(void)
@@ -2446,7 +2448,14 @@ bool Parser::IsImportOrExportStatementValidHere()
         && curFunc->sxFnc.IsModule()
         && this->m_currentBlockInfo->pnodeBlock == curFunc->sxFnc.pnodeBodyScope
         && (this->m_grfscr & fscrEvalCode) != fscrEvalCode
-        && this->m_tryCatchOrFinallyDepth == 0;
+        && this->m_tryCatchOrFinallyDepth == 0
+        && !this->m_disallowImportExportStmt;
+}
+
+bool Parser::IsTopLevelModuleFunc()
+{
+    ParseNodePtr curFunc = GetCurrentFunctionNode();
+    return curFunc->nop == knopFncDecl && curFunc->sxFnc.IsModule();
 }
 
 template<bool buildAST> ParseNodePtr Parser::ParseImportCall()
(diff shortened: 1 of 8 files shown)


curtisman:
>     ListForListIterator* ModuleNamespace::EnsureSortedExportedNames()

No longer have any reference?  Remove?

---
Refers to: lib/Runtime/Language/ModuleNamespace.cpp:306 in 6ce8c61. [](commit_id = 6ce8c612f4ebb21393b61c148da73a93f8ef9f17, deletion_comment = False)

lib/Parser/Parse.cpp
```diff
@@ -10106,7 +10115,7 @@ ParseNodePtr Parser::ParseStatement()
     {
         if (buildAST)
         {
-            if (nullptr == m_currentNodeFunc)
+            if (nullptr == m_currentNodeFunc || IsImportOrExportStatementValidHere())
```
curtisman: This looks weird to me. 
Is this saying that `return` is an error, but 'try { return; } catch {}` is not?

curtisman:
>         ParseNodePtr pnodeBody = ParseStatement<buildAST>();

How about this one?


---
Refers to: lib/Parser/Parse.cpp:9795 in 6ce8c61. [](commit_id = 6ce8c612f4ebb21393b61c148da73a93f8ef9f17, deletion_comment = False)

lib/Runtime/Language/ModuleNamespace.cpp
```diff
@@ -338,28 +330,4 @@ namespace Js
         }
         return sortedExportedNames;
     }
-
-    Var ModuleNamespace::EntrySymbolIterator(RecyclableObject* function, CallInfo callInfo, ...)
```
boingoing: So without this, does property iteration still work just by doing an ordinary walk across the type handler?

suwc: Yes, property iteration still works.

boingoing: Does that affect the iteration order or does that continue to work because we already sorted them ?

suwc: This alone doesn't affect the iteration order. We have a separate issue to address the iteration order: [https://github.com/Microsoft/ChakraCore/issues/3246](https://github.com/Microsoft/ChakraCore/issues/3246)

---
In reply to: [126488250](https://github.com/Microsoft/ChakraCore/pull/3307#discussion_r126488250) [](ancestors = 126488250)

lib/Parser/Parse.cpp
suwc: Clarified 'try { return }' and others (except inside a function body) should be treated as top-level return. Fixed in 2nd iteration.

---
In reply to: [126453545](https://github.com/Microsoft/ChakraCore/pull/3307#discussion_r126453545) [](ancestors = 126453545)

suwc:
Any other comments?

boingoing:
Looks good to me. I wondered if `GetCurrentFunctionNode()` can return null but I see it used all over the parser with loose checking so it's probably fine.

pleath:
GetCurrentFunctionNode returns the current enclosing function node, of which there's always one, even if it's the one that encloses the global statements.

PR #3309, opened 2017-07-10
Fixes #2471 Symbol registration map should support null characters in symbol names.

### LINKED ISSUE #2471: Symbol.for and strings with embedded nulls
Symbol.for internally uses wcscmp for comparing contents of strings used as keys. Thus the following prints  `true` when it should be `false`:
```
console.log(Symbol.for('A\0Z') === Symbol.for('A\0z'))
```
Also verified that v8 prints `false`.

atulkatti:


lib/Runtime/Base/ThreadContext.cpp
```diff
@@ -4291,15 +4293,18 @@ const Js::PropertyRecord* ThreadContext::AddSymbolToRegistrationMap(const char16
 
     Assert(propertyRecord);
 
-    // The key is the PropertyRecord's buffer (the PropertyRecord itself) which is being pinned as long as it's in this map.
-    // If that's ever not the case, we'll need to duplicate the key here and put that in the map instead.
-    this->recyclableData->symbolRegistrationMap->Add(propertyRecord->GetBuffer(), propertyRecord);
+    // We need to support null characters in the Symbol names. For e.g. "A\0Z" is a valid symbol name and is different than "A\0Y".
```
agarwal-sandeep: Do we already have tests? if no please add some.

akroshg:
LGTM - please add tests for both APIs

PR #3310, opened 2017-07-10
xplatInterval: remove exclude_jenkins

### LINKED ISSUE #3010: [Disabled Test] test/Date/xplatInterval.js disabled in Jenkins. Is this by design?
See `test\Date\rlexe.xml`

`test/Date/xplatInterval.js` is marked `exclude_jenkins` with no explanation. It does use `Date` but it doesn't seem to be timezone sensitive at first glance. Is this another instance of #319 or something else?

Please update the test so that it can run in Jenkins or explain in a comment in the `rlexe.xml` why it cannot.

obastemur:
Fixes #3010

### DIFF
--- a/test/Date/rlexe.xml
+++ b/test/Date/rlexe.xml
@@ -120,8 +120,7 @@
   <test>
     <default>
       <files>xplatInterval.js</files>
-      <!-- Fix or add explanation for exclude_jenkins (Microsoft/ChakraCore#3010) -->
-      <tags>exclude_jenkins,Slow</tags>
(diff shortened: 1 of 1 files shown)


dilijev:
Looks like this is just re-enabling the test under Jenkins.
Was the problem just that this test was Slow, or has the underlying problem been resolved previously?

obastemur:
@dilijev https://github.com/Microsoft/ChakraCore/issues/3010#issuecomment-304022862

obastemur:

PR #3312, opened 2017-07-10
Fix proxy method SetProperty

meg-gupta:
Pass requestContext to handle Var marshalling correctly.
Fixes OS#9941606


akroshg:
Can you add a test case? Seems fine to me but JC has done similar fixes around proxy for the request context. Can you check with him?

meg-gupta:
Added testcase. Thanks for reviewing! Thanks @tcare  for the testcase!

PR #3315, opened 2017-07-11
fix oop jit assert on platforms without cfg

MikeHolman:
there was some inconsistency with JIT thunk allocation in case OOP JIT is enabled but running on platform without CFG


MikeHolman:
Note: this is causing breaks on ChakraFull CI.

PR #3317, opened 2017-07-11
IntlICU: remove invalid assert: this failure is thrown as a JS RangeError.

dilijev:


MSLaguana:
How does this show up as a JS range error? I just took a quick look at the code, and I could only see that it would return `undefined` rather than throw an exception. This is based on the call at  https://github.com/dilijev/ChakraCore/blob/586c154650dc27a7bd3a6dc8e31c1e167eaabacf/lib/Runtime/Library/IntlEngineInterfaceExtensionObject.cpp#L606 which returns undefined at https://github.com/dilijev/ChakraCore/blob/586c154650dc27a7bd3a6dc8e31c1e167eaabacf/lib/Runtime/Library/IntlEngineInterfaceExtensionObject.cpp#L620 if there is any error.

dilijev:
Looks like you might be looking at the wrong call. This is for `IcuIntlAdapter::IsWellFormedLanguageTag`, not `IcuIntlAdapter::NormalizeLanguageTag`.

Here's the trace:

```
bool IcuIntlAdapter::IsWellFormedLanguageTag(...)
{

...
        if (!success)
        {
            goto cleanup;
        }
...
cleanup:
        return success;

}


// Called from (C++ call site):

Var IntlEngineInterfaceExtensionObject::EntryIntl_IsWellFormedLanguageTag(...)
{
...
        return TO_JSBOOL(scriptContext, IcuIntlAdapter::IsWellFormedLanguageTag(argString->GetSz(), argString->GetLength()));
...
}


// which is called from: Intl.js call site:

    function CanonicalizeLocaleList(locales) {
        ...
	    if (!platform.isWellFormedLanguageTag(tag)) {
                platform.raiseLocaleNotWellFormed(String(tag));
            }
        ...
    }


// Which throws RangeError with JSERR_LocaleNotWellFormed (jndirectfields.h, rterrors.h)
```


Test case that this fixes:

```
try {
    print(Intl.getCanonicalLocales(['中文']));
} catch (e) {
    print(e);
        console.log("!!! Expected RangeError");
    }
}
```

MSLaguana:
Ah you are right, I was looking at the wrong place. Seems fine to me then!

PR #3321, opened 2017-07-11
Set hasAnyWriteToFormals on the correct function node

rajatd:
Set hasAnyWriteToFormals on the correct function node; potentially unlocking wins from arguments optimization.

### DIFF
--- a/lib/Parser/Parse.cpp
+++ b/lib/Parser/Parse.cpp
@@ -1716,9 +1716,9 @@ void Parser::BindPidRefsInScope(IdentPtr pid, Symbol *sym, int blockId, uint max
         if (ref->IsAssignment())
         {
             sym->PromoteAssignmentState();
-            if (m_currentNodeFunc && sym->GetIsFormal())
+            if (sym->GetIsFormal())
             {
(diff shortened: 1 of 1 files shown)


lib/Parser/Parse.cpp
```diff
@@ -1716,9 +1716,9 @@ void Parser::BindPidRefsInScope(IdentPtr pid, Symbol *sym, int blockId, uint max
         if (ref->IsAssignment())
         {
             sym->PromoteAssignmentState();
-            if (m_currentNodeFunc && sym->GetIsFormal())
+            if (sym->GetIsFormal())
             {
-                m_currentNodeFunc->sxFnc.SetHasAnyWriteToFormals(true);
+                GetCurrentFunctionNode()->sxFnc.SetHasAnyWriteToFormals(true);
```
akroshg: LGTM - how HasAnyWriteToFormals used?

rajatd: It is one of the first checks done to decide whether we can optimize away arguments creation on the heap, aka, stack args optimization.

PR #3322, opened 2017-07-12
OS12507864: Correct assertion introduced in PR#3211

suwc:
The assertions are intended to future-proof the assumption that
isInlineSegment being true implies segment swap.
https://github.com/Microsoft/ChakraCore/pull/3211


agarwal-sandeep:
:shipit:

PR #3323, opened 2017-07-12
Fix #3203, Fix #3204: Intl.getCanonicalLocales: name, toString, cannot call with new

### LINKED ISSUE #3203: `new Intl.getCanonicalLocales()` should fail with `not a constructor`
For example:

```
new Intl.DateTimeFormat.supportedLocalesOf();
```

outputs:

```
TypeError: Function 'Intl.DateTimeFormat.supportedLocalesOf' is not a constructor
   at Global code (e:\work\intl-icu-xplat\test\iswellformedlanguagetag.js:2:1)
```

Similar should happen for

```
new Intl.getCanonicalLocales();
```

Instead, an object is constructed.

### LINKED ISSUE #3204: Name of function `Intl.getCanonicalLocales` is incorrectly `CanonicalizeLocaleList` -- should be hidden.
```
print(Intl.DateTimeFormat.supportedLocalesOf);
print(Intl.getCanonicalLocales);
```

```
function() {
    [native code]
}
function CanonicalizeLocaleList() { [native code] }
```

Should do something similar here to hide name of the implementing function as we did for `supportedLocalesOf`.

dilijev:
Fixes #3203
Fixes #3204


dilijev:
I'm working on adding a test.

dilijev:
@Cellule please review RegenAllByteCodeNoBuild.cmd -- I've been using this to speed up my inner loop while changing Intl.js and I think it will come in handy for others in the future.

RegenAllByteCodeNoBuild.cmd
```diff
@@ -0,0 +1,62 @@
+::-------------------------------------------------------------------------------------------------------
+:: Copyright (C) Microsoft. All rights reserved.
+:: Licensed under the MIT license. See LICENSE.txt file in the project root for full license information.
+::-------------------------------------------------------------------------------------------------------
+
+:: WARNING: be careful when using this script as it assumes that
+:: you already have bytecode-format-compatible builds for all required flavors.
+:: This script helps speed things up when you are only making changes to
```
Cellule: nit: feels like something is missing at the end of this sentence.
`... making changes to scripts, e.g. Intl.js` ?

RegenAllByteCodeNoBuild.cmd
```diff
@@ -0,0 +1,62 @@
+::-------------------------------------------------------------------------------------------------------
```
Cellule: Wouldn't it be better if `RegenAllByteCode.cmd` called this script after building everything needed ?

Also, I was thinking we probably should move these scripts to the `tools` folder instead of leaving them at the root. (Would require a fix of the regen script in chakra full). If you don't want to do this now, we can create an issue and I'll clean it up later.

RegenAllByteCodeNoBuild.cmd
```diff
@@ -0,0 +1,62 @@
...
+    if [%2] == [nojit] (
+      set _nojit=.nojit
+    )
+
+    pushd lib\Runtime\Library\InJavascript
+      set _script=Intl.js
+      set _header=%_script%%_nojit%.bc.%_bits%.h
+      set _bin=%_reporoot%\Build\VcBuild%_nojit%\bin\%_arch%_debug\ch.exe
+
+      echo %_bin% -GenerateLibraryByteCodeHeader:%_header% -Intl %_script%
```
Cellule: Why not call `GenByteCode.cmd` ?

dilijev: Ah good point. I thought the GenByteCode.cmd script forced a build. Made the change.

RegenAllByteCodeNoBuild.cmd
dilijev: Fixed.

RegenAllByteCodeNoBuild.cmd
dilijev: Yeah I agree with this. I think both the refactoring and the script move are good. 
Added https://github.com/Microsoft/ChakraCore/issues/3330

PR #3324, opened 2017-07-12
clean up asm.js/wasm entrypoint management

MikeHolman:
There were a few places where asm.js strayed needlessly from how js manages entrypoints.

In case of NativeAddress usage, there was a race condition that could lead to us calling direct JIT address rather than the JIT thunk, leading to CFG failure.

It also looks like we may have been leaking old entrypoints.

OS: 12228631


Cellule:
Like we talked yesterday, this is most likely breaking templatizedJit even further. But, we need to do a rework of how we handle entry points in asmjs/wasm anyway.
So as long as this fixes your bug, LGTM

PR #3327, opened 2017-07-12
xplat: fix absolute path for cc-toolchain

obastemur:


obastemur:
@kfarnung @MSLaguana Thanks for the reviews

Issue #3217, opened 2017-06-23
Reflect.construct permanently corrupts the invoked constructor


shicks:
Repro:

```
<pre id="log"></pre>
<script>
function log(msg) { document.getElementById('log').textContent += msg + '\n'; }

function Base() {}
function Derived() {}
Derived.prototype = Object.create(Base.prototype);  // OPTIONAL

log(new Base() instanceof Derived);  // correctly logs 'false'
Reflect.construct(Base, [], Derived);
log(new Base() instanceof Derived);  // from now on, logs 'true'!
</script>
```

This bug is pretty elusive - if you ever open the developer tools (even just for a second) everything goes back to normal, and the `Base` constructor again constructs `Base` objects (hence the HTML logging).

It's also irrelevant whether `Derived` is a prototypal subclass of `Base`.  Fun fact: if the `Object.create` call is left out, then in the broken state, `new Base() instanceof Base` actually returns false!

This seems to be a one-way switch: calling `Reflect.construct(Base, [], Other)` afterwards does *not* switch `new Base()` from creating `Derived`s to creating `Other`s.  But it also doesn't work to try to workaround the issue with a proactive `Reflect.construct(Base, [], Base)`, since a future call to `Reflect.construct(Base, [], Derived)` will still cause problems.

EDIT: Also, this bug only triggers if the base class has been instantiated at least once before calling `Reflect.construct`: so if the first log statement is removed then the second one logs 'false'.

PR #3328, opened 2017-07-12
Fix Issue#3217: Reflect.construct permanently corrupts the invoked constructor

suwc:
With the optional newTarget argument, Reflect.construct has a legitimate
user case where cached constructor prototype mismatches prototype of
newly created object. Therefore it is necessary to tighten requirement for
sharing type by checking if constructor prototypes match.
Confirmed no perf impact.

Fixes #3328

### DIFF
--- a/lib/Runtime/Language/JavascriptOperators.cpp
+++ b/lib/Runtime/Language/JavascriptOperators.cpp
@@ -6247,7 +6247,7 @@ CommonNumber:
                 if (typeHandler->IsSharable())
                 {
 #if DBG
-                    bool cachedProtoCanBeCached;
+                    bool cachedProtoCanBeCached = false;
                     Assert(type->GetPrototype() == JavascriptOperators::GetPrototypeObjectForConstructorCache(constructor, requestContext, cachedProtoCanBeCached));
                     Assert(cachedProtoCanBeCached);
                     Assert(type->GetScriptContext() == constructorCache->GetScriptContext());
--- a/lib/Runtime/Library/JavascriptFunction.cpp
+++ b/lib/Runtime/Library/JavascriptFunction.cpp
@@ -938,13 +938,15 @@ namespace Js
                 resultObject,
                 JavascriptFunction::Is(functionObj) && functionObj->GetScriptContext() == scriptContext ?
                 JavascriptFunction::FromVar(functionObj) :
-                nullptr);
+                nullptr,
+                overridingNewTarget != nullptr);
     }
 
     Var JavascriptFunction::FinishConstructor(
         const Var constructorReturnValue,
         Var newObject,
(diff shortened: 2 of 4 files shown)


suwc:
@Microsoft/chakra-developers please review. Thanks!

lib/Runtime/Language/JavascriptOperators.cpp
```diff
@@ -6241,12 +6241,12 @@ namespace Js
 
             if (constructorBody->GetHasOnlyThisStmts())
             {
-                if (typeHandler->IsSharable())
+                bool cachedProtoCanBeCached;
```
agarwal-sandeep: Initialize please

suwc:
Addresses https://github.com/Microsoft/ChakraCore/issues/3217

lib/Runtime/Language/JavascriptOperators.cpp
```diff
@@ -6243,63 +6243,66 @@ namespace Js
             {
                 if (typeHandler->IsSharable())
                 {
+                    bool cachedProtoCanBeCached = nullptr;
+                    if (type->GetPrototype() == JavascriptOperators::GetPrototypeObjectForConstructorCache(constructor, requestContext, cachedProtoCanBeCached)
```
curtisman: Should we check if we have `overridingNewTarget` in JavascriptFunction::CallAsConstructor instead?

suwc: Would that shut cases with 'overridingNewTarget' completely out of the constructor cache, even if the cache is not being used in the first place?

---
In reply to: [127739790](https://github.com/Microsoft/ChakraCore/pull/3328#discussion_r127739790) [](ancestors = 127739790)

lib/Runtime/Language/JavascriptOperators.cpp
```diff
@@ -6243,63 +6243,66 @@ namespace Js
             {
                 if (typeHandler->IsSharable())
                 {
+                    bool cachedProtoCanBeCached = nullptr;
```
akroshg: >nullptr [](start = 50, length = 7)

false?

lib/Runtime/Language/JavascriptOperators.cpp
```diff
@@ -6243,63 +6243,66 @@ namespace Js
...
+                    if (type->GetPrototype() == JavascriptOperators::GetPrototypeObjectForConstructorCache(constructor, requestContext, cachedProtoCanBeCached)
+                        && cachedProtoCanBeCached)
+                    {
+
 #if DBG
```
akroshg: nit: remove this #f if the body has only asserts.

kunalspathak:
:shipit:

PR #3331, opened 2017-07-12
fix bug in BrBool peep where cmp2Src==cmp1Dst

MikeHolman:
Peep stops us from storing to the dst, so we shouldn't do the peep if result is used

OS: 12444855


rajatd:
LGTM

PR #3333, opened 2017-07-13
WASM - int64 split live on back edge fix

Cellule:
Make sure to mark all int64 split symbol as Live on back edge if the original int64 was also live.

The previous solution to mark them as live when we lower the loop top wasn't working because it is possible for a symbol to be live on back edge, but not seen in lower yet.
```js
for(...) {
  int64 myInt64Var;
  for(...) { // myInt64Var is liveOnBackEdge, but not processed yet (backward pass in lowerer)
    // Need to know that the high and low part of myInt64Var is live on back edge to do compensation in RegAlloc
  }
}
```

Move the int64 split map to Func allowing to use post lower.
Also allowing some lowering methods to be static which was prevented before because it needed an instance of lowerer.

Fixes OS#12708317

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/3333)

### DIFF
--- a/lib/Backend/Func.cpp
+++ b/lib/Backend/Func.cpp
@@ -143,6 +143,9 @@ Func::Func(JitArenaAllocator *alloc, JITTimeWorkItem * workItem,
     , m_forInEnumeratorArrayOffset(-1)
     , argInsCount(0)
     , m_globalObjTypeSpecFldInfoArray(nullptr)
+#if LOWER_SPLIT_INT64
+    , m_int64SymPairMap(nullptr)
+#endif
 #ifdef RECYCLER_WRITE_BARRIER_JIT
     , m_lowerer(nullptr)
 #endif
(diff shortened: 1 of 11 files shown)


meg-gupta:
@Cellule  In your example, is there use/def of myInt64Var in the inner loop ? If there's a simple test case for this, can you please add ?

Cellule:
@meg-gupta No there is no use of the var in the inner loop at lower time. If there was, the previous mechanism would solve the issue. There is also no uses of it later.
The bug occurs when there is a spill of the variable in the inner loop hence adding a use at register allocator time

Unfortunately, since it's a register allocator bug, it is really hard to write a reliable test case.

meg-gupta:
Ah ok. But if its getting added to liveOnBackEdge at regalloc time, how is this info available during SccLiveness for use in `func->Int64SplitExtendLoopLifetime(loop)` called from `SccLiveness::Build`
. Don't we call SccLiveness only once before regalloc to number live ranges ?

Cellule:
It's the other way around.
`func->Int64SplitExtendLoopLifetime(loop)` is the thing that adds it to liveOnBackEdge during SCCLiveness so it can then be used further in SCCLiveness and in RegAlloc

Cellule:
FYI, after talking with Louis, we decided to not allow creating new int64 sym map post lower. So I effectively only changed an assert condition to be more conservative

Cellule:

PR #3334, opened 2017-07-13
Add assert too Opnd::Free to ensure it is not in use

sigatrev:


meg-gupta:
LGTM, as long as all tests are run for a long time.

LouisLaf:
:shipit:

PR #3336, opened 2017-07-13
Fix Issue#3064/#3423: Cache conflicts in super property access

### LINKED ISSUE #3064: Caching Error in Chakra
The following code behaves incorrectly (inconsistent with the standard and other engines) in Chakra due to a missing cache update:

```
  function Base() { }

  Base.prototype = {
    f() {
      return "Base " + this.toString();
    },
    x: 15,
    toString() {
      return "this is Base";
    }
  };

  function Derived() {
    this.derivedDataProperty = "xxx";
  }
  Derived.prototype = {
    __proto__: Base.prototype,
    toString() { return "this is Derived"; },
    x: 27,
    f() {

      var a = super.x;

      print(this.x);
      return "Derived";
    }
  };

  print(new Base().f());
  print(new Derived().f());
```

It should print 27 as the value of Derived.x, but instead prints 15. This code hits an assert in the debug build:

ASSERTION 5639: (/home/user/test_everywhere/ChakraCore/lib/Runtime/Types/TypePropertyCache.cpp, line 277) *propertyValue == JavascriptOperators::GetProperty(propertyObject, propertyId, requestContext)
 Failure: (*propertyValue == JavascriptOperators::GetProperty(propertyObject, propertyId, requestContext))
Illegal instruction (core dumped)

### LINKED ISSUE #3423: Repeated calls to class setter triggers assertion
```javascript
class B {
    set x(v) { print("B.prototype.set x"); this._x = v; }
    get x() { print("B.prototype.get x"); return this._x; }
}

class A extends B {
    set x(v) { print("A.prototype.set x"); super.x = v + 100; }
    get x() { print("A.prototype.get x"); return super.x; }
}

var a = new A();
a.x = 100;
print(a.x);

var a1 = new A();
a1.x = 100;
print(a1.x);
```
Expected:
```
A.prototype.set x
B.prototype.set x
A.prototype.get x
B.prototype.get x
200
A.prototype.set x
B.prototype.set x
A.prototype.get x
B.prototype.get x
200
```
Actual
```
A.prototype.set x
B.prototype.set x
A.prototype.get x
B.prototype.get x
200
A.prototype.set x
ASSERTION 15068: (lib\runtime\Language/InlineCache.inl, line 356) setterValue == nullptr || setterValue == function
 Failure: (setterValue == nullptr || setterValue == function)
```

suwc:
Accesses to super properties are cached on 'this' object (vs. the
"super" object), causing conflict between e.g. super.x and this.x.
Similar conflicts cause Issue#3423 for GetProperty cases.
Fix by adding 'isSuper' flag to use appropriate object for caching.

Fixes #3064, Fixes #3423

### DIFF
--- a/lib/Runtime/Language/JavascriptOperators.cpp
+++ b/lib/Runtime/Language/JavascriptOperators.cpp
@@ -1739,7 +1739,7 @@ CommonNumber:
             //         Also we might want to throw here instead of checking it again in the caller
             if (value && !requestContext->IsUndeclBlockVar(*value) && !WithScopeObject::Is(object))
             {
-                CacheOperators::CachePropertyRead(instance, object, isRoot, propertyId, false, info, requestContext);
+                CacheOperators::CachePropertyRead(propertyObject, object, isRoot, propertyId, false, info, requestContext);
             }
 #ifdef TELEMETRY_JSO
             if (TELEMETRY_PROPERTY_OPCODE_FILTER(propertyId))
@@ -1779,7 +1779,7 @@ CommonNumber:
                 }
 
                 PropertyValueInfo::Set(info, requestContext->GetLibrary()->GetMissingPropertyHolder(), 0);
-                CacheOperators::CachePropertyRead(instance, requestContext->GetLibrary()->GetMissingPropertyHolder(), isRoot, propertyId, true, info, requestContext);
+                CacheOperators::CachePropertyRead(propertyObject, requestContext->GetLibrary()->GetMissingPropertyHolder(), isRoot, propertyId, true, info, requestContext);
             }
 #if defined(TELEMETRY_JSO) || defined(TELEMETRY_AddToCache) // enabled for `TELEMETRY_AddToCache`, because this is the property-not-found codepath where the normal TELEMETRY_AddToCache code wouldn't be executed.
             if (TELEMETRY_PROPERTY_OPCODE_FILTER(propertyId))
@@ -7274,7 +7274,7 @@ CommonNumber:
         PropertyValueInfo::SetCacheInfo(&info, functionBody, inlineCache, inlineCacheIndex, !IsFromFullJit);
         Var value;
         if (CacheOperators::TryGetProperty<true, true, true, true, true, true, !TInlineCache::IsPolymorphic, TInlineCache::IsPolymorphic, false>(
-                thisInstance, false, object, propertyId, &value, scriptContext, nullptr, &info))
+                instance, false, object, propertyId, &value, scriptContext, nullptr, &info))
         {
             return value;
         }
@@ -7930,55 +7930,7 @@ CommonNumber:
     template <bool IsFromFullJit, class TInlineCache>
     inline void JavascriptOperators::PatchPutValueNoLocalFastPath(FunctionBody *const functionBody, TInlineCache *const inlineCache, const InlineCacheIndex inlineCacheIndex, Var instance, PropertyId propertyId, Var newValue, PropertyOperationFlags flags)
     {
-        ScriptContext *const scriptContext = functionBody->GetScriptContext();
-
-        if (TaggedNumber::Is(instance))
-        {
-            JavascriptOperators::SetPropertyOnTaggedNumber(instance,
-                                        nullptr,
-                                        propertyId,
-                                        newValue,
-                                        scriptContext,
-                                        flags);
-             return;
-        }
-#if ENABLE_COPYONACCESS_ARRAY
-        JavascriptLibrary::CheckAndConvertCopyOnAccessNativeIntArray<Var>(instance);
-#endif
-        RecyclableObject *object = RecyclableObject::FromVar(instance);
-
-        PropertyValueInfo info;
-        PropertyValueInfo::SetCacheInfo(&info, functionBody, inlineCache, inlineCacheIndex, !IsFromFullJit);
-        if (CacheOperators::TrySetProperty<!TInlineCache::IsPolymorphic, true, true, true, true, !TInlineCache::IsPolymorphic, TInlineCache::IsPolymorphic, false>(
-                object, false, propertyId, newValue, scriptContext, flags, nullptr, &info))
-        {
-            return;
-        }
-
-#if DBG_DUMP
-        if (PHASE_VERBOSE_TRACE1(Js::InlineCachePhase))
-        {
-            CacheOperators::TraceCache(inlineCache, _u("PatchPutValueNoLocalFastPath"), propertyId, scriptContext, object);
-        }
-#endif
-
-        ImplicitCallFlags prevImplicitCallFlags = ImplicitCall_None;
-        ImplicitCallFlags currImplicitCallFlags = ImplicitCall_None;
-        bool hasThisOnlyStatements = functionBody->GetHasOnlyThisStmts();
-        if (hasThisOnlyStatements)
-        {
-            prevImplicitCallFlags = CacheAndClearImplicitBit(scriptContext);
-        }
-        if (!JavascriptOperators::OP_SetProperty(instance, propertyId, newValue, scriptContext, &info, flags))
-        {
-            // Add implicit call flags, to bail out if field copy prop may propagate the wrong value.
-            scriptContext->GetThreadContext()->AddImplicitCallFlags(ImplicitCall_NoOpSet);
-        }
-        if (hasThisOnlyStatements)
-        {
-            currImplicitCallFlags = CheckAndUpdateFunctionBodyWithImplicitFlag(functionBody);
-            RestoreImplicitFlag(scriptContext, prevImplicitCallFlags, currImplicitCallFlags);
-        }
+        PatchPutValueWithThisPtrNoLocalFastPath<IsFromFullJit, TInlineCache>(functionBody, inlineCache, inlineCacheIndex, instance, propertyId, newValue, instance, flags);
     }
     template void JavascriptOperators::PatchPutValueNoLocalFastPath<false, InlineCache>(FunctionBody *const functionBody, InlineCache *const inlineCache, const InlineCacheIndex inlineCacheIndex, Var instance, PropertyId propertyId, Var newValue, PropertyOperationFlags flags);
     template void JavascriptOperators::PatchPutValueNoLocalFastPath<true, InlineCache>(FunctionBody *const functionBody, InlineCache *const inlineCache, const InlineCacheIndex inlineCacheIndex, Var instance, PropertyId propertyId, Var newValue, PropertyOperationFlags flags);
--- a/lib/Backend/IRBuilder.cpp
+++ b/lib/Backend/IRBuilder.cpp
@@ -4648,7 +4648,7 @@ IRBuilder::BuildElementC2(Js::OpCode newOpcode, uint32 offset, Js::RegSlot insta
             value2Opnd = this->BuildDstOpnd(value2Slot);
             regOpnd = this->BuildDstOpnd(regSlot);
 
-            instr = IR::Instr::New(newOpcode, regOpnd, fieldSymOpnd, value2Opnd, m_func);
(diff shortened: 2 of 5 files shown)


suwc:
Addresses https://github.com/Microsoft/ChakraCore/issues/3064

suwc:
@Microsoft/chakra-developers @Microsoft/chakra-es please review. Thanks!

lib/Runtime/Types/TypePropertyCache.cpp
```diff
@@ -172,7 +172,9 @@ namespace Js
                 &propertyIndex,
                 &isInlineSlot,
                 &isMissing,
-                &prototypeObjectWithProperty))
+                &prototypeObjectWithProperty)
+            || (prototypeObjectWithProperty != nullptr
+                && prototypeObjectWithProperty != propertyObject))
```
curtisman: Huh? prototypeObjectWithProperty should never by the propertyObject itself.   Otherwise, we will not hit the "proto" inline cache and have prototypeObjectWithProperty not null

pleath: @suwc , as we discussed, I think we're probably abusing the cache on a byte code level. I'll make some time to investigate today. If that's the case, then we want to fix the code gen rather than working around it in the cache logic.

suwc: The TypePropertyCache can be populated in multiple occasions. In this repro case, calls to “super.x” go through these lines in GetProperty_Internal():

                PropertyValueInfo::Set(info, requestContext->GetLibrary()->GetMissingPropertyHolder(), 0);
                CacheOperators::CachePropertyRead(instance, requestContext->GetLibrary()->GetMissingPropertyHolder(), isRoot, propertyId, true, info, requestContext);

Now that the cache is populated with requestContext->GetLibrary()->GetMissingPropertyHolder() as “prototypeObjectWithProperty”, next call to “this.x” (having same “instance”) will be a cache hit (with prototypeObjectWithProperty != nullptr), whereas it should be a cache miss + update.


---
In reply to: [127714797](https://github.com/Microsoft/ChakraCore/pull/3336#discussion_r127714797) [](ancestors = 127714797)

pleath: Yes, as I thought, the byte code gen for super.x is misusing the inline caches. So let's apply a fix there rather than in the cache logic. We can discuss offline.

suwc:
@dotnet-bot test Ubuntu _no_jit_shared_ubuntu_linux_test

lib/Runtime/Language/JavascriptOperators.cpp
```diff
@@ -1739,7 +1739,7 @@ namespace Js
             //         Also we might want to throw here instead of checking it again in the caller
             if (value && !requestContext->IsUndeclBlockVar(*value) && !WithScopeObject::Is(object))
             {
-                CacheOperators::CachePropertyRead(instance, object, isRoot, propertyId, false, info, requestContext);
+                CacheOperators::CachePropertyRead(isSuper ? propertyObject : instance, object, isRoot, propertyId, false, info, requestContext);
```
curtisman: I think this should always be propertyObject......

suwc: @curtisman I don't have an example where caching on instance is more appropriate when propertyObject != instance. When a super property ("super.x") exists in super object's prototype (vs. the super object itself), it is distinguished by object != prototypeObject. It would be simpler if we can go without 'isSuper'.

@pleath am I missing something?

curtisman: You should be able to find all the cases that caching on the instance is wrong and create test cases for them

pleath: If we can prove that there's no case where propertyObject != instance, and where we want to cache based on instance, or where the fact that propertyObject !=instance is our indication that we have a prototype access, then going without isSuper seems fine. The only case I can think of is where instance is a tagged number, and in that case it seems fine to cache as a local access to Number.prototype (which will be propertyObject, right?).

pleath: (All of which means that the logic we've had since IE9 is subtly wrong, but in a way that never mattered until super was implemented. That may be true, but it would be nice to prove it.)

akroshg:
so what is a state on this PR?

pleath:
A simpler fix along the lines of what Curtis suggests seems possible, but we want to do some work to confirm that.

lib/Runtime/Language/JavascriptOperators.cpp
suwc: Besides super property access, other cases where propertyObject != instance are following:
1) instance is proxy, propertyObject is target object
2) instance is number, propertyObject is Number.prototype
3) instance is boolean, propertyObject is Boolean.prototype

By design, 1) does not cache property access (confirmed) because underlying target object can mutate without proxy knowing it.

2) and 3) are justified to cache on their prototype (propertyObject).


---
In reply to: [130143456](https://github.com/Microsoft/ChakraCore/pull/3336#discussion_r130143456) [](ancestors = 130143456)

suwc:
Besides super property access, other cases where propertyObject != instance are following:
1) instance is proxy, propertyObject is target object
2) instance is number, propertyObject is Number.prototype
3) instance is boolean, propertyObject is Boolean.prototype

By design, 1) does not cache property access (confirmed) because underlying target object can mutate without proxy knowing it.

2) and 3) are justified to cache on their prototype (propertyObject).

lib/Backend/GlobOpt.cpp
```diff
@@ -3489,7 +3489,11 @@ GlobOpt::OptSrc(IR::Opnd *opnd, IR::Instr * *pInstr, Value **indirIndexValRef, I
 
     if (opnd->IsSymOpnd() && opnd->AsSymOpnd()->IsPropertySymOpnd())
     {
-        TryOptimizeInstrWithFixedDataProperty(&instr);
+        if (instr->m_opcode != Js::OpCode::LdSuperFld) // TODO: should profile LdSuperFld?
```
pleath: Yes, I would prefer to profile LdSuperFld. It seems best to let it participate in this optimization. If there's a reason we can't do that, I'd prefer to use some kind of general check on the opcode attributes rather than special-casing for this opcode.

lib/Backend/JnHelperMethodList.h
```diff
@@ -215,10 +215,10 @@ HELPERCALL(Op_Memset, Js::JavascriptOperators::OP_Memset, AttrCanThrow)
 HELPERCALL(Op_Memcopy, Js::JavascriptOperators::OP_Memcopy, AttrCanThrow)
 
 HELPERCALL(Op_PatchGetValue, ((Js::Var (*)(Js::FunctionBody *const, Js::InlineCache *const, const Js::InlineCacheIndex, Js::Var, Js::PropertyId))Js::JavascriptOperators::PatchGetValue<true, Js::InlineCache>), AttrCanThrow)
-HELPERCALL(Op_PatchGetValueWithThisPtr, ((Js::Var(*)(Js::FunctionBody *const, Js::InlineCache *const, const Js::InlineCacheIndex, Js::Var, Js::PropertyId, Js::Var))Js::JavascriptOperators::PatchGetValueWithThisPtr<true, Js::InlineCache>), AttrCanThrow)
+HELPERCALL(Op_PatchGetSuperValue, ((Js::Var(*)(Js::FunctionBody *const, Js::InlineCache *const, const Js::InlineCacheIndex, Js::Var, Js::PropertyId, Js::Var, const bool))Js::JavascriptOperators::PatchGetSuperValue<true, Js::InlineCache>), AttrCanThrow)
```
pleath: Curious...is there a reason for the name change? I kind of like the generic name better than one that suggests this is only for super. But I could be convinced.

lib/Runtime/ByteCode/ByteCodeWriter.cpp
```diff
@@ -2126,7 +2126,7 @@ namespace Js
         return false;
     }
 
-    void ByteCodeWriter::PatchablePropertyWithThisPtr(OpCode op, RegSlot value, RegSlot instance, RegSlot thisInstance, uint cacheId, bool isCtor, bool registerCacheIdForCall)
+    void ByteCodeWriter::PatchableSuperProperty(OpCode op, RegSlot value, RegSlot instance, RegSlot thisInstance, uint cacheId, bool isCtor, bool registerCacheIdForCall)
```
pleath: Again, I generally prefer a generic name.

lib/Backend/JnHelperMethodList.h
suwc: This is just a wrapper. I renamed the generic version to "PatchGetValueHelper" and hid it from JIT, in order to have (current and future) flexibility of adding, changing, or removing a parameter without JIT impact. Same goes for other helper functions. JIT doesn't need to deal with these.

---
In reply to: [132067176](https://github.com/Microsoft/ChakraCore/pull/3336#discussion_r132067176) [](ancestors = 132067176)

lib/Runtime/ByteCode/ByteCodeWriter.cpp
suwc: Although this so far is used only for super, I can see a point for leaving it in a generic name here. I will revert it to the old name.

---
In reply to: [132067519](https://github.com/Microsoft/ChakraCore/pull/3336#discussion_r132067519) [](ancestors = 132067519)

suwc:
@pleath, latest iteration made LdSuperFld & StSuperFld ProfiledInstr. Also reverted name changes in bytecode emitter to remain generic.

lib/Backend/JnHelperMethodList.h
pleath: Well, the JIT isn't really a factor. I'm just wondering why we wouldn't go with a generic name -- or, specifically, why we would change from using a generic name to using a less portable one. Can we revert this?

lib/Backend/JnHelperMethodList.h
```diff
@@ -234,16 +234,16 @@ HELPERCALL(Op_PatchInitValue, ((void (*)(Js::FunctionBody *const, Js::InlineCach
 HELPERCALL(Op_PatchInitValuePolymorphic, ((void (*)(Js::FunctionBody *const, Js::PolymorphicInlineCache *const, const Js::InlineCacheIndex, Js::RecyclableObject*, Js::PropertyId, Js::Var))Js::JavascriptOperators::PatchInitValue<true, Js::PolymorphicInlineCache>), AttrCanThrow)
 
 HELPERCALL(Op_PatchPutValue, ((void (*)(Js::FunctionBody *const, Js::InlineCache *const, const Js::InlineCacheIndex, Js::Var, Js::PropertyId, Js::Var, Js::PropertyOperationFlags))Js::JavascriptOperators::PatchPutValue<true, Js::InlineCache>), AttrCanThrow)
-HELPERCALL(Op_PatchPutValueWithThisPtr, ((void (*)(Js::FunctionBody *const, Js::InlineCache *const, const Js::InlineCacheIndex, Js::Var, Js::PropertyId, Js::Var, Js::Var, Js::PropertyOperationFlags))Js::JavascriptOperators::PatchPutValueWithThisPtr<true, Js::InlineCache>), AttrCanThrow)
+HELPERCALL(Op_PatchPutSuperValue, ((void (*)(Js::FunctionBody *const, Js::InlineCache *const, const Js::InlineCacheIndex, Js::Var, Js::PropertyId, Js::Var, Js::Var, Js::PropertyOperationFlags))Js::JavascriptOperators::PatchPutSuperValue<true, Js::InlineCache>), AttrCanThrow)
```
pleath: Ditto for these name changes. The existing names are preferable, IMO, because they describe what the method does and don't create the impression that the methods are necessarily limited to one opcode.

suwc:
@pleath, I reverted the name change. Thanks!

pleath:
Awesome. Thanks. That leaves us with a nice, simple-looking change.

PR #3338, opened 2017-07-13
assert that function we are JITing has a body

MikeHolman:
There was a bug that caused this not to happen, and it hit null AV on JIT process. This assert should allow us to catch such an issue earlier, when we have more context.

OS: 12457116

### DIFF
--- a/lib/Backend/FunctionJITTimeInfo.cpp
+++ b/lib/Backend/FunctionJITTimeInfo.cpp
@@ -31,7 +31,7 @@ FunctionJITTimeInfo::BuildJITTimeData(
     }
     else
     {
-        // outermost function must have a body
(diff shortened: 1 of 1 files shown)


lib/Backend/FunctionJITTimeInfo.cpp
```diff
@@ -29,6 +29,11 @@ FunctionJITTimeInfo::BuildJITTimeData(
         jitData->bodyData = AnewStructZ(alloc, FunctionBodyDataIDL);
         JITTimeFunctionBody::InitializeJITFunctionData(alloc, body, jitData->bodyData);
     }
+    else
+    {
+        // outermost function must have a body
```
rajatd: could you leave a comment here that this for built-ins?

rajatd:
:clock1:

PR #3339, opened 2017-07-13
xplat: remove redundant tzset calls from DateTime

obastemur:
timelocal / localtime.. already considers local time while converting back.

### DIFF
--- a/lib/Runtime/PlatformAgnostic/Platform/Linux/DateTime.cpp
+++ b/lib/Runtime/PlatformAgnostic/Platform/Linux/DateTime.cpp
@@ -160,7 +160,6 @@ namespace DateTime
         // tm doesn't have milliseconds
         int milliseconds = local->time % 1000;
 
-        tzset();
         time_t utime = timelocal(&local_tm);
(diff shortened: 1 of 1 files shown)


lib/Runtime/PlatformAgnostic/Platform/Linux/DateTime.cpp
```diff
@@ -160,7 +160,6 @@ namespace DateTime
         // tm doesn't have milliseconds
         int milliseconds = local->time % 1000;
 
-        tzset();
```
MSLaguana: This won't have much of an impact since the other time functions like `timelocal` implicitly invoke `tzset`-like functionality when called.

obastemur: > This won't have much of an impact since the other time functions like timelocal implicitly invoke tzset-like functionality when called.

@MSLaguana  That's why these are redundant hence this PR.

obastemur:
@MSLaguana @digitalinfinity @kfarnung Thanks for the review

PR #3340, opened 2017-07-13
OS#11662594 Only lock the type while adding a type to the equivalent type cache to avoid recreating the type handler for non-shared types.

atulkatti:


atulkatti:
This has been reviewed offline in codeflow. 

@Microsoft/chakra-developers

PR #3346, opened 2017-07-13
[OSG 9773504] Silence a noisy assert. 

pleath:
In source rundown mode, we're hitting an assert in byte code gen to the effect that we don't have a proper initial entry point on the function we're compiling. Make CleanupToReparse set an initial entry point before proceeding.


pleath:
Offline review by @agarwal-sandeep . (Thanks.)

PR #3350, opened 2017-07-14
Bug fix in array bounds check elimination

Cellule:
In DetermineSymBoundOffsetOrValueRelativeToLandingPad, handle the case where bounds is null and TryGetIntConstantLowerBound or TryGetIntConstantupperBound 

Fixes OS#11370283

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/3350)


LouisLaf:
:shipit:

PR #3353, opened 2017-07-14
Debugger: keep new jobs while debugger is attached

### LINKED ISSUE #3352: xplat: Enable JIT for debugging
I compiled ChakraCore using the official build guide. 
On ubuntu 16.04, ch crashes with the following test case.
I also noticed that if we slightly change the code (using a proper indentation), the Seg fault will be gone.
The hashtag of my chakracore build is 
commit cb9c8271c8c9e2a0daec635504b5279168d92f1b

$ ~/chakracore/ChakraCore/out/Release/ch abc.js
Segmentation fault (core dumped)
$ cat abc.js
```
function
run
()
{}
WScript.Attach(run)
WScript
    .
Detach
(
run
)
```

obastemur:
~~Fixes #3352~~

If JIT is enabled (in debug mode) force NoNative.

### DIFF
--- a/lib/Runtime/Base/ScriptContext.cpp
+++ b/lib/Runtime/Base/ScriptContext.cpp
@@ -3137,7 +3137,13 @@ namespace Js
 
         } autoRestore(this->GetThreadContext());
 
+        // xplat-todo: (obastemur) Enable JIT on Debug mode
+        // CodeGen entrypoint can be deleted before we are able to unregister
+        // due to how we handle xdata on xplat, resetting the entrypoints below might affect CodeGen process.
+        // it is safer (on xplat) to turn JIT off during Debug for now.
+#ifdef _WIN32
         if (!Js::Configuration::Global.EnableJitInDebugMode())
+#endif
         {
             if (attach)
             {


lib/Runtime/Base/ScriptContext.cpp
```diff
@@ -3137,7 +3137,7 @@ namespace Js
 
         } autoRestore(this->GetThreadContext());
 
-        if (!Js::Configuration::Global.EnableJitInDebugMode())
+        if (Js::Configuration::Global.EnableJitInDebugMode())
```
agarwal-sandeep: >if (Js::Configuration::Global.EnableJitInDebugMode()) [](start = 8, length = 53)

Sorry I didn't understood the logic. Are you saying we disable jit in debug mode?

Run below test in full with "-mic:1 -off:simplejit Test.js -targeted" and you will see the difference

function run() {
  foo();
  foo();
}
function foo() {
  WScript.Echo(Debug.isInJit());
}
run();
WScript.Attach(run);

obastemur: JIT is enabled by default for debug. The problem is background worker is processing  on updated entity while all the attach/detach happens. Reading the comments around code I have edited, IMHO the fix does the right thing and prevents this?

obastemur: I got your point. Flag vs comments did mislead me. Let me take a closer look to underlying problem.

lib/Runtime/Base/ScriptContext.cpp
```diff
@@ -3137,7 +3137,13 @@ namespace Js
 
         } autoRestore(this->GetThreadContext());
 
+        // xplat-todo: (obastemur) Enable JIT on Debug mode
+        // CodeGen entrypoint can be deleted before we are able to unregister
+        // due to how we handle xdata on xplat, resetting the entrypoints below might affect CodeGen process.
+        // it is safer (on xplat) to turn JIT off during Debug for now.
+#ifdef _WIN32
```
agarwal-sandeep: Can we open a tracking issue for this?

obastemur: There is already an issue open. #3352

agarwal-sandeep: I mean open a issue to enable jit debugging after xdata issue is fixed.

digitalinfinity:
@obastemur this should affect Node-ChakraCore debugging on xplat too, correct? If so, can you cherry pick this into release/1.7 so that we can pick it up in the node-chakracore master branch please?

obastemur:
@digitalinfinity retargeted to 1.7 @agarwal-sandeep Thanks for the review!

PR #3354, opened 2017-07-14
Fix issue#3246: Module namespace entries are not sorted properly

suwc:
9.4.6 Module namespace exotic objects' [[OwnPropertyKeys]] should
return a list of property names as sorted with Array.prototype.sort
using undefined as comparefn.

### DIFF
--- a/lib/Runtime/Language/JavascriptOperators.cpp
+++ b/lib/Runtime/Language/JavascriptOperators.cpp
@@ -15,6 +15,9 @@
 
 #include "Library/ForInObjectEnumerator.h"
 #include "Library/ES5Array.h"
+#include "Types/SimpleDictionaryPropertyDescriptor.h"
+#include "Types/SimpleDictionaryTypeHandler.h"
+#include "Language/ModuleNamespace.h"
 
 #ifndef SCRIPT_DIRECT_TYPE
 typedef enum JsNativeValueType: int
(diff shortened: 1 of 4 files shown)


suwc:
This addresses https://github.com/Microsoft/ChakraCore/issues/3246

suwc:
@Microsoft/chakra-developers @Microsoft/chakra-es review please. Thanks!

atulkatti:
:shipit:

lib/Runtime/Language/JavascriptOperators.cpp
```diff
@@ -1145,6 +1148,20 @@ namespace Js
...
 
+        if (ModuleNamespace::Is(object))
+        {
+            ModuleNamespace *ns = ModuleNamespace::FromVar(object);
+            JavascriptArray *names = ns->GetSortedExportedNames();
+            if (names == nullptr)
+            {
+                names = JavascriptObject::CreateOwnStringPropertiesHelper(object, scriptContext);
+                names->Sort(nullptr);
+                ns->SetSortedExportedNames(names);
```
curtisman: Can we actually cache this?  The array is returned to the user and it is mutable right?

lib/Runtime/Language/JavascriptOperators.cpp
```diff
@@ -1145,6 +1148,20 @@ namespace Js
             return proxy->PropertyKeysTrap(JavascriptProxy::KeysTrapKind::GetOwnPropertyNamesKind, scriptContext);
         }
 
+        if (ModuleNamespace::Is(object))
```
curtisman: How about `JavascriptOperators::GetOwnPropertyKeys`?
See 9.4.6.10?

lib/Runtime/Language/JavascriptOperators.cpp
boingoing: Think I agree with this. We should not cache the array or we should return a clone of the cached version if user code can mutate the array.

PR #3358, opened 2017-07-14
make wasm/asm.js event filter more strict

MikeHolman:
We should require that the faulting location was part of the reservation, but we might want to change the BaseOpnd in some cases, so we shouldn't require that it is the wasm memory


Cellule:
LGTM

Issue #2590, opened 2017-02-23
Running native tests xplat


jaredwy:
In an attempt to write some unit tests for #2581 I hit a slight snag. The native tests that test the JSRT api is driven by a windows only build. 

I managed to get the driver to compile, this took a few #defines and an update to catch.hpp. Also I had to use libc++ as the version of stdlib++ doesn't seem to support some of the features that are needed by catch. I would need to look on linux to see what happens here.

The other problem is the rather liberal use of windows only jsrt functions. E.g.

```
REQUIRE(JsGetPropertyNameFromId(name1, &name) == JsNoError); REQUIRE(JsPointerToString(_u("value1"), wcslen(_u("value1")), &value1) == JsNoError);
```

My suggestion is rewrite these to use `JsCopyString` etc. If their is still a need to test JsPointerToString this can be done with platform specific test files.

digitalinfinity:
Thanks @jaredwy - talked offline with @curtisman and we don't see any problems moving these tests to the new Chakra API path- we would welcome a PR here 😄 

sidenote- curious, what was the issue with catch and stdlib++ (cc @akroshg)

jaredwy:
Happy to spin up a PR. Ill have some time this weekend.

Any issues with MS just dropping in the latest version of catch from github? 
I may need a hand just checking the windows side of things, as I don't have a windows machine at present. 


I don't think it is a problem with stdlib++ per say, just an issue with the version that ships in osx. I was getting a bunch of missing symbol errors. I didn't look too far into it. Been on mac long enough to know that there are issues with gnu tooling :)

akroshg:
@jaredwy I haven't dropped the latest version of the catch. You are welcome to try with the new version.

digitalinfinity:
@jaredwy happy to help with the windows side of things- you can share your branch here (or if you open a PR against our master branch, our CI will run the Windows and Linux builds automatically anyway)

jaredwy:
Are these run on linux?  I couldn't find any evidence that they were.

digitalinfinity:
No, the native tests are not run on Linux (they will be after your PR hopefully 😄). The entry point to invoke them on Windows is test\runnativetests.cmd and I believe they run in our Jenkins CI (@dilijev/@akroshg, correct me if I'm wrong here)

From your PR perspective, I think you'll need to do the following:
[...]

PR #3360, opened 2017-07-14
Fixes #3192, allows promise callback to be set to null

### LINKED ISSUE #3192: can't cleanup after JsSetPromiseContinuationCallback
I'd like some method to remove my non-null promise continuation callback from the JavaScriptContext.

However, I get a ``JavaScriptErrorCode.NullArgument`` if I call ``JavaScriptContext.SetPromiseContinuationCallback(null, IntPtr.Zero)``

I presume the argument validation is too strict?

jackhorton:


digitalinfinity:
@dotnet-bot test Tab Check please

lib/Jsrt/ChakraCommon.h
```diff
@@ -2384,7 +2384,7 @@ typedef unsigned short uint16_t;
     /// </returns>
     CHAKRA_API
         JsSetPromiseContinuationCallback(
-            _In_ JsPromiseContinuationCallback promiseContinuationCallback,
+            _In_opt_ JsPromiseContinuationCallback promiseContinuationCallback,
```
jackhorton: @liminzhu This may need a doc update?

bin/NativeTests/JsRTApiTest.cpp
```diff
@@ -1414,9 +1414,10 @@ namespace JsRTApiTest
     static void CALLBACK PromiseContinuationCallback(JsValueRef task, void *callbackState)
     {
         CHECK(callbackState != nullptr);
-        CHECK(*(void **)callbackState == nullptr);
-        // All the task need to finish async, so we need to save the callback.
-        *(void **)callbackState = task;
+        CHECK(*(JsValueRef *) callbackState == JS_INVALID_REFERENCE);
```
kfarnung: I feel like this was more correct before.  callback state is a void pointer so you don't need to cast it to assign the JsValueRef to it.

bin/NativeTests/JsRTApiTest.cpp
```diff
@@ -1428,15 +1429,15 @@ namespace JsRTApiTest
         REQUIRE(JsSetPromiseContinuationCallback(PromiseContinuationCallback, &callback) == JsNoError);
         REQUIRE(JsRunScript(
             _u("new Promise(") \
-            _u("  function(resolve, reject) {") \
-            _u("      resolve('basic:success');") \
-            _u("  }") \
+            _u("    function(resolve, reject) {") \
```
kfarnung: Was this intentional?  It seems like JS tends to use two spaces for indentation anyway, so I'd rather just leave this as it was unless there's a strong opinion to the contrary.

lib/Jsrt/Core/JsrtContextCore.h
```diff
@@ -163,7 +163,6 @@ class ChakraCoreHostScriptContext sealed : public HostScriptContext
 
     HRESULT EnqueuePromiseTask(Js::Var taskVar) override
     {
-        AssertMsg(false, "jsrt should have set the promise callback");
```
kfarnung: I feel like this should be ok as well, it only hits in debug builds when the embedder has not added a promise callback, but tries to use Promises in their code?

bin/NativeTests/JsRTApiTest.cpp
jackhorton: There was one spacing issue with a brace before and that was just how I ended up fixing it. I can revert it back to two spaces.

lib/Jsrt/Jsrt.cpp
```diff
@@ -2932,15 +2932,13 @@ CHAKRA_API JsIdle(_Out_opt_ unsigned int *nextIdleTick)
...
-    return ContextAPINoScriptWrapper_NoRecord([&](Js::ScriptContext * scriptContext) -> JsErrorCode {
-        PARAM_NOT_NULL(promiseContinuationCallback);
-
-        scriptContext->GetLibrary()->SetNativeHostPromiseContinuationFunction((Js::JavascriptLibrary::PromiseContinuationCallback)promiseContinuationCallback, callbackState);
+    return ContextAPINoScriptWrapper_NoRecord([&](Js::ScriptContext *scriptContext) -> JsErrorCode {
+        scriptContext->GetLibrary()->SetNativeHostPromiseContinuationFunction((Js::JavascriptLibrary::PromiseContinuationCallback) promiseContinuationCallback, callbackState);
         return JsNoError;
     },
-    /*allowInObjectBeforeCollectCallback*/true);
+    /*allowInObjectBeforeCollectCallback*/ true);
```
kfarnung: It's not too distracting here, but it's better to minimize the changes you make when completing a bug fix.  These sorts of formatting changes create more opportunities for merge conflicts and can hide the true spirit of the change.  If we want to fix these, I would do a separate PR to take care of them or at least split them out of the bug fix commit and push two separate commits.

lib/Jsrt/Core/JsrtContextCore.h
jackhorton: I took this out after speaking with Jimmy, because we wanted to add a unit test for this and he said testing asserts was problematic.

MSLaguana: Please correct me if I'm wrong, but I recall that when I made a change to `nativetests` and was trying to test an error case that had a debug assert when the test case hit it, the assert caused the process to exit, and the test wasn't even marked as failed. My perspective here was that if we are explicitly supporting un-setting the promise callback, then having a promise complete with no completion handler is no longer assert-worthy, but instead a documented error case.

jackhorton: That's exactly what happened, NativeTests.exe crashes in debug mode with that assert in. In the JSRT overview in the wiki, there is mention that you need to use this function in order to use promises. Should something be added to the JsSetPromiseContinuationCallback documentation as well?

lib/Jsrt/ChakraCommon.h
liminzhu: Will do after the change merges

bin/NativeTests/JsRTApiTest.cpp
```diff
@@ -1472,6 +1474,48 @@ namespace JsRTApiTest
...
 
+    void UnsetPromiseContinuation(JsRuntimeAttributes attributes, JsRuntimeHandle runtime)
+    {
+        JsValueRef result = JS_INVALID_REFERENCE, callbackState = JS_INVALID_REFERENCE, exception = JS_INVALID_REFERENCE;
+        JsValueType cbStateType = JsUndefined;
+
+        // script with no promise continuation callback should error
+        REQUIRE(JsRunScript(_u("new Promise((res, rej) => res()).then(() => 1)"), JS_SOURCE_CONTEXT_NONE, _u(""), &result) == JsErrorScriptException);
+        CHECK(result == JS_INVALID_REFERENCE);
+        REQUIRE(JsGetAndClearException(&exception) == JsNoError);
```
MSLaguana: Perhaps also verify that the exception that was reported matches the one that we expect?

bin/NativeTests/JsRTApiTest.cpp
```diff
@@ -1478,17 +1478,18 @@ namespace JsRTApiTest
     {
         JsValueRef result = JS_INVALID_REFERENCE, callbackState = JS_INVALID_REFERENCE, exception = JS_INVALID_REFERENCE;
         JsValueType cbStateType = JsUndefined;
+        const wchar_t *script = _u("new Promise((res, rej) => res()).then(() => 1)");
```
kfarnung: I think you want to use WCHAR since wchar_t is not a consistent size across Linux and Windows.

kfarnung: Oh, hmm, it looks like maybe these tests are currently Windows only?  In that case I guess it doesn't matter, but we should probably get them running cross-platform.

kfarnung:
Meta comment, you'll probably want to squash your changes before the final merge and I'd update the commit message title with a description of the bug and just mention "Fixes <foo>" in the body.

bin/NativeTests/JsRTApiTest.cpp
jackhorton: if I convert to WCHAR, should I also convert to JsRun? Or, if making these tests xplat is a goal, we should probably just open an issue to do the conversion all at once.

kfarnung: Honestly this method is a little bit of a mess.  It's Windows only and the header uses wchar_t and the implementation uses "WCHAR".  I'd say just leave it for now.

digitalinfinity: I agree- dealing with making NativeTests xplat is a separate issue, and is already tracked with #2590. I'll poke that issue again. For now, you can leave it as wchar_t, we can clean this up later.

PR #3362, opened 2017-07-14
Make PropertyQueryFlags an enum class

suwc:
Change PropertyQueryFlags from enum to enum class
Update references to enum class values
Fix unintended implicit conversions between PropertyQueryFlags and BOOL


tcare:
Is there a bug number associated with this?

suwc:
No, there's no associated bug or issue.

---
In reply to: [315488814](https://github.com/Microsoft/ChakraCore/pull/3362#issuecomment-315488814) [](ancestors = 315488814)

PR #3364, opened 2017-07-15
Marshal return value of GetCallerProperty as per requestContext

meg-gupta:


lib/Runtime/Library/JavascriptFunction.cpp
```diff
@@ -2793,6 +2793,10 @@ void __cdecl _alloca_probe_16()
             {
                 *value = scriptContext->GetLibrary()->GetNull();
             }
+            else
```
agarwal-sandeep: In if is it guaranteed that funcCaller->GetScriptContext() is requestContext here?
Better to use nullValue (from line 2754) here.

meg-gupta: funcCaller->GetScriptContext() != requestContext here. Marshalling was needed due to this.

meg-gupta: @agarwal-sandeep not sure if I understood the question. Do you mean we have to return nullValue instead of marshalling when the contexts don't match ?

agarwal-sandeep: funcCaller->GetScriptContext() != requestContext then in case of cross domain check fail we are returning null from scriptContext (not marshaled to requestContext). That's why I say in if (S_OK != hr) case use *value = nullValue which is from requestContext.

meg-gupta: oh yes! Thanks. Updated the review.

agarwal-sandeep:
Change looks good to me but can you also add a test case in full to exercise cross domain scenario. In full the 5th argument to WScript.LoadScript is the domainId, see example unittest\Debugger\debugEvalSite.js

kunalspathak:
:shipit:

PR #3365, opened 2017-07-15
BSF and BSR should be marked as setting flags

MikeHolman:
Without this, HoistSameInstructionAboveSplit peep could kick in and invalidate flags register, leading to clz/ctz giving the wrong result

OS: 12604639


MikeHolman:
We originally got the table from MSVC, but only took the parts we needed. I think these ops were added in later, and may not have used VC as a reference.

PR #3369, opened 2017-07-17
OS#12724597 Symbol registraion map should use the key from the property record's buffer.

atulkatti:


atulkatti:
@Microsoft/chakra-developers

lib/Runtime/Base/ThreadContext.cpp
```diff
@@ -4297,7 +4297,7 @@ const Js::PropertyRecord* ThreadContext::AddSymbolToRegistrationMap(const char16
     // However, as the key contains a null character we need to hash the symbol name past the null character. The default implementation terminates
     // at the null character, so we use the Js::HashedCharacterBuffer as key. We allocate the key in the recycler memory as it needs to be around
     // for the lifetime of the map.
-    Js::HashedCharacterBuffer<char16> * propertyName = RecyclerNew(GetRecycler(), Js::HashedCharacterBuffer<char16>, stringKey, stringLength);
+    Js::HashedCharacterBuffer<char16> * propertyName = RecyclerNew(GetRecycler(), Js::HashedCharacterBuffer<char16>, propertyRecord->GetBuffer(), propertyRecord->GetLength());
```
agarwal-sandeep: >propertyRecord [](start = 117, length = 14)

How did we find this issue? is there an existing test. If not please add one.

boingoing:
LGTM

lib/Runtime/Base/ThreadContext.cpp
akroshg: what is the difference here, shouldn't both length be same?

atulkatti: They should be the same. The real fix is to use the property record's buffer. I made the length change just to be safe.

---
In reply to: [127781026](https://github.com/Microsoft/ChakraCore/pull/3369#discussion_r127781026) [](ancestors = 127781026)

atulkatti: This was found by crawler. I will see if I can write a reliable test for this. Basically this would need the previous string to be garbage collected. 

---
In reply to: [127757480](https://github.com/Microsoft/ChakraCore/pull/3369#discussion_r127757480) [](ancestors = 127757480)

jianchun: From a glance, "stringKey" was "GetSz()" result from previous JavascriptString arg, and that looks normally recycler allocated and should not be GC freed if referenced here. Not sure why it was freed.

kunalspathak:
:shipit:

lib/Runtime/Base/ThreadContext.cpp
atulkatti: I looked at this briefly, but can't see why the original string is getting collected. The repro isn't straightforward either. However,  crawler has close to 200 hits so far. I will talk to Curtis about this and see if he has any thoughts.

---
In reply to: [127852232](https://github.com/Microsoft/ChakraCore/pull/3369#discussion_r127852232) [](ancestors = 127852232)

PR #3359, opened 2017-07-14
Fix issue#3249: Module namespace object's Symbol.toStringTag has incorrect attributes

suwc:
Module namespace object's Symbol.toStringTag must be:
writable: false
enumerable: false
configurable: false

PR #3377, opened 2017-07-17
Fix module namespace property attributes

suwc:
9.4.6.
   5. Return PropertyDescriptor{[[Value]]: value, [[Writable]]: true, [[Enumerable]]: true, [[Configurable]]: false }.

26.3.1 @@toStringTag
The initial value of the @@toStringTag property is the String value "Module".
This property has the attributes { [[Writable]]: false, [[Enumerable]]: false, [[Configurable]]: false }.

### DIFF
--- a/lib/Runtime/Language/ModuleNamespace.cpp
+++ b/lib/Runtime/Language/ModuleNamespace.cpp
@@ -50,7 +50,7 @@ namespace Js
         if (scriptContext->GetConfig()->IsES6ToStringTagEnabled())
         {
             DynamicObject::SetPropertyWithAttributes(PropertyIds::_symbolToStringTag, library->GetModuleTypeDisplayString(),
-                PropertyConfigurable, nullptr);
+                PropertyNone, nullptr);
         }
 
         ModuleImportOrExportEntryList* localExportList = sourceTextModuleRecord->GetLocalExportEntryList();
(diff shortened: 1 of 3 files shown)


suwc:
Addresses https://github.com/Microsoft/ChakraCore/pull/3359#pullrequestreview-50391127

lib/Runtime/Language/ModuleNamespace.h
```diff
@@ -61,9 +61,9 @@ namespace Js
         virtual BOOL GetEnumerator(JavascriptStaticEnumerator * enumerator, EnumeratorFlags flags, ScriptContext* requestContext, ForInCache * forInCache = nullptr);
         virtual BOOL SetAccessors(PropertyId propertyId, Var getter, Var setter, PropertyOperationFlags flags = PropertyOperation_None) override { return false; }
         virtual BOOL GetAccessors(PropertyId propertyId, Var *getter, Var *setter, ScriptContext * requestContext) override { return false; }
-        virtual BOOL IsWritable(PropertyId propertyId) override { return false; }
+        virtual BOOL IsWritable(PropertyId propertyId) override { return propertyId != PropertyIds::_symbolToStringTag; }
```
curtisman: Although currently a `ModuleNamespace` object only have `@@toStringTag`, but the `[[Get]]` operation is defined to redirect to `OrdinaryGet` for any symbol.  So these IsWritable/IsEnumerable API should of the same.

lib/Runtime/Language/ModuleNamespace.cpp
```diff
@@ -170,6 +170,42 @@ namespace Js
         return HasProperty(propertyId);
     }
 
+    BOOL ModuleNamespace::IsConfigurable(PropertyId propertyId)
+    {
+        SimpleDictionaryPropertyDescriptor<BigPropertyIndex> propertyDescriptor;
+        const Js::PropertyRecord* propertyRecord = GetScriptContext()->GetThreadContext()->GetPropertyName(propertyId);
+        if (propertyMap != nullptr && propertyMap->TryGetValue(propertyRecord, &propertyDescriptor))
```
curtisman: This still doesn't match the order of what GetPropertyQuery does, which check if it is symbol and only then  we will pass it on to DynamicObject?

We should keep these logic in sync.

PR #3378, opened 2017-07-18
TryFinally Fixes

meg-gupta:
-  Dont run region propagation code on deleted/dead blocks
- Set m_hasNonBranchRef on a label in eh region which can become unreferenced after globopt removes BrOnException edges
-  Add ByteCodeOffset to BrOnException added from finally to early exit

### DIFF
--- a/lib/Backend/GlobOpt.cpp
+++ b/lib/Backend/GlobOpt.cpp
@@ -2572,7 +2572,6 @@ GlobOpt::OptInstr(IR::Instr *&instr, bool* isInstrRemoved)
             if (this->RemoveFlowEdgeToFinallyOnExceptionBlock(instr))
             {
                 *isInstrRemoved = true;
-                this->currentBlock->RemoveInstr(instr);
                 return instrNext;
             }
         }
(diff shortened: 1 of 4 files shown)


lib/Backend/GlobOpt.cpp
```diff
@@ -2572,7 +2572,6 @@ GlobOpt::OptInstr(IR::Instr *&instr, bool* isInstrRemoved)
             if (this->RemoveFlowEdgeToFinallyOnExceptionBlock(instr))
             {
                 *isInstrRemoved = true;
-                this->currentBlock->RemoveInstr(instr);
```
LouisLaf: Can we move this whole block of EH handling code to a ProcessExceptionHandling() functions?  I'd like to remove code from OptInstr() to attempt to maintain readability.

LouisLaf:
:shipit:

lib/Backend/GlobOpt.cpp
meg-gupta: Done. Thanks for reviewing!

PR #3379, opened 2017-07-18
fix GetFldInfoAddr

sigatrev:
fldDataAddr is a long long not a pointer type, so we need to account for the size of the struct when getting the offset.


MikeHolman:
would be good to add a test

rajatd:
Looks good.

PR #3381, opened 2017-07-18
Update PR#3354: Module namespace entries are not sorted properly

suwc:
Add 'sortedExportedNames' to ModuleNamespace and update ModuleNamespaceEnumerator.

9.4.6 Module namespace exotic objects' [[OwnPropertyKeys]] should
return a list of property names as sorted with Array.prototype.sort
using undefined as comparefn.

### DIFF
--- a/lib/Runtime/Language/ModuleNamespace.cpp
+++ b/lib/Runtime/Language/ModuleNamespace.cpp
@@ -84,6 +84,7 @@ namespace Js
         // For items that are not in the local export list, we need to resolve them to get it
         ExportedNames* exportedNames = sourceTextModuleRecord->GetExportedNames(nullptr);
         ModuleNameRecord* moduleNameRecord = nullptr;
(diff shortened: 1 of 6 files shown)


lib/Runtime/Language/ModuleNamespace.cpp
```diff
@@ -116,6 +119,21 @@ namespace Js
                 this->AddUnambiguousNonLocalExport(propertyId, moduleNameRecord);
             });
         }
+
+        sortedExportedNames->Sort([](void* context, const void* left, const void* right) ->int {
+            JavascriptString** leftString = (JavascriptString**)(left);
+            JavascriptString** rightString = (JavascriptString**)(right);
+            if (JavascriptString::LessThan(*leftString, *rightString))
```
curtisman: Shouldn't just one call to `JavascriptString::strcmp` and return suffice?

boingoing:
Looks good to me

suwc:
Thanks for the review.

PR #3382, opened 2017-07-18
On .caller, only invalidate cached scopes for nested parents

rajatd:
When a function escapes as the return value of .caller, we currently walk the whole stack and invalidate cached scope on every function on the stack that has one set. However, it seems that we only need to do the invalidation for nested parents of the result of .caller. This PR attempts to do just that.

### DIFF
--- a/lib/Runtime/Types/ActivationObject.h
+++ b/lib/Runtime/Types/ActivationObject.h
@@ -209,6 +209,12 @@ namespace Js
             return VirtualTableInfo<Js::ActivationObjectEx>::HasVirtualTable(instance);
         }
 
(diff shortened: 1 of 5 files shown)


lib/Runtime/Types/ActivationObject.h
```diff
@@ -209,6 +209,12 @@ namespace Js
             return VirtualTableInfo<Js::ActivationObjectEx>::HasVirtualTable(instance);
         }
 
+        static ActivationObjectEx * FromVar(Var instance)
+        {
+            Assert(Is(instance));
```
akroshg: failfast?

akroshg:
LGTM

PR #3384, opened 2017-07-18
OS#12654029: ArrayBuffer throw OOM if unable to allocate memory after GC

agarwal-sandeep:


lib/Runtime/Library/ArrayBuffer.cpp
```diff
@@ -615,17 +615,17 @@ namespace Js
...
-                    JavascriptError::ThrowOutOfMemoryError(GetScriptContext());
-                }
             }
 
             if (buffer != nullptr)
             {
                 bufferLength = length;
                 ZeroMemory(buffer, bufferLength);
             }
+            else
```
akroshg: LGTM - although Curtis can answer history behind why we didn't have this in the first place?

digitalinfinity: Akrosh, this looks like a dupe of the bug we were looking at, yes? cc @curtisman in case he knows why an ArrayBuffer buffer allocation failure doesn't OOM today

akroshg: Yes

---
In reply to: [128093238](https://github.com/Microsoft/ChakraCore/pull/3384#discussion_r128093238) [](ancestors = 128093238)

jianchun: This was probably an error made in https://github.com/Microsoft/ChakraCore/commit/c04bab38bf0d2f09c7e2cc98b200527418321ade#diff-699ac297c4c6339e45f156e1316de619

PR #3397, opened 2017-07-19
Fix crash when a Map is constructed with custom Map.prototype.set

### LINKED ISSUE #2747: JavascriptMap::NewInstance null deference
```javascript
class ExternInt16Array extends Int16Array {
	static get [Symbol.species]() {	return Map;};
};

var m1 = new Map();
var o1 = Object.getPrototypeOf(m1); 
Reflect.defineProperty(Map.prototype, "set", o1); 
var o2 = new ExternInt16Array(new ArrayBuffer(0x100));
var m2 = new Map(o2); 
```

Above repro triggers accessing incomplete Map instance during `JavascriptMap::NewInstance`, resulting in null dereference violation.

(found by external researcher, thanks!)

jackhorton:
Fixes #2747, sets up map instance before finding Map.prototype.set in case set's getter needs to be called on a valid Map instance

### DIFF
--- a/lib/Runtime/Library/JavascriptMap.cpp
+++ b/lib/Runtime/Library/JavascriptMap.cpp
@@ -64,6 +64,15 @@ namespace Js
 
         Var iterable = (args.Info.Count > 1) ? args[1] : library->GetUndefined();
 
+        if (mapObject->map != nullptr)
+        {
+            JavascriptError::ThrowTypeErrorVar(scriptContext, JSERR_ObjectIsAlreadyInitialized, _u("Map"), _u("Map"));
+        }
+
+        // ensure mapObject->map is created before trying to fetch the adder function
+        // see github#2747
+        mapObject->map = RecyclerNew(scriptContext->GetRecycler(), MapDataMap, scriptContext->GetRecycler());
+
         RecyclableObject* iter = nullptr;
         RecyclableObject* adder = nullptr;
 
@@ -77,13 +86,6 @@ namespace Js
             }
             adder = RecyclableObject::FromVar(adderVar);
         }
-
-        if (mapObject->map != nullptr)
-        {
-            JavascriptError::ThrowTypeErrorVar(scriptContext, JSERR_ObjectIsAlreadyInitialized, _u("Map"), _u("Map"));
-        }
-
-        mapObject->map = RecyclerNew(scriptContext->GetRecycler(), MapDataMap, scriptContext->GetRecycler());
 
         if (iter != nullptr)
         {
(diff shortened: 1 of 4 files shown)


lib/Runtime/Library/JavascriptMap.cpp
```diff
@@ -64,6 +64,15 @@ namespace Js
 
         Var iterable = (args.Info.Count > 1) ? args[1] : library->GetUndefined();
 
+        if (mapObject->map != nullptr)
+        {
+            JavascriptError::ThrowTypeErrorVar(scriptContext, JSERR_ObjectIsAlreadyInitialized, _u("Map"), _u("Map"));
+        }
+
+        // ensure mapObject->map is created before trying to fetch the adder function
```
digitalinfinity: I would be more explicit in the comment here- specifically, point out that subsequent operations (like fetching the adder) can access the map so all it's fields need to be initialized before these operations can be invoked.

lib/Runtime/Library/JavascriptMap.cpp
```diff
@@ -64,6 +64,15 @@ namespace Js
...
+        // see github#2747
+        mapObject->map = RecyclerNew(scriptContext->GetRecycler(), MapDataMap, scriptContext->GetRecycler());
```
akroshg: Isn't the same pattern there in the JavascriptSet::NewInstance ?

jackhorton: Yes, I just checked and its the exact same. I can try to repro the issue there.

test/es6/bug_issue_2747.js
```diff
@@ -0,0 +1,29 @@
...
+var m;
+function constructorFunc () {
+    m = new Map([["a", 1], ["b", 2]]);
+}
+
+Object.defineProperty(Map.prototype, "set", {
+    get: Map.prototype.get, // can be any Map.prototype method that depends on `this` being a valid map
+    configurable: true
+});
+assert.throws(function () { return Map.prototype.set });
```
akroshg: throws function takes other params, which actually describes what kind of error and description. It will be useful for readability. 
same true for doesNotThrow as well.

lib/Runtime/Library/JavascriptMap.cpp
jackhorton: Set was vulnerable to the same issue, I just fixed it there. I can look tonight or tomorrow at other classes that may have the same issue -- I would imagine candidates include any where you can give initial data to the constructor that the corresponding NewInstance function simply calls this->insert() with, especially ES2015 APIs like Set and Map.

jackhorton:
WeakSet and WeakMap don't seem to exhibit the same issue, and nothing else uses a similar code path as far as I can tell.

lib/Runtime/Library/JavascriptMap.cpp
akroshg: Thanks

---
In reply to: [128404692](https://github.com/Microsoft/ChakraCore/pull/3397#discussion_r128404692) [](ancestors = 128404692)

akroshg:
:shipit:

PR #3398, opened 2017-07-19
OS12503560: assignment to super[prop] not accessing base class property

suwc:
Assignments to super[expr] where expr contains variable references
are not being applied to super property objects. Fix bytecode emitter.


curtisman:
:shipit:

PR #3399, opened 2017-07-20
Make toGMTString the same function object as toUTCString

### LINKED ISSUE #2808: test262/annexB: toGMTString and toUTCString should be the same function object
Update definition of exposed functions.
toGMTString currently calls toUTCString internally and we can get rid of the separate definition entirely.

    Failures (1 test):
        test262/test/annexB/built-ins/Date/prototype/toGMTString/value.js:
            Expected SameValue("function toGMTString() { [native code] }", "function toUTCString() { [native code] }") to be true

    Resources:
        esid: sec-date.prototype.togmtstring
        es6id: B.2.4.3
        es5id: B.2.6
        description: Value of `Date.prototype.toGMTString`
        https://tc39.github.io/ecma262/#sec-date.prototype.togmtstring

dilijev:
The spec requires that `Date.prototype.toGMTString` and `Date.prototype.toUTCString` are the same function object.

See: https://tc39.github.io/ecma262/#sec-date.prototype.togmtstring

Fixes #2808

### DIFF
--- a/lib/Runtime/Library/JavascriptBuiltInFunctionList.h
+++ b/lib/Runtime/Library/JavascriptBuiltInFunctionList.h
@@ -125,7 +125,6 @@ BUILTIN(JavascriptDate, ToLocaleTimeString, EntryToLocaleTimeString, FunctionInf
 BUILTIN(JavascriptDate, ToString, EntryToString, FunctionInfo::ErrorOnNew | FunctionInfo::HasNoSideEffect)
 BUILTIN(JavascriptDate, ToTimeString, EntryToTimeString, FunctionInfo::ErrorOnNew)
 BUILTIN(JavascriptDate, ToUTCString, EntryToUTCString, FunctionInfo::ErrorOnNew)
-BUILTIN(JavascriptDate, ToGMTString, EntryToGMTString, FunctionInfo::ErrorOnNew)
 BUILTIN(JavascriptDate, UTC, EntryUTC, FunctionInfo::ErrorOnNew)
 BUILTIN(JavascriptDate, ValueOf, EntryValueOf, FunctionInfo::ErrorOnNew | FunctionInfo::HasNoSideEffect)
 BUILTIN(JavascriptDate, SymbolToPrimitive, EntrySymbolToPrimitive, FunctionInfo::ErrorOnNew)
--- a/lib/Runtime/Library/JavascriptDate.cpp
+++ b/lib/Runtime/Library/JavascriptDate.cpp
@@ -1503,16 +1503,6 @@ namespace Js
             DateImplementation::DateTimeFlag::NoDate);
     }
 
-    // CONSIDER: ToGMTString and ToUTCString is the same, but currently the profiler use the entry point address to identify
-    // the entry point. So we will have to make the function different. Consider using FunctionInfo to identify the function
-    Var JavascriptDate::EntryToGMTString(RecyclableObject* function, CallInfo callInfo, ...)
-    {
-        PROBE_STACK(function->GetScriptContext(), Js::Constants::MinStackDefault);
-
-        ARGUMENTS(args, callInfo);
-        return JavascriptFunction::CallFunction<true>(function, JavascriptDate::EntryToUTCString, args);
-    }
-
     Var JavascriptDate::EntryToUTCString(RecyclableObject* function, CallInfo callInfo, ...)
     {
         PROBE_STACK(function->GetScriptContext(), Js::Constants::MinStackDefault);
--- a/lib/Runtime/Library/JavascriptLibrary.cpp
+++ b/lib/Runtime/Library/JavascriptLibrary.cpp
@@ -2595,7 +2595,7 @@ namespace Js
         scriptContext->SetBuiltInLibraryFunction(JavascriptDate::EntryInfo::ToISOString.GetOriginalEntryPoint(),
             library->AddFunctionToLibraryObject(datePrototype, PropertyIds::toISOString, &JavascriptDate::EntryInfo::ToISOString, 0));
         scriptContext->SetBuiltInLibraryFunction(JavascriptDate::EntryInfo::ToJSON.GetOriginalEntryPoint(),
-             library->AddFunctionToLibraryObject(datePrototype, PropertyIds::toJSON, &JavascriptDate::EntryInfo::ToJSON, 1));
+            library->AddFunctionToLibraryObject(datePrototype, PropertyIds::toJSON, &JavascriptDate::EntryInfo::ToJSON, 1));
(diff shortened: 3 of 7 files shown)


lib/Runtime/Library/JavascriptDate.cpp
```diff
@@ -1503,16 +1503,6 @@ namespace Js
             DateImplementation::DateTimeFlag::NoDate);
     }
 
-    // CONSIDER: ToGMTString and ToUTCString is the same, but currently the profiler use the entry point address to identify
-    // the entry point. So we will have to make the function different. Consider using FunctionInfo to identify the function
```
agarwal-sandeep: What about this?

dilijev: So until this morning, I thought this profiling was like the observable information from F12, and that it wouldn't be a big deal to just treat this as the same function for that purpose. Even if that's not true, since it is literally the same function now aside from the name, I think it might still be okay to treat these as the same function. We can discuss offline if this is a deeper issue than I realize.

agarwal-sandeep: Does F12 script profiler shows ToGMTString and ToUTCString after your change? if yes then its ok

lib/Runtime/Library/JavascriptBuiltInFunctionList.h
```diff
@@ -121,7 +121,7 @@ BUILTIN(JavascriptDate, ToLocaleTimeString, EntryToLocaleTimeString, FunctionInf
 BUILTIN(JavascriptDate, ToString, EntryToString, FunctionInfo::ErrorOnNew | FunctionInfo::HasNoSideEffect)
 BUILTIN(JavascriptDate, ToTimeString, EntryToTimeString, FunctionInfo::ErrorOnNew)
 BUILTIN(JavascriptDate, ToUTCString, EntryToUTCString, FunctionInfo::ErrorOnNew)
-BUILTIN(JavascriptDate, ToGMTString, EntryToGMTString, FunctionInfo::ErrorOnNew)
+BUILTIN(JavascriptDate, ToGMTString, EntryToUTCString, FunctionInfo::ErrorOnNew) // Spec stipulates toGMTString must be the same function object as toUTCString
```
boingoing: For other cases where we have the same function object stored under multiple named properties, we don't have BUILTIN entries for both - only the one matching the entry function. I don't know if there's a requirement or not but you should make sure we need to have this entry here.

You can check out `Array.prototype.toStringLocale` or `Array.prototype.toString` for an example. Both are shared between %TypedArray%.prototype and Array.prototype but there are BUILTIN entries only for the Array methods.

test/Date/DateUTC-DateGMT-same.js
```diff
@@ -0,0 +1,11 @@
+//-------------------------------------------------------------------------------------------------------
+// Copyright (C) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE.txt file in the project root for full license information.
+//-------------------------------------------------------------------------------------------------------
+
+const areSame = Date.prototype.toGMTString === Date.prototype.toUTCString;
```
dilijev: I thought loading the UnitTestFramework and writing the boilerplate for the framework for this one tiny test case was overkill.

lib/Runtime/Library/JavascriptLibrary.cpp
```diff
@@ -7714,7 +7716,7 @@ namespace Js
         REG_OBJECTS_LIB_FUNC(toString, JavascriptDate::EntryToString);
         REG_OBJECTS_LIB_FUNC(toTimeString, JavascriptDate::EntryToTimeString);
         REG_OBJECTS_LIB_FUNC(toUTCString, JavascriptDate::EntryToUTCString);
-        REG_OBJECTS_LIB_FUNC(toGMTString, JavascriptDate::EntryToGMTString);
+        REG_OBJECTS_LIB_FUNC(toGMTString, JavascriptDate::EntryToUTCString);
```
dilijev: That's a good question.

lib/Runtime/Library/JavascriptBuiltInFunctionList.h
dilijev: Thanks for pointing me at those functions. I took a stab at it to get the test to pass and tried to ensure other tests wouldn't break but it appears I missed some.

lib/Runtime/Library/JavascriptLibrary.cpp
```diff
@@ -2670,9 +2670,12 @@ namespace Js
...
         scriptContext->SetBuiltInLibraryFunction(JavascriptDate::EntryInfo::ToTimeString.GetOriginalEntryPoint(),
             library->AddFunctionToLibraryObject(datePrototype, PropertyIds::toTimeString, &JavascriptDate::EntryInfo::ToTimeString, 0));
-        scriptContext->SetBuiltInLibraryFunction(JavascriptDate::EntryInfo::ToUTCString.GetOriginalEntryPoint(),
-            library->AddFunctionToLibraryObject(datePrototype, PropertyIds::toUTCString, &JavascriptDate::EntryInfo::ToUTCString, 0));
-        library->AddFunctionToLibraryObject(datePrototype, PropertyIds::toGMTString, &JavascriptDate::EntryInfo::ToGMTString, 0);
+
+        // Spec stipulates toGMTString must be the same function object as toUTCString
+        JavascriptFunction *toUTCStringFunc = library->AddFunctionToLibraryObject(datePrototype, PropertyIds::toUTCString, &JavascriptDate::EntryInfo::ToUTCString, 0);
+        scriptContext->SetBuiltInLibraryFunction(JavascriptDate::EntryInfo::ToUTCString.GetOriginalEntryPoint(), toUTCStringFunc);
+        library->AddMember(datePrototype, PropertyIds::toGMTString, toUTCStringFunc, PropertyBuiltInMethodDefaults);
```
dilijev: I think this issue is causing a test to fail. I'll review and update.

lib/Runtime/Library/JavascriptDate.cpp
dilijev: I'll set up a test to investigate. If not, how can we make it show the data separately while using the same function object for the two functions?

dilijev:
Chrome's profiler reports a call to Date.prototype.toGMTString as a call to a function called toUTCString. The spec supports this behavior. This change will result in that behavior in Edge as well. I'd say there's nothing to worry about here. @boingoing @agarwal-sandeep @bterlson

lib/Runtime/Library/JavascriptDate.cpp
dilijev: @akroshg was this comment just out of date?

PR #3400, opened 2017-07-20
jsrt: cleanup CHAKRACOREBUILD_ symbol in headers

jianchun:
`CHAKRACOREBUILD_` symbol should not be defined in `ChakraCommon.h`. The
later is included by Chakra JSRT (Windows 10 SDK) and would thus
incorrectly define `CHAKRACOREBUILD_` and expose ChakraCore APIs (Note
that `NTBUILD` is not defined when a user includes Windows SDK headers).

Move the def to `ChakraCore.h` to solve this problem. Normally including
`ChakraCore.h` means using ChakraCore only APIs. (We define the symbol
only when `!NTBUILD` because of one exception -- shared implementation
`Jsrt.cpp` includes `ChakraCore.h`.)

Moved `#ifdef CHAKRACOREBUILD_` to top of file to enclose all recent new
ChakraCore APIs, to help ensure those APIs not defined/exposed to Chakra.

Lastly renamed all `CHAKRACOREBUILD_` to `_CHAKRACOREBUILD`. Use this one
symbol to differentiate ChakraCore APIs and implementations.

### DIFF
--- a/lib/Jsrt/Jsrt.cpp
+++ b/lib/Jsrt/Jsrt.cpp
@@ -1559,6 +1559,7 @@ CHAKRA_API JsCreateArrayBuffer(_In_ unsigned int byteLength, _Out_ JsValueRef *r
     });
 }
 
+#ifdef _CHAKRACOREBUILD
 CHAKRA_API JsCreateSharedArrayBufferWithSharedContent(_In_ JsSharedArrayBufferContentHandle sharedContents, _Out_ JsValueRef *result)
 {
     return ContextAPIWrapper<true>([&](Js::ScriptContext *scriptContext, TTDRecorder& _actionEntryPopper) -> JsErrorCode {
@@ -1585,7 +1586,7 @@ CHAKRA_API JsGetSharedArrayBufferContent(_In_ JsValueRef sharedArrayBuffer, _Out
         {
             return JsErrorInvalidArgument;
         }
-        
+
         Js::SharedContents**& content = (Js::SharedContents**&)sharedContents;
         *content = Js::SharedArrayBuffer::FromVar(sharedArrayBuffer)->GetSharedContents();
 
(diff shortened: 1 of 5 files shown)


jianchun:
@curtisman @obastemur As I'm limiting/deprecating the other use of `_CHAKRACOREBUILD`, we can actually merge this one (`CHAKRACOREBUILD_`) to use the same symbol. Do you know of any reason to keep using a different one here? If not, I'm thinking to replace all `CHAKRACOREBUILD_` with `_CHAKRACOREBUILD`, so that we'll use only one symbol to identify ChakraCore APIs and ChakraCore builds.

@agarwal-sandeep Questions:
* I left `ChakraDebug.h` out of my `#ifdef`, because the related implementations are not strictly excluded from Chakra builds. If it is inside, then a bunch of JSRT debugging related code/files need to add `#ifdef _CHAKRACOREBUILD` to be ChakraCore exclusive. Would be noisy but maybe a good change to not leave unused code in Chakra?
* What was the intended include sequence for `ChakraDebug.h`? Looks users should just include `ChakraCore.h`, then `ChakraDebug.h` should not include `ChakraCommon.h`.

lib/Jsrt/Jsrt.cpp
```diff
@@ -1607,6 +1608,7 @@ CHAKRA_API JsReleaseSharedArrayBufferContentHandle(_In_ JsSharedArrayBufferConte
         return JsNoError;
     });
 }
+#endif // _CHAKRACOREBUILD
```
obastemur: nit: commented definition doesn't match.

agarwal-sandeep:
@jianchun initially when JsRT debugging code was written we didn't had ChakraCore specific APIs (the plan was to have them in chakra build as well) but now I think we should limit it to ChakraCore only. This should give some binary size reduction in chakra as well.
If we make debugging APIs ChakraCore specific then ChakraDebug.h should not include ChakraCommon.h.


---
In reply to: [316569321](https://github.com/Microsoft/ChakraCore/pull/3400#issuecomment-316569321) [](ancestors = 316569321)

jianchun:
@obastemur Thanks for review! Do you recall the reason of adding `CHAKRACOREBUILD_` previously when we had `_CHAKRACOREBUILD`? Was it because `_CHAKRACOREBUILD` was misused (we had actually defined it for both ChakraCore and Chakra, so it didn't work)? If so, that problem is now resolved. I am considering to:
1. Replace all `CHAKRACOREBUILD_` with `_CHAKRACOREBUILD`. One symbol in future.
2. Add `_CHAKRACOREBUILD` to top-level cmake lists txt so it is always defined for ChakraCore (consistent to Windows build where we define it in default VS props).

obastemur:
LGTM. @jianchun Thanks for fixing this!

jianchun:
@obastemur  Thanks for review!

PR #3402, opened 2017-07-20
Change release/1.7 to RELEASE mode and update bytecode GUID.

dilijev:
Since we'll release soon, skipping PRERELEASE.


dilijev:
@tcare ensured that built-in library bytecode doesn't use the GUID (the GUID slot is set to all 0's IIRC). In any case, I'll regenerate it to double check.

MSLaguana:
Great :)

PR #3408, opened 2017-07-20
Adding Jsrt function JsCopyStringOneByte

MSLaguana:
For scenarios where a string is known to have values fitting in
one byte, this method allows directly copying those values into a
char* buffer rather than having to copy to a uint16_t* buffer
and then to a char* buffer, and treating the string as raw bytes
rather than a valid utf8 encoding.

This will help reduce overhead in some common nodejs use cases.

### DIFF
--- a/lib/Jsrt/ChakraCore.h
+++ b/lib/Jsrt/ChakraCore.h
@@ -679,5 +679,43 @@ CHAKRA_API
         _In_ JsPropertyIdRef propertyId,
         _Out_ bool *hasOwnProperty);
 
+/// <summary>
+///     Write JS string value into char string buffer without a null terminator
+/// </summary>
+/// <remarks>
+///     <para>
+///         When size of the `buffer` is unknown,
+///         `buffer` argument can be nullptr.
+///         In that case, `written` argument will return the length needed.
+///     </para>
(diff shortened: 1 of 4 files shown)


lib/Jsrt/ChakraCore.h
```diff
@@ -679,5 +679,42 @@ CHAKRA_API
         _In_ JsPropertyIdRef propertyId,
         _Out_ bool *hasOwnProperty);
 
+/// <summary>
+///     Write JS string value into char string buffer without a null terminator
```
digitalinfinity: Can you add some more details about the semantics of this API, specifically what happens if the codepoint being copied is not within the char range

MSLaguana: I had addressed that at https://github.com/Microsoft/ChakraCore/pull/3408/files#diff-ffd84896ce29f73bbba0ed9612cf37b2R699 unless you feel that is unclear?

digitalinfinity: Ah yes, that's great. Also clarify that UTF16 encoded codepoints are being truncated in this case. Actually, what happens in the case of a surrogate pair?

Issue #3376, opened 2017-07-17
Yield created with an escape sequence should be an error in strict mode


bterlson:
```js
'use strict';
var \u0079ield = 1
```
Expected: SyntaxError (yield is a reserved identifier)
Actual: No Error

PR #3409, opened 2017-07-20
Fix Issue #3376:  Escaped yield cannot be an identifier in strict mode

curtisman:
The unescape ID scanning fast path already have the check for strict mode for yield.  Refactor the logic and share it with escaped ID scanning code path.
Also fixed the escaped await used as ID in module case as well.


curtisman:
@pleath Can you code review please?

akroshg:
:shipit:

PR #3411, opened 2017-07-21
Fix -stats:rejit for chakra full

agarwal-sandeep:


agarwal-sandeep:
@Penguinwizzard please take a look

Issue #3368, opened 2017-07-17
Incorrect realization of Symbol.xxx.toString()


sunlili:
run `Symbol.species.toString()`
expected result: **Symbol(Symbol.species)**
ChakraCore-release-1.6 result: **Symbol([Symbol.species])**

run `Symbol.match.toString();`
or `Symbol.replace.toString();`
or `Symbol.search.toString();` 
or `Symbol.split.toString();`
expected result: **Symbol(Symbol.match)** ...
ChakraCore-release-1.6 result: **TypeError: Unable to get property 'toString' of undefined or null reference at Global code**

dilijev:
Confirmed issue:

```
>eshost -its -h ch-1.?.?,ch-dev-ext,d8,sm,jsc,node,node-ch issue3368-symbol-tostring.js
## Source
print(Symbol.species.toString())
print(Symbol.match.toString())
print(Symbol.replace.toString())
print(Symbol.search.toString())
print(Symbol.split.toString())


┌────────────┬─────────────────────────────────────────────────────────────────────────────┐
│ ch-1.4.5   │ Symbol([Symbol.species])                                                    │
│ ch-1.6.0   │ TypeError: Unable to get property 'toString' of undefined or null reference │
│ ch-1.5.3   │                                                                             │
│ ch-1.2.3   │                                                                             │
│ ch-1.3.2   │                                                                             │
│ ch-dev-ext │                                                                             │
├────────────┼─────────────────────────────────────────────────────────────────────────────┤
│ d8         │ Symbol(Symbol.species)                                                      │
│ sm         │ Symbol(Symbol.match)                                                        │
│ jsc        │ Symbol(Symbol.replace)                                                      │
│ node       │ Symbol(Symbol.search)                                                       │
│            │ Symbol(Symbol.split)                                                        │
├────────────┼─────────────────────────────────────────────────────────────────────────────┤
│ node-ch    │ Symbol([Symbol.species])                                                    │
│            │ Symbol(Symbol.match)                                                        │
│            │ Symbol(Symbol.replace)                                                      │
│            │ Symbol(Symbol.search)                                                       │
│            │ Symbol(Symbol.split)                                                        │
└────────────┴─────────────────────────────────────────────────────────────────────────────┘
```

curtisman:
The regex symbols are not enabled by default yet.  Only in experimental.
The species one is a bug.

PR #3412, opened 2017-07-21
Fix Issue #3368:  Symbol.species.toString() shouldn't have bracket around the name

curtisman:


MSLaguana:
Looks like you need to update a few baselines to reflect this change as well.

curtisman:
@MSLaguana Nope.  The accessor function name needs to have the bracket.  toStringTag already have similar pattern and defined propertyId for the function name with [] for all the symbols..  So I updated it to create the accessor with the appropriate name

lib/Runtime/Library/JavascriptLibrary.cpp
```diff
@@ -5142,6 +5115,14 @@ namespace Js
         AddAccessorsToLibraryObject(object, propertyId, getterFunction, setterFunction);
     }
 
+    void JavascriptLibrary::AddSpeciesAccessorsToLibraryObject(DynamicObject* object, FunctionInfo * getterFunctionInfo)
+    {
+        if (scriptContext->GetConfig()->IsES6SpeciesEnabled())
```
akroshg: Nit: species have been enabled since th2. We could get rid of this check.

akroshg:
LGTM

aneeshdk:
LGTM

lib/Runtime/Library/JavascriptLibrary.cpp
curtisman: We probably can do a clean up for all the enabled language feature later.

PR #3413, opened 2017-07-21
Fixes #3065 Generator length property needs to be set of the script function it wraps.

### LINKED ISSUE #3065: Generator length property is inconsistently configurable (+ assert failure in debug build)
The length property of a generator is configurable if you call get OwnPropertyDescriptor, but if you configure the property with a getter and setter, only the setter gets called. If the property is accessed, the read-only length value is returned, even if the setter has been called. This is because the generator class overrides GetProperty for the length property, but not SetProperty. I believe the standard says that this property be configurable, but regardless it should be consistent (either configurable is true, and you can set the getter and setter, or it is false and you can't).

```
The following code shows the problem:


function getSloppyGenerator() {
  return function* f(x) {};
}

function test(testFunction) {
  testFunction(getSloppyGenerator());
}


function testReconfigureAsAccessorProperty(f) {
  print(f.length);
  var length = 2;
  print(Object.getOwnPropertyDescriptor(f, "length").configurable)
  Object.defineProperty(f, 'length', {
    get: function() { return length; },
    set: function(v) { print("in set"); length = v; }
  });
  print("set");
  f.length = 3;
  print("get " + f.length);

}


test(testReconfigureAsAccessorProperty);

```

This prints:

1
true
set
in set
get 1

So configurable is true, the setter gets called, but the getter doesn't. It also triggers an assert in the Debug build:

ASSERTION 5843: (/home/user/test_everywhere/ChakraCore/lib/Runtime/./Language/CacheOperators.inl, line 362) objectWithProperty->IsWritable(propertyId) || (isRoot && RootObjectBase::FromVar(objectWithProperty)->IsLetConstGlobal(propertyId)) || JavascriptFunction::IsBuiltinProperty(objectWithProperty, propertyId)
 Failure: ( objectWithProperty->IsWritable(propertyId) || (isRoot && RootObjectBase::FromVar(objectWithProperty)->IsLetConstGlobal(propertyId)) || JavascriptFunction::IsBuiltinProperty(objectWithProperty, propertyId))

atulkatti:
Fixes #3065

### DIFF
--- a/lib/Runtime/Library/JavascriptGeneratorFunction.cpp
+++ b/lib/Runtime/Library/JavascriptGeneratorFunction.cpp
@@ -378,8 +378,23 @@ namespace Js
         return false;
     }
 
+    BOOL JavascriptGeneratorFunction::SetAccessors(PropertyId propertyId, Var getter, Var setter, PropertyOperationFlags flags)
+    {
+        if (propertyId == PropertyIds::length)
+        {
+            return this->scriptFunction->SetAccessors(propertyId, getter, setter, flags);
+        }
+
+        return JavascriptFunction::SetAccessors(propertyId, getter, setter, flags);
+    }
+
     BOOL JavascriptGeneratorFunction::GetAccessors(PropertyId propertyId, Var *getter, Var *setter, ScriptContext * requestContext)
     {
+        if (propertyId == PropertyIds::length)
+        {
+            return this->scriptFunction->GetAccessors(propertyId, getter, setter, requestContext);
+        }
+
         if (propertyId == PropertyIds::caller || propertyId == PropertyIds::arguments)
         {
             // JavascriptFunction has special case for caller and arguments; call DynamicObject:: virtual directly to skip that.
@@ -397,6 +412,11 @@ namespace Js
             return DynamicObject::GetSetter(propertyId, setterValue, info, requestContext);
         }
 
+        if (propertyId == PropertyIds::length)
+        {
+            return this->scriptFunction->GetSetter(propertyId, setterValue, info, requestContext);
+        }
+
         return JavascriptFunction::GetSetter(propertyId, setterValue, info, requestContext);
     }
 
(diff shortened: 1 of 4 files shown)


atulkatti:
@Microsoft/chakra-developers

aneeshdk:
LGTM

akroshg:
What about other properties? If I set any property will it go to the inner wrapper or to the outside generator function?

test/es6/generators-apis.js
```diff
@@ -50,6 +50,39 @@ var tests = [
         }
     },
     {
+        // Regression Test GitHub Issue #3065: Generator length property is inconsistently configurable (+ assert failure in debug build)
```
kunalspathak: can we add test (if not already) for other built-ins and a non-built-in test for same scenario?

lib/Runtime/Library/JavascriptGeneratorFunction.cpp
```diff
@@ -378,8 +378,23 @@ namespace Js
...
     }
 
+    BOOL JavascriptGeneratorFunction::SetAccessors(PropertyId propertyId, Var getter, Var setter, PropertyOperationFlags flags)
+    {
+        if (propertyId == PropertyIds::length)
+        {
+            return this->scriptFunction->SetAccessors(propertyId, getter, setter, flags);
+        }
+
+        return JavascriptFunction::SetAccessors(propertyId, getter, setter, flags);
```
kunalspathak: I see in `GetAccessors`, we directly call `DynamicObject::` methods for `caller` and `arguments`. Should that be done here as well?

atulkatti:
Any other property will be set on the outer generator function.

---
In reply to: [317085201](https://github.com/Microsoft/ChakraCore/pull/3413#issuecomment-317085201) [](ancestors = 317085201)

lib/Runtime/Library/JavascriptGeneratorFunction.cpp
atulkatti: JavascriptFunction doesn't override SetAccessors and this call will currently go to DynamicObject::SetAccessors directly.

---
In reply to: [128839022](https://github.com/Microsoft/ChakraCore/pull/3413#discussion_r128839022) [](ancestors = 128839022)

test/es6/generators-apis.js
atulkatti: There is good coverage for built-in properties like caller in the other tests in this file. I will add a test case for non-built-in property.

---
In reply to: [128838455](https://github.com/Microsoft/ChakraCore/pull/3413#discussion_r128838455) [](ancestors = 128838455)

PR #3415, opened 2017-07-21
slice helper should check for the ES5array or Proxy in the protototype

akroshg:
This way we can identify if the array's content can be affected. If there is sideeffect expected we can take the slower
to finish the logic.


akroshg:
@rajatd please have a look at the c6851ea commit. Rest commits are ported from rs3 branch.

rajatd:
:shipit:

Issue #3261, opened 2017-06-29
NULL character ignored and exception is not thrown


hackvertor:
```javascript
eval('\0');
```
Expected: Illegal character exception

Actual: undefined

Multiple NULL characters throw an exception however it's possible to call a function before the NULL. E.g
```javascript
eval('alert(1)\0');
```

PR #3416, opened 2017-07-21
Fix Issue #3261: Need to detect invalid null character at the end of the source string

curtisman:


curtisman:
@pleath  Please review.

PR #3417, opened 2017-07-22
abort inlining dom fastpath if it didn't get registered

MikeHolman:
OS: 12599786

### DIFF
--- a/lib/Backend/ServerScriptContext.cpp
+++ b/lib/Backend/ServerScriptContext.cpp
@@ -335,13 +335,12 @@ ServerScriptContext::GetEmitBufferManager(bool asmJsManager)
 IR::JnHelperMethod
 ServerScriptContext::GetDOMFastPathHelper(intptr_t funcInfoAddr)
 {
-    IR::JnHelperMethod helper;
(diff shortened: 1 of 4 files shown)


agarwal-sandeep:
:shipit:

lib/Backend/ServerScriptContext.cpp
```diff
@@ -335,13 +335,12 @@ ServerScriptContext::GetEmitBufferManager(bool asmJsManager)
 IR::JnHelperMethod
 ServerScriptContext::GetDOMFastPathHelper(intptr_t funcInfoAddr)
 {
-    IR::JnHelperMethod helper;
+    IR::JnHelperMethod helper = IR::HelperInvalid;
```
rajatd: >HelperInvalid [](start = 36, length = 13)

should we change ScriptContext::GetDOMFastPathHelper accordingly as well?  I'm thinking in-proc jit here, unless the jit always refers to the ServerScriptContext now irrespective of oop or in proc.

MikeHolman: You're right, I should change both.

PR #3421, opened 2017-07-24
OS#12565074: fix CopyProp phase check

sigatrev:
CopyProp is default on,so the phase check should look for it being off. 
Saw no impact on perf tests


rajatd:
This is very good to fix. Make sure to give exprgen a go with this, though; we are enabling optimizations on previously unoptimized tests.

rajatd:
Looks good.

LouisLaf:
Looks good!

PR #3424, opened 2017-07-25
Give NewScFunc the correct nested function index when we transfrom GetCachedFunc to NewScFunc in the backend

rajatd:
As part of stack args optimization, we remove all references to scope object from the IR. One of the instructions referencing the scope object (cached scope object in this case) is GetCachedFunc, which loads the nested function object from a particulat index from the cached scope. In the deadstore phase, we change it to NewScFunc and create a new function object instead of using the cached one.

Slot values in the cached scope and the function body's nested function array may correspond to different nested functions. In this situation, we were incorrectly setting the index of the nested function on NewScFunc by reusing the index from GetCachedFunc.


dilijev:
Please update GUID in lib\Runtime\ByteCode\ByteCodeCacheReleaseFileVersion.h and rebase on top of the latest release/1.6 to avoid the merge conflict on bytecode.

lib/Runtime/ByteCode/ByteCodeSerializer.cpp
```diff
@@ -63,6 +63,8 @@ namespace Js
     const int magicEndOfAsmJsModuleInfo = *(int*)"]asmmodinfo";
     const int magicStartOfPropIdsOfFormals = *(int*)"propIdOfFormals[";
     const int magicEndOfPropIdsOfFormals = *(int*)"]propIdOfFormals";
+    const int magicStartOfSlotIdToNestedIndexArray = *(int*)"slotIdToNestedIndexArray[";
```
digitalinfinity: Lol, this magic part is not what it's supposed to be :) your usage accidentally works- the dbgscope ones dont- the magic reinterprets the first 4 chars of a string as a uint32 so you really want only 4 char strings- in the case of magicStartOfDebuggerScopes and magicStartOfDebuggerScopeProperties, the first 4 chars are the same- really all of these longer than 4 chars should be fixed to be unique 4 char strings (3 + [)

digitalinfinity:
>         PrependStruct<SerializedFieldList>(builder, _u("Serialized Field List"), &definedFields);

Instead of using a byte for whether your field exists, can you add a bit in this struct?

---
Refers to: lib/Runtime/ByteCode/ByteCodeSerializer.cpp:2228 in db2510c. [](commit_id = db2510c52c14f73bcc13522303fb3177eb957302, deletion_comment = False)

lib/Backend/BackwardPass.cpp
```diff
@@ -3062,8 +3062,11 @@ BackwardPass::DeadStoreOrChangeInstrForScopeObjRemoval(IR::Instr ** pInstrPrev)
                 {
                     instr->m_opcode = Js::OpCode::NewScFunc;
                     IR::Opnd * intConstOpnd = instr->UnlinkSrc2();
+                    Assert(intConstOpnd->IsIntConstOpnd());
 
-                    instr->ReplaceSrc1(intConstOpnd);
+                    uint nestedFuncIndex = instr->m_func->GetJITFunctionBody()->GetNestedFuncIndexForSlotIdInCachedScope(intConstOpnd->AsIntConstOpnd()->AsUint32());
```
LouisLaf: Might as well Free intConstOpnd...

LouisLaf:
:shipit:

PR #3425, opened 2017-07-25
Fix possible infinite loop in String.raw()

Cellule:
I was able to construct a use case where we would end up in an infinite loop.
It takes forever to run to just overflow the uint32, but I figured we should at least avoid possible infinite loops.

A possible solution could be to just cap the length to loop through to UINT32_MAX, but it would be technically not spec compliant.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/3425)

### DIFF
--- a/lib/Runtime/Library/JavascriptString.cpp
+++ b/lib/Runtime/Library/JavascriptString.cpp
@@ -1560,22 +1560,30 @@ case_2:
         // strcon2 /
         // expr2   \__ step 3
         // strcon3 /
-        for (uint32 i = 1; i < length; ++i)
+        const auto append = [&] (Var var)
         {
-            // First append the next substitution expression
-            // If we have an arg at [i+1] use that one, otherwise empty string (which is nop)
-            if (i+1 < args.Info.Count)
+            JavascriptString* string = JavascriptConversion::ToString(var, scriptContext);
+            stringBuilder.Append(string);
+        };
+        uint32 loopMax = length >= UINT_MAX ? UINT_MAX-1 : (uint32)length;
+        uint32 i = 1, argsCount = args.Info.Count;
+        for (; i < loopMax; ++i)
+        {
+            // First append the next substitution expression if available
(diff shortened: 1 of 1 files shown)


Cellule:
@Microsoft/chakra-developers Can someone review this please ?

lib/Runtime/Library/JavascriptString.cpp
```diff
@@ -1548,7 +1548,7 @@ namespace Js
         }
 
         // We aren't going to bail early so let's create a StringBuilder and put the first raw string in there
-        CompoundString::Builder<64 * sizeof(void *) / sizeof(char16)> stringBuilder(scriptContext);
+        CompoundString::Builder<64 * sizeof(void *) / sizeof(char16)> stringBuilder = (scriptContext);
```
dilijev: Why change the clear invocation of a constructor to assignment syntax for constructing the object?

curtisman:
Our JavascriptString has a MaxCharCount = 2^31-1 limit.  We should use that?

Cellule:
Unfortunately, the input is not a string. It's technically anything with a `length` property.
We run out of memory really quick because of our limit, but I was able to construct a case that doesn't run out of memory and runs into an infinite loop.

```js
var a = new Proxy({}, {
  get(target, name) {
    if (name === "length") return 4294967297
    if (name === "4294967297") return "my actual string"
    return "";
  }
});
console.log(String.raw({raw: a}));
```
_I write the use case from memory, there might be some syntax errors_

One could argue that we don't really care about this use case however ;)

MSLaguana:
For what its worth, this case seems to also cause issues for v8. In an unrelated situation, I created a reference cycle of prototypes using a proxy as well that resulted in an infinite loop; it seems like proxy gives a lot of scope for poor behavior.

curtisman:
Well, since user can just write script that infinite loop, we shouldn't guard against it?

PR #3429, opened 2017-07-25
Fixes #586 JsCreateNamedFunction should create a tracked PropertyId for name to ensure toString() works as expected.

### LINKED ISSUE #586: JsCreateNamedFunction toString result is wrong
JsCreateNamedFunction sets "functionNameId" as name. That is wrong in current code. Elsewhere we typically call 

```
if (scriptContext->GetConfig()->IsES6FunctionNameEnabled())
{
    object->SetPropertyWithAttributes(PropertyIds::name, functionName,
                                      PropertyConfigurable, nullptr);
}
```

"functionNameId" is used as toString result cache. Setting "functionNameId" to a string incorrectly makes the runtime use that string as toString result. We get `custom_name` instead of `function custom_name() { [native code] }`.

atulkatti:
Fixes #586

### DIFF
--- a/lib/Runtime/Debug/DiagObjectModel.cpp
+++ b/lib/Runtime/Debug/DiagObjectModel.cpp
@@ -269,7 +269,7 @@ namespace Js
             }
             else
             {
-                Js::JavascriptString *builtInName = returnValue->calledFunction->GetDisplayName();
+                Js::JavascriptString *builtInName = ParseFunctionName(returnValue->calledFunction->GetDisplayName(), pResolvedObject->scriptContext);
                 swprintf_s(finalName, RETURN_VALUE_MAX_NAME, _u("[%s returned]"), builtInName->GetSz());
             }
             pResolvedObject->obj = returnValue->returnedValue;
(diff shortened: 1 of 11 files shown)


atulkatti:
@Microsoft/chakra-developers

lib/Runtime/Library/JavascriptLibrary.cpp
```diff
@@ -6646,12 +6646,31 @@ namespace Js
...
+            PropertyId pid = TaggedInt::ToInt32(name);
+            if (!scriptContext->IsTrackedPropertyId(pid))
+            {
+                scriptContext->TrackPid(pid);
+            }
+        }
+#endif
         JavascriptExternalFunction* function = this->CreateIdMappedExternalFunction(entryPoint, stdCallFunctionWithDeferredPrototypeType);
-        function->SetFunctionNameId(nameId);
+        function->SetFunctionNameId(functionNameOrId);
```
akroshg: >functionNameOrId [](start = 36, length = 16)

Should assert or failfast if not taggedint at this point.

lib/Runtime/Library/RuntimeFunction.h
```diff
@@ -18,10 +18,13 @@ namespace Js
 
         void SetFunctionNameId(Var nameId);
 
-        // this is for cached source string for the function. Possible values are:
+        // This is for cached source string for the function. Possible values are:
         // NULL; initialized for anonymous methods.
         // propertyId in Int31 format; this is used for fastDOM function as well as library function
         // JavascriptString: composed using functionname from the propertyId, or fixed string for anonymous functions.
+        // NOTE: This has a side-effect that after toString() is called for the first time on a built-in function the functionNameId gets replaced with a string like "function foo() { native code }".
```
akroshg: Does this comment require at point?

lib/Runtime/Debug/DiagObjectModel.cpp
```diff
@@ -285,6 +285,52 @@ namespace Js
...
+
+        // Look for the left parenthesis, if we don't find one; return the original string. If we find a left parenthesis, match the rest of the function body pattern, if we don't find an exact match; return the original string.
+        const char16* parenChar = wmemchr(displayNameBuffer, '(', displayNameBufferLength);
+        if ((parenChar == nullptr) || (wmemcmp(parenChar, JS_DISPLAY_STRING_FUNCTION_BODY, _countof(JS_DISPLAY_STRING_FUNCTION_BODY)) != 0))
+        {
+            return displayName;
+        }
+
+        charcount_t actualFunctionNameLength = displayNameBufferLength - templateStringLength;
+        if (actualFunctionNameLength == 0)
```
akroshg: >if (actualFunctionNameLength == 0) [](start = 8, length = 34)

if you check displayNameBufferLength <= templateStringLength above this can be removed, no?

lib/Runtime/Debug/DiagObjectModel.cpp
```diff
@@ -285,6 +285,52 @@ namespace Js
...
+        const charcount_t templateStringLength = funcStringLength + _countof(JS_DISPLAY_STRING_FUNCTION_BODY) - 1; // discount the ending null character in string literal
+        // If the string doesn't meet our expected format; return the original string.
+        if (displayNameBufferLength < templateStringLength || (wmemcmp(displayNameBuffer, JS_DISPLAY_STRING_FUNCTION_HEADER, funcStringLength) != 0))
+        {
+            return displayName;
+        }
+
+        // Look for the left parenthesis, if we don't find one; return the original string. If we find a left parenthesis, match the rest of the function body pattern, if we don't find an exact match; return the original string.
+        const char16* parenChar = wmemchr(displayNameBuffer, '(', displayNameBufferLength);
+        if ((parenChar == nullptr) || (wmemcmp(parenChar, JS_DISPLAY_STRING_FUNCTION_BODY, _countof(JS_DISPLAY_STRING_FUNCTION_BODY)) != 0))
```
akroshg: why are we checking the body, why not just check the header? just checking the paren not sufficient?

lib/Runtime/Library/RuntimeFunction.h
atulkatti: I added it so that anyone using this field in future is aware of this behavior and can avoid a bug like we are currently fixing in debugger.

---
In reply to: [129430780](https://github.com/Microsoft/ChakraCore/pull/3429#discussion_r129430780) [](ancestors = 129430780)

lib/Runtime/Debug/DiagObjectModel.cpp
atulkatti: I had it like that before but Doug pointed out a spec consideration where this could sometimes be "function () { [native code] }". In this case the function name is blank. This happen for built-in or external function from what I have seen so far. But I have handled that case in case we see something like that in future.

---
In reply to: [129432508](https://github.com/Microsoft/ChakraCore/pull/3429#discussion_r129432508) [](ancestors = 129432508)

lib/Runtime/Debug/DiagObjectModel.cpp
atulkatti: I chose to a be a bit stricter about this. But I think we can relax this a little bit.

---
In reply to: [129433553](https://github.com/Microsoft/ChakraCore/pull/3429#discussion_r129433553) [](ancestors = 129433553)

akroshg:
:shipit:

PR #3435, opened 2017-07-26
OS12814968: ASSERTION:(lib\runtime\Base/FunctionBody.h) IsFunctionBody()

suwc:
Function body not available for deferred-parse function.
Add additional checking.

### DIFF
--- a/lib/Jsrt/JsrtDiag.cpp
+++ b/lib/Jsrt/JsrtDiag.cpp
@@ -499,34 +499,43 @@ CHAKRA_API JsDiagGetFunctionPosition(
 
         Js::ScriptFunction* jsFunction = Js::ScriptFunction::FromVar(function);
 
-        Js::FunctionBody* functionBody = jsFunction->GetFunctionBody();
-        if (functionBody != nullptr)
+        BOOL fParsed = jsFunction->GetParseableFunctionInfo()->IsFunctionParsed();
+        if (!fParsed)
         {
-            Js::Utf8SourceInfo* utf8SourceInfo = functionBody->GetUtf8SourceInfo();
-            if (utf8SourceInfo != nullptr && !utf8SourceInfo->GetIsLibraryCode())
+            Js::JavascriptFunction::DeferredParseCore(&jsFunction, fParsed);
+        }
+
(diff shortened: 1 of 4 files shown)


lib/Runtime/Base/FunctionInfo.cpp
```diff
@@ -64,7 +64,7 @@ namespace Js
     FunctionBody *
     FunctionInfo::GetFunctionBody() const
     {
-        return functionBodyImpl == nullptr ? nullptr : functionBodyImpl->GetFunctionBody();
+        return (functionBodyImpl && functionBodyImpl->IsFunctionBody()) ? functionBodyImpl->GetFunctionBody() : nullptr;
```
agarwal-sandeep: There is no harm in putting this additional check here as we already have an assert in FunctionProxy::GetFunctionBody but I think we should rather fix JsDiagGetFunctionPosition in either of 2 ways
1. If the function is not parsed, parse it and then return the information (preferable as user have no knowledge of defer-parse or re-deferal)
2. Check if function have body and if not return failure code

curtisman: Agreed.  I'd much rather you do that in JsDiagGetFunctionPosition and separately change the Assert in GetFunctionBody to fail fast instead.  Or change the API to always to TryGetFunctionBody which a check and make the API explicit, and eliminate all the extra check in the caller.

lib/Jsrt/JsrtDiag.cpp
```diff
@@ -498,6 +498,10 @@ CHAKRA_API JsDiagGetFunctionPosition(
         }
 
         Js::ScriptFunction* jsFunction = Js::ScriptFunction::FromVar(function);
+        if (!jsFunction->GetParseableFunctionInfo()->IsFunctionParsed())
+        {
+            Js::JavascriptFunction::DeferredParse(&jsFunction);
```
agarwal-sandeep: Use DeferredParseCore and assert on fParsed

agarwal-sandeep: @pleath will this work for all nested and re-deferral scenarios?

PR #3436, opened 2017-07-26
OS# 12840992: ARM: Wrong accessor flag offset passed for Getter/Setter inline cache check

agarwal-sandeep:
Fix is in function GenerateFlagInlineCacheCheckForGetterSetter (lib\Backend\arm\LowerMD.cpp)
Removed un-used function GenerateFlagInlineCacheCheckForNoGetterSetter

Gives 5% overall gain in speedometer on ARM


kunalspathak:
:shipit:

PR #3437, opened 2017-07-26
Fixes #3430 overzealous type cast

### LINKED ISSUE #3430: Issues found when building with MSVC + /permissive- - 3
This commit https://github.com/Microsoft/ChakraCore/commit/9020556559a551daccc5df885cfd19f34a4716f5 added new code which casts string literal to 'uint16 *'. This is not allowed by the standard and /permissive- will catch it:
    JsRTApiTest.cpp(2151): error C2440: 'reinterpret_cast': cannot convert from 'const wchar_t [9]' to 'uint16_t *'

The fix is (JsCreateStringUtf16 accepts 'const uint16_t *'):

JsCreateStringUtf16(reinterpret_cast<uint16_t*>(_u("0\x10\x80\xa9\uabcd\U000104377")), 

->

JsCreateStringUtf16(reinterpret_cast<const uint16_t*>(_u("0\x10\x80\xa9\uabcd\U000104377")),

MSLaguana:
The lack of a `const` caused complaints with more stringent compiler flags, as pointed out in #3430

### DIFF
--- a/bin/NativeTests/JsRTApiTest.cpp
+++ b/bin/NativeTests/JsRTApiTest.cpp
@@ -2156,7 +2156,7 @@ namespace JsRTApiTest
         size_t written = 0;
         char buf[10] = {0};
         JsValueRef value;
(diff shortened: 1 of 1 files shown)


kunalspathak:
:shipit:

PR #3443, opened 2017-07-27
Make expensive entryRecord on Stack check is Debug Only

obastemur:
Improves JSRT API calls performance >10%

### DIFF
--- a/lib/Runtime/Base/ThreadContext.cpp
+++ b/lib/Runtime/Base/ThreadContext.cpp
@@ -1741,7 +1741,7 @@ ThreadContext::ProbeStack(size_t size, Js::ScriptContext *scriptContext, PVOID r
         PlatformAgnostic::PerfTrace::WritePerfMap();
     }
 #endif
-    
+
     // BACKGROUND-GC TODO: If we're stuck purely in JITted code, we should have the
     // background GC thread modify the threads stack limit to trigger the runtime stack probe
     if (this->callDispose && this->recycler->NeedDispose())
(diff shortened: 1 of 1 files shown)


obastemur:
Perf comparison made with has/get/set API calls to regular objects

MikeHolman:
I'm surprised this is showing as so expensive. Where is the cost?

obastemur:
@MikeHolman it unfortunately is expensive on xplat. On Windows, not this much.

lib/Runtime/Base/ThreadContext.cpp
```diff
@@ -2206,11 +2206,13 @@ ThreadContext::PushEntryExitRecord(Js::ScriptEntryExitRecord * record)
         lastRecord->hasReentered = true;
         record->next = lastRecord;
 
+#if defined(JSRT_VERIFY_RUNTIME_STATE) || defined(DEBUG)
         // these are on stack, which grows down. if this condition doesn't hold, then the list somehow got messed up
         if (!IsOnStack(lastRecord) || (uintptr_t)record >= (uintptr_t)lastRecord)
         {
             EntryExitRecord_Corrupted_fatal_error();
         }
```
kunalspathak: could you please explain why this check is not necessary for chakracore?

MSLaguana:
I took a look at the assembly on xplat, and I think that the function at https://github.com/Microsoft/ChakraCore/blob/cbe1eee0d670123dbea3d620b83c5fc7f7488ae5/pal/src/thread/pal_thread.cpp#L2823 isn't terribly well optimised. The `GetCurrentSP` function is defined in asm and essentially just copies `rsp`: https://github.com/Microsoft/ChakraCore/blob/master/pal/src/arch/i386/asmhelpers.S#L32

However, that is kept as a function call rather than having the `IsAddressOnStack` function examine `rsp` itself, and it also looks like there's a lot of assembly for essentially `return addr < highLimit && addr > rsp`.

MSLaguana:
Also, on acmeair I see ~3% perf improvement, but it seems to be consistent.

kunalspathak:
@MSLaguana  - Is the perf improvement with @obastemur changes or by optimizing `GetCurrentSP()` the way you suggested?

MSLaguana:
Using @obastemur's changes here. I'm not sure how best to improve the things I pointed out

MikeHolman:
The issue this is trying to prevent is that some old ThreadContext wasn't removed from the list. In that case, IsOnStack would still be true, so the benefit of that check is marginal.

If IsOnStack is expensive on Linux, I'd suggest removing only that check (and maybe only removing it on Linux, since it isn't an issue on Windows). We have actually seen this hit before, and without the failfast such problems could be exploitable.

obastemur:
> However, that is kept as a function call rather than having the IsAddressOnStack function examine rsp itself

@MSLaguana Majority of the perf loss comes from thread locals penalty and we have to keep this call separate. We don't have much to do there :/

@MikeHolman Thanks for the details. I will update the PR accordingly.

Issue #2437, opened 2017-01-31
Yield register change after reparse for debugger


Cellule:
If the debugger is attached while running a yielded function, it is possible to modify the YieldRegister while reparsing for debugger.
When resuming the yield, we use the incorrect register and read wrong data resulting in undefined behavior (crash in the repro code).

Repro Code
```js
function attach() {
  return new Promise(function (r) {
    WScript.Attach(r);
  });
}

async function mainTest() {
  for (let i = 0; i < 1; ++i) {
    await attach();
  }
}
mainTest();
```

MikeHolman:
@agarwal-sandeep FYI

PR #3445, opened 2017-07-27
Keep the same register allocation path for debug and non-debug mode in generators

aneeshdk:
In debug mode we have different register allocation logic for inner scope slots. In case of generators if the debugger is attached dynamically before the generator completes then this can cause issues. Current fix is to avoid the different paths for register allocation.

### DIFF
--- a/lib/Runtime/ByteCode/ByteCodeEmitter.cpp
+++ b/lib/Runtime/ByteCode/ByteCodeEmitter.cpp
@@ -1744,9 +1744,13 @@ void ByteCodeGenerator::FinalizeRegisters(FuncInfo * funcInfo, Js::FunctionBody 
 
     this->SetClosureRegisters(funcInfo, byteCodeFunction);
 
-    if (this->IsInDebugMode())
+    if (this->IsInDebugMode() || byteCodeFunction->IsCoroutine())
     {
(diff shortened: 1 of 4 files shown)


aneeshdk:
@curtisman @akroshg Can you please take a look?

lib/Runtime/ByteCode/ByteCodeEmitter.cpp
```diff
@@ -1744,7 +1744,7 @@ void ByteCodeGenerator::FinalizeRegisters(FuncInfo * funcInfo, Js::FunctionBody
 
     this->SetClosureRegisters(funcInfo, byteCodeFunction);
 
-    if (this->IsInDebugMode())
+    if (this->IsInDebugMode() || byteCodeFunction->IsCoroutine())
```
agarwal-sandeep: Can you add a test case?

aneeshdk:
https://github.com/Microsoft/ChakraCore/issues/2437

akroshg:
LGTM- it would be nice to have some kind of validation where you can say that between non-debug to debug phase same regslot is used.

curtisman:
this would cause some scoped variable to display properly in generators, right?  If so, please add a TODO?

PR #3446, opened 2017-07-27
OS: 12891717 Missing conversion for CopyOnAccess Array on ForInObjectEnumerator

akroshg:
We get copyonaccessnativeintarray when ForInObjectEnumerator. This happens when we are at proxy. Fixed that by converting that array to normal native int array.

### DIFF
--- a/lib/Runtime/Library/JavascriptArray.cpp
+++ b/lib/Runtime/Library/JavascriptArray.cpp
@@ -12143,6 +12143,10 @@ Case0:
 
     JavascriptEnumerator * JavascriptArray::GetIndexEnumerator(EnumeratorFlags flags, ScriptContext* requestContext)
     {
+#if ENABLE_COPYONACCESS_ARRAY
+        JavascriptLibrary::CheckAndConvertCopyOnAccessNativeIntArray<Var>(this);
+#endif
+
         if (!!(flags & EnumeratorFlags::SnapShotSemantics))
         {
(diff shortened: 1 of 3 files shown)


akroshg:
@suwc @kunalspathak please have a look.

suwc:
:shipit:

lib/Runtime/Library/JavascriptArray.cpp
```diff
@@ -12143,6 +12143,10 @@ namespace Js
 
     JavascriptEnumerator * JavascriptArray::GetIndexEnumerator(EnumeratorFlags flags, ScriptContext* requestContext)
     {
+#if ENABLE_COPYONACCESS_ARRAY
+            JavascriptLibrary::CheckAndConvertCopyOnAccessNativeIntArray<Var>(this);
```
suwc: >      [](start = 7, length = 5)

nit:tab

lib/Runtime/Library/JavascriptArray.cpp
```diff
@@ -12152,6 +12156,9 @@ namespace Js
 
     BOOL JavascriptArray::GetNonIndexEnumerator(JavascriptStaticEnumerator * enumerator, ScriptContext* requestContext)
     {
+#if ENABLE_COPYONACCESS_ARRAY
+        JavascriptLibrary::CheckAndConvertCopyOnAccessNativeIntArray<Var>(this);
```
kunalspathak: Can we also add a check inside `JavsacriptArray::GetEnumerator()`?

PR #3447, opened 2017-07-27
Fix #3438: AutoInitLibraryCodeScope: hide Intl.js initialization from debugger in addition to profiler.

### LINKED ISSUE #3438: Intl.js: when the outer function of Intl.js is called (Intl lazy-init), it appears in the debugger dump
It seems an internal anonymous function call during the constructor's execution is being shown in the debugger output.

Run the following with "ch -Intl -debugLaunch"
```js
var options = { ca: "gregory", hour12: true, timeZone:"UTC" }; 
var dateFormat1 = new Intl.DateTimeFormat("en-US", options);    /**bp:resume('step_over');locals();**/
WScript.Echo("");
```
**Actual output:** 
    "functionCallsReturn": {
          "[Anonymous function returned]": "undefined undefined",
          "[Intl.DateTimeFormat returned]": "Object {...}"
**Expected output:**
    "functionCallsReturn": {
          "[Intl.DateTimeFormat returned]": "Object {...}"

dilijev:
Fxies #3438

### DIFF
--- a/lib/Runtime/Base/ScriptContext.cpp
+++ b/lib/Runtime/Base/ScriptContext.cpp
@@ -3981,7 +3981,6 @@ namespace Js
             }
         }
 
-
         __TRY_FINALLY_BEGIN // SEH is not guaranteed, see the implementation
         {
             Assert(!function->IsScriptFunction() || function->GetFunctionProxy());
@@ -3998,7 +3997,7 @@ namespace Js
             OUTPUT_VERBOSE_TRACE(Js::DebuggerPhase, _u("DebugProfileProbeThunk: calling function: %s isWrapperRegistered=%d useDebugWrapper=%d\n"),
                 function->GetFunctionInfo()->HasBody() ? function->GetFunctionBody()->GetDisplayName() : _u("built-in/library"), AutoRegisterIgnoreExceptionWrapper::IsRegistered(scriptContext->GetThreadContext()), useDebugWrapper);
 
-            if (scriptContext->IsScriptContextInDebugMode())
+            if (scriptContext->IsDebuggerRecording())
             {
                 scriptContext->GetDebugContext()->GetProbeContainer()->StartRecordingCall();
             }
@@ -4071,7 +4070,7 @@ namespace Js
             }
 #endif
 
-            if (scriptContext->IsScriptContextInDebugMode())
+            if (scriptContext->IsDebuggerRecording())
             {
                 scriptContext->GetDebugContext()->GetProbeContainer()->EndRecordingCall(aReturn, function);
             }
@@ -5801,6 +5800,23 @@ void ScriptContext::RegisterPrototypeChainEnsuredToHaveOnlyWritableDataPropertie
         return false;
     }
 
+    bool ScriptContext::IsDebuggerRecording() const
+    {
+        if (this->debugContext != nullptr)
+        {
+            return this->GetDebugContext()->IsDebuggerRecording();
+        }
+        return false;
+    }
+
+    void ScriptContext::SetIsDebuggerRecording(bool isDebuggerRecording)
+    {
+        if (this->debugContext != nullptr)
+        {
+            this->GetDebugContext()->SetIsDebuggerRecording(isDebuggerRecording);
+        }
+    }
+
     bool ScriptContext::IsIntlEnabled()
     {
         if (GetConfig()->IsIntlEnabled())
@@ -5813,7 +5829,6 @@ void ScriptContext::RegisterPrototypeChainEnsuredToHaveOnlyWritableDataPropertie
         }
         return false;
     }
-
 
 #ifdef INLINE_CACHE_STATS
     void ScriptContext::LogCacheUsage(Js::PolymorphicInlineCache *cache, bool isGetter, Js::PropertyId propertyId, bool hit, bool collision)
--- a/lib/Runtime/Debug/DebugContext.cpp
+++ b/lib/Runtime/Debug/DebugContext.cpp
@@ -10,7 +10,8 @@ namespace Js
         scriptContext(scriptContext),
         hostDebugContext(nullptr),
         diagProbesContainer(nullptr),
-        debuggerMode(DebuggerMode::NotDebugging)
+        debuggerMode(DebuggerMode::NotDebugging),
+        isDebuggerRecording(true)
     {
         Assert(scriptContext != nullptr);
     }
--- a/lib/Runtime/Debug/DebugContext.h
+++ b/lib/Runtime/Debug/DebugContext.h
@@ -51,10 +51,14 @@ namespace Js
         void SetHostDebugContext(HostDebugContext * hostDebugContext);
 
         void SetDebuggerMode(DebuggerMode mode);
+
         bool IsDebugContextInNonDebugMode() const { return this->debuggerMode == DebuggerMode::NotDebugging; }
         bool IsDebugContextInDebugMode() const { return this->debuggerMode == DebuggerMode::Debugging; }
         bool IsDebugContextInSourceRundownMode() const { return this->debuggerMode == DebuggerMode::SourceRundown; }
         bool IsDebugContextInSourceRundownOrDebugMode() const { return IsDebugContextInSourceRundownMode() || IsDebugContextInDebugMode(); }
+
+        bool IsDebuggerRecording() const { return this->isDebuggerRecording; }
+        void SetIsDebuggerRecording(bool isDebuggerRecording) { this->isDebuggerRecording = isDebuggerRecording; }
 
         ProbeContainer* GetProbeContainer() const { return this->diagProbesContainer; }
 
@@ -63,8 +67,9 @@ namespace Js
     private:
         ScriptContext * scriptContext;
         HostDebugContext* hostDebugContext;
-        DebuggerMode debuggerMode;
         ProbeContainer* diagProbesContainer;
+        DebuggerMode debuggerMode;
+        bool isDebuggerRecording;
 
         // Private Functions
         void WalkAndAddUtf8SourceInfo(Js::Utf8SourceInfo* sourceInfo, JsUtil::List<Js::Utf8SourceInfo *, Recycler, false, Js::CopyRemovePolicy, RecyclerPointerComparer> *utf8SourceInfoList);
--- a/lib/Runtime/Library/IntlEngineInterfaceExtensionObject.cpp
+++ b/lib/Runtime/Library/IntlEngineInterfaceExtensionObject.cpp
@@ -411,7 +411,6 @@ namespace Js
                     AssertMsg(false, "Not a valid intlInitializationType.");
                     // fall thru
                 case IntlInitializationType::Intl:
-
                     IfCOMFailIgnoreSilentlyAndReturn(globAdapter->EnsureNumberFormatObjectsInitialized(library));
                     IfCOMFailIgnoreSilentlyAndReturn(globAdapter->EnsureDateTimeFormatObjectsInitialized(library));
                     initType = scriptContext->GetLibrary()->CreateStringFromCppLiteral(_u("Intl"));
@@ -437,8 +436,10 @@ namespace Js
             {
                 scriptContext->RegisterScript(function->GetFunctionProxy());
             }
-            // Mark we are profiling library code already, so that any initialization library code called here won't be reported to profiler
-            AutoProfilingUserCode autoProfilingUserCode(scriptContext->GetThreadContext(), /*isProfilingUserCode*/false);
+
+            // Mark we are profiling library code already, so that any initialization library code called here won't be reported to profiler.
+            // Also tell the debugger not to record events during intialization so that we don't leak information about initialization.
+            AutoInitLibraryCodeScope autoInitLibraryCodeScope(scriptContext);
 
             Js::Var args[] = { scriptContext->GetLibrary()->GetUndefined(), scriptContext->GetLibrary()->GetEngineInterfaceObject(), initType };
             Js::CallInfo callInfo(Js::CallFlags_Value, _countof(args));
--- a/lib/Runtime/Base/ScriptContext.h
+++ b/lib/Runtime/Base/ScriptContext.h
@@ -500,6 +500,10 @@ namespace Js
         bool IsScriptContextInNonDebugMode() const;
         bool IsScriptContextInDebugMode() const;
         bool IsScriptContextInSourceRundownOrDebugMode() const;
+
+        bool IsDebuggerRecording() const;
+        void SetIsDebuggerRecording(bool isDebuggerRecording);
+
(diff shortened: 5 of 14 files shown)


lib/Runtime/Debug/DebugContext.h
```diff
@@ -37,6 +37,9 @@ namespace Js
         // The debugger is running which means that the engine is
         // running in interpreted mode.
         Debugging,
+
+        // The debugger is running but recording is suspended.
+        RecordingSuspended,
```
agarwal-sandeep: I would suggest not to add a DebuggerMode but use a flag on DebugContext the way kyle did for isReparsingSource
[https://github.com/Microsoft/ChakraCore/commit/1c2d2591d4a73415d71d468264dd8e5676862255](https://github.com/Microsoft/ChakraCore/commit/1c2d2591d4a73415d71d468264dd8e5676862255)

agarwal-sandeep: Further convert that bool to a bit mask

---
In reply to: [129982675](https://github.com/Microsoft/ChakraCore/pull/3447#discussion_r129982675) [](ancestors = 129982675)

dilijev: Okay I'll give that a try!

lib/Runtime/Library/IntlEngineInterfaceExtensionObject.cpp
```diff
@@ -446,7 +445,12 @@ namespace Js
             // Clear disable implicit call bit as initialization code doesn't have any side effect
             Js::ImplicitCallFlags saveImplicitCallFlags = scriptContext->GetThreadContext()->GetImplicitCallFlags();
             scriptContext->GetThreadContext()->ClearDisableImplicitFlags();
-            JavascriptFunction::CallRootFunctionInScript(function, Js::Arguments(callInfo, args));
+
+            {
+                AutoSuspendDebuggerRecording autoSuspendDebuggerRecording(scriptContext);
```
agarwal-sandeep: >autoSuspendDebuggerRecording [](start = 45, length = 28)

Same problem will be in promise as well?

agarwal-sandeep: >autoSuspendDebuggerRecording [](start = 45, length = 28)

Instead of AutoSuspendDebuggerRecording you can rename AutoProfilingUserCode and use that instead. The purpose of AutoProfilingUserCode is same (both profiler and debugger should be not tracking internal library code)

lib/Runtime/Library/IntlEngineInterfaceExtensionObject.cpp
```diff
@@ -446,7 +445,12 @@ namespace Js
...
+                JavascriptFunction::CallRootFunctionInScript(function, Js::Arguments(callInfo, args));
```
agarwal-sandeep: >CallRootFunctionInScript [](start = 36, length = 24)

Can this call user code down the line? if so you won't be recording that?

dilijev: This is the lazy-initialization of the Intl object. It cannot call user code, and everything about it is library-code-internal and unobservable by design.

lib/Runtime/Library/IntlEngineInterfaceExtensionObject.cpp
dilijev: Okay. I'll refactor to use the same object.

lib/Runtime/Debug/DebugContext.cpp
```diff
@@ -139,6 +140,18 @@ namespace Js
         this->debuggerMode = mode;
     }
 
+    void DebugContext::SuspendRecording()
+    {
+        Assert(this->debuggerMode == DebuggerMode::Debugging);
```
agarwal-sandeep: >this->debuggerMode == DebuggerMode::Debugging [](start = 15, length = 45)

Use IsDebugContextInDebugMode() instead

lib/Runtime/Debug/DebugContext.h
dilijev: Bit mask would be good. I'm making this change on 1.6 where Kyle's change is not present. Perhaps factoring the two flags together after they are merged is a better time? @kfarnung

lib/Runtime/Base/ScriptContext.cpp
```diff
@@ -3998,7 +3998,7 @@ namespace Js
             OUTPUT_VERBOSE_TRACE(Js::DebuggerPhase, _u("DebugProfileProbeThunk: calling function: %s isWrapperRegistered=%d useDebugWrapper=%d\n"),
                 function->GetFunctionInfo()->HasBody() ? function->GetFunctionBody()->GetDisplayName() : _u("built-in/library"), AutoRegisterIgnoreExceptionWrapper::IsRegistered(scriptContext->GetThreadContext()), useDebugWrapper);
 
-            if (scriptContext->IsScriptContextInDebugMode())
+            if (scriptContext->IsScriptContextInDebugRecordingMode())
```
agarwal-sandeep: >IsScriptContextInDebugRecordingMode [](start = 31, length = 35)

I think we don't need this method on ScriptContext. Just get the DebugContext and use the method on DebugContext as most of time we will be recording call and call StartRecordingCall

lib/Runtime/Debug/DebugContext.cpp
dilijev: Done

lib/Runtime/Base/ScriptContext.cpp
dilijev: Early on I had some trouble with type visibility in compilation units which prevented using methods on DebugContext. I'll try again.

lib/Runtime/Debug/DebugContext.h
agarwal-sandeep: @kfarnung why we didn't made the isReparsingSource change in release/1.6 or RS3?

lib/Runtime/Library/IntlEngineInterfaceExtensionObject.cpp
dilijev: AutoProfilingUserCode is related to ThreadContext and AutoSuspendDebuggerRecording is in ScriptContext. Having trouble with types in compilation.

lib/Runtime/Debug/DebugContext.h
```diff
@@ -51,10 +51,17 @@ namespace Js
...
+        void ResumeRecording();
+
+        DebuggerMode GetDebuggerMode() { return this->debuggerMode; }
         bool IsDebugContextInNonDebugMode() const { return this->debuggerMode == DebuggerMode::NotDebugging; }
-        bool IsDebugContextInDebugMode() const { return this->debuggerMode == DebuggerMode::Debugging; }
+        bool IsDebugContextInDebugMode() const { return debuggerMode == DebuggerMode::Debugging; }
         bool IsDebugContextInSourceRundownMode() const { return this->debuggerMode == DebuggerMode::SourceRundown; }
         bool IsDebugContextInSourceRundownOrDebugMode() const { return IsDebugContextInSourceRundownMode() || IsDebugContextInDebugMode(); }
+        bool IsDebugContextInDebugRecordingMode() const { return this->IsDebugContextInDebugMode() && this->IsDebuggerRecording(); }
+        bool IsDebugContextInDebugRecordingSuspendedMode() const { return this->IsDebugContextInDebugMode() && !this->IsDebuggerRecording(); }
```
agarwal-sandeep: I don't think there is any need for these 2 functions. 
IsDebuggerRecording should have an assert that we are debugging and done.

lib/Runtime/Base/ScriptContext.cpp
```diff
@@ -3998,7 +3998,7 @@ namespace Js
...
             {
                 scriptContext->GetDebugContext()->GetProbeContainer()->StartRecordingCall();
```
agarwal-sandeep: >StartRecordingCall [](start = 71, length = 18)

In ProbeContainer::StartRecordingCall and ProbeContainer::EndRecordingCall can you Assert(this->pScriptContext->GetDebugContext()->IsDebuggerRecording())

lib/Runtime/Debug/DebugContext.h
dilijev: Good point. I'll refactor that.

lib/Runtime/Library/IntlEngineInterfaceExtensionObject.cpp
dilijev: Moved the class to ScriptContext.h which has all the types needed. Renamed it to AutoInitLibraryCodeScope.

---
In reply to: [129987339](https://github.com/Microsoft/ChakraCore/pull/3447#discussion_r129987339) [](ancestors = 129987339)

lib/Runtime/Base/ScriptContext.cpp
dilijev: Still having problems with type visibility. DebugContext is not visible outside of ScriptContext.cpp where I would need it in order to call the methods. The method is not necessary here, but it is necessary to call outside of ScriptContext. I'll keep trying to refactor to reduce the number of methods needed but I think a couple will be needed.

---
In reply to: [129987071](https://github.com/Microsoft/ChakraCore/pull/3447#discussion_r129987071) [](ancestors = 129987071)

dilijev: Changed to IsDebuggerRecording. Need a method because of type visibility on DebugContext.

---
In reply to: [129986937](https://github.com/Microsoft/ChakraCore/pull/3447#discussion_r129986937) [](ancestors = 129986937)

dilijev:
Thanks for the thorough review, @agarwal-sandeep!

For convenience, here are the remaining issues:

## To Discuss

* Merging the approach taken by @kfarnung in https://github.com/Microsoft/ChakraCore/commit/1c2d2591d4a73415d71d468264dd8e5676862255
* Turning the above two booleans into bitfield to save space
  * Can we just use `bool ... : 1` and let the compiler decide, or would it prefer speed over size?
  * If we're choosing size over speed, is it really worth it for saving a byte in memory layout for a sometimes-initialized pointer on `ScriptContext`?

## TODO

* In `ProbeContainer::StartRecordingCall and ProbeContainer::EndRecordingCall` can you `Assert(this->pScriptContext->GetDebugContext()->IsDebuggerRecording())`

## Check

* Double-check (post-refactor) that type visibility is still a concern motivating the additional methods on `ScriptContext`.

kfarnung:
@dilijev I targeted 1.7 because the fix ultimately benefits node-chakracore.  There were no consumers of 1.6 so I didn't see a need for it.  I don't see anything that would block back-porting if that's what we want to do.  I agree that we could combine the booleans into a single flags, it was discussed with @agarwal-sandeep when I made the change, but wasn't necessary at the time.

agarwal-sandeep:
@kfarnung@microsoft.com if we need to target 1.6 to merge into rs3 then lets port this to 1.6

---
In reply to: [318552127](https://github.com/Microsoft/ChakraCore/pull/3447#issuecomment-318552127) [](ancestors = 318552127)

lib/Runtime/Debug/DebugContext.h
```diff
@@ -51,11 +51,22 @@ namespace Js
         void SetHostDebugContext(HostDebugContext * hostDebugContext);
 
         void SetDebuggerMode(DebuggerMode mode);
+
         bool IsDebugContextInNonDebugMode() const { return this->debuggerMode == DebuggerMode::NotDebugging; }
         bool IsDebugContextInDebugMode() const { return this->debuggerMode == DebuggerMode::Debugging; }
         bool IsDebugContextInSourceRundownMode() const { return this->debuggerMode == DebuggerMode::SourceRundown; }
         bool IsDebugContextInSourceRundownOrDebugMode() const { return IsDebugContextInSourceRundownMode() || IsDebugContextInDebugMode(); }
 
+        bool IsDebuggerRecording() const { return this->IsDebugContextInDebugMode() && this->isDebuggerRecording; }
```
agarwal-sandeep: >this->IsDebugContextInDebugMode() [](start = 50, length = 33)

This should instead be an assert

lib/Runtime/Debug/DebugContext.h
```diff
@@ -51,11 +51,22 @@ namespace Js
...
+        void SetIsDebuggerRecording(bool isDebuggerRecording)
+        {
+            if (this->IsDebugContextInDebugMode())
```
agarwal-sandeep: >if (this->IsDebugContextInDebugMode()) [](start = 12, length = 38)

Same here

dilijev:
@kfarnung @agarwal-sandeep Updated comment above with resolution of offline discussion. If there are further concerns about whether to backport @kfarnung's change, let's speak further offline.

agarwal-sandeep:
LGTM Resolve the last 2 comments and commit

lib/Runtime/Debug/DebugContext.h
dilijev: If we are not in `Debugging` mode, such an assert would fail when we try to get the current debugging state in 
`AutoInitLibraryCodeScope` https://github.com/Microsoft/ChakraCore/pull/3447/files#diff-6858ff2e7f3b97763dd5be71dfc9ae56R1901

At current, not being in Debug mode will result in this returning false, which we then use to reset the recording state in `~AutoInitLibraryCodeScope`.

Considering that `SetIsDebuggerRecording` has no effect if we're not in `Debugging` mode, and that the Assert described above would fail, we could remove this additional condition, because it would have no effect on the state after `~AutoInitLibraryCodeScope` runs.

lib/Runtime/Debug/DebugContext.h
dilijev: Such an Assert would fail when `~AutoInitLibraryCodeScope` is run when not in `Debugging` mode. My thinking was that this check is necessary to protect against changing the state of `isDebuggerRecording` when not in `Debugging` mode. However, if we're saving and restoring the state anyway that might be unnecessary.

lib/Runtime/Debug/DebugContext.h
dilijev: These can't be asserts, but we can remove them to simplify the logic. Besides `AutoInitLibraryCodeScope` save/restore of the `isDebuggerRecording` flag, we only use IsDebuggerRecording in a code path where we already know `IsDebugContextInDebugMode()==true`.



---
In reply to: [130155025](https://github.com/Microsoft/ChakraCore/pull/3447#discussion_r130155025) [](ancestors = 130155025)

lib/Runtime/Debug/DebugContext.h
dilijev: See other comment.

---
In reply to: [130155689](https://github.com/Microsoft/ChakraCore/pull/3447#discussion_r130155689) [](ancestors = 130155689)

PR #3449, opened 2017-07-28
Disable doCollect for lookup and set/get API

obastemur:
Improves has/get/set etc. Jsrt API calls ~15% on xplat. It should also boost Windows
numbers too (will check). This change affects only ChakraCore

Why?

Assumption; Js land calls to a user defined function. This function reaches multiple has/get/set calls before finishing its' task. Setting `doCleanup = true` on each and every API calls effects the performance badly.

Besides, we were updating `tickCountNextDispose` value regardless from doCleanup is true or not. (`SetIsScriptActive`)

### DIFF
--- a/lib/Common/Memory/Recycler.cpp
+++ b/lib/Common/Memory/Recycler.cpp
@@ -947,12 +947,12 @@ Recycler::IsSweeping() const
 }
 
 void
-Recycler::SetIsScriptActive(bool isScriptActive)
+Recycler::SetIsScriptActive(bool isScriptActive, bool doCleanup)
 {
     Assert(this->isInScript);
     Assert(this->isScriptActive != isScriptActive);
     this->isScriptActive = isScriptActive;
-    if (isScriptActive)
+    if (isScriptActive && doCleanup)
     {
         this->tickCountNextDispose = ::GetTickCount() + RecyclerHeuristic::TickCountFinishCollection;
     }
--- a/lib/Jsrt/Jsrt.cpp
+++ b/lib/Jsrt/Jsrt.cpp
@@ -957,7 +957,7 @@ CHAKRA_API JsBooleanToBool(_In_ JsValueRef value, _Out_ bool *boolValue)
 
 CHAKRA_API JsConvertValueToBoolean(_In_ JsValueRef value, _Out_ JsValueRef *result)
 {
-    return ContextAPIWrapper<JSRT_MAYBE_TRUE>([&] (Js::ScriptContext *scriptContext, TTDRecorder& _actionEntryPopper) -> JsErrorCode {
+    return ContextAPIWrapper<JSRT_MAYBE_TRUE, JSRT_MAYBE_TRUE>([&] (Js::ScriptContext *scriptContext, TTDRecorder& _actionEntryPopper) -> JsErrorCode {
         PERFORM_JSRT_TTD_RECORD_ACTION(scriptContext, RecordJsRTVarToBooleanConversion, (Js::Var)value);
 
         VALIDATE_INCOMING_REFERENCE(value, scriptContext);
@@ -1139,7 +1139,7 @@ CHAKRA_API JsNumberToInt(_In_ JsValueRef value, _Out_ int *asInt)
 
 CHAKRA_API JsConvertValueToNumber(_In_ JsValueRef value, _Out_ JsValueRef *result)
 {
-    return ContextAPIWrapper<JSRT_MAYBE_TRUE>([&] (Js::ScriptContext *scriptContext, TTDRecorder& _actionEntryPopper) -> JsErrorCode {
+    return ContextAPIWrapper<JSRT_MAYBE_TRUE, JSRT_MAYBE_TRUE>([&] (Js::ScriptContext *scriptContext, TTDRecorder& _actionEntryPopper) -> JsErrorCode {
         PERFORM_JSRT_TTD_RECORD_ACTION(scriptContext, RecordJsRTVarToNumberConversion, (Js::Var)value);
 
         VALIDATE_INCOMING_REFERENCE(value, scriptContext);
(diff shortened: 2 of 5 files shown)


lib/Jsrt/Jsrt.cpp
```diff
@@ -957,7 +957,7 @@ CHAKRA_API JsBooleanToBool(_In_ JsValueRef value, _Out_ bool *boolValue)
 
 CHAKRA_API JsConvertValueToBoolean(_In_ JsValueRef value, _Out_ JsValueRef *result)
 {
-    return ContextAPIWrapper<JSRT_MAYBE_TRUE>([&] (Js::ScriptContext *scriptContext, TTDRecorder& _actionEntryPopper) -> JsErrorCode {
+    return ContextAPIWrapper<JSRT_MAYBE_TRUE, JSRT_MAYBE_TRUE>([&] (Js::ScriptContext *scriptContext, TTDRecorder& _actionEntryPopper) -> JsErrorCode {
```
curtisman: Can we rename this to what it actually control?
Like JSRT_VERIFY_POLICY

curtisman: If you have identify that these API are so lightweight, might as well just not do clean up for these API anyway?

Or have an count of how many time it was called before it would clean up?

lib/Common/Memory/Recycler.cpp
```diff
@@ -947,12 +947,12 @@ Recycler::IsSweeping() const
...
 
 void
-Recycler::SetIsScriptActive(bool isScriptActive)
+Recycler::SetIsScriptActive(bool isScriptActive, bool doCleanup)
 {
     Assert(this->isInScript);
     Assert(this->isScriptActive != isScriptActive);
     this->isScriptActive = isScriptActive;
-    if (isScriptActive)
+    if (isScriptActive && doCleanup)
```
curtisman: Why do we want to only "delay" dispose if it doCleanup?

Would that make it more likely to have dispose disrupt the script?

curtisman:
>             if (!threadServiceWrapper || !threadServiceWrapper->ScheduleNextCollectOnExit())

We will not be scheduling IdleGC on these calls?  Seems risky

---
Refers to: lib/Runtime/Base/ThreadContext.cpp:1561 in 22f31e7. [](commit_id = 22f31e786f625f5f8621f043e6fc719e48e6e959, deletion_comment = False)

lib/Jsrt/Jsrt.cpp
obastemur: Not sure if this is the answer but it is actually controlled by `JSRT_VERIFY_RUNTIME_STATE` and `JSRT_MAYBE_TRUE` corresponds to `possibly true`

obastemur: > Assumption; Js land calls to a user defined function. This function reaches multiple has/get/set calls before finishing its' 

We know for sure, these are called inside a user defined function and that function will collect on exit. I made them `JSRT_MAYBE_TRUE` in order to keep ChakraFull intact during the process. If we are happy with the result, We might as well remove it? I will add a TODO note there!

lib/Common/Memory/Recycler.cpp
obastemur: Sure. If we don't push the timer, it is more likely the `tick > tickCountNextDispose`

obastemur:
> We will not be scheduling IdleGC on these calls? Seems risky

These methods are being called under a user defined function. We already schedule it at least once?

MSLaguana:
I applied these changes to node-chakracore, and I'm not seeing anything like a 15% gain on the acmeair benchmark, more like a 1% gain.

obastemur:
> I applied these changes to node-chakracore, and I'm not seeing anything like a 15% gain on the xplat

Perhaps you missread the description...

obastemur:
@MSLaguana see the copy-paste of description;

'Improves has/get/set etc. Jsrt API calls ~15% on xplat. '

lib/Jsrt/Jsrt.cpp
curtisman: I know... so don't map it to JSRT_MAYBE_TRUE.  It just means nothing at the call site.

lib/Common/Memory/Recycler.cpp
curtisman: That's kinda opposite to what we want?

lib/Jsrt/Jsrt.cpp
curtisman: What do you mean "user defined function"?

curtisman:
>> We will not be scheduling IdleGC on these calls? Seems risky

> These methods are being called under a user defined function. We already schedule it at least once?

But once it ran, it will not run again until you schedule it.  It is there to detect activity that may need clean up.

lib/Common/Memory/Recycler.cpp
obastemur: No. We just don't want to keep calling the same sub interface unnecessarily. If there is something to dispose etc. we better do it. We were keep updating tick threshold to a future value + all the sub calls.

curtisman:
> @MSLaguana see the copy-paste of description;'Improves has/get/set etc. Jsrt API calls ~15% on xplat. '

I think @MSLaguana is pointing out that overall, it doesn't really benefit AcmeAir much.   
The concern here is that this case can cause memory usage regression in cases, so we have to wonder whether the change is worth it.

lib/Common/Memory/Recycler.cpp
curtisman: Isn't this just a GetTickCount() call?  what sub calls?

lib/Jsrt/Jsrt.cpp
obastemur: How about once we are inside the 'native function'? Shouldn't we doCleanup while we are exiting from that native call? Or doCleanup on specifically marked ones? ie. CallFunction, CreateThisORThat...

lib/Common/Memory/Recycler.cpp
obastemur: Assuming we pushed the timer beyond tickCount. Unfortunately GetTickCount is more than free too :/

obastemur:
@curtisman I will double check the memory regression (if there is) but I don't agree on gains part. Assuming no memory regression, double digit improvement to API calls is not trivial.

curtisman:
>Assuming no memory regression, double digit improvement to API calls is not trivial.

That's is microbenchmarking. In the real world, it depends on how often are these API are called.

Put it in the context of the over all scenario, if it only translate to gain in the noise range, then it is not very useful.

obastemur:
Well, if the whole engine perf assumption translates to AcmeAir, cool closing this one.

PR #3451, opened 2017-07-28
Ensure that LdHeapArguments is the first thing in the bytecode after recording constants

rajatd:
The Jit stack args optimization assumes that there will be no bailout before a LdHeapArguments (or equivalent) opcode. In this case, we had some class constructor related opcodes emitted before LdHeapArguments and when stressing bailouts, we were crashing. Note that the class constructor related opcodes will not get bailouts associated with them without any stress flag. But, its still good to maintain the jit assumption


suwc:
:shipit:

akroshg:
:shipit:

PR #3452, opened 2017-07-28
Splice helper function should check for side-effect in the prototype.

akroshg:
Splice has fast path which did not check the prototype (say you have proxy) properly which resulted to incorrect result. Fixed that by using the side-effect macro to take the slower and observable code path.

### DIFF
--- a/test/Array/protoLookupWithGetters.baseline
+++ b/test/Array/protoLookupWithGetters.baseline
@@ -1,35 +1,35 @@
 Test case 1
-d1,d2,d3,d4,p5,p6,p7,p8,p9
+p0,p1,p2,p3,p4,p5,p6,p7,p8
 9
 p0,p1,p2,p3,p4
 5
 
 Test case 2
-d1,d2,d3,d4,d5,d6,d7,p5,p6,p7,p8,p9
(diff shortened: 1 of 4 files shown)


akroshg:
@rajatd please have a look at this one.

test/Array/protoLookupWithGetters.baseline
```diff
@@ -1,35 +1,35 @@
 Test case 1
-d1,d2,d3,d4,p5,p6,p7,p8,p9
+p0,p1,p2,p3,p4,p5,p6,p7,p8
```
akroshg: earlier result was incorrect.

test/Array/protoLookupWithGetters.js
```diff
@@ -12,7 +12,7 @@ write("Test case 1");
 
 for(var i =0;i<10;i++)
 {
-  Object.defineProperty(Array.prototype, i, { get: function (i) { return function () { return "p"+i; } }(i), configurable: true, enumerable: true });
+  Object.defineProperty(Array.prototype, i, { get: function (i) { return function () { return "p"+i; } }(i), set : function(a) {}, configurable: true, enumerable: true });
```
akroshg: made this change as you need to have setter.

rajatd:
LGTM

PR #3454, opened 2017-07-31
Performance Improvements.

obastemur:
Total AcmeAir (LTO) perf. improvement 3%

Changes:

- Changes to JsHasProperty: This is a very hot method and ContextAPIWrapper is quite expensive. This change improves AcmeAir, 1%

- Changes to PropertyRecord creation: Improves the performance of PropertyRecord by combining tasks. Improves AcmeAir 2%. Downside, `! (!isSymbol && length <= 10 && length > 0)` might turn out to be false positive and we allocate extra sizeof(uint32) space.

- Centralize hashing logic

### DIFF
--- a/lib/Jsrt/Jsrt.cpp
+++ b/lib/Jsrt/Jsrt.cpp
@@ -1469,7 +1469,10 @@ CHAKRA_API JsSetProperty(_In_ JsValueRef object, _In_ JsPropertyIdRef propertyId
 
 CHAKRA_API JsHasProperty(_In_ JsValueRef object, _In_ JsPropertyIdRef propertyId, _Out_ bool *hasProperty)
 {
-    return ContextAPIWrapper<JSRT_MAYBE_TRUE>([&] (Js::ScriptContext *scriptContext, TTDRecorder& _actionEntryPopper) -> JsErrorCode {
+    VALIDATE_JSREF(object);
+    if (!Js::JavascriptOperators::IsObject(object)) return JsErrorArgumentNotObject;
+
+    auto internalHasProperty = [&] (Js::ScriptContext *scriptContext, TTDRecorder& _actionEntryPopper) -> JsErrorCode {
         PERFORM_JSRT_TTD_RECORD_ACTION(scriptContext, RecordJsRTHasProperty, (Js::PropertyRecord *)propertyId, object);
 
         VALIDATE_INCOMING_OBJECT(object, scriptContext);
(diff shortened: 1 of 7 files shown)


lib/Jsrt/Jsrt.cpp
```diff
@@ -1480,7 +1483,16 @@ CHAKRA_API JsHasProperty(_In_ JsValueRef object, _In_ JsPropertyIdRef propertyId
...
-    });
+    };
+
+    if (Js::JavascriptProxy::Is(object))
+    {
+        return ContextAPIWrapper<JSRT_MAYBE_TRUE>(internalHasProperty);
+    }
+    else
+    {
+        return ContextAPINoScriptWrapper(internalHasProperty);
```
kunalspathak: >return ContextAPINoScriptWrapper(internalHasProperty); [](start = 8, length = 54)

you can still call into user script.

```javascript
var obj = {};
var handler = {};
handler.getPrototypeOf = function(target) {
 print('inside getPrototypeOf');
 return Reflect.getPrototypeOf(target);
}
var proxy = new Proxy(obj, handler);

// although foo is not a proxy, the prototype object is a proxy
var foo = Object.create(proxy);

for(var x in foo) {
 print(x);
}
```

lib/Runtime/Base/ThreadContext.cpp
```diff
@@ -1035,37 +1035,18 @@ ThreadContext::UncheckedAddPropertyId(JsUtil::CharacterBuffer<WCHAR> const& prop
...
-
-    uint32 indexVal = 0;
-
-    // Symbol properties cannot be numeric since their description is not to be used!
-    bool isNumeric = !isSymbol && Js::PropertyRecord::IsPropertyNameNumeric(propertyName.GetBuffer(), propertyName.GetLength(), &indexVal);
-
-    uint hash = JsUtil::CharacterBuffer<WCHAR>::StaticGetHashCode(propertyName.GetBuffer(), propertyName.GetLength());
-
-    size_t allocLength = bytelength + sizeof(char16) + (isNumeric ? sizeof(uint32) : 0);
+    size_t allocLength = bytelength + sizeof(char16) + ( (!isSymbol && length <= 10 && length > 0) ? sizeof(uint32) : 0);
```
kunalspathak: >!isSymbol && length <= 10 && length > 0) ? sizeof(uint32)  [](start = 58, length = 58)

isn't that a lot? Basically with this, for every property name with length between 1 and 10, we are allocating extra 4 bytes?

obastemur: sizeof(PropertyRecord) is 24 bytes. In this case, we actually allocate extra only when string length is 3. Still a lot?

kunalspathak: Right, I did that calculation and seems like you will allocate extra 16 bytes in these three cases. Can you do little tracing and see how many times we create property record with these length?

kunalspathak: technically you can get the right size `sizeof(uint32)` vs. `sizeof(uint16)` depending on length and eliminate the difference.

---
In reply to: [130418087](https://github.com/Microsoft/ChakraCore/pull/3454#discussion_r130418087) [](ancestors = 130418087)

lib/Jsrt/Jsrt.cpp
obastemur: Added prototype checking. (saw no harm on performance) @kunalspathak thanks for this example and chat on getter/setter. Current implementation, couldn't trigger JS code while inside a ContextAPINoScriptWrapper. Additional concerns welcome!

lib/Runtime/Base/ThreadContext.cpp
obastemur: This depends to app. I couldn't see observable difference with acmeair

obastemur: > technically you can get the right size sizeof(uint32) vs. sizeof(uint16) depending on length and eliminate the difference.

well, this is much more tricky then how it sounds. no?

lib/Jsrt/Jsrt.cpp
kunalspathak: Tried looking around but I don't see if this can call into user script. So should be good to go. May be in other PR we can convert some hot JSRT APIs to similar mechanism.

lib/Runtime/Base/ThreadContext.cpp
kunalspathak: well, don't we already know that for length 2, 3 and 10 we will allocate extra 16 bytes (at least for x64). Why can't we add extra condition here to return `sizeof(uint32)` vs. sizeof(uint16)`?

---
In reply to: [130432408](https://github.com/Microsoft/ChakraCore/pull/3454#discussion_r130432408) [](ancestors = 130432408)

lib/Jsrt/Jsrt.cpp
kunalspathak: @akroshg , in case he can think of a scenario where this can become re-entrant.

kunalspathak:
:shipit:

obastemur:
@kunalspathak Thanks for the review!

PR #3455, opened 2017-07-31
JSON: Minor improvements

obastemur:
- fast path for constant propertyId
- CompleteInit, do not double copy

### DIFF
--- a/lib/Runtime/Base/ScriptContext.h
+++ b/lib/Runtime/Base/ScriptContext.h
@@ -945,7 +945,7 @@ private:
         }
 #endif
 
-        void SetHasUsedInlineCache(bool value) 
+        void SetHasUsedInlineCache(bool value)
         {
             hasUsedInlineCache = value;
 #if DBG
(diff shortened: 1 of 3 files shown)


obastemur:
@kunalspathak Thanks for the review!

lib/Runtime/Library/JSON.cpp
```diff
@@ -336,17 +332,28 @@ namespace JSON
...
-        wmemset(buffer, _u(' '), JSONspaceSize);
+        static const char16* spaceBuffer = GetGapStringInternal(); // multi threading safe
+
         charcount_t len = 0;
         switch (Js::JavascriptOperators::GetTypeId(space))
         {
         case Js::TypeIds_Integer:
             {
                 len = max(0, min(JSONspaceSize, static_cast<int>(Js::TaggedInt::ToInt32(space))));
+                if (len)
```
akroshg: nit: can you use len> 0 ?

lib/Runtime/Base/ScriptContext.h
```diff
@@ -1424,7 +1424,20 @@ namespace Js
...
         }
 
+    private:
+        Js::PropertyId emptyStringPropertyId;
     public:
+        Js::PropertyId GetEmptyStringPropertyId()
+        {
+            if (emptyStringPropertyId == Js::PropertyIds::_none)
+            {
+                Js::PropertyRecord const * propertyRecord;
```
obastemur: why?

lib/Runtime/Library/JSON.cpp
obastemur: `max(0, ....)`

lib/Runtime/Base/ScriptContext.h
akroshg: kind of good practice in general. what if for Example GetOrAddPropertyRecord not populate the out variable in all path.

---
In reply to: [131467416](https://github.com/Microsoft/ChakraCore/pull/3455#discussion_r131467416) [](ancestors = 131467416)

lib/Runtime/Library/JSON.cpp
akroshg: my point is you are using int as bool expression in the if condition.

---
In reply to: [131467627](https://github.com/Microsoft/ChakraCore/pull/3455#discussion_r131467627) [](ancestors = 131467627)

PR #3456, opened 2017-07-31
xplat: fix multi-thread builds for node-chakracore

obastemur:
Previously node-chakracore build process was failing if -j was used for multi-threaded builds. As a result, build was slow or failing.


obastemur:
rebasing PR to reset the CI (stuck)

obastemur:
@digitalinfinity @kunalspathak Thanks for the review

PR #3457, opened 2017-07-31
Cleanup: Remove unused field in JavascriptLibrary and avoid casting in functionObjectTypesList

curtisman:
FuncitonObjectTypesList in FunctionBody always store RecyclerWeakReference<ScriptFunctionType> instead of RecyclerWeakReference<DynamicType>
Upgrade the type to avoid casting.

The JavascriptLibrary::dyanmicFunctionReference is already referenced by the JavascriptLibrary, so no need to bind reference.

Also typedArrayType field is not used in JavascriptLibaray. Remove it.


atulkatti:
:shipit:

PR #3458, opened 2017-07-31
When ForceSplitScope flag is provided don't assume that we have non-simple parameter list

aneeshdk:
When we force all functions to have split scope we shouldn't assume that those function have non-simple parameter list. Fixed couple of places in EmitOneFunction.


aneeshdk:

akroshg:
LGTM

PR #3459, opened 2017-07-31
fix crashes with cross site typed virtual arrays

sigatrev:
If an object has type CrossSiteObject\<T> then VirtualTableInfo\<T>::HasVirtualTable() will always return false. In the case of CrossSiteObject virtual typed arrays, the vtable check for the virtual array type was returning false, so they were assumed to be non-virtual arrays which is incorrect.

Since no action was performed for the CrossSiteObject arrays anyway, this commit adds specific checks for the non-cross site, non-virtual array types, and removes the redundant CrossSite checks.


rajatd:
:shipit:

PR #3460, opened 2017-07-31
Fix order of generating bailout info in ValueNumberLdElem

meg-gupta:
When there is an early return in try finally, we have to execute the finally block before returning
So we add edges from early return block to finally block and finally block back to early return block.

If an instruction assigning to s0 is type specialized and s0 already exists in bytecodeUpwardExposedUsed
While populating bailout info, we will end up using the type specialized symbol instead of s0 even for pre opt bailout.
Because we don't differentiate between pre-op bailout and post-op bailout while populating bailout info
The type specialized symbol gets added on bytecodeUpwardExposed used, due to this gets added to upwardExposedUsed
If this instruction was in a loop, gets added to liveOnBackEdge syms.
This causes asserts in register allocator, when it extends the lifetime of the type specialized symbol to the entire loop.

This is a problem only with try-finallys. For non-EH functions, the early return block gets moved out of the loop due to break block removal.
With try finallys, these blocks are not moved out due to the early return -> finally and finally -> early return edges.

This happens only with return s0 symbol, because we avoid creating new temporary destination register for the s0 case.

This happens only in ValueNumberLdElem because it first calls TypeSpecializeIntDst and TypeSpecializeFloatDst and then calls GenerateBailAtOperation for the pre-op bailout.

### DIFF
--- a/lib/Backend/GlobOpt.cpp
+++ b/lib/Backend/GlobOpt.cpp
@@ -5346,6 +5346,39 @@ GlobOpt::ValueNumberLdElemDst(IR::Instr **pInstr, Value *srcVal)
             }
             return dstVal;
         }
+
+        if (!this->IsLoopPrePass())
+        {
+            if (instr->HasBailOutInfo())
+            {
+                const IR::BailOutKind oldBailOutKind = instr->GetBailOutKind();
+                Assert(
+                    (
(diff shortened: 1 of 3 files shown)


rajatd:
:shipit:

PR #3462, opened 2017-07-31
Add JsGetDataViewInfo support

kfarnung:
Match the existing JsGetTypedArrayInfo signature in order to support
making similar queries for DataViews.

### DIFF
--- a/lib/Jsrt/ChakraCore.h
+++ b/lib/Jsrt/ChakraCore.h
@@ -733,5 +733,22 @@ JsCopyStringOneByte(
     _Out_opt_ char* buffer,
     _Out_opt_ size_t* written);
 
+/// <summary>
+///     Obtains frequently used properties of a data view.
+/// </summary>
+/// <param name="dataView">The data view instance.</param>
+/// <param name="arrayBuffer">The ArrayBuffer backstore of the view.</param>
+/// <param name="byteOffset">The offset in bytes from the start of arrayBuffer referenced by the array.</param>
+/// <param name="byteLength">The number of bytes in the array.</param>
+/// <returns>
+///     The code <c>JsNoError</c> if the operation succeeded, a failure code otherwise.
+/// </returns>
+CHAKRA_API
+    JsGetDataViewInfo(
+        _In_ JsValueRef dataView,
+        _Out_opt_ JsValueRef *arrayBuffer,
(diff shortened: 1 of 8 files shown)


lib/Runtime/Debug/TTEventLog.cpp
```diff
@@ -2343,6 +2344,15 @@ namespace TTD
         giAction->Result = TTD_CONVERT_JSVAR_TO_TTDVAR(result);
     }
 
+    void EventLog::RecordJsRTGetDataViewInfo(Js::Var var, Js::Var result)
+    {
+        NSLogEvents::JsRTSingleVarArgumentAction* giAction = this->RecordGetInitializedEvent_DataOnly<NSLogEvents::JsRTSingleVarArgumentAction, NSLogEvents::EventKind::GetDataViewInfoActionTag>();
+        NSLogEvents::SetVarItem_0(giAction, TTD_CONVERT_JSVAR_TO_TTDVAR(var));
+
+        //entry/exit status should be set to clead by initialization so don't need to do anything
```
MSLaguana: Typo in comment? I assume you mean `clear`?

kfarnung: Yeah, that was actually a copy/paste so I'll fix the other one as well.

mrkmarron:
lgtm

lib/Jsrt/ChakraCore.h
```diff
@@ -733,5 +733,22 @@ JsCopyStringOneByte(
...
+/// </summary>
+/// <param name="dataView">The data view instance.</param>
+/// <param name="arrayBuffer">The ArrayBuffer backstore of the view.</param>
+/// <param name="byteOffset">The offset in bytes from the start of arrayBuffer referenced by the array.</param>
+/// <param name="byteLength">The number of bytes in the array.</param>
+/// <returns>
+///     The code <c>JsNoError</c> if the operation succeeded, a failure code otherwise.
+/// </returns>
+CHAKRA_API
+    JsGetDataViewInfo(
```
kfarnung: Yeah, I thought the same thing at first, but that method is purely interested in the raw storage.  This method returns the parameters which were passed to the create method and I couldn't see a good way to get them (other than using script).  Since TypedArray and DataView are both views over an ArrayBuffer, I think it makes sense for them to have similar interaction models.

PR #3463, opened 2017-08-01
Bug 11712101: Hoist CheckObjType out of loops only when the operand's containing object type is also invariant

ricobbe:
Check object type of argument to CheckObjType before hoisting it out of a loop, to avoid hoisting it over a DeleteFld that invalidates the type.

### DIFF
--- a/lib/Backend/GlobOpt.cpp
+++ b/lib/Backend/GlobOpt.cpp
@@ -16197,7 +16197,7 @@ GlobOpt::OptIsInvariant(
         }
         break;
     case Js::OpCode::CheckObjType:
-        // Bug 11712101: If the operand is a field, ensure that its containing object type is invariant 
+        // Bug 11712101: If the operand is a field, ensure that its containing object type is invariant
         // before hoisting -- that is, don't hoist a CheckObjType over a DeleteFld on that object.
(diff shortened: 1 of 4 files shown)


lib/Backend/GlobOpt.cpp
```diff
@@ -16196,6 +16196,22 @@ GlobOpt::OptIsInvariant(
...
         }
         break;
+    case Js::OpCode::CheckObjType:
+        // Bug 11712101: If the operand is a field, ensure that its containing object type is invariant 
+        // before hoisting -- that is, don't hoist a CheckObjType over a DeleteFld on that object.
+        // (CheckObjType only checks the operand and its immediate parent, so we don't need to go
+        // any farther up the object graph.)
+        Assert(instr->GetSrc1());
+        PropertySym *propertySym = instr->GetSrc1()->AsPropertySymOpnd()->GetPropertySym();
+        Assert(propertySym != nullptr);  // QUESTION FOR REVIEWER: Can this ever be false?
```
LouisLaf: >Assert(propertySym != nullptr);  / [](start = 8, length = 34)

It cannot.  The As getter above could potentially fail, but I don't believe so.  You can remove this assert.

LouisLaf:
:shipit:

PR #3464, opened 2017-08-01
Handle strict mode in console scope

aneeshdk:
For store operations we were not checking for strict mode in console scope. Added a new opcode to handle strict mode in console scope.


aneeshdk:
@agarwal-sandeep @akroshg Can you please take a look?

agarwal-sandeep:
:shipit:

PR #3465, opened 2017-08-01
JSONStack: Improve performance

obastemur:
1% Acme Air LTO gain. Majority of the time, there is at max 1 item on the stacks and SList too expensive to maintain 1item in/out during the whole JSON.stringify

### DIFF
--- a/lib/Runtime/Library/JSONStack.cpp
+++ b/lib/Runtime/Library/JSONStack.cpp
@@ -6,13 +6,26 @@
 #include "JSONStack.h"
 namespace JSON
 {
+#if defined(_JS_VALUE) || (_DOM_VALUE)
+#error "Something went wrong!"
+#endif
+
+#define _JS_VALUE tempValues[0]
+#define _DOM_VALUE   tempValues[1]
+
     bool StrictEqualsObjectComparer::Equals(Js::Var x, Js::Var y)
     {
         return JSONStack::Equals(x,y);
     }
 
-    JSONStack::JSONStack(ArenaAllocator *allocator, Js::ScriptContext *context) : jsObjectStack(allocator), domObjectStack(nullptr), alloc(allocator), scriptContext(context)
+    JSONStack::JSONStack(ArenaAllocator *allocator, Js::ScriptContext *context)
+      : jsObjectStack(allocator), domObjectStack(nullptr),
+        alloc(allocator), scriptContext(context)
     {
+        // most of the time, the size of the stack is 1 object.
+        // use direct member instead of expensive Push / Pop / Has
+        _JS_VALUE = nullptr;
+        _DOM_VALUE = nullptr;
     }
 
     bool JSONStack::Equals(Js::Var x, Js::Var y)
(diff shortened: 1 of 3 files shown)


lib/Runtime/Library/JSONStack.cpp
```diff
@@ -37,22 +54,68 @@ namespace JSON
     {
         if (bJsObject)
         {
-            return jsObjectStack.Push(data);
+            bool result = true;
```
digitalinfinity: Consider refactoring this into an inline function- something like `PushObject(data, jsObjectStack, &STACK_VALUE)/PushObject(data, domObjectStack, &DOM_VALUE)`

lib/Runtime/Library/JSONStack.h
```diff
@@ -15,6 +15,7 @@ namespace JSON
     class JSONStack
     {
     private:
+        Js::Var        tempValues[2];
```
digitalinfinity: Not really your change, but more a question- are these references to Vars supposed to be kept alive? I see that this class uses an arena allocator so curious who keeps it alive

lib/Runtime/Library/JSONStack.cpp
```diff
@@ -24,10 +33,18 @@ namespace JSON
     {
         if (bJsObject)
         {
+            if (STACK_VALUE)
+            {
+                return (data == STACK_VALUE);
```
kunalspathak: >return (data == STACK_VALUE); [](start = 16, length = 29)

just to guard future changes to this class, can you add assert at appropriate places to check when STACK_VALUE != nullptr, jsobjectStack is always empty and similarly for dom stack?

lib/Runtime/Library/JSONStack.h
obastemur: @digitalinfinity not sure about the answer. I added a destructor to make sure they are nullified.

lib/Runtime/Library/JSONStack.cpp
obastemur: @kunalspathak  not sure I understood. there is an `if(STACK_VALUE)` why should I check STACK_VALUE != nullptr inside an Assert?

lib/Runtime/Library/JSONStack.cpp
obastemur: @digitalinfinity This also came to my mind. Good idea. However, all three functions may end up in similar inline function pairs. Current looked more cleaner (whole file is ~100 lines)

obastemur:

obastemur:
@digitalinfinity @kunalspathak Thanks for the review!

lib/Runtime/Library/JSONStack.h
obastemur: This is not a problem for JSONStack as it always cleans up the references once stringify is done with inner object/array etc.

lib/Runtime/Library/JSONStack.cpp
kunalspathak: I meant to say add assert something like
```c++
if(STACK_VALUE) 
{
  assert(jsObjectStack.Count() == 0);
   return (data == STACK_VALUE)
}
```

PR #3468, opened 2017-08-02
Strengthen WebAssemblyArrayBuffer.GrowMemory

Cellule:
Make reporting external memory allocation cleaner with a wrapper and ensure we don't end up in a bad state after ReAlloc

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/3468)

### DIFF
--- a/lib/Common/Memory/Recycler.h
+++ b/lib/Common/Memory/Recycler.h
@@ -1202,9 +1202,12 @@ public:
 
     template <CollectionFlags flags>
     bool FinishDisposeObjectsNow();
-    BOOL ReportExternalMemoryAllocation(size_t size);
+    bool RequestExternalMemoryAllocation(size_t size);
     void ReportExternalMemoryFailure(size_t size);
     void ReportExternalMemoryFree(size_t size);
+    // ExternalAllocFunc returns true when allocation succeeds
+    template <typename ExternalAllocFunc>
+    bool DoExternalAllocation(size_t size, ExternalAllocFunc externalAllocFunc);
 
 #ifdef TRACE_OBJECT_LIFETIME
 #define DEFINE_RECYCLER_ALLOC_TRACE(AllocFunc, AllocWithAttributesFunc, attributes) \
@@ -2525,3 +2528,35 @@ extern bool IsLikelyRuntimeFalseReference(
 #else
 #define DECLARE_RECYCLER_VERIFY_MARK_FRIEND()
 #endif
+
+template <typename ExternalAllocFunc>
+bool Recycler::DoExternalAllocation(size_t size, ExternalAllocFunc externalAllocFunc)
+{
+    // Request external memory allocation
+    if (!RequestExternalMemoryAllocation(size))
+    {
+        // Attempt to free some memory then try again
+        CollectNow<CollectOnTypedArrayAllocation>();
+        if (!RequestExternalMemoryAllocation(size))
+        {
+            return false;
+        }
+    }
+    struct AutoExternalAllocation
+    {
+        bool allocationSucceeded = false;
+        Recycler* recycler;
+        size_t size;
+        AutoExternalAllocation(Recycler* recycler, size_t size): recycler(recycler), size(size) {}
+        // In case the externalAllocFunc throws or fails, the destructor will report the failure
+        ~AutoExternalAllocation() { if (!allocationSucceeded) recycler->ReportExternalMemoryFailure(size); }
+    };
+    AutoExternalAllocation externalAllocation(this, size);
+    if (externalAllocFunc())
+    {
+        this->AddExternalMemoryUsage(size);
+        externalAllocation.allocationSucceeded = true;
+        return true;
+    }
+    return false;
+}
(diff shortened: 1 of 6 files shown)


lib/Common/Memory/Recycler.h
```diff
@@ -2525,3 +2528,35 @@ extern bool IsLikelyRuntimeFalseReference(
...
 #endif
+
+template <typename ExternalAllocFunc>
+bool Recycler::DoExternalAllocation(size_t size, ExternalAllocFunc externalAllocFunc)
+{
+    // Request external memory allocation
+    if (!RequestExternalMemoryAllocation(size))
+    {
+        // Attempt to free some memory then try again
+        CollectNow<CollectOnTypedArrayAllocation>();
```
Cellule: Right now `CollectOnTypedArrayAllocation` is the only case we would use this allocation method. Should I make it generic anyway ?

lib/Runtime/Library/ArrayBuffer.h
```diff
@@ -159,7 +159,8 @@ namespace Js
         virtual BOOL GetDiagTypeString(StringBuilder<ArenaAllocator>* stringBuilder, ScriptContext* requestContext) override;
         virtual BOOL GetDiagValueString(StringBuilder<ArenaAllocator>* stringBuilder, ScriptContext* requestContext) override;
 
-        virtual ArrayBufferDetachedStateBase* DetachAndGetState();
+        void Detach();
```
akroshg: make it private member function as I don't see it is called externally.

akroshg: or protected.

---
In reply to: [130999054](https://github.com/Microsoft/ChakraCore/pull/3468#discussion_r130999054) [](ancestors = 130999054)

lib/Runtime/Library/ArrayBuffer.cpp
```diff
@@ -1042,68 +1047,85 @@ namespace Js
         }
         uint32 growSize = newBufferLength - this->bufferLength;
 
-        bool failedReport = false;
-        const auto reportFailedFn = [&failedReport] { failedReport = true; };
+        struct AutoFailFastOnBadState
+        {
+            bool isStateValid = true;
+            ~AutoFailFastOnBadState()
```
akroshg: QueryContinue can throw javascript exception. In that case we fail fast?
Check AutoDisableInterrupt in threadcontext.h, which is suited for this need.

Cellule: I'm not sure I understand your point. What is the link between QueryContinue and WebAssembly.GrowMemory ?
After we've called ReAlloc, `this->buffer` is freed. Which means, any kind of exception before we detach the buffer will leave us in a bad state and there isn't much we can do to recover at that point. 
This should be a very rare scenario, most of the time in very intense memory usage and thus it makes sense to failfast.

Cellule: Ok, so you mean, in order to minimize the kind of things that can throw here, I should use `AutoDisableInterrupt` when it becomes a bad state so it won't failfast.
Do I get this right ?

Cellule:
@akroshg can you take a look on my review fixes ?

lib/Runtime/Library/ArrayBuffer.cpp
```diff
@@ -1042,68 +1047,89 @@ namespace Js
...
-                newArrayBuffer = GetLibrary()->CreateWebAssemblyArrayBuffer(this->buffer, 0);
+                newArrayBuffer = this->GetLibrary()->CreateWebAssemblyArrayBuffer(this->buffer, 0);
             }
         }
         else
         {
-            ReportDifferentialAllocation(newBufferLength, reportFailedFn);
-            if (failedReport)
+            // Disable QC while doing a ReAlloc to minimize chances to end up in a bad state
+            AutoDisableInterrupt autoDisableInterrupt(this->GetScriptContext()->GetThreadContext()->GetInterruptPoller(), true);
```
akroshg: >true [](start = 122, length = 4)

I think you need to pass false here - as you don't want explicit completion. True here means that you are going to call the autoDisableInterrupt.Completed() in this block.

Cellule: The `true` means that I want to disable Interrupt polling

```c++
AutoDisableInterrupt(InterruptPoller* interruptPoller, bool disable)
        : interruptPoller(interruptPoller)
    {
        if (interruptPoller != nullptr)
        {
            previousState = interruptPoller->IsDisabled();
            interruptPoller->SetDisabled(disable);
        }
    }
```

Cellule:
I have reworked `WebAssemblyArrayBuffer::GrowMemory` a lot, so I decided to just squash my changes + rebase on top of #3479.
@tcare Could you take a look at how I use `AutoDisableInterrupt` as well

lib/Runtime/Base/ThreadContext.h
```diff
@@ -1798,7 +1798,7 @@ class AutoDisableInterrupt
             AssertOrFailFast(false);
         }
     }
-
+    void RequireExplicitCompletion() { m_explicitCompletion = true; }
```
tcare: nit: SetRequireExplicitCompletion

tcare:
Interrupt use LGTM

akroshg:
:shipit:

PR #3469, opened 2017-08-02
WASM: reserved immediates

### LINKED ISSUE #3450: WASM: Reserved immediate
> The reserved immediate to the current_memory and grow_memory operators is for future 🦄 use and must be 0 in the MVP.
The call_indirect operator takes a list of function arguments and as the last operand the index into the table. Its reserved immediate is for future 🦄 use and must be 0 in the MVP.

Currently we just ignore the reserved immediate, we need to validate that they're 0

Cellule:
Make sure reserved value are 0 
Fixes #3450

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/3469)


rajatd:
:shipit:

PR #3472, opened 2017-08-03
JSON: allocate enough cache

obastemur:
Improves AcmeAir LTO ~1%

### DIFF
--- a/lib/Runtime/Library/JSON.cpp
+++ b/lib/Runtime/Library/JSON.cpp
@@ -641,10 +641,10 @@ namespace JSON
                         precisePropertyCount = propertyCount;
                     }
 
-                    result = Js::ConcatStringBuilder::New(this->scriptContext, propertyCount);    // Reserve initial slots for properties.
-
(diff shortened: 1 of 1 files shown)


obastemur:
previously, concatStringBuilder was being initialized with lower amount of storage hence it was ending up resizing.

lib/Runtime/Library/JSON.cpp
```diff
@@ -641,10 +641,9 @@ namespace JSON
                         precisePropertyCount = propertyCount;
                     }
 
-                    result = Js::ConcatStringBuilder::New(this->scriptContext, propertyCount);    // Reserve initial slots for properties.
-
                     if (ReplacerFunction != replacerType)
                     {
+                        result = Js::ConcatStringBuilder::New(this->scriptContext, propertyCount + 1);    // Reserve initial slots for properties.
```
digitalinfinity: Why +1? Can you update the comment to reflect your thinking here?

obastemur:

PR #3473, opened 2017-08-03
Fixing ninja builds as part of parent make

MSLaguana:
When trying to use ninja builds as part of node-chakracore,
the parent make process was passing in flags like -r to the child
'make' process, which was actually ninja. The -r flag is not
understood by ninja, so it bails out and fails the build.

If we are building using ninja, then we shouldn't pass on flags
from make.


MSLaguana:
@dotnet-bot test this please

MSLaguana:
@dotnet-bot test this please

PR #3479, opened 2017-08-03
Port AutoDisableInterrupt updates from internal branch

tcare:
I made use of Akrosh's auto interrupt disable class, and moved it to ThreadContext to replace the existing one which has less functionality.


akroshg:
:shipit:

PR #3480, opened 2017-08-03
fix minor out of bound read (JavascriptString::ToInteger)

jianchun:
Reorder checks to avoid minor out of bound read.

Fix OS#13058871


jianchun:
Thanks @digitalinfinity !

PR #3485, opened 2017-08-04
OS# 12999605: Fix deleting of DebugContext

agarwal-sandeep:
Don't delete debugContext when ScriptContext is closed as we might be
in middle of a PDM call. When ScriptContext is closed delete all memory
allocated by debugContext and mark it as closed. Delete debugContext in
ScriptContext destructor.

### DIFF
--- a/lib/Runtime/Debug/DebugContext.cpp
+++ b/lib/Runtime/Debug/DebugContext.cpp
@@ -10,6 +10,7 @@ namespace Js
         scriptContext(scriptContext),
         hostDebugContext(nullptr),
         diagProbesContainer(nullptr),
+        isClosed(false),
         debuggerMode(DebuggerMode::NotDebugging),
         isDebuggerRecording(true)
     {
@@ -18,7 +19,7 @@ namespace Js
 
     DebugContext::~DebugContext()
     {
-        Assert(this->scriptContext == nullptr);
+        Assert(this->scriptContext != nullptr);
         Assert(this->hostDebugContext == nullptr);
         Assert(this->diagProbesContainer == nullptr);
     }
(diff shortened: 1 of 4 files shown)


agarwal-sandeep:
@jianchun and @akroshg please take a look
@aneeshdk FYI

jianchun:
:shipit:

lib/Runtime/Debug/DebugContext.cpp
```diff
@@ -49,6 +60,11 @@ namespace Js
         }
     }
 
+    bool DebugContext::IsSelfOrScriptContextClosed() const
+    {
+        return (this->IsClosed() || this->scriptContext->IsClosed());
```
akroshg: do you need to check for this->scriptContext not null?

lib/Runtime/Debug/DebugContext.cpp
```diff
@@ -170,14 +189,10 @@ namespace Js
             return hr;
         }
 
-        // Cache ScriptContext as multiple calls below can go out of engine and ScriptContext can be closed which will delete DebugContext
-        Js::ScriptContext* cachedScriptContext = this->scriptContext;
-
         utf8SourceInfoList->MapUntil([&](int index, Js::Utf8SourceInfo * sourceInfo) -> bool
         {
-            if (cachedScriptContext->IsClosed())
+            if (this->IsSelfOrScriptContextClosed())
```
akroshg: the check was there on the cached script context, are we not using this anymore?

akroshg:
:shipit:

lib/Runtime/Debug/DebugContext.cpp
agarwal-sandeep: No, scriptContext on debugContext is never set to null (in constructor we have assert that initial value is not null)

lib/Runtime/Debug/DebugContext.cpp
agarwal-sandeep: No, that's the crux of this change. If during host call ScriptContext is closed we would have closed debug context but not deleted it so debugcontext will always be valid and in closed state.

PR #3486, opened 2017-08-04
WASM - fix assert in jsrt debugger

Cellule:
Support WebAssembly types in Jsrt debugger. 
Add more wasm debugger test and move them to DebuggerCommon

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/3486)


bin/ch/Helpers.cpp
```diff
@@ -252,74 +252,51 @@ LPCWSTR Helpers::JsErrorCodeToString(JsErrorCode jsErrorCode)
...
+    case JsErrorNotImplemented:                return _u("JsErrorNotImplemented");
+    case JsErrorWrongThread:                   return _u("JsErrorWrongThread");
+    case JsErrorRuntimeInUse:                  return _u("JsErrorRuntimeInUse");
+    case JsErrorBadSerializedScript:           return _u("JsErrorBadSerializedScript");
+    case JsErrorInDisabledState:               return _u("JsErrorInDisabledState");
+    case JsErrorCannotDisableExecution:        return _u("JsErrorCannotDisableExecution");
+    case JsErrorHeapEnumInProgress:            return _u("JsErrorHeapEnumInProgress");
+    case JsErrorArgumentNotObject:             return _u("JsErrorArgumentNotObject");
+    case JsErrorInProfileCallback:             return _u("JsErrorInProfileCallback");
+    case JsErrorInThreadServiceCallback:       return _u("JsErrorInThreadServiceCallback");
```
agarwal-sandeep: Since you are here can you add all missing error codes as well. I see JsCannotSetProjectionEnqueueCallback
JsInvalidModuleHostInfoKind
JsErrorScriptException missing (and more)

lib/Jsrt/JsrtDebugUtils.cpp
```diff
@@ -467,7 +473,12 @@ const char16 * JsrtDebugUtils::GetClassName(Js::TypeId typeId)
     case Js::TypeIds_Promise:           return _u("Promise");
     case Js::TypeIds_GlobalObject:      return _u("Object");
     case Js::TypeIds_SpreadArgument:    return _u("Spread");
-
+#ifdef ENABLE_WASM
+    case Js::TypeIds_WebAssemblyModule:  return _u("WebAssembly.Module");
```
agarwal-sandeep: >WebAssembly.Module [](start = 52, length = 18)

Does other browsers display same string?

test/DebuggerCommon/rlexe.xml
```diff
@@ -1370,4 +1370,25 @@
       <tags>exclude_dynapogo</tags>
     </default>
   </test>
+  <test>
+    <default>
+      <files>wasm_basic.js</files>
```
Cellule: I would like it to run with jshost at one point. I just don't have the time to figure out what needs to be done right now

bin/ch/Helpers.cpp
Cellule: Right, I rewrote it, but I used `JsError` as a find string. But I didn't realize there were other with a different prefix. Will fix

lib/Jsrt/JsrtDebugUtils.cpp
Cellule: I'm not entirely sure what the other browsers do. 
I looked at https://github.com/WebAssembly/design/blob/master/JS.md#webassemblymoduleprototype--tostringtag--property
To determine which string to use, but maybe it's not the right one.

bin/ch/Helpers.cpp
```diff
@@ -252,74 +252,53 @@ LPCWSTR Helpers::JsErrorCodeToString(JsErrorCode jsErrorCode)
...
-    case JsErrorScriptTerminated:
-        return _u("JsErrorScriptTerminated");
-        break;
-
-    case JsErrorFatal:
-        return _u("JsErrorFatal");
-        break;
-
+    case JsNoError:                            return _u("JsNoError");
+    // JsErrorCategoryUsage = 0x10000,
```
agarwal-sandeep: >JsErrorCategoryUsage [](start = 7, length = 20)

TTD code returns this error, good to have string of it.
In fact just add for all other as well

lib/Jsrt/JsrtDebugUtils.cpp
agarwal-sandeep: Can you confirm in other browser

---
In reply to: [131437540](https://github.com/Microsoft/ChakraCore/pull/3486#discussion_r131437540) [](ancestors = 131437540)

Cellule: I confirmed this is what v8 uses at least.

lib/Runtime/Library/WebAssemblyModule.cpp
```diff
@@ -37,12 +36,13 @@ WebAssemblyModule::WebAssemblyModule(Js::ScriptContext* scriptContext, const byt
     m_binaryBufferLength(binaryBufferLength),
     m_customSections(nullptr)
 {
+    m_alloc = HeapNew(ArenaAllocator, _u("WebAssemblyModule"), scriptContext->GetThreadContext()->GetPageAllocator(), Js::Throw::OutOfMemory);
```
MikeHolman: what is this change for? i think we will leak this if the other allocations throw

lib/Runtime/Library/WebAssemblyModule.cpp
```diff
@@ -23,7 +23,6 @@ WebAssemblyModule::WebAssemblyModule(Js::ScriptContext* scriptContext, const byt
     m_memoryMaxSize(0),
     m_tableInitSize(0),
     m_tableMaxSize(0),
-    m_alloc(_u("WebAssemblyModule"), scriptContext->GetThreadContext()->GetPageAllocator(), Js::Throw::OutOfMemory),
```
MikeHolman: should still initialize to nullptr, otherwise check in finalize doesn't work

lib/Runtime/Library/WebAssemblyModule.cpp
Cellule: This is because adding
```cpp
protected:  
    DEFINE_VTABLE_CTOR(WebAssemblyModule, DynamicObject); 
    DEFINE_MARSHAL_OBJECT_TO_SCRIPT_CONTEXT(WebAssemblyModule);
```

Creates a constructor with `VirtualTableInfoCtorEnum` as argument. This means, I need a default constructor for the ArenaAllocator. I tried initially to add a such a constructor to ArenaAllocator, but it proved to be more difficult than I expected. 
I then decided to switch to a allocate the ArenaAllocator on the heap and manage the memory using Finalize.

I could try/catch the other allocators and free the m_alloc memory if an exception occurs. But the only kind of exceptions we can encounter are OOM exception, I doubt it is worth protecting for this case.

Cellule:
I had to rebase because of a merge conflict and decided to squash my commit history because it was really messy.
Notable changes since the last reviews: 
- Removed `m_alloc->clear()` in `WebAssemblyModule::Finalize()`
- Moved the wasm debugger tests back to the wasm folder. It was a pain the deal with all the exclude tags and in general I think it's better to have wasm related tests in their own folders. Might make a wasm debugger specific folder if we write more tests

PR #3487, opened 2017-08-04
WASM - improve some error messages

Cellule:
Improved some error messages to help debugging.
Also, change the CompileErrors offset to be relative to the binary buffer and not the function. This helps a lot to pinpoint where the error is exactly in the source.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/3487)

### DIFF
--- a/lib/Runtime/Library/WebAssemblyInstance.cpp
+++ b/lib/Runtime/Library/WebAssemblyInstance.cpp
@@ -17,6 +17,10 @@ Var GetImportVariable(Wasm::WasmImport* wi, ScriptContext* ctx, Var ffi)
     uint32 modNameLen = wi->modNameLen;
     ctx->GetOrAddPropertyRecord(modName, modNameLen, &modPropertyRecord);
     Var modProp = JavascriptOperators::OP_GetProperty(ffi, modPropertyRecord->GetPropertyId(), ctx);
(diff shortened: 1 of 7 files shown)


lib/Runtime/Library/WebAssemblyInstance.cpp
```diff
@@ -368,13 +367,17 @@ void WebAssemblyInstance::LoadImports(
             {
                 if (!WebAssemblyTable::Is(prop))
                 {
-                    JavascriptError::ThrowWebAssemblyLinkError(ctx, WASMERR_NeedTableObject);
+                    throw Wasm::WasmCompilationException(_u("Invalid import '%s.%s'. Expected a WebAssembly.Table"), import->modName, import->importName);
```
agarwal-sandeep: >Invalid import '%s.%s'. Expected a WebAssembly.Table [](start = 61, length = 52)

Can these error strings be centralized at one place? Can't we use RT_ERROR_MSG macros here?

akroshg: How is the localization work for these strings?

---
In reply to: [131437934](https://github.com/Microsoft/ChakraCore/pull/3487#discussion_r131437934) [](ancestors = 131437934)

agarwal-sandeep:
LGTM

PR #3488, opened 2017-08-04
OS: 11478866 adding a probe_stack to one of GetPrototypeSpecial API.

akroshg:


lib/Runtime/Library/JavascriptProxy.cpp
```diff
@@ -1451,6 +1451,8 @@ namespace Js
 
     RecyclableObject* JavascriptProxy::GetPrototypeSpecial()
     {
+        PROBE_STACK(GetScriptContext(), Js::Constants::MinStackDefault);
```
agarwal-sandeep: Do we have a test?

akroshg: I have a test but that is something exhausting the stack. Pretty much revealing that probe_stack is missing. I don't see a value in it. What do you think?

agarwal-sandeep: I was interested a small snippet, if its huge ignore it

akroshg:

akroshg:

lib/Runtime/Library/JavascriptProxy.cpp
agarwal-sandeep: Have a question related to this. JavascriptProxy::DeleteProperty does have PROBE_STACK but why SetProperty doesn't? Similarly IsExtensible which seems to have same pattern of CallFunction?

akroshg: SetProperty calls GetOwnPropertyDescriptor, which eventually has it. but I don't have test case to back that up. I can add the check there as well. 
I didn't understand the IsExtensible question.

agarwal-sandeep: discussed offline, Akrosh will check all proxy apis and make sure we have stack probe where needed

PR #3495, opened 2017-08-08
OS#13129251: Math.min/max should return an integer value when all of its params are integers

thomasmo:
This change addresses a perf issue where non-inlined Math.min/max are always floating-point vars. In the bug, this causes expensive bailouts in a loop that was setting to a Uin8ClampedArray. The fix is to check whether all of the parameters are tagged integers, and, if so, return an int.
With a reduced repro of the scenario from the original page, there is a significant improvement, where the same function takes 20% of the time it did before. Normal usage of Math.max with 3 int parameters set to a var results in taking 65% of the time it did before. Normal usage of Math.max with a float parameter showed a 1-5% regression, depending on where the first non-int parameter is listed.

### DIFF
--- a/lib/Runtime/Language/TaggedInt.inl
+++ b/lib/Runtime/Language/TaggedInt.inl
@@ -206,7 +206,22 @@ namespace Js
     }
 #endif
 
+    // Returns true when the provided args only contains TaggedInts. Note that
+    // iteration starts from 1 (to account for 'this' at index 0).
+    bool inline TaggedInt::OnlyContainsTaggedInt(Js::Arguments& args)
+    {
+        bool onlyContainsInts = true;
+        for (uint idxArg = 1; idxArg < args.Info.Count; idxArg++)
+        {
+            if (!TaggedInt::Is(args[idxArg]))
+            {
+                onlyContainsInts = false;
+                break;
+            }
+        }
 
+        return onlyContainsInts;
+    }
 
     inline Var TaggedInt::Add(Var aLeft,Var aRight,ScriptContext* scriptContext)
 #ifdef DBG
--- a/lib/Runtime/Library/MathLibrary.cpp
+++ b/lib/Runtime/Library/MathLibrary.cpp
@@ -702,6 +702,7 @@ namespace Js
         ARGUMENTS(args, callInfo);
         AssertMsg(args.Info.Count > 0, "Should always have implicit 'this'");
         ScriptContext* scriptContext = function->GetScriptContext();
+        bool hasOnlyIntegerArgs = false;
 
         Assert(!(callInfo.Flags & CallFlags_New));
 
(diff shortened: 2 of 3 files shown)


lib/Runtime/Library/MathLibrary.cpp
```diff
@@ -714,35 +716,56 @@ namespace Js
...
         }
 
-        double current = JavascriptConversion::ToNumber(args[1], scriptContext);
-        if(JavascriptNumber::IsNan(current))
+        if (hasOnlyIntegerArgs)
         {
-            return scriptContext->GetLibrary()->GetNaN();
-        }
+            // **Assuming that TaggedInt cannot be NaN?
+            // Int32 vs Int64?
```
LouisLaf: Yes, tagged-ints are int32 in 64bits, and int31 in 32bits.  They can't be NaN.

LouisLaf:
:shipit:

lib/Runtime/Language/TaggedInt.inl
```diff
@@ -206,7 +206,23 @@ namespace Js
     }
 #endif
 
+    // Returns true when the provided args only contains TaggedInts. Note that
+    // iteration starts from 0 (to account for 'this' at index 0).
+    // ** Is this too large of a function to inline?
```
thomasmo: @LouisLaf , @pleath -- fyi, most of the comments in this first round are questions about assumptions I made that I need verified

thomasmo:
thanks @dilijev !

lib/Runtime/Language/TaggedInt.inl
dilijev: You can use the REVIEW: prefix to comments in the future to indicate this (by convention) :)

thomasmo:
@pleath is added to the review. #Closed

thomasmo:
@michol is added to the review. #Closed

lib/Runtime/Library/MathLibrary.cpp
```diff
@@ -714,35 +716,56 @@ namespace Js
...
+
+            for (uint idxArg = 2; idxArg < args.Info.Count; idxArg++)
             {
-                current = compare;
+                double compare = JavascriptConversion::ToNumber(args[idxArg], scriptContext);
+                if (JavascriptNumber::IsNan(compare))
+                {
+                    return scriptContext->GetLibrary()->GetNaN();
+                }
+                if ((JavascriptNumber::IsNegZero(current) && compare == 0) ||
```
obastemur: why we check `(JavascriptNumber::IsNegZero(current) && compare == 0)` on each loop? is it necessary?

LouisLaf: Unfortunately, -0.0 === 0.0.  I don't see how we can avoid the check...  It would probably be cheaper to invert the whole thing however:
if (current < compare || (compare == 0 && IsNegZero(current)))

---
In reply to: [133482430](https://github.com/Microsoft/ChakraCore/pull/3495#discussion_r133482430) [](ancestors = 133482430)

lib/Runtime/Library/MathLibrary.cpp
```diff
@@ -714,35 +716,56 @@ namespace Js
...
+        if (hasOnlyIntegerArgs)
         {
-            return scriptContext->GetLibrary()->GetNaN();
-        }
+            // **Assuming that TaggedInt cannot be NaN?
+            // Int32 vs Int64?
+            int32 current = TaggedInt::ToInt32(args[1]);
+            for (uint idxArg = 2; idxArg < args.Info.Count; idxArg++)
+            {
+                // ** do we need JavascriptNumber::IsNegZero check for TaggedInt?
```
LouisLaf: >criptNumber::IsNegZero check for TaggedInt [](start = 38, length = 42)

Nope, floats only

lib/Runtime/Library/MathLibrary.cpp
LouisLaf: And int64 is WASM only, which doesn't have these buijlt-ins.

---
In reply to: [131993866](https://github.com/Microsoft/ChakraCore/pull/3495#discussion_r131993866) [](ancestors = 131993866)

lib/Runtime/Language/TaggedInt.inl
LouisLaf: It shouldn't be too large, especially with PGO.

---
In reply to: [132061017](https://github.com/Microsoft/ChakraCore/pull/3495#discussion_r132061017) [](ancestors = 132061017)

thomasmo: thanks for confirming

lib/Runtime/Library/MathLibrary.cpp
thomasmo: good points:
1) comment added to make this clear
2) comparison order rearranged

PR #3496, opened 2017-08-08
remove invalid special case in regalloc

MikeHolman:
We had special case in regalloc if we are storing to a tmp reg and the next instr has the same dst, we don't need to insert a store. Unfortunately, this is invalid in case the sym is also a source of the next instr.

Benchmark perf is flat.

OS: 11480955


LouisLaf:
Did you measure x86 as well?

LouisLaf:
:shipit:

MikeHolman:
Yup, measured both x86 and x64.

PR #3498, opened 2017-08-08
Fix bug in which we incorrectly copy-propagated away a load from a stack slot after boxing locals onto the heap

ricobbe:
OS:9682944

`test/Closures/copy-prop-stack-slot-test-framework.js` is not interesting for the purposes of this review; I copied it straight out of the original bug report.  Unfortunately, I was unable to find a way to remove this file and still reproduce the bug.


LouisLaf:
:shipit:

lib/Backend/GlobOpt.cpp
```diff
@@ -5180,6 +5181,19 @@ GlobOpt::ValueNumberDst(IR::Instr **pInstr, Value *src1Val, Value *src2Val)
     case Js::OpCode::Typeof:
         return this->NewGenericValue(ValueType::String, dst);
         break;
```
rajatd: >break [](start = 8, length = 5)

lets remove this 'break'.

ricobbe: Done

PR #3499, opened 2017-08-09
Retain subclass valueinfo while optimizing BailOnNotObject in OptTagChecks

meg-gupta:
Fixes OS#13115740.

In OptTagCheks currently, the valueinfo is being replaced by a new valueinfo changing the valuetype turning off Bits::CanBeTaggedValue for a BailOnNotObject instruction.
The initial valueInfo can be an IntRange ranging from 0 - INT_MAX for something like an array length. If the array length symbol is the opnd of BailOnNotObject then, the valueInfo will change to INT_MIN to INT_MAX.
Avoid this by preserving the subclass info while changing the value type.
```
function test0() {
  var ui16 = new Uint16Array();
    for (var _strvar0 in ui16) {
      litObj1.prop0 = ui16.length; // Array Length int specialized with int range 0 - INT_MAX
      if (litObj1.prop0.prop0) { // Generates BailOnNotObject on array length and force sets the value info to INT_MIN - INT_MAX
        ui16.length; // We end up with a negative lower bound on array length, assertion is fired during typespec
      }
    }
}
```

### DIFF
--- a/lib/Backend/GlobOpt.cpp
+++ b/lib/Backend/GlobOpt.cpp
@@ -2820,7 +2820,9 @@ GlobOpt::OptTagChecks(IR::Instr *instr)
             {
                 if (valueType.CanBeTaggedValue())
                 {
-                    ChangeValueType(nullptr, value, valueType.SetCanBeTaggedValue(false), false);
+                    // We're not adding new information to the value other than changing the value type. Preserve any existing
+                    // information and just change the value type.
+                    ChangeValueType(nullptr, value, valueType.SetCanBeTaggedValue(false), true /*preserveSubClassInfo*/);
                     return false;
                 }
                 if (this->byteCodeUses)
--- a/test/Bugs/rlexe.xml
+++ b/test/Bugs/rlexe.xml
@@ -355,4 +355,9 @@
       <files>bug11026788.js</files>
     </default>
   </test>
+  <test>
+    <default>
+      <files>valueInfoLossBug.js</files>
+    </default>
+  </test>
 </regress-exe>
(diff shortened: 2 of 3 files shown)


lib/Backend/GlobOpt.cpp
```diff
@@ -2820,7 +2820,9 @@ GlobOpt::OptTagChecks(IR::Instr *instr)
             {
                 if (valueType.CanBeTaggedValue())
                 {
-                    ChangeValueType(nullptr, value, valueType.SetCanBeTaggedValue(false), false);
+                    // We're not adding new information to the value other than changing the value type. Preserve any existing
+                    // information and just change the value type.
+                    ChangeValueType(nullptr, value, valueType.SetCanBeTaggedValue(false), true);
```
rajatd: Add a test if you have one.

meg-gupta: Added test.

rajatd: >ChangeValueType [](start = 20, length = 15)

The fix by itself seems good. There are other usages of this method that may need fixing or at least understanding why they do not 'preserveSubClassInfo'. Also could you pass the last parameter as "true /*preserveSubClassInfo*/" ?

rajatd:
:shipit:

lib/Backend/GlobOpt.cpp
meg-gupta: Done. I'll try to understand the other usages, some atleast can be elevated to true from the first look.

meg-gupta:
Thanks for reviewing!

PR #3500, opened 2017-08-09
OS:10898061 fix bug with cached scopes and default arguments

sigatrev:
When cached scopes are used with default function arguments, any variable references in the default argument values would erroneously see the merged body and param scope

### DIFF
--- a/lib/Runtime/ByteCode/ByteCodeEmitter.cpp
+++ b/lib/Runtime/ByteCode/ByteCodeEmitter.cpp
@@ -3997,6 +3997,7 @@ void ByteCodeGenerator::StartEmitFunction(ParseNode *pnodeFnc)
                 !funcInfo->Escapes() &&
                 funcInfo->frameObjRegister != Js::Constants::NoRegister &&
                 !ApplyEnclosesArgs(pnodeFnc, this) &&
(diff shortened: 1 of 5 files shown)


lib/Runtime/ByteCode/ByteCodeEmitter.cpp
```diff
@@ -3997,6 +3997,7 @@ void ByteCodeGenerator::StartEmitFunction(ParseNode *pnodeFnc)
                 !funcInfo->Escapes() &&
                 funcInfo->frameObjRegister != Js::Constants::NoRegister &&
                 !ApplyEnclosesArgs(pnodeFnc, this) &&
+                !pnodeFnc->sxFnc.HasDefaultArguments() &&
```
rajatd: Is it ok to do cached scopes in other cases of non-simple params - rest, destructuring patterns?

pleath: Yes. This specifically has to do with lookups from within the default expression.

sigatrev:
opening a new PR based to 1.7, and using one of the flags freed up by #3535 for HasDestructuredParams()

PR #3501, opened 2017-08-09
Perf Improvements

obastemur:
AcmeAIR LTO gain ~3%

- jsrt-internals: use known length JsGetProperty
- numberToString: Use faster implementation
Previously, we had twice faster numberToString implementation is introduced. Use it on remaining places.
- jitFuncMaps: Use initial capacity and update size policy
In dictionary collision drops up to 10 times.
- Do not lookup for empty string record
Empty string propertyRecord shows up very often on benchmarks
- Update hashing algorithm to FNV-1a
Lower the amount of collision in dictionaries and make hash code data
type consistent.

### DIFF
--- a/lib/Common/DataStructures/Comparer.h
+++ b/lib/Common/DataStructures/Comparer.h
@@ -115,10 +115,18 @@ struct RecyclerPointerComparer
 };
 
 // FNV-1a hash -> https://en.wikipedia.org/wiki/Fowler%E2%80%93Noll%E2%80%93Vo_hash_function
-#define CC_HASH_OFFSET_VALUE 2166136261
+// #define CC_HASH_OFFSET_VALUE 2166136261
+// #define CC_HASH_LOGIC(hash, byte) \
+//    hash ^= byte;                  \
+//    hash *= 16777619
+
+// previous hash function.
+// TODO: hash function below is bad for key distribution.
+//       FNV-1a above results better but expensive for lookups in small data sets.
+#define CC_HASH_OFFSET_VALUE 0
 #define CC_HASH_LOGIC(hash, byte) \
-    hash ^= byte;                 \
-    hash *= 16777619
+    hash ^= _rotl(hash, 7);       \
+    hash ^= byte;
 
 template <>
 struct DefaultComparer<GUID>
--- a/lib/Common/DataStructures/Comparer.h
+++ b/lib/Common/DataStructures/Comparer.h
@@ -114,9 +114,11 @@ struct RecyclerPointerComparer
     }
 };
 
+// FNV-1a hash -> https://en.wikipedia.org/wiki/Fowler%E2%80%93Noll%E2%80%93Vo_hash_function
+#define CC_HASH_OFFSET_VALUE 2166136261
 #define CC_HASH_LOGIC(hash, byte) \
-    hash  = _rotl(hash, 7);        \
-    hash ^= byte;
+    hash ^= byte;                 \
+    hash *= 16777619
 
 template <>
 struct DefaultComparer<GUID>
@@ -129,7 +131,7 @@ struct DefaultComparer<GUID>
      inline static hash_t GetHashCode(GUID const& guid)
      {
         char* p = (char*)&guid;
-        int hash = 0;
+        hash_t hash = CC_HASH_OFFSET_VALUE;
         for (int i = 0; i < sizeof(GUID); i++)
         {
             CC_HASH_LOGIC(hash, (uint32)(p[i]));
@@ -148,7 +150,7 @@ struct StringComparer
 
     inline static hash_t GetHashCode(T str)
     {
-        int hash = 0;
+        hash_t hash = CC_HASH_OFFSET_VALUE;
         while (*str)
         {
             CC_HASH_LOGIC(hash, *str);
--- a/lib/Runtime/Base/ThreadContext.cpp
+++ b/lib/Runtime/Base/ThreadContext.cpp
@@ -211,6 +211,7 @@ ThreadContext::ThreadContext(AllocationPolicyManager * allocationPolicyManager, 
 #if ENABLE_JS_REENTRANCY_CHECK
     , noJsReentrancy(false)
 #endif
+    , emptyStringPropertyRecord(nullptr)
 {
     pendingProjectionContextCloseList = JsUtil::List<IProjectionContext*, ArenaAllocator>::New(GetThreadAlloc());
     hostScriptContextStack = Anew(GetThreadAlloc(), JsUtil::Stack<HostScriptContext*>, GetThreadAlloc());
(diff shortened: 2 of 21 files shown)


obastemur:
Additional notes: fnv-1a is up to 15% slower than our previous implementation. (strings with 3-15 chars) However, previous hashing results in more collisions and drops lookup times significantly

lib/Runtime/Base/ThreadContext.cpp
```diff
@@ -910,19 +911,19 @@ ThreadContext::IsNumericProperty(Js::PropertyId propertyId)
...
-    if (IsDirectPropertyName(propertyName, propertyNameLength))
-    {
-        propertyRecord = propertyNamesDirect[propertyName[0]];
-        Assert(propertyRecord == propertyMap->LookupWithKey(Js::HashedCharacterBuffer<char16>(propertyName, propertyNameLength)));
-    }
-    else
+    if (propertyNameLength < 2)
     {
-        propertyRecord = propertyMap->LookupWithKey(Js::HashedCharacterBuffer<char16>(propertyName, propertyNameLength));
+        if (propertyNameLength == 0) return this->GetEmptyStringPropertyRecord();
```
leirocks: >if (propertyNameLength == 0) return this->GetEmptyStringPropertyRecord(); [](start = 8, length = 73)

move this line out? and use {} around the return.

and maybe still keep propertyRecord for all code path, so in chk build it's easier for debugging, and in free build it will by optimized by the compiler anyway

lib/Runtime/Base/ThreadContext.cpp
```diff
@@ -1083,7 +1084,7 @@ ThreadContext::AddPropertyRecordInternal(const Js::PropertyRecord * propertyReco
 #if DBG
     // Only Assert we can't find the property if we are not adding a symbol.
     // For a symbol, the propertyName is not used and may collide with something in the map already.
-    if (!propertyRecord->IsSymbol())
+    if (propertyNameLength && !propertyRecord->IsSymbol())
```
leirocks: >propertyNameLength [](start = 8, length = 18)

please explicitly use propertyNameLength>0

lib/Common/DataStructures/Comparer.h
```diff
@@ -114,9 +114,11 @@ struct RecyclerPointerComparer
     }
 };
 
+// FNV-1a hash -> https://en.wikipedia.org/wiki/Fowler%E2%80%93Noll%E2%80%93Vo_hash_function
+#define CC_HASH_OFFSET_VALUE 2166136261
 #define CC_HASH_LOGIC(hash, byte) \
-    hash  = _rotl(hash, 7);        \
-    hash ^= byte;
+    hash ^= byte;                 \
+    hash *= 16777619
```
kunalspathak: as you mentioned, this seems to be very expensive given that we will do hash calculation every time we look/insert. Can you run browser benchmarks like Octane/speedometer to see how this impacts and is it worth pursuing?

obastemur: I would not say very expensive but a bit slower. Well, the only difference is full register multiply vs left and right bit + or.

kunalspathak: multiple is very expensive than bit operations, no? 😄  . It will be good to get browser benchmark numbers.

obastemur: Surely multiply is slower. However, when we have a hashCode that doesn't distribute well, we end up having multiple hashCode generations per each lookup because we have lots of collisions. Besides, we loop on WCHAR.. Well, neither of the algorithms planned for that. What would happen to our current algorithm if that WCHAR would carry a unicode? Previous algorithm fails at the beginning due to strict 128 ASCII distance from each possibility block. Popularly, for hash/map, we MOD the hash-key by the number of capacity we have. Old algorithm is being in order (128 ASCII blocks), not surprisingly results in a very bad distribution. i.e. 127 far from 128 more like to end up in 127 from 0-128 while the new algorithm generates has codes more likely to spread across the table. That's why the perf gain. I will do more benchmarks but so far so good. PS2: power of 128 approach collides even faster when the length of the data grows.

lib/Runtime/Base/ThreadContext.h
```diff
@@ -423,7 +423,24 @@ class ThreadContext sealed :
 #endif
 #endif
 
+public:
+    Js::PropertyRecord const * GetEmptyStringPropertyRecord()
+    {
+        if (!emptyStringPropertyRecord)
+        {
+            emptyStringPropertyRecord = this->UncheckedAddPropertyId(_u(""), 0);
```
MSLaguana: Is this guaranteed to happen prior to any user code? Or could I write some JS which immediately does something like `global[''] =1` to force the empty string to get a property record, and then this line here would attempt to create an additional one?

MSLaguana:
The CI failed with what looks like assertions about property IDs being internal when they shouldn't be. Sounds like it might be related to the empty property ID change that you've made?

obastemur:

lib/Runtime/Base/ThreadContext.h
obastemur: Yes, this is possible. Updated.

obastemur:
@MSLaguana @kunalspathak I have added a commit after your approvals.

Simply, I'm taking back the hash function update temporarily and making it a TODO for later. I would like to try that change after #3650

MSLaguana:
LGTM

kunalspathak:
LGTM as well.

lib/Runtime/Base/ThreadContext.h
```diff
@@ -423,7 +423,28 @@ class ThreadContext sealed :
...
 
+public:
+    Js::PropertyRecord const * GetEmptyStringPropertyRecord()
+    {
+        if (!emptyStringPropertyRecord)
+        {
+            emptyStringPropertyRecord = propertyMap->LookupWithKey(Js::HashedCharacterBuffer<char16>(_u(""), 0));
+            if (emptyStringPropertyRecord == nullptr)
+            {
+                emptyStringPropertyRecord = this->UncheckedAddPropertyId(_u(""), 0);
```
digitalinfinity: The ThreadContext is heap allocated, but the property record you're allocating here is Recycler allocated so it can cause the PropertyRecord to get collected prematurely and can cause GetEmptyStringPropertyRecord to return a pointer to a record that's already been collected.

I think this is a bug with the existing code that can cause functional issues- for now, the simplest thing to do is call UncheckedAddPropertyId(_u(""), 0, true /* bind */) so that we keep the property record alive for the duration of the thread context.

lib/Runtime/Base/ThreadContext.cpp
```diff
@@ -910,19 +911,22 @@ ThreadContext::IsNumericProperty(Js::PropertyId propertyId)
 const Js::PropertyRecord *
 ThreadContext::FindPropertyRecord(const char16 * propertyName, int propertyNameLength)
 {
-    Js::PropertyRecord const * propertyRecord = nullptr;
-
-    if (IsDirectPropertyName(propertyName, propertyNameLength))
+    if (propertyNameLength < 2)
```
digitalinfinity: Consider adding a comment here to explain that we check for 2 here because empty string property records have 0 length, and propertyNamesDirect is a cache for single character property records

lib/Runtime/Base/ThreadContext.cpp
```diff
@@ -1083,7 +1087,7 @@ ThreadContext::AddPropertyRecordInternal(const Js::PropertyRecord * propertyReco
 #if DBG
     // Only Assert we can't find the property if we are not adding a symbol.
     // For a symbol, the propertyName is not used and may collide with something in the map already.
-    if (!propertyRecord->IsSymbol())
+    if (propertyNameLength > 0 && !propertyRecord->IsSymbol())
```
digitalinfinity: Probably don't need this condition anymore? If we're adding a property record, the old invariant that it shouldn't already be in the map, should still hold

lib/Runtime/Base/ThreadContext.cpp
```diff
@@ -1132,7 +1136,7 @@ ThreadContext::AddPropertyRecordInternal(const Js::PropertyRecord * propertyReco
 #if DBG
     // Only Assert we can find the property if we are not adding a symbol.
     // For a symbol, the propertyName is not used and we won't be able to look the pid up via name.
-    if (!propertyRecord->IsSymbol())
+    if (propertyNameLength && !propertyRecord->IsSymbol())
```
digitalinfinity: Same thing here- the invariant should be that if we add a property record, it should be in the map when we try to retrieve it (if you added the check for perf, its probably not worth it since this is debug only code)

lib/Runtime/Library/JavascriptProxy.cpp
```diff
@@ -1986,11 +1986,10 @@ namespace Js
 
     void JavascriptProxy::PropertyIdFromInt(uint32 index, PropertyRecord const** propertyRecord)
     {
-        char16 buffer[20];
-
-        ::_i64tow_s(index, buffer, sizeof(buffer) / sizeof(char16), 10);
+        char16 buffer[22];
+        int pos = TaggedInt::ToBuffer(index, buffer, 22);
```
digitalinfinity: Nit: use _countof(buffer) here instead of hardcoded 22?

lib/Runtime/Base/ThreadContext.cpp
obastemur: It's for Debug to check `at this point, propertyRecord is not for a symbol`

lib/Runtime/Base/ThreadContext.h
obastemur: Good point. Thanks

PR #3506, opened 2017-08-10
Default case in InliningDecider::GetBuiltInInfoCommon

Cellule:
When turned off `ENABLE_SIMDJS` it changed the behavior of the switch case in `InliningDecider::GetBuiltInInfoCommon` 
Turns out we only use the return value for -off:<some inlining feature>, so this is just a clean up change

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/3506)

### DIFF
--- a/lib/Backend/InliningDecider.cpp
+++ b/lib/Backend/InliningDecider.cpp
@@ -624,26 +624,8 @@ bool InliningDecider::GetBuiltInInfoCommon(
         break;
 #endif
 
-#ifdef ENABLE_SIMDJS
-    // SIMD_JS
-    // we only inline, and hence type-spec on IA
-#if defined(_M_X64) || defined(_M_IX86)
     default:
-    {
-#if 0 // TODO OOP JIT, inline SIMD
-        // inline only if simdjs and simd128 type-spec is enabled.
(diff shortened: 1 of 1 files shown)


lib/Backend/InliningDecider.cpp
```diff
@@ -624,26 +624,8 @@ bool InliningDecider::GetBuiltInInfoCommon(
...
 #endif
 
-#ifdef ENABLE_SIMDJS
-    // SIMD_JS
-    // we only inline, and hence type-spec on IA
-#if defined(_M_X64) || defined(_M_IX86)
     default:
-    {
-#if 0 // TODO OOP JIT, inline SIMD
-        // inline only if simdjs and simd128 type-spec is enabled.
```
agarwal-sandeep: Since we haven't removed any SIMD specific code so far I think we should keep the code and fix this by using #else when ENABLE_SIMDJS is off

Cellule: The code was already behind `#if 0` and I confirmed with @MikeHolman that this is not something we will need in the future.
I don't see why we shouldn't cleanup the code.

agarwal-sandeep: Thanks for confirming. LGTM

PR #3509, opened 2017-08-10
17-08 ChakraCore servicing release

pleath:
Addresses the following:

CVE-2017-0228
CVE-2017-8634
CVE-2017-8635
CVE-2017-8636
CVE-2017-8637
CVE-2017-8638
CVE-2017-8640
CVE-2017-8641
CVE-2017-8645
CVE-2017-8646
CVE-2017-8647
CVE-2017-8655
CVE-2017-8656
CVE-2017-8657
CVE-2017-8658
CVE-2017-8659
CVE-2017-8670
CVE-2017-8671
CVE-2017-8672
CVE-2017-8674

Details:

[CVE-2017-8656] PreVisitCatch doesn't call SetIsCatch for all cases

The catch block has a var declaration with the same name as the destructured parameter. When we emit code for initializing vars we don't emit the initialization code for the symbol in the body as the catch param is not marked. The fix is to mark the destructured param as catch param.

Blind constants more aggressively in jitted code

Let RPC do handle marshalling for the JIT process

[CVE-2017-8647] Fix oop jit diff to check msvcrt instead of ucrt

[CVE-2017-8637] [CVE-2017-8659] Harden JIT process against bad input

[CVE-2017-8655] Prevent restoring inlinee frame with the wrong function object

Invalidate lastUsedSegment of an array at the right point

[CVE-2017-8638] Fix function object inline cache if function body got reparsed

[CVE-2017-8657] Postpone adding deferred top-level functions to the deferred function dictionary. Do this so that if byte code gen is interrupted or restarted, we are not left with orphaned functions in the dictionary that may be semi-initialized.

[CVE-2017-8658] Don't attempt to use deferred function stubs inside a formal argument list, as this leads to issues with arrow functions, nested functions, etc.

[CVE-2017-8635] Don't try to reuse property indexes on re-add of a deleted property if the object is non-extensible. Doing so exhausts the free indexes on repeated re-adds, which violates our assumption that a free index will always be available if a property has been deleted.

[CVE-2017-8671] Consider CallFlags_ExtraArg in Function.call implementation. Extra arg (for proxy, etc.) should not be a factor when copying/eliding arguments to delegate to the real target.

[CVE-2017-8674] Re-map function object types if AutoRestoreFunctionInfo has to restore FunctionInfo on failed re-parse/byte code gen.

[CVE-2017-8640] Mark the correct function node when we're detecting 'arguments' declaration that overrides the built-in declaration.

[CVE-2017-8670] Make sure we mark the correct function node when we're processing formal named 'arguments' in a destructured expression.

[CVE-2017-8672] Avoid passing JsNull as JsFunction*

In JavascriptStackWalker GetCaller we were sending out JsNull as a
JavascriptFunction* in the outparam. I checked all locations that
did a call to this function, and recursively as far as the outparam
escaped. Virtually all of the callers of the function or the other
functions that allowed escape of the outparam were safe, as checks
against the return value of GetCaller are near-ubiquitous. The one
callsite where this wasn't the case was in GlobalObject.cpp L#693,
which is in VEval. This location used the outparam without checks,
which meant that if a JsNull object were returned, we'd call other
functions which wouldn't necessarily be correct.

[CVE-2017-8645] Whitelist the type of functions allow to be asm.js module

[CVE-2017-8646] Correctly use PushPopFrameHelper in ProcessLinkFailedAsmJsModule.
No longer ignore arbitrary javascript code after a `let`

[CVE-2017-8641] Integer overflow in chakra!Js::GlobalObject
Integer overflow can happen in multiple cases of add-mult and allocate.
Fix by using UInt32Math::AddMul() to catch overflows.

[CVE-2017-8634] Type confusion in Concat due to IsConcatSpreadable

[CVE-2017-8636] Limiting args count to a certain limit. This limit will be enforced during the parser time. During the bytecode time we add so many extra args which actually overflowed the initial assumption.

[CVE-2017-0228]  We have attempted to fix the ReverseHelper case where the inline segment was swapped from head to bottom due to reverse loop below. However that fix was done for multi-segment scenario. The EnsureHeadStartsFromZero can also insert head segment in between if the segment's length is less than Array's length (look at the loop above), which will make the code vulnerable. Fixed that scenario by allocating the segment (for the single segment case) if the segment's length is less than array's length.


pleath:
(Reviewed by the contributors offline.)

PR #3512, opened 2017-08-10
OOM / SO exception counter

Cellule:
Keep a counter of OOM and S/O exceptions to help diagnose crash dumps.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/3512)

### DIFF
--- a/lib/Runtime/Language/JavascriptExceptionObject.h
+++ b/lib/Runtime/Language/JavascriptExceptionObject.h
@@ -195,6 +195,12 @@ namespace Js
         Field(JavascriptExceptionObject*) next;  // to temporarily store list of throwing exceptions
 
         PREVENT_COPY(JavascriptExceptionObject)
+    public:
+#if DBG
+        // Count how many Out of Memory and Stack overflow exceptions happened during the execution
+        static uint oomExceptionCount;
+        static uint soExceptionCount;
+#endif
     };
 
     class GeneratorReturnExceptionObject : public JavascriptExceptionObject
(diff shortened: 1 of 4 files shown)


Cellule:
@Microsoft/chakra-developers Can someone review this please ?

agarwal-sandeep:
Do you just want to know that OOM/SO happened? How is the counter helping in debugging?

Cellule:
I want to know that an exception of this type happened. I thought about making it a bool, but a counter could be helpful.
The main reason is that when I look at a dump, I want to help direct my attention to the fact that an OOM or SO exception happened or not to end up in this state.

Most likely the user will do something like
```js
try {
  doSomethingThatOOMs()
} catch (e) {
}
OtherOperationThatCrashes()
```
Since we have the dump at `OtherOperationThatCrashes` and not at `doSomethingThatOOMs` it's hard to pinpoint what could have caused us to get into this state.

lib/Runtime/Language/JavascriptExceptionObject.h
```diff
@@ -195,6 +195,12 @@ namespace Js
         Field(JavascriptExceptionObject*) next;  // to temporarily store list of throwing exceptions
 
         PREVENT_COPY(JavascriptExceptionObject)
+    public:
+#if DBG
+        // Count how many Out of Memory and Stack overflow exceptions happened during the execution
+        static uint oomExceptionCount;
+        static uint soExceptionCount;
+#endif
```
LouisLaf: Don't you want this on the scriptContext?  Otherwise, all it's telling you was that in the past few hours, since the crawler run started, some site got an OOM/SO...

Cellule: Good point

PR #3514, opened 2017-08-10
WASM - preserve nan bits

### LINKED ISSUE #2049: WASM: float32 NAN value representation in JIT

Cellule:
Introduce Float32Const IR opnd
Change a lot of function to use __vectorcall.
Do exact comparison in Wasm spec tests

This builds on top of @Krovatkin's work to pass the NaN tests of the Wasm Spec.

Fixes #2049 #2125 #2037 

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/3514)


lib/Backend/Opnd.cpp
```diff
@@ -1777,6 +1784,73 @@ FloatConstOpnd::FreeInternal(Func *func)
...
+///
+///----------------------------------------------------------------------------
+
+Float32ConstOpnd *
+Float32ConstOpnd::CopyInternal(Func *func)
+{
+    Assert(m_kind == OpndKindFloat32Const);
+    Float32ConstOpnd * newOpnd;
+
+    newOpnd = Float32ConstOpnd::New(m_value, m_type, func);
```
agarwal-sandeep: We had talked about moving away from new construct to initializing the object in constructor it self. This might be a good start?

Cellule: That means we would have to write 
```cpp
JitAnew(func->m_alloc, IR::Float32ConstOpnd, m_value, m_type);
```` 
Do we really want to go down that route ? Or am I missing something ?

Comment (agarwal-sandeep) [inline lib/Backend/Opnd.cpp]: There are static analyzer which can detect uninitialized fields but they don't work well with such construct.

**Comment by MikeHolman** [comment]
>     bool isFloat64 = opndDst->IsFloat64();

would be nice if this condition was also captured by the template

---
Refers to: lib/Backend/LowerMDShared.cpp:7011 in 2706863. [](commit_id = 27068633ba2c1234af36edfc18c9c008a380b5d4, deletion_comment = False)

File: lib/Runtime/Math/WasmMath.inl [hunk 2]
```diff
@@ -105,7 +105,21 @@ inline int WasmMath::Eqz(T value)
 template<>
 inline double WasmMath::Copysign(double aLeft, double aRight)
 {
+#if _M_IX86
+    double a = *(double*)Js::NumberConstants::AbsDoubleCst;
+    double b = *(double*)Js::NumberConstants::SgnDoubleBitCst;
+    __asm {
+        movq xmm0, aLeft;
```
MikeHolman: why is this movq? shouldn't it be movsd?

MikeHolman: in fact, i think we should be able to do this in c++ and not use inline asm at all...

---
In reply to: [132787155](https://github.com/Microsoft/ChakraCore/pull/3514#discussion_r132787155) [](ancestors = 132787155)

test/WasmSpec/convert-test-suite/index.js
```diff
@@ -114,7 +114,7 @@ function main() {
     const runs = specFiles.map(specFile => {
       const ext = path.extname(specFile);
       const basename = path.basename(specFile, ext);
-      const isXplatExcluded = config["xplat-excludes"].indexOf(path.basename(specFile, ext)) !== -1;
+      const isXplatExcluded = true || config["xplat-excludes"].indexOf(path.basename(specFile, ext)) !== -1;
```
MikeHolman: not really a fan of this, because when we enable crossplat we will need to remember to change this true || ... condition, which we almost certainly will forget to do

Cellule: well, when we'll want to enable those tests on xplat, we'll have to find this otherwise the script will always output `exclude_xplat`

Cellule:
I reverted my change to move the content of JavascriptMath to JavascriptMathOperators

Cellule:
I confirmed this change doesn't regress performance on x86.
Also, I just create a code review for changes in private repo (very minimal changes) and confirmed that it builds chakra.dll on our 2 build systems

PR #3516, opened 2017-08-11
perf: Improve hot MOD arithmetic

obastemur:
1 - Once the modulo number is known, compiler can optimize modulo aritmetic
(i.e. doesn't use DIV..) This brings up to 10 times less cpu cycles per
each modulo arithmetic. Used HashTable / Dictionary etc.

2 - For PrimePolicy, we know majority of the possible modulo numbers. Pow2 size policy already uses the special case.

3 - GetLevel2Id (hot function), QuickSort mod operation(s) are optimized to Pow2 special case since the MOD number is power of 2.

Well, we have bunch of other MOD arithmetic usage all over the places. However, these are the hot ones and reasonable to improve.

### DIFF
--- a/lib/Common/DataStructures/SizePolicy.cpp
+++ b/lib/Common/DataStructures/SizePolicy.cpp
@@ -9,8 +9,60 @@ static const uint primes[] = {
     1103, 1327, 1597, 1931, 2333, 2801, 3371, 4049, 4861, 5839, 7013, 8419, 10103, 12143, 14591,
     17519, 21023, 25229, 30293, 36353, 43627, 52361, 62851, 75431, 90523, 108631, 130363, 156437,
     187751, 225307, 270371, 324449, 389357, 467237, 560689, 672827, 807403, 968897, 1162687, 1395263,
-    1674319, 2009191, 2411033, 2893249, 3471899, 4166287, 4999559, 5999471, 7199369
+    1674319, 2009191, 2411033, 2893249, 3471899, 4166287, 4999559, 5999471, 7199369, 10453007, 13169977, 20906033
 };
+
+// If we know the modulo number at compile time, the compiler can optimize instructions.
+// These macros let the compiler do these optimizations for modulus optimizations of known primes
+static_assert(_countof(primes) == UNKNOWN_MOD_INDEX, "Adding / removing prime numbers ?");
+
+#define MOD_FNC_(x) static uint mod##x(const uint k, const uint) { return k % x##llu; }
+#define MOD_FNC__(a,b,c,d,e) MOD_FNC_(a) MOD_FNC_(b) MOD_FNC_(c) MOD_FNC_(d) MOD_FNC_(e)
+#define MOD_FNC(a,b,c,d,e,f,g,h,j,k) MOD_FNC__(a,b,c,d,e) MOD_FNC__(f,g,h,j,k)
+
+MOD_FNC(3, 7, 11, 17, 23, 29, 37, 47, 59, 71)
+MOD_FNC(89, 107, 131, 163, 197, 239, 293, 353, 431, 521)
+MOD_FNC(631, 761, 919, 1103, 1327, 1597, 1931, 2333, 2801, 3371)
+MOD_FNC(4049, 4861, 5839, 7013, 8419, 10103, 12143, 14591, 17519, 21023)
+MOD_FNC(25229, 30293, 36353, 43627, 52361, 62851, 75431, 90523, 108631, 130363)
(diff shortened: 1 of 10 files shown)


lib/Common/DataStructures/SizePolicy.h
```diff
@@ -5,39 +5,44 @@
 #pragma once
 
 
+#define UNKOWN_MOD_INDEX 75 // count of the known mod array (under SizePolicy.cpp)
```
kfarnung: nit: typo `UNKNOWN_MOD_INDEX`

lib/Common/DataStructures/SizePolicy.cpp
```diff
@@ -9,9 +9,59 @@ static const uint primes[] = {
...
     187751, 225307, 270371, 324449, 389357, 467237, 560689, 672827, 807403, 968897, 1162687, 1395263,
-    1674319, 2009191, 2411033, 2893249, 3471899, 4166287, 4999559, 5999471, 7199369
+    1674319, 2009191, 2411033, 2893249, 3471899, 4166287, 4999559, 5999471, 7199369, 10453007, 13169977, 20906033
 };
 
+static_assert(_countof(primes) == UNKOWN_MOD_INDEX, "Adding / removing prime numbers ?");
+
+#define MOD_FNC_(x) static uint mod##x(const uint k, const uint) { return k % x##llu; }
+#define MOD_FNC__(a,b,c,d,e) MOD_FNC_(a) MOD_FNC_(b) MOD_FNC_(c) MOD_FNC_(d) MOD_FNC_(e)
+#define MOD_FNC(a,b,c,d,e,f,g,h,j,k) MOD_FNC__(a,b,c,d,e) MOD_FNC__(f,g,h,j,k)
```
MSLaguana: Is there a reason that `i` is missing from this list?

lib/Common/DataStructures/SizePolicy.cpp
```diff
@@ -9,9 +9,59 @@ static const uint primes[] = {
     1103, 1327, 1597, 1931, 2333, 2801, 3371, 4049, 4861, 5839, 7013, 8419, 10103, 12143, 14591,
     17519, 21023, 25229, 30293, 36353, 43627, 52361, 62851, 75431, 90523, 108631, 130363, 156437,
     187751, 225307, 270371, 324449, 389357, 467237, 560689, 672827, 807403, 968897, 1162687, 1395263,
-    1674319, 2009191, 2411033, 2893249, 3471899, 4166287, 4999559, 5999471, 7199369
+    1674319, 2009191, 2411033, 2893249, 3471899, 4166287, 4999559, 5999471, 7199369, 10453007, 13169977, 20906033
 };
 
+static_assert(_countof(primes) == UNKOWN_MOD_INDEX, "Adding / removing prime numbers ?");
+
+#define MOD_FNC_(x) static uint mod##x(const uint k, const uint) { return k % x##llu; }
```
mike-kaufman: some comments in here RE what you're trying to achieve with the macros will be helpful for new people down the road. e.g., 
```
// If we know the modulo number at compile time, the compiler can optimize instructions.  
// These macros let the compiler do these optimizations for modulus optimizations of known primes
```

mike-kaufman: And I'm not following what the 2nd (unnamed) param is for on line 17.

obastemur: second param carries the hash in case the original number is beyond the range of optimized MOD zone. see UNKNOWN_MOD_INDEX

lib/Common/DataStructures/SizePolicy.cpp
obastemur: well, no particular reason.

lib/Common/DataStructures/SizePolicy.h
obastemur: Thanks will fix

obastemur:
> What is the perf gain from this?

hashing a string, calls MOD op. string.length times. This PR, those prime MOD ops are 10 times faster.

lib/Common/DataStructures/SizePolicy.cpp
obastemur: good one. added!

obastemur:
From our usage: each time we compare hash to buckets, find new buckets, copy buckets.. we use MOD of primes. (Basically, each and every data structure we have with PrimePolicy)

obastemur:
@kfarnung @MSLaguana  @mike-kaufman Thanks for the review. Will merge once the CI checks are done.

Issue #3254, opened 2017-06-28
Duplicate VarDeclaredName and LexicallyDeclaredName is always an error in Modules


bterlson:
```js
var f;
function f() {}
```
Actual: No Error
Expected: Syntax Error (redeclaration of f).

Issue #3275, opened 2017-06-30
Eval containing a function declaration with a same-named lexical declaration in outer scope should throw


bterlson:
```js
let a = 1;
eval('function a() { };');
```
Expected: SyntaxError per https://tc39.github.io/ecma262/#sec-evaldeclarationinstantiation 5.a.i.1
Actual: no syntax error, function is assigned to let variable.

```js
(function () {
  let a = 1;
  eval('function a() { }');
})();
```
Expected: SyntaxError per https://tc39.github.io/ecma262/#sec-evaldeclarationinstantiation 5.d.ii.2.a.i
Actual: no syntax error, function is assigned to let variable.

Note 1: The order of the lexical declaration and eval are not important - it should be the same error in either case.
Note 2: annex B semantics don't come in to play here

PR #3517, opened 2017-08-11
Fix issue#3254: Redeclaration error related to function declaration in module top-level

suwc:
Function declarations in module top-level should be treated as a lexical declarations,
and early error should be thrown upon redeclaration. Fix by checking for function
declaration in module top-level in redeclaration code path.

15.2.1.1Static Semantics: Early Errors
It is a Syntax Error if any element of the LexicallyDeclaredNames of ModuleItemList also
occurs in the VarDeclaredNames of ModuleItemList

15.2.1.11 NOTE 2
At the top level of a Module, function declarations are treated like lexical declarations
rather than like var declarations.

Fixes https://github.com/Microsoft/ChakraCore/issues/3254

### DIFF
--- a/lib/Parser/Parse.cpp
+++ b/lib/Parser/Parse.cpp
@@ -861,6 +861,13 @@ Symbol* Parser::AddDeclForPid(ParseNodePtr pnode, IdentPtr pid, SymbolType symbo
                 // If !errorOnRedecl, (old) let/const hides the (new) var, so do nothing.
                 break;
             case knopVarDecl:
+                // At the top level of a Module, function declarations are treated like lexical declarations rather than like var declarations.
+                if (errorOnRedecl && m_scriptContext->GetConfig()->IsES6ModuleEnabled() && IsImportOrExportStatementValidHere() &&
+                    (symbolType == STFunction || sym->GetSymbolType() == STFunction))
+                {
+                    Error(ERRRedeclaration);
+                }
+
                 // Legal redeclaration. Who wins?
                 if (errorOnRedecl || sym->GetDecl()->sxVar.isBlockScopeFncDeclVar || sym->GetIsArguments())
                 {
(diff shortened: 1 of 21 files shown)


suwc:
@dotnet-bot
test Ubuntu shared_ubuntu_linux_release

lib/Parser/Parse.cpp
```diff
@@ -837,6 +837,14 @@ Symbol* Parser::AddDeclForPid(ParseNodePtr pnode, IdentPtr pid, SymbolType symbo
             }
             break;
         case knopVarDecl:
+
```
pleath: Can you move this down to the switch statement just below that handles redeclarations? And do we not have issues involving function redeclaration errors in other situations? I wonder if it's necessary to have a fix that's specific to modules.

suwc: Moved to below case knopVarDecl.
Treating function declaration like lexical declaration is unique to module top-level.

pleath: Function-in-block is essentially a lexical declaration. The module-top-level issue seems equivalent to the issue in a top-level block. Given:
{
    function f(){};
    var f;
}

both Chrome and FF throw redeclaration errors, and we don't. So we do seem to have a general problem with lexical declarations of functions that is not limited to modules. 

The issue also seems similar in spirit to https://github.com/Microsoft/ChakraCore/issues/3275 .

test/es6/module-syntax.js
```diff
@@ -48,6 +48,15 @@ var tests = [
         }
     },
     {
+        name: "Syntax error of redeclaration in function declaration at module top level",
+        body: function () {
+            testModuleScript('var f; function f() {}', 'Syntax error of redeclaration in function declaration in module\'s top-level', true);
```
aneeshdk: >function f() {} [](start = 37, length = 15)

Add a test case where function definition comes before var declaration. Also a let declaration case too.

lib/Parser/perrors.h
```diff
@@ -54,7 +54,7 @@ LSC_ERROR_MSG( 1048, ERRKeywordNotId  , "The use of a keyword for an identifier
 LSC_ERROR_MSG( 1049, ERRFutureReservedWordNotId, "The use of a future reserved word for an identifier is invalid")
 LSC_ERROR_MSG( 1050, ERRFutureReservedWordInStrictModeNotId, "The use of a future reserved word for an identifier is invalid. The identifier name is reserved in strict mode.")
 LSC_ERROR_MSG( 1051, ERRSetterMustHaveOneParameter, "Setter functions must have exactly one parameter")
-LSC_ERROR_MSG( 1052, ERRRedeclaration  , "Let/Const redeclaration") // "var x; let x;" is also a redeclaration
+LSC_ERROR_MSG( 1052, ERRRedeclaration  , "Identifier redeclaration") // "var x; let x;" is also a redeclaration
```
aneeshdk: >Identifier [](start = 42, length = 10)

If the function definition in this case is treated as lexical declaration then do we really need to change the error string?

PR #3518, opened 2017-08-11
DetermineSymBoundOffsetOrValueRelativeToLandingPad: Handle case where we are unable to find the matching bound

Cellule:
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/3518)


meg-gupta:
This PR fixes OS#13155679 as well. I will dup it to you.

PR #3522, opened 2017-08-14
Fix #2983 WeakMap + HostObject === Sadness (TypeError when using HostObject with WeakMap)

### LINKED ISSUE #2983: WeakMap + HostObject === Sadness (TypeError when using HostObject with WeakMap)
```js
let map = new WeakMap();
map.set(window.location, '1'); // TypeError: WeakMap.prototype.set: 'key' is not an object
```

source: https://github.com/emberjs/ember.js/issues/15249

cc @jdalton / @nolanlawson

tcare:
Our implementation of WeakMap and WeakSet uses internal properties on DynamicObjects to store the <WeakMap pointer, value> pair data. This allows us to follow the lifetime of the object. Unfortunately, HostDispatch objects such as window.location are not DynamicObjects, and so we throw a type error. We did not expect widespread use of HostDispatch objects as keys, so this was left as a TODO.

Since Ember is now encountering this with window.location, we need to support this scenario.

This change allows HostDispatch objects to be used as a WeakMap key. HostDispatch has a matching change to support this WeakMap data. No other RecyclableObjects are supported, and we assert as such when trying to access the object key map data from within it.

### DIFF
--- a/lib/Runtime/Library/JavascriptWeakMap.cpp
+++ b/lib/Runtime/Library/JavascriptWeakMap.cpp
@@ -24,9 +24,12 @@ namespace Js
         return static_cast<JavascriptWeakMap *>(RecyclableObject::FromVar(aValue));
     }
 
-    JavascriptWeakMap::WeakMapKeyMap* JavascriptWeakMap::GetWeakMapKeyMapFromKey(DynamicObject* key) const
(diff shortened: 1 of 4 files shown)


digitalinfinity:
:shipit:

PR #3524, opened 2017-08-14
Added a function to change the callback state of a JS external function

nullcatalyst:
The functionality to change the callback state of a js::JavascriptExternalFunction already existed inside of the library, but there was no externally visible API in ChakraCore to use this functionality.

### DIFF
--- a/lib/Jsrt/ChakraCommon.h
+++ b/lib/Jsrt/ChakraCommon.h
@@ -2138,6 +2138,25 @@ typedef unsigned short uint16_t;
             _In_opt_ void *callbackState,
             _Out_ JsValueRef *function);
 
+
+    /// <summary>
+    ///     Changes the internal callback state for an external function.
+    /// </summary>
+    /// <remarks>
+    ///     Requires an active script context.
+    /// </remarks>
+    /// <param name="function">The function object.</param>
+    /// <param name="callbackState">
+    ///     User provided state that will be passed back to the callback.
+    /// </param>
+    /// <returns>
+    ///     The code <c>JsNoError</c> if the operation succeeded, a failure code otherwise.
+    /// </returns>
+    CHAKRA_API
+        JsSetExternalFunctionState(
+              _In_ JsValueRef function,
+              _In_opt_ void *callbackState);
+
     /// <summary>
     ///     Creates a new JavaScript error object
     /// </summary>
--- a/lib/Jsrt/Jsrt.cpp
+++ b/lib/Jsrt/Jsrt.cpp
@@ -2432,6 +2432,22 @@ CHAKRA_API JsCreateNamedFunction(_In_ JsValueRef name, _In_ JsNativeFunction nat
     });
 }
 
+CHAKRA_API JsSetExternalFunctionState(_In_ JsValueRef function, _In_opt_ void *callbackState)
+{
+    return ContextAPIWrapper<JSRT_MAYBE_TRUE>([&](Js::ScriptContext *scriptContext, TTDRecorder& _actionEntryPopper) -> JsErrorCode {
+        VALIDATE_INCOMING_FUNCTION(function, scriptContext);
+
+        Js::JavascriptFunction *jsFunction = Js::JavascriptFunction::FromVar(function);
+        if(jsFunction != JS_INVALID_REFERENCE && jsFunction->IsExternalFunction())
+        {
+            Js::JavascriptExternalFunction *externalFunction = (Js::JavascriptExternalFunction *) jsFunction;
+            externalFunction->SetCallbackState(callbackState);
+        }
+
+        return JsNoError;
+    });
+}
+
 void SetErrorMessage(Js::ScriptContext *scriptContext, JsValueRef newError, JsValueRef message)
 {
     Js::JavascriptOperators::OP_SetProperty(newError, Js::PropertyIds::message, message, scriptContext);


liminzhu:
Thanks for the PR @nullcatalyst ! Is there a scenario you're hoping to enable with this new API?

nullcatalyst:
So I am currently using ChakraCore as the scripting engine for a game I've been making. I passed some external functions to the JS context so that they can be called to initialize different parts of the game. Also worth noting is that I intend to use JS functions as event handlers, eg: collide with enemy.

The problem: I don't want these external init functions stored in a closure on these event handlers and called after initialization is complete. I would also like to be able to free the data associated with the initialization objects -- ideally, I shouldn't have to use a handle to store the pointer, and check if the value the handle points to has changed to null, I would rather not have the extra allocated handle, nor the extra layer of indirection.

By being able to reset the `callbackState` value, I can free the init objects and handle calling the init JS functions by checking for null, and thus throw an error accordingly.

nullcatalyst:
That was probably more than you needed to know, but context is everything. Maybe you can think of an alternative solution for me :)

liminzhu:
Certainly appreciate the context @nullcatalyst .

Am I to understand the problem as you have a bunch of native init functions (probably gets triggered within event handlers) and want to make sure are only called once (or the second time they throw or do nothing) and you want to be able to free associated data after the initialization? 

I know you mentioned 
> I don't want these external init functions stored in a closure on these event handlers 

but in JS land, wouldn't it be more direct solution to store init a closure to make sure the native functions are never called twice?
```js
let init = (function() {
    let initialized = false;
    return (function() {
        if (!initialized) initNative();
        initialized = true;
    })
})
```

On the native side, you can easily have a variable to track if the native initializer was ever called, and free stuff at the end of the JsNativeFunction.

I'm probably still missing something, but I like having the api interface lean, so I always play devil's advocate to understand and make sure every new API has clear use cases :).

nullcatalyst:
Your snippet of code would be how I would typically do it in JS, but I can't quite do it that way since I would like to allow the script to call the function an unknown number of times, not just once.

I have exposed a function `addEnemyType()` that can be called multiple times to create and initialize new enemy types, but I only want to add new enemy types at the beginning of the level.

I have thought about changing the value of `addEnemyType` on the global object, but that fails to stop the user from storing a reference to its previous value, using the same technique that you used to call the function only once:

```js
let callPrevAddEnemyType = (function() {
    let cachedFn = addEnemyType;
    return (function(...args) {
        cachedFn(...args);
    });
})();

addEnemyType({
    onCollideWithPlayer: function () {
        callPrevAddEnemyType(); // oops
    },
});
```

The only way around it that I can see at this moment is changing my exposed API

liminzhu:
Gotcha. 

You can use `callbackstate` to store a pointer to a data structure that logs whether an enemy type is initialized and what not, and the host will have access to the info without exposing it to JS user land. For example, see how `callbackstate` is used as task queue in PromiseContinuationCallback [sample](PromiseContinuationCallback).

I'm not sure if there's any new scenario unlocked with the API since you can pass a pointer?

nullcatalyst:
That was how I've been temporarily doing it. I had a `bool loaded` property on the passed object that gets checked prior to the native method doing its actual implementation.

Unfortunately, there is still a problem with that -- being that I cannot seem to find a way through the ChakraCoreCommon API to track the lifetime of the object I pass as `callbackState`. When I call `JsCreateNamedFunction`, passing in the object to be used in the callback, I can either free the object after running the script (and leave the callbackState as dangling pointer that I cannot change) or forget about the object, since unlike the `JsCreateExternalObject` function I do not get a `JsFinalizeCallback` to free the memory associated with it.

I am all for a lean API. But there are too many gaps in the CRUD lifecycle at the moment -- I am stuck between a dangling pointer and a memory leak.

nullcatalyst:
Oh and by the way, I am getting a 404 for the sample link you provided. I cannot find https://github.com/Microsoft/ChakraCore/pull/PromiseContinuationCallback

kfarnung:
I think @liminzhu was approximately referring to https://github.com/Microsoft/ChakraCore/wiki/JavaScript-Runtime-%28JSRT%29-Overview#promises which gives an example of callbackState usage.

liminzhu:
@nullcatalyst ugh sorry for the 404 link and thanks @kfarnung for providing the correct one.

I think your function needs to be aware of and check `callbackstate` anyways. Being able to reset `callbackstate` doesn't change the matter that you need to do something to `callbackstate` (call reset or point it to null).

A finalizer is of course not what you want since you need to free memory before Chakra GC deems the associated function in JS out of reach.

nullcatalyst:
The promise callback queue is a neat idea, but still suffers from the same problem of storing a reference to the addEnemyType function and calling it inside an event callback that I store a reference to on the longer lived enemy type object. If the addEnemyType JS function is called at that point, the C function will be passed a non-null pointer (the callbackState) to a stack allocated object that has gone out of scope.

As for the finalizer, I meant a C callback that ChakraCore would call prior to releasing the external JS function, in order to handle deallocating any memory needed to run that function.

For my purposes, I am now statically linking to a custom compiled version of ChakraCore with this added in. I was hoping to avoid that, but I have to keep moving forward. If that becomes cumbersome to keep up to date, I have also added support for JavaScriptCore to port over to iOS, so I also have that.

dilijev:
@boingoing would you be willing to adopt this PR?

dilijev:
Would we want to retarget this for release/1.10 or keep in master?
@jackhorton one of the test failures is `Failed -> Strings/unicode_toUpperCase_toLowerCase.js` -- you have a PR for that, is this a known failure?

boingoing:
I would vote to keep this in master.

jackhorton:
I think this needs a rebase, I am not aware of any failures in that test.

liminzhu:
Sorry I forgot to reply earlier :/.

I thought about this more. One way to deal with this problem is to register a finalizer on the function using [JsSetObjectBeforeCollectCallback](https://github.com/Microsoft/ChakraCore/wiki/JsSetObjectBeforeCollectCallback) to clean up previously passed callbackState or anything else. Therefore we don't have to worry about a dangling pointer. 

I'm a bit resistant to the proposed API b/c we have a whole bunch of callbacks that accept a callbackState and I'm not sure how and if we need to deal with all those.

PR #3526, opened 2017-08-14
OS#9030969 fix stack trace with jitted loops

sigatrev:
native code address was being cleared more often than it should, causing the trace to use the line/column numbers of the last bytecode instruction executed, which is not where the error occurs if it was during jitted code.


rajatd:
>         {

Setting this->previousInterpreterFrameIsForLoopBody to false here would be a better fix. Make sure to run exprgen for a long time

---
Refers to: lib/Runtime/Language/JavascriptStackWalker.cpp:680 in 4cd121b. [](commit_id = 4cd121b9846b8de0c138657801c98452c1cbb2b4, deletion_comment = False)

PR #3527, opened 2017-08-14
clarifying error message when trying to change non-configurable, non-writable property

### LINKED ISSUE #424: Fix error message for changing [[Writable]] to true from false for non-configurable properties

mike-kaufman:
when trying to change e a non-configurable, non-writable property to a writable property, we would emit an error message saying "Cannot modify non-writable property".  Now, we say "Cannot redefine non-configurable property" since it is more accurate.  Also updated tests to check for correct message in the error that is raised. 

Fixes  #424.


rajatd:
Looks good.

Issue #3250, opened 2017-06-28
Module crash in circular import case


bterlson:
```js
// bug.js
WScript.LoadModule(`import "bug-dep.js"`);

// bug-dep.js
import * as ns from './bug-dep.js';

export let c_localUninit1;
export { c_localUninit1 as f_indirectUninit } from './bug-dep2.js';

var stringKeys = Object.getOwnPropertyNames(ns);
print(JSON.stringify(stringKeys));

// bug-dep2.js
export * from "./bug-dep.js";
```

Actual: crash
Expected: ["c_localUninit1","f_indirectUninit"]

PR #3533, opened 2017-08-16
module circular reference GetExportedNames/ResolveExport bugs

jianchun:
GetExportedNames(): module namespace binds all names from
`GetExportedNames(nullptr)`. However we incorrectly cached result from any
`GetExportedNames(exportStarSet)` call. This causes module namespace
having incomplete names, because a previous intermediate call with non-null
exportStarSet did not include names from that set. This fails 3250-ext-a.
Found this problem while investigating #3250. Fixed by only cache/reuse the
result if exportStarSet is nullptr (root call).

ResolveExport(): parameter `exportStarSet` were used to prevent infinite
loop when looking up name under module circular references. However it
broke the case that we lookup this module record recursively for a
different name (indirect). This causes #3250 to throw SyntaxError for
failing to resolve a binding (resolving f_indirectUninit for dep2, reenter
to resolve c_localUninit1 but rejected by `exportStarSet`). Fixed by
removing it -- another parameter 'resolveSet' already does the job well. It
prevents recursion for the same {moduleRecord, name}, but allows looking up
a different name.

Finally, moved a fragment into try/catch and added an ENTER_SCRIPT_IF. The
given call may throw JavaScriptException which requires in script.

### DIFF
--- a/lib/Runtime/Language/ModuleNamespace.cpp
+++ b/lib/Runtime/Language/ModuleNamespace.cpp
@@ -94,7 +94,7 @@ namespace Js
             exportedNames->Map([&](PropertyId propertyId) {
                 JavascriptString* propertyString = scriptContext->GetPropertyString(propertyId);
                 sortedExportedNames->Add(propertyString);
-                if (!moduleRecord->ResolveExport(propertyId, nullptr, nullptr, &moduleNameRecord))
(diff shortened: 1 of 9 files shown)


jianchun:
Thanks @suwc !

PR #3534, opened 2017-08-16
change how we shift addresses from JIT proc to be more futureproof

MikeHolman:
Previously we were statically mandating what dll is used for fixing up JIT helper calls, but it is more resilient if we decide at runtime based on which dll the address falls into.

### DIFF
--- a/lib/Common/Core/SysInfo.cpp
+++ b/lib/Common/Core/SysInfo.cpp
@@ -397,10 +397,33 @@ LPCWSTR AutoSystemInfo::GetJscriptDllFileName()
     return (LPCWSTR)Data.binaryName;
 }
 
+#ifdef _WIN32
 HMODULE AutoSystemInfo::GetCRTHandle()
 {
-    return GetModuleHandle(_u("msvcrt.dll"));
+    if (Data.crtBase == 0)
+    {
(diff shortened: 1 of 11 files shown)


lib/Common/Core/SysInfo.cpp
```diff
@@ -397,10 +397,33 @@ LPCWSTR AutoSystemInfo::GetJscriptDllFileName()
     return (LPCWSTR)Data.binaryName;
 }
 
+#ifdef _WIN32
 HMODULE AutoSystemInfo::GetCRTHandle()
 {
-    return GetModuleHandle(_u("msvcrt.dll"));
```
leirocks: Attackers can corrupt this memory?

lib/Common/Core/SysInfo.cpp
```diff
@@ -397,10 +397,33 @@ LPCWSTR AutoSystemInfo::GetJscriptDllFileName()
...
+    if (Data.crtBase == 0)
+    {
+        Data.crtBase = GetModuleHandle(_u("msvcrt.dll"));
```
leirocks: Attackers can corrupt this memory?

MikeHolman: Fair enough. I was thinking we've already lost if they can corrupt this, but i could see how this might make things a bit harder

PR #3536, opened 2017-08-16
Do not peep closure stack symbols

meg-gupta:


meg-gupta:
This bug came up in swb branch for test32:core\test\Bugs\randomBugs.js. I don't have a repro for without the flags  -forceSoftwareWriteBarrier -verifyBarrierBit yet. It requires a specific pattern of register allocation to raise this bug. With the flags, the result of consecutive LdSlotArr s are assigned 2 different registers and peeps optimizes it incorrectly. Without the flags, am unable to repro such a scenario yet.

lib/Backend/Peeps.cpp
```diff
@@ -387,6 +387,11 @@ Peeps::SetReg(RegNum reg, StackSym *sym)
     this->ClearReg(sym->scratch.peeps.reg);
```
LouisLaf: > this->ClearReg(sym->scratch.peeps.reg); [](start = 3, length = 40)

This should probably move bellow your new code.

meg-gupta: You are right, will move this below.

PR #3537, opened 2017-08-17
Prioritize integer type in Lifetime

Cellule:
No longer force us to check if the lifetime is a float or any of the simd types before determining that it's an integer type.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/3537)

### DIFF
--- a/lib/Backend/LinearScan.cpp
+++ b/lib/Backend/LinearScan.cpp
@@ -2666,72 +2666,28 @@ RegNum
 LinearScan::FindReg(Lifetime *newLifetime, IR::RegOpnd *regOpnd, bool force)
 {
     BVIndex regIndex = BVInvalidIndex;
-    IRType type;
+    bool isRegAvailable = true;
     bool tryCallerSavedRegs = false;
     BitVector callerSavedAvailableBv;
 
     if (newLifetime)
     {
-        if (newLifetime->isFloat)
+        if (newLifetime->IsInt())
         {
-            type = TyFloat64;
-        }
-        else if (newLifetime->isSimd128F4)
-        {
-            type = TySimd128F4;
-        }
-        else if (newLifetime->isSimd128I4)
-        {
-            type = TySimd128I4;
-        }
-        else if (newLifetime->isSimd128I8)
-        {
-            type = TySimd128I8;
-        }
-        else if (newLifetime->isSimd128I16)
-        {
-            type = TySimd128I16;
-        }
-        else if (newLifetime->isSimd128U4)
-        {
-            type = TySimd128U4;
-        }
-        else if (newLifetime->isSimd128U8)
-        {
-            type = TySimd128U8;
-        }
-        else if (newLifetime->isSimd128U16)
-        {
-            type = TySimd128U16;
-        }
-        else if (newLifetime->isSimd128B4)
-        {
-            type = TySimd128B4;
-        }
-        else if (newLifetime->isSimd128B8)
-        {
-            type = TySimd128B8;
-        }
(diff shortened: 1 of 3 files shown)


Cellule:
@Krovatkin FYI

lib/Backend/SccLiveness.cpp
```diff
@@ -399,18 +399,7 @@ SCCLiveness::ProcessSrc(IR::Opnd *src, IR::Instr *instr)
...
-                lifetime->isSimd128I8 = symOpnd->IsSimd128I8();
-                lifetime->isSimd128I16 = symOpnd->IsSimd128I16();
-                lifetime->isSimd128U4 = symOpnd->IsSimd128U4();
-                lifetime->isSimd128U8 = symOpnd->IsSimd128U8();
-                lifetime->isSimd128U16 = symOpnd->IsSimd128U16();
-                lifetime->isSimd128B4 = symOpnd->IsSimd128B4();
-                lifetime->isSimd128B8 = symOpnd->IsSimd128B8();
-                lifetime->isSimd128B16 = symOpnd->IsSimd128B16();
-                lifetime->isSimd128D2 = symOpnd->IsSimd128D2();
+                lifetime->isFloat = symOpnd->IsFloat() || symOpnd->IsSimd128();
```
LouisLaf: >lifetime->isFloat = symOpnd->IsFloat() || symOpnd->IsSimd128(); [](start = 16, length = 63)

Can we move this (and region) to the constructor?  I hate the fact that they are separate from everything else.

Cellule: Can I rely on the StackSym's type or do I have to pass the Opnd ? 
I thought you said there were some bugs where the StackSym's type wouldn't match the Opnd's.

lib/Backend/LinearScan.cpp
```diff
@@ -2909,15 +2865,14 @@ LinearScan::Spill(Lifetime *newLifetime, IR::RegOpnd *regOpnd, bool dontSpillCur
 {
     uint minSpillCost = (uint)-1;
 
-    Assert(!newLifetime || !regOpnd || newLifetime->isFloat == (regOpnd->GetType() == TyMachDouble) || newLifetime->isSimd128() == (regOpnd->IsSimd128()));
```
LouisLaf: >newLifetime->isFloat == (regOpnd->GetType() == TyMachDouble) || newLifetime->isSimd128() == (regOpnd->IsSimd128())); [](start = 39, length = 116)

Can't we keep this assert?  It catches bad typing bugs in the IR...

lib/Backend/SccLiveness.cpp
LouisLaf: You can't...  Alright, I'll fix it.

---
In reply to: [133602310](https://github.com/Microsoft/ChakraCore/pull/3537#discussion_r133602310) [](ancestors = 133602310)

lib/Backend/LinearScan.cpp
Cellule: The lifetime no longer knows the difference between float and simd, so we can't check

Krovatkin:
@Cellule thanks for the FYI!

lib/Backend/LinearScan.cpp
LouisLaf: The point of the assert is that int regOpnd use int registers, and float/simd regOpnd use float registers.

lib/Backend/SccLiveness.cpp
LouisLaf: Nope, you can't... :-/

PR #3539, opened 2017-08-17
WASM: Table grow(0) with initial: 0

Cellule:
OS#13149792 
Do not call RecyclerNew with length 0 in case we call WebAssemblyTable.grow with initial length of 0. 
This crashes in clang, but is fine on msvc since we don't actually access whatever the Recycler returns

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/3539)


Cellule:

PR #3541, opened 2017-08-17
Better Debug coverage, smaller binary size

obastemur:
- add debug symbols to test builds
- keep libwabt debug symbols are being included on release builds.


Cellule:
:shipit:

obastemur:
@Cellule @digitalinfinity Thanks for the review!

Issue #3434, opened 2017-07-26
Heap overflow in CreateTTDDirectoryAsNeeded


rajatd:
Repro:
build chakracore on linux:
./build.sh --cc=/usr/bin/clang --cxx=/usr/bin/clang++ --arch=amd64 --debug --static -j 8 --sanitize=address,undefined,signed-integer-overflow

run `./ch -` from `/out/Debug/bin/ch`

output: 

==67392==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x60200000eec4 at pc 0x7f4ec99064bf bp 0x7ffcc6c0ab30 sp 0x7ffcc6c0ab28
READ of size 2 at 0x60200000eec4 thread T0
    #0 0x7f4ec99064be in PAL_wcslen /home/chakraut/ChakraCore/pal/src/cruntime/wchar.cpp:1212:12
    #1 0x7f4ec9845b78 in Helpers::CreateTTDDirectoryAsNeeded(unsigned long*, char*, char const*, char16_t const*) /home/chakraut/ChakraCore/bin/ch/Helpers.cpp:440:41
    #2 0x7f4ec98462a0 in Helpers::GetTTDDirectory(char16_t const*, unsigned long*, char*, unsigned long) /home/chakraut/ChakraCore/bin/ch/Helpers.cpp:499:5
    #3 0x7f4ec9834a79 in main /home/chakraut/ChakraCore/bin/ch/ch.cpp:986:13
    #4 0x7f4ec776c2b0 in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x202b0)
    #5 0x7f4ec975cf79 in _start (/home/chakraut/ChakraCore/out/Debug/bin/ch/ch+0xb60f79)

0x60200000eec4 is located 12 bytes to the left of 10-byte region [0x60200000eed0,0x60200000eeda)
allocated by thread T0 here:
    #0 0x7f4ec97fb578 in __interceptor_malloc (/home/chakraut/ChakraCore/out/Debug/bin/ch/ch+0xbff578)
    #1 0x7f4ec98cc577 in CorUnix::InternalMalloc(unsigned long) /home/chakraut/ChakraCore/pal/src/cruntime/malloc.cpp:114:20
    #2 0x7f4ec98cc544 in PAL_malloc /home/chakraut/ChakraCore/pal/src/cruntime/malloc.cpp:98:12
[...]

PR #3542, opened 2017-08-17
fixing bug in wcssstr

mike-kaufman:
Fixing https://github.com/Microsoft/ChakraCore/issues/3434..  Bug in wcsstr would return a match if the string being searched ended with a prefix of the search string.  Fixed this to behave correctly.  Also added tests to test_native.sh to verify this.

### DIFF
--- a/test/native-tests/test-pal/sample.cpp
+++ b/test/native-tests/test-pal/sample.cpp
@@ -3,39 +3,34 @@
 // Licensed under the MIT license. See LICENSE.txt file in the project root for full license information.
 //-------------------------------------------------------------------------------------------------------
 
-#include "pal.h"
+extern "C"  int
+__cdecl
+PAL_wcscmp(
+	   const char16_t *string1,
+	   const char16_t *string2);
 
-int wscmp(const char16_t *left, const char16_t *right)
-{
-    if (!left && !right)
-    {
-        return 0;
-    }
-    else if (left && right)
-    {
-        while (*left != u'\0' && *right != u'\0')
-        {
-            int diff = *right - *left;
-            if (diff != 0)
-            {
-                return diff;
-            }
-            ++left;
-            ++right;
-        }
+extern "C" const char16_t *
+__cdecl
+PAL_wcsstr(
+	   const char16_t *string,
+	   const char16_t *strCharSet);
 
-        int diff = *right - *left;
-        return diff;
-    }
-    else
-    {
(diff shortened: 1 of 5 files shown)


test/native-tests/test-pal/sample.cpp
```diff
@@ -0,0 +1,70 @@
+//-------------------------------------------------------------------------------------------------------
+// Copyright (C) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE.txt file in the project root for full license information.
+//-------------------------------------------------------------------------------------------------------
+
+#include "pal.h"
+
+int wscmp(const char16_t *left, const char16_t *right)
```
obastemur: why don't we use PAL_wcscmp?

mike-kaufman: you're right, for some reason I thought this was removed from our copy.  I'll fix...

test/native-tests/test_native.sh
```diff
@@ -42,13 +47,7 @@ TEST () {
...
 }
 
-RES=$(c++ --version)
-if [[ ! $RES =~ "Apple LLVM" ]]; then
-    FIND_CLANG
-else
-    CC="cc"
-    CXX="c++"
-fi
+FIND_CLANG
```
obastemur: why this change?

mike-kaufman: including pal.h causes a bunch of errors around `__declspec` syntax, so you need `-fms_extensions` flag passed to clang.

obastemur: well, if we remove this, we won't be able to test gcc/g++ compiling a test app. You don't need to include PAL.h to use that. define the extern inside the C/xx file and linker will find it for you.

test/native-tests/test-pal/sample.cpp
mike-kaufman: fixed now.

test/native-tests/test_native.sh
mike-kaufman: fixed now.

test/native-tests/test_native.sh
```diff
@@ -1,3 +1,5 @@
+#!/bin/bash
+
```
obastemur: this breaks the Copyright check (CI)

mike-kaufman:
Closing, will re-open though chakra-hub.

PR #3549, opened 2017-08-18
fixing bug in wcssstr

mike-kaufman:
Fixing https://github.com/Microsoft/ChakraCore/issues/3434..  Bug in wcsstr would return a match if the string being searched ended with a prefix of the search string.  Fixed this to behave correctly.  Also added tests to test_native.sh to verify this.

This is re-opened from PR #3542.

### DIFF
--- a/pal/src/cruntime/wchar.cpp
+++ b/pal/src/cruntime/wchar.cpp
@@ -1493,16 +1493,22 @@ PAL_wcsstr(
         i = 0;
         while (1)
         {
-            if (*(string + i) == 0 || *(strCharSet + i) == 0)
+            if (*(strCharSet + i) == 0)
             {
                 ret = (char16_t *) string;
+                goto leave;
+            }
+            else if (*(string + i) == 0)
+            {
+                ret = NULL;
                 goto leave;
             }
             if (*(string + i) != *(strCharSet + i))
             {
                 break;
             }
-        i++;
+
+            i++;
         }
         string++;
     }
--- a/test/native-tests/test-pal/sample.cpp
+++ b/test/native-tests/test-pal/sample.cpp
@@ -1,0 +1,68 @@
+//-------------------------------------------------------------------------------------------------------
+// Copyright (C) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE.txt file in the project root for full license information.
+//-------------------------------------------------------------------------------------------------------
+
+extern "C"  int
+__cdecl
+PAL_wcscmp(
+       const char16_t *string1,
+       const char16_t *string2);
+
+extern "C" const char16_t *
+__cdecl
+PAL_wcsstr(
+       const char16_t *string,
+       const char16_t *strCharSet);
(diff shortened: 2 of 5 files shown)


test/native-tests/test-pal/sample.cpp
```diff
@@ -0,0 +1,65 @@
...
+    result += validate(PAL_wcsstr(u"targetstringxxx", u"searchStringxxx"), NULL, u"test1");
+    result += validate(PAL_wcsstr(u"abcdefg", u"def"), u"defg", u"test2");
+    result += validate(PAL_wcsstr(u"abcdefg", u"a"), u"abcdefg", u"test3");
+    result += validate(PAL_wcsstr(u"abcdefg", u"g"), u"g", u"test4");
+    result += validate(PAL_wcsstr(u"abcdefg", u"gh"), NULL, u"test5");
+    result += validate(PAL_wcsstr(u"abcdefg", u"ah"), NULL, u"test6");
+    result += validate(PAL_wcsstr((const char16_t *)NULL, (const char16_t *)NULL), NULL, u"test7");
+    result += validate(PAL_wcsstr((const char16_t *)NULL, u""), NULL, u"test8");
+    result += validate(PAL_wcsstr(u"", (const char16_t *)NULL), NULL, u"test9");
+    result += validate(PAL_wcsstr(u"abcdefg", u""), u"abcdefg", u"test10");
```
digitalinfinity: These are good- i would also add cases for "abcdefg", "abcdefg", "", "" and "", "def"

mike-kaufman: fixed

mike-kaufman: the `"", ""` case actually uncovered another bug, where we would return `null` instead of the input string.  This was inconsistent with win32 API behavior, so I fixed that as well.

pal/src/cruntime/wchar.cpp
```diff
@@ -1482,9 +1482,8 @@ PAL_wcsstr(
         goto leave;
     }
 
-    if (*strCharSet == 0)
-    {
-        ret = (char16_t *)string;
+    if (*strCharSet == 0) {
```
MSLaguana: This looks like an unintentional change, the curly brace should start on the line after the `if`

mike-kaufman: fixed...

test/native-tests/test-python/helloWorld.py
```diff
@@ -13,7 +13,7 @@
     platform = "so"
 
 build_type = sys.argv[1]
-if sys.argv[2] != None:
+if sys.argv.__len__() > 2 and sys.argv[2] != None:
```
MSLaguana: Is this change required?

mike-kaufman: I believe so.  I was getting failures when running the tests that used this about teh index being out of bounds.  Adding this checked made that error go away.

PR #3553, opened 2017-08-19
Experimental: Introduce "lite" ChakraCore build

curtisman:
Initial attempt to produce a ChakraCore build with a small binary footprint, trading off with performance and some functionality.

Lite build currently will disable
- JIT,
- Script debugging
- Valid pointer table in the recycler
- Intl
- ETW

Add support to build without script debug
Add support to not use precomputed valid pointer table saving memory/binary size.

Disable page heap support in recycler for release build
Disable unused script profiling code in ChakraCore

Also fix the OOPJIT IDL to include CommonDefines.h

On Windows, current lite build is about 50% of the normal build on x86 and 51% on x64
x86
  Core 5516288 bytes
  Lite 2765312 bytes

x64
  Core 7315968 bytes
  Lite 3742720 bytes


obastemur:
Shall we send another PR to enable 'light' mode testing on CI first? Otherwise these changes won't be tested.

curtisman:
@obastemur I am thinking that we can just switch the DisableJIT daily run to test the "Lite" mode instead.
@digitalinfinity @dilijev What do you guys think?

lib/Runtime/Language/AsmJs.cpp
```diff
@@ -580,10 +580,12 @@ namespace Js
                 {
                 case MathBuiltin::Function:{
                     auto func = mathBuiltin.u.func;
+#ifdef ENABLE_SCRIPT_DEBUGGING
                     if (func->GetName() != nullptr)
                     {
                         OutputMessage(m.GetScriptContext(), DEIT_ASMJS_FAILED, _u("Warning: Math Builtin already defined for var %s"), func->GetName()->Psz());
                     }
+#endif
```
agarwal-sandeep: RaiseMessageToDebugger is already guarded we are missing trace due to this. I think its not needed?

curtisman: You are right. Removed.

lib/Common/CommonDefines.h
```diff
@@ -134,22 +137,32 @@
...
+
 
 #ifdef _WIN32
 #define SYSINFO_IMAGE_BASE_AVAILABLE 1
 #define ENABLE_CONCURRENT_GC 1
 #define ENABLE_ALLOCATIONS_DURING_CONCURRENT_SWEEP 1 // Only takes effect when ENABLE_CONCURRENT_GC is enabled.
 #define ENABLE_ALLOCATIONS_DURING_CONCURRENT_SWEEP_USE_SLIST 1 // Use Interlocked SLIST for allocableHeapBlockList
 #define SUPPORT_WIN32_SLIST 1
+#ifndef CHAKRACORE_LITE
 #define ENABLE_JS_ETW                               // ETW support
```
digitalinfinity: How costly is this? I would imagine that we'd want ETW based profiling available even in lite-mode so curious if we could have at least just the rundown events emitted here?

digitalinfinity:
@curtisman I'm in favor of replacing DisableJIT with this mode.

lib/Common/CommonDefines.h
curtisman: I would like the "lite" mode to be as bare bone as possible.  
It is easy for the user to reenable these piece meal
It would also allow us to test builds with these feature disabled as well.

dilijev:
@digitalinfinity @curtisman I agree that we can at least partially replace CI for DisableJIT with CI for ChakraCore-Lite (lol branding). If we replace DisableJIT testing with Lite, since Lite disables Intl as well, there will be no testing for the nojit variant of Intl bytecode. I think it's possible some folks will want to run DisableJIT but still have Intl, so we should continue to test that. Maybe we can have a small set of flavors for each rather than completely replacing all 9 Windows flavors.

Since this is a feature of release binary size, we probably want at least one *_release build, and for testing purposes we would want at least one *_debug build. Think a *_test build is necessary -- there are a handful (currently 10) of `exclude_chk` tests, mostly to do with being too slow to run under *_debug but fast enough under *_test builds. All but 2 of those tests are also tagged Slow. There are many more `exclude_fre` tests (176)?

I'm assuming all of the portions of ChakraCore that are disabled for Lite can be disabled individually. We probably don't need to CI all of those options individually, as there might not be significant advantage, but is that something we would want to test in any capacity?

BTW, do we share the 32b bytecode between ARM and x86? Would an ARM build be useful CI for Lite? Do arm_test builds actually provide any value considering we don't run tests on ARM CI?

curtisman:
@dilijev I don't think we need to test DisableJIT by itself.   The chance that DisableJIT with Intl break while Lite doesn't break is pretty small.  And if it is broken, it shouldn't be too hard to fix?

Also Lite mode doesn't have embedded byte code (because Intl is disabled)

dilijev:
I am worried mainly about keeping at least one CI config with JIT disabled and Intl enabled because Intl has specific bytecode for the DisableJIT config.

```
lib/Runtime/Library/InJavascript/Intl.js.bc.32b.h
lib/Runtime/Library/InJavascript/Intl.js.bc.64b.h
lib/Runtime/Library/InJavascript/Intl.js.nojit.bc.32b.h
lib/Runtime/Library/InJavascript/Intl.js.nojit.bc.64b.h
```

If it is not updated when the rest of the bytecode is updated, it will be broken.

We should keep at least one Intl-enabled DisableJIT build with tests (probably x64_debug).

PR #3560, opened 2017-08-21
WASM: Spec fixes + improve debugging

Cellule:
Fix the remaining spec test failures.
Improve debugging output with `-trace:wasmbytecode` and improve some error messages.

Bugs fixed:
- Disallow invalid utf8 characters in wasm strings
- Disallow `tee_local` of unreachable values
- Must have an `else` block if the `if` is yielding a value
- All the targets of `br_table` must have the same signature
- If `br_if` is yielding `any` type, then push the block's result type to the stack instead. ie: the any type now has an explicit type for further type checking

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/3560)


lib/WasmReader/WasmBinaryReader.cpp
```diff
@@ -1033,6 +1021,14 @@ const char16* WasmBinaryReader::ReadInlineName(uint32& length, uint32& nameLengt
         AssertMsg(UNREACHED, "We calculated the length before decoding, what happened ?");
         ThrowDecodingError(_u("Error while decoding utf8 string"));
     }
+    for (size_t i = 0; i < decodedLength; ++i)
+    {
+        const char16 c = contents[i];
+        if (!utf8::IsValidWideChar(c) || c == UNICODE_UNKNOWN_CHAR_MARK)
+        {
+            ThrowDecodingError(_u("Invalid UTF-8 encoding"));
```
Penguinwizzard: Does this work as intended (and as the error message says) with the whole range of Unicode strings? Additionally, should utf8::DecodeUnitsIntoAndNullTerminate be able to generate invalid char16s? Additionally, is this sort of twobyte-by-twobyte validation possible? The docs seem to imply no, as the official validation involves surrogate pairs.

lib/WasmReader/WasmByteCodeGenerator.cpp
```diff
@@ -1086,33 +1181,40 @@ void WasmBytecodeGenerator::EmitBrTable()
     EmitInfo scrutineeInfo = PopEvalStack(WasmTypes::I32, _u("br_table expression must be of type i32"));
 
     m_writer->AsmReg2(Js::OpCodeAsmJs::BeginSwitch_Int, scrutineeInfo.location, scrutineeInfo.location);
-    EmitInfo yieldInfo;
-    if (ShouldYieldToBlock(defaultEntry))
+    EmitInfo yieldvalue;
```
Penguinwizzard: nit: casing

lib/WasmReader/WasmBinaryReader.cpp
Penguinwizzard: Tagging @dilijev for Unicode wisdom.

dilijev: Doing a char16 at a time means only surrogate pairs would be invalid.  utf8::DecodeUnitsIntoAndNullTerminate should not be able to generate anything invalid (although that API should be the test surface for such a claim), but given invalid utf8 input, would produce � FFFD http://www.fileformat.info/info/unicode/char/0fffd/index.htm (which I'm assuming is this `UNICODE_UNKNOWN_CHAR_MARK` being checked for here).

Cellule: At first I wanted the utf8 decoder to tell me if there was an invalid character, but I didn't find any non intrusive way to do it. Which is why I am checking afterward.
From what @dilijev says, I should only need to check for `UNICODE_UNKNOWN_CHAR_MARK` to see if there was any invalid characters right ?

Since strings in the wasm buffer are relatively rare (only import, export, custom section and debugging info) and small, I don't think doing another pass on the buffer will make a big difference.

dilijev: It sounds like you're saying this possibly-redundant check wouldn't really hurt performance? If so, you can leave it in. Otherwise, I believe it is correct so say that if you are looking for the presence of invalid characters in the utf-8 input buffer by looking at the utf-16 output, then looking for `UNICODE_UNKNOWN_CHAR_MARK` is the only thing that is necessary.

dilijev:
FWIW Revision LGTM

PR #3566, opened 2017-08-22
Disable fixed field and equivalent type spec code from DISABLE_JIT builds

curtisman:
Put equivalent type spec and property guard code under ENABLE_NATIVE_CODEGEN
Also fixed ability to disable SUPPORT_FIXED_FIELDS_ON_PATH_TYPES
Disable fixed field for when JIT is disabled.

Reduced binary size of the lite build by ~41K on x86 and ~65K on x64.


digitalinfinity:
Rubber-stamp LGTM if CI is green

PR #3571, opened 2017-08-22
Globopt Instr string config flag

Cellule:
Add option to disable printing Globopt instr strings post lower.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/3571)

### DIFF
--- a/lib/Backend/IR.cpp
+++ b/lib/Backend/IR.cpp
@@ -4209,7 +4209,7 @@ Instr::DumpByteCodeOffset()
 void
 Instr::DumpGlobOptInstrString()
 {
-    if(this->globOptInstrString)
+    if(this->globOptInstrString && Js::Configuration::Global.flags.PrintGlobOptInstrString)
     {
         Output::Print(_u("\n\n GLOBOPT INSTR: %s\n\n"), this->globOptInstrString);
     }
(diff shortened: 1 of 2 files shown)


Cellule:
@meg-gupta Yeah, when trying to look at the asm code we generate, the globopt instr are in the way and make it hard to wrap your head around the asm code

agarwal-sandeep:
For most cases the globopt string is useful. I think instead of enabling this using a switch you should not print if the switch is passed so that others don't need to pass a new switch and for asm you can pass the switch to not print the string.

thomasmo:
looks like the default behavior is a no-op, so no objections here

Cellule:
> For most cases the globopt string is useful. I think instead of enabling this using a switch you should not print if the switch is passed so that others don't need to pass a new switch and for asm you can pass the switch to not print the string.

@agarwal-sandeep: like @thomasmo said, I haven't changed the default. If you pass `-PrintGlobOptInstrString-` now it won't print them anymore

agarwal-sandeep:
@Cellule thanks. I missed looking at the default true value.

LouisLaf:
Let's use a -off: switch for this.

PR #3574, opened 2017-08-23
Remove unused ScanFlagSuppressIdPid mode, and refactor Ident::Tk to replace HashTbl::TkFromNameLen

curtisman:


lib/Parser/Hash.cpp
```diff
@@ -168,22 +206,8 @@ tokens Ident::Tk(bool isStrictMode)
...
-        #include "kwds_sw.h"
-
-        #define KEYWORD(tk,f,prec2,nop2,prec1,nop1,name) \
-            LEqual_##name: \
-                if (cch == g_ssym_##name.cch && \
-                        0 == memcmp(g_ssym_##name.sz, prgch, cch * sizeof(OLECHAR))) \
-                { \
-                    if (f) \
-                        this->m_grfid |= f; \
-                    this->m_tk = tk; \
```
dilijev:
```
-                    if (f) \
 -                        this->m_grfid |= f; \
 -                    this->m_tk = tk; \
```

Were not present in `HashTbl::TkFromNameLen`. Is the difference in behavior significant?

curtisman: Nope.   Just addition information that get passed out.
That is debug only usage, and we have a wrapper that will throw out those value.

PR #3577, opened 2017-08-24
Remove additional edges in try finally flowgraph transformation

meg-gupta:
Remove additional edges in try finally flowgraph transformation.

  We used to add edge from infinite loops in finally region to end of finally block, so that unreachable block elimination does not remove finally end block.
  We didnt need this edge for globopt. But had one issue that came up with not adding these edges in building loops/processing basic blocks in flowgraphs.
  Because we don't cascade edge removal on calling RemoveBlock after early exit flowgraph transformation.

  This PR avoids adding this additional edge. And adds an extra call to RemoveUnreachableBlocks before FindLoops, so that unreachable blocks are correctly removed before loops are built.

  Fixes asserts in OS#13182223 and OS#12792486.

### DIFF
--- a/lib/Backend/SccLiveness.cpp
+++ b/lib/Backend/SccLiveness.cpp
@@ -165,8 +165,6 @@ SCCLiveness::Build()
                 if (labelInstr->isOpHelper == (this->lastOpHelperLabel != nullptr)
                     && lastLabelInstr && labelInstr->isOpHelper == lastLabelInstr->isOpHelper)
                 {
-                    // No such transition. Remove the label.
-                    Assert(!labelInstr->GetRegion() || labelInstr->GetRegion() == this->curRegion);
                     labelInstr->Remove();
                     continue;
                 }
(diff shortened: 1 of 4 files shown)


lib/Backend/SccLiveness.cpp
```diff
@@ -165,8 +165,6 @@ SCCLiveness::Build()
                 if (labelInstr->isOpHelper == (this->lastOpHelperLabel != nullptr)
                     && lastLabelInstr && labelInstr->isOpHelper == lastLabelInstr->isOpHelper)
                 {
-                    // No such transition. Remove the label.
-                    Assert(!labelInstr->GetRegion() || labelInstr->GetRegion() == this->curRegion);
                     labelInstr->Remove();
```
meg-gupta: Removing this assert. It is possible that a block containing a label gets assigned a region and then dead coded due to unreachable block elimination. With try finally having early exit, blocks can become unreachable later on after first pass of region assignment is done. See test6() in test/EH/tryfinallytests.js

meg-gupta:
This PR removes the additional edges as suggested by @LouisLaf  in https://github.com/Microsoft/ChakraCore/pull/3117.

LouisLaf:
Looks much better! :grinning: 

---
In reply to: [324500978](https://github.com/Microsoft/ChakraCore/pull/3577#issuecomment-324500978) [](ancestors = 324500978)

PR #3578, opened 2017-08-24
Dont update valueInfo of bytecode constants

meg-gupta:


lib/Backend/GlobOptFields.cpp
```diff
@@ -3220,6 +3220,12 @@ GlobOpt::UpdateObjPtrValueType(IR::Opnd * opnd, IR::Instr * instr)
         return;
     }
 
+    ValueInfo *objValueInfo = objVal->GetValueInfo();
```
rajatd: if something is a bytecode constant, shouldn't the valueType be definite (and the check above pass)?

meg-gupta: As discussed offline, added comment below.

meg-gupta:
@rajatd @LouisLaf thanks for looking!

PR #3580, opened 2017-08-24
Remove all code dealing with the unused CallPut

curtisman:


pleath:
What will happen if someone writes f() = 1?

curtisman:
@pleath You already made it an early error in TH2 :)

pleath:
That change had to be reverted for compat reasons. f() = 1 is a runtime error today. Not as I would have it, but there it is.

curtisman:
I believe it always emitted as a runtime error and all the code that support the actually "CallPut" is never emitted any more.

curtisman:
@pleath gave the green light offline.

PR #3583, opened 2017-08-24
Allow cross-site objects too in ActivationObjectEx::Is

rajatd:


lib/Runtime/Types/ActivationObject.h
```diff
@@ -210,7 +210,8 @@ namespace Js
 
         static bool Is(void* instance)
         {
-            return VirtualTableInfo<Js::ActivationObjectEx>::HasVirtualTable(instance);
+            return VirtualTableInfo<ActivationObjectEx>::HasVirtualTable(instance) ||
+                VirtualTableInfo<CrossSiteObject<ActivationObjectEx>>::HasVirtualTable(instance);
```
akroshg: can we create a test for this?

akroshg: Aneesh has test case.

---
In reply to: [135153348](https://github.com/Microsoft/ChakraCore/pull/3583#discussion_r135153348) [](ancestors = 135153348)

rajatd: @aneeshdk , could you share the test case with me?

PR #3587, opened 2017-08-25
Setting internal properties with "enumerable" attribute off.

### LINKED ISSUE #1622: PathTypeHandlerBase::SetPrototype changes hasNoEnumerableProperties of newPrototype
In PathTypeHandler.cpp line 1470 has the call below which eventually ends up modifying the hasNoEnumerableProperties flag of the newPrototype object due to an "internal property" being added to the newPrototype object.

```
newPrototype->SetInternalProperty(Js::InternalPropertyIds::TypeOfPrototypeObjectDictionary, (Var)oldTypeToPromotedTypeMap, PropertyOperationFlags::PropertyOperation_Force, nullptr);
```

If the newPrototype object has hasNoEnumerableProperties as true, then this will call down through SimpleDictionaryTypeHandler.cpp line 1325 which adds the internal property with the default set of attribute flags:

```
return this->AddProperty(instance, propertyRecord, value, PropertyDynamicTypeDefaults, info, flags, SideEffects_Any);
```

This will then reach line 2609 which checks and sets hasNoEnumerableProperties as false:

```
if (attributes & PropertyEnumerable)
{
    instance->SetHasNoEnumerableProperties(false);
}
```

rajatd:
While we don't enumerate internal properties, setting an internal property as enumerable unnecessarily sets the hasNoEnumerableProperties flag to false on the type handler. This may have perf impact as this flag is used to skip prototypes when enumerating properties. 

Changing this to set internal properties as non-enumerable from the beginning. Doing this for non-path type handlers only because, 
1. An object with a path type handler should theoretically have hasNoEnumerableProperties set to false anyway.
2. Setting a non-enumerable property on an object with a path type handler will convert its type handler to a (simple) dictionary type handler.

Fixes #1622

### DIFF
--- a/lib/Runtime/Types/DeferredTypeHandler.h
+++ b/lib/Runtime/Types/DeferredTypeHandler.h
@@ -329,10 +329,7 @@ namespace Js
             return TRUE;
         }
 
-        DynamicTypeHandler * typeHandler = GetCurrentTypeHandler(instance);
-        return typeHandler->IsPathTypeHandler() ?
-            typeHandler->SetProperty(instance, propertyId, value, flags, nullptr) :
-            typeHandler->SetPropertyWithAttributes(instance, propertyId, value, PropertyWritable & PropertyConfigurable, nullptr, flags);
+        return GetCurrentTypeHandler(instance)->SetInternalProperty(instance, propertyId, value, flags);
     }
 
     template <DeferredTypeInitializer initializer, typename DeferredTypeFilter, bool isPrototypeTemplate, uint16 _inlineSlotCapacity, uint16 _offsetOfInlineSlots>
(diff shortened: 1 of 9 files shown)


lib/Runtime/Types/DeferredTypeHandler.h
```diff
@@ -321,6 +322,20 @@ namespace Js
...
     template <DeferredTypeInitializer initializer, typename DeferredTypeFilter, bool isPrototypeTemplate, uint16 _inlineSlotCapacity, uint16 _offsetOfInlineSlots>
+    BOOL DeferredTypeHandler<initializer, DeferredTypeFilter, isPrototypeTemplate, _inlineSlotCapacity, _offsetOfInlineSlots>::SetInternalProperty(DynamicObject* instance, PropertyId propertyId, Var value, PropertyOperationFlags flags)
+    {
+        if (!EnsureObjectReady(instance, DeferredInitializeMode_Set))
+        {
+            return TRUE;
+        }
+
+        DynamicTypeHandler * typeHandler = GetCurrentTypeHandler(instance);
+        return typeHandler->IsPathTypeHandler() ?
```
curtisman: Just call typeHandler->SetInternalProperty here?

rajatd:
Oh, that's right. Thanks!

Get Outlook for iOS<https://aka.ms/o0ukef>

PR #3588, opened 2017-08-25
Keep AutoString::length in sync w/ AutoString::data

Krovatkin:
This fix makes sure that `AutoString::length` and `AutoString::data` are in sync.

### DIFF
--- a/bin/ch/stdafx.h
+++ b/bin/ch/stdafx.h
@@ -201,7 +201,6 @@ public:
             strValue = value;
         }
         int strLen = 0;
-        size_t writtenLen = 0;
         size_t actualLen = 0;
         if (errorCode == JsNoError)
         {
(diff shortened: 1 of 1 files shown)


bin/ch/stdafx.h
```diff
@@ -268,7 +267,7 @@ class AutoString
 
     size_t GetLength()
```
Krovatkin: `AutoString::GetLength` is used only in `Debugger.cpp` (namely, https://github.com/Microsoft/ChakraCore/blob/master/bin/ch/Debugger.cpp#L497). If ChakraFull doesn't have any additional dependencies we could do away with `AutoString::GetLength`

Krovatkin: @Cellule what do you think?

obastemur:
This supposed to cache string length so we don't measure it again and again.. Well, this is not perf critical part of the solution. 

LGTM. Yet, could we do it while caching the `length` ?

Krovatkin:
@obastemur I pondered a bit on the caching aspect. Unfortunately, this design created *enough* complexity to produce a bug. Moreover, some uses in `WScriptJsrt.cpp`  **still** call `strlen`. All the uses of `AutoString` seem to be contained in `WScriptJsrt.cpp` and `Debugger.cpp` and neither one seems to be perf critical. Lastly, users of `AutoString` should be able to cache the result of `GetLength` in most cases in case we ever hit a perf issue. 

Anyways, I'll do whatever you guys tell me to. 

fyi @arunetm

Cellule:
@Krovatkin First of all it won't affect ChakraFull since this file is in ch.exe only
I found the source of the bug.
in commit 356656c26eeec2b74b24abb2f727f23ca3d6d595 (aka #3433) the length was cached from the `ChakraRTInterface::JsCopyString` call in `AutoString::Initialize`. Instead now it is using a local variable instead thus breaking the initial contract of AutoString.

I think we have to use `AutoString::length` instead of `writtenLength` in `AutoString::Initialize` (both CopyString calls) and should be the extent of the fix.

Thoughts ?

Krovatkin:
@Cellule, done

Krovatkin:
@dotnet-bot test Ubuntu static_ubuntu_linux_release please

Cellule:
Thanks @Krovatkin, I will merge it for you

Krovatkin:
@Cellule thank you for your help!

PR #3590, opened 2017-08-26
TTD: Fix leak on top-level function load and bug in external function inflate.

mrkmarron:
- Fix leak of top level scripts by cleaning un-needed references.
- Update inflate code for external functions to ensure snap/inflate result in same heap.

### DIFF
--- a/lib/Runtime/Debug/TTEventLog.cpp
+++ b/lib/Runtime/Debug/TTEventLog.cpp
@@ -385,7 +385,7 @@ namespace TTD
         }
     }
 
-    SnapShot* EventLog::DoSnapshotExtract_Helper(double gcTime)
+    SnapShot* EventLog::DoSnapshotExtract_Helper(double gcTime, JsUtil::BaseHashSet<Js::FunctionBody*, HeapAllocator>& liveTopLevelBodies)
     {
         SnapShot* snap = nullptr;
 
@@ -397,7 +397,7 @@ namespace TTD
         //Phase 2: Evacuate marked objects
         //Allows for parallel execute and evacuate (in conjunction with later refactoring)
 
-        this->m_snapExtractor.EvacuateMarkedIntoSnapshot(this->m_threadContext);
+        this->m_snapExtractor.EvacuateMarkedIntoSnapshot(this->m_threadContext, liveTopLevelBodies);
 
         ///////////////////////////
         //Phase 3: Complete and return snapshot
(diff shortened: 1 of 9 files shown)


lib/Runtime/Debug/TTEventLog.cpp
```diff
@@ -1256,7 +1262,14 @@ namespace TTD
         //Create the event object and add it to the log
         NSLogEvents::SnapshotEventLogEntry* snapEvent = this->RecordGetInitializedEvent_DataOnly<NSLogEvents::SnapshotEventLogEntry, NSLogEvents::EventKind::SnapshotTag>();
         snapEvent->RestoreTimestamp = this->GetLastEventTime();
-        snapEvent->Snap = this->DoSnapshotExtract_Helper((endTime - startTime) / 1000.0);
+
+        JsUtil::BaseHashSet<Js::FunctionBody*, HeapAllocator> liveTopLevelBodies(&HeapAllocator::Instance);
```
mike-kaufman: these few lines are repeated here & above.  Should they be refactored into a method?

lib/Runtime/Debug/TTRuntmeInfoTracker.cpp
```diff
@@ -419,13 +442,33 @@ namespace TTD
         }
     }
 
+    void ScriptContextTTD::ProcessFunctionBodyOnUnLoad(Js::FunctionBody* body, Js::FunctionBody* parent)
+    {
+        //if this is a root (parent is null) then put this in the rootbody pin set so it isn't reclaimed on us
+        if(parent == nullptr)
+        {
+            TTDAssert(this->m_ttdPinnedRootFunctionSet->Contains(body), "We already added this function!!!");
+            this->m_ttdPinnedRootFunctionSet->Remove(body);
```
mike-kaufman: not clear why you're removing this after the assert here.  Should this be a fail-fast condition?

lib/Runtime/Debug/TTEventLog.cpp
```diff
@@ -429,7 +429,13 @@ namespace TTD
             this->SetSnapshotOrInflateInProgress(true);
             this->PushMode(TTDMode::ExcludedExecutionTTAction);
 
-            snap = this->DoSnapshotExtract_Helper(0.0);
+            JsUtil::BaseHashSet<Js::FunctionBody*, HeapAllocator> liveTopLevelBodies(&HeapAllocator::Instance);
```
mike-kaufman: are the `JsUtil::` classes chakra-internal collection classes?  Or are they types exposed through the JS runtime to users' processes?

mrkmarron: The are chakra-internal collections.

lib/Runtime/Debug/TTRuntmeInfoTracker.cpp
mrkmarron: TTDAssert is a fail-fast assert.

lib/Runtime/Debug/TTEventLog.cpp
mrkmarron: Personally I didn't see a clean API/call signature for refactoring around this so preferred a bit of duplication to (perhaps) forced code sharing.

mike-kaufman:
@mrkmarron - can you add tests? :)

mrkmarron:
Hi Mike, existing tests cover the correctness issues here.

PR #3594, opened 2017-08-28
Byte code size and serialized byte code size optimizations (15-20% serialized byte code size reduction)

curtisman:
Byte code size reductions:
- Variable slot index size in layout

Serialized byte code size reductions:
- Move has loopHeaderArray and has asmJsInfo into the bit field (instead of a byte)
- Don't emit field that has non-zero default value.
- Use the defined fields to indicate existing of more field to reduce serialized byte code size
- Use constant type to indicate whether it is property string or not instead of an extra boolean
- Add int8 and int16 constant type for small int encoding in the constant table
- Separate out the byte offset from the character offset in the LineOffsetCache so that we can omit them if the source is purely ASCII-7
    -Reduce by half the memory used by the LineOffsetCache for ASCII-7 sources, and it reduces the serialized byte code size as well.
- Remove unused sourceSpans from byte code serialization.
- Serialize the bit flag with constant encoding (since the bit flag always as a large value)

Also:
- Clean up unused Reg2WithICIndex and REg1Int2 layout
- Move LineOffsetCache to Runtime/Base instead of Common/DataStructures, and de-templatize it as we only use Recycler as the allocator anyway.


lib/Runtime/Language/InterpreterStackFrame.cpp
```diff
@@ -8889,7 +8889,7 @@ const byte * InterpreterStackFrame::OP_ProfiledLoopBodyStart(const byte * ip)
         return JavascriptOperators::OP_LdModuleSlot(playout->SlotIndex1, playout->SlotIndex2, scriptContext);
     }
 
-    inline void InterpreterStackFrame::OP_StModuleSlot(Var instance, int32 slotIndex1, int32 slotIndex2, Var value)
+    inline void InterpreterStackFrame::OP_StModuleSlot(Var instance, uint32 slotIndex1, uint32 slotIndex2, Var value)
```
dilijev: >Var instance [](start = 55, length = 12)

Unused?

lib/Common/DataStructures/LineOffsetCache.h
```diff
@@ -1,269 +0,0 @@
-//-------------------------------------------------------------------------------------------------------
-// Copyright (C) Microsoft. All rights reserved.
-// Licensed under the MIT license. See LICENSE.txt file in the project root for full license information.
-//-------------------------------------------------------------------------------------------------------
-#pragma once
-
-namespace JsUtil
-{
-    template <typename TAllocator>
-    class LineOffsetCache
```
dilijev: >LineOffsetCache [](start = 10, length = 15)

Any changes to this code besides splitting into .h and .cpp under lib/runtime/base and removing the JsUtil namespace?

dilijev:
> #undef DECLARE_SERIALIZABLE_ACCESSOR_FIELD

Move `#undef DECLARE_SERIALIZABLE_ACCESSOR_FIELD_NO_CHECK` after this line?

---
Refers to: lib/Runtime/SerializableFunctionFields.h:112 in 31cc8d6. [](commit_id = 31cc8d6191d153168ced309366c449b06741b449, deletion_comment = False)

lib/Runtime/ByteCode/ByteCodeSerializer.cpp
```diff
@@ -64,7 +64,7 @@ namespace Js
     const int magicStartOfPropIdsOfFormals = *(int*)"pif[";
     const int magicEndOfPropIdsOfFormals = *(int*)"]pif";
     const int magicStartOfSlotIdToNestedIndexArray = *(int*)"sni[";
-    const int magicEndOfSlotIdToNestedIndexArray = *(int*)"]sni"
+    const int magicEndOfSlotIdToNestedIndexArray = *(int*)"]sni";
```
dilijev: >; [](start = 64, length = 1)

wait... how was this not an error?

lib/Runtime/ByteCode/ByteCodeSerializer.cpp
```diff
@@ -123,15 +123,22 @@ enum FileVersionScheme : byte
...
     bool has_m_columnNumber: 1;
+    bool has_attributes : 1;
     bool has_m_nestedCount: 1;
+    bool has_loopHeaderArray : 1;
+    bool has_asmJsInfo : 1;
+    bool has_auxiliary : 1;
+    bool has_propertyIdOfFormals: 1;
     bool has_slotIdInCachedScopeToNestedIndexArray : 1;
+    bool has_debuggerScopeSlotArray : 1;
+
```
dilijev: nit: blank line

lib/Runtime/ByteCode/ByteCodeSerializer.cpp
```diff
@@ -232,7 +239,10 @@ enum ConstantType : byte
     ctStrictNullDisplay = 7,
     ctTrue = 8,
     ctFalse = 9,
-    ctStringTemplateCallsite = 10
+    ctStringTemplateCallsite = 10,
+    ctPropertyString16 = 11,
+    ctInt16 = 12,
+    ctInt8 = 13,
```
dilijev: why not group the three int size constant types?

lib/Runtime/ByteCode/ByteCodeSerializer.cpp
```diff
@@ -1784,45 +1813,35 @@ class ByteCodeBufferBuilder
         return sizeof(serialization_alignment TStructType);
     }
 
-    uint32 AddPropertyIdOfFormals(BufferBuilderList & builder, FunctionBody * function)
+    uint32 AddPropertyIdOfFormals(BufferBuilderList & builder, PropertyIdArray * propIds, FunctionBody * function)
```
dilijev: >propIds [](start = 81, length = 7)

`Assert(propIds != nullptr)` ?

lib/Runtime/ByteCode/ByteCodeSerializer.cpp
```diff
@@ -2365,6 +2413,11 @@ class ByteCodeBufferReader
         return ReadConstantSizedInt32(buffer, remainingBytes, value);
     }
 
+    const byte * ReadConstantSizedUInt32(const byte * buffer, uint * value)
+    {
+        return ReadConstantSizedInt32(buffer, (int *)value);
```
dilijev: >Int32 [](start = 32, length = 5)

mismatch of ReadUInt32 (name) and ReadInt32 (returned)

lib/Common/DataStructures/LineOffsetCache.h
curtisman: Yes.  Splitting the data from charOffset/byteOffset pair to separate list, so that the byteOffset don't need to be stored if it is ASCII-7 (which charOffset == byteOffset)

lib/Runtime/ByteCode/ByteCodeSerializer.cpp
curtisman: The code is only enabled explicitly to debug issues.  No official build has this code enabled.

lib/Runtime/ByteCode/ByteCodeSerializer.cpp
curtisman: The uint return is in out param.

lib/Runtime/ByteCode/ByteCodeSerializer.cpp
curtisman: I guess since the byte code is going to be updated already, I might as well?

lib/Runtime/Language/InterpreterStackFrame.cpp
curtisman: Yeah.. it is borrowing the SET_ELEM_ENVSLOTNonVar in the interpreter dispatch, which to clean up will require more change.  I will do that separately.

lib/Runtime/ByteCode/ByteCodeSerializer.cpp
curtisman: Sigh... I will have to renumber them all

lib/Runtime/ByteCode/ByteCodeSerializer.cpp
curtisman: Doesn't feel like the assert serves much value.....

lib/Runtime/ByteCode/ByteCodeSerializer.cpp
curtisman: removed

PR #3604, opened 2017-08-29
perf: reduce collision in hot maps

obastemur:
Both maps were starting with 4 capacity (powerOf2) and were suffering
from high collision rates.

BuiltInLibraryFunctionMap (~120 items +) hence allocating 131 prime capacity
helps both with startup time and collision management. Prime number
choice is due to hash type. (16 / aligned address)

Property name map is set to 10% init rate of actual property map and
updated to prime size policy from powerOf2. (uses same hash function with property map)

Prime MOD was previously very expensive (comparing to powerOf2) https://github.com/Microsoft/ChakraCore/pull/3516 intends to change that.

### DIFF
--- a/lib/Runtime/Base/ScriptContext.cpp
+++ b/lib/Runtime/Base/ScriptContext.cpp
@@ -4706,7 +4706,7 @@ void ScriptContext::RegisterPrototypeChainEnsuredToHaveOnlyWritableDataPropertie
             {
                 Assert(this->recycler);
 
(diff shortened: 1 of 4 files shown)


obastemur:
> Change looks fine to me as long as it improves perf.

I wouldn't be surprised if this would improve perf solid 1 to 2% on AcmeAir due to large property map.

Thanks for the review.

PR #3605, opened 2017-08-29
perf: use wmemcpy instead of memcpy for wide strings

obastemur:


obastemur:
@MSLaguana Thanks for the review.

Issue #2713, opened 2017-03-20
The `length` property of numerous built-ins are not configurable but should be


bterlson:
A partial list of intrinsics with non-configurable lengths:
* Every built-in constructor
* `async function foo () {}`

akroshg:
closing this issue

PR #3606, opened 2017-08-29
Making length property on built-in constructor to be configurable

akroshg:


akroshg:
Fixes #2713.
@tcare @aneeshdk please review. (test262 run is pending, I'll update here once they are done)

akroshg:
fyi... 112 more test262 tests passed with this change,

PR #3607, opened 2017-08-29
Adding a new enumerator flag to support short-term accesses

MSLaguana:
JSON::Stringify uses enumerators to extract all properties from an object.
By default this enumerator will attempt to populate read/write caches, but
when stringifying an object we are unlikely to access these values again.

### DIFF
--- a/lib/Runtime/Library/JSON.cpp
+++ b/lib/Runtime/Library/JSON.cpp
@@ -638,7 +638,7 @@ namespace JSON
             {
                 uint32 precisePropertyCount = 0;
                 Js::JavascriptStaticEnumerator enumerator;
-                if (object->GetEnumerator(&enumerator, EnumeratorFlags::SnapShotSemantics, scriptContext))
+                if (object->GetEnumerator(&enumerator, EnumeratorFlags::SnapShotSemantics | EnumeratorFlags::EphemeralReference, scriptContext))
                 {
                     bool isPrecise;
                     uint32 propertyCount = GetPropertyCount(object, &enumerator, &isPrecise);
--- a/lib/Runtime/Library/JSON.cpp
+++ b/lib/Runtime/Library/JSON.cpp
@@ -433,9 +433,20 @@ namespace JSON
         // can end up being serialized. Well, that is the ES5 spec word.
         //if(!Js::RecyclableObject::FromVar(holder)->GetType()->GetProperty(holder, keyId, &value))
 
-        if(!Js::JavascriptOperators::GetProperty(Js::RecyclableObject::FromVar(holder),keyId, &value, scriptContext))
+        if (VirtualTableInfo<Js::PropertyString>::HasVirtualTable(key))
         {
-            return scriptContext->GetLibrary()->GetUndefined();
+            PropertyValueInfo info;
+            Js::PropertyString* propertyString = (Js::PropertyString*)key;
+            PropertyValueInfo::SetCacheInfo(&info, propertyString, propertyString->GetLdElemInlineCache(), false);
+            CacheOperators::TryGetProperty<true, false, true, false, true, false, false, true, false>(holder, false, Js::RecyclableObject::FromVar(holder), keyId, &value, scriptContext, nullptr, &info);
+        }
+
+        if (value == nullptr)
+        {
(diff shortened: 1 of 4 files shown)


MSLaguana:
On AcmeAir, this was a ~ 1% perf gain.

curtisman:
Hmm.. interesting.  

JSON does require it to get the value of the property, which is done in StringifySession::Str.
I am wondering whether it is faster to continue to cache and use the cached value in StringifySession::Str (instead of avoiding caching)?

MSLaguana:
I had a quick look at this, changing `StringifySession::Str`  to look like

```cpp
        PropertyValueInfo info;
        PropertyString* propertyString = scriptContext->GetPropertyString(keyId);
        PropertyValueInfo::SetCacheInfo(&info, propertyString, propertyString->GetLdElemInlineCache(), false);
        if (!CacheOperators::TryGetProperty<true, false, true, false, true, true, false, true, false>(holder, false, Js::RecyclableObject::FromVar(holder), keyId, &value, scriptContext, nullptr, &info))
        {
            if (!Js::JavascriptOperators::GetProperty(Js::RecyclableObject::FromVar(holder), keyId, &value, scriptContext))
            {
            }
        }
        return StrHelper(key, value, holder);
```
and removing the other changes in this PR. From what I see so far it doesn't look like it has much effect.

Would you recommend a different approach, or a different tuning of the template parameters?

curtisman:
Sorry. That's not what I meant.  

From the enumerator (without your change), it should have returned a PropertyString, which contains a inline cache that is populated.   Then in StringifySession::Str if the string that got passed in is already a PropertyString, use the inline cache in there to do the look up, which should be able to do the fast path because the info is a match.

MSLaguana:
That's the piece that I was missing :) Adding a conversion from the `Js::JavascriptString` to the `Js::PropertyString` and checking the cache based on that looks like it is improving perf at least a similar amount to disabling caching; benchmark is still running and I'll see how it looks afterwards.

MSLaguana:
Looking at the benchmark results, it seems that there is a slight improvement if we use the cache:

```cpp
if (VirtualTableInfo<Js::PropertyString>::HasVirtualTable(key))
{
  PropertyValueInfo info;
  Js::PropertyString* propertyString = (Js::PropertyString*)key;
  PropertyValueInfo::SetCacheInfo(&info, propertyString, propertyString->GetLdElemInlineCache(), false);
  CacheOperators::TryGetProperty<true, false, true, false, true, false, false, true, false>(holder, false, Js::RecyclableObject::FromVar(holder), keyId, &value, scriptContext, nullptr, &info);
}

if (value == nullptr)
{
  if(!Js::JavascriptOperators::GetProperty(Js::RecyclableObject::FromVar(holder),keyId, &value, scriptContext))
  {
  }
}
return StrHelper(key, value, holder);
```

however not as big a gain as if we avoid caching entirely here (Using caching gave a ~0.2% gain, disabling caching gave ~0.6%). Based on that, unless you have more suggestions for how to tweak the use of caching, I still suggest we go with this PR as it stands and disable caching for JSON property enumerators.

obastemur:
>  but when stringifying an object we are unlikely to access these values again

@MSLaguana could you please elaborate on this?

MSLaguana:
My thinking is that once you `JSON.Stringify(foo)`, you are fairly likely to not do very much with `foo` afterwards; typically I believe you will either parse the stringified result as a kind of "deep copy", or you send the JSON string elsewhere and the original object is probably stale.

It's true that we do touch the values twice in the course of stringifying the object (once enumerating the properties, and then once to actually fetch the property) but from the experiments I've done that doesn't seem to give a large benefit. I also found that when we attempt to use the cache there is something like a 10% cache miss rate, even though we just enumerated the properties.

MikeHolman:
If the cache isn’t getting you anything, I’m ok disabling it. One last thing I’d try is changing to only cache in the ldFld cache for your case.

MSLaguana:
Having it only cache loads as well as using the cache gave similar performance to removing the cache entirely, the difference seemed to be ~noise, but since it makes more sense I'll go with that approach. If the tests are happy, I'll check this in.

PR #3608, opened 2017-08-29
Add a flag to use a stronger array sort for bug analysis.

Penguinwizzard:
The default JS sort often results in an implementation-defined order.
This is not generally a problem, and helps improve our perf. However,
there are cases where our ordering depends on the internal structure,
not the user-visible structure, of the array. This change adds a flag
to make this ordering more consistent, which should help identify our
bugs where the issue is just our sort orders being inconsistent.


lib/Runtime/Library/JavascriptArray.cpp
```diff
@@ -6702,7 +6708,42 @@ namespace Js
         Assert(element1 != NULL);
         Assert(element2 != NULL);
 
+#ifndef DBG
```
akroshg: >DBG [](start = 8, length = 3)

why not do in the enable_debug_config_option?

lib/Runtime/Library/JavascriptArray.cpp
```diff
@@ -6533,7 +6533,13 @@ namespace Js
         TryFinally([&]()
         {
             //The array is a continuous array if there is only one segment
-            if (startSeg->next == nullptr) // Single segment fast path
+            if (startSeg->next == nullptr
+                // If this flag is specified, we want to improve the consistency of our array sorts
+                // by removing missing values from all kinds of arrays before sorting (done here by
+                // using the copy-to-one-segment path for array sorts) and by using a stronger sort
+                // comparer than the spec requires (done in CompareElements).
+                && !Js::Configuration::Global.flags.StrongArraySort
```
akroshg: can you remind me where will this flag be enabled?

lib/Runtime/Library/JavascriptArray.cpp
Penguinwizzard: I've changed it to this.

lib/Runtime/Library/JavascriptArray.cpp
Penguinwizzard: This flag is intended to be used when looking at a bug; if it's something where repro/not repro differs on -forcearraybtree, this flag can identify if it's the known issue of multiple search orders being equivalent, or if it's a new, more interesting issue.

lib/Runtime/Library/JavascriptArray.cpp
```diff
@@ -6702,7 +6710,42 @@ namespace Js
         Assert(element1 != NULL);
         Assert(element2 != NULL);
 
+#ifndef ENABLE_DEBUG_CONFIG_OPTIONS
```
Cellule: This is not needed, since your flag has a default value of `false` the compiler will remove the else branch

Penguinwizzard: I've change to use CONFIG_FLAG such that this holds, and have removed the #ifndefs.

lib/Runtime/Library/JavascriptArray.cpp
```diff
@@ -6702,7 +6708,38 @@ namespace Js
         Assert(element1 != NULL);
         Assert(element2 != NULL);
 
-        return JavascriptString::strcmp(element1->StringValue, element2->StringValue);
+        if (CONFIG_FLAG(StrongArraySort))
```
Cellule: Shouldn't this be `!CONFIG_FLAG(StrongArraySort)` ?

Penguinwizzard: Indeed, fixed.

Cellule:
:shipit:

PR #3609, opened 2017-08-29
Changing utf8 conversion codex to improve perf and safety

MSLaguana:
With this change the utf8Codex becomes aware of the size of the buffer it
is given to write into, and will assert if the buffer it too small.

utf8Codex now also supports just counting the size that a utf8 string
would be without needing a buffer to write to.

To make use of this, the Jsrt JsCopyString method has been modified so that
it can be used to extract a string with only one allocation, down from ~3.

This has a ~1.3% perf gain in node-chakracore acmeair


MSLaguana:

boingoing:
LGTM 👍🏼

PR #3610, opened 2017-08-29
[MSFT 13209141]: Defer default expressions in a formals list if the function containing them is going to be deferred. 

pleath:
We know that the expressions can't be evaluated until the function has been fully compiled, so compiling the expressions up front is wasted work. And generating code for a function nested inside a deferred function creates inconsistencies.


pleath:

pleath:
(Reviewed offline by @aneeshdk .)

PR #3611, opened 2017-08-29
Use dynamic dispatch via function pointer field to call the initializer in the deferred type handler

curtisman:
Reduce core binary size by 7% (~561K) in x64 and 6% (~329K) in x86.
(Lite build is reduced by 454K/264K respectively)
(Chakra.dll is reduced by 1.25M/627K respectively)

### DIFF
--- a/lib/Runtime/Types/DeferredTypeHandler.h
+++ b/lib/Runtime/Types/DeferredTypeHandler.h
@@ -74,7 +74,7 @@ namespace Js
         DEFINE_GETCPPNAME();
 
     private:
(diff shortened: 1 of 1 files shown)


curtisman:
the size reduction is because of ICF, as the code for most of these templates becomes the same, and able to fold into a single definition.

@dilijev  There is a small cause of dynamic dispatch (especially with CFG), but most of the initializer are only call once per library.  The only one that might be called more are for function objects, and it didn't have any measurable impact on a local micro benchmark.

Theoretically, the initializer can be a constructor param now.  The only benefit is that compiled obj size is smaller (because of less template expansion.  It should make not difference after it is linked.
I keep it as is because:
1. Less code churn.
2. Retain the option to tune it by changing a selective number of deferred type handler back to static dispatch if we found some that are perf sensitive.

PR #3613, opened 2017-08-30
Check the return value of the CreateNamedFunction and return appropriately

akroshg:


akroshg:
@jianchun please have a look at this one.

jianchun:
LGTM. (Is there an internal bug number that you could put somewhere in the commit message?)

akroshg:
Yes I do - I'll put in the commit message.

PR #3615, opened 2017-08-30
[MSFT 12988607] Don't bind name references outside the function when completing the function expression scope. 

pleath:
References to outer bindings may be left on the stack if we're reparsing an arrow function's formals.


pleath:
@aneeshdk do you mind looking at this one-liner? (Thanks.)

PR #3621, opened 2017-08-31
Replace ArrayBuffer.transfer for ArrayBuffer.detach

Cellule:
Remove ArrayBuffer.transfer feature and replace it with a test function to only detach the ArrayBuffer.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/3621)


lib/Runtime/Library/ArrayBuffer.cpp
```diff
@@ -429,22 +429,15 @@ namespace Js
...
-        if (args.Info.Count >= 3)
-        {
-            newBufferLength = ToIndex(args[2], JSERR_ArrayLengthConstructIncorrect, scriptContext, MaxArrayBufferLength);
-
-            // ToIndex above can call user script (valueOf) which can detach the buffer
-            if (arrayBuffer->IsDetached())
-            {
-                JavascriptError::ThrowTypeError(scriptContext, JSERR_DetachedTypedArray, _u("ArrayBuffer.transfer"));
-            }
+            JavascriptError::ThrowTypeError(scriptContext, JSERR_DetachedTypedArray, _u("ArrayBuffer.prototype.detach"));
```
akroshg: nit: this should be ArrayBuffer.detach as you have not added this method to prototype, you have added to constructor.

Cellule: Good catch. Initially I put it on the prototype and then decided to keep it similar to what ArrayBuffer.transfer had.

lib/Runtime/Library/ArrayBuffer.h
```diff
@@ -142,15 +142,15 @@ namespace Js
...
+            static FunctionInfo Detach;
         };
 
         static Var NewInstance(RecyclableObject* function, CallInfo callInfo, ...);
         static Var EntrySlice(RecyclableObject* function, CallInfo callInfo, ...);
         static Var EntryIsView(RecyclableObject* function, CallInfo callInfo, ...);
         static Var EntryGetterByteLength(RecyclableObject* function, CallInfo callInfo, ...);
         static Var EntryGetterSymbolSpecies(RecyclableObject* function, CallInfo callInfo, ...);
-        static Var EntryTransfer(RecyclableObject* function, CallInfo callInfo, ...);
+        static Var EntryDetach(RecyclableObject* function, CallInfo callInfo, ...);
```
akroshg: nit: why not add this function in ENABLE_CONFIGOPTION only? same is true with EntryInfo::Detach

Cellule: I guess you're right. I was on the fence about it, but for no good reason.

Penguinwizzard:
This change should fix #2733 (as it removes the broken test, as well as the code that was causing the leak).

PR #3625, opened 2017-09-01
A few random cleanups.

LouisLaf:
Add assert to dbcheck to ensure sym type consistency.
Some better structure packing.  Make RegNum a BYTE instead of an int32.

Free srcs/dst during instr->Remove(). This reduces memory allocated in the func arean by about %2...

### DIFF
--- a/lib/Backend/IR.cpp
+++ b/lib/Backend/IR.cpp
@@ -359,7 +359,6 @@ Instr::Free()
                         return;
                     }
                     Assert(this->m_func->GetTopFunc()->allowRemoveBailOutArgInstr || !stackSym->m_isBailOutReferenced);
-                    stackSym->m_instrDef = nullptr;
                 }
                 else
                 {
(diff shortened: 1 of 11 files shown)


lib/Backend/IR.cpp
```diff
@@ -373,6 +372,15 @@ Instr::Free()
...
+        this->FreeDst();
+    }
+    if (this->GetSrc1())
+    {
+        this->FreeSrc1();
+        if (this->GetSrc2())
+        {
+            this->FreeSrc2();
+        }
     }
```
agarwal-sandeep: Can there be a case where src1 is not present but src2 is present? either an assert or moving src2 check outside will help?

LouisLaf: It should not happen normally, but I found instances where we unlink src1 and then call instr->Remove()...  I added to fix this.

PR #3635, opened 2017-09-03
Numbers up to 10^21 shouldn't be displayed exponentially (Fixes #2751)

### LINKED ISSUE #2751: `1e+19.toString(2)` is displayed exponentially, but should be written out
The output given by ChakraCore is `1.00010101100011100100011000001001000100111101(e+63)`, but in three other engines, the correct answer of `1000101011000111001000110000010010001001111010000000000000000000` is provided. The [spec](https://tc39.github.io/ecma262/#sec-tostring-applied-to-the-number-type) states that with *n* (the base-10 exponent) between 0 and 21, the written out version should be used, rather than the exponential version.

xiaoyinl:
The spec section 7.1.12.1 requires that if a number's base 10 logarithm is <= 21, then it shouldn't use exponential notation when it's converted to String. The original code checks if the base 2 logarithm of the number is <= 60.

This fixes #2751.

### DIFF
--- a/lib/Common/Common/NumberUtilities_strtod.cpp
+++ b/lib/Common/Common/NumberUtilities_strtod.cpp
@@ -2528,7 +2528,7 @@ BOOL Js::NumberUtilities::FNonZeroFiniteDblToStr(double dbl, _In_range_(2, 36) i
     maxOutDigits = g_rgcchSig[radix];
     __analysis_assume(maxOutDigits > 0);
 
-    if (wExp2 < -60 || wExp2 > 60)
+    if (dbl < 1e-21 || dbl > 1e+21)
     {
         // Use exponential notation. Get the exponent and normalize.
         if (cbitDigit != 0)
--- a/test/Number/toString_3.baseline
+++ b/test/Number/toString_3.baseline
@@ -116,10 +116,10 @@ n.toPrecision(20):  0.0000000000000000000
 Test: 1e21
 n.toString():  1e+21
 n.toString(10):  1e+21
-n.toString(8):  1.5432711533427365(e+23)
-n.toString(2):  1.101100011010111001001101011011100010111011110101(e+69)
-n.toString(16):  3.635c9adc5dea(e+17)
-n.toString(25):  1.1l259oooooof(e+15)
+n.toString(8):  154327115334273650000000
+n.toString(2):  1101100011010111001001101011011100010111011110101000000000000000000000
+n.toString(16):  3635c9adc5dea00000
+n.toString(25):  11l259ooooooo5ie
 n.toFixed():  1e+21
 n.toFixed(0):  1e+21
 n.toFixed(2):  1e+21


lib/Common/Common/NumberUtilities_strtod.cpp
```diff
@@ -2525,10 +2525,11 @@ BOOL Js::NumberUtilities::FNonZeroFiniteDblToStr(double dbl, _In_range_(2, 36) i
     // powers of 2, but not for other radixes.
     // REVIEW : round?
     wExp2 = (int)((Js::NumberUtilities::LuHiDbl(dbl) & 0x7FF00000) >> 20) - 0x03FF;
+    wExp10 = (int)floor(log10(dbl));
     maxOutDigits = g_rgcchSig[radix];
     __analysis_assume(maxOutDigits > 0);
 
-    if (wExp2 < -60 || wExp2 > 60)
+    if (wExp10 < -21 || wExp10 > 21)
```
obastemur: Exp2 is already calculated above. Could we use that for comparison instead of calculating expensive Exp10?

xiaoyinl:
@obastemur I changed it to compare `dbl` with 1e+21 directly, since Exp10 isn't used elsewhere. Also #2751 has further discussion on what to do for numbers larger than 10^21. I'll update the code if ChakraCore wants to follow the behavior of other engines.

obastemur:
@xiaoyinl could you please update the output under `test/Number/toString_3.baseline` 

```
09:11:35 Output: (at line 119)
09:11:35 ----------------------------
09:11:35 n.toString(8):  154327115334273650000000
09:11:35 ----------------------------
09:11:35 Expected Output:
09:11:35 ----------------------------
09:11:35 n.toString(8):  1.5432711533427365(e+23)
09:11:35 ----------------------------
09:11:35 exit code: 0
```

MSLaguana:
Isn't the exponential form correct in that case?

obastemur:
> Isn't the exponential form correct in that case?

Correct.

obastemur:
@xiaoyinl @MSLaguana test case should be updated to `154327115334273650000000` ..

The reason is, we ask for `1e21`.toString(8) So, the output we see there is <= 1e21. So the output after this PR is the correct output.

xiaoyinl:
@obastemur @MSLaguana Will do. The remaining question is what to do if the number is > 1e21. @littledan says other engines never switch to exponential notation for radix not 10, regardless of how large it is. Do you want ChakraCore to adopt this behavior?

obastemur:
At this point; if tests pass, squashing commits could be nice

dilijev:
Tests pass. I'll adopt this and merge it in after it passed tests internally.

I have added some follow-up tasks:
* Augment the test with edge cases (the points at which it switches between written-out and exponential): https://github.com/Microsoft/ChakraCore/issues/3740
* Adopt the behavior of the other engines to never use exponential notation for non-radix-10: https://github.com/Microsoft/ChakraCore/issues/3739

Issue #3592, opened 2017-08-26
Add API to store private properties in any object


fatcerberus:
It's possible right now to create an "External Object" which stores a C pointer to internal native data.  However, what's *not* possible as far as I can see is being able to store arbitrary properties on an object which refer to JS values but aren't accessible from JS.  For example in Duktape it's possible to:

```c
duk_push_object(ctx);
/* populate new object with anything needed */
duk_put_prop_string(ctx, -2, "\xff""hiddenObject");
```

That can be done with any object, whereas in Chakra you currently have to create an external object to store private data--and it's not possible to convert an existing standard JS object to an external one, either, which means that in native constructors for external objects, you can't use the automatic `this` and must instead return a fresh object from the constructor.

I propose an API that would either:
* Add an API that converts an existing standard JS object to an External object, **OR**:
* Add an API for storing "hidden" properties that could be accessed using `Js[GS]etProperty()` but not within JavaScript code.  Perhaps something like `JsCreateHiddenPropertyId()`.

obastemur:
I'm currently experimenting on external data support for all the object types. Upcoming days will show...

gashtio:
I'd like to add one more vote for such feature request.

The project I'm working on needs to map JS objects to C++ object and I'm experimenting moving our JS VM to ChakraCore. The external data support for all objects would be perfect. Right now I'm using code like this, setting an internal property (say, "__privdata"), which refers to an external object:

```
inline void SetUserDataChakraCoreImpl(JsValueRef object, JsPropertyIdRef prop, void* userData)
{
	bool hasUserData = false;
	CHAKRA_CALL(JsHasProperty(object, prop, &hasUserData));
	if (hasUserData)
	{
		// Overwrite value
		JsValueRef userObject = JS_INVALID_REFERENCE;
		CHAKRA_CALL(JsGetProperty(object, prop, &userObject));
		CHAKRA_CALL(JsSetExternalData(userObject, userData));
	}
	else
	{
		// Create new value
		JsValueRef userObject = JS_INVALID_REFERENCE;
		CHAKRA_CALL(JsCreateExternalObject(userData, nullptr, &userObject));
		CHAKRA_CALL(JsSetProperty(object, prop, userObject, true));
	}
}
```

That appeared to be quite expensive (mostly due to the access to the property) when I ran a profile on some performance tests that exercised going back and forth from C++ to JS. Being able to simply set external data to any object would be much more convenient (and a lot faster I suppose).

[...]

PR #3639, opened 2017-09-04
Jsrt: External data support for DynamicObject [experiment PR]

obastemur:
Adds API below to ChakraCore Jsrt;

 -  JsObjectHasExternalData
 -  JsObjectGetExternalData
 -  JsObjectSetExternalData
 -  JsObjectSetExternalDataWithCallback

Removes API from ChakraCore common Jsrt;

- JsHasExternalData
- JsGetExternalData
- JsSetExternalData

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -63,6 +63,7 @@ test/benchmarks/*.dpl
 test/benchmarks/*.txt
 testout*
 packages.config
+lib/wabt/built/config.h
 
 # CMake Files
 cmake_install.cmake
--- a/lib/Jsrt/Jsrt.cpp
+++ b/lib/Jsrt/Jsrt.cpp
@@ -2233,56 +2233,6 @@ CHAKRA_API JsStrictEquals(_In_ JsValueRef object1, _In_ JsValueRef object2, _Out
     });
 }
 
-CHAKRA_API JsHasExternalData(_In_ JsValueRef object, _Out_ bool *value)
-{
-    VALIDATE_JSREF(object);
-    PARAM_NOT_NULL(value);
-
-    BEGIN_JSRT_NO_EXCEPTION
-    {
-        *value = JsrtExternalObject::Is(object);
-    }
-    END_JSRT_NO_EXCEPTION
-}
-
-CHAKRA_API JsGetExternalData(_In_ JsValueRef object, _Out_ void **data)
-{
-    VALIDATE_JSREF(object);
-    PARAM_NOT_NULL(data);
-
-    BEGIN_JSRT_NO_EXCEPTION
-    {
-        if (JsrtExternalObject::Is(object))
-        {
-            *data = JsrtExternalObject::FromVar(object)->GetSlotData();
-        }
-        else
-        {
-            *data = nullptr;
-            RETURN_NO_EXCEPTION(JsErrorInvalidArgument);
-        }
-    }
-    END_JSRT_NO_EXCEPTION
-}
-
-CHAKRA_API JsSetExternalData(_In_ JsValueRef object, _In_opt_ void *data)
-{
-    VALIDATE_JSREF(object);
-
-    BEGIN_JSRT_NO_EXCEPTION
-    {
-        if (JsrtExternalObject::Is(object))
-        {
-            JsrtExternalObject::FromVar(object)->SetSlotData(data);
-        }
-        else
-        {
-            RETURN_NO_EXCEPTION(JsErrorInvalidArgument);
-        }
-    }
-    END_JSRT_NO_EXCEPTION
-}
-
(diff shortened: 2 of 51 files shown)


.gitignore
```diff
@@ -63,6 +63,7 @@ test/benchmarks/*.dpl
 test/benchmarks/*.txt
 testout*
 packages.config
+lib/wabt/built/config.h
```
MSLaguana: Why should this header be excluded?

lib/Jsrt/Jsrt.cpp
```diff
@@ -4764,7 +4714,7 @@ CHAKRA_API JsGetAndClearExceptionWithMetadata(_Out_ JsValueRef *metadata)
 
 CHAKRA_API JsHasOwnProperty(_In_ JsValueRef object, _In_ JsPropertyIdRef propertyId, _Out_ bool *hasOwnProperty)
 {
-    return ContextAPIWrapper<true>([&] (Js::ScriptContext *scriptContext, TTDRecorder& _actionEntryPopper) -> JsErrorCode {
+    return ContextAPIWrapper<JSRT_MAYBE_TRUE>([&] (Js::ScriptContext *scriptContext, TTDRecorder& _actionEntryPopper) -> JsErrorCode {
```
MSLaguana: Looks like this is from one of your other changes; but I don't think it's directly relevant to this one.

obastemur: @MSLaguana I saw this one was skipped and updated. Not sure I understood the problem?

fatcerberus:
Does this change allowing setting a finalizer too?

obastemur:
> Does this change allowing setting a finalizer too?

@fatcerberus Yes, `JsObjectSetExternalDataWithCallback`

liminzhu:
Looks solid 👍 . A few feedback/questions,

1. Why put the old APIs in ChakraCommonWindows.h? Either leaving them as-is or deprecating them in ChakraCore (a.k.a. they show up in ChakraFull not here) makes more sense. I tend to leave them as-is, but that's ofc up for debate.

2. Can I put ext data on any JsValueRef, or just objects?

3. I assume every time `JsObjectSetExternalDataWithCallback` is called, it will replace any previous finalizer set on the object, including the one from JsCreateExternalObejct. A following `JsObjectSetExternalData` call should set the finalizer to nothing. Am I correct?

obastemur:
@liminzhu I will put an external post to explain what is what in detail.

About JSRT; IMHO old `JsHasExternalData..` confusing and not needed.

obastemur:
Definitely strange errors from CI

```
/Users/dotnet-bot/j/workspace/Microsoft_ChakraCore/release_1.7/static_osx_osx_debug_prtest/out/Debug/../../lib/Jsrt/ChakraCore.h:300:5: note: candidate function not viable: requires 4 arguments, but 5 were provided
```

Well... It's defined under ChakraCore.h:309.. not 300!..

obastemur:
Looks like the CI issue is related to branch / base difference among the headers... Updated to latest again.

test/native-tests/test-char/sample.cpp
```diff
@@ -3,23 +3,7 @@
 // Licensed under the MIT license. See LICENSE.txt file in the project root for full license information.
 //-------------------------------------------------------------------------------------------------------
 
-#include "ChakraCore.h"
```
obastemur: this is the one causing the CI fails.. (no nullptr is defined.) will update after review.

obastemur:
Reminder; I'm just experimenting here. No promises. If your application depends on external data usage and current ChakraCore Jsrt is not helping, please let us know from https://github.com/Microsoft/ChakraCore/issues/3592 

Current version;

- External data support for non primitive JS types.
- If, needed in the future, same interface can host additional properties on a DynamicObject without increasing its initial size.
- No runtime / memory usage etc. change on anything that doesn't use external data.
- Perf cost is limited to couple of if/else call - bool function (vtable)
- Brings additional TypeHandler types. i.e. We had PathTypeHandler.. now we also have PathTypeHandlerWithExternal. The new interface is all about `HasExternalDataSupport()` virtual function is returning true. No additional logic update to original type handler.
- Centralize the usage of auxSlots and internally manages under DynamicObject
- Supports type handler sharing among the new object instances with external data.
- Added native tests to make sure things work -> https://github.com/Microsoft/ChakraCore/pull/3639/files#diff-f2ba368ddfd5ff52839d16738442a504

What else needs to be done;

- More testing!
- Reduce the amount of ConvertToExternalData calls. (latest update to PR cleared the hot paths already)

obastemur:

fatcerberus:
Without this change, is there a way right now to have a native constructor create an object with external data?  If I return a new object from a constructor directly, then subclassing it doesn't work :(

obastemur:
Was a good experiment and learned a lot. Thanks for the `thumbs up`s!

#3717 may help to implement such feature without changing the type system significantly

PR #3645, opened 2017-09-06
[MSFT 12313182] Detect assignment to const in nested functions that were originally deferred. 

pleath:
Add symbol-is-const flag to Symbol and to ScopeInfo to allow us to detect this case. Don't rely on having access to a parse node for the declaration.

### DIFF
--- a/lib/Runtime/ByteCode/ScopeInfo.h
+++ b/lib/Runtime/ByteCode/ScopeInfo.h
@@ -27,11 +27,12 @@ namespace Js {
                 Field(PropertyRecord const*) name;
             };
             Field(SymbolType) symbolType;
-            Field(bool) hasFuncAssignment;
-            Field(bool) isBlockVariable;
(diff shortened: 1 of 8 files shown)


lib/Runtime/ByteCode/ScopeInfo.h
```diff
@@ -29,6 +29,7 @@ namespace Js {
             Field(SymbolType) symbolType;
             Field(bool) hasFuncAssignment;
             Field(bool) isBlockVariable;
+            Field(bool) isConst;
             Field(bool) isFuncExpr;
             Field(bool) isModuleExportStorage;
             Field(bool) isModuleImport;
```
agarwal-sandeep: Can we make these bit fields. Saves 4 bytes on x86.

pleath: Sure.

PR #3646, opened 2017-09-06
OS#13255723, OS#13255732, OS#13255734, OS#13255735, OS#13255737: A few more uninitialized locals.

LouisLaf:
Fixed a few more benign cases.

### DIFF
--- a/lib/Backend/IntBounds.cpp
+++ b/lib/Backend/IntBounds.cpp
@@ -178,7 +178,7 @@ void IntBounds::SetBound(
 
     // Aggressively merge the constant lower or upper bound of the base value, adjusted by the offset
     ValueInfo const * const baseValueInfo = baseValue->GetValueInfo();
-    int constantBoundBase;
+    int constantBoundBase = 0xCCCCCCCC;
     const bool success =
         Lower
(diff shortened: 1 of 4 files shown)


lib/Backend/IntBounds.cpp
```diff
@@ -178,7 +178,7 @@ void IntBounds::SetBound(
 
     // Aggressively merge the constant lower or upper bound of the base value, adjusted by the offset
     ValueInfo const * const baseValueInfo = baseValue->GetValueInfo();
-    int constantBoundBase;
+    int constantBoundBase = 0xCCCCCCCC;
```
agarwal-sandeep: >0xCCCCCCCC [](start = 28, length = 10)

Why these value?

LouisLaf: I couldn't find a default that made sense for these.  0 is often a bad choice, MIN/MAX_INT were not great.  

0xCCCCCCCC has a few good properties:
o Negative
o Bad pointer
o INT 3 on x86/x64
o Easy to spot in the debugger
o That's what VC uses to initialize the stack with -RTC1

PR #3679, opened 2017-09-07
Update scanner handling of numeric literal suffixes (Fixes #3281)

### LINKED ISSUE #3281: In operator allowed to follow a number
```javascript
1337in{1337:1}
```

Expected: Syntax error. In operator cannot follow number.

Actual: true

Firefox, Chrome and Safari all throw exceptions.

tcare:
The scanner was missing logic to error out in some cases when we have forbidden characters after a numeric literal. The spec states that we cannot have an identifier start or a digit immediately following a numeric literal.

- Changed Scanner::FScanNumber to check for forbidden characters following a valid literal.
- Added an error message to convey the identifier after literal case.
- Removed the unused oFScanNumber function.
- Created a scanner test dir since we don't have one.
- Added a comprehensive test to test both the following literal and digit case (for octal and binary literals.)

### DIFF
--- a/lib/Parser/Scan.cpp
+++ b/lib/Parser/Scan.cpp
@@ -565,9 +565,25 @@ typename Scanner<EncodingPolicy>::EncodedCharPtr Scanner<EncodingPolicy>::FScanN
 {
     EncodedCharPtr last = m_pchLast;
     EncodedCharPtr pchT = nullptr;
+    bool baseSpecified = false;
     likelyInt = true;
     // Reset
     m_OctOrLeadingZeroOnLastTKNumber = false;
+
+    auto baseSpecifierCheck = [&pchT, &pdbl, p, &baseSpecified]()
+    {
+        if (pchT == p + 2)
+        {
+            // An octal token '0' was followed by a base specifier: /0[xXoObB]/
+            // This literal can no longer be a double
+            *pdbl = 0;
+            // Advance the character pointer to the base specifier
+            pchT = p + 1;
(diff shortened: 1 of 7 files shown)


lib/Parser/Scan.cpp
```diff
@@ -565,9 +565,22 @@ typename Scanner<EncodingPolicy>::EncodedCharPtr Scanner<EncodingPolicy>::FScanN
...
+    bool baseSpecified = false;
     likelyInt = true;
     // Reset
     m_OctOrLeadingZeroOnLastTKNumber = false;
+
+    auto octalCheck = [&pchT, &pdbl, p, &baseSpecified]()
+    {
+        if (pchT == p + 2)
+        {
+            // "Octal zero token "0" followed by an identifier token beginning with character 'x'/'X'
```
boingoing: This comment is a little too narrow, right?

pleath:
So is it the case that "123xyz" is always an error, but that "123 xyz" may not be? IOW, is the presence or absence of white space significant?

tcare:
Yes, this is specifically called out in the spec.

https://tc39.github.io/ecma262/#sec-literals-numeric-literals

> The SourceCharacter immediately following a NumericLiteral must not be an IdentifierStart or DecimalDigit.
> Note
> For example: 3in is an error and not the two input elements 3 and in.

tcare:
Renamed octalCheck and made the comments a bit clearer.

pleath:
Okay. I just wanted to make sure that the scanner really needed to be involved in detecting the error, but since white space is a factor I guess it does.

PR #3680, opened 2017-09-07
Intl-ICU: Add LanguageTag RegExp; make Intl.*.supportedLocalesOf work under ICU

### LINKED ISSUE #2981: Intl.js: Add RegExp to high-level test for valid language tags (according to grammar in RFC 5646 Section 2.1)
Add RegExp in Intl.js to high-level test for invalid language tags (according to grammar in RFC 5646 Section 2.1): https://tools.ietf.org/html/rfc5646#section-2.1

This work item pertains to fixing two open issues:

* i18n-library-agnostic (ILA) fix for #2604
* part of an ILA fix for #2961 (which pertains to an additional contraint on language tags as defined in ECMA 402 -- see issue for details).

No error should be thrown in the case of a grammatically correct language tag.

ILA fixes for as many issues as possible is preferred because we will have support for two different libs upon completion of #2919.

### LINKED ISSUE #3648: Intl-ICU: implement Intl.*.supportedLocalesOf
Subtask of #2919

### LINKED ISSUE #3652: Intl.js fallback: platform.getExtensions

### LINKED ISSUE #3653: Intl.js fallback: platform.resolveLocaleLookup

### LINKED ISSUE #3654: Intl.js fallback: platform.resolveLocaleBestFit

dilijev:
Resolves #2981: Intl.js: Add RegExp to high-level test for valid language tags (according to grammar in RFC 5646 Section 2.1)
Resolves #3652: Intl.js fallback: platform.getExtensions
Resolves #3653: Intl.js fallback: platform.resolveLocaleLookup
Resolves #3654: Intl.js fallback: platform.resolveLocaleBestFit
Resolves #3648: Intl-ICU: implement Intl.*.supportedLocalesOf


dilijev:
@kfarnung  0b94c00 contains the version of the LangTag RE's you already reviewed. fd0dd62 contains some fixes (mostly adding word boundaries so that matching actually works correctly).

dilijev:
@dotnet-bot test Ubuntu shared_ubuntu_linux_release please

test/Intl/Basics.js
```diff
@@ -12,11 +12,20 @@ function testRangeError(tag) {
 }
 
 function assertEachIsOneOf(actualList, expectedList, msg) {
+    if (actualList.length === 0) {
```
kunalspathak: wondering if this method checks if `actualList` contains any duplicates? Do we care about it?

lib/Runtime/Library/InJavascript/Intl.js
```diff
@@ -97,6 +97,20 @@
                 "" // another entry for fun
             ]
         }
+        // no return value
+    }
+
+    let IsWellFormedLanguageTag = function (langTag) {
+        let retVal = platform.isWellFormedLanguageTag(langTag);
+        if (retVal === null) {
+            if (!LANG_TAG_RE) {
```
kunalspathak: you have this code pattern in multiple places. Consider moving it to a function and calling it.

lib/Runtime/Library/InJavascript/Intl.js
```diff
@@ -595,16 +707,17 @@
                         // We have no other information from the platform on the cause of the exception.
                         platform.raiseOptionValueOutOfRange();
                     }
-                    if (num === 0) caseFirst = 'false';
-                    else if (num === -1) caseFirst = 'upper';
-                    else caseFirst = 'lower';
+
+                    caseFirst = (num === 0) ? 'false' : (num === -1) ? 'upper' : 'lower';
```
kunalspathak: I personally think `if-else` is more readable.

dilijev: Yeah I think you're right. I wanted to add braces but initially didn't like how long it got. I'll change it to the long form with braces.

dilijev:
@dotnet-bot test Ubuntu _no_jit_shared_ubuntu_linux_test please

test/Intl/Basics.js
dilijev: It's true that there should be no duplicates, but that wasn't the intent of this function. This test is intended to be the basics only. i.e. does each API work at all as intended. Things like no duplicates are checked in other tests.

lib/Runtime/Library/InJavascript/Intl.js
dilijev: I think the end shape of each of these cases may be different enough that adding the overhead of another call for it isn't worthwhile. (I believe that Intl code is not inlined into user code but I don't know to what extent it is internally optimized during runtime.)

dilijev: I assumed you were referring to this pattern.

```
+    let IsWellFormedLanguageTag = function (langTag) {
+        let retVal = platform.isWellFormedLanguageTag(langTag);
```

I think you might also be referring to the check-and-call pattern around `InitializeLangTagREs();`. I figured that the check inline and then calling the function if necessary was cheaper than calling the function and doing the check inside (i.e. move the checks inside and rename `InitializeLangTagREs` to `EnsureLangTagREs`). For now, I'll stand by this pattern of checking outside the init function, but I'll reconsider refactoring this in the future.
 #Pending

kunalspathak: SGTM

dilijev:

PR #3681, opened 2017-09-07
add inlining support for asm.js/wasm

### LINKED ISSUE #2794: WASM - Inlining (2w)
Add support for inlining in WebAssembly

MikeHolman:
Add support for profiling in asm.js, and profile direct internal calls.

Use temp registers rather than hard coded registers for intermediate bytecode values, so backend correctly understands that these are temps and not locals. (Also allowed me to remove the I_Conv_* opcodes.)

Tell inliner to try to inline candidate asm.js/wasm functions. Required some special handling for functions with no parameters and non-var arguments, and skipping function object bailout check.

Simplify how we emit calls in asm.js bytecode generator to evaluate arguments before emitting argouts. Prior code required complicated tracking of call depth during bytecode gen, which led to complicated call nesting (and subsequently made inlining more complicated).

Track calls for asm.js/wasm during globopt, and change inline argument tracking to use inlinee argout size rather than count for frame size calculation. (This will be needed for stackwalking, which we punted for asm.js in general... but is something we should really get around to.)

For inlining heuristics, we are more aggressive at inlining than normal JS, for the reason that bytecode is already low level, so it will blow up in size less than JS bytecode will.

Preliminary perf indicates overall improvements of about 4.5% in Unity and 1% in Jetstream

Closes #2794

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/3681)


MikeHolman:

lib/Backend/Inline.cpp
```diff
@@ -4927,7 +5028,16 @@ Inline::MapFormals(Func *inlinee,
                 }
                 break;
             }
-
+        case Js::OpCode::LdAsmJsEnv:
+            if (instr->m_func == inlinee)
+            {
+                instr->SetSrc1(funcObjOpnd);
```
rajatd: Can this never be an AddrOpnd in asmjs?

rajatd:
>         instr->m_func = this->func;

why not do this for asmjs functions?

---
Refers to: lib/Backend/LinearScan.cpp:2113 in 7baad00. [](commit_id = 7baad0033f4f5e0fb7fe0ffb2471c5a6ccb74faf, deletion_comment = False)

rajatd:
Backend changes look good to me

MikeHolman:
>         instr->m_func = this->func;

i think you're right this one i don't need. the reason for other one is we can introduce call during lowerer in an inlinee which was leaf, where the parent was also leaf, and only the inner func is updated. this leads to problems later where we end up encoding a call from a leaf func.
now though, i'm thinking probably better for LowerCall to do `SetHasCallsOnSelfAndParents` instead of `SetHasCalls`

---
In reply to: [328701050](https://github.com/Microsoft/ChakraCore/pull/3681#issuecomment-328701050) [](ancestors = 328701050)

---
Refers to: lib/Backend/LinearScan.cpp:2113 in 7baad00. [](commit_id = 7baad0033f4f5e0fb7fe0ffb2471c5a6ccb74faf, deletion_comment = False)

lib/Backend/Inline.cpp
MikeHolman: nope, it's always a RegOpnd

---
In reply to: [138230447](https://github.com/Microsoft/ChakraCore/pull/3681#discussion_r138230447) [](ancestors = 138230447)

Cellule:
Reviewed 67 of 74 files at r1, 9 of 10 files at r2.
Review status: all files reviewed at latest revision, 10 unresolved discussions, some commit checks failed.

---

*[a discussion](https://reviewable.io:443/reviews/microsoft/chakracore/3681#-KtrPjGnn3ZeL5A7ITj3:-KtrPjGo1TP2EZwSmSWV:bb6h1dj) (no related file):*
One thing I noticed, is that right now we don't detect that the various `LdAsmJsEnv` in the inlined function are not considered to be the same as the root function (if they come from the same instance aka not a call_import). 
But this sounds like an optimization that we should do in a separate change, just wanted to toss it out there.

---

*[a discussion](https://reviewable.io:443/reviews/microsoft/chakracore/3681#-KtrTAQ64y1rgb_TSoz0:-KtrTAQ64y1rgb_TSoz1:b-ee1r34) (no related file):*
Looks like the flag `-trace:wasminout` is broken with this change (ie: causes assert and/or AV)

---

*[a discussion](https://reviewable.io:443/reviews/microsoft/chakracore/3681#-KtrUSSDVqpxmij_dvin:-KtrUSSDVqpxmij_dvio:b-52b38g) (no related file):*
I found a functional bug.
We can't allow to inline imported function because we share the code for multiple instances.
for instance the following test is broken
```js
const buf = WebAssembly.wabt.convertWast2Wasm(`
(module
    (type $t1 (func (result i32)))
    (type $t2 (func (param i32) (result i32)))
    (import "test" "foo" (func $foo (type $t1)))
    (import "test" "print" (func $print (type $t2)))
    (func $f2 (export "a") (type $t2) (local f32)
      (call $print (call $foo))
    )
    (func (export "f2") (type $t1)
      (i32.const 2)
    )
)`);
const module = new WebAssembly.Module(buf);
const {exports: {a: a1, f2}} = new WebAssembly.Instance(module, {test: {foo: () => 1, print}});
const {exports: {a: a2}} = new WebAssembly.Instance(module, {test: {foo: f2, print}});
a1();
a1();
a2();
```
Without inlining, it prints `1 1 2` and with inling it prints `1 1 1`.
Right now we can't inline import thunks because it might change depending on the instance

---

*[lib/Backend/BackendOpCodeAttrAsmJs.cpp, line 17 at r2](https://reviewable.io:443/reviews/microsoft/chakracore/3681#-KtsR0GJA_67KqDAqebe:-KtsR0GJA_67KqDAqebf:b-53uvi0) ([raw file](https://github.com/microsoft/chakracore/blob/3688d1872942a2fd09ff06ce1450ce4b6c8c8418/lib/Backend/BackendOpCodeAttrAsmJs.cpp#L17)):*
> ```cpp
>         OpHasProfiled = 1 << 2,
>         OpProfiled = 1 << 3,
>         OpByteCodeOnly = 1 << 4
> ```

We seem to only set the `OpByteCodeOnly` on Profiled opcode, but don't actually use it. 
Can we remove this ?

---

*[lib/Backend/Sym.h, line 179 at r2](https://reviewable.io:443/reviews/microsoft/chakracore/3681#-KtsQUPBQPJCs3x003Uo:-KtsQUPBQPJCs3x003Up:b73zwq6) ([raw file](https://github.com/microsoft/chakracore/blob/3688d1872942a2fd09ff06ce1450ce4b6c8c8418/lib/Backend/Sym.h#L179)):*
> ```objc
>     void            SetIsFromByteCodeConstantTable() { this->m_isFromByteCodeConstantTable = true; }
>     Js::ArgSlot     GetArgSlotNum() const { Assert(HasArgSlotNum()); return m_slotNum; }
>     void            SetArgSlotNum(Js::ArgSlot newNum) { m_slotNum = newNum; }
> ```

I can't find any references to this function. Is it still needed ?

---

*[lib/Runtime/Language/DynamicProfileInfo.cpp, line 417 at r2](https://reviewable.io:443/reviews/microsoft/chakracore/3681#-KtsGDGZI3VXu_YwaGNZ:-KtsGDGZI3VXu_YwaGN_:bray9rb) ([raw file](https://github.com/microsoft/chakracore/blob/3688d1872942a2fd09ff06ce1450ce4b6c8c8418/lib/Runtime/Language/DynamicProfileInfo.cpp#L417)):*
> ```cpp
>         if (!callerBody || !callerBody->GetIsAsmjsMode() || !calleeBody || !calleeBody->GetIsAsmjsMode())
>         {
>             AssertMsg(UNREACHED, "Call to RecordAsmJsCallSiteInfo without 2 wasm function body");
> ```

nit: asm.js/wasm function body

---

*[lib/WasmReader/WasmByteCodeGenerator.cpp, line 966 at r2](https://reviewable.io:443/reviews/microsoft/chakracore/3681#-KtrhoKeNGTH4E9NH-cD:-KtrhoKeNGTH4E9NH-cE:b-my10w5) ([raw file](https://github.com/microsoft/chakracore/blob/3688d1872942a2fd09ff06ce1450ce4b6c8c8418/lib/WasmReader/WasmByteCodeGenerator.cpp#L966)):*
> ```cpp
>         WasmFunctionInfo* calleeInfo = m_module->GetWasmFunctionInfo(funcNum);
>         calleeSignature = calleeInfo->GetSignature();
>         if (!isImportCall)
> ```

change check for GetReader()->m_currentNode.call.funcType == FunctionIndexTypes::Function

---

*[lib/WasmReader/WasmByteCodeGenerator.cpp, line 1136 at r2](https://reviewable.io:443/reviews/microsoft/chakracore/3681#-KtrZ7wkbsSdnluHQjyJ:-KtrZ7wlUxTyZ8bWStG0:bkvhiw3) ([raw file](https://github.com/microsoft/chakracore/blob/3688d1872942a2fd09ff06ce1450ce4b6c8c8418/lib/WasmReader/WasmByteCodeGenerator.cpp#L1136)):*
> ```cpp
>                 throw WasmCompilationException(_u("Unknown call return type %u"), retInfo.type);
>             }
>             if (retInfo.type != WasmTypes::Void)
> ```

unnessary check since we have the same check above.

---

*[test/AsmJs/rlexe.xml, line 613 at r2](https://reviewable.io:443/reviews/microsoft/chakracore/3681#-KtrLNm5IqlwJVPE1yqq:-KtrLNm5IqlwJVPE1yqr:b4ubmv6) ([raw file](https://github.com/microsoft/chakracore/blob/3688d1872942a2fd09ff06ce1450ce4b6c8c8418/test/AsmJs/rlexe.xml#L613)):*
> ```XML
>       <files>cseBug.js</files>
>       <baseline>cseBug.baseline</baseline>
>       <compile-flags>-testtrace:asmjs -off:deferparse</compile-flags>
> ```

Why was this needed ?

---

*[test/wasm/inlining.js, line 1 at r2](https://reviewable.io:443/reviews/microsoft/chakracore/3681#-KtrKhZPP4peUjxCA8vf:-KtrKhZPP4peUjxCA8vg:b-rktt2c) ([raw file](https://github.com/microsoft/chakracore/blob/3688d1872942a2fd09ff06ce1450ce4b6c8c8418/test/wasm/inlining.js#L1)):*
> ```JavaScript
> ```

You didn't add this file to wasm/rlexe.xml

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/microsoft/chakracore/3681)*

Cellule:
Reviewed 5 of 5 files at r3.
Review status: all files reviewed at latest revision, 9 unresolved discussions, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/microsoft/chakracore/3681)*

MikeHolman:
Review status: 76 of 77 files reviewed at latest revision, 10 unresolved discussions, some commit checks failed.

---

*[a discussion](https://reviewable.io:443/reviews/microsoft/chakracore/3681#-KtrTAQ64y1rgb_TSoz0:-KtsgzRz1wFXlYsTLNHQ:b-uophqh) (no related file):*
<details><summary><i>Previously, Cellule (Michael Ferris) wrote…</i></summary><blockquote>

Looks like the flag `-trace:wasminout` is broken with this change (ie: causes assert and/or AV)
</blockquote></details>

Disabled inlining/profiling when this is enabled and seems to work now

---

*[a discussion](https://reviewable.io:443/reviews/microsoft/chakracore/3681#-KtrPjGnn3ZeL5A7ITj3:-Ktsh8krO3pEsM2o2hwf:blvruup) (no related file):*
<details><summary><i>Previously, Cellule (Michael Ferris) wrote…</i></summary><blockquote>

One thing I noticed, is that right now we don't detect that the various `LdAsmJsEnv` in the inlined function are not considered to be the same as the root function (if they come from the same instance aka not a call_import). 
But this sounds like an optimization that we should do in a separate change, just wanted to toss it out there.
</blockquote></details>

Yeah, I thought about this, but seems like something we can optimize later

---

*[lib/Backend/BackendOpCodeAttrAsmJs.cpp, line 17 at r2](https://reviewable.io:443/reviews/microsoft/chakracore/3681#-KtsR0GJA_67KqDAqebe:-KtsYqrPC-CmpW-aALn9:b-896fix) ([raw file](https://github.com/microsoft/chakracore/blob/3688d1872942a2fd09ff06ce1450ce4b6c8c8418/lib/Backend/BackendOpCodeAttrAsmJs.cpp#L17)):*
<details><summary><i>Previously, Cellule (Michael Ferris) wrote…</i></summary><blockquote>

We seem to only set the `OpByteCodeOnly` on Profiled opcode, but don't actually use it. 
Can we remove this ?
</blockquote></details>

Done.

---

*[lib/Backend/Sym.h, line 179 at r2](https://reviewable.io:443/reviews/microsoft/chakracore/3681#-KtsQUPBQPJCs3x003Uo:-KtsZ2S_ZBzRuHHV1bge:b-896fix) ([raw file](https://github.com/microsoft/chakracore/blob/3688d1872942a2fd09ff06ce1450ce4b6c8c8418/lib/Backend/Sym.h#L179)):*
<details><summary><i>Previously, Cellule (Michael Ferris) wrote…</i></summary><blockquote>

I can't find any references to this function. Is it still needed ?
</blockquote></details>

Done.

---

*[lib/Runtime/Language/DynamicProfileInfo.cpp, line 417 at r2](https://reviewable.io:443/reviews/microsoft/chakracore/3681#-KtsGDGZI3VXu_YwaGNZ:-KtsZLkeJoNZUYwlhidx:b-896fix) ([raw file](https://github.com/microsoft/chakracore/blob/3688d1872942a2fd09ff06ce1450ce4b6c8c8418/lib/Runtime/Language/DynamicProfileInfo.cpp#L417)):*
<details><summary><i>Previously, Cellule (Michael Ferris) wrote…</i></summary><blockquote>

nit: asm.js/wasm function body
</blockquote></details>

Done.

---

*[lib/WasmReader/WasmByteCodeGenerator.cpp, line 966 at r2](https://reviewable.io:443/reviews/microsoft/chakracore/3681#-KtrhoKeNGTH4E9NH-cD:-KtsZMsz5IVX2E9UGKBC:b-896fix) ([raw file](https://github.com/microsoft/chakracore/blob/3688d1872942a2fd09ff06ce1450ce4b6c8c8418/lib/WasmReader/WasmByteCodeGenerator.cpp#L966)):*
<details><summary><i>Previously, Cellule (Michael Ferris) wrote…</i></summary><blockquote>

change check for GetReader()->m_currentNode.call.funcType == FunctionIndexTypes::Function
</blockquote></details>

Done.

---

*[lib/WasmReader/WasmByteCodeGenerator.cpp, line 1136 at r2](https://reviewable.io:443/reviews/microsoft/chakracore/3681#-KtrZ7wkbsSdnluHQjyJ:-KtsZa5dwnkFESZm126J:b-896fix) ([raw file](https://github.com/microsoft/chakracore/blob/3688d1872942a2fd09ff06ce1450ce4b6c8c8418/lib/WasmReader/WasmByteCodeGenerator.cpp#L1136)):*
<details><summary><i>Previously, Cellule (Michael Ferris) wrote…</i></summary><blockquote>

unnessary check since we have the same check above.
</blockquote></details>

Done.

---

*[test/AsmJs/rlexe.xml, line 613 at r2](https://reviewable.io:443/reviews/microsoft/chakracore/3681#-KtrLNm5IqlwJVPE1yqq:-KtsZb6MLiRSUX4h6_E2:bwnzzr3) ([raw file](https://github.com/microsoft/chakracore/blob/3688d1872942a2fd09ff06ce1450ce4b6c8c8418/test/AsmJs/rlexe.xml#L613)):*
<details><summary><i>Previously, Cellule (Michael Ferris) wrote…</i></summary><blockquote>

Why was this needed ?
</blockquote></details>

something changed now that we are profiling causing different deferral between interpreter/dynapogo so trace was printing in different spot

---

*[test/wasm/inlining.js, line 1 at r2](https://reviewable.io:443/reviews/microsoft/chakracore/3681#-KtrKhZPP4peUjxCA8vf:-Kts_oKveX4SHCJ0ZqcA:b-896fix) ([raw file](https://github.com/microsoft/chakracore/blob/3688d1872942a2fd09ff06ce1450ce4b6c8c8418/test/wasm/inlining.js#L1)):*
<details><summary><i>Previously, Cellule (Michael Ferris) wrote…</i></summary><blockquote>

You didn't add this file to wasm/rlexe.xml
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/microsoft/chakracore/3681)*

Cellule:
:lgtm:

---

Reviewed 8 of 8 files at r5.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/microsoft/chakracore/3681#-:-KtvmZUcpMf-Fcfusxmf:bnfp4nl)*

PR #3682, opened 2017-09-07
Fix #2987: Per spec, skip over holes when iterating over localeList

### LINKED ISSUE #2987: Intl: CanonicalizeLocaleList incorrectly errs when holes are present in input array
https://tc39.github.io/ecma402/#sec-canonicalizelocalelist step 7.b.

Object pretending to be an array via getters, with a hole at 0:

```
eshost -h d8,sm,ch-intl-gcl,node -is Intl-gcl-weird.js
## Source
class x {
    // get 0() { return 'en'; } // culture[0] is a hole
    get 1() { return 'en'; }
    get length() { return 2; }
}
let culture = new x();
print(Intl.getCanonicalLocales(culture)[0]);

#### d8, sm, node
en

#### ch-intl-gcl
TypeError: Locale is not an object or a string
```

Which reduces to problems with arrays with holes:

```
eshost -h d8,sm,ch-intl-gcl,node -is intl-gcl-array-with-holes.js
## Source
let a = [];
a[1] = "en";
print(Intl.getCanonicalLocales(a)[0]);

#### d8, sm, node
en

#### ch-intl-gcl
TypeError: Locale is not an object or a string
```

@bterlson

dilijev:
The iteration strategy over localeList in the spec (https://tc39.github.io/ecma402/#sec-canonicalizelocalelist 7.b.) requires testing `HasProperty(O,Pk)` which can be written as `key in obj`. Correctly skipping over holes avoids errors that would arise when testing the value "undefined" that comes from reading from a non-existent key.


lib/Runtime/Library/InJavascript/Intl.js
```diff
@@ -99,11 +99,12 @@
         }
     }
 
-    var forEach = function (obj, length, func) {
-        var current = 0;
-
+    var forEachIfPresent = function (obj, length, func) {
```
akroshg: curious, why not use Array.prototype.forEach ?

dilijev: This was the pattern I inherited. Might be a slight difference between `Array.prototype.forEach` and the spec iteration. Might be because of surface area for tainting (would need to use `platform.ArrayInstanceForEach`). Might be a bit of both.

dilijev:
Squashed and pushed e905806

PR #3684, opened 2017-09-07
Fix broken config parsing in ChakraFull

digitalinfinity:
ChakraFull had ifdef-ed the value of FINISHED. When the config parsing code was changed to use fget*c instead of fwscanf, the ifdef should have been removed. Not doing so caused us to fill in EOFs in the config buffer causing us to incorrectly parse the config file.

This change removes the old version of FINISHED and switches to using EOF or WEOF depending on the platform.


Cellule:
:shipit:

PR #3686, opened 2017-09-07
OS13108399 Fix TVS warning in GetDir function

akroshg:
GetDir function used to pass the drive array whose size _splitpath_s did not like. Fixed that


akroshg:
@suwc @jianchun please have a look

akroshg:
I have manually pushed this changes.

Issue #3603, opened 2017-08-29
Null pointer dereference when Array.prototype.concat.call


DaramG:
This is a PoC code.
```
var p = new Proxy([], {});
class MyArray extends Array {
    static get [Symbol.species]() {
        return function() { return p; }
    };
}
size = 0xffffffff;
w = new MyArray(size);
x = Array.prototype.concat.call(w);

akroshg:
thanks @DaramG for reporting this. I can verify there is indeed a null pointer dereference when we convert to big index in the concat code path.

akroshg:
fixed.

PR #3688, opened 2017-09-07
Concat method should pass the pDestObj when we take the concat path using bigindex

akroshg:


akroshg:
fixes #3603.
It takes lot of time of test finish the test so I haven't added that.
@suwc @aneeshdk please have a look at this one.

lib/Runtime/Library/JavascriptArray.cpp
```diff
@@ -3161,7 +3161,7 @@ namespace Js
                         // in EntryConcat like we do with Arrays because a getProperty on an object Length
                         // is observable. The result is we have to check for overflows separately for
                         // spreadable objects and promote to a bigger index type when we find them.
-                        ConcatArgs<BigIndex>(pDestArray, remoteTypeIds, args, scriptContext, idxArg, idxDest, /*firstPromotedItemIsSpreadable*/true, length);
+                        ConcatArgs<BigIndex>(pDestObj, remoteTypeIds, args, scriptContext, idxArg, idxDest, /*firstPromotedItemIsSpreadable*/true, length);
```
suwc: >pDestObj [](start = 45, length = 8)

Should we check pDestArray and send pDestObj down the generic path down below instead?

PR #3689, opened 2017-09-07
Refactoring JavascriptFunction::CallFunction callsites

rajatd:
Refactoring a bunch of JavascriptFunction::CallFunction callsites to use CALL_FUNCTION macro. I'm refactoring the callsites where the number of arguments is known at compile time.

### DIFF
--- a/lib/Runtime/Debug/DiagStackFrame.cpp
+++ b/lib/Runtime/Debug/DiagStackFrame.cpp
@@ -359,8 +359,7 @@ namespace Js
             varThis = scriptContext->GetLibrary()->GetNull();
         }
 
-        Js::Arguments args(1, (Js::Var*) &varThis);
-        varResult = pfuncScript->CallFunction(args);
+        varResult = CALL_FUNCTION(pfuncScript->GetScriptContext()->GetThreadContext(), pfuncScript, CallInfo(1), varThis);
 
         debugManager->UpdateConsoleScope(dummyObject, scriptContext);
 
--- a/lib/Runtime/Library/JavascriptProxy.cpp
+++ b/lib/Runtime/Library/JavascriptProxy.cpp
@@ -228,20 +228,15 @@ namespace Js
         }
 
         PropertyId propertyId = getPropertyId();
-        CallInfo callInfo(CallFlags_Value, 3);
-        Var varArgs[3];
-        Js::Arguments arguments(callInfo, varArgs);
-        varArgs[0] = handlerObj;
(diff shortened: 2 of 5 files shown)


lib/Runtime/Debug/DiagStackFrame.cpp
```diff
@@ -359,8 +359,7 @@ namespace Js
             varThis = scriptContext->GetLibrary()->GetNull();
         }
 
-        Js::Arguments args(1, (Js::Var*) &varThis);
-        varResult = pfuncScript->CallFunction(args);
+        varResult = CALL_FUNCTION(pfuncScript->GetScriptContext()->GetThreadContext(), pfuncScript, CallInfo(1), varThis);
```
agarwal-sandeep: >CALL_FUNCTION [](start = 20, length = 13)

Noticed that CallFunction does stack probe but CALL_FUNCTION macro doesn't, will it be safe to not have probe stack in all new code?

rajatd: I believe so, since the number of arguments are well known and small

lib/Runtime/Library/JavascriptProxy.cpp
```diff
@@ -228,20 +228,15 @@ namespace Js
...
-        varArgs[2] = GetName(requestContext, propertyId);
+        Var propertyName = GetName(requestContext, propertyId);
 
-        Assert(JavascriptString::Is(varArgs[2]) || JavascriptSymbol::Is(varArgs[2]));
+        Assert(JavascriptString::Is(propertyName) || JavascriptSymbol::Is(propertyName));
         //8. Let trapResultObj be the result of calling the[[Call]] internal method of trap with handler as the this value and a new List containing target and P.
         //9. ReturnIfAbrupt(trapResultObj).
         //10. If Type(trapResultObj) is neither Object nor Undefined, then throw a TypeError exception.
 
         Js::ImplicitCallFlags saveImplicitCallFlags = threadContext->GetImplicitCallFlags();
```
agarwal-sandeep: >Js::ImplicitCallFlags saveImplicitCallFlags = threadContext->GetImplicitCallFlags(); [](start = 8, length = 84)

Isn't it better to have all such calls in this file use ExecuteImplicitCall instead? we don't want to allow this call if disableImplicitCall is set?

rajatd: Yes, that was going to be my next PR :)

PR #3693, opened 2017-09-08
Implement an additional bitflag to the debugger to distinguish variables still in TDZ

### LINKED ISSUE #3690: Variables in TDZ reported as strings in debugger
Minor cosmetic issue, block-scoped variables still in TDZ are reported as having a string value in the debugger:
```
@/scripts/main.js:37 game()
(ssj) v
pig: number = 812
cow: string = it eats cats
ape: string = [Uninitialized block variable]
dayNight: string = [Uninitialized block variable]
session: string = [Uninitialized block variable]
```

I think it would be cleaner if these were reported as `TDZ` or `uninitialized` instead of `string`.

fatcerberus:
When examining local variables through the JSON debug API, variables which are still in the temporal dead zone now get an additional bitflag in their propertyAttributes, `IN_TDZ`, to allow the host to quickly distinguish them from variables already in play.

Fixes #3690 (in a sense).

### DIFF
--- a/lib/Jsrt/ChakraDebug.h
+++ b/lib/Jsrt/ChakraDebug.h
@@ -414,6 +414,7 @@ typedef unsigned __int32 uint32_t;
     ///         NONE = 0x1,
     ///         HAVE_CHILDRENS = 0x2,
     ///         READ_ONLY_VALUE = 0x4,
+    ///         IN_TDZ = 0x8,
     ///     </para>
     ///     <para>
     ///     {
--- a/lib/Jsrt/JsrtDebugUtils.cpp
+++ b/lib/Jsrt/JsrtDebugUtils.cpp
@@ -145,6 +145,7 @@ void JsrtDebugUtils::AddPropertyType(Js::DynamicObject * object, Js::IDiagObject
     bool addValue = false;
 
     Js::Var varValue = objectDisplayRef->GetVarValue(FALSE);
+    Js::IDiagObjectAddress* varAddress = objectDisplayRef->GetDiagAddress();
 
     if (varValue != nullptr)
     {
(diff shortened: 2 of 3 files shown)


agarwal-sandeep:
@fatcerberus In all cases the type property in the JSON represents one of JavaScript primitive types (undefined/function/Boolean/number/string/symbol/object) and the value debugger is giving is infact a string ("[Uninitialized block variable]") so I don't think that we should change the type to "tdz". I would suggest you add a new enum value to [JsrtDebugPropertyAttribute](https://github.com/Microsoft/ChakraCore/blob/36bd20a7863c3dcd08fd9cb663bdc86ff68e3beb/lib/Jsrt/JsrtDebugUtils.h#L13) and add that to [propertyAttributes](https://github.com/Microsoft/ChakraCore/blob/36bd20a7863c3dcd08fd9cb663bdc86ff68e3beb/lib/Jsrt/JsrtDebugUtils.cpp#L369) if the value is tdz.

fatcerberus:
Yes, I agree that a flag would be cleaner.  Is there a better way to determine whether a variable is in TDZ than comparing the value to the known string?

agarwal-sandeep:
@fatcerberus you can do
```C++
bool isTdz = (objectDisplayRef->GetDiagAddress() != nullptr) ? objectDisplayRef->GetDiagAddress()->IsInDeadZone() : false;
```

agarwal-sandeep:
@fatcerberus I mean do something like below [here](https://github.com/Microsoft/ChakraCore/blob/36bd20a7863c3dcd08fd9cb663bdc86ff68e3beb/lib/Jsrt/JsrtDebugUtils.cpp#L368)
```c++
if (objectDisplayRef->GetDiagAddress() != nullptr && objectDisplayRef->GetDiagAddress()->IsInDeadZone())
{
    propertyAttributes |= JsrtDebugPropertyAttribute::TDZ;
}
```

fatcerberus:
At @agarwal-sandeep's suggestion I changed the PR to use a property attribute to signify TDZ variables rather than changing the type.

agarwal-sandeep:
@fatcerberus can you also add the new enum value to [ChakraDebug.h](https://github.com/Microsoft/ChakraCore/blob/a25a575ba058d47b0bac2390b02343649f9a48f5/lib/Jsrt/ChakraDebug.h#L416) for documentation.

fatcerberus:
@agarwal-sandeep I thought I did?

Issue #3530, opened 2017-08-15
Incorrect behavior of Array.prototype.filter()


sunlili:
Hello,
The following code behaves incorrectly (inconsistent with the standard and other engines) in Chakra due to incorrect realization of HasItem() functions:
````
var get_cnt = 0;
var handler = {
        get: function(oTarget, sKey) {
		print("get " + sKey.toString());
		if (sKey.toString()=="constructor") {
			return Reflect.get(oTarget, sKey);
		} else if (Number(sKey.toString()) != NaN) {
			get_cnt++;	
			return "hahahahaha";
		}
		return Reflect.get(oTarget, sKey);
        },
	has: function (oTarget, sKey) {
		print("has " + sKey.toString());
		return Reflect.has(oTarget, sKey);
	},
};

var words1 = ["abc", "aaaaaaa", "abcd", "aaaaabcd", "abcde", "aaaaabcdef"];
var words2 = ["abc", "aaaaaaa", "abcd", "aaaaabcd", "abcde", "aaaaabcdef"];

var p = new Proxy(words1, handler);
words2.__proto__ = p;
words2.length = 10;

var longWords = words1.filter.call(words2, function(word){
  return word.length > 6;
});
print(longWords);
````

Expected result:
```
get constructor
has 6
has 7
has 8
has 9
aaaaaaa,aaaaabcd,aaaaabcdef
```
Chakra 1.7-release result:
```
get constructor
get 6
get 7
get 8
get 9
aaaaaaa,aaaaabcd,aaaaabcdef,hahahahaha,hahahahaha,hahahahaha,hahahahaha
```

ICT BT group
2017.8.15

akroshg:
Thanks @sunlili for reporting the issue.
@suwc it seems we are directly calling the DirectGetItem in the FilterHelper. which by-pass calling the has trap.
Best to confirm with HasAnyES5ArrayInPrototypeChain first and take the slower path.

PR #3694, opened 2017-09-08
Fix Issue#3530: Incorrect behavior of Array.prototype.filter()

suwc:
Fastpath of Array.prototype.filter() is missing following operation:
   b. Let kPresent be ? HasProperty(O, Pk).
Fix by adding call to JavascriptOperators::HasItem()

### DIFF
--- a/lib/Runtime/Library/JavascriptArray.cpp
+++ b/lib/Runtime/Library/JavascriptArray.cpp
@@ -9587,6 +9587,12 @@ Case0:
             Assert(length <= UINT_MAX);
             for (uint32 k = 0; k < (uint32)length; k++)
             {
+                JS_REENTRANT(jsReentLock, BOOL hasItem = JavascriptOperators::HasItem(pArr, k));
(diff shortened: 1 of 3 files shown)


suwc:
Fixes https://github.com/Microsoft/ChakraCore/issues/3530

lib/Runtime/Library/JavascriptArray.cpp
```diff
@@ -9587,6 +9587,12 @@ namespace Js
             Assert(length <= UINT_MAX);
             for (uint32 k = 0; k < (uint32)length; k++)
             {
+                JS_REENTRANT(jsReentLock, BOOL hasItem = JavascriptOperators::HasItem(pArr, k));
```
akroshg: you are making hasItem call for array case which is not required if the array is not observable. Why not check first HasAnyES5ArrayInPrototypeChain to determine whether we need to come to this path?
Also there are other places you need to consider where observable behavior can happen.

PR #3696, opened 2017-09-08
OS#10614839 Relax an assert for verifying cached value against current property value

agarwal-sandeep:
If disable implicit bit is set GetProperty returns a null value for external objects. Relax the assert to handle this.


agarwal-sandeep:
@pleath please take a look

PR #3699, opened 2017-09-10
Dont give a (backend) Value to the property sym for 'arguments' property

rajatd:
We dont want to field copy prop loads of 'arguments' property from a function object (<function>.arguments). This was achieved by marking 'arguments' property load in the interpreter as having implicit calls. However, this misses some cases when the 'arguments' load wasn't profiled. In this change I'm trying to fix the original problem in the globopt by not giving a value to the property sym for the 'arguments' property.


LouisLaf:
LGTM

PR #3702, opened 2017-09-11
Call SetInternalProperty and fix flags

sigatrev:
SetPropertyInternal was never being called, and flags were &'d instead of |'d


rajatd:
I'm pretty sure SetInternalProperty was being called. Maybe a direct call instead of through DynamicObject::SetInternalProperty. In any case, LGTM

PR #3705, opened 2017-09-11
Fix assert in byte code serializer

curtisman:
The assert that the string template array can't have property string is not true, as we create them for single char strings.


curtisman:
Duplicate PR #3704

PR #3707, opened 2017-09-12
Remove unneeded symbol that is breaking RecordInlineeFrameInfo copy prop

sigatrev:
the float sym can safely have to the original var sym as it's equivalent var sym.


rajatd:
LGTM. Make sure exprgen's happy

PR #3713, opened 2017-09-12
Refactoring some code in JavascriptProxy.cpp

rajatd:


agarwal-sandeep:
>         Var getResult;

Can you move this to place of use. Other places as well.

---
Refers to: lib/Runtime/Library/JavascriptProxy.cpp:221 in 19372a0. [](commit_id = 19372a00227acf9f759592ebf16d69002c0add68, deletion_comment = False)

PR #3714, opened 2017-09-12
OS#13641974 Asm.js internal call thunk

Cellule:
Refactor CallAsmJsFunction to offset most of the work in C++ instead of the assembly thunks

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/3714)

### DIFF
--- a/lib/Runtime/Library/amd64/JavascriptFunctionA.asm
+++ b/lib/Runtime/Library/amd64/JavascriptFunctionA.asm
@@ -219,122 +219,67 @@ ifdef _ENABLE_DYNAMIC_THUNKS
 extrn ?GetStackSizeForAsmJsUnboxing@Js@@YAHPEAVScriptFunction@1@@Z: PROC
 extrn ?GetArgsSizesArray@Js@@YAPEAIPEAVScriptFunction@1@@Z : PROC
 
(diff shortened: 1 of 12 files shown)


Cellule:
Ping

lib/Runtime/Library/amd64/JavascriptFunctionA.asm
```diff
@@ -219,198 +219,118 @@ ifdef _ENABLE_DYNAMIC_THUNKS
...
         .pushreg rsi
         push rdi
         .pushreg rdi
-        push r12
-        .pushreg r12
-        push r13
-        .pushreg r13
+        ; push r12 unused
+        ; .pushreg r12
+        ; push r13 unused
```
MikeHolman: nit: remove these comments

Cellule: Wouldn't it be better to leave it so if someone starts using r12,r13,... they can see that they're not currently saved by the callee ?

MikeHolman: i guess so, i don't really care much either way so i'm fine leaving it if you think there is value.

---
In reply to: [139541099](https://github.com/Microsoft/ChakraCore/pull/3714#discussion_r139541099) [](ancestors = 139541099)

PR #3717, opened 2017-09-13
remove old interceptors code and open a space under TypeFlagMask

obastemur:
CanHaveInterceptors was being triggered heavily on some of the benchmarks and looking deeper shown that there is no reason keeping this code around.

Besides.. Removing this, comes with a very useful space within `TypeFlagMask` that I could use for ExternalData support.

So.. unless someone comes and say that this particular mask is in use at some place (which I could not find), I will update the external data PR too.

Reminder: Both TypeFlagMask_External and TypeFlagMask_CanHaveInterceptors are set at the same place in full code. There is no place that they were being set separately.

### DIFF
--- a/lib/Runtime/Debug/DiagObjectModel.cpp
+++ b/lib/Runtime/Debug/DiagObjectModel.cpp
@@ -2141,7 +2141,7 @@ namespace Js
                     auto funcPtr = [&]()
                     {
                         IGNORE_STACKWALK_EXCEPTION(scriptContext);
-                        if (object->CanHaveInterceptors())
+                        if (object->IsExternal())
                         {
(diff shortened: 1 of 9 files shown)


obastemur:
ChakraCore code uses TypeFlagMask_External and TypeFlagMask_CanHaveInterceptors interchangeable. That's why the CI errors.

digitalinfinity:
Maybe @curtisman or @Yongqu remember how this is used...

mrkmarron:
It should be fine from a TTD perspective. My asserts were in there as a purely defensive measure.

obastemur:
@curtisman ping

obastemur:

obastemur:
Why it's waiting? This will target master branch.

obastemur:
Will rebase this to r1.8.. looks like it's going to fix CI timeouts for `misc_bugs.js`

PR #3719, opened 2017-09-13
fix JavascriptFunction::CheckValidDebugThunk CrossSite check

jianchun:
Comparing with scriptContext->CurrentCrossSiteThunk is unreliable. The
function being checked could be CrossSite and its context closed, thus
could be using a cross site thunk different to scriptContext's (note we
have 2 cross site thunks: DefaultThunk and ProfileThunk).

Change to CrossSite::IsThunk to cover both.

Fix OS#11699485


jianchun:
Thanks @akroshg

PR #3722, opened 2017-09-13
OS11907290 InitPorperty inside eval code should have configurable true

akroshg:
As per EvalDeclarationInstantiation CreateGlobalFunctionBinding will make sure to have global function to be configurable. But we didn't set that which breaks a web-page
 https://spb.tele2.ru/mobile/internet/skylink, where we throw error for such scenario. Fixed that.

### DIFF
--- a/lib/Runtime/Language/JavascriptOperators.cpp
+++ b/lib/Runtime/Language/JavascriptOperators.cpp
@@ -8057,6 +8057,15 @@ CommonNumber:
 
         Type *typeWithoutProperty = object->GetType();
 
+        if (functionBody->IsEval())
+        {
+            if (object->InitPropertyInEval(propertyId, newValue, flags, &info))
+            {
+                CacheOperators::CachePropertyWrite(object, false, typeWithoutProperty, propertyId, &info, scriptContext);
+                return;
+            }
+        }
+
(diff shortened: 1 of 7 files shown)


akroshg:
@tcare @pleath - please have a look

pleath:
This is only for indirect eval, right? The base eval case still works the way it always has?

tcare:
Please make sure to have the bug number look like OS: 123456 otherwise automation will not pick it up.

---
In reply to: [329299188](https://github.com/Microsoft/ChakraCore/pull/3722#issuecomment-329299188) [](ancestors = 329299188)

tcare:
I believe base eval already had the correct behavior. Akrosh, how does the base eval work for this case?

---
In reply to: [329303273](https://github.com/Microsoft/ChakraCore/pull/3722#issuecomment-329303273) [](ancestors = 329303273)

pleath:
It uses ScopedInitFld instead of normal InitFld, because it has to access the scope chain instead of going straight to the global object.

akroshg:
@pleath yes this is for indirect eval (the test case mentions the scenario)

tcare:
Does the normal eval case explicitly set the correct configurability?

akroshg:
@tcare yes

tcare:
Cool, good to know.

PR #3723, opened 2017-09-13
Push and pop paren expr depth counter for deferred rest and spread errors

tcare:
Fixes OS: 12896447

There exists a situation where deferred parsing can miss a deferred spread error that will get found in the full parse. This results because we get a different paren expr depth when we reparse the nested function inside the paren expr.

The fix is to push/pop the paren expr count every time we parse a new function. We know that when parsing a new function, any potential rest/spread errors we see are not going to be resolved in the outer function.

I also changed the naming of the paren count to make it clearer that it is for paren exprs only and per function.

### DIFF
--- a/lib/Parser/Parse.cpp
+++ b/lib/Parser/Parse.cpp
@@ -96,7 +96,7 @@ Parser::Parser(Js::ScriptContext* scriptContext, BOOL strictMode, PageAllocator 
     m_pCurrentAstSize = nullptr;
     m_arrayDepth = 0;
     m_funcInArrayDepth = 0;
-    m_parenDepth = 0;
+    m_funcParenExprDepth = 0;
     m_funcInArray = 0;
     m_tryCatchOrFinallyDepth = 0;
     m_UsesArgumentsAtGlobal = false;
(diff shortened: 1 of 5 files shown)


tcare:
I'm in the process of writing tests, but I wanted @pleath to take a look before he goes OOF.

lib/Parser/Parse.cpp
```diff
@@ -3055,6 +3056,10 @@ ParseNodePtr Parser::ParseTerm(BOOL fAllowCall,
         {
             m_deferEllipsisError = false;
         }
+
+        // Restore the deferred ellipsis error state now that an error has been ruled out.
+        m_deferEllipsisError = deferEllipsisErrorSave;
```
pleath: Do we also need to save/restore the RestorePoint (m_EllipsisErrorLoc) if there is one? (And only if there is one?)

tcare: That depends how much we care about emitting the first error we find. I'm thinking maybe just the paren depth save/restore should be enough.

pleath: Well...given that it's unusual to have one of these, it's probably a negligible cost to copy the restore point if it exists. I can imagine a compliance test failing if the wrong error is reported.

tcare: I mean more in the case of something like (...a, (...b, (...c etc. Presumably you'd want to emit the first one and not the innermost one, which is why we currently keep the first restore point even if we find subsequent errors. The only exception is if we find another error before we get back up to the spread/rest error, in which case that will be emitted first.

pleath: So does that mean that if ...b turned out to be an error, we would report the error at the location of ...a?

tcare: Hmm.. We might already have that problem if we have multiple paren depth >1 potential errors. 

The easiest option seems to be reporting the innermost confirmed error, because we don't really know what we're going to get as we finish parsing those nested paren exprs. We might get to the first one we captured, but we could have found a completely different syntax error in the mean time.

We could be smarter about using more context info to error out early, but again this is a rare situation. I might see what other engines do with some weird cases.

pleath: It's definitely more important that we report the correct location of the error we detected than it is to report the one that is lexically first, if that makes sense.

tcare:
I incorporated Paul's feedback. I added some tests but really I think extended fuzzer coverage will be better, and it took a long time to find this bug last time.

tcare:
Ping

tcare:
Ping @boingoing @akroshg

PR #3725, opened 2017-09-14
xplat: improve virtual memory region lookups

obastemur:
virtual memory region linked list reaches to ~400 items during AcmeAir
warmup and successful lookups are mostly around the last found item.

### DIFF
--- a/pal/src/map/virtual.cpp
+++ b/pal/src/map/virtual.cpp
@@ -57,6 +57,7 @@ typedef struct FREE_BLOCK {
 #endif  // MMAP_IGNORES_HINT
 
 // The first node in our list of allocated blocks.
+static PCMI pVirtualMemoryLastFound;
 static PCMI pVirtualMemory;
 
 #if MMAP_IGNORES_HINT
@@ -126,6 +127,7 @@ VIRTUALInitialize( void )
     InternalInitializeCriticalSection(&virtual_realloc);
 
     pVirtualMemory = NULL;
+    pVirtualMemoryLastFound = NULL;
 
     return TRUE;
 }
@@ -165,6 +167,7 @@ void VIRTUALCleanup()
         InternalFree(pTempEntry );
     }
     pVirtualMemory = NULL;
+    pVirtualMemoryLastFound = NULL;
 
 #if MMAP_IGNORES_HINT
     // Clean up the free list.
@@ -534,6 +537,19 @@ static PCMI VIRTUALFindRegionInformation( IN UINT_PTR address )
 
     pEntry = pVirtualMemory;
 
+    if (pVirtualMemoryLastFound && pVirtualMemoryLastFound->startBoundary <= address)
+    {
+        pEntry = pVirtualMemoryLastFound;
+        if (pEntry->startBoundary == address)
+        {
+            return pEntry;
+        }
+    }
+    else
+    {
+        pEntry = pVirtualMemory;
+    }
+
     while( pEntry )
     {
         if ( pEntry->startBoundary > address )
@@ -549,6 +565,8 @@ static PCMI VIRTUALFindRegionInformation( IN UINT_PTR address )
 
         pEntry = pEntry->pNext;
     }
+
+    if (pEntry) pVirtualMemoryLastFound = pEntry;
     return pEntry;
 }
 
@@ -611,6 +629,11 @@ static BOOL VIRTUALReleaseMemory( PCMI pMemoryToBeReleased )
         {
             pMemoryToBeReleased->pNext->pLast = pMemoryToBeReleased->pLast;
         }
+    }
+
+    if (pVirtualMemoryLastFound->startBoundary >= pMemoryToBeReleased->startBoundary)
+    {
+        pVirtualMemoryLastFound = NULL;
     }
 
 #if MMAP_IGNORES_HINT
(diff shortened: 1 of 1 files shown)


digitalinfinity:
@obastemur interesting- will review this- is there a win on AcmeAir here? How much?

pal/src/map/virtual.cpp
```diff
@@ -2002,6 +2025,7 @@ VirtualProtect(
     MemSize = (((UINT_PTR)(dwSize) + ((UINT_PTR)(lpAddress) & VIRTUAL_PAGE_MASK)
                 + VIRTUAL_PAGE_MASK) & ~VIRTUAL_PAGE_MASK);
 
+#if DEBUG
```
digitalinfinity: Curious: Why is this under DEBUG? Is the claim that we'll always call VirtualProtect with valid flags (and fail on debug if we don't?) Is it too costly to do a runtime check here? VirtualProtect calls should be rare (not on the hot path), and given the rest of the cost of VirtualProtect, having this runtime protection seems worthwhile? (Because otherwise, the behavior with invalid flags is more dangerous, and could cause security issues if an attacker can somehow corrupt the flags we pass into VirtualProtect)

pal/src/map/virtual.cpp
```diff
@@ -57,6 +57,7 @@ typedef struct FREE_BLOCK {
 #endif  // MMAP_IGNORES_HINT
 
 // The first node in our list of allocated blocks.
```
digitalinfinity: Nit: Move this comment one line below- and consider adding a new comment to describe what your new variable is

pal/src/map/virtual.cpp
```diff
@@ -534,6 +537,19 @@ static PCMI VIRTUALFindRegionInformation( IN UINT_PTR address )
 
     pEntry = pVirtualMemory;
 
+    if (pVirtualMemoryLastFound && pVirtualMemoryLastFound->startBoundary <= address)
+    {
+        pEntry = pVirtualMemoryLastFound;
```
digitalinfinity: Sorry, I don't understand the intent here- correct me if I'm wrong but this is how I understand the current code:
- pVirtualMemory is a list of allocated "regions", stored in increasing order of startAddress (hence the check on line 555)
- Your idea (which makes sense) is to have a 1 element MRU cache that you check first before walking the list- that is, check if the address is within the most recently used region

Is that right? If so, why are you checking for exact start instead of `address <= pEntry->startBoundary + pEntry->memSize`? It seems like the current condition is much more restrictive than it needs to be?

Also, minor nit but I personally find the assignment to pEntry redundant- I would rather check pVirtualMemoryLastFound directly instead of assigning to pEntry first but that's just me :)

pal/src/map/virtual.cpp
```diff
@@ -534,6 +537,19 @@ static PCMI VIRTUALFindRegionInformation( IN UINT_PTR address )
...
+        if (pEntry->startBoundary == address)
+        {
+            return pEntry;
+        }
+    }
+    else
+    {
+        pEntry = pVirtualMemory;
```
digitalinfinity: Nit: This assignment seems redundant- can remove

pal/src/map/virtual.cpp
```diff
@@ -613,6 +631,11 @@ static BOOL VIRTUALReleaseMemory( PCMI pMemoryToBeReleased )
         }
     }
 
+    if (pVirtualMemoryLastFound->startBoundary >= pMemoryToBeReleased->startBoundary)
```
digitalinfinity: Why >=? Should this not be == instead?

pal/src/map/virtual.cpp
obastemur: memSize not needed. This is to help us if the address we are looking is further from the previous search/find point on the linked list or not. Only condition we need for that is startBoundary <= address.

pal/src/map/virtual.cpp
obastemur: == condition is always true. Consumers of this method calls for lookup hence we setup the pVirtualMemoryLastFound first. I do check for `>` condition because of calls to this interface from `error` handling of PAL's virtual memory handling. Actually we should discard the value either it is equal or not. However, no point discarding it in case if it's `<`. So, the cost of `==` and `<=` is almost the same here but we might save the cache in case the problem happened in new memory request.

pal/src/map/virtual.cpp
obastemur: No, not because it is too costly. Because it's kinda redundant. Debug check should be just fine. no?

digitalinfinity: Curious- where is the cost here? I assume other things this function does (like call mprotect and VIRTUALFindRegionInformation) would be way more expensive than checking if the flags are valid, no?

pal/src/map/virtual.cpp
digitalinfinity: Can you explain why you're doing `pEntry->startBoundary == address` instead of within a region? For example, consider this case- your list has `A (startBoundary: 1000), B (startBoundary: 2000). `pVirtualMemoryLastFound` is set to `A`. You call `VIRTUALFindRegionInformation` with 1500- this should be a cache hit, but you'd miss the cache in this case, no? What did I misunderstand?

pal/src/map/virtual.cpp
obastemur: You are right. `mprotect` call is incomparably more expensive. As I mentioned, took them off because of they are rather redundant and not sure how cool we are with them on prod. mode.

pal/src/map/virtual.cpp
obastemur: Not sure I understood. Let's say we've been looping from the beginning. Wouldn't we loop until the pVirtualMemoryLastFound? and loop further or not? so, we set the loop start point here and do the rest similarly?

pal/src/map/virtual.cpp
digitalinfinity: Hmm, again, I didn't completely understand- sorry! The case I'm imagining is this:
```pVirtualMemory = 1000 -> 2000 -> 3000
pVirtualMemoryLastFound = 1000
```

Now you call `VIRTUALReleaseMemory` with 1000. In your case, wouldn't you invalidate the cache unnecessarily?

PR #3733, opened 2017-09-14
OS:13674598 use type specialized sym for upward exposed uses check

sigatrev:


sigatrev:
rebasing problems. trying again

PR #3736, opened 2017-09-14
OS#13255820 OS#13255822: Fix uninitialized locals and check return value of function that initializes them.

dilijev:


dilijev:

PR #3737, opened 2017-09-14
WASM: Implement new Sign Extend operators

Cellule:
Implement new operators as describe in
https://github.com/WebAssembly/threads/blob/master/proposals/threads/Overview.md#new-sign-extending-operators

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/3737)


MikeHolman:
IMO this should be behind a feature flag, as it is not yet in wasm proper.

Cellule:
> IMO this should be behind a feature flag, as it is not yet in wasm proper.

Fair enough, that's what I did initially, but turned it on by default at the last minute

Cellule:
Closing for now. I found a bug with const folding that I want to take care before moving on.

PR #3738, opened 2017-09-14
ConfigParser fix: handle leading whitespace

digitalinfinity:
The jscript.config used by IB has a leading whitespace- this change fixes an issue where ConfigParser doesn't deal with that.

Also added a mechanism to test parsing of jscript.config (Windows i.e rl.exe only for now)


Cellule:
I'm wondering why did you choose to implement testing custom config file in rl.exe instead of simply passing `-CustomConfigFile:myfile` by using `<compile-flags>` ?
It sounds like it would have been easier to implement the flag in jshost.exe instead of rl.exe

digitalinfinity:
rl's compile flags are a composition of whats in the compile flag node in rlexe.xml and whats passed in through other mechanisms (environment variables etc). Which means that we have different flags for interpreter/dynopogo etc. The new node is a mechanism to basically force RL to ignore the compile flags and just use the `-CustomConfigFile` flag

Cellule:
Are the flags passed by rl normally causing a problem with the test ?
The part I find confusing, is that we'll run those tests in both interpreter and dynapogo run, but they won't actually have different flags. 

Maybe I'm just being too picky, but I feel the point of this is to exercise `ConfigParser::ParseConfig` which could easily have been done through a WScript hook, or even extend `WScript.Flag` like `WScript.Flag("CustomConfig.config", true)` where true means the string represents a filename and not the actually flags.

digitalinfinity:
Sorry, forgot about this PR 😄 - yes, the additional flags being passed in cause problem. The test fundamentally relies on -Verbose to output which flags were passed in to compare with the baseline. When rl passes in different (additional) flags, the baseline match fails. We probably could have done this using WScript too- I just picked one option and didn't spend too much time thinking about it since it's a test for a test-only hook 😄

test/ConfigParsing/dummy.js
```diff
@@ -0,0 +1,6 @@
+//-------------------------------------------------------------------------------------------------------
+// Copyright (C) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE.txt file in the project root for full license information.
+//-------------------------------------------------------------------------------------------------------
+
+(function() {})();
```
dilijev: Does this test require a no-op function? Copyright check is fine with completely empty files with no copyright, so if you just need a dummy.js file you can have it as a completely empty file.

test/ConfigParsing/rlexe.xml
```diff
@@ -0,0 +1,39 @@
...
+    </default>
+  </test>
+  <test>
+    <default>
+      <files>dummy.js</files>
+      <baseline>dummy.baseline</baseline>
+      <custom-config-file>interspersednewline_leadingwhitespace.testconfig</custom-config-file>
+    </default>
+  </test>
+
```
dilijev: nit: unnecessary blank line

lib/Common/Core/ConfigParser.cpp
```diff
@@ -306,13 +306,23 @@ void ConfigParser::ParseRegistryKey(HKEY hk, CmdLineArgsParser &parser)
...
 {
 #if defined(ENABLE_DEBUG_CONFIG_OPTIONS) && CONFIG_PARSE_CONFIG_FILE
-    Assert(!_hasReadConfig);
+    Assert(!_hasReadConfig || strCustomConfigFile != nullptr);
     _hasReadConfig = true;
 
+    const char16* configFileName = strCustomConfigFile;
+    const char16* configFileExt = _u(""); /* in the custom config case,
+                                             ext is expected to be passed
+                                             in as part of the filename */
```
dilijev: nit: with a slight rewording you probably could fit this comment on the same line without it being too long

PR #3741, opened 2017-09-15
WASM: Implement new Sign Extend operators

Cellule:
Implement new operators as describe in
https://github.com/WebAssembly/threads/blob/master/proposals/threads/Overview.md#new-sign-extending-operators

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/3741)


Cellule:
Fixed the cse case and added tests, ready to review

lib/Backend/IRBuilderAsmJs.cpp
```diff
@@ -2394,6 +2395,13 @@ IRBuilderAsmJs::BuildInt2(Js::OpCodeAsmJs newOpcode, uint32 offset, Js::RegSlot
     case Js::OpCodeAsmJs::GrowMemory:
         instr = IR::Instr::New(Js::OpCode::GrowWasmMemory, dstOpnd, BuildSrcOpnd(AsmJsRegSlots::WasmMemoryReg, TyVar), srcOpnd, m_func);
         break;
+
+    case Js::OpCodeAsmJs::I32Extend8_s:  signExtendFromType = TyInt8;  goto make_sign_extend;
+    case Js::OpCodeAsmJs::I32Extend16_s: signExtendFromType = TyInt16; goto make_sign_extend;
+make_sign_extend:
+        // Src2 is a dummy source, used only to carry the type to cast from
+        instr = IR::Instr::New(Js::OpCode::Conv_Prim, dstOpnd, srcOpnd, IR::IntConstOpnd::New(signExtendFromType, signExtendFromType, m_func), m_func);
```
MikeHolman: this really needs a comment for why the int value is the type

lib/Backend/amd64/LowererMDArch.cpp
```diff
@@ -2564,6 +2564,31 @@ LowererMDArch::EmitUIntToFloat(IR::Opnd *dst, IR::Opnd *src, IR::Instr *instrIns
...
+    {
+        op = Js::OpCode::MOVSXW;
+    }
+    else if (src2->GetSize() == 4)
+    {
+        op = Js::OpCode::MOVSXD;
+    }
+
+    IR::RegOpnd * tempReg = IR::RegOpnd::New(fromType, this->m_func);
+    instr->InsertBefore(IR::Instr::New(Js::OpCode::MOV_TRUNC, tempReg, src1, m_func));
```
MikeHolman: why do you need this?

lib/Backend/i386/LowererMDArch.cpp
```diff
@@ -2802,6 +2802,52 @@ LowererMDArch::EmitLoadInt32(IR::Instr *instrLoad, bool conversionFromObjectAllo
...
+    IR::Opnd* src1 = instr->GetSrc1();
+    IR::Opnd* src2 = instr->GetSrc2();
+
+    // Src2 is used to determine what's the from type size
+    Assert(src2->GetSize() < dst->GetSize());
+    IRType fromType = src2->GetType();
+    Js::OpCode op = Js::OpCode::MOVSX;
+    if (src2->GetSize() == 2)
+    {
+        op = Js::OpCode::MOVSXW;
```
MikeHolman: should assert else size == 1

test/wasm/signextend.js
```diff
@@ -0,0 +1,77 @@
...
+        }
+      }
+    }
+  };
+}
+
+const tests = [
+  makeCSETest("i32", "extend8_s" , null, [0xFF, 1, 0x1FF]),
+  makeCSETest("i32", "extend16_s", null, [0xFF, 1, 0xFFFF, 0x1FFFF]),
+  makeCSETest("i64", "extend8_s" , null, [0xFF, 1, 0x1FF]),
```
MikeHolman: nice. very exhaustive :)

lib/Backend/amd64/LowererMDArch.cpp
Cellule: Initially, I thought we needed to explicitly zero out the upper bits of the register, but MOVSX, takes care of setting the bits, so it is not actually needed.

PR #3742, opened 2017-09-15
Never use exponential if radix not 10 (Fix #3739, #3740)

### LINKED ISSUE #3739: Follow-up #2751: For Number#toString with a radix other than 10, never switch to exponential display
Follow-up to #2751 (PR #3635)

Compatibility: align with other engines' implementation here.

See @littledan's comment https://github.com/Microsoft/ChakraCore/issues/2751#issuecomment-326902301

/cc @xiaoyinl @littledan @bterlson

### LINKED ISSUE #3740: Follow-up #2751: Number#toString: add tests on the edge cases between written-out and exponential notation
Follow-up to #2751 (PR #3635)

See comments on PR: https://github.com/Microsoft/ChakraCore/pull/3635

/cc @xiaoyinl

xiaoyinl:
Other engines (SpiderMonkey, JSC and V8) never use exponential display for num.toString(radix) if radix is not 10, no matter how large num is. This PR updates the code and the test cases to match that behavior.

This PR fixes #3739 and fixes #3740.

See also: #2751, #3635.

/cc @dilijev @littledan @bterlson


test/Number/toString.js
```diff
@@ -9,6 +9,7 @@ runTest('new Number("444" + "123.789123456789875436")');
 runTest('new Number(-444123.78963636363636363636)');
 runTest('new Number(0)');
 runTest('1e21');
+runTest('1.1e21');
```
dilijev: Maybe also add tests for `1e21+1` and `1e21-1` (integers one before and after the boundary) and compare output with other engines. Maybe `1e21+0.1` and `1e21-0.1` also?

xiaoyinl: These numbers all equal to `1e21` due to the loss of precision. How about adding `0.9999999999999999e21` and `1.0000000000000001e21`?

xiaoyinl:

test/Number/toString.js
dilijev: Yeah, that sounds good. :)

jackhorton:
Any updates on this? I believe it is blocking https://github.com/nodejs/node-chakracore/issues/390.

MikeHolman:
So just for fun I did some archeology to see why we are doing this, and turns out this code goes all the way back to at least 2002 (when it was copied between source trees). I don't think anyone has considered this behavior in a very long time.

Seems to me like a fine change to get compat with the other browsers.

dilijev:
@MikeHolman can you adopt and merge this PR?

dilijev:
Spoke with @MikeHolman - I'll test this one internally and then merge it.

dilijev:
Note: merging to release/1.7 to unblock the node-chakracore scenario: https://github.com/Microsoft/ChakraCore/pull/3742#issuecomment-335887549

PR #3743, opened 2017-09-15
Binding for special named properties

boingoing:
There are a few special properties in Javascript which have proper names such as `this`, `super`, and `new.target`. We have historically not done name binding for these properties instead hardcoding locations for them to live in the ByteCodeGenerator and loading from those locations when we emit specially-crafted parse nodes referring to those properties.

Originally we simply tracked a register where we stored the `this` value during the function prologue but over time the set of things we needed to keep track of has drastically increased. First came lexical capture of `this` for lambda and then `super` and the special handling for derived class constructor calls to `super()`. At this point juggling all of the state to specially handle these values is becoming a problem and hindering further development of frontend optimizations.

Instead of constructing special parse nodes for each of these special names, we now create ordinary-ish name parse nodes and push references to the well-known pids corresponding to those names. This acts basically exactly the same as how we handle references to ordinary identifiers. We use values for these pids which are not valid identifiers for Javascript code such that user code cannot declare them.

We create symbols for these special names in a slightly-special way. Ordinary name binding works by binding references to names upwards to their var decls. Special named properties like `this` do not have var decls, though. We fix this by adding var decls for special named properties to functions when they contain references to the names. When we are finished parsing a function, we check to see if it contains references to each special name. If it does, we create a var decl for the special name in the parameter scope and bind those references to it. The logic to do this is in `CreateSpecialSymbolDeclarations()`.

Global function never sets a binding for `super` or `new.target` but it does create a binding for `this` which will load the global `this` value. This binding is not stored in a root field for reasons explained below.

One key of this design is that lambda functions do not bind any of the special names. In this way, they are able to automatically capture the bindings from their parent. There is one caveat here which is the case of lambda with global function as the parent (global lambda). In this case, the lambda should capture the value of `this` by ordinary capture but it cannot because the global scope is not an object. We don't need to worry about the other special symbol capture here because `super` and `new.target` are invalid syntax at global scope. In this case, we create a binding for `this` in the global lambda and load the root `this` object into it.

Another benefit of doing binding of these special names is that `eval` doesn't have to do special work to capture these values, either. Direct `eval` doesn't need to create bindings for any of these special names and doesn't need to perform a stack walk, either. It can just do a scoped property lookup to fetch the outer bindings. There is a caveat here, as well, for the global `this` binding. If a direct `eval` captures global `this`, the scoped property lookup will not find any value for `this` in the scope chain. Instead we'll try to look in the root object but there can't be a `this` binding there because it would be observable, would be a reference to itself essentially, and would cause other problems. So instead, we detect this case when the scoped property lookup for the `this` property id fails and we load and return the global `this` value directly. With these improvements, `eval` no longer needs to load the `this` value via the first argument.

Indirect eval gets a binding for `this` which loads the global `this` value. It otherwise works the same as direct eval.

To keep track of the additional information we need to know which names are special, which nodes contain references to additional special names, etc we add three new parse nodes.

- `PnSpecialName` is a subclass of `PnPid` and is used as a holder for flags to indicate which special name this parse node represents. This is very similar to a regular `PnPid` and only needs to exist because we don't want to bloat `PnPid` itself with extra flags. There will also be more of these special names in the future. All `PnSpecialName` nodes have their opcode field set to `knopName` to avoid replacing all cases of `knopName` in conditionals with a boolean or of the two. We use a flag on `ParseNode` (`ParseNode::isSpecialName`) to indicate that the node has an `sxSpecialName` field.

- `PnSuperReference` is a subclass of `PnBin` and is used to indicate a super reference via dot access (`knopDot`) or index access (`knopIndex`). We need to have a reference to `this` at every super reference site because super references maintain the `this` binding through their calls. We don't have a flag on `ParseNode` or `PnBin` to indicate that the node has an `sxSuperReference` field. Instead we determine this by looking at the `pnode1` property and checking to see if it is a special name node for the `super` name.

- `PnSuperCall` is a subclass of `PnCall` and is used to indicate a call to `super()` in the context of a derived class constructor. The super call itself will need to use both the `this` and `new.target` values so we stash them on this parse node. All `PnSuperCall` nodes have their opcode field set to `knopCall` so as to avoid needing to change all the conditional statements currently checking for `knopCall` to include both (ie: `if(nop == knopCall)` => `if(nop == knopCall || nop == knopSuperCall)`). We use a flag on `PnCall` (`PnCall::isSuperCall`) to indicate the parse node has the `sxSuperCall` field.

In `ByteCodeGenerator` we treat these special symbols pretty much like ordinary symbols. We do store references to them directly in FuncInfo because we sometimes want to access them directly instead of during regular symbol enumeration. We collect the special symbols in `AddVarsToScope` similar to what we do for `arguments`.

There are a couple of pieces which are special we handle while emitting byte code. We do not check the special symbols for redeclarations. It should not be possible for them to be redeclared. Similarly, we skip initializing these special symbols via `LdUndef` because we will always immediately load a value in `LoadAllConstants` when the symbol exists. Because of that, we always assign a register for these special symbols. The pattern they have is to load a value into that register (via `LdThis`, `LdNewTarget`, `LdHomeObj`, etc) and then immediately store to the prop which will put the value into a scope slot if there is one.

The `super()` in a derived class constructor has a big impact on how these special symbols work. Because of `super` calls we need to have a TDZ for `this` in derived class constructors. This can't work like an ordinary block scope binding so we have a couple of concessions. When we emit a `knopName` node for the special name `this`, we'll emit a `ChkUndecl` if the symbol is marked as needing a declaration (we set this in the parser when `this` is bound in a derived class constructor) or if `this` was loaded from a scoped property lookup (we don't know where the `this` binding came from in that case - could have been a derived class constructor). In other cases we can skip this check because we know `this` was bound to some value. The other special work for `this` as part of `super()` has to do with loading the `this` value to make sure it has not already been assigned and assigning the result of `super()` back to `this`. When we load `this` to check that it is undeclared, this property load will throw a `ReferenceError` if it is, in fact, undecl. But we want to find out if the value is undeclared here, not throw if it is. This is accomplished via a flag added to `EmitPropLoad` named `skipUseBeforeDeclarationCheck`. When we pass this flag, we skip throwing a `ReferenceError` if the value is undecl. It is only used for this load of `this` in `super()`. The opposite is a flag added to `EmitPropStore` named `skipUseBeforeDeclarationCheck`. `EmitPropStore` will throw a `ReferenceError` at runtime if we try to store to a property which has an undecl value. That's exactly what we want to do when we assign to `this` as part of `super()`, though.

This change also makes a rough attempt to clean up all the now unused code we previously needed to keep around to manage the special values. There is probably a lot more code which can be safely removed in future changes.


tcare:
Huge but important change. It might take me a while to digest this. :)

lib/Parser/ptree.h
```diff
@@ -615,6 +618,23 @@ struct PnFinally : PnStmt
     ParseNodePtr pnodeBody;
 };
 
+struct PnSpecialName : PnPid
+{
+    bool isThis : 1;
+    bool isSuper : 1;
```
tcare: enum (nit)

lib/Parser/Parse.cpp
```diff
@@ -3508,8 +3680,30 @@ ParseNodePtr Parser::ParsePostfixOperators(
...
+                        {
+                            pnode = CreateSuperCallNode(pnode, pnodeArgs);
+                            Assert(pnode);
+
+                            PidRefStack* ref = this->PushPidRef(wellKnownPropertyPids._this);
+                            pnode->sxSuperCall.pnodeThis = CreateSpecialNameNode(wellKnownPropertyPids._this);
+                            pnode->sxSuperCall.pnodeThis->ichMin = pnode->ichMin;
+                            pnode->sxSuperCall.pnodeThis->ichLim = m_pscan->IchLimTok();
+                            pnode->sxSuperCall.pnodeThis->sxPid.SetSymRef(ref);
+                            pnode->sxSuperCall.pnodeThis->sxSpecialName.isThis = true;
```
tcare: Helper to hide pid ref stuff

lib/Parser/Parse.cpp
```diff
@@ -3534,6 +3734,11 @@ ParseNodePtr Parser::ParsePostfixOperators(
                         if (pToken->tk == tkID && pToken->pid == wellKnownPropertyPids.eval && count > 0) // Detect eval
                         {
                             this->MarkEvalCaller();
+
+                            this->PushPidRef(wellKnownPropertyPids._this);
```
tcare: >this->PushPidRef(wellKnownPropertyPids._this); [](start = 28, length = 46)

Hide behind something like "Reserve this"

lib/Parser/Parse.cpp
```diff
@@ -7783,7 +8078,7 @@ LPCOLESTR Parser::ConstructNameHint(ParseNodePtr pNode, uint32* fullNameHintLeng
     {
         leftNode = ConstructNameHint(pNode->sxBin.pnode1, fullNameHintLength, pShortNameOffset);
     }
-    else if (pNode->sxBin.pnode1->nop == knopName)
+    else if (pNode->sxBin.pnode1->nop == knopName && !pNode->sxBin.pnode1->isSpecialName)
```
tcare: Comment as to why?

boingoing: Good point, this is not clear

lib/Parser/Parse.cpp
boingoing: 👍

lib/Parser/ptree.h
boingoing: Agree but I'm going to punt that until we add another special name. Unless you feel strongly. 😃

lib/Parser/ptree.h
```diff
@@ -769,43 +796,46 @@ struct ParseNode
...
+const int kcbPnTri            = kcbPnNone + sizeof(PnTri);
+const int kcbPnTry            = kcbPnNone + sizeof(PnTry);
+const int kcbPnTryCatch       = kcbPnNone + sizeof(PnTryCatch);
+const int kcbPnTryFinally     = kcbPnNone + sizeof(PnTryFinally);
+const int kcbPnUni            = kcbPnNone + sizeof(PnUni);
+const int kcbPnUniSlot        = kcbPnNone + sizeof(PnUniSlot);
+const int kcbPnVar            = kcbPnNone + sizeof(PnVar);
+const int kcbPnWhile          = kcbPnNone + sizeof(PnWhile);
+const int kcbPnWith           = kcbPnNone + sizeof(PnWith);
+const int kcbPnParamPattern   = kcbPnNone + sizeof(PnParamPattern);
```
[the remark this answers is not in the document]
boingoing: Thanks, I moved the `const int` table around to put the entries in order. The entries between these two tables are actually slightly different (there is no `kcbPnLoop` while there is `PnLoop`, for example). 👍

PR #3747, opened 2017-09-15
[Preview][Archive] Enable WScript.RegisterModuleSource() in ch.exe

suwc:
Disclaimer: this is just to archive one way to do it in case it is useful. Not intended for merge as is.


suwc:
@boingoing, FYI, feel free to own it or ditch it. Thanks!

dilijev:
@boingoing do you think we should merge this?

boingoing:
@dilijev Yeah, we should probably clean it up and get it checked-in. We need this if we want to run exprgen with ch and test module code. We could also move all the full module tests into core. Not a high priority, but this should get us most of the way there.

rhuanjl:
@dilijev  @boingoing this is now a duplicate the feature has already been added by obastemur in #4582

PR #3753, opened 2017-09-18
Phase on/off flag default values

sigatrev:
This commit gives each phase a default on/off value. PHASE_ON and PHASE_OFF will use that value unless overriden with flags.

Phases that checked PHASE_ON previously default to off, all other phases default to on. There were 3 phases that checked both. They now default to on, and the former PHASE_ON behavior is now PHASE_FORCE

A const table is used for build flavors that ignore flags so that the checks optimize out.


lib/Runtime/ByteCode/Symbol.cpp
```diff
@@ -88,7 +88,7 @@ bool Symbol::IsInSlot(FuncInfo *funcInfo, bool ensureSlotAlloc)
 
 bool Symbol::GetIsCommittedToSlot() const
 {
-    if (!PHASE_ON1(Js::DelayCapturePhase))
+    if (!PHASE_FORCE1(Js::DelayCapturePhase))
```
rajatd: It almost seems like we're missing an optimization opportunity here. Have someone with frontend knowledge  evaluate.

PR #3756, opened 2017-09-18
fix bug with zero argument inlinees on ARM

MikeHolman:
With my wasm inlining change I made some conditions use `func->HasInlinee()` rather than `func->GetMaxInlineeArgOutCount() != 0`. This was causing an assert to fire on ARM.

I reverted there and another suspicious location to use the old check.

OS: 13782295


Cellule:
So that means there are cases where `GetMaxInlineeArgOutCount` is not zero, but it doesn't have an `inlinee` ?

Right now, we only call `UpdateMaxInlineeArgOutCount` on `Js::OpCode::InlineeStart` instr in GlobOpt.
Therefore, it would mean that it is possible to have an InlineeStart instr, but `func->HasInlinee()` is false.

MikeHolman:
It was the other way around, we have an Inlinee but GetMaxInlineeArgOutCount was 0.

Cellule:
Oh I see, it makes more sense, I couldn't find any way for my theory to be true.

LGTM

MikeHolman:
@dotnet-bot test Windows arm_test

PR #3767, opened 2017-09-20
fix bug with inlining and int64 syms on x86

MikeHolman:
When splitting int64 syms on x86 we map them to a dictionary of associated int32 syms, which is on the func... but with inlining, we might have the sym in the dictionary of 1 inlinee and other inlinees won't know about it.

Fix is to always look at topFunc.

OS: 13801633


Cellule:
@dotnet-bot test 
Ubuntu static_ubuntu_linux_test

Cellule:

Issue #3660, opened 2017-09-06
(1d,P3) Intl-ICU: Intl.DateTimeFormat: platform.validateAndCanonicalizeTimeZone


dilijev:
* [ ] Harden Intl logic to tolerate falsy returns, and add a fallback value. (See #3769)
* [ ] **Unblock other work:** return the argument string with no validation so that "good inputs" will work (check this in ASAP so that work on Intl.DateTimeFormat can go ahead without a proper implementation).
* [ ] Implement validation and canonicalization.

Issue #3659, opened 2017-09-06
(4h,P3) Intl-ICU: platform.getDefaultTimeZone


dilijev:
* [ ] Fallback logic (default value) (See #3769)
* [ ] Implement with ICU

jackhorton:
Fixed by #4645

PR #3769, opened 2017-09-20
Ref #3660: platform.GetDefaultTimeZone fallback logic

dilijev:
Ref #3660


lib/Runtime/Library/InJavascript/Intl.js
```diff
@@ -78,6 +78,15 @@
         return defaultLocale;
     }
 
+    let getDefaultTimeZone = function () {
+        let defaultTimeZone = platform.getDefaultTimeZone();
+        if (defaultTimeZone === null) {
```
agarwal-sandeep: >defaultTimeZone [](start = 12, length = 15)

If GetDefaultTimeZoneId fails in EntryIntl_GetDefaultTimeZone undefined is returned.

Since valid return value is a string why not check for typeof?

lib/Runtime/Library/InJavascript/Intl.js
```diff
@@ -1347,13 +1356,14 @@
                 var matcher = GetOption(options, "localeMatcher", "string", ["lookup", "best fit"], "best fit");
                 var timeZone = GetOption(options, "timeZone", "string", undefined, undefined);
```
agarwal-sandeep: >var [](start = 16, length = 3)

nit: Nice opportunity to change all vars to let (and const where applicable) :-)

lib/Runtime/Library/InJavascript/Intl.js
dilijev: A return of `undefined` was, in general, accounted for in the Intl.js logic for WinGlob. I added returns of null for unimplemented platform APIs and the fallback behavior was meant to catch that. I imagine that the fallback paths could be generalized to any implementation returning something other than the expected value. Will take note of this for a refactoring later on. Thanks!

lib/Runtime/Library/InJavascript/Intl.js
dilijev: Definitely planning on it, but that level of churn would delay productive work on Intl-ICU at the moment. Sometimes changing `var` to `let` causes hard-to-see issues (due to the permissiveness of `var`) and the Intl.js code is hard to test directly so it takes a long time to find out and that can be disruptive. I'll add a backlog task for this kind of refactoring.

dilijev:
@jackhorton I'm not sure what the issue is with the bytecode causing tests to fail.

I'll close this PR since it's not worth putting effort into resolving a change this small by itself. Feel free to include this JS change in your branch as you work on #3659

PR #3772, opened 2017-09-20
SetPropertyWithAttributes assert does not cover static constructor case

tcare:
Fixes OS: 12681861

We have a condition in an assert in SetPropertyWithAttributes that checks to see if we are coming from an InitClassMember opcode. We check this by getting the constructor property on the class and checking if it has the constructor flag set.

This misses the case where we have a static method named 'constructor'. In this case we will have the class method flag set. I added this case to the assert.

### DIFF
--- a/lib/Runtime/Types/DictionaryTypeHandler.cpp
+++ b/lib/Runtime/Types/DictionaryTypeHandler.cpp
@@ -1856,6 +1856,9 @@ namespace Js
             {
                 if (descriptor->IsAccessor && !(attributes & PropertyLetConstGlobal))
                 {
+#if DEBUG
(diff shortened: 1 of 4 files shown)


lib/Runtime/Types/DictionaryTypeHandler.cpp
```diff
@@ -1856,6 +1856,9 @@ namespace Js
             {
                 if (descriptor->IsAccessor && !(attributes & PropertyLetConstGlobal))
                 {
+#if DEBUG
+                    Var ctor = JavascriptOperators::GetProperty(instance, PropertyIds::constructor, scriptContext);
```
boingoing: How does this ctor work below if it's defined in `#if DEBUG`?

tcare: It's only used in the assert

PR #3774, opened 2017-09-20
perf: less collision - less comparison

obastemur:
AcmeAIR LTO gain ~2%

- Do not CrossSite::MarshalVar if scriptContext == requestContext
   Saves 3 compare + arithmetic ops per each change.

 - Use prime policy for inline cache dictionaries
    Less collision rate, more responsive


lib/Runtime/Language/JavascriptOperators.cpp
```diff
@@ -9367,7 +9370,12 @@ namespace Js
 
             Var thisVar = RootToThisObject(object, scriptContext);
 
-            RecyclableObject* marshalledFunction = RecyclableObject::FromVar(CrossSite::MarshalVar(requestContext, function));
+            RecyclableObject* marshalledFunction = function;
+            if (scriptContext != requestContext)
```
leirocks: can you make an overload of CrossSite::MarshalVar() to take both scriptContext as input parameter, to make the caller code cleaner

obastemur: @leirocks good point. I was thinking about it, let me update the PR.

tcare: +1, I like that convention.

lib/Runtime/Types/TypePropertyCache.cpp
```diff
@@ -296,8 +299,10 @@ namespace Js
                 requestContext);
             return true;
         }
-
-        *propertyValue = CrossSite::MarshalVar(requestContext, *propertyValue);
+        else
+        {
+            *propertyValue = CrossSite::MarshalVar(requestContext, *propertyValue);
```
tcare: Do we need to marshal here? We already had this check before and didn't marshal in the false case..

obastemur: code reaches here only when `! if(prototypeObjectWithProperty->GetScriptContext() == requestContext)`

tcare: Right, but it already did before without marshalling. Is there an explicit need now?

obastemur: Not sure. I just added `else` there.

tcare: Oh sorry the diff wasn't showing properly. That's fine.

obastemur:

PR #3775, opened 2017-09-21
Implement Dump method on part of DynamicTypeHandler hierarchy

ricobbe:
As part of my current investigation, I need to be able to dump a human-readable represntation of DictionaryTypeHandlerBase objects.  Add a virtual Dump method to DynamicTypeHandle to allow this, and provide stub implementations for other subtypes of DynamicTypeHandler.  (As the need arises, I may provide full implementations for these stubs; others are welcome to do so if they get to it before I do.)

### DIFF
--- a/lib/Runtime/Base/PropertyRecord.cpp
+++ b/lib/Runtime/Base/PropertyRecord.cpp
@@ -142,7 +142,7 @@ namespace Js
 
         Output::Print(_u("%*sPropertyRecord (0x%p):\n"), indent, padding, this);
         Output::Print(_u("%*spid: %d\n"), fieldIndent, padding, this->pid);
-        Output::Print(_u("%*shash: %#08x\n"), fieldIndent, padding, this->hash);
+        Output::Print(_u("%*shash: 0x%08x\n"), fieldIndent, padding, this->hash);
         Output::Print(_u("%*sisNumeric: %d\n"), fieldIndent, padding, this->isNumeric);
         Output::Print(_u("%*sIsBound: %d\n"), fieldIndent, padding, this->isBound);
         Output::Print(_u("%*sIsSymbol: %d\n"), fieldIndent, padding, this->isSymbol);
--- a/lib/Runtime/Base/PropertyRecord.cpp
+++ b/lib/Runtime/Base/PropertyRecord.cpp
@@ -135,7 +135,7 @@ namespace Js
     }
 
 #if DBG_DUMP
-	void PropertyRecord::Dump(unsigned indent) const
+    void PropertyRecord::Dump(unsigned indent) const
     {
         const auto padding(_u(""));
         const unsigned fieldIndent(indent + 2);
--- a/lib/Runtime/Base/PropertyRecord.cpp
+++ b/lib/Runtime/Base/PropertyRecord.cpp
@@ -134,6 +134,22 @@ namespace Js
         }
     }
 
+#if DBG_DUMP
+	void PropertyRecord::Dump(unsigned indent) const
+    {
+        const auto padding(_u(""));
+        const unsigned fieldIndent(indent + 2);
+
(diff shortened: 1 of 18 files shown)


lib/Runtime/Base/PropertyRecord.cpp
```diff
@@ -134,6 +134,22 @@ namespace Js
         }
     }
 
+#if DBG_DUMP
+	void PropertyRecord::Dump(unsigned indent) const
```
obastemur: nit: space

lib/Runtime/Types/DictionaryTypeHandler.cpp
```diff
@@ -2909,6 +2909,41 @@ namespace Js
     }
 #endif
 
+#if DBG_DUMP
+	template<typename T> void DictionaryTypeHandlerBase<T>::Dump(unsigned indent) const {
```
obastemur: nit: padding

lib/Runtime/Types/DictionaryTypeHandler.cpp
```diff
@@ -2909,6 +2909,41 @@ namespace Js
...
+        const auto padding(_u(""));
+        const unsigned fieldIndent(indent + 2);
+        const unsigned mapLabelIndent(indent + 4);
+        const unsigned mapValueIndent(indent + 6);
+
+        Output::Print(_u("%*sDictionaryTypeHandlerBase (0x%p):\n"), indent, padding, this);
+        if (this->propertyMap == nullptr)
+        {
+            Output::Print(_u("%*spropertyMap: null\n"), fieldIndent, padding);
```
obastemur: nit: `<null>` instead? to be similar with others?

obastemur:
I've been looking into Type system a bit recently. I'm glad to see this is coming.

couple of nits.

LGTM

obastemur:
BTW; about the Clang CI errors. Try casting to `(void*)` i.e. `(void*)this->propertyMap`

lib/Runtime/Base/PropertyRecord.cpp
ricobbe: Fixed.  (My emacs config was incorrectly not indenting definitions inside a namespace -- holdover from previous job -- so I was doing it by hand via C-q C-i, also a holdover from the previous job.  Thx for the comment; this motivated me to fix the emacs config to indent within a namespace!)

lib/Runtime/Types/DictionaryTypeHandler.cpp
ricobbe: Fixed

lib/Runtime/Types/DictionaryTypeHandler.cpp
ricobbe: Fixed

lib/Runtime/Types/DictionaryTypeHandler.cpp
```diff
@@ -2909,6 +2909,41 @@ namespace Js
...
+
+        Output::Print(_u("%*sDictionaryTypeHandlerBase (0x%p):\n"), indent, padding, this);
+        if (this->propertyMap == nullptr)
+        {
+            Output::Print(_u("%*spropertyMap: <null>\n"), fieldIndent, padding);
+        }
+        else
+        {
+            Output::Print(_u("%*spropertyMap: 0x%p\n"), fieldIndent, padding, static_cast<void*>(this->propertyMap));
+            for (auto iter = this->propertyMap->GetIterator(); iter.IsValid(); iter.MoveNext())
```
digitalinfinity: Can you use Map here instead of using an iterator?

ricobbe: Fixed.  I'd overlooked Map when reading through BaseDictionary; thanks!

PR #3780, opened 2017-09-21
Improving perf map support for dynamic code

MSLaguana:
Prior to this change the perf map did not reference which file a function came from, so anonymous functions or utility functions were very difficult to identify. I'm also changing from parens to square brackets for `[Interpreted]` etc since perf output includes parens by default for some metadata, and other tools were stripping out their contents.

### DIFF
--- a/lib/Runtime/PlatformAgnostic/Platform/Linux/PerfTrace.cpp
+++ b/lib/Runtime/PlatformAgnostic/Platform/Linux/PerfTrace.cpp
@@ -83,13 +83,19 @@ void  PerfTrace::WritePerfMap()
 
             scriptContext->MapFunction([=] (FunctionBody* body)
             {
+              char16 const * url = body->GetSourceContextInfo()->url;
+              if (body->GetSourceContextInfo()->IsDynamic()) {
+                url = _u("dynamic");
+              }
+
 #if DYNAMIC_INTERPRETER_THUNK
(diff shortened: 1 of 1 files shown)


lib/Runtime/PlatformAgnostic/Platform/Linux/PerfTrace.cpp
```diff
@@ -84,13 +84,19 @@ void  PerfTrace::WritePerfMap()
 
             scriptContext->MapFunction([=] (FunctionBody* body)
             {
+              char16 const * url = body->GetSourceContextInfo()->url;
+              if (body->GetSourceContextInfo()->IsDynamic()) {
+                url = _u("eval");
```
digitalinfinity: nit: i'd change this to dyn or something like that since strictly speaking we can have other dynamic code (eg. `new Function("...")`)

MikeHolman:
IMO it would be nice if we shared this with etw registration code so we name things consistently.

MSLaguana:
@MikeHolman What do you mean? You'd prefer if there was a single place that did tracing (such as `EtwTrace.cpp`) and produced common output between windows and xplat? Or you just want to see a consistent way of formatting names of functions?

MikeHolman:
Oh, I was just thinking about consistent naming.

dilijev:
@MSLaguana are you still planning to merge this change/

PR #3781, opened 2017-09-21
Update build system for more generic include paths

### LINKED ISSUE #3773: intl-icu branch needs an update to bytecode version GUID
#3750 changed the bytecode version by introducing a new PropertyId via adding to JnDirectFields.h
Since 1.7 (the parent branch) is in RELEASE mode (see lib/Common/ChakraCoreVersion.h), this means we need to update the GUID.

See lib/Runtime/ByteCode/ByteCodeCacheReleaseFileVersion.h

@jackhorton we spoke offline about how to do this.

jackhorton:
This will eventually be useful for standardizing how embedded ICU works across windows/xplat and ChakraCore/Node-ChakraCore. This also updates the bytecode version as a precaution.

Fixes #3773

### DIFF
--- a/Build/Chakra.Build.Paths.props
+++ b/Build/Chakra.Build.Paths.props
@@ -11,9 +11,9 @@
   </PropertyGroup>
   <PropertyGroup Label="ICU">
     <!-- IcuLibDir can be set in environment or with "/p:IcuLibDir=..." -->
-    <IcuLibDir Condition="'$(IcuLibDir)'=='' AND '$(Configuration)'=='Debug'">$(ChakraCoreRootDirectory)\deps\static\icu-small\debug</IcuLibDir>
-    <IcuLibDir Condition="'$(IcuLibDir)'=='' AND ('$(Configuration)'=='Test' or '$(Configuration)'=='Release')">$(ChakraCoreRootDirectory)\deps\static\icu-small\release</IcuLibDir>
+    <!-- <IcuLibDir Condition="'$(IcuLibDir)'=='' AND '$(Configuration)'=='Debug'">$(ChakraCoreRootDirectory)\deps\static\icu-small\debug</IcuLibDir> -->
+    <!-- <IcuLibDir Condition="'$(IcuLibDir)'=='' AND ('$(Configuration)'=='Test' or '$(Configuration)'=='Release')">$(ChakraCoreRootDirectory)\deps\static\icu-small\release</IcuLibDir> -->
     <!-- IcuIncludeDir can be set in environment or with "/p:IcuIncludeDir=..." -->
-    <IcuIncludeDir Condition="'$(IcuIncludeDir)'==''">$(ChakraCoreRootDirectory)\deps\icu-small\source</IcuIncludeDir>
+    <!-- <IcuIncludeDir Condition="'$(IcuIncludeDir)'==''">$(ChakraCoreRootDirectory)\deps\icu-small\source</IcuIncludeDir> -->
   </PropertyGroup>
 </Project>
--- a/Build/Chakra.Build.Paths.props
+++ b/Build/Chakra.Build.Paths.props
@@ -11,9 +11,9 @@
   </PropertyGroup>
   <PropertyGroup Label="ICU">
     <!-- IcuLibDir can be set in environment or with "/p:IcuLibDir=..." -->
-    <!-- <IcuLibDir Condition="'$(IcuLibDir)'=='' AND '$(Configuration)'=='Debug'">$(ChakraCoreRootDirectory)\deps\static\icu-small\debug</IcuLibDir> -->
-    <!-- <IcuLibDir Condition="'$(IcuLibDir)'=='' AND ('$(Configuration)'=='Test' or '$(Configuration)'=='Release')">$(ChakraCoreRootDirectory)\deps\static\icu-small\release</IcuLibDir> -->
+    <IcuLibDir Condition="'$(IcuLibDir)'=='' AND '$(Configuration)'=='Debug'">$(ChakraCoreRootDirectory)\deps\static\icu-small\debug</IcuLibDir>
+    <IcuLibDir Condition="'$(IcuLibDir)'=='' AND ('$(Configuration)'=='Test' or '$(Configuration)'=='Release')">$(ChakraCoreRootDirectory)\deps\static\icu-small\release</IcuLibDir>
     <!-- IcuIncludeDir can be set in environment or with "/p:IcuIncludeDir=..." -->
-    <!-- <IcuIncludeDir Condition="'$(IcuIncludeDir)'==''">$(ChakraCoreRootDirectory)\deps\icu-small\source</IcuIncludeDir> -->
+    <IcuIncludeDir Condition="'$(IcuIncludeDir)'==''">$(ChakraCoreRootDirectory)\deps\icu-small\source</IcuIncludeDir>
   </PropertyGroup>
 </Project>
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -456,7 +456,8 @@ include_directories(
     pal
     pal/inc
     pal/inc/rt
-    ${ICU_INCLUDE_PATH}
+    ${ICU_INCLUDE_PATH}/common
+    ${ICU_INCLUDE_PATH}/i18n
     )
 
 if(ICU_INCLUDE_PATH)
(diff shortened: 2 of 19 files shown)


jackhorton:
Fixes #3773

Also, has there ever been a discussion about refactoring all of these include directories? The CMakeLists and vcxprojs look totally different here, and at the cost of possibly introducing some header naming conflicts (which seems unlikely to me right now), I think it would be nice if we could simplify the system at least enough for every project to look the same. If its an item at all, it would be fairly low-priority, but still useful IMO

jackhorton:
Oh and lastly, @dilijev , I think I removed the default lib and include path if ICU was true mistakenly, I can add that back in if you still need it. I can also make some of the conditions more strict (like IntlICU==true AND IcuIncludePath=='' for setting default include paths, just to be sure)

Build/Chakra.Build.Paths.props
```diff
@@ -11,9 +11,9 @@
   </PropertyGroup>
   <PropertyGroup Label="ICU">
     <!-- IcuLibDir can be set in environment or with "/p:IcuLibDir=..." -->
-    <IcuLibDir Condition="'$(IcuLibDir)'=='' AND '$(Configuration)'=='Debug'">$(ChakraCoreRootDirectory)\deps\static\icu-small\debug</IcuLibDir>
-    <IcuLibDir Condition="'$(IcuLibDir)'=='' AND ('$(Configuration)'=='Test' or '$(Configuration)'=='Release')">$(ChakraCoreRootDirectory)\deps\static\icu-small\release</IcuLibDir>
+    <!-- <IcuLibDir Condition="'$(IcuLibDir)'=='' AND '$(Configuration)'=='Debug'">$(ChakraCoreRootDirectory)\deps\static\icu-small\debug</IcuLibDir> -->
```
MSLaguana: What is the purpose of keeping these lines around commented out? Are they intended as documentation of how to specify the `IcuLibDir` parameter to msbuild?

jackhorton: The real reason is because I commented them out to ensure nothing was being set behind my back and then made the pull request without remembering to either comment their removal better or add them back. My comment above asking doug what to do here is why it hasn't been fixed yet.

lib/Runtime/ByteCode/ByteCodeCacheReleaseFileVersion.h
```diff
@@ -4,6 +4,6 @@
 //-------------------------------------------------------------------------------------------------------
 // NOTE: If there is a merge conflict the correct fix is to make a new GUID.
 
-// {5C44723E-F7DF-4397-A24B-AC1A90D221F9}
+// {131ae129-1863-4b97-b9b2-4802d60acefb}
 const GUID byteCodeCacheReleaseFileVersion =
-{ 0x5C44723E, 0xF7DF, 0x4397, { 0xA2, 0x4B, 0xAC, 0x1A, 0x90, 0xD2, 0x21, 0xF9 } };
+{ 0x131ae129, 0x1863, 0x4b97, { 0xb9, 0xb2, 0x48, 0x02, 0xd6, 0x0a, 0xce, 0xfb } };
```
dilijev: :+1: Thanks!

Note: Fixes #3773

dilijev:
@jackhorton 

> I think I removed the default lib and include path if ICU was true mistakenly, I can add that back in if you still need it. 

It is convenient to have them. Let's clean up things like that only after this work lands to make development easier for the time being.

Build/Chakra.Build.Paths.props
dilijev: Yeah, let's keep them for now.

dilijev:
@digitalinfinity Any objections to moving these dependencies into a single place at the top level instead of in the ~1/2 of projects I had added it to before. IIRC we had a conversation about minimal dependencies at some point and that was the reason for that somewhat unweildy decision. FWIW I like this better and it matches the way we do it for CMake/xplat which is one reason to unify.

jackhorton:
@dilijev Re-added the default lib and include values

digitalinfinity:
@dilijev no, no objections as long as whatever technique you use works fine with the Windows build systems

CMakeLists.txt
```diff
@@ -456,7 +456,8 @@ include_directories(
     pal
     pal/inc
     pal/inc/rt
-    ${ICU_INCLUDE_PATH}
+    ${ICU_INCLUDE_PATH}/common
```
obastemur: why this is needed?

jackhorton: Because @dilijev is working on things right now that have headers in <icu>/source/i18n in addition to source/common. Its easier if we normalize everything off of just icu/source

obastemur: Well, this is not a trivial change. Better if we keep it as is. There is a reason they kept the header files in two sep. folders. We shouldn't normalize them.

dilijev: Not sure what you mean. `ICU_INCLUDE_PATH` will have a known directory tree with files in known places. The dependency on ICU headers is effectively the directory that needs to be specified. In my WIP*, I'm adding `${ICU_INCLUDE_PATH}/i18n` because that's where those headers are.

* see: https://github.com/dilijev/ChakraCore/tree/intl-numfmt and this commit in particular:
https://github.com/dilijev/ChakraCore/commit/a04cde6ce70c1a9d92f2add6dc55b51a19a98cb4

https://github.com/dilijev/ChakraCore/blob/a04cde6ce70c1a9d92f2add6dc55b51a19a98cb4/lib/Runtime/Base/Chakra.Runtime.Base.vcxproj#L31-L32

If not this strategy, what would you propose? Have the build system specify two directories that are within the same parent folder for semantically the same dependency?

obastemur: 1 - This is the place where you set the search path for compiler. If you put both `i18n` and `common` separately, you would normalize all the header files under.

How it's different than copying all the header files from those folders into a common folder?
Well, worst case, it might break and we won't even notice.

What's wrong with `#include <i18n/THIS_FILE>` and `#include <common/THAT_FILE>` ?

2 - ICU is optional. There is a reason we created a variable for `ICU_INCLUDE_PATH`... Once `${ICU_INCLUDE_PATH}` is not defined, what exactly do expect from compiler for `${ICU_INCLUDE_PATH}/common` or `${ICU_INCLUDE_PATH}/....` ??

If we didn't do this for Windows project files, well, it's time we make it right?

dilijev: I see your point. FWIW from what I've seen, including them this way seems to be common practice, and in the documentation is even shown as just the filename itself, implying that they would go even further to specify all of:

* `${ICU_INCLUDE_PATH}/common/` (has not been applicable in the project yet but is supported by the current scheme)
* `${ICU_INCLUDE_PATH}/common/unicode`
* `${ICU_INCLUDE_PATH}/i18n/unicode`

However, I understand where you are coming from, and agree there's no reason to keep adding more directories when we can specify the top level one instead.

This would mean that we need to update all of the ICU includes to start scoped to the true ICU directory.

Jack, do you think we should do that now or make a revision to this in a subsequent pull request? Up to you based on whether you feel like landing this immediately is important and an update can come later.

obastemur: Alternatively, only if INTL is enabled, add `${ICU_INCLUDE_PATH}../i18/`.. Mind `../` because `ICU_INCLUDE_PATH` == `common` already.

However, we should track from CMAKE and make sure INTL is enabled before adding that.

So, we won't be updating build scripts command lines (it's not just us!) and make the header sharing limited to INTL usage. I guess, this is what you want?

jackhorton: I think that both of the following are fine, all things being considered equal: 

1. put the paths in the #includes
```CMake
include_directories(
  ...
  ${ICU_INCLUDE_PATH}
  )
```
```C
#include "common/unicode/uloc.h"
#include "i18n/header.h"
```
2. Scope the include
```CMake
if(SET ${ICU_INCLUDE_PATH})
  include_directories(
    ${ICU_INCLUDE_PATH}/common
    ${ICU_INCLUDE_PATH/i18n
    )
endif()
```
The latter takes fewer changes, so I would probably prefer the former at least for now.

jackhorton: Note that ICU_INCLUDE_PATH will only be set when we are not using the system ICU -- it will be worth looking at where the i18n headers are when theyre installed on the system and probably choose whatever is necessary to conform to that

dilijev: > The latter takes fewer changes, so I would probably prefer the former at least for now.

Prefer the former because it takes more changes?

jackhorton: I have no idea what I was on about there, I prefer solution #2.

dilijev:
Note for @jackhorton and @obastemur:

I'm closing this PR and will include @jackhorton's changes to the Windows build (and GUID update) in my next PR. This is because the Windows changes do not regress any current scenario, and are necessary to unblock my work. The Linux changes will be left to a future change after the proper path forward can be investigated in more detail.

PR #3783, opened 2017-09-21
Disable vector call temporarily

agarwal-sandeep:
Disable vector call temporarily


agarwal-sandeep:
@Cellule this is part of moving chakra full nugget package

PR #3788, opened 2017-09-22
perf: Improve internal type discovery and getPropertyInternal

obastemur:
Acme AIR LTO gain ~2%

- Fast path for IsNull / IsUndefined (in case the argument is
recyclableObject, or we know for sure object's context). Especially
NullCheck was pretty common within the internal loops

- Introduce internal GetProperty calls by introducing
GetPropertyNoCache that serves only for RecyclableObject+ and when info ==
nullptr.

- Improve logic in related places. (i.e. use less GetTypeId calls,
GetPropertyName calls...)


CMakeLists.txt
```diff
@@ -155,7 +155,7 @@ if(CC_TARGETS_AMD64)
     add_definitions(-D_M_X64_OR_ARM64)
     add_compile_options(-msse4.2)
 
-    set(CAN_BUILD_WABT 1)
+    #set(CAN_BUILD_WABT 1)
```
digitalinfinity: Why this change? (cc @Cellule)

lib/Runtime/Language/JavascriptOperators.cpp
```diff
@@ -10479,11 +10502,43 @@ namespace Js
...
     }
 
+    BOOL JavascriptOperators::IsNull(Var instance, ScriptContext* scriptContext)
+    {
+        return JavascriptOperators::IsNull(instance, scriptContext->GetLibrary());
+    }
+
+    BOOL JavascriptOperators::IsNull(Var instance, JavascriptLibrary* library)
+    {
+        Assert(!RecyclableObject::Is(instance) ? 1 : ((RecyclableObject*)instance)->GetScriptContext()->GetLibrary() == library );
```
digitalinfinity: Use TRUE/FALSE instead of hardcoding constants here

lib/Runtime/Language/JavascriptOperators.cpp
```diff
@@ -10494,20 +10549,27 @@ namespace Js
...
 
     BOOL JavascriptOperators::IsUndefinedObject(Var instance, ScriptContext *scriptContext)
     {
-        return JavascriptOperators::IsUndefinedObject(instance, scriptContext->GetLibrary()->GetUndefined());
+        return JavascriptOperators::IsUndefinedObject(instance, scriptContext->GetLibrary());
     }
 
     BOOL JavascriptOperators::IsUndefinedObject(Var instance, JavascriptLibrary* library)
     {
+        Assert(!RecyclableObject::Is(instance) ? 1 : ((RecyclableObject*)instance)->GetScriptContext()->GetLibrary() == library );
```
digitalinfinity: Use TRUE/FALSE instead of magic numbers

CMakeLists.txt
Cellule: Indeed, we need wabt for some of our test on xplat (agreed very few passes).
What we need to do is not set this on release builds, but leave it on test/debug builds

obastemur: no worries! was experimenting with the CI. libwabt is no longer part of the binary for xplat release builds

lib/Runtime/Language/JavascriptOperators.cpp
```diff
@@ -1679,6 +1679,34 @@ namespace Js
         return GetProperty_Internal<false>(instance, RecyclableObject::FromVar(instance), true, propertyId, value, requestContext, info);
     }
 
+    BOOL JavascriptOperators::GetProperty_InternalSimple(Var instance, RecyclableObject* object, PropertyId propertyId, Var* value, ScriptContext* requestContext)
```
Penguinwizzard: Could you SAL-annotate the outparam?

lib/Runtime/Language/JavascriptOperators.cpp
```diff
@@ -10706,6 +10768,28 @@ namespace Js
...
     }
 
+    Var JavascriptOperators::GetPropertyNoCache(RecyclableObject* instance, PropertyId propertyId, ScriptContext* requestContext)
+    {
+        return JavascriptOperators::GetPropertyNoCache(instance, instance, propertyId, requestContext);
+    }
+
+    Var JavascriptOperators::GetPropertyNoCache(Var instance, RecyclableObject* propertyObject, PropertyId propertyId, ScriptContext* requestContext)
+    {
+        Var value;
```
Penguinwizzard: nit: I see that the current implementation of GetProperty_InternalSimple definitely sets value; however, could you initialize it to nullptr here?

obastemur:
@Cellule after rebasing to latest r1.7, I started to get wasm related CI errors (Windows) for this PR. Could you please take a look what may cause?

lib/Runtime/Language/JavascriptOperators.cpp
```diff
@@ -10494,20 +10550,27 @@ namespace Js
...
 
     BOOL JavascriptOperators::IsUndefinedObject(Var instance, ScriptContext *scriptContext)
     {
-        return JavascriptOperators::IsUndefinedObject(instance, scriptContext->GetLibrary()->GetUndefined());
+        return JavascriptOperators::IsUndefinedObject(instance, scriptContext->GetLibrary());
     }
 
     BOOL JavascriptOperators::IsUndefinedObject(Var instance, JavascriptLibrary* library)
     {
+        Assert(!RecyclableObject::Is(instance) ? TRUE : ((RecyclableObject*)instance)->GetScriptContext()->GetLibrary() == library );
```
Cellule: From what I understand, this Assert is wrong in Cross Context scenario.
That wasm test just happen to exercice that code path, but it's not a wasm related bug.

obastemur:

PR #3792, opened 2017-09-22
Utilize PropertyString cache for Object.assign

MikeHolman:
Actual change is pretty straightforward. Most of this is refactoring to centralize the PropertyString cache code.

Improves react-redux perf by about 2%.

### DIFF
--- a/lib/Runtime/Language/JavascriptConversion.cpp
+++ b/lib/Runtime/Language/JavascriptConversion.cpp
@@ -290,24 +290,28 @@ CommonNumber:
             // For all other types, convert the key into a string and use that as the property name
             JavascriptString * propName = JavascriptConversion::ToString(key, scriptContext);
 
-            if (VirtualTableInfo<Js::PropertyString>::HasVirtualTable(propName))
+            PropertyString * propertyString = PropertyString::TryFromVar(propName);
+            if (propertyString == nullptr)
             {
-                PropertyString * propertyString = (PropertyString *)propName;
-                *propertyRecord = propertyString->GetPropertyRecord();
+                LiteralStringWithPropertyStringPtr * strWithPtr = LiteralStringWithPropertyStringPtr::TryFromVar(propName);
+                if (strWithPtr != nullptr)
+                {
+                    if (!strWithPtr->GetPropertyString())
+                    {
+                        propertyString = strWithPtr->GetPropertyString();
+                    }
+                    else
+                    {
+                        scriptContext->GetOrAddPropertyRecord(propName->GetString(), propName->GetLength(), propertyRecord);
+                        propertyString = scriptContext->GetPropertyString((*propertyRecord)->GetPropertyId());
+                        strWithPtr->SetPropertyString(propertyString);
+                    }
+                }
             }
-            else if (VirtualTableInfo<Js::LiteralStringWithPropertyStringPtr>::HasVirtualTable(propName))
+
+            if (propertyString != nullptr)
             {
-                LiteralStringWithPropertyStringPtr * str = (LiteralStringWithPropertyStringPtr *)propName;
-                if (str->GetPropertyString())
-                {
-                    *propertyRecord = str->GetPropertyString()->GetPropertyRecord();
-                }
-                else
-                {
-                    scriptContext->GetOrAddPropertyRecord(propName->GetString(), propName->GetLength(), propertyRecord);
-                    PropertyString * propStr = scriptContext->GetPropertyString((*propertyRecord)->GetPropertyId());
-                    str->SetPropertyString(propStr);
-                }
+                *propertyRecord = propertyString->GetPropertyRecord();
             }
             else
             {
--- a/lib/Runtime/Language/JavascriptConversion.cpp
+++ b/lib/Runtime/Language/JavascriptConversion.cpp
@@ -290,32 +290,37 @@ CommonNumber:
             // For all other types, convert the key into a string and use that as the property name
             JavascriptString * propName = JavascriptConversion::ToString(key, scriptContext);
 
+            // Check if we have one of the JavascriptString types which allow us to directly read the PropertyRecord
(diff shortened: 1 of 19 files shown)


lib/Runtime/Library/PropertyString.h
```diff
@@ -6,58 +6,86 @@
...
-    };
+public:
+    virtual VTableValue DummyVirtualFunctionToHinderLinkerICF()
+    {
+        return VTableValue::VtablePropertyString;
+    }
+};
+
+
+template <typename T> inline
```
thomasmo: why does this need to be templated? is it not guaranteed to always be a Var?

lib/Runtime/Language/JavascriptConversion.cpp
```diff
@@ -301,25 +301,29 @@ namespace Js
...
+                    else
+                    {
+                        scriptContext->GetOrAddPropertyRecord(propName->GetString(), propName->GetLength(), propertyRecord);
+                        propertyString = scriptContext->GetPropertyString((*propertyRecord)->GetPropertyId());
+                        strWithPtr->SetPropertyString(propertyString);
+                    }
                 }
             }
+
+            if (propertyString != nullptr)
```
thomasmo: this seems to be changing the logic of this function beyond simple refactoring.
should this be an else-if instead of just if? I think that keeps the same behavior.

thomasmo: as in, with the original if/else-if/else, only one block was executed. now, more than one can execute, logically. not sure if there is some property/invariant of the code itself that ensures the same outcome.

lib/Runtime/Language/JavascriptOperators.cpp
```diff
@@ -4346,31 +4320,26 @@ namespace Js
         }
 
         // fastpath for PropertyStrings only if receiver == object
-        if (!TaggedInt::Is(index) && JavascriptString::Is(index) &&
-            (VirtualTableInfo<Js::PropertyString>::HasVirtualTable(index) || VirtualTableInfo<Js::LiteralStringWithPropertyStringPtr>::HasVirtualTable(index)))
+        PropertyString * propertyString = PropertyString::TryFromVar(index);
```
thomasmo: why don't we need to check that index is a TaggedInt?

thomasmo: oh--because it can't be a taggedint and javascriptstring?

lib/Runtime/Library/PropertyString.cpp
```diff
@@ -48,6 +60,88 @@ namespace Js
...
+        Var const instance,
+        RecyclableObject *const object,
+        Var *const propertyValue,
+        ScriptContext *const requestContext,
+        PropertyValueInfo *const propertyValueInfo)
+    {
+        if (ShouldUseCache())
+        {
+            PropertyValueInfo::SetCacheInfo(propertyValueInfo, this, GetLdElemInlineCache(), true);
+            if (CacheOperators::TryGetProperty<true, true, true, true, true, true, false, true, false>(
```
thomasmo: please add a comment that translates this cryptic template parameter list into English.
same with the other function below

lib/Runtime/Library/PropertyString.cpp
```diff
@@ -48,6 +60,88 @@ namespace Js
...
+        Var propertyValue,
+        ScriptContext *const requestContext,
+        const PropertyOperationFlags propertyOperationFlags,
+        PropertyValueInfo *const propertyValueInfo)
+    {
+        if (ShouldUseCache())
+        {
+            PropertyValueInfo::SetCacheInfo(propertyValueInfo, this, GetStElemInlineCache(), true);
+            if (CacheOperators::TrySetProperty<true, true, true, true, true, false, true, false>(
+                object,
```
thomasmo: nit: can we be consistent with params on the same vs new line between this and the other function?

lib/Runtime/Library/PropertyString.h
MikeHolman: You're right. The idea is that it might be a RecyclableObject, in which case we don't need check for RecyclableObject in the Is method. I'll add a comment.

lib/Runtime/Library/JavascriptObject.cpp
```diff
@@ -1540,15 +1540,28 @@ namespace Js
                 scriptContext->GetOrAddPropertyRecord(propertyName->GetString(), propertyName->GetLength(), &propertyRecord);
                 nextKey = propertyRecord->GetPropertyId();
             }
+            PropertyString * propertyString = PropertyString::TryFromVar(propertyName);
 
-            if (!JavascriptOperators::GetOwnProperty(from, nextKey, &propValue, scriptContext))
+            PropertyValueInfo getPropertyInfo;
+            if (!propertyString || !propertyString->TryGetPropertyFromCache(from, from, &propValue, scriptContext, &getPropertyInfo))
```
thomasmo: this set of changes seem to be the core of the perf improvement?
please add a brief comment for how/why this is a win. (or, update the comment at the top of the while loop?)

lib/Runtime/Language/JavascriptConversion.cpp
MikeHolman: semantically it's equivalent -- no matter what we are going to get the propertyRecord, but some of the paths have different perf characteristics. and you're right the new code isn't doing things quite right. i'll try to better organize this

---
In reply to: [141236387](https://github.com/Microsoft/ChakraCore/pull/3792#discussion_r141236387) [](ancestors = 141236387)

lib/Runtime/Library/PropertyString.h
thomasmo: thanks!

lib/Runtime/Language/JavascriptConversion.cpp
thomasmo: thanks for the followup

PR #3794, opened 2017-09-22
Cleanup Wasm tracing flags

Cellule:
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/3794)

### DIFF
--- a/lib/WasmReader/WasmBinaryReader.cpp
+++ b/lib/WasmReader/WasmBinaryReader.cpp
@@ -214,6 +214,17 @@ SectionHeader WasmBinaryReader::ReadSectionHeader()
     return header;
 }
 
+#if ENABLE_DEBUG_CONFIG_OPTIONS
+Js::FunctionBody* WasmBinaryReader::GetFunctionBody() const
+{
+    if (m_readerState == READER_STATE_FUNCTION)
+    {
+        return m_funcState.body;
+    }
+    return nullptr;
+}
+#endif
+
 #if DBG_DUMP
 void WasmBinaryReader::PrintOps()
 {
@@ -244,6 +255,9 @@ void WasmBinaryReader::PrintOps()
             --j;
         }
     }
+
+    uint32 moduleId = m_module->GetWasmFunctionInfo(0)->GetBody()->GetSourceContextId();
+    Output::Print(_u("Module #%u's current opcode distribution\n"), moduleId);
     for (i = 0; i < count; ++i)
     {
         switch (ops[i])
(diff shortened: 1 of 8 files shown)


lib/WasmReader/WasmBinaryReader.cpp
```diff
@@ -214,6 +215,17 @@ SectionHeader WasmBinaryReader::ReadSectionHeader()
...
+    if (m_readerState == READER_STATE_FUNCTION)
+    {
+        return m_funcState.body;
+    }
+    return nullptr;
+}
+#endif
+
 #if DBG_DUMP
 void WasmBinaryReader::PrintOps()
```
Cellule: I wasn't sure if we wanted to keep PrintOps.
In the current code base in doesn't do anything, because we print before actually parsing any opcode.

lib/WasmReader/WasmBinaryReader.cpp
```diff
@@ -9,6 +9,7 @@
 #include "WasmLimits.h"
 #if ENABLE_DEBUG_CONFIG_OPTIONS
 #include "Codex/Utf8Helper.h"
+#include <inttypes.h>
```
MikeHolman: Why is this needed?

lib/WasmReader/WasmByteCodeGenerator.h
```diff
@@ -159,6 +159,7 @@ namespace Wasm
         Js::WebAssemblyModule* GenerateModule();
         void GenerateFunctionHeader(uint32 index);
     private:
+        Js::FunctionBody* GetFunctionBody() const { return nullptr; }
```
MikeHolman: this seems bad

lib/WasmReader/WasmReader.h
```diff
@@ -20,23 +20,19 @@
     }
 
 // Level of tracing
-#define DO_WASM_TRACE_ALL       PHASE_TRACE1(Js::WasmPhase)
-#define DO_WASM_TRACE_BYTECODE  DO_WASM_TRACE_ALL || PHASE_TRACE(Js::WasmBytecodePhase, GetFunctionBody())
-#define DO_WASM_TRACE_DECODER   DO_WASM_TRACE_ALL || PHASE_TRACE1(Js::WasmReaderPhase)
-  #define DO_WASM_TRACE_SECTION DO_WASM_TRACE_DECODER || PHASE_TRACE1(Js::WasmSectionPhase)
-#define DO_WASM_TRACE_LEB128    PHASE_TRACE1(Js::WasmLEB128Phase)
+#define WASM_TRACE_BODY_CHECK(phase) (GetFunctionBody() ? PHASE_TRACE(phase, GetFunctionBody()) : PHASE_TRACE1(phase))
```
MikeHolman: I guess this is why you added the GetFunctionBody method, but without this context it is confusing

lib/WasmReader/WasmByteCodeGenerator.h
Cellule: Agreed, I think there is only one case that needs it for the macro, so I'll just rewrite that part accordingly

PR #3796, opened 2017-09-23
Fix ES module filename not visible in stack trace/debugger

### LINKED ISSUE #3618: No filename information for errors in ES modules
A runtime error in an ES module produces a stack trace like:

```
ReferenceError: 'getEaten' is not defined
   at Showcase.prototype.on_update (:49:3)
   at Anonymous function (#/game_modules/thread.js:134:26)
   at Anonymous function (#/game_modules/thread.js:195:3)
   at next (#/runtime/from.js:267:4)
   at nullOp (#/runtime/from.js:706:9)
   at Anonymous function (#/runtime/from.js:101:3)
   at _updateAll (#/game_modules/thread.js:187:2)
```

The frames above that have filename information were in scripts loaded using `require()`.  ES module frames like the top one `Showcase.prototype.on_update` do not.

fatcerberus:
JsParseModuleSource() had been creating the source context with url set to `nullptr`, which caused the module name not to be visible in error backtraces or debugger JSON, e.g.:
```
Error: PIG!
   at mp3Demo (:28:3)
```

With this patch the URL for the module is set to the host-normalized module specifier, so that:
```
Error: PIG!
   at mp3Demo (@/bin/main.mjs:28:3)
```

I'm not sure if I implemented this properly; in particular I wasn't sure if SourceTextModuleRecord::GetSpecifier() could ever return nullptr, in which case the fix is incorrect.  Please advise on that point.

Fixes #3618.

### DIFF
--- a/lib/Jsrt/Core/JsrtCore.cpp
+++ b/lib/Jsrt/Core/JsrtCore.cpp
@@ -77,7 +77,10 @@ JsParseModuleSource(
         SourceContextInfo* sourceContextInfo = scriptContext->GetSourceContextInfo(sourceContext, nullptr);
         if (sourceContextInfo == nullptr)
         {
-            sourceContextInfo = scriptContext->CreateSourceContextInfo(sourceContext, nullptr, 0, nullptr, nullptr, 0);
+            Js::JavascriptString *specifier = Js::JavascriptString::FromVar(moduleRecord->GetSpecifier());
+            const char16 *urlSz = specifier->GetSz();
+            size_t urlLength = specifier->GetLength();
+            sourceContextInfo = scriptContext->CreateSourceContextInfo(sourceContext, urlSz, urlLength, nullptr, nullptr, 0);
         }
         SRCINFO si = {
             /* sourceContextInfo   */ sourceContextInfo,
--- a/lib/Runtime/Language/SourceTextModuleRecord.cpp
+++ b/lib/Runtime/Language/SourceTextModuleRecord.cpp
@@ -171,7 +171,7 @@ namespace Js
         {
             if (*exceptionVar == nullptr)
             {
-                *exceptionVar = JavascriptError::CreateFromCompileScriptException(scriptContext, &se);
+                *exceptionVar = JavascriptError::CreateFromCompileScriptException(scriptContext, &se, this->GetSpecifierSz());
             }
             if (this->parser)
             {
@@ -831,7 +831,7 @@ namespace Js
         this->rootFunction = scriptContext->GenerateRootFunction(parseTree, sourceIndex, this->parser, this->pSourceInfo->GetParseFlags(), &se, _u("module"));
         if (rootFunction == nullptr)
         {
-            this->errorObject = JavascriptError::CreateFromCompileScriptException(scriptContext, &se);
+            this->errorObject = JavascriptError::CreateFromCompileScriptException(scriptContext, &se, this->GetSpecifierSz());
             OUTPUT_TRACE_DEBUGONLY(Js::ModulePhase, _u("\t>NotifyParentAsNeeded rootFunction == nullptr\n"));
             NotifyParentsAsNeeded();
         }


fatcerberus:
...aaaaannnnddd I broke all the tests.  Will have to fix it later tonight.

fatcerberus:
This is still failing CI tests after my fix, but it looks like the same tests fail in master so I’m not sure whether this change is good or not... someone?

I’m not yet sure who’s responsible for what around here :)

fatcerberus:
I'm failing CI tests, but I don't understand why--do the tests somehow rely on ES modules not having a filename?  Because that's literally the only change I've made here.

fatcerberus:
I took a look into the tests I'm failing and it seems that this is what has happened:
```
22:05:39 Output:
22:05:39 ----------------------------
22:05:39 WARNING: DynamicProfileStorage: Dynamic Profile Data corrupted: 'undefined'
22:05:39 WARNING: DynamicProfileStorage: Dynamic Profile Data corrupted: 'null'
22:05:39 WARNING: DynamicProfileStorage: Dynamic Profile Data corrupted: 'true'
22:05:39 WARNING: DynamicProfileStorage: Dynamic Profile Data corrupted: 'false'
22:05:39 WARNING: DynamicProfileStorage: Dynamic Profile Data corrupted: '1234567890'
22:05:39 WARNING: DynamicProfileStorage: Dynamic Profile Data corrupted: 'abc789cde'
22:05:39 WARNING: DynamicProfileStorage: Dynamic Profile Data corrupted: 'number1007.2'
22:05:39 WARNING: DynamicProfileStorage: Dynamic Profile Data corrupted: '[object Promise]'
22:05:39 WARNING: DynamicProfileStorage: Dynamic Profile Data corrupted: 'Infinity'
22:05:39 WARNING: DynamicProfileStorage: Dynamic Profile Data corrupted: '[object Promise][object Promise]'
22:05:39 WARNING: DynamicProfileStorage: Dynamic Profile Data corrupted: 'NaN'
22:05:39 WARNING: DynamicProfileStorage: Dynamic Profile Data corrupted: '0'
22:05:39 WARNING: DynamicProfileStorage: Dynamic Profile Data corrupted: '-12345'
22:05:39 WARNING: DynamicProfileStorage: Dynamic Profile Data corrupted: '1.1234567890123457'
22:05:39 WARNING: DynamicProfileStorage: Dynamic Profile Data corrupted: '-1.1234567890123457'
22:05:39 WARNING: DynamicProfileStorage: Dynamic Profile Data corrupted: '[object Object]'
22:05:39 pass
```

That just seems like something went wrong in the test framework itself rather than any problem with my change *per se*.  Note the `pass` at the end, but the test framework still fails the test because the output doesn't match.

MSLaguana:

fatcerberus:
@MSLaguana It looks like the issue is that the failing test does things like `import(undefined)` in which case the file doesn't exist and the promise should be rejected - however ch.exe still calls JsParseModuleSource anyway.  That seems like a bug - should I fix it?

fatcerberus:
On further investigation, it turns out that a module which fails to parse for *any reason* (a syntax error, for example) will produce DynamicProfileStorage warnings in the dynapogo tests causing them to fail, but *only if the module's source context has an URL associated with it*.  This actually seems like it might be a bug in the test harness. as the test output indeed contains the word "pass".

fatcerberus:
@MSLaguana 
On the advice of @rhuanjl [here](https://github.com/Microsoft/ChakraCore/issues/3618#issuecomment-337377812), I changed the patch to set the URL *after* calling CreateSourceContextInfo, which seems to have allowed the tests to pass.  It's an ugly hack but *does* seem to work, and it would be nice to get further insight into the warnings mentioned above.  If there is a better way to avoid triggering them than what I have here, I wasn't able to figure it out.

fatcerberus:
Closing in favor of championing #4080.

PR #3802, opened 2017-09-25
Add cache for RegExp.test

MikeHolman:
It seems to be a common enough pattern where people will create a RegExp and test a small number of unique strings many times.

For this, I added a simple cache of size 8 on the RegExp object for RegExp.test, which caches the result for some given input string.

I disable the cache it in case of using global or sticky, since the result can change depending on lastIndex and caching likely wouldn't help those patterns.

After experimentation I decided on size 8, because there were rapidly diminishing returns at size 16 and at size 4 I saw a big increase in cache misses.

I tried dynamic sizing as well, but megamorphic cases would quickly max out the cache and the polymorphic cases didn't seem to get much benefit above size 8, so it seemed better to avoid the overhead of resizing and stick with a fixed sized cache.

Improves perf of angular by about 6%.

### DIFF
--- a/lib/Runtime/Library/JavascriptRegularExpression.h
+++ b/lib/Runtime/Library/JavascriptRegularExpression.h
@@ -9,7 +9,7 @@ namespace Js
     struct RegExpTestCache
     {
         bool result;
-        Field(JavascriptString*) input;
+        Field(RecyclerWeakReference<JavascriptString> *) input;
     };
     class JavascriptRegExp : public DynamicObject
     {
--- a/lib/Runtime/Library/JavascriptRegularExpression.h
+++ b/lib/Runtime/Library/JavascriptRegularExpression.h
@@ -6,6 +6,11 @@
 
 namespace Js
 {
+    struct RegExpTestCache
+    {
+        bool result;
+        Field(JavascriptString*) input;
+    };
     class JavascriptRegExp : public DynamicObject
     {
         friend class JavascriptRegularExpressionType;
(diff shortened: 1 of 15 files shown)


lib/Runtime/Library/RegexHelper.cpp
```diff
@@ -665,24 +665,57 @@ namespace Js
...
 #if ENABLE_REGEX_CONFIG_OPTIONS
         RegexHelperTrace(scriptContext, UnifiedRegex::RegexStats::Test, regularExpression, input);
+        JavascriptRegExp::TraceTestCache(cachedValue == input, input, cachedValue, !useCache);
 #endif
-        const bool isGlobal = pattern->IsGlobal();
-        const bool isSticky = pattern->IsSticky();
+
+        if (useCache && cachedValue == input)
+        {
+            return JavascriptBoolean::ToVar(cache[cacheIndex].result, scriptContext);
```
agarwal-sandeep: >return JavascriptBoolean::ToVar(cache[cacheIndex].result, scriptContext); [](start = 12, length = 73)

In case of a cache hit Is it still worth to do the slow lookup in chk build and validate against cached value?

MikeHolman: seems like a good idea

lib/Runtime/Library/JavascriptRegularExpression.h
```diff
@@ -6,6 +6,11 @@
 
 namespace Js
 {
+    struct RegExpTestCache
+    {
+        bool result;
+        Field(JavascriptString*) input;
```
rajatd: What do you think about making this a weak reference to the string?

MikeHolman: I originally tried this and had problems with this due to ArenaLiteralStrings. These are only used in a couple of places and really don't seem that useful, so I decided to remove them altogether and replace them with normal recycler managed strings.

lib/Runtime/Library/RegexHelper.cpp
```diff
@@ -665,24 +665,68 @@ namespace Js
...
+            cacheHit = cachedInput == input;
+        }
 #if ENABLE_REGEX_CONFIG_OPTIONS
         RegexHelperTrace(scriptContext, UnifiedRegex::RegexStats::Test, regularExpression, input);
+        JavascriptRegExp::TraceTestCache(cacheHit, input, cachedInput, !useCache);
 #endif
-        const bool isGlobal = pattern->IsGlobal();
-        const bool isSticky = pattern->IsSticky();
+
+        if (useCache && cacheHit)
```
agarwal-sandeep: >useCache && [](start = 12, length = 11)

nit: no need of this as cacheHit can only be true if useCache was true.

PR #3805, opened 2017-09-26
OS#13894912 WASM: Yielding in unreachable code

Cellule:
Allow the eval stack to be empty when yielding values in Unreachable state.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/3805)

### DIFF
--- a/lib/Runtime/Language/RuntimeLanguagePch.h
+++ b/lib/Runtime/Language/RuntimeLanguagePch.h
@@ -5,9 +5,9 @@
 #pragma once
 
 #include "Parser.h"
-#include "WasmReader.h"
 
 #include "Runtime.h"
+#include "WasmReader.h"
 
 #include "Language/AsmJsUtils.h"
 #include "Language/AsmJsLink.h"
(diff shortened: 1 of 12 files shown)


Cellule:
ping

Cellule:
ping

MikeHolman:
>         EmitInfo retExprInfo = explicitRetInfo ? *explicitRetInfo : PopEvalStack();

I believe this also needs to be changed

---
Refers to: lib/WasmReader/WasmByteCodeGenerator.cpp:1376 in 4987bb4. [](commit_id = 4987bb48f604a002daaa2fd79dbcfe291b09b8f1, deletion_comment = False)

Cellule:
> I believe this also needs to be changed

I don't think we should change it. Since `return` is an explicit operator, I believe it needs to always have a value to pop. I think the rule applies only to implicit pop (ie yields)

Cellule:
@MikeHolman Can you take a look at the last fix ?

PR #3806, opened 2017-09-26
OSX: Use the max between rsp vs USRSTACK** 

obastemur:
Check and manually fix if `USRSTACK**` < rsp

Targets the issue with OSX 12.6 / xcode9 (ChakraCore crashes on startup)

### DIFF
--- a/pal/src/misc/sysinfo.cpp
+++ b/pal/src/misc/sysinfo.cpp
@@ -125,6 +125,9 @@ Note:
 Others are set to zero.
 
 --*/
+
+int GetCurrentThreadStackLimits(ULONG_PTR* lowLimit, ULONG_PTR* highLimit);
+
 VOID
 PALAPI
 GetSystemInfo(
@@ -167,6 +170,15 @@ GetSystemInfo(
     TRACE("dwNumberOfProcessors=%d\n", nrcpus);
     lpSystemInfo->dwNumberOfProcessors = nrcpus;
 
+// TODO: (BSD) is it a good idea to take maxOf(GetCurrentThreadStackLimits, USRSTACKXX) ?
+#ifdef __APPLE__ // defined(USRSTACK64) || defined(USRSTACK)
+ULONG_PTR lowl, highl;
+GetCurrentThreadStackLimits(&lowl, &highl);
+#ifndef PAL_MAX
+#define PAL_MAX(a, b) ((a > b) ? (a) : (b))
+#endif
+#endif
+
 #ifdef VM_MAXUSER_ADDRESS
     lpSystemInfo->lpMaximumApplicationAddress = (PVOID) VM_MAXUSER_ADDRESS;
 #elif defined(__LINUX__)
@@ -175,12 +187,12 @@ GetSystemInfo(
     lpSystemInfo->lpMaximumApplicationAddress = (PVOID) USERLIMIT;
 #elif defined(_WIN64)
 #if defined(USRSTACK64)
-    lpSystemInfo->lpMaximumApplicationAddress = (PVOID) USRSTACK64;
+    lpSystemInfo->lpMaximumApplicationAddress = (PVOID) PAL_MAX(highl, USRSTACK64);
 #else // !USRSTACK64
 #error How come USRSTACK64 is not defined for 64bit?
 #endif // USRSTACK64
 #elif defined(USRSTACK)
-    lpSystemInfo->lpMaximumApplicationAddress = (PVOID) USRSTACK;
+    lpSystemInfo->lpMaximumApplicationAddress = (PVOID) PAL_MAX(highl, USRSTACK);
 #else
 #error The maximum application address is not known on this platform.
 #endif


pal/src/misc/sysinfo.cpp
```diff
@@ -175,12 +186,12 @@ GetSystemInfo(
     lpSystemInfo->lpMaximumApplicationAddress = (PVOID) USERLIMIT;
 #elif defined(_WIN64)
 #if defined(USRSTACK64)
```
MSLaguana: Are these conditions correct? Won't `highl` be undefined (and undeclared) except when `__APPLE__` is `#define`d, which would not be the case if `_WIN64` is defined?

obastemur: Yes. `USRSTACK64` defines only when  `__APPLE__` is defined.

MSLaguana: If this section is intended to only be for osx builds, would it make sense to replace the `#elif defined(_WIN64)` with `#elif defined(__APPLE__)` to make it clear that this is for osx only? If that is semantically correct, I'd prefer that.

obastemur: > If this section is intended to only be for osx builds, would it make sense to replace the #elif defined(_WIN64) with #elif defined(__APPLE__)

@MSLaguana _WIN64 != __APPLE__ .. It's a definition PAL uses for amd64.

MSLaguana: I must be misunderstanding something here then. From what I read here, it looks like at this point we know

* It is not `__LINUX__`
* There is no hard-coded `VM_MAXUSER_ADDRESS` or `USERLIMIT`
* It is 64bit due to `_WIN64`

Does that mean that it is definitely `__APPLE__`? If so, perhaps move the `#ifdef __APPLE__` section from above into this case, so it's clear that this case only applies to osx builds? Both for the `USRSTACK64` and `USRSTACK` case together?

obastemur: Agree, we could do some cleanup in PAL code..

Just a reminder; This PR didn't change anything there (no previously defined place) and has no desire to redesign the DEFS in that file. I just added `ifdef __APPLE__` because the both of those relevant places are `__APPLE__` only. If you say that there was a better way to do this, well let me do that.

MSLaguana: e.g.

```c++
#ifdef VM_MAXUSER_ADDRESS
    lpSystemInfo->lpMaximumApplicationAddress = (PVOID) VM_MAXUSER_ADDRESS;
#elif defined(__LINUX__)
    lpSystemInfo->lpMaximumApplicationAddress = (PVOID) MAX_PROCESS_VA_SPACE_LINUX;
#elif defined(USERLIMIT)
    lpSystemInfo->lpMaximumApplicationAddress = (PVOID) USERLIMIT;
#elif defined(__APPLE__)

    ULONG_PTR lowl, highl;
    GetCurrentThreadStackLimits(&lowl, &highl);
#ifndef PAL_MAX
#define PAL_MAX(a, b) ((a > b) ? (a) : (b))
#endif
#ifdef _WIN64
#if defined(USRSTACK64)
    lpSystemInfo->lpMaximumApplicationAddress = (PVOID) PAL_MAX(highl, USRSTACK64);
#else // !USRSTACK64
#error How come USRSTACK64 is not defined for 64bit?
#endif // USRSTACK64
#else // !_WIN64
#ifdef USRSTACK
    lpSystemInfo->lpMaximumApplicationAddress = (PVOID) PAL_MAX(highl, USRSTACK);
#else // !USRSTACK
#error How come USRSTACK is not defined for 32bit?
#endif // USRSTACK
#endif // _WIN64

#else
#error The maximum application address is not known on this platform.
#endif

```

obastemur: @MSLaguana  I just saw that I didn't answer your question below so.. probably that's why.. you came up with the code above.

> Does that mean that it is definitely __APPLE__?

Well, `USRSTACK` etc. are BSD based. 

However, we don't support BSD yet. Couldn't be nice to have a compiler error there and don't do this `PAL_MAX` comparison on a target we don't support yet? Well, that what this PR does?

I'm not sure what's the argument here... So, I can make it `#if defined(USRSTACK) || defined(USRSTACK64)` ? and also `#ifndef __APPLE__ -> #error 'Are you sure you want to compare this and that?` ....

Makes sense?

MSLaguana: My main concerns are consistency and clarity. With the code as you have it now, it was not clear to me that the `#ifdef __APPLE__` section would be run (and thus declare/define highl) in the same circumstances as the `USRSTACK`/`USRSTACK64` sections.

If you change it from `#ifdef __APPLE__` to `#if defined(USRSTACK) || defined(USRSTACK64)` then that would make it clear to me.

obastemur: @MSLaguana as I mentioned on my comment above, that `ifdef` version needs additional error check. However, updated the PR and added comments there.

release-57-1.zip
```diff
@@ -0,0 +1,34 @@
+<?xml version="1.0"?>
```
MSLaguana: Looks like you've accidentally included this file.

obastemur: yep..

PR #3807, opened 2017-09-26
WASM.SIMD Float64x2 Ops

Krovatkin:
This PR includes 

* Float64x2 Ops + tests


Krovatkin:
@Cellule , could you please take a look when you have a moment?

Cellule:
I noticed something you might want to put on your radar (ie not for this particular PR)
When we move a float64 into a simd128 we end up not peeping out those useless moves.
for instance in Splat we sometimes end up with 
```
19729F1103B   s16(xmm0)[Simd128Float32x4].simd128 = MOVSD  s11(xmm0).f64
```
Peep doesn't know it can remove this instr because it looks at the dst's type and from this check if the opcode is `MOVUPS`. I tried to relax `LowererMD::IsAssign` to only check if the opcode is any kind of mov instr, but it fired other asserts that we need to verify.

---

Reviewed 23 of 23 files at r1.
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

*[lib/Backend/LowerMDSharedSimd128.cpp, line 929 at r1](https://reviewable.io:443/reviews/microsoft/chakracore/3807#-Kv3m0cxAO0HZFSY8icE:-Kv3m0cxAO0HZFSY8icF:bhhxupg) ([raw file](https://github.com/microsoft/chakracore/blob/5587cb491614332fda9ffa922054c078dc4a9ff8/lib/Backend/LowerMDSharedSimd128.cpp#L929)):*
> ```cpp
>     Assert(dst->IsSimd128() && src1->IsSimd128());
> 
>     if (instr->m_opcode == Js::OpCode::Simd128_ReplaceLane_D2)
> ```

I think we can do better here
With `SHUFPD` we can directly place lane 0 of src3 into lane 1 of dst and keep lane 0 of dst intact.
The part that I am less fond is that src3 is a Float64 and thus doesn't match the expected types of `SHUFPD`.
In practice it works because Float64 uses xmm register and thus have "128bits".
If the float64 has been spilled for some reason, I think it still works because `SHUFPD` would only read the first 8 bytes of a memory location in that shuffle.

```cpp
if (instr->m_opcode == Js::OpCode::Simd128_ReplaceLane_D2)
    {
        AssertMsg(AutoSystemInfo::Data.SSE2Available(), "SSE2 not supported");
        Assert(src3->IsFloat64());
        m_lowerer->InsertMove(dst, src1, instr);
        if (lane == 0)
        {
            instr->InsertBefore(IR::Instr::New(Js::OpCode::MOVSD, dst, src3, m_func));
        }
        else
        {
            Assert(lane == 1);
            instr->InsertBefore(IR::Instr::New(Js::OpCode::SHUFPD, dst, src3, IR::IntConstOpnd::New(0 | 1 << 2, TyInt8, m_func, true), m_func));
        }
        return removeInstr(instr);
    }
```

---

*[lib/Runtime/Language/SimdUtils.h, line 235 at r1](https://reviewable.io:443/reviews/microsoft/chakracore/3807#-Kv3egEM36ZlEH22Ho3v:-Kv3egEM36ZlEH22Ho3w:b-4osfe8) ([raw file](https://github.com/microsoft/chakracore/blob/5587cb491614332fda9ffa922054c078dc4a9ff8/lib/Runtime/Language/SimdUtils.h#L235)):*
> ```objc
>         static inline SIMDValue SIMD128InnerReplaceLaneD2(SIMDValue simdVal, const uint32 lane, const double value)
>         {
>             simdVal.f64[lane] = value;
> ```

nit: add `Asset(lane < 2)`

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/microsoft/chakracore/3807#-:-Kv3oB7x4ajqkFeKNy9X:b-5wnpcy)*

Krovatkin:
>  noticed something you might want to put on your radar (ie not for this particular PR)
When we move a float64 into a simd128 we end up not peeping out those useless moves.
for instance in Splat we sometimes end up with

Could you please elaborate on this a little bit more? 
It seems to me that `s16[Simd128Float32x4].simd128 = MOVSD  s12.f64 ` moves a seed value into a destination register which is then broadcasted across all the lanes `s16[Simd128Float32x4].simd128 = SHUFPD  s16[Simd128Float32x4].simd128, 0 (0x0).i8 #`. It doesn't look like a useless move to me :-( 

```
    s12.f64         =  MOVSD          prm2<16>.f64!                           #


 GLOBOPT INSTR:     s16[Simd128Float32x4].simd128 = Simd128_Splat_D2  s12.f64!                #0017


    s16[Simd128Float32x4].simd128 = MOVSD  s12.f64                            #
    s16[Simd128Float32x4].simd128 = SHUFPD  s16[Simd128Float32x4].simd128, 0 (0x0).i8 #
```

Cellule:
In Lower you are right the instruction makes sense and we shouldn't remove it because we don't know yet that it is a useless move.
However, after Register allocator we have
```
19729F1103B   s16(xmm0)[Simd128Float32x4].simd128 = MOVSD  s11(xmm0).f64
```
where `movsd xmm0, xmm0` doesn't do anything since the dst and src are the same for that movsd

---

Reviewed 1 of 1 files at r2, 1 of 1 files at r3.
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---

*[lib/Backend/LowerMDSharedSimd128.cpp, line 936 at r3](https://reviewable.io:443/reviews/microsoft/chakracore/3807#-Kv4UalDRWG7NVJwGF_s:-Kv4UalDRWG7NVJwGF_t:boawoid) ([raw file](https://github.com/microsoft/chakracore/blob/d4d65f01b4066faff11a6c5d76c26807eedfd24e/lib/Backend/LowerMDSharedSimd128.cpp#L936)):*
> ```cpp
>         if (lane) 
>         {
>             instr->InsertBefore(IR::Instr::New(Js::OpCode::SHUFPD, dst, src3, IR::IntConstOpnd::New(0, TyInt8, m_func, true), m_func));
> ```

Shouldn't the immediate should be `0b10` aka `2`
I think my original code was buggy regarding this too.
Not sure why it passes unittest though

From the doc
```
IF SELECT[0] = 0
THEN DEST[63:0] ← DEST[63:0]; // This is the case we want
ELSE DEST[63:0] ← DEST[127:64]; FI;
IF SELECT[1] = 0
THEN DEST[127:64] ← SRC[63:0];
ELSE DEST[127:64] ← SRC[127:64]; FI; // This is the case we want
```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/microsoft/chakracore/3807#-:-Kv4WAphZa_xMPimfujA:bm5c7pi)*

Krovatkin:
@Cellule 

> ELSE DEST[127:64] ← SRC[127:64]; FI; // This is the case we want

Are we sure about this? `SRC` is actually a 64-bit float in our case. `SRC[127:64]` is undefined. Let me run this case under debugger, I'll get back to you.

Krovatkin:
@Cellule Looks like we are doing the right thing. `src3` a.k.a. `SRC` is indeed stored in the lower 64 bits.
This is one of your unittests. 

image

Cellule:
Ohh You are right!!! I got confused.. hahaha all good

Cellule:
:lgtm:

---

Review status: all files reviewed at latest revision, all discussions resolved, some commit checks pending.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/microsoft/chakracore/3807#-:-Kv4pF8bzFimEV0BgGvJ:bnfp4nl)*

Krovatkin:
> In Lower you are right the instruction makes sense and we shouldn't remove it because we don't know yet that it is a useless move.
However, after Register allocator we have

I see what you mean! I'll address this shortly!

```asm 
036A103E F2 0F 10 45 0C       movsd       xmm0,mmword ptr [ebp+0Ch]  
036A1043 F2 0F 10 4D 14       movsd       xmm1,mmword ptr [ebp+14h]  

036A1048 F2 0F 10 C0          movsd       xmm0,xmm0  

036A104C 66 0F C6 C0 00       shufpd      xmm0,xmm0,0  
036A1051 66 0F C6 C1 00       shufpd      xmm0,xmm1,0  
036A1056 0F 11 01             movups      xmmword ptr [ecx],xmm0  
```

Krovatkin:
@Cellule could you please help us merge this PR?

Krovatkin:
@Cellule @MikeHolman could you please help us merge this PR?

Cellule:
I have merged the PR.
We have some outage in our infra to mirror our repo to GitHub, it might take a little while before this PR is actually closed

PR #3809, opened 2017-09-26
Intl-ICU: implement Intl.NumberFormat under ICU

### LINKED ISSUE #3661: (1d) Intl-ICU: platform.currencyDigits
* [x] What is this used for?
  * platform.currencyDigits is the number of fractional digits for this currency
* [x] Return a "reasonable" default (2) as a temporary measure.
* [x] Determine if there is an equivalent in ICU
* [x] Implement in ICU

### LINKED ISSUE #3664: (1d) Intl-ICU: platform.formatNumber

### LINKED ISSUE #3670: (4h) Intl-ICU: Intl.NumberFormat (ctor)

### LINKED ISSUE #3671: (4h) Intl-ICU: Intl.NumberFormat.prototype.format (getter)

### LINKED ISSUE #3716: (1d) Intl-ICU: platform.cacheNumberFormat

### LINKED ISSUE #3773: intl-icu branch needs an update to bytecode version GUID

dilijev:
Closes #3661 (platform.currencyDigits)
Closes #3664 (platform.formatNumber)
Closes #3716 (platform.cacheNumberFormat) 
Closes #3670 (Intl.NumberFormat ctor)
Closes #3671 (Intl.NumberFormat.prototype.format)

Includes some of @jackhorton's changes from PR #3781:

* Windows ICU dependency updates
* GUID update (Fixes #3773)


lib/Runtime/Base/WindowsGlobalizationAdapter.cpp
```diff
@@ -829,6 +718,7 @@ if (this->object) \
         return ResolveLocaleLookup(scriptContext, locale, resolved);
     }
 
+    // REVIEW (doilij): Return type of this function doesn't match the value returned. Should probably explicitly cast at the call site if needed.
```
jackhorton: I changed the return type of getDefaultLocale to be `size_t` in my pull, which is fine because its only used by immediately checking it against 0

lib/Runtime/Library/InJavascript/Intl.js
```diff
@@ -191,6 +191,11 @@
 
         let matches = platform.builtInRegexMatch(locale, LANG_TAG_RE); // String.prototype.match i.e. string.match(re)
         let extensionsString = matches[4];
+        // Note: extensions are /((${extension})-)*/ and are made up of \\b(?:${singleton}(?:-${alphanum}{2,8})+)\\b
+        // where the ${alphanum}{2,8} fields are of the form `${key}-${value}`.
+        // We will return an array of those `${key}-${value}` pairs
+
+        // TODO (doilij): return an array of  `${key}-${value}` pairs
```
jackhorton: Resolving this todo will eliminate the need for the ICU special case in `resolveLocaleHelper` in my pull

lib/Runtime/PlatformAgnostic/Chakra.Runtime.PlatformAgnostic.vcxproj
```diff
@@ -52,12 +52,19 @@
...
   <ItemGroup>
     <ClInclude Include="ChakraPlatform.h" />
     <ClInclude Include="EventTrace.h" />
     <ClInclude Include="PerfTrace.h" />
     <ClInclude Include="RuntimePlatformAgnosticPch.h" />
     <ClInclude Include="UnicodeText.h" />
+    <ClInclude Include="IPlatformAgnosticResource.h" />
+  </ItemGroup>
+  <ItemGroup Condition="'$(IntlICU)'=='true'">
+    <ClInclude Include="Intl.h" />
```
jackhorton: The naming here is a bit confusing: If IntlICU is true, we include Intl.h, but if IntlWinGlob (I know it doesn't exist but regardless) is true, we don't include Intl.h?

lib/Runtime/PlatformAgnostic/IPlatformAgnosticResource.h
```diff
@@ -0,0 +1,63 @@
...
+    template <typename T>
+    class StringBufferAutoPtr
+    {
+    private:
+        const T *strBuf;
+    public:
+        StringBufferAutoPtr(const T *buffer) : strBuf(buffer) {}
+        ~StringBufferAutoPtr()
+        {
+            delete[] strBuf;
```
jackhorton: I feel like these two classes should look exactly the same except `StringBufferAutoPtr` should be renamed to `AutoArrayPtr` and should use `delete[]` instead of `delete`

lib/Runtime/PlatformAgnostic/Intl.h
```diff
@@ -0,0 +1,70 @@
...
+
+    template <typename Func>
+    HRESULT CreateFormatter(Func function, _In_z_ const char16 *languageTag, _In_ const charcount_t cch, _Out_ IPlatformAgnosticResource **resource);
+    HRESULT CreateNumberFormatter(_In_z_ const char16 *languageTag, _In_ const charcount_t cch, _Out_ IPlatformAgnosticResource **resource);
+    HRESULT CreatePercentFormatter(_In_z_ const char16 *languageTag, _In_ const charcount_t cch, _Out_ IPlatformAgnosticResource **resource);
+    HRESULT CreateCurrencyFormatter(_In_z_ const char16 *languageTag, _In_ const charcount_t cch,
+        _In_z_ const char16 *currencyCode, _In_ const uint32 currencyDisplay, _Out_ IPlatformAgnosticResource **resource);
+
+    HRESULT SetNumberFormatSignificantDigits(IPlatformAgnosticResource *resource, const uint16 minSigDigits, const uint16 maxSigDigits);
+    HRESULT SetNumberFormatIntFracDigits(IPlatformAgnosticResource *resource, const uint16 minFracDigits, const uint16 maxFracDigits, const uint16 minIntDigits);
```
jackhorton: Is returning HRESULT from PlatformAgnostic a thing that's common, even though HRESULT is a very Windows-y thing as far as I know?

lib/Runtime/Library/IntlEngineInterfaceExtensionObject.cpp
```diff
@@ -818,26 +819,147 @@ namespace Js
             return scriptContext->GetLibrary()->GetUndefined();
         }
 
-#ifdef INTL_WINGLOB
+        DynamicObject *options = DynamicObject::FromVar(args.Values[1]);
+
+        HRESULT hr = S_OK;
+        Var propertyValue = nullptr; // set by the GetTypedPropertyBuiltInFrom macro
```
jackhorton: Not sure if this is a macro that you made or not but I found it super confusing to reason about, even with this comment.

lib/Runtime/Library/IntlEngineInterfaceExtensionObject.cpp
```diff
@@ -818,26 +819,147 @@ namespace Js
...
+        else
+        {
+            // Use the number formatter (0 or default)
+            IfFailThrowHr(PlatformAgnostic::Intl::CreateNumberFormatter(locale, cch, &numberFormatter));
+        }
+
+        Assert(numberFormatter);
+        // REVIEW (doilij): AutoPtr will call delete on IPlatformAgnosticResource and complain of non-virtual dtor. There are no IfFailThrowHr so is this still necessary?
+        // TODO (doilij): If necessary, introduce an PlatformAgnosticResourceAutoPtr that calls Cleanup() instead of delete on the pointer.
+        // PlatformAgnostic::Resource::AutoPtr<PlatformAgnostic::Resource::IPlatformAgnosticResource> numberFormatterGuard(numberFormatter);
```
jackhorton: Are these comments intentionally left in?

lib/Runtime/Library/IntlEngineInterfaceExtensionObject.cpp
```diff
@@ -818,26 +819,147 @@ namespace Js
...
+        // TODO (doilij): find out whether numberFormat->getLocale() has relevant extension tags for things like numeral system (-nu-)
+
+        if (HasPropertyBuiltInOn(options, __minimumSignificantDigits) || HasPropertyBuiltInOn(options, __maximumSignificantDigits))
+        {
+            // Do significant digit rounding
+            uint16 minSignificantDigits = 1, maxSignificantDigits = 21;
+
+            if (GetTypedPropertyBuiltInFrom(options, __minimumSignificantDigits, TaggedInt))
+            {
+                minSignificantDigits = max<uint16>(min<uint16>(TaggedInt::ToUInt16(propertyValue), 21), 1);
```
jackhorton: Will this be impacted by https://github.com/Microsoft/ChakraCore/pull/3799 ?

dilijev:
@jackhorton I'd imagine the way things are going that IcuIntlAdapter will go away in a refactoring (which TBH now is a good time to do). PlatformAgnostic::Intl now exists to hold everything which is ICU-specific. IntlEngineInterfaceExtensionObject.cpp is where all non-ICU (i.e. Runtime) logic should go, which includes getting JS values from the arguments passed into the entry points and as well as extracting the values stored on their properties, and any logic involving JS values. JS values should not be passed into PlatformAgnostic::Intl to keep the separation of ICU and Runtime logic.

Basically, there's no reason for IcuIntlAdapter now except that it used to exist.

I had a vague notion that it would be a good place to keep long logic that doesn't cleanly fit into PlatformAgnostic::Intl but can be abstracted out of IntlEngineInterfaceExtensionObject.cpp to keep that file shorter. However, this is not a useful abstraction anymore.

lib/Runtime/Base/WindowsGlobalizationAdapter.cpp
dilijev: Okay. Sounds like I'll probably need to resolve that when I rebase.

lib/Runtime/Library/IntlEngineInterfaceExtensionObject.cpp
dilijev: For review, yes -- I was hoping to discuss this design with @digitalinfinity and get his sign off before removing the code and comments (or otherwise updating the design).

lib/Runtime/Library/IntlEngineInterfaceExtensionObject.cpp
dilijev: Yes, I believe so. I'll take note, thanks!

lib/Runtime/Library/IntlEngineInterfaceExtensionObject.cpp
dilijev: Yes, I agree. I did not write this macro and the WinGlob code has super confusing logic surrounding it, so I didn't want to rewrite or change anything there at this time. I think this is worth a refactoring at some point.

lib/Runtime/PlatformAgnostic/Intl.h
dilijev: Yeah I agree it feels pretty dirty. The thing is `HRESULT` and `IfFailThrowHr` (which calls `MapAndThrowError(scriptContext, hr)`) is a really useful thing to have. I think the type isn't really a problem and we rely on `HRESULT` on xplat as well for `IfFailThrowHr` / `MapAndThrowError` (probably among other things).

lib/Runtime/Library/InJavascript/Intl.js
dilijev: Noted. I'll keep this TODO for future work.

lib/Runtime/PlatformAgnostic/Chakra.Runtime.PlatformAgnostic.vcxproj
dilijev: Right now `Intl.h` is only needed for `IntlICU` -- in the future we will probably want to move the `WindowsGlobalizationAdapter` into `PlatformAgnostic` and then have i18n-library-specific paths here. The ICU version is not platform-specific but is an implementation detail that can certainly be abstracted over using `PlatformAgnostic` and conditional options (in the same way we choose between platforms in PlatformAgnostic code). For this, I'd probably call the folders `Common` (for WinGlob -- or `Common.WinGlob`) and `Common.ICU` (as we already have `.ICU` as a way to switch over ICU or not ICU in `PlatformAgnostic`).

lib/Runtime/PlatformAgnostic/IPlatformAgnosticResource.h
dilijev: Yeah, basically this is AutoArrayPtr in general (already does `delete[]` instead of `delete`). I could update this but I'm kind of meh on that. There's kind of a logical abstraction between arrays and strings in general parlance even though they are both pointers in practice in C++.

lib/Runtime/PlatformAgnostic/Platform/Common/Intl.cpp
```diff
@@ -0,0 +1,487 @@
...
+            else
+            {
+                AssertMsg(false, "uloc_toLanguageTag: unexpected error (besides U_ILLEGAL_ARGUMENT_ERROR)");
+            }
+
+            goto LReturn;
+        }
+
+    LReturn:
+        return success;
```
jackhorton: I am assuming this used to have a `delete[]` that got removed with the addition of StringBufferAutoPtr, but it may be worth it to remove this label/goto path as a result.

lib/Runtime/PlatformAgnostic/Platform/Common/Intl.cpp
```diff
@@ -0,0 +1,487 @@
...
+        if (!success)
+        {
+            if (error == UErrorCode::U_ILLEGAL_ARGUMENT_ERROR)
+            {
+                AssertMsg(false, "uloc_toLanguageTag: error U_ILLEGAL_ARGUMENT_ERROR");
+            }
+            else
+            {
+                AssertMsg(false, "uloc_toLanguageTag: unexpected error (besides U_ILLEGAL_ARGUMENT_ERROR)");
+            }
```
jackhorton: We may want to consider making this a macro in the future, it shows up a lot

lib/Runtime/PlatformAgnostic/Platform/Common/Intl.cpp
dilijev: Yes, you're correct. Now no explicit cleanup is needed so we can change back to multiple returns instead of using goto/label. #Pending

lib/Runtime/PlatformAgnostic/Platform/Common/Intl.cpp
```diff
@@ -0,0 +1,485 @@
...
+        // explicitly zero the allocated array
+        for (size_t index = 0; index < inputLangTagUtf8SizeAllocated; ++index)
+        {
+            inputLangTagUtf8[index] = 0;
+        }
+
+        inputLangTagUtf8SizeActual = utf8::EncodeIntoAndNullTerminate(inputLangTagUtf8, languageTag, cch);
+
+        // Convert input language tag to a locale ID for use in uloc_toLanguageTag API.
+        forLangTagResultLength = uloc_forLanguageTag(reinterpret_cast<const char *>(inputLangTagUtf8),
```
boingoing: I'd like it if we didn't need to have this `reinterpret_cast` but I suppose `utf8::EncodeIntoAndNullTerminate` takes an `unsigned char*`, right? Can you leave out the `const` inside the cast, at least?

lib/Runtime/PlatformAgnostic/Platform/Common/Intl.cpp
```diff
@@ -0,0 +1,485 @@
...
+        int32_t parsedLength = 0;
+        int32_t forLangTagResultLength = 0;
+        int32_t toLangTagResultLength = 0;
+
+        // Allocate memory for the UTF8 output buffer. Need 3 bytes for each (code point + null) to satisfy SAL.
+        const size_t inputLangTagUtf8SizeAllocated = AllocSizeMath::Mul(AllocSizeMath::Add(cch, 1), 3);
+        unsigned char *inputLangTagUtf8 = new unsigned char[inputLangTagUtf8SizeAllocated];
+        StringBufferAutoPtr<unsigned char> guard(inputLangTagUtf8);
+
+        if (!inputLangTagUtf8)
```
boingoing: Do this check before we instantiate the `StringBufferAutoPtr`?

lib/Runtime/PlatformAgnostic/Platform/Common/Intl.cpp
```diff
@@ -0,0 +1,485 @@
...
+        {
+            AssertOrFailFastMsg(false, "OOM: failed to allocate inputLangTagUtf8.");
+            goto LReturn;
+        }
+
+        // explicitly zero the allocated array
+        for (size_t index = 0; index < inputLangTagUtf8SizeAllocated; ++index)
```
boingoing: Can we get away with using a `memset` or assert that `utf8::EncodeIntoAndNullTerminate` wrote a null byte at the end?

lib/Runtime/PlatformAgnostic/Platform/Common/Intl.cpp
```diff
@@ -0,0 +1,485 @@
...
+        }
+
+        // Try to convert icuLocaleId (locale ID version of input locale string) to BCP47 language tag, using strict checks
+        toLangTagResultLength = uloc_toLanguageTag(icuLocaleId, icuLangTag, ULOC_FULLNAME_CAPACITY, TRUE, &error);
+        success = toLangTagResultLength && U_SUCCESS(error);
+        if (!success)
+        {
+            if (error == UErrorCode::U_ILLEGAL_ARGUMENT_ERROR)
+            {
+                AssertMsg(false, "uloc_toLanguageTag: error U_ILLEGAL_ARGUMENT_ERROR");
```
boingoing: Are these really errors we cannot recover from? Are the arguments provided by user code? If not, I suppose this is probably fine. We might want to consider returning a more specific error code instead of just `E_INVALIDARG`, though.

lib/Runtime/PlatformAgnostic/Platform/Common/Intl.cpp
```diff
@@ -0,0 +1,485 @@
...
+            reinterpret_cast<const utf8char_t *>(icuLangTag), reinterpret_cast<utf8char_t *>(icuLangTag + toLangTagResultLength), utf8::doDefault);
+
+    LReturn:
+        return success ? S_OK : E_INVALIDARG;
+    }
+
+    int32_t GetCurrencyFractionDigits(_In_z_ const char16 * currencyCode)
+    {
+        UErrorCode error = UErrorCode::U_ZERO_ERROR;
+        const UChar *uCurrencyCode = reinterpret_cast<const UChar *>(currencyCode); // UChar, like char16, is guaranteed to be 2 bytes on all platforms.
```
boingoing: If it's guaranteed, can we use a static cast?

lib/Runtime/PlatformAgnostic/Platform/Common/Intl.cpp
```diff
@@ -0,0 +1,485 @@
...
+        // explicitly zero the allocated array
+        for (size_t index = 0; index < inputLangTagUtf8SizeAllocated; ++index)
+        {
+            inputLangTagUtf8[index] = 0;
+        }
+
+        utf8::EncodeIntoAndNullTerminate(inputLangTagUtf8, languageTag, cch);
+
+        char *localeName = reinterpret_cast<char *>(inputLangTagUtf8);
+        auto locale = icu::Locale::createFromName(localeName);
```
boingoing: Just a nitpick but I'm not a big fan of `auto`.

lib/Runtime/PlatformAgnostic/Platform/Common/Intl.cpp
```diff
@@ -0,0 +1,485 @@
...
+    }
+
+    HRESULT CreateCurrencyFormatter(_In_z_ const char16 *languageTag, _In_ const charcount_t cch,
+        _In_z_ const char16 *currencyCode, _In_ const uint32 currencyDisplay, _Out_ IPlatformAgnosticResource **resource)
+    {
+        UErrorCode error = UErrorCode::U_ZERO_ERROR;
+
+        unsigned char *inputLangTagUtf8 = nullptr;
+
+        {
```
boingoing: Intentional block?

lib/Runtime/PlatformAgnostic/Platform/Common/Intl.cpp
```diff
@@ -0,0 +1,485 @@
...
+        if (!formatterHolder)
+        {
+            AssertOrFailFastMsg(formatterHolder, "Formatter did not hold an object of type `FinalizableIntlObject<icu::NumberFormat>`");
+            return nullptr;
+        }
+
+        icu::NumberFormat *numberFormatter = formatterHolder->GetInstance();
+
+        // __formatterToUse: 0 (default) is number, 1 is percent, 2 is currency
+        if (formatterToUse == 0 || formatterToUse == 1)
```
boingoing: Is `formatterToUse` only used within ChakraCore? I don't see a reason not to use an enum for it. 👍

lib/Runtime/PlatformAgnostic/Platform/Common/Intl.cpp
```diff
@@ -0,0 +1,485 @@
...
+        {
+            numberFormatter->format(val, result);
+        }
+        else if (formatterToUse == 2)
+        {
+            UErrorCode error = UErrorCode::U_ZERO_ERROR;
+
+            // const UChar *currencyCode = reinterpret_cast<const UChar *>(numberFormatter->getCurrency());
```
boingoing: Remove this line?

lib/Runtime/PlatformAgnostic/Platform/Common/Intl.cpp
```diff
@@ -0,0 +1,485 @@
...
+            if (U_FAILURE(error))
+            {
+                AssertMsg(false, "numberFormatter->getLocale failed");
+            }
+
+            UBool isChoiceFormat = false;
+            int32_t currencyNameLen = 0;
+
+            // __currencyDisplayToUse: 0 (default) is symbol, 1 is code, 2 is name
+            if (currencyDisplay == 0) // symbol ($42.00)
```
boingoing: Replace magic numbers with an enum?

lib/Runtime/PlatformAgnostic/Platform/Common/Intl.cpp
jackhorton: All of the UTF8 stuff seems to work in terms of utf8char_t, which is typedef'd from BYTE, which is typedef'd from unsigned char. I don't know what our default warnings are but I think that could cause some (I know signed/unsigned comparison is a gcc/clang warning that could come up)

lib/Runtime/PlatformAgnostic/Platform/Common/Intl.cpp
jackhorton: For these specific errors, I could see us not wanting to assert in the case of an illegal argument, which may simply mean that the input tag couldn't be normalized. That's an expected error and probably shouldn't stop a debug build from running. With that being said, the only "recovery" would be to propagate the error back to the user and let them decide what to do with an un-normalizable langtag.

lib/Runtime/PlatformAgnostic/Platform/Common/Intl.cpp
```diff
@@ -0,0 +1,485 @@
...
+        StringBufferAutoPtr<unsigned char> guard(inputLangTagUtf8);
+
+        if (!inputLangTagUtf8)
+        {
+            AssertOrFailFastMsg(false, "OOM: failed to allocate inputLangTagUtf8");
+            goto LReturn;
+        }
+
+        // explicitly zero the allocated array
+        for (size_t index = 0; index < inputLangTagUtf8SizeAllocated; ++index)
```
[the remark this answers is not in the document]
dilijev: No -- that would be a smart thing to do :P

lib/Runtime/PlatformAgnostic/Platform/Common/Intl.cpp
dilijev: I think this cast is necessary to some degree because of the way the types are set up in existing code, yeah. Not sure I see why leaving off the `const` is an improvement?

lib/Runtime/PlatformAgnostic/Platform/Common/Intl.cpp
dilijev: Also, the logic should already have taken care of that situation in a spec-complaint way. This Assert is because that sort of issue shouldn't get to this point.

lib/Runtime/PlatformAgnostic/Platform/Common/Intl.cpp
dilijev: The whole check-if-else block? Yeah maybe.

lib/Runtime/PlatformAgnostic/Platform/Common/Intl.cpp
dilijev: For some reason I thought `static_cast` has some runtime overhead. Otherwise, yeah I can do that.

lib/Runtime/PlatformAgnostic/Platform/Common/Intl.cpp
dilijev: Some of this code was more readable with `auto` -- this one is debatable. It's an `icu::Locale` and the type is on the RHS so I thought maybe. But yeah, I don't like this one either. I'll change it.

lib/Runtime/PlatformAgnostic/Platform/Common/Intl.cpp
```diff
@@ -0,0 +1,486 @@
...
+            {
+                AssertOrFailFastMsg(false, "OOM: failed to allocate inputLangTagUtf8.");
+                return E_OUTOFMEMORY;
+            }
+
+            // explicitly zero the allocated array
+            for (size_t index = 0; index < inputLangTagUtf8SizeAllocated; ++index)
+            {
+                inputLangTagUtf8[index] = 0;
+            }
```
dilijev: Note: use memset here too

lib/Runtime/PlatformAgnostic/Platform/Common/Intl.cpp
dilijev: Yeah, whoops

lib/Runtime/PlatformAgnostic/Platform/Common/Intl.cpp
dilijev: Will do

lib/Runtime/PlatformAgnostic/Platform/Common/Intl.cpp
dilijev: Yeah I had some dumb idea about limiting variable scope (in this case only inputLangTagUtf8SizeAllocated). I might have had an AutoPtr in mind at some point. Didn't use it. I'll remove the block.

lib/Runtime/PlatformAgnostic/Platform/Common/Intl.cpp
dilijev: These numbers are also used on the JS side -- the effect was that it moved parsing String options to the JS side and made it a simple comparison on the native side as an int.

I can mirror an enum (or constants since JS doesn't really have enums) over to the JS as well so that we stop leaving these magic constants with comments everywhere :)

lib/Runtime/PlatformAgnostic/Platform/Common/Intl.cpp
dilijev: (These numbers are also used on the JS side -- will mirror the enum over there too.)

lib/Runtime/PlatformAgnostic/Platform/Common/Intl.cpp
jackhorton: Shorter line length? That's why I have been leaving them off.

dilijev: Ah, yeah. I'm adding a `const` -- no need for that, the function call does it. I was worried in general about dropping a const, although IIRC I think casting away `const` is the only thing `reinterpret_cast` won't do compared to a C-style cast, and that's a good thing :)

Anyway I'll drop `const` here and in similar places.

lib/Runtime/PlatformAgnostic/IPlatformAgnosticResource.h
```diff
@@ -0,0 +1,63 @@
+//-------------------------------------------------------------------------------------------------------
+// Copyright (C) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE.txt file in the project root for full license information.
+//-------------------------------------------------------------------------------------------------------
+
+#ifndef RUNTIME_PLATFORM_AGNOSTIC_COMMON_IFINALIZABLE
```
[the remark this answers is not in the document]
dilijev: Yes but for some reason in PlatformAgnostic we use this pattern instead in many instances. Might be a mistake. I'll ask around.

---
In reply to: [141529369](https://github.com/Microsoft/ChakraCore/pull/3809#discussion_r141529369) [](ancestors = 141529369)

lib/Runtime/PlatformAgnostic/Platform/Common/Intl.cpp
dilijev: Marking WontFix for now because I don't want to undertake any superficial refactorings at this point.

---
In reply to: [141681648](https://github.com/Microsoft/ChakraCore/pull/3809#discussion_r141681648) [](ancestors = 141681648)

lib/Runtime/PlatformAgnostic/Platform/Common/Intl.cpp
dilijev: No runtime overhead, but because on Windows char16 reduces to the native wchar_t and UChar to char16_t.
The key element here is that they are not the same native type and wchar_t may be wider on other platforms (on linux it's 4 bytes, maybe elsewhere too).
Because of typedefs in the strategic places we know that ultimately on any platform char16 will be 2 bytes, but the compiler can't tell that.

I added an explainer to the source.

---
In reply to: [141681856](https://github.com/Microsoft/ChakraCore/pull/3809#discussion_r141681856) [](ancestors = 141681856)

dilijev:
I'm not going to make this change until both of our PRs are merged so that we don't disrupt each other here.

---
In reply to: [332697924](https://github.com/Microsoft/ChakraCore/pull/3809#issuecomment-332697924) [](ancestors = 332697924)

lib/Runtime/Library/IntlEngineInterfaceExtensionObject.cpp
dilijev: Looks like that was updated in such a way that there will be no change here. We'll find out when everything is merged.

---
In reply to: [141507935](https://github.com/Microsoft/ChakraCore/pull/3809#discussion_r141507935) [](ancestors = 141507935)

lib/Runtime/PlatformAgnostic/Chakra.Runtime.PlatformAgnostic.vcxproj
dilijev: Marking as WontFix for now. Will clean up in future iterations.

---
In reply to: [141508775](https://github.com/Microsoft/ChakraCore/pull/3809#discussion_r141508775) [](ancestors = 141508775)

lib/Runtime/PlatformAgnostic/Intl.h
dilijev: Might update in the future along with other refactorings.

---
In reply to: [141508287](https://github.com/Microsoft/ChakraCore/pull/3809#discussion_r141508287) [](ancestors = 141508287)

lib/Runtime/PlatformAgnostic/IPlatformAgnosticResource.h
dilijev: I'm updating to check for nullptr first. Otherwise leaving alone.

---
In reply to: [141508964](https://github.com/Microsoft/ChakraCore/pull/3809#discussion_r141508964) [](ancestors = 141508964)

PR #3812, opened 2017-09-26
Use Unicode Windows API to print ChakraCore version (fix #3178)

### LINKED ISSUE #3178: ch.exe -v doesn't work correctly if path contains non-ascii characters

xiaoyinl:
This PR adds a new WIN32-only function GetChakraDllNameW() to ChakraRtInterface.cpp, which returns wide string "ChakraCore.dll". In ch.cpp, the function PrintChakraCoreVersion() uses GetChakraDllNameW() and Unicode Windows APIs to obtain the file version of ChakraCore.dll, so that it will work even if the pathname contains non-ASCII characters.

Fixes #3178.

unicodetest

### DIFF
--- a/bin/ch/ChakraRtInterface.cpp
+++ b/bin/ch/ChakraRtInterface.cpp
@@ -6,6 +6,7 @@
 
 #ifdef _WIN32
 LPCSTR chakraDllName = "chakracore.dll";
+LPCWSTR chakraDllNameW = _u("chakracore.dll");
 #else
 #include <dlfcn.h>
 #ifdef __APPLE__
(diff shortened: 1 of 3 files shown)


bin/ch/ChakraRtInterface.cpp
```diff
@@ -23,10 +24,16 @@ ChakraRTInterface::ArgInfo* ChakraRTInterface::m_argInfo = nullptr;
 TestHooks ChakraRTInterface::m_testHooks = { 0 };
 JsAPIHooks ChakraRTInterface::m_jsApiHooks = { 0 };
 
-LPCSTR GetChakraDllName()
+LPCSTR GetChakraDllName()
+{
+	return chakraDllName;
+}
```
agarwal-sandeep: There s no usage of this now. Remove it.

xiaoyinl:

bin/ch/ChakraRtInterface.cpp
xiaoyinl: @agarwal-sandeep Done. Thanks!

dilijev:
Testing internally.

PR #3813, opened 2017-09-26
Fix #3040: Incorrect parsing of class expression

### LINKED ISSUE #3040: Incorrect parsing of class expression
```js
1,class extends[]/print(1){}
```

Expected: Syntax error unexpected token `/`
Actual: prints 1, complains about the class not having a valid prototype value.

tcare:
Class extends exprs should only be LHS exprs. Change to ParseTerm instead of blanket ParseExpr.


tcare:

PR #3815, opened 2017-09-26
ChakraFull ARM64 build support

agarwal-sandeep:
Enable ChakraFull to compile for ARM64 using VS preview builds

### DIFF
--- a/bin/GCStress/GCStress.vcxproj
+++ b/bin/GCStress/GCStress.vcxproj
@@ -33,7 +33,7 @@
         $(ChakraCommonLinkDependencies);
         Advapi32.lib;
         %(AdditionalDependencies)</AdditionalDependencies>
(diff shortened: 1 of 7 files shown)


bin/GCStress/GCStress.vcxproj
```diff
@@ -33,7 +33,7 @@
         $(ChakraCommonLinkDependencies);
         Advapi32.lib;
         %(AdditionalDependencies)</AdditionalDependencies>
-      <AdditionalLibraryDirectories>$(OutDir);%(AdditionalLibraryDirectories)</AdditionalLibraryDirectories>
+      <AdditionalLibraryDirectories>$(OutDir);%(AdditionalLibraryDirectories);$(SdkLibPath)</AdditionalLibraryDirectories>
```
tcare: Curious, what are the dependencies that require this path?

agarwal-sandeep: Somehow for ARM64 Advapi32.lib was not picked from correct location.

PR #3817, opened 2017-09-27
ARM64 Encoder Initial Commit (Thanks to Aaron Giles <aarongi@microsoft.com>)

agarwal-sandeep:


lib/Backend/arm64/md.h
```diff
@@ -17,25 +17,27 @@
 const int MachChar = 1;
 const int MachShort = 2;
 const int MachInt = 4;
-const int MachRegInt = 8;
-__declspec(selectany) const int MachPtr = 8;
+const int MachRegInt = 4;
+__declspec(selectany) const int MachPtr = 4;
```
Cellule: Registers are 32 bits in Arm64 ?

agarwal-sandeep: Nope 64 bit.
FYI that this code only compiles and doesn't run yet so there might be bugs.

sigatrev:
LGTM

PR #3825, opened 2017-09-28
fix some errors in stack traces with inlining

sigatrev:
This commit fixes two issues with stack traces matching the wrong statements with inlined functions.

### DIFF
--- a/lib/Runtime/Base/FunctionBody.cpp
+++ b/lib/Runtime/Base/FunctionBody.cpp
@@ -3365,6 +3365,15 @@ namespace Js
         SmallSpanSequence *spanSequence = entryPoint ? entryPoint->GetNativeThrowSpanSequence() : nullptr;
         int statementIndex = GetStatementIndexFromNativeOffset(spanSequence, offset);
 
+        // the offset passed in here is the offset of the Inlinee Start instruction. In some cases, the first statement of the inlinee may be mapped at that offset and would be matched 
+        // instead of matching the statement that made the call to the inlinee. In that case, to find the statement associated with the call to the inlinee we need to look earlier.
+        // We cannot alawys use the lower offset because for inlinees of inlinees the call statement and the inlinee start statement may occur at the same offset, in which case the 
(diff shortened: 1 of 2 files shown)


lib/Runtime/Base/FunctionBody.cpp
```diff
@@ -3365,6 +3365,15 @@ namespace Js
         SmallSpanSequence *spanSequence = entryPoint ? entryPoint->GetNativeThrowSpanSequence() : nullptr;
         int statementIndex = GetStatementIndexFromNativeOffset(spanSequence, offset);
 
+        // the offset passed in here is the offset of the Inlinee Start instruction. In some cases, the first statement of the inlinee may be mapped at that offset and would be matched 
+        // instead of matching the statement that made the call to the inlinee. In that case, to find the statement associated with the call to the inlinee we need to look earlier.
+        // We cannot alawys use the lower offset because for inlinees of inlinees the call statement and the inlinee start statement may occur at the same offset, in which case the 
+        // call statement is the one that gets matched and looking earlier would result in matching a statement from the enclosing function.
+        if (statementIndex == 0)
+        {
+            statementIndex = GetStatementIndexFromNativeOffset(spanSequence, offset - 1);
```
agarwal-sandeep: >statementIndex [](start = 12, length = 14)

Can you add test case?

agarwal-sandeep: >offset [](start = 77, length = 6)

offset is guaranteed to be > 0?

sigatrev: the only caller checks that the offset it is non-zero. https://github.com/Microsoft/ChakraCore/blob/master/lib/Runtime/Language/JavascriptStackWalker.cpp#L459

agarwal-sandeep: Add a assert here

---
In reply to: [141506849](https://github.com/Microsoft/ChakraCore/pull/3825#discussion_r141506849) [](ancestors = 141506849)

PR #3835, opened 2017-09-28
Fix null dereference in LoadAllConstants when code has no 'this' binding

boingoing:
Fixes https://microsoft.visualstudio.com/DefaultCollection/OS/_workitems?id=13977382&_a=edit


akroshg:
Did you find out the scenario? good to add in our unittest.

boingoing:
@akroshg It's the case where an event handler doesn't have a `this` binding. We have test coverage for this, but not in ChakraCore.

akroshg:
LGTM

PR #3838, opened 2017-09-28
OS#13885461 WASM: Jit loop body yield restore

Cellule:
Correctly restore yielded values from jit loop bodies.
Use `Return_****` opcodes when yielding.
Made restoring loop values more generic to include `const` (aka return) and `tmps` (aka yielded values)
Some cleanup in IRBuilderAsmJs

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/3838)

### DIFF
--- a/lib/WasmReader/WasmByteCodeGenerator.cpp
+++ b/lib/WasmReader/WasmByteCodeGenerator.cpp
@@ -1600,7 +1600,7 @@ void WasmBytecodeGenerator::YieldToBlock(BlockInfo blockInfo, EmitInfo expr)
         if (!IsUnreachable())
         {
             blockInfo.yieldInfo->didYield = true;
-            m_writer->AsmReg2(GetLoadOp(expr.type), yieldInfo.location, expr.location);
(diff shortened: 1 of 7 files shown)


lib/WasmReader/WasmByteCodeGenerator.cpp
```diff
@@ -1600,7 +1600,7 @@ void WasmBytecodeGenerator::YieldToBlock(BlockInfo blockInfo, EmitInfo expr)
         if (!IsUnreachable())
         {
             blockInfo.yieldInfo->didYield = true;
-            m_writer->AsmReg2(GetLoadOp(expr.type), yieldInfo.location, expr.location);
+            m_writer->AsmReg2(GetReturnOp(expr.type), yieldInfo.location, expr.location);
```
Cellule: I considered making a `Yield_***` opcode, but I was able to get the same result with `Return_***` might need to revisit when we'll do multi-return

Cellule:
Ping

Cellule:
ping

PR #3841, opened 2017-09-29
fix ProfileModeDeferredParsingThunk for arm64

sigatrev:


lib/Runtime/Language/arm64/arm64_Thunks.asm
```diff
@@ -88,6 +88,7 @@
     stp   x6, x7, [sp, #64]
 
     mov  x0, sp                   ; Pass the address of the function at the saved x0 in case it need to be boxed
+    add  x0, x0, #16
```
agarwal-sandeep: Why?

sigatrev: I tried this because it was what was done by DeferredParsingThunk, and it fixed the crash happened whenever ProfileModeDeferredParsingThunk was called. 

Looking at the disassembly of compiled arm64 functions in windbg, they start with
`sub         sp,sp,#0x10`, so adding 16 would provide the address of the start of the function.

agarwal-sandeep: Add that as a comment here so that's its easy for others to understand it later. 

In general it will be good practice to write comments for change we make for ARM64 so that people who later add additional code can understand easily.

PR #3844, opened 2017-09-30
Support int64 type in ChangeToAssign and InsertMove

Cellule:
Move check for int64 assign into the relevant place to minimize burden when trying to move an int64 on x86.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/3844)


Cellule:
@Microsoft/chakra-jit Can someone review this please ?

PR #3845, opened 2017-09-30
CanonicalizeAccessor returns wrong results on cross thread

obastemur:
`accessor` might be initialized under a different context hence `library->Undefined != accessor's->library->Undefined`

Recently added an Assert to IsUndefinedObject** checks. That assert helped revealing this issue.

See also https://github.com/Microsoft/ChakraCore/blame/master/lib/Runtime/Types/DictionaryTypeHandler.cpp#L2020


obastemur:
Related to OS14010656

obastemur:
@agarwal-sandeep @Penguinwizzard Thanks for the reviews

PR #3853, opened 2017-10-02
Correct exception specification mismatch in Memory::Recycler::AddMark

### LINKED ISSUE #3862: Exception specification mismatches will cause build breaks in new versions of VS 2017
Memory::Recycler::AddMark has a mismatched exception specification between the declaration and definition. Starting with version 19.12 of MSVC this will be diagnosed under /permissive- or /std:c++17.

I've submitted PR #3853 to correct this, but I note that none of the other declarations in this class have exception specifications, so it may be preferable to remove throw() from the definition instead of adding it to the declaration. I can make that change if you prefer.

joemmett:
Memory::Recycler::AddMark() is declared with no exception specification
but defined as `throw()`. Upcoming versions of MSVC will diagnose this
mismatch in /permissive- or C++17 modes. Add `throw()` to the
declaration to bring the declaration in line with the definition.

Fixes #3862

### DIFF
--- a/lib/Common/Memory/Recycler.h
+++ b/lib/Common/Memory/Recycler.h
@@ -1574,9 +1574,9 @@ private:
     inline void ScanMemoryInline(void ** obj, size_t byteCount);
     template <bool doSpecialMark>
     void ScanMemory(void ** obj, size_t byteCount) { if (byteCount != 0) { ScanMemoryInline<doSpecialMark>(obj, byteCount); } }
-    bool AddMark(void * candidate, size_t byteCount);
+    bool AddMark(void * candidate, size_t byteCount) throw();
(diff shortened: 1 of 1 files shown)


Penguinwizzard:
Could you rebase this PR on top of master? We have fixes for the CI issues that this PR has hit in master.

joemmett:
Rebased, thanks!

dilijev:
@joemmett Sorry for the delay -- could you rebase and resolve the conflict? Then I'll get this merged.

joemmett:
Rebased, but it looks like the newly added AddPreciselyTracedMark() function will have exactly the same issue. I've updated the PR to change this signature as well.

dilijev:
@joemmett Thanks!

PR #3855, opened 2017-10-03
Remove support for VS2013 (dev12)

Penguinwizzard:
This change removes support for Visual Studio 2013. Doing this will
afford us the use of a much wider set of constructs in C++, as more
recent versions of Visual Studio provide full support for C++11.

I'm going to leave this PR open for a week (at least) to invite any
comments about the change that may be out there.

Note: Although this change is to move to support C++11, we could be
a C++14 code base with little extra effort; our minimum versions of
clang (3.9) and VS (dev14) are both sufficient for C++14 support. I
think the only thing holding that back is that we may want to allow
a 4.9 version of GCC.

### DIFF
--- a/Build/Chakra.Core.sln
+++ b/Build/Chakra.Core.sln
@@ -2,7 +2,7 @@
 Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 15
 VisualStudioVersion = 15.0.26228.4
-MinimumVisualStudioVersion = 10.0.40219.1
+MinimumVisualStudioVersion = 14.0.00000.0
 Project("{8BC9CEB8-8B4A-11D0-8D11-00A0C91BC942}") = "ChakraCore", "..\bin\ChakraCore\ChakraCore.vcxproj", "{EA882C8D-81FC-42FE-ABD5-2666DB933FDB}"
 	ProjectSection(ProjectDependencies) = postProject
 		{1876E800-AD77-48C4-A2F7-E5265F24AC38} = {1876E800-AD77-48C4-A2F7-E5265F24AC38}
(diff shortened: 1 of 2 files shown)


Cellule:
Don't forget to remove the CI trigger definitions for dev12.
@dilijev Is it possible to keep the dev12 build only for older releases or should we just drop it completely ?

MSLaguana:
The CI builds are defined based on the netci.groovy script in that branch. If we leave that target there in an older branch, then (as long as we still care to track that branch, which is defined elsewhere) it'll keep the definition and run it when PRs are made.

Penguinwizzard:
I've added another commit to remove dev12 from the netci.groovy file. Remaining failures will be resolved when CI fix PR #3854 gets pumped up and this PR gets rebased.

dilijev:
Do we still want a Win7 build with Dev14 (since we still claim to support building and running on Win7)?

Should we make Dev14 on Server 2012 R2 the legacy build and move to Dev15 on Server 2016 for the primary Windows CI checks?

Penguinwizzard:
We do want a Dev14 build; I'm not sure about whether Win7 or Server 2012 R2 is more used in the community though.

dilijev:
Yes, I may have misspoken. For our purposes:

1. Server 2008 R2 ~= Windows 7
2. Server 2012 R2 ~= Windows 8.1
3. Server 2016 = Windows 10 (unsure of a build correspondence)

Legacy builds used (Dev12 on 2008 R2)
Current builds use (Dev14 on 2012 R2)

We could update so that:

* Current builds use (Dev15 on 2016)
* "Compat" builds (reduced matrix) on (Dev14 on 2012 R2) -- or (Dev14 + 2016)
* Legacy = (Dev14 on 2008 R2)

Or something similar.

Basically we want to support the following configs with a minimal matrix: { Dev15, Dev14 } x { Server 2016, 2012 R2, 2008 R2 }

PR #3856, opened 2017-10-03
Fixing an issue with duplicate let definition of a destructured catch param

aneeshdk:
The destructured catch param are let bindings. Any duplicate definition of them inside the catch block is an error. So when assigning registers we don't have to look for duplicate symbols in the function body.


aneeshdk:
@akroshg @pleath Can you please take a look?

PR #3857, opened 2017-10-03
Fix issue with calling JSON.stringify on an object with internal properties

jackhorton:


digitalinfinity:
Not sure what’s up with the CI. Unit test probably needs a copyright?

jackhorton:
Added the copyright, sorry about that. The prefast warnings on the Windows CI is something we have been seeing for a few days now, @rajatd mentioned that that code hasn't been touched in quite a while. Ubuntu CI seemed to just have a bit of a fit. Retesting now.

MSLaguana:

obastemur:
@jackhorton good catch and Thanks for the test case!

test/JSON/stringifyobjects.js
```diff
@@ -0,0 +1,20 @@
...
+if (this.WScript && this.WScript.LoadScriptFile) { // Check for running in ch
+    this.WScript.LoadScriptFile("..\\UnitTestFramework\\UnitTestFramework.js");
+}
+
+const tests = [
+    {
+        name: "Stringify error",
+        body: function () {
+            const err = new Error("message");
+            assert.areEqual('{"description":"message","message":"message"}', JSON.stringify(err));
```
akroshg: does it make sense to call stringify on every object chakra can produce?

MSLaguana:

MSLaguana:

MSLaguana:

test/JSON/stringifyobjects.js
jackhorton: Possibly? I took a quick look through all of our JSON tests to try to find a place to put this one, and it looked like every other JSON test we have JSON.parse()s a string and then JSON.stringify()s it to make sure the strings look similar. One possible downside of adding exhaustive tests here are because they depend on the order of enumeration, if we ever change that then we would have go go through a ton of tests to update it as well. This is a problem we have with a lot of node-chakracore tests right now, where many tests they add fail because they depend on V8's current enumeration ordering for output.

PR #3858, opened 2017-10-03
Preparing release of 1.7.3

MSLaguana:


MSLaguana:
Need to wait until after #3857 before merging this in.

MSLaguana:
We're going to back this out for now; too much churn in the past few days, especially with CI being a bit wonky.

dilijev:
In general you wouldn't need to close a PR like this due to churn in the target branch. The CI will run against `merge((PR), (latest commit from the target branch))` at the time the tests are triggered. You wouldn't even need to rebase. You can simply re-test when ready before merging.

PR #3861, opened 2017-10-03
jsrt: JsConvertValueToString exit fast if arg. is JavascriptString

obastemur:
Entering into ContextAPIWrapper is expensive. node-chakracore, (epscially acme air app)
majority of the callers triggers this method with JavascriptString.

### DIFF
--- a/lib/Jsrt/Jsrt.cpp
+++ b/lib/Jsrt/Jsrt.cpp
@@ -1215,12 +1215,22 @@ CHAKRA_API JsStringToPointer(_In_ JsValueRef stringValue, _Outptr_result_buffer_
 
 CHAKRA_API JsConvertValueToString(_In_ JsValueRef value, _Out_ JsValueRef *result)
 {
+    PARAM_NOT_NULL(result);
+    *result = nullptr;
+
+    if (value != nullptr && Js::JavascriptString::Is(value))
+    {
+        return ContextAPINoScriptWrapper_NoRecord([&](Js::ScriptContext *scriptContext) -> JsErrorCode {
+            VALIDATE_INCOMING_REFERENCE(value, scriptContext);
+
+            *result = value;
+            return JsNoError;
+        });
+    }
+
     return ContextAPIWrapper<JSRT_MAYBE_TRUE>([&] (Js::ScriptContext *scriptContext, TTDRecorder& _actionEntryPopper) -> JsErrorCode {
         PERFORM_JSRT_TTD_RECORD_ACTION(scriptContext, RecordJsRTVarToStringConversion, (Js::Var)value);
-
         VALIDATE_INCOMING_REFERENCE(value, scriptContext);
-        PARAM_NOT_NULL(result);
-        *result = nullptr;
 
         *result = (JsValueRef) Js::JavascriptConversion::ToString((Js::Var)value, scriptContext);
 


lib/Jsrt/Jsrt.cpp
```diff
@@ -1215,6 +1215,12 @@ CHAKRA_API JsStringToPointer(_In_ JsValueRef stringValue, _Outptr_result_buffer_
 
 CHAKRA_API JsConvertValueToString(_In_ JsValueRef value, _Out_ JsValueRef *result)
 {
+    if (value != nullptr && Js::JavascriptString::Is(value))
+    {
+        *result = value;
+        return JsNoError;
```
MSLaguana: Does this have any TTD impact? I suppose since this case explicitly doesn't have side effects, it should probably be fine.

lib/Jsrt/Jsrt.cpp
```diff
@@ -1215,6 +1215,12 @@ CHAKRA_API JsStringToPointer(_In_ JsValueRef stringValue, _Outptr_result_buffer_
 
 CHAKRA_API JsConvertValueToString(_In_ JsValueRef value, _Out_ JsValueRef *result)
 {
+    if (value != nullptr && Js::JavascriptString::Is(value))
```
MikeHolman: i'd think we need to also ensure that ScriptContext for value is the same as the current context?

MikeHolman:
I also noticed this has very high overhead. is there anything we can do to improve it?

lib/Jsrt/Jsrt.cpp
obastemur: done.

obastemur:
> I also noticed ContextAPIWrapper has very high overhead. is there anything we can do to improve it?

So far I did multiple attempts there (especially on ChakraCore end) i.e. (added `JSRT_MAYBE_TRUE` to skip some really expensive and unnecessary check) Also some hot interface (on node-chakracore) now have specialized conditions before entering into ContextAPIWrapper.

`JsConvertValueToString` was pretty hot on node-chakracore `acme-air` runs and couldn't resist doing some improvement. However, I'm pretty sure, it won't change anything on benchmarks.

We could improve ContextAPIWrapper but not sure the `measurable` potential gain on the current goals.

lib/Jsrt/Jsrt.cpp
```diff
@@ -1215,6 +1215,15 @@ CHAKRA_API JsStringToPointer(_In_ JsValueRef stringValue, _Outptr_result_buffer_
...
+    {
+        return ContextAPINoScriptWrapper_NoRecord([&](Js::ScriptContext *scriptContext) -> JsErrorCode {
+            *result = Js::CrossSite::MarshalVar(scriptContext, value,
```
akroshg: >result [](start = 13, length = 6)

result can be null, it will crash in that case?

MSLaguana: We should move the `PARAM_NOT_NULL(result);` from inside the `ContextAPIWrapper` to up front.

MSLaguana: We probably also want the `VALIDATE_INCOMING_REFERENCE(value, scriptContext)` from the next wrapper in here?

obastemur: > `result` can be null..

`PARAM_NOT_NULL(result)` check.. well, I'm okay adding all sort of those checks but what purpose this method has if the `result` argument is nullptr? Well, isn't that a programatical mistake?

MSLaguana: Yes it is, and the `PARAM_NOT_NULL` check is what reports that back to the embedder calling the function.

MSLaguana: Note that the check was already there originally, it was just inside the context wrapper.

akroshg: You don't want null point exception in that case right?

---
In reply to: [142538359](https://github.com/Microsoft/ChakraCore/pull/3861#discussion_r142538359) [](ancestors = 142538359)

obastemur: I know that. My point was, some stuff inside the JSRT.cpp feels like they are there because of copy-paste from another API method... This particular interface, what is difference between returning `hey result is null, so you called this interface with something you definitely shouldn't, probably you have something terribly wrong` vs `crash`) because, whatever the application logic is.. something is wrong.

Imagine..

scenario A)
```
JsValueRef result;
JsConvertValueToString(Some_JS_Value, &result)
```

if `&result is null?`... 

scenario b)
```
JsConvertValueToString(Some_JS_Value, nullptr)
```

why do you call this interface then?

scenario c)
```
JsConvertValueToString(Some_JS_Value, &someClass->result)
```

so... `someClass` is probably dead? It will crash anyway?

I will apply the checks..

jackhorton: I would suggest that crashing with a reason is always more helpful than crashing without a reason. If there is something wrong with an embedder's logic, we should at least do simple checks like this to let them know that the crash is their fault, not ours.

lib/Jsrt/Jsrt.cpp
```diff
@@ -1215,12 +1215,22 @@ CHAKRA_API JsStringToPointer(_In_ JsValueRef stringValue, _Outptr_result_buffer_
...
 CHAKRA_API JsConvertValueToString(_In_ JsValueRef value, _Out_ JsValueRef *result)
 {
+    PARAM_NOT_NULL(result);
+    *result = nullptr;
+
+    if (value != nullptr && Js::JavascriptString::Is(value))
+    {
+        return ContextAPINoScriptWrapper_NoRecord([&](Js::ScriptContext *scriptContext) -> JsErrorCode {
+            VALIDATE_INCOMING_REFERENCE(value, scriptContext);
+            *result = Js::CrossSite::MarshalVar(scriptContext, value,
```
MSLaguana: I believe that this isn't required after the `VALIDATE_INCOMING_REFERENCE`, it does some marshaling on the `value` parameter already. We should just be able to do `*result = value` here.

obastemur: Well, actually `VALIDATE_INCOMING_REFERENCE` totally unnecessary here.

obastemur: It enters here only when `Js::JavascriptString::Is` `true`

MSLaguana: `VALIDATE_INCOMING_REFERENCE` also does some other checks too though; I'm not sure if they are strictly required here, but it checked that it was the same runtime as well I think. Are you sure that it isn't necessary here?

obastemur:
```
#define VALIDATE_INCOMING_OBJECT(p, scriptContext) \
        { \
            VALIDATE_JSREF(p); \
            if (!Js::JavascriptOperators::IsObject(p)) \
            { \
                return JsErrorArgumentNotObject; \
            } \
            MARSHAL_OBJECT(p, scriptContext)   \
        }
```

```
#define VALIDATE_JSREF(p) \
    if (p == JS_INVALID_REFERENCE) \
    { \
        return JsErrorInvalidArgument; \
    } \
```

>  it checked that it was the same runtime as well I think.

?

Well, as I mentioned, we do `!nullptr && Js::JavascriptString::Is` check.. and Marshal but we don't do `JS_INVALID_REFERENCE` check.

```
#ifdef __cplusplus
    const JsRef JS_INVALID_REFERENCE = 0;
#else
    #define JS_INVALID_REFERENCE (JsRef)0
#endif
```

MSLaguana:
```cpp
#define MARSHAL_OBJECT(p, scriptContext) \
        Js::RecyclableObject* __obj = Js::RecyclableObject::FromVar(p); \
        if (__obj->GetScriptContext() != scriptContext) \
        { \
            if(__obj->GetScriptContext()->GetThreadContext() != scriptContext->GetThreadContext()) \
            {  \
                return JsErrorWrongRuntime;  \
            }  \
            p = Js::CrossSite::MarshalVar(scriptContext, __obj); \
}
```

That's the part that I'm referring to; the comparison of the thread context. I'm not familiar with what the consequences of ignoring that are.

obastemur:
@dotnet-bot  test Windows 7 ci_dev12_x64_debug please

lib/Jsrt/Jsrt.cpp
obastemur: Since we know it will pass `recyclableObject::FromVar` straight, this one looks okay too.

PR #3864, opened 2017-10-03
json: allow totalNumPropertyCount == 0 (assert)

obastemur:


lib/Runtime/Library/JSON.cpp
```diff
@@ -695,10 +695,11 @@ namespace JSON
                         if (dynamicObject->HasObjectArray())
                         {
                             int totalNumPropertyCount = this->GetPropertyCount(object, &enumerator) - propertyCount;
-                            Assert(totalNumPropertyCount > 0);
+                            Assert(totalNumPropertyCount >= 0);
```
akroshg: what is the scenario here? if the object has the array, it should have some properties. Did remove property in between? can you add a test if it is handy?

obastemur: > if the object has the array, it should have some properties. 

@akroshg could you provide more details on this claim?

I did introduce this assert recently in order to make sure `this->GetPropertyCount` and `object->propertyCount` are consistent. However, there is an unusual case that object may have an objectArray but that Array may no longer have any items inside.

Also added a test case.

akroshg: @obastemur it was more like a question, that how did item/property get removed. Looked at the test case and made sense.

PR #3865, opened 2017-10-03
Defer destructured params if the parent function is getting deferred

aneeshdk:
Following up on the PR https://github.com/Microsoft/ChakraCore/pull/3610/files, this change addresses the same issue for destructured params. When parent function is deferred the destructured params should also get the deferred parse flag. In this case we are creating a dummy node for the pattern as the param count is needed for function length calculation.

Note: The change in StartEmit function is to do the scope slot allocation logic for the params only after the function is parsed. The scopes needs to be pushed to the stack in both cases.


akroshg:
Looks good to me. I'd look for Paul's opinion on this one though.

pleath:
Seems fine. I remember discussing with you how we could handle destructured params in a case like this.

PR #3868, opened 2017-10-03
cleanup unit tests for arm64 with nonative

sigatrev:
A few tests that expect jitted code and needed exclude_nonative were using require_backend or had nothing to disable the tests.

One test was for asmjs which is not currently compiled for arm or arm64.

One test was hitting stack overflow, which seems to happen at a lower threashold on arm64.

A few slower tests were timing out before completing.

### DIFF
--- a/test/inlining/rlexe.xml
+++ b/test/inlining/rlexe.xml
@@ -139,6 +139,7 @@
       <files>bug515849.js</files>
       <baseline>bug515849.baseline</baseline>
       <compile-flags>-minInterpretCount:1 -maxInterpretCount:1 -msjrc:0 -force:inline</compile-flags>
+      <tags>exclude_arm64</tags>
     </default>
   </test>
   <test>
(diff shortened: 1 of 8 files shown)


test/inlining/rlexe.xml
```diff
@@ -139,6 +139,7 @@
       <files>bug515849.js</files>
       <baseline>bug515849.baseline</baseline>
       <compile-flags>-minInterpretCount:1 -maxInterpretCount:1 -msjrc:0 -force:inline</compile-flags>
+      <tags>exclude_arm64</tags>
```
agarwal-sandeep: >exclude_arm64 [](start = 12, length = 13)

Why? if this is failing due to slowness/bug add a comment to re-enable later

sigatrev: The test is somewhat recursive and on arm64 it's hitting "Out of stack space" before it reaches it's normal completion.

Issue #3790, opened 2017-09-22
perf: JsrtGet/Set property from JsValueRef and more.


obastemur:
Using Jsrt, there is no direct way to use a `JavascriptString` object for property identifier.

`expensive loop` : Get the string from JsValueRef -> Convert to Utf16 -> Lookup / convert to JsPropertyRef... -> etc.. and finally Get OR Set the property!

So.. If we had an interface that accepts JsValueRef, we wouldn't have to do all these stuff.
This alone should change the things for node-chakracore. (AcmeAir perf wise.)

However, I would like to take it one step further and implement a string type that is very similar to Js::PropertyString with some differences.

`Js::JsrtString` (or however we are going to call)

Structure-wise, apart from being a regular `JavascriptString`... it will be holding a pointer to `PropertyRecord` + another pointer for extension / Jsrt call caching purposes. (Memory wise, it will cost +16 bytes on top of standard JavascriptString)

`PropertyRecord` ptr by default will be `nullptr`... It will be filled, once this`JsrtString` object is used for any property related purposes. Additional usages will be saving from `string TO propertyId` hashing / lookup... (name it!) And... from GC point.. Yes, as long as this string object is alive, PropertyRecord will be alive too.

`JsrtCreateString...` will create this new `JsrtString` by default. (In case the string length is bigger than 128 char, it will create the good old `JavascriptString` instead.`

** Q/A **

- Why not using `Js::PropertyString`?
That would cost +32 bytes. It also expects every string variable to be property string... and will ask for PropertyRecord from the beginning... etc. ( <- some cache related stuff )

- Expected gain?
Somewhere between 2% to 10% AcmeAIR LTO gain.

- 2% or 10% ? How did you calculate this?
Average AcmeAIR benchmark request forces the above mentioned "expensive loop" aprox. 50+ times.. In other words, ~80 times each milliseconds. (considering 1730 requests/second rate on the machine I do the benchmark)...

1 ms ==> 2.7Ghz => 2.7 e6 ticks (aprox CPU hertz) / 80times == > So... per each `33750` cpu ticks, we do one of these heavy loops. One loop costs somewhere between 2% to 10% of the available cpu ticks... Hence, this work may end-up giving 2% gain or something around 10%... or less...

- How much change needed on Chakrashim end?
Not much.

- Chakracore?
Not much.

I will start hacking soon on this.

Opinions? Objections?

MikeHolman:
[...]

PR #3875, opened 2017-10-04
jsrt: JsObject Delete/Get/Has/OwnP../Set Property

obastemur:
AcmeAIR LTO gain (3%)

New JSRT property interface that API user can use without going through convert-to-property-id process.

See #3790 for details.

Also added a note;
```
TODO: Can we make PropertyString && LiteralStringWithPropertyStringPtr
share the string buffer?
```

Once this PR is merged, there will be an additional PR on node-chakracore end to benefit this new interface.

### DIFF
--- a/lib/Common/Codex/Utf8Helper.h
+++ b/lib/Common/Codex/Utf8Helper.h
@@ -86,30 +86,28 @@ namespace utf8
         return WideStringToNarrow(Allocator::allocate, sourceString, sourceCount, destStringPtr, destCount, allocateCount);
     }
 
-    ///
-    /// Use the codex library to encode a UTF8 string to UTF16.
-    /// The caller is responsible for freeing the memory, which is allocated
-    /// using Allocator.
-    /// The returned string is null terminated.
-    ///
-    template <typename AllocatorFunction>
-    HRESULT NarrowStringToWide(_In_ AllocatorFunction allocator,_In_ LPCSTR sourceString, size_t sourceCount, _Out_ LPWSTR* destStringPtr, _Out_ size_t* destCount, size_t* allocateCount = nullptr)
+    inline HRESULT NarrowStringToWideNoAlloc(_In_ LPCSTR sourceString, size_t sourceCount,
+        __out_ecount(destBufferCount) LPWSTR destString, size_t destBufferCount, _Out_ size_t* destCount)
     {
-        size_t cbSourceString = sourceCount;
         size_t sourceStart = 0;
-        size_t cbDestString = (sourceCount + 1) * sizeof(WCHAR);
-        if (cbDestString < sourceCount) // overflow ?
+        size_t cbSourceString = sourceCount;
+
+        if (sourceCount >= MAXUINT32)
         {
+            destString[0] = WCHAR(0);
             return E_OUTOFMEMORY;
         }
 
-        WCHAR* destString = (WCHAR*)allocator(cbDestString);
         if (destString == nullptr)
         {
-            return E_OUTOFMEMORY;
+            return E_INVALIDARG;
         }
 
-        if (allocateCount != nullptr) *allocateCount = cbDestString;
+        if (sourceCount >= destBufferCount)
+        {
+            destString[0] = WCHAR(0);
+            return E_INVALIDARG;
+        }
 
         for (; sourceStart < sourceCount; sourceStart++)
         {
@@ -127,7 +125,6 @@ namespace utf8
         {
             *destCount = sourceCount;
             destString[sourceCount] = WCHAR(0);
-            *destStringPtr = destString;
         }
         else
         {
@@ -136,18 +133,54 @@ namespace utf8
 
             charcount_t cchDestString = utf8::ByteIndexIntoCharacterIndex(remSourceString, cbSourceString - sourceStart);
             cchDestString += (charcount_t)sourceStart;
-            Assert (cchDestString <= sourceCount);
+            if (cchDestString > sourceCount)
+            {
+                return E_OUTOFMEMORY;
+            }
 
             // Some node tests depend on the utf8 decoder not swallowing invalid unicode characters
             // instead of replacing them with the "replacement" chracter. Pass a flag to our
             // decoder to require such behavior
             utf8::DecodeUnitsIntoAndNullTerminateNoAdvance(remDestString, remSourceString, (LPCUTF8) sourceString + cbSourceString, DecodeOptions::doAllowInvalidWCHARs);
-            Assert(destString[cchDestString] == 0);
+
             static_assert(sizeof(utf8char_t) == sizeof(char), "Needs to be valid for cast");
-            *destStringPtr = destString;
             *destCount = cchDestString;
         }
+
+        Assert(destString[*destCount] == 0);
+
         return S_OK;
+    }
+
+    ///
+    /// Use the codex library to encode a UTF8 string to UTF16.
+    /// The caller is responsible for freeing the memory, which is allocated
+    /// using Allocator.
+    /// The returned string is null terminated.
+    ///
+    template <typename AllocatorFunction>
+    HRESULT NarrowStringToWide(_In_ AllocatorFunction allocator,_In_ LPCSTR sourceString,
+        size_t sourceCount, _Out_ LPWSTR* destStringPtr, _Out_ size_t* destCount, size_t* allocateCount = nullptr)
+    {
+        size_t cbDestString = (sourceCount + 1) * sizeof(WCHAR);
+        if (cbDestString < sourceCount) // overflow ?
+        {
+            return E_OUTOFMEMORY;
+        }
+
+        WCHAR* destString = (WCHAR*)allocator(cbDestString);
+        if (destString == nullptr)
+        {
+            return E_OUTOFMEMORY;
+        }
+
+        if (allocateCount != nullptr)
+        {
+            *allocateCount = cbDestString;
+        }
+
+        *destStringPtr = destString;
+        return NarrowStringToWideNoAlloc(sourceString, sourceCount, destString, sourceCount + 1, destCount);
     }
 
     template <class Allocator>
--- a/lib/Jsrt/Jsrt.cpp
+++ b/lib/Jsrt/Jsrt.cpp
@@ -1383,51 +1383,466 @@ CHAKRA_API JsPreventExtension(_In_ JsValueRef object)
     });
 }
 
+static JsErrorCode InternalGetPropertyRecord(Js::ScriptContext * scriptContext,
+    Js::RecyclableObject * key, _Out_ const Js::PropertyRecord ** propertyRecord)
+{
+    Assert(propertyRecord != nullptr);
+    *propertyRecord = nullptr;
+
+    if (key->GetTypeId() != Js::TypeIds_String)
+    {
+        return JsErrorInvalidArgument;
+    }
+
+    scriptContext->GetOrAddPropertyRecord(Js::JavascriptString::FromVar(key),
+        (Js::PropertyRecord const **)propertyRecord);
+    return JsNoError;
+}
+
+CHAKRA_API JsHasOwnPropertyCommon(Js::ScriptContext * scriptContext, _In_ JsValueRef object,
+    _In_ const Js::PropertyRecord * propertyRecord, _Out_ bool *hasOwnProperty,
+    TTDRecorder& _actionEntryPopper)
+{
+    PERFORM_JSRT_TTD_RECORD_ACTION(scriptContext, RecordJsRTHasOwnProperty, propertyRecord, object);
+
+    *hasOwnProperty = Js::JavascriptOperators::OP_HasOwnProperty(object,
+        propertyRecord->GetPropertyId(), scriptContext) != 0;
+
+    return JsNoError;
+}
+
+CHAKRA_API JsHasOwnProperty(_In_ JsValueRef object, _In_ JsPropertyIdRef propertyId,
+    _Out_ bool *hasOwnProperty)
+{
+    return ContextAPIWrapper<true>([&] (Js::ScriptContext *scriptContext,
+        TTDRecorder& _actionEntryPopper) -> JsErrorCode {
+
+        VALIDATE_INCOMING_OBJECT(object, scriptContext);
+        VALIDATE_INCOMING_PROPERTYID(propertyId);
+        PARAM_NOT_NULL(hasOwnProperty);
+        *hasOwnProperty = false;
+
+        return JsHasOwnPropertyCommon(scriptContext, object,
+            (const Js::PropertyRecord *)propertyId, hasOwnProperty, _actionEntryPopper);
+    });
+}
+
+#ifdef _CHAKRACOREBUILD
+CHAKRA_API JsObjectHasOwnProperty(_In_ JsValueRef object, _In_ JsValueRef propertyId, _Out_ bool *hasOwnProperty)
+{
+    return ContextAPIWrapper<true>([&] (Js::ScriptContext *scriptContext,
+        TTDRecorder& _actionEntryPopper) -> JsErrorCode {
+
+        VALIDATE_INCOMING_OBJECT(object, scriptContext);
+        VALIDATE_INCOMING_RECYCLABLE(propertyId, scriptContext);
+        PARAM_NOT_NULL(hasOwnProperty);
+        *hasOwnProperty = false;
+
+        const Js::PropertyRecord *propertyRecord = nullptr;
+        JsErrorCode errorValue = InternalGetPropertyRecord(scriptContext,
+            Js::RecyclableObject::FromVar(propertyId), &propertyRecord);
+
+        if (errorValue != JsNoError)
+        {
+            return errorValue;
+        }
+
+        return JsHasOwnPropertyCommon(scriptContext, object, propertyRecord, hasOwnProperty, _actionEntryPopper);
+    });
+}
+#endif
+
+static JsErrorCode JsGetPropertyCommon(Js::ScriptContext * scriptContext,
+    _In_ Js::RecyclableObject * object,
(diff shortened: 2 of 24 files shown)


lib/Common/Codex/Utf8Helper.h
```diff
@@ -83,31 +83,21 @@ namespace utf8
...
 
-    ///
-    /// Use the codex library to encode a UTF8 string to UTF16.
-    /// The caller is responsible for freeing the memory, which is allocated
-    /// using Allocator.
-    /// The returned string is null terminated.
-    ///
-    template <typename AllocatorFunction>
-    HRESULT NarrowStringToWide(_In_ AllocatorFunction allocator,_In_ LPCSTR sourceString, size_t sourceCount, _Out_ LPWSTR* destStringPtr, _Out_ size_t* destCount, size_t* allocateCount = nullptr)
+    inline HRESULT NarrowStringToWideNoAlloc(_In_ LPCSTR sourceString, size_t sourceCount, _Out_ LPWSTR destString, _Out_ size_t* destCount)
```
MSLaguana: For the no-alloc case, you should pass in the size of `destString` so we can check for and avoid overflows.

lib/Jsrt/Jsrt.cpp
```diff
@@ -1367,102 +1367,199 @@ CHAKRA_API JsPreventExtension(_In_ JsValueRef object)
...
+}
+
+static JsErrorCode JsGetPropertyCommon(Js::ScriptContext * scriptContext, _In_ Js::RecyclableObject * object,
+    _In_ const Js::PropertyRecord * propertyRecord, _Out_ JsValueRef *value,
+    TTDRecorder& _actionEntryPopper)
+{
+    AssertMsg(scriptContext->GetThreadContext()->IsScriptActive(), "Caller is expected to be under ContextAPIWrapper!");
+    PERFORM_JSRT_TTD_RECORD_ACTION(scriptContext, RecordJsRTGetProperty, propertyRecord, object);
+
+    *value = Js::JavascriptOperators::GetPropertyNoCache(object, propertyRecord->GetPropertyId(), scriptContext);
```
MSLaguana: Why `GetPropertyNoCache`?

lib/Jsrt/Jsrt.cpp
```diff
@@ -1367,102 +1367,199 @@ CHAKRA_API JsPreventExtension(_In_ JsValueRef object)
     });
 }
 
+static JsErrorCode InternalGetPropertyRecord(Js::ScriptContext * scriptContext, JsValueRef key, _Out_ const Js::PropertyRecord ** propertyRecord)
+{
```
MSLaguana: Is this deliberately weaker than being able to compute `x[y]`,  e.g. `var y = 2, var x = [1,2,3]; x[y]`? That is we deliberately do not support getting property records for numbers, null, true, false, undefined, or arbitrary objects which get converted to a string?

obastemur: We have an interface for number `JsGetIndexed...` (also works similarly) for the time being.. Once / if we ever  drop that interface, we may combine here. I was designing in a way that they are handled in the same place. Then I realized, consumer may have a better chance to know it hence no point wasting it here.

lib/Jsrt/Jsrt.cpp
obastemur: We know the type and no cache info ptr. So, no waste time checking and cache related stuff.

obastemur:

lib/Runtime/Library/ConcatString.h
```diff
@@ -9,18 +9,30 @@ namespace Js
     class LiteralStringWithPropertyStringPtr : public LiteralString
     {
     private:
+        // TODO: Can we make PropertyString && LiteralStringWithPropertyStringPtr share the string buffer?
         PropertyString * propertyString;
```
MSLaguana: @MikeHolman do you know why this isn't a `Field`? The CI seems to be complaining about it now, and I'm inclined to agree with it.

MikeHolman: Yeah, it should be one.

lib/Runtime/Language/JavascriptOperators.cpp
```diff
@@ -4413,18 +4413,8 @@ namespace Js
...
-                    scriptContext->GetOrAddPropertyRecord(str->GetString(), str->GetLength(), &propertyRecord);
-                    propertyString = scriptContext->GetPropertyString(propertyRecord->GetPropertyId());
-                    str->SetPropertyString(propertyString);
-                }
-                else
-                {
-                    propertyRecord = propertyString->GetPropertyRecord();
-                }
-
+                propertyString = str->GetPropertyStringLazy();
```
MSLaguana: Why not also use `GetPropertyStringLazy` in `OP_GetElementI`?

obastemur: ? sorry not sure I understood.

MSLaguana: Also looks like `JavascriptConversion::ToPropertyKey` could have some of it's code dealing with `LiteralStringWithPropertyStringPtr` replaced with a call to `GetPropertyStringLazy`

obastemur: `JavascriptConversion::ToPropertyKey` and similar places good idea. Could you please clarify -> `Why not also use GetPropertyStringLazy in OP_GetElementI` ?

MSLaguana: I was looking for where you had used `GetPropertyStringLazy`, and I saw some other places that `GetPropertyString` was used but you had not changed it to use `GetPropertyStringLazy`; is there a reason for that, or should we also update the other locations? Specifically for `OP_GetElementI` I was referring to https://github.com/Microsoft/ChakraCore/blob/master/lib/Runtime/Language/JavascriptOperators.cpp#L3697

lib/Common/Codex/Utf8Helper.h
```diff
@@ -141,12 +132,43 @@ namespace utf8
...
+        WCHAR* destString = (WCHAR*)allocator(cbDestString);
+        if (destString == nullptr)
+        {
+            return E_OUTOFMEMORY;
+        }
+
+        if (allocateCount != nullptr) *allocateCount = cbDestString;
+
+        *destStringPtr = destString;
+        HRESULT result = NarrowStringToWideNoAlloc(sourceString, sourceCount, destString, sourceCount, destCount);
```
MSLaguana: I believe that the `destBufferCount` parameter here should be `sourceCount+1` to allow for the null

lib/Runtime/Library/ConcatString.h
```diff
@@ -9,18 +9,30 @@ namespace Js
     class LiteralStringWithPropertyStringPtr : public LiteralString
     {
     private:
-        PropertyString * propertyString;
+        // TODO: Can we make PropertyString && LiteralStringWithPropertyStringPtr share the string buffer?
+        Field(PropertyString*) propertyString;
 
     public:
         PropertyString * GetPropertyString() const;
+        PropertyString * GetPropertyStringLazy(); // Get if it's there, otherwise bring it in.
```
MikeHolman: nit: this name is confusing to me. i think we could come up with something better, maybe `GetOrAddPropertyString`?

lib/Runtime/Library/ConcatString.h
```diff
@@ -9,18 +9,30 @@ namespace Js
     class LiteralStringWithPropertyStringPtr : public LiteralString
     {
     private:
-        PropertyString * propertyString;
+        // TODO: Can we make PropertyString && LiteralStringWithPropertyStringPtr share the string buffer?
```
MikeHolman: I think you can just overwrite the buffer on this once we populate the PropertyString field (and then PropertyString, LiteralStringWithPropertyStringPtr, and PropertyRecord will all be sharing same buffer)

lib/Runtime/Library/ConcatString.cpp
```diff
@@ -17,11 +17,127 @@ namespace Js
...
+
+    LiteralStringWithPropertyStringPtr *LiteralStringWithPropertyStringPtr::
+      NewFromCString(const char * cString, const CharCount charCount, JavascriptLibrary *const library)
+    {
+        Assert(library != nullptr && cString != nullptr);
+
+        switch (charCount)
+        {
+            case 0:
+                return (LiteralStringWithPropertyStringPtr*) library->GetEmptyString();
```
MikeHolman: Could you have GetEmptyString return LiteralStringWithPropertyStringPtr* so we can avoid this cast? otherwise someone could change the emptyString to some other type of string, and compiler won't complain

obastemur: I could put an assert on the returning type (VT check)

MikeHolman: Sure that would work, but I think it is strictly better to have more type information represented at compile time rather than relying on run time asserts.

MikeHolman:
Just FYI I have some changes in master that this will conflict with.

Also, any reason this is going in 1.7? Seems odd to be doing perf work in a release branch.

liminzhu:
We should have an equivalent for [JsHasOwnProperty](https://github.com/Microsoft/ChakraCore/wiki/JsHasOwnProperty) for completeness. Also, do those APIs work with symbols?

lib/Runtime/Library/ConcatString.cpp
obastemur: Aren't we going to cast this type eventually? (library-GetEmptyString)

lib/Runtime/Library/ConcatString.h
obastemur: Indeed, this is a trivial TODO. However, I would like to keep it separate. (That's why TODO) BTW; the issue is still open and it also mentions this improvement.

lib/Runtime/Library/ConcatString.h
obastemur: I'm not strong on naming and what you suggest sounds good.

obastemur:
> Also, any reason this is going in 1.7? Seems odd to be doing perf work in a release branch.

and stressful.. 

@digitalinfinity ?

obastemur:
> We should have an equivalent for JsHasOwnProperty for completeness.

Good catch! Done.

> Also, do those APIs work with symbols?

Not for now. Documented (strictly) that it requires a JavascriptString.

liminzhu:
> Not for now. Documented (strictly) that it requires a JavascriptString.

Yep, was asking b/c I saw the comments on header. Do you plan to support symbol eventually with this set of API? It makes sense to be able to handle symbols and string props the same way.

obastemur:
> Do you plan to support symbol eventually with this set of API? It makes sense to be able to handle symbols and string props the same way.

Counter question; how does a JSRT user can use as a Symbol to define a property?

digitalinfinity:
@MikeHolman @obastemur we were targeting 1.7 for Node perf fixes but I think we can go back to master for now (unless explicitly fixing a bug in 1.7).

obastemur:
@digitalinfinity sounds good. Targets ‘master’ now.

obastemur:
Ping?

liminzhu:
> Counter question; how does a JSRT user can use as a Symbol to define a property?

Got side-tracked by other stuff, but there are a few ways users can get a Symbol natively. All of them return a `JsValueRef`.

1. Create one through `JsCreateSymbol`
2. `JsGetOwnPropertySymbols`
3. Grab a symbol that is a property of another var (e.g. `Symbol.toStringTag`)

After having the symbol `JsValueRef`, it's similar to name properties
```cpp
JsGetPropertyIdFromSymbol(sym, &symPropId);
JsSetProperty(o, symPropId, val, useStrictRules);
```

Given JSRT users always deal with symbols in `JsValueRef`, a `JsObjectSetProperty` (and other property-related APIs) working with Symbol would be great.

obastemur:
@liminzhu thanks for the detailed answers.

User needs to go through multiple steps to use Symbol. I'm +1 on making Symbol also part of the supported options.

`Ping` was for mostly reviewers.

mrkmarron:
TTD parts look good to me. I am not sure if they have been run through the TTD top NPM tests yet which will test this API more thoroughly.

obastemur:
self note for the PR;

- `JavascriptOperators::GetElementIHelper` (etc.) could use `PropertyString || LiteralStringWithPropertyStringPtr` to identify the type.

obastemur:
Made some additional improvements. Will measure the perf separately but need review on this.

/cc @MikeHolman @MSLaguana

lib/Common/Codex/Utf8Helper.h
```diff
@@ -86,31 +86,24 @@ namespace utf8
...
+
+        if (sourceCount >= MAXUINT32)
         {
+            destString[0] = WCHAR(0);
             return E_OUTOFMEMORY;
         }
 
-        WCHAR* destString = (WCHAR*)allocator(cbDestString);
-        if (destString == nullptr)
+        if (destString == nullptr || sourceCount >= destBufferCount)
```
MSLaguana: If `destString == nullptr` then we shouldn't be writing to it at all; perhaps put that check first?

MSLaguana: Also isn't `sourceCount >= destBufferCount` legal? E.g. if the input narrow string has more 3-byte utf8 codepoints than 1-bye utf8 codeopints, the utf16 representation can be shorter than the input.

lib/Common/Codex/Utf8Helper.h
```diff
@@ -142,14 +134,44 @@ namespace utf8
             // instead of replacing them with the "replacement" chracter. Pass a flag to our
             // decoder to require such behavior
             utf8::DecodeUnitsIntoAndNullTerminateNoAdvance(remDestString, remSourceString, (LPCUTF8) sourceString + cbSourceString, DecodeOptions::doAllowInvalidWCHARs);
```
MSLaguana: The assert just before here that `cchDestString <= sourceCount` should be `cchDestString <= destBufferCount` I believe (maybe strictly less to allow for the null terminator), and if this function is exposed outside of our code then I think it should be an `if(...) { return E_OUTOFMEMORY;}` and not an assert.

lib/Common/Codex/Utf8Helper.h
obastemur: I didn't understand your first comment.

Second comment; -> sourceCount represents the length of `char16` buffer. Allocator is supposed to allocate equal or more (prior to calculating exact correspondence). So we can convert using the fast path below.

lib/Common/Codex/Utf8Helper.h
obastemur: Not sure.. This PR didn't make any logical changes within the Utf8Helper stuff. Just split an existing interface for reuse purposes.

lib/Common/Codex/Utf8Helper.h
MSLaguana: This condition is saying `if (destString == nullptr || ...) { destString[0] = 0}` which will (try to) write to the null pointer. Also the previous condition will write to `destString` without checking whether it is null. We should either be checking if `destString == nullptr` up front and exiting with an error without writing to it, or we should be relying on SAL to make sure that we never pass in a null pointer I think.

Regarding the second point; Requiring that `sourceCount >= destBufferCount` is safe, but it may lead to excess memory usage. Prior to this PR the over-estimating allocations are (I believe) always temporary, and the string is copied to a more accurately sized buffer for long-term storage, or freed once the variable goes out of scope. The new usage that you're adding in recycler-allocates a (potentially) over-estimated amount and hands that off to to be stored in a `LiteralString`.

Perhaps it is worth being clear about what the contract of this function is? Are you supposed to supply a buffer with `destBufferCount == sourceCount`, and then the caller is responsible for adjusting to `destCount` when the call returns? Is the caller supposed to supply a buffer with size at least `destCount` which is initially unknown, and it is an error if `destBufferCount < destCount`?

obastemur: Second part; Yes, the idea is/was, instead of double pass, allocate ASCII amount of memory. OR we will endup having double pass for all the strings.

lib/Common/Codex/Utf8Helper.h
obastemur: Added `if (cchDestString > sourceCount)`

lib/Runtime/Library/ConcatString.h
obastemur: done.

lib/Runtime/Language/JavascriptOperators.cpp
```diff
@@ -10449,15 +10442,9 @@ namespace Js
...
-        JsUtil::CharacterBuffer<WCHAR> propertyName(propertyNameString->GetString(), propertyNameString->GetLength());
-        if (Js::BuiltInPropertyRecords::__proto__.Equals(propertyName))
-        {
-            return CheckPrototypesForAccessorOrNonWritablePropertyCore<JavascriptString*, false, false>(instance, propertyNameString, setterValue, flags, info, scriptContext);
-        }
-        else
-        {
-            return CheckPrototypesForAccessorOrNonWritablePropertyCore<JavascriptString*, true, false>(instance, propertyNameString, setterValue, flags, info, scriptContext);
-        }
+        PropertyString * propertyString = nullptr;
```
MSLaguana: This variable isn't used and that's causing some compile errors on windows.

obastemur:
Made some additional changes, looks like `FindPropertyRecord` was also taking the slow path all the time. Will measure the perf separately.

lib/Runtime/Library/JSON.cpp
```diff
@@ -431,10 +431,16 @@ namespace JSON
 
         if (value == nullptr)
         {
-            Js::PropertyString* propertyString = Js::PropertyString::TryFromVar(key);
-            if (propertyString)
+            Js::JavascriptString * propertyKey = key;
```
MSLaguana: This is going to conflict with another JSON change that went in earlier today, might want to rebase your change and reapply if necessary

PR #3881, opened 2017-10-05
keep the toolchain path from android-configure 

miaulightouch:
node's [android-configure](https://github.com/nodejs/node-chakracore/blob/master/android-configure#L50) script had set the $TOOLCHAIN to the right path and exported the needed variables.

only configure toolchain path when $TOOLCHAIN is undefined.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -273,17 +273,22 @@ while [[ $# -gt 0 ]]; do
         _TARGET_OS="${_TARGET_OS:9}"
         if [[ $_TARGET_OS =~ "android" ]]; then
             OLD_PATH=$PATH
-            export TOOLCHAIN=$PWD/android-toolchain-arm
+            if [[ -z "$TOOLCHAIN" ]]; then
+                export TOOLCHAIN=$PWD/android-toolchain-arm
+                export PATH=$TOOLCHAIN/bin:$OLD_PATH
+                export AR=arm-linux-androideabi-ar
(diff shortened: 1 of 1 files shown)


build.sh
```diff
@@ -273,17 +273,22 @@ while [[ $# -gt 0 ]]; do
         _TARGET_OS="${_TARGET_OS:9}"
         if [[ $_TARGET_OS =~ "android" ]]; then
             OLD_PATH=$PATH
```
obastemur: Better move this line too?

miaulightouch: I will keep this, because this is **target** option. consider android is not official supported, it could be a **temporary** solution.

I would wish this option has ability like android-configure:
1. check external toolchain (set by android-configure) and apply
2. build standalone toolchain if toolchain undefined and set PATH

obastemur: @miaulightouch what I mean is... You moved everything but `OLD_PATH=$PATH` into an `IF` section. Only use I can see for `OLD_PATH` is inside that `IF` section.

So, could you please move that line too?

obastemur: Or.. if you are concerned with the TARGET_OS thing.. you may move it up too.

miaulightouch:
I think is fine now, I forget to move OLD_PATH in.

miaulightouch:
I think I had post another new PR for accidently branch merge.

PR #3883, opened 2017-10-05
ARM64: Copy XDataAllocator implementation from ARM32 to get code working

agarwal-sandeep:
ARM64: Copy XDataAllocator implementation from ARM32 to get code working


pleath:
Is the plan going forward to share the implementation?

agarwal-sandeep:
@pleath sure if there are things in common and deadline permit.
For now the goal is to get basic things running so that others are unblocked and can make progress.

pleath:
Well, if the implementation is being copied, then it sounds like it could also be shared, which would be desirable. The same goes for legalization.

PR #3884, opened 2017-10-05
keep toolchain path that set by android-configure

miaulightouch:
node's [android-configure](https://github.com/nodejs/node-chakracore/blob/master/android-configure#L50) script had set the $TOOLCHAIN to the right path and exported the needed variables.

only configure toolchain path when $TOOLCHAIN is undefined.


miaulightouch:
hope there's no problem this time.

~~that "Update branch" button is a trap.~~

obastemur:
@miaulightouch Thanks for the contribution!

LGTM

PR #3888, opened 2017-10-05
Introduce native unittests for FunctionExecutionStateMachine

thomasmo:
Introduce native unittests for FunctionExecutionStateMachine
This change introduces a set of CATCH native unittests for the state machine to emulate a handful of scenarios. This will be the first round of defense with modifying the state machine after the refactoring. In addition to tests, several stubs classes and files are added to let FunctionExecutionStateMachine compile as well as to control its behavior.
The tests in this first change include
- BasicTest
- NormalExecution
- JSUnitTestInterpreted
- JSUnitTestDynapogo
More tests will be added as more code is updated.


bin/NativeTests/FunctionExecutionTest.cpp
```diff
@@ -0,0 +1,313 @@
...
+        //15 chakra!ScriptEngine::ExecutePendingScripts
+        //16 chakra!ScriptEngine::ParseScriptTextCore
+        //17 chakra!ScriptEngine::ParseScriptText
+        f.TryTransitionToNextExecutionMode();
+        CHECK(f.GetExecutionMode() == ExecutionMode::FullJit);
+        CHECK(f.GetInterpretedCount() == 0);
+    }
+
+    // test what happens when we jit a Loop Body
+    TEST_CASE("FuncExe_NormalExecutionOfLoop")
```
akroshg: is this work in progress?

akroshg:
Good to go if the tests are stable.

bin/NativeTests/FunctionExecutionTest.cpp
thomasmo: yes--i'll followup with this alongside product code

PR #3889, opened 2017-10-05
Encoder reorganization

aaronsgiles:
Reorganized the encoder to support multiple operand sizes. Removed explicit 64-bit opcodes. Added support for byte and halfword loads/stores. Added legalizer support for converting CMP/CMN/TST into SUBS/ADDS/ANDS as appropriate.

### DIFF
--- a/lib/Backend/arm64/ARM64Encoder.h
+++ b/lib/Backend/arm64/ARM64Encoder.h
@@ -1921,12 +1921,12 @@ EmitBitfieldCommon(
     Arm64CodeEmitter &Emitter,
     Arm64SimpleRegisterParam Dest,
     Arm64SimpleRegisterParam Src,
-    ULONG Immr,
-    ULONG Imms,
+    ULONG64 Immr,
+    ULONG64 Imms,
     ULONG Opcode
     )
 {
-    return Emitter.EmitFourBytes(Opcode | (Immr << 16) | (Imms << 10) | (Src.RawRegister() << 5) | Dest.RawRegister());
+    return Emitter.EmitFourBytes(Opcode | (ULONG(Immr) << 16) | (ULONG(Imms) << 10) | (Src.RawRegister() << 5) | Dest.RawRegister());
 }
 
 inline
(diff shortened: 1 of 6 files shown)


lib/Backend/arm64/EncoderMD.cpp
```diff
@@ -186,555 +186,497 @@ EncoderMD::DecodeMemoryOpnd(IR::Opnd* opnd, ARM64_REGISTER &baseRegResult, ARM64
...
-        else
-        {
-            Assert(src1->IsRegOpnd());
-            bytes = EmitAddsRegister(Emitter, ARMREG_ZR, this->GetRegEncode(dst->AsRegOpnd()), this->GetRegEncode(src1->AsRegOpnd()));
-        }
+        Assert(false);
         break;
 
     case Js::OpCode::CMP_ASR31:
         dst = instr->GetDst();
```
agarwal-sandeep: CMP compares src1 and src2 and not dst. @sigatrev was trying to fix the same in #3886

lib/Backend/arm64/ARM64Encoder.h
```diff
@@ -1921,12 +1921,12 @@ EmitBitfieldCommon(
...
     Arm64SimpleRegisterParam Src,
-    ULONG Immr,
-    ULONG Imms,
+    ULONG64 Immr,
+    ULONG64 Imms,
     ULONG Opcode
     )
 {
-    return Emitter.EmitFourBytes(Opcode | (Immr << 16) | (Imms << 10) | (Src.RawRegister() << 5) | Dest.RawRegister());
+    return Emitter.EmitFourBytes(Opcode | (ULONG(Immr) << 16) | (ULONG(Imms) << 10) | (Src.RawRegister() << 5) | Dest.RawRegister());
```
agarwal-sandeep: Will truncating cause data loss?

lib/Backend/arm64/EncoderMD.cpp
aaronsgiles: Ah, sorry for the misunderstanding. Will address this.

lib/Backend/arm64/ARM64Encoder.h
aaronsgiles: Not for this case, no. Upstream functions ensure it won't happen. I just switched to 64-bit parameters here to prevent excessive warnings, and there's no perf difference.

PR #3892, opened 2017-10-06
Top level event handlers have an implicit 'this' binding

boingoing:


lib/Parser/Parse.cpp
```diff
@@ -1682,10 +1682,12 @@ void Parser::CreateSpecialSymbolDeclarations(ParseNodePtr pnodeFnc, bool isGloba
 {
     Assert(!(isGlobal && (this->m_grfscr & fscrEval)));
 
+    bool isTopLevelEventHandler = (this->m_grfscr & fscrImplicitThis || this->m_grfscr & fscrImplicitParents);
```
pleath: This is actually too conservative. We only have the implicit "this" reference in the top lexical level of an event handler. Functions declared inside the enclosing function don't need this.

lib/Parser/Parse.cpp
```diff
@@ -1682,10 +1682,12 @@ void Parser::CreateSpecialSymbolDeclarations(ParseNodePtr pnodeFnc, bool isGloba
...
+
     // Create a 'this' symbol for global lambda, indirect eval, ordinary functions with references to 'this', and all class constructors.
     PidRefStack* ref = wellKnownPropertyPids._this->GetTopRef();
     if (((pnodeFnc->sxFnc.IsLambda() && GetCurrentNonLambdaFunctionNode() == nullptr && !(this->m_grfscr & fscrEval)) || !pnodeFnc->sxFnc.IsLambda())
```
pleath: As we've discussed, this check can be simpler. Let me take a crack at it.

boingoing: @pleath As part of the work to defer parse lambdas, I have tried to simplify this further by removing the 'this' binding for global and lambda capturing global altogether. That turns this check into:

```c++
if ((ref && ref->GetScopeId() >= m_currentBlockInfo->pnodeBlock->sxBlock.blockId) || pnodeFnc->sxFnc.IsClassConstructor())
```

Seems to work ok; this does break some assumption in the optimizer about calling LdThis more than once, though.

```js
() => {
    this.o = 'first';
    this.o = 'second';
}
```

pleath: Ah. Cool. I thought for some reason the global "this" change was already done. Okay, I'll leave that for you, then. Thanks.

lib/Parser/Parse.cpp
boingoing: Thanks, Paul. We made this change yesterday to try and unblock the DRT. 

Is the event handler treated like global code?

PR #3893, opened 2017-10-06
json: use type handler property index

obastemur:
Add a test case for stringify of an object with a getter.


obastemur:
Added more test cases then simplified JSON.stringify further with extra perf gain. Also simplified  propertyRecord lookups.. will send another PR instead.

PR #3894, opened 2017-10-06
perf: simplify stringify + GetOrAddPropertyRecord

obastemur:
AcmeAIR LTO (~1%)

- further simplify JSON.stringify (add more test cases and fix compat issues)
- Use PropertyString->PropertyRecord instead of a lookup using string buffer

### DIFF
--- a/lib/Runtime/Base/ScriptContext.cpp
+++ b/lib/Runtime/Base/ScriptContext.cpp
@@ -837,6 +837,18 @@ namespace Js
         return propertyRecord->GetPropertyId();
     }
 
+    void ScriptContext::GetOrAddPropertyRecord(Js::JavascriptString * propertyString, PropertyRecord const** propertyRecord)
+    {
+        if (VirtualTableInfo<Js::PropertyString>::HasVirtualTable(propertyString))
+        {
+            *propertyRecord = ((Js::PropertyString*)propertyString)->GetPropertyRecord();
+        }
+        else
+        {
+            GetOrAddPropertyRecord(propertyString->GetString(), propertyString->GetLength(), propertyRecord);
+        }
+    }
+
     void ScriptContext::GetOrAddPropertyRecord(JsUtil::CharacterBuffer<WCHAR> const& propertyName, PropertyRecord const ** propertyRecord)
     {
         threadContext->GetOrAddPropertyId(propertyName, propertyRecord);
--- a/lib/Runtime/Base/ScriptContext.cpp
+++ b/lib/Runtime/Base/ScriptContext.cpp
@@ -839,7 +839,7 @@ namespace Js
 
     void ScriptContext::GetOrAddPropertyRecord(Js::JavascriptString * propertyString, PropertyRecord const** propertyRecord)
     {
-        if (VirtualTableInfo<Js::PropertyString>::HasVirtualTable(propertyString))
+        if (VirtualTableInfo<Js::PropertyString>::HasVirtualTable(propertyString) && propertyString->GetScriptContext() == this)
         {
             *propertyRecord = ((Js::PropertyString*)propertyString)->GetPropertyRecord();
         }
--- a/lib/Runtime/Debug/DiagObjectModel.cpp
+++ b/lib/Runtime/Debug/DiagObjectModel.cpp
@@ -2483,7 +2483,7 @@ namespace Js
                                         else
                                         {
                                             const PropertyRecord* propertyRecord;
-                                            objectContext->GetOrAddPropertyRecord(obj->GetSz(), obj->GetLength(), &propertyRecord);
+                                            objectContext->GetOrAddPropertyRecord(obj, &propertyRecord);
                                             propertyId = propertyRecord->GetPropertyId();
                                         }
                                     }
--- a/lib/Runtime/Library/ForInObjectEnumerator.cpp
+++ b/lib/Runtime/Library/ForInObjectEnumerator.cpp
@@ -194,7 +194,7 @@ namespace Js
                     else
                     {
                         ScriptContext* scriptContext = currentIndex->GetScriptContext();
-                        scriptContext->GetOrAddPropertyRecord(currentIndex->GetString(), currentIndex->GetLength(), &propRecord);
+                        scriptContext->GetOrAddPropertyRecord(currentIndex, &propRecord);
                         propertyId = propRecord->GetPropertyId();
 
                         // We keep the track of what is enumerated using a bit vector of propertyID.
--- a/lib/Runtime/Library/JSON.cpp
+++ b/lib/Runtime/Library/JSON.cpp
@@ -137,7 +137,7 @@ namespace JSON
             Js::JavascriptString *propertyName = Js::JavascriptString::FromVar(value);
             nameTable[tableLen].propName = propertyName;
             Js::PropertyRecord const * propertyRecord;
-            scriptContext->GetOrAddPropertyRecord(propertyName->GetString(), propertyName->GetLength(), &propertyRecord);
+            scriptContext->GetOrAddPropertyRecord(propertyName, &propertyRecord);
             nameTable[tableLen].propRecord = propertyRecord;        // Keep the property id alive.
             tableLen++;
         }
@@ -453,6 +453,7 @@ namespace JSON
                 }
             }
         }
+
         return StrHelper(key, value, holder);
     }
 
@@ -688,24 +689,19 @@ namespace JSON
 
                         result = Js::ConcatStringBuilder::New(this->scriptContext, propertyCount);
                         Js::DynamicTypeHandler * typeHandler = dynamicObject->GetTypeHandler();
-                        enumerator.Reset();
-                        Js::PropertyId previousId;
+
                         // if object has an objectarray, (aka indexed properties)
                         // we need to loop them first since they won't show up on the second loop below
                         if (dynamicObject->HasObjectArray())
                         {
-                            int totalNumPropertyCount = this->GetPropertyCount(object, &enumerator) - propertyCount;
-                            Assert(totalNumPropertyCount >= 0);
-
-                            int index = 0;
                             enumerator.Reset();
-                            while (index < totalNumPropertyCount && (propertyName = enumerator.MoveAndGetNext(id)) != NULL)
+                            while ((propertyName = enumerator.MoveAndGetNext(id)) != NULL)
                             {
                                 // if unsuccessful get propertyId from the string
-                                scriptContext->GetOrAddPropertyRecord(propertyName->GetString(), propertyName->GetLength(), &propRecord);
+                                scriptContext->GetOrAddPropertyRecord(propertyName, &propRecord);
+                                if (!propRecord->IsNumeric()) break;
                                 id = propRecord->GetPropertyId();
                                 StringifyMemberObject(propertyName, id, value, (Js::ConcatStringBuilder*)result, indentString, memberSeparator, isFirstMember, isEmpty);
-                                index++;
                             }
                         }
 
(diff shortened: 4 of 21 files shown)


lib/Runtime/Base/ScriptContext.cpp
```diff
@@ -837,6 +837,18 @@ namespace Js
         return propertyRecord->GetPropertyId();
     }
 
+    void ScriptContext::GetOrAddPropertyRecord(Js::JavascriptString * propertyString, PropertyRecord const** propertyRecord)
+    {
+        if (VirtualTableInfo<Js::PropertyString>::HasVirtualTable(propertyString))
+        {
+            *propertyRecord = ((Js::PropertyString*)propertyString)->GetPropertyRecord();
```
MSLaguana: Are there any concerns about passing strings cross-contexts here? Should we require that the `propertyString` comes from this script context?

obastemur: good question and IMHO cross context here matters. I will add that.

lib/Runtime/Library/JSON.cpp
```diff
@@ -688,55 +689,45 @@ namespace JSON
...
-
-                            int index = 0;
                             enumerator.Reset();
-                            while (index < totalNumPropertyCount && (propertyName = enumerator.MoveAndGetNext(id)) != NULL)
+                            while ((propertyName = enumerator.MoveAndGetNext(id)) != NULL)
                             {
                                 // if unsuccessful get propertyId from the string
-                                scriptContext->GetOrAddPropertyRecord(propertyName->GetString(), propertyName->GetLength(), &propRecord);
+                                scriptContext->GetOrAddPropertyRecord(propertyName, &propRecord);
+                                if (!propRecord->IsNumeric()) break;
```
MSLaguana: Do we guarantee that numeric properties always come first in enumeration order? Is that enforced by spec, or an implementation detail?

obastemur: if we don't https://github.com/Microsoft/ChakraCore/pull/3894/files#diff-8092540edf441c597a868bfdb4789f0aR58 will break.

lib/Runtime/Library/JSON.cpp
```diff
@@ -688,55 +689,45 @@ namespace JSON
...
                         for (uint32 i = 0; i < propertyCount; i++)
                         {
                             id = typeHandler->GetPropertyId(scriptContext, (Js::PropertyId)i);
-
-                            if (id < Js::InternalPropertyIds::Count)
+                            if (id == Js::Constants::NoProperty)
                             {
-                                // if the property ID is internal, we cannot assume dynamicObject->getSlot(i) is a var later
-                                previousId = id;
                                 continue;
```
MSLaguana: Why is this case now a no-op? Previously (both immediately prior to this PR and also before your original stringify changes) this case tried to fetch the id from the enumerator

obastemur: because, now implicitly checks `enumerable` and `callable` on premises while reading the property using typeHandler lookup.

lib/Runtime/Library/JSON.cpp
```diff
@@ -688,55 +689,45 @@ namespace JSON
...
+                                propRecord = propertyString->GetPropertyRecord();
+                                propertyName = (Js::JavascriptString*) propertyString;
                             }
 
-                            previousId = id;
                             if (!propRecord->IsSymbol())
                             {
-                                Js::Var property = dynamicObject->GetSlot(i);
+                                PropertyIndex index = typeHandler->GetPropertyIndex(propRecord);
+                                Js::Var property = index != Constants::NoSlot ?
```
MSLaguana: How does this work with getters? I believe that `GetPropertyIndex` only gets the `Data` property of the `DictionaryPropertyDescriptor` in that case, which would be `NoSlots` if it has an accessor.

obastemur: That's why `index != Constants::NoSlot`

lib/Runtime/Library/JSON.cpp
```diff
@@ -688,55 +689,45 @@ namespace JSON
...
                                 id = propRecord->GetPropertyId();
                                 StringifyMemberObject(propertyName, id, value, (Js::ConcatStringBuilder*)result, indentString, memberSeparator, isFirstMember, isEmpty);
-                                index++;
                             }
                         }
 
                         // loop over the named properties
                         for (uint32 i = 0; i < propertyCount; i++)
                         {
                             id = typeHandler->GetPropertyId(scriptContext, (Js::PropertyId)i);
```
jackhorton: This line has confused me for a while, and looking into it more, I realized why -- the second argument, if its going to be casted, should be casted to a `Js::PropertyIndex` rather than a `Js::PropertyId`

lib/Runtime/Library/JSON.cpp
```diff
@@ -688,55 +689,45 @@ namespace JSON
...
-                                if ((propertyName = enumerator.MoveAndGetNext(previousId)) == NULL) break;
-
-                                scriptContext->GetOrAddPropertyRecord(propertyName->GetString(), propertyName->GetLength(), &propRecord);
-                                id = propRecord->GetPropertyId();
-                            }
                             else
                             {
-                                propRecord = scriptContext->GetPropertyName(id);
-                                propertyName = scriptContext->GetPropertyString(id);
+                                if (!object->IsEnumerable(id)) continue;
```
jackhorton: nit: for consistency, shouldn't this be `if (!dynamicObject->IsEnumerable(id))`?

lib/Runtime/Library/JSON.cpp
```diff
@@ -688,55 +689,45 @@ namespace JSON
...
+                                Js::PropertyString * propertyString = scriptContext->GetPropertyString(id);
+                                propRecord = propertyString->GetPropertyRecord();
+                                propertyName = (Js::JavascriptString*) propertyString;
```
jackhorton: nit: We have so many dynamic casts in the codebase that it may not matter, but using `static_cast` whenever possible at least helps newer people like myself understand what is going on a bit better.

test/JSON/toJSON.js
```diff
@@ -15,12 +15,141 @@ fnc.prototype.toJSON = function() {
...
-TEST(JSON.stringify(obj), '{"0":2,"a":1}');
-delete obj[0];
-TEST(JSON.stringify(obj), '{"a":1}');
+{
+  var obj = {"a":1};
+  obj[0] = 2;
+  TEST(JSON.stringify(obj), '{"0":2,"a":1}');
+  delete obj[0];
+  TEST(JSON.stringify(obj), '{"a":1}');
+}
```
jackhorton: You may want to merge these tests with the one I added in `stringifyobject.js`, just for simplicity's sake. Also, using the unit test framework may help pinpoint exact problems, since we can actually name the tests there?

test/JSON/toJSON.js
```diff
@@ -15,12 +15,141 @@ fnc.prototype.toJSON = function() {
...
 
+// test getter - no-enum
+{
+  var obj = {};
+  Object.defineProperty(obj, "c", { get: () => 3, enumerable: false });
+
+  obj.a = 1;
+  obj.b = 2;
+  obj.d = 4;
+  TEST('{"a":1,"b":2,"d":4}', JSON.stringify(obj));
```
akroshg: interesting to see that we didn't have any test which does stringify on getter/setter object. Thanks for adding it.

test/JSON/toJSON.js
obastemur: Well, we had this discussion with Tom on the original PR that I brought this test file.

test/JSON/toJSON.js
MSLaguana: That reminds me actually; can you please add a test case where a single property has both a getter and a setter? That specifically is what caused issues for node, since it meant one property used up more than one slot.

test/JSON/toJSON.js
```diff
@@ -15,12 +15,141 @@ fnc.prototype.toJSON = function() {
...
+  }
+  var obj = {a:1, b:2};
+  Object.defineProperty(obj, "c", { get: () => 3, enumerable: true });
+
+  obj.d = 4;
+  TEST('{"a":99,"b":2,"d":4}', JSON.stringify( JSON.parse(JSON.stringify(obj), rev) ));
+}
+
+// test getter with revive && objectArray
+{
```
akroshg: Is it possible to mix proxy with JSON.stringify? don't have test handy but changing the state when trap is invoked? although that would be irrelevant if this PR is meant to improve/address perf.

obastemur: proxy is handled in separate code place and I didn't change that part so far. However, i.e. `Date .toString` depends on internal `toJSON` and I had added tests to cover those.

obastemur:
@MSLaguana @akroshg @jackhorton Thanks for the reviews

lib/Runtime/Debug/DiagObjectModel.cpp
```diff
@@ -2483,7 +2483,7 @@ namespace Js
                                         else
                                         {
                                             const PropertyRecord* propertyRecord;
-                                            objectContext->GetOrAddPropertyRecord(obj->GetSz(), obj->GetLength(), &propertyRecord);
+                                            objectContext->GetOrAddPropertyRecord(obj, &propertyRecord);
```
akroshg: this is an else part, the 'if' part already checking the VirtualTableInfo thing. That code is not needed anymore right? Could you remove it>

lib/Runtime/Library/ForInObjectEnumerator.cpp
```diff
@@ -194,7 +194,7 @@ namespace Js
                     else
                     {
                         ScriptContext* scriptContext = currentIndex->GetScriptContext();
-                        scriptContext->GetOrAddPropertyRecord(currentIndex->GetString(), currentIndex->GetLength(), &propRecord);
+                        scriptContext->GetOrAddPropertyRecord(currentIndex, &propRecord);
```
akroshg: the 'if' already checks for VirtualTable - now that is not needed anymore now, right?

lib/Runtime/Base/ScriptContext.cpp
```diff
@@ -837,6 +837,18 @@ namespace Js
         return propertyRecord->GetPropertyId();
     }
 
+    void ScriptContext::GetOrAddPropertyRecord(Js::JavascriptString * propertyString, PropertyRecord const** propertyRecord)
+    {
+        if (VirtualTableInfo<Js::PropertyString>::HasVirtualTable(propertyString) && propertyString->GetScriptContext() == this)
```
akroshg: at certain places we have checks for         ```VirtualTableInfo<Js::LiteralStringWithPropertyStringPtr>``` do you need to handle that part as well?

obastemur: https://github.com/Microsoft/ChakraCore/pull/3875

lib/Runtime/Library/ForInObjectEnumerator.cpp
obastemur: @akroshg sorry this one is not clear to me?

lib/Runtime/Debug/DiagObjectModel.cpp
obastemur: good idea. will do separately. (since this PR is merged)

lib/Runtime/Library/ForInObjectEnumerator.cpp
akroshg: the 'if' condition above has a check already
``` if (VirtualTableInfo<Js::PropertyString>::HasVirtualTable(currentIndex))``` (line 187)

---
In reply to: [143252170](https://github.com/Microsoft/ChakraCore/pull/3894#discussion_r143252170) [](ancestors = 143252170)

PR #3903, opened 2017-10-08
ARM64: Fix prolog/epilog

agarwal-sandeep:
Fix BLR and RET opcode encoding. Seems like they were reversed.

Fix prolog and epilog. There is still some work left to optimize STR/LDR to STP/LDP. Also need to make sure stack probe code works correctly.

### DIFF
--- a/lib/Backend/arm64/ARM64Encoder.h
+++ b/lib/Backend/arm64/ARM64Encoder.h
@@ -998,7 +998,7 @@ EmitBlr(
     Arm64SimpleRegisterParam Reg
     )
 {
-    return Emitter.EmitFourBytes(0xd65f0000 | (Reg.RawRegister() << 5));
+    return Emitter.EmitFourBytes(0xd63f0000 | (Reg.RawRegister() << 5));
 }
 
 inline
@@ -1008,7 +1008,7 @@ EmitRet(
     Arm64SimpleRegisterParam Reg
     )
 {
-    return Emitter.EmitFourBytes(0xd63f0000 | (Reg.RawRegister() << 5));
+    return Emitter.EmitFourBytes(0xd65f0000 | (Reg.RawRegister() << 5));
 }
 
 //
--- a/lib/Backend/arm64/LowerMD.cpp
+++ b/lib/Backend/arm64/LowerMD.cpp
@@ -750,11 +750,6 @@ LowererMD::LoadDoubleHelperArgument(IR::Instr * instr, IR::Opnd * opndArg)
 void
 LowererMD::GenerateStackProbe(IR::Instr *insertInstr, bool afterProlog)
 {
-    // ARM64_WORKITEM
-    __debugbreak();
-
-#if 0
-
     //
     // Generate a stack overflow check. This can be as simple as a cmp esp, const
     // because this function is guaranteed to be called on its base thread only.
@@ -803,11 +798,11 @@ LowererMD::GenerateStackProbe(IR::Instr *insertInstr, bool afterProlog)
             this->CreateAssign(scratchOpnd, IR::IndirOpnd::New(scratchOpnd, 0, TyMachReg, this->m_func), insertInstr);
         }
 
-        if (EncoderMD::CanEncodeModConst12(frameSize))
+        if (IS_CONST_00000FFF(frameSize))
         {
             // If the frame size is small enough, just add the constant.
             // Does this ever happen with the size of the MinStackJIT constant?
-            instr = IR::Instr::New(Js::OpCode::ADDS, scratchOpnd, scratchOpnd,
+            instr = IR::Instr::New(Js::OpCode::ADD, scratchOpnd, scratchOpnd,
                                    IR::IntConstOpnd::New(frameSize, TyMachReg, this->m_func), this->m_func);
             insertInstr->InsertBefore(instr);
         }
@@ -823,11 +818,16 @@ LowererMD::GenerateStackProbe(IR::Instr *insertInstr, bool afterProlog)
             if (!afterProlog)
             {
                 opnd = IR::IndirOpnd::New(IR::RegOpnd::New(nullptr, RegSP, TyMachReg, this->m_func), (int32)0, TyMachReg, this->m_func);
-                instr = IR::Instr::New(Js::OpCode::PUSH, opnd, this->m_func);
+                instr = IR::Instr::New(Js::OpCode::STR, opnd, this->m_func);
                 scratchBit.Set(RegEncode[SP_ALLOC_SCRATCH_REG]);
                 opnd = IR::RegBVOpnd::New(scratchBit, TyMachReg, this->m_func);
                 instr->SetSrc1(opnd);
                 insertInstr->InsertBefore(instr);
+
+                IR::RegOpnd* regSpOpnd = IR::RegOpnd::New(nullptr, RegSP, TyMachReg, this->m_func);
+                IR::Instr * instrSub = IR::Instr::New(Js::OpCode::SUB, regSpOpnd, regSpOpnd, IR::IntConstOpnd::New(MachRegInt, TyMachReg, this->m_func), this->m_func);
+                insertInstr->InsertBefore(instrSub);
+
             }
 
             IR::Opnd *scratchOpnd2 = IR::RegOpnd::New(nullptr, SP_ALLOC_SCRATCH_REG, TyMachReg, this->m_func);
@@ -840,10 +840,14 @@ LowererMD::GenerateStackProbe(IR::Instr *insertInstr, bool afterProlog)
             {
                 Assert(scratchBit.Test(RegEncode[SP_ALLOC_SCRATCH_REG]));
                 opnd = IR::RegBVOpnd::New(scratchBit, TyMachReg, this->m_func);
-                instr = IR::Instr::New(Js::OpCode::POP, opnd, this->m_func);
+                instr = IR::Instr::New(Js::OpCode::LDR, opnd, this->m_func);
                 opnd = IR::IndirOpnd::New(IR::RegOpnd::New(nullptr, RegSP, TyMachReg, this->m_func), (int32)0, TyMachReg, this->m_func);
                 instr->SetSrc1(opnd);
                 insertInstr->InsertBefore(instr);
+
+                IR::RegOpnd* regSpOpnd = IR::RegOpnd::New(nullptr, RegSP, TyMachReg, this->m_func);
+                IR::Instr * instrSub = IR::Instr::New(Js::OpCode::ADD, regSpOpnd, regSpOpnd, IR::IntConstOpnd::New(MachRegInt, TyMachReg, this->m_func), this->m_func);
+                insertInstr->InsertBefore(instrSub);
             }
         }
 
@@ -896,8 +900,6 @@ LowererMD::GenerateStackProbe(IR::Instr *insertInstr, bool afterProlog)
 
     insertInstr->InsertBefore(doneLabelInstr);
     Security::InsertRandomFunctionPad(doneLabelInstr);
-
-#endif
 }
 
 //
@@ -1027,17 +1029,9 @@ LowererMD::GenerateStackDeallocation(IR::Instr *instr, uint32 allocSize)
 IR::Instr *
 LowererMD::LowerEntryInstr(IR::EntryInstr * entryInstr)
 {
-
     IR::Instr *insertInstr = entryInstr->m_next;
-
-    // ARM64_WORKITEM
-    //__debugbreak();
-
-//#if 0
-
     BYTE regEncode;
     BOOL hasTry = this->m_func->HasTry();
-    AssertMsg(!hasTry, "ToDo (SaAgarwa): prolog not implemented for try");
 
     // Begin recording info for later pdata/xdata emission.
     UnwindInfoManager *unwindInfo = &this->m_func->m_unwindInfo;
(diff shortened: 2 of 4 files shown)


agarwal-sandeep:
@aaronsgiles please look at commit d7de8f7 (Fix BLR and RET Encoding)

lib/Backend/arm64/ARM64Encoder.h
```diff
@@ -998,7 +998,7 @@ EmitBlr(
     Arm64SimpleRegisterParam Reg
     )
 {
-    return Emitter.EmitFourBytes(0xd65f0000 | (Reg.RawRegister() << 5));
+    return Emitter.EmitFourBytes(0xd63f0000 | (Reg.RawRegister() << 5));
```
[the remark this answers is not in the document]
agarwal-sandeep: For BLR opcode disassembler is showing a RET and PC+4 is not being saved in X30/LR. After swapping the encodings I see BLR and PC+4 is being saved correctly. So I made the change and wanted to verify with you that its correct.

lib/Backend/arm64/LowerMD.cpp
```diff
@@ -1353,89 +1341,95 @@ LowererMD::LowerEntryInstr(IR::EntryInstr * entryInstr)
...
         IR::LabelInstr *prologStartLabel = IR::LabelInstr::New(Js::OpCode::Label, this->m_func);
         insertInstr->InsertBefore(prologStartLabel);
         this->m_func->m_unwindInfo.SetPrologStartLabel(prologStartLabel->m_id);
     }
 
+    int32 regPushedSize = 0;
+
     if (!paramRegs.IsEmpty())
     {
         for (int i = LAST_INT_ARG_REG; i >= FIRST_INT_ARG_REG; --i)
```
aaronsgiles: I think you need a TODO: here to indicate that we will eventually push in pairs. (Unless I missed that code.)

lib/Backend/arm64/LowerMD.cpp
```diff
@@ -1353,89 +1341,95 @@ LowererMD::LowerEntryInstr(IR::EntryInstr * entryInstr)
...
-        frameRegs.Set(RegEncode[RegFP]);
-        frameRegs.Set(RegEncode[RegLR]);
-
-        // Generate PUSH {r11,lr}
-        IR::Instr * instrPush = IR::Instr::New(Js::OpCode::STP, this->m_func);
-        instrPush->SetDst(IR::IndirOpnd::New(IR::RegOpnd::New(nullptr, RegSP, TyMachReg, this->m_func), (int32)0, TyMachReg, this->m_func));
-        instrPush->SetSrc1(IR::RegBVOpnd::New(frameRegs, TyMachReg, this->m_func));
-        insertInstr->InsertBefore(instrPush);
+        // Generate STP {fp,lr}
+        IR::Instr * instrStp = IR::Instr::New(Js::OpCode::STP_PRE, this->m_func);
```
aaronsgiles: The limit of pre-indexed STP is -512, so unless you can guarantee that, you will need to add fallback logic to do a manual stack allocation.

lib/Backend/arm64/LowerMD.cpp
```diff
@@ -1353,89 +1341,95 @@ LowererMD::LowerEntryInstr(IR::EntryInstr * entryInstr)
...
     }
 
     if (!usedDoubleRegs.IsEmpty())
     {
-        // Generate VPUSH {d8-d15}
-        IR::Instr * instrPush = IR::Instr::New(Js::OpCode::STP, this->m_func);
-        instrPush->SetDst(IR::IndirOpnd::New(IR::RegOpnd::New(nullptr, RegSP, TyMachReg, this->m_func), (int32)0, TyMachReg, this->m_func));
-        instrPush->SetSrc1(IR::RegBVOpnd::New(usedDoubleRegs, TyMachReg, this->m_func));
-        insertInstr->InsertBefore(instrPush);
+        for (RegNum reg = LAST_CALLEE_SAVED_DBL_REG; reg >= FIRST_CALLEE_SAVED_DBL_REG; reg = (RegNum)(reg - 1))
```
aaronsgiles: You need to use FSTR here. Also floating point registers can be stored in pairs as well, so pairwise logic should eventually be added to use FSTP.

lib/Backend/arm64/LowerMD.cpp
```diff
@@ -1506,40 +1495,35 @@ LowererMD::LowerEntryInstr(IR::EntryInstr * entryInstr)
         if (r17Opnd == nullptr || isScratchRegisterThrashed)
         {
             r17Opnd = r17Opnd ? r17Opnd : IR::RegOpnd::New(nullptr, SCRATCH_REG, TyMachReg, this->m_func);
-            // mov r12, 0
-            IR::Instr * instrMov = IR::Instr::New(Js::OpCode::MOV, r17Opnd, IR::IntConstOpnd::New(0, TyMachReg, this->m_func), this->m_func);
+            // mov r17, 0
+            IR::Instr * instrMov = IR::Instr::New(Js::OpCode::MOVZ, r17Opnd, IR::IntConstOpnd::New(0, TyMachReg, this->m_func), this->m_func);
```
aaronsgiles: Again, watch the limits of MOVZ (0-65535). You should at least have an Assert in place here to catch it in case.

lib/Backend/arm64/LowerMD.cpp
```diff
@@ -1646,82 +1630,68 @@ LowererMD::LowerExitInstr(IR::ExitInstr * exitInstr)
...
+
     // 2. Restore saved double registers. Generate vpop {d8-d15}
     BVUnit savedDoubleRegs(this->m_func->m_unwindInfo.GetDoubleSavedRegList());
     if (!savedDoubleRegs.IsEmpty())
     {
-        IR::Instr * instrVPop = IR::Instr::New(Js::OpCode::VPOP, this->m_func);
-        instrVPop->SetDst(IR::RegBVOpnd::New(savedDoubleRegs, TyMachReg, this->m_func));
-        instrVPop->SetSrc1(IR::IndirOpnd::New(IR::RegOpnd::New(nullptr, RegSP,TyMachReg, this->m_func), (int32)0, TyMachReg, this->m_func));
-        exitInstr->InsertBefore(instrVPop);
+        for (RegNum reg = FIRST_CALLEE_SAVED_DBL_REG; reg <= LAST_CALLEE_SAVED_DBL_REG; reg = (RegNum)(reg + 1))
```
aaronsgiles: Same comment as above, use LDP to load in pairs for efficiency/code size. It can be a TODO for now, but should be notated.

aaronsgiles: Also since this is the floating point case, use FLDR instead of LDR.

lib/Backend/arm64/LowerMD.cpp
```diff
@@ -1646,82 +1630,68 @@ LowererMD::LowerExitInstr(IR::ExitInstr * exitInstr)
...
-        // SP now points to the location where we saved LR.
-        // So return by doing a LDR pc,[sp],#n, where the postincrement of SP deallocates what remains of the stack.
-        // Note: the offset on this indir indicates the postincrement, which is the homed param area plus the size
-        // of LR itself.
-        IR::IndirOpnd * spIndir = IR::IndirOpnd::New(
-            IR::RegOpnd::New(nullptr, RegSP, TyMachReg, this->m_func),
-             (homedParams + 1) * MachRegInt,
-            TyMachPtr, this->m_func);
+    // Generate LDP {fp,lr}
+    IR::Instr * instrStp = IR::Instr::New(Js::OpCode::LDP_POST, this->m_func);
```
aaronsgiles: Again, LDP is limited to +512 byte offset, so you will need a fallback case for bigger allocations. TODO item for sure.

lib/Backend/arm64/LowerMD.cpp
agarwal-sandeep: Added ToDo comment

---
In reply to: [143539271](https://github.com/Microsoft/ChakraCore/pull/3903#discussion_r143539271) [](ancestors = 143539271)

agarwal-sandeep: Changed

---
In reply to: [143539446](https://github.com/Microsoft/ChakraCore/pull/3903#discussion_r143539446) [](ancestors = 143539446)

lib/Backend/arm64/LowerMD.cpp
agarwal-sandeep: Changed to FSTR

---
In reply to: [143538894](https://github.com/Microsoft/ChakraCore/pull/3903#discussion_r143538894) [](ancestors = 143538894)

lib/Backend/arm64/LowerMD.cpp
```diff
@@ -803,7 +798,7 @@ LowererMD::GenerateStackProbe(IR::Instr *insertInstr, bool afterProlog)
             this->CreateAssign(scratchOpnd, IR::IndirOpnd::New(scratchOpnd, 0, TyMachReg, this->m_func), insertInstr);
         }
 
-        if (EncoderMD::CanEncodeModConst12(frameSize))
+        if (EncoderMD::CanEncodeLogicalConst(frameSize, 4))
```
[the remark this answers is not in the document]
agarwal-sandeep: Changed to ADD

---
In reply to: [143535757](https://github.com/Microsoft/ChakraCore/pull/3903#discussion_r143535757) [](ancestors = 143535757)

lib/Backend/arm64/LowerMD.cpp
agarwal-sandeep: That is the plan. Added a ToDo

---
In reply to: [143538276](https://github.com/Microsoft/ChakraCore/pull/3903#discussion_r143538276) [](ancestors = 143538276)

lib/Backend/arm64/LowerMD.cpp
agarwal-sandeep: Relative to SP? The offset here is only 16

---
In reply to: [143539590](https://github.com/Microsoft/ChakraCore/pull/3903#discussion_r143539590) [](ancestors = 143539590)

aaronsgiles: Sorry you're right, it is only a fixed amount. My bad.

lib/Backend/arm64/LowerMD.cpp
agarwal-sandeep: Only moving 0, should be ok?

---
In reply to: [143539114](https://github.com/Microsoft/ChakraCore/pull/3903#discussion_r143539114) [](ancestors = 143539114)

lib/Backend/arm64/LowerMD.cpp
```diff
@@ -1279,25 +1267,25 @@ LowererMD::LowerEntryInstr(IR::EntryInstr * entryInstr)
...
 
-        //Note: Separate push instruction is generated for r11 & lr push and hence usedRegs mask is not updated with
+        //Note: Separate push instruction is generated for fp & lr push and hence usedRegs mask is not updated with
         //bit mask for these registers.
 
         if (!IsSmallStack(stackAdjust) || useDynamicStackProbe)
         {
             unwindInfo->SetSavedScratchReg(true);
-            if (!usedRegs.Test(RegEncode[SP_ALLOC_SCRATCH_REG])) //If its a large stack and RegR4 is not already saved.
+            if (!usedRegs.Test(RegEncode[SP_ALLOC_SCRATCH_REG])) //If its a large stack and RegR18 is not already saved.
```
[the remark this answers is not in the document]
agarwal-sandeep: Changed SP_ALLOC_SCRATCH_REG to RegR19 as RegR17 is used as SCRATCH_REG.

I will revisit this whole SCRATCH_REG and GenerateStackAllocation logic as it seems on AMR64 _chkstk doesn't have the requirement to pass stack size in R4.

---
In reply to: [143537349](https://github.com/Microsoft/ChakraCore/pull/3903#discussion_r143537349) [](ancestors = 143537349)

lib/Backend/arm64/LowerMD.cpp
aaronsgiles: Yes, that's fine. Sorry still getting used to reading the syntax of these and missed that it was a constant 0.

lib/Backend/arm64/LowerMD.cpp
```diff
@@ -1353,89 +1341,95 @@ LowererMD::LowerEntryInstr(IR::EntryInstr * entryInstr)
...
-
     // Zero-initialize dedicated arguments slot
     if (hasCalls)
     {
         //R17 acts a dummy zero register which we push to arguments slot
         //mov r17, 0
         Assert(r17Opnd == nullptr);
         r17Opnd = IR::RegOpnd::New(nullptr, SCRATCH_REG, TyMachReg, this->m_func);
-        IR::Instr * instrMov = IR::Instr::New(Js::OpCode::MOV, r17Opnd, IR::IntConstOpnd::New(0, TyMachReg, this->m_func), this->m_func);
+        IR::Instr * instrMov = IR::Instr::New(Js::OpCode::MOVZ, r17Opnd, IR::IntConstOpnd::New(0, TyMachReg, this->m_func), this->m_func);
```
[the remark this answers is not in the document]
aaronsgiles: This is OK as well since it's a constant 0.

lib/Backend/arm64/LowerMD.cpp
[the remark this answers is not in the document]
aaronsgiles: Ok, just remember that x19 is callee-save, so you probably can't use it in the prolog. x16/x17 are probably better choices. On ARM64, _chkstk is called with the amount to allocate divided by 16 in x15.

;        The typical calling sequence from the prologue is:
;
;        mov    x15, #immed               ; set requested stack frame size
;        bl     __chkstk                  ; verify stack
;        sub    sp, sp, x15, lsl #4       ; apply stack adjust

I'll need to add some way to encode that SP subtract.

lib/Backend/arm64/LowerMD.cpp
agarwal-sandeep: We are only pushing R0-R7, fp, lr so its 10*8 which is < 512. Added an assert for now.

---
In reply to: [143538688](https://github.com/Microsoft/ChakraCore/pull/3903#discussion_r143538688) [](ancestors = 143538688)

aaronsgiles: As long as you can guarantee it, you're good. Just wanting to be careful.

lib/Backend/arm64/LowerMD.cpp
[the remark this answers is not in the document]
aaronsgiles: New opcode SUB_LSL4 now implemented in commit 764546be7fb30c48586b6260e2b998878f2600b8

PR #3908, opened 2017-10-09
ARM64: convert some uses of TyInt32 to TyMachReg

sigatrev:


lib/Backend/arm64/LowerMD.cpp
```diff
@@ -3323,7 +3323,7 @@ LowererMD::GenerateFastAdd(IR::Instr * instrAdd)
     }
 
     // s1  = SUB src1, 1     -- get rid of one of the tag
-    opndReg = IR::RegOpnd::New(TyInt32, this->m_func);
```
agarwal-sandeep: >TyInt32 [](start = 31, length = 7)

We are doing a Int add TyInt32 is ok here. What was the purpose of making it MachReg?

agarwal-sandeep:
Commit the change to fix the operands mismatch issue. Will fix logic issue next.

PR #3913, opened 2017-10-10
Use Utf8Helper classes in Intl.cpp, clean up error handling

### LINKED ISSUE #3851: Use Utf8Codex helper classes in intl-icu
See @MSLaguana's review of https://github.com/Microsoft/ChakraCore/pull/3840. We may not need to use the StringBufferAutoPtr class at all, and we can get rid of our own math here.

/cc @dilijev

jackhorton:
Fixes #3851

### DIFF
--- a/lib/Common/Codex/Utf8Helper.h
+++ b/lib/Common/Codex/Utf8Helper.h
@@ -3,6 +3,7 @@
 // Licensed under the MIT license. See LICENSE.txt file in the project root for full license information.
 //-------------------------------------------------------------------------------------------------------
 #pragma once
+#include "CommonPal.h" // MAXUINT32
 #include "Utf8Codex.h"
 
 namespace utf8
--- a/lib/Runtime/Library/IntlEngineInterfaceExtensionObject.cpp
+++ b/lib/Runtime/Library/IntlEngineInterfaceExtensionObject.cpp
@@ -566,7 +566,7 @@ namespace Js
         }
 
 #if DEBUG
-#ifdef INTL_ICU_DEBUG
+#if INTL_ICU_DEBUG
         Output::Print(_u("EntryIntl_RaiseAssert\n"));
 #endif
         JavascriptExceptionOperators::Throw(JavascriptError::FromVar(args.Values[1]), scriptContext);
@@ -654,7 +654,7 @@ namespace Js
             return JavascriptString::NewCopySz(resolvedLocaleName, scriptContext);
         }
 
-#ifdef INTL_ICU_DEBUG
+#if INTL_ICU_DEBUG
         Output::Print(_u("Intl::ResolveLocaleLookup returned false: EntryIntl_ResolveLocaleLookup returning null to fallback to JS\n"));
 #endif
         return scriptContext->GetLibrary()->GetNull();
@@ -693,7 +693,7 @@ namespace Js
             return JavascriptString::NewCopySz(resolvedLocaleName, scriptContext);
         }
 
-#ifdef INTL_ICU_DEBUG
+#if INTL_ICU_DEBUG
         Output::Print(_u("Intl::ResolveLocaleBestFit returned false: EntryIntl_ResolveLocaleBestFit returning null to fallback to JS\n"));
 #endif
         return scriptContext->GetLibrary()->GetNull();
@@ -798,7 +798,7 @@ namespace Js
         return toReturn;
 #else
         // TODO (doilij): implement INTL_ICU version
-#ifdef INTL_ICU_DEBUG
+#if INTL_ICU_DEBUG
         Output::Print(_u("EntryIntl_GetExtensions > returning null: fallback to JS function getExtensionSubtags\n"));
 #endif
         return scriptContext->GetLibrary()->GetNull(); // fallback to Intl.js
@@ -1173,7 +1173,7 @@ namespace Js
         return scriptContext->GetLibrary()->GetUndefined();
 #else
         // TODO (doilij): implement INTL_ICU version
-#ifdef INTL_ICU_DEBUG
+#if INTL_ICU_DEBUG
         Output::Print(_u("EntryIntl_CreateDateTimeFormat > returning null, fallback to JS\n"));
 #endif
         return scriptContext->GetLibrary()->GetNull();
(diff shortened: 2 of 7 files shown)


lib/Runtime/PlatformAgnostic/Platform/Common/Intl.cpp
```diff
@@ -62,116 +51,47 @@ namespace Intl
...
-        forLangTagResultLength = uloc_forLanguageTag(reinterpret_cast<char *>(inputLangTagUtf8),
-            icuLocaleId, ULOC_FULLNAME_CAPACITY, &parsedLength, &error);
-        success = (forLangTagResultLength > 0) && (parsedLength > 0) &&
-            U_SUCCESS(error) && ((size_t)parsedLength == inputLangTagUtf8SizeActual);
-        if (!success)
-        {
-            return false;
-        }
+        int32_t forLangTagResultLength = uloc_forLanguageTag(languageTagUtf8, icuLocaleId, ULOC_FULLNAME_CAPACITY, &parsedLength, &error);
+        ICU_RETURN(error, forLangTagResultLength > 0 || parsedLength > 0 || ((size_t) parsedLength) == languageTagUtf8.Length(), false);
```
dilijev: why the change from && to || ?

lib/Runtime/Library/RuntimeLibraryPch.h
```diff
@@ -95,8 +95,16 @@
...
 #ifdef INTL_ICU
+
+#ifndef U_STATIC_IMPLEMENTATION
 #define U_STATIC_IMPLEMENTATION
+#endif
+
+#ifdef U_SHOW_CPLUSPLUS_API
+#undef U_SHOW_CPLUSPLUS_API
 #define U_SHOW_CPLUSPLUS_API 0
+#endif
```
dilijev: These defines are only necessary if this compilation unit uses the ICU headers. I think you can safely remove this block now.

jackhorton: Yeah I think they were necessitated by moving the includes to intl.h but that doesn't work on Linux so I am reverting it now.

lib/Runtime/PlatformAgnostic/Platform/Common/Intl.cpp
jackhorton: Because I had ICU_RETURN checking for true rather than false for a while and had to negate the entire boolean expression, but then I switched it back and only half-un-negated it. Good catch!

jackhorton:
@obastemur would you mind taking a look at the OS X build failures? I had to change some of the includes to make SAL _and_ Utf8Helpers happy, and it seems like something in PAL is conflicting with stdlib stuff on OS X in a way it doesn't on Linux. I don't have a Mac handy, or else I would look into this myself.

lib/Runtime/PlatformAgnostic/RuntimePlatformAgnosticPch.h
```diff
@@ -7,7 +7,7 @@
 #ifdef _WIN32
 #include "CommonMin.h"
 #else
-#include "pal.h"
+#include "CommonPal.h"
```
obastemur: ?

lib/Common/Codex/Utf8Helper.h
```diff
@@ -3,6 +3,7 @@
 // Licensed under the MIT license. See LICENSE.txt file in the project root for full license information.
 //-------------------------------------------------------------------------------------------------------
 #pragma once
+#include "CommonPal.h" // MAXUINT32
```
obastemur: ?

obastemur:
@jackhorton could be nice to have some context on `includes` you have introduced.

lib/Common/Codex/Utf8Helper.h
jackhorton: Utf8Helper.h uses MAXUINT32, which was probably getting defined by accident each time before Utf8Helper.h got included. Including Utf8Helper.h in PlatformAgnostic/Common/Intl.cpp caused this to come up.

lib/Runtime/PlatformAgnostic/RuntimePlatformAgnosticPch.h
jackhorton: Utf8Helper.h uses wcslen, so it has to be available _and_ properly `#define`d to PAL_wcslen on xplat. Simply including pal.h didn't get that done, since there needs to be a combination of #defines in order to make that work. I could try to recreate the defines, but my initial stab caused other errors on Linux, and including CommonPal.h seemed easier (and it matches what we do on win32, where we include CommonMin.h)

jackhorton:
The new includes are _definitely_ why the CI is failing, that's for certain. I am still trying to work through how we were able to use Utf8Codex.h before with no issues whatsoever, but now switching that to Utf8Helper.h causes all of these issues.

jackhorton:
Switching to Utf8Helper classes proved to be much more difficult than I thought it would be. I will close this for now and may re-open something later to use those classes. In the meantime, any objections to factoring the new asserts out of this PR into a separate one, since those should be much easier to integrate?

lib/Runtime/Library/IntlEngineInterfaceExtensionObject.cpp
```diff
@@ -566,7 +566,7 @@ namespace Js
         }
 
 #if DEBUG
-#ifdef INTL_ICU_DEBUG
+#if INTL_ICU_DEBUG
```
dilijev: Side note: FWIW I'm not a fan of this change. I commented on your other PR as well about the idea here, but the fact that there is a value given to this macro in `#define` is defensive -- the point is actually that it's defined at all.

dilijev:
@jackhorton to which asserts are you referring?

lib/Runtime/Library/IntlEngineInterfaceExtensionObject.cpp
jackhorton: The reason for the change was to allow `if (INTL_ICU_DEBUG) Output::Print()` in the ICU_ASSERT/ICU_RETURN macros, since _as far as I know_ (I may be very wrong), I cant have #ifdefs in macros (I would need to conditionally define the entire macro just for printing purposes). Though, thinking about it, I suppose I could just macro the Output::Print part conditionally on its own?

jackhorton:
"Factoring out the new asserts" was supposed to mean factoring out the new assert macros. Basically, taking the majority of this PR aside from the WideToNarrow class usage and splitting it out.

dilijev:
Oh yeah, definitely.

PR #3922, opened 2017-10-10
Fix benign UAF in peeps.

Penguinwizzard:
We've already called ->Remove on instr, which means that the uses of
it in the condition for the if block are technically UAFs. This does
not result in an exploitable path, though, since we do no allocation
here and don't give back the reference. Additionally, the old opcode
can only be ADD, SUB, or OR, so the check against NOP fails.


Penguinwizzard:
Was reviewed internally previously.

PR #3923, opened 2017-10-10
json.stringify: simplify array interface and make sure object is dynamicObject 

obastemur:
- stringify(array)   implementation was including some dead code and redundant type checks. Partly handled by LTO, partly not.
- stringify(object) implementation adds more strict checking around dynamicObject.
- stringify(object) measure ConcatString item count precisely so it doesn't go through expensive resize path

### DIFF
--- a/lib/Runtime/Library/JSON.cpp
+++ b/lib/Runtime/Library/JSON.cpp
@@ -397,23 +397,22 @@ namespace JSON
         objectStack = Anew(tempAlloc, JSONStack, tempAlloc, scriptContext);
     }
 
-    Js::Var StringifySession::Str(uint32 index, Js::Var holder)
+    Js::Var StringifySession::Str(uint32 index, Js::RecyclableObject * holder)
     {
         Js::Var value = nullptr;
         Js::RecyclableObject *undefined = scriptContext->GetLibrary()->GetUndefined();
 
-        if (Js::JavascriptArray::Is(holder) && !Js::JavascriptArray::FromVar(holder)->IsCrossSiteObject())
+        if (Js::JavascriptArray::Is(holder->GetTypeId()) && !Js::JavascriptArray::FromAnyArray(holder)->IsCrossSiteObject())
         {
-            if (Js::JavascriptOperators::IsUndefinedObject(value = Js::JavascriptArray::FromVar(holder)->DirectGetItem(index), undefined))
+            value = Js::JavascriptArray::FromAnyArray(holder)->DirectGetItem(index);
+            if (Js::JavascriptOperators::IsUndefinedObject(value, undefined))
             {
                 return value;
             }
         }
         else
         {
-            Assert(Js::JavascriptOperators::IsArray(holder));
-            Js::RecyclableObject *arr = RecyclableObject::FromVar(holder);
-            if (!Js::JavascriptOperators::GetItem(arr, index, &value, scriptContext))
+            if (!Js::JavascriptOperators::GetItem(holder, index, &value, scriptContext))
             {
                 return undefined;
             }
@@ -585,7 +584,7 @@ namespace JSON
 
         default:
             Js::Var ret = undefined;
-            if(Js::JavascriptOperators::IsJsNativeObject(value))
+            if (Js::JavascriptOperators::IsJsNativeObject(value))
             {
                 if (!Js::JavascriptConversion::IsCallable(value))
                 {
@@ -595,9 +594,9 @@ namespace JSON
                     }
                     objectStack->Push(value);
 
-                    if(Js::JavascriptOperators::IsArray(value))
+                    if (Js::JavascriptOperators::IsArray(value))
                     {
-                        ret = StringifyArray(value);
+                        ret = StringifyArray(Js::RecyclableObject::FromVar(value));
                     }
                     else
                     {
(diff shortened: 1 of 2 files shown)


lib/Runtime/Library/JSON.cpp
```diff
@@ -397,23 +397,22 @@ namespace JSON
         objectStack = Anew(tempAlloc, JSONStack, tempAlloc, scriptContext);
     }
 
-    Js::Var StringifySession::Str(uint32 index, Js::Var holder)
+    Js::Var StringifySession::Str(uint32 index, Js::JavascriptArray * holder)
```
MSLaguana: Why have you changed the `holder` to be a `JavascriptArray`? From what I recall of the spec, it's potentially an object, but not necessarily an array.

MSLaguana: See, I think, https://www.ecma-international.org/ecma-262/8.0/index.html#sec-serializejsonproperty

obastemur: Agree with comments. I'm more interested with the optimized path for javascriptArray types only. i.e. better consider proxy separately?

This wasn't planned to be part of 1.7 and no rush.

lib/Runtime/Library/JSON.cpp
```diff
@@ -682,56 +681,90 @@ namespace JSON
...
-                        // if object has an objectarray, (aka indexed properties)
+                        // if object has an objectArray, (aka indexed properties)
                         // we need to loop them first since they won't show up on the second loop below
-                        if (dynamicObject->HasObjectArray())
+                        // besides, we might be talking to a non dynamicObject
+                        // in that case, we will also take the enumerator path
+                        if (!isDynamicObject || hasObjectArray)
                         {
+                            // get the precise count because resizing ConcatString will be expensive!
+                            propertyCount = (uint32) this->GetPropertyCount(object, &enumerator);
```
MSLaguana: Why do you cast to `uint32` here? Doesn't the function already return a `uint32`?

lib/Runtime/Library/JSON.cpp
```diff
@@ -682,56 +681,90 @@ namespace JSON
...
-                    Js::DynamicObject * dynamicObject = ((Js::DynamicObject*)object);
-                    if (ReplacerFunction != replacerType || !dynamicObject->HasObjectArray())
+                    bool isDynamicObject = Js::DynamicObject::Is(object);
+                    Js::DynamicObject * dynamicObject = nullptr;
+                    bool hasObjectArray = false;
+                    if (isDynamicObject)
                     {
-                        uint32 propertyCount = dynamicObject->GetPropertyCount();
+                        dynamicObject = Js::DynamicObject::FromVar(object);
+                        if (dynamicObject->HasObjectArray())
```
MSLaguana: I'd suggest just `hasObjectArray = dynamicObject->hasObjectArray()`, without the redundant conditional.

lib/Runtime/Library/JSON.cpp
```diff
@@ -682,56 +681,90 @@ namespace JSON
...
+                            result = Js::ConcatStringBuilder::New(this->scriptContext, propertyCount);
+
                             enumerator.Reset();
-                            while ((propertyName = enumerator.MoveAndGetNext(id)) != NULL)
+                            while (propertyCount > 0 && (propertyName = enumerator.MoveAndGetNext(id)) != NULL)
                             {
                                 // if unsuccessful get propertyId from the string
                                 scriptContext->GetOrAddPropertyRecord(propertyName, &propRecord);
-                                if (!propRecord->IsNumeric()) break;
+                                if (isDynamicObject && !propRecord->IsNumeric())
```
MSLaguana: What happens if a non-`DynamicObject` has non-numeric properties here? Why do we treat that differently to a `DynamicObject` with non-numeric properties?

obastemur: when it's non-dynamicObject, we want to use this enumator loop only because overridden type might behave differently.

lib/Runtime/Library/JSON.cpp
```diff
@@ -682,56 +681,90 @@ namespace JSON
...
+                        dynamicObject = Js::DynamicObject::FromVar(object);
+                        if (dynamicObject->HasObjectArray())
+                        {
+                            hasObjectArray = dynamicObject->HasObjectArray();
+                        }
+                    }
 
-                        result = Js::ConcatStringBuilder::New(this->scriptContext, propertyCount);
-                        Js::DynamicTypeHandler * typeHandler = dynamicObject->GetTypeHandler();
+                    if (ReplacerFunction != replacerType || !hasObjectArray)
```
MSLaguana: Could you please add the `!isDynamicObject` condition here? I know it is redundant since that requires that `!hasObjectArray`, but that's not immediately clear and I spent a while trying to work out what combination of this conditional and the following conditional could happen.

obastemur: `!hasObjectArray` != `!isDynamicObject`

obastemur: Condition here is.. Everything but `ES5 && ReplacerFunction == replacerType. && object->HasObjectArray`

obastemur: Because, when there is a replacer function && we need to do an enumerate-loop (because of objectArray), we will need to store the existing properties. That's what the `else` of `this if` does

MSLaguana: Ah; I was mis-reading the indentation depth. Let me take a closer look and make sure I understand.

obastemur: Actually, ruling out `!dynamicObject && replacerType != replacerFunction` was a good idea

obastemur:
@MSLaguana Thanks for the review! Sounds good.

PR #3924, opened 2017-10-11
Avoid potential future uninitialized read in Regex

Penguinwizzard:
We assume in a couple places that these unicode equivalence classes
have no pure identity rows (i.e. rows where all four values are the
same). In these cases, some other parts of our code may leak unsafe
data to the running script. This static assert should help avoid an
error here getting through and causing issues elsewhere.

I also added an Assert at the immediately visible location where an
incorrect identity row would allow uninitialized reads, so that the
analyzers know about this information.

### DIFF
--- a/lib/Parser/CaseInsensitive.cpp
+++ b/lib/Parser/CaseInsensitive.cpp
@@ -581,7 +581,20 @@ END {
 
         static constexpr bool isTransformEntryValid(int offset)
         {
-            return offset >= numTransforms || ((transforms[offset].delta0 != transforms[offset].delta1 || transforms[offset].delta0 != transforms[offset].delta2 || transforms[offset].delta0 != transforms[offset].delta3) && isTransformEntryValid(offset + 1));
+            return
+                // If we've made it to the end of the table, return true. (base case)
+                offset >= numTransforms ||
+                // Otherwise, the transformation table is valid if
+                (
+                    // there is at least one delta that differs from delta0 in the transform entry that we're currently looking at
+                    (transforms[offset].delta0 != transforms[offset].delta1 || transforms[offset].delta0 != transforms[offset].delta2 || transforms[offset].delta0 != transforms[offset].delta3)
+                    // and the transformation range is ordered properly
+                    && (transforms[offset].lo <= transforms[offset].hi)
+                    // and the length of the range (inclusive of ends) is a multiple is the skipcount
+                    && ( ((1 + transforms[offset].hi - transforms[offset].lo) % transforms[offset].skipCountOfRange) == 0)
+                    // and the rest of the transformation table is valid. (compile-time recursive call)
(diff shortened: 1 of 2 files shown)


lib/Parser/CaseInsensitive.cpp
```diff
@@ -579,9 +579,18 @@ END {
         static const int numTransforms = sizeof(transforms) / sizeof(Transform);
         static const Transform lastTransform = transforms[numTransforms - 1];
 
+        static constexpr bool isTransformEntryValid(int offset)
+        {
+            return offset >= numTransforms || ((transforms[offset].delta0 != transforms[offset].delta1 || transforms[offset].delta0 != transforms[offset].delta2 || transforms[offset].delta0 != transforms[offset].delta3) && isTransformEntryValid(offset + 1));
```
dilijev: Break up this line to make it more readable, and add a comment about the static recursive call.

lib/Parser/CaseInsensitive.cpp
```diff
@@ -579,9 +579,18 @@ END {
...
+        }
+
         template <typename Char, typename Fn>
         bool RangeToEquivClass(uint& tblidx, uint l, uint h, uint& acth, Char equivl[EquivClassSize], Fn acceptSource)
         {
+            // Note: There's a few places where we assume that there's no equivalence set
+            // with only one actual member. If you fail this check, double-check the data
+            // in the table above - one line likely has the same value for all deltas.
+            static_assert(isTransformEntryValid(0), "Invalid Transform code - check for 4 identical deltas!");
```
dilijev: Add comment about the 0 base case for static recursive call.

lib/Parser/RegexCompileTime.cpp
```diff
@@ -1281,6 +1281,7 @@ namespace UnifiedRegex
         {
             Char uniqueEquivs[CaseInsensitive::EquivClassSize];
             CharCount uniqueEquivCount = FindUniqueEquivs(cs, uniqueEquivs);
+            Assert(uniqueEquivCount >= 2);
             switch (uniqueEquivCount)
             {
             case 2:
```
dilijev: Maybe defensively add a case 1? (`EMIT(compiler, MatchChar2Inst, uniqueEquivs[0])`

Penguinwizzard: Unfortunately, there's no MatchChar1Inst; I've made it a failfast, however.

lib/Parser/CaseInsensitive.cpp
Penguinwizzard: done.

lib/Parser/CaseInsensitive.cpp
Penguinwizzard: Done, and I added checks for two more properties of the table.

PR #3931, opened 2017-10-11
Add JSRT API JsLessThan (Fix #3568)

### LINKED ISSUE #3568: JSRT API to access javascript relational comparison function
It would be tremendously convenient if there were a JSRT API that could compare the magnitude of two JsValueRef's, as per the the ecmascript [abstract relational comparison algorithm](https://tc39.github.io/ecma262/#sec-abstract-relational-comparison).

xiaoyinl:
This PR implements the API JsLessThan and JsLessThanOrEqual, proposed in issue #3568.

Fixes #3568

Todo:
- [x] Add more test cases
- [x] Update documents (Microsoft/ChakraCore-wiki#43)

### DIFF
--- a/lib/Jsrt/Jsrt.cpp
+++ b/lib/Jsrt/Jsrt.cpp
@@ -2221,6 +2221,36 @@ CHAKRA_API JsGetIndexedPropertiesExternalData(
     END_JSRT_NO_EXCEPTION
 }
 
+CHAKRA_API JsLessThan(_In_ JsValueRef object1, _In_ JsValueRef object2, _Out_ bool *result)
+{
+    return ContextAPIWrapper<JSRT_MAYBE_TRUE>([&](Js::ScriptContext *scriptContext, TTDRecorder& _actionEntryPopper) -> JsErrorCode {
+        PERFORM_JSRT_TTD_RECORD_ACTION(scriptContext, RecordJsRTLessThan, object1, object2, false);
+
+        VALIDATE_INCOMING_REFERENCE(object1, scriptContext);
+        VALIDATE_INCOMING_REFERENCE(object2, scriptContext);
+        PARAM_NOT_NULL(result);
+
+        *result = Js::JavascriptOperators::Less((Js::Var)object1, (Js::Var)object2, scriptContext) != 0;
+
+        return JsNoError;
+    });
+}
+
+CHAKRA_API JsLessThanOrEqual(_In_ JsValueRef object1, _In_ JsValueRef object2, _Out_ bool *result)
+{
+    return ContextAPIWrapper<JSRT_MAYBE_TRUE>([&](Js::ScriptContext *scriptContext, TTDRecorder& _actionEntryPopper) -> JsErrorCode {
+        PERFORM_JSRT_TTD_RECORD_ACTION(scriptContext, RecordJsRTLessThan, object1, object2, true);
+
+        VALIDATE_INCOMING_REFERENCE(object1, scriptContext);
+        VALIDATE_INCOMING_REFERENCE(object2, scriptContext);
+        PARAM_NOT_NULL(result);
+
+        *result = Js::JavascriptOperators::LessEqual((Js::Var)object1, (Js::Var)object2, scriptContext) != 0;
+
+        return JsNoError;
+    });
+}
+
 CHAKRA_API JsEquals(_In_ JsValueRef object1, _In_ JsValueRef object2, _Out_ bool *result)
 {
     return ContextAPIWrapper<JSRT_MAYBE_TRUE>([&] (Js::ScriptContext *scriptContext, TTDRecorder& _actionEntryPopper) -> JsErrorCode {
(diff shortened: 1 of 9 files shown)


xiaoyinl:

dilijev:
Would it make sense to add the other operators as well? (LessThanOrEqual, GreaterThan, GreaterThanOrEqual) I know that they can all be implemented in terms of JsLessThan and JsEqual, but adding wrappers for these primitives doesn't seem like a bad idea. @liminzhu @Cellule opinions?

dilijev:
@xiaoyinl When you make your next update to add documentation (assuming you still have a work item there), could you rebase this PR to a single commit?

xiaoyinl:
@dilijev The way this PR implements LessThan is that, as proposed by @liminzhu, it takes a parameter `allowsEqual` to control whether this API is LessThan or LessThanOrEqual. I also think maybe having two separate functions `JsLessThan` and `JsLessThanOrEqual` makes code more readable than `JsLessThan(..., true, ...)`.

If you have decided to separate `JsLessThan` and `JsLessThanOrEqual`, I'll update this PR, and rebase it at that time.

dilijev:
@xiaoyinl I didn't realize the discussion on the issue at first. I'm not familiar with the discussion here. In general we want to keep the API surface small so let's not add a bunch more operations without more unanimous approval. But I agree JsLessThan(..., true, ...) doesn't do a good job of capturing the idea of `<=`.

/cc @liminzhu @bterlson @marktjsonar who were involved in the issue discussion.

xiaoyinl:
@dilijev If I remember it correctly, there is an edge case that makes `JsLessThan(...) || JsEqual(...)` not equivalent to `JsLessThanOrEqual(...)`:

    let a = undefined;
    a <= a;  // false
    a < a;   // false
    a == a;  // true
    a === a; // true

liminzhu:
Ole! Thanks for the PR @xiaoyinl ! 

@xiaoyinl @dilijev having separate ```JsLessThan``` and ```JsLessThanOrEqual``` sounds fine to me. I'd hold off on > and >=. The only time someone would want it with the presence of < and <= is when the sequence of evaluating lhs and rhs matters, which does not sound common to me. It's probably better if we wait till a legitimate scenario presents itself before we add extra APIs.

```JsLessThan || JsEquals``` is not a good proxy for <=. Besides what @xiaoyinl mentioned, JS is all messed up when comparing objects. Example, 

```js
let x = {a: 1};
let y = {b: 2};
x > y  // false
x < y // false
x == y // false
x === y // false
x >= y // true
x <= y // true
```

liminzhu:
@akroshg who is a good person to code-review JSRT changes?

dilijev:
@liminzhu I would have thought you. :) Maybe @boingoing?

liminzhu:
@dilijev well I care much about the interface but usually don't look into .cpp's so another pair of eyes would be nice :)

lib/Jsrt/Jsrt.cpp
```diff
@@ -2221,6 +2221,36 @@ CHAKRA_API JsGetIndexedPropertiesExternalData(
     END_JSRT_NO_EXCEPTION
 }
 
+CHAKRA_API JsLessThan(_In_ JsValueRef object1, _In_ JsValueRef object2, _Out_ bool *result)
+{
+    return ContextAPIWrapper<JSRT_MAYBE_TRUE>([&](Js::ScriptContext *scriptContext, TTDRecorder& _actionEntryPopper) -> JsErrorCode {
+        PERFORM_JSRT_TTD_RECORD_ACTION(scriptContext, RecordJsRTEquals, object1, object2, false);
```
boingoing: I'm pretty sure we will need a new TTD action for recording these. Right now we're recording doing a non-strict-equals with `RecordJsRTEquals` and the flag value `false` to indicate not `is-strict`. We probably want to add a new action like `RecordJsRTLessThan` with a flag to say if it should be `or-equal`.

lib/Jsrt/Jsrt.cpp
```diff
@@ -2239,7 +2239,7 @@ CHAKRA_API JsLessThan(_In_ JsValueRef object1, _In_ JsValueRef object2, _Out_ bo
 CHAKRA_API JsLessThanOrEqual(_In_ JsValueRef object1, _In_ JsValueRef object2, _Out_ bool *result)
 {
     return ContextAPIWrapper<JSRT_MAYBE_TRUE>([&](Js::ScriptContext *scriptContext, TTDRecorder& _actionEntryPopper) -> JsErrorCode {
-        PERFORM_JSRT_TTD_RECORD_ACTION(scriptContext, RecordJsRTEquals, object1, object2, false);
+        PERFORM_JSRT_TTD_RECORD_ACTION(scriptContext, RecordJsRTLessThan, object1, object2, false);
```
boingoing: Looks like this should pass true for `allowsEqual`, right?

xiaoyinl:
@boingoing Thank you for the detailed explanation. I have done what you mentioned. Hopefully this time I didn't miss anything.

xiaoyinl:
Thank you again for the review! I have squashed the commits.

dilijev:
Testing internally; then will merge.

PR #3934, opened 2017-10-11
Notify host Dynamic Module ready (Fixes #3902)

### LINKED ISSUE #3902: Dynamic import from within an ES module never resolves
A dynamic `import()` of a module that wasn't already imported statically will never resolve.  This only affects `import()` done in an ES module--dynamic imports loaded using the FetchImportedModuleFromScript strategy resolve just fine.

Repro:
```js
// test.js
console.log("running test.js");
WScript.LoadScriptFile('./pig.js', 'module');
```
```js
// pig.js
console.log("running pig.js");
console.log("starting to load cow.js");
import("./cow.js").then(module => {
	console.log("successfully loaded cow.mjs");
	console.log(`the cow says: ${module.Cow}`);
}).catch(e => {
	console.log(`error: ${e}`);
});
```
```js
// cow.js
console.log("running cow.js");
export const Cow = "cowz eat kittiez";
```

Output:
```
D:\src\ChakraCore\Build\VcBuild\bin\x64_debug>ch test.js
running test.js
running pig.js
starting to load cow.js
```

By contrast, `import()`ing from global code:
```
D:\src\ChakraCore\Build\VcBuild\bin\x64_debug>ch pig.js
running pig.js
starting to load cow.js
running cow.js
successfully loaded cow.mjs
the cow says: cowz eat kittiez
```

rhuanjl:
Previously the call to notify the Host that a module was ready was made only for:
a) Errors AND
b) Parent level modules

A module loaded with a dynamic import may not be a parent and hopefully won't always be an error - this condition change means that the notification is made for any dynamic import when completed.

Condition check used of having an attached promise was chosen to be the same as the condition used for Dynamics on the failed import path on line 194.

Fixes: https://github.com/Microsoft/ChakraCore/issues/3902

### DIFF
--- a/lib/Runtime/Language/SourceTextModuleRecord.cpp
+++ b/lib/Runtime/Language/SourceTextModuleRecord.cpp
@@ -307,7 +307,7 @@ namespace Js
             OUTPUT_TRACE_DEBUGONLY(Js::ModulePhase, _u("\t>NotifyParentsAsNeeded\n"));
             NotifyParentsAsNeeded();
 
-            if (!WasDeclarationInitialized() && isRootModule)
+            if (!WasDeclarationInitialized() && (isRootModule || this->promise != nullptr))
             {
                 // TODO: move this as a promise call? if parser is called from a different thread
                 // We'll need to call the bytecode gen in the main thread as we are accessing GC.
--- a/test/es6/bug_issue_3257_script/script0.js
+++ b/test/es6/bug_issue_3257_script/script0.js
@@ -4,5 +4,5 @@
 //-------------------------------------------------------------------------------------------------------
 
 console.log("script0");
-import('../bug_issue_3257_mod1.js');
-import('../bug_issue_3257_mod2/mod2.js');
+import('bug_issue_3257_mod1.js');
+import('bug_issue_3257_mod2/mod2.js');
--- a/test/es6/dynamic-module-functionality.js
+++ b/test/es6/dynamic-module-functionality.js
@@ -59,6 +59,25 @@ function testDynamicImport(importFunc, thenFunc, catchFunc, _asyncEnter, _asyncE
     });
 }
 
+function testDoubleDynamicImport(importFunc, secondFunc, thirdFunc, catchFunc, _asyncEnter, _asyncExit) {
+	let promise = importFunc();
+    assert.isTrue(promise instanceof Promise);
+    promise.then((v)=>{
+        _asyncEnter();
+		let secondPromise = secondFunc(v);
+		secondPromise.then((v2)=>{
+			_asyncEnter();
+			thirdFunc(v2);
+			_asyncExit();
+		});
+        _asyncExit();
+	}).catch((err)=>{
+        _asyncEnter();
+        catchFunc(err);
+        _asyncExit();
+	});
+}
+
 var tests = [
     {
         name: "Runtime evaluation of module specifier",
@@ -367,6 +386,21 @@ var tests = [
             testModuleScript(functionBody, "Test importing a simple exported function", false, true);
         }
     },
+    {
+        name: "Test dynamic import of an un-parsed module from a module",
+        body: function () {
+            let functionBody =
+                `testDoubleDynamicImport(
+                    ()=>{ return import('testDynamicImportfromModule.js'); },
+					(v)=>{ return v.foo; },
+					(v2)=>{ assert.areEqual(true, v.foo.success, "Failed to load module dynamicly from module");},
+                    (err)=>{ assert.fail(err.message); }, _asyncEnter, _asyncExit
+                )`;
+			testModuleScript(functionBody, "Test dynamic import of an un-parsed module from a module", false, true);
+			//note exclusion of testScript case intentional - running the code from a script loads the module
+			//then the test from Module uses the one loaded by the script - do not add testScript here
+        }
+	},
 ];
 
 testRunner.runTests(tests, { verbose: WScript.Arguments[0] != "summary" });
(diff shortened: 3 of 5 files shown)


lib/Runtime/Language/SourceTextModuleRecord.cpp
```diff
@@ -307,7 +307,7 @@ namespace Js
             OUTPUT_TRACE_DEBUGONLY(Js::ModulePhase, _u("\t>NotifyParentsAsNeeded\n"));
             NotifyParentsAsNeeded();
 
-            if (!WasDeclarationInitialized() && isRootModule)
+            if (!WasDeclarationInitialized() && (isRootModule || this->promise != nullptr))
```
akroshg: @rhuanjl thanks for fixing this code. I would like the test for this issue to be added so that we could catch future regression.  Could you add test for this issue? Preferably in test\es6\dynamic-module-functionality.js file.

rhuanjl: @akroshg I've written a test which I think covers this.

Concerned about the CI fails though, as the test it's failing is not entirely unrelated. This PR was to fix static imports that lead to dynamic imports - which were failing silently.

The CI test this is failing on is assessing dynamic imports which lead to static imports, and is apparently leading to ch segfaulting.

If I run the test individually with ch offline it passes - however if I run the whole framework offline it does fail, not sure what could of caused this - particularly confused as the individual test seems to run fine it's only in the context of the testing framework that it fails.

akroshg: you are hitting this assert 
script0
3>ASSERTION 2300: (D:\j\workspace\x86_debug_prtest06343b35\lib\Runtime\Language\SourceTextModuleRecord.cpp, line 806) moduleRecord->WasParsed()

could you run the test directly with ```-forcenative``` flag to see if this reproes?

akroshg: btw you could repro without forcenative as well, no?
```
core\Build\VcBuild\bin\x64_debug\ch.exe bug_issue_3257.js
script0
ASSERTION 24864: (C:\cg\core\lib\Runtime\Language\SourceTextModuleRecord.cpp, line 806) moduleRecord->WasParsed()
 Failure: (moduleRecord->WasParsed())
```

rhuanjl: Sorry I was testing with a test build switching to a debug build I now see this assert fail.

Having looked into this further I think my fix may have introduced a different error - I'll need to go through this again in more detail - oddly it runs fine for me with a TestBuild or a Release build but fails the assert with a Debug build and if I comment out the assert the Debug build Segfaults.

I'll endeavour work out I'm missing and fix later this week or early next; thanks for the help.

rhuanjl: Ok Two actions required before this can be merged, posting this now Because I would like some input on point 1:
1. Need to understand the correct fix to issue number 3306 - this change is reversing part of a previous fix to that issue - I think there is another fix needed that was hidden by the change I'm reversing (more below).

2. I need to learn how your test framework works my attempts so far have failed.

**More detail on point 1**
My fix is actually reversing part of this previous fix: https://github.com/Microsoft/ChakraCore/issues/3306

The fix posted for that issue involved removing the case I've added back - this suppressed the test failure but left no path for the host to be notified about a dynamic module that was not a root module.

The specific test failure I was seeing, bug_issue_3257, (also referenced in 3306) was due to CC attempting to execute a module that it didn't believe had been parsed.

**Process of the test - prior to any edit:**
1. Load a module which loads other modules staticly (these are all loaded and the host notified for them)
2. Load a script which dynamicly loads 2 of the already loaded modules

Execution process for point 2:
a) ch normalises the names of the modules noting it has already loaded modules by those names - returns the same specifier
b) the host is not asked to fetch them as there is already a specifier for a module that has been fetched
c) the pre-existing specifier is a child-specifier as they were previously loaded through static imports from a different module
d) the host is never notified that they're ready because the impact of issue 3902 is that the host is never notified for child dynamic imports

**Result of just adding in my bugfix for ChakraCore:**
Point(d) changes to:
The host is notified for the dynamic modules -> Assert failure (or segfault depending on build) as it can't find these modules

**Result of editing out the static imports in the test**
- The first module import is now the dynamic import
- ch throws an error "file not found" as it is attempting to evaluate them relative to the CWD - whereas they're given as relative to the script file

**Conclusion**
- currently Ch appears to correctly use paths relative to modules for static module imports 
- BUT ch uses paths relative to CWD for dynamic module imports from a script
- I suppressed this error by changing the script in the test to use paths relative to the CWD
-> The correct fix is probably to update ch's path handling

akroshg:
@boingoing will be able to review the change? 
@rhuanjl can you squash your commits to 1 and push it again?

rhuanjl:
All commits squashed, also created new issue 3964 for the path handling error in ch that was causing the CI test fail which required me to update bug_issue_3257.js.

I believe that is an unrelated error in ch which should be looked at separately but it's useful background for understanding the test change I've included in this PR.

rhuanjl:
@akroshg @boingoing is there anything else I need to do to help get this merged? (Other than perhaps rebasing to be against 1.7 not master as it’s a bug fix)

rhuanjl:
@boingoing ok, rebased to latest master- in an ideal world this would go to 1.7 I think but I didn't think of that before making the PR and it looks like it would be pretty painful to try and swap it now.

rhuanjl:
Sorry I'm unsure of the etiquette/practise here, I rebased the PR to master following boinging's review above but a different PR has been merged ahead of this one should I rebase again or leave it?

boingoing:
Looks like you should rebase on top of master one more time. 👍

rhuanjl:
Is this waiting on any other signoff/is there anything else I should do with it? @boingoing

boingoing:
@rhuanjl I will test this internally and try to merge it after that. 👍

PR #3935, opened 2017-10-11
ARM64: Implemented inline ceil/floor/round operations.

aaronsgiles:
Also added new opcodes needed, including move to/from FP status & control registers and conditional select.


sigatrev:

PR #3936, opened 2017-10-11
Add cache for @@toStringTag; allow DictionaryTypeHandlerBase into missing-property caches and invalidate when necessary.

ricobbe:
Should give us a ~3% improvement in EmberJS on Speedometer 2, as that test involves a lot of toString operations each of which looks up the @@toStringTag property and (as it happens) fails to find it.

### DIFF
--- a/lib/Runtime/Language/JavascriptOperators.cpp
+++ b/lib/Runtime/Language/JavascriptOperators.cpp
@@ -1808,12 +1808,10 @@ CommonNumber:
                 Output::Print(_u("MissingPropertyCaching: Missing property %d on slow path.\n"), propertyId);
             }
 
-            // Only cache missing property lookups for non-root field loads on objects that have PathTypeHandlers, because only these objects guarantee a type change when the property is added,
-            // which obviates the need to explicitly invalidate missing property inline caches.
-            //
-            // ricobbe: we now invalidate prototype caches on missing properties for DictionaryTypeHandlerBase as well, so allow those too.  Eventually we'll want to go back and
-            // add support for all kinds of types here.
-            if (!PHASE_OFF1(MissingPropertyCachePhase) && !isRoot && DynamicObject::Is(instance) && 
+            // Only cache missing property lookups for non-root field loads on objects that have PathTypeHandlers and DictionaryTypeHandlers, because only these types have the right behavior
+            // when the missing property is later added: path types guarantee a type change, and DictionaryTypeHandlerBase::AddProperty explicitly invalidates all prototype caches for the
+            // property.  (We don't support other types only because we haven't needed to yet; we do not anticipate any difficulties in adding the cache-invalidation logic there if that changes.)
+            if (!PHASE_OFF1(MissingPropertyCachePhase) && !isRoot && DynamicObject::Is(instance) &&
                 (DynamicObject::FromVar(instance)->GetDynamicType()->GetTypeHandler()->IsPathTypeHandler() || DynamicObject::FromVar(instance)->GetDynamicType()->GetTypeHandler()->IsDictionaryTypeHandler()))
             {
 #ifdef MISSING_PROPERTY_STATS
--- a/lib/Runtime/Language/JavascriptOperators.cpp
+++ b/lib/Runtime/Language/JavascriptOperators.cpp
@@ -1810,7 +1810,11 @@ CommonNumber:
 
             // Only cache missing property lookups for non-root field loads on objects that have PathTypeHandlers, because only these objects guarantee a type change when the property is added,
             // which obviates the need to explicitly invalidate missing property inline caches.
-            if (!PHASE_OFF1(MissingPropertyCachePhase) && !isRoot && DynamicObject::Is(instance) && ((DynamicObject*)instance)->GetDynamicType()->GetTypeHandler()->IsPathTypeHandler())
+            //
+            // ricobbe: we now invalidate prototype caches on missing properties for DictionaryTypeHandlerBase as well, so allow those too.  Eventually we'll want to go back and
+            // add support for all kinds of types here.
+            if (!PHASE_OFF1(MissingPropertyCachePhase) && !isRoot && DynamicObject::Is(instance) && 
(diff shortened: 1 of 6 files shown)


MikeHolman:
Can be a separate change, but seems like it would be a good idea to do the same thing with @@toPrimitive

lib/Runtime/Language/JavascriptOperators.cpp
```diff
@@ -1808,9 +1808,11 @@ namespace Js
                 Output::Print(_u("MissingPropertyCaching: Missing property %d on slow path.\n"), propertyId);
             }
 
-            // Only cache missing property lookups for non-root field loads on objects that have PathTypeHandlers, because only these objects guarantee a type change when the property is added,
-            // which obviates the need to explicitly invalidate missing property inline caches.
-            if (!PHASE_OFF1(MissingPropertyCachePhase) && !isRoot && DynamicObject::Is(instance) && ((DynamicObject*)instance)->GetDynamicType()->GetTypeHandler()->IsPathTypeHandler())
+            // Only cache missing property lookups for non-root field loads on objects that have PathTypeHandlers and DictionaryTypeHandlers, because only these types have the right behavior
```
obastemur: can't we invalidate this for all types by default?

pros -> use this approach for all + don't do type checking.
cons -> ??

ricobbe: We could, and I think the only costs would be the implementation cost of adding the prototype-cache invalidation logic to the other kinds of types, and any potential performance costs that arise because we invalidate caches more often than previously.  IB and benchmark results, however, indicate that this latter cost is likely to be negligible.

ricobbe:

lib/Runtime/Language/JavascriptOperators.cpp
obastemur: copy paste from your comments below;

```
However, if the new property is already in the cache as a missing property, we have to invalidate the prototype caches.
```
Just asking;
How this might affect the perf badly? It sounds like we better do it already for others too?

obastemur: If we are able to say that the only cost here is to implement this for others.. well, could be nice to at least have an open `performance` issue on this. Curious, how that could affect Acme.

ricobbe: It might affect perf because we're invalidating the caches more frequently than previously; this could in principle cause operations that currently hit the cache to start missing.  As above, though, we don't expect this cost to be significant.

I assume that we're already invalidating prototype caches correctly for other type handlers (that is, under the same conditions in which we invalidate the prototype caches for dictionary type handlers in the code currently on master), but I haven't verified this explicitly.  This should be sufficient as long as we don't allow the other type handlers into missing-property caches.

akroshg:
@aneeshdk how do we make use this to help @@toprimitive @@isconcatspreadable and @@hasintance  ?

lib/Runtime/Library/JavascriptLibrary.h
```diff
@@ -471,6 +472,8 @@ namespace Js
 
         mutable Field(CharStringCache) charStringCache;
 
+        Field(ScriptContextPolymorphicInlineCache*) toStringTagCache;
```
Penguinwizzard: Can you add setting this to nullptr in the initializer list for the constructor?

lib/Runtime/Library/JavascriptObject.cpp
```diff
@@ -357,6 +357,32 @@ namespace Js
...
+    {
+        const PropertyId toStringTagId(PropertyIds::_symbolToStringTag);
+        PolymorphicInlineCache *cache = scriptContext->GetLibrary()->GetToStringTagCache();
+        PropertyValueInfo info;
+        // We don't allow cache resizing, at least for the moment: it's more work, and since there's only one
+        // cache per script context, we can afford to create each cache with the maximum size.
+        PropertyValueInfo::SetCacheInfo(&info, nullptr, cache, false);
+        Var value;
+        if (CacheOperators::TryGetProperty<true, true, true, true, true, true, !PolymorphicInlineCache::IsPolymorphic, PolymorphicInlineCache::IsPolymorphic, false>(
+                thisArg, false, thisArg, toStringTagId, &value, scriptContext, nullptr, &info))
```
Penguinwizzard: Filed #3939 about the readability of this template; should probably get inspection from someone more familiar with the caching.

ricobbe: Added initializer to JavascriptLibrary ctor; will update PR momentarily.

Yes, I'd love to get feedback on the template parameters.  I'm fairly confident that most of them are correct, but I'm not sure about CheckTypePropertyCache, though I think this should only affect perf, not functional correctness.  @MikeHolman?

lib/Runtime/Language/JavascriptOperators.cpp
MikeHolman: Have you tried SimpleDictionaryTypeHandlerBase as well? Those are generally used more often than DictionaryTypeHandlerBase, so I think you stand to get an even bigger perf benefit from including them 

---
In reply to: [144170887](https://github.com/Microsoft/ChakraCore/pull/3936#discussion_r144170887) [](ancestors = 144170887)

lib/Runtime/Library/JavascriptLibrary.h
MikeHolman: nit: there is a `Cache` struct up above, might be better to include this in there

---
In reply to: [144174060](https://github.com/Microsoft/ChakraCore/pull/3936#discussion_r144174060) [](ancestors = 144174060)

lib/Runtime/Library/JavascriptObject.cpp
MikeHolman: I think these arguments are fine. For clarity though, you might want to annotate them (e.g. something like I did for [PropertyString cache](https://github.com/Microsoft/ChakraCore/blob/master/lib/Runtime/Library/PropertyString.h#L108))

---
In reply to: [144175487](https://github.com/Microsoft/ChakraCore/pull/3936#discussion_r144175487) [](ancestors = 144175487)

lib/Runtime/Library/JavascriptLibrary.h
ricobbe: Done; will update PR shortly.

PR #3937, opened 2017-10-12
Fix #3816 Class methods cannot be called static

### LINKED ISSUE #3816: Class methods cannot be called static
Currently it seems to not be possible to name a regular (non static) class method `static`. Edge errors out with `Expected identifier, string or number` so I assume that it gets interpreted as the static keyword instead of the method name.  
I've encountered this problem in Edge 15 and the Edge 16 preview.  

Here's a small reproduction: 

```js
class Foo {
  static() {
    console.log('this does not work in Edge');
  }
}

const foo = new Foo();
foo.static();
```

I am not 100% sure if the name is allowed, but it seems to work in other JavaScript engines

tcare:
'static' should be supported as a class method name, even in strict mode. Currently we eat the static token and just mark a flag.

Added detection similar to async and added a test case. Also included 'let' in the test case, as it is the only other semantically determined keyword.

### DIFF
--- a/lib/Parser/Parse.cpp
+++ b/lib/Parser/Parse.cpp
@@ -7605,10 +7605,26 @@ ParseNodePtr Parser::ParseClassDecl(BOOL isDeclaration, LPCOLESTR pNameHint, uin
             break;
         }
 
-        bool isStatic = m_token.tk == tkSTATIC;
-        if (isStatic)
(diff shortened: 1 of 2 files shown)


akroshg:
LGTM

tcare:

PR #3940, opened 2017-10-12
Correctly order special symbol initialization for default arguments

boingoing:
We incorrectly emit the special symbol initialization at the top of the
body scope when the function has default arguments but merged body and
param scopes. To be correct, we need to always emit the special symbol
initialization in the presence of default arguments - and always emit
those initializations before we emit the default arguments themselves.

Change here handles the case by walking the body scope var list manually
(because symbols in the body scope sym list are in the opposite order)
and emits each special symbol. We stop walking this list when we come to
the first non-special symbol as we know the parser has put all of the
special symbols at the front of the var list.

Fixes:
https://microsoft.visualstudio.com/OS/_workitems?id=13977464&_a=edit

### DIFF
--- a/lib/Runtime/ByteCode/ByteCodeEmitter.cpp
+++ b/lib/Runtime/ByteCode/ByteCodeEmitter.cpp
@@ -2975,9 +2975,12 @@ void ByteCodeGenerator::EmitOneFunction(ParseNode *pnode)
             {
                 for (ParseNodePtr pnodeVar = pnode->sxFnc.pnodeVars; pnodeVar; pnodeVar = pnodeVar->sxVar.pnodeNext)
                 {
+#if DBG
+                    bool reachedEndOfSpecialSymbols = false;
+#endif
                     Symbol* sym = pnodeVar->sxVar.sym;
 
-                    if (sym != nullptr && !(pnodeVar->sxVar.isBlockScopeFncDeclVar && sym->GetIsBlockVar()) && sym->IsSpecialSymbol())
+                    if (sym != nullptr && sym->IsSpecialSymbol())
                     {
                         EmitPropStoreForSpecialSymbol(sym->GetLocation(), sym, sym->GetPid(), funcInfo, true);
                         if (ShouldTrackDebuggerMetadata() && !sym->IsInSlot(funcInfo))
@@ -2987,10 +2990,21 @@ void ByteCodeGenerator::EmitOneFunction(ParseNode *pnode)
                     }
                     else
                     {
-                        // All of the special symbols exist at the beginning of the var list (parser guarantees this) so we can quit walking 
-                        // at the first non-special one we see.
+#if DBG
+                        reachedEndOfSpecialSymbols = true;
+#else
+                        // All of the special symbols exist at the beginning of the var list (parser guarantees this and debug build asserts this)
+                        // so we can quit walking at the first non-special one we see.
                         break;
+#endif
                     }
+
+#if DBG
+                    if (reachedEndOfSpecialSymbols)
+                    {
+                        Assert(sym == nullptr || !sym->IsSpecialSymbol());
+                    }
+#endif
                 }
             }
             else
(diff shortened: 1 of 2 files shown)


lib/Runtime/ByteCode/ByteCodeEmitter.cpp
```diff
@@ -2964,11 +2965,35 @@ void ByteCodeGenerator::EmitOneFunction(ParseNode *pnode)
...
+                    {
+                        EmitPropStoreForSpecialSymbol(sym->GetLocation(), sym, sym->GetPid(), funcInfo, true);
+                        if (ShouldTrackDebuggerMetadata() && !sym->IsInSlot(funcInfo))
+                        {
+                            byteCodeFunction->InsertSymbolToRegSlotList(sym->GetName(), sym->GetLocation(), funcInfo->varRegsCount);
+                        }
+                    }
+                    else
+                    {
+                        // All of the special symbols exist at the beginning of the var list (parser guarantees this) so we can quit walking
```
pleath: Can we have a debug-only check here that walks the remainder of the var list and asserts that no special symbols remain?

lib/Runtime/ByteCode/ByteCodeEmitter.cpp
```diff
@@ -2964,11 +2965,35 @@ void ByteCodeGenerator::EmitOneFunction(ParseNode *pnode)
...
+            // If the param and body scope are merged, the special symbol vars are located in the body scope so we
+            // need to walk over the var list.
+            if (funcInfo->IsBodyAndParamScopeMerged())
+            {
+                for (ParseNodePtr pnodeVar = pnode->sxFnc.pnodeVars; pnodeVar; pnodeVar = pnodeVar->sxVar.pnodeNext)
+                {
+                    Symbol* sym = pnodeVar->sxVar.sym;
 
-            if (paramScope)
+                    if (sym != nullptr && !(pnodeVar->sxVar.isBlockScopeFncDeclVar && sym->GetIsBlockVar()) && sym->IsSpecialSymbol())
```
pleath: What's the case where one of the special symbols is also the var binding of a block-scoped function?

lib/Runtime/ByteCode/ByteCodeEmitter.cpp
```diff
@@ -2977,6 +3002,11 @@ void ByteCodeGenerator::EmitOneFunction(ParseNode *pnode)
                     }
                 });
             }
+        }
+
+        if (pnode->sxFnc.HasNonSimpleParameterList() || !funcInfo->IsBodyAndParamScopeMerged())
```
pleath: Do we need to check IsBodyAndParamScopeMerged here? Can that be false if HasNonSimpleParameterList is also false? Is this just for ForceSplitScope?

lib/Runtime/ByteCode/ByteCodeEmitter.cpp
boingoing: Yes, that's a good idea.

lib/Runtime/ByteCode/ByteCodeEmitter.cpp
boingoing: I don't think we can have split scope without a non-simple parameter list unless ForceSplitScope is turned on. I moved our code to emit the special symbol storage above, though, because we don't really care if the scopes are merged. We just need to know if the parameter list is non-simple.

lib/Runtime/ByteCode/ByteCodeEmitter.cpp
boingoing: I don't think there is one. This code was lifted from DefineUserVars, which does this check. I'll try to remove it.

lib/Runtime/ByteCode/ByteCodeEmitter.cpp
```diff
@@ -2964,11 +2965,49 @@ void ByteCodeGenerator::EmitOneFunction(ParseNode *pnode)
...
+            {
+                for (ParseNodePtr pnodeVar = pnode->sxFnc.pnodeVars; pnodeVar; pnodeVar = pnodeVar->sxVar.pnodeNext)
+                {
+#if DBG
+                    bool reachedEndOfSpecialSymbols = false;
+#endif
+                    Symbol* sym = pnodeVar->sxVar.sym;
 
-            if (paramScope)
+                    if (sym != nullptr && sym->IsSpecialSymbol())
```
aneeshdk: >sym != nullptr [](start = 24, length = 14)

With this change can it ever be null?

boingoing: Yes, I think so. The sym could still be a non-special symbol of any flavor.

boingoing:
Thanks, all!

PR #3941, opened 2017-10-12
Simplify FunctionExecutionStateMachine::TryTransitionToNextExecutionMode

thomasmo:
Simplify FunctionExecutionStateMachine::TryTransitionToNextExecutionMode
This change makes significant changes to FunctionExecutionStateMachine::TryTransitionToNextExecutionMode. The previous implementation of this function was based around the ExecutionMode enum and would jump between states. With the new implementation, a separate enum, ExecutionState, is introduced to have a linear of moving across states. Common code paths across all states are also better shared. With this change, the invariants are consistent across states, and the exceptions stand out more. Further, the API boundary for FunctionExecutionStateMachine and some of FunctionBody is tightened to keep some of the internals of FunctionExecutionStateMachine abstracted. Some of the native unittests in FunctionExecutionTest are refactored to support behavior when FullJit is enabled vs not enabled for better code coverage.

### DIFF
--- a/lib/Runtime/Base/FunctionBody.cpp
+++ b/lib/Runtime/Base/FunctionBody.cpp
@@ -6619,9 +6619,14 @@ namespace Js
         return executionState.IncreaseInterpretedCount();
     }
 
-    ExecutionMode FunctionBody::GetDefaultInterpreterExecutionMode() const
+    void FunctionBody::SetAsmJsExecutionMode()
     {
-        return executionState.GetDefaultInterpreterExecutionMode();
+        executionState.SetAsmJsExecutionMode();
+    }
+
+    void FunctionBody::SetDefaultInterpreterExecutionMode()
+    {
+        executionState.SetDefaultInterpreterExecutionMode();
     }
 
     ExecutionMode FunctionBody::GetExecutionMode() const
@@ -6632,11 +6637,6 @@ namespace Js
     ExecutionMode FunctionBody::GetInterpreterExecutionMode(const bool isPostBailout)
     {
         return executionState.GetInterpreterExecutionMode(isPostBailout);
-    }
-
-    void FunctionBody::SetExecutionMode(const ExecutionMode executionMode)
-    {
-        executionState.SetExecutionMode(executionMode);
     }
 
     bool FunctionBody::IsInterpreterExecutionMode() const
@@ -6651,10 +6651,7 @@ namespace Js
 
     void FunctionBody::TryTransitionToNextInterpreterExecutionMode()
     {
-        Assert(IsInterpreterExecutionMode());
-
-        TryTransitionToNextExecutionMode();
-        SetExecutionMode(GetInterpreterExecutionMode(false));
+        executionState.TryTransitionToNextInterpreterExecutionMode();
     }
 
     void FunctionBody::SetIsSpeculativeJitCandidate()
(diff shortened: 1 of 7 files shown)


lib/Runtime/Base/FunctionExecutionStateMachine.cpp
```diff
@@ -376,147 +482,91 @@ namespace Js
...
+                    if (newState != ExecutionState::FullJit || !PHASE_OFF(FullJitPhase, owner))
+                    {
+                        isTransitionNeeded = true;
+                    }
+                    else
+                    {
+                        // If FullJit is the next state, but FullJit is disabled, then no transition
+                        // is needed.
+                        isTransitionNeeded = false;
+                    }
```
rajatd: To me, this would be more readable if the conditional logic was flipped and isTransitionNeeded not changed if the (flipped) condition failed.

lib/Runtime/Base/FunctionBody.cpp
```diff
@@ -9256,7 +9253,6 @@ namespace Js
                 {
                     newEntryPoint = simpleJitEntryPointInfo;
                     functionBody->SetDefaultFunctionEntryPointInfo(simpleJitEntryPointInfo, newEntryPoint->GetNativeEntrypoint());
-                    functionBody->SetExecutionMode(ExecutionMode::SimpleJit);
```
rajatd: Are we missing an udpate to the execution state here?

thomasmo: nope--it was moved into FunctionExecutionStateMachine::ResetSimpleJitLimit (the next function call)

lib/Runtime/Base/FunctionExecutionStateMachine.cpp
thomasmo: actually--I can do better and simply set the bool to the conditional statement. not sure why I didn't think of that before, but i'll do it now :-)

thomasmo: nope, I change my mind--I like your idea better. will be in the next update

lib/Runtime/Base/FunctionExecutionStateMachine.h
```diff
@@ -15,50 +15,73 @@ namespace Js
...
+        
+        // Debug functions
         void PrintLimits() const;
         void AssertIsInitialized() const;
 
-
     private:
+        // This enum creates a linear view of the progression of ExecutionModes, as
+        // described in the banner comment for TryTransitionToNextExecutionMode
+        enum ExecutionState
```
LouisLaf: Does it make a difference for it to be int8 like executionMode was?  Consider using an enum class which avoids enum collisions.

lib/Runtime/Base/FunctionExecutionStateMachine.h
```diff
@@ -15,50 +15,73 @@ namespace Js
...
+        {
+            Interpreter,
```
LouisLaf: ExecutionModes.h had good comments about the different modes.  It would be nice to have the same.
Do we still need ExecutionMode?  Should it be removed?

rajatd: I was also thinking about do we still need ExecutionMode, but was on the fence with it.

---
In reply to: [145004520](https://github.com/Microsoft/ChakraCore/pull/3941#discussion_r145004520) [](ancestors = 145004520)

lib/Runtime/Base/FunctionExecutionStateMachine.h
thomasmo: updated in latest commit

lib/Runtime/Base/FunctionExecutionStateMachine.h
thomasmo: So, I didn't add as much about the modes here because I didn't remove ExecutionModes. In the latest commit, I did add a comment saying to refer to Executionmodes.h for more details.

I wanted ExecutionState to remain an internal detail to the state machine and not be something that's exposed outside of it. If I remove ExecutionMode, I'll need to also update places throughout the code that uses this enum and also come up with a way to represent how some modes are repeated in the state diagram. I'd rather not change that, but won't push back if you say I should.

PR #3943, opened 2017-10-12
ARM64: Prolog/Epilog fixes

agarwal-sandeep:
Fix generating stack allocation – GenerateStackAllocation/_chkstk
Fix generating stack probe - GenerateStackProbe
Take care of 16-byte stack alignment
Use STP/LDP for all register store/load

### DIFF
--- a/lib/Backend/arm64/LowerMD.cpp
+++ b/lib/Backend/arm64/LowerMD.cpp
@@ -768,8 +768,8 @@ LowererMD::GenerateStackProbe(IR::Instr *insertInstr, bool afterProlog)
     // If the check fails call ThreadContext::ProbeCurrentStack which will check again and must throw.
     //
     //       LDIMM r12, ThreadContext::scriptStackLimit + frameSize //Load to register first, as this can be more than 12 bit supported in CMP
-    //       CMP  sp, r12
-    //       BGT  done
+    //       CMP   sp, r12
+    //       BHI   done
     // begin:
     //       LDIMM  r0, frameSize
     //       LDIMM  r1, scriptContext
@@ -781,10 +781,11 @@ LowererMD::GenerateStackProbe(IR::Instr *insertInstr, bool afterProlog)
     // For thread context with script interrupt enabled:
     //       LDIMM r12, &ThreadContext::scriptStackLimitForCurrentThread
     //       LDR   r12, [r12]
-    //       ADD   r12, frameSize
+    //       MOV   r15, frameSize
+    //       ADDS  r12, r12, r15
     //       BVS   $helper
-    //       CMP  sp, r12
-    //       BGT  done
+    //       CMP   sp, r12
+    //       BHI   done
     // $helper:
     //       LDIMM  r0, frameSize
     //       LDIMM  r1, scriptContext
@@ -803,65 +804,22 @@ LowererMD::GenerateStackProbe(IR::Instr *insertInstr, bool afterProlog)
 
     if (doInterruptProbe || !m_func->GetThreadContextInfo()->IsThreadBound())
     {
-        // Load the current stack limit and add the current frame allocation.
-        {
-            intptr_t pLimit = m_func->GetThreadContextInfo()->GetThreadStackLimitAddr();
-            this->CreateAssign(scratchOpnd, IR::AddrOpnd::New(pLimit, IR::AddrOpndKindDynamicMisc, this->m_func), insertInstr);
-            this->CreateAssign(scratchOpnd, IR::IndirOpnd::New(scratchOpnd, 0, TyMachReg, this->m_func), insertInstr);
-        }
+        // LDIMM r12, &ThreadContext::scriptStackLimitForCurrentThread
+        intptr_t pLimit = m_func->GetThreadContextInfo()->GetThreadStackLimitAddr();
+        this->CreateAssign(scratchOpnd, IR::AddrOpnd::New(pLimit, IR::AddrOpndKindDynamicMisc, this->m_func), insertInstr);
 
-        if (IS_CONST_00000FFF(frameSize))
-        {
-            // If the frame size is small enough, just add the constant.
-            // Does this ever happen with the size of the MinStackJIT constant?
-            instr = IR::Instr::New(Js::OpCode::ADD, scratchOpnd, scratchOpnd,
-                                   IR::IntConstOpnd::New(frameSize, TyMachReg, this->m_func), this->m_func);
-            insertInstr->InsertBefore(instr);
-        }
-        else
-        {
-            // We need a second scratch reg.
-            // If we're probing after the prolog, the reg has already been saved and will be restored.
-            // If not, push and pop it here, knowing that we'll never throw while the stack is whacked.
-            Assert(!afterProlog || this->m_func->m_unwindInfo.GetSavedScratchReg());
+        // LDR   r12, [r12, #0]
+        this->CreateAssign(scratchOpnd, IR::IndirOpnd::New(scratchOpnd, 0, TyMachReg, this->m_func), insertInstr);
 
-            BVUnit scratchBit;
-            IR::Opnd *opnd;
-            if (!afterProlog)
-            {
-                opnd = IR::IndirOpnd::New(IR::RegOpnd::New(nullptr, RegSP, TyMachReg, this->m_func), (int32)0, TyMachReg, this->m_func);
-                instr = IR::Instr::New(Js::OpCode::STR, opnd, this->m_func);
-                scratchBit.Set(RegEncode[SP_ALLOC_SCRATCH_REG]);
-                opnd = IR::RegBVOpnd::New(scratchBit, TyMachReg, this->m_func);
-                instr->SetSrc1(opnd);
-                insertInstr->InsertBefore(instr);
+        AssertMsg(!IS_CONST_00000FFF(frameSize), "For small size we can just add frameSize to r12");
 
-                IR::RegOpnd* regSpOpnd = IR::RegOpnd::New(nullptr, RegSP, TyMachReg, this->m_func);
-                IR::Instr * instrSub = IR::Instr::New(Js::OpCode::SUB, regSpOpnd, regSpOpnd, IR::IntConstOpnd::New(MachRegInt, TyMachReg, this->m_func), this->m_func);
-                insertInstr->InsertBefore(instrSub);
+        // MOV r15, frameSize
+        IR::Opnd* spAllocRegOpnd = IR::RegOpnd::New(nullptr, SP_ALLOC_SCRATCH_REG, TyMachReg, this->m_func);
+        this->CreateAssign(spAllocRegOpnd, IR::IntConstOpnd::New(frameSize, TyMachReg, this->m_func), insertInstr);
 
-            }
-
-            IR::Opnd *scratchOpnd2 = IR::RegOpnd::New(nullptr, SP_ALLOC_SCRATCH_REG, TyMachReg, this->m_func);
-            this->CreateAssign(scratchOpnd2, IR::IntConstOpnd::New(frameSize, TyMachReg, this->m_func), insertInstr);
-
-            instr = IR::Instr::New(Js::OpCode::ADDS, scratchOpnd, scratchOpnd, scratchOpnd2, this->m_func);
-            insertInstr->InsertBefore(instr);
-
-            if (!afterProlog)
-            {
-                Assert(scratchBit.Test(RegEncode[SP_ALLOC_SCRATCH_REG]));
-                opnd = IR::RegBVOpnd::New(scratchBit, TyMachReg, this->m_func);
-                instr = IR::Instr::New(Js::OpCode::LDR, opnd, this->m_func);
-                opnd = IR::IndirOpnd::New(IR::RegOpnd::New(nullptr, RegSP, TyMachReg, this->m_func), (int32)0, TyMachReg, this->m_func);
-                instr->SetSrc1(opnd);
-                insertInstr->InsertBefore(instr);
-
-                IR::RegOpnd* regSpOpnd = IR::RegOpnd::New(nullptr, RegSP, TyMachReg, this->m_func);
-                IR::Instr * instrSub = IR::Instr::New(Js::OpCode::ADD, regSpOpnd, regSpOpnd, IR::IntConstOpnd::New(MachRegInt, TyMachReg, this->m_func), this->m_func);
-                insertInstr->InsertBefore(instrSub);
-            }
-        }
+        // ADDS r12, r12, r15
+        instr = IR::Instr::New(Js::OpCode::ADDS, scratchOpnd, scratchOpnd, spAllocRegOpnd, this->m_func);
+        insertInstr->InsertBefore(instr);
 
         // If this add overflows, we have to call the helper.
         instr = IR::BranchInstr::New(Js::OpCode::BVS, helperLabel, this->m_func);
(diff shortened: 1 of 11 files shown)


lib/Backend/arm64/LowerMD.cpp
```diff
@@ -791,64 +791,47 @@ LowererMD::GenerateStackProbe(IR::Instr *insertInstr, bool afterProlog)
...
-            }
+        AssertMsg(!IS_CONST_00000FFF(frameSize), "For small size we can just add frameSize to r17");
 
-            IR::Opnd *scratchOpnd2 = IR::RegOpnd::New(nullptr, SP_ALLOC_SCRATCH_REG, TyMachReg, this->m_func);
-            this->CreateAssign(scratchOpnd2, IR::IntConstOpnd::New(frameSize, TyMachReg, this->m_func), insertInstr);
+        IR::Opnd *spAllocRegOpnd = nullptr;
+        IR::Opnd *unusedRegForAlignOpnd = nullptr;
+        if (!afterProlog)
+        {
+            // STP r15,r14,[sp, #-16]
```
aaronsgiles: This very unusual. First, x14, x15 are volatile registers, no need to save them. Second, if I'm reading this right, it looks like the saves are done in the function body. You are not allowed to make changes to SP in the middle of a function, unless you are saving SP in the prolog to a reserved nonvolatile register.

lib/Backend/arm64/LowerMD.cpp
```diff
@@ -791,64 +791,47 @@ LowererMD::GenerateStackProbe(IR::Instr *insertInstr, bool afterProlog)
...
 
-                IR::RegOpnd* regSpOpnd = IR::RegOpnd::New(nullptr, RegSP, TyMachReg, this->m_func);
-                IR::Instr * instrSub = IR::Instr::New(Js::OpCode::ADD, regSpOpnd, regSpOpnd, IR::IntConstOpnd::New(MachRegInt, TyMachReg, this->m_func), this->m_func);
-                insertInstr->InsertBefore(instrSub);
-            }
+        if (!afterProlog)
+        {
+            // LDP r15,r14,[sp, #16]
+            spAllocRegOpnd = IR::RegOpnd::New(nullptr, SP_ALLOC_SCRATCH_REG, TyMachReg, this->m_func);
+            unusedRegForAlignOpnd = IR::RegOpnd::New(nullptr, UNUSED_REG_FOR_STACK_ALIGN, TyMachReg, this->m_func);
```
aaronsgiles: Can't you re-use the operands from earlier?

lib/Backend/arm64/LowerMD.cpp
```diff
@@ -865,18 +849,21 @@ LowererMD::GenerateStackProbe(IR::Instr *insertInstr, bool afterProlog)
...
     {
+        // CMP sp, r17 
         instr = IR::Instr::New(Js::OpCode::CMP, this->m_func);
         instr->SetSrc1(IR::RegOpnd::New(nullptr, GetRegStackPointer(), TyMachReg, this->m_func));
         instr->SetSrc2(scratchOpnd);
         insertInstr->InsertBefore(instr);
         LegalizeMD::LegalizeInstr(instr, false);
 
+        // BGT  done
         instr = IR::BranchInstr::New(Js::OpCode::BGT, doneLabelInstr, this->m_func);
```
aaronsgiles: It doesn't matter in practice, but this should be an unsigned compare and thus the branch should be BHI instead of BGT.

lib/Backend/arm64/LowerMD.cpp
```diff
@@ -1356,32 +1365,47 @@ LowererMD::LowerEntryInstr(IR::EntryInstr * entryInstr)
...
     }
 
-    int32 regPushedSize = 0;
-
     if (!paramRegs.IsEmpty())
     {
-        // ToDo (SaAgarwa): Change STR to STP
+        IR::RegOpnd* src1 = nullptr;
+        IR::RegOpnd* src2 = nullptr;
         for (int i = LAST_INT_ARG_REG; i >= FIRST_INT_ARG_REG; --i)
```
aaronsgiles: Why the backwards order here? This is non-standard for ARM64.

lib/Backend/arm64/LowerMD.cpp
```diff
@@ -1392,47 +1416,75 @@ LowererMD::LowerEntryInstr(IR::EntryInstr * entryInstr)
...
-
-        regPushedSize = 0; // STP_PRE Advanced the SP
     }
 
     if (!usedRegs.IsEmpty())
     {
+        IR::RegOpnd* src1 = nullptr;
+        IR::RegOpnd* src2 = nullptr;
         for (RegNum reg = LAST_CALLEE_SAVED_GP_REG; reg > LAST_INT_ARG_REG; reg = (RegNum)(reg - 1))
         {
```
aaronsgiles: Again, weird backward order.

lib/Backend/arm64/LowerMD.cpp
```diff
@@ -1392,47 +1416,75 @@ LowererMD::LowerEntryInstr(IR::EntryInstr * entryInstr)
...
+                }
+            }
+
+            if (src1 != nullptr && src2 != nullptr)
+            {
+                IR::Instr * instrStp = IR::Instr::New(Js::OpCode::STP_PRE,
+                    IR::IndirOpnd::New(IR::RegOpnd::New(nullptr, RegSP, TyMachReg, this->m_func), -(2 * MachRegDouble), TyMachReg, this->m_func),
+                    src2, src1, this->m_func);
+                insertInstr->InsertBefore(instrStp);
+                src1 = src2 = nullptr;
```
aaronsgiles: Ugh, this isn't going to work with the way the unwind codes operate. You have to pre-allocate the space then use FSTP (no predecrement) with a positive offset to store them.

lib/Backend/arm64/LowerMD.cpp
```diff
@@ -1392,47 +1416,75 @@ LowererMD::LowerEntryInstr(IR::EntryInstr * entryInstr)
...
+                }
+            }
+
+            if (src1 != nullptr && src2 != nullptr)
+            {
+                IR::Instr * instrStp = IR::Instr::New(Js::OpCode::STP_PRE,
+                    IR::IndirOpnd::New(IR::RegOpnd::New(nullptr, RegSP, TyMachReg, this->m_func), -(2 * MachRegInt), TyMachReg, this->m_func),
+                    src2, src1, this->m_func);
+                insertInstr->InsertBefore(instrStp);
+                src1 = src2 = nullptr;
```
aaronsgiles: Sorry, this won't work with the unwind codes. You must pre-allocate all the space up front, then use STP (no _PRE) to store the values with a positive offset.

lib/Backend/arm64/LowerMD.cpp
```diff
@@ -1475,17 +1532,18 @@ LowererMD::LowerEntryInstr(IR::EntryInstr * entryInstr)
...
         // Now push the reg we used above to save the address of the top of the locals area.
-        IR::Instr * instrPush = IR::Instr::New(Js::OpCode::STR,
-            IR::IndirOpnd::New(IR::RegOpnd::New(nullptr, RegSP, TyMachReg, this->m_func), (int32)0, TyMachReg, this->m_func),
-            IR::RegOpnd::New(EH_STACK_SAVE_REG, TyMachReg, this->m_func),
-            this->m_func);
+        IR::RegOpnd* ehSaveOpnd = IR::RegOpnd::New(EH_STACK_SAVE_REG, TyMachReg, this->m_func);
+        IR::Instr * instrPush = IR::Instr::New(Js::OpCode::STP_PRE,
+            IR::IndirOpnd::New(IR::RegOpnd::New(nullptr, RegSP, TyMachReg, this->m_func), - 2 * MachRegInt, TyMachReg, this->m_func),
+            ehSaveOpnd, ehSaveOpnd, this->m_func);
         insertInstr->InsertBefore(instrPush);
```
aaronsgiles: Again, preallocate space then STP.

lib/Backend/arm64/LowerMD.cpp
```diff
@@ -1611,78 +1690,104 @@ LowererMD::LowerExitInstr(IR::ExitInstr * exitInstr)
...
+            if (reg1Opnd != nullptr && reg2Opnd != nullptr)
+            {
+                IR::Instr * instrLoadRegister = IR::Instr::New(Js::OpCode::LDP_POST,
+                    reg1Opnd,
+                    IR::IndirOpnd::New(IR::RegOpnd::New(nullptr, RegSP, TyMachReg, this->m_func), 2 * MachRegDouble, TyMachReg, this->m_func),
+                    reg2Opnd,
+                    this->m_func);
+                exitInstr->InsertBefore(instrLoadRegister);
+                reg1Opnd = reg2Opnd = nullptr;
             }
```
aaronsgiles: Mirror comments from above. You can't incrementally load and pop. You must do all the loads with LDP first then release the stack all at once at the end.

lib/Backend/arm64/LowerMD.cpp
agarwal-sandeep: SP is being changed only when afterProlog is false, so this is in prolog. 
You are right we don't need to save x14, x15 so I removed this code

---
In reply to: [144412145](https://github.com/Microsoft/ChakraCore/pull/3943#discussion_r144412145) [](ancestors = 144412145)

lib/Backend/arm64/LowerMD.cpp
agarwal-sandeep: Removed the code to save/restore x14,x15

---
In reply to: [144412422](https://github.com/Microsoft/ChakraCore/pull/3943#discussion_r144412422) [](ancestors = 144412422)

lib/Backend/arm64/LowerMD.cpp
agarwal-sandeep: Updated to BHI

---
In reply to: [144412622](https://github.com/Microsoft/ChakraCore/pull/3943#discussion_r144412622) [](ancestors = 144412622)

lib/Backend/arm64/LowerMD.cpp
```diff
@@ -936,32 +898,29 @@ LowererMD::GenerateStackAllocation(IR::Instr *instr, uint32 allocSize, uint32 pr
...
+    //chkstk expects the stacksize argument in R15 register
+    IR::RegOpnd *spAllocOpnd = IR::RegOpnd::New(nullptr, SP_ALLOC_SCRATCH_REG, TyMachReg, this->m_func);
     IR::RegOpnd *targetOpnd = IR::RegOpnd::New(nullptr, SCRATCH_REG, TyMachReg, this->m_func);
 
-    IR::IntConstOpnd *  stackSizeOpnd   = IR::IntConstOpnd::New((allocSize/MachPtr), TyMachReg, this->m_func, true);
+    IR::IntConstOpnd *  stackSizeOpnd   = IR::IntConstOpnd::New((allocSize / MachStackAlignment), TyMachReg, this->m_func, true);
 
-    IR::Instr *movInstr = IR::Instr::New(Js::OpCode::LDIMM, r4Opnd, stackSizeOpnd, this->m_func);
+    IR::Instr *movInstr = IR::Instr::New(Js::OpCode::LDIMM, spAllocOpnd, stackSizeOpnd, this->m_func);
     instr->InsertBefore(movInstr);
```
aaronsgiles: Is it ok to use LDIMM in the prolog, which might get expanded to multiple instructions? (Just checking, I don't know for sure if we're more restricted here.)

lib/Backend/arm64/LowerMD.cpp
```diff
@@ -936,32 +898,29 @@ LowererMD::GenerateStackAllocation(IR::Instr *instr, uint32 allocSize, uint32 pr
...
 
     IR::Instr *movHelperAddrInstr = IR::Instr::New(Js::OpCode::LDIMM, targetOpnd, IR::HelperCallOpnd::New(IR::HelperCRT_chkstk, this->m_func), this->m_func);
     instr->InsertBefore(movHelperAddrInstr);
 
-    IR::Instr * callInstr = IR::Instr::New(Js::OpCode::BLR, r4Opnd, targetOpnd, this->m_func);
+    IR::Instr * callInstr = IR::Instr::New(Js::OpCode::BLR, spAllocOpnd, targetOpnd, this->m_func);
     instr->InsertBefore(callInstr);
 
     // Generate SUB SP, SP, R4
```
aaronsgiles: I recommend updating the comment here to reflect the special SUB we're doing.

lib/Backend/arm64/LowerMD.cpp
```diff
@@ -1298,24 +1181,45 @@ LowererMD::LowerEntryInstr(IR::EntryInstr * entryInstr)
...
+            if (!usedRegs.Test(RegEncode[SP_ALLOC_SCRATCH_REG])) //If its a large stack and RegR15 is not already saved.
             {
                 // If it is not a small stack we have to call __chkstk.
-                // __chkstk has special calling convention and trashes R4
+                // __chkstk has special calling convention and is called with the amount to allocate divided by 16 in x15
                 // And if we're probing the stack dynamically, we need an extra reg to do the frame size calculation.
                 //
                 // Note that it's possible that we now no longer need a dynamic stack probe because
                 // m_localStackHeight may be encodable in Mod12. However, this is a chicken-and-egg
                 // problem, so we're going to stick with saving R4 even though it's possible it
```
aaronsgiles: Comment needs updating. Mod12 is not the encoding limitation here and R4 is not the reg in question.

lib/Backend/arm64/LowerMD.cpp
```diff
@@ -1368,32 +1294,19 @@ LowererMD::LowerEntryInstr(IR::EntryInstr * entryInstr)
         this->m_func->m_unwindInfo.SetPrologStartLabel(prologStartLabel->m_id);
     }
 
-    int32 regPushedSize = 0;
+    // STP fp, lr, [sp, -preIndexedOffset]!
+    // STP r8 to r28 (in pairs), [sp, +ve offset]
+    // STP d16 to d29 (in pairs), [sp, +ve offset]
```
aaronsgiles: Should only be saving r19 to r28. Please either update the comment or fix the logic.

lib/Backend/arm64/LowerMD.cpp
```diff
@@ -1405,46 +1318,114 @@ LowererMD::LowerEntryInstr(IR::EntryInstr * entryInstr)
...
+        preIndexedOffset = 2 * MachRegInt;
     }
 
     if (!usedRegs.IsEmpty())
     {
-        for (RegNum reg = LAST_CALLEE_SAVED_GP_REG; reg > LAST_INT_ARG_REG; reg = (RegNum)(reg - 1))
+        IR::RegOpnd* src1 = nullptr;
+        IR::RegOpnd* src2 = nullptr;
+        for (RegNum reg = RegR8; reg <= LAST_CALLEE_SAVED_GP_REG; reg = (RegNum)(reg + 1))
         {
```
aaronsgiles: Should be RegR19 here, not RegR8.

lib/Backend/arm64/LowerMD.cpp
agarwal-sandeep: Is there some other instruction which can achieve the same?

aaronsgiles: You can use MOVZ directly if you know the immediate is less than 64k. But again, if it's correct to use LDIMM and allow for expansion, feel free to keep it as-is. For some reason had it in my head that there were some limitations in this situation, but I can't point to how I suspect this. :)

agarwal-sandeep: Thanks! for now will keep it this way.

PR #3945, opened 2017-10-12
Remove 'this' binding for global and global lambda functions

boingoing:
Do not create a 'this' symbol for global functions and global lambda functions. Instead emit a LdThis into a special register stored in the FuncInfo during the prolog of these functions and load that register at every name node referencing 'this'. 

Add `PnFnc::IsDerivedClassConstructor` and refactor `Parser::CreateSpecialSymbolDeclarations`
    
`PnFnc::IsDerivedClassConstructor` is just a convenience method for `(PnFnc::IsClassConstructor() && !PnFnc::IsBaseClassConstructor())`.
    
`Parser::CreateSpecialSymbolDeclarations` had the code which does the check for special symbol references factored-out so it's a bit simpler.
    
Simplify the logic to construct the special symbol var declarations a little bit. Also fix emitting a special symbol as a call target. This fixes:
https://microsoft.visualstudio.com/OS/_workitems?id=13977724&_a=edit

### DIFF
--- a/lib/Parser/Parse.cpp
+++ b/lib/Parser/Parse.cpp
@@ -1688,6 +1688,8 @@ void Parser::CreateSpecialSymbolDeclarations(ParseNodePtr pnodeFnc, bool isGloba
         return;
     }
     Assert(!(isGlobal && (this->m_grfscr & fscrEval)));
+    Assert(!isGlobal || (this->m_grfscr & fscrEvalCode));
+
     bool isTopLevelEventHandler = (this->m_grfscr & fscrImplicitThis || this->m_grfscr & fscrImplicitParents);
     // Create a 'this' symbol for indirect eval, non-lambda functions with references to 'this', and all class constructors.
     PidRefStack* ref = wellKnownPropertyPids._this->GetTopRef();
@@ -1726,8 +1728,9 @@ void Parser::CreateSpecialSymbolDeclarations(ParseNodePtr pnodeFnc, bool isGloba
         superNode->sxPid.sym->SetIsSuper(true);
     }
 
+    // Create a 'super' (as the call target for super()) symbol only for class constructors.
     ref = wellKnownPropertyPids._superConstructor->GetTopRef();
-    if (ref && ref->GetScopeId() >= m_currentBlockInfo->pnodeBlock->sxBlock.blockId)
+    if ((ref && ref->GetScopeId() >= m_currentBlockInfo->pnodeBlock->sxBlock.blockId) && pnodeFnc->sxFnc.IsClassConstructor())
     {
         // Insert the decl for 'super (as the call target for super())'
         ParseNodePtr superNode = this->CreateSpecialVarDeclNode(pnodeFnc, wellKnownPropertyPids._superConstructor);
@@ -2252,6 +2255,14 @@ void Parser::EnsureStackAvailable()
 
 void Parser::ThrowNewTargetSyntaxErrForGlobalScope()
 {
+    // If we are parsing a previously deferred function, we can skip throwing the SyntaxError for `new.target` at global scope.
+    // If we are at global scope, we would have thrown a SyntaxError when we did the Upfront parse pass and we would not have
+    // deferred the function in order to come back now and reparse it.
+    if (m_parseType == ParseType_Deferred)
+    {
+        return;
+    }
+
     if (GetCurrentNonLambdaFunctionNode() != nullptr)
     {
         return;
@@ -11484,8 +11495,8 @@ ParseNodePtr Parser::Parse(LPCUTF8 pszSrc, size_t offset, size_t length, charcou
     if (tkEOF != m_token.tk)
         Error(ERRsyntax);
 
-    // Direct eval doesn't need to have special symbols created
-    if (!(this->m_grfscr & fscrEval))
+    // We only need to create special symbol bindings for 'this' for indirect eval
+    if ((this->m_grfscr & fscrEvalCode) && !(this->m_grfscr & fscrEval))
     {
         CreateSpecialSymbolDeclarations(pnodeProg, true);
     }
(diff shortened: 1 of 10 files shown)


lib/Parser/Parse.cpp
```diff
@@ -1680,56 +1680,73 @@ ParseNodePtr Parser::ReferenceSpecialName(IdentPtr pid, charcount_t ichMin, char
...
 {
+    // Lambda function cannot have any special bindings.
+    if (pnodeFnc->sxFnc.IsLambda())
+    {
+        return;
+    }
     Assert(!(isGlobal && (this->m_grfscr & fscrEval)));
+    Assert(!isGlobal || (this->m_grfscr & fscrEvalCode));
 
     bool isTopLevelEventHandler = (this->m_grfscr & fscrImplicitThis || this->m_grfscr & fscrImplicitParents);
```
pleath: We're still planning to make isTopLevelEventHandler non-conservative, right? (Doesn't have to be this change, I'm just checking.)

lib/Parser/Parse.cpp
```diff
@@ -2250,6 +2267,14 @@ void Parser::EnsureStackAvailable()
 
 void Parser::ThrowNewTargetSyntaxErrForGlobalScope()
 {
+    // If we are parsing a previously deferred function, we can skip throwing the SyntaxError for `new.target` at global scope.
+    // If we are at global scope, we would have thrown a SyntaxError when we did the Upfront parse pass and we would not have
+    // deferred the function in order to come back now and reparse it.
+    if (m_parseType == ParseType_Deferred)
```
pleath: Is this just a shortcut, or does it fix some issue?

lib/Parser/Parse.cpp
```diff
@@ -1680,56 +1680,73 @@ ParseNodePtr Parser::ReferenceSpecialName(IdentPtr pid, charcount_t ichMin, char
...
+
 void Parser::CreateSpecialSymbolDeclarations(ParseNodePtr pnodeFnc, bool isGlobal)
 {
+    // Lambda function cannot have any special bindings.
+    if (pnodeFnc->sxFnc.IsLambda())
+    {
+        return;
+    }
     Assert(!(isGlobal && (this->m_grfscr & fscrEval)));
+    Assert(!isGlobal || (this->m_grfscr & fscrEvalCode));
```
pleath: We still shouldn't get here if fscrEval is set, right? Can we still assert that?

lib/Parser/Parse.cpp
```diff
@@ -1680,56 +1680,73 @@ ParseNodePtr Parser::ReferenceSpecialName(IdentPtr pid, charcount_t ichMin, char
...
 
     bool isTopLevelEventHandler = (this->m_grfscr & fscrImplicitThis || this->m_grfscr & fscrImplicitParents);
```
pleath: (Nice simplification in here. Thanks.)

lib/Runtime/ByteCode/ByteCodeEmitter.cpp
```diff
@@ -2342,47 +2340,15 @@ void ByteCodeGenerator::EmitBaseClassConstructorThisObject(FuncInfo *funcInfo)
     this->Writer()->Reg2(Js::OpCode::NewScObjectNoCtorFull, funcInfo->GetThisSymbol()->GetLocation(), funcInfo->GetNewTargetSymbol()->GetLocation());
 }
 
-void ByteCodeGenerator::GetEnclosingNonLambdaScope(FuncInfo *funcInfo, Scope * &scope, Js::PropertyId &envIndex)
```
pleath: Hah!

lib/Parser/Parse.cpp
boingoing: Yes, I'm testing that and I'll send a separate PR.

lib/Runtime/ByteCode/ByteCodeEmitter.cpp
```diff
@@ -4872,14 +4838,34 @@ ByteCodeGenerator::GetLdSlotOp(Scope *scope, int envIndex, Js::RegSlot scopeLoca
...
 
-void ByteCodeGenerator::EmitPropLoadThis(Js::RegSlot lhsLocation, ParseNode *pnode, FuncInfo *funcInfo, bool chkUndecl)
+bool ByteCodeGenerator::ShouldLoadConstThis(FuncInfo* funcInfo)
 {
-    this->EmitPropLoad(lhsLocation, pnode->sxPid.sym, pnode->sxPid.pid, funcInfo, true);
+    // Load a const 'this' binding if the following holds
+    // - The function has a 'this' name node
+    // - We are in a global or global lambda function
+    // - The function has no 'this' symbol (an indirect eval would have this symbol)
+    return funcInfo->thisConstantRegister != Js::Constants::NoRegister
```
pleath: Instead of checking that we're in a global context (global func or global lambda), can we assert it? I.e., if "this" has no symbol, we should be in a global "this" context.

lib/Parser/Parse.cpp
boingoing: This is from the 'defer parse lambda' branch which is where this change to remove the global 'this' binding was pulled out of. The change here avoids the need to figure out if the lambda is actually global at defer parse time by skipping this unneeded check.

lib/Runtime/ByteCode/ByteCodeEmitter.cpp
boingoing: Isn't it possible for an ordinary function to not have a 'this' binding? I guess we could just check to see if `funcInfo->thisConstantRegister != Js::Constants::NoRegister` and then assert and skip this check function.

pleath: That sounds right.

lib/Parser/Parse.cpp
boingoing: We are asserting this, right?

lib/Parser/Parse.cpp
boingoing: 👍

lib/Parser/Parse.cpp
boingoing: Strike that. Went ahead and added it into this change as it's very small.

lib/Parser/Parse.cpp
pleath: OK. Thanks.

lib/Parser/Parse.cpp
pleath: Oh, of course. I misread the diff. This is fine.

PR #52, opened 2016-01-12
Add %AsyncFunction% and %AsyncFunctionPrototype% built-ins

ianwjhalliday:
Add the async function built-in objects, %AsyncFunction% constructor and %AsyncFunctionPrototype% object.

Hooked up async function instances to %AsyncFunctionPrototype%.

Just like generators, neither are these objects are accessible from the global object.  Only from async function instances via the prototype.

Added tests and fixed a couple bugs as well:
 - toString wasn't showing async keyword for async functions
 - async function objects incorrectly had caller and arguments properties

Issue #11, opened 2016-01-06
[Continuous Integration] Add check for Microsoft copyright headers in committed files


dilijev:
Include support for exceptions for files that should not contain the Microsoft copyright notice for various reasons (e.g. would be rendered in markdown, format of file doesn't support comments, included in this repo under a different license, etc.)

This should run as a separate Jenkins CI task from the already existing tasks to make it immediately clear what the issue is.

Also provide a script to easily fix up files that do not have the notice properly displayed.

PR #3946, opened 2017-10-12
ARM64: enable tagged floats

sigatrev:


lib/Backend/Lower.cpp
```diff
@@ -16482,6 +16484,20 @@ Lowerer::GenerateFastLdElemI(IR::Instr *& ldElem, bool *instrIsInHelperBlockRef)
...
+#elif _M_ARM64
+                {
+                    IR::Instr *const instr = IR::Instr::New(Js::OpCode::FCMP, this->m_func);
+                    instr->SetSrc1(reg);
+                    instr->SetSrc2(reg);
+                    ldElem->InsertBefore(instr);
+                }
+
+                // JP $helper
+                {
```
aaronsgiles: Comment should be fixes for consistency.

lib/Backend/arm64/LowerMD.cpp
```diff
@@ -6008,73 +5958,70 @@ LowererMD::GenerateClz(IR::Instr * instr)
...
-        StackSym * tempNumberSym = this->m_lowerer->GetTempNumberSym(instrOrig->GetDst(), instrOrig->dstIsTempNumberTransferred);
-
-        // LEA dst, &tempSym
-        IR::SymOpnd * symTempSrc = IR::SymOpnd::New(tempNumberSym, TyMachPtr, this->m_func);
-        newInstr = IR::Instr::New(Js::OpCode::LEA, dstOpnd, symTempSrc, this->m_func);
+
+    // s1 = MOV opndFloat
+    IR::RegOpnd *s1 = IR::RegOpnd::New(TyMachReg, m_func);
+    IR::Instr *mov = IR::Instr::New(Js::OpCode::FMOV, s1, opndFloat, m_func);
+    instrInsert->InsertBefore(mov);
```
aaronsgiles: You need to use FMOV_GEN to move from a floating point register to an integer register.

lib/Backend/arm64/LowerMD.cpp
```diff
@@ -6008,73 +5958,70 @@ LowererMD::GenerateClz(IR::Instr * instr)
...
+        Lowerer::InsertBranch(Js::OpCode::Br, done, helper);
+
+        IR::RegOpnd* tmp2 = IR::RegOpnd::New(TyMachReg, m_func);
 
-        // STR dst->vtable, JavascriptNumber::vtable
-        newInstr = IR::Instr::New(Js::OpCode::STR, symVTableDst, jsNumberVTable, this->m_func);
-        numberInitInsertInstr->InsertBefore(newInstr);
-        LegalizeMD::LegalizeInstr(newInstr, false);
+        newInstr = IR::Instr::New(Js::OpCode::AND, tmp2, s1, IR::AddrOpnd::New((Js::Var)0x000FFFFFFFFFFFFFull, IR::AddrOpndKindConstantVar, m_func, true), m_func);
+        done->InsertBefore(newInstr);
```
aaronsgiles: Better here to use UBFX dst, src, #0, #52 which is guaranteed to be a single instruction (parameters would by tmp2, s1, IR::AddrOpnd::New(BITFIELD(0, 52), ...)

lib/Backend/arm64/LowerMD.cpp
```diff
@@ -6008,73 +5958,70 @@ LowererMD::GenerateClz(IR::Instr * instr)
...
+        // tmp = NOT s1
+        // tmp = AND tmp, 0x7FF0000000000000ull
+        // cbz tmp, helper
+        // jmp done
+        // helper:
+        // tmp2 = AND s1, 0x000FFFFFFFFFFFFFull
+        // cbz tmp2, done
+        // s1 = JavascriptNumber::k_Nan
+        // done:
+
```
aaronsgiles: Here's an alternative:
s1 = FMOV_GEN src    ; assuming src is in an FP register
tmp = UBFX s1, #52, #11   ; extract exponent, bits 52-62
CMP tmp, #0x7ff    ; check for 0x7ff
BNE done    ; if not, done; otherwise fall through
tmp2 = UBFX s1, #0, #52   ; extract mantissa, bits 0-51
CBZ tmp2, done    ; if zero, done
s1 = JavascriptNumber::k_Nan    ; load NaN
done:

lib/Backend/arm64/LowerMD.cpp
```diff
@@ -8372,6 +8316,40 @@ void LowererMD::GenerateFloatTest(IR::RegOpnd * opndSrc, IR::Instr * insertInstr
     insertInstr->InsertBefore(instr);
 }
 
+IR::RegOpnd* LowererMD::CheckFloatAndUntag(IR::RegOpnd * opndSrc, IR::Instr * insertInstr, IR::LabelInstr* labelHelper)
+{
+    IR::Opnd* floatTag = IR::AddrOpnd::New((Js::Var)Js::FloatTag_Value, IR::AddrOpndKindConstantVar, this->m_func, /* dontEncode = */ true);
+    IR::RegOpnd* regOpndFloatTag = IR::RegOpnd::New(TyUint64, this->m_func);
```
aaronsgiles: Why is this an AddrOpnd and not an IntConstOpnd?

aaronsgiles: Note that FloatTag_Value is directly encodeable as an immediate to logical operations (AND/OR/EOR), so might be most optimal to leave it as an immediate in the operations below.

lib/Backend/arm64/LowerMD.cpp
```diff
@@ -8372,6 +8316,40 @@ void LowererMD::GenerateFloatTest(IR::RegOpnd * opndSrc, IR::Instr * insertInstr
...
+    }
+
+    // untaggedFloat = EOR floatTagReg, s1 // where untaggedFloat == floatTagReg; use floatTagReg temporarily for the untagged float
+    IR::RegOpnd* untaggedFloat = regOpndFloatTag;
+    instr = IR::Instr::New(Js::OpCode::EOR, untaggedFloat, regOpndFloatTag, opndSrc, this->m_func);
+    insertInstr->InsertBefore(instr);
+
+    IR::RegOpnd *floatReg = IR::RegOpnd::New(TyMachDouble, this->m_func);
+    instr = IR::Instr::New(Js::OpCode::FMOV, floatReg, untaggedFloat, this->m_func);
+    insertInstr->InsertBefore(instr);
```
aaronsgiles: Again need to use FMOV_GEN to move from/to an integer register.

lib/Backend/arm64/LowerMD.cpp
sigatrev: It believe the reason we use AddrOpnd is to help with IR dumping. An addr that matches that is uniquely FloatTag, where as a number that matches that might not be.

lib/Backend/arm64/LowerMD.cpp
sigatrev: During the Layout phase later on helper blocks are moved to the end of the function, so by treating the special case as a helper block the emitted code will take the shape(with the other corrections applied)

s1 = FMOV_GEN src
tmp = UBFX s1, #52, #11
cmp = tmp, 0x7FF
beq tmp, helper
done:
...
helper:
tmp2 = tmp2 = UBFX s1, #0, #52
cbz tmp2, done
s1 = JavascriptNumber::k_Nan

aaronsgiles: Ah, didn't know that.

lib/Backend/arm64/LowerMD.cpp
```diff
@@ -6008,73 +5958,71 @@ LowererMD::GenerateClz(IR::Instr * instr)
...
+        // cbz tmp2, done
+        // s1 = JavascriptNumber::k_Nan
+        // done:
 
-    if (numberInitInsertInstr)
-    {
-        IR::Opnd *jsNumberVTable = m_lowerer->LoadVTableValueOpnd(numberInitInsertInstr, VTableValue::VtableJavascriptNumber);
+        IR::RegOpnd* tmp = IR::RegOpnd::New(TyMachReg, m_func);
+
+        IR::Instr* newInstr = IR::Instr::New(Js::OpCode::UBFX, tmp, s1, IR::IntConstOpnd::New(((52<<16)|11), TyMachReg, m_func, true), m_func);
```
aaronsgiles: It just landed, but there is a BITFIELD() macro you can use to say BITFIELD(52, 11) instead of hand-assembling the constant here.

lib/Backend/arm64/LowerMD.cpp
```diff
@@ -8372,6 +8317,35 @@ void LowererMD::GenerateFloatTest(IR::RegOpnd * opndSrc, IR::Instr * insertInstr
...
+        insertInstr->InsertBefore(instr);
+        LegalizeMD::LegalizeInstr(instr, false);
+
+        // BZ $helper
+        instr = IR::BranchInstr::New(Js::OpCode::BEQ /* BZ */, labelHelper, this->m_func);
+        insertInstr->InsertBefore(instr);
+    }
+
+    IR::RegOpnd* untaggedFloat = IR::RegOpnd::New(TyMachReg, this->m_func);
+    IR::Instr* instr = IR::Instr::New(Js::OpCode::EOR, untaggedFloat, floatTag, opndSrc, this->m_func);
```
aaronsgiles: Think you need to swap floatTag and opndSrc here, unless the system is smart enough to move the immediate for a commutative operation.

lib/Backend/arm64/LowerMD.cpp
```diff
@@ -6008,73 +5958,71 @@ LowererMD::GenerateClz(IR::Instr * instr)
...
-        IR::Opnd *typeOpnd = m_lowerer->LoadLibraryValueOpnd(numberInitInsertInstr, LibraryValue::ValueNumberTypeStatic);
-        newInstr = IR::Instr::New(Js::OpCode::STR, symTypeDst, typeOpnd, this->m_func);
-        numberInitInsertInstr->InsertBefore(newInstr);
-        LegalizeMD::LegalizeInstr(newInstr, false);
+        Lowerer::InsertBranch(Js::OpCode::Br, done, helper);
+
+        IR::RegOpnd* tmp2 = IR::RegOpnd::New(TyMachReg, m_func);
+
+        newInstr = IR::Instr::New(Js::OpCode::UBFX, tmp2, s1, IR::IntConstOpnd::New(((0<<16)|52), TyMachReg, m_func, true), m_func);
+        done->InsertBefore(newInstr);
```
aaronsgiles: Can use BITFIELD(0, 52) here now.

PR #3949, opened 2017-10-13
Avoid varargs leak

Penguinwizzard:
va_start needs va_end pairing.


Penguinwizzard:
Going to re-submit as part of a set of other small fixes.

PR #3950, opened 2017-10-13
Fix NumberFormatOptions test in ICU, minor cleanup

### LINKED ISSUE #3948: Intl-ICU: useGrouping option for NumberFormat is not respected
```js
new Intl.NumberFormat('en-US', { useGrouping: false }).format(123456)
```

Expected return: `"123456"`
Actual return: `"123,456"`

jackhorton:
Fixes #3948


lib/Runtime/Library/IntlEngineInterfaceExtensionObject.cpp
```diff
@@ -164,7 +165,10 @@ namespace Js
             }
         }
     };
+#endif
 
+// Defining Finalizable wrappers for Intl data
+#ifdef INTL_WINGLOB
```
dilijev: The above endif is for INTL_WINGLOB also. Why have two separate blocks?

lib/Runtime/Library/IntlEngineInterfaceExtensionObject.cpp
```diff
@@ -203,7 +207,57 @@ namespace Js
             return instance;
         }
     };
-#endif
+
+#elif INTL_ICU
```
dilijev: #elif doesn't semantically match #ifdef

Could change both to use `#if defined(...)` and `#elif defined(...)`

I've noticed you tending to depend on the fact that `INTL_ICU` is set to something non-zero as opposed to being defined. My setting it to `1` was defensive against this kind of conflation of defined-ness versus truthy-ness, but `#ifdef / #if defined(...)` is more defensive in general. We know that if this macro is defined it tells us we're using the feature. It's actual value (or lack thereof) should not be important (or could theoretically become important in the future, but unlikely). I'm more concerned that someone will see the `1` and think this is strange style since that `1` has no semantic meaning as a constant and could end up breaking an uncommon config (`INTL_ICU`) -- another reason to get some CI coverage but that's another story.

lib/Runtime/Library/IntlEngineInterfaceExtensionObject.cpp
```diff
@@ -203,7 +207,57 @@ namespace Js
...
+        {
+        }
+
+        T * GetInstance()
+        {
+            return instance;
+        }
+    };
+
+#endif // Defining Finalizable wrappers for Intl data
```
dilijev: I think this makes some sense as-is since the concept of this block is wrapper objects, but this is a strange way to comment an `#endif`.

Consider the perspective of someone unfamiliar with the code and reading through preprocessor config, `#endif // INTL_ICU` is more helpful to see which `#ifdef` or `#elif` it matches when you're reading up to find the beginning of the block, since preprocessor directives are not indented and "navigate to beginning of block" is still a little-supported feature of IDEs, especially for preprocessor blocks.

Documentation of this block is already given in the comment at the beginning, so that is sufficient.

lib/Runtime/Library/IntlEngineInterfaceExtensionObject.cpp
```diff
@@ -866,23 +920,19 @@ namespace Js
...
-
         if (GetTypedPropertyBuiltInFrom(options, __isDecimalPointAlwaysDisplayed, JavascriptBoolean))
         {
             isDecimalPointAlwaysDisplayed = JavascriptBoolean::FromVar(propertyValue)->GetValue();
         }
 
         if (GetTypedPropertyBuiltInFrom(options, __useGrouping, JavascriptBoolean))
         {
-            useGrouping = JavascriptBoolean::FromVar(propertyValue)->GetValue();
+            bool useGrouping = JavascriptBoolean::FromVar(propertyValue)->GetValue();
```
dilijev: Oh hm... useGrouping was not read at all previously? I'm surprised we didn't get a compiler warning for unused var. Maybe because it was declared and then written to?

test/Intl/IntlReturnedValueTests.js
```diff
@@ -6,59 +6,36 @@
...
 
 /////////////////// DateFormat ////////////////////
+
 var options = { ca: "gregory", hour12: true, timeZone:"UTC" };
 var dateFormat1 = new Intl.DateTimeFormat("en-US", options);    /**bp:resume('step_over');locals();**/
 WScript.Echo("");  // Dummy line to get desired debugger logging behavior. Required due to the above bug.
 var date1 = new Date(2000, 1, 1);
 
-// Bug: GitHub Issue#3439: Intl function call has 2 entries in the debugger functionCallsReturn.
-// https://github.com/Microsoft/ChakraCore/issues/3439
```
dilijev: Thanks for cleaning up!

test/Intl/CollatorUnsupportedLocales.js
```diff
@@ -3,35 +3,28 @@
 // Licensed under the MIT license. See LICENSE.txt file in the project root for full license information.
 //-------------------------------------------------------------------------------------------------------
 
-var passed = true;
+WScript.LoadScriptFile("..\\UnitTestFramework\\UnitTestFramework.js");
```
dilijev: Appreciate the change to use the UnitTestFramework -- but just checking, was this test broken before?

test/Intl/IntlReturnedValueTests.js
dilijev: Was this just cleanup or was something broken?

test/Intl/NumberFormatOptions.js
```diff
@@ -82,18 +82,20 @@ var tests = [
...
+            const n = 123.1;
 
-            var actual1 = new Intl.NumberFormat("en-US", extendOptions(usdBaseOptions, {})).format(123.1);
-            assert.isTrue(expectedRegex1.test(actual1), "currency: USD (default display) actual: '" + actual1 + "' expected(Regex): '" + expectedRegex1 + "'");
-            var actual2 = new Intl.NumberFormat("en-US", extendOptions(usdBaseOptions, { currencyDisplay: "code" })).format(123.1);
-            assert.isTrue(expectedRegex2.test(actual2), "currency code: USD actual: '" + actual2 + "' expected(Regex): '" + expectedRegex2 + "'");
+            const defaultActual = new Intl.NumberFormat("en-US", extendOptions(usdBaseOptions, {})).format(n);
+            assert.isTrue(symbolFormat.test(defaultActual), `USD (default) actual: '${defaultActual}', which did not match ${symbolFormat.source}`);
+            const codeActual = new Intl.NumberFormat("en-US", extendOptions(usdBaseOptions, { currencyDisplay: "code" })).format(n);
+            assert.isTrue(codeFormat.test(codeActual), `USD (code) actual: '${codeActual}', which did not match ${codeFormat.source}`);
```
dilijev: Maybe it's too niche (IMO not really) but -- if we end up reusing this text-match-regex pattern (which I think we will) maybe UnitTestFramework could use an `assert.matches`. Just something to keep in mind for your next change like this.

test/Intl/NumberFormatOptions.js
```diff
@@ -82,18 +82,20 @@ var tests = [
         name: "Test Valid Options - Formatting Currency with Significant Digits",
         body: function () {
             const usdBaseOptions = { minimumSignificantDigits: 3, maximumSignificantDigits: 3, style: "currency", currency: "USD" };
-            var expectedRegex1 = new RegExp("\\$[\x20\u00a0]?123");
-            var expectedRegex2 = new RegExp("USD[\x20\u00a0]?123");
+            const symbolFormat = /\$[\x20\u00a0]?123/;
```
dilijev: Do you need to update tags in `test/Intl/rlexe.xml`?

dilijev:
Oh, please add a test for `useGrouping` (#3948)

lib/Runtime/Library/IntlEngineInterfaceExtensionObject.cpp
jackhorton: I think I did this yesterday while taking notes on all of the different classes, where they all lived, etc so that the elif of this if had the same corresponding functionality, in that they enabled auto-pointers for their respective backing library (I think it was specifically because I wanted to add that comment).

TL;DR I can revert it, as it was an artifact of re-organizing things so I could understand them better. Now that I understand them, if you don't think there is much value added, I can remove it.

lib/Runtime/Library/IntlEngineInterfaceExtensionObject.cpp
jackhorton: Noted, I can switch this back to elif defined()

lib/Runtime/Library/IntlEngineInterfaceExtensionObject.cpp
jackhorton: Yeah I am not sure why it didn't warning either, but I couldn't find a single other place where it was used in ICU code.

test/Intl/IntlReturnedValueTests.js
jackhorton: This was just cleanup, I was trying to understand what all of the tests did. The resolution of that bug was that the function call was supposed to have 2 entries by design, so the comment was irrelevant now.

test/Intl/CollatorUnsupportedLocales.js
jackhorton: Sortof -- note that it used to test best fit twice, rather than best fit followed by lookup (which is what the assert messages suggested it should do). Since its the same code both times and it was obviously a copy-paste bug, I thought itd be best to factor it out regardless.

test/Intl/NumberFormatOptions.js
jackhorton: I thought I already did, but this comment may have been from before I committed that. I updated a bunch of tags, though this test was already not marked as require_winglob, so it didn't need to be updated.

test/Intl/NumberFormatOptions.js
jackhorton: I like it, and will definitely consider it next time I have to edit a regex assert.

jackhorton:
`useGrouping` is tested by an existing part of NumberFormatOptions.js already, though I think there is only one assert for it. Should I add more or is that enough?

lib/Runtime/Library/IntlEngineInterfaceExtensionObject.cpp
```diff
@@ -866,23 +920,19 @@ namespace Js
         }
 
         Assert(numberFormatter);
-        // REVIEW (doilij): AutoPtr will call delete on IPlatformAgnosticResource and complain of non-virtual dtor. There are no IfFailThrowHr so is this still necessary?
-        // TODO (doilij): If necessary, introduce an PlatformAgnosticResourceAutoPtr that calls Cleanup() instead of delete on the pointer.
-        // AutoPtr<IPlatformAgnosticResource> numberFormatterGuard(numberFormatter);
 
         // TODO (doilij): Render signed zero.
-
+        // TODO(jahorto): isDecimalPointAlwaysDisplayed seems like a Windows-only construct
```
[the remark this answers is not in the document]
jackhorton: It also doesn't seem to be part of the spec currently, if it was before.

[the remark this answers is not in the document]
dilijev: It probably wasn't? Not all of our internal properties are designed to be surfaced via e.g. resolvedOptions. Anyway it's not now and it doesn't seem to be used. I think we can drop it.

test/Intl/IntlReturnedValueTests.js
dilijev: Yes you're correct about this comment/cleanup -- I meant the file in general since other lines changed. Looked like just formatting but I wanted to be sure.

dilijev:
@jackhorton 

> Should I add more or is that enough?

Your call

test/Intl/NumberFormatOptions.js
dilijev: When you do, note that both core and full tests will likely have baseline updates due to function IDs and/or line numbers in JS tests for stack traces and debug output.

PR #3951, opened 2017-10-13
Add unit tests for @@toStringTag caching, DictionaryTypeHandlerBases in missing-prop caches

ricobbe:
Should have been part of my last commit, but got overlooked.

For reviewers:

1. This is as much a test of whether I got missing-prop caching for DictionaryTypeHandlerBase right as it is a test of toStringTag caching.  As such, should I move these tests to a sibling of test/es6?

2. Are there any other crazy ways to add @@toStringTag to an existing object that I overlooked?

### DIFF
--- a/test/es6/toStringTag.js
+++ b/test/es6/toStringTag.js
@@ -262,6 +262,25 @@ var tests = [
             p[Symbol.toStringTag] = "prototypeTag";
             assert.areEqual("[object prototypeTag]", o.toString(), "second call to toString picks up @@toStringTag in prototype");
         }
+    },
+    {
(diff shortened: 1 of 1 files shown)


Penguinwizzard:
The path test/es6 is fine; to be honest it's not like we've kept terribly tight constraints on where tests should go.

Proxies:

```javascript
var handler = {
    get: function(target, name) {
        return name == Symbol.toStringTag ? "TEST" : target[name];
    }
};

var testobject = {};
var prox = new Proxy(testobject, handler);

console.log(testobject.toString());
console.log(prox.toString());
// Output:
// [object Object]
// [object TEST]
```

ricobbe:
Added proxy test.

PR #3953, opened 2017-10-13
Remove undefined functions from JavascriptOperators.h

xiaoyinl:
VS2017 marks these functions as not defined. This PR removes the unused function declarations. The rest three of them are defined in JavascriptOperators.cpp, but the parameter types are not the same (e.g. int vs int32), so this PR changes the typename so that VS can match the declaration and the definition.

### DIFF
--- a/lib/Runtime/Language/JavascriptOperators.h
+++ b/lib/Runtime/Language/JavascriptOperators.h
@@ -95,8 +95,6 @@ namespace Js
     {
     // Methods
     public:
-        static void FreeTemp(Var aValue);
-
         static BOOL IsArray(Var instanceVar);
         static BOOL IsConstructor(Var instanceVar);
         static BOOL IsConcatSpreadable(Var instanceVar);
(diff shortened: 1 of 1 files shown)


lib/Runtime/Language/JavascriptOperators.h
```diff
@@ -342,7 +333,7 @@ namespace Js
         static BOOL OP_SetElementI_JIT(Var instance, Var aElementIndex, Var aValue, ScriptContext* scriptContext, PropertyOperationFlags flags = PropertyOperation_None);
 
         static BOOL OP_SetElementI_UInt32(Var instance, uint32 aElementIndex, Var aValue, ScriptContext* scriptContext, PropertyOperationFlags flags = PropertyOperation_None);
-        static BOOL OP_SetElementI_Int32(Var instance, int aElementIndex, Var aValue, ScriptContext* scriptContext, PropertyOperationFlags flags = PropertyOperation_None);
+        static BOOL OP_SetElementI_Int32(Var instance, int32 aElementIndex, Var aValue, ScriptContext* scriptContext, PropertyOperationFlags flags = PropertyOperation_None);
```
dilijev: Assuming these type changes were made to reflect the types in the definitions in the .cpp file?

dilijev:
@Cellule might want to take a look as well?

lib/Runtime/Language/JavascriptOperators.h
xiaoyinl: Yes.

Cellule:
Looks good to me, it's the kind of change that if it builds then it's good :)

dilijev:
Testing internally. Will merge when complete.

PR #3955, opened 2017-10-13
ARM64: fix a couple FP checks

sigatrev:


lib/Backend/arm64/LowerMD.cpp
```diff
@@ -8037,12 +8037,9 @@ LowererMD::CheckOverflowOnFloatToInt32(IR::Instr* instrInsert, IR::Opnd* intOpnd
...
 
-    IR::RegOpnd *regOpnd= IR::RegOpnd::New(TyMachReg, this->m_func);
+    IR::RegOpnd *regOpnd= IR::RegOpnd::New(TyUint32, this->m_func);
 
-    instr = IR::Instr::New(Js::OpCode::MVN,
-        regOpnd,
-        IR::IntConstOpnd::New(0x80000000, TyInt32, this->m_func, true),
-        this->m_func);
+    instr = IR::Instr::New(Js::OpCode::MOVN, regOpnd, IR::IntConstOpnd::New(0x80000000, TyUint32, this->m_func, true), this->m_func);
     instrInsert->InsertBefore(instr);
```
aaronsgiles: This can be collapsed into a CMN against 0x80000000, instead of allocating a reg and MVN the value into it.

lib/Backend/arm64/ARM64Encoder.h
```diff
@@ -150,6 +150,8 @@ enum
     NEONREG_INVALID = 99,
 };
 
+#define LAST_FLOAT_REG_ENCODE NEONREG_D31
+
```
aaronsgiles: Since you trimmed off a few FP registers, we might consider loading them with commonly-used constant values (0, 0.5) as a future optimization.

lib/Backend/arm64/LowerMD.cpp
aaronsgiles: Actually, given the encoding support for SUBS (aka CMP), it looks like 0x80000000 can't be encoded directly either, so the legalizer will move that into a register.

aaronsgiles: If we can allocate a temporary register, this does it in 3 instructions:
tmp = EOR_ASR31 src, src    ; tmp = src ^ ((int32)src >> 31) -- converts 0x80000000 into 0x7fffffff
EOR tmp, 0x7fffffff    ; this is encodeable
CBZ tmp, label    ; branch if tmp == 0x7fffffff

aaronsgiles: An equivalent option, also 3 instructions:
tmp = EOR src, 0x80000000   ; gives 0 or -1 for overflow values
EOR_ASR31 tmp, tmp   ; tmp = tmp ^ ((int32)tmp >> 31) -- converts -1 to 0
CBZ tmp, label   ; branch if tmp was -1 or 0

PR #3956, opened 2017-10-13
Allow remaining DynamicTypeHandlers into missing-property caches

ricobbe:
As @MikeHolman and @obastemur requested, allow the remaining DynamicTypeHandlers into missing-property caches.

I've got a OneCirro run pending; I'll wait for that to finish before I land this.

I'm pretty sure that I've covered all of the subclasses of DynamicTypeHandler, but it's possible that I missed one.  The assert on line 1821 of JavascriptOperators.cpp guards against this, although it's fairly _ad hoc_.

Questions for reviewers:
1. Should I allow NullTypeHandlerBase and NullTypeHandler in JavascriptOperators.cpp?  They don't come up in the unit tests, but it should be safe, because adding a property to either of these types causes us to create a new type, so cache invalidation isn't a problem.
2. Because the assertion is fairly _ad hoc,_ I'm not sure how much value it adds.  Keep it?

### DIFF
--- a/lib/Runtime/Language/JavascriptOperators.cpp
+++ b/lib/Runtime/Language/JavascriptOperators.cpp
@@ -1808,12 +1808,19 @@ CommonNumber:
                 Output::Print(_u("MissingPropertyCaching: Missing property %d on slow path.\n"), propertyId);
             }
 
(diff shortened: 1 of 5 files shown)


ricobbe:
Internal testing indicates that I introduced an AV with these changes.  Unfortunately, I don't have time to investigate this further right now, so I'm going to withdraw the PR.  I'll try to get back to this after RS4 feature freeze.

PR #3957, opened 2017-10-13
External memory pressure: Add a collect threshold

obastemur:
`uncollectedExternalBytes` weren't in any specific use. Put a basic threashold for external data and improve handling the requests


lib/Common/Memory/Recycler.cpp
```diff
@@ -1098,20 +1098,31 @@ Recycler::Prime()
...
+        }
 
-bool Recycler::RequestExternalMemoryAllocation(size_t size)
-{
-    return recyclerPageAllocator.RequestAlloc(size);
+        // Generally normal GC can cleanup the uncollectedAllocBytes. But if external components
+        // do fast large allocations in a row, normal GC might not kick in. Let's force the GC
+        // here if we need to collect anyhow.
+
+        if (collectOnFail || this->autoHeap.uncollectedExternalBytes > KB128 || size >= KB64)
```
mike-kaufman: so, if I'm reading this correctly, if ever we end up in a place where `uncollectedExternalBytes is > 128KB`, every call to this function will trigger a GC.  Is there some way we can avoid this?  E.g., is there an API to "suggest a GC", and let some centralized GC logic worry about when to schedule that?

obastemur: No it won’t trigger GC. It will check whether it’s the time to do GC by checking other heuristics as is. If it really collects, well, uncollectedExternalBytes will be set to 0

mike-kaufman: sorry, not following why `uncollectedExternalBytes` will be 0.  Won't they only be 0 if GC could actually collect those?

obastemur: How do we plan to do that? (Since some of it could be really external) and the only reason for this parameter to be exist is that we could measure the external memory pressure and try GC

mike-kaufman: sorry, not following.  I have two questions:

1.  What is `uncollectedExternalBytes` and when does it go to 0?
2.  Are we ever subject to a condition where `uncollectedExternalBytes > KB128` and calling `CollectNow` will trigger GCs that are unable to reduce `uncollectedExternalBytes`?  (i.e., does this change subject us to overly-aggressive GC?)

obastemur: `uncollectedExternalBytes` symbolize an extra pressure number. i.e. host app allocated a piece of memory and notified us with the size.

At the moment, we do try to GC, regardless what `uncollectedExternalBytes` is..

So what I did is not limited to putting a `128KB` threshold. Whenever `requestAlloc` fails, we do actually try to GC, so adding up `uncollectedExternalBytes` and trying again was waste of cpu tick. Cleaned up there a bit and put a threshold.

dilijev:
@mike-kaufman @VSadov Has this scenario been sufficiently covered by changes to memory policy or should we consider merging this as well?

PR #3959, opened 2017-10-13
 Update Intl.cpp error handling, utf8 conversion, currency formatter creation

jackhorton:
This pull request takes the assert and CreateCurrencyFormatter changes from https://github.com/Microsoft/ChakraCore/pull/3913 and adds a smaller abstraction layer over Utf8 conversion than the Utf8Helper changes introduced there.

### DIFF
--- a/lib/Runtime/PlatformAgnostic/Platform/Common/Intl.cpp
+++ b/lib/Runtime/PlatformAgnostic/Platform/Common/Intl.cpp
@@ -54,124 +54,178 @@ typedef uint64_t uint64;
 #include <unicode/decimfmt.h>
 #pragma warning(pop)
 
+#include "CommonDefines.h" // INTL_ICU_DEBUG
+
+#if defined(__GNUC__) || defined(__clang__)
+// Output gets included through some other path on Windows
+class Output
+{
+public:
+    static size_t __cdecl Print(const char16 *form, ...);
+};
+#endif
+
+#define ICU_ERROR_FMT _u("INTL: %S failed with error code %S\n")
+#define ICU_EXPR_FMT _u("INTL: %S failed expression check %S\n")
+
+#ifdef INTL_ICU_DEBUG
+#define ICU_DEBUG_PRINT(fmt, msg) Output::Print(fmt, __func__, (msg))
+#else
+#define ICU_DEBUG_PRINT(fmt, msg)
+#endif
+
+#define ICU_RETURN(e, expr, r)                                                \
+    do                                                                        \
+    {                                                                         \
+        if (U_FAILURE(e))                                                     \
+        {                                                                     \
+            ICU_DEBUG_PRINT(ICU_ERROR_FMT, u_errorName(e));                   \
+            return r;                                                         \
+        }                                                                     \
+        else if (!(expr))                                                     \
+        {                                                                     \
+            ICU_DEBUG_PRINT(ICU_EXPR_FMT, u_errorName(e));                    \
+            return r;                                                         \
+        }                                                                     \
+    } while (false)
+
+#define ICU_ASSERT(e, expr)                                                   \
+    do                                                                        \
+    {                                                                         \
+        if (U_FAILURE(e))                                                     \
+        {                                                                     \
+            ICU_DEBUG_PRINT(ICU_ERROR_FMT, u_errorName(e));                   \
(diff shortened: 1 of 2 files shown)


lib/Runtime/PlatformAgnostic/Platform/Common/Intl.cpp
```diff
@@ -54,225 +54,224 @@ typedef uint64_t uint64;
...
+    // - char16 => char16_t -- see pal/inc/pal_mstypes.h
+    // All platforms:
+    // - UChar -> char16_t (all platforms)
+    static_assert(sizeof(UChar) == sizeof(char16), "ICU-based Intl logic assumes that UChar is compatible with char16");
+
+    // This function duplicates the Utf8Helper.h classes/methods
+    // See https://github.com/Microsoft/ChakraCore/pull/3913 for a (failed) attempt to use those directly
+    // Use CHECK_UTF_CONVERSION macro at the callsite to check the return value
+    // The caller is responsible for delete[]ing the memory returned here
+    static utf8char_t *Utf16ToUtf8(_In_ const char16 *src, _In_ const charcount_t srcLength, _Out_ size_t *destLength)
```
dilijev: Probably a better approach is to factor the bits that don't have unfortunate dependencies into a separate header and include that from Utf8Helper.h as well as here.

dilijev: (Rather than duplicating code which may become out of sync, leaving bugs in one of the versions.)

jackhorton: Factor the bits of Utf8Helper with bad dependencies? I can look into it, but in the grand scheme of things, this code was concerning to me both because it was duplicated within this file, in addition to the fact that it duplicated the helper class somewhere else. I dont know how tricky it will be to get Utf8Helper to a state that it can be included here -- thoughts on logging it as its own issue (if it isnt already?) and coming back to it later?

lib/Runtime/PlatformAgnostic/Platform/Common/Intl.cpp
```diff
@@ -54,225 +54,218 @@ typedef uint64_t uint64;
...
+public:
+    static size_t __cdecl Print(const char16 *form, ...);
+};
+#endif
+
+#define ICU_ERROR_FMT _u("INTL: %S failed with error code %S\n")
+#define ICU_EXPR_FMT _u("INTL: %S failed expression check %S\n")
+
+#ifdef INTL_ICU_DEBUG
+#define ICU_DEBUG_PRINT(fmt, msg) if (INTL_ICU_DEBUG) Output::Print(fmt, __func__, (msg))
```
MSLaguana: I'd suggest going with 
```cpp
#if defined(INTL_ICU_DEBUG) && INTL_ICU_DEBUG
#define ICU_DEBUG_PRINT(fmt, msg) Output::Print(fmt, __func__, (msg))
#else
#define ICU_DEBUG_PRINT(fmt, msg)
#endif
```

lib/Runtime/PlatformAgnostic/Platform/Common/Intl.cpp
dilijev: Not sure if we're saying the same thing, but this function works here, and it is logically a dependency of this code and of Utf8Helper.h (in the sense that that is where it lived before) -- so maybe you can factor out that dependency to another file.

lib/Runtime/PlatformAgnostic/Platform/Common/Intl.cpp
dilijev: Agreed. Keep macro logic in macros.

Also there's not currently and not intended to be any distinction between truthy INTL_ICU_DEBUG versus just being defined in the first place so you can keep the first line as just `#ifdef INTL_ICU_DEBUG` -- unless @jackhorton is intentionally adding a distinction here with the intent of changing the `#define` value of `INTL_ICU_DEBUG` to 0 while keeping it `#define`'d ?

jackhorton: Nope, I just brainfarted and left the if() in. Jimmy and I had discussed this offline and I hadn't intended to keep it.

dilijev:
Fixes https://github.com/Microsoft/ChakraCore/pull/3988

PR #3961, opened 2017-10-14
ARM64: legalize reg opnds

sigatrev:


agarwal-sandeep:
>         // No legalization possible, just report error.

nit: Comment is no longer valid?

---
Refers to: lib/Backend/arm64/LegalizeMD.cpp:210 in 2ef375c. [](commit_id = 2ef375c8bca2aecc09c356a7e0ec9857dc33462e, deletion_comment = False)

lib/Backend/arm64/LegalizeMD.cpp
```diff
@@ -71,6 +71,26 @@ void LegalizeMD::LegalizeInstr(IR::Instr * instr, bool fPostRegAlloc)
...
+    {
+        if (IRType_IsSignedInt(opnd->GetType()))
+        {
+            opnd->UnUse();
+            opnd->UseWithNewType(TyInt32, instr->m_func);
+        }
+        else
+        {
+            opnd->UnUse();
+            opnd->UseWithNewType(TyUint32, instr->m_func);
```
agarwal-sandeep: >TyUint32 [](start = 33, length = 8)

Float will work correctly?

aaronsgiles: Agree with the float concern here.

PR #3962, opened 2017-10-14
Initialize Intl/JsBuiltin extension objects when needed only

meg-gupta:
When we add JsBuiltins to JavascriptLibrary we do not want Intl to be loaded every time the JsBuiltins are needed.
Make Intl loading on-demand.

### DIFF
--- a/lib/Runtime/Library/EngineInterfaceObject.cpp
+++ b/lib/Runtime/Library/EngineInterfaceObject.cpp
@@ -159,7 +159,7 @@ namespace Js
 
         for (uint i = 0; i <= MaxEngineInterfaceExtensionKind; i++)
         {
-            if (engineExtensions[i] != nullptr)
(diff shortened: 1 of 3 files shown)


dilijev:
Curious of a repro case where Intl was being loaded unnecessarily? You could probably add a debugger test case to show the Intl initialization calls happening inappropriately to confirm that this fixes the scenario.

meg-gupta:
@dilijev Intl is not loaded unnecessarily now, I am working on bringing https://github.com/Microsoft/ChakraCore/tree/builtins to master with implementation of Array Iterator. 
With that, we will need to make sure Intl is not loaded when javascript array iterators are used.

dilijev:
@meg-gupta Gotcha, thanks!

dilijev:
@meg-gupta the PR said this task was incomplete. Was other work done to finish this task?

meg-gupta:
@dilijev Actually, we don't need this PR, the issue described in the PR was fixed.

PR #3963, opened 2017-10-14
Remove Intl from Builtin names on EngineInterface

meg-gupta:
Use a common name for global object functions that will be used by both Intl and JsBuiltins.
This change only removes the "Intl" prefix on these methods

### DIFF
--- a/lib/Runtime/Library/EngineInterfaceObject.h
+++ b/lib/Runtime/Library/EngineInterfaceObject.h
@@ -88,12 +88,12 @@ namespace Js
 
 #ifndef GlobalBuiltIn
 #define GlobalBuiltIn(global, method) \
-            static NoProfileFunctionInfo Intl_BuiltIn_##global##_##method##; \
+            static NoProfileFunctionInfo BuiltIn_##global##_##method##; \
 
 #define GlobalBuiltInConstructor(global)
 
 #define BuiltInRaiseException(exceptionType, exceptionID) \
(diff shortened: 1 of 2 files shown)


lib/Runtime/Library/EngineInterfaceObject.h
```diff
@@ -123,7 +123,7 @@ namespace Js
 #define GlobalBuiltInConstructor(global)
 
 #define BuiltInRaiseException(exceptionType, exceptionID) \
-        static Var EntryIntl_BuiltIn_raise##exceptionID(RecyclableObject *function, CallInfo callInfo, ...);
+        static Var EntryBuiltIn_raise##exceptionID(RecyclableObject *function, CallInfo callInfo, ...);
```
dilijev: Let's use `Entry_` to match the convention used for other native entry points.

dilijev:
Also might be a good idea to run `RegenAllByteCode.cmd` to ensure that the library bytecode doesn't change as a result (I expect that it shouldn't, but just in case).

lib/Runtime/Library/EngineInterfaceObject.h
meg-gupta: Will do.

meg-gupta:
@digitalinfinity yes that's right. Sorry this PR fell off my radar!

meg-gupta:
@dilijev checked that, bytecode doesn't change, because it relies on the property ids and not the name.

PR #3970, opened 2017-10-16
ARM64: shift opnd type

sigatrev:
fix type mismatch

### DIFF
--- a/lib/Backend/arm64/LegalizeMD.cpp
+++ b/lib/Backend/arm64/LegalizeMD.cpp
@@ -71,6 +71,37 @@ void LegalizeMD::LegalizeInstr(IR::Instr * instr, bool fPostRegAlloc)
     LegalizeSrc(instr, instr->GetSrc2(), 2, fPostRegAlloc);
 }
 
+void LegalizeMD::LegalizeRegOpnd(IR::Instr* instr, IR::Opnd* opnd)
+{
+    // Arm64 does not support 8 or 16 bit register usage, so promote anything smaller than 32 bits up to 32 bits.
+    // Arm64 does not support 32 bit float register usage, so promote 32 bit floats to 64 bit.
+    IRType ty = opnd->GetType();
+    switch(ty)
+    {
+    case TyInt8:
+    case TyInt16:
+        ty = TyInt32;
+        break;
+    
+    case TyUint8:
+    case TyUint16:
+        ty = TyUint32;
+        break;
+
+    case TyFloat32:
+        ty = TyFloat64;
+        break;
+    }
+
+    if (ty != opnd->GetType())
+    { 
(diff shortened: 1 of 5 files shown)


lib/Backend/arm64/LowerMD.cpp
```diff
@@ -2680,7 +2680,7 @@ void LowererMD::ChangeToShift(IR::Instr *const instr, const bool needFlags)
     {
         // In the variable case, generate code to do the mask.
         IR::Opnd *const src2 = instr->UnlinkSrc2();
-        instr->SetSrc2(IR::RegOpnd::New(TyMachReg, func));
+        instr->SetSrc2(IR::RegOpnd::New(TyInt32, func));
```
agarwal-sandeep: Should this be TyInt8 and then legalizer take care of it?

lib/Backend/arm64/LegalizeMD.cpp
```diff
@@ -71,6 +71,37 @@ void LegalizeMD::LegalizeInstr(IR::Instr * instr, bool fPostRegAlloc)
     LegalizeSrc(instr, instr->GetSrc2(), 2, fPostRegAlloc);
 }
 
+void LegalizeMD::LegalizeRegOpnd(IR::Instr* instr, IR::Opnd* opnd)
+{
+    // Arm64 does not support 8 or 16 bit register usage, so promote anything smaller than 32 bits up to 32 bits.
+    // Arm64 does not support 32 bit float register usage, so promote 32 bit floats to 64 bit.
+    IRType ty = opnd->GetType();
```
aaronsgiles: This comment is confusing to me. Arm64 as an architecture doesn't support 8 or 16 bit integer operations on registers, this is true. However, it does fully support 32 bit float operations. Now, we may not currently be doing anything with that (though I understand WASM will need it), but auto-promoting 32-bit floats to 64-bit seems dubious to me.

sigatrev: That change is from an already merged PR, I guess it's being shown here because the PR hasn't been pumped back to git yet. Either way, I must have misinterpreted the some of the comments on that PR, so I'll have to fix the logic/comment

lib/Backend/arm64/LowerMD.cpp
sigatrev: While the result of the operation is guaranteed to fit into int8, the src1 operand of the AND is int32 so I think int32 dst makes more sense semantically?

agarwal-sandeep: Fine

lib/Backend/arm64/LegalizeMD.cpp
aaronsgiles: Ah, that explains why I saw it before. I didn't see your final solution before it merged. I think the concern was that your previous check only looked at the size, and therefore might also impact float parameters. I don't necessarily think we were suggesting you do any kind of float promotion, just that you don't accidentally do something by only checking the size and not the type.

PR #3971, opened 2017-10-16
Do not load 'this' explicitly for class expression call target

boingoing:
We had a special case for when a class expression is emitted as a call
target which tried to load the 'this' value for the enclosing function.
Sometimes enclosing function has no 'this' binding (as in global
function or global lambda) which would mean there is no 'this' symbol
for us to load leading to null dereference.

Since class constructor cannot be called without the new keyword and we
always construct the value assigned to 'this' during function prolog,
there doesn't seem to be any reason at all to even load the 'this' value
here.

Fixes:
https://microsoft.visualstudio.com/OS/_workitems?id=14256595&_a=edit

### DIFF
--- a/lib/Runtime/ByteCode/ByteCodeEmitter.cpp
+++ b/lib/Runtime/ByteCode/ByteCodeEmitter.cpp
@@ -7665,14 +7665,6 @@ void EmitCallTarget(
         break;
     }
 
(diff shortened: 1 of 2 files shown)


boingoing:
Thanks, Tom.

PR #3972, opened 2017-10-16
Fix multi-unit characters after numeric literal

tcare:
Fixes OS: 14164940

I introduced a regession with my recent change to numeric literal spec complicance. After parsing a numeric literal, we were not checking if there was a multi-byte character before checking it as an IdStart character. In this case, we need to read the full character before checking.

This affected a handful of Test262 tests which used multi-byte trailing whitespace.


boingoing:
Are these ARM64 commits related?

tcare:
I pushed my review before they were pushed to GitHub. They're there now: https://github.com/Microsoft/ChakraCore/commits/master

tcare:
@dotnet-bot test OSX please

PR #3973, opened 2017-10-16
Fixing a bad merge

aneeshdk:
Fixing a bad merge


lib/Runtime/ByteCode/ByteCodeEmitter.cpp
```diff
@@ -3495,9 +3498,9 @@ void ByteCodeGenerator::StartEmitFunction(ParseNode *pnodeFnc)
     Scope * const bodyScope = funcInfo->GetBodyScope();
     Scope * const paramScope = funcInfo->GetParamScope();
 
-    if (funcInfo->byteCodeFunction->IsFunctionParsed() && funcInfo->GetParsedFunctionBody()->GetByteCode() == nullptr)
+    if (funcInfo->byteCodeFunction->IsFunctionParsed())
```
pleath: Do we still need this part of the change?

PR #3975, opened 2017-10-17
Add RTTI support for ICU

jackhorton:
Also creates consistent method names in Intl.cpp

### DIFF
--- a/lib/Runtime/Library/IntlEngineInterfaceExtensionObject.cpp
+++ b/lib/Runtime/Library/IntlEngineInterfaceExtensionObject.cpp
@@ -920,18 +920,10 @@ namespace Js
 
         Assert(numberFormatter);
 
-        // TODO (doilij): Render signed zero.
-        // TODO(jahorto): isDecimalPointAlwaysDisplayed seems like a Windows-only construct
-        bool isDecimalPointAlwaysDisplayed = false;
-        if (GetTypedPropertyBuiltInFrom(options, __isDecimalPointAlwaysDisplayed, JavascriptBoolean))
-        {
-            isDecimalPointAlwaysDisplayed = JavascriptBoolean::FromVar(propertyValue)->GetValue();
-        }
-
         if (GetTypedPropertyBuiltInFrom(options, __useGrouping, JavascriptBoolean))
         {
             bool useGrouping = JavascriptBoolean::FromVar(propertyValue)->GetValue();
-            SetGroupingUsed(numberFormatter, useGrouping);
+            SetNumberFormatGroupingUsed(numberFormatter, useGrouping);
         }
 
         // Numeral system is in the locale and is therefore already set on the icu::NumberFormat
--- a/lib/Runtime/PlatformAgnostic/Platform/Common/Intl.cpp
+++ b/lib/Runtime/PlatformAgnostic/Platform/Common/Intl.cpp
@@ -131,11 +131,20 @@ public:
         }                                                                     \
     } while (false)
 
+#define UNWRAP_RESOURCE(resource, innerType) reinterpret_cast<PlatformAgnosticIntlObject<innerType> *>(resource)->GetInstance();
+
 namespace PlatformAgnostic
 {
 namespace Intl
 {
     using namespace PlatformAgnostic::Resource;
+
+    // This file uses dynamic_cast and RTTI
+    // While most of ChakraCore has left this feature disabled for a number of reasons, ICU uses it heavily internally
+    // For instance, in creating NumberFormats, the documentation recommends the use of the NumberFormat::create* factory methods
+    // However, to use attributes and significant digits, you must be working with a DecimalFormat
+    // NumberFormat offers no indicator fields or convenience methods to convert to a DecimalFormat, as ICU expects RTTI to be enabled
+    // As such, RTTI is enabled only in the PlatformAgnostic project and only if IntlICU=true
 
     // lots of logic below requires utf8char_t ~= char and UChar ~= char16
     static_assert(sizeof(utf8char_t) == sizeof(char), "ICU-based Intl logic assumes that utf8char_t is compatible with char");
@@ -269,6 +278,14 @@ namespace Intl
         icu::NumberFormat *nf = formatterFactory(locale, error);
         ICU_ASSERT_RETURN(error, true, E_INVALIDARG);
 
+        // If the formatter produced a DecimalFormat, force it to round up
+        icu::DecimalFormat *df = dynamic_cast<icu::DecimalFormat *>(nf);
+        if (df)
+        {
+            df->setAttribute(UNUM_ROUNDING_MODE, UNUM_ROUND_HALFUP, error);
+            ICU_ASSERT(error, true);
+        }
+
         IPlatformAgnosticResource *formatterResource = new PlatformAgnosticIntlObject<icu::NumberFormat>(nf);
         if (!formatterResource)
         {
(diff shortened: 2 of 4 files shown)


jackhorton:
I mention it a bit in a comment, but for reference, the implementation of `unum_clone` [here](http://source.icu-project.org/repos/icu/icu/tags/release-58-rc/source/i18n/unum.cpp) is a good example of why we need RTTI/dynamic_cast to safely convert NumberFormats to DecimalFormats, which is required for certain operations like working with sigfigs or rounding modes (both of which relate to a test failure in NumberFormat.js)

digitalinfinity:
The change seems to have a lot more commits than what the PR describes- do you need to rebase?

dilijev:
@digitalinfinity Medusa is timing out and thus is behind. @tcare is looking into it.

lib/Runtime/PlatformAgnostic/Platform/Common/Intl.cpp
```diff
@@ -327,36 +342,39 @@ namespace Intl
...
+        // TODO(jahorto): Determine if we could ever have a NumberFormat that isn't a DecimalFormat (and if so, what to do here)
+        icu::DecimalFormat *df = dynamic_cast<icu::DecimalFormat *>(nf);
+        if (df)
+        {
+            df->setMinimumSignificantDigits(minSigDigits);
+            df->setMaximumSignificantDigits(maxSigDigits);
+        }
+        else
+        {
+            AssertMsg(false, "Could not cast an icu::NumberFormat to an icu::DecimalFormat");
```
dilijev: Add a comment explaining why this is an Assert and not FailFast.

To wit: if we can't use DecimalFormat-specific features because we didn't get a DecimalFormat, we should not crash. Best effort is good enough for Intl outputs, and we'd assume this is a transient issue. However, non-DecimalFormat-based output might be regarded as buggy, especially if consistently wrong.  We'd like to use Debug builds to detect if this is the case and how prevalent it is, and we can make a further determination if and when we see failures here.

lib/Runtime/Library/IntlEngineInterfaceExtensionObject.cpp
```diff
@@ -920,18 +920,10 @@ namespace Js
 
         Assert(numberFormatter);
 
-        // TODO (doilij): Render signed zero.
-        // TODO(jahorto): isDecimalPointAlwaysDisplayed seems like a Windows-only construct
-        bool isDecimalPointAlwaysDisplayed = false;
-        if (GetTypedPropertyBuiltInFrom(options, __isDecimalPointAlwaysDisplayed, JavascriptBoolean))
-        {
-            isDecimalPointAlwaysDisplayed = JavascriptBoolean::FromVar(propertyValue)->GetValue();
```
dilijev: Did we determine this is unused in WinGlob, too? Maybe you can remove it from Intl.js and the WinGlob path.

lib/Runtime/PlatformAgnostic/Platform/Common/Intl.cpp
```diff
@@ -327,36 +342,39 @@ namespace Intl
...
         nf->setMinimumIntegerDigits(minIntDigits);
         nf->setMinimumFractionDigits(minFracDigits);
         nf->setMaximumFractionDigits(maxFracDigits);
-        return S_OK;
     }
 
-    void SetGroupingUsed(IPlatformAgnosticResource *resource, const bool isGroupingUsed)
+    void SetNumberFormatGroupingUsed(IPlatformAgnosticResource *resource, const bool isGroupingUsed)
     {
         icu::NumberFormat *nf = reinterpret_cast<PlatformAgnosticIntlObject<icu::NumberFormat> *>(resource)->GetInstance();
```
dilijev: Might as well use the standard pattern (used above) here as well.

```
         // We know what actual type we stored in the IPlatformAgnosticResource*, so cast to it.
          auto *numFormatResource = reinterpret_cast<PlatformAgnosticIntlObject<icu::NumberFormat> *>(resource);
          icu::NumberFormat *nf = numFormatResource->GetInstance();
```

Or better, refactor to a method to call that takes `IPlatformAgnosticResource *resource` and returns `icu::NumberFormat *`

dilijev:
@jackhorton Note that the osx queue is really backed up. Try not to trigger too many extra builds as you iterate here (best effort) *shrug*

lib/Runtime/PlatformAgnostic/Platform/Common/Intl.cpp
jackhorton: That is at least as good if not better than anything I would have to say about it, so I will just copy/paste that.

lib/Runtime/Library/IntlEngineInterfaceExtensionObject.cpp
jackhorton: It is never referenced from Intl.js at all --  the only references in the entire project are the GetTypedProperty call that remains in WinGlob-specific code and the entry in JnDirectFields.h. So, I don't think its ever actually set, and we can probably remove it for WinGlob too. Should that happen in this pull or a separate one?

boingoing: Your call on removing the dead code. If it's truly not used, I would lean towards removing it now.

lib/Runtime/PlatformAgnostic/Platform/Common/Intl.cpp
```diff
@@ -131,12 +131,21 @@ class Output
...
+#define UNWRAP_RESOURCE(resource, innerType) reinterpret_cast<PlatformAgnosticIntlObject<innerType> *>(resource)->GetInstance();
+
 namespace PlatformAgnostic
 {
 namespace Intl
 {
     using namespace PlatformAgnostic::Resource;
 
+    // This file uses dynamic_cast and RTTI
+    // While most of ChakraCore has left this feature disabled for a number of reasons, ICU uses it heavily internally
```
boingoing: Is there a plan to enable RTTI for ChakraCore broadly at some point? I wonder if it's worth adding a TODO somewhere in `PlatformAgnostic` somewhere to make sure we remove the explicit enablement here at that point.

lib/Runtime/PlatformAgnostic/Platform/Common/Intl.cpp
```diff
@@ -327,36 +344,42 @@ namespace Intl
...
-        // REVIEW (doilij): Without RTTI support, we can't do dynamic_cast,
-        // which is the preferred way to know whether it was indeed created as a DecimalFormat,
-        // which is not something the API will indicate to us directly (create returns a NumberFormat which may or may not be a DecimalFormat)
-        icu::DecimalFormat *df = reinterpret_cast<icu::DecimalFormat *>(nf);
-        df->setMinimumSignificantDigits(minSigDigits);
-        df->setMaximumSignificantDigits(maxSigDigits);
-        return S_OK;
+        icu::NumberFormat *nf = UNWRAP_RESOURCE(resource, icu::NumberFormat);
+
+        // TODO(jahorto): Determine if we could ever have a NumberFormat that isn't a DecimalFormat (and if so, what to do here)
```
boingoing: It sounds like we've determined this via the else clause below?

lib/Runtime/PlatformAgnostic/Platform/Common/Intl.cpp
jackhorton: @Penguinwizzard felt very strongly about not (and, I think, never) doing so.

lib/Runtime/PlatformAgnostic/Platform/Common/Intl.cpp
jackhorton: Well, the `else { assert }` is just to allow us to hopefully find out _when_ its not the case, but I still don't know what the proper resolution to that scenario is.

lib/Runtime/PlatformAgnostic/Platform/Common/Intl.cpp
dilijev: I think the sentiment there is pretty strong. There's a binary size cost and runtime cost that make it a non-workable option for Chakra probably. No reason to do it in ChakraCore if we're not going to enable it for Chakra.

lib/Runtime/PlatformAgnostic/Platform/Common/Intl.cpp
dilijev: Personally I think continuing with a less-than-optimal format output is fine. The assert to give us some visibility on when/if it happens will be helpful.

PR #3978, opened 2017-10-17
ListOpnd for lowered IR

Cellule:
Create new Opnd ListOpnd. 
Currently available only at lower and after.
Allows to set a list of dst and src.

---
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/3978)

### DIFF
--- a/lib/Backend/Opnd.cpp
+++ b/lib/Backend/Opnd.cpp
@@ -160,6 +160,9 @@ Opnd::CloneDef(Func *func)
     case OpndKindIndir:
         return static_cast<IndirOpnd*>(this)->CloneDefInternal(func);
 
+    case OpndKindList:
+        return static_cast<ListOpnd*>(this)->CloneDefInternal(func);
+
     default:
         return this->Copy(func);
     };
@@ -190,6 +193,9 @@ Opnd::CloneUse(Func *func)
 
     case OpndKindIndir:
         return static_cast<IndirOpnd*>(this)->CloneUseInternal(func);
+
+    case OpndKindList:
+        return static_cast<ListOpnd*>(this)->CloneUseInternal(func);
 
     default:
         return this->Copy(func);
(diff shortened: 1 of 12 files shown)


Cellule:
@Microsoft/chakra-jit FYI
@agarwal-sandeep & @sigatrev , I believe you guys could use this for arm64.

Cellule:
I have tested the behavior in my change for WebAssembly Threads, but I am splitting that change into multiple PRs.

lib/Backend/Opnd.cpp
```diff
@@ -2288,6 +2304,110 @@ AddrOpnd::SetAddress(Js::Var address, AddrOpndKind addrOpndKind)
...
+        }
+    }
+    return true;
+}
+
+Opnd * ListOpnd::CloneUseInternal(Func * func)
+{
+    Assert(m_kind == OpndKindList);
+    int count = Count();
+    ListOpndType** opnds = JitAnewArray(func->m_alloc, RegOpnd*, count);
```
rajatd: >RegOpnd [](start = 55, length = 7)

Use ListOpndType for consistency? (here and below)

lib/Backend/Opnd.h
```diff
@@ -1467,6 +1472,105 @@ class AddrOpnd sealed : public Opnd
...
+    template<typename K1, typename... K>
+    void insert(int index, K1 arg, K... rest)
+    {
+        values[index] = arg;
+        insert(index + 1, rest...);
+    }
+    template<typename K>
+    void insert(int index, K last)
+    {
+        values[index] = last;
```
agarwal-sandeep: >values[index] = last; [](start = 8, length = 21)

Is overwriting allowed?

Cellule: No, this structure should be a type safe variable arguments buffer that can be optimized away at compile time.
@penguinwizzard originally wrote this struct.
I was thinking of making this generic for possibly other purposes.
Also, you're making me realize that the insert method should probably be private

Penguinwizzard: You could probably even go one step further and make ListOpndInit a private class inside ListOpnd.

PR #3982, opened 2017-10-17
ARM64: convert usage of RegR12 to SCRATCH_REGISTER

sigatrev:


lib/Backend/arm64/EncoderMD.cpp
```diff
@@ -843,22 +843,22 @@ EncoderMD::GenerateEncoding(IR::Instr* instr, BYTE *pc)
         break;
 
     case Js::OpCode::CMP_ASR31:
-        dst = instr->GetDst();
         src1 = instr->GetSrc1();
```
agarwal-sandeep: dst should be null right? if so add an assert.

lib/Backend/arm64/EncoderMD.h
```diff
@@ -36,7 +36,7 @@ enum InstructionType {
 #define LAST_CALLEE_SAVED_GP_REG    RegR28
 
 #define UNUSED_REG_FOR_STACK_ALIGN  RegR11
-#define SCRATCH_REG                 RegR12
+#define SCRATCH_REG                 RegR17
```
agarwal-sandeep: >RegR17 [](start = 36, length = 6)

I kept it as r12 because 

[http://infocenter.arm.com/help/topic/com.arm.doc.ihi0055b/IHI0055B_aapcs64.pdf](http://infocenter.arm.com/help/topic/com.arm.doc.ihi0055b/IHI0055B_aapcs64.pdf)

mentions that it can be used by linker

Registers r16 (IP0) and r17 (IP1) may be used by a linker as a scratch register between a routine and any subroutine it calls (for details, see §5.3.1.1, Use of IP0 and IP1 by the linker). They can also be used within a routine to hold intermediate values between subroutine calls. 

So didn't used it explicitly. @Aaron Giles please confirm if it's advisable to use r17

agarwal-sandeep: Make sense to use R17 as that's marked as RA_DONTALLOCATE

aaronsgiles: Yes I think r17 is better. The linker only matters for static compiled code and even only between function calls so it's totally save to use as a short term scratch register.

PR #3983, opened 2017-10-17
ARM64: Update tagging functions based on AMD64 logic. Implement fast add/sub.

aaronsgiles:


lib/Backend/arm64/LowerMD.cpp
```diff
@@ -3267,75 +3268,109 @@ LowererMD::GenerateFastAdd(IR::Instr * instrAdd)
...
     }
 
-    if (!opndSrc2->IsRegOpnd())
+    // Tagged ints?
+    bool isTaggedInts = false;
+    if (opndSrc1->IsTaggedInt())
     {
-        opndSrc2 = IR::RegOpnd::New(opndSrc2->GetType(), this->m_func);
-        LowererMD::CreateAssign(opndSrc2, instrAdd->GetSrc2(), instrAdd);
+        if (opndSrc2->IsTaggedInt())
```
agarwal-sandeep: > if (opndSrc2->IsTaggedInt()) [](start = 7, length = 29)

nit: 2 if can be combined in &&

aaronsgiles: There is a method to the madness, however. I started with the code for x86/amd64 from LowerMDShared.cpp and tried to make pretty minimal changes. The if structure you mention comes from there. In the interest of keeping the code more closely aligned, I explicitly chose not to clean up or modify the structure so that the two functions can be easily compared to track changes going forward.

PR #3984, opened 2017-10-17
ARM64: use two sources for CMP_ASR31

sigatrev:


lib/Backend/arm64/EncoderMD.cpp
```diff
@@ -843,22 +843,22 @@ EncoderMD::GenerateEncoding(IR::Instr* instr, BYTE *pc)
         break;
 
     case Js::OpCode::CMP_ASR31:
-        dst = instr->GetDst();
         src1 = instr->GetSrc1();
-        Assert(dst->IsRegOpnd());
+        src2 = instr->GetSrc2();
```
agarwal-sandeep: dst should be null right? if so add an assert.

sigatrev: assert added

PR #3985, opened 2017-10-17
fixing build break on arm64 fre

mike-kaufman:


lib/Backend/arm64/EncoderMD.cpp
```diff
@@ -1208,8 +1208,10 @@ EncoderMD::GenerateEncoding(IR::Instr* instr, BYTE *pc)
 
     // Opcode not yet implemented
     default:
+#ifdef DBG
```
rajatd: Please use DBG_DUMP instead of DBG here.

mike-kaufman: OK, should this be `DBG && DBG_DUMP`?  The build break was because `instr->Dump` wasn't defined in fre builds.

rajatd: Just using DBG_DUMP should be fine. It is defined in debug builds only, but we have it such that we could define it in test builds if we wanted to.

PR #3988, opened 2017-10-17
Make Date.parse() support milliseconds (Fix #3980)

### LINKED ISSUE #3980: Date.parse() not working with milliseconds
Calling `Date.parse()` with the format `YYYY-MM-DD HH:mm:ss.sss`  returns `NaN`:
`Date.parse('2017-10-17 08:00:00.000')`

I'm not sure whether the specification requires this format, but it works in Chrome and Firefox.

Tested in Edge 40.15063.674.0

xiaoyinl:
As reported in #3980, `Date.parse()` doesn't recongize the format `YYYY-MM-DD HH:mm:ss.sss`. This PR updates `DateImplementation::UtcTimeFromStrCore()` to correctly parse this format which is supported by other engines.

Fixes #3980

### DIFF
--- a/lib/Runtime/Library/DateImplementation.cpp
+++ b/lib/Runtime/Library/DateImplementation.cpp
@@ -1032,6 +1032,7 @@ Error:
             ssNil,
             ssMinutes,
             ssSeconds,
+            ssMillisecond,
             ssAddOffset,
             ssSubOffset,
             ssDate,
@@ -1041,6 +1042,7 @@ Error:
 
         char16 *pchBase;
         char16 *pch;
+        char16 *pchEndOfDigits = nullptr;
         char16 ch;
         char16 *pszSrc = nullptr;
 
@@ -1052,6 +1054,7 @@ Error:
         int32 lwMonth = lwNil;
         int32 lwDate = lwNil;
         int32 lwTime = lwNil;
+        int32 lwMillisecond = lwNil;
         int32 lwZone = lwNil;
         int32 lwOffset = lwNil;
 
@@ -1257,15 +1260,15 @@ Error:
 
             for (lwT = ch - '0'; !FBig(*pch) && isdigit(*pch); pch++)
             {
+                // to avoid overflow
+                if (pch - pchBase > 6)
+                {
+                    goto LError;
+                }
                 lwT = lwT * 10 + *pch - '0';
             }
 
-            // to avoid overflow
-            if (pch - pchBase > 6)
-            {
-                goto LError;
-            }
-
+            pchEndOfDigits = pch;
             // skip to the next real character
             while (0 != (ch = *pch) && (ch <= ' ' || classifier->IsBiDirectionalChar(ch)))
             {
@@ -1306,6 +1309,35 @@ Error:
                     if (lwT >= 60)
                         goto LError;
                     lwTime += lwT;
+                    ss = (ch == '.') ? (pch++, ssMillisecond) : ssNil;
+                    break;
+                }
+                case ssMillisecond:
+                {
+                    AssertMsg(isNextFieldDateNegativeVersion5 == false, "isNextFieldDateNegativeVersion5 == false");
+
+                    size_t numOfDigits = static_cast<size_t>(pchEndOfDigits - pchBase);
+                    if (numOfDigits <= 1)
+                    {
+                        // 1 digit only, treat it as hundreds
+                        lwMillisecond = lwT * 100;
+                    }
+                    else if (numOfDigits <= 2)
+                    {
+                        // 2 digit only, treat it as tens
+                        lwMillisecond = lwT * 10;
+                    }
+                    else if (numOfDigits <= 3)
+                    {
+                        // canonical 3 digit, per EcmaScript spec
+                        lwMillisecond = lwT;
+                    }
+                    else
+                    {
+                        // ignore any digits beyond the third
+                        lwMillisecond = (pchBase[0] - '0') * 100 + (pchBase[1] - '0') * 10 + (pchBase[2] - '0');
+                    }
+
                     ss = ssNil;
                     break;
                 }
@@ -1508,6 +1540,11 @@ Error:
             lwTime = 0;
         }
 
+        if (lwNil == lwMillisecond)
+        {
+            lwMillisecond = 0;
+        }
+
         if (lwNil != lwZone)
         {
             lwTime -= lwZone * 60;
@@ -1523,7 +1560,7 @@ Error:
         }
 
         // Rebuild time.
-        tv = TvFromDate(lwYear, lwMonth, lwDate - 1, (double)lwTime * 1000);
+        tv = TvFromDate(lwYear, lwMonth, lwDate - 1, (double)lwTime * 1000 + lwMillisecond);
         if (!fUtc)
         {
             tv = GetTvUtc(tv, scriptContext);
--- a/test/Date/DateParse.js
+++ b/test/Date/DateParse.js
@@ -108,8 +108,8 @@ for (var ts in toStrings)
 // No milliseconds
 WScript.Echo(Date.parse("2011-11-08T19:48:43"));
 
-// Missing digits after .
-WScript.Echo(Date.parse("2011-11-08T19:48:43."));
+// Missing digits after .: invalid
+WScript.Echo(Date.parse("2011-11-08T19:48:43.Z"));
 
 // milliseconds, 1 digit only, treat it as hundreds
 WScript.Echo(Date.parse("2011-11-08T19:48:43.1"));
--- a/test/Date/DateParse3.js
+++ b/test/Date/DateParse3.js
@@ -1,0 +1,42 @@
+//-------------------------------------------------------------------------------------------------------
+// Copyright (C) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE.txt file in the project root for full license information.
+//-------------------------------------------------------------------------------------------------------
+
+// Test non-ISO format with milliseconds
+// using colon as millisecond separator is not allowed
+runTest("2011-11-08 19:48:43:", "2011-11-08T19:48:43.000");  // valid, last colon is ignored
+runTest("2011-11-08 19:48:43:1", null);
+runTest("2011-11-08 19:48:43:10", null);
+runTest("2011-11-08 19:48:43:100", null);
+
+// use dot as millisecond separator
+runTest("2011-11-08 19:48:43.", "2011-11-08T19:48:43.000");
+runTest("2011-11-08 19:48:43.1", "2011-11-08T19:48:43.100");
+runTest("2011-11-08 19:48:43.1 ", "2011-11-08T19:48:43.100");
+runTest("2011-11-08 19:48:43. 1", "2011-11-08T19:48:43.100");
+runTest("2011-11-08 19:48:43.01", "2011-11-08T19:48:43.010");
+runTest("2011-11-08 19:48:43.001", "2011-11-08T19:48:43.001");
+runTest("2011-11-08 19:48:43.0001", "2011-11-08T19:48:43.000");
+runTest("2011-11-08 19:48:43.00000001", null);  // having more than 7 consecutive digits causes overflow
+runTest("2011-11-08 19:48:43.10", "2011-11-08T19:48:43.100");
+runTest("2011-11-08 19:48:43.100", "2011-11-08T19:48:43.100");
+runTest("2011-11-08 19:48:43.1000", "2011-11-08T19:48:43.100");
+runTest("2011-11-08 19:48:43.12345", "2011-11-08T19:48:43.123");
+
+function runTest(dateToTest, isoDate)
+{
+    if (isoDate === null) {
+        if (isNaN(Date.parse(dateToTest))) {
+            console.log("PASS");
+        } else {
+            console.log("Wrong date parsing result: Date.parse(\"" + dateToTest + "\") should return NaN");
+        }        
+    } else {
+        if (Date.parse(dateToTest) === Date.parse(isoDate)) {
+            console.log("PASS");            
+        } else {
+            console.log("Wrong date parsing result: Date.parse(\"" + dateToTest + "\") should equal Date.parse(\"" + isoDate + "\")");
+        }
+    }
+}
--- a/test/Date/parseISO.js
+++ b/test/Date/parseISO.js
@@ -60,7 +60,7 @@ runTest("+300000-01-01T01:01:01.001Z");
 writeStats();
 
 echo("////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////");
-echo("// Potential cross-browser compatilibity issues");
+echo("// Potential cross-browser compatibility issues");
 echo("");
 
 echo("// Leading and trailing whitespace, nulls, or non-whitespace non-nulls");
--- a/test/Date/rlexe.xml
+++ b/test/Date/rlexe.xml
@@ -35,6 +35,11 @@
   </test>
   <test>
     <default>
+      <files>DateParse3.js</files>
+    </default>
+  </test>
+  <test>
+    <default>
       <files>toISO_3.js</files>
       <baseline>toISO_3.baseline</baseline>
     </default>


dilijev:
Thanks for the contribution! Initially, looks good. I'll review more closely once tests are added.

xiaoyinl:
Thank you for the review @dilijev! I have already found two bugs in my implementation: 1 or 2-digit millisecond specifies the hundreds and tens digits; more than 4 digit millisecond should ignore digits after the third digit. 

I'm fixing them now.

test/Date/DateParse.js
```diff
@@ -174,3 +174,57 @@ WScript.Echo(Date.parse("2011/11-08 19:48:43"));
...
+WScript.Echo(Date.parse("2011-11-08 19:48:43.00001"));
+WScript.Echo(Date.parse("2011-11-08 19:48:43.00001a"));
+WScript.Echo(Date.parse("2011-11-08 19:48:43.00000"));
+WScript.Echo(Date.parse("2011-11-08 19:48:43.00000a"));
+
+WScript.Echo(Date.parse("2011-11-08 19:48:43.00000000000001"));
+WScript.Echo(Date.parse("2011-11-08 19:48:43.00000000000001a"));
+
+WScript.Echo(Date.parse("2011-11-08 19:48:43.12340000000001"));
+WScript.Echo(Date.parse("2011-11-08 19:48:43.12340000000001a"));
```
dilijev: nit: newline at end of file
You probably need to update the test baseline as well.

lib/Runtime/Library/DateImplementation.cpp
```diff
@@ -1306,6 +1307,27 @@ namespace Js {
...
                     lwTime += lwT;
+                    ss = (ch == ':' || ch == '.') ? (pch++, ssMillisecond) : ssNil;
+                    break;
+                }
+                case ssMillisecond:
+                {
+                    AssertMsg(isNextFieldDateNegativeVersion5 == false, "isNextFieldDateNegativeVersion5 == false");
+
+                    if (lwT >= 0 && lwT < 10)
+                        lwMillisecond = lwT * 100;
```
dilijev: Please always put braces around if-else bodies, even if it's a single statement.

xiaoyinl:
@dilijev Both commends are addressed. There's an unrelated small change in the previous commit: the overflow check is moved to L1262, inside the for loop, so that `lwT` won't get overflowed, avoiding an undefined behavior.

dilijev:
@Cellule can you review, especially with regards to ignoring everything after the 3rd millisecond digit?

lib/Runtime/Library/DateImplementation.cpp
```diff
@@ -1306,6 +1307,35 @@ namespace Js {
...
                         goto LError;
                     lwTime += lwT;
+                    ss = (ch == ':' || ch == '.') ? (pch++, ssMillisecond) : ssNil;
+                    break;
+                }
+                case ssMillisecond:
+                {
+                    AssertMsg(isNextFieldDateNegativeVersion5 == false, "isNextFieldDateNegativeVersion5 == false");
+
+                    size_t numOfDigits = static_cast<size_t>(pch - pchBase);
```
Cellule: This is incorrect.
First of all, you can't detect the number of digits from `pch` since there can be ignored characters like 
`console.log(Date.parse("2011-11-08 19:48:43:1"));` needs to have the same result as `console.log(Date.parse("2011-11-08 19:48:43:1 "));`

If only 1 digit is present, then it is not a hundred, but a digit `19:48:43:1` has `1`ms not `100`  similarly `19:48:43:15` has `15`ms not `150`

You can rewrite this whole case with
```cpp
                case ssMillisecond:
                {
                    AssertMsg(isNextFieldDateNegativeVersion5 == false, "isNextFieldDateNegativeVersion5 == false");
                    // ignore any digits beyond the third also ignores leading zeros
                    lwMillisecond = lwT % 1000;
                    ss = ssNil;
                    break;
                }
```

test/Date/DateParse.js
```diff
@@ -174,3 +174,58 @@ WScript.Echo(Date.parse("2011/11-08 19:48:43"));
 
 // Multiple 'T's aren't allowed
 WScript.Echo(Date.parse("2011-11-08TT19:48:43"));
+
+// Verifying non-standard delimiters with milliseconds
+
+// Missing digits after . or :
+WScript.Echo(Date.parse("2011-11-08 19:48:43."));
```
Cellule: I personally really dislike these types of tests, because I have no idea if the baseline contains the right answer or not.
I would prefer something like
```js
function doTest(format, expected) {
  if (Date.parse(format) !== expected) {
    console.log(`Invalid date parsing result, expected ${format} to give ${expected}`);
  }
}

// Notice this test shows the expected result which doesn't match the current baseline output
doTest("2011-11-08 19:48:43:1", 1320810523001);
...
```

Cellule:
I am not familiar with the spec, but v8 accepts format to have leading zeros in the milliseconds like `console.log(Date.parse("2011-11-08 19:48:43:00111"));`

And also, instead of ignoring values past 3 digits, it returns `NaN` if the result is greater than `1000` like
`console.log(isNan(Date.parse("2011-11-08 19:48:43:1000")));` instead of if there are more than 6 digits

Similarly, `console.log(Date.parse("2011-11-08 19:48:43:000000001")); ` is fine on V8, but we would show `NaN` and if you keep adding zeros, it just gets ignored at a certain point.

I would have to dig into the spec to figure out who is right here.

lib/Runtime/Library/DateImplementation.cpp
xiaoyinl: I'm pretty sure the `1` in `Date.parse("2011-11-08 19:48:43.1")` is 100 ms, not 1 ms. Both SpiderMonkey and v8 return 1320799723100. If it meant to be 1 ms, it should be 1320799723001.

Cellule: Interesting
`Date.parse("2011-11-08 19:48:43.1")` is indeed `1320810523100`
However, `Date.parse("2011-11-08 19:48:43:1")` is `1320810523001`
It seems, depending on wether we use `.` or `:` as a separator, it changes the meaning the next value.

Also, the following
```
console.log(Date.parse("2011-11-08 19:48:43.0010"));
console.log(Date.parse("2011-11-08 19:48:43:0010"));
```
is also very different.

It would seem that your implementation matches more the `.` pattern and mine the `:` pattern.
Therefore, we need to differenciate between the 2 when interpreting the value.

Note, that we still need a different mechanism than `static_cast<size_t>(pch - pchBase);` to determine the number of digits to take spaces into account.
We could probably just calculate it before `// skip to the next real character`

xiaoyinl: Sorry I think I misunderstood your comment. Now I see `19:48:43.1` and `19:48:43:1` represent different time.

Cellule: I think I was also confused about that fact when I first wrote my original comment as I only tested with `:` initially

test/Date/DateParse.js
dilijev: I agree, but this whole test has that problem.

Also having a multiplied-out value showing the represented value would add more semantic meaning to the expected output.

The tests being added can use the better style but fixing the rest of the test is out of scope and up to @xiaoyinl whether to update.

xiaoyinl:
Thank you @Cellule and @dilijev! In the latest commit, I test it by comparing if the non-standard format gives the same time as the corresponding ISO format. I think this has the benefits you mentioned, and it doesn't depend on time zone, so it doesn't need to be disabled.

lib/Runtime/Library/DateImplementation.cpp
```diff
@@ -1339,6 +1342,18 @@ namespace Js {
                     ss = ssNil;
                     break;
                 }
+                case ssMillisecondColon:
+                {
+                    AssertMsg(isNextFieldDateNegativeVersion5 == false, "isNextFieldDateNegativeVersion5 == false");
+
+                    if (lwT >= 1000)
```
Cellule: can't we just trim to 1000 with `lwMillisecond = lwT % 1000` ?
It would effectively ignore the characters after the third digit

xiaoyinl: The thing is v8 returns NaN when millisecond is >= 1000 and `:` is used as delimiter.

dilijev:
In response to thread here: https://github.com/Microsoft/ChakraCore/pull/3988#discussion_r146093193
But as a non-review thread so context isn't lost in future updates.

This is a question for how the spec defines the behavior. However, now that I'm reading the spec on Date parse, it doesn't look like there is an option to use `:` as a separator before milliseconds.

https://tc39.github.io/ecma262/#sec-date-time-string-format

Notable quotes (emphasis mine)

> YYYY-MM-DDTHH:mm:ss.sssZ

> ":" (colon) appears literally ***twice*** in the string.

> . | "." (dot) appears literally in the string.

> sss | is the number of complete milliseconds since the start of the second ***as three decimal digits.***

https://tc39.github.io/ecma262/#sec-date.parse

> Unrecognizable Strings or dates containing illegal element values in the format String shall cause Date.parse to return NaN.

So it appears that the colon is not allowed at all, and should result in `NaN`. However, there seems to be a compat issue between engines, and that this feature is beyond the spec text. @bterlson can you confirm? Is there a proposal to add `:` in progress?

SpiderMonkey does not support `:`
v8 supports it, and NaN if more than 3 digits (makes sense given the semantics we're seeing here).

All engines support ms even if there are more than 3 digits, and even when there is a timezone specified afterwards (so the whole string was parsed). Seems like the spec language could use a little clarification about whether or not extra digits are legal here. @bterlson 

```
## Source
print(Date.parse('2011-11-08 19:48:43:123'));
print(Date.parse('2011-11-08 19:48:43:1234'));
print(Date.parse('2011-11-08 19:48:43.123'));
print(Date.parse('2011-11-08 19:48:43.1234'));
print(Date.parse('2011-11-08 19:48:43.1234+06:00'));

┌──────────────────┬───────────────┐
│ d8               │ 1320810523123 │
│ node             │ NaN           │
│                  │ 1320810523123 │
│                  │ 1320810523123 │
│                  │ 1320760123123 │
├──────────────────┼───────────────┤
│ ch-1.7.1         │ NaN           │
│ ch-master-latest │ NaN           │
│ jsc              │ NaN           │
│ node-ch          │ NaN           │
│                  │ NaN           │
├──────────────────┼───────────────┤
│ sm               │ NaN           │
│                  │ NaN           │
│                  │ 1320810523123 │
│                  │ 1320810523123 │
│                  │ 1320760123123 │
└──────────────────┴───────────────┘
```

It is true that we didn't support ms correctly before, we must at a minimum add that.

dilijev:
Not sure if this is current, and ECMA 262 differs from ISO 8601 (superset AFAICT), but the format indicated here for ISO 8601 date strings also doesn't allow `:` as the separator before ms. https://www.w3.org/TR/NOTE-datetime

lib/Runtime/Library/DateImplementation.cpp
Cellule: Right, looks good then

Cellule:
Looks good in term of implementation.
I would like to get confirmation on the spec part however

bterlson:
Gah sorry for missing this issue. A couple points:

1. I am not convinced that we need to support colon for ms. We are not in the business of implementing every non-standard extension v8 implements, but only those that have either a super strong use case or are necessary for compat/interop with existing real world code.
2. From a spec perspective, we are free to either support this extension or not. Date.parse is allowed to parse whatever formats we feel like.
3. For parsing ms in the standard date format, if it's fewer than or more than 3 characters it is not valid. However, we can still support these forms due to point 2 above.

My suggestion is to only implement the minimal semantics required to be spec compliant and run real world code. Thus, at most, parsing ms using '.' seems like something we should tackle. The rest I would prefer not to support.

xiaoyinl:
@dilijev @bterlson Done. I have removed the support for colon separator, and updated the test cases. If the code now looks good to you I will squash the commits.

dilijev:
As a curiosity: to the point about whether we should support millis with other than exactly 3 digits with `.`, and other behavior I've observed, here's some tests showing the similarities and differences in behavior:

```js
// jsc requires T instead of space, does not support : at all
// jsc's default timezone handling is different from everyone else, so specify Z explicitly
// nobody supports : with T in this form
print(Date.parse('2011-11-08T19:48:43Z'));
print(Date.parse('2011-11-08T19:48:43:1Z'));
print(Date.parse('2011-11-08T19:48:43:12Z'));
print(Date.parse('2011-11-08T19:48:43:123Z'));
print(Date.parse('2011-11-08T19:48:43:1234Z'));
```

```
┌──────────────────┬───────────────┐
│ ch-1.7.1         │ 1320781723000 │
│ ch-master-latest │ NaN           │
│ d8               │ NaN           │
│ jsc              │ NaN           │
│ node             │ NaN           │
│ node-ch          │               │
│ sm               │               │
└──────────────────┴───────────────┘
```

----

```js
// d8 only supports : when T is used
print(Date.parse('2011-11-08 19:48:43Z'));
print(Date.parse('2011-11-08 19:48:43:1Z'));
print(Date.parse('2011-11-08 19:48:43:12Z'));
print(Date.parse('2011-11-08 19:48:43:123Z'));
print(Date.parse('2011-11-08 19:48:43:1234Z')); // (nobody)
```

```
┌──────────────────┬───────────────┐
│ ch-1.7.1         │ 1320781723000 │
│ ch-master-latest │ NaN           │
│ node-ch          │ NaN           │
│ sm               │ NaN           │
│                  │ NaN           │
├──────────────────┼───────────────┤
│ d8               │ 1320781723000 │
│ node             │ 1320781723001 │
│                  │ 1320781723012 │
│                  │ 1320781723123 │
│                  │ NaN           │
├──────────────────┼───────────────┤
│ jsc              │ NaN           │
│                  │ NaN           │
│                  │ NaN           │
│                  │ NaN           │
│                  │ NaN           │
└──────────────────┴───────────────┘
```

----

```js
## Source
print(Date.parse('2011-11-08T19:48:43Z'));
print(Date.parse('2011-11-08T19:48:43.1Z'));
print(Date.parse('2011-11-08T19:48:43.12Z'));
print(Date.parse('2011-11-08T19:48:43.123Z'));
print(Date.parse('2011-11-08T19:48:43.1234Z'));
print(Date.parse('2011-11-08T19:48:43.123456789Z'));
```

```
┌──────────────────┬────────────────────┐
│ ch-1.7.1         │ 1320781723000      │
│ ch-master-latest │ 1320781723100      │
│ d8               │ 1320781723120      │
│ node             │ 1320781723123      │
│ node-ch          │ 1320781723123      │
│ sm               │ 1320781723123      │
├──────────────────┼────────────────────┤
│ jsc              │ 1320781723000      │
│                  │ 1320781723100      │
│                  │ 1320781723120      │
│                  │ 1320781723123      │
│                  │ 1320781723123.4    │
│                  │ 1320781723123.4568 │
└──────────────────┴────────────────────┘
```

dilijev:
```
00:05:23.975 Output:
00:05:23.975 ----------------------------
00:05:23.975 Wrong date parsing result: Date.parse("2011-11-08 19:48:43:") should return NaN
```

Yeah we apparently will parse everything that we understand in certain cases and ignore trailing characters. If you want to disambiguate whether it should be allowed, put the explicit Z on the end to force parsing the entire string.

xiaoyinl:
@dilijev I will leave date like "2011-11-08 19:48:43:" valid. It's valid even in the current version of Edge, so I don't think it's necessary to explicitly reject that.

dilijev:
@xiaoyinl Fair enough. Could you rebase this PR to a single commit?

xiaoyinl:
@dilijev Done. Thank you for your reviews!

dilijev:
Testing and will merge soon.

dilijev:
Failed DateParse.js which is tagged as `exclude_jenkins`

```
test/Date/rlexe.xml-19-  <test>
test/Date/rlexe.xml-20-    <default>
test/Date/rlexe.xml:21:      <files>DateParse.js</files>
test/Date/rlexe.xml:22:      <baseline>DateParse_es5.baseline</baseline>
test/Date/rlexe.xml-23-      <!-- test is timezone-sensitive; remove exclude_jenkins after fix (Microsoft/ChakraCore#319) -->
test/Date/rlexe.xml-24-      <tags>exclude_jenkins</tags>
test/Date/rlexe.xml-25-    </default>
test/Date/rlexe.xml-26-  </test>
```

If your timezone is set to Pacific Time, you should be able to run that test locally and update the test to fix it up.

xiaoyinl:
@dilijev I've fixed the test case. It's a quite interesting error. `Date.parse("2011-11-08T19:48:43."));` treats the `T` as the military time zone for UTC-7. The mismatch was actually caused by #4016. I added a `Z` to the end which returns NaN as expected. Now the test should pass.

dilijev:
@xiaoyinl Sorry for the delay. Running an internal test run and will merge if it passes.

dilijev:
@xiaoyinl I sent you a message on Gitter. Did you get it?

dilijev:
@xiaoyinl I retargeted this PR to release/1.8. Would you mind cherry-picking or rebasing your changes on top of release/1.8 (to leave out the master-only changes that are not in release/1.8), and then force-pushing your branch?

dilijev:
Nevermind. I forgot about the feature where maintainers can push to your PR branches. I took care of it.

PR #3992, opened 2017-10-18
Legalize CMP in GenerateSmIntTest and disable inlining for ARM64

agarwal-sandeep:


sigatrev:
LGTM. Worth noting that GenerateSmIntPairTest also needs the CMP legalization

PR #3993, opened 2017-10-18
Disable inlining for x86 when there is try

meg-gupta:


digitalinfinity:
Curious- what is the motivation behind this change?

meg-gupta:
https://github.com/Microsoft/ChakraCore/pull/3832 - enabled inlining into functions with try. On x86, additional analysis is needed to compute arg offsets and stack offsets when there is a bailout in EH region, due to x86 specific calling conventions. Disabling till all corner cases are fixed.

PR #3994, opened 2017-10-18
Change var slot allocation order to avoid loading special symbols from incorrect slots

boingoing:
If a function contains nested named function declarations which require a binding, these will be bound to scope slots before scope slots are allocated for the var list. This logic is in `ByteCodeGenerator::StartEmitFunction` and calls `EnsureFncDeclScopeSlot` which ensures a scope slot for each such nested named function declaration.
    
Immediately after ensuring scope slots for function declarations, we then ensure scope slots for the user var list. Because scope slots must be initialized in the same order they are ensured (allocated) we need to make sure the logic in `ByteCodeGenerator::StartEmitFunction` matches the logic to emit storage for symbols in `ByteCodeGenerator::EmitOneFunction`.

The bug is that we are emitting the slot initialization in a different order from the order in which we ensured the slots. We are initializing some of the slots ensured for var list symbols before we initialize the slots ensured for nested function declarations. Specifically, these var list slots are ensured for the special symbols.

We emit storage for the special symbols (which are part of the var list) before the default arguments (because default arguments might access the special named values). However, we cannot emit storage for nested function declarations before the default arguments because the default arguments are not supposed to be able to access those nested functions as they are declared in the body.
    
Fix is to just ensure the scope slots in the same order we are going to initialize them. That means walking the var list and ensuring scope slots for those (including special symbols) before ensuring scope slots for nested named function declarations.
    
Fixes:
https://microsoft.visualstudio.com/OS/_workitems?id=14212259&_a=edit

### DIFF
--- a/test/Basics/SpecialSymbolCapture.js
+++ b/test/Basics/SpecialSymbolCapture.js
@@ -901,6 +901,15 @@ var tests = [
                 assert.areEqual(_this, this, "Regular 'this' binding is correct");
             }
             foo.call(_this)
+            
+            function bar(a = this) {
+                eval('');
+                assert.areEqual(_this, a, "Correct default value was assigned");
+                assert.areEqual(_this, this, "Regular 'this' binding is correct");
+                function b() { return 'b'; }
+                assert.areEqual('b', b(), "Nested functions are bound to the correct slot");
+            }
+            bar.call(_this)
         }
     },
     {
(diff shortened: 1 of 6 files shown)


test/Basics/SpecialSymbolCapture.js
```diff
@@ -901,6 +901,15 @@ var tests = [
...
+            function bar(a = this) {
+                eval('');
+                assert.areEqual(_this, a, "Correct default value was assigned");
+                assert.areEqual(_this, this, "Regular 'this' binding is correct");
+                function b() { return 'b'; }
+                assert.areEqual('b', b(), "Nested functions are bound to the correct slot");
+            }
+            bar.call(_this)
         }
     },
```
aneeshdk: Can you add one test case with a function definition in the param scope (non-split scope)?

boingoing:
Thank you!

PR #3996, opened 2017-10-18
Allow getter/setter calls to be inlined if the object is polymorphic but the identity of the getter/setter is consistent. 

pleath:
Add helpers to get accessors directly from the object if the accessor inline cache misses. Delete all the machine-specific lowerings related to accessor inline caches and use a shared lowering.


lib/Backend/Lower.cpp
```diff
@@ -6305,6 +6305,202 @@ Lowerer::LowerDelFld(IR::Instr *delFldInstr, IR::JnHelperMethod helperMethod, bo
...
+        opndType = IR::RegOpnd::New(TyMachReg, this->m_func);
+    }
+
+    opndBase = propertySymOpnd->CreatePropertyOwnerOpnd(m_func);
+    opndDst = instrLdFld->GetDst();
+    opndInlineCache = IR::RegOpnd::New(TyMachPtr, this->m_func);
+
+    labelFallThru = IR::LabelInstr::New(Js::OpCode::Label, this->m_func);
+    // Label to jump to (or fall through to) when bailing out
+    bailOutLabel = IR::LabelInstr::New(Js::OpCode::Label, instrLdFld->m_func, true /* isOpHelper */);
```
rajatd: nit: could you rename bailoutLabel to, say, helperLabel now that we won't be bailing out on a failed type check but call the helper?

dilijev:
@pleath are you still planning to merge this?

pleath:
I may. I need to see if it still contributes to perf now that we share types with accessors.

PR #3997, opened 2017-10-18
ARM64: Finished auditing all VarTag_Shift usage for ARM64, porting AMD64 as needed.

aaronsgiles:


lib/Backend/arm64/LowerMD.cpp
```diff
@@ -6001,92 +5863,75 @@ LowererMD::GenerateFastAbs(IR::Opnd *dst, IR::Opnd *src, IR::Instr *callInstr, I
         IR::AddrOpnd *varOpnd = src->AsAddrOpnd();
         Assert(varOpnd->IsVar() && Js::TaggedInt::Is(varOpnd->m_address));
 
-        int absValue = abs(Js::TaggedInt::ToInt32(varOpnd->m_address));
+        __int64 absValue = ::_abs64(Js::TaggedInt::ToInt32(varOpnd->m_address));
```
sigatrev: I think we'd prefer the typedef'd ``int64`` here instead.

PR #4003, opened 2017-10-19
Added assert.matches, centralized NumberFormat tests, removed other unnecessary tests

jackhorton:
This pull request makes the NumberFormat tests pass for ICU, bringing the total pass rate to 13/17


jackhorton:
One thing to note is that our intl tests are _very_ en-US specific. Should it be a goal to add more tests to ensure that things are working properly in other locales? There are some low-hanging fruit, like making sure the decimal point and grouping separator are switched in some (most?) european locales, and some basic tests for other alphabets.

Also, I just realized that this tests Intl.NumberFormat.format a bunch and never n.toLocaleString, should I add that @dilijev ?

dilijev:
@jackhorton Might be a good idea to exercise the same params more than once to exercise the caching behavior. It occurs to me I probably should add that kind of test with my PR.

Also look for tests of toLocaleString in other folders -- some may already exist. It makes sense to have tests for `-Intl-` behavior as well as the Intl implementation.

jackhorton:
Perhaps the new caches should be tested by ChakraFull by manually toggling the useCaches switch?

PR #4005, opened 2017-10-19
Improving handling of internal properties in JSON.stringify

MSLaguana:


lib/Runtime/Library/JSON.cpp
```diff
@@ -738,7 +738,7 @@ namespace JSON
                             for (uint32 i = 0; i < propertyCount; i++)
                             {
                                 id = typeHandler->GetPropertyId(scriptContext, (Js::PropertyId)i);
-                                if (id == Js::Constants::NoProperty)
```
jackhorton: Is this instead of or in addition to? Even if `Js::Constants::NoProperty < Js::InternalPropertyIds::Count` right now, it seems clearer to me and more future-proof to check both conditions.

MSLaguana: Done.

MSLaguana: On a side note, that seems to have fixed some debug-mode test failures, even though `Js::Constants::NoProperty == -1` and so the condition _should_ be identical as far as I can tell.

PR #4006, opened 2017-10-19
OS#14163877: Add caches for Date.prototype.toLocaleString etc, with default params.

### LINKED ISSUE #4009: Intl.DateTimeFormat().format.length is incorrect (0, should be 1)
```
## Source
print(new Intl.DateTimeFormat().format.length)

┌──────────────────┬───┐
│ d8               │ 1 │
│ jsc              │   │
│ node             │   │
│ sm               │   │
├──────────────────┼───┤
│ ch-1.7.1         │ 0 │
│ ch-dev-che3      │   │
│ ch-master-latest │   │
│ node-ch          │   │
└──────────────────┴───┘
```

`Intl.NumberFormat().format.length` is correct.

```
## Source
print(new Intl.NumberFormat().format.length)

┌──────────────────┬───┐
│ ch-1.7.1         │ 1 │
│ ch-dev-che3      │   │
│ ch-master-latest │   │
│ d8               │   │
│ jsc              │   │
│ node             │   │
│ node-ch          │   │
│ sm               │   │
└──────────────────┴───┘
```

### LINKED ISSUE #4012: Update EntryIntl_RegisterBuiltInFunction and Intl.js to use an enum instead of magic constants for entry point IDs
/cc @jackhorton

dilijev:
When calling `toLocaleString` functions that are implemented in Intl like `Date.prototype.toLocaleString`, that call is essentially sugar for creating an Intl.DateTimeFormat object, and formatting `this`. We naively throw away that object on the assumption that users will cache an `Intl.DateTimeFormat` object if they plan on making the same format call repeatedly. At a minimum, we can cache the formatter object created when the calls are made with no arguments passed in.

In particular, these two code patterns are semantically identical:

```
let d = new Date();
let formatter = new Intl.DateTimeFormat();
print(formatter.format(d));
```

```
let d = new Date();
print(d.toLocaleString());
```

This change significantly improves the running time when Date.prototype.toLocaleString is called in a loop with default options. #4007 has been opened to track further-reaching improvements related to caching behavior on implicitly-created i18n objects (applicable to both WinGlob and ICU).

The remaining gap in the measurement below between ch and d8 is likely related to debug versus release and WinGlob versus ICU.

```
>eshost -h d8,ch-dev-che3,ch-master-latest caching.js
#### d8
time spent 29 ms
time spent 19 ms
time spent 18 ms

#### ch-dev-che3 # (x64_debug, with this change to add caching behavior)
time spent 106 ms
time spent 43 ms
time spent 45 ms

#### ch-master-latest # (x64_debug, baseline)
time spent 2173 ms
time spent 1676 ms
time spent 1168 ms
```

caching.js

```
const iterations = 1000;
var arr = [];

function testMethod(method) {
    var start = new Date();
    testDateMethod(method);
    var diff = (new Date()) - start;
    print('time spent ' + diff + ' ms');
}

function testDateMethod(method) {
    arr = [];
    var d = new Date();
    for (var i = 0; i < iterations; i++) {
        arr.push(method.call(d));
    }
    return arr.length;
}

function test()
{
    testMethod(Date.prototype.toLocaleString);
    testMethod(Date.prototype.toLocaleDateString);
    testMethod(Date.prototype.toLocaleTimeString);
}

test();
```

No difference in runtime between calling the functions directly versus using `call`.

Additionally:
* Made some refactorings to improve readability of this code and similar code throughout Intl.js
* Fixed #4009: Intl.DateTimeFormat().format.length should be 1
* Fixed #4012: Use an enum instead of magic constants for builtin function IDs


dilijev:
>eshost -h d8,ch-dev-che3,ch-master-latest caching.js
#### d8
time spent 29 ms
time spent 19 ms
time spent 18 ms

#### ch-dev-che3 # (with this change to add caching behavior)
time spent 106 ms
time spent 43 ms
time spent 45 ms

#### ch-master-latest # (baseline)
time spent 2173 ms
time spent 1676 ms
time spent 1168 ms

lib/Runtime/Library/InJavascript/Intl.js
```diff
@@ -1593,44 +1593,122 @@
...
-                }
-                var value = callInstanceFunc(DateInstanceGetDate, new Date(this));
-                if (isNaN(value) || !isFinite(value)) {
-                    return "Invalid Date";
-                }
-                var stateObject = setPrototype({}, null);
-                InitializeDateTimeFormat(stateObject, arguments[0], ToDateTimeOptions(arguments[1], "any", "all"));
-                return String(platform.formatDateTime(Internal.ToNumber(this), stateObject));
-            }), 0);
+            {
```
jackhorton: What is the goal of creating new scopes for these?

lib/Runtime/Library/InJavascript/Intl.js
```diff
@@ -1593,44 +1593,122 @@
...
+                // scope for the cache variable
+                let __date_prototype_toLocaleString_default_formatterStateObject_cache = undefined;
+                platform.registerBuiltInFunction(tagPublicFunction("Date.prototype.toLocaleString", function () {
+                    if (typeof this !== 'object' || !(this instanceof Date)) {
+                        platform.raiseNeedObjectOfType("Date.prototype.toLocaleString", "Date");
+                    }
+                    var value = callInstanceFunc(DateInstanceGetDate, new Date(this));
```
jackhorton: nit: if we are adding new code, we should probably be using let/const consistently

jackhorton:
I mentioned this on my pull request as well, but perhaps the new caches should be tested by ChakraFull by manually toggling the useCaches switch?

lib/Runtime/Library/InJavascript/Intl.js
```diff
@@ -1593,44 +1593,122 @@
...
+                platform.registerBuiltInFunction(tagPublicFunction("Date.prototype.toLocaleTimeString", function () {
+                    if (typeof this !== 'object' || !(this instanceof Date)) {
+                        platform.raiseNeedObjectOfType("Date.prototype.toLocaleTimeString", "Date");
+                    }
+                    var value = callInstanceFunc(DateInstanceGetDate, new Date(this));
+                    if (isNaN(value) || !isFinite(value)) {
+                        return "Invalid Date";
+                    }
+
+                    let stateObject = undefined;
```
akroshg: refactor all these three instanced to one? this function is already so big to follow.

akroshg: well I get confused with other function - I take back that this function is big. But still good to refactor the code. All I see is different the cached var.

---
In reply to: [145771668](https://github.com/Microsoft/ChakraCore/pull/4006#discussion_r145771668) [](ancestors = 145771668)

lib/Runtime/Library/InJavascript/Intl.js
```diff
@@ -1593,44 +1593,122 @@
...
-                    return "Invalid Date";
-                }
-                var stateObject = setPrototype({}, null);
-                InitializeDateTimeFormat(stateObject, arguments[0], ToDateTimeOptions(arguments[1], "any", "all"));
-                return String(platform.formatDateTime(Internal.ToNumber(this), stateObject));
-            }), 0);
+            {
+                // scope for the cache variable
+                let __date_prototype_toLocaleString_default_formatterStateObject_cache = undefined;
+                platform.registerBuiltInFunction(tagPublicFunction("Date.prototype.toLocaleString", function () {
```
akroshg: curious - why not define locale and options as formals?

lib/Runtime/Library/InJavascript/Intl.js
dilijev: Defensive coding: ensure that we don't accidentally use the cache variable somewhere other than where it was intended.

lib/Runtime/Library/InJavascript/Intl.js
dilijev: The call to instantiate the formatter takes different parameters. There's no way to pass a pointer in for the cache so assigning it would not propagate outside of the call. I'll see if there's something I can do here.

lib/Runtime/Library/InJavascript/Intl.js
dilijev: The length of the function is required to be 0 by spec (and the length is equal to the number of formals), so we have to get arguments this way. This is done elsewhere. I'll add comments.

lib/Runtime/Library/InJavascript/Intl.js
dilijev: Yes, I meant to update these. Thanks.

dilijev:
@jackhorton 

> I mentioned this on my pull request as well, but perhaps the new caches should be tested by ChakraFull by manually toggling the useCaches switch?

Good point. I'll add that logic. I'm not sure about the testing strategy for this other than to ensure the behavior is consistent. I'll see what I can do in the way of tests.

(We really should add the Debug object functionality to core and migrate those tests.)

lib/Runtime/Library/InJavascript/Intl.js
dilijev: Spoke offline -- We can use default params to capture these values in a more canonical way and keep the function length as intended. I'll make this update.

lib/Runtime/Library/InJavascript/Intl.js
dilijev: I made a mistake in copy-paste. The parameters to the instantiation were supposed to be different (toLocaleString formats DateTime; toLocaleDateString, Date only; toLocaleTimeString, Time only).

Apparently this behavior was not covered by any test.

```
>eshost -h d8,ch-dev-che3,ch-master-latest -e "new Date().toLocaleString()"
#### d8
10/19/2017, 12:46:06 PM

#### ch-master-latest
‎10‎/‎19‎/‎2017‎ ‎12‎:‎46‎:‎06‎ ‎PM

#### ch-dev-che3
‎10‎/‎19‎/‎2017‎ ‎12‎:‎46‎:‎06‎ ‎PM

>eshost -h d8,ch-dev-che3,ch-master-latest -e "new Date().toLocaleTimeString()"
#### ch-master-latest
‎12‎:‎43‎:‎17‎ ‎PM

#### ch-dev-che3
‎10‎/‎19‎/‎2017‎ ‎12‎:‎43‎:‎17‎ ‎PM

#### d8
12:43:17 PM


>eshost -h d8,ch-dev-che3,ch-master-latest -e "new Date().toLocaleDateString()"
#### d8
10/19/2017

#### ch-dev-che3
‎10‎/‎19‎/‎2017‎ ‎12‎:‎43‎:‎38‎ ‎PM

#### ch-master-latest
‎10‎/‎19‎/‎2017
```

I will fix this and add tests to cover it.

lib/Runtime/Library/InJavascript/Intl.js
dilijev: @akroshg mentioned that this creates a capture scope which might have a minor memory/perf impact. No good reason to have that just for defensive reasons. My incoming refactoring covers the main concern of accidentally reusing a variable name.

lib/Runtime/Library/InJavascript/Intl.js
dilijev: I figured out a refactoring that I think works well here.

dilijev:
Thanks for your patience with my various updates. I've refactored a few things to get it to pass ChakraFull debugger tests. I've now rebased it to a few commits grouping the relevant changes into easier chunks (ended up being mostly in the first commit).

I'll add additional tests of Date.prototype.toLocale*String in a follow-up.

PR #4011, opened 2017-10-19
ARM64: fix double register handling for prolog/epilog

sigatrev:


sigatrev:
messed up history to include a different PR as well, closing to create a clean PR

PR #4013, opened 2017-10-19
ARM64: GenerateFloatTest

sigatrev:


lib/Backend/arm64/LowerMD.cpp
```diff
@@ -7973,33 +7973,16 @@ void LowererMD::GenerateFloatTest(IR::RegOpnd * opndSrc, IR::Instr * insertInstr
...
-
-    IR::RegOpnd *vt = IR::RegOpnd::New(TyMachPtr, this->m_func);
-    IR::Opnd* opnd = IR::IndirOpnd::New(opndSrc, (int32)0, TyMachPtr, this->m_func);
-    LowererMD::CreateAssign(vt, opnd, insertInstr);
-
-    // CMP [number], JavascriptNumber::vtable
-    IR::Instr* instr = IR::Instr::New(Js::OpCode::CMP, this->m_func);
-    instr->SetSrc1(vt);
-    instr->SetSrc2(m_lowerer->LoadVTableValueOpnd(insertInstr, VTableValue::VtableJavascriptNumber));
+    // TST s1, floatTagReg
```
agarwal-sandeep: This is doing a tagged float check and LoadFloatValue relies on this and loads value from offset which is correct?

sigatrev: This is doing a tagged float check, but LoadFloatValue is still ARM implementation, expected non-tagged floats, so that will have to be updated as well.

agarwal-sandeep: add a note there and you can commit this PR.

---
In reply to: [145843700](https://github.com/Microsoft/ChakraCore/pull/4013#discussion_r145843700) [](ancestors = 145843700)

PR #4016, opened 2017-10-20
Military time zone calculation is wrong

xiaoyinl:
According to source [1] and [2], the calculation of military time zone is wrong in ChakraCore. It should be the opposite: a corresponds to UTC+1, b corresponds to UTC+2, ... e.g. Eastern standard time is R, or UTC-5.

From [2]:
> The suffix indicates the correction (for description, see table) which must be applied to the time as expressed in order to convert to GMT.

Test cases are added.

[1] https://www.timeanddate.com/time/zones/military
[2] https://web.archive.org/web/20160304061301/http://jcs.dtic.mil/j6/cceb/acps/acp121/ACP121I.pdf, pp. 51-52


fatcerberus:
Hmm... if `r` means UTC-5, then you'd have to *add* 5 to the time-as-expressed to get UTC.  So unless I'm misreading something, the current code seems correct...?

xiaoyinl:
@fatcerberus Unless I miss something, "2011-11-08 19:48:43 r" should represent the same time as "2011-11-08T19:48:43-05:00", but this is what the current code gives:
    
    > Date.parse("2011-11-08 19:48:43 r") === Date.parse("2011-11-08T19:48:43+05:00")
    true
    > Date.parse("2011-11-08 19:48:43 r") === Date.parse("2011-11-08T19:48:43-05:00")
    false

fatcerberus:
> `lwTime -= lwZone * 60;`

That was the part I missed, thanks :)

dilijev:
Yes, we were incorrect. See our inconsistency with the other engines:

```
>eshost -itse "Date.parse('2011-11-08 19:48:43 r') === Date.parse('2011-11-08T19:48:43+05:00')"

## Source
print(Date.parse('2011-11-08 19:48:43 r') === Date.parse('2011-11-08T19:48:43+05:00'))

┌──────────────────┬───────┐
│ d8               │ false │
│ jsc              │       │
│ node             │       │
│ sm               │       │
├──────────────────┼───────┤
│ ch-1.7.1         │ true  │
│ ch-master-latest │       │
│ node-ch          │       │
└──────────────────┴───────┘
```

xiaoyinl:
@dilijev I'll definitely add some tests, but can I add the tests after #3988 is merged, if that'll happen soon? The reason is I want to add these tests to the file `test/Date/DateParse3.js` that I created for #3988.

dilijev:
@xiaoyinl I understand the desire to add more tests to the same file but it makes sense to split up unit tests between different files when they effectively cover different features. Having separate files also allows us to merge these PRs independently from each other.

dilijev:
This is correcting existing behavior, so I think we should merge it but military time zone doesn't seem to be in the spec and we are the only engine that implements it. @bterlson what do you think of this?

```
## Source
print(Date.parse('2011-11-08 19:48:43 y'))

┌──────────────────┬───────────────┐
│ d8               │ NaN           │
│ jsc              │               │
│ node             │               │
│ sm               │               │
├──────────────────┼───────────────┤
│ ch-1.7.1         │ 1320738523000 │
│ ch-master-latest │               │
│ node-ch          │               │
└──────────────────┴───────────────┘
```

dilijev:
Side note: interesting that we don't allow capital letters here. Might conflict with the logic we use to interpret Z for Zulu time or +00:00 which IS part of the spec.

xiaoyinl:
Another side note: we only support military time zone, but not military time, which has a different format than the normal one. It doesn't use colon separator, e.g. "1700R" represents "17:00 UTC-5", and "091630Z JUL 11" represents "16:30 UTC on July 9, 2011".

bterlson:
I think it's strange we support this, but we're a bit stuck now and I'm ok keeping it if it's not impacting anything else. If we want to support it this change seems correct. As long as this doesn't impact other "more standard" parse formats, seems pretty light weight. Every implementation has impl-specific parsing rules after all.

PR #4017, opened 2017-10-20
Fix clearing callinfo on inlined frames in a try

meg-gupta:
We inline functions into try, and clear the callinfo of all inlined frames until the enclosing try frame on a throw.
But if the inlined function calls into the interpreter and a throw happens, we do not want to clear the inlinee frame info of outer inlined frames.

In the failing test case, we inlined a function that called into an async function in the interpreter, we incorrectly cleared all the inlined frames until the enclosing try.
And on a following bailout from inside the inlined code in the try, we ended up with cleared callinfo while executing BailOutRecord::BailOutInlinedHelper.

Fixes OS#14226300

### DIFF
--- a/lib/Runtime/Language/JavascriptExceptionOperators.cpp
+++ b/lib/Runtime/Language/JavascriptExceptionOperators.cpp
@@ -95,16 +95,32 @@ namespace Js
         JavascriptExceptionObject *exception = nullptr;
 
         PROBE_STACK(scriptContext, Constants::MinStackDefault + spillSize + argsSize);
+        Js::JavascriptExceptionOperators::TryCatchFrameAddrStack tryCatchFrameAddrStack(scriptContext, frame);
 
         try
         {
             Js::JavascriptExceptionOperators::AutoCatchHandlerExists autoCatchHandlerExists(scriptContext);
-            Js::JavascriptExceptionOperators::TryCatchFrameAddrStack tryCatchFrameAddrStack(scriptContext, frame);
             continuation = amd64_CallWithFakeFrame(tryAddr, frame, spillSize, argsSize);
         }
         catch (const Js::JavascriptException& err)
         {
             exception = err.GetAndClear();
+
+            // We need to clear callinfo on inlinee virtual frames on an exception.
+            // We now allow inlining of functions into callers that have try-catch/try-finally.
+            // When there is an exception inside the inlinee with caller having a try-catch, clear the inlinee callinfo by walking the stack.
+            // If not, we might have the try-catch inside a loop, and when we execute the loop next time in the interpreter on BailOnException,
+            // we will see inlined frames as being present even though they are not, because we depend on FrameInfo's callinfo to tell if an inlinee is on the stack,
+            // and we haven't cleared those bits due to the exception
+            // When we start inlining functions with try, we have to track the try addresses of the inlined functions as well.
+
+#if ENABLE_NATIVE_CODEGEN
+            Assert(scriptContext->GetThreadContext()->GetTryCatchFrameAddr() != nullptr);
+            if (exception->GetExceptionContext() && exception->GetExceptionContext()->ThrowingFunction())
+            {
+                 WalkStackForCleaningUpInlineeInfo(scriptContext, nullptr /* start stackwalk from the current frame */);
+            }
+#endif
         }
 
         if (exception)
@@ -224,11 +240,11 @@ namespace Js
         JavascriptExceptionObject *exception = nullptr;
 
         PROBE_STACK(scriptContext, Constants::MinStackDefault + argsSize);
+        Js::JavascriptExceptionOperators::TryCatchFrameAddrStack tryCatchFrameAddrStack(scriptContext, framePtr);
 
         try
         {
             Js::JavascriptExceptionOperators::AutoCatchHandlerExists autoCatchHandlerExists(scriptContext);
-            Js::JavascriptExceptionOperators::TryCatchFrameAddrStack tryCatchFrameAddrStack(scriptContext, framePtr);
 #if defined(_M_ARM)
             continuation = arm_CallEhFrame(tryAddr, framePtr, localsPtr, argsSize);
 #elif defined(_M_ARM64)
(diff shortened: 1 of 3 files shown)


meg-gupta:
This fix does that stack walk in the catch handler of the JIT TryCatch helper as opposed to doing it in the interpreter. Because right now, we do not mark try frames
  of the interpreter on the stack of try-catches we maintain. Doing it in the catch handler we initialize the stack walker from the throwing function and clear
  inlinee frames till the most recent enclosing try jit frame.

lib/Runtime/Language/JavascriptExceptionOperators.cpp
```diff
@@ -95,16 +95,32 @@ namespace Js
         JavascriptExceptionObject *exception = nullptr;
 
         PROBE_STACK(scriptContext, Constants::MinStackDefault + spillSize + argsSize);
+        Js::JavascriptExceptionOperators::TryCatchFrameAddrStack tryCatchFrameAddrStack(scriptContext, frame);
```
meg-gupta: Moved this out of the try, because the destructor executes before the catch.

meg-gupta:
@rajatd The latest commit implements the design we discussed. I have kicked off all the tests as well.

lib/Runtime/Language/JavascriptExceptionOperators.cpp
```diff
@@ -95,16 +95,34 @@ namespace Js
...
+            // When there is an exception inside the inlinee with caller having a try-catch, clear the inlinee callinfo by walking the stack.
+            // If not, we might have the try-catch inside a loop, and when we execute the loop next time in the interpreter on BailOnException,
+            // we will see inlined frames as being present even though they are not, because we depend on FrameInfo's callinfo to tell if an inlinee is on the stack,
+            // and we haven't cleared those bits due to the exception
+            // When we start inlining functions with try, we have to track the try addresses of the inlined functions as well.
+
+#if ENABLE_NATIVE_CODEGEN
+            if (scriptContext->GetThreadContext()->GetTryCatchFrameAddr() != nullptr)
+            {
+                if (exception->GetExceptionContext() && exception->GetExceptionContext()->ThrowingFunction())
```
rajatd: do you still need this check?

rajatd:
>     void JavascriptExceptionOperators::WalkStackForCleaningUpInlineeInfo(ScriptContext *scriptContext, PVOID returnAddress)

Do you still need this parameter? I would leave a comment to add return address of throw to the exception context and pass that in here.

---
Refers to: lib/Runtime/Language/JavascriptExceptionOperators.cpp:927 in 8d7bf1c. [](commit_id = 8d7bf1c66fac92f56c45e6bea9b9f0763586db33, deletion_comment = False)

lib/Runtime/Language/JavascriptExceptionOperators.cpp
```diff
@@ -238,6 +256,14 @@ namespace Js
         catch (const Js::JavascriptException& err)
         {
             exception = err.GetAndClear();
+
+#if ENABLE_NATIVE_CODEGEN
+            if (scriptContext->GetThreadContext()->GetTryCatchFrameAddr() != nullptr)
+            {
+                Assert(exception->GetExceptionContext() && exception->GetExceptionContext()->ThrowingFunction());
+                WalkStackForCleaningUpInlineeInfo(scriptContext, exception->GetExceptionContext()->ThrowingFunction());
```
rajatd: >exception->GetExceptionContext()->ThrowingFunction( [](start = 65, length = 51)

Clean this up as this parameter is not really used and is the wrong thing to pass here anyway

lib/Runtime/Language/JavascriptExceptionOperators.cpp
```diff
@@ -440,6 +466,13 @@ namespace Js
         catch(const Js::JavascriptException& err)
         {
             pExceptionObject = err.GetAndClear();
+#if ENABLE_NATIVE_CODEGEN
+            if (scriptContext->GetThreadContext()->GetTryCatchFrameAddr() != nullptr)
+            {
+                Assert(pExceptionObject->GetExceptionContext() && pExceptionObject->GetExceptionContext()->ThrowingFunction());
+                WalkStackForCleaningUpInlineeInfo(scriptContext, pExceptionObject->GetExceptionContext()->ThrowingFunction());
```
rajatd: >pExceptionObject->GetExceptionContext()->ThrowingFunction() [](start = 65, length = 59)

Clean this up as this parameter is not really used and is the wrong thing to pass here anyway

lib/Runtime/Language/JavascriptExceptionOperators.cpp
LouisLaf: Don't you want the dtor to run before the user catch code however?  I think you want to use {} to make that happen.

---
In reply to: [146085699](https://github.com/Microsoft/ChakraCore/pull/4017#discussion_r146085699) [](ancestors = 146085699)

lib/Runtime/Language/JavascriptExceptionOperators.cpp
```diff
@@ -95,16 +95,32 @@ namespace Js
...
+            // We now allow inlining of functions into callers that have try-catch/try-finally.
+            // When there is an exception inside the inlinee with caller having a try-catch, clear the inlinee callinfo by walking the stack.
+            // If not, we might have the try-catch inside a loop, and when we execute the loop next time in the interpreter on BailOnException,
+            // we will see inlined frames as being present even though they are not, because we depend on FrameInfo's callinfo to tell if an inlinee is on the stack,
+            // and we haven't cleared those bits due to the exception
+            // When we start inlining functions with try, we have to track the try addresses of the inlined functions as well.
+
+#if ENABLE_NATIVE_CODEGEN
+            Assert(scriptContext->GetThreadContext()->GetTryCatchFrameAddr() != nullptr);
+            if (exception->GetExceptionContext() && exception->GetExceptionContext()->ThrowingFunction())
```
LouisLaf: I'm curious as to why ThrowingFunction would ever be null?

meg-gupta: For OOMs the exception object is not well formed. It could be null.

lib/Runtime/Language/JavascriptExceptionOperators.cpp
meg-gupta: We want the dtor to run after the catch. We don't want to pop off the top most try catch addr.

LouisLaf: I don't think it is right, but doesn't matter until we start inlining functions with try/catches.  When running the user catch, the try isn't "active" anymore.  If you get an exception during the user catch, you need to unwind the inlined frames until the **next** try block.  But for now, if the try block is in the same function, we can't have inlined frames in between the catch and that try block...
We should still do it the right way in case we do enable inlining of functions with EH.

PR #4018, opened 2017-10-20
fix modFunctionIndex is missing on copy

obastemur:
OS14289876 credit goes to @jianchun

### DIFF
--- a/lib/Common/DataStructures/BaseDictionary.h
+++ b/lib/Common/DataStructures/BaseDictionary.h
@@ -143,7 +143,8 @@ namespace JsUtil
             }
         }
 
-        BaseDictionary(const BaseDictionary &other) : alloc(other.alloc)
+        BaseDictionary(const BaseDictionary &other) :
+          alloc(other.alloc)
         {
             if(other.Count() == 0)
             {
@@ -153,6 +154,7 @@ namespace JsUtil
                 entries = nullptr;
                 count = 0;
                 freeCount = 0;
+                modFunctionIndex = UNKNOWN_MOD_INDEX;
 
 #if PROFILE_DICTIONARY
                 stats = nullptr;
@@ -182,6 +184,7 @@ namespace JsUtil
             count = other.count;
             freeList = other.freeList;
             freeCount = other.freeCount;
+            modFunctionIndex = other.modFunctionIndex;
 
             CopyArray(buckets, bucketCount, other.buckets, bucketCount);
             CopyArray<EntryType, Field(ValueType, TAllocator), TAllocator>(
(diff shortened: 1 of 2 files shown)


jianchun:
>                 bucketCount = 0;

If `other.Count() == 0` (because `count == freeCount` but `bucketCount != 0`), this could result in bad modFunctionIndex.

---
Refers to: lib/Common/DataStructures/BaseDictionary.h:153 in 1339070. [](commit_id = 1339070d172f6dc7da9aa3e1d5a2a21a6bc43e90, deletion_comment = False)

jianchun:
>             this->freeCount = 0;

Here modFunctionIndex missing update.

---
Refers to: lib/Common/DataStructures/BaseDictionary.h:302 in 1339070. [](commit_id = 1339070d172f6dc7da9aa3e1d5a2a21a6bc43e90, deletion_comment = False)

jianchun:
>                 bucketCount = 0;

Missing modFunctionIndex update.

---
Refers to: lib/Common/DataStructures/BaseDictionary.h:311 in 1339070. [](commit_id = 1339070d172f6dc7da9aa3e1d5a2a21a6bc43e90, deletion_comment = False)

jianchun:
>             uint newBucketCount = SizePolicy::GetBucketSize(newSize, &modFunctionIndex);

In theory this is not safe. `this->modFunctionIndex` is changed while other variables are not (`newSize`, `newBucketCount` are using temporaries). If anything below failed in the middle we'd end up with bad `modFunctionIndex`. To be consistent and safer, consider also use a temporary "newModFunctionIndex" and be sure use it during rehash.

---
Refers to: lib/Common/DataStructures/BaseDictionary.h:1021 in 1339070. [](commit_id = 1339070d172f6dc7da9aa3e1d5a2a21a6bc43e90, deletion_comment = False)

jianchun:
Please also fix `WeaklyReferencedKeyDictionary::Resize() and Initialize()`, both have the same unsafe pattern of changing `modFunctionIndex` before everything succeeds.

obastemur:
@jianchun please consider that the mod index set by previous dictionary was working for that dictionary.

jianchun:
> @jianchun please consider that the mod index set by previous dictionary was working for that dictionary.

Yes it is a bug. But that's not the only problem. Several from my comments are apparent bugs too. The observation is that we have similar amount of failures in `WeaklyReferencedKeyDictionary`, which was not covered by current change. Why not do a thorough review and fix?

obastemur:
Well, it wasn’t a through review but your comments on vso with a credit on description.

obastemur:
@jianchun Thank you for the review and pointers. Is it good to go?

PR #4021, opened 2017-10-20
ARM64: Enable moving ArgOuts next to call for arm64

agarwal-sandeep:


agarwal-sandeep:
@sigatrev please take second look

lib/Common/Common/SmartFpuControl.cpp
```diff
@@ -16,7 +16,7 @@
 static errno_t SetFPUControlDefault(void)
 {
 #if _WIN32
-#if _M_AMD64 || _M_ARM
+#if _M_AMD64 || _M_ARM || _M_ARM64
     return _controlfp_s(0, _RC_NEAR + _DN_SAVE + _EM_INVALID + _EM_ZERODIVIDE +
```
aaronsgiles: Why did you add || _M_ARM64 here instead of converting to !defined(_M_IX86) like above?

PR #4023, opened 2017-10-20
ARM64: legalize VFP indir

sigatrev:


lib/Backend/arm64/LegalizeMD.cpp
```diff
@@ -316,6 +316,13 @@ IR::Instr * LegalizeMD::LegalizeLoad(IR::Instr *instr, uint opndNum, LegalForms
...
 void LegalizeMD::LegalizeIndirOffset(IR::Instr * instr, IR::IndirOpnd * indirOpnd, LegalForms forms, bool fPostRegAlloc)
 {
+    if (indirOpnd->IsFloat())
+    {
+        // Vfp doesn't support register indirect operation
+        LegalizeMD::LegalizeIndirOpndForVFP(instr, indirOpnd, fPostRegAlloc);
+        return;
+    }
+
     int32 offset = indirOpnd->GetOffset();
```
aaronsgiles: LegalizeIndirForVFP has not been updated for ARM64. I actually think the indir logic is the same for both FP and integer except for the no register variant, so I'd just add a check here for index and if it's present call HoistIndirIndexOpndAsAdd on it.

sigatrev:
@aaronsgiles Updated. Want to take another look?

agarwal-sandeep:
Looks ok to me.

PR #4024, opened 2017-10-20
Disable RTTI on x-plat

jackhorton:
My RTTI change "just worked" on Linux, which seemed wrong. Turns out, we never disabled RTTI to begin with. This change brings a release build of libChakraCoreStatic.a down from 484MB to 463MB.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -346,8 +346,13 @@ if(CLR_CMAKE_PLATFORM_XPLAT)
     add_compile_options(
         -fasm-blocks
         -fms-extensions
-        -fwrapv                 # Treat signed integer overflow as two's complement
+        -fwrapv # Treat signed integer overflow as two's complement
     )
+
+    # Only disable RTTI in release builds so that TrackAlloc works for debug and test builds
+    if(CMAKE_BUILD_TYPE STREQUAL Release)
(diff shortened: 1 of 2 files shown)


jackhorton:
That build break doesn't happen when i build locally in Linux... Looking into it now.

jackhorton:
Update: apparently on Linux and macOS, debug and test builds require RTTI to be enabled throughout the codebase (anywhere that uses Allocator.h) because of a call to `typeid()` in the allocation tracking code. MSVC seems to optimize this out because our calls to typeid are always on static types rather than runtime objects, but clang doesn't seem to perform this optimization (for better or worse).

CMakeLists.txt
```diff
@@ -346,9 +346,14 @@ if(CLR_CMAKE_PLATFORM_XPLAT)
     add_compile_options(
         -fasm-blocks
         -fms-extensions
-        -fwrapv                 # Treat signed integer overflow as two's complement
+        -fwrapv # Treat signed integer overflow as two's complement
     )
 
+    # Only disable RTTI in release builds to that TrackAlloc works for debug and test builds
```
digitalinfinity: s/to/so?

jackhorton: Yup, typo. Will fix in a second.

PR #4033, opened 2017-10-21
Make sure we correctly track if a function has calls even in inlinee

Cellule:
This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/4033)


Cellule:
Unfortunately, my strategy to track calls wasn't fool proof as other phases could insert calls after reg alloc.
I decided to fall back to our good old ways and just fix the issues that we currently had.
I wish I could find a clean way to enforce setting that bit whenever a call is added

lib/Backend/LowerMDShared.cpp
```diff
@@ -7722,6 +7723,7 @@ LowererMD::EmitInt64toFloat(IR::Opnd *dst, IR::Opnd *src, IR::Instr *instr)
     LoadInt64HelperArgument(instr, srcOpnd);
 
     IR::Instr* callinstr = IR::Instr::New(Js::OpCode::CALL, dst, this->m_func);
+    this->m_func->SetHasCallsOnSelfAndParents();
```
rajatd: you probably dont need this as we call ChangeToHelperCall on callInstr below

MikeHolman:
Out of curiousity, where can we insert calls after reg alloc? I'd think just the stack probe, but that probably shouldn't count?

Cellule:
I think in RegAlloc there are 1 or 2 cases where we can add Calls and my initial checks seems to have been too late somehow

Issue #4008, opened 2017-10-19
perf: new Date(STRING) is slow


obastemur:
`new Date(STRING)` is a hot call on acme-air.

Micro-bench

```
var arr_dates = [];

var days = ['Sun', 'Mon', 'Tue', 'Wed', 'Sat'];
var mons = ['Oct', 'Nov', 'Sep', 'Jun', 'Jan'];

for (var i = 1980; i < 2050; i++)
{
	for(var mday = 1; mday < 29; mday++)
	{
		for(var j = 0; j < days.length; j++)
		{
			for(var k = 0; k < mons.length; k++)
			{
				arr_dates.push(days[j] + " " + mons[k] + " " + mday + " 00:00:00 GMT " + i);
			}
		}
	}
}

// 50K+ dates.. we won't be measuring the cached results

var start = Date.now();
var len = arr_dates.length;
var tt = 0;
for(var i = 0; i < len; i++)
{
	tt += new Date(arr_dates[i]).getMonth();
	tt = tt & 63; // MOD 64
}
var end = Date.now() - start;

console.log("Total: ", end, tt) // tt should be 0!
```

CC is 6x slower than v8 on above micro bench.

akroshg:
When I ran the test with the jshost on the master I get 
```Total:  45 18```
This seems quick enough, no?

rhuanjl:
@obastemur thought I'd try and be helpful

Running with ch on master, release build and lto I get "Total: 380 0"

I then tried running it with instruments' time profiler and found that 348ms of that was spent in this call to _wcslwr:
https://github.com/Microsoft/ChakraCore/blob/2e15b3995061d9dfaf16e0951407c3fbf4139ace/lib/Runtime/Library/DateImplementation.cpp#L1076

It seems to make it faster the key thing to do would be try and make it handle uppercase as well as lowercase.

I'd be happy to have a go at this if you've not already started it (that is if it's ok for an external contributor to look at this.)

obastemur:
@rhuanjl go ahead if you want!

@akroshg Thanks for looking at Windows numbers. Looks like the origin of this issue is PAL `towlower`

rhuanjl:
I’ve had a quick go hacking around with Pal and got the time down from 380ms to 27 ms.
Hi 
I’ve noticed a few optimisation opportunities in pal, will aim to submit a PR over the weekend if it’s ok to edit pal?

rhuanjl:
[...]

PR #4035, opened 2017-10-21
Optimise xplat UtcTimeFromStrCore

rhuanjl:
**edit:** This description (other than this first paragraph) is out of date - see comments below, this is now an update to PAL_towlower to give it a fast path for ascii characters. Key outcome is the ~70% performance gain for DateImplementation::UtcTimeFromStrCore on xplat.

Added a fast inline conversion from upper case to lowercase for xplat in DateImplementation::UtcTimeFromStrCore

Note this method will not work for non-latin alphabet - but the function only uses the latin alphabet and this is ~100x faster than the PAL _wcslwr_s.

SIDENOTE: I did not revise _wcslwr_s as it appears to be used nowhere else in the source tree - I note that if it is important for any other applications in any performance critical path it should be revised, it uses a number of malloc operations equal to the length of the string + 1.

Resolves https://github.com/Microsoft/ChakraCore/issues/4008

### DIFF
--- a/pal/src/cruntime/wchar.cpp
+++ b/pal/src/cruntime/wchar.cpp
@@ -885,51 +885,67 @@ char16_t
 __cdecl
 PAL_towlower( char16_t c )
 {
-#if HAVE_COREFOUNDATION
     PERF_ENTRY(towlower);
     ENTRY("towlower (c=%d)\n", c);
-    if (!PAL_iswlower(c))
-    {
-        CFMutableStringRef cfString = CFStringCreateMutable(
-                                            kCFAllocatorDefault, 1);
-        if (cfString != NULL)
+    if(c < 128)
+    {//fast path for ascii characters
+        if(c >= 'A' && c <= 'Z')
         {
-            CFStringAppendCharacters(cfString, (const UniChar*)&c, 1);
-            CFStringLowercase(cfString, NULL);
-            c = CFStringGetCharacterAtIndex(cfString, 0);
-            CFRelease(cfString);
+            c += ('a' - 'A');
+            PERF_EXIT(towlower);
+            LOGEXIT("towlower returns int %d\n", c );
+            return c;
         }
-    }
-    LOGEXIT("towlower returns int %d\n", c );
-    PERF_EXIT(towlower);
-    return c;
-#else   /* HAVE_COREFOUNDATION */
-    UnicodeDataRec dataRec;
-
-    PERF_ENTRY(towlower);
-    ENTRY("towlower (c=%d)\n", c);
-
-    if (!GetUnicodeData(c, &dataRec))
-    {
-        TRACE( "Unable to retrieve unicode data for the character %c.\n", c );
-        LOGEXIT("towlower returns int %d\n", c );
-        PERF_EXIT(towlower);
-        return c;
-    }
-
-    if ( (dataRec.C1_TYPE_FLAGS & C1_LOWER) || (dataRec.nOpposingCase ==  0 ))
-    {
-        LOGEXIT("towlower returns int %d\n", c );
-        PERF_EXIT(towlower);
-        return c;
+        else
+        {
+            PERF_EXIT(towlower);
+            LOGEXIT("towlower returns int %d\n", c );
+            return c;
+        }
     }
     else
     {
-        LOGEXIT("towlower returns int %d\n", dataRec.nOpposingCase );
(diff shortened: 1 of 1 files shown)


rhuanjl:
@obastemur Not the prettiest but it does the job.

With the micro benchmark from https://github.com/Microsoft/ChakraCore/issues/4008 without this fix average time was around 380ms - with this fix 48ms. (testing on a 2.5ghz macbook pro intel i7 running macos Sierra)

lib/Runtime/Library/DateImplementation.cpp
```diff
@@ -1073,7 +1073,22 @@ namespace Js {
...
         js_memcpy_s(pszSrc, size, psz, size);
 
+        #ifdef _WIN32
         _wcslwr_s(pszSrc,ulength+1);
+        #else
+        //fast xplat conversion from upper case to lower case
+        //only targets the UTF16 Latin alphabet - but that's all we need here
+        //used instead of the PAL _wcslwr_s which takes ~100x as long to run
+        int i;
+        for (i = 0 ; pszSrc[i] != 0; i++)
```
Penguinwizzard: We actually have a length here, I'd prefer it were used rather than the null terminator.

lib/Runtime/Library/DateImplementation.cpp
```diff
@@ -1073,7 +1073,22 @@ namespace Js {
...
+        {
+            if(pszSrc[i] > 64 && pszSrc[i] < 91)
```
Penguinwizzard: Can you change this to >= 'A' and <= 'Z', as that's a little clearer?

lib/Runtime/Library/DateImplementation.cpp
```diff
@@ -1073,7 +1073,22 @@ namespace Js {
...
+            {
+                pszSrc[i] += 32;
```
Penguinwizzard: += ('a' - 'A')

lib/Runtime/Library/DateImplementation.cpp
```diff
@@ -1073,7 +1073,22 @@ namespace Js {
         size_t size = sizeof(char16) * (ulength + 1);
         js_memcpy_s(pszSrc, size, psz, size);
 
+        #ifdef _WIN32
```
Penguinwizzard: Is this necessarily platform-dependent?

obastemur:
@rhuanjl Thanks for the effort! In order this to be merged;

1- Please make the changes under PAL wcslwr...
2- your changes apply only when the char16 data is ASCII. Otherwise we use the existing path.

obastemur:
FYI; `PAL_towlower` is the exact place for these changes.

lib/Runtime/Library/DateImplementation.cpp
rhuanjl: Native windows _wcslwr_s is already fast - whilst this alternate path would work on windows it wouldn't be an optimisation there.

I'll look at putting it into PAL wcslwr as suggested by obastemur below - just will have to add some extra checks to only put it in when wanted.

rhuanjl:
@Penguinwizzard I've revised in line with obastemur's instruction to do the change within PAL_towlower in wchar.cpp.

Couple of things I'm unsure on:
1. I note - your point on doing a loop using the length; following my revision this is now using the existing loop in the function _wcslwr which doesn't receive the length parameter - I could revise to take the loop to the the function _wcslwr_s which does have the length parameter (though would require moving it to wchar.cpp or making wtolower visible in palrt.h)

2. the wrapper of _wcslwr_s (within palrt.h) I can't work out what this is doing of any use - it copies the full buffer, sends the copy to be edited by _wcslwr then copies the result back - the overhead of these two copies and the associated malloc and free add a detectable amount of time at execution and I can't work out the benefit - if it's meant to be to avoid a buffer overflow the key check I'd want would be for the existence of a null terminator (or just iterate vs the length instead) I can't see how copying it twice helps - left in for now as I assume there is a purpose to the action that I'm just not aware of.

Penguinwizzard:
You're fine leaving the code where it is for now; we will most likely want to revise _wsclwr_s separately based on 1. and 2.

Penguinwizzard:
@MSLaguana: Can you pick this up for node-cc perf testing and check-in?

MSLaguana:
Change looks good to me and passes tests; thanks for your contribution @rhuanjl !

PR #4040, opened 2017-10-23
FromVar hardening

rajatd:
This PR hardens the FromVar method on Chakra's runtime classes to fail fast if the var passed in is not of the expected type. Although, there are a lot of instances in the code where we call FromVar only after ensuring that the var has the expected type. Those cases will now have to do a redundant check in FromVar and that has perf impact. To counter such perf loss, I have introduced two new methods:

1. `UnsafeFromVar` - it is identical to FromVar that we have today (checks the type, but only in debug builds). Used in cases where the type check is done by a switch on typeId.

2. `JavascriptOperators::TryFromVar<T>` - Performs both the type check and pointer cast (as dictated by the template parameter) for the passed in var. Used in cases where T::Is() is immediately followed by a T::FromVar.

### DIFF
--- a/lib/Runtime/Library/ArrayBuffer.cpp
+++ b/lib/Runtime/Library/ArrayBuffer.cpp
@@ -13,6 +13,12 @@ namespace Js
 
     ArrayBufferBase* ArrayBufferBase::FromVar(Var value)
     {
+        AssertOrFailFast(ArrayBufferBase::Is(value));
+        return static_cast<ArrayBuffer *> (value);
+    }
+
+    ArrayBufferBase* ArrayBufferBase::UnsafeFromVar(Var value)
+    {
         Assert(ArrayBufferBase::Is(value));
         return static_cast<ArrayBuffer *> (value);
     }
@@ -60,7 +66,7 @@ namespace Js
                         VirtualTableInfo<CrossSiteObject<Int8Array>>::SetVirtualTable(parent);
                     }
                 }
-                TypedArrayBase::FromVar(parent)->ClearLengthAndBufferOnDetach();
+                TypedArrayBase::UnsafeFromVar(parent)->ClearLengthAndBufferOnDetach();
                 break;
 
         case TypeIds_Uint8Array:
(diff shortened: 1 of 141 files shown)


lib/Runtime/Library/ArrayBuffer.cpp
```diff
@@ -568,11 +574,18 @@ namespace Js
 
     ArrayBuffer* ArrayBuffer::FromVar(Var aValue)
     {
-        AssertMsg(Is(aValue), "var must be an ArrayBuffer");
+        AssertOrFailFastMsg(Is(aValue), "var must be an ArrayBuffer");
 
         return static_cast<ArrayBuffer *>(RecyclableObject::FromVar(aValue));
```
MikeHolman: This can be UnsafeFromVar, since there is already failfast right above

MikeHolman: Looks like a number of the FromVars have this. But seems like in some they just directly static_cast a var to the type. If that's possible, it seems preferable here

---
In reply to: [147226615](https://github.com/Microsoft/ChakraCore/pull/4040#discussion_r147226615) [](ancestors = 147226615)

rajatd: Makes sense. I'll try to make it uniform across FromVars, either with or without RecyclableObject::UnsafeFromVar

lib/Runtime/Math/JavascriptMath.cpp
```diff
@@ -901,9 +901,9 @@ namespace Js
             ScriptContext * scriptContext = function->GetScriptContext();
 
             TypeId typeId = JavascriptOperators::GetTypeId(arrayArg);
-            if (!JavascriptNativeArray::Is(typeId) && !(TypedArrayBase::Is(typeId) && typeId != TypeIds_CharArray && typeId != TypeIds_BoolArray))
```
MikeHolman: is this change in semantics? or were the extra conditions just redundant?

rajatd: TypedArrayBase::Is does a range check for typeId and TypeIds_CharArray and TypeIds_BoolArray fall in that range. Presumably, they didn't fall in the range when this code was written.

MSLaguana: If `TypedArrayBase::Is(typeId)` returns true for `TypeIds_CharArray`, then this is a change in semantics: the logic before (pushing in the negation) was
```cpp
!JavascriptNativeArray::Is(typeId) &&
  (!TypedArrayBase::Is(typeId) || typeId == TypeIds_CharArray || typeId == TypeIds_BoolArray)
```

which means that if it was a `CharArray` or `BoolArray` (and not a native array) the condition would succeed, while now it would fail.

rajatd: That is true; my bad. Will revert this. Thanks for catching this.

PR #4041, opened 2017-10-23
Adding readline support

JohnathanNorman:
V8 and Mozilla's javascript engines both support readline which is useful in testing. This pull request adds readline support.


dilijev:
@JohnathanNorman please see https://github.com/Microsoft/ChakraCore/wiki/Jenkins-CI-Checks#faq about fixing the failed style checks (EOL Check and Tab Check)

bin/ch/WScriptJsrt.cpp
```diff
@@ -1073,6 +1074,115 @@ JsValueRef __stdcall WScriptJsrt::LoadTextFileCallback(JsValueRef callee, bool i
...
+		goto Error;
+
+	int gotlength;
+	while ((gotlength = js_fgets(buf + buflength, bufsize - buflength, from)) > 0) {
+		buflength += gotlength;
+
+		/* Are we done? */
+		if (buf[buflength - 1] == '\n') {
+			buf[buflength - 1] = '\0';
+			sawNewline = true;
```
Penguinwizzard: Why not do the buflength -1 here? Unless I'm reading it incorrectly, that's effectively the purpose of this flag.

bin/ch/WScriptJsrt.cpp
```diff
@@ -1073,6 +1074,115 @@ JsValueRef __stdcall WScriptJsrt::LoadTextFileCallback(JsValueRef callee, bool i
     return returnValue;
 }
 
+int js_fgets(char* buf, int size, FILE* file)
+{
+	int  i, n, c;
+	bool crflag;
```
Penguinwizzard: Can you initialize these or move the declaration to the point of initialization? This function doesn't have gotos, so there's not as much the desire to avoid jumping past declarations that there is with other code.

JohnathanNorman: good idea.

bin/ch/WScriptJsrt.cpp
JohnathanNorman: I guess i don't understand the question here. Can you maybe show me what you are suggesting?

JohnathanNorman:
killing request.. will redo after i address EOL/TAB issues later.

bin/ch/WScriptJsrt.cpp
Penguinwizzard: As per how I understand the function, the only use of sawNewline is to keep track that we have an extra null character at the end of the string, and that buflength represents the string length -1. This means that a line with just "\n" won't go through the buflength == 0 path, despite being 0x0000 0x0000 in the binary representation, with a wstrlen of 0.

Issue #4038, opened 2017-10-23
ES6 modules not working with try/catch in module


herrernst:
Hi,
I've found a case where ES6 modules are not working for me in current MS Edge (Creators Update 41.16299.15.0), that is when there is a try/catch in an imported module. It's also reproducable in ChakraCore.

Here is a hosted reduced test case:
https://rawgit.com/herrernst/es6-module-try-catch/master/index.html
It works in current Chrome, Safari, and Firefox (with dom.moduleScripts.enabled), but not in MS Edge (nothing happens). If you [remove the try/catch](https://github.com/herrernst/es6-module-try-catch/blob/master/utils.js#L4) it works.
When running in Chakra (tested with 1.7.3 OSX binary release from GitHub) via `ch ./chakra-wrap.js`, this is printed:
```
thrown at main.js:
^
SyntaxError: Module import or export statement unexpected here
```

akroshg:
@herrernst thanks for reporting this.
I can repro this. @boingoing  do you happen to know why this not working?
although if I move the try/catch to a function it works.

boingoing:
I don't see any obvious issues; seems like that should work.

akroshg:
ok - found the problem. I'll fix and send PR.

rhuanjl:
Also noticed it works if you move the try/catch down to the bottom of the file - specifically have it after the exports rather than before - possible work around until the fix flows through to release.

PR #4046, opened 2017-10-23
Try-catch on top level module source was throwing syntax error.

akroshg:
The bookkeeping went wrong. Fixed a field when we just have try/catch.


akroshg:
fixes #4038 . @boingoing  can you take a look?

PR #4047, opened 2017-10-23
ARM64: add MOV_TRUNC opcode

sigatrev:
MOV_TRUNC is an alias of MOV that won't get removed by peeps when dst == src. Used for 64 bit to 32 bit conversions


sigatrev:
@aaronsgiles I don't see any MOV_TRUNCs in the AMD64 GenerateFast* functions

PR #4049, opened 2017-10-24
Use correct OpndType when doing charLength compare and Legalize TST opcode

agarwal-sandeep:


lib/Backend/arm64/LowerMD.cpp
```diff
@@ -3800,6 +3800,7 @@ LowererMD::GenerateTaggedZeroTest( IR::Opnd * opndSrc, IR::Instr * insertInstr,
         instr->SetSrc1(opndSrc);
         instr->SetSrc2(opndSrc);
         insertInstr->InsertBefore(instr);
+        LegalizeMD::LegalizeInstr(instr, false);
     }
```
aaronsgiles: Isn't there a this->Legalize(instr) you can use?

rajatd: +1

agarwal-sandeep: This file have mixed usage of both (way too many) keeping it LegalizeMD::LegalizeInstr for now. Will have another PR in which I will remove most of static usage.

PR #4050, opened 2017-10-24
Fix arm debug assert allowing base offset to be R7 when functions have try

meg-gupta:
We always used R7 as base register in arm when there is a try. We now have inlined functions within try.
Fix assert in Encoder which should now take care of case where there are inlined functions in try.

### DIFF
--- a/lib/Backend/arm/EncoderMD.cpp
+++ b/lib/Backend/arm/EncoderMD.cpp
@@ -2241,7 +2241,7 @@ EncoderMD::BaseAndOffsetFromSym(IR::SymOpnd *symOpnd, RegNum *pBaseReg, int32 *p
         if (func->GetMaxInlineeArgOutSize() != 0)
         {
             Assert(func->HasInlinee());
-            Assert(baseReg == RegSP);
+            Assert(baseReg == (func->HasTry() ?  RegR7 : RegSP));
             if (stackSym->IsArgSlotSym() && !stackSym->m_isOrphanedArg)
             {
                 Assert(stackSym->m_isInlinedArgSlot);
@@ -2250,7 +2250,7 @@ EncoderMD::BaseAndOffsetFromSym(IR::SymOpnd *symOpnd, RegNum *pBaseReg, int32 *p
             else
             {
                 AssertMsg(stackSym->IsAllocated(), "StackSym offset should be set");
-                Assert((uint)offset > func->m_argSlotsForFunctionsCalled * MachRegInt + func->GetMaxInlineeArgOutSize());
+                Assert(offset > (func->HasTry() ? (int32)func->GetMaxInlineeArgOutSize() : (int32)(func->m_argSlotsForFunctionsCalled * MachRegInt + func->GetMaxInlineeArgOutSize())));
             }
         }
         // TODO: restore the following assert (very useful) once we have a way to tell whether prolog/epilog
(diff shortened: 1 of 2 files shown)


lib/Backend/arm/EncoderMD.cpp
```diff
@@ -2241,7 +2241,7 @@ EncoderMD::BaseAndOffsetFromSym(IR::SymOpnd *symOpnd, RegNum *pBaseReg, int32 *p
         if (func->GetMaxInlineeArgOutSize() != 0)
         {
             Assert(func->HasInlinee());
-            Assert(baseReg == RegSP);
+            Assert(baseReg == RegSP || (func->HasTry() && baseReg == RegR7));
```
MSLaguana: If the function has a try, but the `baseReg == RegSP` is that ok, or is that also an error? e.g. should the `baseReg` be `RegR7` if and only if the function has a try?

meg-gupta: The baseReg is always R7 when there is a try.

MSLaguana: In that case, I suggest something like `Assert(baseReg == (func->HasTry() ? RegR7 : RegSP))` to enforce that

lib/Backend/arm64/EncoderMD.cpp
```diff
@@ -1405,7 +1405,7 @@ EncoderMD::BaseAndOffsetFromSym(IR::SymOpnd *symOpnd, RegNum *pBaseReg, int32 *p
 
         if (func->HasInlinee())
         {
-            Assert(baseReg == RegSP);
+            Assert(baseReg == RegSP || (func->HasTry() && baseReg == RegR7));
```
agarwal-sandeep: >RegR7 [](start = 69, length = 5)

R0 to R7 are parameter register in ARM64

meg-gupta: Added TODO.

lib/Backend/arm/EncoderMD.cpp
```diff
@@ -2241,7 +2241,7 @@ EncoderMD::BaseAndOffsetFromSym(IR::SymOpnd *symOpnd, RegNum *pBaseReg, int32 *p
         if (func->GetMaxInlineeArgOutSize() != 0)
         {
             Assert(func->HasInlinee());
-            Assert(baseReg == RegSP);
+            Assert(baseReg == func->HasTry() ?  RegR7 : RegSP);
```
MSLaguana: I believe you need the parentheses here, otherwise this is interpreted as `(baseReg == func->HasTry()) ? RegR7 : RegSP`

lib/Backend/arm/EncoderMD.cpp
```diff
@@ -2250,7 +2250,7 @@ EncoderMD::BaseAndOffsetFromSym(IR::SymOpnd *symOpnd, RegNum *pBaseReg, int32 *p
             else
             {
                 AssertMsg(stackSym->IsAllocated(), "StackSym offset should be set");
-                Assert((uint)offset > func->m_argSlotsForFunctionsCalled * MachRegInt + func->GetMaxInlineeArgOutSize());
+                Assert(offset > func->HasTry() ? (int32)func->GetMaxInlineeArgOutSize() : func->m_argSlotsForFunctionsCalled * MachRegInt + func->GetMaxInlineeArgOutSize());
```
MSLaguana: Same here, the precedence of the ternary operator is low, so this doesn't have the intended meaning.

lib/Backend/arm/EncoderMD.cpp
meg-gupta: ah, yes! fixed.

PR #4051, opened 2017-10-24
   Clear inlinee callinfo from the catch of OP_TryFinally as well

meg-gupta:
When we hit an exception and we have a try finally within a try catch. We will execute finally code by calling BailOnException.
   Since we inline into functions with try now, we will end up in inconsistent inlinee callinfo when we call the bailout code from finally.
   So add the walker and clear inlinee callinfo. Also set a bit that the callinfo's were cleared so that OP_TryCatch does not walk again to do the same.

Fixes OS# 14336922


LouisLaf:
This unfortunately does not work with nested exception:
try { 
    try { 
        inlinee1(); 
    } finally { 
        try { 
            inlinee2(); 
        } catch() {
        }
    }
} catch() {
}

The catch in the finally will reset the threadcontext...

You can fix this by having your constructor save the previous state on the threadcontext, which gives you a stack.

Of course, this still won't be sufficient for inlining functions with try blocks...  If we ever do our own EH frames, we could treat inlinees as finallys and unwind them then.

meg-gupta:
Will close this and reopen against release/1.8

PR #4053, opened 2017-10-24
Remove redundant class name in function declaration (Fix #4052)

### LINKED ISSUE #4052: Issues found when building Chakra with MSVC + /permissive-
Building Chakra with MSVC + /permissive- failed with error C4596, the issue is in https://github.com/Microsoft/ChakraCore/blob/master/lib/Backend/Opnd.h#L1502, and involved by commit https://github.com/Microsoft/ChakraCore/commit/fa9151ead415e84ec0c07f9acf87db2f36826a9b 

Using full-qualified name to declare members inside class is ill-formed and incorrectly allowed by MSVC. MSVC now gives error when /permissive- is used. The fix is remove ‘ListOpnd::’

Failure like:
d:\chakra\src\lib\backend\Opnd.h(1502): error C4596: 'New': illegal qualified name in member declaration

xiaoyinl:
This PR removes "ListOpnd::" from the member function declaration in Opnd.h.

Fixes #4052

### DIFF
--- a/lib/Backend/Opnd.h
+++ b/lib/Backend/Opnd.h
@@ -1499,7 +1499,7 @@ class ListOpnd : public Opnd
     };
 public:
     ~ListOpnd();
-    static ListOpnd* ListOpnd::New(Func *func, __in_ecount(count) ListOpndType** opnds, DECLSPEC_GUARD_OVERFLOW int count);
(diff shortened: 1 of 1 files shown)


dilijev:
I'm testing this internally and will then merge.

PR #4054, opened 2017-10-24
Updating Stats:Recycler output

MSLaguana:
Stats:Recycler functionality was no longer working, with most of the
counters not updated correctly. This fixes up most of the mark tracking
section, so each TryMark should correspond to a mark or an attributed
non-mark. I believe that the SweptBlk section is also correct without
any changes in my part.

Reported memory is still not fully correct; the Unused column will
be larger than it should unless a partial GC completes and updates
the smallNonLeafHeapBlockPartialReuseBytes counters properly, but
I believe that the Total and Live numbers are representative.


dilijev:
@MSLaguana are you still planning to merge this?

PR #4062, opened 2017-10-25
treat two digit years less than 50 as 21st century years

xiaoyinl:
One year ago, Mozilla changed the way Firefox interprets dates in the format `mm/dd/yy`. In Firefox version 49 or later, if `yy` is less than 50, the year is regarded as `20yy`; if `yy` is 50~99, the year is regarded as `19yy`. The old Firefox behavior matches IE and Edge, and its new behavior matches Chrome and Safari. Given that, I wonder if ChakraCore wants to follow them.

This PR changes the behavior of parsing two digit years to align with SpiderMonkey and v8.

I will update test cases, if you'd like to adopt this behavior change.

/cc @bterlson @dilijev

### DIFF
--- a/lib/Runtime/Library/DateImplementation.cpp
+++ b/lib/Runtime/Library/DateImplementation.cpp
@@ -1464,6 +1464,10 @@ Error:
                 lwYear = -lwYear + 1;
             }
         }
+        else if (lwYear < 50 && isDateNegativeVersion5 == false)
+        {
+            lwYear += 2000;
+        }
         else if (lwYear < 100 && isDateNegativeVersion5 == false)
         {
             lwYear += 1900;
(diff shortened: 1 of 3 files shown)


xiaoyinl:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1265136

bterlson:
If there are no compat concerns with this it seems fine to make this change. It's technically breaking, but only for code vulnerable to Y2K style bugs - hopefully not very common these days. Aligning with the majority of browsers seems like a good idea, and we have great evidence this change is web compatible given how broadly implemented it is.

+1.

bterlson:
Btw, @xiaoyinl your PRs have been awesome so far!

xiaoyinl:
@bterlson Thank you for your compliment!
@dilijev Tests are added.

xiaoyinl:
Thanks! Just squashed the commits.

PR #4063, opened 2017-10-25
adding readline support

JohnathanNorman:
Adding readline support. V8 and other engines have it and i find it useful for testing.


bin/ch/WScriptJsrt.cpp
```diff
@@ -1073,6 +1074,112 @@ JsValueRef __stdcall WScriptJsrt::LoadTextFileCallback(JsValueRef callee, bool i
     return returnValue;
 }
 
+int js_fgets(char* buf, int size, FILE* file)
```
MikeHolman: nit: prefer CamelCase method names

bin/ch/WScriptJsrt.cpp
```diff
@@ -1073,6 +1074,112 @@ JsValueRef __stdcall WScriptJsrt::LoadTextFileCallback(JsValueRef callee, bool i
...
+    return i;
+}
+
+JsValueRef __stdcall WScriptJsrt::ReadLineStdinCallback(JsValueRef callee, bool isConstructCall, JsValueRef *arguments, unsigned short argumentCount, void *callbackState)
+{
+    HRESULT hr = E_FAIL;
+    JsValueRef returnValue = JS_INVALID_REFERENCE;
+    JsErrorCode errorCode = JsNoError;
+
+#define BUFSIZE 256
```
MikeHolman: nit: prefer const int over #define

bin/ch/WScriptJsrt.cpp
```diff
@@ -1073,6 +1074,112 @@ JsValueRef __stdcall WScriptJsrt::LoadTextFileCallback(JsValueRef callee, bool i
...
+    }
+
+    buf = tmp;
+
+    /*
+    * Turn buf into a JSString. Note that buflength includes the trailing null
+    * character.
+    */
+    JsValueRef stringObject;
+    IfJsrtErrorSetGo(ChakraRTInterface::JsCreateString(buf, sawNewline ? buflength - 1 : buflength, &stringObject));
```
MikeHolman: JsCreateString copies the buffer, so you need to call to free

bin/ch/WScriptJsrt.cpp
```diff
@@ -1073,6 +1074,112 @@ JsValueRef __stdcall WScriptJsrt::LoadTextFileCallback(JsValueRef callee, bool i
...
+        }
+        else {
+            JsValueRef emptyStringObject;
+            IfJsrtErrorSetGo(ChakraRTInterface::JsCreateString(buf, buflength, &emptyStringObject));
+            return emptyStringObject;
+        }
+    }
+
+    /* Shrink the buffer to the real size. */
+    tmp = static_cast<char*>(realloc(buf, buflength));
```
MikeHolman: Since JsCreateString copies the buffer, you don't need to realloc here

bin/ch/WScriptJsrt.cpp
```diff
@@ -1073,6 +1074,112 @@ JsValueRef __stdcall WScriptJsrt::LoadTextFileCallback(JsValueRef callee, bool i
...
+        if (bufsize > buflength) {
+            tmp = static_cast<char*>(realloc(buf, bufsize));
+        }
+        else {
+            goto Error;
+        }
+
+        if (!tmp) {
+            free(buf);
+            goto Error;
```
MikeHolman: if error label isn't doing anything, you can just return here. however, i'd suggest having the error label do something like `if(buf) free(buf);` instead of on each branch, since it looks like a couple error blocks are missing call to free

bin/ch/WScriptJsrt.cpp
JohnathanNorman: changed.

bin/ch/WScriptJsrt.cpp
JohnathanNorman: done

bin/ch/WScriptJsrt.cpp
```diff
@@ -1073,6 +1074,114 @@ JsValueRef __stdcall WScriptJsrt::LoadTextFileCallback(JsValueRef callee, bool i
...
+        else {
+            JsValueRef emptyStringObject;
+            IfJsrtErrorSetGo(ChakraRTInterface::JsCreateString(buf, buflength, &emptyStringObject));
+            free(buf);
+            return emptyStringObject;
+        }
+    }
+
+    /* Shrink the buffer to the real size. */
+    tmp = static_cast<char*>(buf);
```
MikeHolman: can you just remove this block? doesn't really make sense anymore

bin/ch/WScriptJsrt.cpp
```diff
@@ -1073,6 +1074,114 @@ JsValueRef __stdcall WScriptJsrt::LoadTextFileCallback(JsValueRef callee, bool i
...
+    * character.
+    */
+    JsValueRef stringObject;
+    IfJsrtErrorSetGo(ChakraRTInterface::JsCreateString(buf, sawNewline ? buflength - 1 : buflength, &stringObject));
+    free(buf);
+    return stringObject;
+
+Error:
+    if (buf)
+        free(buf);
```
MikeHolman: nit: prefer brackets even for single line blocks

bin/ch/WScriptJsrt.cpp
```diff
@@ -1073,6 +1074,114 @@ JsValueRef __stdcall WScriptJsrt::LoadTextFileCallback(JsValueRef callee, bool i
...
+    if (!buf)
+        goto Error;
+
+    int gotlength;
+    while ((gotlength = JsFgets(buf + buflength, bufsize - buflength, from)) > 0) {
+        buflength += gotlength;
+
+        /* Are we done? */
+        if (buf[buflength - 1] == '\n') {
+            buf[buflength - 1] = '\0';
```
MikeHolman: it seems like the intention is to remove newline from input. but in case of `\r\n`, won't this replacement give you `\r\0`. also it looks like there is a bit of inconsistency. above you treat `\r` alone as ending line (e.g. even if not followed by `\n`), but here you only consider `\n`

bin/ch/WScriptJsrt.cpp
```diff
@@ -1073,6 +1074,114 @@ JsValueRef __stdcall WScriptJsrt::LoadTextFileCallback(JsValueRef callee, bool i
...
+
+    int gotlength;
+    while ((gotlength = JsFgets(buf + buflength, bufsize - buflength, from)) > 0) {
+        buflength += gotlength;
+
+        /* Are we done? */
+        if (buf[buflength - 2] == '\r' && buf[buflength - 1] == '\n') {
+            buf[buflength - 1] = '\0';
+            buf[buflength - 2] = '\0';
+            sawNewline = true;
```
Penguinwizzard: Here you take two characters off the buffer, but sawNewline only indicates that you should take one character off the buffer on L1173. Why not just modify buflength here rather than trying to keep a flag alive for this?

bin/ch/WScriptJsrt.cpp
```diff
@@ -1073,6 +1074,114 @@ JsValueRef __stdcall WScriptJsrt::LoadTextFileCallback(JsValueRef callee, bool i
...
+{
+    HRESULT hr = E_FAIL;
+    JsValueRef returnValue = JS_INVALID_REFERENCE;
+    JsErrorCode errorCode = JsNoError;
+
+    const int BUFSIZE = 256;
+    FILE* from = stdin;
+    int buflength = 0;
+    int bufsize = BUFSIZE;
+    char* buf = (char*)malloc(bufsize);
```
Penguinwizzard: Can you use c-style or static_cast for both this malloc and the realloc below? I'd rather not have them mixed in the same function.

bin/ch/WScriptJsrt.cpp
```diff
@@ -1073,6 +1074,114 @@ JsValueRef __stdcall WScriptJsrt::LoadTextFileCallback(JsValueRef callee, bool i
...
+    bool sawNewline = false;
+    char* tmp;
+    if (!buf)
+        goto Error;
+
+    int gotlength;
+    while ((gotlength = JsFgets(buf + buflength, bufsize - buflength, from)) > 0) {
+        buflength += gotlength;
+
+        /* Are we done? */
```
jianchun: >/* Are we done? */ [](start = 8, length = 18)

nit pick: Existing source rarely uses this comment style. Could you change them to "//" so the code looks a bit more consistent?

bin/ch/WScriptJsrt.cpp
```diff
@@ -1073,6 +1074,114 @@ JsValueRef __stdcall WScriptJsrt::LoadTextFileCallback(JsValueRef callee, bool i
...
+int JsFgets(char* buf, int size, FILE* file)
+{
+    int n = size - 1;
+    if (n < 0)
+        return -1;
+
+    bool crflag = false;
+    int c, i = 0;
+
+    for (i = 0; i < n && (c = getc(file)) != EOF; i++) {
```
jianchun: >{ [](start = 55, length = 1)

nit pick: Could you please change this "{ at end of line" style to "{ at next line by its own", to be consistent with rest of the code?

MikeHolman:
Looks like few UTs are unhappy with the new global readline method (which is why CI is all red). Can you update the baselines for these failing tests?

bin/ch/WScriptJsrt.cpp
```diff
@@ -1073,6 +1074,114 @@ JsValueRef __stdcall WScriptJsrt::LoadTextFileCallback(JsValueRef callee, bool i
...
+    /* Treat the empty string specially. */
+    if (buflength == 0) {
+        if (feof(from)) {
+            goto Error;
+        }
+        else {
+            JsValueRef emptyStringObject;
+            IfJsrtErrorSetGo(ChakraRTInterface::JsCreateString(buf, buflength, &emptyStringObject));
+            free(buf);
+            return emptyStringObject;
```
jianchun: This looks very similar to below; may be able to merge this empty string result with normal non-empty string result.

dilijev:
ping @JohnathanNorman

JohnathanNorman:
Ok i'm lame. I had some problems updating the unit tests and then got busy
with other projects. Unfortunately i wont have time to finish this up, you
guys can just kill the pull request

dilijev:
@JohnathanNorman we won't kill it just yet. :) Thanks for the contribution. I'll leave it open for now in case someone wants to adopt it, fix it up, and merge it.

Cellule:
Closing as this was merged by #4576

PR #4066, opened 2017-10-25
RegExp Bytecode: Remove HardFailMixin and WordBoundaryTest's isNegation, converting bools to template parameters.

dilijev:
Follow up to #3915 as part of the "remove booleans" opportunity. Other work on removing the rest of the booleans is in progress but more invasive.

This PR also:

* removes virtual Print from `Inst` so that Debug build `Inst`s don't have a vtable pointer to disrupt the bytecode layout.
* includes other changes to -RegexBytecodeDebug output to help view important aspects of bytecode layout.

I don't yet have data about runtime and memory improvement with this change, but since we are removing some branches through the use of a boolean template instead of a boolean member variable and decreasing memory by removing the boolean and all of the associated padding (4 or 8 bytes, depending) any time these bytecodes are used, there should be moderate gains in both runtime and memory.

BOITest / BOIHardFailTest / EOITest / EOIHardFailTest / WordBoundaryTest / NegatedWordBoundaryTest go from 0x8 to 0x4 bytes. (Opcode is 32 bit and the bool gets padded out 4 bytes.)

Before/after snippet for BOITest, for `/(^token|foo)/` (running as `ch.exe -RegexDebug -RegexBytecodeDebug test.js`):

Before:
```
        L0010: (0x008 bytes) BOITest(hardFail: false)
            0x000001F4591AC0B0[+0x000](0x004) [Inst]: 0c000000
            0x000001F4591AC0B4[+0x004](0x001) [HardFailMixin]:00
        L0018: (0x00c bytes) MatchLiteral(literal: "token")
```

After:
```
        L0010: (0x004 bytes) BOITest(<hardFail>: false)
            0x0000020E2BECC0B0[+0x000](0x004)(size:0x04)(align:0x04) [Inst]: 0d000000
        L0014: (0x00c bytes) MatchLiteral(literal: "token")
```

Again for `/\b\B/`:

Before:
```
        L0000: (0x008 bytes) WordBoundaryTest()
            0x0000023973F34100[+0x000](0x004) [Inst]: 10000000
            0x0000023973F34100[+0x000](0x008) [WordBoundaryTestInst]: 10000000 00000000
        L0008: (0x008 bytes) WordBoundaryTest()
            0x0000023973F34108[+0x000](0x004) [Inst]: 10000000
            0x0000023973F34108[+0x000](0x008) [WordBoundaryTestInst]: 10000000 01000000
        L0010: (0x004 bytes) Succ()
```

After:
```
        L0000: (0x004 bytes) WordBoundaryTest()
            0x0000019F284AC190[+0x000](0x004)(size:0x04)(align:0x04) [Inst]: 13000000
        L0004: (0x004 bytes) NegatedWordBoundaryTest()
            0x0000019F284AC194[+0x000](0x004)(size:0x04)(align:0x04) [Inst]: 12000000
        L0008: (0x004 bytes) Succ()
```


lib/Parser/RegexRuntime.h
```diff
@@ -851,39 +840,44 @@ namespace UnifiedRegex
     // Built-in assertions
     //
 
-    struct BOITestInst : Inst, HardFailMixin
+    // BOI = Beginning of Input
+    template <bool canHardFail>
+    struct BOITestInst : Inst
     {
-        inline BOITestInst(bool canHardFail) : Inst(BOITest), HardFailMixin(canHardFail) {}
+        inline BOITestInst();
```
Penguinwizzard: nit: I don't think this needs to be declared as inline if you're putting the implementation in the cpp.

dilijev: I think it's helpful for the reader to know this is defined elsewhere (at least that's how I read this).

dilijev:
@Penguinwizzard I decided to include removal of virtual Print into this PR. Update to message incoming.

dilijev:
@Penguinwizzard updated now.

lib/Parser/RegexRuntime.h
dilijev: Turns out this declaration is also required or the compiler assumes it was deleted and explicit template overrides conflict with that.

`note: 'UnifiedRegex::BOITestInst<true>::BOITestInst(void)': function was implicitly deleted because a base class 'UnifiedRegex::Inst' has either no appropriate default constructor or overload resolution was ambiguous`

PR #4067, opened 2017-10-25
Make Date.to(UTC)String() outputs match web reality and spec language

xiaoyinl:
Ref #4178

[Sec 20.3.3.2](https://tc39.github.io/ecma262/#sec-date.parse) requires `Date.parse` should be able to parse the values returned by `Date.toString()`, `Date.toUTCString()` and `Date.toISOString()` in each engine's own implementation. In ChakraCore, if the year is <100, `toString()` and `toUTCString()` don't pad the year to 4 digits, and thus `Date.parse` can't parse them.

This PR makes two changes:
1. If the absolute value of the year is less than 1000, pad it to 4 digits
2. For a BC year, output negative year, rather than positive year + "B.C."

### DIFF
--- a/lib/Runtime/Library/DateImplementation.cpp
+++ b/lib/Runtime/Library/DateImplementation.cpp
@@ -402,14 +402,15 @@ namespace Js {
         bs->AppendChars(_u(' '));
 
         // Add the year.
-        if (pymd->year > 0)
+        if ((pymd->year) > 0)
         {
-            bs->AppendChars(pymd->year, 10, ConvertLongToString);
+            bs->AppendChars(pymd->year, 10, ConvertUInt32ToString_ZeroPad_4);
         }
         else
         {
-            bs->AppendChars(1 - pymd->year, 10, ConvertLongToString);
-            bs->AppendChars(_u(" B.C."));
+            int positiveYear = -(pymd->year); // pymd->year is negative
+            bs->AppendChars(_u('-'));
+            bs->AppendChars(positiveYear, 10, ConvertUInt32ToString_ZeroPad_4);
         }
 
         // Add the time.
--- a/lib/Runtime/Library/DateImplementation.h
+++ b/lib/Runtime/Library/DateImplementation.h
@@ -382,6 +382,40 @@ namespace Js {
         return GetDateDefaultString<StringBuilder>(&ymd, &tzd, DateTimeFlag::None, scriptContext, newStringBuilder);
     }
 
+    const auto ConvertUInt32ToString_ZeroPad_4 = [](const uint32 value, char16 *const buffer, const CharCount charCapacity)
+    {
+        Assert(charCapacity >= 4);
+        if (value < 10)
+        {
+            buffer[0] = _u('0');
+            buffer[1] = _u('0');
+            buffer[2] = _u('0');
+            buffer[3] = static_cast<char16>(value + _u('0'));
+            buffer[4] = 0;
+        }
+        else if (value < 100)
+        {
+            buffer[0] = _u('0');
+            buffer[1] = _u('0');
+            buffer[2] = static_cast<char16>((value / 10) + _u('0'));
+            buffer[3] = static_cast<char16>((value % 10) + _u('0'));
+            buffer[4] = 0;
+        }
+        else if (value < 1000)
+        {
+            buffer[0] = _u('0');
+            buffer[1] = static_cast<char16>((value / 100) + _u('0'));
+            buffer[2] = static_cast<char16>(((value / 10) % 10) + _u('0'));
+            buffer[3] = static_cast<char16>((value % 10) + _u('0'));
+            buffer[4] = 0;
+        }
+        else
+        {
+            const errno_t err = _ultow_s(value, buffer, charCapacity, 10);
+            Assert(err == 0);
+        }
+    };
+
     //
     // Get default date string, shared with hybrid debugging.
     //  StringBuilder: A Js::StringBuilder/CompoundString like class, used to build strings.
@@ -418,7 +452,16 @@ namespace Js {
             bs->AppendChars(_u(' '));
 
             // year is directly after day, month, daydigit for IE11+
-            bs->AppendChars(pymd->year, 10, ConvertLongToString);
+            if ((pymd->year) > 0)
+            {
+                bs->AppendChars(pymd->year, 10, ConvertUInt32ToString_ZeroPad_4);
+            }
+            else
+            {
+                int positiveYear = -(pymd->year); // pymd->year is negative
+                bs->AppendChars(_u('-'));
+                bs->AppendChars(positiveYear, 10, ConvertUInt32ToString_ZeroPad_4);
+            }
 
             if(!(noDateTime & DateTimeFlag::NoTime))
             {
--- a/test/Date/toStringAndToUTCStringYearPadding.js
+++ b/test/Date/toStringAndToUTCStringYearPadding.js
@@ -1,0 +1,47 @@
+//-------------------------------------------------------------------------------------------------------
+// Copyright (C) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE.txt file in the project root for full license information.
+//-------------------------------------------------------------------------------------------------------
+
+// For dates whose years are less than 1000, Date.toString() and Date.toUTCString() should pad the years
+// to 4 digits.
+// See https://github.com/Microsoft/ChakraCore/pull/4067
+
+/// <reference path="../UnitTestFramework/UnitTestFramework.js" />
+if (this.WScript && this.WScript.LoadScriptFile) { // Check for running in ch
+    this.WScript.LoadScriptFile("..\\UnitTestFramework\\UnitTestFramework.js");
+}
+
+function testDate(DateObj, toStringExpected, toUTCStringExpected) {
+    assert.areEqual(toStringExpected, DateObj.toString(), "Date.toString() returns invalid value.");
+    assert.areEqual(toUTCStringExpected, DateObj.toUTCString(), "Date.toUTCString() returns invalid value.");
+}
+
+let tests = [{
+    name: "test if Date.toString() and Date.toUTCString() pad positive years to four digits",
+    body: function () {
+        testDate(new Date("0001-10-13T05:16:33Z"), "Fri Oct 12 0001 22:16:33 GMT-0700 (Pacific Daylight Time)",
+            "Sat, 13 Oct 0001 05:16:33 GMT");
+        testDate(new Date("0011-10-13T05:16:33Z"), "Wed Oct 12 0011 22:16:33 GMT-0700 (Pacific Daylight Time)",
+            "Thu, 13 Oct 0011 05:16:33 GMT");
+        testDate(new Date("0111-10-13T05:16:33Z"), "Mon Oct 12 0111 22:16:33 GMT-0700 (Pacific Daylight Time)",
+            "Tue, 13 Oct 0111 05:16:33 GMT");
+        testDate(new Date("1111-10-13T05:16:33Z"), "Thu Oct 12 1111 22:16:33 GMT-0700 (Pacific Daylight Time)",
+            "Fri, 13 Oct 1111 05:16:33 GMT");
+    }
+},
+{
+    name: "test if Date.toString() and Date.toUTCString() pad negative years to four digits",
+    body: function () {
+        testDate(new Date("-000001-10-13T05:16:33Z"), "Tue Oct 12 -0001 22:16:33 GMT-0700 (Pacific Daylight Time)",
+            "Wed, 13 Oct -0001 05:16:33 GMT");
+        testDate(new Date("-000011-10-13T05:16:33Z"), "Thu Oct 12 -0011 22:16:33 GMT-0700 (Pacific Daylight Time)",
+            "Fri, 13 Oct -0011 05:16:33 GMT");
+        testDate(new Date("-000111-10-13T05:16:33Z"), "Sat Oct 12 -0111 22:16:33 GMT-0700 (Pacific Daylight Time)",
+            "Sun, 13 Oct -0111 05:16:33 GMT");
+        testDate(new Date("-001111-10-13T05:16:33Z"), "Wed Oct 12 -1111 22:16:33 GMT-0700 (Pacific Daylight Time)",
+            "Thu, 13 Oct -1111 05:16:33 GMT");
+    }
+}];
+
+testRunner.run(tests, { verbose: WScript.Arguments[0] != "summary" });
--- a/test/Date/DateParse2.js
+++ b/test/Date/DateParse2.js
@@ -8,7 +8,8 @@
 myPrint("A --");
 testDate(new Date(-2012, 1, 2, 1, 2, 3));
 testParseDate(new Date(-2012, 1, 2, 1, 2, 3).toString());
-testParseDate(new Date(-2012, 1, 2, 1, 2, 3).toUTCString());
+// Disabled due to https://github.com/Microsoft/ChakraCore/issues/4300
+//testParseDate(new Date(-2012, 1, 2, 1, 2, 3).toUTCString());
 testParseDate(new Date(-2012, 1, 2, 1, 2, 3).toISOString());
 
 myPrint("B --");
@@ -26,7 +27,7 @@ testParseDate(new Date(99999, 1, 2, 1, 2, 3).toISOString());
 myPrint("D --");
 testDate(new Date(-99999, 1, 2, 1, 2, 3));
 testParseDate(new Date(-99999, 1, 2, 1, 2, 3).toString());
-testParseDate(new Date(-99999, 1, 2, 1, 2, 3).toUTCString());
+//testParseDate(new Date(-99999, 1, 2, 1, 2, 3).toUTCString());
 testParseDate(new Date(-99999, 1, 2, 1, 2, 3).toISOString());
 
 myPrint("E --");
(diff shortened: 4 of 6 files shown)


lib/Runtime/Library/DateImplementation.cpp
```diff
@@ -404,7 +404,7 @@ namespace Js {
         // Add the year.
         if (pymd->year > 0)
         {
-            bs->AppendChars(pymd->year, 10, ConvertLongToString);
+            bs->AppendChars(static_cast<uint32>(pymd->year), 10, ConvertUInt32ToString_ZeroPad_4);
```
dilijev: Did something change that now this static_cast is necessary?

lib/Runtime/Library/DateImplementation.h
```diff
@@ -382,6 +382,39 @@ namespace Js {
...
+            buffer[2] = ((static_cast<uint16>(value) / 10) % 10) + _u('0');
+            buffer[3] = (static_cast<uint16>(value) % 10) + _u('0');
+            buffer[4] = 0;
+        }
+        else
+        {
+            const errno_t err = _ultow_s(value, buffer, charCapacity, 10);
+            Assert(err == 0);
+        }
+    };
```
dilijev: nit: newline after this

dilijev: In general the above function is a bit ugly but I understand the need to implement ~ this way.

lib/Runtime/Library/DateImplementation.h
```diff
@@ -418,7 +451,14 @@ namespace Js {
             bs->AppendChars(_u(' '));
 
             // year is directly after day, month, daydigit for IE11+
-            bs->AppendChars(pymd->year, 10, ConvertLongToString);
+            if (pymd->year > 0)
```
dilijev: nit: wrap the `pymd->year` in parens for readability.

dilijev:
Ah, I noticed the test code is assuming a complete implementation. If we take this as-is we should disable the failing test cases for now.

dilijev:
Also looks like the code doesn't yet handle negative years for the year padding:

`Date.parse can't parse "Sat Nov 13 -1 17:26:37 GMT-0800 (Pacific Standard Time)", returned by toString() on Date -000001-11-14T01:26:37.343Z`

(Note the year `-000001`)

xiaoyinl:
@dilijev Thanks for the review! I prefer to merge the current code change in this PR, and I will open another PR for making `Date.parse` recognize zero-padded years. I will address your comments, and update the tests so that they only test the outputs of `toString`.

dilijev:
Would be nice to compare the `toString` output with how other engines pad out the year, in case there's a majority agreement on how it should be done.

Here's my test:

```
┌──────────────────┬───────────────────────────────────────────────────────────┐
│ ch-1.7.1         │ Fri Oct 12 1 22:16:33 GMT-0700 (Pacific Daylight Time)    │
│ ch-master-latest │ Wed Oct 12 11 22:16:33 GMT-0700 (Pacific Daylight Time)   │
│ node-ch          │ Mon Oct 12 111 22:16:33 GMT-0700 (Pacific Daylight Time)  │
│                  │ Thu Oct 12 1111 22:16:33 GMT-0700 (Pacific Daylight Time) │
├──────────────────┼───────────────────────────────────────────────────────────┤
│ d8               │ Fri Oct 12 0001 22:16:33 GMT-0700 (Pacific Daylight Time) │
│ jsc              │ Wed Oct 12 0011 22:16:33 GMT-0700 (Pacific Daylight Time) │
│ sm               │ Mon Oct 12 0111 22:16:33 GMT-0700 (Pacific Daylight Time) │
│                  │ Thu Oct 12 1111 22:16:33 GMT-0700 (Pacific Daylight Time) │
├──────────────────┼───────────────────────────────────────────────────────────┤
│ node             │ Fri Oct 12    1 22:16:33 GMT-0700 (Pacific Daylight Time) │
│                  │ Wed Oct 12   11 22:16:33 GMT-0700 (Pacific Daylight Time) │
│                  │ Mon Oct 12  111 22:16:33 GMT-0700 (Pacific Daylight Time) │
│                  │ Thu Oct 12 1111 22:16:33 GMT-0700 (Pacific Daylight Time) │
└──────────────────┴───────────────────────────────────────────────────────────┘
```

I wonder why node uses spaces instead of leading 0's on the year?
/cc @bterlson: do you have any insight here?

dilijev:
Handling of negative years has less inter-engine agreement:

```
## Source
let year1BC = new Date("0001-10-13T05:16:33Z");
let year11BC = new Date("0001-10-13T05:16:33Z");
year1BC.setFullYear(-1, 10, 13);
year11BC.setFullYear(-11, 10, 13);

print(year1BC.toString());
print(year11BC.toString());


┌──────────────────┬────────────────────────────────────────────────────────────┐
│ d8               │ Sat Nov 13 -001 22:16:33 GMT-0800 (Pacific Standard Time)  │
│ jsc              │ Mon Nov 13 -011 22:16:33 GMT-0800 (Pacific Standard Time)  │
├──────────────────┼────────────────────────────────────────────────────────────┤
│ ch-1.7.1         │ Sat Nov 13 -1 22:16:33 GMT-0800 (Pacific Standard Time)    │
│ ch-master-latest │ Mon Nov 13 -11 22:16:33 GMT-0800 (Pacific Standard Time)   │
│ node-ch          │                                                            │
├──────────────────┼────────────────────────────────────────────────────────────┤
│ sm               │ Sat Nov 13 -0001 22:16:33 GMT-0800 (Pacific Standard Time) │
│                  │ Mon Nov 13 -0011 22:16:33 GMT-0800 (Pacific Standard Time) │
├──────────────────┼────────────────────────────────────────────────────────────┤
│ node             │ Sat Nov 13   -1 22:16:33 GMT-0800 (Pacific Standard Time)  │
│                  │ Mon Nov 13  -11 22:16:33 GMT-0800 (Pacific Standard Time)  │
└──────────────────┴────────────────────────────────────────────────────────────┘
```

test/Date/parseToStringResults.js
```diff
@@ -0,0 +1,49 @@
...
+// Date.parse must be able to parse the strings returned by Date.toString(), Date.toUTCString() and Date.toISOString()
+
+runTest("0001-10-13T05:16:33Z");
+runTest("0011-10-13T05:16:33Z");
+runTest("0111-10-13T05:16:33Z");
+runTest("1111-10-13T05:16:33Z");
+
+// test BC years
+let year1BC = new Date();
+let year11BC = new Date();
```
dilijev: Would be better to do `new Date("0001-10-13T05:16:33Z")` (with a specific initialization string) so that the values and output is more predictable.

dilijev:
Added bug #4178

lib/Runtime/Library/DateImplementation.h
xiaoyinl: Fixed.

lib/Runtime/Library/DateImplementation.h
xiaoyinl: Done.

lib/Runtime/Library/DateImplementation.cpp
xiaoyinl: I remember I put `static_cast` there only for stylistic reason (making `int` to `uint32` conversion explicit). It's unnecessary so I removed it.

xiaoyinl:
@dilijev I have updated the code, and it now also pads negative years to 4 digits, matching SpiderMonkey, because I think this might make `Date.parse()` easier to parse these years.

dilijev:
@xiaoyinl Thanks -- also I happen to think that format makes the most sense. :)

lib/Runtime/Library/DateImplementation.h
```diff
@@ -385,28 +385,29 @@ namespace Js {
     const auto ConvertUInt32ToString_ZeroPad_4 = [](const uint32 value, char16 *const buffer, const CharCount charCapacity)
     {
         Assert(charCapacity >= 4);
+        uint16 value_uint16 = static_cast<uint16>(value);
```
dilijev: Not sure I see the value here. Should be the same logic whether 32 or 16 bit?

xiaoyinl: I remember I put it there, because otherwise I need to cast `value + _u('0')` to 16 bit every time it's assigned to `buffer[]`, which takes char16.

lib/Runtime/Library/DateImplementation.h
dilijev: It wasn't super clear from the read that the reason for converting the input number to uint16 is so that it is compatible with the char16 _u('0'). Can you add a comment explaining? What will behavior look like for years > 4 digits and > MAX_UINT16 ?

lib/Runtime/Library/DateImplementation.h
```diff
@@ -418,7 +453,15 @@ namespace Js {
...
             // year is directly after day, month, daydigit for IE11+
-            bs->AppendChars(pymd->year, 10, ConvertLongToString);
+            if ((pymd->year) > 0)
+            {
+                bs->AppendChars(pymd->year, 10, ConvertUInt32ToString_ZeroPad_4);
+            }
+            else
+            {
+                bs->AppendChars(_u('-'));
+                bs->AppendChars(-(pymd->year), 10, ConvertUInt32ToString_ZeroPad_4);
```
dilijev: Rather than passing in a negated negative number, maybe use a local variable named to make the intent more clear? Like
`positiveYear = -(pymd->year); // pymd->year is negative`

test/Date/toStringAndToUTCStringYearPadding.js
```diff
@@ -0,0 +1,56 @@
...
+// Licensed under the MIT license. See LICENSE.txt file in the project root for full license information.
+//-------------------------------------------------------------------------------------------------------
+
+// For dates whose years are less than 1000, Date.toString() and Date.toUTCString() should pad the years
+// to 4 digits.
+// See https://github.com/Microsoft/ChakraCore/pull/4067
+
+runTest(new Date("0001-10-13T05:16:33Z"),
+    "Fri Oct 12 0001 22:16:33 GMT-0700 (Pacific Daylight Time)",
+    "Sat, 13 Oct 0001 05:16:33 GMT");
```
dilijev: I think this test would read better if the definition of `runTest` were at the top of the file.

Also, you might consider using UnitTestFramework.js (see usage in other tests) and using asserts instead of rolling your own messages for failures.

test/Date/toStringAndToUTCStringYearPadding.js
```diff
@@ -0,0 +1,56 @@
...
+    "Tue, 13 Oct 0111 05:16:33 GMT");
+
+runTest(new Date("1111-10-13T05:16:33Z"),
+    "Thu Oct 12 1111 22:16:33 GMT-0700 (Pacific Daylight Time)",
+    "Fri, 13 Oct 1111 05:16:33 GMT");
+
+// Negative years are also padded to four digits
+runTest(new Date("-000001-10-13T05:16:33Z"),
+    "Tue Oct 12 -0001 22:16:33 GMT-0700 (Pacific Daylight Time)",
+    "Wed, 13 Oct 0002 B.C. 05:16:33 GMT");
```
dilijev: I wonder why `toUTCString` not only prints GMT timezone (as opposed to UTC), but also switches from negative year to B.C. year. I think we should consult the spec and other engines on what the outputs should look like.

lib/Runtime/Library/DateImplementation.h
xiaoyinl: Agreed. I fixed it by moving the `static_cast` to where the int32 value is assigned to char16, for readability and avoided > MAX_UINT16 issue.

xiaoyinl:
@dilijev Sorry for the delay! I was a little busy these days. I think I have addressed the issues you mentioned in the above comments. UnitTestFramework.js is used. It's indeed much more convenient.

test/Date/toStringAndToUTCStringYearPadding.js
xiaoyinl: For this date (`-000001-10-13T05:16:33Z`), both v8 and SpiderMonkey output GMT timezone and negative years: `Wed, 13 Oct -001 05:16:33 GMT` and `Wed, 13 Oct -0001 05:16:33 GMT` respectively.

xiaoyinl:
I just realized [Section 20.3.4.43](https://tc39.github.io/ecma262/#sec-date.prototype.toutcstring) specifies the exact format of `toUTCString()`. It says:

> [...]
> 7. Let year be the String representation of YearFromTime(tv), formatted as a number of at least four digits, padded to the left with zeroes if necessary.
> 8. Return the string-concatenation of weekday, ",", the code unit 0x0020 (SPACE), month, the code unit 0x0020 (SPACE), day, the code unit 0x0020 (SPACE), year, the code unit 0x0020 (SPACE), and TimeString(tv).

So it seems `toUTCString()` should print a negative year, instead of a positive year + "B.C.".

Also I'm confused that based on item 8, it seems an UTCString should look like `Mon, Nov 20 2017 15:16:33 GMT` (month before day), but v8, SpiderMonkey and ChakraCore all print `Mon, 20 Nov 2017 15:16:33 GMT` (day before month). Do I misunderstand the spec or should we follow that?

/cc @bterlson

dilijev:
We generally go with the flow of 3/4 agreement (web reality). In this case looks like we all already agree, and the spec should be updated to match the web reality. I'll open a PR on the spec. /cc @bterlson 

Printing B.C. definitely seems to be a bug.

```
## Source
let posYear = new Date('2008-10-13T05:16:33Z').toUTCString();
print(posYear);

┌──────────────────┬───────────────────────────────┐
│ ch-1.6-latest    │ Mon, 13 Oct 2008 05:16:33 GMT │
│ ch-1.7-latest    │                               │
│ ch-1.7.4         │                               │
│ ch-1.8-latest    │                               │
│ ch-master-latest │                               │
│ d8               │                               │
│ jsc              │                               │
│ node             │                               │
│ node-ch          │                               │
│ sm               │                               │
└──────────────────┴───────────────────────────────┘

## Source
let negYear = new Date((new Date('2008-10-13T05:16:33Z')).setFullYear(-2008, 10, 13)).toUTCString();
print(negYear);

┌──────────────────┬────────────────────────────────────┐
│ ch-1.6-latest    │ Sat, 14 Nov 2009 B.C. 06:16:33 GMT │
│ ch-1.7-latest    │                                    │
│ ch-1.7.4         │                                    │
│ ch-1.8-latest    │                                    │
│ ch-master-latest │                                    │
│ node-ch          │                                    │
├──────────────────┼────────────────────────────────────┤
│ d8               │ Sat, 14 Nov -2008 06:16:33 GMT     │
│ jsc              │                                    │
│ node             │                                    │
│ sm               │                                    │
└──────────────────┴────────────────────────────────────┘
```

dilijev:
@xiaoyinl I retargeted this PR to release/1.8. Would you mind cherry-picking or rebasing your changes on top of release/1.8?

dilijev:
@xiaoyinl Thanks!

xiaoyinl:
@dilijev I've cherry-picked on top of release/1.8 and I also changed the B.C. to printing negative years, so that it should now completely match SpiderMonkey.

dilijev:
Hm... for the title, I think it would be better to say "Make Date.to(UTC)String() outputs match web reality and spec language".

Spec language re: negative years: the year is padded to 4 digits as indicated in spec, and preceeded by a `-`, which is not discussed in the spec, and I'll make a clarifying PR to the spec to address that point.

It's generally not our goal to match any particular other engine, regardless of whether we agree with them. Either we follow consensus by all engines, or where that's not clear, we follow the spec.

In this case, the implementation being updated will make 2/4 consensus on how to format negative years with the other engines differing.

dilijev:
As for negative years, besides negative sign and then the absolute value padded to 4 digits, the other logical interpretation of the spec is to behave as `printf` formatting a negative number with `printf("%04d", negativeNumber);`, which for `int negativeNumber = -3;` would print `-003`. (Java's `String.format` also agrees with that interpretation of `%04d` with a negative argument.)

This is what `d8` and `jsc` have done in terms of output, and I'd imagine that the implemented the formatting behavior with something akin to `sprintf` instead of rolling their own as we've done here.

We don't have to change this PR at this time since it's unclear from the spec, but we should bring up this point as a point that lacks clarity in the spec.

dilijev:
If the spec clarification ends up favoring `-003` over `-0003` then it will be easy to make that change later.

dilijev:
These two tests failed. They are tagged exclude_jenkins so they didn't run on GitHub's CI. Can you look into the failures?

    x64_debug\dynapogo\rl.results.log:Date (DateParse2.js () exe) -- failed : exec time=0
    x64_debug\dynapogo\rl.results.log:Date (toStringAndToUTCStringYearPadding.js () exe) -- failed : exec time=0

xiaoyinl:
I have fixed the test failure for toStringAndToUTCStringYearPadding.js. The other failure is a little complicated. It's due to the change of the BC year to negative year. DateParse2.js tests if `Date.parse` is able to parse outputs of Date.toString() and Date.toUTCString(). The thing is the current `Date.parse` can parse string like `Tue, 02 Feb 2013 B.C. 09:02:03 GMT`, but not `Tue, 02 Feb -2012 09:02:03 GMT`. Thus the change in this PR breaks the test.

But as we discussed earlier fixing `Date.parse` isn't the scope of this PR, so I wonder if I can disable `DateParse2.js` for now, and I will send another PR to fix that and at that time re-enable this test?

dilijev:
@xiaoyinl That sounds fair. You can disable just the smallest part of that test that fails (by commenting it out). Also open an issue to track the disabled portion of the test.

xiaoyinl:
@dilijev I commented out 2 tests and removed corresponding outputs in the baseline file. I run the tests locally and all tests are passed now.

dilijev:
Looks great. I'll merge soon.

xiaoyinl:
@dilijev Thank you!

PR #4070, opened 2017-10-26
Increase the Wasm function size limit to 7654321 bytes

Cellule:
Bumping limit according to discussion on https://github.com/WebAssembly/design/issues/1138

Fixes OS#14291237

This change is [Reviewable](https://reviewable.io/reviews/microsoft/chakracore/4070)

### DIFF
--- a/lib/WasmReader/WasmLimits.h
+++ b/lib/WasmReader/WasmLimits.h
@@ -26,7 +26,7 @@ namespace Wasm {
         static const uint32 MaxMemoryInitialPages = 16384;
         static const uint32 MaxMemoryMaximumPages = 65536;
         static const uint32 MaxModuleSize = 1024 * 1024 * 1024;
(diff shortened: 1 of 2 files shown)


MikeHolman:
I thought we were doing 7654321 bytes? That sounded like fun

dilijev:
Sounds like the 7654321 number was for the spec footnote so that it's clear it's arbitrary. 7MB is a nice round number (better for checking length against the limit?) and "close enough" to that limit. Depends on whether adherence to the exact value is important.

dilijev:
nit: title is megabits, not megabytes, and anyway we're talking about 7 MiB (mebibytes) = 7 * 2^10

Cellule:
I got focused on the 7MB limits discussed and somehow discarded the 7654321 bytes limit discussed after.
I updated to reflect the new limit

PR #4071, opened 2017-10-26
RegExp Bytecode: Optimize memory for Switch instructions.

dilijev:
The number of cases supported by each switch opcode seems to have been chosen somewhat arbitrarily in the original implementation, and for the common case of few alternates, has a lot of unused space. This PR replaces the old Switch opcodes (10, 20 cases) with a new set of Switch opcodes following a pseudo-doubling scheme (2, 4, 8, 16, 24*). 

\* The final increment was chosen to ensure that unused memory is never greater than the predecessor, we support at least as many switch cases as before before switching to chained layout.

This has a slightly higher cost at compile time because there are now 5 Switch opcode alternatives as opposed to the previous 2.

Memory usage is improved on average across all cases (fewer unused slots on average), especially in the presumed common case of a small number of alternatives (human edited regexps) when the switch has 8 or fewer cases. I don't yet have data about the distribution of the various numbers of cases.

# Analysis of Memory Improvement

In my test suite running locally (1000 tests from RegexTestCollateral), this change improves overall memory usage of Switch statements in the RegExp by 65.5% (now 34.5% of their former size).

Unfortunately my analysis also revealed that Switch statements are a relatively small contributor to total bytecode size, saving only 0.25% of overall RegExp Bytecode size in the above test suite.

# Future Work

Currently, each case takes up 2+4 [8 when padded] bytes (2 [4] for the char16, 4 bytes for the Jump target which is a uint32_t to support any size of regexp program). This could be improved with reduced layout options of 16-bit jump-targets (2+2 [4] bytes per case).

This could be further improved with a narrow layout for a single byte char for the ASCII case, and further by reducing the size of the jumps (using offset from next instruction would make this more feasible but for large numbers of cases the later targets are spaced out enough that jumps would probably not be able to fit into a single byte -- this would be improved by better packing and ASCII layout for common instructions like MatchChar).

In order to use the reduced layouts, we would have to keep track of the highest values for the various fields at compile time and use the opcode layout that best fits them.

Using a switch instruction with an exact number of cases encoded would make the `numCases` field less redundant, reduce code, and maximize memory efficiency, but would mean that Switch statements are no longer a fixed size. Could solve this by serializing a number of "SwitchCase" structs after the Switch instruction itself without it needing to be part of the same struct.

# Example

2 cases. `/(hello|world)/`
Before (scroll to the right to note the unused slots):

```
        L0020: (0x058 bytes) SwitchAndConsume10(
            case 'h': Jump(L007c)
            case 'w': Jump(L0090)
        )
            0x000002862F6F8020[+0x000](0x004) [Inst]: 0a000000
            0x000002862F6F8024[+0x004](0x054) [SwitchMixin<10>]: 02000000 68000000 7c000000 77000000 90000000 ffff0000 ffffffff ffff0000 ffffffff ffff0000 ffffffff ffff0000 ffffffff ffff0000 ffffffff ffff0000 ffffffff ffff0000 ffffffff ffff0000 ffffffff
```

After:

```
        L0020: (0x018 bytes) SwitchAndConsume2(
            case 'h': Jump(L003c)
            case 'w': Jump(L0050)
        )
            0x0000022F95158090[+0x000](0x004) [Inst]: 0d000000
            0x0000022F95158094[+0x004](0x014) [SwitchMixin<2>]: 02000000 68000000 3c000000 77000000 50000000
```


lib/Parser/RegexRuntime.cpp
```diff
@@ -914,11 +908,11 @@ namespace UnifiedRegex
     }
 #endif
 
-    template <int n>
+    template <uint8_t n>
     void SwitchMixin<n>::AddCase(char16 c, Label targetLabel)
     {
         Assert(numCases < MaxCases);
```
akroshg: nit: You can change this assert to AnalysisAssert and remove below statement.

lib/Parser/RegexRuntime.cpp
```diff
@@ -1186,270 +1187,73 @@ namespace UnifiedRegex
...
+
+#define SwitchAndConsumeInstImpl(BaseName, n) \
+    inline bool BaseName##n##Inst::Exec(REGEX_INST_EXEC_PARAMETERS) const                       \
+    {                                                                                           \
+        if (inputOffset >= inputLength)                                                         \
+        {                                                                                       \
+            return matcher.Fail(FAIL_PARAMETERS);                                               \
+        }                                                                                       \
+                                                                                                \
+        const int localNumCases = numCases;                                                     \
```
akroshg: >int [](start = 14, length = 3)

should this be unit_8 ?

lib/Parser/RegexCompileTime.cpp
```diff
@@ -2683,28 +2706,65 @@ namespace UnifiedRegex
...
+                        }
+                        else if (switchSize <= Switch8Inst::MaxCases)
+                        {
+                            EMIT(compiler, SwitchAndConsume8Inst);
+                        }
+                        else if (switchSize <= Switch16Inst::MaxCases)
+                        {
+                            EMIT(compiler, SwitchAndConsume16Inst);
+                        }
+                        else if (switchSize <= Switch24Inst::MaxCases)
```
MSLaguana: Should this just be an `else` to match previous behavior? I know we have an assert above, but could Bad Things happen in release builds if none of these cases are hit?

lib/Parser/RegexCompileTime.cpp
```diff
@@ -2659,7 +2680,7 @@ namespace UnifiedRegex
                     //     L3:    <item3>
```
Penguinwizzard: Note: you may want to update this comment; it's still referring to 10|20

lib/Parser/RegexRuntime.cpp
```diff
@@ -1180,270 +1187,73 @@ namespace UnifiedRegex
...
-        PRINT_BYTES(SwitchMixin<20>);
-        PRINT_RE_BYTECODE_END();
-    }
-#endif
+#define COMP_STATS
+#define SwitchAndConsumeInstPrintImpl(BaseName, n)
+#endif
+
+#define SwitchAndConsumeInstImpl(BaseName, n) \
+    inline bool BaseName##n##Inst::Exec(REGEX_INST_EXEC_PARAMETERS) const                       \
```
Penguinwizzard: Why preprocessor rather than template?

lib/Parser/RegexRuntime.h
```diff
@@ -671,20 +671,20 @@ namespace UnifiedRegex
 #endif
     };
 
-    template <int n>
+    template <uint8_t n>
     struct SwitchMixin
     {
-        static const int MaxCases = n;
+        static const uint8_t MaxCases = n;
```
Penguinwizzard: nit: should this be constexpr?

lib/Parser/RegexRuntime.cpp
```diff
@@ -934,29 +934,36 @@ namespace UnifiedRegex
...
+            SwitchMixin<n> x;                               \
+            x.AddCase(0, 0);                                \
+            SWITCH_FORCE_INSTANTIATION_PRINT;               \
         }
+
+        SWITCH_FORCE_INSTANTIATION(2);
+        SWITCH_FORCE_INSTANTIATION(4);
+        SWITCH_FORCE_INSTANTIATION(8);
+        SWITCH_FORCE_INSTANTIATION(16);
+        SWITCH_FORCE_INSTANTIATION(24);
```
Penguinwizzard: Why this instead of explicit instantiation? "template SwitchMixin<2>;" outside of a function should force instantiation of the class and its members.

lib/Parser/RegexRuntime.cpp
dilijev: Yep, thanks.

lib/Parser/RegexCompileTime.cpp
dilijev: Fair enough. However, Bad Things (more like wrong behavior) might also happen if we try to serialize as a switch something with more than the max number of cases. Maybe else FailFast?

lib/Parser/RegexCompileTime.cpp
dilijev: Thanks.

lib/Parser/RegexRuntime.cpp
dilijev: Yeah not sure. This is what was done previously, I just templatized it.

lib/Parser/RegexRuntime.cpp
dilijev: Arbitrary... Template would be better, but requires some other shenanigans that have made my other refactorings take longer. I wanted to get eyes on this change soon. I'll see how hard it would be to refactor to templates here.

lib/Parser/RegexRuntime.h
dilijev: Yes. Thanks.

lib/Parser/RegexCompileTime.cpp
dilijev: Also there's logic to prevent entering here if we have more than the max cases required here, so in theory it can't happen. I think `else FailFast` is the right thing.

lib/Parser/RegexRuntime.cpp
dilijev: I'll fix this in a follow-up refactoring. This is fine for now. Want to get this merged ASAP without spending effort fighting with the xplat compilers over templates.

---
In reply to: [147281727](https://github.com/Microsoft/ChakraCore/pull/4071#discussion_r147281727) [](ancestors = 147281727)

lib/Parser/RegexRuntime.cpp
dilijev: Will fix this in a follow-up refactoring.

---
In reply to: [147281404](https://github.com/Microsoft/ChakraCore/pull/4071#discussion_r147281404) [](ancestors = 147281404)

dilijev:
Addressed review comments and rebased. @Penguinwizzard can you take a last look before I merge?

Penguinwizzard:
LGTM, should be shippable as-is.

PR #4072, opened 2017-10-26
Do not skip emitting the nested functions as the assumption is broken in VisiteNestedScopes

aneeshdk:
Do not skip emitting the nested functions as the assumption is broken in VisiteNestedScopes

In VisitNestedScopes while visiting a function we create new FunctionBody for all nested functions inside that. This breaks the assumption made in the PR https://github.com/Microsoft/ChakraCore/pull/3973/files. So reverting back that change. I will revisit this later for a complete fix.

### DIFF
--- a/lib/Runtime/ByteCode/ByteCodeEmitter.cpp
+++ b/lib/Runtime/ByteCode/ByteCodeEmitter.cpp
@@ -3413,16 +3413,13 @@ void ByteCodeGenerator::EmitScopeList(ParseNode *pnode, ParseNode *breakOnBodySc
 
                 PushFuncInfo(_u("StartEmitFunction"), funcInfo);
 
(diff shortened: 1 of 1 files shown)


lib/Runtime/ByteCode/ByteCodeEmitter.cpp
```diff
@@ -3816,12 +3813,9 @@ void ByteCodeGenerator::StartEmitFunction(ParseNode *pnodeFnc)
 
         PushScope(paramScope);
 
-        if (funcInfo->byteCodeFunction->IsFunctionParsed() && funcInfo->GetParsedFunctionBody()->GetByteCode() == nullptr)
```
akroshg: >funcInfo->byteCodeFunction->IsFunctionParsed() && funcInfo->GetParsedFunctionBody()->GetByteCode() == nullptr [](start = 12, length = 109)

just for future note : we should have better state machine enums, to illustrate, deferred, undeferred, redeffered, neverdeffered. something along the line. it will make the debugging easier.

PR #4073, opened 2017-10-26
Addressing WASM.SIMD xplat failure.

arunetm:
Temp fix for WASM.SIMD xplat

### DIFF
--- a/lib/Runtime/Language/InterpreterStackFrame.cpp
+++ b/lib/Runtime/Language/InterpreterStackFrame.cpp
@@ -3803,12 +3803,14 @@ namespace Js
         case AsmJsRetType::Uint32x4:
         case AsmJsRetType::Uint16x8:
         case AsmJsRetType::Uint8x16:
+#if _WIN32 || _WIN64 //WASM.SIMD ToDo: Enable thunk for Xplat
 #if _M_X64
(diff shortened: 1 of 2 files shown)


lib/Runtime/Language/InterpreterStackFrame.cpp
```diff
@@ -3803,13 +3803,15 @@ namespace Js
         case AsmJsRetType::Uint32x4:
         case AsmJsRetType::Uint16x8:
         case AsmJsRetType::Uint8x16:
+#if _WIN32 || _WIN64 //WASM.SIMD ToDo: Enable thunk for Xplat
```
obastemur: Not sure if PAL was defining `_WIN64`.. If all we want is to rule `xplat` out, can we use `#ifdef _WIN32` instead?

arunetm: Yes. This change is incorporated here
https://github.com/Microsoft/ChakraCore/pull/4074/files#diff-c1e1755ec9c1d66c55cceabcbdb92bc2R3806

I didnt know this earlier. Thanks for pointing it out!

arunetm:
https://github.com/Microsoft/ChakraCore/pull/4074 incorporates this fix. Closing this PR

PR #4075, opened 2017-10-26
ARM64: fix default indir scale

sigatrev:


lib/Backend/arm64/ARM64Encoder.h
```diff
@@ -3596,7 +3596,7 @@ EmitLdrRegister64(
     Arm64RegisterParam Index
     )
 {
-    return EmitLdrStrRegisterCommon(Emitter, Dest, Addr, Index, 2, 0xf8600800);
+    return EmitLdrStrRegisterCommon(Emitter, Dest, Addr, Index, 3, 0xf8600800);
```
agarwal-sandeep: >, [](start = 65, length = 1)

Can we replace these 0,1,2,3 etc with a explanatory enum value

sigatrev: Yeah. Updated PR

lib/Backend/arm64/ARM64Encoder.h
```diff
@@ -3500,7 +3500,7 @@ EmitLdrbRegister(
     Arm64RegisterParam Index
     )
 {
-    return EmitLdrStrRegisterCommon(Emitter, Dest, Addr, Index, 0, 0x38600800);
+    return EmitLdrStrRegisterCommon(Emitter, Dest, Addr, Index, IndirScale1, 0x38600800);
 }
```
aaronsgiles: Where is IndirScale1 defined? ARM64Encoder.h is self-contained. I'm leery of introducing dependencies on other chakra components when it isn't necessary.

aaronsgiles: Haha, I see Sandeep suggested you make this change. I'm going to disagree. If you'd like a constant, please defined a new one in ARM64Encoder.h and use it there. Otherwise, I don't think it makes sense to apply this change and create circular dependencies.

agarwal-sandeep: I like #defines instead of numbers. Define it in ARM64Encoder.h and use.

sigatrev: updated to use enums declared in ARM64Encoder.h

